PR #59, opened 2021-06-17
Implement LIMA_HOME to override ~/.lima location

jandubois:
The `limactl start` command now verifies that the combination of LimaDir, instance name, and longest socket name is less than 104 characters, and refuses to create the instance otherwise.

This is intended for applications that want to use lima internally to manage Linux VMs, and don't want to mix those with the VMs managed by the user directly (e.g. [Rancher Desktop](https://github.com/rancher-sandbox/rancher-desktop)).


AkihiroSuda (approved):

AkihiroSuda:
Thanks

PR #60, opened 2021-06-17
Check that files in the guest home directory persist across restarts

jandubois:
Could/should probably be extended to check that containers and images persist across reboots too, but this is good enough for me (for now).


AkihiroSuda (approved):
Thanks

PR #62, opened 2021-06-18
Follow-up to "Implement LIMA_HOME to override ~/.lima location"

AkihiroSuda:
Follow-up to #59 

cc @jandubois

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -82,10 +82,10 @@ func loadOrCreateInstance(clicontext *cli.Context) (*store.Instance, error) {
 		return nil, err
 	}
 
-	// all socket names should be 6 characters or less
-	maxSockName := filepath.Join(instDir, "serial.sock")
-	if len(maxSockName) >= 104 {
-		return nil, errors.Errorf("Instance name %q too long: %q must be less than 104 characers, but is %d",
+	// the full path of the socket name can be at most 104 chars.
+	maxSockName := filepath.Join(instDir, filenames.LongestSock)
+	if len(maxSockName) > 104 {
+		return nil, errors.Errorf("instance name %q too long: %q can be at most 104 characers, but is %d",
 			instName, maxSockName, len(maxSockName))
 	}
 	if _, err := os.Stat(instDir); !errors.Is(err, os.ErrNotExist) {
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -1,6 +1,15 @@
 # Internal data structure
 
-## Instance directory (`~/.lima/<INSTANCE>`)
+## Lima home directory (`${LIMA_HOME}`)
+
+Defaults to `~/.lima`.
+
+Note that we intentionally avoid using `~/Library/Application Support/Lima` on macOS.
+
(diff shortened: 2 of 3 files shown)


docs/internal.md
```diff
@@ -31,9 +40,21 @@ Host agent:
 - `ha.stdout.log`: hostagent stdout (JSON lines, see `pkg/hostagent/api.Events`)
 - `ha.stderr.log`: hostagent stderr (human-readable messages)
 
-## Cache directory (`~/Library/Caches/lima/download/by-url-sha256/<SHA256_OF_URL>`)
+## Lima cache directory (`~/Library/Caches/lima`)
+
+Currently hard-coded to `~/Library/Caches/lima` on macOS.
```
AkihiroSuda, line 45: @jandubois Do you want this directory to be separated out for Rancher Desktop too?

If so, what should be the env var for this?  Can we (ab)use `$XDG_CACHE_HOME`? Or should we define a new env var like `$LIMA_CACHE`?

cmd/limactl/start.go
```diff
@@ -82,10 +82,10 @@ func loadOrCreateInstance(clicontext *cli.Context) (*store.Instance, error) {
...
 	}
 
-	// all socket names should be 6 characters or less
-	maxSockName := filepath.Join(instDir, "serial.sock")
-	if len(maxSockName) >= 104 {
-		return nil, errors.Errorf("Instance name %q too long: %q must be less than 104 characers, but is %d",
+	// the full path of the socket name can be at most 104 chars.
+	maxSockName := filepath.Join(instDir, filenames.LongestSock)
+	if len(maxSockName) > 104 {
+		return nil, errors.Errorf("instance name %q too long: %q can be at most 104 characers, but is %d",
```
jandubois, line 88: I thought 104 characters includes the trailing NUL char, so the actual length must be **less than** 104.

AkihiroSuda: Thanks, fixed

docs/internal.md
jandubois: I don't know yet, I suspect that Rancher Desktop may not be using the lima cache (e.g. it will include the ISO inside the app). But if we do need to use it, it should be separate, so a user running `limactl prune` will not clean out the RD cache.

Personally I would prefer `$LIMA_CACHE` over using `$XDG_CACHE_HOME`, which feels alien on macOS...

AkihiroSuda: Or maybe we can just use `~/Library/Caches/lima/<hash of LIMA_HOME>` so that we do not need to care about multiple env vars.

jandubois: I would prefer to use a separate directory. E.g. if a user runs into a problem with lima and manually removes both `~/.lima` and `~/Library/Caches/lima` to start over with a fresh install, it would not be obvious that this would also remove the RD cache as well.

AkihiroSuda: Opened https://github.com/AkihiroSuda/lima/issues/63 for further discussion

cmd/limactl/start.go
AkihiroSuda: https://github.com/AkihiroSuda/lima/pull/64

PR #68, opened 2021-06-21
cidata: split user-data

### LINKED ISSUE #58: Refactor `user-data.TEMPLATE`
> The `user-data.TEMPLATE` is becoming too long.
> Eventually we should just place the scripts as plain files in the `cidata` drive.
> (And we can execute scripts from the `cidata` drive directly, without copying them into `/var/lib/lima-guestagent`.)
> 
> Can be another PR, though.

_Originally posted by @AkihiroSuda in https://github.com/AkihiroSuda/lima/pull/52#discussion_r652297752_

AkihiroSuda:
Fix #58 

- - -

Upgrading from Lima v0.3.0 probably requires recreating instances.


jandubois:
Thank you, this is awesome! Now I can replace `cloud-init` in the alpine image with a 30 line shell script! 😄 

🚧 https://github.com/rancher-sandbox/alpine-lima/commit/9a51e59eaffd173484ab312a09918ca506b459f9

(it works, but needs some support for logging, in case something goes wrong)

PR #72, opened 2021-06-22
guestagent: run as the root

AkihiroSuda:
The guestagent will soon require root privileges (mostly for virtio stuff), so the guestagent will have to run as the root.


AkihiroSuda:
virtserial turned out to be slower than ssh (https://github.com/AkihiroSuda/lima/issues/20#issuecomment-866796939), so I'm not likely to implement virtserial soon, and probably I don't need to run the guestagent as the root.

So I'm withdrawing this PR, but I can reopen if this is beneficial for Alpine. 

cc @jandubois

jandubois:
> I can reopen if this is beneficial for Alpine.

The only reason it would have been needed was to implement `lima-guestagent install-openrc`, but I think creating the openrc script in the cloud-init data is more straight-forward, so wouldn't want to change it.

PR #73, opened 2021-06-22
Don't try to remove host keys when ~/.ssh/known_hosts doesn't exist

jandubois:
`ssh-keygen -R` will exit with a non-0 status and the hostagent would abort. If the file doesn't exist, then there isn't anything to do.

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -1,7 +1,9 @@
 package sshutil
 
 import (
+	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/sshutil/sshutil.go
```diff
@@ -46,6 +48,25 @@ func DefaultPubKeys() []PubKey {
 	return res
 }
 
+func RemoveHostKeys(sshLocalPort int) error {
```
AkihiroSuda, line 51: Looks like a misnomer.

Can we call this RemoveKnownHostsEntries()?

jandubois: Sure, done.

AkihiroSuda (approved):

PR #80, opened 2021-06-24
allow overriding qemu binary ($QEMU_SYSTEM_X86_64, $QEMU_SYSTEM_AARCH64)

### LINKED ISSUE #14: Customize qemu parameters
Hey! 

I’m asking myself whether there is an config option to customize the used qemu parameters? 
For example to add -nographic and other settings. 

Thx and greets Nico

### LINKED ISSUE #27: An option to specify custom QEMU installation (prefix)
There should be an option to use custom QEMU installation that is located in some directory. It is not always convenient to have system-global QEMU.

This setting needs to be configured per VM (in config).

Currently there's 3 things that are used from QEMU:
- `qemu-system-`
- `qemu-img`
- firmware (UEFI)

AkihiroSuda:
Passing args is also possible (only for debugging purpose), e.g., "QEMU_SYSTEM_X86_64=/opt/qemu/bin/qemu-system-x86_64 -snapshot" .

Fix #27 
Fix #14

### DIFF
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -59,6 +59,12 @@ The directory contains the following files:
 - `$LIMA_INSTANCE`: `lima ...` is expanded to `limactl shell ${LIMA_INSTANCE} ...`.
   - Default : `default`
 
+- `$QEMU_SYSTEM_X86_64`: path of `qemu-system-x86_64`
+  - Default: `qemu-system-x86_64` in `$PATH`
+
+- `$QEMU_SYSTEM_AARCH64`: path of `qemu-system-aarch64`
+  - Default: `qemu-system-aarch64` in `$PATH`
+
 ## `cidata.iso`
 `cidata.iso` contains the following files:
 
--- a/go.mod
+++ b/go.mod
@@ -14,6 +14,7 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/mattn/go-isatty v0.0.13
+	github.com/mattn/go-shellwords v1.0.12
 	github.com/norouter/norouter v0.6.3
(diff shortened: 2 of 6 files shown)


jandubois:
I think it would be good if `qemu.CmdLine` would check for certain options if they are already provided by the override, and then don't add the built-in values for them (e.g. for `-cpu`, `-machine`, `-smp`, `-boot`, `-display` from a cursory scan). Since you are prefixing with the environment variable, there is no way to overwrite the builtin settings otherwise.

Maybe write a custom `appendIfMissing(args, "-option", ...)` that only appends the additional arguments when `args` doesn't yet have an entry equal to `"-option"`?

jandubois:
BTW, providing options to qemu for debugging without recompiling lima is the main point of this change for me.

For providing the location of an alternate qemu binary `$PATH` works just as well as `$QEMU_SYSTEM_X86_64`.

AkihiroSuda:
> e.g. for -cpu, -machine, -smp, -boot, -display from a cursory scan

I'm not sure we want to support overriding these flags, as most of them are configurable via the YAML, though I guess it wouldn't hurt to support overriding `-cpu` and `-machine`.

A workaround is to specify a custom script like this via the env.
```sh
#!/bin/sh
set -eu
exec qemu-system-x86_64 (echo $@ | sed -e 's/-cpu [^ ]* /-cpu SOMETHING /')
```

AkihiroSuda:
added commit `qemu: allow overriding -cpu, -machine, .. as $QEMU_SYSTEM_X86_64 `

jandubois (approved):
LGTM

Also tested with `QEMU_SYSTEM_X86_64="qemu-system-x86_64 -display cocoa" limactl start --tty=false` and worked as expected.

PR #83, opened 2021-06-26
Use user identify from $LIMA_HOME/_config/user

### LINKED ISSUE #78: Make ssh keys configurable
Right now lima adds all `~/.ssh/*.pub` keys as authorized keys to the guest OS and lets ssh pick a key to authenticate.

lima will not work unless the user has at least one key pair, and asks them to run `ssh-keygen` when there isn't one.

For embedding lima in Rancher Desktop, we would want to create a key pair independently of what the user does, and not store it under `~/ssh` because that should only be done by the user themselves.

I'm proposing to authorize a key from `$LIMA_HOME/_config/user.pub` and use `$LIMA_HOME/_config/user.key` as the identity file, if those files exist, and fall back on the current behaviour if they don't.

I've choosen `$LIMA_HOME/_config` because there seemed to be tentative agreement already in #63 on using `$LIMA_HOME/_cache` for the cache.

I'm happy to implement this, but wanted to make sure first that this approach is agreeable.

jandubois:
By default lima uses the public keys from `~/.ssh/*.pub`.

This change makes it try `$LIME_HOME/_config/user.pub` first and only fall back on the `~/.ssh/*.pub` keys when that one does not exist. It will create the key under `$LIMA_HOME/_config` if there aren't any public keys under `~/.ssh`.

Fixes #78

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/AkihiroSuda/lima/pkg/osutil"
+	"github.com/AkihiroSuda/lima/pkg/store"
 	"github.com/AkihiroSuda/lima/pkg/store/filenames"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
@@ -19,34 +20,71 @@ type PubKey struct {
 	Content  string
 }
 
-// DefaultPubKeys finds ssh public keys from ~/.ssh
-func DefaultPubKeys() []PubKey {
+func readPublicKey(f string) (PubKey, error) {
+	entry := PubKey{
+		Filename: f,
+	}
+	content, err := os.ReadFile(f)
+	if err == nil {
+		entry.Content = strings.TrimSpace(string(content))
+	} else {
+		err = errors.Wrapf(err, "failed to read ssh public key %q", f)
+	}
+	return entry, err
+}
+
+// DefaultPubKeys returns the public key from $LIMA_HOME/_config/user.pub.
+// The key will be created if it does not yet exist. All public keys
+// ~/.ssh/*.pub will be appended to make the VM accessible without specifying
+// and identity explicitly.
+func DefaultPubKeys() ([]PubKey, error) {
+	// Read $LIMA_HOME/_config/user.pub
+	configDir, err := store.LimaConfigDir()
+	if err != nil {
+		return nil, err
+	}
+	_, err = os.Stat(filepath.Join(configDir, filenames.UserPrivateKey))
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return nil, err
+		}
+		if err := os.MkdirAll(configDir, 0700); err != nil {
+			return nil, errors.Wrapf(err, "could not create %q directory", configDir)
+		}
+		keygenCmd := exec.Command("ssh-keygen", "-t", "ed25519", "-q", "-N", "", "-f",
+			filepath.Join(configDir, filenames.UserPrivateKey))
+		logrus.Debugf("executing %v", keygenCmd.Args)
+		if out, err := keygenCmd.CombinedOutput(); err != nil {
+			return nil, errors.Wrapf(err, "failed to run %v: %q", keygenCmd.Args, string(out))
+		}
+	}
+	entry, err := readPublicKey(filepath.Join(configDir, filenames.UserPublicKey))
+	if err != nil {
+		return nil, err
+	}
+	res := []PubKey{entry}
+
+	// Append all of ~/.ssh/*.pub
 	homeDir, err := os.UserHomeDir()
 	if err != nil {
-		logrus.Warn(err)
-		return nil
+		return nil, err
 	}
 	files, err := filepath.Glob(filepath.Join(homeDir, ".ssh/*.pub"))
 	if err != nil {
-		logrus.Warn(err)
-		return nil
+		panic(err) // Only possible error is ErrBadPattern, so this should be unreachable.
 	}
-	var res []PubKey
 	for _, f := range files {
 		if !strings.HasSuffix(f, ".pub") {
 			panic(errors.Errorf("unexpected ssh public key filename %q", f))
 		}
-		entry := PubKey{
-			Filename: f,
+		entry, err := readPublicKey(f)
+		if err == nil {
+			res = append(res, entry)
+		} else if !errors.Is(err, os.ErrNotExist) {
+			return nil, err
 		}
-		if content, err := os.ReadFile(f); err == nil {
-			entry.Content = strings.TrimSpace(string(content))
-		} else {
-			logrus.WithError(err).Warningf("failed to read ssh public key %q", f)
-		}
-		res = append(res, entry)
 	}
-	return res
+	return res, nil
 }
 
 func RemoveKnownHostEntries(sshLocalPort int) error {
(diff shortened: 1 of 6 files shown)


pkg/sshutil/sshutil.go
```diff
@@ -18,34 +19,74 @@ type PubKey struct {
...
 
-// DefaultPubKeys finds ssh public keys from ~/.ssh
+func readPublicKey(f string) (PubKey, error) {
+	entry := PubKey{
+		Filename: f,
+	}
+	content, err := os.ReadFile(f)
+	if err == nil {
+		entry.Content = strings.TrimSpace(string(content))
+	} else {
```
AkihiroSuda, line 29: Return err if !error.Is(err, os.ErrNorExist)

pkg/sshutil/sshutil.go
```diff
@@ -18,34 +19,74 @@ type PubKey struct {
...
-	homeDir, err := os.UserHomeDir()
+	// Read $LIMA_HOME/_config/user.pub
+	configDir, err := store.LimaConfigDir()
+	privateKeyPath := filepath.Join(configDir, "user")
 	if err != nil {
 		logrus.Warn(err)
-		return nil
+	} else {
+		// Check that private key exists and then try to read public key
+		if _, err := os.Stat(privateKeyPath); err == nil {
```
AkihiroSuda, line 47: Return err if !error.Is(err, os.ErrNorExist)

pkg/sshutil/sshutil.go
```diff
@@ -83,5 +124,12 @@ func SSHArgs(instDir string) ([]string, error) {
 		"-o", "Compression=no",
 		"-o", "BatchMode=yes",
 	}
+	configDir, err := store.LimaConfigDir()
+	if err == nil {
```
AkihiroSuda, line 128: Return err if !error.Is(err, os.ErrNorExist)

pkg/sshutil/sshutil.go
```diff
@@ -83,5 +124,12 @@ func SSHArgs(instDir string) ([]string, error) {
...
+		privateKeyPath := filepath.Join(configDir, "user")
+		if _, err := os.Stat(privateKeyPath); err == nil {
```
AkihiroSuda, line 130: Return err if !error.Is(err, os.ErrNorExist)

pkg/sshutil/sshutil.go
```diff
@@ -18,34 +19,74 @@ type PubKey struct {
...
+			}
 		}
-		res = append(res, entry)
 	}
-	return res
+	// Create $LIMA_HOME/_config/user (and $LIMA_HOME/_config/user.pub)
+	if configDir == "" {
+		return nil
+	}
+	keygenCmd := exec.Command("ssh-keygen", "-q", "-N", "", "-f", privateKeyPath)
```
AkihiroSuda, line 81: Can we specify algo explicitly?

pkg/sshutil/sshutil.go
```diff
@@ -18,34 +19,74 @@ type PubKey struct {
...
+	if out, err := keygenCmd.CombinedOutput(); err != nil {
```
AkihiroSuda, line 82: logrus.Debugf(“executing %v”, keygenCmd.Args)

pkg/sshutil/sshutil.go
AkihiroSuda: Perhaps ed25519 is fine

pkg/store/store.go
```diff
@@ -29,6 +29,20 @@ func LimaDir() (string, error) {
 	return dir, nil
 }
 
+// LimaConfigDir returns the path of the config directory, $LIMA_HOME/_config.
+// It will create the directory if it doesn't yet exist.
```
AkihiroSuda, line 33: I’m not sure this function should make the dir.

pkg/store/store.go
```diff
@@ -29,6 +29,20 @@ func LimaDir() (string, error) {
...
+func LimaConfigDir() (string, error) {
+	limaDir, err := LimaDir()
+	if err != nil {
+		return "", err
+	}
+	configDir := filepath.Join(limaDir, "_config")
```
AkihiroSuda, line 39: “_config” should be a constant.

pkg/sshutil/sshutil.go
jandubois: Why? If `ReadFile` returns an error, `content` cannot be used, and we should return the error, regardless of what it is.

It will just be written to the log, and then ignored because `DefaultPubKeys` doesn't return an error. Maybe it should?

pkg/sshutil/sshutil.go
jandubois: `DefaultPubKeys` doesn't return any errors; it just ignores key files it cannot read. So not sure what to do here.

In the end the user would be told to run `ssh-keygen` when `DefaultPubKeys` returns an empty list, so not sure that aborting on the first error is really helpful.

pkg/store/store.go
jandubois: Yeah, it is inconsistent with what `LimaDir` does, but I was thinking about letting that one create the directory too. Would that simplify the calling code?

pkg/store/store.go
jandubois: Yes, I was tempted to put it into `store/filenames`, but that package claims it only defines names for the instance directories. So where should `_config` be defined? New package?

AkihiroSuda: pkg/store.ConfigDir

jandubois:
@AkihiroSuda I believe I've addresses all issues you pointed out. Can you please confirm?

But I'm wondering now if the `user` filename shouldn't be a constant too, but again don't know where to put it.

Should `store/filenames` be extended to cover both config and instance filenames?

pkg/sshutil/sshutil.go
jandubois: This is now handled by the caller; any error is reported back to the caller because we are only calling `readPublicKey` on files that are supposed to exist.

jandubois:
Sorry, I refactored `DefaultPubKeys` once more, so that it can actually reflect an error back to the caller.

By tightening the error handling I've also reduced the logic nesting level a bit, as it was getting out of hand.

I've also gone ahead and added other file and directory names to the `filenames` package. Let me know, if this is not how you want it to be!

I believe I'm done with it now (unless you have more feedback).

jandubois:
@AkihiroSuda The latest CI failure is different, but I've seen a lot of errors like this before:

```
time="2021-06-28T21:49:58Z" level=fatal msg="failed to resolve reference \"ghcr.io/stargz-containers/nginx:1.19-alpine-org\": failed to do request: Head \"https://ghcr.io/v2/stargz-containers/nginx/manifests/1.19-alpine-org\": net/http: TLS handshake timeout"
```

Any idea why that is happening? You would think the connection between gha and ghcr should be pretty solid...

AkihiroSuda:
I guess macOS instances might be running on a different infra that isn't optimal for ghcr?

pkg/sshutil/sshutil.go
```diff
@@ -19,34 +20,71 @@ type PubKey struct {
...
+	}
+	_, err = os.Stat(filepath.Join(configDir, filenames.UserPrivateKey))
+	if err != nil {
+		if !errors.Is(err, os.ErrNotExist) {
+			return nil, err
+		}
+		if err := os.MkdirAll(configDir, 0700); err != nil {
+			return nil, errors.Wrapf(err, "could not create %q directory", configDir)
+		}
+		keygenCmd := exec.Command("ssh-keygen", "-t", "ed25519", "-q", "-N", "", "-f",
```
AkihiroSuda, line 54: We should lock the directory. I can open a follow-up PR.

AkihiroSuda (approved):

pkg/sshutil/sshutil.go
```diff
@@ -77,7 +115,17 @@ func SSHArgs(instDir string) ([]string, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	privateKeyPath := filepath.Join(configDir, filenames.UserPrivateKey)
+	_, err = os.Stat(privateKeyPath)
+	if err != nil {
+		return nil, err
+	}
 	args := []string{
+		"-i", privateKeyPath,
```
AkihiroSuda, line 128: Question: would it be possible to let ssh read both `~/.lima/_config` keys and `~/.ssh` keys?

Otherwise this will break existing instances.

jandubois: > Question: would it be possible to let ssh read both `~/.lima/_config` keys and `~/.ssh` keys?

Yes, you can specify multiple `-i` options, and the identities will be tried in sequence. I think the server may be limiting the number of attempts to 3, so it could still fail if the user added multiple keys after creating the instance. But that is already a problem with the current implementation, so doesn't change anything. Just another reason to have a lima-specific identity, and always specify that one first.

jandubois: > Question: would it be possible to let ssh read both `~/.lima/_config` keys and `~/.ssh` keys?

I've pushed the change in a separate commit for easier review; lmk if you want me to squash it!

AkihiroSuda (approved):
Thanks

PR #87, opened 2021-06-28
sshutil: specify username explicitly

### LINKED ISSUE #85: ssh should always specify the user name
Hypothetical situation: I use a company laptop, and device management software forces me to use some LDAP identity as the local user name, say `JDubois`. But for all my VMs and lab machines I use `jan` as the user name. Therefore I create an `~/.ssh/config` with explicit entries for a couple of official servers using my LDAP identify, and a fallback for my lab:

```
Host *
     User jan
     ForwardAgent yes
     UseKeychain yes
     AddKeysToAgent yes
     IdentityFile ~/.ssh/jan
     IdentitiesOnly yes
```

This currently breaks lima because it would create a `JDubois` user in the guest VM, but tries to connect with a `jan` user because of the config definition.

Always specifying the user in ssh commands avoids this potential error situation.

AkihiroSuda:
Fix #85

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"os/user"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


jandubois (approved):
LGTM

Also verified that it fixes the issue from #85

PR #88, opened 2021-06-28
examples: add archlinux.yaml

AkihiroSuda:


examples/archlinux.yaml
```diff
@@ -0,0 +1,15 @@
+arch: "x86_64"
+images:
+  - location: "https://linuximages.de/openstack/arch/arch-openstack-LATEST-image-bootstrap.qcow2"
```
jandubois, line 3: Maybe add some comment here that this URL points to a symlink on the server. So Lima will cache this the first time it downloads the image, and will never refresh it unless the cache is manually pruned.

I noticed that e.g. the Debian sample points to snapshot from a specific day and not to a "latest" URL like https://cloud.debian.org/images/cloud/bullseye/daily/latest/debian-11-generic-amd64-daily.qcow2

So maybe Debian should be changed to, but I still wonder if there should be some mechanism to do some kind of cache validation? At least I think the user should see a message that they are using an outdated local cache version, and that a newer release is available for download.

AkihiroSuda: > I noticed that e.g. the Debian sample points to snapshot from a specific day

I wanted to do the same for archlinux, but https://linuximages.de/openstack/arch seems to retain old images only for a week 😞 

> So maybe Debian should be changed to, but I still wonder if there should be some mechanism to do some kind of cache validation? At least I think the user should see a message that they are using an outdated local cache version, and that a newer release is available for download.

I think we can use [ETag](https://en.wikipedia.org/wiki/HTTP_ETag).
(Side topic: we should also support validating SHA256)

jandubois: > (Side topic: we should also support validating SHA256)

Yes, I've been wondering if https://github.com/hashicorp/go-getter would be a good match for lima, but haven't really looked into it further. That is the library minikube is using for downloading files and validating their checksums.

PR #92, opened 2021-06-29
Add copy/cp/scp command

jandubois:
This is a wrapper around `scp`. User doesn't have to lookup / specify the port number, and doesn't have to prefix remote names with `127.0.0.1:`; a single `:` is enough. And the connection goes through the ControlPath.

Example:

```
limactl cp default :/etc/os-release .
```

### DIFF
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"os/user"
+	"strings"
+
+	"github.com/AkihiroSuda/lima/pkg/sshutil"
+	"github.com/AkihiroSuda/lima/pkg/store"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+)
+
+var copyCommand = &cli.Command{
+	Name:        "copy",
+	Aliases:     []string{"cp"},
+	Usage:       "Copy files between host and guest",
+	Description: "Prefix guest filenames with the instance name and a colon.\nExample: limactl copy default:/etc/os-release .",
+	ArgsUsage:   "SOURCE ... TARGET",
+	Action:      copyAction,
+}
+
+func copyAction(clicontext *cli.Context) error {
+	if clicontext.NArg() < 2 {
+		return errors.Errorf("requires at least 2 arguments: SOURCE DEST")
+	}
+	arg0, err := exec.LookPath("scp")
+	if err != nil {
+		return err
+	}
+	u, err := user.Current()
+	if err != nil {
+		return err
+	}
+
+	args, err := sshutil.CommonArgs()
+	if err != nil {
+		return err
+	}
+	args = append(args, "-3", "--")
+	for _, arg := range clicontext.Args().Slice() {
+		path := strings.Split(arg, ":")
+		switch len(path) {
+		case 1:
+			args = append(args, arg)
+		case 2:
+			instName := path[0]
+			inst, err := store.Inspect(instName)
+			if err != nil {
+				if errors.Is(err, os.ErrNotExist) {
+					return errors.Errorf("instance %q does not exist, run `limactl start %s` to create a new instance", instName, instName)
+				}
+				return err
+			}
+			if inst.Status == store.StatusStopped {
+				return errors.Errorf("instance %q is stopped, run `limactl start %s` to start the instance", instName, instName)
+			}
+			args = append(args, fmt.Sprintf("scp://%s@127.0.0.1:%d/%s", u.Username, inst.SSHLocalPort, path[1]))
+		default:
+			return errors.Errorf("Path %q contains multiple colons", arg)
+		}
+	}
+	cmd := exec.Command(arg0, args...)
+	cmd.Stdin = os.Stdin
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	logrus.Debugf("executing scp (may take a long)): %+v", cmd.Args)
+
+	// TODO: use syscall.Exec directly (results in losing tty?)
+	return cmd.Run()
+}
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -106,15 +106,7 @@ func RemoveKnownHostEntries(sshLocalPort int) error {
 	return nil
 }
 
-func SSHArgs(instDir string) ([]string, error) {
-	u, err := user.Current()
-	if err != nil {
-		return nil, err
-	}
-	controlSock := filepath.Join(instDir, filenames.SSHSock)
-	if len(controlSock) >= osutil.UnixPathMax {
-		return nil, errors.Errorf("socket path %q is too long: >= UNIX_PATH_MAX=%d", controlSock, osutil.UnixPathMax)
-	}
+func CommonArgs() ([]string, error) {
 	configDir, err := store.LimaConfigDir()
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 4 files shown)


cmd/limactl/copy.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/AkihiroSuda/lima/pkg/sshutil"
+	"github.com/AkihiroSuda/lima/pkg/store"
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+	"github.com/urfave/cli/v2"
+)
+
+var copyCommand = &cli.Command{
+	Name:        "copy",
+	Aliases:     []string{"cp", "scp"},
```
AkihiroSuda, line 20: I’m not sure we need “scp” alias

AkihiroSuda:
Thanks, but I heard scp is deprecated https://lists.mindrot.org/pipermail/openssh-unix-dev/2019-March/037672.html

Can we use sftp or rsync?

cmd/limactl/copy.go
```diff
@@ -0,0 +1,75 @@
...
+	Usage:       "Copy files between host and guest",
+	Description: "Prefix guest filenames with a colon. Example: limactl copy default :/etc/os-release .",
```
AkihiroSuda, line 22: Do we really need colon?

cmd/limactl/copy.go
jandubois: Sure, I'll remove it.

cmd/limactl/copy.go
jandubois: We need some mechanism to tell which file is on the host, and which is on the guest. If we don't use the colon, how would you specify the direction:

```sh
# copy from guest to host
limactl cp default :/etc/hostname hostname
# copy from host to guest
limactl cp default file1 file2 file3 :.
```

jandubois:
> Thanks, but I heard scp is deprecated https://lists.mindrot.org/pipermail/openssh-unix-dev/2019-March/037672.html

It is my understanding that the scp *protocol* is deprecated, not the cli tool: https://github.com/openssh/openssh-portable/pull/194

> Can we use sftp or rsync?

I prefer to use `scp` because it is part of OpenSSH, so we don't need to require additional software to be installed on the guest.

cmd/limactl/copy.go
AkihiroSuda: An alternative syntax: `limactl cp default:/etc/hostname hostname`.

This syntax can be also extended to `limactl cp inst1:/foo inst2:/foo` to copy files across instances.

This can be also extended to `limactl cp inst1 inst2` to copy instances, though it might be confusing.
But given that `limactl ls` lists instances and `limactl delete` deletes instances, `limactl cp inst1 inst2` might be probably acceptable.

jandubois: > This syntax can be also extended to `limactl cp inst1:/foo inst2:/foo` to copy files across instances.

I like it. I just did a manual test, and it seems to work:

```console
$ scp -i ~/.lima/_config/user -o NoHostAuthenticationForLocalhost=yes -3 scp://jan@127.0.0.1:60020//etc/hostname scp://jan@127.0.0.1:60022/myname
$ limactl shell default sh -c 'cat ~/myname'
lima-alpine
```

I was confused for a bit about the obscure syntax: the first slash after the port is just a separator and not part of the path. You need a double-slash to specify an absolute path. At least it means relative paths are still supported.

The copy still has to go through the host (using `-3`) and not directly between the instances because they use different ports.

I guess theoretically you could use `ssh` to run the `scp` command from within the source instances, but that seems like unnecessary complexity.

jandubois: I don't think it will be possible to use the `ControlPath` when we copy between instances, so this will complicate the logic some more...

I'll take a look into it tomorrow.

AkihiroSuda: Thanks, `limactl cp inst1:/foo inst2:/foo` can be implemented later, we just have to determine the CLI syntax here.

jandubois: I just implemented it:

```console
$ limactl cp alpine:/etc/os-release opensuse:release-info
$ limactl shell opensuse sh -c 'cat ~/release-info'
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.13.5
PRETTY_NAME="Alpine Linux v3.13"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://bugs.alpinelinux.org/"
```

scp will not use the `ControlPath`, but I think that is fine; it is just an optimization.

pkg/sshutil/sshutil.go
```diff
@@ -162,3 +150,25 @@ func SSHArgs(instDir string) ([]string, error) {
...
+func SSHArgs(instDir string) ([]string, error) {
+	controlSock := filepath.Join(instDir, filenames.SSHSock)
+	if len(controlSock) >= osutil.UnixPathMax {
+		return nil, errors.Errorf("socket path %q is too long: >= UNIX_PATH_MAX=%d", controlSock, osutil.UnixPathMax)
+	}
+	u, err := user.Current()
+	if err != nil {
+		return nil, err
+	}
+	args, err := SCPArgs()
```
AkihiroSuda, line 163: This is confusing, can we just use a single function like `func SSHArgs(instDir string, useControlMaster bool)` ?

cmd/limactl/copy.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		return err
+	}
+
+	args, err := sshutil.SCPArgs()
+	if err != nil {
+		return err
+	}
+	args = append(args, "-3", "--")
+	for _, arg := range clicontext.Args().Slice() {
```
AkihiroSuda, line 44: I think we will eventually want unit tests for the parser, but can be added in separate PR in future

AkihiroSuda, line 44: We may want to limit the length of the slice to be 2?

cmd/limactl/copy.go
```diff
@@ -0,0 +1,74 @@
...
+		path := strings.Split(arg, ":")
+		switch len(path) {
+		case 1:
```
AkihiroSuda, line 47: Can we require `/` to be included in a host file path?

Because we may want `limactl cp inst1 inst2` to duplicate an instance in future.

pkg/sshutil/sshutil.go
jandubois: > This is confusing,

Not sure why this is confusing: the SCP arguments are just a subset of the SSH arguments. There are also potential differences (not relevant here), like `-p` for  SSH is `-P` for SCP because `-p` for SCP means "preserve access mode and file times"...

> can we just use a single function like func SSHArgs(instDir string, useControlMaster bool) ?

So you would call it with an empty string as the `instDir` from the `copyCommand`? Because we no longer have a global `instDir`, it can be different for each source argument.

Please reconsider! I think the current implementation is less confusing than forcing both commands to use the same function.

AkihiroSuda: I was confused because "SCPArgs" is called even when we do not use scp.

Just renaming `SCPArgs()` to `CommonArgs()` and let `var SCPArgs = CommonArgs` might be less confusing, but anyway I don't have a strong opinion.

cmd/limactl/copy.go
jandubois: Why? You can specify multiple source arguments when the destination is a directory:

```console
$ limactl cp alpine:/etc/os-release default:/etc/hostname opensuse:.
jan@mactop:~/suse/lima [readme %]
$ limactl shell opensuse sh -c 'ls -l ~'
total 12
drwxr-xr-x 2 jan users   6 Mar  3 11:52 bin
-rw-r--r-- 1 jan users  13 Jun 29 07:47 hostname
-rw-r--r-- 1 jan users 164 Jun 29 07:47 os-release
```

AkihiroSuda: Ah, that's awesome :+1:

cmd/limactl/copy.go
jandubois: > Can we require `/` to be included in a host file path?

We could, but it feels slightly wrong to do so. Wouldn't it make more sense to say if none of the arguments contain a colon (i.e. all arguments are host references), then the arguments must be 2 instance names, of which the first one must already exist, and the second one must not?

Because `limactl cp inst1 inst2` would otherwise just be a local copy between 2 files, which doesn't need `limactl cp`, you can use just plain `cp` for that. 😄

AkihiroSuda: Ok, I guess we are unlikely going to overload `limactl cp` for copying instances.
For copying instances we can just have `limactl clone inst1 inst2` or maybe `limactl copy-instance inst1 inst2`.

If a user attempted to run `limactl cp inst1 inst2`, we can just print a warning.

AkihiroSuda (approved):

cmd/limactl/copy.go
jandubois: > Ok, I guess we are unlikely going to overload `limactl cp` for copying instances.
> For copying instances we can just have `limactl clone inst1 inst2` or maybe `limactl copy-instance inst1 inst2`.

Good. I don't really like overloading commands with unrelated functionality, but couldn't think of an "obvious" name for copying an instance. I think `clone` is perfect for this.

> If a user attempted to run `limactl cp inst1 inst2`, we can just print a warning.

Not sure if we need to warn; `limectl cp foo bar` really does the same thing as `cp foo bar` right now, and makes sense to me: if source and dest can be both remote (guest), then why shouldn't they both be allowed to be local (host) as well? And if the source file does not exist, we already get a warning:

```console
$ limactl cp foo bar
cp: foo: No such file or directory
exit status 1
```

pkg/sshutil/sshutil.go
jandubois: I've renamed `SCPArgs` to `CommonArgs` now.

PR #98, opened 2021-06-29
README updates: add openSUSE, remove outdated FAQs

jandubois:
I've removed the ssh related FAQs that should no longer be relevant because the code deals with them correctly now. Not sure if you may want to keep them in the FAQ for a while though, for users of earlier lima versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Lima is expected to be used on macOS hosts, but can be used on Linux hosts as we
 
 ✅ Intel on ARM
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -210,13 +212,13 @@ The current default spec:
 Password is disabled and locked by default.
```
AkihiroSuda, line 212: Could you run [`doctoc`](https://github.com/thlorenz/doctoc) to update the TOC section few lines above?

AkihiroSuda (approved):
Thanks

PR #105, opened 2021-07-02
ssh: allow `ssh.loadDotSSHPubKeys = false`

AkihiroSuda:
`ssh.loadDotSSHPubKeys` (boolean) specifies whether to load `~/.ssh/*.pub` in addition to `$LIMA_HOME/_config/user`.

The default value remains true.

 - - -
I'll release v0.5.0 after merging this and https://github.com/AkihiroSuda/lima/pull/106

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -35,10 +35,11 @@ func readPublicKey(f string) (PubKey, error) {
 }
 
 // DefaultPubKeys returns the public key from $LIMA_HOME/_config/user.pub.
-// The key will be created if it does not yet exist. All public keys
-// ~/.ssh/*.pub will be appended to make the VM accessible without specifying
-// and identity explicitly.
-func DefaultPubKeys() ([]PubKey, error) {
+// The key will be created if it does not yet exist.
+//
+// When loadDotSSH is true, ~/.ssh/*.pub will be appended to make the VM accessible without specifying
+// an identity explicitly.
+func DefaultPubKeys(loadDotSSH bool) ([]PubKey, error) {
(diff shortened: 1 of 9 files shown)


pkg/sshutil/sshutil.go
```diff
@@ -35,10 +35,11 @@ func readPublicKey(f string) (PubKey, error) {
 }
 
 // DefaultPubKeys returns the public key from $LIMA_HOME/_config/user.pub.
-// The key will be created if it does not yet exist. All public keys
-// ~/.ssh/*.pub will be appended to make the VM accessible without specifying
+// The key will be created if it does not yet exist.
+//
+// When loadDotSSH is true, ~/.ssh/*.pub will be appended to make the VM accessible without specifying
 // and identity explicitly.
```
jandubois, line 41: `s/and/an/`

jandubois (approved):
LGTM and worked as expected when I tested it.

As I wrote before, I'm not sure why this option is needed. There is at least a small cost to this, as new users will read the config file, and they need to understand what the various options mean to them.

So I think there should be a use case for why the option is necessary beyond just some abstract cleanliness. It is not this option by itself, but more about having a policy to keep complexity to a minimum, especially the part that is exposed to the user.

But I'm fine with adding it if you feel that there are situations where it will prove beneficial/necessary.

pkg/sshutil/sshutil.go
AkihiroSuda: fixed

AkihiroSuda:
`ssh.loadDotSSHPubKeys = false` would be recommended when the user have an insecure key under `~/.ssh` or facing some random SSH error.

PR #106, opened 2021-07-02
Support digest validation

AkihiroSuda:
Also removes dependency on `curl` binary

### DIFF
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -63,6 +63,8 @@ The directory contains the following files:
 
 - `url`: raw url text, without "\n"
 - `data`: data
+- `<ALGO>.digest`: digest of the data, in OCI format.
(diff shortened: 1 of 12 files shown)


jandubois (commented):
Looks fine to me, but I'm in a bit of a hurry right now and didn't look too closely. Will be mostly offline the next 2 days.

One thought (for a later PR) would be to support downloadable checksums. They obviously don't protect against a website spoofing attack, but still help with truncated downloads due to networking errors.

E.g. checking https://cloud-images.ubuntu.com/hirsute/current/SHA256SUMS, but maybe there are too many different formats to support for that?

AkihiroSuda:
Yes, I was thinking of that.
Not too many format, but probably in another PR in future

AkihiroSuda:
Merging and releasing v0.5.0.  Support for`SHA256SUMS` file can be added later.

PR #108, opened 2021-07-05
examples/archlinux.yaml: use more reliable image URL

AkihiroSuda:


AkihiroSuda:
```
+ modprobe tap
modprobe: FATAL: Module tap not found in directory /lib/modules/5.12.11-arch1-1
+ true
+ modprobe ip_tables
modprobe: FATAL: Module ip_tables not found in directory /lib/modules/5.12.11-arch1-1
```

CI is failing probably because the kernel is updated during running cloud-init 😞

PR #110, opened 2021-07-06
Always disregard ~/.ssh/config; only use specified identity file

jandubois:
This change is from debugging a setup issue with lima 0.4.0 by @mook-as.

It was no longer required with 0.5.0, but it shouldn't hurt, and might be useful to avoid other conflicts.


AkihiroSuda (approved):
thanks

AkihiroSuda:
(Maybe we should have a documentation about that ~/.ssh/config is discarded)

PR #111, opened 2021-07-06
Don't useDotSSHPubKeys for limactl copy command

jandubois:
It fails if there is a public key without a corresponding private key in `~/.ssh`. `useDotSSHPubKeys` should not be needed for the copy command except when using it on legacy instances. Document how to make those instances compatible.


AkihiroSuda (approved):
Thanks

Issue #45, opened 2021-06-11
Support port forwarding for privileged ports (1-1023)


AkihiroSuda:
https://twitter.com/_AkihiroSuda_/status/1403403845842075648

I'm planning to support privileged ports using authbind (https://github.com/Castaglia/MacOSX-authbind) if it works.

AkihiroSuda:
Another way is to replace `ssh -L` with https://github.com/norouter/norouter and set SETUID bit on the `norouter` binary.
(We could just set SETUID on `ssh` binary, but perhaps it may cause some unexpected side effect)

AkihiroSuda:
Or maybe we don't need any SETUID stuff. Just bind 0.0.0.0:80, and reject connections from non-loopback addresses.

jandubois:
> reject connections from non-loopback addresses

Maybe make the rejection configurable? I think it can be useful to expose an app running in lima to the local network, so you can test it e.g. as a service from another machine.

AkihiroSuda:
Yes, we will eventually need config file and (REST?) API to control flexible forwarding

markomitranic:
Until
this gets done, I hope this old little snippet helps someone :)

```
alias port80="sudo ifconfig lo0 127.0.0.1 alias
echo \"rdr pass on lo0 inet proto tcp from any to 127.0.0.1 port 80 -> 127.0.0.1 port 8080\" | sudo pfctl -ef -"
```

To revert the mappings back to their original state, just empty the file.

sudo pfctl -f /etc/pf.conf

Same for 443

AkihiroSuda:
PR: https://github.com/lima-vm/lima/pull/283

PR #114, opened 2021-07-09
Allow forwarding of specified port ranges to all interfaces

jandubois:
This allows to expose e.g. an ingress port outside of the local host while still forwarding other ports only to localhost.

Example:

```yaml
allInterfaces:
- port: 443
- port: 4000
  until: 4999
```

I'm not quite sure how this will interact with #45, but I think/hope it will be independent (the filtering of incoming connections would be done via `iptables` inside the guest?).

I'm also not really happy about the `allInterfaces` key name, but couldn't come up with anything better.

### DIFF
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -73,6 +73,24 @@ containerd:
   # Default: true
   user: true
 
+# port forwarding rules.
+# By default guest ports are forwarded to the same port on the 127.0.0.1 interface on the host.
+# ports:
+
+#   - guestPort: 443
+#     hostIP: "0.0.0.0" # overrides the default value "127.0.0.1"
+#   # default: hostPort: 443
+#   # default: guestIP: "127.0.0.1" (only valid value right now)
+#   # default: proto: "tcp" (only valid value right now)
+
+#   - guestPortRange: [4000, 4999]
+#     hostIP:  "0.0.0.0" # overrides the default value "127.0.0.1"
+#   # default: hostPortRange: [4000, 4999] (must specify same number of ports as guestPortRange)
+
+#   - guestPort: 80
+#     hostPort: 8080 # overrides the default value 80
+
+
 # Provisioning scripts need to be idempotent because they might be called
 # multiple times, e.g. when the host VM is being restarted.
 # provision:
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -90,6 +90,8 @@ containerd:
 #   - guestPort: 80
 #     hostPort: 8080 # overrides the default value 80
 
+#   - guestPort: 8888
+#     ignore: true (don't forward this port)
 
 # Provisioning scripts need to be idempotent because they might be called
 # multiple times, e.g. when the host VM is being restarted.
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -73,25 +73,31 @@ containerd:
   # Default: true
   user: true
 
-# port forwarding rules.
-# By default guest ports are forwarded to the same port on the 127.0.0.1 interface on the host.
-# ports:
-
+# Port forwarding rules. Forwarding between ports 22 and ssh.localPort cannot be overridden.
+# Rules are checked sequentially until the first one matches.
+# portForwards:
 #   - guestPort: 443
-#     hostIP: "0.0.0.0" # overrides the default value "127.0.0.1"
-#   # default: hostPort: 443
-#   # default: guestIP: "127.0.0.1" (only valid value right now)
+#     hostIP: "0.0.0.0" # overrides the default value "127.0.0.1"; allows privileged port forwarding
+#   # default: hostPort: 443 (same as guestPort)
+#   # default: guestIP: "127.0.0.1" (also matches bind addresses "0.0.0.0", "::", and "::1")
 #   # default: proto: "tcp" (only valid value right now)
-
 #   - guestPortRange: [4000, 4999]
 #     hostIP:  "0.0.0.0" # overrides the default value "127.0.0.1"
 #   # default: hostPortRange: [4000, 4999] (must specify same number of ports as guestPortRange)
-
 #   - guestPort: 80
 #     hostPort: 8080 # overrides the default value 80
-
+#   - guestIP: "127.0.0.2" # overrides the default value "127.0.0.1"
+#     hostIP: "127.0.0.2" # overrides the default value "127.0.0.1"
+#   # default: guestPortRange: [1024, 65535]
+#   # default: hostPortRange: [1024, 65535]
 #   - guestPort: 8888
 #     ignore: true (don't forward this port)
+#   # Lima internally appends this fallback rule at the end:
+#   - guestIP: "127.0.0.1"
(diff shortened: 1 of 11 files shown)


AkihiroSuda:
> I'm also not really happy about the allInterfaces key name, but couldn't come up with anything better.

What about `ports` (or `portForwards`) ?

```yaml
ports:
- guestPort: 443
  hostIP: "0.0.0.0" # overrides the default value "127.0.0.1"
# default: hostPort: 443
# default: guestIP: "127.0.0.1" (implementation can be done later, but we have to reserve the field name now)
# default: proto: "tcp" (same as above)

- guestPortRange: [4000, 4999]
  hostIP:  "0.0.0.0" # overrides the default value "127.0.0.1"
# default: hostPortRange: [4000, 4999]
# default: guestIP: "127.0.0.1"
# default: proto: "tcp"

- guestPort: 80
  hostPort: 8080 # overrides the default value 80
# default: hostIP: "127.0.0.1"
# default: guestIP: "127.0.0.1"
# default: proto: "tcp"
```

> I'm not quite sure how this will interact with #45, but I think/hope it will be independent (the filtering of incoming connections would be done via iptables inside the guest?).

No, iptables inside the guest cannot be used, as the guest is unaware of the src IP.
So the filtering will be implemented in a usermode proxy process on the host.

I assume #45 will not need a new YAML field. "Try `ssh -L` first and then fall back to our own forwarder` would be fine. We can also consider completely removing `ssh -L`, depending on benchmarking result.

jandubois:
> What about `ports` (or `portForwards`) ?

Yes, much better and more flexible. I've updated the PR to match your samples.

jandubois:
Sorry, did one more push to fix a trivial error in an error message:

```diff
        case port > 65535:
-               return errors.Errorf("field `%s` must be < 65535", field)
+               return errors.Errorf("field `%s` must be < 65536", field)
```

But now I'm really done...

jandubois:
@AkihiroSuda I've been wondering how `GuestIP` is supposed to be used. Is given that we only forward ports from `0.0.0.0` or `127.0.0.1`, is there a reason to every use anything else but `127.0.0.1` as the guest IP?

Even if the port is bound to `::1` it is still reachable via `127.0.0.1` as well, as both addresses are just aliases on the `lo` adapter.

I tried to make the logic deal with `GuestIP` as well, but it quickly got confusing, and I don't actually see any functionality it would give us, so I dropped it again. Maybe this will be different when we get an actual IP address via `vde_vmnet`?

AkihiroSuda (commented):
I'd like to see unit tests and/or integration tests, but tests can be added later

AkihiroSuda:
I guess the "guestIP" isn't going to be used in most cases. The field is only useful when an application inside VM calls `bind("192.168.X.X:Y")`.

jandubois:
> I'd like to see unit tests and/or integration tests, but tests can be added later

I've been trying to create integration tests today, but can't make them work reliably because there is no way to flush the hostagent logs to disk. While looking into this I noticed that `logrus` is in maintenance mode, but all the newer libraries don't seem to have a `Flush` API either (but maybe one can implement it at the app level, which doesn't seem to be possible with `logrus`).

Anyways, will change the tests to shut down the instance, and analyze the logs afterwards.

BTW, I noticed that virtually *all* the output from the hostagent except for status messages goes to stderr. If this is intentional, then I think it would make more sense to just have a single log stream for the hostagent.

AkihiroSuda:
> I've been trying to create integration tests today, but can't make them work reliably because there is no way to flush the hostagent logs to disk. 

Implemented in commit `hostagent: sync every (io.Writer).Write`. Accidentally pushed to master without opening a PR. 😅 
https://github.com/AkihiroSuda/lima/commit/722b7d1b1f38eaca397bf29f8f83062a72fc78d1

> BTW, I noticed that virtually all the output from the hostagent except for status messages goes to stderr. If this is intentional, then I think it would make more sense to just have a single log stream for the hostagent.

Stdout is guaranteed to be unmarshallable as JSON lines, but stderr is no, because imported libraries may print something to stderr without using logrus. So we have to maintain the two streams.

jandubois:
> Implemented in commit `hostagent: sync every (io.Writer).Write`.

Thanks! It turns out that I also mis-remembered that the guest-agent checks for port changes every second; it only checks every 3 seconds; so my strategic `sleep 2` pauses need to be `sleep 5` to make sure forwards are removed before they are added back...

I see that the `lima-guestagent` has a `--tick` option to specify a different poll frequency, but there is no way to configure it via lima. I'm inclined to add an env variable to set it (`LIMA_GUESTAGENT_TICK`?), or do you want it to be part of the yaml file?

(Wild idea for some distant time in the future: make the polling adaptive: poll more frequently whenever a new process or container is being created, as ports are typically opened early in the lifetime of a server process. Back off later, and especially when running on battery. Not sure if it is worth it; maybe it is just related to me seeing a demo using eBPF/LSM hooks yesterday 😄 ).

> Stdout is guaranteed to be unmarshallable as JSON lines, but stderr is not, 

Yes, but stdout seems to contain only 2 lines:

```json
{"time":"2021-07-21T00:09:03.305227-07:00","status":{"sshLocalPort":60020}}
{"time":"2021-07-21T00:09:25.48178-07:00","status":{"running":true,"sshLocalPort":60020}}
```

I don't think they are important enough to require a separate JSON-unmarshallable file. Is there anything else written to it?

Anyways, my argument was that all the tracing and logging should go into stdout as well, and only errors should be written to stderr. Or maybe I'm still confused about what each file is supposed to be used for.

AkihiroSuda:
> I see that the lima-guestagent has a --tick option to specify a different poll frequency, but there is no way to configure it via lima. I'm inclined to add an env variable to set it (LIMA_GUESTAGENT_TICK?), or do you want it to be part of the yaml file?

Either is fine to me, but I think we should clarify that the field is likely to be removed in future.

> (Wild idea for some distant time in the future: make the polling adaptive: poll more frequently whenever a new process or container is being created, as ports are typically opened early in the lifetime of a server process. Back off later, and especially when running on battery. Not sure if it is worth it; maybe it is just related to me seeing a demo using eBPF/LSM hooks yesterday 😄 ).

Yes, actually I was planning to replace ticks with eBPF retprobes for `bind`, though it might be too much complicated.
We can also consider patching CNI portmap to inject a tick to the lima guestagent.

pkg/limayaml/default.yaml
```diff
@@ -73,6 +73,32 @@ containerd:
...
+#   - guestPort: 80
+#     hostPort: 8080 # overrides the default value 80
+#   - guestIP: "127.0.0.2" # overrides the default value "127.0.0.1"
+#     hostIP: "127.0.0.2" # overrides the default value "127.0.0.1"
+#   # default: guestPortRange: [1024, 65535]
+#   # default: hostPortRange: [1024, 65535]
+#   - guestPort: 8888
+#     ignore: true (don't forward this port)
+#  # Lima internally appends this fallback rule at the end:
+#  - guestIP: "127.0.0.1"
```
AkihiroSuda, line 96: nit: indentation seems inconsistent here

jandubois: thanks, fixed

AkihiroSuda (approved):
Thank you!

PR #123, opened 2021-07-23
fix: support for staged installs (DESTDIR prefixed)

leodido:
Ensuring that `$(DESTDIR)/$(PREFIX)` is a canonical name referring to an existing directory (via `$(realpath ...)`) on the `install` and `uninstall` make targets.

```console
make install # -> installs in /usr/local

PREFIX=/usr/local/ make install # -> installs in /usr/local
DESTDIR=/tmp/ PREFIX=/ciao/ make install # -> installs in /tmp/ciao (assuming it already exists)
DESTDIR=/tmp/ PREFIX=ciao make install # -> installs in /tmp/ciao (assuming it already exists)
DESTDIR=/tmp/ PREFIX=ciao/ make install # -> installs in /tmp/ciao (assuming it already exists)
DESTDIR=/tmp/ PREFIX=/ciao/ make install # -> installs in /tmp/ciao (assuming it already exists)
DESTDIR=/tmp PREFIX=/ciao/ make install # -> installs in /tmp/ciao (assuming it already exists)
```

As you can see the goal is to make the `make install` (and `make uninstall`) process more robust w.r.t. to paths the user might express (containing or not leading/trailing slashes). :)

Signed-off-by: Leonardo Di Donato <leodidonato@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 # Files are installed under $(DESTDIR)/$(PREFIX)
 PREFIX ?= /usr/local
+DEST := $(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's:///*:/:g; s://*$$::')
 
 GO ?= go
 
@@ -52,17 +53,19 @@ _output/share/lima/lima-guestagent.Linux-aarch64:
 
 .PHONY: install
 install:
-	cp -av _output/* "$(DESTDIR)/$(PREFIX)/"
-	if [[ $(shell uname -s ) != Linux && ! -e "$(DESTDIR)/$(PREFIX)/bin/nerdctl" ]]; then ln -sf nerdctl.lima "$(DESTDIR)/$(PREFIX)/bin/nerdctl"; fi
+	mkdir -p "$(DEST)"
+	cp -av _output/* "$(DEST)"
+	if [[ $(shell uname -s ) != Linux && ! -e "$(DEST)/bin/nerdctl" ]]; then ln -sf nerdctl.lima "$(DEST)/bin/nerdctl"; fi
 
 .PHONY: uninstall
 uninstall:
+	@test -f "$(DEST)/bin/lima" || (echo "lima not found in $(DEST) prefix"; exit 1)
 	rm -rf \
-		"$(DESTDIR)/$(PREFIX)/bin/lima" \
-		"$(DESTDIR)/$(PREFIX)/bin/limactl" \
-		"$(DESTDIR)/$(PREFIX)/bin/nerdctl.lima" \
-		"$(DESTDIR)/$(PREFIX)/share/lima" "$(DESTDIR)/$(PREFIX)/share/doc/lima"
-	# TODO: remove $(DESTDIR)/$(PREFIX)/bin/nerdctl only when it is a symlink to nerdctl.lima
+		"$(DEST)/bin/lima" \
+		"$(DEST)/bin/limactl" \
+		"$(DEST)/bin/nerdctl.lima" \
+		"$(DEST)/share/lima" "$(DEST)/share/doc/lima"
+	# TODO: remove $(DEST)/bin/nerdctl only when it is a symlink to nerdctl.lima
 
 .PHONY: clean
 clean:


Makefile
```diff
@@ -51,18 +51,22 @@ _output/share/lima/lima-guestagent.Linux-aarch64:
 	chmod 644 $@
 
 .PHONY: install
+install: DEST := $(realpath $(DESTDIR)/$(PREFIX))
 install:
-	cp -av _output/* "$(DESTDIR)/$(PREFIX)/"
-	if [[ $(shell uname -s ) != Linux && ! -e "$(DESTDIR)/$(PREFIX)/bin/nerdctl" ]]; then ln -sf nerdctl.lima "$(DESTDIR)/$(PREFIX)/bin/nerdctl"; fi
+	@test "$(DEST)" || (echo "invalid destination: $(DESTDIR)$(PREFIX)"; exit 1)
```
AkihiroSuda, line 56: Can we mkdir -p when DEST does not exist?

leodido: We could use `$(abspath ...)`. Lemme send another commit with it, 1 min :)

leodido: Please take a look at the following commit :)

AkihiroSuda (approved):
Thanks

Makefile
```diff
@@ -51,18 +51,26 @@ _output/share/lima/lima-guestagent.Linux-aarch64:
 	chmod 644 $@
 
 .PHONY: install
+install: | $(abspath $(DESTDIR)/$(PREFIX))
+install: DEST = $(realpath $(DESTDIR)/$(PREFIX))
```
jandubois, line 55: Since setting `DEST` is repeated on the `uninstall` target, I think it should instead be set globally at the top, just after the line providing the default for `PREFIX`.

Makefile
```diff
@@ -51,18 +51,26 @@ _output/share/lima/lima-guestagent.Linux-aarch64:
 	chmod 644 $@
 
 .PHONY: install
+install: | $(abspath $(DESTDIR)/$(PREFIX))
```
jandubois, line 54: I feel that this is needless complexity/obfuscation; just include `mkdir -p ...` as the first line under the `install` target. It will not produce any error message and will still return a `0` exit code, even if the directory already exists. It is idempotent by design.

Makefile
```diff
@@ -51,18 +51,26 @@ _output/share/lima/lima-guestagent.Linux-aarch64:
...
+install: DEST = $(realpath $(DESTDIR)/$(PREFIX))
 install:
-	cp -av _output/* "$(DESTDIR)/$(PREFIX)/"
-	if [[ $(shell uname -s ) != Linux && ! -e "$(DESTDIR)/$(PREFIX)/bin/nerdctl" ]]; then ln -sf nerdctl.lima "$(DESTDIR)/$(PREFIX)/bin/nerdctl"; fi
+	@test -d "$(DEST)" || (echo "invalid destination: $(DESTDIR)/$(PREFIX)"; exit 1)
```
jandubois, line 57: Given that the directory has just been created, this should never actually trigger. If `mkdir -p` cannot create the directory, it will already fail:

```console
$ mkdir -p /bin/foo;echo $?
mkdir: /bin/foo: Operation not permitted
1
```

And if the directory does already exist, but is not writable, then the `cp` command is going to fail with a relevant error message.

So I think the test serves no useful purpose and should be removed.

Makefile
```diff
@@ -51,18 +51,26 @@ _output/share/lima/lima-guestagent.Linux-aarch64:
...
+	cp -av _output/* "$(DEST)/"
+	if [[ $(shell uname -s ) != Linux && ! -e "$(DEST)/bin/nerdctl" ]]; then ln -sf nerdctl.lima "$(DEST)/bin/nerdctl"; fi
 
 .PHONY: uninstall
+uninstall: DEST = $(realpath $(DESTDIR)/$(PREFIX))
 uninstall:
+	@test -d "$(DEST)" || (echo "invalid destination: $(DESTDIR)/$(PREFIX)"; exit 1)
```
jandubois, line 64: I would maybe check for `-f "${DEST}/bin/lima` instead, to verify not just that the directory exists, but also that lima has previously been installed into this location.

jandubois (commented):
I think the PR will work as-is, but I would prefer that the comments I made were addressed.

Makefile
leodido: I can do it, and I agree that simple is better :)

But it would not work for a destination that does not already exist, because `realpath` returns an empty string in that case

leodido: What we can do is: move DEST to the top scope but using `abspath` (which does not return an empty string for directories that do not already exist), similar to what I was doing below in the target prerequisite.

jandubois: > What we can do is: move DEST to the top scope but using `abspath`

I realize the original intent of this PR was to use `realpath` to check if the path exists, and fail if it doesn't. But now that the path is created when it doesn't exist, is there any reason to call `abspath`? Does anything user-observable work differently?

Sorry for getting pretty close to nit-picking territory now, but if the call doesn't serve any functional purpose, and doesn't cause any user-visible change in behaviour, then maybe it should be removed too? Because code that does not exist cannot have bugs. 😄 

And thinking about it some more, `abspath` may actually introduce bugs because it can generate the wrong path when symlinks and `../` path segments are involved:

```console
$ grep -A1 abspath: Makefile
abspath:
	@echo $(DEST)
$ ln -s /usr/local ~/foo
$ export DESTDIR=~/foo
$ export PREFIX=../lib
$ make abspath
/Users/jan/lib
$ (cd $DESTDIR/$PREFIX;pwd)
/usr/lib
```

So now I think the `abspath` call should be removed unless I've missed some reason it is useful/required.

leodido: Hello Jan,

yes, the scope of the PR was that one.

The reason I used `abspath` is the same I used `realpath` in the first iteration of this PR.

Ie., to make the process of staged installs (via `DESTDIR`) more robust and clear (and actually work) w.r.t. to trailing slashes, etc.

This is what the user will see if we remove `abspath`:

```diff
diff --git a/Makefile b/Makefile
index 0b74f88..8e36c16 100644
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Files are installed under $(DESTDIR)/$(PREFIX)
 PREFIX ?= /usr/local
-DEST := $(abspath $(DESTDIR)/$(PREFIX))
+DEST := $(DESTDIR)/$(PREFIX)

 GO ?= go
```

Outputs:

```console
DESTDIR=/tmp/ make install
mkdir -p "/tmp///usr/local"
cp -av _output/* "/tmp///usr/local"
_output/bin -> /tmp///usr/local/bin
...
```

If this is what you want, I'll do it. Even if do not totally like it atm :)

jandubois: Hi Leo,

> If this is what you want, I'll do it.

I'm only providing code review feedback. If you disagree, please push back! In the end it depends on what @AkihiroSuda wants.

As for the multiple slashes, isn't that a purely cosmetic issue? POSIX pathname definition says:

> Multiple successive &lt;slash> characters are considered to be the same as one &lt;slash>, except for the case of exactly two leading <slash> characters.

So paths starting with two leading slashes are potentially a concern, but I believe neither Linux nor macOS implement any special semantics in this case.

I do understand where you are coming from, as I'm always annoyed that e.g. `find /etc/ ...` shows names as `/etc//...` (and `find /etc ...` doesn't work because `/etc` is a symlink to `private/etc` on macOS).

If you are purely concerned about the successive slashes, maybe a strict textual substitution is preferable:

```Makefile
DEST := $(patsubst %/,%,$(DESTDIR))/$(patsubst /%,%,$(PREFIX))
```

But again this introduces subtle bugs, in this case if the paths contain consecutive whitespace:

```console
$ make abspath DESTDIR="/foo bar  baz/xyzzy/"
/foo bar baz/xyzzy/usr/local
```

So my personal preference would be to favour correctness over aesthetics and accept the untidy output, especially since this is about `make install` and not related to using `lima`.

Let's wait for @AkihiroSuda to make a decision.

jandubois: > this introduces subtle bugs, in this case if the paths contain consecutive whitespace

This is also true for `DEST := $(DESTDIR)/$(PREFIX)`; the right hand side must be quoted to preserve whitespace:

```Makefile
DEST := "$(DESTDIR)/$(PREFIX)"
```

But this doesn't help with the `patsubst` expression, because `patsubst` itself does whitespace folding (and drops leading/trailing whitespace).

Here is one more attempt that seems to work correctly in all cases I can think off:

```Makefile
DEST := "$(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's:///*:/:g')"
```

Samples:

```console
$ make abspath
/usr/local
$ make abspath DESTDIR="/foo bar  baz//xyzzy/"
/foo bar  baz/xyzzy/usr/local
```

Can you see anything wrong with this one?

leodido: Hi Jan,

that final sentence - not being a native speaker - probably came out wrong!

Please excuse me in case it sounded rude. It wasn't meant to.

> I'm only providing code review feedback. If you disagree, please push back!

At all! On the contrary, I want to thank you for your reviews, which have been very on the point and helpful.

Yes, it's mainly cosmetic 😃 

Such details bother me. I believe you can totally feel me.
But I also agree with you we should put aesthetics in second place in this case. Meaning, try to achieve it only in case correctness it's achieved.

In this context, I believe that `sed s://*:/:g` would make it.

Another option would be `readlink -m "/foo bar  baz//xyzzy/"` but it canonicalizes symlinks (thus I guess it's not what we want, right?).

leodido: Also, are we fine introducing `sed` as "dependency"?

jandubois: > Please excuse me in case it sounded rude.

Oh no, it didn't. I just wanted to point out that I also am just voicing out my opinions, and I can be wrong or overlook stuff. So if you feel that I'm wrong, please push back and don't just change something because "I want it". Code review is a collaborative process, with the end result hopefully better than what either of the participants would achieve by themselves.

> Another option would be readlink -m "/foo bar baz//xyzzy/"

I would prefer not to do this. To begin with because it doesn't work on macOS. 😄 

> Also, are we fine introducing sed as "dependency"?

I would think so; it is a pretty basic requirement. And I intentionally used an expression that works with the basic pattern support. Normally I would prefer to use `perl` because I know it works the same everywhere (unlike `sed`), but I don't wan't to rely on it being installed.

leodido: I probably found one last edge (cosmetic) case: DEST to contain a trailing slash. In this case the `uninstall` command looks this way:

```console
PREFIX=/usr/local/ make uninstall
rm -rf \
	"/usr/local//bin/lima" \
	"/usr/local//bin/limactl" \
	"/usr/local//bin/nerdctl.lima" \
	"/usr/local//share/lima" "/usr/local//share/doc/lima"
```

Thus, I'm thinking to define DEST like this:

```makefile
DEST := $(patsubst %/,%,$(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's://*:/:g'))
```

Or with only sed:

```makefile
DEST := $(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's#//*#/#g;s#/*$$##')
```

Results:

```console
PREFIX="/uno/due/foo bar baz//xyzzy/" make prova
/uno/due/foo bar baz/xyzzy
```

```console
PREFIX=/usr/local make prova
/usr/local
```

What do you think? Which one do you prefer? :)

jandubois: > Thus, I'm thinking to define DEST like this:
> 
> ```makefile
> DEST := $(patsubst %/,%,$(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's://*:/:g'))
> ```

As I explained before, `patsubst` will fold multiple spaces into a single space, so should not be used. Maybe it was too subtle, but I intentionally had 2 spaces after `bar`. You dropped the multiple spaces in your test case, so you didn't notice.

 ```
/foo bar  baz
     ^   ^^
````

> Or with only sed:
> 
> ```makefile
> DEST := $(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's#//*#/#g;s#/*$$##')
> ```

Yes, just `sed` is fine. For some reason your command doesn't work for me:

```console
Makefile:3: *** unterminated call to function `shell': missing `)'.  Stop.
```

It has something to do with `#` also being a comment character, and at least "GNU Make 3.81" does not parse it correctly.

You are also still missing the quotes around the right hand side to preserve whitespace.

I've extended my sample, and it worked for me, so I would suggest to use that version (assuming it works for you as well):

```makefile
DEST := "$(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's:///*:/:g; s://*$$::')"
```

(You also dropped one slash from `///*`, which is just a basic pattern way to write `//+`, which only works with `-E`. Of course you can replace a single slash by another identical single slash, but that it kind of pointless. Not that it really matters either way, but since this PR is about aesthetics...)

AkihiroSuda:
Could you consider squashing commits?

Makefile
leodido: Thank you, Jan!

Yes, I copied and pasted the commands from the shell not from the Makefile I was actually using, silly me 🤦 😄

leodido:
> Could you consider squashing commits?

Of course, will do!

jandubois:
Can you please fix your sign-off so that the email address matches the commit author?

```
commit 6aa2d56b34f2fa64375e6eb141b584a1add7a169 (HEAD -> fix/destdir+prefix)
Author: Leonardo Di Donato <leodidonato@gmail.com>
Date:   Fri Jul 23 17:39:16 2021 +0200

    fix: support for staged installs (DESTDIR)

    Signed-off-by: Leonardo Di Donato <leodido@linux.com>
```

Makefile
```diff
@@ -52,17 +53,19 @@ _output/share/lima/lima-guestagent.Linux-aarch64:
 
 .PHONY: install
 install:
-	cp -av _output/* "$(DESTDIR)/$(PREFIX)/"
-	if [[ $(shell uname -s ) != Linux && ! -e "$(DESTDIR)/$(PREFIX)/bin/nerdctl" ]]; then ln -sf nerdctl.lima "$(DESTDIR)/$(PREFIX)/bin/nerdctl"; fi
+	mkdir -p $(DEST)
```
jandubois, line 56: Every expression that interpolates `$(DEST)` (or any other variable) needs to be put in quotes. Otherwise it will break when the path contains spaces. This applies also to virtually all new lines below (see the original lines, that did include the quotes; not sure why you removed them).

leodido: I did it because of:

```console
$ DESTDIR=/tmp/ PREFIX="/foo bar  baz" make install
mkdir -p "/tmp/foo bar  baz"
cp -av _output/* "/tmp/foo bar  baz"
_output/bin -> /tmp/foo bar  baz/bin
_output/bin/lima -> /tmp/foo bar  baz/bin/lima
_output/bin/nerdctl.lima -> /tmp/foo bar  baz/bin/nerdctl.lima
_output/bin/limactl -> /tmp/foo bar  baz/bin/limactl
_output/share -> /tmp/foo bar  baz/share
_output/share/lima -> /tmp/foo bar  baz/share/lima
_output/share/lima/lima-guestagent.Linux-x86_64 -> /tmp/foo bar  baz/share/lima/lima-guestagent.Linux-x86_64
_output/share/lima/lima-guestagent.Linux-aarch64 -> /tmp/foo bar  baz/share/lima/lima-guestagent.Linux-aarch64
_output/share/doc -> /tmp/foo bar  baz/share/doc
_output/share/doc/lima -> /tmp/foo bar  baz/share/doc/lima
_output/share/doc/lima/LICENSE -> /tmp/foo bar  baz/share/doc/lima/LICENSE
_output/share/doc/lima/docs -> /tmp/foo bar  baz/share/doc/lima/docs
_output/share/doc/lima/docs/internal.md -> /tmp/foo bar  baz/share/doc/lima/docs/internal.md
_output/share/doc/lima/README.md -> /tmp/foo bar  baz/share/doc/lima/README.md
_output/share/doc/lima/VERSION -> /tmp/foo bar  baz/share/doc/lima/VERSION
_output/share/doc/lima/examples -> /tmp/foo bar  baz/share/doc/lima/examples
_output/share/doc/lima/examples/debian.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/debian.yaml
_output/share/doc/lima/examples/alpine.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/alpine.yaml
_output/share/doc/lima/examples/ubuntu.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/ubuntu.yaml
_output/share/doc/lima/examples/fedora.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/fedora.yaml
_output/share/doc/lima/examples/README.md -> /tmp/foo bar  baz/share/doc/lima/examples/README.md
_output/share/doc/lima/examples/archlinux.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/archlinux.yaml
_output/share/doc/lima/examples/opensuse.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/opensuse.yaml
_output/share/doc/lima/examples/k3s.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/k3s.yaml
_output/share/doc/lima/examples/k3s-rootless.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/k3s-rootless.yaml
_output/share/doc/lima/examples/default.yaml -> /tmp/foo bar  baz/share/doc/lima/examples/default.yaml
if [[ Darwin != Linux && ! -e "/tmp/foo bar  baz"/bin/nerdctl ]]; then ln -sf nerdctl.lima "/tmp/foo bar  baz"/bin/nerdctl; fi
```

```console
DESTDIR=/tmp/ PREFIX="/foo bar  baz" make uninstall
rm -rf \
	"/tmp/foo bar  baz"/bin/lima \
	"/tmp/foo bar  baz"/bin/limactl \
	"/tmp/foo bar  baz"/bin/nerdctl.lima \
	"/tmp/foo bar  baz"/share/lima "/tmp/foo bar  baz"/share/doc/lima
```

While, with quotes:

```diff
diff --git a/Makefile b/Makefile
index 3e3dd4b..c47c5b0 100644
--- a/Makefile
+++ b/Makefile
@@ -53,18 +53,18 @@ _output/share/lima/lima-guestagent.Linux-aarch64:
 
 .PHONY: install
 install:
-       mkdir -p $(DEST)
-       cp -av _output/* $(DEST)
-       if [[ $(shell uname -s ) != Linux && ! -e $(DEST)/bin/nerdctl ]]; then ln -sf nerdctl.lima $(DEST)/bin/nerdctl; fi
+       mkdir -p "$(DEST)"
+       cp -av _output/* "$(DEST)"
+       if [[ $(shell uname -s ) != Linux && ! -e "$(DEST)/bin/nerdctl" ]]; then ln -sf nerdctl.lima "$(DEST)/bin/nerdctl"; fi
 
 .PHONY: uninstall
 uninstall:
-       @test -f $(DEST)/bin/lima || (echo "lima not found in $(DEST) prefix"; exit 1)
+       @test -f "$(DEST)/bin/lima" || (echo "lima not found in $(DEST) prefix"; exit 1)
        rm -rf \
-               $(DEST)/bin/lima \
-               $(DEST)/bin/limactl \
-               $(DEST)/bin/nerdctl.lima \
-               $(DEST)/share/lima $(DEST)/share/doc/lima
+               "$(DEST)/bin/lima" \
+               "$(DEST)/bin/limactl" \
+               "$(DEST)/bin/nerdctl.lima" \
+               "$(DEST)/share/lima $(DEST)/share/doc/lima"
        # TODO: remove $(DEST)/bin/nerdctl only when it is a symlink to nerdctl.lima
```

```console
DESTDIR=/tmp/ PREFIX="/foo bar  baz" make install
mkdir -p ""/tmp/foo bar  baz""
cp -av _output/* ""/tmp/foo bar  baz""
_output/bin -> baz/bin
_output/bin/lima -> baz/bin/lima
_output/bin/nerdctl.lima -> baz/bin/nerdctl.lima
_output/bin/limactl -> baz/bin/limactl
_output/share -> baz/share
_output/share/lima -> baz/share/lima
_output/share/lima/lima-guestagent.Linux-x86_64 -> baz/share/lima/lima-guestagent.Linux-x86_64
_output/share/lima/lima-guestagent.Linux-aarch64 -> baz/share/lima/lima-guestagent.Linux-aarch64
_output/share/doc -> baz/share/doc
_output/share/doc/lima -> baz/share/doc/lima
_output/share/doc/lima/LICENSE -> baz/share/doc/lima/LICENSE
_output/share/doc/lima/docs -> baz/share/doc/lima/docs
_output/share/doc/lima/docs/internal.md -> baz/share/doc/lima/docs/internal.md
_output/share/doc/lima/README.md -> baz/share/doc/lima/README.md
_output/share/doc/lima/VERSION -> baz/share/doc/lima/VERSION
_output/share/doc/lima/examples -> baz/share/doc/lima/examples
_output/share/doc/lima/examples/debian.yaml -> baz/share/doc/lima/examples/debian.yaml
_output/share/doc/lima/examples/alpine.yaml -> baz/share/doc/lima/examples/alpine.yaml
_output/share/doc/lima/examples/ubuntu.yaml -> baz/share/doc/lima/examples/ubuntu.yaml
_output/share/doc/lima/examples/fedora.yaml -> baz/share/doc/lima/examples/fedora.yaml
_output/share/doc/lima/examples/README.md -> baz/share/doc/lima/examples/README.md
_output/share/doc/lima/examples/archlinux.yaml -> baz/share/doc/lima/examples/archlinux.yaml
_output/share/doc/lima/examples/opensuse.yaml -> baz/share/doc/lima/examples/opensuse.yaml
_output/share/doc/lima/examples/k3s.yaml -> baz/share/doc/lima/examples/k3s.yaml
_output/share/doc/lima/examples/k3s-rootless.yaml -> baz/share/doc/lima/examples/k3s-rootless.yaml
_output/share/doc/lima/examples/default.yaml -> baz/share/doc/lima/examples/default.yaml
/tmp/foo -> baz/foo
bar -> baz/bar
if [[ Darwin != Linux && ! -e ""/tmp/foo bar  baz"/bin/nerdctl" ]]; then ln -sf nerdctl.lima ""/tmp/foo bar  baz"/bin/nerdctl"; fi
/bin/sh: -c: line 0: syntax error in conditional expression
/bin/sh: -c: line 0: syntax error near `bar'
/bin/sh: -c: line 0: `if [[ Darwin != Linux && ! -e ""/tmp/foo bar  baz"/bin/nerdctl" ]]; then ln -sf nerdctl.lima ""/tmp/foo bar  baz"/bin/nerdctl"; fi'
make: *** [Makefile:58: install] Error 2
```

And assuming it was installed at the step before...

```console
DESTDIR=/tmp/ PREFIX="/foo bar  baz" make uninstall
/bin/sh: line 0: test: too many arguments
lima not found in /tmp/foo bar baz prefix
make: *** [Makefile:62: uninstall] Error 1
```

---

```console
make --version
GNU Make 4.3
```

jandubois: You are absolutely right, and it is actually my fault for (incorrectly) insisting that

```
DEST := "$(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's:///*:/:g; s://*$$::')"
```

must include quotes on the right hand side (because I was testing with `@echo $(DEST)` instead of `@echo "$(DEST)"`). That was wrong; the quotes should not be part of the value, but should be added when the variable is used.

So please remove the quotes from the `DEST := ...` line and restore them in all the original places!

Sorry for causing this confusion and extra work!

jandubois (approved):
Thank you! Looks good to merge now.

AkihiroSuda (approved):
Thanks

PR #129, opened 2021-07-27
trap handlers don't stack; there is only one per signal

jandubois:
Each new `trap ... EXIT` command simply replaces the previous trap handler with the new one, so only the very last one will be executed.

Side-question: why do we create all the temp directories and files in the `$HOME` directory? Is `/tmp` not writable? Can't we use `mktemp`?


AkihiroSuda (approved):
Thanks

Issue #81, opened 2021-06-25
Support vmnet.framework (using vde_vmnet)


AkihiroSuda:
~~With this qemu patch: https://lore.kernel.org/qemu-devel/20210617143246.55336-1-yaroshchuk2000@gmail.com/~~ https://github.com/AkihiroSuda/vde_vmnet

AkihiroSuda:
https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_vm_networking
```
This entitlement is restricted to developers of virtualization software. To request this entitlement, contact your Apple representative.
```

Looks like VMnet (without root) requires agreement with Apple 😞 

So I guess vmnet-enabled version of QEMU isn't likely to be available in Homebrew repo....

AkihiroSuda:
Experimental `vmnet.framework` support for unpatched rootless QEMU: https://github.com/AkihiroSuda/vde_vmnet
This `vde_vmnet` does not require QEMU to run as the root user, though `vde_vmnet` itself has to run as the root

PR #133, opened 2021-07-30
Add networking options for vde_vmnet support

jandubois:
Ref #81

### DIFF
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -31,6 +31,19 @@ memory: "4GiB"
 # Default: "100GiB"
 disk: "100GiB"
 
+network:
+  # The instance can get a routable IP address from the vmnet framework using
+  # https://github.com/AkihiroSuda/vde_vmnet. Both vde_switch and vde_vmnet
+  # daemons must be running before the instance is started. The interface type
+  # (host, shared, or bridged) is configured in vde_vmnet and not lima.
+  vde:
+    # vde_switch socket directory, normally /var/run/vde.ctl
+    # Default: "" (vmnet is not being used)
+    url: ""
+    # MAC address of the instance; lima will pick one based on the instance name,
+    # so DHCP assigned ip addresses should remain constant over instance restarts.
+    macAddress: ""
+
 # Expose host directories to the guest
 # Default: none
 mounts:
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -32,17 +32,18 @@ memory: "4GiB"
 disk: "100GiB"
 
 network:
-  # The instance can get a routable IP address from the vmnet framework using
+  # The instance can get routable IP addresses from the vmnet framework using
   # https://github.com/AkihiroSuda/vde_vmnet. Both vde_switch and vde_vmnet
   # daemons must be running before the instance is started. The interface type
   # (host, shared, or bridged) is configured in vde_vmnet and not lima.
   vde:
-    # vde_switch socket directory, normally /var/run/vde.ctl
-    # Default: "" (vmnet is not being used)
-    url: ""
-    # MAC address of the instance; lima will pick one based on the instance name,
-    # so DHCP assigned ip addresses should remain constant over instance restarts.
-    macAddress: ""
+    # url points to the vde_switch socket directory
+    # - url: "/var/run/vde.ctl"
+    #   # MAC address of the instance; lima will pick one based on the instance name,
+    #   # so DHCP assigned ip addresses should remain constant over instance restarts.
+    #   macAddress: ""
+    #   # Interface name, defaults to "vde0", "vde1", etc.
+    #   name: ""
 
 # Expose host directories to the guest
 # Default: none
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -1,13 +1,16 @@
 package limayaml
 
 import (
+	"crypto/sha256"
 	"fmt"
+	"net"
 	"runtime"
+	"strconv"
 
 	"github.com/AkihiroSuda/lima/pkg/guestagent/api"
 )
 
-func FillDefault(y *LimaYAML) {
+func FillDefault(y *LimaYAML, filePath string) {
 	y.Arch = resolveArch(y.Arch)
 	for i := range y.Images {
 		img := &y.Images[i]
@@ -54,6 +57,21 @@ func FillDefault(y *LimaYAML) {
 	for i := range y.PortForwards {
 		FillPortForwardDefaults(&y.PortForwards[i])
 		// After defaults processing the singular HostPort and GuestPort values should not be used again.
+	}
+	for i := range y.Network.VDE {
+		vde := &y.Network.VDE[i]
+		if vde.MACAddress == "" {
+			// every interface in every limayaml file must get its own unique MAC address
+			uniqueID := fmt.Sprintf("%s#%d", filePath, i)
+			sha := sha256.Sum256([]byte(uniqueID))
+			// According to https://gitlab.com/wireshark/wireshark/-/blob/master/manuf
+			// no well-known MAC addresses start with 0x22.
+			hw := append(net.HardwareAddr{0x22}, sha[0:5]...)
+			vde.MACAddress = hw.String()
+		}
+		if vde.Name == "" {
+			vde.Name = "vde" + strconv.Itoa(i)
+		}
 	}
 }
 
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -133,10 +133,14 @@ func appendArgsIfNoConflict(args []string, k, v string) []string {
 	return append(args, k, v)
 }
 
-func macAddress(cfg Config) string {
-	addr := cfg.LimaYAML.Network.VDE.MACAddress
+const (
+	SlirpMACAddress = "22:11:11:11:11:11"
+)
+
+func MACAddress(instanceDir string, y *limayaml.LimaYAML) string {
+	addr := y.Network.VDE.MACAddress
 	if addr == "" {
-		sha := sha256.Sum256([]byte(cfg.InstanceDir))
+		sha := sha256.Sum256([]byte(instanceDir))
 		// According to https://gitlab.com/wireshark/wireshark/-/blob/master/manuf
 		// no well-known MAC addresses start with 0x22.
 		hw := append(net.HardwareAddr{0x22}, sha[0:5]...)
@@ -209,12 +213,12 @@ func Cmdline(cfg Config) (string, []string, error) {
 
 	// Network
 	if y.Network.VDE.URL != "" {
-		args = append(args, "-device", fmt.Sprintf("virtio-net-pci,netdev=net0,mac=%s", macAddress(cfg)))
 		args = append(args, "-netdev", fmt.Sprintf("vde,id=net0,sock=%s", y.Network.VDE.URL))
+		args = append(args, "-device", fmt.Sprintf("virtio-net-pci,netdev=net0,mac=%s", MACAddress(cfg.InstanceDir, y)))
 	}
 	// CIDR is intentionally hardcoded to 192.168.5.0/24, as each of QEMU has its own independent slirp network.
-	args = append(args, "-device", "virtio-net-pci,netdev=net1")
 	args = append(args, "-netdev", fmt.Sprintf("user,id=net1,net=192.168.5.0/24,hostfwd=tcp:127.0.0.1:%d-:22", y.SSH.LocalPort))
+	args = append(args, "-device", "virtio-net-pci,netdev=net1,mac="+SlirpMACAddress)
 
 	// virtio-rng-pci acceralates starting up the OS, according to https://wiki.gentoo.org/wiki/QEMU/Options
 	args = append(args, "-device", "virtio-rng-pci")
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -1,9 +1,7 @@
 package qemu
 
 import (
-	"crypto/sha256"
 	"fmt"
-	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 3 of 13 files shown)


pkg/limayaml/default.yaml
```diff
@@ -31,6 +31,19 @@ memory: "4GiB"
...
 disk: "100GiB"
 
+network:
+  # The instance can get a routable IP address from the vmnet framework using
+  # https://github.com/AkihiroSuda/vde_vmnet. Both vde_switch and vde_vmnet
+  # daemons must be running before the instance is started. The interface type
+  # (host, shared, or bridged) is configured in vde_vmnet and not lima.
+  vde:
+    # vde_switch socket directory, normally /var/run/vde.ctl
+    # Default: ~ (vmnet is not being used)
```
AkihiroSuda, line 41: The default should be empty.

~ means home directory.

pkg/limayaml/default.yaml
```diff
@@ -31,6 +31,19 @@ memory: "4GiB"
...
+    vdeSocketDir: ~
+    # MAC address of the instance; lima will pick one based on the instance name,
+    # so DHCP assigned ip addresses should remain constant over instance restarts.
+    macAddress: ~
```
AkihiroSuda, line 45: Why ~ ?

pkg/qemu/qemu.go
```diff
@@ -194,8 +196,19 @@ func Cmdline(cfg Config) (string, []string, error) {
 	args = append(args, "-cdrom", filepath.Join(cfg.InstanceDir, filenames.CIDataISO))
 
 	// Network
+	if y.Network.VDE.SocketDir != "" {
+		macAddr := y.Network.VDE.MACAddress
+		if macAddr == "" {
+			sha1 := sha1.Sum([]byte(cfg.InstanceDir))
```
AkihiroSuda, line 202: Any reason not to use sha256?

pkg/qemu/qemu.go
```diff
@@ -194,8 +196,19 @@ func Cmdline(cfg Config) (string, []string, error) {
...
+			hw := append(net.HardwareAddr{0x22}, sha1[0:5]...)
+			macAddr = hw.String()
+		}
+		args = append(args, "-device", fmt.Sprintf("virtio-net-pci,netdev=net0,mac=%s", macAddr))
+		args = append(args, "-netdev", fmt.Sprintf("vde,id=net0,sock=%s", y.Network.VDE.SocketDir))
+	}
 	// CIDR is intentionally hardcoded to 192.168.5.0/24, as each of QEMU has its own independent slirp network.
-	// TODO: enable bridge (with sudo?)
 	args = append(args, "-net", "nic,model=virtio")
 	args = append(args, "-net", fmt.Sprintf("user,net=192.168.5.0/24,hostfwd=tcp:127.0.0.1:%d-:22", y.SSH.LocalPort))
```
AkihiroSuda, line 213: We should specify “id=net1” here?

pkg/qemu/qemu.go
AkihiroSuda: Also, can we split function?

pkg/limayaml/default.yaml
jandubois: I assume you are just talking about the comment, and want it to look like this:

```
    # Default: "" (vmnet is not being used)
```

Do you want the YAML values changed as well? Because there `~` means `null` and not `"~"`.

AkihiroSuda: Aha, didn’t know that… thanks for the info.

 I prefer to use empty string, as ~ really looks like home directory.

pkg/limayaml/default.yaml
jandubois: It is `null` or undefined: [YAML core schema](http://yaml.org/spec/1.2/spec.html#id2805071)

pkg/qemu/qemu.go
jandubois: No reason. It doesn't really matter, as it is not a security thing. I'll change it.

pkg/qemu/qemu.go
jandubois: Good idea. Because the ordering alone doesn't map the VDE interface to `eth0`; it is still `eth1` for me...

jandubois: > Because the ordering alone doesn't map the VDE interface to `eth0`; it is still `eth1` for me...

Adding the id didn't help; SLIRP continues to be `eth0`.

pkg/limayaml/default.yaml
jandubois: I've put `null` for the value, but `""` in the comment. I can put `""` as the values as well, as that's what they'll eventually end up being anyways.

pkg/limayaml/default.yaml
```diff
@@ -31,6 +31,19 @@ memory: "4GiB"
...
 
+network:
+  # The instance can get a routable IP address from the vmnet framework using
+  # https://github.com/AkihiroSuda/vde_vmnet. Both vde_switch and vde_vmnet
+  # daemons must be running before the instance is started. The interface type
+  # (host, shared, or bridged) is configured in vde_vmnet and not lima.
+  vde:
+    # vde_switch socket directory, normally /var/run/vde.ctl
+    # Default: "" (vmnet is not being used)
+    vdeSocketDir: null
```
AkihiroSuda, line 42: Calling this field “location” might be more consistent with other fields in the YAML.

Or just “vdeSocket” or “vdeURL”, as this path can be non-directory socket when we support PTP mode (switchless mode).

Also, we will be able to support more non-directory  networking types when QEMU upgrades VDE2 to VDEplug4

https://github.com/rd235/vdeplug4/blob/089984200f447abb0e825eb45548b781ba1ebccd/man/libvdeplug.3#L41

jandubois: I went with `url` for future-proofing. I can't bring myself to call it a socket when it is actually a directory containing sockets. 😄

jandubois:
I think I've addressed all comments now.

I don't know how I can force QEMU to use eth0 for vde and eth1 for slirp; any suggestions?

pkg/qemu/qemu.go
```diff
@@ -194,10 +208,13 @@ func Cmdline(cfg Config) (string, []string, error) {
 	args = append(args, "-cdrom", filepath.Join(cfg.InstanceDir, filenames.CIDataISO))
 
 	// Network
+	if y.Network.VDE.URL != "" {
+		args = append(args, "-device", fmt.Sprintf("virtio-net-pci,netdev=net0,mac=%s", macAddress(cfg)))
+		args = append(args, "-netdev", fmt.Sprintf("vde,id=net0,sock=%s", y.Network.VDE.URL))
+	}
 	// CIDR is intentionally hardcoded to 192.168.5.0/24, as each of QEMU has its own independent slirp network.
-	// TODO: enable bridge (with sudo?)
 	args = append(args, "-net", "nic,model=virtio")
```
AkihiroSuda, line 216: This line probably needs netdev=net1

jandubois: Thanks! This fixed the interface ordering issue.

AkihiroSuda:
Thanks, but the slirp eth does not link up for me 

```console
suda@lima-default:~$ ip a
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: enp0s1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 22:34:71:8a:4e:62 brd ff:ff:ff:ff:ff:ff
    inet 192.168.105.3/24 brd 192.168.105.255 scope global dynamic enp0s1
       valid_lft 85382sec preferred_lft 85382sec
    inet6 fd69:599c:c913:ae3f:2034:71ff:fe8a:4e62/64 scope global dynamic mngtmpaddr noprefixroute 
       valid_lft 2591992sec preferred_lft 604792sec
    inet6 fe80::2034:71ff:fe8a:4e62/64 scope link 
       valid_lft forever preferred_lft forever
3: enp0s2: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
    link/ether 52:54:00:12:34:56 brd ff:ff:ff:ff:ff:ff
```

Config: the default template but `network.vde.url = "/var/run/vde.ctl"`

jandubois:
> Thanks, but the slirp eth does not link up for me

I'll take a look; I've done all my testing with a modified Alpine image (just added an entry for `eth1` to `/etc/network/interfaces`). Will try with `default` image.

AkihiroSuda:
https://cloudinit.readthedocs.io/en/latest/topics/network-config.html#fallback-network-configuration

> Cloud-init will attempt to determine which of any attached network devices is most likely to have a connection and then generate a network configuration to issue a DHCP request on that interface.

Looks like it only enables the first interface by default.

So we need to generate network config explicitly 😞 

https://cloudinit.readthedocs.io/en/latest/topics/network-config-format-v2.html#examples

jandubois:
> So we need to generate network config explicitly

Yes, that's what I've been thinking too. I will try if a simple config is going to work:

```
network:
  version: 2
  ethernets:
    id0:
      match:
        macaddress: '00:11:22:33:44:55'
      dhcp4: true
    id1:
      match:
        macaddress: '00:11:22:33:44:66'
      dhcp4: true
```

jandubois:
@AkihiroSuda you said (on Slack): "vde will be eth0 and slirp will be eth1". Does it really matter?

If not then I would prefer to always use `eth0` for slirp, so it has a predictable interface name. Otherwise it will be `eth1` when you configure vde, but `eth0` if you don't.

AkihiroSuda:
> If not then I would prefer to always use eth0 for slirp, so it has a predictable interface name. Otherwise it will be eth1 when you configure vde, but eth0 if you don't.

SGTM

jandubois:
I've updated the PR to allow the specification of multiple interfaces (e.g. both a host and a bridged network).

For testing I defined 2 interfaces on the same vde_switch (which doesn't make much sense, but shows the feature):

```yaml
network:
  vde:
  - url: "/var/run/vde.ctl"
  - url: "/var/run/vde.ctl"
    name: "yo0"
```

With the `default` template it produces (IPv6 lines removed for brevity):

```console
$ lima ip addr
bash: line 1: cd: /Users/jan/suse/lima: No such file or directory
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 22:11:11:11:11:11 brd ff:ff:ff:ff:ff:ff
    altname enp0s1
    inet 192.168.5.15/24 brd 192.168.5.255 scope global dynamic eth0
       valid_lft 85741sec preferred_lft 85741sec
3: vde0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 22:0b:6d:67:d6:48 brd ff:ff:ff:ff:ff:ff
    altname enp0s2
    inet 192.168.105.11/24 brd 192.168.105.255 scope global dynamic vde0
       valid_lft 84877sec preferred_lft 84877sec
4: yo0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    link/ether 22:59:7a:b1:ad:46 brd ff:ff:ff:ff:ff:ff
    altname enp0s3
    inet 192.168.105.10/24 brd 192.168.105.255 scope global dynamic yo0
       valid_lft 84877sec preferred_lft 84877sec
```

This required some refactoring, as `limayaml.FillDefault` now needs the file path of the YAML file to calculate the default MAC addresses. I've also removed `limayaml.Validate` and renamed `ValidateRaw` to `Validate`. The places that used to call the original `Validate` function had already called `FillDefaults` anyways, so there was no need for it (unless I missed something).

pkg/limayaml/defaults.go
```diff
@@ -55,6 +58,21 @@ func FillDefault(y *LimaYAML) {
 		FillPortForwardDefaults(&y.PortForwards[i])
 		// After defaults processing the singular HostPort and GuestPort values should not be used again.
 	}
+	for i := range y.Network.VDE {
+		vde := &y.Network.VDE[i]
+		if vde.MACAddress == "" {
+			// every interface in every limayaml file must get its own unique MAC address
+			uniqueID := fmt.Sprintf("%s#%d", filePath, i)
```
AkihiroSuda, line 65: We should also put host UUID to avoid MAC collision when multiple Mac hosts with bridged mode are connected to the same physical network. (Can be another PR)

https://apple.stackexchange.com/questions/342042/how-can-i-query-the-hardware-uuid-of-a-mac-programmatically-from-a-command-line

I guess we can just use hostname instead, but UUID seems more stable than hostname.

We should probably also put user ID to the hasher.

jandubois: > We should also put host UUID to avoid MAC collision when multiple Mac hosts with bridged mode are connected to the same physical network. (Can be another PR)

I've create #135 for this. Does it make sense to use VDE on Linux too?

> We should probably also put user ID to the hasher.

Why would we need this; the full path to the `lima.yaml` file is already unique for each machine. If 2 users refer to the same config file, they will refer to the same instance. Also for regular operation the filename already includes the user name: `/Users/$USER/.lima/$INSTANCE/lima.yaml`.

AkihiroSuda (approved):
Thanks, merging and I'll open a follow-up PR to cover https://github.com/AkihiroSuda/lima/issues/135

PR #137, opened 2021-08-02
Add validation to network.vde[*].name fields

jandubois:
Not really happy about the `qemuconst` package, but `limayaml` importing `qemu` would create a circular dependency.

### DIFF
--- a/pkg/qemu/qemuconst/qemuconst.go
+++ b/pkg/qemu/qemuconst/qemuconst.go
@@ -1,0 +1,6 @@
+package qemuconst
+
+const (
+	SlirpMACAddress  = "22:11:11:11:11:11"
+	SlirpNICName = "eth0"
+)
--- a/pkg/qemu/qemuconst/qemuconst.go
+++ b/pkg/qemu/qemuconst/qemuconst.go
@@ -1,6 +1,5 @@
 package qemuconst
 
 const (
-	SlirpMACAddress  = "22:11:11:11:11:11"
 	SlirpNICName = "eth0"
 )
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -72,7 +72,8 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML) error {
 		args.Mounts = append(args.Mounts, expanded)
 	}
 
-	args.Networks = append(args.Networks, Network{MACAddress: qemuconst.SlirpMACAddress, Name: qemuconst.SlirpNICName})
+	slirpMACAddress := limayaml.MACAddress(instDir)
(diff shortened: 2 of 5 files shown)


pkg/qemu/qemuconst/qemuconst.go
```diff
@@ -0,0 +1,6 @@
+package qemuconst
+
+const (
+	SlirpMACAddress  = "22:11:11:11:11:11"
+	SlirpNetworkName = "eth0"
```
AkihiroSuda, line 5: I'd call this "SlirpNICName"

pkg/qemu/qemuconst/qemuconst.go
```diff
@@ -0,0 +1,6 @@
+package qemuconst
+
+const (
+	SlirpMACAddress  = "22:11:11:11:11:11"
```
AkihiroSuda, line 4: Maybe we should hash SlirpMACAddress too, but can be another PR.

pkg/qemu/qemuconst/qemuconst.go
jandubois: changed

pkg/qemu/qemuconst/qemuconst.go
jandubois: We could, but does it matter? Isn't it always on a network by itself?

AkihiroSuda (approved):
Thanks

pkg/qemu/qemuconst/qemuconst.go
AkihiroSuda: Some apps may want eth0 MAC to be unique.

jandubois: Ok, I'll add it to this PR.

Is the name `eth0` fine, or should it be something like `slirp` instead?

I think I prefer to always have an `eth0` interface though; it feels more traditional. 😄

AkihiroSuda: eth0 looks good to me

jandubois: And any suggestion about where to put `qemuconst.SlirpNICName`? It feels a little silly to have a separate package just for one constant.

AkihiroSuda: >  It feels a little silly to have a separate package just for one constant.

I think this is fine

jandubois: > Maybe we should hash SlirpMACAddress too

Implemented in 5f4d187.

I'm slightly worried that the SlirpMACAddress is calculated in 2 different places, and that if only one of them is updated later, the addresses would no longer match. Not sure how to avoid this though.

PR #139, opened 2021-08-02
Put machineID to the MAC address hasher + update docs + misc

### LINKED ISSUE #135: Include the machine-id in the hasher to generate unique MAC addresses
Create `osutil.MachineID()` function to return a unique id for the host system.

More info in https://github.com/AkihiroSuda/lima/pull/133#pullrequestreview-719642788

On Linux we can probably use `/var/lib/dbus/machine-id` (fallback to `/etc/machine-id`) instead.

AkihiroSuda:
Fix #135

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,8 +70,8 @@ jobs:
         run: make
       - name: Install
         run: sudo make install
-      - name: Validate examples
-        run: limactl validate ./examples/*.yaml
+      - name: Validate examples (except vmnet.yaml)
+        run: find examples -name '*.yaml' | grep -v 'vmnet.yaml' | xargs limactl validate
       - name: Uninstall
         run: sudo make uninstall
 
@@ -81,9 +81,9 @@ jobs:
     timeout-minutes: 40
     strategy:
       matrix:
-        # GHA macOS is slow and flaky, so we only test "default.yaml" here.
+        # GHA macOS is slow and flaky, so we only test a few YAMLS here.
         # Other yamls are tested on Linux instances of Cirrus.
-        example: [default.yaml]
+        example: [default.yaml, vmnet.yaml]
     steps:
       - uses: actions/setup-go@v2
         with:
@@ -100,6 +100,16 @@ jobs:
         # bash:      required by test-example.sh (OS version of bash is too old)
         # coreutils: required by test-example.sh for the "timeout" command
         run: brew install qemu bash coreutils
+      - name: Install vde_vmnet
+        if: matrix.example == 'vmnet.yaml'
+        env:
+          VDE_VMNET_VERSION: v0.4.0
+        run: |
+          git clone https://github.com/lima-vm/vde_vmnet
+          cd vde_vmnet
+          git checkout $VDE_VMNET_VERSION
+          make
+          sudo make install
       - name: Prepare ssh
         run: |
           if [ -e ~/.ssh/id_rsa ]; then
--- a/README.md
+++ b/README.md
@@ -202,6 +202,8 @@ The current default spec:
   - ["Port forwarding does not work"](#port-forwarding-does-not-work)
   - [stuck on "Waiting for the essential requirement 1 of X: "ssh"](#stuck-on-waiting-for-the-essential-requirement-1-of-x-ssh)
   - ["permission denied" for `limactl cp` command](#permission-denied-for-limactl-cp-command)
+- [Networking](#networking)
+  - ["Cannot access the guest IP 192.168.5.15 from the host"](#cannot-access-the-guest-ip-192168515-from-the-host)
 - ["Hints for debugging other problems?"](#hints-for-debugging-other-problems)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -303,6 +305,15 @@ The `copy` command only works for instances that have been created by lima 0.5.0
 ```console
 < ~/.lima/_config/user.pub limactl shell INSTANCE sh -c 'tee -a ~/.ssh/authorized_keys'
 ```
+
+### Networking
+#### "Cannot access the guest IP 192.168.5.15 from the host"
+
+The default guest IP 192.168.5.15 is not accessible from the host and other guests.
+
+To add another IP address that is accessible from the host and other virtual machines, enable [`vde_vmnet`](https://github.com/lima-vm/vde_vmnet).
+
+See [`./docs/network.md`](./docs/network.md).
 
 ### "Hints for debugging other problems?"
 - Inspect logs:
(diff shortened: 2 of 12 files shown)


AkihiroSuda:
@jandubois 

After merging this one I'm gonna move the repo to https://github.com/lima-vm and release v0.6.0

.github/workflows/test.yml
```diff
@@ -70,8 +70,8 @@ jobs:
         run: make
       - name: Install
         run: sudo make install
-      - name: Validate examples
-        run: limactl validate ./examples/*.yaml
+      - name: Validate examples (except vmnet.yaml)
+        run: find examples -name '*.yaml' | grep -v 'vmnet.yaml' | xargs limactl validate
```
jandubois, line 74: Why does `vmnet.yaml` need to be excluded? It seems to validate fine.

README.md
```diff
@@ -304,6 +306,15 @@ The `copy` command only works for instances that have been created by lima 0.5.0
 < ~/.lima/_config/user.pub limactl shell INSTANCE sh -c 'tee -a ~/.ssh/authorized_keys'
 ```
 
+### Networking
+#### "Cannot access the guest IP 192.168.5.15 from the host"
+
+The default guest IP 192.168.5.15 is not accessible from the host and other guests.
+
+To allow accessing the guest IP from the host and other virtual machines, enable [`vde_vmnet`](https://github.com/lima-vm/vde_vmnet).
```
jandubois, line 314: Maybe reword to make clear that this will add one or more additional routable IPs to the guest. It will still not allow access "the" guest IP.

docs/network.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Guest IP (192.168.5.15)
+
+The guest IP address is typically set to 192.168.5.15.
+
+This IP address is not accessible from the host by design.
+
+Use `vde_vmnet` to allow accessing the guest IP from the host and other guests.
+
+### Host IP (192.168.5.2)
```
jandubois, line 15: Unrelated to this PR, but I wonder if we shouldn't add an entry to `/etc/hosts` to provide a symbolic name for it?

Challenge is to find a good name, as the meaning of "host" is reversed inside the guest. I guess `hostlocalhost` is rather confusing. 😄

docs/network.md
```diff
@@ -0,0 +1,44 @@
...
+
+The loopback addresses of the host is accessible from the guest as 192.168.5.2.
+
+### DNS (192.168.5.3)
+
+The DNS.
+
+## `vde_vmnet` (192.168.105.0/24)
```
jandubois, line 23: Add a note that the IP range is just the default setting and can be configured on the vde_vmnet side?

pkg/limayaml/default.yaml
```diff
@@ -146,3 +102,67 @@ containerd:
...
+  # The instance can get routable IP addresses from the vmnet framework using
+  # https://github.com/lima-vm/vde_vmnet. Both vde_switch and vde_vmnet
+  # daemons must be running before the instance is started. The interface type
+  # (host, shared, or bridged) is configured in vde_vmnet and not lima.
+  vde:
+    # url points to the vde_switch socket directory, optionally with vde:// prefix
+    # - url: "vde:///var/run/vde.ctl"
+    #   # VDE Switch port number (not TCP/UDP port number). Set to 65535 for PTP mode.
+    #   # Default: 0
+    #   port: 0
```
jandubois, line 132: Maybe call it `switchPort` to make it more explicit? Because people may delete the comments in their own config files and then later get confused.

We already use `macAddress` below, so we are not aiming for the shortest possible names.

pkg/limayaml/validate.go
```diff
@@ -160,28 +162,56 @@ func Validate(y LimaYAML) error {
...
+				if fi.Mode()&os.ModeSocket == 0 {
+					return errors.Errorf("field `%s.url` %q is not a directory nor a UNIX socket", field, vdeSwitch)
+				}
+				if vde.Port != 65535 {
+					return errors.Errorf("field `%s.url` points to a PTP (switchless) socket %q, so the port number has to be 65535 (got %d)",
+						field, vdeSwitch, vde.Port)
+				}
+			}
+		} else if runtime.GOOS != "linux" {
+			logrus.Warnf("field %s.url is unlikely to work for %s (unless you managed to port over libvdeplug4 to %s by yourself)",
```
jandubois, line 213: Put `%s.url` in backquotes.

The quoted text feels slightly snarky to me. Maybe "(unless libvdeplug4 has been ported to %s and is installed)" is more neutral?

pkg/osutil/machineid.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+func MachineID() string {
+	machineIDOnce.Do(func() {
+		x, err := machineID()
+		if err == nil && x != "" {
+			machineIDCached = x
+			return
+		}
+		logrus.WithError(err).Debug("failed to get machine ID, falling back to use hostname instead")
```
jandubois, line 29: I feel slightly uneasy about the fallback, and would rather tell the user to fix the cause why the machine id cannot be determined. Because once they fix it, all the hashes will change.

jandubois (commented):
Finished my review

.github/workflows/test.yml
AkihiroSuda: Because it needs VDE socket to be present

pkg/osutil/machineid.go
AkihiroSuda: In a container, dbus is almost always missing, so lacking machine ID is completely fine.

.github/workflows/test.yml
jandubois: > Because it needs VDE socket to be present

Ah yes, I completely forgot that I have the socket locally when I tested this.

docs/network.md
AkihiroSuda: The hostname should be like `host.lima.internal` to follow the `host.docker.internal` / `host.containers.internal` convention.

Not in this PR, though.

docs/network.md
AkihiroSuda: updated

README.md
AkihiroSuda: updated

pkg/limayaml/default.yaml
AkihiroSuda: updated

pkg/limayaml/validate.go
AkihiroSuda: updated

AkihiroSuda:
btw maybe we should rename `url` to `vnl` (Virtual Network Locator) to follow the latest VDE nomenclature

https://github.com/rd235/vdeplug4/commit/089984200f447abb0e825eb45548b781ba1ebccd

jandubois (approved):
LGTM

PR #165, opened 2021-08-20
SSH: Set UserKnownHostsFile to /dev/null

mook-as:
As we are dealing with temporary SSH host keys, we should not pollute the users's known hosts file with ephemeral data.  Additionally, this avoids issues where the user has an invalid known hosts file, as in https://github.com/rancher-sandbox/rancher-desktop/issues/504.

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -155,6 +155,7 @@ func CommonArgs(useDotSSH bool) ([]string, error) {
 
 	args = append(args,
 		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
 		"-o", "NoHostAuthenticationForLocalhost=yes",
 		"-o", "GSSAPIAuthentication=no",
 		"-o", "PreferredAuthentications=publickey",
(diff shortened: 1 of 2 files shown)


pkg/sshutil/sshutil.go
```diff
@@ -155,6 +155,7 @@ func CommonArgs(useDotSSH bool) ([]string, error) {
 
 	args = append(args,
 		"-o", "StrictHostKeyChecking=no",
+		"-o", "UserKnownHostsFile=/dev/null",
```
AkihiroSuda, line 158: Do we now want to remove RemoveKnownHostEntries() in L101?

jandubois: > Do we now want to remove RemoveKnownHostEntries() in L101?

Yes, that function should not be needed anymore if we don't use a known hosts file.

@mook-as can you update the PR  to include removal of `RemoveKnownHostEntries()` (and the call location, of course)?

mook-as: Thanks for catching that! Fixed in 08d4bc3770410158242eb953c1eed38514e15015

AkihiroSuda (approved):
Thanks, LGTM if CI green

jandubois (approved):
LGTM

PR #166, opened 2021-08-22
remove direct dependency on github.com/pkg/errors

AkihiroSuda:
Used https://github.com/AkihiroSuda/go-wrap-to-percent-w for converting `errors.Wrap` to `%w`

### DIFF
--- a/pkg/cidata/template.go
+++ b/pkg/cidata/template.go
@@ -3,7 +3,8 @@ package cidata
 import (
 	"bytes"
 	"embed"
-	_ "embed"
+	"errors"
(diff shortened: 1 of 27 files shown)


pkg/cidata/template.go
```diff
@@ -3,15 +3,15 @@ package cidata
 import (
 	"bytes"
 	"embed"
-	_ "embed"
```
AkihiroSuda, line 6: (This fix is unrelated to pkg/errors)

jandubois (approved):
LGTM

PR #173, opened 2021-08-28
Run `brew update` to make sure we always get the latest dependencies

jandubois:
Also run `brew upgrade`, just so everything is fresh. Collect timings, as we may drop `brew upgrade` again if it takes significant time.


AkihiroSuda (approved):
LGTM, but needs rebase

AkihiroSuda (approved):

PR #174, opened 2021-08-28
qemu-img 6.1.0 no longer supports using -b without -F

jandubois:
> Backing file specified without backing format
> Detected format of qcow2.

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -78,7 +78,7 @@ func EnsureDisk(cfg Config) error {
 	}
 	args := []string{"create", "-f", "qcow2"}
 	if !isBaseDiskISO {
-		args = append(args, "-b", baseDisk)
+		args = append(args, "-F", "qcow2", "-b", baseDisk)
 	}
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

pkg/qemu/qemu.go
```diff
@@ -78,7 +78,7 @@ func EnsureDisk(cfg Config) error {
 	}
 	args := []string{"create", "-f", "qcow2"}
 	if !isBaseDiskISO {
-		args = append(args, "-b", baseDisk)
+		args = append(args, "-F", "qcow2", "-b", baseDisk)
```
AkihiroSuda, line 81: Thx

question: does -F qcow2 work when the base image is non-qcow2 (especially raw)?

jandubois: I don't know, but I suspect it won't. :(

AkihiroSuda: Can we inspect the format?

AkihiroSuda: opened https://github.com/lima-vm/lima/pull/176 for detecting the format

PR #176, opened 2021-08-28
pkg/qemu: detect basedisk format

AkihiroSuda:
Follow-up to #174 `qemu-img 6.1.0 no longer supports using -b without -F`


jandubois (approved):
LGTM

PR #186, opened 2021-09-01
Fix `lima (start|stop|shell)` suggestion to remove instance name

### LINKED ISSUE #185: Wrong hint when using `lima start default` instead of `limactl start default`
When using `lima start [x]...` (instead of `limactl`) the hint adds these arguments behind the correct command:
`lima start default` -> `WARN[0000] Perhaps you meant "limactl start default default"?`
Which looks IMO a bit confusing for new users.

jandubois:
Because the lima wrapper turns `lima start foo` into `limactl shell default start foo` and the instance name should not
be part of the suggestion "Perhaps you meant `limactl start foo`".

Fixes #185


AkihiroSuda (approved):
Thanks

PR #118, opened 2021-07-21
[WIP/DNM] replace reverse SSHFS with Samba

AkihiroSuda:
Sambas are not automatically mounted yet. `mount -t cifs -o credentials=/tmp/credentials //192.168.5.4/lima-0 /mnt/tmp-0`

FIXME: mount takes 25 secs, due to hostname resolution errors
```
[2021/07/21 18:47:12.871531,  3] ../../lib/util/util_net.c:257(interpret_string_addr_internal)
  interpret_string_addr_internal: getaddrinfo failed for name suda-mbp.local (flags 1026) [nodename nor servname provided, or not known]
[2021/07/21 18:47:12.871626,  3] ../../source3/lib/util_sock.c:1026(get_mydnsfullname)
  get_mydnsfullname: getaddrinfo failed for name suda-mbp.local [Unknown error]
```

A workaround is to add `127.0.0.1 localhost suda-mbp.local` to `/etc/hosts` on the host, but that requires sudo.

- - -

TODOs:
- supply the credential to the guest
  - via serial? via ISO?
  - Using an ACPI table is not an option, because ACPI is N/A for qemu-system-aarch64

- auto mount samba mounts

- remove sshfs

- decrease debug level

PR #188, opened 2021-09-02
Replace reverse SSHFS with Samba

### LINKED ISSUE #20: Filesystem Sharing
Hello. I would like to give some ideas and advices about filesystem sharing.

So, since we are using QEMU, lets see which options do we have:
- VirtioFS. Looking very cool, seems to have really good performance, but works only on Linux hosts. It is very optimised for using in virtual machines, it even uses DAX (direct access) for files, so there's no need to copy files over network, they're just in the shared RAM between VM and host.
- VirtFS (9P). I've tried to use it, but it's incredible slow. Really. Using just `git status` in shared directory with middle size project takes at least half a minute. I would rather just place files in VM and access them via some remote file access protocol and use vscode with remote access (sad, but they're proprietary). I think that it is so slow because it is sync. Whenever you read some file, do stat call etc. you have to wait for this operation to end.
- Just sync filesystem state between VM and host.
- Write custom FUSE driver with asynchronous protocol and multithreading support. Theoretically this can be more performant than 9P, but not sure.

### LINKED ISSUE #187: Freezing when trying to map a volume
Using the default configuration, lima/nerdctl/containerd is freezing when I run a container with `-v`. Running the container without mapping a volume works fine, but as soon as I add the `-v` flag, nerdctl freezes when trying to start it.

## To reproduce
* `brew install lima`
* `limactl start` (using default configuration)
* `lima nerdctl run --rm -it -v $PWD:/app alpine`
* At this point, lima freezes entirely - does not respond to ^C, ^Z, or the like. Only way to exit out of it is with the SSH escape (`~.`).

## Full logs
```
╭─ ~/code/lima-testing
╰─❯ cat index.html 
This is a test!

╭─ ~/code/lima-testing
╰─❯ lima                            
kopinsky@lima-default:/Users/kopinsky/code/lima-testing$ cat index.html 
This is a test!
kopinsky@lima-default:/Users/kopinsky/code/lima-testing$ nerdctl run --rm -it alpine
/ # uname -a
Linux ec71a982be17 5.11.0-31-generic #33-Ubuntu SMP Wed Aug 11 13:19:04 UTC 2021 x86_64 Linux
/ # 
kopinsky@lima-default:/Users/kopinsky/code/lima-testing$ nerdctl --debug-full run --rm -it alpine
DEBU[0000] rootless parent main: executing "/usr/bin/nsenter" with [-r/ -w/Users/kopinsky/code/lima-testing --preserve-credentials -m -n -U -t 798 -F nerdctl --debug-full run --rm -it alpine] 
DEBU[0000] found CNI isolation plugin                   
/ # uname -a
Linux f761dc56f016 5.11.0-31-generic #33-Ubuntu SMP Wed Aug 11 13:19:04 UTC 2021 x86_64 Linux
/ # 
kopinsky@lima-default:/Users/kopinsky/code/lima-testing$ nerdctl --debug-full run --rm -it -v $PWD:/app alpine
DEBU[0000] rootless parent main: executing "/usr/bin/nsenter" with [-r/ -w/Users/kopinsky/code/lima-testing --preserve-credentials -m -n -U -t 798 -F nerdctl --debug-full run --rm -it -v /Users/kopinsky/code/lima-testing:/app alpine] 
```

Relevant output from `lima ps auxf`:
```
root         711  0.0  0.1  13288  7104 ?        Ss   21:14   0:00 sshd: /usr/sbin/sshd -D [listener] 0 of 10-100 startups
root        1552  0.0  0.2  14856  9152 ?        Ss   21:14   0:00  \_ sshd: kopinsky [priv]
kopinsky    1679  0.0  0.1  15016  6300 ?        S    21:14   0:00      \_ sshd: kopinsky@pts/0,pts/1
kopinsky    1693  0.0  0.0 301552  3336 ?        Ssl  21:14   0:00          \_ sshfs :/Users/kopinsky /Users/kopinsky -o slave -o ro -o allow_root
kopinsky    1708  0.0  0.0 162148  3116 ?        Ssl  21:14   0:00          \_ sshfs :/tmp/lima /tmp/lima -o slave -o allow_root
kopinsky    1939  0.0  0.1  10460  5392 pts/0    Ss   21:16   0:00          \_ bash --login
kopinsky    2351  0.2  0.5 724624 23612 pts/0    Sl+  21:17   0:00          |   \_ nerdctl --debug-full run --rm -it -v /Users/kopinsky/code/lima-testing:/app alpine
kopinsky    2362  0.0  0.0  11864  3688 pts/1    Rs+  21:19   0:00          \_ ps auxf
```

AkihiroSuda:
🔴 Current blocker: nls_utf8.ko is missing in openSUSE 15.3 https://bugzilla.opensuse.org/show_bug.cgi?id=1190797
🔴 I also have to rewrite this PR to make Samba non-default (https://github.com/lima-vm/lima/pull/188#discussion_r718677030)

- - -
Replace previous PR #118
Fix #20  (`Filesystem sharing`)

See the changes of `docs/internal.md` for the design.

- On macOS hosts, `/usr/local/sbin/samba-dot-org-smbd` is used as the `smbd` binary.
  This binary can be installed with `brew install samba`.
  Apple's version of `/usr/sbin/smbd` cannot be used.
  The binary path can be overridden with `$SMBD` env var.

- smbd is connected to QEMU via smb's stdio.
  The samba address 192.168.5.4:445 is only accessible from the guest, not from the host.

 - When the host's hostname is not present in `/etc/hosts` on the host filesystem, the Lima hostagent launches a mDNS to help looking up the hostname.
    Otherwise starting Samba takes 25 secs with "getaddrinfo failed" error.
    See the `pkg/samba/getaddrinfoworkaround` package for the further information.


AkihiroSuda:
Now this should be almost ready to merge (after passing CI).

cc @jandubois

docs/internal.md
```diff
@@ -105,6 +111,56 @@ The volume label is "cidata", as defined by [cloud-init NoCloud](https://cloudin
...
+
+## FAQs
+### Why use stdio?
+For hiding the Samba port from the host processes, such as malicious WebSocket apps that can
+connect to arbitrary IP including 127.0.0.1.
+
+Run `sudo lsof -i  -P | grep LISTEN` to confirm that no real TCP port on the host is listening for Samba.
+
+(If you carefully read the code of [`slirp_socketpair_with_oob()`](https://gitlab.freedesktop.org/slirp/libslirp/-/blob/v4.6.1/src/misc.c#L80-L133),
+you might notice that it actually bind on `INADDR_ANY`, but the socket is closed before executing `smbd` process.)
```
AkihiroSuda, line 144: (This section needs careful double-checking)

AkihiroSuda: (I'm having second thought on this, we can avoid using `slirp_socketpair_with_oob()`, by launching the smb process by ourselves, and connecting its stdio to a UNIX socket)

pkg/hostagent/hostagent.go
```diff
@@ -144,6 +145,23 @@ func (a *HostAgent) Run(ctx context.Context) error {
...
+			// Attempt to drop supplemental groups, for workarounding Samba issue
+			//  `get_user_groups: failed to get the unix group list` that occurs
+			// when the user belongs to more than 16 groups
+			// https://github.com/Homebrew/legacy-homebrew/issues/5954
+			//
+			// However, this turned out NOT to work, because Samba calls getgrouplist(3)
+			// which talks to opendirectoryd(8) to get the group list.
+			//
+			// https://github.com/samba-team/samba/blob/samba-4.14.7/source3/lib/system_smbd.c#L185
+			NoSetGroups: true,
```
AkihiroSuda, line 161: This is the current blocker 😞

yangyang198703:
@AkihiroSuda , when would we expect this to be release and how much would the performance increase? 

During the evaluation of replacing Docker Desktop using lima-vm, we observed the sshfs performance significantly bring down the site speed of the local development since we mounted the PHP files from Mac OSX to lima, and each web session requiring 300+ PHP files to be read has lead to 30 seconds + time for loading the files. Greatly hope that switching to Samba could address this performance issue.

AkihiroSuda:
>  when would we expect this to be release

A couple of weeks

> how much would the performance increase?

Should be roughly two times faster than sshfs https://github.com/lima-vm/lima/issues/20#issuecomment-869509863

yangyang198703:
> > when would we expect this to be release
> 
> A couple of weeks
> 
> > how much would the performance increase?
> 
> Should be roughly two times faster than sshfs [#20 (comment)](https://github.com/lima-vm/lima/issues/20#issuecomment-869509863)

Thanks for the quick answer! 

I think this performance improvement will be able to convince our technical community in Houzz to switch to use lima-vm in the near future, excited to see this happening soon!

pkg/cidata/cidata.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/lima-vm/lima/pkg/localpathutil"
 	"github.com/lima-vm/lima/pkg/osutil"
-	"github.com/lima-vm/lima/pkg/qemu/const"
+	qemu "github.com/lima-vm/lima/pkg/qemu/const"
```
yangyang198703, line 19: typo?

afbjorklund:
It seems like using "virtfs" will be faster, as soon as the 9p-darwin branch is merged into qemu.

https://github.com/containers/podman/issues/8016#issuecomment-920037754

The _real_ solution is virtio-fs and vsock, but that is further away - especially on darwin (macOS)

https://virtio-fs.gitlab.io/

AkihiroSuda:
Yes, I know, but the virtfs patch is still unmerged in the upstream.
Is anyone trying to resubmit the patch to the upstream recently?

afbjorklund:
>  Is anyone trying to resubmit the patch to the upstream recently?

I have no contact with the qemu team, and they are not using GitHub

afbjorklund:
Using reverse sshfs is somewhat new to me, decided to use "forward" sshfs...
I think `vboxsf` was the only one that actually was used by default, back then.

The main problem was getting someone to support Windows, using "Dokan".
Seems like there are better options there now, like using SSHFS-Win/WinFsp

pkg/cidata/cidata.TEMPLATE.d/boot/15-pkg-cifs.sh
```diff
@@ -0,0 +1,31 @@
...
+	# shellcheck disable=SC2086
+	apt-get install -y -q $pkgs
+elif command -v dnf >/dev/null 2>&1; then
+	dnf install -y cifs-utils
+elif command -v pacman >/dev/null 2>&1; then
+	pacman -Syu --noconfirm cifs-utils
+elif command -v zypper >/dev/null 2>&1; then
+	zypper install -y cifs-utils
+	# openSUSE 15.3 doesn't seem to provide nls_utf8.ko, though /proc/config.gz contains CONFIG_NLS_UTF8=m
+	# https://bugzilla.opensuse.org/show_bug.cgi?id=1190797
```
AkihiroSuda, line 28: @jandubois Any chance to get your colleagues to look into this? 🙏

jandubois: > @jandubois Any chance to get your colleagues to look into this? 🙏

Trying to find somebody, but they are in a different timezone. But I've send some requests...

AkihiroSuda: Thank you!

jandubois: The issue is not openSUSE, but that we are using the JeOS image:

>Because of size constraints, JeOS images use the `kernel-default-base` as the default kernel. Since the kernel does not contain drivers for bare metal, this allows for a smaller footprint.
>If a specific usage scenario requires any of the omitted modules, install the `kernel- default` package to get a full tree of the kernel modules. This is the same kernel used by a regular SUSE Linux Enterprise Server installation.

The problem is that installing `kernel-default` requires a reboot. 😞 

Will see if there are any regular openSUSE images with cloud-init support.

yangyang198703: Would this be a block for this PR, since other distributions can be satisfied? We could first drop the openSUSE support and later add it back, wonder how many ppl are relying on the openSUSE image now.

jandubois: I'm getting the same error with the default Ubuntu image:

```console
root@lima-default:~# modprobe nls_utf8
modprobe: FATAL: Module nls_utf8 not found in directory /lib/modules/5.11.0-31-generic
```

What am I missing?

jandubois: I could get it to work with this provisioning script:

```yaml
provision:
- mode: system
  script: |
    #!/bin/sh
    set -o errexit -o nounset -o xtrace
    if ! modprobe nls_utf8; then
      zypper in -y kernel-default
      reboot
    fi
```

Of course this is a bit wonky, as it reboots too late, so you may already connect to the instance. But I'm sure this could be moved into the boot script as well and would be mostly invisible. Maybe this is good enough?

AkihiroSuda: Thanks, rebooting is okay, but I guess we should make samba non-default then.

jandubois: > Thanks, rebooting is okay, but I guess we should make samba non-default then.

idk, it is a one-time thing during first boot only. So if samba is otherwise much better, I think we could still make it the default.

Was there a way to make the mounting of the file systems faster? It seemed like the first mount took a very long time; the second one was much faster.

AkihiroSuda: > It seemed like the first mount took a very long time; the second one was much faster.

That’s another reason that we should make samba optional 😞

satmandu:
@afbjorklund 
> Minikube bundles a [go 9p](https://github.com/kubernetes/minikube/tree/master/third_party/go9p) server, but it's not very well maintained either

Looks like 9p is [available in](https://www.kraxel.org/blog/2019/06/macos-qemu-guest/) macos Monterey?

```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/share/man/man8/mount_9p.8
/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/share/man/man8/mount_9p.8
/Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/share/man/man8/mount_9p.8
/Library/Developer/CommandLineTools/SDKs/MacOSX11.1.sdk/usr/share/man/man8/mount_9p.8
/Library/Developer/CommandLineTools/SDKs/MacOSX11.3.sdk/usr/share/man/man8/mount_9p.8
/sbin/mount_9p
/usr/share/man/man8/mount_9p.8
```

afbjorklund:
> Looks like 9p is [available in](https://www.kraxel.org/blog/2019/06/macos-qemu-guest/) macos Monterey?

Most likely that is the 9p client, not the 9p server ?

satmandu:
@afbjorklund Here's some discussion of a recently worked on 9p server: https://9fans.topicbox.com/groups/9fans/T5addc17a0e19cd91

afbjorklund:
Not sure that 9p is the best long term solution, just thought that using "virtfs" would be best short term:

https://www.kernel.org/doc/ols/2010/ols2010-pages-109-120.pdf

virtfs-performance

But after that, using "virtiofs" should be better (in terms of performance) - once it is generally available:

https://archive.fosdem.org/2020/schedule/event/vai_virtio_fs/

virtiofs-performance

satmandu:
There is maybe sufficiently working virtio-fs functionality in https://github.com/containers/libkrun for macos as per https://twitter.com/slpnix/status/1351969466847010818 ?

DestroyerXyz:
Isn't the virtio driver in the linux kernel?

AkihiroSuda:
> Isn't the virtio driver in the linux kernel?

Yes, but QEMU lacks it on macOS hosts.

DestroyerXyz:
So you will replace it with smb and then with virtio in the future?

AkihiroSuda:
> So you will replace it with smb and then with virtio in the future?

I was planning to do so, but Samba seems still unstable on macOS (https://bugzilla.samba.org/show_bug.cgi?id=14900), so I'm now looking for switching to virtio-9p-pci, possibly with a patched version of QEMU.
https://github.com/lima-vm/lima/issues/20#issuecomment-1011798955

DestroyerXyz:
So pre-compiled modified binary or self-compiled?
And don't know about Macports and nix but Homebrew will probably allow custom QEMU binaries.

jandubois:
I don't know about Macports or nix either, but even if homebrew doesn't allow embedded binaries, there are always the options of creating a cask instead, or using a tap and maintain the packages ourselves.

Of course the best option would be to get things merged upstream. Is there a specific holdup?

DestroyerXyz:
Also, we can create a separate homebrew formula for the modified qemu which can also help other projects in need of virtio.

AkihiroSuda:
Let's continue the discussion in:
- https://github.com/lima-vm/lima/issues/20

PR #194, opened 2021-09-02
copy: use SSHArgs if copying to a single remote host.

mook-as:
This lets us use `ControlPath` with the connection, so that if a small number of copies run in parallel we can reuse the connection, rather than failing when too many connections are in flight.  Note that this only works if there's a single remote host; if the copy involves multiple hosts (`limactl copy one:/some/path two:/some/path`), the syntax does not allow for separate configuration, so we have to fall back to the old way without a `ControlPath`.

Having too many copies in parallel will still hit any limits imposed by `MaxSessions` (default 10), at which point `scp` falls back to not using the control connection (and may hit `MaxStartups` instead).

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -172,7 +172,7 @@ func SSHArgs(instDir string, useDotSSH bool) ([]string, error) {
 		return nil, err
 	}
 	args = append(args,
-		"-l", u.Username, // guest and host have the same username, but we should specify the username explicitly (#85)
+		"-o", fmt.Sprintf("User=%s", u.Username), // guest and host have the same username, but we should specify the username explicitly (#85)
 		"-o", "ControlMaster=auto",
 		"-o", fmt.Sprintf("ControlPath=\"%s\"", controlSock),
 		"-o", "ControlPersist=5m",
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -64,6 +64,9 @@ func copyAction(clicontext *cli.Context) error {
 
 	sshArgs := []string{}
 	if len(instDirs) == 1 {
+		// Only one (instance) host is involved; we can use the instance-specific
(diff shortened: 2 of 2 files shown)


jandubois (commented):
The code looks fine, except I think it would be useful to have a comment in there that explains why the single instance code path is preferable.

Do I understand correctly that this change will have no effect at all unless you also modify the `MaxSessions` setting as well?

mook-as:
There will be _some_ effect: `MaxSessions` defaults to 10.  But most of the effect will be from bumping that up too.

jandubois:
> There will be _some_ effect: `MaxSessions` defaults to 10.

Your PR description said:

> Having too many copies in parallel will still hit any limits imposed by MaxSessions (default 1),

which made it look like there is no effect (but I think I also remember that `10` is the actual default).

mook-as:
Sorry, that `default 1` was a typo; fixed :)

pkg/sshutil/sshutil.go
```diff
@@ -172,7 +172,7 @@ func SSHArgs(instDir string, useDotSSH bool) ([]string, error) {
 		return nil, err
 	}
 	args = append(args,
-		"-l", u.Username, // guest and host have the same username, but we should specify the username explicitly (#85)
+		"-o", fmt.Sprintf("User=%s", u.Username), // guest and host have the same username, but we should specify the username explicitly (#85)
```
AkihiroSuda, line 175: Any specific reason for this change?

AkihiroSuda (approved):

pkg/sshutil/sshutil.go
mook-as: `scp` doesn't support `-l` (that means `-l limit`, `Limit the used bandwidth` instead).  However, both `ssh` and `scp` does `-o User=xxx` because it's the format from `~/.ssh/config`.

jandubois (approved):

PR #195, opened 2021-09-03
Allow users to define custom environment variables to be loaded on VM start

loganprice:
In many corporate environments proxy environment variables `HTTPS_PROXY,HTTP_PROXY,NO_PROXY` may need to be supplied to allow access to the internet. These changes would allow users too append any additional environment variables they may want at VM start.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/lima.env
+++ b/pkg/cidata/cidata.TEMPLATE.d/lima.env
@@ -15,3 +15,6 @@ LIMA_CIDATA_CONTAINERD_SYSTEM=1
 LIMA_CIDATA_CONTAINERD_SYSTEM=
 {{- end}}
 LIMA_CIDATA_SLIRP_GATEWAY={{ .SlirpGateway }}
+{{- range $key, $val := .Env}}
+{{$key}}={{$val}}
+{{- end}}
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -57,6 +57,8 @@ ssh:
   # Default: true
   loadDotSSHPubKeys: true
 
+
+
 # ===================================================================== #
 # ADVANCED CONFIGURATION
 # ===================================================================== #
@@ -162,6 +164,10 @@ network:
 #     hostIP: "127.0.0.1"
 #     hostPortRange: [1024, 65535]
 #   # Any port still not matched by a rule will not be forwarded (ignored)
+
+# # extra environment variables that will be loaded into the VM at start up
+# env:
+#   KEY: value
 
 # ===================================================================== #
 # END OF TEMPLATE
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -7,20 +7,21 @@ import (
 )
 
 type LimaYAML struct {
-	Arch         Arch          `yaml:"arch,omitempty"`
-	Images       []File        `yaml:"images"` // REQUIRED
-	CPUs         int           `yaml:"cpus,omitempty"`
-	Memory       string        `yaml:"memory,omitempty"` // go-units.RAMInBytes
-	Disk         string        `yaml:"disk,omitempty"`   // go-units.RAMInBytes
-	Mounts       []Mount       `yaml:"mounts,omitempty"`
-	SSH          SSH           `yaml:"ssh,omitempty"` // REQUIRED (FIXME)
-	Firmware     Firmware      `yaml:"firmware,omitempty"`
-	Video        Video         `yaml:"video,omitempty"`
-	Provision    []Provision   `yaml:"provision,omitempty"`
-	Containerd   Containerd    `yaml:"containerd,omitempty"`
-	Probes       []Probe       `yaml:"probes,omitempty"`
-	PortForwards []PortForward `yaml:"portForwards,omitempty"`
-	Network      Network       `yaml:"network,omitempty"`
+	Arch         Arch               `yaml:"arch,omitempty"`
+	Images       []File             `yaml:"images"` // REQUIRED
+	CPUs         int                `yaml:"cpus,omitempty"`
+	Memory       string             `yaml:"memory,omitempty"` // go-units.RAMInBytes
+	Disk         string             `yaml:"disk,omitempty"`   // go-units.RAMInBytes
+	Mounts       []Mount            `yaml:"mounts,omitempty"`
+	SSH          SSH                `yaml:"ssh,omitempty"` // REQUIRED (FIXME)
+	Firmware     Firmware           `yaml:"firmware,omitempty"`
+	Video        Video              `yaml:"video,omitempty"`
+	Provision    []Provision        `yaml:"provision,omitempty"`
+	Containerd   Containerd         `yaml:"containerd,omitempty"`
+	Probes       []Probe            `yaml:"probes,omitempty"`
+	PortForwards []PortForward      `yaml:"portForwards,omitempty"`
+	Network      Network            `yaml:"network,omitempty"`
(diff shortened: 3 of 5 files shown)


pkg/cidata/cidata.TEMPLATE.d/lima.env
```diff
@@ -15,3 +15,6 @@ LIMA_CIDATA_CONTAINERD_SYSTEM=1
 LIMA_CIDATA_CONTAINERD_SYSTEM=
 {{- end}}
 LIMA_CIDATA_SLIRP_GATEWAY={{ .SlirpGateway }}
+{{- range $key, $val := .EnvironmentVariables}}
```
jandubois, line 18: This isn't going to work; the `lima.env` file is only used during `cloud-init`; you still need to copy them into a profile script, e.g. `/etc/profile.d/lima.sh`.

In `lima.env` the variables should be defined like this (similar to how `.Mounts` are defined earlier):

```shell
LIMA_CIDATA_ENVS=2
LIMA_CIDATA_ENVS_0_NAME=http_proxy
LIMA_CIDATA_ENVS_0_VALUE=my.proxy.com:8080
LIMA_CIDATA_ENVS_1_NAME=https_proxy
LIMA_CIDATA_ENVS_1_VALUE=my.proxy.com:8080
```

and then in `pkg/cidata/boot/25-guestagent-base.sh` you need a loop that writes them to `/etc/profile.d/lima.sh`.

pkg/limayaml/limayaml.go
```diff
@@ -7,20 +7,21 @@ import (
...
+	Mounts               []Mount           `yaml:"mounts,omitempty"`
+	SSH                  SSH               `yaml:"ssh,omitempty"` // REQUIRED (FIXME)
+	Firmware             Firmware          `yaml:"firmware,omitempty"`
+	Video                Video             `yaml:"video,omitempty"`
+	Provision            []Provision       `yaml:"provision,omitempty"`
+	Containerd           Containerd        `yaml:"containerd,omitempty"`
+	Probes               []Probe           `yaml:"probes,omitempty"`
+	PortForwards         []PortForward     `yaml:"portForwards,omitempty"`
+	Network              Network           `yaml:"network,omitempty"`
+	EnvironmentVariables map[string]string `yaml:"environmentVariables,omitempty"`
```
AkihiroSuda, line 24: I think we can just call it "Env".
Also, the type should be probably `map[string]*string`

pkg/limayaml/default.yaml
```diff
@@ -57,6 +57,10 @@ ssh:
   # Default: true
   loadDotSSHPubKeys: true
 
+# # extra environment variables that will be loaded into the VM at start up
+# environmentVariables:
+#   KEY: value
```
AkihiroSuda, line 62: I think this should be moved to the `ADVANCED CONFIGURATION` section

pkg/limayaml/default.yaml
```diff
@@ -57,6 +57,10 @@ ssh:
   # Default: true
   loadDotSSHPubKeys: true
 
+# # extra environment variables that will be loaded into the VM at start up
+# environmentVariables:
```
jandubois, line 61: I would prefer the key to be just `env` here. I care less about how we call it inside the Go code.

Hidden agenda: I hope that one day we can support some kind of override mechanism on the commandline, so you can write:

```
limactl start examples/arch.yaml --set env.HTTP_PROXY=my.proxy.com:8080
```

(kind of like the way `helm` allows you to override `values.yaml` from the commandline). And for that I want the name to be short. 😄

jandubois (changes requested):

pkg/cidata/cidata.TEMPLATE.d/lima.env
```diff
@@ -15,3 +15,6 @@ LIMA_CIDATA_CONTAINERD_SYSTEM=1
...
+{{$key}}={{$val}}
+{{- end}}
```
AkihiroSuda, line 20: The env vars here are only used for the init script of Lima.
To apply the env vars globally, `/etc/environment` need to be updated.

The file path may differ across guest distros, so we should have an integration tests. (`hack/test-example.sh`)

pkg/cidata/cidata.TEMPLATE.d/lima.env
AkihiroSuda: > /etc/profile.d/lima.sh

Should the path be `/etc/environment`?

jandubois: > Should the path be `/etc/environment`?

I don't think it really matters. The only advantage I see for a profile script is that you can extend `PATH`:

```
env:
    PATH="/opt/bin:$PATH"
```

But you can always do this with a provisioning script as well.

Another thought: `/etc/environment` is not a POSIX feature but is implemented by PAM. So not sure if that always works. I guess I should try the minimal Alpine image...

loganprice: So I was thinking more during the cloud-init phase. In my local testing I added them directly to lima.env and was about resolve the issues I was having during the apt-get update and apt-get install. Anything after that I figured if a user needed them they could add anything they need during the provision stage. 

But can add the additional logic so they can get added to the profile.

jandubois: > So I was thinking more during the cloud-init phase.

Ah yes, thank you! I forgot that these also need to be in effect during cloud-init.

This could be done by sourcing `/etc/profile.d/lima.sh` in `pkg/cidata/boot/30-install-packages.sh`.

jandubois: > But can add the additional logic so they can get added to the profile.

I would prefer this to be a global conf setting and not just for the cloud-init phase.

But wait for @AkihiroSuda to decide on `/etc/environment` vs. `/etc/profile.d/lima.sh`

AkihiroSuda: Either is fine, as long as we can have CI. The latter seems more friendly to Alpine.

jandubois: Yes, alpine-virt ignores `/etc/environment` because it doesn't use PAM by default.

`limactl shell` has a hardcoded `exec bash --login`, so using `/etc/profile.d` seems appropriate.

jandubois:
@loganprice Can you please make sure you sign your commits; we cannot merge PRs unless they pass the DCO check: https://github.com/lima-vm/lima/pull/195/checks?check_run_id=3503216089

Also maybe squash all the "chore" commits with the base commit while you are at it?

Finally, let me know if you prefer that I do the remaining changes that I asked about (setting the variables in `/etc/profile.d/lima.sh`). In that case we can merge this PR as-is once you have supplied the DCO signature(s).

jandubois (approved):
@AkihiroSuda I'm fine with merging this PR as-is, as it solves the problem of making proxy settings available during cloud-init processing.

I'll create a separate PR to make the variables available via `/etc/profile.d/lima.sh` once this PR is merged.

AkihiroSuda:
Thanks.

Next time please consider cleaning up commit message.

pkg/limayaml/limayaml.go
jandubois: > Also, the type should be probably `map[string]*string`

@AkihiroSuda Why did you ask for the type to be changed? I noticed that in template expansion you cannot pass a `*string` to e.g. `{{ printf "%s=%q" $var $key }}`, you can only interpolate directly, i.e. `{{$key}}` (or at least that's how I remember it when I last tried it; I may have made some other mistake).

Just asking about this for general understanding; we don't actually need to pass the values to `printf`.

Issue #189, opened 2021-09-02
Add support for HTTP Proxy or any generic environment variables


legege:
Starting a VM when behind an HTTP Proxy doesn't work right now. There are various errors from the boot scripts.

However, adding these environment variables to `lima.env` and compiling Lima from source works:

```
...
LIMA_CIDATA_SLIRP_GATEWAY={{ .SlirpGateway }}
http_proxy=proxy_host:3128
https_proxy=proxy_host:3128
no_proxy=.internal.example.com
HTTP_PROXY=proxy_host:3128
HTTPS_PROXY=proxy_host:3128
NO_PROXY=.internal.example.com
```

One way to solve this would be to allow generic environment variables to be specified in the `lima.yaml`. Something like:

```
env:
  http_proxy: "proxy_host:3128"
  https_proxy: "proxy_host:3128"
  no_proxy: ".internal.example.com"
  HTTP_PROXY: "proxy_host:3128"
  HTTPS_PROXY: "proxy_host:3128"
  NO_PROXY: ".internal.example.com"
```

We could as well automatically add these variables to a profile file, like `/etc/profile.d/env.sh`, to make sure the variables are also in the user environment. That would avoid having to create a provision script like this one:

```
provision:
  - mode: system
    script: |
      #!/bin/bash
      echo "http_proxy=proxy_host:3128" > /etc/profile.d/proxy.sh
      echo "https_proxy=proxy_host:3128" >> /etc/profile.d/proxy.sh
      echo "no_proxy=.internal.example.com" >> /etc/profile.d/proxy.sh
      echo "HTTP_PROXY=proxy_host:3128" >> /etc/profile.d/proxy.sh
      echo "HTTPS_PROXY=proxy_host:3128" >> /etc/profile.d/proxy.sh
      echo "NO_PROXY=.internal.example.com" >> /etc/profile.d/proxy.sh
```

Let me know what you think.

AkihiroSuda:
SGTM.

 I think we should import all proxy envs by default.

jandubois:
👍 on explicit support for setting environment variables without having to write a provisioning script.

> I think we should import all proxy envs by default.

Not sure how to figure out the correct settings though. Proxies are defined separately for each network. Here I've defined separate (fake) `http` proxies for `en0` and `en1`:

```console
$ system_profiler SPNetworkDataType -json | jq '.SPNetworkDataType[] | [._name, .interface, .Proxies]'
[
  "Ethernet",
  "en0",
  {
[...]

PR #199, opened 2021-09-04
Pass DNS addresses from en0 interface to guest instance

### LINKED ISSUE #150: [IPv6] DNS failure during install of either 0.5.0 or 0.6.0 results in sshfs not found or installed.
Installation on Big Sur amd64 system....
Logs:
[ha.stderr.log](https://github.com/lima-vm/lima/files/6948365/ha.stderr.log)
[ha.stdout.log](https://github.com/lima-vm/lima/files/6948366/ha.stdout.log)
[serial.log](https://github.com/lima-vm/lima/files/6948367/serial.log)

jandubois:
qemu user-mode networking is forwarding DNS to just a single DNS entry from the host. So if that entry does not resolve
the query, there is no fallback.

Fixes #150 (I hope)

Please review the code updating the DNS entries in the guest. The direct updates to `/etc/resolv.conf` are probably useless because the file will still be regenerated after cloud-init is done. I've included code to configure `systemd-resolved`, and `udhcpc` (used by Alpine). Any other mechanisms needing support?

The `sysprof` code is expected to be used to provide default proxy information as well (see https://github.com/lima-vm/lima/issues/189#issuecomment-911972759)

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/network-config
+++ b/pkg/cidata/cidata.TEMPLATE.d/network-config
@@ -6,4 +6,11 @@ ethernets:
       macaddress: '{{$nw.MACAddress}}'
     dhcp4: true
     set-name: {{$nw.Name}}
+    {{- if and (eq $nw.Name $.SlirpNICName) (gt (len $.DNSAddresses) 0) }}
+    nameservers:
+      addresses:
+      {{- range $ns := $.DNSAddresses }}
+      - {{$ns}}
+      {{- end }}
+    {{- end }}
   {{- end }}
--- a/pkg/osutil/dns_darwin.go
+++ b/pkg/osutil/dns_darwin.go
@@ -1,0 +1,25 @@
+package osutil
+
+import "github.com/lima-vm/lima/pkg/sysprof"
+
+func DNSAddresses() ([]string, error) {
+	nwData, err := sysprof.NetworkData()
+	if err != nil {
+		return nil, err
+	}
+	var addresses []string
+	if len(nwData) > 0 {
+		// Return DNS addresses from en0 interface
+		for _, nw := range nwData {
+			if nw.Interface == "en0" {
+				addresses = nw.DNS.ServerAddresses
+				break
+			}
+		}
+		// In case "en0" is not found, use the addresses of the first interface
+		if len(addresses) == 0 {
+			addresses = nwData[0].DNS.ServerAddresses
+		}
+	}
+	return addresses, nil
+}
--- a/pkg/osutil/dns_others.go
+++ b/pkg/osutil/dns_others.go
@@ -1,0 +1,9 @@
+//go:build !darwin
+// +build !darwin
+
+package osutil
+
+func DNSAddresses() ([]string, error) {
+	// TODO: parse /etc/resolv.conf?
+	return []string{}, nil
+}
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -16,6 +16,7 @@ import (
 	"github.com/lima-vm/lima/pkg/iso9660util"
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/lima-vm/lima/pkg/localpathutil"
+	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/qemu/qemuconst"
 	"github.com/lima-vm/lima/pkg/sshutil"
 	"github.com/lima-vm/lima/pkg/store/filenames"
@@ -51,6 +52,7 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML) error {
 		User:         u.Username,
 		UID:          uid,
 		Containerd:   Containerd{System: *y.Containerd.System, User: *y.Containerd.User},
+		SlirpNICName: qemuconst.SlirpNICName,
 		SlirpGateway: qemuconst.SlirpGateway,
 		Env:          y.Env,
 	}
@@ -78,6 +80,11 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML) error {
 	args.Networks = append(args.Networks, Network{MACAddress: slirpMACAddress, Name: qemuconst.SlirpNICName})
 	for _, vde := range y.Network.VDE {
 		args.Networks = append(args.Networks, Network{MACAddress: vde.MACAddress, Name: vde.Name})
+	}
+
+	args.DNSAddresses, err = osutil.DNSAddresses()
+	if err != nil {
+		return err
 	}
 
 	if err := ValidateTemplateArgs(args); err != nil {
--- a/pkg/cidata/template.go
+++ b/pkg/cidata/template.go
@@ -35,8 +35,10 @@ type TemplateArgs struct {
 	Mounts       []string // abs path, accessible by the User
 	Containerd   Containerd
 	Networks     []Network
+	SlirpNICName string
 	SlirpGateway string
 	Env          map[string]*string
+	DNSAddresses []string
 }
 
 func ValidateTemplateArgs(args TemplateArgs) error {
--- a/pkg/osutil/osutil_others.go
+++ b/pkg/osutil/osutil_others.go
@@ -1,3 +1,4 @@
+//go:build !linux
 // +build !linux
 
 package osutil
--- a/pkg/sysprof/network_darwin.go
+++ b/pkg/sysprof/network_darwin.go
@@ -1,0 +1,14 @@
+package sysprof
+
+type SPNetworkDataType struct {
+	SPNetworkDataType []NetworkDataType `json:"SPNetworkDataType"`
+}
+
+type NetworkDataType struct {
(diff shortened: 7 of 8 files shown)


pkg/cidata/cidata.TEMPLATE.d/lima.env
```diff
@@ -4,6 +4,10 @@ LIMA_CIDATA_MOUNTS={{ len .Mounts }}
 {{- range $i, $val := .Mounts}}
 LIMA_CIDATA_MOUNTS_{{$i}}_MOUNTPOINT={{$val}}
 {{- end}}
+LIMA_CIDATA_DNS={{ len .DNSAddresses }}
+{{- range $i, $val := .DNSAddresses }}
+LIMA_CIDATA_DNS_{{$i}}_ADDRESS={{$val}}
+{{- end}}
```
AkihiroSuda, line 10: Please add these vars to docs/internal. md

pkg/cidata/cidata.TEMPLATE.d/lima.env
```diff
@@ -4,6 +4,10 @@ LIMA_CIDATA_MOUNTS={{ len .Mounts }}
 {{- range $i, $val := .Mounts}}
 LIMA_CIDATA_MOUNTS_{{$i}}_MOUNTPOINT={{$val}}
 {{- end}}
+LIMA_CIDATA_DNS={{ len .DNSAddresses }}
```
AkihiroSuda, line 7: nit: LIMA_CIDATA_DNS_COUNT might be less confusing

pkg/cidata/cidata.TEMPLATE.d/boot/15-setup-dns.sh
```diff
@@ -0,0 +1,29 @@
...
+# NOTE: Busybox sh does not support `for ((i=0;i<$N;i++))` form
+for f in $(seq 0 $((LIMA_CIDATA_DNS - 1))); do
+	dnsaddressvar="LIMA_CIDATA_DNS_${f}_ADDRESS"
+	dnsaddress="$(eval echo \$"$dnsaddressvar")"
+	if [ -z "${DNS}" ]; then
+		DNS="${dnsaddress}"
+	else
+		DNS="${DNS} ${dnsaddress}"
+	fi
+	echo "nameserver ${dnsaddress}" >>/etc/resolv.conf
```
AkihiroSuda, line 14: This should be only on Alpine, I guess

AkihiroSuda: and for other hosts, can we depend on cloud-init? Sed is not robust.

jandubois: > This should be only on Alpine, I guess

It is not needed on Alpine either. `udhcpc` will generate `/etc/resolv.conf` later via script, so any changes we make here are lost. I'm fine removing direct edits to `/etc/resolv.conf` completely, as I think they will likely never have any effect.

> can we depend on cloud-init

Absolutely! Trying to understand all the different ways the resolver can be configured seems to be a nightmare, which is why I asked if we need to support other mechanisms. Happy to delegate this to cloud-init (after checking that it actually works).

pkg/cidata/cidata.TEMPLATE.d/lima.env
jandubois: Sure. I copied the naming scheme from `LIMA_CIDATA_MOUNTS`, so I will append `_COUNT` there as well for consistency.

pkg/osutil/dns_darwin.go
```diff
@@ -0,0 +1,25 @@
+package osutil
+
+import "github.com/lima-vm/lima/pkg/sysprof"
+
+func DNSAddresses() ([]string, error) {
```
AkihiroSuda, line 5: QEMU seems using libresolv to get DNS addresses https://gitlab.freedesktop.org/slirp/libslirp/-/blob/v4.6.1/src/slirp.c#L151

Why isn’t this working for us?

jandubois: Because it then only picks the [first one it finds](https://gitlab.freedesktop.org/slirp/libslirp/-/blob/v4.6.1/src/slirp.c#L159). It forwards DNS to that address instead of adding all the servers to the resolver inside the guest (probably because it isn't clear how it could do the latter).

So if the choosen server is not working, or is an [IPv6 address in a broken IPv6 setup](https://github.com/rancher-sandbox/rancher-desktop/issues/572#issuecomment-911805379), then no names will resolve because there is [no fallback](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=625689#28).

pkg/cidata/cidata.TEMPLATE.d/lima.env
jandubois: The variables are no longer needed; the settings are added to `network-data` for cloud-init instead. That means the Alpine logic will have to be implemented in `lima-init`.

pkg/cidata/cidata.TEMPLATE.d/lima.env
jandubois: Since the `DNS` variables are gone, I'm not going to rename the `MOUNTS` variables in this PR.

jandubois:
Temporarily switched to Draft mode because I think the name servers should only be added to the QEMU slirp interface, and not the the vde interfaces. Will update once more.

jandubois:
Seems to be working fine now:

```console
jan@lima-default:/Users/jan$ sudo cat /mnt/lima-cidata/network-config
version: 2
ethernets:
  eth0:
    match:
      macaddress: '52:55:55:e9:a1:b4'
    dhcp4: true
    set-name: eth0
    nameservers:
      addresses:
      - 192.168.123.53
      - 192.168.123.17
  vde0:
    match:
      macaddress: '52:55:55:be:2a:6a'
    dhcp4: true
    set-name: vde0
    
jan@lima-default:/Users/jan$ systemd-resolve --status
Global
       Protocols: -LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
resolv.conf mode: stub

Link 2 (eth0)
    Current Scopes: DNS
         Protocols: +DefaultRoute +LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
Current DNS Server: 192.168.123.53
       DNS Servers: 192.168.123.53 192.168.123.17 192.168.5.3

Link 3 (vde0)
    Current Scopes: DNS
         Protocols: +DefaultRoute +LLMNR -mDNS -DNSOverTLS DNSSEC=no/unsupported
Current DNS Server: 192.168.105.1
       DNS Servers: 192.168.105.1 fe80::864:900c:7ae3:f83c%22011
        DNS Domain: home
```

pkg/osutil/dns_others.go
```diff
@@ -0,0 +1,8 @@
+// +build !darwin
```
AkihiroSuda, line 1: Could you gofmt with Go 1.17

AkihiroSuda:
LGTM after gofmt, thanks

pkg/osutil/dns_others.go
jandubois: > Could you gofmt with Go 1.17

Sure. That also updated `pkg/osutil/osutil_others.go`

AkihiroSuda (approved):
Thanks

pkg/cidata/cidata.TEMPLATE.d/network-config
```diff
@@ -6,4 +6,11 @@ ethernets:
       macaddress: '{{$nw.MACAddress}}'
     dhcp4: true
     set-name: {{$nw.Name}}
+    {{- if eq $nw.Name $.SlirpNICName }}
```
AkihiroSuda, line 9: && DNSAddress is not empty

AkihiroSuda:
ArchLinux CI is failing consistently after retrying several times

pkg/cidata/cidata.TEMPLATE.d/network-config
jandubois: I'll change it, but I don't think it should make any difference. Setting a key to `nil` or not having the key at all should have the same effect. But having the check means we won't have to worry if there is a bug in dealing with this in cloud-init...

jandubois:
> ArchLinux CI is failing consistently after retrying several times

Any idea what might be causing it? I can't repro it locally, even when I pass an empty `Addresses` list to the instance.

pkg/cidata/cidata.TEMPLATE.d/network-config
jandubois: [Checking len(DNSAddresses) now](https://github.com/lima-vm/lima/compare/e23912b405ceb80280360fd070db7b4109b7682c..fd4346f725153d7557264038bfccea451550d31f)

jandubois:
> ArchLinux CI is failing consistently after retrying several times

Worked first time now, but Fedora had to be restarted. I don't think this is related to the `len($.DNSAddresses)` check, but anyways, all green now. 😄

AkihiroSuda (approved):
Thanks

PR #200, opened 2021-09-04
Make env vars persistent 

loganprice:
In relation to https://github.com/lima-vm/lima/pull/195

Changed env vars to follow the same style of mounts

Close if this is not needed


pkg/limayaml/default.yaml
```diff
@@ -167,7 +167,8 @@ network:
 
 # # extra environment variables that will be loaded into the VM at start up
 # env:
-#   KEY: value
+#   - key: key1
+#     value: value
```
AkihiroSuda, line 171: I’d prefer to keep the previous YAML syntax

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -10,6 +10,10 @@ update_fuse_conf() {
 	fi
 }
 
+if [ "${LIMA_CIDATA_ENVS}" -gt 0 ]; then
+	source /etc/profile.d/lima.sh
```
AkihiroSuda, line 14: Isn’t this automatically sourced?

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -13,6 +13,20 @@ done
 # Install or update the guestagent binary
 install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent /usr/local/bin/lima-guestagent
 
+# Set up Environment Variable to be loaded via profile
+if [[ ! -e /etc/profile.d/lima.sh ]]; then
+    touch /etc/profile.d/lima.sh
+	echo "#! /bin/sh -" > /etc/profile.d/lima.sh
```
AkihiroSuda, line 19: Extra space after ‘!’

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -13,6 +13,20 @@ done
 # Install or update the guestagent binary
 install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent /usr/local/bin/lima-guestagent
 
+# Set up Environment Variable to be loaded via profile
+if [[ ! -e /etc/profile.d/lima.sh ]]; then
+    touch /etc/profile.d/lima.sh
```
AkihiroSuda, line 18: Why?

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -13,6 +13,20 @@ done
...
+	echo "#! /bin/sh -" > /etc/profile.d/lima.sh
+fi
+
+for f in $(seq 0 $((LIMA_CIDATA_ENVS - 1))); do
+	keyvar="LIMA_CIDATA_ENVS_${f}_KEY"
+	valuevar="LIMA_CIDATA_ENVS_${f}_VALUE"
+    key="$(eval echo \$"$keyvar")"
+    value="$(eval echo \$"$valuevar")"
+    echo "export "$key"="$value"" >> test.txt
```
AkihiroSuda, line 27: File name seems wrong

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
jandubois: There is no need for a hashbang line; the files are sourced by `/etc/profile`, not loaded as scripts:

```
if [ -d /etc/profile.d ]; then
  for i in /etc/profile.d/*.sh; do
    if [ -r $i ]; then
      . $i
    fi
  done
  unset i
fi
```

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
loganprice: yup little sleep and copy and paste error

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
jandubois: I'm not sure now. I think `/etc/profile` is only sourced for interactive shells. Which makes me rethink now if `/etc/profile.d` is really the right place to set these variables. I guess needs more research...

pkg/limayaml/default.yaml
jandubois: > I’d prefer to keep the previous YAML syntax

Yes, me too. The list form only makes sense when you can legitimately have multiple entries with the same key. But the environment is a map/hash, not a slice/array.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
AkihiroSuda: If not automatically sourced, this needs to be sourced in every boot script, not just 30-install-packages.sh

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
loganprice: it's not needed will remove

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
jandubois: Note that the file is only created in `25-guestagent-base.sh`, so it can only be sourced in that script itself, and any later ones (currently `30-install-packages.sh` and `40-install-containerd.sh`).

AkihiroSuda: Other scripts should source too, on the next boot IMHO.

Or maybe create env sh in 00- script

loganprice: Would it make sense to keep whats currently there rename it too `cloudIntEnv` and add an additional section to the yaml for env vars that would be added to the profile for use during runtime.

jandubois: > Would it make sense to keep whats currently there rename it too `cloudIntEnv` and add an additional section to the yaml for env vars that would be added to the profile for use during runtime.

I don't understand what that means.

> Or maybe create env sh in 00- script

That makes sense to me. At first I thought it should happen after `05-persistent-data-volume.sh`, but I realized it doesn't matter.

Somebody should do some testing on when `/etc/profile` is sourced, and when it isn't. I can do that, but not until tomorrow...

loganprice: What I meant by that. There are two different use cases for users needing environment variables:

1) during the cloud-init 
2) during the actual running during the VM. 

There might be times where you dont need the same env vars set for cloud-init and during the actual runtime of the VM. 

Also all the files in `${LIMA_CIDATA_MNT}"/boot/` are each run in their own sub-shell branching from boot.sh. A 00-env.sh script creating a profile  will not be loaded by subsequent scripts sub-shell executions without being sourced.

So the idea is to have a configuration item that loads and sets up any env vars needs during the cloud-init section (the previous PR) and rename that in the yaml as `cloudInitEnv` and create another configuration item for env vars that will be added to the profile.

AkihiroSuda: > There might be times where you dont need the same env vars set for cloud-init and during the actual runtime of the VM.

Do you have actual examples?

AkihiroSuda: I still think we should just have `env` that is applied universally to both the boot scripts and the user shell.

If the user wants to override the env for the user shell, the user should just modify `.bash_profile` in the provision script.

jandubois: > Also all the files in `${LIMA_CIDATA_MNT}"/boot/` are each run in their own sub-shell branching from boot.sh.

Thank you, I think this actually points out the proper place to implement this: at the top of `boot.sh`. It can then source the file once it has written it, and all the `boot/*.sh` scripts will automatically inherit the variables.

> I still think we should just have `env` that is applied universally to both the boot scripts and the user shell.

I agree. Having separate settings makes it more complex for the users that want to set both to the same set of variables and values, which I expect to be the vast majority of users.

loganprice: > 
> Do you have actual examples?

For most of our users we try to encourage that they dont the use the proxy but instead pull packages/images/etc from our artifact store but with how the the cloud-init is structured we need to proxy to install the initial packages before we can override things during the provision stage.

jandubois: Thank you for the example! You can still achieve this by replacing `/etc/profile.d/lima.sh` in a provisioning script, which is little more effort than having 2 separate blocks of environment variables.

So I would say having a shared `env` block causes a little bit of extra effort in this case, but would cause a lot of effort for users that want the same settings in both situations. And duplicating settings here also makes it harder to maintain them, as you will have to apply future changes to 2 places to keep them in sync.

So I'm still in favour of using `env` globally.

loganprice: Add all of lima.env to lima.sh, changed permissions, and sourced it. If you are both ok with this solution. I will create a new PR as I messed up my branch rebasing and would be easier to just create a new branch and open a new PR.

jandubois: > Add all of lima.env to lima.sh, changed permissions, and sourced it.

I think adding the `env.*` variables to `lima.env` is redundant, as you are going to source them again anyways. And I don't want the `LIMA_CIDATA_*` variables in the user environment.

So instead of adding the `env.*` variables to `lima.env`, I would move this to a separate template for `lima.sh` (with added `export` keyword):

```
export {{$key}}={{$val}}
{{- end}}
``` 

Then your change to `boot.sh` becomes something like:

```
cp "${LIMA_CIDATA_MNT}"/lima.sh /etc/profile.d/
chmod 644 /etc/profile.d/lima.sh
. /etc/profile.d/lima.sh
```

Note I changed your permissions from `775` (why not `755`?) to `644`. 

> If you are both ok with this solution. I will create a new PR as I messed up my branch rebasing and would be easier to just create a new branch and open a new PR.

Is is fine with me, but you could also try to reset your current branch:

```console
$ git fetch --all
$ git checkout persistant_env_vars
# Assuming your upstream remote is called `remote`
$ git reset --hard upstream/master
# If you wanted to salvage your old commit, but I think you should just create a new one
# git cherry-pick 299a0159
$ git push -f
```

jandubois:
Superseded by #207

PR #202, opened 2021-09-04
CI: bump up macos-10.15 to macos-11

AkihiroSuda:


.github/workflows/test.yml
```diff
@@ -55,7 +55,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-10.15, ubuntu-20.04]
+        os: [macos-11, ubuntu-20.04]
```
jandubois, line 58: Any reason not to test on both? Given that macOS is the primary deployment target.

AkihiroSuda: Updated to support both, but if it makes CI even slower, we should consider removing 10.15.

jandubois: Will they not run in parallel?

AkihiroSuda: Seems sometimes yes, sometimes no

jandubois: There are limits on the number of concurrent jobs per org, but I don't think we should be hitting these limits: https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits

AkihiroSuda: Our test are still very flaky, and almost impossible to make 4 jobs green. And we can’t restart jobs one by one.

So I guess we should just drop 10.15 CI.

jandubois: > So I guess we should just drop 10.15 CI.

I guess we have to then; I thought it was mostly the Linux builds failing though, but maybe I'm misremembering.

AkihiroSuda:
Retried several time, but CI can't be greeen.

Seems we have to deflake our tests (and maybe rewrite in Go) before merging this PR...

jandubois (approved):

PR #205, opened 2021-09-04
yaml: mark "Env" experimental for now

AkihiroSuda:
The syntax and the behavior of this field is still under discussion (PR #200), so this field is treated as experimental in the next upcoming release (v0.6.3).

Probably we can move this out of experimental in v0.6.4.


AkihiroSuda:
cc @loganprice @jandubois

jandubois (approved):

PR #210, opened 2021-09-05
examples: add {docker,podman,singularity}.yaml

AkihiroSuda:
Docker, Podman, Singularity, or whatever can be used instead of the default containerd/nerdctl .

LXD can be used too, of course, but there is no `lxd.yaml`, because LXD is included in the default Ubuntu image.

Example usage:
```
$ limactl start ./docker.yaml
$ limactl shell docker docker run -it --rm -v $HOME:$HOME alpine
```

```
$ limactl start ./podman.yaml
$ limactl shell podman podman run -it --rm -v $HOME:$HOME:ro docker.io/library/alpine
```

```
$ limactl start ./singularity.yaml
$ limactl shell singularity singularity run -u -B $HOME:$HOME docker://alpine
```

- - -

`docker` CLI on macOS can connect to `dockerd` inside the guest VM as well: `DOCKER_HOST=ssh://127.0.0.1:60006 docker run -it --rm -v $HOME:$HOME alpine` .

This requires `NoHostAuthenticationForLocalhost yes` in `~/.ssh/config`. (Alternatively you can just login to `ssh -p 60006 127.0.0.1` to initialize `~/.ssh/known_hosts`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Lima is expected to be used on macOS hosts, but can be used on Linux hosts as we
 
 ✅ Automatic port forwarding
 
-✅ Built-in support for [containerd](https://containerd.io)
+✅ Built-in support for [containerd](https://containerd.io) ([Other container engines can be used too](./examples))
 
 ✅ Intel on Intel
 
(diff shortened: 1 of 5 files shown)


csantanapr:
podman on macos doesn't support volumes `-v` so this solution would be perfect.

jandubois:
The `DOCKER_HOST=...` examples only work when rebasing on top of #220. Sorry about the breakage! 😞 

I suspect there are various ways this could also be broken by `.ssh/config` settings. I think we should support forwarding sockets in addition to ports from the host agent; that way it could even be linked to the default socket location. See e.g. https://github.com/abiosoft/colima/blob/main/colima#L151

I've tried all the sample commands, and they all work for me.

Have not yet found time to review the example files themselves.

jandubois (approved):

PR #214, opened 2021-09-06
Use "lima" username if the local user is not a valid Linux name

### LINKED ISSUE #191: Username with '@' causes fatal error in `limactl start`
Executing `limactl start` with a username containing an "@" will trigger following error

```
DEBU[0000] interpreting argument "default" as an instance name "default" 
FATA[0000] identifier "xxx@xxxx.com" must match ^[A-Za-z0-9]+(?:[._-](?:[A-Za-z0-9]+))*$: invalid argument 
```

jandubois:
At least some companies set local users via directory services and choose the email address as the user name. The '@' character is not valid in Linux usernames.

Fixes #191

### DIFF
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -5,9 +5,9 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"os/user"
 	"strings"
 
+	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/sshutil"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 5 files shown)


jandubois:
For testing I've forced the username to be an email address like this:

```diff
+                       cache.u.Username = "user@example.com"
+                       if err := identifiers.Validate(cache.u.Username); err != nil {
```

This is the output when starting an instance:

```console
$ l start default
WARN[0000] local user "user@example.com" is not a valid Linux user name: identifier "user@example.com" must match ^[A-Za-z0-9]+(?:[._-](?:[A-Za-z0-9]+))*$: invalid argument; using "lima" username instead
[...]
INFO[0023] READY. Run `lima` to open the shell.

$ lima id
uid=501(lima) gid=1000(lima) groups=1000(lima)
```

AkihiroSuda (approved):
Thanks

PR #217, opened 2021-09-07
Add manage_resolv_conf to cloud-init

jandubois:
This has no effect on instances managed by systemd, but is necessary on e.g. Alpine.


pkg/cidata/cidata.TEMPLATE.d/user-data
```diff
@@ -30,3 +30,14 @@ write_files:
    owner: root:root
    path: /var/lib/cloud/scripts/per-boot/00-lima.boot.sh
    permissions: '0755'
+
+{{- if .DNSAddresses }}
+manage_resolv_conf: true
```
AkihiroSuda, line 35: > This has no effect on instances managed by systemd, but is necessary on e.g. Alpine.

Could you add this comment here

AkihiroSuda (approved):
Thanks

PR #218, opened 2021-09-07
Provide explicit DNS override

jandubois:
In some situations (e.g. VPN) the user may need to explicitly set the DNS addresses; even the fallback to the "en0" settings
does not work.

This cannot be done in a provisioning script because they run after the boot scripts, but working DNS is required to install
additional dependencies.


AkihiroSuda (approved):
Thanks

Issue #104, opened 2021-07-01
FATA[0607] did not receive an event with the "running" status


ndeloof:
follow-up https://github.com/AkihiroSuda/lima/issues/18

```console
$ brew info libslirp
libslirp: stable 4.6.1 (bottled)

$ limactl start default
? Creating an instance "default" Proceed with the default configuration
INFO[0001] Downloading "https://github.com/containerd/nerdctl/releases/download/v0.9.0/nerdctl-full-0.9.0-linux-amd64.tar.gz" 
INFO[0001] Using cache "/Users/nicolas/Library/Caches/lima/download/by-url-sha256/4c048fab91ba88527ad58a6b4dc4c3bfa35e828b8147565456b76d362ec2e6cb/data" 
INFO[0004] Attempting to download the image from "~/Downloads/hirsute-server-cloudimg-amd64.img" 
INFO[0004] Attempting to download the image from "https://cloud-images.ubuntu.com/hirsute/current/hirsute-server-cloudimg-amd64.img" 
INFO[0006] Using cache "/Users/nicolas/Library/Caches/lima/download/by-url-sha256/e1fed960ebd29619676c7ab7535bc83f7fb2ad71739edb6fde4e17bce0b61a47/data" 
INFO[0006] [hostagent] Starting QEMU (hint: to watch the boot progress, see "/Users/nicolas/.lima/default/serial.log") 
INFO[0006] SSH Local Port: 60022                        
INFO[0007] [hostagent] Waiting for the essential requirement 1 of 4: "ssh" 
...
INFO[0578] [hostagent] Waiting for the essential requirement 1 of 4: "ssh" 
INFO[0588] [hostagent] Waiting for the essential requirement 1 of 4: "ssh" 
INFO[0598] [hostagent] Waiting for the essential requirement 1 of 4: "ssh" 
FATA[0607] did not receive an event with the "running" status 
```

[serial.log](https://github.com/AkihiroSuda/lima/files/6749252/serial.log)

AkihiroSuda:
https://github.com/AkihiroSuda/lima/blob/86d6a938c1595aff1f352ff5c3344958311494b7/pkg/limayaml/default.yaml#L54

Could you try `ssh.loadDotSSHPubKeys=false`? (requires Lima v0.5.0)

ndeloof:
created default.yaml file with updated `  loadDotSSHPubKeys: false`
```
$ limactl start default.yaml
...
[...]

PR #219, opened 2021-09-07
Ignore public keys that are not in ssh format

jandubois:
It seems that some ~/.ssh/*.pub files are public SSL keys, so add some minimal validation before using them as authorized_keys.

Ref https://github.com/lima-vm/lima/issues/104#issuecomment-914030614

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -90,7 +90,11 @@ func DefaultPubKeys(loadDotSSH bool) ([]PubKey, error) {
 		}
 		entry, err := readPublicKey(f)
 		if err == nil {
-			res = append(res, entry)
+			if strings.ContainsRune(entry.Content, '\n') || !strings.HasPrefix(entry.Content, "ssh-") {
+				logrus.Warnf("public key %q doesn't seem to be in ssh format", entry.Filename)
+			} else {
+				res = append(res, entry)
+			}
 		} else if !errors.Is(err, os.ErrNotExist) {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/sshutil/sshutil.go
```diff
@@ -90,7 +90,11 @@ func DefaultPubKeys(loadDotSSH bool) ([]PubKey, error) {
 		}
 		entry, err := readPublicKey(f)
 		if err == nil {
-			res = append(res, entry)
+			if strings.ContainsRune(entry.Content, '\n') || !strings.HasPrefix(entry.Content, "ssh-") {
```
AkihiroSuda, line 93: Containing \n perhaps doesn’t matter, I guess

jandubois: I think so too, but it doesn't hurt, I think.

Leading/trailing whitespace is already stripped, so if there is still a newline in the file, then it has multiple lines with content, which we don't want (I think). E.g. this protects us against somebody doing a `cp ~/.ssh/authorized_keys ~/.ssh/mykeys.pub`, where each line is a valid public key in ssh format.

Anyways, do you want me to remove it? I would prefer to keep it, at least as long as `lima-init` is written in shell and awk. 😄

AkihiroSuda (approved):

pkg/sshutil/sshutil.go
AkihiroSuda: We can keep it. We may change it later if we get a request to do so.

asdfghjkl403622:
still getting this issue

jandubois:
@asdfghjkl403622 We would need to see the log files to check this further, but I wonder if this isn't related to your all-numeric username, so maybe test again once #243 has been merged?

PR #224, opened 2021-09-08
qemu: use `-cpu host` by default

AkihiroSuda:
`-cpu host` seemed to cause kernel panic with QEMU 5.2 on Intel Mac, but seems to work well with QEMU 6.1 now

The behavior can be rolled back by specifying an env var like `QEMU_SYSTEM_X86_64=qemu-system-x86_64 -cpu Haswell-v4` or `QEMU_SYSTEM_AARCH64=qemu-system-aarch64 -cpu cortex-a72` .


jandubois (approved):
LGTM

PR #230, opened 2021-09-09
[WIP] feat(#131): support ssh.localPort autoassign free port

### LINKED ISSUE #131: Assign `ssh.localPort` automatically to an available port
The port may change when the instance is stopped and restarted because the original port may have been taken by a different application (or lima instance) in between.

Question: where should the port be recorded? Do we load `lima.yaml`, modify the port, and store the file back in YAML format? This would drop any comments from the file, but I think it would be the right thing to do.

Long term I would like to be able to run e.g. `lima restart default --memory 12GB` and have it modify the config file for me automatically, so there needs to be a machine-writable version of the config eventually anyways.

ysicing:
Fixed #131 
support ssh.localPort autoassign free port

Signed-off-by: ysicing <i@ysicing.me>

### DIFF
--- a/pkg/osutil/port_test.go
+++ b/pkg/osutil/port_test.go
@@ -14,7 +14,7 @@ func TestCheckOrGetFreePort(t *testing.T) {
 
 	testCases := []testCase{
 		{
-			// 默认macOS 53端口被占用
+			// By default, macOS 53  are already used
 			port:       53,
 			expectedOK: false,
 		},
--- a/pkg/osutil/port_test.go
+++ b/pkg/osutil/port_test.go
@@ -1,0 +1,30 @@
+package osutil
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestCheckOrGetFreePort(t *testing.T) {
+	type testCase struct {
+		port       int
+		expectedOK bool
+	}
+
+	testCases := []testCase{
+		{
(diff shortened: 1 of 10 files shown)


pkg/osutil/port_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+func TestCheckOrGetFreePort(t *testing.T) {
+	type testCase struct {
+		port       int
+		expectedOK bool
+	}
+
+	testCases := []testCase{
+		{
+			// 默认macOS 53端口被占用
```
AkihiroSuda, line 17: 请用英语

pkg/store/instance.go
```diff
@@ -63,7 +63,11 @@ func Inspect(instName string) (*Instance, error) {
 	}
 	inst.Dir = instDir
 	inst.Arch = y.Arch
-	inst.SSHLocalPort = y.SSH.LocalPort
+	if *y.SSH.AutoPort {
+		inst.SSHLocalPort = osutil.CheckOrGetFreePort(y.SSH.LocalPort)
```
AkihiroSuda, line 67: We have to return the actual port in use, not a new port.

We should let hostagent speak REST API over an UNIX socket to provide `GET /v1/info` endpoint, which would contain port information. (See the guest agent code)

AkihiroSuda:
> Signed-off-by: ysicin

Please use real name for signing

PR #239, opened 2021-09-13
Implement vde_vmnet management

### LINKED ISSUE #175: Integrated VDE networking support
I think Lima should manage the VDE networking setup for the user. I would like to make it work like this:

## Lima Network Configuration

Managed networks are declared in `$LIMA_HOME/_config/networks.yaml`. Lima will provide a default configuration:

```yaml
# Paths to vde executables. Because vde_vmnet is invoked via sudo it should be
# installed where only root can modify/replace it. This means also none of the
# parent directories should be writable by the user.
#
# The var_run directory also must not be writable by the user because it will
# include the vde_vmnet pid files. Those will be terminated via sudo, so replacing
# the pid files would allow killing of arbitrary privileged processes.
paths:
  vde_switch: /opt/vde/bin/vde_switch
  vde_vmnet: /opt/vde/bin/vde_vmnet
  var_run: /var/run/lima

group: staff

networks:
  shared:
    mode: shared
    gateway: 192.168.105.1
    dhcp-end: 129.168.105.254
    netmask: 255.255.255.0
  bridged:
    mode: bridged
    interface: en0
    # bridged mode doesn't has a gateway; dhcp is managed by outside network
  host:
    mode: host
    gateway: 192.168.106.1
    dhcp-end: 129.168.106.254
    netmask: 255.255.255.0
```

Users can modify existing networks (e.g. change the dhcp range), or define additional ones.

## Instance Network Configuration

The managed networks can be referenced from an instance's `lima.yaml` using a custom `lima://` VNL scheme:

```
networks:
  vde:
  - vnl: lima://shared
    name: nat
  - vnl: lima://host
```

The default name for networks with a `lima://` VNL will be the network name, not `vdeN`. So the config above will create a `nat` and a `host` interface inside the instance.

Whenever Lima starts an instance, it will check if all referenced networks are already running, and if not, start both `vde_switch` and `vde_vmnet` daemons for that network. Sockets and pid files will be stored under `/var/run/lima`. Log files will be stored under `$LIMA_HOME/_vde`. (Or should we put them into `/var/run/lima` as well? Maybe needs another setting under `paths`.)

When an instance is stopped, Lima will check if there are networks that are no longer being used, and shut them down as well.

## Managing root privileges

`vde_vmnet` has to be run as root, so Lima will start and stop it using `sudo`. This will ask the user for their admin password unless they have enabled password-less `sudo`.

A helper command for more locked-down systems will be implemented to allow Lima to manage networks without asking for the password every time a network is started or stopped:

```console
$ limactl sudoers | sudo tee /etc/sudoers.d/lima
```

This will write the specific commands needed to start and stop the daemons to `/etc/sudoers.d/lima`, e.g.

```
Cmnd_Alias SHARED_SWITCH_START = /opt/vde/bin/vde_switch --sock=/var/run/lima/shared.ctl \
                                 --pidfile=/var/run/lima/shared_switch.pid --group=staff --dirmode=0770
Cmnd_Alias SHARED_SWITCH_STOP  = /usr/bin/pkill -F /var/run/lima/shared_switch.pid

Cmnd_Alias SHARED_VMNET_START  = /opt/vde/bin/vde_vmnet --vmnet-mode=shared /var/run/lima/shared.ctl
Cmnd_Alias SHARED_VMNET_STOP   = /usr/bin/pkill -F /var/run/lima/shared_vmnet.pid

%staff ALL=(root:root)    NOPASSWD:NOSETENV: /bin/mkdir -m 755 -p /var/run/lima
%staff ALL=(daemon:staff) NOPASSWD:NOSETENV: SHARED_SWITCH_START, SHARED_SWITCH_STOP
%staff ALL=(root:root)    NOPASSWD:NOSETENV: SHARED_VMNET_START,  SHARED_VMNET_STOP
```

(Real settings will include `gateway`, `dhcp-end`, `netmask`, etc too. Also `vde_vmnet` doesn't yet support writing a pid file.)

Question: do we really need to run `vde_switch` using the `daemon` user? We run all of `qemu` under the actual user account; is there anything in the switch that makes it more risky than other utilities, so it needs to be segregated?

Every time the `network.yaml` file is edited, the sudoers file must be regenerated.

jandubois:
Lima uses ~/.lima/_config/networks.yaml to define managed networks for instances. By default "shared", "bridged", and "host" are defined. They can be referenced from `lima.yaml` via e.g. `lima://shared`.

The networks will be automatically started by lime as long as any instance referencing them are running, and once the last instance using a network is stopped, the network is brought down as well.

A new command `limactl sudoers` can be used to write a sudoers file to allow lima to start/stop the daemons via `sudo`. Lima does not support prompting for a sudo password; either the sudoers file must
be maintained, or the user must have password-less sudo configured.

Every time the `networks.yaml` configuration is changed, the sudoers file must be regenerated to match.

Resolves #175

### DIFF
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -16,8 +16,9 @@ func newSudoersCommand() *cobra.Command {
 		Args:  cobra.MaximumNArgs(1),
 		RunE:  sudoersAction,
 	}
+	configFile, _ := networks.ConfigFile()
 	sudoersCommand.Flags().Bool("check", false,
-		"check that the sudoers file is up-to-date with $LIMA_HOME/_config/networks.yaml")
+		fmt.Sprintf("check that the sudoers file is up-to-date with %q", configFile))
 	return sudoersCommand
 }
 
@@ -30,39 +31,44 @@ func sudoersAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	if check {
-		var file string
-		switch len(args) {
-		case 0:
-			config, err := networks.Config()
-			if err != nil {
-				return err
-			}
-			file = config.Paths.Sudoers
-			if file == "" {
-				return errors.New("no sudoers file defined in ~/.lima/_config/networks.yaml")
-			}
-		case 1:
-			file = args[0]
-		default:
-			return errors.New("can check only a single sudoers file")
-		}
-		config, err := networks.Config()
-		if err != nil {
-			return err
-		}
-		if err := config.Validate(); err != nil {
-			return err
-		}
-		if err := config.VerifySudoAccess(file); err != nil {
-			return err
-		}
-		fmt.Printf("%q is up-to-date (or sudo doesn't require a password)\n", file)
-		return nil
+		return verifySudoAccess(args)
 	}
 	sudoers, err := networks.Sudoers()
 	if err != nil {
 		return err
 	}
 	fmt.Print(sudoers)
+	return nil
+}
+
+func verifySudoAccess(args []string) error {
+	var file string
+	switch len(args) {
+	case 0:
+		config, err := networks.Config()
+		if err != nil {
+			return err
+		}
+		file = config.Paths.Sudoers
+		if file == "" {
+			configFile, _ := networks.ConfigFile()
+			return fmt.Errorf("no sudoers file defined in %q", configFile)
+		}
+	case 1:
+		file = args[0]
+	default:
+		return errors.New("can check only a single sudoers file")
+	}
+	config, err := networks.Config()
+	if err != nil {
+		return err
+	}
+	if err := config.Validate(); err != nil {
+		return err
+	}
+	if err := config.VerifySudoAccess(file); err != nil {
+		return err
+	}
+	fmt.Printf("%q is up-to-date (or sudo doesn't require a password)\n", file)
 	return nil
 }
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -3,16 +3,18 @@ package main
 import (
 	"errors"
 	"fmt"
+	"runtime"
+
 	"github.com/lima-vm/lima/pkg/networks"
 	"github.com/spf13/cobra"
 )
 
 func newSudoersCommand() *cobra.Command {
 	sudoersCommand := &cobra.Command{
-		Use:               "sudoers [SUDOERSFILE]",
-		Short:             "Generate /etc/sudoers.d/lima file.",
-		Args:              cobra.MaximumNArgs(1),
-		RunE:              sudoersAction,
+		Use:   "sudoers [SUDOERSFILE]",
+		Short: "Generate /etc/sudoers.d/lima file.",
+		Args:  cobra.MaximumNArgs(1),
+		RunE:  sudoersAction,
 	}
 	sudoersCommand.Flags().Bool("check", false,
 		"check that the sudoers file is up-to-date with $LIMA_HOME/_config/networks.yaml")
@@ -20,11 +22,14 @@ func newSudoersCommand() *cobra.Command {
 }
 
 func sudoersAction(cmd *cobra.Command, args []string) error {
+	if runtime.GOOS != "darwin" {
+		return errors.New("sudoers command is only supported on macOS right now")
+	}
 	check, err := cmd.Flags().GetBool("check")
 	if err != nil {
 		return err
 	}
-	if check	{
+	if check {
 		var file string
 		switch len(args) {
 		case 0:
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -1,0 +1,56 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"github.com/lima-vm/lima/pkg/networks"
+	"github.com/spf13/cobra"
+)
+
+func newSudoersCommand() *cobra.Command {
+	sudoersCommand := &cobra.Command{
+		Use:               "sudoers [SUDOERSFILE]",
+		Short:             "Generate /etc/sudoers.d/lima file.",
+		Args:              cobra.MaximumNArgs(1),
+		RunE:              sudoersAction,
+	}
+	sudoersCommand.Flags().Bool("check", false,
+		"check that the sudoers file is up-to-date with $LIMA_HOME/_config/networks.yaml")
+	return sudoersCommand
+}
+
+func sudoersAction(cmd *cobra.Command, args []string) error {
+	check, err := cmd.Flags().GetBool("check")
+	if err != nil {
+		return err
+	}
+	if check	{
+		var file string
+		switch len(args) {
+		case 0:
+			config, err := networks.Config()
+			if err != nil {
+				return err
+			}
+			file = config.Paths.Sudoers
+			if file == "" {
+				return errors.New("no sudoers file defined in ~/.lima/_config/networks.yaml")
+			}
+		case 1:
+			file = args[0]
+		default:
+			return errors.New("can check only a single sudoers file")
+		}
+		if err := networks.CheckSudoers(file); err != nil {
+			return err
+		}
+		fmt.Printf("%q is up-to-date\n", file)
+		return nil
+	}
+	sudoers, err := networks.Sudoers()
+	if err != nil {
+		return err
+	}
+	fmt.Print(sudoers)
+	return nil
+}
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -46,10 +46,17 @@ func sudoersAction(cmd *cobra.Command, args []string) error {
 		default:
 			return errors.New("can check only a single sudoers file")
 		}
-		if err := networks.CheckSudoers(file); err != nil {
+		config, err := networks.Config()
+		if err != nil {
 			return err
 		}
-		fmt.Printf("%q is up-to-date\n", file)
+		if err := config.Validate(); err != nil {
+			return err
+		}
+		if err := config.VerifySudoAccess(file); err != nil {
+			return err
+		}
+		fmt.Printf("%q is up-to-date (or sudo doesn't require a password)\n", file)
 		return nil
 	}
 	sudoers, err := networks.Sudoers()
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -42,13 +42,16 @@ func sudoersAction(cmd *cobra.Command, args []string) error {
 }
 
 func verifySudoAccess(args []string) error {
+	config, err := networks.Config()
+	if err != nil {
+		return err
+	}
+	if err := config.Validate(); err != nil {
+		return err
+	}
 	var file string
 	switch len(args) {
 	case 0:
-		config, err := networks.Config()
-		if err != nil {
-			return err
-		}
 		file = config.Paths.Sudoers
 		if file == "" {
 			configFile, _ := networks.ConfigFile()
@@ -58,13 +61,6 @@ func verifySudoAccess(args []string) error {
 		file = args[0]
 	default:
 		return errors.New("can check only a single sudoers file")
-	}
-	config, err := networks.Config()
-	if err != nil {
-		return err
-	}
-	if err := config.Validate(); err != nil {
-		return err
 	}
 	if err := config.VerifySudoAccess(file); err != nil {
 		return err
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -2,6 +2,7 @@ package limayaml
 
 import (
 	"fmt"
+	"github.com/lima-vm/lima/pkg/networks"
 	"net"
 	"os"
 	"path/filepath"
@@ -188,7 +189,13 @@ func validateNetwork(y LimaYAML) error {
 			if nw.SwitchPort != 0 {
 				return fmt.Errorf("field `%s.switchPort` cannot be used with field `%s.lima`", field, field)
 			}
-			// TODO: validate network name (we can't use networks and store packages right now)
+			config, err := networks.Config()
+			if err != nil {
+				return err
+			}
+			if config.Check(nw.Lima) != nil {
+				return fmt.Errorf("field `%s.lima` references network %q which is not defined in networks.yaml", field, nw.Lima)
+			}
 		} else {
 			if nw.VNL == "" {
 				return fmt.Errorf("field `%s.lima` or field `%s.vnl` must be set", field, field)
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -18,7 +18,7 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func Validate(y LimaYAML) error {
+func Validate(y LimaYAML, warn bool) error {
 	switch y.Arch {
 	case X8664, AARCH64:
 	default:
@@ -161,17 +161,18 @@ func Validate(y LimaYAML) error {
 		// processed sequentially and the first matching rule for a guest port determines forwarding behavior.
 	}
 
-	if err := validateNetwork(y); err != nil {
+	if err := validateNetwork(y, warn); err != nil {
 		return err
 	}
-
 	return nil
 }
 
-func validateNetwork(y LimaYAML) error {
+func validateNetwork(y LimaYAML, warn bool) error {
 	if len(y.Network.VDEDeprecated) > 0 {
 		if y.Network.migrated {
-			logrus.Warnf("field `network.VDE` is deprecated; please use `networks` instead")
+			if warn {
+				logrus.Warnf("field `network.VDE` is deprecated; please use `networks` instead")
+			}
 		} else {
 			return fmt.Errorf("you cannot use deprecated field `network.VDE` together with replacement field `networks`")
 		}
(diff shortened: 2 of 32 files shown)


jandubois:
@AkihiroSuda Could you please do a first-pass review?  The PR is fully functional, but I want to add a few things still:

* support for network metrics, so the guest OS will prefer using the `vde` devices over `eth0`.
* some validation for `networks.yaml`
* update documentation

You set up your sudoers file with:

```console
$ limactl sudoers | sudo tee /etc/sudoers.d/lima
```

Sample network declaration in `lima.yaml`:

```yaml
network:
  vde:
  - vnl: lima://shared
```

pkg/limayaml/validate.go
```diff
@@ -177,7 +177,14 @@ func validateNetwork(yNetwork Network) error {
 		// The field is called VDE.VNL in anticipation of QEMU upgrading VDE2 to VDEplug4,
 		// but right now the only valid value on macOS is a path to the vde_switch socket directory,
 		// optionally with vde:// prefix.
-		if !strings.Contains(vde.VNL, "://") || strings.HasPrefix(vde.VNL, "vde://") {
+		// TODO: use networks.LimaSchema after solving circular dependency
+		if strings.HasPrefix(vde.VNL, "lima://") {
```
AkihiroSuda, line 181: Sorry, I still think that we shouldn't contaminate the VNL namespace.
`lima://` is reserved for a potential VDE4 plugin that would be called "lima".

https://github.com/lima-vm/lima/issues/175#issuecomment-907589748

pkg/networks/commands.go
```diff
@@ -0,0 +1,84 @@
+package networks
+
+import (
+	"fmt"
+	"github.com/lima-vm/lima/pkg/store"
+)
+
+const (
+	Switch = "switch"
```
AkihiroSuda, line 9: For performance reason, we may just ditch the switch and use PTP mode everywhere: https://github.com/lima-vm/vde_vmnet/issues/26#issuecomment-917392360

That may also beneficial for simplicity.

pkg/networks/commands.go
```diff
@@ -0,0 +1,84 @@
...
+
+func (config *NetworksConfig) StartCmd(name, daemon string) string {
+	var cmd string
+	switch daemon {
+	case Switch:
+		cmd = fmt.Sprintf("%s --pidfile=%s --sock=%s --group=%s --dirmode=0770 --nostdin",
+			config.Paths.VDESwitch, config.PIDFile(name, Switch), config.VDESock(name), config.Group)
+	case VMNet:
+		nw := config.Networks[name]
+		cmd = fmt.Sprintf("%s --pidfile=%s --vde-group=%s --vmnet-mode=%s",
```
AkihiroSuda, line 68: Can we error out when the `vde_vmnet` binary (or its DLLs) is modifiable by non-root?

cmd/limactl/sudoers.go
```diff
@@ -0,0 +1,54 @@
...
+	Flags: []cli.Flag{
+		&cli.BoolFlag{
+			Name:  "check",
+			Usage: "check that the sudoers file is up-to-date with ~/.lima/_config/networks.yaml",
+		},
+	},
+	Action: sudoersAction,
+}
+
+func sudoersAction(clicontext *cli.Context) error {
```
AkihiroSuda, line 23: error out when GOOS != "darwin"

pkg/networks/commands.go
jandubois: Yes, I intended to do this as part of `some validation for networks.yaml`. Not sure about the dylibs though; I don't want to shell out to `otool` to look at the import paths, and then figure out how to resolve `@executable_path/...` references etc. Let me know how you would want to determine the library dependencies!

I was planning to make sure all `config.Path` entries don't have any whitespace, and the files and all parent etc. directories are only writable by root.

AkihiroSuda: The otool stuff can be another PR in future.

pkg/limayaml/validate.go
jandubois: Ok, how do you want to implement it. As a separate list under `network`, e.g.

```yaml
network:
  lima:
  - shared
  - host
```

pkg/networks/commands.go
jandubois: I would add support for PTP mode in a later PR; I believe it can be added as just another network, albeit one that can only be used by one instance at a time.

And e.g. a host-only network doesn't work in PTP mode because you can't connect multiple instances to it. Maybe it doesn't make sense to consider this, but I was thinking about being able have one instance on a host-only network, and a second instance running as a proxy server/jumpbox, to test that an application works correctly in a fully proxied environment.

pkg/limayaml/validate.go
AkihiroSuda: SGTM

jandubois: It will have to be:

```yaml
network:
  lima:
  - network: bridged
    macAddress: 11:22:33:44:55:66
    name: lima0
  - network: host
```

etc. It is unfortunate that we use `vde.name` for the interface name. I wish it was `interface`, so we could use `lima.name` and `lima.interface` instead of `lima.network` and `lima.name`, but I guess it is too late to change this now.

Any strong feelings about the field names?

AkihiroSuda: I don't have a strong opinion.

If we want to change `name` to `interface`, we can safely do that by deprecating `name` with a warning message.

jandubois: This turns out to be really painful to implement, as we would have to duplicate so much code in validation. And the code would still not be independent as we would still have to verify that there are no duplicate interface names between the `VDE` and the `lima` lists.

So if we can still change the `network` layout, then I would suggest going with:

```yaml
network:
- vnl: /var/run/vde.ctl
  interface: vde0
- lima: shared
  interface: lima0
```

Either `vnl` or `lima` would be required, and `switchPort` would only be valid when you use `vnl`.

The problem is that we can't really do a proper deprecation message, as `network` would turn from a `map` to a `slice`.

I still think reserving the `lima://` scheme for our internal use is the most elegant solution, as I think it will be rather unlikely that there ever will be a VDE4 plugin called `lima`. But if we can change the `network` layout, then I'm fine with that too.

jandubois: Ok, one more thought: we could call the new structure `networks`, and deprecate `network`.

jandubois: > we could call the new structure `networks`, and deprecate `network`.

I've implemented this in https://github.com/lima-vm/lima/pull/239/commits/03e90a2 (separate commit for review; will squash later).

It throws an error when you use both the new and the deprecated fields:

```
$ l start alpine.yaml
? Creating an instance "alpine" Proceed with the default configuration
FATA[0001] the YAML is invalid, saved the buffer as "lima.REJECTED.yaml": you cannot use deprecated field `network.VDE` together with replacement field `networks`
```

And will show a deprecation warning when the file uses the old field:

```
$ l start alpine.yaml
? Creating an instance "alpine" Proceed with the default configuration
WARN[0001] field `network.VDE` is deprecated; please use `networks` instead
WARN[0001] field `network.VDE` is deprecated; please use `networks` instead
WARN[0001] field `network.VDE` is deprecated; please use `networks` instead
WARN[0001] field `network.VDE` is deprecated; please use `networks` instead
WARN[0001] field `network.VDE` is deprecated; please use `networks` instead
WARN[0001] public key "/Users/jan/.ssh/foo.pub" doesn't seem to be in ssh format
INFO[0001] Attempting to download the image from "https://github.com/lima-vm/alpine-lima/releases/download/v0.1.4/alpine-lima-std-3.13.5-x86_64.iso"
...
```

The warning is show multiple times because the file is validated several times. I'm somewhat surprised it happens 5 times, but I think it is unavoidable to do it multiple times, unless we cache all the instance data returned by `store.Inspect`.

Haven't done the other changes (network validation, metrics, help), but let me know if you think this is now an acceptable way to declare the networks for an instance!

This is now a working minimal network config:

```yaml
networks:
- lima: shared
```

And I changed the default interface names from `vde0` etc. to `lima0` etc because the use of vde is kind of an implementation details now.

cmd/limactl/sudoers.go
jandubois: > error out when GOOS != "darwin"

Done in 97d3a2b9256ab5503ab8176a1741deea48cfb6f8

pkg/networks/commands.go
jandubois: > Can we error out when the `vde_vmnet` binary (or its DLLs) is modifiable by non-root?

Done in b6bcc80ba862b348839f62842bee9dd019550dec

(otool logic to come in a later PR)

cmd/limactl/sudoers.go
```diff
@@ -0,0 +1,66 @@
...
+		var file string
+		switch clicontext.NArg() {
+		case 0:
+			config, err := networks.Config()
+			if err != nil {
+				return err
+			}
+			file = config.Paths.Sudoers
+			if file == "" {
+				return errors.New("no sudoers file defined in ~/.lima/_config/networks.yaml")
```
AkihiroSuda, line 39: This message would be confusing when LIMA_HOME is set to a non-default path

pkg/networks/commands.go
jandubois: > (otool logic to come in a later PR)

I found that `/usr/bin/otool` is just a shim that tries to locate `XCode.app`, and if it can't find it presents a dialog offering to install the "commandline developer tools".

So it is going to be tricky to figure out if it is even safe to call `otool`. And what do we do when it is not available? Fall back on checking the "known filenames"?

AkihiroSuda: Maybe we can parse macho directly (in another PR)

AkihiroSuda: This should work https://pkg.go.dev/debug/macho#File.ImportedLibraries

AkihiroSuda:
What are the remaining tasks to move this out of draft?

jandubois:
> What are the remaining tasks to move this out of draft?

Things that I think *must* be done as part of this PR are:

* improve `networks.Validate`:
  *  currently allows all paths to be writable by `daemon`; only `varRun` should be writable
  * the lowest level parent directory of `varRun` that already exists *must* be writable by `daemon`, so the `varRun` directory can be created
* update docs
  * document `$LIMA_HOME/_config/networks.yaml`
  * document `$LIMA_HOME/_networks`

One thing I would like to still address in the PR:

* `lima.yaml` is validated 3-5 times, which shows any warnings just as often. Either reduce the number of calls, or cache the results.

Things that can be done in another PR:

* check dylib dependencies of `vde_*` binaries to be only root-writable
* set interface metrics so that the vde interfaces will be prioritized over slirp for outgoing connections
* add support for PTP networks

Things that must be a separate issue/PR:

* managed network support for Linux (using `tap`).

jandubois:
> * `lima.yaml` is validated 3-5 times, which shows any warnings just as often. Either reduce the number of calls, or cache the results.

Haven't found a non-convoluted way of doing this, so I added a `warn` parameter to `limayaml.Validate` to only issue warnings when called by directly from the `start` command. Validation is quick, so it doesn't hurt to call it multiple times.

I'm going to move the PR out of draft now; I think everything else can wait for follow-up PRs.

@AkihiroSuda Do you want me to squash everything to a single commit, or do you want to keep anything separate (e.g. the re-arranging of packages to prevent cyclical import dependencies). I do not want to re-arrange the order of commits though, I would rather squash everything into a single mega-commit.

pkg/networks/config.go
```diff
@@ -0,0 +1,88 @@
...
+		if cache.err != nil {
+			return
+		}
+		_, cache.err = os.Stat(configFile)
+		if cache.err != nil {
+			if !errors.Is(cache.err, os.ErrNotExist) {
+				return
+			}
+			configDir := filepath.Dir(configFile)
+			cache.err = os.MkdirAll(configDir, 0700)
```
jandubois, line 48: Not sure why I picked `0700`; it probably should be `0755`.

pkg/qemu/qemu.go
```diff
@@ -257,33 +258,49 @@ func Cmdline(cfg Config) (string, []string, error) {
 
 	// Network
 	args = append(args, "-netdev", fmt.Sprintf("user,id=net0,net=%s,dhcpstart=%s,hostfwd=tcp:127.0.0.1:%d-:22",
-		qemuconst.SlirpNetwork, qemuconst.SlirpIPAddress, y.SSH.LocalPort))
+		qemu.SlirpNetwork, qemu.SlirpIPAddress, y.SSH.LocalPort))
 	args = append(args, "-device", "virtio-net-pci,netdev=net0,mac="+limayaml.MACAddress(cfg.InstanceDir))
-	if len(y.Network.VDE) > 0 && !strings.Contains(string(features.NetdevHelp), "vde") {
+	usingVDE := false
+	for _, nw := range y.Networks {
+		if nw.VNL != "" {
```
jandubois, line 265: Oops, that was a thinko, the managed networks also require `netdev "vde"` support.

pkg/networks/sudoers.go
```diff
@@ -0,0 +1,93 @@
...
+		sb.WriteString(fmt.Sprintf("# Manage %q network daemons\n", name))
+		for _, daemon := range []string{Switch, VMNet} {
+			sb.WriteRune('\n')
+			sb.WriteString(fmt.Sprintf("%%%s ALL=(%s:%s) NOPASSWD:NOSETENV: \\\n",
+				config.Group, config.DaemonUser(daemon), config.DaemonGroup(daemon)))
+			sb.WriteString(fmt.Sprintf("    %s, \\\n", config.StartCmd(name, daemon)))
+			sb.WriteString(fmt.Sprintf("    %s\n", config.StopCmd(name, daemon)))
+		}
+	}
+	return sb.String(), nil
```
AkihiroSuda, line 41: We should have unit tests to validate the generated strings.
Can be another PR, though.

pkg/networks/validate.go
```diff
@@ -0,0 +1,107 @@
...
+			return findBaseDirectory(filepath.Dir(path))
+		}
+	}
+	return path
+}
+
+const (
+	RootUID   = 0
+	WheelGID  = 0
+	DaemonGID = 1
```
AkihiroSuda, line 54: These should be moved to `*_darwin.go`

pkg/limayaml/validate.go
```diff
@@ -12,12 +12,13 @@ import (
...
 	"github.com/lima-vm/lima/pkg/localpathutil"
+	"github.com/lima-vm/lima/pkg/networks"
 	"github.com/lima-vm/lima/pkg/osutil"
-	"github.com/lima-vm/lima/pkg/qemu/qemuconst"
+	"github.com/lima-vm/lima/pkg/qemu/const"
 	"github.com/sirupsen/logrus"
 )
 
-func Validate(y LimaYAML) error {
+func Validate(y LimaYAML, warn bool) error {
```
AkihiroSuda, line 21: Maybe we should make this like `Validate(y LimaYAML)(warns []error, err error)

AkihiroSuda:
> Do you want me to squash everything to a single commit, or do you want to keep anything separate (e.g. the re-arranging of packages to prevent cyclical import dependencies). I do not want to re-arrange the order of commits though, I would rather squash everything into a single mega-commit.

Either is fine to me

pkg/networks/validate.go
jandubois: > These should be moved to `*_darwin.go`

Just those constants, or the whole code?

The logic is not really darwin specific; it is just that we only call it on darwin.

Even the constants are kind of universal; except that "wheel" should be called "root" on Linux. I'm fine with using `RootGID`. So if anything we should have platform specific constants for the string values (e.g. `RootGroup = "wheel"` on darwin).

AkihiroSuda: > Just those constants, or the whole code?

Either is fine.

At least `DaemonGID=1` is specific to Darwin.
On Linux, the daemon GID is often 2.

AkihiroSuda:
Could you update `examples/vmnet.yaml` to the new syntax to test this on CI?

pkg/limayaml/validate.go
jandubois: > Maybe we should make this like `Validate(y LimaYAML)(warns []error, err error)`

We could, but I'm not keen on it: there are over 50 `return` statements that would all need to deal with it:

```
$ grep return pkg/limayaml/validate.go | wc -l
      54
```

Just for the benefit of 3 possible warnings:

```
$ grep 'logrus\.' pkg/limayaml/validate.go
				logrus.Warnf("field `network.VDE` is deprecated; please use `networks` instead")
					logrus.WithError(err).Debugf("field `%s.vnl` %q failed stat", field, vdeSwitch)
					logrus.Warnf("field `%s.vnl` is unlikely to work for %s (unless libvdeplug4 has been ported to %s and is installed)",
```

And the warnings don't even come from `Validate` directly, but from `validateNetworks`.

I'll do it if you really think it makes a difference, but I think the code is clearer the way it currently is.

pkg/networks/validate.go
jandubois: I haven't split the code into platform-specific pieces, but made it look all user and group names / id dynamically: https://github.com/lima-vm/lima/pull/239/commits/c96f16b802ae8cf3566761717ca423cbbc01d6ef

pkg/networks/sudoers.go
jandubois: > We should have unit tests to validate the generated strings.

Added in https://github.com/lima-vm/lima/pull/239/commits/42fcd35d4ad0bee35253f089d728d20f7a9f42d4

AkihiroSuda (approved):
Thanks

PR #241, opened 2021-09-14
Use numeric uid/gid with chown

jandubois:
Because chown does not work if the name includes a dot, like "firstname.lastname".

See also https://github.com/lima-vm/alpine-lima/pull/13

Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/605

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
+
 set -eux
 
 # Create mount points
@@ -7,7 +8,8 @@ for f in $(seq 0 $((LIMA_CIDATA_MOUNTS - 1))); do
 	mountpointvar="LIMA_CIDATA_MOUNTS_${f}_MOUNTPOINT"
 	mountpoint="$(eval echo \$"$mountpointvar")"
 	mkdir -p "${mountpoint}"
-	chown "${LIMA_CIDATA_USER}" "${mountpoint}"
+	gid=$(id -g "${LIMA_CIDATA_USER}")
+	chown "${LIMA_CIDATA_UID}:${gid}" "${mountpoint}"
 done
 
 # Install or update the guestagent binary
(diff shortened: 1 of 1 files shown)


jandubois:
I first missed this, but the issue affects all guest OSes, not just Alpine.

jandubois:
> I first missed this, but the issue affects all guest OSes, not just Alpine.

Hmm, testing on Ubuntu shows that its `chown` handles dots in usernames fine.

coreutils documentation states:

>Some older scripts may still use ‘.’ in place of the ‘:’ separator. POSIX 1003.1-2001 (see Standards conformance) does not require support for that, but for backward compatibility GNU chown supports ‘.’ so long as no ambiguity results. New scripts should avoid the use of ‘.’ because it is not portable, and because it has undesirable results if the entire owner‘.’group happens to identify a user whose name contains ‘.’.

So using numeric uid/gid seem safer on all OSes.

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -7,7 +10,8 @@ for f in $(seq 0 $((LIMA_CIDATA_MOUNTS - 1))); do
 	mountpointvar="LIMA_CIDATA_MOUNTS_${f}_MOUNTPOINT"
 	mountpoint="$(eval echo \$"$mountpointvar")"
 	mkdir -p "${mountpoint}"
-	chown "${LIMA_CIDATA_USER}" "${mountpoint}"
+	LIMA_CIDATA_GID=$(id -g "${LIMA_CIDATA_USER}")
```
AkihiroSuda, line 13: var name should be just “gid” because this is not really from cidata

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -7,7 +8,8 @@ for f in $(seq 0 $((LIMA_CIDATA_MOUNTS - 1))); do
 	mountpointvar="LIMA_CIDATA_MOUNTS_${f}_MOUNTPOINT"
 	mountpoint="$(eval echo \$"$mountpointvar")"
 	mkdir -p "${mountpoint}"
-	chown "${LIMA_CIDATA_USER}" "${mountpoint}"
+	GID=$(id -g "${LIMA_CIDATA_USER}")
```
AkihiroSuda, line 11: lower chars should be used for local vars consistently

AkihiroSuda (commented):
LGTM but a single tiny nit, thanks

AkihiroSuda (approved):

PR #242, opened 2021-09-14
Enable sudo nerdctl run to expose ports to localhost

### LINKED ISSUE #140: port forwarding does not work for `sudo nerdctl` (i.e. rootful CNI portmap)
```console
$ limactl --version
limactl version 0.5.2-19-g2180618

$ cat default.yaml 
images:
  - location: "https://cloud-images.ubuntu.com/hirsute/current/hirsute-server-cloudimg-amd64.img"
    arch: "x86_64"
ssh:
  localPort: 60022
containerd:
  system: true
  user: true

$ limactl start default.yaml

$ lima nerdctl run -d -p 127.0.0.1:8080:80 nginx:alpine

$ curl localhost:8080
...
<title>Welcome to nginx!</title>
...

$ lima sudo nerdctl run -d -p 127.0.0.1:8081:80 nginx:alpine

$ curl localhost:8081
curl: (7) Failed to connect to localhost port 8081: Connection refused
```

Port forwarding does not seem working for `sudo nerdctl`, i.e. rootful CNI portmap, with iptables.
This is because iptables ports do not appear in `/proc/net/tcp`, `/proc/net/tcp6`.

Note that rootless ports are not affected by this issue because RootlessKit (or slirp4netns, depending on the configuration) manages the ports in userspace.

mattfarina:
A couple changes were made to make this possible including:
- Adding iptables package to read the ports exposed by portmap
  CNI plugin. This was done so that 1 call to iptables is all
  that is needed on each run.
- Move the lima-guestagent to run as root. This is needed for
  permission to use iptables.

Note, this implementation only includes ipv4 support. Does not work for ipv6.

Closes #140 

Note, this is my first pass. Happy to make changes as needed.

### DIFF
--- a/cmd/lima-guestagent/daemon_linux.go
+++ b/cmd/lima-guestagent/daemon_linux.go
@@ -5,7 +5,6 @@ import (
 	"net"
 	"net/http"
 	"os"
-	"path/filepath"
 	"time"
 
 	"github.com/gorilla/mux"
@@ -21,28 +20,18 @@ func newDaemonCommand() *cobra.Command {
 		Short: "run the daemon",
 		RunE:  daemonAction,
 	}
-	daemonCommand.Flags().String("socket", socketDefaultValue(), "the unix socket to listen on")
 	daemonCommand.Flags().Duration("tick", 3*time.Second, "tick for polling events")
 	return daemonCommand
 }
 
 func daemonAction(cmd *cobra.Command, args []string) error {
-	socket, err := cmd.Flags().GetString("socket")
-	if err != nil {
-		return err
-	}
-	if socket == "" {
-		return errors.New("socket must be specified")
-	}
+	socket := "/run/lima-guestagent.sock"
 	tick, err := cmd.Flags().GetDuration("tick")
 	if err != nil {
 		return err
 	}
 	if tick == 0 {
 		return errors.New("tick must be specified")
-	}
-	if os.Geteuid() == 0 {
-		return errors.New("must not run as the root")
 	}
 	logrus.Infof("event tick: %v", tick)
 
@@ -69,14 +58,9 @@ func daemonAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	if err := os.Chmod(socket, 0777); err != nil {
+		return err
+	}
 	logrus.Infof("serving the guest agent on %q", socket)
 	return srv.Serve(l)
-}
-
-func socketDefaultValue() string {
-	if xrd := os.Getenv("XDG_RUNTIME_DIR"); xrd != "" {
-		return filepath.Join(xrd, "lima-guestagent.sock")
-	}
-	logrus.Warn("$XDG_RUNTIME_DIR is not set, cannot determine the socket name")
-	return ""
 }
--- a/cmd/lima-guestagent/daemon_linux.go
+++ b/cmd/lima-guestagent/daemon_linux.go
@@ -33,6 +33,9 @@ func daemonAction(cmd *cobra.Command, args []string) error {
 	if tick == 0 {
 		return errors.New("tick must be specified")
 	}
+	if os.Geteuid() != 0 {
+		return errors.New("must run as the root")
+	}
 	logrus.Infof("event tick: %v", tick)
 
 	newTicker := func() (<-chan time.Time, func()) {
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
@@ -13,7 +13,6 @@ for f in $(seq 0 $((LIMA_CIDATA_MOUNTS - 1))); do
 done
 
 # Install or update the guestagent binary
-rm -f /usr/local/bin/lima-guestagent
 install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent /usr/local/bin/lima-guestagent
 
 # Launch the guestagent service
@@ -36,5 +35,8 @@ EOF
 	rc-update add lima-guestagent default
 	rc-service lima-guestagent start
 else
+	# Remove legacy systemd service
+	rm -f "/home/${LIMA_CIDATA_USER}.linux/.config/systemd/user/lima-guestagent.service"
+
 	sudo lima-guestagent install-systemd
 fi
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
@@ -13,6 +13,7 @@ for f in $(seq 0 $((LIMA_CIDATA_MOUNTS - 1))); do
 done
 
 # Install or update the guestagent binary
+rm -f /usr/local/bin/lima-guestagent
 install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent /usr/local/bin/lima-guestagent
 
 # Launch the guestagent service
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
@@ -17,11 +17,6 @@ install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent /usr/local/bin/lima-guestage
 
 # Launch the guestagent service
 if [ -f /etc/alpine-release ]; then
-	# Create directory for the lima-guestagent socket (normally done by systemd)
-	mkdir -p /run/user/"${LIMA_CIDATA_UID}"
-	gid=$(id -g "${LIMA_CIDATA_USER}")
-	chown "${LIMA_CIDATA_UID}:${gid}" /run/user/"${LIMA_CIDATA_UID}"
-	chmod 700 /run/user/"${LIMA_CIDATA_UID}"
 	# Install the openrc lima-guestagent service script
 	cat >/etc/init.d/lima-guestagent <<'EOF'
 #!/sbin/openrc-run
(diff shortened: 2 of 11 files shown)


cmd/lima-guestagent/daemon_linux.go
```diff
@@ -40,17 +28,11 @@ var daemonCommand = &cli.Command{
...
-	socket := clicontext.String("socket")
-	if socket == "" {
-		return errors.New("socket must be specified")
-	}
+	socket := "/run/lima-guestagent.sock"
 	tick := clicontext.Duration("tick")
 	if tick == 0 {
 		return errors.New("tick must be specified")
 	}
-	if os.Geteuid() == 0 {
```
jandubois, line 51: Maybe reverse the check and throw an error when **not** running as root?

mattfarina: That's a good idea

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -30,18 +25,16 @@ supervisor=supervise-daemon
...
+pidfile="/run/lima-guestagent.pid"
 EOF
 	chmod 755 /etc/init.d/lima-guestagent
 
 	rc-update add lima-guestagent default
 	rc-service lima-guestagent start
 else
-	until [ -e "/run/user/${LIMA_CIDATA_UID}/systemd/private" ]; do sleep 3; done
-	sudo -iu "${LIMA_CIDATA_USER}" "XDG_RUNTIME_DIR=/run/user/${LIMA_CIDATA_UID}" lima-guestagent install-systemd
+	until [ -e "/run" ]; do sleep 3; done
```
jandubois, line 38: This should not be needed; the `/run/user/$UID/systemd/private` directory is created by systemd, so we need to wait for it to exists, but `/run` should always be there by the time cloud-init executes.

mattfarina: done.

cmd/lima-guestagent/daemon_linux.go
mattfarina: done

pkg/guestagent/iptables/iptables.go
```diff
@@ -0,0 +1,111 @@
...
+//    -A CNI-DN-2e2f8d5b91929ef9fc152 -d 127.0.0.1/32 -p tcp -m tcp --dport 8081 -j DNAT --to-destination 10.4.0.7:80
+//    -A CNI-DN-04579c7bb67f4c3f6cca0 -p tcp -m tcp --dport 8082 -j DNAT --to-destination 10.4.0.10:80
+// The -A on the front is to amend the rule that was already created. portmap
+// ensures the rule is created before creating this line so it is always -A.
+// CNI-DN- is the prefix used for rule for an individual container.
+// -d is followed by the IP address. The regular expression looks for a valid
+// ipv4 IP address. We need to detect this IP.
+// --dport is the destination port. We need to detect this port
+// -j DNAT this tells us it's the line doing the port forwarding.
+var findPortRegex = regexp.MustCompile(`.*-A\W+CNI-DN-\w*\W+(?:-d ((?:\b25[0-5]|\b2[0-4][0-9]|\b[01]?[0-9][0-9]?)(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}))?(?:/\d+\W+)?-p.*--dport (\d+) -j DNAT`)
```
jandubois, line 30: Some nit-picking:

* Remove the leading `.*` because it isn't anchored to the beginning of the line.
* Replace all `\W` with `\s` to match only whitespace. E.g. you don't want to match an inverted rule like `! -d 127.0.0.1/32`, but `\W` does match the `!`.
* Added a space after the `-p`.

```suggestion
var findPortRegex = regexp.MustCompile(`-A\s+CNI-DN-\w*\s+(?:-d ((?:\b25[0-5]|\b2[0-4][0-9]|\b[01]?[0-9][0-9]?)(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}))?(?:/\d+\s+)?-p .*--dport (\d+) -j DNAT`)
```

I suspect all `\s+` sequences can be replaced by single spaces (at least it still passes your tests):

```suggestion
var findPortRegex = regexp.MustCompile(`-A CNI-DN-\w* (?:-d ((?:\b25[0-5]|\b2[0-4][0-9]|\b[01]?[0-9][0-9]?)(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}))?(?:/\d+ )?-p .*--dport (\d+) -j DNAT`)
```

I'm also tempted to say that `/\d+ ` should be replaced by `/32 ` because I don't know what this would mean for any other mask value. But then I don't think portmap will specify any mask, so it may not matter.

pkg/guestagent/iptables/iptables.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+type Entry struct {
+	IP   net.IP
+	Port int
+}
+
+// This regex can detect a line in the iptables added by portmap to do the
+// forwarding. The following two are examples of lines (notice that one has the
+// destinateion IP and the other does not):
```
jandubois, line 20:
```suggestion
// destination IP and the other does not):
```

pkg/guestagent/iptables/iptables.go
mattfarina: > Remove the leading .* because it isn't anchored to the beginning of the line.

oops, at one point it was anchored with `^` but I removed that and failed to remove the `.*` along with it. Fixed.

> Replace all \W with \s to match only whitespace. E.g. you don't want to match an inverted rule like ! -d 127.0.0.1/32, but \W does match the !.

Great catch on the ! case. Fixed.

> Added a space after the -p.

Updated

> I suspect all \s+ sequences can be replaced by single spaces (at least it still passes your tests):

I used the `\s+` to be on the safe side. I was attempting to be loose in case something added more space in there. It should work with a single space. I'm leaving it that way for now to be loose with the parsing. If I really do need to tighten it please let me know.

> I'm also tempted to say that /\d+ should be replaced by /32 because I don't know what this would mean for any other mask value. But then I don't think portmap will specify any mask, so it may not matter.

I discovered the `/32` in my testing. I did not track down exactly what was adding it. I've updated it to only look for `/32` as that's the single IP and we should not have a range.

pkg/guestagent/iptables/iptables.go
mattfarina: fixed.

pkg/guestagent/iptables/iptables.go
```diff
@@ -0,0 +1,111 @@
...
+//    -A CNI-DN-2e2f8d5b91929ef9fc152 -d 127.0.0.1/32 -p tcp -m tcp --dport 8081 -j DNAT --to-destination 10.4.0.7:80
+//    -A CNI-DN-04579c7bb67f4c3f6cca0 -p tcp -m tcp --dport 8082 -j DNAT --to-destination 10.4.0.10:80
+// The -A on the front is to amend the rule that was already created. portmap
+// ensures the rule is created before creating this line so it is always -A.
+// CNI-DN- is the prefix used for rule for an individual container.
+// -d is followed by the IP address. The regular expression looks for a valid
+// ipv4 IP address. We need to detect this IP.
+// --dport is the destination port. We need to detect this port
+// -j DNAT this tells us it's the line doing the port forwarding.
+var findPortRegex = regexp.MustCompile(`-A\s+CNI-DN-\w*\s+(?:-d ((?:\b25[0-5]|\b2[0-4][0-9]|\b[01]?[0-9][0-9]?)(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}))?(?:/32\s+)?-p .*--dport (\d+) -j DNAT`)
```
AkihiroSuda, line 30: I don't think we should limit the regex to CNI rules.
We should cover non-CNI rules such as Docker libnetwork rules with `userland-proxy: false`.

We can just overestimate the set of the listening ports, and probably try `connect()`-ing to verify whether the port is actually open (for TCP ports).

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -17,11 +17,6 @@ install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent /usr/local/bin/lima-guestage
```
AkihiroSuda, line 17: We probably need to uninstall the previous version of the guest agent

cmd/lima-guestagent/daemon_linux.go
```diff
@@ -40,16 +28,13 @@ var daemonCommand = &cli.Command{
 }
```
AkihiroSuda, line 28: Probably this ticking is too heavy for watching iptables. Instead we should wait for `NETLINK_AUDIT` events or some eBPF stuff, but I can work on that in a separate PR after merging this.

pkg/guestagent/iptables/iptables.go
mattfarina: I tested lima with Docker and `userland-proxy: false`. The port forwarding with the latest release of lima and this patch works without any changes. The iptables rule for forwarding is listed and reads...

```
-A DOCKER -p tcp -m tcp --dport 8081 -j DNAT --to-destination 172.17.0.2:80
```

When the userland proxy is in place the iptables rule is different...

```
-A DOCKER -d 127.0.0.1/32 ! -i docker0 -p tcp -m tcp --dport 8081 -j DNAT --to-destination 172.17.0.2:80
```

Docker (I think moby) is doing something to make the ports work. For example, running `netstat -plunt` will show the forwarded ports for docker but not containerd (running as root).

Do we need to support Docker if it already works?

Note, I tested docker running rootless and as root.

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
mattfarina: Done

pkg/guestagent/iptables/iptables.go
mattfarina: I updated the iptables check to verify TCP ports are open before returning them.

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
jandubois: > We probably need to uninstall the previous version of the guest agent

Just out of curiosity: why do we need to remove it? `install` should be able to overwrite the file. What am I missing?

AkihiroSuda: Not overwritten because we switched away from non-root to root

AkihiroSuda: I meant we have to remove non-root systemd unit

jandubois: > I meant we have to remove non-root systemd unit

Ok, thanks, because that is not what https://github.com/mattfarina/lima/commit/3aba8f39fcccce1102e963db4c0ae34809db02e2 is doing...

mattfarina: This makes sense. Just pushed a change to remove the legacy systemd configuration.

AkihiroSuda (approved):
Thanks, LGTM but let me release v0.6.4 before merging this

mattfarina:
The latest push is a rebase on master after the cobra changes landed.

AkihiroSuda (approved):
Thanks!

Issue #253, opened 2021-09-17
I wrote a xbar plugin for limactl


unixorn:
I wrote an [xbar](https://xbarapp.com/) plugin script for `limactl` - it shows your lima VMs in a dropdown menu from xbar and lets you start or stop the individual lima VMs.

I don't see a section in the readme for extra tooling around lima, is it ok if I add one and include a link to [lima-xbar-plugin](https://github.com/unixorn/lima-xbar-plugin)?

AkihiroSuda:
Thanks, PR is welcome around here https://github.com/lima-vm/lima/tree/v0.6.4#adopters (Maybe s/Adopters/Relevant projects/)

PR #256, opened 2021-09-18
Add link to lima xbar plugin

unixorn:
- Add link to [lima-xbar-plugin](https://github.com/unixorn/lima-xbar-plugin)
- Sort adopters list

Closes https://github.com/lima-vm/lima/issues/253

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,8 +36,9 @@ to Mac users, but Lima can be used for non-container applications as well.
 
 ## Adopters
 
-- [Rancher Desktop](https://rancherdesktop.io/): Kubernetes and container management to the desktop
 - [Colima](https://github.com/abiosoft/colima): Docker (and Kubernetes) on macOS with minimal setup
+- [Lima xbar plugin](https://github.com/unixorn/lima-xbar-plugin) - [xbar](https://xbarapp.com/) plugin to start/stop VMs from the menu bar and see their running status.
+- [Rancher Desktop](https://rancherdesktop.io/): Kubernetes and container management to the desktop
 
 ## Examples
 
--- a/README.md
+++ b/README.md
@@ -36,9 +36,12 @@ to Mac users, but Lima can be used for non-container applications as well.
 
 ## Adopters
 
-- [Colima](https://github.com/abiosoft/colima): Docker (and Kubernetes) on macOS with minimal setup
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -36,8 +36,9 @@ to Mac users, but Lima can be used for non-container applications as well.
 
 ## Adopters
 
-- [Rancher Desktop](https://rancherdesktop.io/): Kubernetes and container management to the desktop
 - [Colima](https://github.com/abiosoft/colima): Docker (and Kubernetes) on macOS with minimal setup
+- [Lima xbar plugin](https://github.com/unixorn/lima-xbar-plugin) - Plugin for [xbar](https://xbarapp.com/) to create a Lima menu in your macOS menubar to start/stop your Lima VMs and see their running status.
```
jandubois, line 40: Seems slightly verbose compared to the other entries; it is kind of implied that this is about Lima. 😄 

```suggestion
- [Lima xbar plugin](https://github.com/unixorn/lima-xbar-plugin) - [xbar](https://xbarapp.com/) plugin to start/stop VMs from the menu bar and see their running status.
```

On second thoughts, this is just nit-picking, and the existing entry is fine too.

jandubois (approved):
LGTM

README.md
```diff
@@ -36,8 +36,9 @@ to Mac users, but Lima can be used for non-container applications as well.
 
 ## Adopters
 
-- [Rancher Desktop](https://rancherdesktop.io/): Kubernetes and container management to the desktop
 - [Colima](https://github.com/abiosoft/colima): Docker (and Kubernetes) on macOS with minimal setup
+- [Lima xbar plugin](https://github.com/unixorn/lima-xbar-plugin) - [xbar](https://xbarapp.com/) plugin to start/stop VMs from the menu bar and see their running status.
+- [Rancher Desktop](https://rancherdesktop.io/): Kubernetes and container management to the desktop
```
AkihiroSuda, line 41: I don’t think the list has to be alphabetically sorted.

Also, could you squash commits?

jandubois: > I don’t think the list has to be alphabetically sorted.

You'll need to let @unixorn know which order you prefer. Is it by original release date or something else?

> Also, could you squash commits?

Yes, the change done via the Github website is also missing the DCO line.

unixorn: Squashed the commits and signed the new squashed one.

I'm happy to revert the sort to the original order. It doesn't have to be alphabetical, I just feel that makes it easier to find things - not an issue now with three items, but for the future.

AkihiroSuda (approved):
Added a follow-up commit

PR #257, opened 2021-09-18
Add awesomebot github action

unixorn:
Awesomebot will check all the links in README.md and ensure they're still alive.

Includes example (img.shields.io has been a problem on my own repos) to whitelist sites with intermittent issues so they don't block PR checks.

Use `--request-delay 1` so that it doesn't hit rate limits for sites that have multiple links pointing at them.


.github/workflows/awesomebot.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: docker://dkhamsing/awesome_bot:latest
+      with:
+        args: /github/workspace/README.md --allow-dupe --allow-redirect --request-delay 1 --white-list https://github,https://img.shields.io,http://127.0.0.1:8080
```
AkihiroSuda, line 18: Why https://github has to be whitelisted?

.github/workflows/awesomebot.yml
```diff
@@ -0,0 +1,18 @@
...
+  pull_request:
+    branches: [ '*' ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
```
AkihiroSuda, line 15: `@v2`

.github/workflows/awesomebot.yml
```diff
@@ -0,0 +1,18 @@
+name: Check links in README.md
```
AkihiroSuda, line 1: We can have this in https://github.com/lima-vm/lima/blob/master/.github/workflows/test.yml

.github/workflows/awesomebot.yml
```diff
@@ -0,0 +1,18 @@
...
+on:
+  push:
+    branches: [ '*' ]
+  pull_request:
+    branches: [ '*' ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
AkihiroSuda, line 12: ubuntu-20.04

.github/workflows/awesomebot.yml
unixorn: Sorry, I missed that when I copied it from one of my other repos

.github/workflows/awesomebot.yml
unixorn: Done

.github/workflows/awesomebot.yml
unixorn: Done

.github/workflows/awesomebot.yml
unixorn: Removed

AkihiroSuda:
Added a follow-up commit

PR #263, opened 2021-09-23
Fix broken brackets

ocadaruma:
## Summary
- There's a broken brackets in `40-install-containerd.sh`, which makes containerd-setup process to be always executed


jandubois (approved):
LGTM

AkihiroSuda (approved):
Thanks

PR #266, opened 2021-09-23
Setup proxy environment variables from system settings

jandubois:
Implement the changes suggested in https://github.com/lima-vm/lima/issues/189#issuecomment-911972759.

`env.*` settings from `lima.yaml` will override system settings.

Process environment during instance creation override all.

Fixes https://github.com/lima-vm/lima/issues/189#issuecomment-911972759

### DIFF
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -7,23 +7,23 @@ import (
 )
 
 type LimaYAML struct {
-	Arch         Arch               `yaml:"arch,omitempty"`
-	Images       []File             `yaml:"images"` // REQUIRED
-	CPUs         int                `yaml:"cpus,omitempty"`
-	Memory       string             `yaml:"memory,omitempty"` // go-units.RAMInBytes
-	Disk         string             `yaml:"disk,omitempty"`   // go-units.RAMInBytes
-	Mounts       []Mount            `yaml:"mounts,omitempty"`
-	SSH          SSH                `yaml:"ssh,omitempty"` // REQUIRED (FIXME)
-	Firmware     Firmware           `yaml:"firmware,omitempty"`
-	Video        Video              `yaml:"video,omitempty"`
-	Provision    []Provision        `yaml:"provision,omitempty"`
-	Containerd   Containerd         `yaml:"containerd,omitempty"`
-	Probes       []Probe            `yaml:"probes,omitempty"`
-	PortForwards []PortForward      `yaml:"portForwards,omitempty"`
-	Networks     []Network          `yaml:"networks,omitempty"`
-	Network      NetworkDeprecated  `yaml:"network,omitempty"` // DEPRECATED, use `networks` instead
-	Env          map[string]*string `yaml:"env,omitempty"`     // EXPERIMENAL, see default.yaml
-	DNS          []net.IP           `yaml:"dns,omitempty"`
+	Arch         Arch              `yaml:"arch,omitempty"`
+	Images       []File            `yaml:"images"` // REQUIRED
+	CPUs         int               `yaml:"cpus,omitempty"`
+	Memory       string            `yaml:"memory,omitempty"` // go-units.RAMInBytes
+	Disk         string            `yaml:"disk,omitempty"`   // go-units.RAMInBytes
+	Mounts       []Mount           `yaml:"mounts,omitempty"`
+	SSH          SSH               `yaml:"ssh,omitempty"` // REQUIRED (FIXME)
+	Firmware     Firmware          `yaml:"firmware,omitempty"`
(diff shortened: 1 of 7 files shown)


jandubois:
This change is fully functional (afaict), but I'm still thinking about it's limitations: environment settings are "baked in" to the instance at the time it is created.

So how do we deal with e.g. a laptop that is moving between a company office requiring certain proxies and a public Wifi setup, using a VPN?

There is no way to update proxy settings even by stopping and restarting the instance, so maybe we need a different mechanism.

jandubois:
> There is no way to update proxy settings even by stopping and restarting the instance

One way to deal with this would be to re-create the cidata volume on each instance start. This would also allow you to e.g. edit the provisioning scripts, or the `dns` and `env` settings without having to destroy and recreate the instance.

@AkihiroSuda what do you think?

AkihiroSuda:
Aren’t we already always recreating cidata.iso ?

jandubois:
> Aren’t we already always recreating cidata.iso ?

Looks like it. When I tested last night, I didn't see changes to proxy settings after a restart. The only explanation I can come up with is that I ran `make all` instead of `make all install` before doing my tests (I was testing with an older pre-existing instance).

But that's good, so maybe this PR is sufficient then.

jandubois:
Given that we recreate cidata on each instance start, I think the PR is good as-is.

pkg/sysprof/network_darwin.go
```diff
@@ -5,10 +5,27 @@ type SPNetworkDataType struct {
 }
 
 type NetworkDataType struct {
-	DNS       DNS    `json:"DNS"`
-	Interface string `json:"interface"`
+	DNS       DNS     `json:"DNS"`
+	Interface string  `json:"interface"`
+	Proxies   Proxies `json:"proxies"`
```
AkihiroSuda, line 10: nit: `json:”Proxies”` for consistency with other json fields

jandubois: Interesting! I didn't know that JSON marshalling will use a case-insensitive match if it can't find an exact match, which is why this still worked despite not having the right capitalization. Fixed.

jandubois: I've left `interface` lowercase because that is how it comes in from `system_profiler`:

```console
$ system_profiler SPNetworkDataType -json | jq -r '.SPNetworkDataType[0]|keys[]' | sort
DNS
Ethernet
IPv4
IPv6
Proxies
_name
dhcp
hardware
interface
ip_address
sleep_proxies
spnetwork_service_order
type
```

pkg/limayaml/limayaml.go
```diff
@@ -7,23 +7,23 @@ import (
...
+	SSH          SSH               `yaml:"ssh,omitempty"` // REQUIRED (FIXME)
+	Firmware     Firmware          `yaml:"firmware,omitempty"`
+	Video        Video             `yaml:"video,omitempty"`
+	Provision    []Provision       `yaml:"provision,omitempty"`
+	Containerd   Containerd        `yaml:"containerd,omitempty"`
+	Probes       []Probe           `yaml:"probes,omitempty"`
+	PortForwards []PortForward     `yaml:"portForwards,omitempty"`
+	Networks     []Network         `yaml:"networks,omitempty"`
+	Network      NetworkDeprecated `yaml:"network,omitempty"` // DEPRECATED, use `networks` instead
+	Env          map[string]string `yaml:"env,omitempty"`
```
AkihiroSuda, line 25: Why was *string changed to string?

jandubois: There was some reason, but I don't remember, and it no longer applies; I can easily change it back. Please confirm if you want me to do that!

I asked back in https://github.com/lima-vm/lima/pull/195#discussion_r703894040 why it was a `*string` to begin with because I was having issues with them in templates elsewhere, but didn't receive a response.

So I'm still curious why it should be a `*string`. Theoretically we could use `nil` vs `""` to "unset" an environment variable, but that doesn't make much sense in `/etc/environment`, so I still have no idea...

jandubois: So it doesn't matter because (1) we no longer pass `y.Env` directly to the template, and (2) because we don't use `printf` on the values, but if we did, here is the difference (using this template):

```yaml
#LIMA-START
{{$key}}={{$val}}
{{printf "%s=%q" $key $val}}
{{- end}}
#LIMA-END
```

produces this output:

```
#LIMA-START
KEY=value"value
KEY=%!q(*string=0xc0001bb2a0)
#LIMA-END
```

which is what prejudices me against the `*string`. 😸

AkihiroSuda (approved):
Thanks

PR #273, opened 2021-09-28
Change instance id on every start

### LINKED ISSUE #270: Cisco AnyConnect VPN with DNS
AnyConnect will setup `utunN` interface as well as DNS resolvers I could see with `scutil --dns`.

This doesn't seem to be propagated into VMs. Is there a way to make it work, do I miss some configuration or it's just not supported?

I also tried to put some known DNS servers to my `~/.lima/default/lima.yaml`:

```yaml
dns:
- x.x.x.x
- y.y.y.y
```

And then `limactl stop && limactl start` - but my VM still doesn't have it in the `/etc/resolv.conf` neither am I able to use `curl` to get from a private host meaning it is not being injected to the VM in any other way. What am I missing?

jandubois:
That way cloud-init will do all the first-boot processing again to apply e.g. updated network configuration.

Fixes #270

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/meta-data
+++ b/pkg/cidata/cidata.TEMPLATE.d/meta-data
@@ -1,1 +1,2 @@
+instance-id: {{.IID}}
 local-hostname: lima-{{.Name}}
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks

PR #276, opened 2021-09-29
Implement starting instances from a HTTP URL

### LINKED ISSUE #237: Allow running start directly from an URL
It would be neat if you could start from an URL directly:

`limactl start https://raw.githubusercontent.com/lima-vm/lima/master/examples/podman.yaml`

USAGE:
   limactl start [command options] NAME|FILE.yaml

It's just a `curl` command away, but still an extra step to do.

For the paranoid, might want to look at embedding signing ?

afbjorklund:
Closes #237

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -3,9 +3,13 @@ package main
 import (
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"strings"
 
@@ -25,7 +29,7 @@ import (
 
 func newStartCommand() *cobra.Command {
 	var startCommand = &cobra.Command{
-		Use:               "start NAME|FILE.yaml",
+		Use:               "start NAME|FILE.yaml|URL",
 		Short:             fmt.Sprintf("Start an instance of Lima. If the instance does not exist, open an editor for creating new one, with name %q", DefaultInstanceName),
 		Args:              cobra.MaximumNArgs(1),
 		ValidArgsFunction: startBashComplete,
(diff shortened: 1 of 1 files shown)


jandubois (approved):
LGTM

Tested with

```
limactl start https://raw.githubusercontent.com/lima-vm/lima/master/examples/podman.yaml
```

cmd/limactl/start.go
```diff
@@ -49,7 +53,22 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string) (*store.Instance, e
...
+		if err != nil {
+			return nil, err
+		}
+		logrus.Debugf("interpreting argument %q as a http url for instance %q", arg, instName)
+		resp, err := http.Get(arg)
+		if err != nil {
+			return nil, err
+		}
+		defer resp.Body.Close()
+		yBytes, err = io.ReadAll(resp.Body)
```
AkihiroSuda, line 67: Should check body size before calling ReadAll

jandubois: Sorry for jumping the gun on merging already 😞

afbjorklund: Good call, hopefully it would return an error in that case ?

afbjorklund: Just used the https://pkg.go.dev/net/http sample code

AkihiroSuda: We can use https://pkg.go.dev/io#LimitedReader and it raises io.EOF when it reaches the limit (e.g., 1MB).
We should wrap the io.EOF error with a human-readable text.

afbjorklund: ah, so it was limit "protection" rather than looking for zero bytes

afbjorklund: but wouldn't the same thing happen with a local file as well ?

AkihiroSuda: PR: https://github.com/lima-vm/lima/pull/279

PR #280, opened 2021-09-30
yaml: support overriding the location of `nerdctl-full-X.Y.Z-linux-ARCH.tar.gz`

AkihiroSuda:
e.g.,
```yaml
containerd:
  system: false
  user: true
  archives:
    - location: "~/Downloads/nerdctl-full-X.Y.Z-linux-amd64.tar.gz"
       arch: "x86_64"
       digest: "sha256:..."
```

The default archive URL can be inspected with `limactl info`
```console
$ limactl info
{
    "version": "v0.6.4-63-g8dfd9d8",
    "defaultTemplate": {
        "arch": "x86_64",
        "images": [
            {
                "location": "~/Downloads/hirsute-server-cloudimg-amd64.img",
                "arch": "x86_64"
            },
            {
                "location": "~/Downloads/hirsute-server-cloudimg-arm64.img",
                "arch": "aarch64"
            },
            {
                "location": "https://cloud-images.ubuntu.com/hirsute/current/hirsute-server-cloudimg-amd64.img",
                "arch": "x86_64"
            },
            {
                "location": "https://cloud-images.ubuntu.com/hirsute/current/hirsute-server-cloudimg-arm64.img",
                "arch": "aarch64"
            }
        ],
        "cpus": 4,
        "memory": "4GiB",
        "disk": "100GiB",
        "mounts": [
            {
                "location": "~"
            },
            {
                "location": "/tmp/lima",
                "writable": true
            }
        ],
        "ssh": {
            "localPort": 60022,
            "loadDotSSHPubKeys": true
        },
        "firmware": {},
        "video": {
            "display": "none"
        },
        "containerd": {
            "system": false,
            "user": true,
            "archives": [
                {
                    "location": "https://github.com/containerd/nerdctl/releases/download/v0.11.2/nerdctl-full-0.11.2-linux-amd64.tar.gz",
                    "arch": "x86_64",
                    "digest": "sha256:27dbb238f9eb248ca68f11b412670db51db84905e3583834400305b2149915f2"
                },
                {
                    "location": "https://github.com/containerd/nerdctl/releases/download/v0.11.2/nerdctl-full-0.11.2-linux-arm64.tar.gz",
                    "arch": "aarch64",
                    "digest": "sha256:fe6322a88cb15d8a502e649827e3d1570210bb038b7a4a52820bce0fec86a637"
                }
            ]
        },
        "network": {}
    }
}
```

### DIFF
--- a/cmd/limactl/info.go
+++ b/cmd/limactl/info.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/lima-vm/lima/pkg/limayaml"
+	"github.com/lima-vm/lima/pkg/version"
+	"github.com/spf13/cobra"
+)
+
+func newInfoCommand() *cobra.Command {
+	infoCommand := &cobra.Command{
+		Use:   "info",
+		Short: "Show diagnostic information",
+		Args:  cobra.NoArgs,
+		RunE:  infoAction,
+	}
+	return infoCommand
+}
+
+type Info struct {
+	Version         string             `json:"version"`
+	DefaultTemplate *limayaml.LimaYAML `json:"defaultTemplate"`
+	// TODO: add diagnostic info of QEMU
+}
+
+func infoAction(cmd *cobra.Command, args []string) error {
+	y, err := limayaml.Load(limayaml.DefaultTemplate, "")
+	if err != nil {
+		return err
+	}
+	info := &Info{
+		Version:         version.Version,
+		DefaultTemplate: y,
+	}
+	j, err := json.MarshalIndent(info, "", "    ")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(cmd.OutOrStdout(), string(j))
+	return err
+}
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -11,6 +11,25 @@ import (
 	"github.com/lima-vm/lima/pkg/osutil"
 )
 
+func defaultContainerdArchives() []File {
+	const nerdctlVersion = "0.12.0"
+	location := func(goarch string) string {
+		return "https://github.com/containerd/nerdctl/releases/download/v" + nerdctlVersion + "/nerdctl-full-" + nerdctlVersion + "-linux-" + goarch + ".tar.gz"
+	}
+	return []File{
+		{
+			Location: location("amd64"),
+			Arch:     X8664,
+			Digest:   "sha256:7789800cfdd19fa9eccadb5e4a911e4ba759799ad9ec0b7929c983b9d149bc98",
+		},
+		{
+			Location: location("arm64"),
+			Arch:     AARCH64,
+			Digest:   "sha256:ebb05e22ac6a3c25ac88ca4f747feed89bfae8e447a626d0fedf3b4f40ac3303",
+		},
+	}
+}
+
 func MACAddress(uniqueID string) string {
 	sha := sha256.Sum256([]byte(osutil.MachineID() + uniqueID))
 	// "5" is the magic number in the Lima ecosystem.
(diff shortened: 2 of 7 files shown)


cmd/limactl/info.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/lima-vm/lima/pkg/limayaml"
+	"github.com/lima-vm/lima/pkg/version"
+	"github.com/spf13/cobra"
+)
+
+func newInfoCommand() *cobra.Command {
+	infoCommand := &cobra.Command{
+		Use:   "info",
+		Short: "Show diagnosis information",
```
jandubois, line 15: I'm not a native speaker, but I think "diagnostic information" would be a better term. "Diagnosis" is a noun (it is a "diagnostic procedure"), and using it as an adjective feels wrong to me.

AkihiroSuda: Thanks, fixed

cmd/limactl/info.go
```diff
@@ -0,0 +1,43 @@
...
+
+func infoAction(cmd *cobra.Command, args []string) error {
+	y, err := limayaml.Load(limayaml.DefaultTemplate, "")
+	if err != nil {
+		return err
+	}
+	info := &Info{
+		Version:         version.Version,
+		DefaultTemplate: y,
+	}
```
jandubois, line 36: I'm not really sure what the use case is for this command; when would you need the default template in JSON format? I have no objections to adding this, but without additional diagnostics included, I can't think of a use case.

AkihiroSuda: When potentially the nerdctl repo gets compromised, user can verify whether the default template is affected or not.

pkg/limayaml/defaults.go
```diff
@@ -59,6 +78,16 @@ func FillDefault(y *LimaYAML, filePath string) {
 	if y.Containerd.User == nil {
 		y.Containerd.User = &[]bool{true}[0]
 	}
+	needsContainerdArchives := (y.Containerd.User != nil && *y.Containerd.User) || (y.Containerd.System != nil && *y.Containerd.System)
```
jandubois, line 81: Why check the pointers for `nil`? The previous 6 lines just made sure that they can't be `nil`.

pkg/limayaml/defaults.go
```diff
@@ -59,6 +78,16 @@ func FillDefault(y *LimaYAML, filePath string) {
...
+	if needsContainerdArchives && len(y.Containerd.Archives) == 0 {
```
jandubois, line 82: Not sure why we need to check `needsContainerdArchives` at all; we could always set the default value, even if we know it won't be used.

pkg/limayaml/defaults.go
AkihiroSuda: removed

pkg/limayaml/defaults.go
AkihiroSuda: removed

pkg/limayaml/defaults.go
```diff
@@ -59,6 +78,13 @@ func FillDefault(y *LimaYAML, filePath string) {
 	if y.Containerd.User == nil {
 		y.Containerd.User = &[]bool{true}[0]
 	}
+	y.Containerd.Archives = defaultContainerdArchives()
```
jandubois, line 81: This still needs to be protected by `if len(y.Containerd.Archives) == 0`.

pkg/limayaml/validate.go
```diff
@@ -109,6 +109,10 @@ func Validate(y LimaYAML, warn bool) error {
 				i, ProvisionModeSystem, ProvisionModeUser)
 		}
 	}
+	needsContainerdArchives := (y.Containerd.User != nil && *y.Containerd.User) || (y.Containerd.System != nil && *y.Containerd.System)
```
jandubois, line 112: If we support calling `Validate()` without calling `FillDefaults()` first (and I think it is good to be robust here to avoid a panic, even though virtually all `lima.yaml` files will fail validation without `FillDefaults`), then `y.UseHostResolver` should be checked for `nil` as well before it is dereferenced.

jandubois (changes requested):

pkg/limayaml/defaults.go
AkihiroSuda: fixed 🤦‍♂️

pkg/limayaml/validate.go
AkihiroSuda: fixed

jandubois (approved):

PR #281, opened 2021-10-01
Add DNS server to host-agent to use native host resolver

### LINKED ISSUE #270: Cisco AnyConnect VPN with DNS

jandubois:
This is is required to correctly resolve hostnames while using conditional forwarding (split-DNS) when connected to a VPN.

The hostagent must be compiled with CGO_ENABLED=1 to use the native resolver.

The DNS server is listening for UDP connections on the same port number used for SSH:

```console
jan@lima-default:~$ dig @192.168.5.2 -p 60022 google.com
; <<>> DiG 9.16.8-Ubuntu <<>> @192.168.5.2 -p 60022 google.com
[...]
;; ANSWER SECTION:
google.com.		0	IN	A	172.217.14.206
```

It is then forwarded via iptables to `192.168.5.3:53`, replacing the DNS supplied by QEMU.

Fixes #270 
See also https://github.com/rancher-sandbox/rancher-desktop/issues/702


pkg/cidata/template.go
```diff
@@ -39,6 +39,7 @@ type TemplateArgs struct {
 	SlirpNICName string
 	SlirpGateway string
 	SlirpDNS     string
+	LocalPort    int // host port for ssh (tcp) and dns (udp)
```
AkihiroSuda, line 42: `SSHLocalPort`, and `UDPDNSLocalPort`

pkg/hostagent/dns.go
```diff
@@ -0,0 +1,123 @@
+// This file has been adapted from https://github.com/norouter/norouter/blob/master/pkg/agent/dns/dns.go
```
AkihiroSuda, line 1: s/master/v0.6.4/

Makefile
```diff
@@ -39,16 +39,16 @@ _output/bin/nerdctl.lima:
 
 .PHONY: _output/bin/limactl
 _output/bin/limactl:
-	$(GO_BUILD) -o $@ ./cmd/limactl
+	CGO_ENABLED=1 $(GO_BUILD) -o $@ ./cmd/limactl
```
AkihiroSuda, line 42: Please add a comment line to explain the reason

pkg/hostagent/hostagent.go
```diff
@@ -143,6 +143,12 @@ func (a *HostAgent) Run(ctx context.Context) error {
 		a.emitEvent(ctx, exitingEv)
 	}()
 
+	dnsServer, err := a.StartDNS()
```
AkihiroSuda, line 146: Could you add the design document to `internal.md`?

AkihiroSuda: And IIUC, we are going to let guestagent to listen on 127.X.X.X:53 on the guest and forward it to 192.168.5.2:XXXXX (which is forwarded to 127.0.0.0.1:XXXXX of the host), right?

pkg/cidata/template.go
jandubois: Should I still add both? We only need `UPDDNSLocalPort` in the template (once I figure out how to do the iptables rule).

If `iptables` doesn't work out, then we may not need it at all, but just create a little forwarder inside the guest-agent. But using iptables should be more efficient.

pkg/hostagent/hostagent.go
AkihiroSuda: And I guess it is not necessarily to listen on 127.0.0.1:XXXXX of the host.
We can just listen it on an UNIX socket on the host and use QEMU guestfwd. (`-nic user,guestfwd=tcp:192.168.5.2:XXXXX-unix:dns.sock)

jandubois: > And IIUC, we are going to let guestagent to listen on 127.X.X.X:53 on the guest

We will, unless this can be done automatically via `iptables`.

> 192.168.5.2:XXXXX (which is forwarded to 127.0.0.0.1:XXXXX of the host), right?

`192.168.5.2` **is** the gateway/host, so all ports from the host are automatically routable through it without any forwarding.

jandubois: > And I guess it is not necessarily to listen on 127.0.0.1:XXXXX of the host.
> We can just listen it on an UNIX socket on the host and use QEMU guestfwd. (`-nic user,guestfwd=tcp:192.168.5.2:XXXXX-unix:dns.sock)

This is not necessary; it already works without any further configuration.

jandubois: Also this is UDP; and we can't transport UDP over a domain socket (at least I don't think you can; I may be wrong).

AkihiroSuda: The guestagent should listen on 53/udp, but we can use arbitrary protocol for the communication between the guestagent and the hostagent.

Makefile
AkihiroSuda: cgo conflicts with cross-compilation.

I'd suggest shelling out nslookup/dig/getent or whatever (with caching the result

jandubois:
I have not been able to redirect this to port 53 on a loopback IP, to put it into `/etc/resolve.conf`. My iptables-fu is weak; I tried this (but it didn't work):

```console
root@lima-default:~# iptables --flush
root@lima-default:~# iptables -A FORWARD -d 192.168.5.2 -i eth0 -p udp -m udp --dport 8053 -j ACCEPT
root@lima-default:~# iptables -t nat -A PREROUTING -d 192.168.5.2 -p udp -m udp --dport 8053 -j DNAT --to-destination 127.0.53.53:53
root@lima-default:~# iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
root@lima-default:~# iptables -vnxL
# Warning: iptables-legacy tables present, use iptables-legacy to see them
Chain INPUT (policy ACCEPT 0 packets, 0 bytes)
    pkts      bytes target     prot opt in     out     source               destination

Chain FORWARD (policy ACCEPT 0 packets, 0 bytes)
    pkts      bytes target     prot opt in     out     source               destination
       0        0 ACCEPT     udp  --  eth0   *       0.0.0.0/0            192.168.5.2          udp dpt:8053

Chain OUTPUT (policy ACCEPT 0 packets, 0 bytes)
    pkts      bytes target     prot opt in     out     source               destination
root@lima-default:~# dig @127.0.53.53 google.com

; <<>> DiG 9.16.8-Ubuntu <<>> @127.0.53.53 google.com
; (1 server found)
;; global options: +cmd
;; connection timed out; no servers could be reached
```

Help wanted!

Also cross-compiling for Linux with `CGO_ENABLED=1` throws this error:

```
GOOS=linux GOARCH=amd64 make clean binaries
rm -rf _output
mkdir -p _output/bin
cp -a ./cmd/lima _output/bin/lima
CGO_ENABLED=1 go build -ldflags="-s -w -X github.com/lima-vm/lima/pkg/version.Version=v0.6.4-61-gf56f620.m" -o _output/bin/limactl ./cmd/limactl
# runtime/cgo
linux_syscall.c:67:13: error: implicit declaration of function 'setresgid' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
linux_syscall.c:67:13: note: did you mean 'setregid'?
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/unistd.h:593:6: note: 'setregid' declared here
linux_syscall.c:73:13: error: implicit declaration of function 'setresuid' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
linux_syscall.c:73:13: note: did you mean 'setreuid'?
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/unistd.h:595:6: note: 'setreuid' declared here
```

jandubois:
I forgot to mention that Ubuntu also needs this:

```
sysctl -w net.ipv4.ip_forward=1
```

Still didn't work though...

jandubois:
I believe the only outstanding issue with this PR is the cross-compilation problem with `CGO_ENABLED=1`.

It looks like cross-compilation for `arm64` on `amd64` works as long as `GOOS` for host and target are the same, so I'm planning to look into using separate builders for the Linux and macOS artifacts on Github actions.

But otherwise I'm pretty pleased with how this turned out. When `useHostResolver` is true, the QEMU DNS is transparently replaced with the hostagent DNS as soon as `iptables` is installed, and all names are resolved on the host from then on.

AkihiroSuda:
> It looks like cross-compilation for arm64 on amd64 works as long as GOOS for host and target are the same, so I'm planning to look into using separate builders for the Linux and macOS artifacts on Github actions.

SGTM.

`actions/upload-artifact@v2` and `actions/download-artifact@v2` can be used for copying files from the macOS job to the Linux job.

https://github.com/rootless-containers/usernetes/blob/f1845d2048de2a83aada679d1d294bf22ceb747d/.github/workflows/main.yaml#L37-L41

AkihiroSuda:
BTW if we can just fix libslirp to use the sane resolver, that will be helpful for non-Lima QEMU users and we can avoid CGO.

https://gitlab.freedesktop.org/slirp/libslirp/-/blob/v4.6.1/src/slirp.c#L244

jandubois:
> BTW if we can just fix libslirp to use the sane resolver, that will be helpful for non-Lima QEMU users and we can avoid CGO.
> https://gitlab.freedesktop.org/slirp/libslirp/-/blob/v4.6.1/src/slirp.c#L244

I'm afraid that is not as easy as it sounds. Afaict libslirp doesn't implement a DNS server; it simply looks up nameserver addresses, picks one (at random?), and then forwards the traffic to that server. And since it is just forwarding packets, it can't even fail over to another server if the chosen one fails.

The line you linked to gets the nameserver address from `/etc/resolv.conf` and is only used on Linux. The macOS/iOS code is using libresolv, but the end effect is the same: it picks an address and forwards all DNS traffic to that address:

https://gitlab.freedesktop.org/slirp/libslirp/-/blob/v4.6.1/src/slirp.c#L138

Afaik the local resolver on macOS is not exposed via DNS, so you would have to include a DNS server & client implementation inside libslirp, and we obviously couldn't use the Go library for that.

So while this would be great to have, I'm not volunteering to work on it, given that it seems to be a rather huge effort.

With systemd-resolved you get this out of the box: you configure the resolver, and it then exposes the resolver via DNS again at `127.0.0.53:53`. But we don't have this for macOS.

For Lima I think this PR provides all the functionality we need (famous last words alert).

You can even add names to `/etc/hosts` on the host, and have them immediately resolve inside the guest, or inside containers without having to restart anything.

jandubois:
@AkihiroSuda I think the PR is now ready for final review. I will test the darwin-aarch64 binaries tomorrow; but don't have real hardware to test the linux-arm64 binaries right now. 

If you think everything looks fine, maybe you could create a `test-action-release-*` tag, so we can do validation with the CI builds, to make sure the cross-compiled bits actually work?

pkg/cidata/cidata.TEMPLATE.d/boot/07-host-dns-setup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -eux
+
+# Wait until iptables has been installed; 30-install-packages.sh will call this script again
+if command -v iptables >/dev/null 2>&1; then
+	if [ "${LIMA_CIDATA_UDP_DNS_LOCAL_PORT}" -ne 0 ]; then
```
AkihiroSuda, line 6: `[ -n ${...} ] && [ ${...} -ne 0 ]` might be more robust

pkg/cidata/cidata.TEMPLATE.d/boot/07-host-dns-setup.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -eux
+
+# Wait until iptables has been installed; 30-install-packages.sh will call this script again
```
AkihiroSuda, line 4: Why not just call this script after installing iptables?

pkg/hostagent/dns.go
```diff
@@ -0,0 +1,123 @@
...
+		h.handleDefault(w, req)
+	}
+}
+
+func (a *HostAgent) StartDNS() (*dns.Server, error) {
+	h, err := newHandler()
+	if err != nil {
+		panic(err)
+	}
+	addr := fmt.Sprintf("127.0.0.1:%d", a.y.SSH.LocalPort)
```
AkihiroSuda, line 115: `a.y.SSH.LocalPort` will need to be changed to `a.sshLocalPort` after merging https://github.com/lima-vm/lima/pull/285

pkg/cidata/cidata.TEMPLATE.d/boot/07-host-dns-setup.sh
jandubois: I'll change it, but the variable is set from an `int` in the template, so can't really be empty.

pkg/cidata/cidata.TEMPLATE.d/boot/07-host-dns-setup.sh
jandubois: When `useHostResolver` is true, we can't set the `en0` nameservers explicitly, so have to rely on slirp picking a working one. If the user starts an image that has `iptables`, but is missing one of the other packages (say `sshfs`), then this allows them to install them, even if slirp DNS is broken.

jandubois:
I've tested the `CGO_ENABLED=1` build cross-compiled for M1 and it works fine (entry has `A` record, but no `AAAA` or `MX` record):

```console
jan@m1 ~ % echo "1.2.3.4 foobar.example.com" | sudo tee -a /etc/hosts
Password:
1.2.3.4 foobar.example.com
jan@m1 ~ % lima host foobar.example.com
foobar.example.com has address 1.2.3.4
Host foobar.example.com not found: 3(NXDOMAIN)
Host foobar.example.com not found: 3(NXDOMAIN)
```

So I think this is ready for merging 😄

AkihiroSuda (approved):
Thanks

PR #283, opened 2021-10-01
Support port forwarding for privileged ports (1-1023)

### LINKED ISSUE #45: Support port forwarding for privileged ports (1-1023)

AkihiroSuda:
On macOS, listening on 127.0.0.1:80 requires root while 0.0.0.0:80 does not require root.

For such ports, the hostagent launches "pseudoloopback" forwarder that listens on 0.0.0.0:80 but rejects connections from non-loopback src IP.

Fix #45


AkihiroSuda:
cc @jandubois @mattfarina

pkg/hostagent/hostagent.go
```diff
@@ -357,6 +358,146 @@ func (a *HostAgent) processGuestAgentEvents(ctx context.Context, localUnix strin
 	return io.EOF
 }
 
+func forwardTCP(ctx context.Context, sshConfig *ssh.SSHConfig, port int, local, remote string, cancel bool) error {
+	// on macOS, listening on 127.0.0.1:80 requires root while 0.0.0.0:80 does not require root.
+	// https://twitter.com/_AkihiroSuda_/status/1403403845842075648
+	// We use "pseudoloopback" forwarder that listens on 0.0.0.0:80 but rejects connections from non-loopback src IP.
+	if runtime.GOOS == "darwin" {
```
jandubois, line 365: This is a very long "true" branch, with a very short "else" branch a long time later. Maybe change to

```
if runtime.GOOS != "darwin" {
    return forwardSSH(ctx, sshConfig, port, local, remote, cancel)
}
```

and then outdent the rest of the function by one level (don't use an "else" branch)?

pkg/hostagent/hostagent.go
```diff
@@ -357,6 +358,146 @@ func (a *HostAgent) processGuestAgentEvents(ctx context.Context, localUnix strin
...
+		localIPStr, localPortStr, err := net.SplitHostPort(local)
+		if err != nil {
+			return err
+		}
+		localIP := net.ParseIP(localIPStr)
+		localPort, err := strconv.Atoi(localPortStr)
+		if err != nil {
+			return err
+		}
+		if net.ParseIP("127.0.0.1").Equal(localIP) && localPort < 1024 {
```
jandubois, line 375: I'm confused that there is no "else" branch for this condition, just a `return nil`. Doesn't that mean non-localhost, or non-privileged ports are no longer forwarded at all?

jandubois: So I think this should have a short block:

```
if !net.ParseIP("127.0.0.1").Equal(localIP) || localPort >= 1024 {
    return forwardSSH(ctx, sshConfig, port, local, remote, cancel)
}

and then the rest of the code could be outdented one level again.
```

AkihiroSuda: fixed

pkg/hostagent/hostagent.go
AkihiroSuda: Moved to `port_darwin.go`

PR #285, opened 2021-10-04
Assign `ssh.localPort` automatically to an available port

### LINKED ISSUE #131: Assign `ssh.localPort` automatically to an available port

AkihiroSuda:
- The hostagent now speaks REST API over `ha.sock` to provide the port information.  See `pkg/hostagent/api`.
- For backward compatibility, the "default" instance uses port 60022 by default.
    
Close #131


jandubois:
The current implementation disables the hostdns lookup because `ssh.localPort` is `0`.

If the port is only determined by the hostagent, then the cidata ISO creation must also be delegated to the hostagent, as it needs to include those port settings, so they are known during cloud-init processing.

AkihiroSuda:
Fixed to pick up a free UDP DNS port in `pkg/hostagent`

PR #286, opened 2021-10-04
nits

AkihiroSuda:
```
00891cd limactl: improve help output
7591b14 alpine: start networking earlier
021556e hostagent: propagate --debug
ac46219 README.md: update the status of the QEMU patches
f81883a pkg/hostagent/api -> pkg/hostagent/events
```

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -2,7 +2,9 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/lima-vm/lima/pkg/store/dirnames"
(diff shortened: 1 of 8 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot/10-alpine-prep.sh
```diff
@@ -32,3 +32,6 @@ rc-service cgroups start
 # `limactl stop` tells acpid to powerdown
 rc-update add acpid
 rc-service acpid start
+
+# enable networking explicitly here for apk
+rc-service networking start
```
AkihiroSuda, line 37: cc @jandubois

jandubois: This is really a bug in `lima-init`, as it works correctly with `cloud-init`. So I think it should just be fixed in the lima-init openrc script instead.

cmd/limactl/main.go
```diff
@@ -23,10 +25,30 @@ func main() {
...
+  $ limactl start
+
+  Open a shell:
+  $ lima
+
+  Run a container:
+  $ lima nerdctl run -d --name nginx -p 8080:80 nginx:alpine
+
+  Stop the default instance:
+  $ limactl stop
```
jandubois, line 49: This does not work right now, as the stop command has `cobra.ExactArgs(1)`. So you have to write `limactl stop ""`. This should be fixed by using `cobra.MaximumNArgs(1)` instead.

pkg/start/start.go
```diff
@@ -72,10 +72,17 @@ func Start(ctx context.Context, inst *store.Instance) error {
...
+	}
+
 	haCmd := exec.CommandContext(ctx, self,
-		"hostagent",
-		"--pidfile", haPIDPath,
-		inst.Name)
+		append(debugFlags, []string{
+			"hostagent",
+			"--pidfile", haPIDPath,
+			inst.Name}...)...)
```
jandubois, line 84: I find the `}...)...)` construct not very readable. Wouldn't it be better to just build up an `args` list upfront and then call:

```
	var args []string
	if logrus.GetLevel() >= logrus.DebugLevel {
		args = append(args, "--debug")
	}
	args = append(args,
		"hostagent",
		"--pidfile", haPIDPath,
		inst.Name)
	haCmd := exec.CommandContext(ctx, self, args...)
```

pkg/cidata/cidata.TEMPLATE.d/boot/10-alpine-prep.sh
jandubois: I've implemented the fix in https://github.com/lima-vm/alpine-lima/pull/19

AkihiroSuda: dropped

cmd/limactl/main.go
AkihiroSuda: fixed

pkg/start/start.go
AkihiroSuda: fixed

jandubois (approved):

PR #288, opened 2021-10-04
Check for openrc-init instead of alpine-release

afbjorklund:
When talking about the init and not the distro

(there are some other places with apk and such)

Otherwise it will hang forever, waiting for systemd


jandubois (approved):
LGTM

PR #289, opened 2021-10-04
Make sure that environment file exists

afbjorklund:
Before trying to append new things to it

Think I only had `/etc/profile` around ?

So it failed the boot.sh shell script...


jandubois (approved):
LGTM

PR #322, opened 2021-10-12
Add SSHIdentityFile to the list output

afbjorklund:
In order to connect to lima with an external ssh tool,
nice to have both the ssh port and the ssh identity...

Avoids hardcoding any dependencies on lima internals,
and the identity can be empty if using default keys.

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -102,13 +102,22 @@ func DefaultPubKeys(loadDotSSH bool) ([]PubKey, error) {
 	return res, nil
 }
 
-func CommonArgs(useDotSSH bool) ([]string, error) {
+// PrivateKeyPath returns the path to the private key
+func PrivateKeyPath() (string, error) {
 	configDir, err := dirnames.LimaConfigDir()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


afbjorklund:
You are right, there are lots of such options needed. And slighly different syntax, for ssh/scp/etc

e.g. I _still_ don't know how to the key to a DOCKER_HOST, so for now I have just used `ssh-add'.

For this time, I think I just wanted to move the construction of the URL from the template to the struct

afbjorklund:
And as long as we are hard-coding various paths on the remote host, the $XDG_RUNTIME_DIR isn't known to us either...

Everyone is assuming systemd and UID 1000, but this is going to crash some day. `/run/user/1000/podman/podman.sock`

afbjorklund:
> Note that we are planning to implement forwarding of unix sockets as part of the hostagent

Doing the unix socket tunneling is "nice", but I think it just shows that `ssh:` urls are not working...

But you are probably right in doing it, because even if introduced in Docker 18.09 no API but python supports it
They had do the temporary local socket workaround also for Podman "compability", mostly for Java API users.

PR #324, opened 2021-10-12
Implement custom go format for list command

### LINKED ISSUE #318: Add custom output format for the limactl list
It would be nice if it was possible to format the output with a Go template.
That would save have to roundtrip into JSON, for customizing the output...

```
Flags:
  -h, --help           help for list
  -f, --format string  Format the output using the given Go template
      --json           JSONify output
  -q, --quiet          Only show names
```

It would use the `lima/pkg/store.Instance` struct.

afbjorklund:
Closes #318

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"reflect"
+	"strings"
 	"text/tabwriter"
 	"text/template"
 
@@ -12,6 +14,16 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
+
+func instanceFields() []string {
+	fields := []string{}
+	var instance store.Instance
+	t := reflect.TypeOf(instance)
+	for i := 0; i < t.NumField(); i++ {
+		fields = append(fields, t.Field(i).Name)
+	}
+	return fields
+}
 
 func newListCommand() *cobra.Command {
 	listCommand := &cobra.Command{
@@ -24,6 +36,7 @@ func newListCommand() *cobra.Command {
 	}
 
 	listCommand.Flags().StringP("format", "f", "", "Format the output using the given Go template")
+	listCommand.Flags().Bool("list-fields", false, "List fields available for format")
 	listCommand.Flags().Bool("json", false, "JSONify output")
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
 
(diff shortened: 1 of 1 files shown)


afbjorklund:
> At first I expected the output to still have proper column format etc, but that is of course not possible when using generic go templates

I think Podman made up some home-grown `tables` function for this, they showed on the last community meeting...

afbjorklund:
> To be useful there needs to be a way to list all available template variables, either as part of `limactl ls --help`, or via a separate command. That information should then be reusable for the `--labels` or `--fields` options.

Added as "additonal help topic" command:

```console
$ _output/bin/limactl list fields
Name
Status
Dir
Arch
CPUs
Memory
Disk
Networks
SSHLocalPort
HostAgentPID
QemuPID
Errors


```

cmd/limactl/list.go
```diff
@@ -22,9 +35,18 @@ func newListCommand() *cobra.Command {
...
 	listCommand.Flags().Bool("json", false, "JSONify output")
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
 
+	fieldsCommand := &cobra.Command{
+		Use:   "fields",
+		Short: "Fields available for format",
+		Long:  strings.Join(instanceFields(), "\n"),
+	}
+
+	listCommand.AddCommand(fieldsCommand)
```
AkihiroSuda, line 48: I don't think this is correct.
We can just have a documentation to explain the fields.

afbjorklund: It uses https://pkg.go.dev/github.com/spf13/cobra#Command.IsAdditionalHelpTopicCommand

afbjorklund:
```
List instances of Lima.

Usage:
  limactl list [flags]

Aliases:
  list, ls

Flags:
  -f, --format string   Format the output using the given Go template
  -h, --help            help for list
      --json            JSONify output
  -q, --quiet           Only show names

Global Flags:
      --debug   debug mode

Additional help topics:
  limactl list fields Fields available for format

```

AkihiroSuda: The problem is that this makes "fields" to be a reserved instance name (in conjunction with #413).

afbjorklund: There is probably some other way to do it in cobra

jandubois: > The problem is that this makes "fields" to be a reserved instance name (in conjunction with #413).

I would expect it to work if you use `limactl ls -- fields`, but it is an ugly exception.

I think I would prefer an additional option: `limactl ls --list-fields` to show the fields (that is mutually exclusive with `--format` and `--json`).

jandubois (approved):
LGTM, although I have a slight preference to have `--list-fields` return them in sorted order. Maybe you can still update?

Not going to merge in case @AkihiroSuda has further feedback.

AkihiroSuda (approved):
Thanks

PR #330, opened 2021-10-13
limactl delete: delete default instance by default

knkarthik:
`start` and `stop` use the default instance if no instance is specified. Make `delete` do the same.

Signed-off-by: Karthik KN <knkarthik@users.noreply.github.com>


AkihiroSuda:
For safety `delete` is intentionally designed not to delete the default instance. Thanks for contribution anyway.

PR #331, opened 2021-10-13
Compatibility with Ubuntu 18.04 LTS (bionic) as a host OS

jandubois:
### Compatibility with qemu 2.11.1

* `-accel help` writes to stderr, not stdout
* `-netdev help` is not implemented

The `-netdev help` info is only used to detect vde support, which isn't used on Linux, so this doesn't affect Lima.

### Compatibility with OpenSSH 7.1p1

The feature to add algorithms to the front of the default set by using a leading `^` has been added in OpenSSH 8.0 and causes an error on earlier versions.

Specifying a different cipher is just a (minor) performance tweak, not a requirement.

<hr>

This PR also moves the SSH info gathering from the init phase to "on demand" when first used. That way the logging system is initialized when the `detect*` functions are running.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/cheggaaa/pb/v3 v3.0.8
 	github.com/containerd/containerd v1.5.7
 	github.com/containerd/continuity v0.2.0
+	github.com/coreos/go-semver v0.3.0
 	github.com/digitalocean/go-qemu v0.0.0-20210326154740-ac9e0b687001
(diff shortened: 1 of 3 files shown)


jandubois:
> We should have a CI, but it can be added later

What did you have in mind? Another Cirrus CI task that runs on 18.04 instead of 21.04? Testing just one of the guests, or all of them?

I'm kind of wary adding more CI platforms while the current ones are still flaky. It will just add to the frustration.

PR #348, opened 2021-10-18
Add example of running centos distribution

afbjorklund:
~~Requires all the other patches, before it actually _works_.~~

But otherwise very similar to fedora (it's based on fc28)

### CentOS Linux

This is supported until the end of the year.
https://cloud.centos.org/centos/8/
(downstream of RHEL)

https://www.centos.org/centos-linux/

### CentOS Stream

And then it can be substituted for "Stream".
https://cloud.centos.org/centos/8-stream/
(upstream of RHEL)

https://www.centos.org/centos-stream/

----

See https://blog.centos.org/2020/12/future-is-centos-stream/

And https://www.centos.org/centos-stream/ for Stream 9 status

### DIFF
--- a/examples/centos.yaml
+++ b/examples/centos.yaml
@@ -1,4 +1,8 @@
 # This example requires Lima v0.7.0 or later.
+
+# Hint: If QEMU crashes with error "vmx_write_mem: mmu_gva_to_gpa XXXXXXXXXXXXXXXX failed" on Intel Mac,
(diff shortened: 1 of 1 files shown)


jandubois:
@AkihiroSuda Should this config also be added to `.cirrus.yml`?

PR will need to be rebased for the test to pass though.

AkihiroSuda:
We have to skip `containerd-rootles-setuptool.sh install-stargz` for CentOS 8, as its kernel is older than 5.11

PR #362, opened 2021-10-23
Write /etc/environment before the lingering session is started

### LINKED ISSUE #365: http_proxy et.al. not set in a new VM until restart
Version 0.7.2

http_proxy, https_proxy, ftp_proxy, and no_proxy  are not set in new VM until the VM is stopped and restarted.
This seems to be a small regression from 0.7.1. 

Steps to reproduce.

```
export http_proxy=http://myproxy:12345
limactl stop 
limactl delete default
limactl start </dev/null   # new vm with defaults
lima bash -c '[[ -z "$http_proxy" ]] && echo http_proxy UNSET || echo http_proxy OK'
  ==> http_proxy UNSET
limactl stop
limactl start
lima bash -c '[[ -z "$http_proxy" ]] && echo http_proxy UNSET || echo http_proxy OK'
  ==> http_proxy OK
```

jandubois:
Should address https://github.com/lima-vm/lima/pull/351#discussion_r734459563
Fixes #365 

Also makes sure we are not reusing a session that was started by a requirements check before `/etc/environment` was updated.

This is a race condition that seems to happen maybe 25% of the time when an instance is started:

```
+ loginctl user-status 501
jan (501)
           Since: Sat 2021-10-23 07:17:02 UTC; 2s ago
           State: active
        Sessions: *2
          Linger: no
            Unit: user-501.slice
                  ├─session-2.scope
                  │ ├─1341 sshd: jan [priv]
                  │ ├─1438 sshd: jan@notty
                  │ ├─1440 /bin/bash
                  │ ├─1441 timeout 30s bash -c until command -v sshfs; do sleep 3; done
                  │ ├─1442 bash -c until command -v sshfs; do sleep 3; done
                  │ └─1443 sleep 3
                  └─user@501.service
                    └─init.scope
                      ├─1344 /lib/systemd/systemd --user
                      └─1345 (sd-pam)
```

If we don't terminate it, then it will be reused for starting rootless components (which will then be missing proxy settings), and also for later ssh connections because we `enable-linger`.


AkihiroSuda:
Is this still WIP?

jandubois:
> Is this still WIP?

Yes, it is, but it is also possible that I have fixed the remaining issue with the refactoring of the shutdown logic I did for the socket forwarding PR. Will try to confirm later today.

jandubois:
> Yes, but the guest agent connection can only be established once the ga is running (after `25-guest-agent-base`),

This turned out to be not correct, and I had to add another check that the instance is "ssh-ready" before setting up sshfs mounts etc.

I believe this PR is now good (assuming I didn't break CI).

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -64,5 +59,9 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.user ]; then
 	done
 fi
 
+# Signal that provisioning is done. The instance-id in the meta-data file changes on every boot,
+# so any copy from a previous boot cycle will have different content.
+cp "${LIMA_CIDATA_MNT}"/meta-data /etc/lima-boot-done
```
AkihiroSuda, line 64: Can we use `/run/lima` rather than `/etc`?

AkihiroSuda: If you need persistence it should be `/var/lib/lima`

jandubois: No, `/run` is perfect.

I want the opposite of persistence, which is why I copy the `meta-data` file, which I know has different content on every restart. Because we have no way to "reset" the ready markers before cloud-init allows ssh access, so the requirements checks can move past them before they have been reset. Having a different content for the marker on every boot solves this.

Updated

AkihiroSuda (approved):
Thanks, merging, but please consider updating `docs/internal.md` to explain the `/run` files

PR #369, opened 2021-10-29
Implement unix socket forwarding

### LINKED ISSUE #221: Add support for forwarding UNIX sockets from the guest to the host
For example the docker port forwarding in #210 would be more robust if it would just forward the socket instead of letting docker connect via ssh. The latter is brittle because user names and keys all must resolve correctly for this to work.

See https://github.com/abiosoft/colima/blob/9adcf3c/colima#L151 for a mechanism that should be built into the host agent.

jandubois:
This is an alternative to #367. It doesn't parse `/proc/net/unix` at all, but just sets up any configured socket forwards when the guestagent is started, and tears them down when the hostagent exits.

This is now my preferred implementation.

```console
$ limactl start --tty=false examples/docker.yaml
[...]
$ DOCKER_HOST=unix://$HOME/docker.sock docker info | grep -E 'Root|System'
 Operating System: Ubuntu Impish Indri (development branch)
 Docker Root Dir: /home/jan.linux/.local/share/docker
WARNING: Running in rootless-mode without cgroups. To enable cgroups in rootless-mode, you need to boot the system in cgroup v2 mode.
```

Fixes #221

### DIFF
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -2,13 +2,7 @@
 # $ limactl start ./docker.yaml
 # $ limactl shell docker docker run -it -v $HOME:$HOME --rm alpine
 
-# Hint: To allow `docker` CLI on the host to connect to the Docker daemon running inside the guest,
-# add `NoHostAuthenticationForLocalhost yes` in ~/.ssh/config , and then run the following commands:
-# $ export DOCKER_HOST=ssh://localhost:60006
-# $ docker ...
-
-# If ssh:// ... does not work, try the following commands:
-# $ ssh -f -N -p 60006 -i ~/.lima/_config/user -o NoHostAuthenticationForLocalhost=yes -L $HOME/docker.sock:/run/user/$(id -u)/docker.sock 127.0.0.1
+# To run `docker` on the host (assumes docker-cli is installed):
 # $ export DOCKER_HOST=unix://$HOME/docker.sock
 # $ docker ...
 
@@ -63,3 +57,6 @@ probes:
         exit 1
       fi
     hint: See "/var/log/cloud-init-output.log". in the guest
+portForwards:
+  - guestSocket: "/run/user/{{.UID}}/docker.sock"
+    hostSocket: "{{.Home}}/docker.sock"
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -11,6 +11,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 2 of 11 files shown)


examples/docker.yaml
```diff
@@ -63,3 +63,6 @@ probes:
         exit 1
       fi
     hint: See "/var/log/cloud-init-output.log". in the guest
+portForwards:
+  - guestSocket: /run/user/${UID}/docker.sock
+    hostSocket: ${HOME}/docker.sock
```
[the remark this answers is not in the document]
jandubois: I've switched to Go templates and also added `{{.Instance}}` for `hostSocket`.

examples/docker.yaml
```diff
@@ -63,3 +63,6 @@ probes:
         exit 1
       fi
     hint: See "/var/log/cloud-init-output.log". in the guest
+portForwards:
+  - guestSocket: "/run/user/{{.UID}}/docker.sock"
+    hostSocket: "{{.Home}}/docker.sock"
```
AkihiroSuda, line 68: > 5: # Hint: To allow `docker` CLI on the host to connect to the Docker daemon running inside the guest,

Could you update this part too?

AkihiroSuda: Do we want to update `podman.yaml` too?

jandubois: > Do we want to update `podman.yaml` too?

Fine with me; just let me know. I've updated `docker.yaml` because I used it for testing the PR. 😄

pkg/hostagent/hostagent.go
```diff
@@ -485,8 +506,35 @@ func forwardSSH(ctx context.Context, sshConfig *ssh.SSHConfig, port int, local,
...
+			logrus.Infof("Forwarding %q (guest) to %q (host)", remote, local)
+			if err := os.RemoveAll(local); err != nil {
+				logrus.WithError(err).Warnf("Failed to clean up %q (host) before setting up forwarding", local)
+			}
+			if err := os.MkdirAll(filepath.Dir(local), 0750); err != nil {
+				return fmt.Errorf("can't create directory for local socket %q: %w", local, err)
+			}
+		case verbCancel:
+			defer func() {
+				logrus.Infof("Stopping forwarding %q (guest) to %q (host)", remote, local)
```
jandubois, line 521: The logging should be outside the deferred cleanup.

examples/docker.yaml
jandubois: > Do we want to update `podman.yaml` too?

I tried, but it doesn't work when I set `DOCKER_HOST=$HOME/podman.sock`, and neither does `PODMAN_HOST` or `CONTAINER_HOST`.

I have some suspision that `podman` is trying to be too smart, and knows that it can't be running natively on macOS, so only connects over ssh, or something.

Maybe @afbjorklund knows and can fix this in a followup PR?

AkihiroSuda: > `DOCKER_HOST=$HOME/podman.sock`

Did you mean `DOCKER_HOST=unix://$HOME/podman.sock`

jandubois: > > `DOCKER_HOST=$HOME/podman.sock`
> 
> Did you mean `DOCKER_HOST=unix://$HOME/podman.sock`

Yes:

```console
$ CONTAINER_HOST=ssh://$(id -un)@localhost:60906/run/user/$(id -u)/podman/podman.sock CONTAINER_SSHKEY=$HOME/.lima/_config/user podman ps
CONTAINER ID  IMAGE       COMMAND     CREATED     STATUS      PORTS       NAMES

$ DOCKER_HOST=unix://$HOME/podman.sock podman ps
Cannot connect to Podman. Please verify your connection to the Linux system using `podman system connection list`, or try `podman machine init` and `podman machine start` to manage a new Linux VM
Error: unable to connect to Podman socket: Get "http://d/v3.4.1/libpod/_ping": dial unix ///var/folders/wm/pq6w7kf54215fxz392ff3rxr0000gn/T/podman-run--1/podman/podman.sock: connect: no such file or directory

$ ls -l ~/*.sock
srw-------  1 jan  staff  0  1 Nov 17:47 /Users/jan/podman.sock
```

jandubois: Hold on, there is a typo in the forward definition 😞

jandubois: Yes, typo. Now works with `DOCKER_HOST=... docker ps` as well as `CONTAINER_HOST=... podman ps`.

I've updated the PR

AkihiroSuda (approved):
Thanks

PR #371, opened 2021-11-02
Follow-up to "Implement unix socket forwarding"

AkihiroSuda:
Follow up to #369

### DIFF
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -6,6 +6,7 @@
 # $ export DOCKER_HOST=unix://$HOME/docker.sock
 # $ docker ...
 
+# This example requires Lima v0.7.3 or later
 images:
   # Hint: run `limactl prune` to invalidate the "current" cache
   - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-amd64.img"
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -38,11 +38,12 @@ provision:
       curl -fsSL https://get.docker.com | sh
       # NOTE: you may remove the lines below, if you prefer to use rootful docker, not rootless
       systemctl disable --now docker
-      apt-get install -y uidmap
+      apt-get install -y uidmap dbus-user-session
   - mode: user
     script: |
       #!/bin/bash
       set -eux -o pipefail
+      systemctl --user start dbus
       dockerd-rootless-setuptool.sh install
       docker context use rootless
 probes:
--- a/examples/podman.yaml
+++ b/examples/podman.yaml
@@ -10,6 +10,7 @@
 # $ export DOCKER_HOST=unix://$HOME/podman.sock
 # $ docker ...
 
+# This example requires Lima v0.7.3 or later
 images:
   # Hint: run `limactl prune` to invalidate the "current" cache
   - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-amd64.img"


AkihiroSuda:
```console
$ DOCKER_HOST=unix://$HOME/docker.sock docker run -it --rm alpine
docker: Error response from daemon: OCI runtime create failed: container_linux.go:380: starting container process caused: process_linux.go:385: applying cgroup configuration for process caused: error while starting unit "docker-f635a58e3ec4f5ff87cdd7b16054f26776c0aba0d6db961c1f1bf3c12cf1fd46.scope" with properties [{Name:Description Value:"libcontainer container f635a58e3ec4f5ff87cdd7b16054f26776c0aba0d6db961c1f1bf3c12cf1fd46"} {Name:Slice Value:"user.slice"} {Name:PIDs Value:@au [5181]} {Name:Delegate Value:true} {Name:MemoryAccounting Value:true} {Name:CPUAccounting Value:true} {Name:IOAccounting Value:true} {Name:TasksAccounting Value:true} {Name:DefaultDependencies Value:false}]: read unix @->/run/systemd/private: read: connection reset by peer: unknown.
ERRO[0000] error waiting for container: context canceled 
```

😞

jandubois:
Just out of curiosity, why is `dbus-user-session` required? Doesn't `loginctl enable-linger` already do everything needed to keep the daemon running?

AkihiroSuda:
Because runc needs to connect to dbus

https://github.com/moby/moby/issues/42793
https://github.com/opencontainers/runc/pull/3186

jandubois:
> Because runc needs to connect to dbus

So why does it work for me even without `dbus-user-session` with just #369:

```
$ limactl start --tty=false examples/docker.yaml
[...]
$ DOCKER_HOST=unix://$HOME/docker.sock docker run -it --rm alpine uname -a
Linux 9d6f08041b9f 5.13.0-14-generic #14-Ubuntu SMP Mon Aug 2 12:43:35 UTC 2021 x86_64 Linux
```

jandubois:
> So why does it work for me even without `dbus-user-session` with just #369:

Looks like it is installed already:

```console
root@lima-docker:~# apt -qq list dbus-user-session
dbus-user-session/impish 1.12.20-2ubuntu2 amd64
root@lima-docker:~# apt install aptitude
[...]
root@lima-docker:~# aptitude why dbus-user-session
i   ubuntu-server Depends  dirmngr
i A dirmngr       Suggests dbus-user-session
```

AkihiroSuda:
`dbus-user-session` wasn't installed in http://cloud-images.ubuntu.com/impish/20211014/impish-server-cloudimg-amd64.img (sha256:fce2af66787ccfed5eef0de2ac38eaedacdd3d8ee8a645c26d5a736559265478) when `containerd.user` is set to `false`.

Maybe your cache revision is different.

jandubois:
> Maybe your cache revision is different.

It is:

```console
$ ag -G url impish
download/by-url-sha256/ac74da77a6828e35de7edaa06fdbb33d12ef97cce2726550017e3c1066c88fb1/url
1:https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-amd64.img
```

Still weird that the installed packages are changing, but I think there is little value in digging any deeper....

jandubois:
Even weirder that my cached version seems older, but has a package that your version seems to be missing:

```console
$ ls -l ~/Library/Caches/lima/download/by-url-sha256/ac74da77a6828e35de7edaa06fdbb33d12ef97cce2726550017e3c1066c88fb1
total 1105208
-rw-r--r--  1 jan  staff  565051392  7 Sep 17:09 data
-rw-r--r--  1 jan  staff         79  7 Sep 17:08 url
```

Not sure how to determine the original download URL; I guess we don't store the endpoint of a redirect chain.

PR #38, opened 2021-06-05
Implement system and user provisioning scripts

jandubois:
Fixes #33 

In addition this PR implements a `bare-vm` setting that disables installation of optional requirements (`containerd` and `nerdctl`), and also skips installation of `sshfs` if there are no directories to mount.

Example use to provision kubernetes via `k3s` (although not in rootless mode):

```console
$ cat k3s.yaml
images:
- location: "~/Downloads/hirsute-server-cloudimg-amd64.img"
  arch: "x86_64"

mounts: []

ssh:
  localPort: 60022

bare-vm: true

provision:
  system: |
    #!/bin/sh
    curl -sfL https://get.k3s.io | sh -
$ limactl start --tty=false k3s.yaml
[...]
INFO[0032] READY. Run `lima bash` to open the shell.
INFO[0032] Forwarding "/run/user/501/lima-guestagent.sock" (guest) to "/Users/jan/.lima/k3s/ga.sock" (host)
INFO[0051] Forwarding TCP port 6444
INFO[0051] Forwarding TCP port 6443
INFO[0054] Forwarding TCP port 10010
[...]
```

And in a second shell:

```console
$ export KUBECONFIG=$PWD/kubeconfig.yaml
$ limactl shell k3s sudo cat /etc/rancher/k3s/k3s.yaml >$KUBECONFIG
$ kubectl get no
NAME       STATUS   ROLES                  AGE   VERSION
lima-k3s   Ready    control-plane,master   69s   v1.21.1+k3s1
$ kubectl get po -A
NAMESPACE     NAME                                      READY   STATUS      RESTARTS   AGE
kube-system   coredns-7448499f4d-gsv6h                  1/1     Running     0          67s
kube-system   local-path-provisioner-5ff76fc89d-kbmtj   1/1     Running     0          67s
kube-system   metrics-server-86cbb8457f-g7bjg           1/1     Running     0          67s
kube-system   helm-install-traefik-crd-sr9m8            0/1     Completed   0          67s
kube-system   helm-install-traefik-f587z                0/1     Completed   1          67s
kube-system   svclb-traefik-sqj88                       2/2     Running     0          47s
kube-system   traefik-97b44b794-j8t8s                   1/1     Running     0          47s
```

PR #375, opened 2021-11-04
Remove default mounts for orchestrators

afbjorklund:
Make k8s and nomad more similar to k3s

### DIFF
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -17,11 +17,8 @@ images:
     arch: "x86_64"
   - location: "https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-arm64.img"
     arch: "aarch64"
-mounts:
-  - location: "~"
-    writable: false
-  - location: "/tmp/lima"
-    writable: true
+# Mounts are disabled in this example, but can be enabled optionally.
+mounts: []
 containerd:
   system: true
   user: false
--- a/examples/nomad.yaml
+++ b/examples/nomad.yaml
@@ -15,11 +15,8 @@ images:
     arch: "x86_64"
   - location: "https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-arm64.img"
     arch: "aarch64"
-mounts:
-  - location: "~"
-    writable: false
-  - location: "/tmp/lima"
-    writable: true
+# Mounts are disabled in this example, but can be enabled optionally.
+mounts: []
 containerd:
   system: true
   user: false


AkihiroSuda:
Any specific reason to do so?

jandubois:
I have no strong opinion either way. I guess having no mounts at all saves you from installing `sshfs` during first boot, but otherwise it doesn't really cost much.

On the other hand is seems unlikely that you are making use of them when you are using `k8s`. Using `hostpath` volumes is not really good practice, but given that we are running single-node dev environments, they should work reliably, and can occasionally be useful.

So I guess I have a slight preference to keep them unless there is a specific reason to drop them.

afbjorklund:
My main goal was to make k3s and k8s more similar...

Changing whitespace and mounts were just casualties.

Similarly, I wanted Kubernetes and Nomad to be similar too.

afbjorklund:
Why were the defaults dropped for k3s ?

jandubois:
> Why were the defaults dropped for k3s ?

That script started out as a PR comment when I implemented provisioning scripts: https://github.com/lima-vm/lima/pull/38#issuecomment-856497148

Since it was supposed to show how provisioning works, I kept it otherwise as minimal as possible. It was not originally meant to end up in the `examples` directory.

I agree that the `mounts` should be added to `k3s` as well, if we keep them for `k8s` and `nomad`.

afbjorklund:
I just wanted the file to be less complicated, I don't think that they will be used ? Especially now when moving `kubectl` outside, then it less likely to be doing "apply" on any file from the home directory or something

jandubois:
> I just wanted the file to be less complicated, I don't think that they will be used ? Especially now when moving `kubectl` outside, then it is less likely to be doing "apply" on any file from the home directory or something

Yeah, the only use case I see it mounting a host directory into a container, to persist data outside the VM. I expect it to be rarely used too. As I said, I don't have a strong opinion, just a slight preference.

I'm fine with whatever @AkihiroSuda decides...

examples/k8s.yaml
```diff
@@ -10,11 +10,6 @@ images:
     arch: "x86_64"
   - location: "https://cloud-images.ubuntu.com/focal/current/focal-server-cloudimg-arm64.img"
     arch: "aarch64"
-mounts:
-  - location: "~"
-    writable: false
-  - location: "/tmp/lima"
-    writable: true
 containerd:
```
AkihiroSuda, line 13: Please add this note, then LGTM
https://github.com/lima-vm/lima/blob/cf42676298bb29c762ea0cb4803d4c6150147e9c/examples/k3s.yaml#L21-L22

afbjorklund:
Ok, I will remove the mounts again and update

AkihiroSuda (approved):
Thanks

PR #376, opened 2021-11-04
Add lint target for running yamllint

afbjorklund:
There were some whitespace errors in the example files.


AkihiroSuda:
Thanks, can we test this in CI?

https://github.com/lima-vm/lima/blob/2a618333137819ea7483423f1994dfd5d098a554/.github/workflows/test.yml#L10-L63

jandubois:
> There were some whitespace errors in the example files.

I somewhat disagree with this. I know that many tools seem to generate the double-indentation of sequences, but to me it looks rather odd. So I would prefer to set `indent-sequences` to `false` and format accordingly.

To quote the YAML spec:

>Since people perceive the “-” indicator as indentation, nested block sequences may be indented by one less space to compensate, except, of course, if nested inside another block sequence ([BLOCK-OUT context] versus [BLOCK-IN context]).

For me a properly indented sequence looks like this:

```yaml
list:
- 1
- 2
- 3
```

and not like this:

```yaml
list:
  - 1
  - 2
  - 3
```

But I realize that this may be a personal preference.

afbjorklund:
I think I might "odd" rather than "wrong". As with any linter, it would be good if you could figure one style and stick with it...

Just that yaml, like python, has synthetic whitespace - so it would be nice to guard against syntax errors ?

Tried using `limactl validate`

But it failed in the Mac file

afbjorklund:
Changed it to be consistent within each file instead (of enforcing a global sequence indentation)

jandubois:
> I think I meant "odd" rather than "wrong". As with any linter, it would be good if you could figure one style and stick with it...

I'm fine with that as long as the style matches the "proper" way a hand-written YAML file is supposed to look, and not like something that comes out of a simple serializer. 😄 

Just look at the https://yaml.org website to see how sequences are supposed to be written...

> Just that yaml, like python, has synthetic whitespace - so it would be nice to guard against syntax errors ?

Completely agree!
 
> Tried using `limactl validate`
> But it failed on the Mac file

I don't understand what this means; what fails, and how?

I believe @AkihiroSuda was just asking to add a `yamllint` step to the Github Action workflow. It should be easy to do now that the Github runners already have the latest version pre-installed: https://github.com/actions/virtual-environments/issues/1657#issuecomment-717953802

> Changed it to be consistent within each file instead (of enforcing a global sequence indentation)
> 
> `indent-sequences: consistent`

Yes, that is probably the best for this PR.

I do wonder though if we have (or can reach) consensus on changing some of the other YAML files to match, to eventually use the same style throughout the project?

afbjorklund:
> I don't understand what this means; what fails, and how?

`FATA[0000] failed to load YAML file "examples/vmnet.yaml": field `networks[0].lima` is only supported on macOS right now`

afbjorklund:
> I do wonder though if we have (or can reach) consensus on changing some of the other YAML files to match, to eventually use the same style throughout the project?

As long as there is a tool to automatically reformat it, I think that could work. Perhaps build one into limactl ?

afbjorklund:
> Thanks, can we test this in CI?

Copied the existing "shellcheck"

examples/k3s.yaml
```diff
@@ -12,11 +12,11 @@
...
-  # Hint: run `limactl prune` to invalidate the "current" cache
-  - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-amd64.img"
-    arch: "x86_64"
-  - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
-    arch: "aarch64"
+# Hint: run `limactl prune` to invalidate the "current" cache
+- location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-amd64.img"
+  arch: "x86_64"
+- location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
+  arch: "aarch64"
```
AkihiroSuda, line 19: This indentation doesn't seem consistent with other files in the same directory

jandubois: Is is consistent with `alpine.yaml`, and is using the "proper" indentation of sequences in YAML. I would like to eventually convert the other files to match (I asked about this earlier in this "thread").

AkihiroSuda: Seems we have to fix `default.yaml`.

Can we automatically detect this kind of issues?

jandubois: > Can we automatically detect this kind of issues?

Yes, in `.yamllint` we have to set `indent-sequences` to `false` instead of `consistent` (the consistency check only works within a file, not across files).

The changes will have to be made manually though, as all the reformatters I'm aware of will drop comments. Also in `default.yaml` much of the sequences are commented out, so wouldn't be reformatted automatically anyways.

I can volunteer to make those edits (and change `indent-sequences`) in a follow-up PR, if we agree to do them. Unless @afbjorklund wants to make them as part of this PR, of course.

afbjorklund: I thought this would be a simple check :-) I'm OK with either one-standard-per-file, or wait until a global one emerges

So far I have found that `strictyaml` preserves comments, but it also adds drops the quotes from all string values...

.yamllint
```diff
@@ -0,0 +1,10 @@
+---
+
+extends: default
+
+rules:
+  indentation:
+    indent-sequences: consistent
+  comments-indentation: disable
+  document-start: disable
+  line-length: disable
```
jandubois, line 10: `yamllint` seems to be stuck on YAML 1.1 definitions of truthy values, so complains that `on` should not be used because it is a boolean: `truthy value should be one of [false, true]`.

This hasn't been true since YAML spec 1.2 (released in 2009), where only `true`, `True`, `TRUE`, `false`, `False`, and `FALSE` are treated as boolean unless quoted.

Our `release.yml` file uses an unquoted `on` in 2 places:

```yaml
on:
  push:
    tags:
      - 'v*'
      - 'test-action-release-*'
env:
  GO111MODULE: on
```

We can tell `yamllint` to ignore it via:

```yaml
rules:
  truthy:
    allowed-values: ['true', 'false', 'on']
```

This is kind of bogus for YAML 1.2, but does turn off the warnings, so maybe it is good enough for us.

examples/k3s.yaml
jandubois: > So far I have found that `strictyaml` preserves comments, but it also adds drops the quotes from all string values...

I would want to test this. `ruamel.yaml` also claimed to retain comments, but broke badly when I tried it. And I think `strictyaml` may be based on `ruamel`, but I didn't look too closely.

> https://hitchdev.com/strictyaml/why/implicit-typing-removed/

I'm somewhat sympathetic with this issue, but the "Norway problem" has been fixed by the YAML 1.2 spec in 2009...
 
> I think indenting them is consistent with machine-readable "markup" (like XML), and not indenting them is consistent with human-readable "language" (like MD). Will probably be hard to keep everyone happy ?

Yes, that sounds right to me. I think the double-indented form comes from machine-generated YAML because it can be produced by recursive tree walking without the writer having to be context aware. Similar to simplistic C code generation:

```c
if (condition)
    {
        statement-1;
        statement-2;
    }
```

But a human would normally write this as:

```c
if (condition) {
    statement-1;
    statement-2;
}
```

Our YAML files are supposed to be read by humans, so should be formatted for that purpose. Otherwise we could be using JSON instead and avoid all these issues (any valid JSON file is automatically also a valid YAML file).

jandubois:
> This had some insightful comments on YAML, about why Unmarshal and Marshal probably is not the way to do it:

We should not be overthinking this; it is a one-time manual edit, and then we have the `yamllint` rules to make sure we stick with it. I could probably have already done all the manual conversion in the time we spent discussing this. 😄

afbjorklund:
It's all yaml engineering

jandubois:
@afbjorklund What is the status on the PR? Are you going to extend the YAML linting to the `*.yml` files used by Github Actions (see my feedback above)?

Or would you prefer if we add this separately later?

afbjorklund:
I can add it (later), or if you want to do it after merge that is fine too.

jandubois:
> I can add it (later)

That's fine; it is not urgent. I just wanted to make sure you are not waiting for the PR to be merged, while I was waiting for you to extend the PR to cover `*.yml` files as well.

Let me know if you decide you won't get around to it anytime soon, and I'll have another look at merging it as-is.

afbjorklund:
I noticed yamllint has a built-in file filter, which made the invocation simpler (and also checks .yamllint itself)

```yaml
---

yaml-files:
  - '*.yaml'
  - '*.yml'
  - '.yamllint'
```

.github/workflows/test.yml
```diff
@@ -21,6 +21,20 @@ jobs:
...
           args: --verbose
 
+  yamllint:
+    runs-on: ubuntu-20.04
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+      - name: Install yamllint
```
jandubois, line 31: This shouldn't be necessary; all Github runners should have the latest `yamllint` pre-installed (via `pip`): https://github.com/actions/virtual-environments/issues/1657#issuecomment-712228497

afbjorklund: Not keeping up to speed on the latest github developments, just matching the local environment

jandubois: I don't think it makes a big difference, but the apt version of `yamllint` is typically older than what you get via `pip` install. I don't really care that much; but removing redundant code and potentially having a more up-to-date version seem to favour just deleting this step.

Put let me know if you prefer to merge as-is; we can always change later.

jandubois (approved):
Thanks, LGTM

AkihiroSuda (approved):
Thanks, but I still would like to see all yaml files to have same indentation style.

Can be another PR.

afbjorklund:
You (anyone) change them, and then change the `.yamllint` accordingly.

I think the votes were in favor of "human" indent, rather than "machine" ?

jandubois:
> You (anyone) can change them, and then change the `.yamllint` accordingly.

Yes. I will eventually do it (because it irritates me), if nobody else beats me to it.

> I think the votes were in favor of "human" indent, rather than "machine" ?

Absolutely! `indent-sequences: false` it the goal. 😄

PR #387, opened 2021-11-05
read ecdsa public keys to ssh-authorized-keys

enihsyou:
I use [Secretive](https://github.com/maxgoedjen/secretive) to manage SSH keys, and it defaults to `ecdsa-sha2-nistp256` algorithm. 
After following the instruction about [putting public key in your ~/.ssh/ directory](https://github.com/maxgoedjen/secretive/blob/main/FAQ.md#how-do-i-tell-ssh-to-use-a-specific-key), `limactl start` give these output
```
WARN[0000] [hostagent] public key "/Users/enihsyou/.ssh/id_rsa.pub" doesn't seem to be in ssh format 
WARN[0000] [hostagent] public key "/Users/enihsyou/.ssh/secretive.pub" doesn't seem to be in ssh format 
```

Those OpenSSH public key format should be picked up, so I learned from what [golang.org/x/crypto/ssh.ParsePublicKey](https://github.com/golang/crypto/blob/089bfa5675191fd96a44247682f76ebca03d7916/ssh/keys.go#L265) do to make a soft public key check.

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -2,6 +2,8 @@ package sshutil
 
 import (
 	"bytes"
+	"encoding/base64"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -94,7 +96,7 @@ func DefaultPubKeys(loadDotSSH bool) ([]PubKey, error) {
 		}
 		entry, err := readPublicKey(f)
 		if err == nil {
-			if strings.ContainsRune(entry.Content, '\n') || !strings.HasPrefix(entry.Content, "ssh-") {
+			if !detectValidPublicKey(entry.Content) {
 				logrus.Warnf("public key %q doesn't seem to be in ssh format", entry.Filename)
 			} else {
 				res = append(res, entry)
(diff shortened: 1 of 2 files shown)


pkg/sshutil/sshutil.go
```diff
@@ -261,3 +263,26 @@ func detectOpenSSHVersion() semver.Version {
 	}
 	return v
 }
+
+// detectValidPublicKey returns whether content represent a public key.
+// OpenSSH public key format have the structure of '<algorithm> <key> <comment>'.
+// By checking 'algorithm' with signature format identifier in 'key' part,
+// this function may report false positive but provide better compatibility.
+func detectValidPublicKey(content string) bool {
+	// content have been space-trimmed
```
jandubois, line 272: I would prefer to keep the check for a newline character in here. We added it because having a newline breaks the generated cloud-init YAML and doesn't allow connecting to the instance via SSH at all.

I guess this would now be caught by the signature check anyways, but keeping the check costs little, even if it may be redundant.

enihsyou: Looks good to me, branch added

jandubois (approved):
Thanks! LGTM

PR #399, opened 2021-11-07
Use crun instead of runc for podman example

### LINKED ISSUE #398: Volume mounts not working with podman runtime
Volume mounts are working with nerdctl and docker, but not with podman:

* default
* docker
* podman

```console
$ lima nerdctl run -v $PWD/hello:/hello docker.io/busybox cat /hello/world
Hello, World!
$ export DOCKER_HOST=unix://$HOME/docker.sock
$ docker run -v $PWD/hello:/hello docker.io/busybox cat /hello/world
Hello, World!
$ export CONTAINER_HOST=unix://$HOME/podman.sock
$ podman --remote run -v $PWD/hello:/hello docker.io/busybox cat /hello/world
Trying to pull docker.io/library/busybox:latest...
Copying blob sha256:01c2cdc137396a9b78be86c47e5773388e6eed953dec79e257777518134b45f1
Copying config sha256:cabb9f684f8ba3edb303d578bfd7d709d853539ea1b420a3f6c81a08e85bb3d7
Storing signatures
Error: error preparing container beb192f8ed88b02e3f3fcbdb3551d29de12770f9aa590e7987af6ec0d4d5e659 for attach: container_linux.go:380: starting container process caused: process_linux.go:545: container init caused: rootfs_linux.go:76: mounting "/home/anders/lima/hello" to rootfs at "/hello" caused: operation not permitted: OCI permission denied
```

It does work with regular root containers, but not with rootless containers.

```console
$ limactl shell podman sudo podman run -v $PWD/hello:/hello docker.io/busybox cat /hello/world
Hello, World!
```

(For some reason it pulls the image every time, probably an upstream bug?
It doesn't do it when running locally, so some quirk with the remote API etc)

afbjorklund:
It seems like podman has issues using runc with cgroups v2.

So use crun instead, which is the podman default (with v2).

```go
        if conf.Engine.OCIRuntime == "" {
                conf.Engine.OCIRuntime = "runc"
                // If we're running on cgroups v2, default to using crun.
                if onCgroupsv2, _ := cgroups.IsCgroup2UnifiedMode(); onCgroupsv2 {
                        conf.Engine.OCIRuntime = "crun"
                }
        }
```

Closes #398

### DIFF
--- a/examples/podman.yaml
+++ b/examples/podman.yaml
@@ -35,7 +35,7 @@ provision:
       command -v podman >/dev/null 2>&1 && exit 0
       export DEBIAN_FRONTEND=noninteractive
       apt-get update
-      apt-get install -y podman
+      apt-get install -y podman crun
   - mode: user
     script: |
       #!/bin/bash


AkihiroSuda (approved):
LGTM, but I guess this is an issue of Podman side rather than runc

PR #400, opened 2021-11-08
Add LIMA_HOME to `limactl info`

jandubois:
This is useful because the directory will have all symlinks resolved, so can be used to determine the path to a relative hostSocket.

### DIFF
--- a/cmd/limactl/info.go
+++ b/cmd/limactl/info.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/lima-vm/lima/pkg/limayaml"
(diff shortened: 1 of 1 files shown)


cmd/limactl/info.go
```diff
@@ -22,6 +23,7 @@ func newInfoCommand() *cobra.Command {
 type Info struct {
 	Version         string             `json:"version"`
 	DefaultTemplate *limayaml.LimaYAML `json:"defaultTemplate"`
+	LimaHome        string             `json:"LimaHome"`
```
AkihiroSuda, line 26: json:”limaHome”

AkihiroSuda (approved):
Thanks

PR #413, opened 2021-11-16
Add possibility to list particular instances

afbjorklund:
As mentioned in https://github.com/lima-vm/lima/pull/324#issuecomment-942611305

`limactl list default`

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -14,18 +14,28 @@ import (
 
 func newListCommand() *cobra.Command {
 	listCommand := &cobra.Command{
-		Use:               "list",
+		Use:               "list [flags] [INSTANCE]...",
 		Aliases:           []string{"ls"},
 		Short:             "List instances of Lima.",
-		Args:              cobra.NoArgs,
+		Args:              cobra.ArbitraryArgs,
 		RunE:              listAction,
-		ValidArgsFunction: cobra.NoFileCompletions,
+		ValidArgsFunction: listBashComplete,
 	}
 
 	listCommand.Flags().Bool("json", false, "JSONify output")
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
 
 	return listCommand
+}
+
+func instanceMatches(arg string, instances []string) []string {
+	matches := []string{}
+	for _, instance := range instances {
+		if instance == arg {
+			matches = append(matches, instance)
+		}
+	}
+	return matches
 }
 
 func listAction(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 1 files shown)


cmd/limactl/list.go
```diff
@@ -14,10 +14,10 @@ import (
...
 func newListCommand() *cobra.Command {
 	listCommand := &cobra.Command{
-		Use:               "list",
+		Use:               "list [flags] [INSTANCE]...",
 		Aliases:           []string{"ls"},
 		Short:             "List instances of Lima.",
-		Args:              cobra.NoArgs,
+		Args:              cobra.ArbitraryArgs,
 		RunE:              listAction,
 		ValidArgsFunction: cobra.NoFileCompletions,
```
jandubois, line 22: Shouldn't this complete instance names?

```suggestion
		ValidArgsFunction: listBashComplete,
```

and

```go
func listBashComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
}
```

@AkihiroSuda I don't quite understand why we have a separate wrapper around `bashCompleteInstanceNames` for each command; why don't we share a single implementation?

cmd/limactl/list.go
```diff
@@ -73,8 +93,10 @@ func listAction(cmd *cobra.Command, args []string) error {
 	w := tabwriter.NewWriter(cmd.OutOrStdout(), 4, 8, 4, ' ', 0)
 	fmt.Fprintln(w, "NAME\tSTATUS\tSSH\tARCH\tCPUS\tMEMORY\tDISK\tDIR")
 
-	if len(instances) == 0 {
+	if len(allinstances) == 0 {
 		logrus.Warn("No instance found. Run `limactl start` to create an instance.")
+	} else if len(instances) == 0 {
```
jandubois, line 98: This will not show an error when some instances exist, but others don't:

```console
$ limactl ls foo bar
WARN[0000] No instance matching [foo bar] found.
NAME    STATUS    SSH    ARCH    CPUS    MEMORY    DISK    DIR
$ limactl ls foo bar alpine
NAME      STATUS     SSH                ARCH      CPUS    MEMORY    DISK      DIR
alpine    Running    127.0.0.1:51115    x86_64    4       4GiB      100GiB    /Users/jan/.lima/alpine
```

cmd/limactl/list.go
AkihiroSuda: We can just share the single impl

cmd/limactl/list.go
afbjorklund:
```console
$ _output/bin/limactl list foo bar alpine
WARN[0000] No instance matching foo found.              
WARN[0000] No instance matching bar found.              
NAME      STATUS     SSH            ARCH      CPUS    MEMORY    DISK      DIR
alpine    Stopped    127.0.0.1:0    x86_64    4       4GiB      100GiB    /home/anders/.lima/alpine

```

jandubois (approved):
Thanks, LGTM

PR #433, opened 2021-11-24
Change the default prompt color from green to lime

### LINKED ISSUE #427: Make the shell prompt `$PS1` distinguishable from the host shell
Maybe we should change the color of `$PS1` by default

afbjorklund:
Closes #427

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/lima.env
+++ b/pkg/cidata/cidata.TEMPLATE.d/lima.env
@@ -1,3 +1,4 @@
+LIMA_CIDATA_NAME={{ .Name }}
 LIMA_CIDATA_USER={{ .User }}
 LIMA_CIDATA_UID={{ .UID }}
 LIMA_CIDATA_MOUNTS={{ len .Mounts }}
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -102,6 +102,7 @@ Max file name length = 30
 The volume label is "cidata", as defined by [cloud-init NoCloud](https://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html).
 
 ### Environment variables
+- `LIMA_CIDATA_NAME`: the lima instance name
 - `LIMA_CIDATA_MNT`: the mount point of the disk. `/mnt/lima-cidata`.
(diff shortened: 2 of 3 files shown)


pkg/cidata/cidata.TEMPLATE.d/lima.env
```diff
@@ -1,3 +1,4 @@
+LIMA_CIDATA_NAME={{ .Name }}
```
jandubois, line 1: Please add this to the ""Environment variables" section in `docs/internal.md`

afbjorklund:
Apparently this gives a warning, the second time you run it:

```
LIMA| WARNING: Failed to execute /mnt/lima-cidata/boot/08-shell-prompt.sh
```

```
patching file .bashrc
Reversed (or previously applied) patch detected!  Skipping patch.
2 out of 2 hunks ignored
```

Should probably leave some kind of text that could be grepped ?

pkg/cidata/cidata.TEMPLATE.d/lima.env
afbjorklund: Added

AkihiroSuda:
Thanks, but this doen't seem to work with the default configuration of tmux (version 3.2a)?

(Adding `set-option -g default-terminal "screen-256color"` to `~/.tmux.conf` worked for me)

afbjorklund:
Hmm, it shouldn't be any different from the previous (ubuntu) setup ? What is the tmux invocation for lima ?

AkihiroSuda:
Maybe, just `PS1="[LIMA] $PS1"` might be fine

afbjorklund:
And for UTF-8 it would be 🄻🄸🄼🄰 or something

afbjorklund:
Then again, that will break a dozen _other_ scenarios - and doesn't even seem to work on GNOME Terminal

jandubois (approved):
LGTM

I probably would have appended the new logic at the end, overwriting the default prompt when needed. That would be somewhat more robust for future Ubuntu upgrades, that might break this patch. But we can also change to that approach if/when that happens.

PR #435, opened 2021-11-24
Add COLORTERM to the lima shell environment

### LINKED ISSUE #434: Forward the setting of `COLORTERM` from the host to lima shell
See https://github.com/termstandard/colors and https://github.com/lima-vm/lima/issues/427#issuecomment-975934793

Only TERM (256) is set by default, not COLORTERM (true color)

```
TERM=xterm-256color
COLORTERM=truecolor
```

----

See https://en.wikipedia.org/wiki/ANSI_escape_code#Colors

afbjorklund:
If the variable is set on the host, forward it.

Typical value if set would be COLORTERM=truecolor

Closes #434

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+set -eux
+
+if grep -q "COLORTERM" /etc/ssh/sshd_config; then
+	exit 0
+fi
+
+# accept any incoming COLORTERM environment variable
+sed -i 's/^AcceptEnv LANG LC_\*$/AcceptEnv COLORTERM LANG LC_*/' /etc/ssh/sshd_config
+if [ -f /sbin/openrc-init ]; then
+	rc-service --ifstarted sshd reload
+elif command -v systemctl >/dev/null 2>&1; then
+	if systemctl -q is-active ssh; then
+		systemctl reload ssh
+	elif systemctl -q is-active sshd; then
+		systemctl reload sshd
+	fi
+fi
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -127,6 +127,10 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		// required for showing the shell prompt: https://stackoverflow.com/a/626574
 		sshArgs = append(sshArgs, "-t")
 	}
+	if _, present := os.LookupEnv("COLORTERM"); present {
+		// SendEnv config is cumulative, with already existing options in ssh_config
+		sshArgs = append(sshArgs, "-o", "SendEnv=\"COLORTERM\"")
+	}
 	sshArgs = append(sshArgs, []string{
 		"-q",
 		"-p", strconv.Itoa(inst.SSHLocalPort),


AkihiroSuda:
Thanks, but CI failing

```
LIMA| Executing /mnt/lima-cidata/boot/11-colorterm-environment.sh
+ grep -q COLORTERM /etc/ssh/sshd_config
+ sed -i 's/^AcceptEnv LANG LC_\*$/AcceptEnv COLORTERM LANG LC_*/' /etc/ssh/sshd_config
+ test -f /sbin/openrc-init
+ command -v systemctl
+ systemctl reload ssh
Failed to reload ssh.service: Unit ssh.service not found.
LIMA| WARNING: Failed to execute /mnt/lima-cidata/boot/11-colorterm-environment.sh
LIMA| Executing /mnt/lima-cidata/boot/20-rootless-base.sh
```

https://cirrus-ci.com/task/4980033568636928

afbjorklund:
Apparently opensuse has a different name for `ssh.service` :-(

afbjorklund:
I fixed the ssh/sshd naming controversy, the current archlinux failure seems to be something else ?

AkihiroSuda:
```
+ command -v systemctl
+ systemctl reload 'ssh*'
Failed to reload sshdgenkeys.service: Job type reload is not applicable for unit sshdgenkeys.service.
LIMA| WARNING: Failed to execute /mnt/lima-cidata/boot/11-colorterm-environment.sh
```

https://cirrus-ci.com/task/5709533226991616?logs=test#L478

afbjorklund:
Ah, Ok. Thanks for spotting it.

pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -eux
+
+! grep -q "COLORTERM" /etc/ssh/sshd_config || exit 0
```
jandubois, line 4: I find `cond && action` simpler than `! cond || action`:

```suggestion
grep -q "COLORTERM" /etc/ssh/sshd_config && exit 0
```

But really, why does this need to be so compact? This isn't `perl`, and we don't have to pay extra per line! The action is more important than the condition, so should be at the beginning of a line. The long form still seems the clearest to me:

```suggestion
if grep -q "COLORTERM" /etc/ssh/sshd_config; then
    exit 0
fi
```

Now personally I would be fine with

```shell
exit 0 if grep -q "COLORTERM" /etc/ssh/sshd_config
```

but that isn't valid shell code, and more importantly, many people would disagree with me. 😸

afbjorklund: I thought it was the error condition, that the script will stop when it "fails" ? But maybe `if` and `&&` are excluded.

afbjorklund: Your "short form" sounded more ruby than perl ;-)

AkihiroSuda (approved):

jandubois:
Looks good to me, but for some reason doesn't work:

```console
$ COLORTERM=truecolor limactl --debug shell default
DEBU[0000] changeDirCmd="cd \"/Users/jan\" || cd \"/Users/jan\""
DEBU[0000] OpenSSH version 8.1.1 detected
DEBU[0000] executing ssh (may take a long)): [/usr/bin/ssh -F /dev/null -o IdentityFile="/Users/jan/.lima/_config/user" -o IdentityFile="/Users/jan/.ssh/[REDACTED]" -o IdentityFile="/Users/jan/.ssh/[REDACTED]" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o NoHostAuthenticationForLocalhost=yes -o GSSAPIAuthentication=no -o PreferredAuthentications=publickey -o Compression=no -o BatchMode=yes -o IdentitiesOnly=yes -o Ciphers="^aes128-gcm@openssh.com,aes256-gcm@openssh.com" -o User=jan -o ControlMaster=auto -o ControlPath="/Users/jan/.lima/default/ssh.sock" -o ControlPersist=5m -t -o SendEnv="COLORTERM" -q -p 60022 127.0.0.1 -- cd "/Users/jan" || cd "/Users/jan" ; exec bash --login]
jan@lima-default:/Users/jan$ sudo grep -B6 reload.ssh /var/log/cloud-init-output.log
LIMA| Executing /mnt/lima-cidata/boot/11-colorterm-environment.sh
+ grep -q COLORTERM /etc/ssh/sshd_config
+ sed -i s/^AcceptEnv LANG LC_\*$/AcceptEnv COLORTERM LANG LC_*/ /etc/ssh/sshd_config
+ [ -f /sbin/openrc-init ]
+ command -v systemctl
+ systemctl -q is-active ssh
+ systemctl reload ssh
jan@lima-default:/Users/jan$ grep COLORTERM /etc/ssh/sshd_config
AcceptEnv COLORTERM LANG LC_*
jan@lima-default:/Users/jan$ env | grep TERM
TERM=xterm-256color
```

I wonder if this is somehow related to this being a "lingering" session, so maybe environment variables aren't updated? I haven't looked further; @afbjorklund does this actually work for you?

Also wondering if this should be documented somewhere. Afaict, the only setting that would make any kind of difference would be `COLORTERM=truecolor`; all other values are ignored, right?

afbjorklund:
If you apply it to a running instance, you need to kill the ssh mux socket.

afbjorklund:
> Afaict, the only setting that would make any kind of difference would be `COLORTERM=truecolor`; all other values are ignored, right?

I haven't seen any documentation, or any other value. The `TERMTHEME` variable I totally made up.

AkihiroSuda (approved):
Thanks

Eventually, we should ensure this to be configured before starting sshd

PR #437, opened 2021-11-26
Bump Alpine from 3.13.5 → 3.14.3

jandubois:
Also switches `qemu-aarch64` from Alpine repo to `tonistiigi/binfmt` to include additional patches, and install `qemu-x86_64` in the `aarch64` ISO.


AkihiroSuda (approved):
Thanks

PR #438, opened 2021-11-26
limactl --list-fields should list them in sort order

jandubois:
```console
$ limactl ls --list-fields
Arch
CPUs
Dir
Disk
Errors
HostAgentPID
Memory
Name
Networks
QemuPID
SSHLocalPort
Status
```


AkihiroSuda (approved):
Thanks

PR #439, opened 2021-11-29
Add `IdentityFile` and `LimaHome` fields to `limactl list` format

jandubois:
Based on the suggestion in https://github.com/lima-vm/lima/pull/324#issuecomment-942611305.

Update examples/docker.yaml to show how this can be used to construct the instance-specific path of forwarded sockets.

```console
$ export DOCKER_HOST=$(limactl list docker --format 'unix://{{.LimaHome}}/{{.Name}}/sock/docker')
$ docker info
```

When the port forwards don't specify an absolute path:

```yaml
portForwards:
    hostSocket: docker
```

### DIFF
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -3,7 +3,7 @@
 # $ limactl shell docker docker run -it -v $HOME:$HOME --rm alpine
 
 # To run `docker` on the host (assumes docker-cli is installed):
-# $ export DOCKER_HOST=unix://$HOME/docker.sock
+# $ export DOCKER_HOST=$(limactl list docker --format 'unix://{{.InstanceDir}}/sock/docker.sock')
 # $ docker ...
 
 # This example requires Lima v0.7.3 or later
@@ -61,4 +61,4 @@ probes:
     hint: See "/var/log/cloud-init-output.log". in the guest
 portForwards:
   - guestSocket: "/run/user/{{.UID}}/docker.sock"
-    hostSocket: "{{.Home}}/docker.sock"
+    hostSocket: "{{.InstanceDir}}/sock/docker.sock"
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -195,11 +195,14 @@ func FillPortForwardDefaults(rule *PortForward, instDir string) {
 			home, _ := os.UserHomeDir()
 			limaHome, _ := dirnames.LimaDir()
 			data := map[string]string{
-				"Home":     home,
-				"Instance": filepath.Base(instDir),
-				"LimaHome": limaHome,
-				"UID":      user.Uid,
-				"User":     user.Username,
+				"Home":        home,
+				"InstanceDir": instDir,
+				"Name":        filepath.Base(instDir),
+				"UID":         user.Uid,
+				"User":        user.Username,
+
+				"Instance": filepath.Base(instDir), // DEPRECATED, use `{{.Name}}`
+				"LimaHome": limaHome,               // DEPRECATED, (use `InstanceDir` instead of `{{.LimaHome}}/{{.Instance}}`
 			}
 			var out bytes.Buffer
 			if err := tmpl.Execute(&out, data); err == nil {
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"path/filepath"
 	"reflect"
 	"sort"
 	"strings"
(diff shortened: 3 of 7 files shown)


jandubois:
Example: The `halog` command from https://github.com/lima-vm/lima/pull/324#issuecomment-942633294 can now use the additional filtering by instance name:

```console
$ function halog { tail -f "$(limactl ls ${1:-default} -f '{{.LimaHome}}/{{.Name}}/ha.stderr.log')"; }
$ halog docker
{"level":"info","msg":"Waiting for the optional requirement 1 of 1: \"user probe 1/1\"","time":"2021-11-29T11:58:44-08:00"}
{"level":"debug","msg":"executing script \"user probe 1/1\"","time":"2021-11-29T11:58:44-08:00"}
[...]
```

jandubois:
> ```yaml
> portForwards:
>   - guestSocket: "/run/user/{{.UID}}/docker.sock"
>     hostSocket: docker
> ```

Switching to a relative host socket name also exposed a bug in the `limactl validate` function: We must always use an absolute path to load a `lima.yaml` file because the `hostSocket` may be located relative to it, and the validation function makes sure it is absolute.

afbjorklund:
Probably want to use "docker.sock" also for the host socket, to stay consistent with "ssh.sock" and friends ?

And then use the same for "podman.sock", in the podman.yaml (again to make it more similar throughout lima)

examples/docker.yaml
```diff
@@ -61,4 +61,4 @@ probes:
     hint: See "/var/log/cloud-init-output.log". in the guest
 portForwards:
   - guestSocket: "/run/user/{{.UID}}/docker.sock"
-    hostSocket: "{{.Home}}/docker.sock"
+    hostSocket: docker
```
AkihiroSuda, line 64: "docker" may look like either a reserved word or an abstract file path in the "current" directory.

Can we rather use a form like `hostSocket: "${{.InstanceDir}}/sock/docker"` ?

afbjorklund: I think "docker.sock" is quite established by now

jandubois:
> Probably want to use "docker.sock" also for the host socket, to stay consistent with "ssh.sock" and friends ?

Ok, will do that (also to avoid the "reserved word" impression), but it is kind of redundant: The idea was that `FOO.sock` would be stored by default as `sock/FOO` inside the instance directory, but keeps the length the same. I didn't want user-forwarded sockets directly in the instance directory to avoid collision with the lima sockets (ha, ga, serial, ssh).

> Can we rather use a form like hostSocket: "${{.InstanceDir}}/sock/docker" ?

I was intentionally using the relative name to show how it defaults to `{{.LimaHome}}/{{.Instance}}/sock`. I wondering if supporting that was a mistake, and we should have always required an absolute path?

Should we "undocument" the relative path case, and changes all samples and docs to use absolute paths instead?

jandubois:
> Can we rather use a form like hostSocket: "${{.InstanceDir}}/sock/docker" ?

We didn't actually have `{{.InstanceDir}}`; you had to write `{{.LimaHome}}/{{.Instance}}`.

I have now added `{{.InstanceDir}}` to the `limactl list` fields, but not (yet) to the `hostSocket`. Ideally we should use the same field names in both places, but we already have the conflict that `{{.Instance}}` in `hostSocket` is called `{{.Name}}` in `limactl list`.

I didn't use `{{.Name}}` in `hostSocket` because it is ambiguous: it sounds like the name of the socket itself (but that doesn't make sense, as it would be a recursive reference).

My suggestion:

* remove documentation for relative `hostSocket` and the `{{.Instance}}` fields (but keep them working).
* add `{{.Name}}` and `{{.InstanceDir}}` fields to `hostSocket`.
* change samples to use `{{.InstanceDir}}/sock/FOO.sock` everywhere
* document that sockets should *not* go directly into the instance dir, but only into the `sock` subdir

Thoughts?

jandubois:
> Thoughts?

I'm going to take the absence of objections as agreement. 😄 

@AkihiroSuda PTAL

afbjorklund:
Having a variable for the dir sounds like a good idea, similar to $HOME vs /home/$USER (or something).

I like keeping the file extension on the unix socket, even if not "needed". Also makes it easier to filter out...

pkg/limayaml/defaults.go
```diff
@@ -195,11 +195,14 @@ func FillPortForwardDefaults(rule *PortForward, instDir string) {
...
-				"UID":      user.Uid,
-				"User":     user.Username,
+				"Home":        home,
+				"InstanceDir": instDir,
+				"Name":        filepath.Base(instDir),
+				"UID":         user.Uid,
+				"User":        user.Username,
+
+				"Instance": filepath.Base(instDir), // DEPRECATED, use `{{.Name}}`
+				"LimaHome": limaHome,               // DEPRECATED, (use `InstanceDir` instead of `{{.LimaHome}}/{{.Instance}}`
```
AkihiroSuda, line 205: Can we print warnings when deprecated variables are used? (Can be another PR)

AkihiroSuda (approved):

pkg/limayaml/defaults.go
jandubois: idk, seems like a lot of effort for little gain: how do you detect that the variable was used in a template expansion?

When I wrote "DEPRECATED" I just meant that we discourage people from using it (by not mentioning it in the docs/sample lima.yaml), and having the comments in the source.

But the variables will still work, and be correct, and I don't see any reason why we would ever have to remove them; they are not in the way of implementing some other feature. It is just that there should be no good reason why a user would ever need them, so there is no reason to make them aware that they exist for backwards compatibility reasons only.

So I think a warning is not really necessary.

jandubois: I already merged, but we could change the comment from "DEPRECATED" to "HIDDEN" if that sounds better?

I would just sneak it in via another PR...

PR #441, opened 2021-12-02
add example to deploy faasd

### LINKED ISSUE #440: Add an example for faasd
For the faasd project, we tend to use Multipass which now has support for preloaded examples and cloud-init.

I'd like to see someone from the Lima or OpenFaaS community contribute an example for faasd:

https://github.com/lima-vm/lima/tree/master/examples

I'd probably suggest using Ubuntu as a base image.

Here's a 3 line installation script, so I think it'd look just like one of the existing examples, plus this in the `provision:` section of the YAML file.

https://github.com/openfaas/faasd#deploy-faasd

jsiebens:
Add an example to deploy faasd

fixes #440

### DIFF
--- a/examples/faasd.yaml
+++ b/examples/faasd.yaml
@@ -1,0 +1,52 @@
+# Deploy faasd (which installs a bundled containerd).
+#
+# It can be accessed from the host by authenticating with faas-cli;
+# the ports are already forwarded automatically by lima:
+#
+# Get the faas-cli:
+# $ curl -sLS https://cli.openfaas.com | sh
+#
+# You can now log into your gateway:
+# $ export OPENFAAS_URL=http://localhost:8080
+# $ limactl shell faasd sudo cat /var/lib/faasd/secrets/basic-auth-password | faas-cli login -u admin --password-stdin
+# $
+# $ faas-cli store deploy NodeInfo
+#
+# Deployed. 200 OK.
+# URL: http://localhost:8080/function/nodeinfo
+#
+# This example requires Lima v0.7.0 or later.
+
+images:
+# Image is set to focal (20.04 LTS) for long-term stability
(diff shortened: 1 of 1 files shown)


examples/faasd.yaml
```diff
@@ -0,0 +1,69 @@
...
+#  version: 0.21.0
+#  sha:     539f0a2c946ee6096ef2b6ab26f81295eed49d54
+#
+#
+# Provider
+#  name:          faasd
+#  orchestration: containerd
+#  version:       0.14.3
+#  sha:           ea62c1b12dd1ae1de794e6dd260351cfbd1a6759
+# Your faas-cli version (0.13.13) may be out of date. Version: 0.13.15 is now available on GitHub.
```
afbjorklund, line 31: I think you want to remove the banner and the version, so that the yaml stays relevant for longer ?

jsiebens: yes, you're right, I've changed it with another example command to make it more future proof and less verbose

examples/faasd.yaml
```diff
@@ -0,0 +1,48 @@
+# Deploy faasd (which installs a bundled containerd).
+#
+# It can be accessed from the host by authenticating with faas-cli;
+# the ports are already forwarded automatically by lima:
+#
+# $ export OPENFAAS_URL=http://localhost:8080
+# $ limactl shell faasd sudo cat /var/lib/faasd/secrets/basic-auth-password | faas-cli login -u admin --password-stdin
```
AkihiroSuda, line 7: nit: Could you add how to install `faas-cli`  on the host?

jsiebens: done

AkihiroSuda (approved):
Thanks

PR #444, opened 2021-12-02
Add config command for viewing yaml config

### LINKED ISSUE #443: Add feature to show resulting config
Currently the default template is included in `limactl info`, but in a different format (JSON instead of YAML)

It would be more useful if it was shown in the _same_ format, without the comments and with values replaced...

```
limactl config [FILE.yaml...]
```

Also add a small helper function, to compare a template against the default.

This is more handy than dumping two yaml files and running `diff` on them.

```
limactl config --diff FILE.yaml...
```

afbjorklund:
Without arguments, show the parsed default config.

Add diff flag for comparing against default config.

Closes #443

Example:
```console
$ limactl config --diff examples/ubuntu.yaml 
  (
  	"""
  	arch: x86_64
  	images:
- 	- location: ~/Downloads/impish-server-cloudimg-amd64.img
- 	  arch: x86_64
- 	- location: ~/Downloads/impish-server-cloudimg-arm64.img
- 	  arch: aarch64
  	- location: https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-amd64.img
  	  arch: x86_64
  	... // 27 identical lines
  	"""
  )
```

Shows the difference between "default" and "ubuntu" (not much)

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -76,6 +76,7 @@ func newApp() *cobra.Command {
 		newListCommand(),
 		newDeleteCommand(),
 		newValidateCommand(),
+		newConfigCommand(),
 		newSudoersCommand(),
 		newPruneCommand(),
 		newHostagentCommand(),
--- a/pkg/store/store.go
+++ b/pkg/store/store.go
@@ -44,6 +44,15 @@ func InstanceDir(name string) (string, error) {
 	return dir, nil
 }
 
+// LoadYAML loads the yaml.
+func LoadYAML(b []byte) (*limayaml.LimaYAML, error) {
+	y, err := limayaml.Load(b, "")
+	if err != nil {
+		return nil, err
+	}
+	return y, nil
+}
+
 // LoadYAMLByFilePath loads and validates the yaml.
 func LoadYAMLByFilePath(filePath string) (*limayaml.LimaYAML, error) {
 	yContent, err := os.ReadFile(filePath)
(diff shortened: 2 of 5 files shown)


cmd/limactl/main.go
```diff
@@ -76,6 +76,7 @@ func newApp() *cobra.Command {
 		newListCommand(),
 		newDeleteCommand(),
 		newValidateCommand(),
+		newConfigCommand(),
```
AkihiroSuda, line 79: Maybe under a `debug` subcommand

https://github.com/lima-vm/lima/issues/443#issuecomment-985221726

pkg/store/store.go
```diff
@@ -44,6 +44,15 @@ func InstanceDir(name string) (string, error) {
...
 }
 
+// LoadYAML loads the yaml.
+func LoadYAML(b []byte) (*limayaml.LimaYAML, error) {
+	y, err := limayaml.Load(b, "")
+	if err != nil {
+		return nil, err
+	}
+	return y, nil
+}
```
AkihiroSuda, line 54: This function doesn't seem meaningful

pkg/store/store.go
```diff
@@ -59,3 +68,28 @@ func LoadYAMLByFilePath(filePath string) (*limayaml.LimaYAML, error) {
...
 }
+
+// SaveYAML saves the yaml.
+func SaveYAML(y *limayaml.LimaYAML) ([]byte, error) {
+	b, err := limayaml.Save(y)
+	if err != nil {
+		return nil, err
+	}
+	return b, nil
+}
```
AkihiroSuda, line 79: This function doesn't seem meaningful

pkg/store/store.go
afbjorklund: Ok

pkg/store/store.go
afbjorklund: Ok

cmd/limactl/main.go
afbjorklund: I don't think it is related to debugging lima, so maybe just leave it out then

PR #446, opened 2021-12-02
Add command for disk usage in home and cache

### LINKED ISSUE #445: Show disk usage of lima directories
Add a lima command similar to `du`, to show the disk usage of home and cache.

Use the standard output format, for tools like https://github.com/zz85/space-radar

`du -k ~/.lima`

Also show urls...

afbjorklund:
Closes #445

### DIFF
--- a/cmd/limactl/du.go
+++ b/cmd/limactl/du.go
@@ -1,0 +1,88 @@
+package main
+
+import (
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/docker/go-units"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func newDiskUsageCommand() *cobra.Command {
+	diskUsageCommand := &cobra.Command{
+		Use:               "du",
+		Short:             "Show lima disk usage",
+		Args:              cobra.NoArgs,
+		RunE:              diskUsageAction,
+		ValidArgsFunction: cobra.NoFileCompletions,
+	}
+	return diskUsageCommand
+}
+
+func diskUsageAction(cmd *cobra.Command, args []string) error {
+	return showCache()
+}
+
+var cache int64
+
+func showCache() error {
+	ucd, err := os.UserCacheDir()
+	if err != nil {
+		return err
+	}
+	cacheDir := filepath.Join(ucd, "lima")
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


cmd/limactl/main.go
```diff
@@ -77,6 +77,7 @@ func newApp() *cobra.Command {
 		newDeleteCommand(),
 		newValidateCommand(),
 		newSudoersCommand(),
+		newDiskUsageCommand(),
```
AkihiroSuda, line 80: Maybe under a `debug` subcommand

https://github.com/lima-vm/lima/issues/443#issuecomment-985221726

cmd/limactl/du.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/docker/go-units"
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/lima-vm/lima/pkg/store/dirnames"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func newDiskUsageCommand() *cobra.Command {
+	diskUsageCommand := &cobra.Command{
+		Use:               "du",
```
AkihiroSuda, line 19: Bikeshedding: maybe `df`

cmd/limactl/du.go
```diff
@@ -0,0 +1,158 @@
...
+		Short:             "Show lima disk usage",
+		Args:              cobra.NoArgs,
+		RunE:              diskUsageAction,
+		ValidArgsFunction: cobra.NoFileCompletions,
+	}
+	diskUsageCommand.Flags().Bool("cache", false, "Show cache usage")
+	return diskUsageCommand
```
AkihiroSuda, line 26: Cane another PR, but we should eventually have `--json`

cmd/limactl/du.go
afbjorklund: I know that Docker calls it "df", but it doesn't make sense since it doesn't show how much is free/available ?

cmd/limactl/main.go
afbjorklund: I think docker put it under `system`

afbjorklund:
You can use `du` instead, and do your own summary.

```
du -hxcs ~/.lima/*
```

```
du -hxcs ~/.cache/lima/download/by-url-sha256/*
```

Or with `du -ak`, if you want the graphical breakdown.

lima-du

PR #452, opened 2021-12-04
Add make target for running more checks

afbjorklund:
Might as well catch those locally instead

Seem to be getting the same GitHub warnings over and over, probably would be a good idea to check it locally first...

I am not sure if there is a better way to share tests, the other YAML looked pretty specific (including custom "actions")

Just don't want to waste test resources, on whitespace and typos.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,8 @@ uninstall:
 lint:
 	golangci-lint run ./...
 	yamllint .
+	find . -name '*.sh' | xargs shellcheck
+	find . -name '*.sh' | xargs shfmt -s -d
 
 .PHONY: clean
 clean:


Makefile
```diff
@@ -74,6 +74,11 @@ lint:
 	golangci-lint run ./...
 	yamllint .
 
+.PHONY: check
+check:
+	find . -name '*.sh' | xargs shellcheck
+	find . -name '*.sh' | xargs shfmt -s -d
+
```
AkihiroSuda, line 81: Can we merge this into the `lint` target?

AkihiroSuda (approved):
Thanks

Issue #458, opened 2021-12-07
Can't find docker.sock file after star limactl with docker.yaml


phuna:
Hi, 
I'm on macos 12.0.1 Intel.
I started lima with `docker.yaml` but can set `DOCKER_HOST` because I can't find `docker.sock` file. Below are details:

```
$ limactl start https://raw.githubusercontent.com/lima-vm/lima/master/examples/docker.yaml
```
Once it done, I run the below command (taking from `docker.yaml` file above) to determine path to `docker.sock` but it failed.
```
$ limactl list docker --format 'unix://{{.InstanceDir}}/sock/docker.sock'
unix://FATA[0000] template: format:1:9: executing "format" at <.InstanceDir>: can't evaluate field InstanceDir in type *store.Instance
```
Trying to look into docker sub directory of lima I can't find `docker.sock` file either:
```
$ ls -a ~/.lima/docker/sock
.		..		<no value>
```
So instead of `docker.sock`, there's a file with `<no value>` as name.
Please help. 
Thank you.

jandubois:
Did you build and install lima from the latest `master` branch of the repo?

The errors you are getting sound like you are using an older lima version (maybe the last released version) with the config file from the HEAD of repository, which will not work, as the `InstanceDir` fields, both for `hostSocket` and for `limactl list --format` have only been implemented very recently.

phuna:
@jandubois  oh, I use the Homebrew version, that could be the cause. I'll try with the `master` branch. Thank you for your quick answer.

phuna:
Haven't got time to build `master` branch myself so I use the `docker.yaml` corresponding to the current binary version (0.7.4) and I can use `docker` CLI on host now.

PR #459, opened 2021-12-07
fix port forwarding of docker socket inside docker example file

anschnapp:
## Issue i try to fix
for me it seems like the current docker.example file has some socket forwarding issues with the latest lima version
(at least that's what i have tested)

The port forward was bind to: `~/.lima/docker/sock/<no value>/sock/docker.sock`
I guess this `<no value>` is related to an unknown mapping for `{{.InstanceDir}}` which is used in the forward definition.

This assumption fits to another issue i had, when i call:
`limactl list docker --format 'unix://{{.InstanceDir}}/sock/docker.sock`

I get this kind of error message:

```
template: format:1:9: executing "format" at <.InstanceDir>: can't evaluate field InstanceDir in type *store.Instance
```

So, it seems that the variable is not defined in the current version.

## How I fixed it
it seems like the correct folder is already used for `hostSocket` if i just define the filename inside `hostSocket` section.

In addition I also adjusted the DOCKER_HOST export comment.
I define now the main lima docker configuration folder via variable `.Dir`  and then i rely on the convention (i hope this is really a convention) that the socket is in the subfolder `sock` which finally looks like this:

```
export DOCKER_HOST="`limactl list docker --format 'unix://{{.Dir}}/sock/docker.sock'`"
```

This new setup works know out of the box (for me) and from my understanding this should also work for everyone else with this adjustments.

### DIFF
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -3,7 +3,7 @@
 # $ limactl shell docker docker run -it -v $HOME:$HOME --rm alpine
 
 # To run `docker` on the host (assumes docker-cli is installed):
-# $ export DOCKER_HOST=$(limactl list docker --format 'unix://{{.InstanceDir}}/sock/docker.sock')
+# $ export DOCKER_HOST="`limactl list docker --format 'unix://{{.Dir}}/sock/docker.sock'`"
 # $ docker ...
 
 # This example requires Lima v0.7.3 or later
@@ -65,4 +65,4 @@ probes:
     hint: See "/var/log/cloud-init-output.log". in the guest
 portForwards:
   - guestSocket: "/run/user/{{.UID}}/docker.sock"
-    hostSocket: "{{.InstanceDir}}/sock/docker.sock"
+    hostSocket: "docker.sock"


jandubois:
> for me it seems like the current docker.example file has some socket forwarding issues with the latest lima version

It is not clear to me what you mean by "current" and "latest". Both the example file and the lima version should either be built from the latest commit on `master`, or use the latest release. You cannot expect the sample file from an unreleased version to be compatible with an older release. This would be a duplicate of #458. The support for `InstanceDir` was only added last week: https://github.com/lima-vm/lima/pull/439#issuecomment-982919803.

If you are using a release version, then you also should use the installed sample files, e.g.

```
limactl start /usr/local/share/doc/lima/examples/docker.yaml
```

But thank you for pointing out that `limactl list` already has a `{{.Dir}}` field, so `{{.InstanceDir}}` is redundant. We should modify `hostSocket` to use `{{.Dir}}` and drop `{{InstanceDir}}`. Since `{{.InstanceDir}}` has never been supported in a released version, we can still change this.

I can do this later today, unless you want to modify your PR to do this; please let me know.

anschnapp:
> It is not clear to me what you mean by "current" and "latest". Both the example file and the lima version should either be built from the latest commit on master, or use the latest release. You cannot expect the sample file from an unreleased version to be compatible with an older release. This would be a duplicate of #458. The support for InstanceDir was only added last week: #439 (comment).

You are totally right. I got the and update my lima installation via Homebrew but used the example file from the latest master version... Wasn't aware of this mixup but now it makes totally sense.

> But thank you for pointing out that limactl list already has a {{.Dir}} field, so {{.InstanceDir}} is redundant. We should modify hostSocket to use {{.Dir}} and drop {{InstanceDir}}. Since {{.InstanceDir}} has never been supported in a released version, we can still change this.

I tried out my current setup with `{{.Dir}}` inside the hostSocket section (instead of `{{.InstanceDir}}`) but it ended up again in a directory entry like 
```
~/.lima/docker/sock/<no value>/sock/docker.sock
```

Of course this could be also an issue with my version mixup, but it's unexpected that it works with `limactl list` but not within this section.

So, I would rather leave the field to you. It looks also for me that this `.Dir` vs `.InstanceDir` topic would most properly effect multiple files and it would be better if someone take a look at it who is really into this project.

jandubois:
> It looks also for me that this `.Dir` vs `.InstanceDir` topic would most properly effect multiple files

Yes, it does. I've opened #462 for this change.

PR #461, opened 2021-12-07
Update and simplify the podman example

afbjorklund:
- The mounting of home works OK now with crun
- The name of the instance is podman not docker
- Need to use --remote on Linux until podman v4
- Make the usage hints shorter and less scary


jandubois (approved):
LGTM

afbjorklund:
It's unfortunate that `podman` and `podman-remote` have different syntax on different platforms, but getting fixed in 4.0.

Calling different binaries has even less chance of succeeding, so the parameter is the least evil (even if unnecessary...)

Anyway, if you call just `podman` on Linux it will show the status of the local host instead of the remote Lima VM.

When using Docker you only have to change DOCKER_HOST, and with Nerdctl you don't have the luxury of remote.

PR #462, opened 2021-12-07
Replace `InstanceDir` with `Dir`

jandubois:
`Dir` was already supported by `limactl list`, so the `InstanceDir` is redundant. It was never part of a release, so can be removed.

For `hostSocket` this commit renames `InstanceDir` to `Dir`, as again there never was a release supporting `InstanceDir` either.

Updated docker and podman examples to match.

PR #466, opened 2021-12-08
Fix required version for examples/{docker,podman}.yaml

jandubois:
They use new templating variables that don't exist in older releases.

Ref https://github.com/lima-vm/lima/pull/462#pullrequestreview-826003326

### DIFF
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -6,7 +6,7 @@
 # $ export DOCKER_HOST=$(limactl list docker --format 'unix://{{.Dir}}/sock/docker.sock')
 # $ docker ...
 
-# This example requires Lima v0.7.3 or later
+# This example requires Lima v0.7.5 or later
 images:
   # Hint: run `limactl prune` to invalidate the "current" cache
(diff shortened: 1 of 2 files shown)


afbjorklund:
Can we add this to the "specification" or the "yaml", so that it is detectable ?

jandubois:
> Can we add this to the "specification" of the "yaml", so that it is detectable ?

Yes, I like that. `limactl` then also needs to know which version it is compatible with, so it can reject incompatible configs.

I think this should be a separate PR; maybe file an issue for it if you aren't going to implement it yourself. 😄

afbjorklund:
I'm a bit reluctant, since it can lead to a whole standardization commitee and whatnot. :-) Just an idea, after seeing these.

AkihiroSuda:
Having `version` property sounds good, but probably we should decouple the versioning scheme of the YAMLs from the versioning scheme of Lima implementation itself.

AkihiroSuda (approved):
Thanks

PR #470, opened 2021-12-12
Fix port forwarding of docker.sock in docker example

michaelferwer:
When we start a lima-vm with the docker example file, we get the following output in the console.

> INFO[0038] [hostagent] Forwarding "/run/user/502/docker.sock" (guest) to "/Users/my-user/.lima/docker/sock/\<no value\>/sock/docker.sock" (host)

With the modification, now we have the following output

> INFO[0030] [hostagent] Forwarding "/run/user/502/docker.sock" (guest) to "/Users/my-user/.lima/docker/sock/docker.sock" (host)

After we can export DOCKER_HOST variable :

`export DOCKER_HOST=$(limactl list docker --format 'unix://{{.Dir}}/sock/docker.sock')`

Run the command `docker version` to get Docker server information.

### DIFF
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -65,4 +65,4 @@ probes:
     hint: See "/var/log/cloud-init-output.log". in the guest
 portForwards:
   - guestSocket: "/run/user/{{.UID}}/docker.sock"
-    hostSocket: "{{.Dir}}/sock/docker.sock"
+    hostSocket: "docker.sock"


afbjorklund:
This information could also be good to move, from comment to metadata.

```yaml
message: |
  $ export DOCKER_HOST=unix://{{.Dir}}/sock/docker.sock
  $ docker ...
```

This way it will be shown to the user with the current values already in place.

```
INFO[0100] READY. Run `limactl shell docker` to open the shell. 
INFO[0100] To run `docker` on the host (assumes docker-cli is installed):
INFO[0100] $ export DOCKER_HOST=unix:///home/anders/.lima/docker/sock/docker.sock
INFO[0100] $ docker ...
```

jandubois:
This is a version mismatch between lima and the docker.yaml file. See https://github.com/lima-vm/lima/pull/459#issuecomment-988145281 for an identical report less than a week ago (which was in turn a duplicate of #458).

We need to both add version checking to the limayaml file parsing, and (maybe?) improve the docs to tell people where to locate the example configs matching their installed lima version to avoid users getting confused about this.

michaelferwer:
Hello @jandubois 

Your proposition for a documentation to find the good version for corresponding lima version will be very helpful.

PR #472, opened 2021-12-12
Add optional message to be shown to template

afbjorklund:
Show example with current values already in place.

Currently just shown as regular INFO log messages.

As mentioned in https://github.com/lima-vm/lima/pull/470#issuecomment-991951841

Example:

```
INFO[0032] [hostagent] The final requirement 1 of 1 is satisfied 
INFO[0032] READY. Run `limactl shell docker` to open the shell. 
INFO[0032] To run `docker` on the host (assumes docker-cli is installed): 
INFO[0032] $ export DOCKER_HOST=unix:///home/anders/.lima/docker/sock/docker.sock 
INFO[0032] $ docker ...                                 
```


jandubois:
I like it!

1. Needs (commented out) sample setting/documentation in `pkg/limayaml/default.yaml`.

2. Should use the `addGlobalFields` logic from `cmd/limactl/list.go` (move it to the `store` package, so it can be shared). That way we have the same set of template variables (adds `IdentityFile` and `LimaHome` for now).

pkg/limayaml/default.yaml
```diff
@@ -196,6 +196,11 @@ networks:
 #     hostPortRange: [1, 65535]
 #   # Any port still not matched by a rule will not be forwarded (ignored)
 
+# Message. Information to be shown to the user, given as a Go template for the instance.
+# The same template variables as for the socket above can be used, for instance {{.Dir}}.
```
jandubois, line 200: Technically it is the same list of fields as shown by `limactl ls --list-fields`. The fields available in the socket path are just a subset of those.

jandubois (approved):
Thanks, LGTM

pkg/store/instance.go
```diff
@@ -164,3 +168,32 @@ func ReadPIDFile(path string) (int, error) {
...
 	return pid, nil
 }
+
+type FormatData struct {
+	Instance
+	HostOS       string
+	HostArch     string
+	LimaHome     string
+	IdentityFile string
+}
```
AkihiroSuda, line 178: I think we have to consolidate this struct into what we have for sockets

https://github.com/lima-vm/lima/blob/7d3eca52b06ae052c5e09d5e6b8e113eab0b1ee9/pkg/limayaml/defaults.go#L197-L206

Having two similar but different structs is confusing.

AkihiroSuda (changes requested):
Thanks, but I think we have to consolidate the template struct into what we already have for sockets

https://github.com/lima-vm/lima/pull/472/files#r767601007

pkg/store/instance.go
afbjorklund: We consolidated the one from "limactl list", but didn't do anything about the socket templates

jandubois:
> Thanks, but I think we have to consolidate the template struct into what we already have for sockets

I don't think that works too well.

For `guestSocket` the `Home` and `User` fields have different values; the user inside the VM may be `lima` if the host user is not valid, and `Home` is `/home/$USER.linux`.

For `hostSocket` we support additional "deprecated" fields, that we don't want to show up in `limactl list --list-fields`.

If we want to consolidate, then we can only do this for `hostSocket`, but we would also have to add tags on the deprecated fields, and check for them in the reflection code that lists field names.

jandubois:
> For `hostSocket` we support additional "deprecated" fields, that we don't want to show up in `limactl list --list-fields`.

Turns out `InstanceDir` (redundant with plain `Dir`) is the only field we want to hide, as `LimaHome` is used by `limactl list`.

jandubois:
> I found another issue with architecture, it was `amd64` on the host but translated to `x86_64` in the VM.

Can you point to where it is used as `amd64` (except in `HostArch` in this PR)?

I would say that `HostArch` should be set to `x86_64` to be consistent with the values we use in `lima.yaml` (`aarch64` and `x86_64`.

I don't think we need to include the GOARCH values as well; you can use templating to translate, if you really have to:

```
curl ... tarball.{{if eq .HostArch "x86_64"}}amd64{{else})arm64{{end}}.tar.tz
```

We can always add more names later, if they seem to be useful, but we can't remove them without breaking compatibility, so I would not add more than we need. I'm fine with `HostArch` and `HostOS` though.

pkg/store/instance.go
```diff
@@ -164,3 +168,32 @@ func ReadPIDFile(path string) (int, error) {
...
+
+func AddGlobalFields(inst *Instance) (FormatData, error) {
+	var data FormatData
+	data.Instance = *inst
+	// Add HostOS
+	data.HostOS = runtime.GOOS
+	// Add HostArch
+	data.HostArch = runtime.GOARCH
```
jandubois, line 186: I think we need to translate the `GOARCH` values (`amd64`, `arm64`) here to the `Arch` values we use everywhere else (`x86_64`, `aarch64`).

jandubois (changes requested):

pkg/store/instance.go
afbjorklund: Should it use limayaml, or move the constants somewhere else ?

```
pkg/qemu/qemu.go:	nativeX8664 := arch == limayaml.X8664 && runtime.GOARCH == "amd64"
pkg/qemu/qemu.go:	nativeAARCH64 := arch == limayaml.AARCH64 && runtime.GOARCH == "arm64"
```

jandubois: Using the `limayaml` constants sounds fine to me. After all, that's what we want to be consistent with. 😄

afbjorklund: Ok, just need to export the translator and it should be done

jandubois (approved):

AkihiroSuda (approved):
Thanks

PR #475, opened 2021-12-13
Allow Proxy ports to be marshalled as strings

### LINKED ISSUE #403: limactl start error: cannot unmarshal string into Go struct field Proxies.SPNetworkDataType.Proxies.HTTPPort of type int
My lima suddenly can't be started. It was working until I did some thing stupid (can't remember exactly), could be updated docker or lima. Help needed.

Error:
```
ERRO[0000] [hostagent] json: cannot unmarshal string into Go struct field Proxies.SPNetworkDataType.Proxies.HTTPPort of type int  fields.level=fatal
FATA[0000] host agent process has exited: exit status 1
```

jandubois:
It is not clear how/why this can happen, but we have seen at least 2 reports from users where a missing port was encoded as an empty string, and not as the number 0. It is unclear if on those systems non-0 ports would be encoded as strings or numbers, so this commit will handle both.

Fixes #403

### DIFF
--- a/pkg/osutil/dns_darwin.go
+++ b/pkg/osutil/dns_darwin.go
@@ -29,12 +29,14 @@ func DNSAddresses() ([]string, error) {
 	return addresses, nil
 }
 
-func proxyURL(proxy string, port int) string {
+func proxyURL(proxy string, port interface{}) string {
 	if !strings.Contains(proxy, "://") {
 		proxy = "http://" + proxy
 	}
-	if port != 0 {
-		proxy = fmt.Sprintf("%s:%d", proxy, port)
+	if portNumber, ok := port.(float64); ok && portNumber != 0 {
+		proxy = fmt.Sprintf("%s:%.0f", proxy, portNumber)
+	} else if portString, ok := port.(string); ok && portString != "" {
+		proxy = fmt.Sprintf("%s:%s", proxy, portString)
 	}
 	return proxy
 }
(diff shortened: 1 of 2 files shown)


pkg/osutil/dns_darwin.go
```diff
@@ -29,12 +29,14 @@ func DNSAddresses() ([]string, error) {
...
 }
 
-func proxyURL(proxy string, port int) string {
+func proxyURL(proxy string, port interface{}) string {
 	if !strings.Contains(proxy, "://") {
 		proxy = "http://" + proxy
 	}
-	if port != 0 {
-		proxy = fmt.Sprintf("%s:%d", proxy, port)
+	if portNumber, ok := port.(float64); ok && portNumber != 0 {
```
AkihiroSuda, line 36: Is there any chance that we may get `int` too?

Also, can we print an error when we get an unexpected type?

jandubois: > Is there any chance that we may get `int` too?

No, there isn't (or shouldn't). From the `json` package docs:

>To unmarshal JSON into an interface value, Unmarshal stores one of these in the interface value:
>
> ```
> bool, for JSON booleans
> float64, for JSON numbers
> string, for JSON strings
> []interface{}, for JSON arrays
> map[string]interface{}, for JSON objects
> nil for JSON null
> ```

>Also, can we print an error when we get an unexpected type?

We could, but given the possible types, I feel like this cannot really happen. Also, the best we could do is call `logrus`.

AkihiroSuda (approved):
Thanks

PR #486, opened 2021-12-16
Only set the prompt color on a dark terminal background

### LINKED ISSUE #485: Don't set the prompt color when running in light terminal theme
As noted in https://github.com/lima-vm/lima/issues/427#issuecomment-979440137, there are some usability issues with lime-on-white.

lima-light

It should also be possible to opt out of it (color), even for the default instance.

Use a better library, than the proposed port:

https://github.com/muesli/termenv

----

* Issue #427

afbjorklund:
The lima color has usability issues in light theme

Closes #485


jandubois:
I thought I had already written this for the previous PR, but cannot find the comment again:

I feel using `patch` to make so many complex changes to `.bashrc` is the wrong approach, and not maintenance friendly (updating the patch will be error-prone if/when the upstream source is going to change).

I would rather see this change appended as a single block at the end of `.bashrc`, or maybe even in a separate file, so it is just called from it. We should not modify the upstream logic, but just determine the conditions under which we want to modify the prompt, and then just overwrite the setting that has already been setup earlier. That way our logic should become much simpler (I hope).

I also feel that we are trying to support too many different configurations.

afbjorklund:
I think the plan was to have the patch fail, if/when the original ever changes...

There is a `unset color_prompt force_color_prompt` which makes it harder.

afbjorklund:
> I also feel that we are trying to support too many different configurations.

Currently there are 4 configurations: Ascii, ANSI, ANSI256, TrueColor

afbjorklund:
Could revert the lima color, and go back to changing the prompt instead...

That `[LIMA] `  would make it work in monochrome as well: https://github.com/lima-vm/lima/pull/433#issuecomment-978881137

afbjorklund:
Who uses light themes anyway ;-)

afbjorklund:
New cyan color works in both.

Issue #355, opened 2021-10-20
Run qemu-guest-agent to keep time in sync with host?


jandubois:
There has been a question about containers falling behind in time when the host has been suspended: https://github.com/rancher-sandbox/rancher-desktop/discussions/836

I believe this would be fixed by installing the `qemu-guest-agent`, but haven't tested it (because I'm not sure how to set it up, but I've only looked at Alpine so far).

totallyGreg:
I would like to be be able to do this so shutdowns are communicated properly and various stats are provided.  I've had success installing the qemu-guest-agent in a k3os image running on proxmox, by adding this to the cloud-config file. 
```
write_files:
- path: /etc/conf.d/qemu-guest-agent
  content: |-
    # Specifies the transport method used to communicate to QEMU on the host side
    # Default: virtio-serial
    #GA_METHOD="virtio-serial"
    # Specifies the device path for the communications back to QEMU on the host
    # Default: /dev/virtio-ports/org.qemu.guest_agent.0
    #GA_PATH="/dev/virtio-ports/org.qemu.guest_agent.0"
      GA_PATH=/dev/vport2p1
  owner: root
  permissions: '0644'
```
I am not sure how lima is communicating with qemu but if the device port  path is known I can try it out.

ko185061:
Im putting a little time towards trying some things with QEMU, I will update the thread if I figure anything out!

dmccaffery:
@AkihiroSuda - I looked into this and noticed a few things:

1. The qemu-guest-agent socket (virtio-serial) is not configured -- this is easy to resolve
2. There is no detection of guest pause/resume from the host

I'm happy to implement support for setting the clock but am not sure what the best way to go about it is within the design of this project. Should we implement settimeofday in the lima guest agent with a signal from the host agent? -- this removes the need for `yet-another-socket` (qga) to be set -- but it does require reinventing the wheel re: setting the time and the hwclock.

Alternatively, we can enable the qemu-guest-agent:

```diff
diff --git a/pkg/qemu/qemu.go b/pkg/qemu/qemu.go
index 71d633b..03a9c7f 100644
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -345,6 +345,17 @@ func Cmdline(cfg Config) (string, []string, error) {
        args = append(args, "-chardev", fmt.Sprintf("socket,id=%s,path=%s,server=on,wait=off,logfile=%s", serialChardev, serialSock, serialLog))
[...]

PR #490, opened 2021-12-21
Add hwclock sync loop to the guestagent

mikluko:
Addresses #355

Running qemu guest agent alone doesn't solve the problem. With that approach, there must be an actor on the host side which catches wakeup events and executes RPC calls on the guest agent socket. As it stands, lima doesn't have any permanently running process to relay wake up events.

This PR offers different approach. It turns out that hardware clock device (`/dev/rtc0`) forwarded into a guest by QEMU always contains correct host time. The PR adds a loop into lima guest agent. That loop periodically checks the difference between hw clock time and system time. If the diff turns out to be over certain limit (hardcoded 5 seconds for now), the loop launches time sync from hw clock to system time.

### DIFF
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -250,26 +250,27 @@ func (a *agent) Info(ctx context.Context) (*api.Info, error) {
 const deltaLimit = 2 * time.Second
 
 func (a *agent) fixSystemTimeSkew() {
-	ticker := time.NewTicker(10 * time.Second)
-	defer ticker.Stop()
-	for now := range ticker.C {
-		rtc, err := timesync.GetRTCTime()
-		if err != nil {
-			logrus.Warnf("fixSystemTimeSkew: lookup error: %s", err.Error())
-			continue
-		}
-		d := rtc.Sub(now)
-		logrus.Debugf("fixSystemTimeSkew: rtc=%s systime=%s delta=%s",
-			rtc.Format(time.RFC3339), now.Format(time.RFC3339), d)
-		if d > deltaLimit || d < -deltaLimit {
-			err = timesync.SetSystemTime(rtc)
+	for {
+		ticker := time.NewTicker(10 * time.Second)
+		for now := range ticker.C {
+			rtc, err := timesync.GetRTCTime()
 			if err != nil {
-				logrus.Warnf("fixSystemTimeSkew: set system clock error: %s", err.Error())
+				logrus.Warnf("fixSystemTimeSkew: lookup error: %s", err.Error())
 				continue
 			}
-			logrus.Infof("fixSystemTimeSkew: system time synchronized with rtc")
-			break
+			d := rtc.Sub(now)
+			logrus.Debugf("fixSystemTimeSkew: rtc=%s systime=%s delta=%s",
+				rtc.Format(time.RFC3339), now.Format(time.RFC3339), d)
+			if d > deltaLimit || d < -deltaLimit {
+				err = timesync.SetSystemTime(rtc)
+				if err != nil {
+					logrus.Warnf("fixSystemTimeSkew: set system clock error: %s", err.Error())
+					continue
+				}
+				logrus.Infof("fixSystemTimeSkew: system time synchronized with rtc")
+				break
+			}
 		}
+		ticker.Stop()
 	}
-	go a.fixSystemTimeSkew()
 }
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -13,6 +13,7 @@ import (
 	"github.com/lima-vm/lima/pkg/guestagent/api"
 	"github.com/lima-vm/lima/pkg/guestagent/iptables"
 	"github.com/lima-vm/lima/pkg/guestagent/procnettcp"
+	"github.com/lima-vm/lima/pkg/guestagent/timesync"
 	"github.com/sirupsen/logrus"
 	"github.com/yalue/native_endian"
 )
@@ -37,6 +38,7 @@ func New(newTicker func() (<-chan time.Time, func()), iptablesIdle time.Duration
 	}
 
 	go a.setWorthCheckingIPTablesRoutine(auditClient, iptablesIdle)
+	go a.fixSystemTimeSkew()
 	return a, nil
 }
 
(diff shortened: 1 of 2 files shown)


pkg/guestagent/hwclock/hwclock_linux.go
```diff
@@ -0,0 +1,23 @@
...
+	err = cmd.Run()
+	if err != nil {
+		return
+	}
+	t, err = time.Parse("2006-01-02 15:04:05.999999-07:00", strings.TrimSpace(cmd.Stdout.(*bytes.Buffer).String()))
+	return
+}
+
+func HcToSys() error {
+	return exec.Command("bash", "-c", "hwclock --hctosys").Run()
```
AkihiroSuda, line 22: No need to use bash

pkg/guestagent/hwclock/hwclock_linux.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"bytes"
+	"os/exec"
+	"strings"
+	"time"
+)
+
+func Get() (t time.Time, err error) {
+	cmd := exec.Command("hwclock")
```
AkihiroSuda, line 11: Can we directly call ioctl?

https://man7.org/linux/man-pages/man4/rtc.4.html

pkg/guestagent/hwclock/hwclock_linux.go
mikluko: Unfortunately, bash is required here. hwclock doesn't reset system clock without environment set.

pkg/guestagent/hwclock/hwclock_linux.go
mikluko: Good point. I will give it a try.

pkg/guestagent/hwclock/hwclock_linux.go
AkihiroSuda: What environment is needed?
Setting `cmd.Env` should work?

pkg/guestagent/hwclock/hwclock_linux.go
jandubois: @mikluko Do you need any help? I'm interested in getting this merged within the next week or two, so if there is anything I can do to help this move forward, please let me know!

pkg/guestagent/guestagent_linux.go
```diff
@@ -244,3 +246,30 @@ func (a *agent) Info(ctx context.Context) (*api.Info, error) {
 	}
 	return &info, nil
 }
+
+const deltaLimit = time.Second * 5
```
jandubois, line 250: Not sure what the right value is, but unless the price for adjusting the time is high, I would set the threshold to 2 seconds.

And this is just personal preference, but I think `2 * time.Second` reads nicer than `time.Second * 2` because it follows the standard form of "numerical value" followed by "unit".

pkg/guestagent/guestagent_linux.go
```diff
@@ -244,3 +246,30 @@ func (a *agent) Info(ctx context.Context) (*api.Info, error) {
...
+	for now := range ticker.C {
+		rtc, err := timesync.GetRTCTime()
+		if err != nil {
+			logrus.Warnf("fixSystemTimeSkew: lookup error: %s", err.Error())
+			continue
+		}
+		d := rtc.Sub(now)
+		logrus.Debugf("fixSystemTimeSkew: rtc=%s systime=%s delta=%s",
+			rtc.Format(time.RFC3339), now.Format(time.RFC3339), d)
+		if (d > 0 && d > deltaLimit) || (d < 0 && d < -deltaLimit) {
```
jandubois, line 264: Aren't the comparisons with `0` redundant?

```suggestion
		if d > deltaLimit || d < -deltaLimit {
```

I'm also slightly concerned about adjusting the time backwards, as that can create subtle problems, but in practice I can only see the guest time lagging behind the host time, so it doesn't really matter.

jandubois (dismissed):
LGTM, and worked in my testing.

I do think the threshold should be lower, but that is more a gut feeling without supporting data.

pkg/guestagent/guestagent_linux.go
mikluko: I chose 5 seconds because I had never saw an evidence that skew below that might cause any problems. 

I have no benchmark data at hand, but to my best knowledge ioctl calls here are relatively cheap. Furthermore, tuning that delta won't really affect the number of update calls much. Updating according to your suggestion.

jandubois (dismissed):
Thank you, LGTM. Leaving for @AkihiroSuda to merge.

pkg/guestagent/guestagent_linux.go
```diff
@@ -244,3 +246,30 @@ func (a *agent) Info(ctx context.Context) (*api.Info, error) {
...
+			err = timesync.SetSystemTime(rtc)
+			if err != nil {
+				logrus.Warnf("fixSystemTimeSkew: set system clock error: %s", err.Error())
+				continue
+			}
+			logrus.Infof("fixSystemTimeSkew: system time synchronized with rtc")
+			break
+		}
+	}
+	go a.fixSystemTimeSkew()
```
AkihiroSuda, line 274: What's this recursion?

jandubois: I assumed this is due to a concern that the time adjustment could cause the ticker not to trigger again. E.g. when you move back the clock by one hour, will the next tick happen in 10s or 1h and 10s?

So any time the clock is adjusted, the current goroutine exits and a new instance sets up a new ticker based on the current clock settings.

I don't know if this is necessary, and I don't think the clock will be adjusted backwards, but it seems like sensible defensive coding to me.

AkihiroSuda: Doesn't this cause infinite recursion?

jandubois: > Doesn't this cause infinite recursion?

Depends what you mean by that. It is an endless loop, but there are only at most 2 instances at any point in time: the routine loops forever, unless there is a time delta. If there is no, it adjusts the time, starts a new instance of itself, and then stops, which in turn will also remove the current ticker.

In the end, there is just a single check every 10s to see if there is time drift. If there is time drift every single time, then the time will be adjusted every 10s and the job passed on to the next instance. But conceptually this isn't different from an endless loop.

I guess given the confusion this is causing, using a nested loop for this would be better (but functionally equivalent):

```
func timeskew {
    forever {
        every 10 seconds {
            if time drift {
                adjust time
                break
            }
        }
    }
}
```

The current implementation is using "recursion" instead of the outer loop:

```
func timeskew {
    every 10 seconds {
        if time drift {
            adjust time
            break
        }
    }
    call timeskew // with tailcall recursion optimization
}
```

jandubois: Since the current implementation uses goroutines, it is not really recursion because the old instance terminates just as the new instance starts. Which also triggers the `defer` statements to `Stop` the old ticker, so there is no resource leakage.

jandubois: I've not tested it, but this would be the version without "recursion":

```diff
$ g diff -b
diff --git pkg/guestagent/guestagent_linux.go pkg/guestagent/guestagent_linux.go
index b4721b5..22e1e31 100644
--- pkg/guestagent/guestagent_linux.go
+++ pkg/guestagent/guestagent_linux.go
@@ -250,8 +250,8 @@ func (a *agent) Info(ctx context.Context) (*api.Info, error) {
 const deltaLimit = 2 * time.Second

 func (a *agent) fixSystemTimeSkew() {
+       for {
                ticker := time.NewTicker(10 * time.Second)
-       defer ticker.Stop()
                for now := range ticker.C {
                        rtc, err := timesync.GetRTCTime()
                        if err != nil {
@@ -271,5 +271,6 @@ func (a *agent) fixSystemTimeSkew() {
                                break
                        }
                }
-       go a.fixSystemTimeSkew()
+               ticker.Stop()
+       }
 }
```

jandubois: @AkihiroSuda Having looked at both versions, I would agree that the nested loop is easier to understand, so preferable, but I leave it up to you to request the change.

AkihiroSuda: Thanks for explanation, I prefer to request the change to remove "recursion"

AkihiroSuda (changes requested):
Please remove this "recursion"
https://github.com/lima-vm/lima/pull/490#discussion_r781746271

pkg/guestagent/guestagent_linux.go
mikluko: Frankly, I don't see how nested loops make code more readable. But ok.

AkihiroSuda (approved):
Thanks

PR #494, opened 2021-12-24
Fix #471: Improve x86_64 emulation on arm64 platform

### LINKED ISSUE #471: Improve x86_64 emulation on arm64 platform
The current x86_64 emulation try to use the apple-hypervisor as the accelerator which only use 1 cpu on the host machine regardless how much cpu core are assigned.

But changing the accelerator parameter from apple hypervisor to tcg with few more extra parameters which UTM use when emulating improved the emulation performance up to the point docker qemu userspace emulation.

gayanper:
This change will use qemu64 cpu for better emulation
and add some extra parameters like tcg accel and
force multi threading where multiple host threads will
be used for vcpus according to qemu documentation.
Added few more extra parameters which are used by
UTM for better emulation of x86_64

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -208,12 +208,24 @@ func Cmdline(cfg Config) (string, []string, error) {
 	}
 	switch *y.Arch {
 	case limayaml.X8664:
-		cpu := "Haswell-v4"
+		cpu := "qemu64"
 		if isNativeArch(*y.Arch) {
 			cpu = "host"
 		}
 		args = appendArgsIfNoConflict(args, "-cpu", cpu)
-		args = appendArgsIfNoConflict(args, "-machine", "q35,accel="+accel)
+		if isNativeArch(*y.Arch) {
+			args = appendArgsIfNoConflict(args, "-machine", "q35,accel="+accel)
+		} else {
+			// use q35 machine with vmware io port disabled.
+			args = appendArgsIfNoConflict(args, "-machine", "q35,vmport=off")
+			// use tcg accelerator with multi threading with 512MB translation block size
+			// https://qemu-project.gitlab.io/qemu/devel/multi-thread-tcg.html?highlight=tcg
+			// https://qemu-project.gitlab.io/qemu/system/invocation.html?highlight=tcg%20opts
+			// this will make sure each vCPU will be backed by 1 host user thread.
+			args = appendArgsIfNoConflict(args, "-accel", "tcg,thread=multi,tb-size=512")
+			// This will disable CPU S3 state.
+			args = append(args, "-global", "ICH9-LPC.disable_s3=1")
+		}
 	case limayaml.AARCH64:
 		cpu := "cortex-a72"
 		if isNativeArch(*y.Arch) {


pkg/qemu/qemu.go
```diff
@@ -208,12 +210,19 @@ func Cmdline(cfg Config) (string, []string, error) {
...
 		args = appendArgsIfNoConflict(args, "-cpu", cpu)
-		args = appendArgsIfNoConflict(args, "-machine", "q35,accel="+accel)
+		if isNativeArch(*y.Arch) {
+			args = appendArgsIfNoConflict(args, "-machine", "q35,accel="+accel)
+		} else {
+			args = appendArgsIfNoConflict(args, "-machine", "q35,vmport=off")
+			args = appendArgsIfNoConflict(args, "-accel", "tcg,thread=multi,tb-size=512")
+			args = appendArgsIfNoConflict(args, "-global", "PIIX4_PM.disable_s3=1")
+			args = appendArgsIfNoConflict(args, "-global", "ICH9-LPC.disable_s3=1")
+		}
```
AkihiroSuda, line 225: Could you add comment lines to explain these flags?

pkg/qemu/qemu.go
```diff
@@ -134,9 +134,11 @@ func appendArgsIfNoConflict(args []string, k, v string) []string {
 		return append(args, k)
 	}
 
-	if origV, ok := argValue(args, k); ok {
-		logrus.Warnf("Not adding QEMU argument %q %q, as it conflicts with %q %q", k, v, k, origV)
-		return args
+	if k != "-global" {
```
AkihiroSuda, line 137: Why this check?

gayanper: Since we need to allow multiple -global arguments which doesn't seems to conflict according to UTM.

AkihiroSuda: Then you can use `args = append(args, "-global", ...)` without using `appendArgsIfNoConflict`

pkg/qemu/qemu.go
```diff
@@ -208,12 +210,24 @@ func Cmdline(cfg Config) (string, []string, error) {
...
+			args = appendArgsIfNoConflict(args, "-machine", "q35,accel="+accel)
+		} else {
+			// use q35 machine with vmware io port disabled.
+			args = appendArgsIfNoConflict(args, "-machine", "q35,vmport=off")
+			// use tcg accelerator with multi threading with 512MB translation block size
+			// https://qemu-project.gitlab.io/qemu/devel/multi-thread-tcg.html?highlight=tcg
+			// https://qemu-project.gitlab.io/qemu/system/invocation.html?highlight=tcg%20opts
+			// this will make sure each vCPU will be backed by 1 host user thread.
+			args = appendArgsIfNoConflict(args, "-accel", "tcg,thread=multi,tb-size=512")
+			// This will disable CPU S3 state.
```
AkihiroSuda, line 228: For what purpose?

pkg/qemu/qemu.go
```diff
@@ -208,12 +210,24 @@ func Cmdline(cfg Config) (string, []string, error) {
...
+			args = appendArgsIfNoConflict(args, "-global", "ICH9-LPC.disable_s3=1")
```
AkihiroSuda, line 229:
```suggestion
			args = append(args, "-global", "ICH9-LPC.disable_s3=1")
```

pkg/qemu/qemu.go
gayanper: When running emulations I found that the VM doesn't use more than 125%.  Which seems that nor more than 1 thread is used from host vm for emulation. But adding this tcg accel cause the VM to use more host threads (CPUs). After this change when running the VM load i found the CPU usage spike up to 300% - 400% on the same apple silicon machine. I found this when comparing the load test on the VM created using UTM App and Lima. The last s3 state toggle might not effect the performance directly. But may be it will help the VM to run without going into s3 state.

AkihiroSuda (approved):
Thanks

pkg/qemu/qemu.go
```diff
@@ -208,12 +208,24 @@ func Cmdline(cfg Config) (string, []string, error) {
...
+		} else {
+			// use q35 machine with vmware io port disabled.
+			args = appendArgsIfNoConflict(args, "-machine", "q35,vmport=off")
+			// use tcg accelerator with multi threading with 512MB translation block size
+			// https://qemu-project.gitlab.io/qemu/devel/multi-thread-tcg.html?highlight=tcg
+			// https://qemu-project.gitlab.io/qemu/system/invocation.html?highlight=tcg%20opts
+			// this will make sure each vCPU will be backed by 1 host user thread.
+			args = appendArgsIfNoConflict(args, "-accel", "tcg,thread=multi,tb-size=512")
+			// This will disable CPU S3 state.
+			args = append(args, "-global", "ICH9-LPC.disable_s3=1")
```
jandubois, line 227: Is there any reason you don't also disable the S4 state?

gayanper: s4 state is for hibernate if i'm not mistaken. It was not part of the UTM parameter list and also i thought that the hibernate state is not something that will the VM automatically go into. Thats why i didn't thought of adding it.

AkihiroSuda:
https://www.qemu.org/docs/master/system/i386/cpu.html

qemu64 does not support x86_64-v3 needed by RHEL 10.
AlmaLinux 10 supports v2, but qemu64 does not seem to even support v2.

Can we switch back to Haswell-v4 or something newer?
- https://github.com/lima-vm/lima/issues/3063

PR #497, opened 2021-12-26
XHCI over EHCI for better performance as well as aarch64 support

byepolr:
Update to use a more friendly version USB that works with aarch64

Notes:
https://qemu-project.gitlab.io/qemu/system/devices/usb.html
```
XHCI controller support
QEMU has XHCI host adapter support. The XHCI hardware design is much more virtualization-friendly when compared to EHCI and UHCI, thus XHCI emulation uses less resources (especially CPU). So if your guest supports XHCI (which should be the case for any operating system released around 2010 or later) we recommend using 
```

Signed-off-by: Jason W. Ehrlich <jwehrlich@outlook.com>

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -324,9 +324,9 @@ func Cmdline(cfg Config) (string, []string, error) {
 	default:
 		// QEMU does not seem to support virtio-vga for aarch64
 		args = append(args, "-vga", "none", "-device", "ramfb")
-		args = append(args, "-device", "usb-ehci")
-		args = append(args, "-device", "usb-kbd")
-		args = append(args, "-device", "usb-mouse")
+		args = append(args, "-device", "qemu-xhci,id=usb-bus")
+		args = append(args, "-device", "usb-kbd,bus=usb-bus.0")
+		args = append(args, "-device", "usb-mouse,bus=usb-bus.0")
 	}
 
 	// Parallel


byepolr:
Resolves https://github.com/lima-vm/lima/issues/499

pkg/qemu/qemu.go
```diff
@@ -324,9 +324,9 @@ func Cmdline(cfg Config) (string, []string, error) {
 	default:
 		// QEMU does not seem to support virtio-vga for aarch64
 		args = append(args, "-vga", "none", "-device", "ramfb")
-		args = append(args, "-device", "usb-ehci")
-		args = append(args, "-device", "usb-kbd")
-		args = append(args, "-device", "usb-mouse")
+		args = append(args, "-device", "qemu-xhci,id=usb-bus")
```
AkihiroSuda, line 327: This requires QEMU >= 2.9.
I guess this is fine, but cc-ing @jandubois FYI
https://github.com/qemu/qemu/commit/72a810f411abaabc55f375533220adf69e059c89

AkihiroSuda (approved):
Thanks

pkg/qemu/qemu.go
jandubois: > I guess this is fine, but cc-ing @jandubois FYI

Yes, seems fine, and works for me.

But as I mentioned in #499, I cannot repro the failure on macOS on aarch64 with the `default.yaml` config.

And I was also wondering why we need USB devices at all? What happens if we just drop them?

AkihiroSuda: > And I was also wondering why we need USB devices at all? What happens if we just drop them?

We will lose keyboard and mouse. Probably negligible for non-GUI mode though.

jandubois: >  Probably negligible for non-GUI mode though.

Do we have a GUI mode? I know we can get a console (with horrible performance impact), but is there a way to actually run a GUI?

On the other hand, I would hope that performance impact from emulating a USB device that is not being used is really minimal as well (really only thinking about battery impact when running on a laptop).

AkihiroSuda: By "GUI" I just meant setting `.video.display = "cocoa"`

jandubois: > By "GUI" I just meant setting .video.display = "cocoa"

So maybe we drop the USB devices with the `display: none` setting. I may try that when I'm bored. 😄

PR #507, opened 2021-12-31
remove $(DEST)/bin/nerdctl only when it is a symlink to nerdctl.lima

20100074:
This patch removes $(DEST)/bin/nerdctl only when it is a symlink to nerdctl.lima.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,7 +67,7 @@ uninstall:
 		"$(DEST)/bin/limactl" \
 		"$(DEST)/bin/nerdctl.lima" \
 		"$(DEST)/share/lima" "$(DEST)/share/doc/lima"
-	# TODO: remove $(DEST)/bin/nerdctl only when it is a symlink to nerdctl.lima
+		if [ "$$(readlink "$(DEST)/bin/nerdctl")" = "nerdctl.lima" ]; then rm "$(DEST)/bin/nerdctl"; fi
 
 .PHONY: lint
 lint:


Makefile
```diff
@@ -65,9 +65,12 @@ uninstall:
 	rm -rf \
 		"$(DEST)/bin/lima" \
 		"$(DEST)/bin/limactl" \
-		"$(DEST)/bin/nerdctl.lima" \
```
AkihiroSuda, line 68: No need to change this line

20100074: Thanks for the comment. Fixed.

Makefile
```diff
@@ -67,7 +67,11 @@ uninstall:
 		"$(DEST)/bin/limactl" \
 		"$(DEST)/bin/nerdctl.lima" \
 		"$(DEST)/share/lima" "$(DEST)/share/doc/lima"
-	# TODO: remove $(DEST)/bin/nerdctl only when it is a symlink to nerdctl.lima
+	@test -f "$(DEST)/bin/nerdctl.lima" && \
```
jandubois, line 70: This condition is always false because you just deleted the file in the line above. Maybe it should be:

```suggestion
	@test -L "$(DEST)/bin/nerdctl" && \
```

Makefile
```diff
@@ -67,7 +67,11 @@ uninstall:
...
+		(if [ "$(shell readlink $(DEST)/bin/nerdctl)" = "$(DEST)/bin/nerdctl.lima" ]; then \
+			unlink "$(DEST)/bin/nerdctl"; \
+			unlink "$(DEST)/bin/nerdctl.lima"; \
```
jandubois, line 73: This file has already been removed above.

jandubois (changes requested):

Makefile
20100074: Thanks for the comment. I didn't know `-L` option. Fixed.

Makefile
```diff
@@ -67,7 +67,10 @@ uninstall:
 		"$(DEST)/bin/limactl" \
 		"$(DEST)/bin/nerdctl.lima" \
 		"$(DEST)/share/lima" "$(DEST)/share/doc/lima"
-	# TODO: remove $(DEST)/bin/nerdctl only when it is a symlink to nerdctl.lima
+	@test -L "$(DEST)/bin/nerdctl" && \
+		(if [ "$(shell readlink $(DEST)/bin/nerdctl)" = "$(DEST)/bin/nerdctl.lima" ]; then \
+			unlink "$(DEST)/bin/nerdctl"; \
+		fi)
```
jandubois, line 73: I don't think this code works because the symlink value is just `nerdctl.lima` and does not include the absolute path.

I'm also somewhat bothered by the fact that the `$(shell  readlink ...)` call is executed while the Makefile is being expanded while the check that the file is a symlink only happens during command execution, i.e. afterwards.

But since `readlink` just returns the empty string when the target is not a symlink, we don't really need that test at all, and the expression could be simply:

```suggestion
	if [ "$$(readlink "$(DEST)/bin/nerdctl")" = "nerdctl.lima" ]; then rm "$(DEST)/bin/nerdctl"; fi
```

20100074: Thanks for the additional comment. Fixed.

jandubois (approved):
LGTM

PR #508, opened 2022-01-01
Replace localhost in proxy settings with gateway address

jandubois:
Previously this replacement was only applied to values from the current process environment. Now it also applies to system proxy settings, and values from the `env` section of `lima.yaml`.


AkihiroSuda (approved):
Thanks

PR #509, opened 2022-01-01
Don't include default settings in example configs

jandubois:
Setting the fields to `null` instead makes them overridable via the `_config/default.yaml` mechanism.

For example, I want to make the home directory writable by default. Right now the only mechanism is:

```console
$ cat ~/.lima/_config/override.yaml
mounts:
- location: "~"
  writable: true
```

But that means it is now impossible to make the home directory read-only in a specific VM; `override.yaml` always takes precedence. With the changes in this PR, the setting above can be moved to `default.yaml`, and can still be overridden by individual VMs.

### DIFF
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -2,9 +2,13 @@
 # BASIC CONFIGURATION
 # ===================================================================== #
 
+# Default values are specified by `null` instead of the builtin default value,
+# so they can be overridden by the default.yaml mechanism documented at the
+# end of this file.
+
 # Arch: "default", "x86_64", "aarch64".
-# "default" corresponds to the host architecture.
-arch: "default"
+# Default: "default" (corresponds to the host architecture)
+arch: null
 
 # An image must support systemd and cloud-init.
 # Ubuntu and Fedora are known to work.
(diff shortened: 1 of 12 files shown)


pkg/limayaml/default.yaml
```diff
@@ -25,34 +29,34 @@ images:
...
 # Default: none
 mounts:
   - location: "~"
     # CAUTION: `writable` SHOULD be false for the home directory.
     # Setting `writable` to true is possible, but untested and dangerous.
-    writable: false
+    writable: null
     # SSHFS has an optional flag called 'follow_symlinks'. This allows mounts
     # to be properly resolved in the guest os and allow for access to the
     # contents of the symlnk. This defaults to false if not supplied. As a result,
```
AkihiroSuda, line 51: Can we add “Default: false” line?
The default value is already explained in the existing comment lines, but the format is not consistent with other properties, so hard to find.

pkg/limayaml/default.yaml
```diff
@@ -25,34 +29,34 @@ images:
...
-memory: "4GiB"
+memory: null
 
 # Disk size
 # Default: "100GiB"
-disk: "100GiB"
+disk: null
 
 # Expose host directories to the guest, the mount point might be accessible from all UIDs in the guest
 # Default: none
```
AkihiroSuda, line 43: “Default: null”

AkihiroSuda (approved):
Thanks

PR #517, opened 2022-01-02
Don't set localPort in example configs

### LINKED ISSUE #511: Remove explicit localPort setting from docker, podman, and singularity examples
These are the only examples that still set a fixed port number instead of letting lima pick an available one:

```console
$ grep -r localPort examples
examples/default.yaml:  localPort: 0
examples/default.yaml:# Port forwarding rules. Forwarding between ports 22 and ssh.localPort cannot be overridden.
examples/docker.yaml:  localPort: 60006
examples/podman.yaml:  localPort: 60906
examples/singularity.yaml:  localPort: 62045
```

I believe this is a left-over from the time before we had unix socket forwarding (or the `limactl show-ssh` command), so access via a fixed port was simpler.

I think this is no longer necessary, and we should remove the port assignments.

In addition, I don't think `examples/docker.yaml` should set `loadDotSSHPubKeys: true` either anymore.

cc/ @afbjorklund

jandubois:
lima will automatically pick an unused port.

Fixes #511

### DIFF
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -17,12 +17,6 @@ mounts:
   - location: "~"
   - location: "/tmp/lima"
     writable: true
-ssh:
-  localPort: 60006
-  # Load ~/.ssh/*.pub in addition to $LIMA_HOME/_config/user.pub , for allowing DOCKER_HOST=ssh:// .
-  # This option is enabled by default.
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks

PR #513, opened 2022-01-01
Running kubectl requires using sudo kubectl

afbjorklund:
A bit of mismatch between k3s and k8s, and needs better documentation...

Make it easier to set up user config, and make `sudo kubectl` work as well.

i.e. running directly on the guest, is _supposed_ to require sudo

running remotely from the host, is configured to **not** use sudo

----

k3s, when not using `sudo`
```
WARN[0000] Unable to read /etc/rancher/k3s/k3s.yaml, please start server with --write-kubeconfig-mode to modify kube config permissions 
error: error loading config file "/etc/rancher/k3s/k3s.yaml": open /etc/rancher/k3s/k3s.yaml: permission denied
```

k8s, without $KUBECONFIG
```
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

PR #518, opened 2022-01-02
Add LIMA_CIDATA_SLIRP_IP_ADDRESS setting to lima.env

jandubois:
So it can be used by system provisioning scripts. User provisioning scripts don't have these sourced yet, but we should probably change that.

Would be useful for #513


AkihiroSuda (approved):
Thanks

PR #519, opened 2022-01-02
Fix indentation

jandubois:
I only noticed that #507 was not indented correctly after I already merged it. :(


AkihiroSuda (approved):
Thanks

PR #523, opened 2022-01-03
Use "everyone" group instead of "staff"

jandubois:
The "staff" group only includes local accounts whereas "everyone" also includes accounts from directory services.

See also https://github.com/rancher-sandbox/rancher-desktop/issues/1152


AkihiroSuda (approved):
Thanks

Issue #514, opened 2022-01-01
Add support for setting up configuration files automatically


afbjorklund:
Currently we hide the instructions in the YAML file, on how to set up the cluster access.

```yaml
# $ export KUBECONFIG=$PWD/kubeconfig.yaml
# $ limactl shell k3s sudo cat /etc/rancher/k3s/k3s.yaml >$KUBECONFIG
```
```yaml
# $ export KUBECONFIG=$PWD/kubeconfig.yaml
# $ limactl shell k8s sudo cat /etc/kubernetes/admin.conf >$KUBECONFIG
```

Could make this actionable for Kubernetes, like it was done earlier for Docker / Podman ?

* #472

Adds set up for config files, to be automatically copied from the guest to the host: (with "mode")

```yaml
configFiles:
  - guestConfig: "/etc/rancher/k3s/k3s.yaml"
    hostConfig: "{{.Dir}}/conf/kubeconfig.yaml"
message: |
  To run `kubectl` on the host (assumes kubectl is installed):
  $ export KUBECONFIG={{.Dir}}/conf/kubeconfig.yaml
  $ kubectl
```
```yaml
configFiles:
  - guestConfig: "/etc/kubernetes/admin.conf"
    hostConfig: "{{.Dir}}/conf/kubeconfig.yaml"
message: |
  To run `kubectl` on the host (assumes kubectl is installed):
  $ export KUBECONFIG={{.Dir}}/conf/kubeconfig.yaml
  $ kubectl
```

Then the setup is a simple one-liner, to set up the KUBECONFIG variable ?

jandubois:
I feel this is both too powerful, and not powerful enough:

* Too powerful: When I run `limactl start https://lima.vm/k8s.yaml`, I don't want it to be able to overwrite just any file on the host.

* Not powerful enough: As your examples show, it does not do the full setup; you still have to setup the environment variable.

  And this is just one possible configuration; maybe I want to configure a docker context or kube context to be able to use the VM without any environment variable setup (e.g. to use it from VS.Code).

So I think copying the config variable should be just another line in the `message`:

```yaml
message: |
  To run `kubectl` on the host (assumes kubectl is installed):
  $ mkdir -p "{{.Dir}}/conf"
  $ limactl cp "{{.name}}:/etc/kubernetes/admin.conf" "{{.Dir}}/conf/kubeconfig.yaml"
  $ export KUBECONFIG="{{.Dir}}/conf/kubeconfig.yaml"
[...]

PR #536, opened 2022-01-09
Add message on how to run kubectl remotely

afbjorklund:
Currently this information is hidden in the YAML file.

Show "actionable" output, similar to docker and podman.

Instead of: https://github.com/lima-vm/lima/pull/516

See: https://github.com/lima-vm/lima/issues/514#issuecomment-1008290440


jandubois (approved):
LGTM

PR #541, opened 2022-01-10
Allow installing system containerd for alpine

### LINKED ISSUE #489: How to install containerd with alpine-lima
It's not fully clear how to add containerd to alpine, since the scripts require systemd:

```
ERRO[0018] [1 error occurred:
	* failed to satisfy the optional requirement 1 of 2 "systemd must be available": systemd is required to run containerd, but does not seem to be available.
Make sure that you use an image that supports systemd. If you do not want to run
containerd, please make sure that both 'container.system' and 'containerd.user'
are set to 'false' in the config file.
```

Installing containerd is quite simple, but packages for "nerdctl" and "buildkit" are missing...

```bash
provision:
  - mode: system
    script: |
      #!/bin/bash
      set -eux -o pipefail
      command -v containerd >/dev/null 2>&1 && exit 0
      apk add runc containerd cni-plugins
      sudo rc-update add containerd default
      sudo service containerd start
```

They can be added from the tarballs, but there doesn't seem to be any support available ?

That is, the current support is for nerdctl-full only (and not for nerdctl and buildkit archives)

----

Would it be better to make custom aports, or perhaps to add support for extra archives ?

containerd
* https://github.com/containerd/containerd/archive/v1.5.8.tar.gz
* https://github.com/alpinelinux/aports/tree/3.14-stable/community/containerd

nerdctl
* nerdctl-0.15.0-linux-amd64.tar.gz
* nerdctl-0.15.0-linux-arm64.tar.gz

buildkit
* buildkit-v0.9.3.linux-amd64.tar.gz
* buildkit-v0.9.3.linux-arm64.tar.gz
* buildkit.confd
* buildkit.initd

Both `lima sudo nerdctl run` and `lima sudo nerdctl build` seem to be doing just fine...

```
init-+-acpid
     |-7*[getty]
     |-sshd.pam---sshd.pam---sshd.pam-+-pstree
     |                                `-2*[sshfs---3*[{sshfs}]]
     |-supervise-daemo---lima-guestagent---7*[{lima-guestagent}]
     |-supervise-daemo---containerd---10*[{containerd}]
     |-supervise-daemo---buildkitd---8*[{buildkitd}]
     |-syslogd
     |-udevd
     `-udhcpc
```

afbjorklund:
Allows installing containerd/buildkitd, and then using `sudo nerdctl`:

```yaml
containerd:
  system: true
  user: false
```

Unfortunately it doesn't respect the socket variables and give weird errors:

```console
$ nerdctl ps
WARN[0000] environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
WARN[0000] environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
FATA[0000] rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): stat /tmp/runtime-anders/containerd-rootless: no such file or directory 
$ containerd-rootless-setuptool.sh install
[ERROR] Needs systemd (systemctl --user)
```

The error messages from `ctr` and `buildctl` are much more understandable.

`ctr: failed to dial "/run/containerd/containerd.sock": connection error: desc = "transport: error while dialing: dial unix /run/containerd/containerd.sock: connect: permission denied"`

`error: failed to list workers: Unavailable: connection error: desc = "transport: error while dialing: dial unix /run/buildkit/buildkitd.sock: connect: permission denied"`

Closes #489

----

There is a backlog item in alpine-lima, to provide packages for buildkit.

* https://github.com/lima-vm/alpine-lima/issues/37

Including a proper OpenRC service, to run with the `supervise-daemon`:

```
init-+-acpid
     |-buildkitd---10*[{buildkitd}]
     |-7*[getty]
     |-sshd.pam---sshd.pam---sshd.pam-+-ash---pstree
     |                                |-sshfs---4*[{sshfs}]
     |                                `-sshfs---3*[{sshfs}]
     |-supervise-daemo---lima-guestagent---7*[{lima-guestagent}]
     |-supervise-daemo---containerd---11*[{containerd}]
     |-syslogd
     |-udevd
     `-udhcpc
```

But otherwise, seems to work OK:

```console
$ sudo nerdctl version
Client:
 Version:	v0.15.0
 Git commit:	b72b5ca14550b2e23a42787664b6182524c5053f

Server:
 containerd:
  Version:	v1.5.8
  GitCommit:	1e5ef943eb76627a6d3b6de8cd1ef6537f393a71
```

One nice feature is when building:

```
sudo nerdctl build --output type=image,name=test:latest .
```

https://github.com/moby/buildkit#containerd-image-store

To not having to save the image to a tarball and load it.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
@@ -5,6 +5,24 @@ if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" != 1 ] && [ "${LIMA_CIDATA_CONTAINERD_US
 	exit 0
 fi
 
+if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" = 1 ] && [ -f /etc/alpine-release ]; then
+	# need to use the containerd packages from alpine, because the
+	# installation from nerdctl requires glibc (alpine uses musl)
+	apk add runc containerd cni-plugins
+	rc-update add containerd default
+	rc-service containerd start
+
+	# no buildkit package in alpine yet, but the installation from
+	# nerdctl is statically linked (but misses an openrc script)
+	tar -xz -C /usr/local -f "${LIMA_CIDATA_MNT}"/nerdctl-full.tgz bin/buildctl bin/buildkitd
+	nohup buildkitd --oci-worker=false --containerd-worker=true >/var/log/buildkitd.log 2>&1 &
+
+	# alpine uses busybox, so the "tar" syntax is somewhat different
+	# don't need the rootless files, since they only work in systemd
+	tar -xz -C /usr/local -f "${LIMA_CIDATA_MNT}"/nerdctl-full.tgz bin/nerdctl
+	exit 0
+fi
+
 # This script does not work unless systemd is available
 command -v systemctl >/dev/null 2>&1 || exit 0
 
--- a/pkg/hostagent/requirements.go
+++ b/pkg/hostagent/requirements.go
@@ -137,7 +137,7 @@ A possible workaround is to run "lima-guestagent install-systemd" in the guest.
 
 func (a *HostAgent) optionalRequirements() []requirement {
 	req := make([]requirement, 0)
-	if *a.y.Containerd.System || *a.y.Containerd.User {
+	if *a.y.Containerd.User {
 		req = append(req,
 			requirement{
 				description: "systemd must be available",
@@ -149,10 +149,10 @@ if ! command -v systemctl 2>&1 >/dev/null; then
     exit 1
 fi
 `,
-				debugHint: `systemd is required to run containerd, but does not seem to be available.
+				debugHint: `systemd is required to run containerd rootless, but does not seem to be available.
 Make sure that you use an image that supports systemd. If you do not want to run
-containerd, please make sure that both 'container.system' and 'containerd.user'
-are set to 'false' in the config file.
+containerd, please make sure that 'containerd.user' is set to 'false' in the
+config file.
 `,
 			},
 			requirement{
--- a/pkg/hostagent/requirements.go
+++ b/pkg/hostagent/requirements.go
@@ -137,6 +137,25 @@ A possible workaround is to run "lima-guestagent install-systemd" in the guest.
 
 func (a *HostAgent) optionalRequirements() []requirement {
 	req := make([]requirement, 0)
+	if *a.y.Containerd.System {
+		req = append(req,
+			requirement{
+				description: "systemd or alpine must be available",
+				fatal:       true,
+				script: `#!/bin/bash
+set -eux -o pipefail
+if ! command -v systemctl 2>&1 >/dev/null -a ! test -f /etc/alpine-release; then
+    echo >&2 "systemd or alpine is not available on this OS"
+    exit 1
+fi
+`,
+				debugHint: `systemd or alpine is required to run containerd, but does not seem to be available.
+Make sure that you use alpine or an image that supports systemd. If you do not
+want to run containerd, please make sure that 'containerd.system' is set to 'false'
+in the config file.
+`,
+			})
+	}
 	if *a.y.Containerd.User {
 		req = append(req,
 			requirement{
@@ -154,7 +173,10 @@ Make sure that you use an image that supports systemd. If you do not want to run
 containerd, please make sure that 'containerd.user' is set to 'false' in the
 config file.
 `,
-			},
+			})
+	}
+	if *a.y.Containerd.System || *a.y.Containerd.User {
+		req = append(req,
 			requirement{
 				description: "containerd binaries to be installed",
 				script: `#!/bin/bash


pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
```diff
@@ -5,6 +5,28 @@ if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" != 1 ] && [ "${LIMA_CIDATA_CONTAINERD_US
 	exit 0
 fi
 
+if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" = 1 ] && [ -f /etc/alpine-release ]; then
+	# need to use the containerd packages
```
AkihiroSuda, line 9: Please explain the reason (glibc) in the comments

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
```diff
@@ -5,6 +5,28 @@ if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" != 1 ] && [ "${LIMA_CIDATA_CONTAINERD_US
...
+	apk add runc containerd cni-plugins
+	rc-update add containerd default
+	rc-service containerd start
+
+	tar -xz -C /usr/local -f "${LIMA_CIDATA_MNT}"/nerdctl-full.tgz bin/buildctl bin/buildkitd
+	nohup buildkitd --oci-worker=false --containerd-worker=true >/var/log/buildkitd.log 2>&1 &
```
AkihiroSuda, line 15: Can we use openrc

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
```diff
@@ -5,6 +5,28 @@ if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" != 1 ] && [ "${LIMA_CIDATA_CONTAINERD_US
...
+
+	# normally set up by systemd-logind
+	cat >"/etc/profile.d/xdg_runtime.sh" <<EOF
+	if ! [ -d /tmp/runtime-\$USER ]; then
+	  mkdir -p /tmp/runtime-\$USER
+	  chmod 700 /tmp/runtime-\$USER
+	fi
+	export XDG_RUNTIME_DIR=/tmp/runtime-\$USER
```
AkihiroSuda, line 23: Why not use /run/user/$UID

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
afbjorklund: Sure, but I thought it was a bit long to include the scripts and config here so was saving it for the apk package...

https://github.com/lima-vm/alpine-lima/pull/36/files

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
afbjorklund: I wasn't really looking to reimplement loginctl and all the tmpfs and things, but maybe could just move it there.

afbjorklund: systemd compatibility seems to only be needed for rootless mode, so I will remove this shell script instead.

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
afbjorklund: Can explain that containerd is dynamic and buildkitd is static

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
afbjorklund: just added a placeholder for now

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
```diff
@@ -5,6 +5,24 @@ if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" != 1 ] && [ "${LIMA_CIDATA_CONTAINERD_US
 	exit 0
 fi
 
+if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" = 1 ] && [ -f /etc/alpine-release ]; then
+	# need to use the containerd packages from alpine, because the
+	# installation from nerdctl requires glibc (alpine uses musl)
+	apk add runc containerd cni-plugins
```
jandubois, line 11: Should this be done in `30-install-packages.sh` and protected by a check that `containerd` or the plugins aren't already installed?

pkg/hostagent/requirements.go
```diff
@@ -137,7 +137,7 @@ A possible workaround is to run "lima-guestagent install-systemd" in the guest.
 
 func (a *HostAgent) optionalRequirements() []requirement {
 	req := make([]requirement, 0)
-	if *a.y.Containerd.System || *a.y.Containerd.User {
+	if *a.y.Containerd.User {
```
jandubois, line 140: Isn't `systemd` still required on any OS except Alpine, even for system-containerd?

afbjorklund: Hmm, might need to check then if there is any other system not using systemd.

But you are right, there should probably be an `isAlpine()` somewhere in there...

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
afbjorklund: According to my experiments, we need one set of plugins for containerd and one set of plugins for nerdctl

afbjorklund: Actually I think it is more or less the _same_ CNI plugins, but installed in different paths (/usr vs /usr/local)

pkg/hostagent/requirements.go
afbjorklund: Hmm, I'm not sure if the Linux distribution is "known" by the time we run these requirements

Maybe it would be OK to let it pass, and then fail the installation if it is not Alpine Linux ?

afbjorklund: Added a check for alpine

AkihiroSuda:
@jandubois What should we do with this?

pkg/hostagent/requirements.go
```diff
@@ -137,7 +137,26 @@ A possible workaround is to run "lima-guestagent install-systemd" in the guest.
...
 	req := make([]requirement, 0)
-	if *a.y.Containerd.System || *a.y.Containerd.User {
+	if *a.y.Containerd.System {
+		req = append(req,
+			requirement{
+				description: "systemd or alpine must be available",
+				fatal:       true,
+				script: `#!/bin/bash
+set -eux -o pipefail
+if ! command -v systemctl 2>&1 >/dev/null -a ! test -f /etc/alpine-release; then
```
jandubois, line 147: This doesn't work for me; did you mean:

```suggestion
if ! command -v systemctl 2>&1 >/dev/null && ! test -f /etc/alpine-release; then
```

jandubois:
> @jandubois What should we do with this?

The PR doesn't work for me as-is (see above).

Even after I fixed that error, it still doesn't match the requirements:

```
INFO[0058] [hostagent] Waiting for the optional requirement 1 of 2: "systemd or alpine must be available"
INFO[0058] [hostagent] Forwarding "/run/lima-guestagent.sock" (guest) to "/Users/jan/.lima/alpine/ga.sock" (host)
INFO[0059] [hostagent] The optional requirement 1 of 2 is satisfied
INFO[0059] [hostagent] Not forwarding TCP [::]:22
INFO[0059] [hostagent] Not forwarding TCP 0.0.0.0:22
INFO[0059] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0059] [hostagent] Forwarding TCP from 127.0.0.1:42457 to 127.0.0.1:42457
INFO[0102] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0145] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0188] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0231] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0274] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0317] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0360] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0404] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0447] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0490] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0533] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
INFO[0576] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed"
FATA[0614] did not receive an event with the "running" status
```

Not sure what is going on, as the `nerdctl` file does exist:

```console
lima-alpine:/var/log# if ! timeout 30s bash -c "until command -v nerdctl; do sleep 3; done"; then echo NOT; else echo DOES;
fi
/usr/local/bin/nerdctl
DOES
```

So it should have passed, but didn't. Unfortunately I don't have time to debug this now.

I also continue to think that any packages installed via package manager (like `apk add runc containerd cni-plugins`) should happen in `30-install-packages.sh`. Maybe it doesn't matter, but I think `40-install-containerd` should only install things that have been bundled in cidata, but not download things over the internet.

In general I would feel better if `buildkitd` could also installed via `apk`, but I'm fine with doing the hacky thing until that is in place.

afbjorklund:
It is probably not urgent. Supporting `apk add nerdctl` or `containerd-rootless-setuptool.sh install` will be a longer journey, I think we solved most of the technical issues but it is always about the little details (as shown here).

In this world of horribly confusing names, `ctr` and `containerd` are supplied by package "containerd" but `buildctl` and `buildkitd` are supplied by package "buildkit" (no d). Fortunately, nerdctl is still in package "nerdctl". :-)

> I also continue to think that any packages installed via package manager (like apk add runc containerd cni-plugins) should happen in 30-install-packages.sh. Maybe it doesn't matter, but I think 40-install-containerd should only install things that have been bundled in cidata, but not download things over the internet.

This means that it needs to use a different download (than the "nerdctl-full" installer), from upstream. Currently that installation only works on distributions with systemd and glibc, which is probably a reasonable requirement.

So it seems that there are some issues that could be driven upstream (packaging nerdctl), and then the lima implementation would be simpler. It _could_ also support rootless installation, as was demonstrated earlier.

I will close this issue, for now.

Feel free to-reopen as a "tracker".

PR #547, opened 2022-01-13
HostResolver shouldn't return IPv6 addresses by default

jandubois:
They don't work over the SLIRP interface, and sometimes don't even work over a bridged vmnet interface.

See also https://github.com/rancher-sandbox/rancher-desktop/issues/1193

This PR deprecates the `useHostResolver` setting and replaces it with `hostResolver.enabled`. The old setting can still be used, but if the new property is set, any old setting in the same file is being ignored. Generating proper warnings gets pretty complex when the interaction between `default.yaml`, `lima.yaml`, and `override.yaml` have to be considered.

It could be argued that `HostResolver.LookupIPv6` should default to `true` for compatibility with previous releases. However, given that it seems to be broken in some setups, disabling IPv6 by default seems like a safer choice to me.

Also includes one bug fix for `limactl debug dns 5533 5533`, which previously ignored the TCP port.

Aside: I needed to include the TCP port during testing because `limactl debug dns 5533` (without `-6`) option resulted in:

```console
$ dig @127.0.0.1 -p 5533 AAAA google.com
;; Truncated, retrying in TCP mode.
;; Connection to 127.0.0.1#5533(127.0.0.1) for google.com failed: connection refused.
```

### DIFF
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -236,9 +236,14 @@ propagateProxyEnv: null
 # The host agent implements a DNS server that looks up host names on the host
 # using the local system resolver. This means changing VPN and network settings
 # are reflected automatically into the guest, including conditional forward,
-# and mDNS lookup:
-# Default: true
-useHostResolver: null
+# and mDNS lookup. By default only IPv4 addresses will be returned. IPv6 addresses
+# can only work when using a vmnet network interface and the host has working
+# IPv6 configured as well.
+hostResolver:
+  # Default: true
+  enabled: null
+  # Default: false
+  ipv6: null
 
 # If useHostResolver is false, then the following rules apply for configuring dns:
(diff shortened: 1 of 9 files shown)


pkg/limayaml/default.yaml
```diff
@@ -236,9 +236,14 @@ propagateProxyEnv: null
...
-# Default: true
-useHostResolver: null
+# and mDNS lookup. By default only IPv4 addresses will be returned. IPv6 addresses
+# can only work when using a vmnet network interface and the host has working
+# IPv6 configured as well.
+hostResolver:
+  # Default: true
+  enabled: null
+  # Default: false
+  lookupIPv6: null
```
AkihiroSuda, line 246: nit: `lookup` -> `lookUp` ?

AkihiroSuda (approved):
Thanks

PR #553, opened 2022-01-14
docs(README): mention nerdctl.lima shortcut

### LINKED ISSUE #552: mention nerdctl.lima in README
In the blog posts I've read about lima and nerdctl, I noticed that people are not aware of the feature in lima that they can use nerdctl. Lima shortcut instead of typing "lima nerdctl." So, I thought it'd be great to mention this tool in README.

> please feel free to assign it to me. I'm willing to work on it.

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Fixes #552

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,8 @@ $ cat /tmp/lima/another-file
 $ lima nerdctl run -d --name nginx -p 127.0.0.1:8080:80 nginx:alpine
 ```
 
+> You don't need to run "lima nerdctl" everytime, instead you can use special shortcut called "nerdctl.lima" to do the same thing. By default, it'll be installed along with the lima, so, you don't need to do anything extra. There will be a symlink called nerdctl pointing to nerdctl.lima. This is only created when there is no nerdctl entry in the directory already though. It worths to mention that this is created only via make install. Not included in Homebrew/MacPorts/nix packages.
+
 http://127.0.0.1:8080 is accessible from both macOS and Linux.
 
 For the usage of containerd and nerdctl (contaiNERD ctl), visit https://github.com/containerd/containerd and https://github.com/containerd/nerdctl.


jandubois:
If you mention `nerdctl.lima`, then you should also mention that there will be a symlink called `nerdctl` pointing to `nerdctl.lima`. This is only created when there is no `nerdctl` entry in the directory already though.

developer-guy:
Thank you @jandubois, I've updated the documentation according to your feedback; would you mind reviewing it again?

developer-guy:
> > There will be a symlink called nerdctl pointing to nerdctl.lima.
> 
> This is created only via `make install`. Not included in Homebrew/MacPorts/nix packages.

Should I add this to the information?

jandubois:
> This is created only via `make install`.
> Not included in Homebrew/MacPorts/nix packages.

Isn't this a bug in the packaging (or the `Makefile`)?

If `make binaries` would create the symlink, then at least `brew` would install it:

```ruby
  def install
    system "make", "VERSION=#{version}", "clean", "binaries"

    bin.install Dir["_output/bin/*"]
    share.install Dir["_output/share/*"]
```

Is there a reason why you wouldn't want to include the symlink?

AkihiroSuda:
> Isn't this a bug in the packaging (or the Makefile)?

Not a bug. Including `nerdctl -> nerdctl.lima` symlink conflicts with native `nerdctl`, Rancher Desktop's `nerdctl`, and colima's `nerdctl`.
(We have lots of `nerdctl` ... 😅 )

developer-guy:
> > > There will be a symlink called nerdctl pointing to nerdctl.lima.
> > 
> > 
> > This is created only via `make install`. Not included in Homebrew/MacPorts/nix packages.
> 
> Should I add this to the information?

So, what do you say then? Do we need to mention that as additional info?

AkihiroSuda:
> So, what do you say then? Do we need to mention that as additional info?

Yes, that would be appreciated.

AkihiroSuda (approved):
Thanks

PR #565, opened 2022-01-19
README.md: clarify that Lima is similar to WSL2 but not similar to WSL1

AkihiroSuda:


afbjorklund:
@AkihiroSuda : I learned that WSL2 is more similar to something like LXD, I think it runs a _single_ shared Hyper-V VM and then does "system containers" for the distributions ? So while it is still more similar to WSL2 than WSL1, not sure "equivalent" - maybe "similar to" ?

AkihiroSuda:
Thanks, updated the wording

README.md
```diff
@@ -3,9 +3,9 @@
 
 # Lima: Linux virtual machines (on macOS, in most cases)
 
-Lima launches Linux virtual machines with automatic file sharing, port forwarding, and [containerd](https://containerd.io).
+Lima launches Linux virtual machines with automatic file sharing, port forwarding (similar to WSL2), and [containerd](https://containerd.io).
```
jandubois, line 6: Just nit picking, but I would maybe word it like this, to make it clear that both file sharing and port forwarding are similar to WSL2. But I'm also fine with leaving it as-is.

```suggestion
Lima launches Linux virtual machines with automatic file sharing and port forwarding (similar to WSL2), and [containerd](https://containerd.io).
```

jandubois (approved):
Thanks!

README.md
AkihiroSuda: Thanks, updated

Issue #431, opened 2021-11-23
archlinux.yaml: seems slower, no serial.log


efx:
I'm noticing the archlinux version declared [here](https://github.com/lima-vm/lima/blob/67e465e2a57ad673cb41e5c8b51695dc783e401b/examples/archlinux.yaml):
- takes significantly longer than the debian example.
- has nothing in `serial.log`

You can reproduce this with:
```
limactl start  --tty=false examples/archlinux.yaml
```

# expectations

I am going to switch to using the debian base image since I am not wed to a given Linux distribution for the project I am working on. That is also related to #97. 
Someone with greater Arch Linux expertise would be most welcome to determine why these things are off.

motecshine:
![image](https://user-images.githubusercontent.com/7615270/146324960-a4dcfef0-09f8-4e4d-9625-65cd954f27d2.png)

PR #566, opened 2022-01-19
dont upgrade pkg when install sshf

Junnplus:
~~maybe fix https://github.com/lima-vm/lima/issues/431~~
Not the main reason, but useful.


AkihiroSuda (approved):
Thanks

PR #569, opened 2022-01-20
The podman-remote binary might be missing

afbjorklund:
Some distributions only ship "podman" on Linux.

But it has a `--remote` flag to make it a client...

(Normally, `podman` runs directly on the host)

For Mac/Win, the "podman" binary _is_ remote.


AkihiroSuda (approved):

afbjorklund:
Thanks!

PR #570, opened 2022-01-20
Use DNS and proxy settings from first interface with an IPv4 address

jandubois:
Instead of always preferring `en0`. This means the settings will now be picked up from e.g. the wireless interface when the ethernet is not plugged in.

I've shown in https://github.com/rancher-sandbox/rancher-desktop/issues/1125#issuecomment-996442224 that the interfaces in the system profiler data are returned in service order.


AkihiroSuda (approved):
Thanks

Issue #190, opened 2021-09-02
Feature request: ability to edit an instance


mkopinsky:
As far as I can tell, the only time I can edit the configuration of an instance is the first time I run `limactl start`. It would be nice if I could do something like:

* `limactl stop default`
* `limactl edit default` (opens editor just like the first time it's created)
* `limactl start default`

Maybe this is already possible somehow? If so, I can't find the documentation in either the repo or in CLI help.

AkihiroSuda:
This is already (unofficially) possible with `vi ~/.lima/default/lima.yaml`.

https://github.com/lima-vm/lima/blob/master/docs/internal.md#instance-directory-lima_homeinstance

I agree we should have the proper CLI command for this.

afbjorklund:
Do you need a command `limactl restart default`, as well ?

afbjorklund:
Would it be possible to do this in two steps too, so that we can use an external editor - like for instance a GUI ?

One command to output the current YAML for an instance, one command (or edit param) to update from a file.

jandubois:
> Would it be possible to do this in two steps too, so that we can use an external editor - like for instance a GUI ?
> 
> One command to output the current YAML for an instance, one command (or edit param) to update from a file.

I think that is too much scaffolding. You can already get the name of the config with our API:

```console
$ limactl ls -f '{{.LimaHome}}/{{.Name}}/lima.yaml' default
/Users/jan/.lima/default/lima.yaml
```

So from a GUI you get the filename, edit the content, and restart the instance (using `stop` and `start`). So all the building blocks already exist.

Or maybe I misunderstand what you are trying to do.

afbjorklund:
Sure, that would work too.

jandubois:
> Sure, that would work too.

I'm just trying to keep our API as small as possible, to avoid overwhelming new users with complexity. Tools like `qemu` or `git ` create a lot of anxiety when people try to wrap their heads around them for the first time, and some simply give up.

So having a small set of basic building blocks that people can use to create their own integrations is preferable to me. Every new API needs to justify why it is necessary! 😄

afbjorklund:
I think the `{{.LimaHome}}/{{.Name}}` dir was already being provided by `limactl list --json`, so easy enough.

PR #572, opened 2022-01-21
Add edit command

Junnplus:
Signed-off-by: ye.sijun <junnplus@gmail.com>

Fixes: https://github.com/lima-vm/lima/issues/190

### DIFF
--- a/cmd/limactl/edit.go
+++ b/cmd/limactl/edit.go
@@ -7,7 +7,10 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/AlecAivazis/survey/v2"
 	"github.com/lima-vm/lima/pkg/limayaml"
+	networks "github.com/lima-vm/lima/pkg/networks/reconcile"
+	"github.com/lima-vm/lima/pkg/start"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/sirupsen/logrus"
@@ -40,12 +43,16 @@ func editAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if inst.Status == store.StatusRunning {
+		return errors.New("Cannot edit a running instance")
+	}
+
 	filePath := filepath.Join(inst.Dir, filenames.LimaYAML)
 	yContent, err := os.ReadFile(filePath)
 	if err != nil {
 		return err
 	}
-	hdr := fmt.Sprintf("# Please edit the folling configuration for Lima instance %q\n", instName)
+	hdr := fmt.Sprintf("# Please edit the following configuration for Lima instance %q\n", instName)
 	hdr += "# and an empty file will abort the edit.\n"
 	hdr += "\n"
 	yBytes, err := openEditor(cmd, instName, yContent, hdr)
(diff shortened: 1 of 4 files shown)


afbjorklund:
This could need a matching `restart` command, perhaps.

jandubois:
I think the edit command should restart the instance automatically if it is already running (but don't restart when the editing was aborted). Leaving the instance running with the old config is confusing, as the `limactl ls` command will report incorrect settings.

You could add `--restart=false` option if you really want to, but I think it is not needed. People who want to do this can always edit the files under `$LIMA_HOME` directly; `limactl edit` should leave the system in a consistent state.

AkihiroSuda:
Automatic restart may surprise users, so I prefer —restart to be false by default

jandubois:
> Automatic restart may surprise users, so I prefer —restart to be false by default

I fail to see the surprise: you edit the config to change the instance; why would you not expect the instance to reflect the modified settings? To me it is the opposite: When I change the config, e.g. to increase the amount of memory to the instance, I expect those changes to be active when I "save" them.

Think of it as a crude "preferences" or "settings" dialog. It is highly unusual to change settings and not have them take effect when you close that dialog. Some dialogs may have an "Apply" button, so changes don't take effect immediately. But once the dialog is closed, the desired and actual state are reconciled. 😄 

How does it make sense to have a running instance that doesn't match its configuration data?

jandubois:
Also think about the confusion if you don't apply the changes on save, and then forget that you have edited the config: if you edit the instance again a week later, the config file no longer reflects the state of the instance at all, and now you wonder how it all got into this state.

If you want to have a modified config file separate from the running instance, you should make a copy, and edit the copy while it is in stopped state. (I know we don't have a copy/clone command yet).

Junnplus:
> I fail to see the surprise: you edit the config to change the instance; why would you not expect the instance to reflect the modified settings? To me it is the opposite: When I change the config, e.g. to increase the amount of memory to the instance, I expect those changes to be active when I "save" them.

I agree with @jandubois . Or can we only allow editing of stopped instances?

AkihiroSuda:
Not all users expect automatic restart because they might be running some apps with unsaved data. Also, some properties could be updated without restarting (in future).

> Or can we only allow editing of stopped instances?

This also SGTM

mikluko:
IMO there should be a promt about restart. E.g., `limactl start` asks interactive questions upon vm creation unless `--tty=false`. Why not use same approch here?

jandubois:
> IMO there should be a promt about restart. E.g., `limactl start` asks interactive questions upon vm creation unless `--tty=false`. Why not use same approch here?

I assume you want to ask a question here **after** the editor has been closed, and not ask if the user actually wants to edit the file.

I would agree with this if we discard the changes when the user decides not to restart the VM, e.g.

```console
$ limactl edit alpine
Waiting for editor...

--- current settings
+++ net settings
@@ -9,6 +9,7 @@ images:

 mounts:
 - location: "~"
+  writable: true
 - location: "/tmp/lima"
   writable: true

? Instance "alpine" configuration has been changed  [Use arrows to move, type to filter]
> Apply changes and restart instance
  Edit to make more changes
  Discard changes and exit
```

I would still disagree with letting the user keep changes in the file without applying them: you either apply or discard.

jandubois:
I think `limactl edit` should also validate the changes, and only allow you to apply them, when they pass validation.

Junnplus:
@jandubois I added a patch that only allows editing of non-running instances, what do you think?

jandubois:
> I added a patch that only allows editing of non-running instances, what do you think?

I think this is fine. It addresses my concern that the `lima.yaml` and running instance are out of sync. Also matches what e.g. VMware Fusion does: it simply disallows you changing certain configurations unless the VM is stopped.

Given that `limactl edit` is what `git` developers would call a "porcelain" command, I still think it would be useful to allow editing of running instances, and handle the stopping and restarting automatically, in the way I've outlined above (showing the diff; requiring the user to either restart, discard, or continue to edit).

But I understand that this is outside the scope of your current PR, so will file this as an enhancement request (suggestion, really), so maybe somebody may implement it at a later time.

Still waiting to hear what others think.

cmd/limactl/edit.go
```diff
@@ -40,12 +43,16 @@ func editAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if inst.Status == store.StatusRunning {
+		return errors.New("Cannot edit a running instance")
+	}
```
AkihiroSuda, line 48: We should take a lock, but can be another PR

https://github.com/lima-vm/lima/blob/v0.8.1/pkg/lockutil/lockutil.go

AkihiroSuda (approved):

AkihiroSuda:
@jandubois LGTY?

jandubois (approved):
Thanks, LGTM.

Philosophically it feels weird to me that we don't allow editing of a running instance by stopping it first, but we do ask to start the stopped instance after editing it. But we can debate and change it some other time.

sildvauresberg:
Hello everyone, I have a problem with running a command `limactl edit <INSTANCE>`
I guess you updated the master README without releasing a new version with this functionality?

```
$ limactl --version
limactl version 0.8.1

$ limactl edit default
FATA[0000] unknown command "edit" for "limactl"

$ limactl help edit
Unknown help topic [`edit`]
```

afbjorklund:
You can read the old README, if you want: https://github.com/lima-vm/lima/tree/v0.8.1

afbjorklund:
I implemented the same in the GUI, to disable the "Edit" button for running instances.

PR #582, opened 2022-01-24
Add the Slack link

AkihiroSuda:
Rancher Users slack has `#lima` channel, colocated with `#rancher-desktop` channel (Lima is not Rancher's product)

- New account: https://slack.rancher.io/
- Login: https://rancher-users.slack.com/

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,7 @@
+blank_issues_enabled: true
+contact_links:
+- name: Ask a question (GitHub Discussions)
+  url: https://github.com/lima-vm/lima/discussions
+- name: Chat with Lima users and developers
+  url: https://slack.rancher.io/
+  about: Rancher Users slack has `#lima` channel, colocated with `#rancher-desktop` channel (Lima is not a Rancher project)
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,7 @@
+blank_issues_enabled: true
+contact_links:
+- name: Ask a question (GitHub Discussions)
+  url: https://github.com/lima-vm/lima/discussions
+- name: Chat with Lima users and developers
+  url: https://slack.rancher.io/
+  about: Rancher Users slack has `#lima` channel, colocated with `#rancher-desktop` channel (Lima is not Rancher's product)
```
jandubois, line 7: I would prefer: "(Lima is not a Rancher project)".

Technically it should also be "SUSE" as Rancher does not exist as a separate identity anymore, but I think it is fine here because it is in the "Rancher Users Slack" name. But Rancher/SUSE is providing professional services (support, consulting); all the software is in FOSS projects.

So this should just be pointing out that Lima is not a Rancher project.

Well, I think. Now it feels like nit-picking. Change or apply as-is! 😄

jandubois (dismissed):
LGTM

.github/ISSUE_TEMPLATE/config.yml
AkihiroSuda: fixed

jandubois (approved):
LGTM

PR #596, opened 2022-01-26
Add kubernetes with kind example

afbjorklund:
### kind (kubernetes in docker)
https://kind.sigs.k8s.io/

Sets up an instance with docker, kind and kubectl.

Then uses "kind create cluster" to start a container.

Requested by some people on the Kubernetes slack.

Mostly based on docker.yaml and the upstream docs.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -20,6 +20,7 @@ Container engines:
 Container orchestration:
 - [`k3s.yaml`](./k3s.yaml): Kubernetes via k3s
 - [`k8s.yaml`](./k8s.yaml): Kubernetes via kubeadm
+- [`kind.yaml`](./kind.yaml): Kubernetes via [kind](https://kind.sigs.k8s.io/)
 - [`nomad.yaml`](./nomad.yaml): Nomad
 - [`faasd.yaml`](./faasd.yaml): [Faasd](https://docs.openfaas.com/deployment/faasd/)
 
--- a/examples/kind.yaml
+++ b/examples/kind.yaml
@@ -1,0 +1,109 @@
+# Deploy kubernetes via kind (kubernetes-in-docker)
+# $ limactl start ./kind.yaml
+# $ limactl shell kind kubectl cluster-info
+# $ limactl shell kind docker container ps
+
+# This example requires Lima v0.8.0 or later
+images:
+# Hint: run `limactl prune` to invalidate the "current" cache
+- location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-amd64.img"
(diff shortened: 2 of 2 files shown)


afbjorklund:
Maybe not obvious, but there is "containerd" in there as well:

```console
$ kubectl get nodes -o wide
NAME                 STATUS   ROLES                  AGE   VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE       KERNEL-VERSION      CONTAINER-RUNTIME
kind-control-plane   Ready    control-plane,master   20h   v1.23.1   172.18.0.2    <none>        Ubuntu 21.10   5.13.0-27-generic   containerd://1.5.9
$ docker ps
CONTAINER ID   IMAGE                  COMMAND                  CREATED        STATUS          PORTS                       NAMES
69ec00dd23b4   kindest/node:v1.23.1   "/usr/local/bin/entr…"   20 hours ago   Up 10 minutes   127.0.0.1:40169->6443/tcp   kind-control-plane
```

AkihiroSuda:
I'm not sure we want this.
We already have `k8s.yaml` and `k3s.yaml` to run Kubernetes.
If users really need to use `kind`, they can install `kind` into `docker.yaml` or `podman.yaml` manually.

jandubois:
> I'm not sure we want this.

I agree. I expected this to work:

```console
$ brew install kind
$ limactl start examples/docker.yaml
$ kind create cluster
ERROR: failed to create cluster: running kind with rootless provider requires cgroup v2, see https://kind.sigs.k8s.io/docs/user/rootless/
```

So shouldn't we fix that instead of creating another instance?

afbjorklund:
> I'm not sure we want this.

I think it was more that _they_ wanted it, since people wanted to avoid running Docker Desktop but still run `kind`...

The alternative (for kind) would be Podman Desktop, but it hasn't been released yet (expecting 4.0 during Feb)

But people can run with `k8s.yaml`.

> So shouldn't we fix that instead of creating another instance?

I think it is work ongoing (with "rootless"), which is documented at https://kind.sigs.k8s.io/docs/user/rootless/

I'm not sure what kind of modifications are required to the lima instances, in order to use it ? Nor very interested.

I will continue with minikube instead.

----

There is also a small project to support `nerdctl` as a provider, that could also be an option for lima (when completed).

* https://github.com/kubernetes-sigs/kind/issues/2317

If all you want is to run Kubernetes, then there is already both Docker Desktop and Rancher Desktop (k3s) today.

Eventually there will also be a new minikube with M1 support, but at the moment the driver and iso are struggling.

PR #606, opened 2022-01-29
Change default for `mounts.sshfs.cache` to `true`

### LINKED ISSUE #600: Reconsider disabling sshfs caching by default
This was already proposed in PR #556.

So far we just made the setting configurable, but the default is still `cache=no`.

Given the problems reported in #556 and elsewhere, e.g. https://github.com/rancher-sandbox/rancher-desktop/issues/1349, I wonder if we shouldn't have reverted the default as well when we made the setting configurable.

jandubois:
This was changed to `false` in lima 0.8.1:

* https://github.com/lima-vm/lima/pull/538

It has generated multiple reports of breaking PHP package manager, unit tests, and file system access itself:

* https://github.com/lima-vm/lima/pull/556
* https://github.com/abiosoft/colima/issues/129
* https://github.com/rancher-sandbox/rancher-desktop/issues/1349

Given that the use cases that triggered the change in #538 were observed in automated environment / CI, and that the PHP breakages happens to people using Lima for interactive development, I think the default should be reverted.

Closes #600

### DIFF
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -50,7 +50,9 @@ mounts:
   sshfs:
     # Enabling the SSHFS cache will increase performance of the mounted filesystem, at
     # the cost of potentially not reflecting changes made on the host in a timely manner.
-    # Default: false
+    # Warning: It looks like PHP filesystem access does not work correctly when
+    # the cache is disabled.
+    # Default: true
     cache: null
     # SSHFS has an optional flag called 'follow_symlinks'. This allows mounts
     # to be properly resolved in the guest os and allow for access to the
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -339,7 +339,7 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 	for i := range y.Mounts {
 		mount := &y.Mounts[i]
 		if mount.SSHFS.Cache == nil {
-			mount.SSHFS.Cache = pointer.Bool(false)
+			mount.SSHFS.Cache = pointer.Bool(true)
 		}
 		if mount.SSHFS.FollowSymlinks == nil {
 			mount.SSHFS.FollowSymlinks = pointer.Bool(false)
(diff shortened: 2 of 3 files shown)


jandubois:
**Poll:** please add a 👍 or 👎 for this PR!

Pinging @AkihiroSuda @rfay @abiosoft, but everyone who has an opinion, please "vote" and maybe add a comment supporting your vote.

rfay:
I'm ok with going back to the previous default.

AkihiroSuda:
Has anyone straced why PHP Composer needs this?

Is there any other `sshfs` option that works for both Composer and https://github.com/abiosoft/colima/issues/99 ?

pkg/limayaml/default.yaml
```diff
@@ -50,7 +50,9 @@ mounts:
   sshfs:
     # Enabling the SSHFS cache will increase performance of the mounted filesystem, at
     # the cost of potentially not reflecting changes made on the host in a timely manner.
-    # Default: false
+    # Warning: It looks like PHP filesystem access does not work correctly when
```
AkihiroSuda, line 53: > PHP filesystem

All PHP apps, or just PHP Composer?

AkihiroSuda (approved):

pkg/limayaml/default.yaml
jandubois: The failing samples from https://github.com/abiosoft/colima/issues/129 seem to be just filesystem iterators, but do not involve composer. If I read the comments correctly, then the code worked up to lima 0.8.0, but has been broken in 0.8.1. On reporter claims to have verified this on 3 different machines.

Personally I've only experimented with composer. It fails because it doesn't see any of the unzipped files. However, the files are not there when composer ends, so I assumed they never got written. But maybe composer deletes the directories again?

I'm not the best person to investigate this; I've never written a PHP program in my life, and this is all nested layers of packaged things, afaict. All manual attempts to unzip things seem to have worked fine for me, as has using `git clone` instead of `unzip` to install dependencies (but it takes forever).

The fact that it affects user programs and not just composer is what makes me want to revert the default.

PR #616, opened 2022-01-31
editor: print a warning when override.yaml exists

### LINKED ISSUE #612: `limactl edit` should print a warning (error?) when override.yaml exists
### Description

`limactl edit` (https://github.com/lima-vm/lima/pull/572) does not work as expected when override.yaml exists.
At least we should print a warning when override.yaml exists.

AkihiroSuda:
Fix #612


AkihiroSuda:
I'll release Lima v0.8.2 after merging this and https://github.com/containerd/nerdctl/pull/749 (nerdctl v0.16.1)

jandubois:
One thing I noticed is: when the user edits the header (e.g. to mistakenly uncomment something from the default or override file), then the whole header remains in the file.

I wonder if we should display a warning when the header has been modified?

Issue #632, opened 2022-02-06
`lima bash -c "$(echo -e '\n\techo balls\n')"` doesn't work as expected


AkihiroSuda:
### Description

```console
$ bash -c "$(echo -e '\n\techo balls\n')"
balls

$ lima bash -c "$(echo -e '\n\techo balls\n')"
bash: line 1: ntecho: command not found
```

This was originally reported as an issue of `nerdctl` in https://github.com/containerd/nerdctl/issues/769 (Thanks to @jdblack and @buu700), but turned out to be an issue of `lima`, not `nerdctl`.

PR #633, opened 2022-02-06
Fix SSH command escaping

buu700:
Addresses https://github.com/lima-vm/lima/issues/632.

I haven't tested very thoroughly, and may be making some incorrect assumptions about how shellescape works, but I can confirm that this at least fixes the `lima bash -c "$(echo -e '\n\techo balls\n')"` example for me.

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -110,7 +110,10 @@ func shellAction(cmd *cobra.Command, args []string) error {
 
 	script := fmt.Sprintf("%s ; exec bash --login", changeDirCmd)
 	if len(args) > 1 {
-		script += fmt.Sprintf(" -c %q", shellescape.QuoteCommand(args[1:]))
+		script += fmt.Sprintf(
+			" -c %s",
+			shellescape.Quote(shellescape.QuoteCommand(args[1:])),
+		)
 	}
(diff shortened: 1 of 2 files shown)


jandubois (changes requested):
This is not sufficient, as it doesn't deal with commands that include quotes. E.g. it breaks:

```console
$ lima bash -c 'echo "foo bar"'
foo bar
```

With the current patch is breaks like this:

```console
$ lima bash -c 'echo "foo bar"'
bar': -c: line 1: unexpected EOF while looking for matching `''
bar': -c: line 2: syntax error: unexpected end of file
```

buu700:
All right, should be good now. Thanks @jandubois!

As an aside, both of these examples would be good test to add for the future.

jandubois (approved):
LGTM, but since you volunteered to add tests (😄 ), maybe add them to this PR?

buu700:
lol, all right, done. I confirmed that the new tests currently pass, and each one also fails as expected with the two previous versions of shell.go.

AkihiroSuda:
https://github.com/apps/dco

buu700:
Done.

jandubois (approved):
Thanks!

PR #642, opened 2022-02-09
Implement the suspend and resume commands

### LINKED ISSUE #598: Implement suspend and resume of VM
### Description

Seems like QMP has these nice commands:

**stop**
Suspend execution of VM

**cont**
Reverse a previous stop command - resume execution of VM.

https://en.wikibooks.org/wiki/QEMU/Monitor#Virtual_machine

Allows you to suspend and resume the VM, through qmp.sock:

```
Welcome to the QMP low-level shell!
Connected to QEMU 4.2.1
(QEMU) stop
{"return": {}}
(QEMU) cont
{"return": {}}
```

afbjorklund:
For halting the emulation and putting the VM into paused run state.

Also query the status of running instances, when showing in "list".

Closes #598

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,10 @@ Detailed usage:
 
 - Run `limactl delete [--force] <INSTANCE>` to delete the instance.
 
+- Run `limactl suspend <INSTANCE>` to pause the instance.
+
+- Run `limactl resume <INSTANCE>` to unpause the instance.
+
 - Run `limactl edit <INSTANCE>` to edit the instance.
 
 - To enable bash completion, add `source <(limactl completion bash)` to `~/.bash_profile`.
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -83,6 +83,8 @@ func newApp() *cobra.Command {
 		newShowSSHCommand(),
 		newDebugCommand(),
 		newEditCommand(),
+		newSuspendCommand(),
+		newResumeCommand(),
 	)
 	return rootCmd
 }
--- a/cmd/limactl/resume.go
+++ b/cmd/limactl/resume.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/lima-vm/lima/pkg/pause"
+	"github.com/lima-vm/lima/pkg/store"
+
+	"github.com/spf13/cobra"
+)
+
+func newResumeCommand() *cobra.Command {
+	var resumeCmd = &cobra.Command{
+		Use:               "resume INSTANCE",
+		Short:             "Resume (unpause) an instance",
+		Aliases:           []string{"unpause"},
+		Args:              cobra.MaximumNArgs(1),
+		RunE:              resumeAction,
+		ValidArgsFunction: resumeBashComplete,
+	}
+
+	return resumeCmd
+}
+
+func resumeAction(cmd *cobra.Command, args []string) error {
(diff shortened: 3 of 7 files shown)


afbjorklund:
Tested with default instance and a busy loop, which went "silent" after putting the instance into pause.

There might be an issue with ssh getting disconnected if sleeping too long, but can be fixed by a restart.

AkihiroSuda:
> There might be an issue with ssh getting disconnected if sleeping too long, but can be fixed by a restart.


I think we should switch away from sshfs to 9p before merging this PR.

jandubois:
> I think we should switch away from sshfs to 9p before merging this PR.

I agree, but the volume mounts are not the only part affected by SSH disconnects. Unix socket forwarding is also sometimes breaking when the host is sleeping. I hope to look into this whole topic (host sleeping) next week, but I still have hopes that 9p would solve/improve other issues, like performance.

afbjorklund:
Actually I meant the guest sleeping, suspending the host probably has similar (or worse issues).

pkg/hostagent/api/api.go
```diff
@@ -3,3 +3,8 @@ package api
 type Info struct {
 	SSHLocalPort int `json:"sshLocalPort,omitempty"`
 }
+
+type Status struct {
+	Running bool `json:"running"`
+	Paused  bool `json:"paused"`
+}
```
AkihiroSuda, line 10: Can we merge this into `Info`?

afbjorklund: The main difference at the moment is that Info has all the information available, but Status relies on an external call (to the driver)

It would be possible to combine them, but then it would always do the external call... I'm also not sure on when to use the hostagent, and when to call the driver directly ?

AkihiroSuda: > I'm also not sure on when to use the hostagent, and when to call the driver directly ?

The hostagent API should be used only when it is unavoidable.
e.g., The `SSHLocalPort` is communicated over the hostagent API, because the `SSHLocalPort` is determined by the hostagent.

I guess the `Status` does not need to be part of the hostagent API.

AkihiroSuda: If the `Status` really has to be a part of the hostagent API, it should be a single string because `Running` and `Paused` are mutually exclusive.

afbjorklund: Can change the API into a string, but probably should make them more like enums and "translate".

afbjorklund: Here are the known states: (for the qemu driver)

```go
// RunState implements the "RunState" QMP API type.
type RunState int

// Known values of RunState.
const (
        RunStateDebug RunState = iota
        RunStateInmigrate
        RunStateInternalError
        RunStateIOError
        RunStatePaused
        RunStatePostmigrate
        RunStatePrelaunch
        RunStateFinishMigrate
        RunStateRestoreVM
        RunStateRunning
        RunStateSaveVM
        RunStateShutdown
        RunStateSuspended
        RunStateWatchdog
        RunStateGuestPanicked
        RunStateColo
)

```

afbjorklund:
I rebased, but haven't updated the API.

afbjorklund:
Maybe a new method to reset/restore all networking ?

pkg/hostagent/api/api.go
afbjorklund: Converted booleans to string.

Issue #592, opened 2022-01-24
Provide a `cpu` setting to override default `host` setting


jandubois:
It seems that sometimes the `-cpu host` option configures a CPU that cannot be used to create a VM. See for example https://github.com/rancher-sandbox/rancher-desktop/issues/1291#issuecomment-1020527274.

Changing to e.g. `-cpu Haswell` could fix the issue, but right now the only mechanism we have to change the setting is via the clumsy `$QEMU_SYSTEM_X86_64` environment variable, that is really only meant for debugging.

I propose adding a `cpu` field to `lima.yaml` that defaults to `host`, but allows the user to switch to a different CPU.

AkihiroSuda:
Having `cpu` (singular) in addition to `cpus` (plural) might be confusing.

Maybe `cpuType`?

PR #643, opened 2022-02-10
add cpuType for specify cpu

Junnplus:
closes: https://github.com/lima-vm/lima/issues/592

- [x] doc

### DIFF
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -29,8 +29,8 @@ images:
 
 # Specify QEMU CPU emulation, e.g., "host", "qemu64", "Haswell".
 # You can see what options are available for host emulation with: `qemu-system-$(arch) -cpu help`.
-# And set of instructions is supported, like: "qemu64,+ssse3".
-# Default: "host" (native arch) or "qemu64" (Intel on ARM) or "cortex-a72" (ARM on Intel)
+# Setting of instructions is supported like this: "qemu64,+ssse3".
+# Default: "host" (native arch), "qemu64" (Intel on ARM), or "cortex-a72" (ARM on Intel)
 cpuType: null
 
 # CPUs: if you see performance issues, try limiting cpus to 1.
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -27,6 +27,12 @@ images:
 - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
   arch: "aarch64"
 
+# Specify QEMU CPU emulation, e.g., "host", "qemu64", "Haswell".
+# You can see what options are available for host emulation with: `qemu-system-$(arch) -cpu help`.
+# And set of instructions is supported, like: "qemu64,+ssse3".
+# Default: "host" (native arch) or "qemu64" (Intel on ARM) or "cortex-a72" (ARM on Intel)
+cpuType: null
+
 # CPUs: if you see performance issues, try limiting cpus to 1.
 # Default: 4
 cpus: null
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -81,6 +81,24 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 		}
 	}
 
+	if y.CPUType == nil {
+		y.CPUType = d.CPUType
+	}
+	if o.CPUType != nil {
+		y.CPUType = o.CPUType
+	}
+	if y.CPUType == nil || *y.CPUType == "" {
+		if IsNativeArch(*y.Arch) {
+			y.CPUType = pointer.String("host")
+		} else if *y.Arch == X8664 {
+			// Intel on ARM
+			y.CPUType = pointer.String("qemu64")
+		} else {
+			// ARM on Intel
+			y.CPUType = pointer.String("cortex-a72")
+		}
+	}
+
 	if y.CPUs == nil {
 		y.CPUs = d.CPUs
 	}
@@ -460,4 +478,10 @@ func ResolveArch(s *string) Arch {
 		return NewArch(runtime.GOARCH)
 	}
 	return *s
+}
+
+func IsNativeArch(arch Arch) bool {
+	nativeX8664 := arch == X8664 && runtime.GOARCH == "amd64"
+	nativeAARCH64 := arch == AARCH64 && runtime.GOARCH == "arm64"
+	return nativeX8664 || nativeAARCH64
 }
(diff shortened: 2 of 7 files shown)


pkg/limayaml/default.yaml
```diff
@@ -27,6 +27,8 @@ images:
 - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
   arch: "aarch64"
 
+cpuType: null
```
jandubois, line 30: Needs documentation and default value.

pkg/qemu/qemu.go
```diff
@@ -206,16 +206,20 @@ func Cmdline(cfg Config) (string, []string, error) {
 		}
 		return "", nil, errors.New(errStr)
 	}
+
+	var cpu string
```
jandubois, line 210: I think the whole default handling to determine the definitive cpu type should happen in `pkg/limayaml/defaults.go`, including the normal `default.yaml` and `override.yaml` rules. The final value should be reported by `limactl ls --json`.

jandubois (commented):
I know the PR is still in draft mode, but I wanted to get in some early feedback already.

pkg/qemu/qemu.go
Junnplus: Thanks for the suggestion, I will improve it.

cmd/limactl/list.go
```diff
@@ -165,7 +165,7 @@ func listAction(cmd *cobra.Command, args []string) error {
 	}
 
 	w := tabwriter.NewWriter(cmd.OutOrStdout(), 4, 8, 4, ' ', 0)
-	fmt.Fprintln(w, "NAME\tSTATUS\tSSH\tARCH\tCPUS\tMEMORY\tDISK\tDIR")
+	fmt.Fprintln(w, "NAME\tSTATUS\tSSH\tARCH\tCPU TYPE\tCPUS\tMEMORY\tDISK\tDIR")
```
AkihiroSuda, line 168: too verbose

pkg/qemu/qemu.go
```diff
@@ -206,16 +206,12 @@ func Cmdline(cfg Config) (string, []string, error) {
...
 	case limayaml.X8664:
-		cpu := "qemu64"
-		if isNativeArch(*y.Arch) {
-			cpu = "host"
-		}
-		args = appendArgsIfNoConflict(args, "-cpu", cpu)
-		if isNativeArch(*y.Arch) {
-			args = appendArgsIfNoConflict(args, "-machine", "q35,accel="+accel)
-		} else {
+		if strings.HasPrefix(cpu, "qemu64") {
```
AkihiroSuda, line 214: Why not just compare `cpu != "host"`

pkg/limayaml/default.yaml
```diff
@@ -27,6 +27,12 @@ images:
 - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
   arch: "aarch64"
 
+# Specify QEMU CPU emulation, e.g., "host", "qemu64", "Haswell".
+# You can see what options are available for host emulation with: `qemu-system-$(arch) -cpu help`.
+# And set of instructions is supported, like: "qemu64,+ssse3".
+# Default: "host" (native arch) or "qemu64" (Intel on ARM) or "cortex-a72" (ARM on Intel)
```
jandubois, line 33:
```suggestion
# Setting of instructions is supported like this: "qemu64,+ssse3".
# Default: "host" (native arch), "qemu64" (Intel on ARM), or "cortex-a72" (ARM on Intel)
```

pkg/limayaml/defaults.go
```diff
@@ -81,6 +81,24 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
+	if y.CPUType == nil || *y.CPUType == "" {
+		if IsNativeArch(*y.Arch) {
+			y.CPUType = pointer.String("host")
+		} else if *y.Arch == X8664 {
+			// Intel on ARM
+			y.CPUType = pointer.String("qemu64")
+		} else {
+			// ARM on Intel
+			y.CPUType = pointer.String("cortex-a72")
+		}
```
jandubois, line 99: If we have specific reasons for picking these CPU types, could you add them in the comments?

pkg/qemu/qemu.go
jandubois: Yeah, seems prudent. I don't know either if these work for all cpu types (and if the reasoning to use them applies).

pkg/limayaml/defaults.go
Junnplus: Sure, or we need additional documentation to explain.

jandubois: I would prefer some explanation in the comments that answers "Why have these values been chosen as the defaults?"

If we have additional documentation, it should expand on this and also offer which alternative values you might want to chose, and why.

I'm not sure if we actually do have the answers to either of these questions, but if we do, it would be nice to record them.

pkg/limayaml/defaults.go
```diff
@@ -81,6 +81,25 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
+	}
+	if o.CPUType != nil {
+		y.CPUType = o.CPUType
+	}
+	if y.CPUType == nil || *y.CPUType == "" {
+		if IsNativeArch(*y.Arch) {
+			y.CPUType = pointer.String("host")
+		} else if *y.Arch == X8664 {
+			// Intel on ARM
+			// Since https://github.com/lima-vm/lima/pull/494, we use qemu64 cpu for better emulation of x86_64.
```
Junnplus, line 95: @jandubois I added new comments in here.

Junnplus: For `cortex-a72`, I don't have any answer. :(

jandubois (approved):
Thanks, LGTM. Leaving for @AkihiroSuda to make final decision and merge.

AkihiroSuda (approved):
Thanks

pkg/limayaml/limayaml.go
```diff
@@ -9,6 +9,7 @@ import (
 type LimaYAML struct {
 	Arch              *Arch             `yaml:"arch,omitempty" json:"arch,omitempty"`
 	Images            []File            `yaml:"images" json:"images"` // REQUIRED
+	CPUType           *string           `yaml:"cpuType,omitempty" json:"cpuType,omitempty"`
```
AkihiroSuda, line 12: I already merged this PR, but on second thought we could define this as `[]struct{Type string, Arch Arch}` so that a single YAML can be used on multiple archs?

AkihiroSuda: Changed to `map[Arch]string` in https://github.com/lima-vm/lima/pull/656 (thanks @jandubois )

PR #650, opened 2022-02-15
Allow adding static names to hostresolver

### LINKED ISSUE #622: Allow adding static names to hostresolver
The host resolver looks up names on the host, not inside the guest, so will not be able to resolve names from `/etc/hosts` inside the guest. This doesn't matter for programs running directly inside the guest, as they use the local resolver, and the host resolver is just a nameserver.

It does however affect containers, which only have access to the nameserver, so can't resolve e.g. `host.lima.internal`.

I suggest to add a configuration option:

```
hostResolver:
  hosts:
    "host.docker.internal": host
    "example.com": 127.0.0.1
```

The string `host` is special and will be translated to the gateway address to the host.

The name `host.lima.internal` is built-in and will always be defined as a name of the host.

jandubois:
This is an alternative to adding static names to `/etc/hosts` on the **host**, to make them available inside containers running in the guest. In addition this change allows creating aliases to `host.lima.internal` to point to the host gateway from inside the guest.

Manual testing with `examples/docker.yaml` with these changes:

```yaml
hostResolver:
  hosts:
    host.docker.internal: host.lima.internal
    cname.docker.internal: host.docker.internal
    guest.internal: 127.1.1.1
```

Testing inside an Alpine container running on Docker:

```console
$ docker context create lima --docker "host=unix:///Users/jan/.lima/docker/sock/docker.sock"
$ docker context use lima
$ docker run -it --rm alpine
/ # nslookup host.docker.internal
Server:		10.0.2.3
Address:	10.0.2.3:53

Non-authoritative answer:

Non-authoritative answer:
host.docker.internal	canonical name = host.lima.internal
Name:	host.lima.internal
Address: 192.168.5.2

/ # nslookup cname.docker.internal
Server:		10.0.2.3
Address:	10.0.2.3:53

Non-authoritative answer:

Non-authoritative answer:
cname.docker.internal	canonical name = host.lima.internal
Name:	host.lima.internal
Address: 192.168.5.2

/ # nslookup guest.internal
Server:		10.0.2.3
Address:	10.0.2.3:53

Non-authoritative answer:
Name:	guest.internal
Address: 127.1.1.1

Non-authoritative answer:
```

In addition this PR contains a bugfix for the `hostResolver.ipv6: false` configuration: instead of forwarding `AAAA` queries to the default resolvers, we now return an empty reply to avoid returning `NXDOMAIN`. During testing I found that that was causing names to resolve correctly on Alpine even when `dig` and `nslookup` seemed to show proper DNS information. It might be a bug in the musl resolver.

Finally this PR also adds the current hostname with the slirp IP address to the host resolver. This is done automatically by systemd-resolved, but necessary for e.g. Alpine.

Fixes #622

### DIFF
--- a/examples/docker.yaml
+++ b/examples/docker.yaml
@@ -23,6 +23,10 @@ containerd:
   user: false
 provision:
 - mode: system
+  # This script defines the host.docker.internal hostname when hostResolver is disabled.
+  # It is also needed for lima 0.8.2 and earlier, which does not support hostResolver.hosts.
+  # Names defined in /etc/hosts inside the VM are not resolved inside containers when
+  # using the hostResolver; use hostResolver.hosts instead (requires lima 0.8.3 or later).
   script: |
     #!/bin/sh
     sed -i 's/host.lima.internal.*/host.lima.internal host.docker.internal/' /etc/hosts
@@ -56,6 +60,11 @@ probes:
       exit 1
     fi
   hint: See "/var/log/cloud-init-output.log". in the guest
+hostResolver:
+  # hostResolver.hosts requires lima 0.8.3 or later. Names defined here will also
+  # resolve inside containers, and not just inside the VM itself.
+  hosts:
+    host.docker.internal: host.lima.internal
 portForwards:
 - guestSocket: "/run/user/{{.UID}}/docker.sock"
   hostSocket: "{{.Dir}}/sock/docker.sock"
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -256,6 +256,12 @@ hostResolver:
   enabled: null
   # Default: false
   ipv6: null
+  # Static names can be defined here as an alternative to adding them to the hosts /etc/hosts.
+  # Values can be either other hostnames, or IP addresses. The host.lima.internal name is
+  # predefined to specify the gateway address to the host.
+  hosts:
+    # guest.name: 127.1.1.1
+    # host.name: host.lima.internal
 
 # If useHostResolver is false, then the following rules apply for configuring dns:
 # Explicitly set DNS addresses for qemu user-mode networking. By default qemu picks *one*
--- a/cmd/limactl/debug.go
+++ b/cmd/limactl/debug.go
@@ -47,7 +47,7 @@ func debugDNSAction(cmd *cobra.Command, args []string) error {
 			return err
 		}
 	}
-	srv, err := dns.Start(udpLocalPort, tcpLocalPort, ipv6)
+	srv, err := dns.Start(udpLocalPort, tcpLocalPort, ipv6, map[string]string{})
 	if err != nil {
 		return err
 	}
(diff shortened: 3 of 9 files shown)


pkg/limayaml/default.yaml
```diff
@@ -264,7 +270,7 @@ hostResolver:
 # that has an IPv4 address, to the list. In case this still doesn't work (e.g. VPN
 # setups), the servers can be specified here explicitly. If nameservers are specified
 # here, then the configuration from network preferences will be ignored.
-# dns:
+dns:
```
AkihiroSuda, line 273: Why?

jandubois: Just to make it more visible when looking at the file with syntax highlighting. I thought other list properties were not commented out either, like `networks`, but now I see this is inconsistent, e.g. `provision`, `probes`, and `portForwards` are all commented out.

Personally I would prefer to uncomment all the top-level list properties, but it doesn't really belong into this PR. I will revert it, and then we can decide if we either want to comment out `networks`, or to uncomment them all.

AkihiroSuda:
Seems accidentally containing https://github.com/lima-vm/lima/pull/643

Could you try rebasing with the master?

jandubois:
> Could you try rebasing with the master?

Should be already fixed; maybe you need to refresh?

jandubois:
Sorry, I did one more force-push to change the hostname address from `127.0.1.1` to `192.168.5.15` (the slirp IP address). This seems to be more in line with the [Debian hostname resolution](https://www.debian.org/doc/manuals/debian-reference/ch05.en.html#_the_hostname_resolution) documentation, given that the slirp IP is pretty permanent:

>The IP address 127.0.1.1 in the second line of this example may not be found on some other Unix-like systems. The [Debian Installer](https://en.wikipedia.org/wiki/Debian-Installer) creates this entry for a system without a permanent IP address as a workaround for some software (e.g., GNOME) as documented in the [bug #719621](http://bugs.debian.org/719621).

examples/docker.yaml
```diff
@@ -56,6 +52,9 @@ probes:
       exit 1
     fi
   hint: See "/var/log/cloud-init-output.log". in the guest
+hostResolver:
+  hosts:
+    host.docker.internal: host.lima.internal
 portForwards:
 - guestSocket: "/run/user/{{.UID}}/docker.sock"
```
AkihiroSuda, line 59: Can we split this yaml to another PR and hold it until right before the release of vNext, to avoid seeing “this example does not work” issues?

jandubois: I've restored the original script because I realized it is still required when `hostResolver.enabled` is `false`.

I've added comments to the file that should make it obvious that this feature requires lima 0.8.3 or later.

I've also added tests to the PR.

pkg/limayaml/default.yaml
jandubois: @AkihiroSuda Do you still want to split out the change, even with the additional comments in place? I think it is pretty clear now. And it's not like the example doesn't work; it should continue to work as before. It just won't have the new functionality.

I can split it still, if you prefer, but I really don't think this is necessary.

AkihiroSuda (approved):
Thanks

PR #653, opened 2022-02-16
feat: add buildkit.yaml to examples

### LINKED ISSUE #567: add buildkit.yaml to examples
We've developed a tool called buildkit-machine that allows to forward sock or open a TCP to sock to let macOS users can access buildkitd easily without running buildkitd in a container. Then, we noticed that there is an alternative way of doing that as lima did for docker sock like the following:

```YAML
portForwards:
  hostSocket: "{{.Dir}}/sock/docker.sock"
```
> https://github.com/lima-vm/lima/blob/2b6db749c5ee63f758741e0891f1915978611b16/examples/docker.yaml#L59-L61

So, AFAIK, lima also runs the buildkitd in a rootles way which means that users may access it through /run/user/{{.UID}}/buildkit/buildkitd.sock. What we are proposing is that we can add a file to the examples folder for the buildkit to achieve the same thing.

<details>

<summary> buildkit.yaml </summary>

```YAML
portForwards:
- guestSocket: "/run/user/{{.UID}}/buildkit/buildkitd.sock"
  hostSocket: "{{.Dir}}/buildkitd.sock"
```
</details>

```shell
$  buildctl --addr unix:///Users/batuhan.apaydin/.lima/buildkit/buildkitd.sock \
   build \
    --frontend=dockerfile.v0 \
    --local context=. \
    --local dockerfile=. \
    --output type=image,name=docker.io/devopps/hello-world-buildctl,push=true
```

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>

Fixes #567


examples/buildkit.yaml
```diff
@@ -0,0 +1,30 @@
...
+- location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
+  arch: "aarch64"
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+
+containerd:
+  system: null
```
afbjorklund, line 25: false ?

examples/buildkit.yaml
```diff
@@ -0,0 +1,30 @@
...
+  user: null
```
afbjorklund, line 26: true ?

examples/buildkit.yaml
```diff
@@ -0,0 +1,30 @@
+# Example to use Buildkit
+# $ limactl start ./buildkit.yaml
+
+# To run `buildkit` on the host (assumes buildctl is installed):
+# $ export BUILDKIT_HOST=unix://$HOME/.lima/buildkit/buildkitd.sock
```
afbjorklund, line 5: To work even if the template is used for a different setup, use a variable instead:

`export BUILDKIT_HOST=$(limactl list buildkit --format 'unix://{{.Dir}}/sock/buildkitd.sock')`

Note that the location needs to match whatever was forwarded below, to work...

examples/buildkit.yaml
```diff
@@ -0,0 +1,30 @@
...
+  arch: "x86_64"
+- location: "~/Downloads/impish-server-cloudimg-arm64.img"
+  arch: "aarch64"
+
+- location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-amd64.img"
+  arch: "x86_64"
+- location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
+  arch: "aarch64"
+
+mounts:
```
afbjorklund, line 19: are these mounts even needed ? normally everything would be in the build context, no ?

afbjorklund:
For the actual user-facing message, maybe a runnable example would be best.

Similar to the "hello-world" added to the docker and podman examples, perhaps.

examples/buildkit.yaml
developer-guy: ah thank you, really appreciate this kind of awesome informations 🤩

examples/buildkit.yaml
```diff
@@ -0,0 +1,30 @@
+# Example to use Buildkit
+# $ limactl start ./buildkit.yaml
+
+# To run `buildkit` on the host (assumes buildctl is installed):
+# $ export BUILDKIT_HOST=unix://$HOME/.lima/buildkit/buildkitd.sock
+# $ buildctl debug workers
+
+images:
+- location: "~/Downloads/impish-server-cloudimg-amd64.img"
```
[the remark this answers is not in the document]
developer-guy: thank you 🙋🏻‍♂️

examples/buildkit.yaml
developer-guy: yes you are right, we don't need to specify them here.

examples/buildkit.yaml
developer-guy: yep

examples/buildkit.yaml
developer-guy: agree 🙈

AkihiroSuda:
Please update README.md

developer-guy:
> Please update README.md

with the example of @afbjorklund mentioned above ?

AkihiroSuda:
> > Please update README.md
> 
> with the example of @afbjorklund mentioned above ?

No, just please follow the current style of `README.md` in the `examples` directory.
(Suggestion to change the style is welcome, but that should be a separate PR, and all the examples have to be updated at once)

examples/README.md
```diff
@@ -20,6 +20,7 @@ Container engines:
 - [`podman.yaml`](./podman.yaml): Podman
 - [`singularity.yaml`](./singularity.yaml): Singularity
 - LXD is installed in the default Ubuntu template, so there is no `lxd.yaml`
+- [`buildkit.yaml`](./buildkit.yaml): Buildkit
```
AkihiroSuda, line 23: I'm not sure BuildKit can be categorized as "Container engines".
Maybe "Container image builders"?

developer-guy: Sure

AkihiroSuda (dismissed):
Thanks

afbjorklund:
I was going to suggest turning containerd off, since it is not supported yet by rootless buildkitd anyway...

But unfortunately it seems that doing so also stops buildkitd, since they share user namespaces or something ?

developer-guy:
kindly ping here 🙋🏻‍♂️

afbjorklund:
I think it is spelled "BuildKit"

developer-guy:
I changed "Buildkit" to "BuildKit" wherever I found @afbjorklund 🙋🏻‍♂️

AkihiroSuda (approved):

examples/buildkit.yaml
```diff
@@ -0,0 +1,19 @@
+# Example to use BuildKit
+# $ limactl start ./buildkit.yaml
+
+# To run `buildkit` on the host (assumes buildctl is installed):
+# $ export BUILDKIT_HOST=$(limactl list buildkit --format 'unix://{{.Dir}}/sock/buildkitd.sock')
+# $ buildctl debug workers
```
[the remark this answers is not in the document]
developer-guy: ah thank, I didn't notice that lima has `message` property, it should be ok right now 🙋🏻‍♂️

examples/buildkit.yaml
```diff
@@ -0,0 +1,26 @@
...
+  user: true
+
+portForwards:
+- guestSocket: "/run/user/{{.UID}}/buildkit/buildkitd.sock"
+  hostSocket: "{{.Dir}}/sock/buildkitd.sock"
+
+message: |
+ To run `buildkit` on the host (assumes buildctl is installed), run the following commands:
+ -------
+ export BUILDKIT_HOST=$(limactl list buildkit --format 'unix://{{.Dir}}/sock/buildkitd.sock')
```
jandubois, line 24: There is no need to call a subprocess; just use the template expansion of the `message`:

```suggestion
 export BUILDKIT_HOST="unix://{{.Dir}}/sock/buildkitd.sock"
```

examples/buildkit.yaml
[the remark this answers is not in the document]
jandubois: I feel like having the comments are now redundant with the `message` property and should be removed. That way they won't diverge when further changes are applied to only one of them in the future. 😄

jandubois (approved):
I still think the comments in L4-6 are redundant with the `message` and should be deleted, but it looks like the other examples have them too. Maybe we can change them all together in a later PR. We can just move the `message` to the top of the file instead.

So, LGTM now, thanks!

AkihiroSuda (approved):
Thanks

PR #656, opened 2022-02-17
Make cpuType a map instead of a single value

### LINKED ISSUE #647: Revisit the cpuType type definition
> I already merged this PR, but on second thought we could define this as `[]struct{Type string, Arch Arch}` so that a single YAML can be used on multiple archs?

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/643#discussion_r805548077_

jandubois:
That way a template lima.yaml can specify alternate cpu types for multiple archs. Also makes it possible to specify defaults and overrides for cpu type globally.

Fixes #647

### DIFF
--- a/pkg/limayaml/default.yaml
+++ b/pkg/limayaml/default.yaml
@@ -27,12 +27,6 @@ images:
 - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
   arch: "aarch64"
 
-# Specify QEMU CPU emulation, e.g., "host", "qemu64", "Haswell".
-# You can see what options are available for host emulation with: `qemu-system-$(arch) -cpu help`.
-# Setting of instructions is supported like this: "qemu64,+ssse3".
-# Default: "host" (native arch), "qemu64" (Intel on ARM), or "cortex-a72" (ARM on Intel)
-cpuType: null
-
 # CPUs: if you see performance issues, try limiting cpus to 1.
 # Default: 4
 cpus: null
@@ -140,6 +134,15 @@ containerd:
 # ===================================================================== #
 # FURTHER ADVANCED CONFIGURATION
 # ===================================================================== #
+
+# Specify desired QEMU CPU type for each arch.
+# You can see what options are available for host emulation with: `qemu-system-$(arch) -cpu help`.
+# Setting of instructions is supported like this: "qemu64,+ssse3".
+cpuType:
+  # Default: "cortex-a72" (or "host" when running on arm64)
+  aarch64: null
+  # Default: "qemu64" (or "host" when running on amd64)
+  x86_64: null
 
 firmware:
   # Use legacy BIOS instead of UEFI.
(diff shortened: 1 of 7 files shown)


pkg/limayaml/default.yaml
```diff
@@ -27,11 +27,15 @@ images:
 - location: "https://cloud-images.ubuntu.com/impish/current/impish-server-cloudimg-arm64.img"
   arch: "aarch64"
 
-# Specify QEMU CPU emulation, e.g., "host", "qemu64", "Haswell".
+# Specify QEMU CPU type for each arch. Only used for emulation; cpu type is always "host"
+# when the desired arch matches the host arch.
```
AkihiroSuda, line 31: We have to support overriding "host" to "Haswell-v4":

https://github.com/lima-vm/lima/blob/71861bcc2c05c88c9cc2e74793df5354729c57d3/examples/rocky.yaml#L3-L5

pkg/limayaml/default.yaml
```diff
@@ -27,11 +27,15 @@ images:
...
 # You can see what options are available for host emulation with: `qemu-system-$(arch) -cpu help`.
 # Setting of instructions is supported like this: "qemu64,+ssse3".
-# Default: "host" (native arch), "qemu64" (Intel on ARM), or "cortex-a72" (ARM on Intel)
-cpuType: null
+cpuType:
+  # Default: "cortex-a72"
+  aarch64: null
+  # Default: "qemu64"
+  x86_64: null
```
AkihiroSuda, line 38: Not new in this PR, but I feel `cpuType` should be explained in the "FURTHER ADVANCED" section

pkg/limayaml/default.yaml
jandubois: Thanks! Should be fixed now.

pkg/limayaml/default.yaml
jandubois: I agree! That's why I asked for it in https://github.com/lima-vm/lima/pull/643#discussion_r805525700.

Unfortunately I don't have the answers either; otherwise I would add something.

jandubois: Oh, do you just mean it should be moved down in the file? I'm going to do that!

jandubois: Done

pkg/limayaml/defaults_test.go
```diff
@@ -282,11 +289,14 @@ func TestFillDefault(t *testing.T) {
...
 
 	o = LimaYAML{
-		Arch:    pointer.String(arch),
-		CPUType: pointer.String("host"),
-		CPUs:    pointer.Int(12),
-		Memory:  pointer.String("7GiB"),
-		Disk:    pointer.String("117GiB"),
+		Arch: pointer.String(arch),
+		CPUType: map[Arch]string{
+			AARCH64: "uber-arm",
```
AkihiroSuda, line 294: ?

AkihiroSuda (approved):
Thanks

pkg/limayaml/defaults_test.go
jandubois: Just random strings; they only need to be different from the lower priority settings, to verify that they are not being overwritten. When I wrote it, it felt like a funny fantasy-cpu name, but now I can't remember why I thought it was funny, so it probably isn't. I can change it if you want; suggestions welcome!

PR #660, opened 2022-02-17
Add portForwards.guestIPMustBeZero property

### LINKED ISSUE #658: `hostIP: "127.0.0.1"` shouldn't be hardcoded (to support exposing  `<nerdctl|docker|podman> run -p 0.0.0.0:80:80`)
### Description

Currently `<nerdctl|docker|podman> run -p 0.0.0.0:80:80` does not listen the port on host 0.0.0.0, because `hostIP: "127.0.0.1"` is hardcoded:

https://github.com/lima-vm/lima/blob/71861bcc2c05c88c9cc2e74793df5354729c57d3/pkg/limayaml/default.yaml#L218-L222

We should probably use 0.0.0.0 when the guest uses 0.0.0.0.

The guest agent is already aware of "0.0.0.0" IP.

https://github.com/lima-vm/lima/blob/71861bcc2c05c88c9cc2e74793df5354729c57d3/pkg/guestagent/api/api.go#L13-L28

Background:
- https://github.com/abiosoft/colima/issues/186#issuecomment-1042571055

cc @abiosoft

jandubois:
When `guestIPMustBeZero` is set the rule will only match if the bind is against literal `0.0.0.0` and not any other interface.

To forward any ports bound to `0.0.0.0` in the guest to `0.0.0.0` on the host, but any other ports bound to `127.0.0.1` only to localhost on the host use these rules:

```yaml
portForwards:
- guestPortRange: [1, 65535]
  guestIPMustBeZero: true
  hostIP: "0.0.0.0"
- guestPortRange: [1, 65535]
  guestIP: "127.0.0.1"
  hostIP: "127.0.0.1"
```

Fixes #658

### DIFF
--- a/hack/test-port-forwarding.pl
+++ b/hack/test-port-forwarding.pl
@@ -292,3 +292,15 @@ portForwards:
   # forward: 0.0.0.0      4033 → ipv4 4033
   # forward: ::           4034 → ipv4 4034
   # forward: ::1          4035 → ipv4 4035
+
+- guestIPMustBeZero: true
+  guestPortRange: [4040, 4049]
+
+- guestIP: "0.0.0.0"
+  guestPortRange: [4040, 4049]
+  ignore: true
+
+  # forward: 0.0.0.0        4040 → 127.0.0.1 4040
+  # forward: ::             4041 → 127.0.0.1 4041
+  # ignore:  127.0.0.1      4043 → 127.0.0.1 4043
+  # ignore:  192.168.5.15   4044 → 127.0.0.1 4044
--- a/pkg/hostagent/port.go
+++ b/pkg/hostagent/port.go
@@ -52,7 +52,9 @@ func (pf *portForwarder) forwardingAddresses(guest api.IPPort) (string, string) 
 		case guest.IP.IsUnspecified():
 		case guest.IP.Equal(rule.GuestIP):
 		case guest.IP.Equal(net.IPv6loopback) && rule.GuestIP.Equal(api.IPv4loopback1):
-		case rule.GuestIP.IsUnspecified():
+		case rule.GuestIP.IsUnspecified() && !rule.GuestIPMustBeZero:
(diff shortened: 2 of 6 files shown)


AkihiroSuda:
> guestIPZero: true
> guestIP: "0.0.0.0"

Can we come up with a better name to distinguish them?
e.g., `propagateExposure: true`

jandubois:
> Can we come up with a better name to distinguish them?

I've been pleading for suggestions in #658 already...

> e.g., `propagateExposure: true`

Sorry, but that name doesn't really work for me; I have no idea what it means. If anything, this option would reduce exposure, as the rule would no longer match any interface, but only a bind to `0.0.0.0` (so it is not treated as a wildcard).

I have one other suggestion now: `guestIPExactMatchRequired`:

```yaml
portForwards:
- guestIP: "0.0.0.0"
  guestIPExactMatchRequired: true
  hostIP: "0.0.0.0"
```

 In the end, the documentation should clarify this, but ideally the name should be self-explanatory.

jandubois:
I do have an issue with "exact match required" though: it is hard to argue that it should imply that `guestIP` must be `0.0.0.0` because every specific IP address also requires an exact match.

So you could say that setting is just ignored for other `guestIP` values. Or you say it defaults to `true`, except for `0.0.0.0`, where it defaults to `false`, which also feels weird.

AkihiroSuda:
guestIPExactMatchRequired seems better than guestIPZero

AkihiroSuda:
Please update the PR title and description (when this PR is out of draft)

jandubois:
> Please update the PR title and description (when this PR is out of draft)

Done. I've also changed the second rule in the description to only forward all the `127.0.0.1` binds. I think forwarding all ports from all interfaces is generally not a good idea.

AkihiroSuda (approved):
Thanks.

Maybe we should document this in `docs/network.md` too

PR #667, opened 2022-02-19
iptables rule without an ip range applies to all interfaces

jandubois:
The current logic pretends it only applies to `127.0.0.1`, which means the new `guestIPMustBeZero` rule does not detect it properly.

This is a problem with `nerdctl` and `containerd` on Alpine. It works "by accident" on Ubuntu because the port was also bound to `[::]`, which had an entry in `/proc/net/tcp6`.

This issue does not affect moby because moby doesn't rely on `iptables` rules but has proper entries in `/proc/net/tcp`.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,7 +30,7 @@ task:
   - cat /proc/cpuinfo
   install_deps_script:
   - apt-get update
-  - apt-get install -y --no-install-recommends ca-certificates curl git golang openssh-client make netcat ovmf sudo qemu-system-x86 qemu-utils
+  - apt-get install -y --no-install-recommends ca-certificates curl git golang jq openssh-client make netcat ovmf sudo qemu-system-x86 qemu-utils
   go_cache:
     fingerprint_script: uname -s ; cat go.sum
     folder: $GOPATH/pkg/mod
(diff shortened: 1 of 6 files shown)


AkihiroSuda:
Thanks, can we have a test?

jandubois:
> Thanks, can we have a test?

The PR includes updates to the parsing test, so this logic is covered by a test.

If you mean an integration test for nerdctl, then it gets more complicated. As I mentioned, this work accidentally on Ubuntu because of IPv6 having an entry in `/proc/net/tcp6`. And the Alpine image, which does show the issue does not have `containerd` installed.

I guess I can `apk add containerd && rc-service containerd start` for this, but it is a bit weird that we can't use `containerd.user: true` for this.

jandubois:
> Thanks, can we have a test?

I've added a test and verified that the test fails on with the Alpine image on both the macOS and the Linux runner.

Since it failed on both, I've removed `alpine.yaml` from the macOS runners again; let me know if you want to run Alpine on macOS too. It makes some sense, as Alpine has probably the largest difference to all other images, and it is the basis for both colima and Rancher Desktop.

AkihiroSuda:
> Since it failed on both, I've removed alpine.yaml from the macOS runners again; let me know if you want to run Alpine on macOS too. It makes some sense, as Alpine has probably the largest difference to all other images, and it is the basis for both colima and Rancher Desktop.

Can be worked on a separate PR

AkihiroSuda (approved):
Thanks

PR #670, opened 2022-02-20
Add Alpine to the macOS test matrix

jandubois:
It is different from the other examples because it doesn't use systemd and cloud-init, and also boots from an ISO and uses a separate data volume. So it makes sense to test it on both Linux and macOS.


AkihiroSuda (approved):
LGTM if green

PR #674, opened 2022-02-22
limactl start: support `limactl start template://TEMPLATE`

AkihiroSuda:
e.g., `limactl start --name=default template://docker`

Available templates:
```console
$ limactl start --list-templates
almalinux
alpine
archlinux
debian
default
docker
faasd
fedora
k3s
k8s
nomad
opensuse
podman
rocky
singularity
ubuntu-lts
ubuntu
vmnet
```

The template can be also chosen from TUI
```console
$ limactl start
? Creating an instance "default"  [Use arrows to move, type to filter]
  Proceed with the current configuration
  Open an editor to review or modify the current configuration
> Choose another example (docker, podman, archlinux, fedora, ...)
  Exit

? Choose an example  [Use arrows to move, type to filter]
> almalinux
  alpine
  archlinux
  debian
  default
  docker
  faasd
...
```


Makefile
```diff
@@ -23,8 +23,10 @@ binaries: \
 	_output/bin/nerdctl.lima \
 	_output/share/lima/lima-guestagent.Linux-x86_64 \
 	_output/share/lima/lima-guestagent.Linux-aarch64
+	cp -aL examples _output/share/lima
 	mkdir -p _output/share/doc/lima
-	cp -aL README.md LICENSE docs examples _output/share/doc/lima
+	cp -aL README.md LICENSE docs _output/share/doc/lima
+	ln -sf ../../lima/examples _output/share/doc/lima
```
jandubois, line 29: This breaks if you have a previous version already compiled and don't run `make clean` before building the new version:

```console
$ make all
[...]
cp -aL README.md LICENSE docs _output/share/doc/lima
ln -sf ../../lima/examples _output/share/doc/lima
ln: _output/share/doc/lima/examples: Operation not permitted
make: *** [binaries] Error 1
```

I think this is ok, but wanted to point it out, in case you want to fix it.

AkihiroSuda: Yes, I know, and not worth fixing IMHO

jandubois: Also fails to during install:

```console
$ make install
mkdir -p "/usr/local"
# Use tar rather than cp, for better symlink handling
( cd _output && tar c * | tar Cxv "/usr/local" )
x bin/
[...]
x share/doc/lima/examples: Can't replace existing directory with non-directory
[...]
tar: Error exit delayed from previous errors.
make: *** [install] Error 1
```

You have to run `make uninstall` first to fix it.

jandubois: I guess this is not our problem, but I wonder how this would affect updates via `brew`.

AkihiroSuda: Added commit `[Makefile: binaries depends on clean; install depends on uninstall](https://github.com/lima-vm/lima/pull/674/commits/d6343d5fae629212662d71c7ad25b1d18e620884)`

Homebrew is unlikely to be affected, as they have versioned directories under `Cellar`

cmd/limactl/start.go
```diff
@@ -29,16 +30,46 @@ import (
...
 	var startCommand = &cobra.Command{
-		Use:               "start NAME|FILE.yaml|URL",
-		Short:             fmt.Sprintf("Start an instance of Lima. If the instance does not exist, open an editor for creating new one, with name %q", DefaultInstanceName),
+		Use: "start NAME|FILE.yaml|URL",
+		Example: `
+Create an instance "default" (if not created yet) from the default Ubuntu template, and start it:
+$ limactl start
+
+Create an instance "default" from a template "docker":
+$ limactl start --name=default template://docker
```
jandubois, line 39: I wonder if the text should explain where the templates are located.

The README says:

> Older releases require `limactl start /usr/local/share/doc/lima/examples/TEMPLATE.yaml` instead.

which kind of documents it implicitly, but `limactl start --help` doesn't tell you which templates exist, not where they can be found.

Ideally (not in this PR) I would like to see something like `limactl start --list-templates` (similar to `limactl ls --list-fields`) that lists all the template names (maybe with a single-line summary for each, but maybe that doesn't make sense).

AkihiroSuda: Added `limactl start --list-templates`

AkihiroSuda:
Merging this. Feel free to open a follow-up issues/PRs

PR #694, opened 2022-03-03
cidata: automatically upgrade existing containerd/nerdctl ; nerdctl: update to v0.17.1

### LINKED ISSUE #303: cidata: support upgrading nerdctl
https://github.com/lima-vm/lima/blob/3ff1a15530f5f4b5ce9eb98600c72ea00fd914c6/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh#L11

Currently nerdctl is not upgraded even when nerdctl in cidata is newer than nerdctl in the root disk.

AkihiroSuda:
When the timestamp of `/mnt/lima-cidata/nerdctl-full.tgz:bin/nerdctl` is newer than
`/usr/local/bin/nerdctl`, containerd/nerdctl is automatically upgraded.

Drawback: increases boot time by 4-5 seconds.

Fix #303

- - -

Also bumps up the default nerdctl version to 0.17.1: https://github.com/containerd/nerdctl/releases/tag/v0.17.1

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -eux
 
 if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" != 1 ] && [ "${LIMA_CIDATA_CONTAINERD_USER}" != 1 ]; then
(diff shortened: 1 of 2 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
```diff
@@ -8,14 +8,31 @@ fi
 # This script does not work unless systemd is available
 command -v systemctl >/dev/null 2>&1 || exit 0
 
-if [ ! -x /usr/local/bin/nerdctl ]; then
+# Extract bin/nerdctl and compare whether it is newer than the current /usr/local/bin/nerdctl (if already exists).
+# Takes 4-5 seconds. (FIXME: optimize)
```
jandubois, line 12: Idea for optimization:

```console
root@lima-default:/mnt/lima-cidata# gunzip -c nerdctl-full.tgz | head -c 512 | tar tv | awk '/ bin\/$/ {print $4}'
2022-03-03
root@lima-default:/mnt/lima-cidata# date -r /usr/local/bin/nerdctl "+%F"
2022-03-03
```

Then do simple string comparison on the dates.

This relies on a tar header block being 512 bytes (so `tar` will not throw an error), and `bin/` being the first entry in the tarball. I think these are reasonable assumptions, but am not totally sure. If the first line doesn't return a non-empty string, fall back on the expensive check by extracting the tarball.

jandubois (approved):
LGTM, but I'm not merging in case you want to include the suggested optimization in this PR.

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
AkihiroSuda: > I think these are reasonable assumptions, but am not totally sure. 

Not guaranteed. https://github.com/containerd/nerdctl/blob/20233c26d26f11ca73a9a775fba87ca884ab14d2/Makefile#L67

jandubois: Yeah, not guaranteed, but seems likely to be true, and with the fallback it should still work.

My only other idea would be to extract the nerdctl version from the download URL and then do a version comparison (pass it in as `LIMA_CIDATA_NERDCTL_VERSION`).

Still not guaranteed if somebody uses a custom download location that doesn't include the version, or worse, a different version, but you can argue that they will get what they deserve if they do this. 😄

AkihiroSuda:
Let me merge and open an issue ticket for optimization in future

PR #695, opened 2022-03-04
nerdctl: use fuse-overlayfs only when kernel < 5.13

### LINKED ISSUE #383: nerdctl: disable fuse-overlayfs unless really needed
fuse-overlayfs shouldn't be needed when `(kernel >= 5.13) || (kernel >= 5.11 && !selinux) || (kernel is ubuntu's variant)`

AkihiroSuda:
The real overlayfs can be safely used on kernel >= 5.13.

Instances created with older releases of Lima will continue to use fuse-overlayfs.

Fix #383

### DIFF
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -92,6 +92,7 @@ The directory contains the following files:
 - `nerdctl-full.tgz`: [`nerdctl-full-<VERSION>-linux-<ARCH>.tar.gz`](https://github.com/containerd/nerdctl/releases)
 - `boot.sh`: Boot script
 - `boot/*`: Boot script modules
+- `util/*`: Utility command scripts, executed in the boot script modules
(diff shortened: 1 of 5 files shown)


jandubois:
#383 mentions 3 conditions:

> fuse-overlayfs shouldn't be needed when (kernel >= 5.13) || (kernel >= 5.11 && !selinux) || (kernel is ubuntu's variant)

This PR seems to only implement the first one: kernel >= 5.13

AkihiroSuda:
> #383 mentions 3 conditions:
> 
> > fuse-overlayfs shouldn't be needed when (kernel >= 5.13) || (kernel >= 5.11 && !selinux) || (kernel is ubuntu's variant)
> 
> This PR seems to only implement the first one: kernel >= 5.13

Yes, for simplicity

jandubois (approved):

PR #700, opened 2022-03-08
Add support for oracle linux 8.

chrisperelstein:
Oracle's yum repos are fairly slow for me so this sometimes times out. More problematically, the containerd install portion doesn't work on ol8 and I haven't quite tracked the problem down yet.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -11,6 +11,7 @@ Distro:
 - [`debian.yaml`](./debian.yaml): ⭐Debian GNU/Linux
 - [`fedora.yaml`](./fedora.yaml): ⭐Fedora
 - [`opensuse.yaml`](./opensuse.yaml): ⭐openSUSE Leap
+- [`oraclelinux.yaml`](./oraclelinux.yaml): Oracle Linux
 - [`rocky.yaml`](./rocky.yaml): Rocky Linux
 - [`ubuntu.yaml`](./ubuntu.yaml): Ubuntu (same as `default.yaml` but without extra YAML lines)
 - [`ubuntu-lts.yaml`](./ubuntu-lts.yaml): Ubuntu LTS (same as `ubuntu.yaml` but pinned to an LTS version)
--- a/examples/oraclelinux.yaml
+++ b/examples/oraclelinux.yaml
@@ -1,0 +1,18 @@
+# Oracle image licence: https://www.oracle.com/downloads/licenses/oracle-linux-license.html
+# Image source: https://yum.oracle.com/oracle-linux-templates.html
+arch: "x86_64"
+
+images:
+- location: "https://yum.oracle.com/templates/OracleLinux/OL8/u5/x86_64/OL8U5_x86_64-olvm-b113.qcow2"
+  arch: "x86_64"
+  digest: "sha256:550db4100b38825d1778f23f8c457f3fc971d96be1326fb512a5e607032b2a24"
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+firmware:
+  legacyBIOS: true
+cpuType:
+  # Workaround for "vmx_write_mem: mmu_gva_to_gpa XXXXXXXXXXXXXXXX failed" on Intel Mac
+  # https://bugs.launchpad.net/qemu/+bug/1838390
+  x86_64: "Haswell-v4"
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
@@ -10,6 +10,14 @@ update_fuse_conf() {
 	fi
 }
 
+dnf_install() {
+	if grep -q "Oracle Linux Server release 8" /etc/system-release; then
+		dnf install --repo ol8_baseos_latest --repo ol8_codeready_builder "$@"
+	else
+		dnf install "$@"
+	fi
+}
+
 INSTALL_IPTABLES=0
 if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" = 1 ] || [ "${LIMA_CIDATA_CONTAINERD_USER}" = 1 ]; then
 	INSTALL_IPTABLES=1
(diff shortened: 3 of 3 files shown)


examples/oracle.yaml
```diff
@@ -0,0 +1,16 @@
+arch: "x86_64"
```
AkihiroSuda, line 1: No support for aarch64?

examples/oracle.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+images:
+- location: "https://yum.oracle.com/templates/OracleLinux/OL8/u5/x86_64/OL8U5_x86_64-olvm-b113.qcow2"
```
AkihiroSuda, line 4: Where to find the latest image?
Also, please link the license https://www.oracle.com/downloads/licenses/oracle-linux-license.html

examples/oracle.yaml
AkihiroSuda, line 1: Please link `oracle.yaml` in `examples/README.md` too

examples/oracle.yaml
AkihiroSuda: Googled "OL8U5_x86_64-olvm-b113.qcow2" but no result.

I guess this URL isn't expected to be directly linked?

examples/oracle.yaml
chrisperelstein: Oddly not. Oracle publishes arm docker images, install isos, and raspberry pi image, but not arm ova/qcow vm images.

examples/oracle.yaml
chrisperelstein: Will add license. I just used the links listed here: https://yum.oracle.com/oracle-linux-templates.html

examples/oracle.yaml
chrisperelstein: Fixed in the latest push: https://github.com/lima-vm/lima/pull/700/files#diff-49aaa2819e35a856818ecec8c9fa7e1c79ad028d3f44bd749736353cfb51bac9R14

examples/oracle.yaml
AkihiroSuda: Yes, comment in the YAML is probably fine

examples/README.md
```diff
@@ -11,6 +11,7 @@ Distro:
 - [`debian.yaml`](./debian.yaml): ⭐Debian GNU/Linux
 - [`fedora.yaml`](./fedora.yaml): ⭐Fedora
 - [`opensuse.yaml`](./opensuse.yaml): ⭐openSUSE Leap
+- [`oracle.yaml`](./oracle.yaml): Oracle Linux
```
AkihiroSuda, line 14: Maybe, the file name should be “oraclelinux.yaml”?
“oracle” means the database product to me.
(For the same reason the ArchLinux yaml is named “archlinux.yaml”, not “arch.yaml”)

examples/oracle.yaml
chrisperelstein: License: https://github.com/lima-vm/lima/pull/700/files#diff-ec074ef274eaabe9a71610791c8ed66c88b59ba7bb648390f2c34426d8130a07R1

examples/README.md
chrisperelstein: Yeah fair point. Moved the file and updated readme.

examples/oraclelinux.yaml
```diff
@@ -0,0 +1,17 @@
+# Licence: https://www.oracle.com/downloads/licenses/oracle-linux-license.html
```
AkihiroSuda, line 1: Please clarify that this license applies to the image, not to the YAML itself, which is licensed under the terms of Apache License 2.0

examples/oraclelinux.yaml
```diff
@@ -0,0 +1,17 @@
...
+arch: "x86_64"
+
+images:
+- location: "https://yum.oracle.com/templates/OracleLinux/OL8/u5/x86_64/OL8U5_x86_64-olvm-b113.qcow2"
```
AkihiroSuda, line 5: Could you add the link to https://yum.oracle.com/oracle-linux-templates.html as a comment line?

examples/README.md
```diff
@@ -11,6 +11,7 @@ Distro:
 - [`debian.yaml`](./debian.yaml): ⭐Debian GNU/Linux
 - [`fedora.yaml`](./fedora.yaml): ⭐Fedora
 - [`opensuse.yaml`](./opensuse.yaml): ⭐openSUSE Leap
+- [`oraclelinux.yaml`](./oracle.yaml): Oracle Linux
```
AkihiroSuda, line 14:
```suggestion
- [`oraclelinux.yaml`](./oraclelinux.yaml): Oracle Linux
```

chrisperelstein: Fixed in latest push.

examples/oraclelinux.yaml
chrisperelstein: Clarified: https://github.com/lima-vm/lima/pull/700/files#diff-ec074ef274eaabe9a71610791c8ed66c88b59ba7bb648390f2c34426d8130a07R1

AkihiroSuda (dismissed):
Thanks

chrisperelstein:
I'd like to propose a bit of a refactor that gets the containerd stuff working and improves the user experience by getting around some of Oracle's slow repos: https://github.com/chrisperelstein/lima/commit/9f69c67d966d9c374d9e28f578660f3a5f24cda6

* setting different flags for ol8 fixes slow repo issues but using an unquoted variable for this triggers https://github.com/koalaman/shellcheck/wiki/SC2086. i could do a bunch of conditionals but the wrapper function seemed cleaner.
* containerd/nerdctl install didn't work due to ol8 apparently not including tar, so added a block to ensure that's present

jandubois (approved):
Thanks, LGTM

jandubois:
> I think it may be much better.

It is better if it works better :)

I've added some stylistic comments on the commit.

> * including tar, so added a block to ensure that's present

I don't see any block dealing with `tar`.

chrisperelstein:
I addressed the comments and squashed everything down. This is the tar block I was referring to: https://github.com/lima-vm/lima/pull/700/files#diff-4437bfd3c932c2cb558d7385a2505f76bfeb6ec8675520940cf0fdb1b60239deR50. I've re-tested the current version locally and it worked for both Oracle Linux 8 and Fedora.

> > I think it may be much better.
> 
> It is better if it works better :)
> 
> I've added some stylistic comments on the commit.
> 
> > * including tar, so added a block to ensure that's present
> 
> I don't see any block dealing with `tar`.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -39,26 +47,29 @@ if command -v apt-get >/dev/null 2>&1; then
...
 	fi
 elif command -v dnf >/dev/null 2>&1; then
+	if ! command -v tar >/dev/null 2>&1; then
+		dnf_install -y tar
+	fi
 	if [ "${LIMA_CIDATA_MOUNTS}" -gt 0 ]; then
 		if ! command -v sshfs >/dev/null 2>&1; then
-			if grep -q "release 8" /etc/system-release; then
-				dnf install --enablerepo powertools -y fuse-sshfs
+			if grep -q "release 8" /etc/system-release && grep -qv "^Oracle" /etc/system-release; then
```
AkihiroSuda, line 55: Why not move this to `dnf_install()` ?

jandubois: How would you do that? Add `--enablerepo powertools` to **all** dnf install command that are not on Oracle?

Maybe that is fine, but then you need to comment that this is really only done for `fuse-sshfs`?

AkihiroSuda: Yes

chrisperelstein: This has a separate conditional due to the `--enablerepo powertools` part. That's not valid for Oracle Linux. If it's `release 8` and not Oracle, powertools is needed, if it's `release 8` and Oracle, ol8_codeready_builder is needed, otherwise we can just install sshfs directly. Combined with needing to keep repo flags limited for Oracle Linux due to repo/dnf performance, this seemed the best way. I could see an arguement for doing the same repo filtering for all rhel8 like linux releases though, so if you want, I can move that conditional into the dnf_install function.

AkihiroSuda (approved):

PR #24, opened 2021-05-19
Make it working on Apple Silicon

Tarik02:
Without `highmem=off` in `-machine` flag QEMU exits with error "qemu-system-aarch64: VCPU supports less PA bits (36) than requested by the memory map (40)".

PR #703, opened 2022-03-09
aarch64: unset highmem=off for QEMU >= 7.0

### LINKED ISSUE #680: Memory allocation fails on darwin arm64 with latest qemu
### Description

Starting the default with 4GiB fails:
`qemu-system-aarch64: Addressing limited to 32 bits, but memory exceeds it by 1073741824 bytes`

Need to reduce memory, to e.g. 2GiB.

Since https://github.com/qemu/qemu/commit/3715c251cc781b263fbd7f7c535fde7a4da401c3

Most likely the `highmem=off` parameter to qemu should be parametrized, and not used with later versions.

The reasons that caused it to be needed (for ARM) earlier, seems to no longer be true with QEMU 7.0.

This was: `QEMU emulator version 6.2.50 (v6.1.0-3974-gf6620d0b34-dirty)`

AkihiroSuda:
- QEMU >= 7.0 requires highmem=off NOT to be set, otherwise fails with "Addressing limited to 32 bits, but memory exceeds it by 1073741824 bytes"
- QEMU <  7.0 requires highmem=off to be set, otherwise fails with "VCPU supports less PA bits (36) than requested by the memory map (40)"

See #680 and #24

Fix #680

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -150,6 +150,10 @@ type features struct {
 	// e.g. "Available netdev backend types:\nsocket\nhubport\ntap\nuser\nvde\nbridge\vhost-user\n"
 	// Not machine-readable, but checking strings.Contains() should be fine.
 	NetdevHelp []byte
+	// MachineHelp is the output of `qemu-system-x86_64 -machine help`
+	// e.g. "Supported machines are:\nakita...\n...virt-6.2...\n...virt-7.0...\n...\n"
+	// Not machine-readable, but checking strings.Contains() should be fine.
+	MachineHelp []byte
 }
 
 func inspectFeatures(exe string) (*features, error) {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Tested on Intel Mac, with qemu/qemu@2ad7624900

@jandubois @afbjorklund PTAL

jandubois (dismissed):
I haven't tested it, but LGTM.

AkihiroSuda:
Can anybody test this with M1 macOS?

afbjorklund:
> Can anybody test this with M1 macOS?

I can see if I can find a machine tonight, we're doing some testing of the minikube ISO (for aarch64)

AkihiroSuda:
Merging and releasing v0.9.0-beta.0 for ease of testing

PR #706, opened 2022-03-09
cidata: Install packages at once

AkihiroSuda:


pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -28,89 +20,93 @@ fi
...
 # Install minimum dependencies
 if command -v apt-get >/dev/null 2>&1; then
-	DEBIAN_FRONTEND=noninteractive
-	export DEBIAN_FRONTEND
-	apt-get update
-	if [ "${LIMA_CIDATA_MOUNTS}" -gt 0 ]; then
-		if ! command -v sshfs >/dev/null 2>&1; then
-			apt-get install -y sshfs
-		fi
+	pkgs=
```
jandubois, line 23: This is just personal preference, but a dangling `=` sign looks to me like the right side might have been forgotten, or accidentally deleted. So I prefer to explicitly set to the empty string:

```suggestion
	pkgs=""
```

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -28,89 +20,93 @@ fi
...
+	if [ "${LIMA_CIDATA_MOUNTS}" -gt 0 ] && ! command -v sshfs >/dev/null 2>&1; then
```
jandubois, line 24: I wonder if `command -v $1 >/dev/null 2>&1` isn't used often enough to turn it into a function

```shell
function exists() { command -v $1 >/dev/null 2>&1; }
```

and then you can shorten conditions everywhere

```suggestion
	if [ "${LIMA_CIDATA_MOUNTS}" -gt 0 ] && ! exists sshfs; then
```

jandubois (dismissed):
LGTM

I cannot concentrate enough right now to verify that the `dnf_install_flags` settings are equivalent to the old logic, but everything else seems fine.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
AkihiroSuda: Fixed

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
AkihiroSuda: I think ``command -v` is short enough

jandubois: > I think ``command -v` is short enough

Well, it also hides the ugly (to me) `>/dev/null 2>&1` redirection, but yeah, it is just aesthetics.

AkihiroSuda: We can add that `function` (or a separate shell script under `/mnt/lima-cidata/util`), but we should do it for all `*.sh` at once (in another PR)

PR #707, opened 2022-03-09
examples/default.yaml: clarify YAML default vs builtin default

AkihiroSuda:


examples/default.yaml
```diff
@@ -2,17 +2,16 @@
...
-# Default: "default" (corresponds to the host architecture)
+# 🟢 Builtin default: "default" (corresponds to the host architecture)
 arch: null
 
-# An image must support systemd and cloud-init.
-# Ubuntu and Fedora are known to work.
-# Default: none (must be specified)
+# OpenStack-compatible disk image.
+# 🟢 Builtin default: null (must be specified)
+# 🔵 YAML default: Ubuntu 21.10 Impish Indri images
```
AkihiroSuda, line 14: "YAML default" might be still confusing, but can't come up with a better terminology

jandubois: Yeah, I don't know, "YAML default:" sounds weird to me.

Maybe "This file:"?

Still weird, but at least makes sure that it is the value from this specific file, and not a generic "YAML default" (what does that even mean?). Like, if there is a "YAML default", could there be a different "JSON default", or "commandline default", or what?

examples/default.yaml
```diff
@@ -28,37 +27,38 @@ images:
...
 # Expose host directories to the guest, the mount point might be accessible from all UIDs in the guest
-# Default: null
+# 🟢 Builtin default: null (Mount nothing)
+# 🔵 YAML default: Mount the home as read-only, /tmp/lima as writable
 mounts:
 - location: "~"
   # CAUTION: `writable` SHOULD be false for the home directory.
   # Setting `writable` to true is possible, but untested and dangerous.
-  # Default: false
+  # Builtin default: false
```
jandubois, line 48: Why no 🟢 for this one?

examples/default.yaml
AkihiroSuda: Changed to "This file"

examples/default.yaml
AkihiroSuda: fixed

examples/default.yaml
```diff
@@ -85,20 +87,21 @@ ssh:
...
-  # Default: true
+  # 🟢 Builtin default: true
   user: null
 #  # Override containerd archive
-#  # Default: hard-coded URL with hard-coded digest (see the output of `limactl info | jq .defaultTemplate.containerd.archives`)
+#  # 🟢 Builtin default: hard-coded URL with hard-coded digest (see the output of `limactl info | jq .defaultTemplate.containerd.archives`)
 #  archives:
 #  - location: "~/Downloads/nerdctl-full-X.Y.Z-linux-amd64.tar.gz"
 #    arch: "x86_64"
 #    digest: "sha256:..."
```
jandubois, line 100: I wonder if this would look better as:

```yaml
containerd:
  [...]
  user: null
  # Override containerd archive
  # 🟢 Builtin default: hard-coded URL with hard-coded digest (see the output of `limactl info | jq .defaultTemplate.containerd.archives`)
  archives:
  # - location: "~/Downloads/nerdctl-full-X.Y.Z-linux-amd64.tar.gz"
  #   arch: "x86_64"
  #   digest: "sha256:..."
```

and of course the same for every empty list. This way the property name shows up more prominently, especially with syntax highlighting. And we get rid of the awkward double comments with `# # comment`.

Maybe should go into a separate PR? I can give it a try tomorrow, if you want me to.

jandubois (approved):

examples/default.yaml
AkihiroSuda: Thanks, I guess we can try it in a separate PR

PR #710, opened 2022-03-09
CI: reduce GHA parallelism to avoid limit

AkihiroSuda:


AkihiroSuda:
@jandubois 

I'm going to merge this and update https://github.com/lima-vm/lima/settings/branches , but feel free to revert if you don't like this

AkihiroSuda:
@jandubois 

I'm going to merge this and update https://github.com/lima-vm/lima/settings/branches , but feel free to revert if you don't like this

PR #718, opened 2022-03-11
Sync the containerd files for Kubernetes

afbjorklund:
Comparing with:
~~https://github.com/containerd/containerd/blob/main/docs/cri/installation.md~~

Unfortunately the upstream documentation has been deprecated,
and no longer includes the configuration needed for CRI and CNI.

It is also no longer available from the kubernetes [documentation](https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd),
so it is up to the user to figure out /etc/crictl.yaml and /etc/cni/net.d

https://github.com/containerd/containerd/blob/main/script/setup/install-critools

https://github.com/containerd/containerd/blob/main/script/setup/install-cni

Upstream includes configuration for CRI and CNI,
so make sure that is added to the nerdctl install.
Currently using "[flannel](https://github.com/flannel-io/flannel)" for multi-node and VXLAN,
but template can be edited to use "containerd-net".

The image endpoint defaults to the runtime endpoint,
so use the same configuration file as upstream has.
Currently we use `kubernetes-cni` which is a slightly
older version so don't bump the `cniVersion` (yet).

----

Reviewer note: this is basically a no-op, unless edited

We create the default file, and then we delete it again

### DIFF
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -54,10 +54,46 @@ provision:
     apt-get install -y cri-tools
     cat  <<EOF | sudo tee /etc/crictl.yaml
     runtime-endpoint: unix:///run/containerd/containerd.sock
-    image-endpoint: unix:///run/containerd/containerd.sock
     EOF
     # cni-plugins
     apt-get install -y kubernetes-cni
+    mkdir -p /etc/cni/net.d
+    cat << EOF | tee /etc/cni/net.d/10-containerd-net.conflist
+    {
+      "cniVersion": "0.4.0",
+      "name": "containerd-net",
+      "plugins": [
+        {
+          "type": "bridge",
+          "bridge": "cni0",
+          "isGateway": true,
+          "ipMasq": true,
+          "promiscMode": true,
+          "ipam": {
+            "type": "host-local",
+            "ranges": [
+              [{
+                "subnet": "10.88.0.0/16"
+              }],
+              [{
+                "subnet": "2001:4860:4860::/64"
+              }]
+            ],
+            "routes": [
+              { "dst": "0.0.0.0/0" },
+              { "dst": "::/0" }
+            ]
+          }
+        },
+        {
+          "type": "portmap",
+          "capabilities": {"portMappings": true}
+        }
+      ]
+    }
+    EOF
+    # To use flannel, delete the default CNI network
+    # To use containerd-net, comment this and below
     rm -f /etc/cni/net.d/*.conf*
     apt-get install -y kubelet kubeadm kubectl && apt-mark hold kubelet kubeadm kubectl
     systemctl enable --now kubelet


examples/k8s.yaml
```diff
@@ -54,10 +54,46 @@ provision:
...
+        {
+          "type": "portmap",
+          "capabilities": {"portMappings": true}
+        }
+      ]
+    }
+    EOF
+    # To use flannel, delete the default CNI network
+    # To use containerd-net, comment this and below
     rm -f /etc/cni/net.d/*.conf*
```
AkihiroSuda, line 97: This rm -f is confusing

afbjorklund: I think I inherited from the installation where we also have podman and cri-o

Perhaps remove the glob and use the above only ? Hope nothing else is there.
Kubernetes only supports having **one** file in CNI, and uses the first in alphabet.

afbjorklund: In the recent release, the cni conf dir flag has been removed - which broke our workaround

afbjorklund: Anyway, we get a default CNI with the container runtime installation.

It must be deleted, if we want to install our own CNI - like "flannel"

AkihiroSuda: I think we should only support flannel.
Not really useful for single-node cluster, but akin to production env.

afbjorklund:
The end goal here was to make the installation more similar as to what is documented upstream:

https://kubernetes.io/docs/setup/production-environment/container-runtimes/#containerd

There is a new version of this document for k8s 1.24, changing from docker.com to containerd.io

https://deploy-preview-32738--kubernetes-io-main-staging.netlify.app/docs/setup/production-environment/container-runtimes/#containerd

And it recommends using the "cri-containerd-cni" tarball:

https://containerd.io/docs/getting-started/#starting-containerd

i.e. including the /etc/crictl.yaml and /etc/cni/net.d (and plugins)

But the configuration is **not** included, in the deb packages...

afbjorklund:
When **not** using flannel, then the pod-network-cidr should be removed too (to default to 10.88.0.0/16)

Ideally this should have some kind of boolean configuration, but I think that is overkill for a yaml file...

### containerd

The default configuration in `/etc/cni/net.d/10-containerd-net.conflist` works, without modifications.

```
    systemctl enable --now kubelet
```

```
    kubeadm init --cri-socket=/run/containerd/containerd.sock --apiserver-cert-extra-sans 127.0.0.1
```

### flannel

The default bridge CNI must be **deleted**, and the flannel pod will create one once the cluster is booted.

```
    rm -f /etc/cni/net.d/*.conf*

    systemctl enable --now kubelet
```

```
    # Initializing your control-plane node
    kubeadm init --cri-socket=/run/containerd/containerd.sock --pod-network-cidr=10.244.0.0/16 --apiserver-cert-extra-sans 127.0.0.1
    # Installing a Pod network add-on
    kubectl apply -f https://raw.githubusercontent.com/flannel-io/flannel/v0.14.0/Documentation/kube-flannel.yml
```

afbjorklund:
Side note: the new version of flannel will install _another_ copy of cni-plugins (1.0.1), because the kubernetes-cni are too old (0.8.7)

It is copied to the host from the container image, overwriting any existing `/opt/cni/bin/flannel`.

```
REPOSITORY                           TAG        IMAGE ID       CREATED         SIZE
flannelcni/flannel                   v0.17.0    9247abf08677   2 weeks ago     59.8MB
flannelcni/flannel-cni-plugin        v1.0.1     ac40ce625740   8 weeks ago     8.1MB
```

https://raw.githubusercontent.com/flannel-io/flannel/master/Documentation/kube-flannel.yml

```yaml
      initContainers:
      - name: install-cni-plugin
       #image: flannelcni/flannel-cni-plugin:v1.0.1 for ppc64le and mips64le (dockerhub limitations may apply)
        image: rancher/mirrored-flannelcni-flannel-cni-plugin:v1.0.1
        command:
        - cp
        args:
        - -f
        - /flannel
        - /opt/cni/bin/flannel
        volumeMounts:
        - name: cni-plugin
          mountPath: /opt/cni/bin
      - name: install-cni
       #image: flannelcni/flannel:v0.17.0 for ppc64le and mips64le (dockerhub limitations may apply)
        image: rancher/mirrored-flannelcni-flannel:v0.17.0
        command:
        - cp
        args:
        - -f
        - /etc/kube-flannel/cni-conf.json
        - /etc/cni/net.d/10-flannel.conflist
        volumeMounts:
        - name: cni
          mountPath: /etc/cni/net.d
        - name: flannel-cfg

```

afbjorklund:
The deployment for flannel is not included in the release, so installing 0.17.0 actually installs 0.16.3

https://raw.githubusercontent.com/flannel-io/flannel/v0.17.0/Documentation/kube-flannel.yml

```yaml
      containers:
      - name: kube-flannel
       #image: flannelcni/flannel:v0.16.3 for ppc64le and mips64le (dockerhub limitations may apply)
        image: rancher/mirrored-flannelcni-flannel:v0.16.3
```

examples/k8s.yaml
afbjorklund: Main reason for containerd-net was to not depend on any third-party images

afbjorklund:
https://raw.githubusercontent.com/flannel-io/flannel/v0.17.0-1-gbf9dbdb0/Documentation/kube-flannel.yml

examples/k8s.yaml
```diff
@@ -54,10 +54,46 @@ provision:
     apt-get install -y cri-tools
     cat  <<EOF | sudo tee /etc/crictl.yaml
     runtime-endpoint: unix:///run/containerd/containerd.sock
-    image-endpoint: unix:///run/containerd/containerd.sock
     EOF
     # cni-plugins
     apt-get install -y kubernetes-cni
+    mkdir -p /etc/cni/net.d
+    cat << EOF | tee /etc/cni/net.d/10-containerd-net.conflist
```
AkihiroSuda, line 61: Looks like we are going to support multi-node example (https://github.com/lima-vm/lima/pull/1183), so probably we do not need to use non-flannel bridge?

afbjorklund: It would have to be versioned, at least. That is, only apply when number of nodes == 1. Otherwise flannel.

AkihiroSuda: Is it harmless to use flannel on single-node mode too?

afbjorklund: Yeah, just overkill. Mostly it was triggered by the kubeadm team not wanting to support flannel anymore.

afbjorklund:
I guess it can default to using flannel on both Kubernetes distributions, as a Lima feature.

Upstream containerd seems to be deprecating the bundled CNI configuration anyway...

> https://github.com/containerd/containerd/blob/main/docs/getting-started.md
> 
> The cri-containerd-... archives are deprecated, do not work on old Linux distributions, and will be removed in containerd 2.0.

So in case the bridge network doesn't work anymore then, don't add another copy here.

Will do a new PR, just to remove the redundant (default) line from the CRI configuration...

```diff
@@ -62,7 +62,6 @@ provision:
     apt-get install -y cri-tools
     cat  <<EOF | sudo tee /etc/crictl.yaml
     runtime-endpoint: unix:///run/containerd/containerd.sock
-    image-endpoint: unix:///run/containerd/containerd.sock
     EOF
     # cni-plugins
     apt-get install -y kubernetes-cni

```

examples/k8s.yaml
afbjorklund: The registry changed from k8s.gcr.io to registry.k8s.io and from quay.io to docker.io, but anyway.

```
IMAGE                                     TAG                 IMAGE ID            SIZE
docker.io/flannel/flannel-cni-plugin      v1.1.2              7a2dcab94698c       3.84MB
docker.io/flannel/flannel                 v0.21.2             7b7f3acab868d       24.2MB
registry.k8s.io/coredns/coredns           v1.9.3              5185b96f0becf       14.8MB
registry.k8s.io/etcd                      3.5.6-0             fce326961ae2d       103MB
registry.k8s.io/kube-apiserver            v1.26.1             deb04688c4a35       35.3MB
registry.k8s.io/kube-controller-manager   v1.26.1             e9c08e11b07f6       32.2MB
registry.k8s.io/kube-proxy                v1.26.1             46a6bb3c77ce0       21.5MB
registry.k8s.io/kube-scheduler            v1.26.1             655493523f607       17.5MB
registry.k8s.io/pause                     3.9                 e6f1816883972       322kB

```

afbjorklund:
* #1384

Issue #287, opened 2021-10-04
There should be an option to run provisioning scripts after mounts are done


jandubois:
Right now provisioning scripts cannot use files from the host filesystem because the mounts are created by the hostagent on request of the guestagent. There should be a mechanism to run additional provisioning scripts once the instance is deemed to be fully initialized.

AkihiroSuda:
Provisioning scripts should be executed always after mounts are setup

jandubois:
> Is this issue still accurate?

I believe this issue is accurate when you are using `mountType: reverse-sshfs`. If you are using e.g. `9p` then the mounts are already in place before the provisioning scripts run.

PR #720, opened 2022-03-11
Allow RHEL-like distros below version 8

alisson276:
This MR allows you create VMs using CentOS 7

My centos7.yaml file:
```yaml
images:
- location: "https://cloud.centos.org/altarch/7/images/CentOS-7-x86_64-GenericCloud-2009.qcow2"
  arch: "x86_64"
  digest: "sha256:e38bab0475cc6d004d2e17015969c659e5a308111851b0e2715e84646035bdd3"
- location: "https://cloud.centos.org/altarch/7/images/CentOS-7-aarch64-GenericCloud-2009.qcow2"
  arch: "aarch64"
  digest: "sha256:51c0222aa4bc7d966fc044eb6ce9182993a1dc398eaa595e58abd0d361439baf"
containerd:
  system: false
  user: false
mounts:
- location: "~"
- location: "/tmp/lima"
  writable: true
firmware:
  legacyBIOS: true
cpuType:
  # Workaround for "vmx_write_mem: mmu_gva_to_gpa XXXXXXXXXXXXXXXX failed" on Intel Mac
  # https://bugs.launchpad.net/qemu/+bug/1838390
  x86_64: "Haswell-v4
```


alisson276:
Yeah, you're right (unfortunately). I'm closing this one. I will just fork lima and use on my own repo since the maintainer doesn't want keep support for CentOS 7 and older Fedora that don't have dnf yet.

afbjorklund:
I don't think using "yum" was the main issue, but yeah

afbjorklund:
@alisson276 if you really want CentOS 7 to be supported, I think opening a new issue and describing why would be helpful

alisson276:
Never mind. I solved my problem doing this:
```
provision:
  - mode: system
    script: |
      #!/bin/bash
      set -eux -o pipefail
      pkgs=""
      yum_install_flags="-y"
      if ! rpm -ql epel-release > /dev/null 2>&1; then
          yum install ${yum_install_flags} epel-release
      fi
      if ! command -v dnf >/dev/null 2>&1; then
          pkgs="${pkgs} dnf"
      fi
      if ! command -v sshfs >/dev/null 2>&1; then
          pkgs="${pkgs} fuse-sshfs"
      fi
      if ! command -v mount.fuse3 >/dev/null 2>&1; then
          pkgs="${pkgs} fuse3"
      fi
      if [ -n "${pkgs}" ]; then
          yum install ${yum_install_flags} ${pkgs}
      fi
```

My use case is very particular, we still use CentOS7 in our company (we have a lot of work to migrate everything to AlmaLinux yet) and some people started to use Macs with Apple M1 chip, so we don't have vagrant and virtualbox available. Then we use lima to create test virtual machines, actually.

alisson276:
Thank you for the warning @jandubois , actually I have a:
```
      until ls /srv/lima/provision &> /dev/null; do
          echo "Waiting for the filesystem to be mounted..."
          sleep 3
      done
```
in my provision code exactly to wait for this... It will break my machines 😭

jandubois:
I've added a comment to #287 that we should support both kinds of provisioning scripts: the ones we have right now, that run as part of cloud-init, and the new kind that would be triggered by the hostagent once the mounts are setup.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -71,6 +71,27 @@ elif command -v dnf >/dev/null 2>&1; then
...
+	fi
+	if [ "${LIMA_CIDATA_CONTAINERD_USER}" = 1 ]; then
+		if ! command -v mount.fuse3 >/dev/null 2>&1; then
+			pkgs="${pkgs} fuse3"
+		fi
+	fi
+	if [ -n "${pkgs}" ]; then
+		# shellcheck disable=SC2086
+		yum install ${yum_install_flags} ${pkgs}
+	fi
```
AkihiroSuda, line 94: Why not install other packages such as shadow-utils and iptables?

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -71,6 +71,27 @@ elif command -v dnf >/dev/null 2>&1; then
...
 		ln -s fusermount3 /usr/bin/fusermount
 	fi
+elif command -v yum >/dev/null 2>&1; then
+	pkgs=""
+	yum_install_flags="-y"
+	if ! rpm -ql epel-release >/dev/null 2>&1; then
+		yum install ${yum_install_flags} epel-release
+	fi
+	if ! command -v dnf >/dev/null 2>&1; then
+		pkgs="${pkgs} dnf"
```
AkihiroSuda, line 81: Why?

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -71,6 +71,27 @@ elif command -v dnf >/dev/null 2>&1; then
 		# Workaround for https://github.com/containerd/stargz-snapshotter/issues/340
 		ln -s fusermount3 /usr/bin/fusermount
 	fi
+elif command -v yum >/dev/null 2>&1; then
```
AkihiroSuda, line 74: Please add a comment line to explain that CentOS 7 support is deprecated

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
alisson276: Yeah, it doesn't make sense. I added to take advantage of the `elif command -v dnf` (when it would be updated). Removing.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
alisson276: This packages was already there by default, that's why I didn't feel the need to add them. Do you think it's relevant anyway? The only situation it's a problem is if someone decides to remove the package (no matter why).

AkihiroSuda: I'd prefer to add them for consistency with EL8 family.
Also, some variant of EL7 may not have them installed by default.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
alisson276: Done!

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
alisson276: Alright. Added.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -71,6 +71,34 @@ elif command -v dnf >/dev/null 2>&1; then
...
+    echo "DEPRECATED: CentOS7 and others RHEL-like version 7 are unsupported and might be removed in future lima releases"
```
AkihiroSuda, line 75: Indent seems broken.
Also, could you squash commits?

afbjorklund:
Side note: EPEL needs to be enabled also for EL9, since it doesn't have "sshfs" in the regular repositories

AkihiroSuda:
> Side note: EPEL needs to be enabled also for EL9, since it doesn't have "fuse-sshfs" in the regular repositories
> 
> `dnf install epel-release epel-next-release`
> 
> Separate issue, though.
> 
> Not convinced that modifying the standard config is a _good_ thing, but...
> 
> Maybe config can be set up, to only install sshfs ? (from EPEL)
> 
> Or disable EPEL after install, but that will also block updates to sshfs

Please open a separate issue/PR

AkihiroSuda:
Looks good but please squash commits (https://www.git-tower.com/learn/git/faq/git-squash)

afbjorklund:
> Please open a separate issue/PR

I thought the plan was to ignore CentOS Stream and use Rocky etc instead ?

Just wondered if it was OK to add third-party repos like this, to the config ?

AkihiroSuda:
> I thought the plan was to ignore CentOS Stream and use Rocky etc instead ?

We can unofficially support Stream (in another PR)

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
alisson276: Will do.

AkihiroSuda (approved):
Thanks

alisson276:
Should I add the centos7.yaml?

AkihiroSuda:
> Should I add the centos7.yaml?

The file name should be like `deprecated-centos-7.yaml` or `deprecated/centos-7.yaml`

examples/README.md
```diff
@@ -33,6 +33,7 @@ Container orchestration:
 
 Others:
 - [`vmnet.yaml`](./vmnet.yaml): ⭐enable [`vmnet.framework`](../docs/network.md)
+- [`centos-7.yaml`](./deprecated/centos-7.yaml): [deprecated] CentOS 7
```
AkihiroSuda, line 36:
```suggestion
- [`deprecated/centos-7.yaml`](./deprecated/centos-7.yaml): [deprecated] CentOS 7
```

examples/deprecated/centos-7.yaml
```diff
@@ -0,0 +1,22 @@
...
+images:
+- location: "https://cloud.centos.org/altarch/7/images/CentOS-7-x86_64-GenericCloud-2009.qcow2"
+  arch: "x86_64"
+  digest: "sha256:e38bab0475cc6d004d2e17015969c659e5a308111851b0e2715e84646035bdd3"
+- location: "https://cloud.centos.org/altarch/7/images/CentOS-7-aarch64-GenericCloud-2009.qcow2"
+  arch: "aarch64"
+  digest: "sha256:51c0222aa4bc7d966fc044eb6ce9182993a1dc398eaa595e58abd0d361439baf"
+containerd:
+  system: false
+  user: false
```
AkihiroSuda, line 12: Please explain the reason. IIRC, CentOS 7 lacked user-scoped systemd?

examples/deprecated/centos-7.yaml
```diff
@@ -0,0 +1,22 @@
...
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+firmware:
+  legacyBIOS: true
+cpuType:
+  # Workaround for "vmx_write_mem: mmu_gva_to_gpa XXXXXXXXXXXXXXXX failed" on Intel Mac
+  # https://bugs.launchpad.net/qemu/+bug/1838390
+  x86_64: "Haswell-v4
```
AkihiroSuda, line 22:
```suggestion
  x86_64: "Haswell-v4"
```

examples/README.md
alisson276: Solved.

examples/deprecated/centos-7.yaml
alisson276: It has. Let me remove this and test just to be sure.

alisson276: You're right. I've just tried use nerdctl in the instance and got a `containerd-rootless-setuptool.sh install`, but CentOS7 lack of support to `systemctl --user`. Using `containerd:system` as true works, but only if you use as root. I will comment in the file

alisson276: Could you check again?

AkihiroSuda (dismissed):
Thanks

afbjorklund:
@alisson276 do you think EPEL should be left enabled ? Or should the VM only have the OS repos ?

alisson276:
> @alisson276 do you think EPEL should be left enabled ? Or should the VM only have the OS repos ?

I don't see as a problem let it enabled. Most of the "good" packages come from there since EL, in general, has a more conservative approach, but I can set it to disable if you prefer. I can add a `yum-config-manager --disable epel`

afbjorklund:
I wasn't sure, that is why I asked. I recall having two different templates, one "centos" and one "epel"

So was more wondering what VM users would expect ?

If it was up to me, I should probably have installed but disabled. And use it with `--enablerepo`

But I haven't used CentOS actively in years, either

alisson276:
> I wasn't sure, that is why I asked.

Just to be more conservative I've set to disabled `epel` right after installing fuse-sshfs. It can be enabled by the user anytime.

AkihiroSuda:
> New changes since you last viewed
> [Merge branch 'lima-vm:master' into master](https://github.com/lima-vm/lima/pull/720/commits/47068dfa8fafbaaf20b63deb68b55ac6c033a0d2)

Still LGTM but please remove this

alisson276:
> Still LGTM but please remove this

I think it's ok now.

jandubois:
> I think it's ok now.

No, it is not; it has other commits merged in, and then squashed, e.g. https://github.com/lima-vm/lima/commit/45c98e71db5202a19b77fda0156101d4ea7a9110.

See also the merge conflicts in `go.{mod,sum}`.

I would recommend to rebase the PR on latest `master`

alisson276:
> No, it is not; it has other commits merged in, and then squashed, e.g. [45c98e7](https://github.com/lima-vm/lima/commit/45c98e71db5202a19b77fda0156101d4ea7a9110).
> 
> See also the merge conflicts in `go.{mod,sum}`.
> 
> I would recommend to rebase the PR on latest `master`

Sorry, I didn't check the "Files changed". I just rebased based on the upstream. Is this right now?

AkihiroSuda (approved):
Thanks

examples/deprecated/centos-7.yaml
```diff
@@ -0,0 +1,24 @@
+# This example requires Lima v0.10.0 or later.
+# WARNING: RHEL-like below version 8 is unsupported, use at your own risk!
+images:
+- location: "https://cloud.centos.org/altarch/7/images/CentOS-7-x86_64-GenericCloud-2009.qcow2"
```
AkihiroSuda, line 4: Was there any specific reason to avoid 2111?
https://cloud.centos.org/altarch/7/images/CentOS-7-x86_64-GenericCloud-2111.qcow2

alisson276: Actually no. The only reason I didn't put was I thought all images should have a checksum to compare against and that image doesn't have on https://cloud.centos.org/altarch/7/images/sha256sum.txt. Also, there is no centos-7.2111 ISO. My conclusion was the image 2111 was an error.

I just test it and it worked. In fact, the build 2111 is associate with CentOS 8, as you can see on https://wiki.centos.org/Manuals/ReleaseNotes/CentOS8.2111, but it is Centos 7 as I could confirm.

AkihiroSuda: Thanks, I agree it might be an error 👀

PR #732, opened 2022-03-14
Fix curl url

Junnplus:

URL copied to the terminal will be escaped.


AkihiroSuda (approved):
Thanks

Issue #743, opened 2022-03-20
Support start cmd with recreate flag


Junnplus:
### Description

Now there is no way for us to use the existing yaml to recreate a vm, we can only manually clear the existing files.

Consider introducing a recreate flag for start cmd.
```
limactl start k8s --recreate
```

PR #744, opened 2022-03-20
Add recreate flag for start cmd

Junnplus:

Fixes: https://github.com/lima-vm/lima/issues/743

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,7 @@ Linux
 ### Command reference
 
 #### `limactl start`
-`limactl start [--name=NAME] [--tty=false] <template://TEMPLATE>`: start the Linux instance
+`limactl start [--name=NAME] [--tty=false] [--recreate=false] <template://TEMPLATE>`: start the Linux instance
 
 ```console
 $ limactl start
@@ -192,6 +192,11 @@ $ limactl start --name=default /usr/local/share/lima/examples/fedora.yaml
 To create an instance "default" from a remote URL (use carefully, with a trustable source):
 ```console
 $ limactl start --name=default https://raw.githubusercontent.com/lima-vm/lima/master/examples/alpine.yaml
+```
+
+To recreate an existing instance:
+```console
+$ limactl start --recreate
 ```
 
 #### `limactl shell`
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -194,6 +194,11 @@ To create an instance "default" from a remote URL (use carefully, with a trustab
 $ limactl start --name=default https://raw.githubusercontent.com/lima-vm/lima/master/examples/alpine.yaml
 ```
 
+To recreate an existing instance:
```
jandubois, line 197: Maybe it is just me, but I would want to be explicit that this action will delete the existing instance. In some ways it is implied, but I think it doesn't hurt to spell it out that all content is lost:

```suggestion
To destroy an instance and recreate it with identical settings (resets the content of the instance):
```

jandubois (commented):
The `--recreate` option should either stop a running instance, or issue an error message that tells the user to stop the instance before it can be recreated. The current code simply ignores the `--recreate` option:

```console
$ limactl start default --recreate
INFO[0000] The instance "default" is already running. Run `lima` to open the shell.
```

AkihiroSuda:
On second thought, maybe we should have this as `limactl factory-reset INSTANCE` rather than overloading `limactl start` command?

jandubois:
> maybe we should have this as `limactl factory-reset INSTANCE` rather than overloading `limactl start` command?

I'm fine with this PR now, but I also prefer to use a separate command. We already overload `start` with 2 meanings:

1. Create a new instance and start it
2. Start an existing instance

So maybe adding a 3rd meaning which will actually destroy (and then recreate) an instance is a step too far. 😄 

Also the word "factory-reset" makes the destructive nature much clearer than the `--recreate` parameter, at least in my mind.

jandubois (approved):
Approving, because the PR is fine now, but I won't merge because I also prefer to use `limactl factory-reset` instead.

Junnplus:
How about adding a `limactl reset` command? @AkihiroSuda @jandubois

AkihiroSuda:
> How about adding a `limactl reset` command? @AkihiroSuda @jandubois

That may surprise some users who except `reset` to just reboot the instance without wiping the data

AkihiroSuda:
Maybe `limactl reset-data`?

jandubois:
> Maybe `limactl reset-data`?

I still prefer "factory reset". It may feel a bit weird applied to pure software, but I think everybody intuitively understands it means "reset everything as if you just opened the box", which is precisely what it does.

Junnplus:
I remade a PR https://github.com/lima-vm/lima/pull/757 for `limactl factory-reset`, close the current.

PR #754, opened 2022-03-24
generate qemu vm uuid

dictcp:
# Background

Some application may requires the `product_uuid` from sysfs, OR it will be broken, eg.

- https://github.com/nestybox/sysbox/issues/439
- https://bugs.launchpad.net/ubuntu/+source/qemu-kvm/+bug/959308

The patch will generate a UUID, based of the MD5 sum of the VM name, and append as qemu `--uuid` parameter.
It should be safe and no impact to others usecase.

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -22,6 +22,8 @@ import (
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/mattn/go-shellwords"
 	"github.com/sirupsen/logrus"
+
+	"github.com/google/uuid"
 )
 
 type Config struct {
@@ -425,6 +427,7 @@ func Cmdline(cfg Config) (string, []string, error) {
 
 	// QEMU process
 	args = append(args, "-name", "lima-"+cfg.Name)
+	args = append(args, "-uuid", uuid.NewMD5(uuid.NameSpaceOID, []byte(cfg.Name)).String() )
 	args = append(args, "-pidfile", filepath.Join(cfg.InstanceDir, filenames.QemuPID))
 
 	return exe, args, nil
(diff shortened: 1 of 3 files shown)


pkg/qemu/qemu.go
```diff
@@ -425,6 +427,7 @@ func Cmdline(cfg Config) (string, []string, error) {
 
 	// QEMU process
 	args = append(args, "-name", "lima-"+cfg.Name)
+	args = append(args, "-uuid", uuid.NewMD5(uuid.NameSpaceOID, []byte(cfg.Name)).String() )
```
AkihiroSuda, line 430: For universal uniqueness, can we generate a random UUID and save it as “uuid” file in the instance dir?
https://github.com/lima-vm/lima/blob/3aca16d2a762fdb1ba9f50d6749cb03240d0a963/pkg/store/filenames/filenames.go#L25-L45

AkihiroSuda:
https://github.com/apps/dco

jandubois:
To have a globally unique, but persistent (accross VM restarts) id we include both the machine id and the full path to the VM directory:

https://github.com/lima-vm/lima/blob/3aca16d2a762fdb1ba9f50d6749cb03240d0a963/pkg/limayaml/defaults.go#L44

The instance dir alone isn't even locally unique because you can have multiple `LIMA_HOME` settings on the same machine.

I would suggest to extract the `sha` function from `limayaml.MACAddress` into a separate function and also use it as input for the qemu UUID. The `uniqueID` passed into the function is the instance directory.

pkg/qemu/qemu.go
AkihiroSuda: ( no need to save the uuid as a file if you adopt Jan’s approach https://github.com/lima-vm/lima/pull/754#issuecomment-1077774078 )

junnplus:
@dictcp Hi, are you still working on this PR? I need it too.

AkihiroSuda:
@junnplus Feel free to open a new PR to carry this (if you would like to)

PR #756, opened 2022-03-25
nerdctl: update to v0.18.0; configure BuildKit with containerd worker

AkihiroSuda:
BuildKit daemon is now configured to use containerd worker, not OCI worker for efficiency and for supporting using local base images

https://github.com/containerd/nerdctl/blob/v0.18.0/docs/build.md#setting-up-buildkit-with-containerd-worker

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
@@ -33,14 +33,27 @@ fi
 
 rm -rf "${tmp_extract_nerdctl}"
 
+: "${CONTAINERD_NAMESPACE:=default}"
+# Overridable in .bashrc
+: "${CONTAINERD_SNAPSHOTTER:=overlayfs}"
+
 if [ "${LIMA_CIDATA_CONTAINERD_SYSTEM}" = 1 ]; then
-	mkdir -p /etc/containerd
+	mkdir -p /etc/containerd /etc/buildkit
 	cat >"/etc/containerd/config.toml" <<EOF
   version = 2
   [proxy_plugins]
     [proxy_plugins."stargz"]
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
@ktock PTAL

ktock:
LGTM

ktock:
Maybe we should add the following example to the integration test.

https://github.com/containerd/nerdctl/blob/v0.18.0-beta.0/docs/build.md

```console
$ mkdir -p /tmp/ctx && cat <<EOF > /tmp/ctx/Dockerfile
FROM ghcr.io/stargz-containers/ubuntu:20.04-org
RUN echo hello
EOF
$ nerdctl build -t foo /tmp/ctx
$ cat <<EOF > /tmp/ctx/Dockerfile
FROM foo
RUN echo bar
EOF
$ nerdctl build -t bar /tmp/ctx
````

AkihiroSuda:
> Maybe we should add the following example to the integration test.
> 
> https://github.com/containerd/nerdctl/blob/v0.18.0-beta.0/docs/build.md
> 
> ```
> $ mkdir -p /tmp/ctx && cat <<EOF > /tmp/ctx/Dockerfile
> FROM ghcr.io/stargz-containers/ubuntu:20.04-org
> RUN echo hello
> EOF
> $ nerdctl build -t foo /tmp/ctx
> $ cat <<EOF > /tmp/ctx/Dockerfile
> FROM foo
> RUN echo bar
> EOF
> $ nerdctl build -t bar /tmp/ctx
> ```

That should be probably in the nerdctl repo

ktock:
> That should be probably in the nerdctl repo

:+1:

We already have it in nerdctl repo https://github.com/containerd/nerdctl/blob/10aae5ab75dc8a3156a7853e6cf5da624463bdb9/cmd/nerdctl/build_test.go#L51-L87

PR #757, opened 2022-03-25
Add factory reset cmd

Junnplus:

Add `limactl factory-reset` command for wiping the data of instance.

### DIFF
--- a/cmd/limactl/factory-reset.go
+++ b/cmd/limactl/factory-reset.go
@@ -6,8 +6,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	networks "github.com/lima-vm/lima/pkg/networks/reconcile"
-	"github.com/lima-vm/lima/pkg/start"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


AkihiroSuda (dismissed):
Thanks

jandubois:
Thanks!

I'm wondering if we should have a way to factory-reset an instance without getting prompted to restart it (like `--tty=false` for the `start` command).

This is a bit ugly:

```console
$ l factory-reset </dev/null
INFO[0000] The QEMU process seems already stopped
INFO[0000] Removing *.pid *.sock under "/Users/jan/.lima/default"
INFO[0000] Instance "default" has been factory reset
? Do you want to start the instance now?  (Y/n) FATA[0000] EOF
^[[60;156R^[[60;49R
```

Junnplus:
> I'm wondering if we should have a way to factory-reset an instance without getting prompted to restart it (like --tty=false for the start command).

Unlike start command, we cannot determine if the user wants to start the instance after a factory reset.
The simple solution is to remove the ask, use the `--start=true` flag.

cmd/limactl/factory-reset.go
```diff
@@ -0,0 +1,76 @@
...
+	for _, f := range fi {
+		path := filepath.Join(inst.Dir, f.Name())
+		if !strings.HasSuffix(path, ".yaml") && !strings.HasSuffix(path, ".yml") {
+			logrus.Infof("Removing %q", path)
+			if err := os.Remove(path); err != nil {
+				logrus.Error(err)
+			}
+		}
+	}
+	logrus.Infof("Instance %q has been factory reset", instName)
```
AkihiroSuda, line 57: We can just exit here. No need to start, as factory-reset would be very rarely used command.

Junnplus: SGTM.

AkihiroSuda (approved):

jandubois (approved):
Thanks, LGTM

AkihiroSuda:
Next time please squash commits 🙏

PR #759, opened 2022-03-26
Add option to add a data disk to an instance

### LINKED ISSUE #722: Support configuring multiple disks for a Lima VM
### Description

Use case:

I want to persist data within the guest without a network volume mount to the host on a separate disk from the Lima guest's root disk.

This would:
- Allow higher performance than network volumes while still retaining the ability to persist files even if you recreate your VM.
- Allow copying the backing disk's image on the host between different VMs, perhaps allowing one to pre-seed VMs with some content separate from the image.

afbjorklund:
The data disk will show up after the root disk

Typically as /dev/vdb, when running with virtio

Closes #722

User needs to partition, format and mount the disk...

----

```yaml
# Disk size
# 🟢 Builtin default: "100GiB"
disk: null

# Data size
# 🟢 Builtin default: "0"
data: "10GiB"
```

```
Disk /dev/vda: 100 GiB, 107374182400 bytes, 209715200 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes

Disk /dev/vdb: 10 GiB, 10737418240 bytes, 20971520 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
```

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -39,6 +39,10 @@ memory: null
 # 🟢 Builtin default: "100GiB"
 disk: null
 
+# Data size
+# 🟢 Builtin default: "0"
+data: null
+
 # Expose host directories to the guest, the mount point might be accessible from all UIDs in the guest
 # 🟢 Builtin default: null (Mount nothing)
 # 🔵 This file: Mount the home as read-only, /tmp/lima as writable
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -35,6 +35,7 @@ cloud-init:
 disk:
 - `basedisk`: the base image
 - `diffdisk`: the diff image (QCOW2)
+- `datadisk`: the data image (QCOW2), optional
 
 QEMU:
 - `qemu.pid`: QEMU PID
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -139,6 +139,16 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 		y.Disk = pointer.String("100GiB")
 	}
 
+	if y.Data == nil {
+		y.Data = d.Data
+	}
+	if o.Data != nil {
+		y.Data = o.Data
+	}
+	if y.Data == nil || *y.Data == "" {
+		y.Data = pointer.String("0")
+	}
+
 	if y.Video.Display == nil {
 		y.Video.Display = d.Video.Display
 	}
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -55,6 +55,7 @@ func TestFillDefault(t *testing.T) {
 		CPUs:   pointer.Int(4),
 		Memory: pointer.String("4GiB"),
 		Disk:   pointer.String("100GiB"),
+		Data:   pointer.String("0"),
 		Containerd: Containerd{
 			System:   pointer.Bool(false),
 			User:     pointer.Bool(true),
@@ -190,6 +191,7 @@ func TestFillDefault(t *testing.T) {
 		CPUs:   pointer.Int(7),
 		Memory: pointer.String("5GiB"),
 		Disk:   pointer.String("105GiB"),
+		Data:   pointer.String("10GiB"),
 		Containerd: Containerd{
 			System: pointer.Bool(true),
 			User:   pointer.Bool(false),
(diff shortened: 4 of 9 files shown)


jandubois:
I think we would want to exclude the "data disk" from the list of files deleted by the `factory-reset` command.

I have not tested this with Alpine yet; I want to make sure that the "implict" data volume continues to be the "diff disk" and not the additional "data disk".

examples/default.yaml
```diff
@@ -39,6 +39,10 @@ memory: null
 # 🟢 Builtin default: "100GiB"
 disk: null
 
+# Data size
```
jandubois, line 42: I think this needs additional documentation to explain that this is a supplemental data volume, and the parameter is the size, and can use size-related units like "GiB". Maybe the builtin default should change to "0GiB" to make this obvious?

jandubois:
I'm not sure if there would ever be a reason to create more than one additional data volume? In which case `data` should be a list and not a scalar.

I can't think of anything right now, but don't want us to paint ourselves into a corner that will require incompatible schema changes in the future.

examples/default.yaml
jandubois: The `datadisk` also needs to be mentioned in `docs/internal.md`.

afbjorklund:
> I have not tested this with Alpine yet; I want to make sure that the "implict" data volume continues to be the "diff disk" and not the additional "data disk".

The persistent disk for distributions that boot from a LiveCD was not considered as a "data" disk, it was more regarded as being a part of the root (or "OS") disk - using the cloud parlance*. Even though it mounts under `/mnt/data` and is called data-volume...

```console
lima-alpine:/home/anders/lima$ sudo fdisk -l
Disk /dev/vda: 100 GB, 107374182400 bytes, 209715200 sectors
366634 cylinders, 13 heads, 44 sectors/track
Units: sectors of 1 * 512 = 512 bytes

Device  Boot StartCHS    EndCHS        StartLBA     EndLBA    Sectors  Size Id Type
/dev/vda1    2,0,33      178,12,44         2048  209715199  209713152 99.9G 83 Linux
Disk /dev/vdb: 10 GB, 10737418240 bytes, 20971520 sectors
20805 cylinders, 16 heads, 63 sectors/track
Units: sectors of 1 * 512 = 512 bytes

Disk /dev/vdb doesn't contain a valid partition table
```

But you are right, the naming could be confusing. When compare with `05-persistent-data-volume.sh` and friends. It was normally just called the "persistence partition" and mounted under `/dev/sda1` or somesuch back when I was using [machine](https://github.com/docker/machine).

And it could well be more than one.

----

\* https://docs.microsoft.com/en-us/azure/virtual-machines/managed-disks-overview#disk-roles
* OS disk
* Temporary disk
* **Data disk(s)**

AkihiroSuda:
> Closes https://github.com/lima-vm/lima/issues/722

722 probably requires supporting custom disk image path

examples/default.yaml
```diff
@@ -39,6 +39,10 @@ memory: null
...
+# 🟢 Builtin default: "0"
+data: null
```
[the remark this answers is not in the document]
afbjorklund: Well, "disk" is also kind of vague. And "data" was ambiguous.

chancez:
Here are my thoughts:
- I agree `data` is a bit vague. I think `disks` is better, or `devices` and using an object you could define the type of device, and other options (maybe you expose all the -drive options directly). Eg: `devices: [{type: virtio, path: "/path/to/somewhere"}]`.
- It would be good to support more than 1 disk. So instead of `disk:` being a single value, perhaps `disks` could become a list of objects. Eg: `data: [ {size: 8GiB }]`
- In the disk's configuration you could configure if it should be wiped on a factory reset. Eg: `{size: 8GiB, deletionPolicy: "delete|keep"}`

jandubois:
> * In the disk's configuration you could configure if it should be wiped on a factory reset. Eg: `{size: 8GiB, deletionPolicy: "delete|keep"}`

As I wrote in https://github.com/lima-vm/lima/issues/722#issuecomment-1079760724, I think it will be better if we manage data volumes independent of instances. Because I may want to keep the data even when I delete the instance, and not just factory reset it. And I may want to read the data from a different instance than the one that created it.

So I think `limectl volume create|ls|destroy` make more sense, and then you can attach the disk to your instance via `lima.yaml`.  And we can eventually extend the volume functionality by supporting things like resizing and cloning volumes.

afbjorklund:
Okay, I guess we can consider the proof-of-concept concluded then (can do an array of disks, but seems it wants more)

PR #766, opened 2022-03-29
[wip] change limactl start timeout as command args

Abirdcfly:
The source code has a hard-coded 10-minute timeout, which interferes with situations such as when the initialization command is time-consuming, or when the network conditions are simply not good

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -1,6 +1,7 @@
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


AkihiroSuda (changes requested):
The timeout has to be set after completion of downloading the image and the containerd archive.

Otherwise downloading will easily fail.

PR #772, opened 2022-04-03
login with default shell

faceair:
image

I am used to using the fish shell and I would like to be able to use the current user's default shell in lima vm as well.

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -108,7 +108,7 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	logrus.Debugf("changeDirCmd=%q", changeDirCmd)
 
-	script := fmt.Sprintf("%s ; exec bash --login", changeDirCmd)
+	script := fmt.Sprintf("%s ; exec $SHELL --login", changeDirCmd)
 	if len(args) > 1 {
 		script += fmt.Sprintf(
 			" -c %s",


jandubois (approved):
Thanks you, LGTM

jandubois:
> I would like to be able to use the current user's default shell in lima vm as well.

At first I thought you wanted to automatically configure the shell the user is running on the host, which would have been problematic because e.g. `fish` is not installed in the images by default. So you still require the user to install and change the shell themselves. That seems fine, as any issues would be the fault of the user changing the configuration. 😄

AkihiroSuda:
> Signed-off-by: faceair

Please use real name

cmd/limactl/shell.go
```diff
@@ -108,7 +108,7 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	logrus.Debugf("changeDirCmd=%q", changeDirCmd)
 
-	script := fmt.Sprintf("%s ; exec bash --login", changeDirCmd)
+	script := fmt.Sprintf("%s ; exec $SHELL --login", changeDirCmd)
```
AkihiroSuda, line 111: Does this work with zsh?

jandubois: > Does this work with zsh?

Looks like it:

```console
$ lima
jan@lima-default:/Users/jan$ exec /usr/bin/zsh
lima-default% echo $0
/usr/bin/zsh
```

faceair:
> Please use real name
This is my English ID, which I use on various sites including github. I have been using this id for many years and I do not use my real name on the internet.

cmd/limactl/shell.go
AkihiroSuda: But not for tcsh? (Anybody still uses?)
```console
$ tcsh --login
Unknown option: `--login'
Usage: tcsh [ -bcdefilmnqstvVxX ] [ argument ... ].
```

AkihiroSuda:
Sorry, we are not accepting anonymous contributions

faceair:
OK. If there are no problems with the code, you can submit this change yourselves.

cmd/limactl/shell.go
jandubois: > But not with tcsh? (Anybody still uses?)

The only way to run a command via a `tcsh` login shell seems to be:

```console
echo $CMD | tcsh -l
```

Users still have to install and configure the shell inside the VM (and then reboot) before they can use it:

```console
lima sudo apt install tcsh
lima sudo chsh -s /usr/bin/tcsh $USER
limactl stop
limactl start
```

I guess the problem is that they are now trapped because they can't undo this change anymore. I wonder if we should provide a `limactl shell --bash` option to force the shell to use `bash` instead of the configured default shell?

faceair: I have read the documentation for [fish](https://fishshell.com/docs/current/cmds/fish.html?highlight=login), [zsh](https://zsh.sourceforge.io/Guide/zshguide02.html) and [bash](https://www.gnu.org/software/bash/manual/html_node/Invoking-Bash.html), all of which explicitly support the '--login' parameter and as an [login shell](https://unix.stackexchange.com/questions/38175/difference-between-login-shell-and-non-login-shell).
It seems that tcsh does not support use as a login shell, so it may not be appropriate to use tcsh as the default shell in this case. But I think you could submit an issue to this project for discussion.

jandubois: > But I think you could submit an issue to this project for discussion.

I don't really care about supporting `tcsh` as a login shell for lima; but I'm concerned that somebody would try it, and gets locked out of their VM. That's why I was suggesting the `--bash` option to be able to still get a shell to fix the issue.

faceair: But we're just opening a new interactive shell after each ssh session, which shouldn't be a login shell, right?

jandubois: > which shouldn't be a login shell

I'm not sure what you are saying here. It won't read `~/.profile` (or equivalent) unless it is a login shell.

faceair: image
It seems that '-l' is a more generic argument than '--login'.

faceair: image

Alternatively, we can consider using the login command to open a new session, which will be opened using the default shell.

PR #775, opened 2022-04-04
regard linux default shell

xfoxfu:
This PR let lima uses the default shell set in Linux.

By using

```bash
chsh -s $(which zsh) $USER
```

The user can set default shell to zsh, and with this PR, lima will regard this setting and use the default shell by reading `/etc/passwd`.

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -108,7 +108,7 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	logrus.Debugf("changeDirCmd=%q", changeDirCmd)
 
-	script := fmt.Sprintf("%s ; exec bash --login", changeDirCmd)
+	script := fmt.Sprintf("%s ; exec $SHELL --login", changeDirCmd)
 	if len(args) > 1 {
 		script += fmt.Sprintf(
 			" -c %s",


jandubois:
I don't think we want this. There is no way to guarantee that the host shell is available within the guest; you will also need to install it, either manually, or via a provisioning script.

I think #772 is a safer way to support this.

xfoxfu:
I've tried that approach, but I found on ArchLinux and zsh, it does not work.

xfoxfu:
I've updated my commit to align with the approach in #772. However, I've found this approach seems only to work after a guest restart or sshd reloading.

jandubois:
> I've found this approach seems only to work after a guest restart or sshd reloading.

Ah yes, because we run `loginctl enable-linger $USER`.

You could solve this by rebooting in your provisioning script. Something like this (untested, so probably not going to work as-is):

```
provision:
  mode: system
  script: |
    #!/bin/sh
    SHELL=$(getent passwd ${LIMA_CIDATA_USER} | awk -F: '{print $NF}')
    if [ "$SHELL" -ne /usr/bin/fish ]; then
      apt install -y fish
      chsh -s /usr/bin/fish ${LIMA_CIDATA_USER}
      reboot
    fi
```

If you put this into `~/.lima/_config/default.yaml` (after testing that it works with all the templates you are using), then it should automatically apply to all your instances as soon as you restart them.

Actually, put it into `override.yaml` instead, so it is executed first, so you reboot as early as possible and not waste time doing other provisioning scripts twice.

jandubois (approved):
LGTM, thanks!

I still think I would like to see a `limactl shell --shell /bin/sh` option (both for predictability in scripting, and to rescue an instance after a botched `chsh` command), but that can be a separate PR.

PR #779, opened 2022-04-06
Add --shell option to limactl shell

jandubois:
This allows to use a predictable shell for scripting, and also provides a way to recover an instance when the user shell has
been misconfigured:

```console
$ limactl shell --shell /bin/ash alpine uname -a
Linux lima-alpine 5.10.93-0-virt #1-Alpine SMP Thu, 27 Jan 2022 09:34:38 +0000 x86_64 Linux
```

This is a followup to https://github.com/lima-vm/lima/pull/775#pullrequestreview-932678689

I'm not happy about the repetition of the word "shell" in `limactl shell --shell /bin/sh ...`, but it still seems like the most appropriate name for the option.

<hr>

This part has been split off into #780

Also adds `LIMA_SHELL` support to the `lima` command:

```console
$ export LIMA_INSTANCE=alpine
$ lima sudo apk add fish
[...]
$ export LIMA_SHELL=/usr/bin/fish
$ lima
jan@lima-alpine /U/jan>
```

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -36,6 +36,7 @@ func newShellCommand() *cobra.Command {
 
 	shellCmd.Flags().SetInterspersed(false)
 
+	shellCmd.Flags().String("shell", "", "shell interpreter, e.g. /bin/bash")
 	shellCmd.Flags().String("workdir", "", "working directory")
 	return shellCmd
 }
@@ -108,7 +109,16 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	logrus.Debugf("changeDirCmd=%q", changeDirCmd)
 
-	script := fmt.Sprintf("%s ; exec $SHELL --login", changeDirCmd)
+	shell, err := cmd.Flags().GetString("shell")
+	if err != nil {
+		return err
+	}
+	if shell == "" {
+		shell = `"$SHELL"`
+	} else {
+		shell = shellescape.Quote(shell)
+	}
+	script := fmt.Sprintf("%s ; exec %s --login", changeDirCmd, shell)
 	if len(args) > 1 {
 		script += fmt.Sprintf(
 			" -c %s",
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -84,19 +84,19 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	if workDir != "" {
-		changeDirCmd = fmt.Sprintf("cd %q || exit 1", workDir)
+		changeDirCmd = fmt.Sprintf("cd %s || exit 1", shellescape.Quote(workDir))
 		// FIXME: check whether y.Mounts contains the home, not just len > 0
 	} else if len(y.Mounts) > 0 {
 		hostCurrentDir, err := os.Getwd()
 		if err == nil {
-			changeDirCmd = fmt.Sprintf("cd %q", hostCurrentDir)
+			changeDirCmd = fmt.Sprintf("cd %s", shellescape.Quote(hostCurrentDir))
 		} else {
 			changeDirCmd = "false"
 			logrus.WithError(err).Warn("failed to get the current directory")
 		}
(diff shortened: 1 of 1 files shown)


jandubois:
FYI @faceair @xfoxfu

faceair:
Based on [my previous comment](https://github.com/lima-vm/lima/pull/772#discussion_r841463202) is that the `-l` option seems to be more generic than `--login`, so is it possible to use `-l` instead of the `--login` option.

cmd/lima
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 set -eu
 : "${LIMA_INSTANCE:=default}"
+: "${LIMA_SHELL:=}"
```
AkihiroSuda, line 4: I'm not sure we want to have this env var.
A user should just chsh.

jandubois: > A user should just chsh.

Sure, but you have to reboot before it takes effect (because we enable "linger"), and it is not obvious to the user, so it looks like it doesn't work at first. Using the env variable makes it easy to switch between shells.

AkihiroSuda: Wondering rather we should have a generic way to support `limactl shell` options in the `lima` script. Perhaps like `LIMA_SHELL_ARGS=--workdir=/ --shell=/bin/zsh`. But on the second thought maybe users should just set `alias lima=limactl shell ....`.


Could you split the `lima` script into another PR for further discussion?

jandubois:
> Based on [my previous comment](https://github.com/lima-vm/lima/pull/772#discussion_r841463202) is that the `-l` option seems to be more generic than `--login`, so is it possible to use `-l` instead of the `--login` option.

I don't know, the only shell where it makes a difference (that I'm aware off) is `tcsh`, and there is has to be the only argument, so it only works for interactive mode:

```console
$ tcsh -c "echo foo"
foo
$ tcsh -l -c "echo foo"
Unknown option: `-l'
Usage: tcsh [ -bcdefilmnqstvVxX ] [ argument ... ].
$ echo "echo foo" | tcsh -l
foo
```

So I'm not sure if we should bother; any modern shell should support `--login`.

cmd/lima
jandubois: > Could you split the lima script part into another PR for further discussion?

Moved to #780

AkihiroSuda (dismissed):
Thanks

cmd/limactl/shell.go
```diff
@@ -108,7 +109,14 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
 
-	script := fmt.Sprintf("%s ; exec $SHELL --login", changeDirCmd)
+	shell, err := cmd.Flags().GetString("shell")
+	if err != nil {
+		return err
+	}
+	if shell == "" {
+		shell = "$SHELL"
+	}
+	script := fmt.Sprintf("%s ; exec %s --login", changeDirCmd, shell)
```
AkihiroSuda, line 119: Better to quote?

jandubois: I don't think this should ever be necessary, but why not...

I've also squeezed in unrelated calls to `shellescape.Quote` directory names. We used to use `%q` for them, but that is wrong if the name contains a `$` sign: `"foo$bar"` needs to be `'foo$bar'`.

cmd/limactl/shell.go
```diff
@@ -108,7 +109,16 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	logrus.Debugf("changeDirCmd=%q", changeDirCmd)
 
-	script := fmt.Sprintf("%s ; exec $SHELL --login", changeDirCmd)
+	shell, err := cmd.Flags().GetString("shell")
+	if err != nil {
+		return err
+	}
+	if shell == "" {
+		shell = "$SHELL"
```
AkihiroSuda, line 117: Can we quote this too?

jandubois: > Can we quote this too?

Sure, but here we need `"$SHELL"` quotes and not `shellescape.Quote` because in this case we do want variable interpolation.

AkihiroSuda (approved):

PR #780, opened 2022-04-06
Add LIMA_SHELL env var to lima command

jandubois:
```console
$ export LIMA_INSTANCE=alpine
$ lima sudo apk add fish
[...]
$ export LIMA_SHELL=/usr/bin/fish
$ lima
jan@lima-alpine /U/jan>
```

Split off from #779.

### DIFF
--- a/cmd/lima
+++ b/cmd/lima
@@ -1,6 +1,8 @@
 #!/bin/sh
 set -eu
 : "${LIMA_INSTANCE:=default}"
+: "${LIMA_SHELL:=}"
+: "${LIMA_WORKDIR:=}"
 : "${LIMACTL:=limactl}"
 
 if [ "$#" -eq 1 ]; then
@@ -11,6 +13,9 @@ if [ "$#" -eq 1 ]; then
     echo "${base} is an alias for \"${LIMACTL} shell ${LIMA_INSTANCE}\"."
     echo "The instance name (\"${LIMA_INSTANCE}\") can be changed by specifying \$LIMA_INSTANCE."
     echo
+    echo "The shell and initial workdir inside the instance can be specified via \$LIMA_SHELL"
+    echo "and \$LIMA_WORKDIR."
+    echo
     echo "See \`${LIMACTL} shell --help\` for further information."
     exit 0
   elif [ "$1" = "-v" ] || [ "$1" = "--version" ]; then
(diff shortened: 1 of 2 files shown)


cmd/lima
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
 set -eu
 : "${LIMA_INSTANCE:=default}"
+: "${LIMA_SHELL:=}"
```
AkihiroSuda, line 4: Wondering rather we should have a generic way to support `limactl shell` options in the `lima` script. Perhaps like `LIMA_SHELL_ARGS=--workdir=/ --shell=/bin/zsh`. But on the second thought maybe users should just set `alias lima=limactl shell ....`.


_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/779#discussion_r843467813_

jandubois: > should have a generic way to support `limactl shell` options in the `lima` script

Practically speaking, we only have 2 options: `--shell` and `--workdir`, and I don't really expect somebody to hard-code the workdir for each `lima` call. But even then, I would rather add another `LIMA_WORKDIR` variable instead, so the options can be manipulated independent of each other.

AkihiroSuda: > we only have 2 options

Eventually we will have more

jandubois: Yeah, but I still like having individual variables. This is how I would do it:

```shell
set - "$LIMA_INSTANCE" "$@"
if [ -n "${LIMA_SHELL}" ]; then
  set - --shell "$LIMA_SHELL" "$@"
fi
if [ -n "${LIMA_WORKDIR}" ]; then
  set - --workdir "$LIMA_WORKDIR" "$@"
fi
exec "$LIMACTL" shell "$@"
```

This is trivially extensible to additional settings.

jandubois: I've updated the PR to include support `LIMA_WORKDIR` as well (and rebased on top of `master`, so the `--shell` option is actually functional.

AkihiroSuda (approved):

AkihiroSuda:
CI failing
```
Warning: Attempt 1 failed. Reason: Child_process exited with error code 1

bash: line 1: ./hack/test-upgrade.sh: No such file or directory
Warning: Attempt 2 failed. Reason: Child_process exited with error code 127

bash: line 1: ./hack/test-upgrade.sh: No such file or directory
Error: Final attempt failed. Child_process exited with error code 127
```
https://github.com/lima-vm/lima/runs/5864063545?check_suite_focus=true

jandubois:
> CI failing

Seems like CI flakiness; it passed on re-run.

PR #783, opened 2022-04-08
Add trusted CA certificate support

### LINKED ISSUE #594: Provisioning should allow files and directories to be copied
### Description

I ran into a slight problem trying to provision a VM with k3s.yaml from the examples.    The provision script executes 'curl', (and subsequently the get.k3s.io script does too), but fails in my environment due to http proxy certificates not installed.    I was able to make this work by modifying the provision script to first create a /root/.curlrc file containing "--insecure --location-trusted".   This could be avoided if I could install a .curlrc or better still, TLS certificates, during provisioning.  Unfortunately, different Linux distributions install certificates in different locations, and use different scripts to update the master certs file.

I'm thinking of a more generalized method for this: - allow files and directories to be copied into the VM before the provisioning scripts run.  If a source specifies a directory, it should be copied recursively.  Source files/directories may be relative or absolute paths.  Target directories in the VM should be absolute paths.  

This would allow things like a .curlrc file, and directory containing certificates, whatever to be available to the provisioning script.

Proposed yaml syntax:
```
provision:
- copy: 
  - required: true
    source: ~/.lima/certs/
    destination: /path/in/vm
  - required: false
    source: vm-curl.rc
    destination: /root/.curlrc
- mode: system
  script: |
    #!/bin/sh
    curl -sfL https://get.k3s.io | sh -
```

Something like this could be included in the override.yaml to perform this for all provisioned VMs.

nickpetrovic:
resolves #594

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -103,6 +103,27 @@ ssh:
 # ADVANCED CONFIGURATION
 # ===================================================================== #
 
+caCerts:
+  # If set to `true`, this will remove all the default trusted CA certificates that
+  # are normally shipped with the OS.
+  # 🟢 Builtin default: false
+  removeDefaults: null
+
+  # A list of trusted CA certificate files. The files will be read and passed to cloud-init.
+  files:
+  # - examples/hello.crt
+
+  # A list of trusted CA certificates. These are directly passed to cloud-init.
+  certs:
+  # - |
+  #   -----BEGIN CERTIFICATE-----
+  #   YOUR-ORGS-TRUSTED-CA-CERT-HERE
+  #   -----END CERTIFICATE-----
+  # - |
+  #   -----BEGIN CERTIFICATE-----
+  #   YOUR-ORGS-TRUSTED-CA-CERT-HERE
+  #   -----END CERTIFICATE-----
+
 containerd:
   # Enable system-wide (aka rootful)  containerd and its dependencies (BuildKit, Stargz Snapshotter)
   # 🟢 Builtin default: false
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -186,6 +186,28 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 		}
 	}
 
+	args.CACerts.RemoveDefaults = y.CACertificates.RemoveDefaults
+
+	for _, path := range y.CACertificates.Files {
+		expanded, err := localpathutil.Expand(path)
+		if err != nil {
+			return err
+		}
+
+		content, err := os.ReadFile(expanded)
+		if err != nil {
+			return err
+		}
+
+		cert := getCert(string(content))
+		args.CACerts.Trusted = append(args.CACerts.Trusted, cert)
+	}
+
+	for _, content := range y.CACertificates.Certs {
+		cert := getCert(content)
+		args.CACerts.Trusted = append(args.CACerts.Trusted, cert)
+	}
+
 	if err := ValidateTemplateArgs(args); err != nil {
 		return err
 	}
@@ -243,4 +265,16 @@ func GuestAgentBinary(arch string) (io.ReadCloser, error) {
 	}
 	gaPath := filepath.Join(dir, "lima-guestagent.Linux-"+arch)
 	return os.Open(gaPath)
+}
+
+func getCert(content string) Cert {
+	lines := []string{}
+	for _, line := range strings.Split(content, "\n") {
+		if line == "" {
+			continue
+		}
+		lines = append(lines, strings.TrimSpace(line))
+	}
+	// return lines
+	return Cert{Lines: lines}
 }
--- a/pkg/cidata/template.go
+++ b/pkg/cidata/template.go
@@ -19,6 +19,15 @@ var templateFS embed.FS
 
 const templateFSRoot = "cidata.TEMPLATE.d"
 
+type CACerts struct {
+	RemoveDefaults *bool
+	Trusted        []Cert
+}
+
+type Cert struct {
+	Lines []string
+}
+
 type Containerd struct {
 	System bool
 	User   bool
@@ -51,6 +60,7 @@ type TemplateArgs struct {
 	TCPDNSLocalPort int
 	Env             map[string]string
 	DNSAddresses    []string
+	CACerts         CACerts
 }
 
 func ValidateTemplateArgs(args TemplateArgs) error {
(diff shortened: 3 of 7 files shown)


examples/default.yaml
```diff
@@ -82,6 +82,27 @@ ssh:
   # 🟢 Builtin default: false
   forwardAgent: null
 
+ca_certs:
```
jandubois, line 85: The naming convention would be `caCerts`.

We only have a single exception for `cpuType.x86_64` because that is the string used as the `arch` value in other places.

examples/default.yaml
```diff
@@ -82,6 +82,27 @@ ssh:
...
+  # If set to `true`, this will remove all the default trusted CA certificates that
+  # are normally shipped with the OS.
+  # 🟢 Builtin default: true
+  remove_defaults: null
```
jandubois, line 89: Should be `removeDefaults`.

pkg/cidata/template.go
```diff
@@ -19,6 +19,15 @@ var templateFS embed.FS
...
 const templateFSRoot = "cidata.TEMPLATE.d"
 
+type CACerts struct {
+	RemoveDefaults bool
+	Trusted        []Cert
+}
+
+type Cert struct {
+	Lines []string
+}
```
jandubois, line 29: Having a slice of lines to represent a cert seems slightly weird to me; that is a representation detail. I do see that it is convenient for the template processing though, so maybe it is fine.

jandubois (changes requested):
This PR is missing the `default.yaml` and `override.yaml` handling in `limayaml.FillDefault()`, and then should get corresponding test data filled in for `default_test.go` as well.

jandubois:
Also please add DCO signature to your commit: https://github.com/lima-vm/lima/pull/783/checks?check_run_id=5879125698

examples/default.yaml
```diff
@@ -82,6 +82,27 @@ ssh:
   # 🟢 Builtin default: false
   forwardAgent: null
 
+ca_certs:
+  # If set to `true`, this will remove all the default trusted CA certificates that
+  # are normally shipped with the OS.
+  # 🟢 Builtin default: true
```
AkihiroSuda, line 88: false

examples/default.yaml
AkihiroSuda: And should be moved to the `FURTHER ADVANCED CONFIGURATION` section

pkg/limayaml/limayaml.go
```diff
@@ -142,6 +143,12 @@ type HostResolver struct {
 	Hosts   map[string]string `yaml:"hosts,omitempty" json:"hosts,omitempty"`
 }
 
+type CACertificates struct {
+	RemoveDefaults bool     `yaml:"remove_defaults,omitempty" json:"remove_defaults,omitempty"`
+	Files          []string `yaml:"files,omitempty" json:"files,omitempty"`
+	Certs          []string `yaml:"certs,omitempty" json:"certs,omitempty"`
+}
+
```
AkihiroSuda, line 151: https://github.com/lima-vm/lima/blob/939f79b4bd7e1e391e416bf96b4e325fddbbdc98/pkg/limayaml/defaults.go#L75 needs to be updated too?

jandubois: Yes, I mentioned that in my review at https://github.com/lima-vm/lima/pull/783#pullrequestreview-935895549

pkg/cidata/template.go
```diff
@@ -19,6 +19,15 @@ var templateFS embed.FS
 
 const templateFSRoot = "cidata.TEMPLATE.d"
 
+type CACerts struct {
+	RemoveDefaults bool
```
jandubois, line 23: This needs to be a pointer for the default processing to work properly:

```suggestion
	RemoveDefaults *bool
```

examples/default.yaml
nickpetrovic: Fixed.

examples/default.yaml
nickpetrovic: Fixed.

pkg/limayaml/limayaml.go
nickpetrovic: Should be fixed. For the CA files and certs, I'm appending and de-duping them.

pkg/cidata/template.go
nickpetrovic: Fixed.

examples/default.yaml
nickpetrovic: Fixed.

nickpetrovic:
> Also please add DCO signature to your commit: [#783 (checks)](https://github.com/lima-vm/lima/pull/783/checks?check_run_id=5879125698)

Added.

nickpetrovic:
> This PR is missing the `default.yaml` and `override.yaml` handling in `limayaml.FillDefault()`, and then should get corresponding test data filled in for `default_test.go` as well.

@jandubois I think I did this correctly. The tests were a little confusing to me. Let me know if its sufficient.

pkg/cidata/cidata.go
```diff
@@ -186,6 +186,28 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 		}
 	}
 
+	args.CACerts.RemoveDefaults = y.CACertificates.RemoveDefaults
```
nickpetrovic, line 189: Missed setting this earlier.

pkg/limayaml/defaults.go
```diff
@@ -450,6 +451,22 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
+	}
+	if y.CACertificates.RemoveDefaults == nil {
+		y.CACertificates.RemoveDefaults = pointer.Bool(false)
+	}
+
+	caFiles := unique(append(append(d.CACertificates.Files, y.CACertificates.Files...), o.CACertificates.Files...))
+	y.CACertificates.Files = caFiles
+
+	caCerts := unique(append(append(d.CACertificates.Certs, y.CACertificates.Certs...), o.CACertificates.Certs...))
+	y.CACertificates.Certs = caCerts
```
jandubois, line 469: It would probably be better to check for duplicates ones all the files have been read and combined with the literal certs. You might have multiple files containing the same cert, or having a file and a literal entry be the same. Or you might have `~/ca.pem` and `/Users/jan/ca.pem`, which would be the same file, but not recognized as duplicates.

On the other hand I don't think having duplicate certs is a problem, so not marking this as a blocker.

jandubois (approved):
LGTM, thanks!

jandubois:
Note that this feature will not work with alpine-lima until https://github.com/lima-vm/alpine-lima/issues/54 is implemented!

PR #791, opened 2022-04-08
limactl: build in experimental/deprecated templates

### LINKED ISSUE #785: limactl start templates don't support subdirectories
### Description

We now have `examples/deprecated/centos-7.yaml` and `examples/experimental/9p.yaml`.

Neither of them are listed by `limactl start --list-templates` and they cannot be started with the `template://` syntax either:

```console
$ limactl start --name centos7 template://deprecated/centos-7
FATA[0000] open /usr/local/share/lima/examples/deprecated.yaml: no such file or directory
```

AkihiroSuda:
The template name does not contain '/'.
e.g., `limactl start template://9p`, not `limactl start template://experimental/9p`.

Fix #785

### DIFF
--- a/cmd/limactl/info.go
+++ b/cmd/limactl/info.go
@@ -21,8 +21,10 @@ func newInfoCommand() *cobra.Command {
 }
 
 type TemplateYAML struct {
-	Name     string `json:"name"`
-	Location string `json:"location"`
+	Name         string `json:"name"`
+	Location     string `json:"location"`
+	Experimental bool   `json:"experimental,omitempty"`
+	Deprecated   bool   `json:"deprecated,omitempty"`
 }
 
 type Info struct {
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"net/http"
 	"net/url"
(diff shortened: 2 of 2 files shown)


jandubois:
This doesn't feel right to me. Why put the templates into subdirectories, and then hide the subdirectories from the user.

* `limactl start --list-templates` now sorts them to the end, but it isn't clear to the user what that means.
* How do we differentiate between `experimental/9p.yaml` and `deprecated/9p.yaml`.
* Why can't the user create additional directories, like `my/default.yaml`?

So we essentially still have a flat namespace for template names, and use hard-coded subdirectories to "tag" some templates, but without really exposing it to the user.

This seems to have limited use, but may make it hard in the future to extend (e.g. I think we need a user template directory in the future where people can store their custom templates, that will take precedence over the bundled ones).

If we want to stick with this mechanism, then at least `--list-templates` need to show the additional tags on each template.

<hr>

I would prefer that we store tag-like metadata like `experimental` and `deprecated` inside the YAML file itself. Then we can eventually extend this other tags, like `ubuntu`, or `k8s`, or `docker`, and the user could filter templates based on all of those. Of course this requires actually reading all the YAML files, but that is so fast to be unnoticeable.

And then I don't think we need to have subdirectories in the templates at all, given that we don't treat them as independent namespaces anyways.

PR #40, opened 2021-06-08
Update logic for shell workdir

jandubois:
Changes to the previous implementation are:

* If there are no volume mounts then there is no fallback to the host `$HOME` directory (because we know it isn't mounted, so no need to display an additional error).

* If the `--workdir` does not exist in the guest, we fallback to the host `$PWD` first before trying the host `$HOME` (because it is less surprising to stay in the current directory than to switch to one that is different from the requested one).

### Testing with no mounts

No error message when `--workdir` is not specified, and only a single error message when the requested workdir does not exist:

```console
$ limactl shell k3s bash
jan@lima-default:~$ pwd
/home/jan.linux
jan@lima-k3s:~$ exit
$ limactl shell --workdir /does-not-exist k3s bash
bash: line 1: cd: /does-not-exist: No such file or directory
jan@lima-default:~$ pwd
/home/jan.linux
```

### Testing with a `/tmp` mount, but no home directory mount

Falls back to the host `$PWD` and `$HOME`. Uses guest `$HOME` when none of them exist:

```console
$ limactl shell --workdir /does-not-exist default bash
bash: line 1: cd: /does-not-exist: No such file or directory
bash: line 1: cd: /Users/jan/suse/lima: No such file or directory
bash: line 1: cd: /Users/jan: No such file or directory
jan@lima-default:~$ pwd
/home/jan.linux
```

PR #797, opened 2022-04-14
CI: fix 'fatal: unsafe repository ('/Users/runner/work/lima/lima/vde_vmnet' is owned by someone else)'

AkihiroSuda:
CI was failing due to the recent breaking change of git: https://github.com/git/git/commit/8959555cee7ec045958f9b6dd62e541affb7e7d9

> Cloning into 'vde_vmnet'...
> HEAD is now at a4b489e Merge pull request #40 from AkihiroSuda/dev
> git submodule update --init
> fatal: unsafe repository ('/Users/runner/work/lima/lima/vde_vmnet' is owned by someone else)
> To add an exception for this directory, call:
>
> 	git config --global --add safe.directory /Users/runner/work/lima/lima/vde_vmnet
> make: *** [install.vde-2] Error 128

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -88,9 +88,11 @@ jobs:
       run: |
         (
           brew install autoconf automake
+          cd ~
           git clone https://github.com/lima-vm/vde_vmnet
           cd vde_vmnet
           git checkout $VDE_VMNET_VERSION
+          sudo git config --global --add safe.directory /Users/runner/vde_vmnet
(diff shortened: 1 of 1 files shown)


jandubois:
Just out of curiosity: **Why** are the owners different? Is the checkout action not cloning the repo using the current user?

AkihiroSuda:
> Just out of curiosity: **Why** are the owners different? Is the checkout action not cloning the repo using the current user?

Because we clone the repo as the `runner` user but run `make install` as the `root`

PR #805, opened 2022-04-19
limactl start: support subdirectories

### LINKED ISSUE #785: limactl start templates don't support subdirectories

AkihiroSuda:
e.g., `limactl start template://deprecated/centos-7` creates an instance named `centos-7`

Fix #785

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"net/http"
 	"net/url"
 	"os"
@@ -14,6 +15,7 @@ import (
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/containerd/containerd/identifiers"
+	securejoin "github.com/cyphar/filepath-securejoin"
 	"github.com/lima-vm/lima/pkg/limayaml"
 	networks "github.com/lima-vm/lima/pkg/networks/reconcile"
 	"github.com/lima-vm/lima/pkg/osutil"
@@ -63,10 +65,10 @@ func readTemplate(name string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	if strings.Contains(name, string(os.PathSeparator)) {
-		return nil, fmt.Errorf("invalid template name %q", name)
+	defaultYAMLPath, err := securejoin.SecureJoin(filepath.Join(dir, "examples"), name+".yaml")
+	if err != nil {
+		return nil, err
 	}
-	defaultYAMLPath := filepath.Join(dir, "examples", name+".yaml")
 	return os.ReadFile(defaultYAMLPath)
 }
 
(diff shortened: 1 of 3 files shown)


cmd/limactl/start.go
```diff
@@ -92,9 +94,16 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string) (*store.Instance, e
 
 	if ok, u := argSeemsTemplateURL(arg); ok {
 		templateName := u.Host
+		if u.Path != "" {
+			templateName, err = securejoin.SecureJoin(templateName, u.Path)
+			if err != nil {
+				return nil, err
+			}
+		}
```
jandubois, line 102: It doesn't make sense to use `SecureJoin` here because it is still a relative path. And `filepath.Join` will just ignore empty elements, so no need for the conditional either:

```suggestion
		templateName = filepath.Join(u.Host, u.Path)
```

The real `SecureJoin` check happens later in `readTemplate` when this relative name is joined with the full path to the examples directory.

jandubois, line 102: It doesn't make sense to use `SecureJoin` here because it is still a relative path. And `filepath.Join` will just ignore empty elements, so no need for the conditional either:

```suggestion
		templateName = filepath.Join(u.Host, u.Path)
```

The real `SecureJoin` check happens later in `readTemplate` when this relative name is joined with the full path to the examples directory.

cmd/limactl/start.go
```diff
@@ -340,21 +349,26 @@ func listTemplateYAMLs() ([]TemplateYAML, error) {
...
-	for _, f := range globbed {
-		base := filepath.Base(f)
-		if strings.HasPrefix(base, ".") {
-			continue
+	walkDirFn := func(p string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		base := filepath.Base(p)
+		if strings.HasPrefix(base, ".") || !strings.HasSuffix(base, ".yaml") {
```
jandubois, line 359: I wonder why the `strings.HasPrefix(base, ".")` check is needed.

Do you expect to have `.hidden.yaml` templates that don't show up in `limactl start --list-templates` but could still be used as `template://.hidden`?

jandubois, line 359: I wonder why the `strings.HasPrefix(base, ".")` check is needed.

Do you expect to have `.hidden.yaml` templates that don't show up in `limactl start --list-templates` but could still be used as `template://.hidden`?

jandubois (approved):
LGTM, but I think the simplification from my first comment should be applied.

cmd/limactl/start.go
AkihiroSuda: IMHO it still makes sense to use `SecureJoin` 
https://go.dev/play/p/DYA2yxk4iBr


> u=url.Parse("template://foo/../bar"), u.Host="foo", u.Path="/../bar", filepath.Join(u.Host, u.Path)="bar", securejoin.SecureJoin(u.Host, u.Path="foo/bar")

cmd/limactl/start.go
AkihiroSuda: yes

cmd/limactl/start.go
jandubois: `SecureJoin` is kind of a light-weight `chroot` jail. It only makes sense when the left side of the "join" is internal to the app. When both parts are generated from user input, then secure join normally doesn't provide any additional protection.

So yes, `foo/../bar` would be treated as `bar`, but how is this a problem? The user could have written `template://bar` with the same result, so the file is already accessible.

The "chroot" for this use case is `/usr/local/share/lima/examples`, so any file under that root is accessible to the user, but since this root directory is built into the app itself, the user cannot escape it. This is enforced in `readTemplate`, because the root directory there is not based on user input.

Why should `foo/../bar` be `foo/bar`, but `foo/bar/../baz` is `foo/baz` and not `foo/bar/baz`? What makes the first directory level under `.../examples/` special, so it needs different treatment? It is breaking path semantics for no actual security benefit.

AkihiroSuda: fixed

jandubois (approved):

PR #811, opened 2022-04-22
examples: update Ubuntu (22.04), Debian, Arch; nerdctl: update to v0.19.0

AkihiroSuda:
nerdctl v0.19.0 implements the long-wanted `nerdctl cp`: https://github.com/containerd/nerdctl/releases/tag/v0.19.0

TODO: update Alpine too:
- https://github.com/lima-vm/alpine-lima/pull/58

PR #815, opened 2022-04-25
Bump alpine-lima example from Alpine 3.14 to 3.15

jandubois:


AkihiroSuda:
Thanks, but covered in https://github.com/lima-vm/lima/pull/811

PR #818, opened 2022-04-25
Set the kubelet cgroup driver to systemd (v2)

### LINKED ISSUE #817: k8s.yaml example does not work after upgrade to Ubuntu 22.04
### Description

I discovered lima yesterday and I was glad to find that `k8s.yaml` provides everything I need to run a disposable test instance of Kubeadm that can successfully connect to my synology-csi through their open source driver, (other solutions like k3d, kind all did not succeed in my trials).

However this morning I saw an upgrade was made from 20.04 to 22.04 ubuntu, and now etcd seems to restart for some unknown reason every 3-5 minutes and it results in cluster services dropping out and all sorts of chaos inside the cluster.

It's easy enough to revert to [the previous commit](https://github.com/lima-vm/lima/blob/ae3e43ef9339b3e20c272f4d99865a9a784c9246/examples/k8s.yaml) and I have done that, and everything seems to be fine again. But I thought I should report this, since I see a lot of things were upgraded at once, it is possible that nobody is aware, (or maybe this is an issue with my machine, but it seems not.)

I am using an M1 Macbook and the only changes I made to the VM config were to add:

```
cpus: 5
memory: 8GiB
```

Just from glancing at the htop statistics in the VM before things go to crap, it does not look like we're running out of memory or anything else jumping out at me as the reason for this fault.

afbjorklund:
See https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/configure-cgroup-driver/

```
NAME       STATUS   ROLES                  AGE     VERSION   INTERNAL-IP    EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION      CONTAINER-RUNTIME
lima-k8s   Ready    control-plane,master   2m32s   v1.23.6   192.168.5.15   <none>        Ubuntu 22.04 LTS   5.15.0-25-generic   containerd://1.6.2
```

Closes #817

### DIFF
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -110,6 +110,10 @@ provision:
       - "127.0.0.1"
     networking:
       podSubnet: "10.244.0.0/16" # --pod-network-cidr
+    ---
+    kind: KubeletConfiguration
+    apiVersion: kubelet.config.k8s.io/v1beta1
+    cgroupDriver: systemd
     EOF
     kubeadm init --config kubeadm-config.yaml
     # Installing a Pod network add-on
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -75,6 +75,23 @@ provision:
   script: |
     #!/bin/bash
     set -eux -o pipefail
+    grep SystemdCgroup /etc/containerd/config.toml && exit 0
+    # Configuring a cgroup driver
+    cat <<EOF >>/etc/containerd/config.toml
+      [plugins]
+        [plugins."io.containerd.grpc.v1.cri"]
+          [plugins."io.containerd.grpc.v1.cri".containerd]
+            [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
+              [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
+                runtime_type = "io.containerd.runc.v2"
+                [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
https://github.com/apps/dco

afbjorklund:
I don't think this change is needed, it was already defaulting to cgroups v2 (systemd).

examples/k8s.yaml
```diff
@@ -71,6 +71,23 @@ provision:
     rm -f /etc/cni/net.d/*.conf*
     apt-get install -y kubelet kubeadm kubectl && apt-mark hold kubelet kubeadm kubectl
     systemctl enable --now kubelet
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    grep SystemdCgroup /etc/containerd/config.toml && exit 0
+    # Configuring a cgroup driver
+    cat <<EOF >>/etc/containerd/config.toml
```
AkihiroSuda, line 80: Maybe we also have to ensure that the config.toml has `version = 2` header?

afbjorklund: Sure, I don't know what to do if it is not - but we could have the script "exit 1"

It has only been tested with the current config:

```toml
  version = 2
  [proxy_plugins]
    [proxy_plugins."stargz"]
      type = "snapshot"
      address = "/run/containerd-stargz-grpc/containerd-stargz-grpc.sock"
```

Not sure why the extra indent ?

AkihiroSuda (approved):
Thanks

PR #827, opened 2022-04-28
experimental support for riscv64

AkihiroSuda:
```
$ limactl start template://experimental/riscv64
...
INFO[0201] READY. Run `limactl shell riscv64` to open the shell.  

$ limactl shell riscv64 uname -a
Linux lima-riscv64 5.15.0-1007-generic #7-Ubuntu SMP Tue Apr 12 16:07:30 UTC 2022 riscv64 riscv64 riscv64 GNU/Linux

$ limactl shell riscv64 cat /proc/cpuinfo
processor       : 0
hart            : 0
isa             : rv64imafdcsu
mmu             : sv48

processor       : 1
hart            : 1
isa             : rv64imafdcsu
mmu             : sv48

processor       : 2
hart            : 2
isa             : rv64imafdcsu
mmu             : sv48

processor       : 3
hart            : 3
isa             : rv64imafdcsu
mmu             : sv48
```

No support for containerd yet.

Tested on macOS 12.3.1 Intel with QEMU 6.2.0_1 (Homebrew).
Took 200 seconds for completion of the entire boot script.


afbjorklund:
Nice!

This probably makes more sense long-term, than the PowerPC experiments that I did...
(someone was wondering if lima supported `ppc64le` and `s390x`, since ubuntu does)

afbjorklund:
There only seems to be one board available ?

https://www.sifive.com/boards/hifive-unmatched

At least when glancing through the docs:

https://wiki.ubuntu.com/RISC-V

pkg/limayaml/validate.go
```diff
@@ -18,41 +18,63 @@ import (
...
+	return nil
+}
+
 func Validate(y LimaYAML, warn bool) error {
 	switch *y.Arch {
-	case X8664, AARCH64:
+	case X8664, AARCH64, RISCV64:
 	default:
-		return fmt.Errorf("field `arch` must be %q or %q , got %q", X8664, AARCH64, *y.Arch)
+		return fmt.Errorf("field `arch` must be %q, %q or %q, got %q", X8664, AARCH64, RISCV64, *y.Arch)
```
jandubois, line 48: Nitpicking: mixing oxford comma:

```suggestion
		return fmt.Errorf("field `arch` must be %q, %q, or %q; got %q", X8664, AARCH64, RISCV64, *y.Arch)
```

pkg/limayaml/validate.go
```diff
@@ -18,41 +18,63 @@ import (
...
+			return err
 		}
-		switch f.Arch {
-		case X8664, AARCH64:
-		default:
-			return fmt.Errorf("field `images.arch` must be %q or %q, got %q", X8664, AARCH64, f.Arch)
+		if f.Kernel != nil {
+			if err := validateFile(f.Kernel.File, fmt.Sprintf("images[%d].kernel", i)); err != nil {
+				return err
+			}
```
jandubois, line 61: Should there be a requirement that the kernel arch matches the image arch?

Same question for Initrd.

Given that all these are loaded before binfmt_misc, I don't see how you could mix architectures between those 3 files.

pkg/qemu/qemu.go
```diff
@@ -398,7 +430,31 @@ func Cmdline(cfg Config) (string, []string, error) {
...
+	kernelCmdline := filepath.Join(cfg.InstanceDir, filenames.KernelCmdline)
+	initrd := filepath.Join(cfg.InstanceDir, filenames.Initrd)
+	if _, err := os.Stat(kernel); err == nil {
+		args = appendArgsIfNoConflict(args, "-kernel", kernel)
+	}
+	if b, err := os.ReadFile(kernelCmdline); err == nil {
+		args = appendArgsIfNoConflict(args, "-append", string(b))
+	}
+	if _, err := os.Stat(initrd); err == nil {
+		args = appendArgsIfNoConflict(args, "-initrd", kernel)
```
jandubois, line 456: Shouldn't this be `initrd`?

```suggestion
		args = appendArgsIfNoConflict(args, "-initrd", initrd)
```

jandubois (commented):
Looks good to me, except for the question about `-initrd` parameter.

pkg/limayaml/validate.go
AkihiroSuda: fixed

pkg/limayaml/validate.go
AkihiroSuda: fixed

pkg/qemu/qemu.go
AkihiroSuda: thanks, fixed

jandubois (approved):

Issue #834, opened 2022-04-30
Forward unix domain socket from the host into the guest


dee-kryvenko:
### Description

```
portForwards:
  - guestSocket: "/foo/test.sock"
    hostSocket: "/Users/me/tmp/test.sock"
```

I was trying something like this only to realize to works the opposite direction I need. Is there a way to forward FROM host TO the guest?

Also, with 9p - will unix sockets just work via mounts without extra forwarding?

afbjorklund:
The most straight-forward is to use `tcp`, but you can use `ssh -L` in the guest to tunnel `unix` sockets from the host.

The 9p file system will show the socket files, but they don't really work for connecting. I don't think sshfs works either...

Note that you would have to set up the authentication in the guest (i.e. the keys), so that it can access the host.

Normally this is only provided for the sftp server in reverse sshfs, there is no generic ssh login method provided.

----

For `ssh-agent` in particular, one would normally use `-A`.

> Connections to ssh-agent may be forwarded from further remote hosts using the -A option to ssh(1) (but see the caveats documented therein), avoiding the need for authentication data to be stored on other machines.  Authentication passphrases and private keys never go over the network: the connection to the agent is forwarded over SSH remote connections and the result is returned to the requester, allowing the user access to their identities anywhere in the network in a secure fashion.

Where the `ssh` man page just has the usual warnings:

> `-A`      Enables forwarding of connections from an authentication agent such as ssh-agent(1).  This can also be specified on a per-host basis in a configuration file.
>
> Agent forwarding should be enabled with caution.  Users with the ability to bypass file permissions on the remote host (for the agent's UNIX-domain socket) can access the local agent through the forwarded connection.  An attacker cannot obtain key material from the agent, however they can perform operations on the keys that enable them to authenticate using the identities loaded into the agent.

afbjorklund:
If you want this feature, the most straight-forward would be to add it to the `limactl shell` command...

A workaround in the meantime, would be to use `limactl show-ssh` and add the `-A` option to the args.

dee-kryvenko:
The thing is... I don't have keys and I don't want to have them. They are in my yubikey where no one can really get them... I am trying to forward my GPG agent that pretends to be SSH agent into my containers, where I want to build some code. But I was trying to keep this open - who knows why someone might need to forward sockets, exactly. Should be available on demand pretty much like port forwarding...

afbjorklund:
[...]

PR #836, opened 2022-05-01
Enable reverse forwarding of portForwards

afbjorklund:
By default, the portForwards will be forwarded
from local (host) to remote (guest). Flag: `-L`

With reverse, the portForwards will be forwarded
from remote (guest) to local (host). Flag: `-R`

Only implemented for unix sockets at the moment.

I kept the confusing wording, of the logging entries.

Issue #834

Tested by forwarding a memcached socket from the host.

`memcached -s mc.sock`

```yaml
portForwards:
 - guestSocket: "/tmp/mc.sock"
   hostSocket: "{{.Home}}/mc.sock"
   reverse: true
```

`lima sudo apt install -y libmemcached-tools`
`lima MEMCACHED_SERVERS=/tmp/mc.sock memcping`

----

ssh says:

`-L local_socket:remote_socket`
Specifies that connections to the given TCP port or Unix socket on the local (client) host are to be forwarded to the given host and port, or Unix socket, on the remote side.
i.e. from host, to guest
`-R remote_socket:local_socket`
Specifies that connections to the given TCP port or Unix socket on the remote (server) host are to be forwarded to the local side.
i.e. from guest, to host

lima says:

reverse: false
`logrus.Infof("Forwarding %q (guest) to %q (host)", remote, local)`
reverse: true
`logrus.Infof("Forwarding %q (host) to %q (guest)", local, remote)`

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -266,8 +266,10 @@ networks:
 #
 # - guestSocket: "/run/user/{{.UID}}/my.sock"
 #   hostSocket: mysocket
+# # default: reverse: false
 # # "guestSocket" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
 # # "hostSocket" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
+# # "reverse" can only be used for unix sockets right now, not for tcp sockets.
 # # Put sockets into "{{.Dir}}/sock" to avoid collision with Lima internal sockets!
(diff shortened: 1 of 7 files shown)


AkihiroSuda (approved):
Thanks

jandubois:
> Only implemented for unix sockets at the moment.

I think the "documentation" in `default.yaml` should mention that this option is only supported for unix sockets, and the code in `pkg/limayaml/validate.go` should throw an error when it is set for regular TCP ports.

jandubois (approved):
Thanks, LGTM, but I only read the code, I didn't actually test it.

PR #838, opened 2022-05-01
Add option to show messages in limactl list

### LINKED ISSUE #614: Add command to show the start message again
### Description

In case you missed it and it scrolled off, or if you are doing it from a GUI and want to show a dialog to the user.

Currently the information is shown with the rest of the clutter, which can make it hard to screen scrape (or read)

```
INFO[0000] Using the existing instance "docker"         
INFO[0000] [hostagent] Starting QEMU (hint: to watch the boot progress, see "/home/anders/.lima/docker/serial.log") 
INFO[0000] SSH Local Port: 43151                        
INFO[0000] [hostagent] Waiting for the essential requirement 1 of 5: "ssh" 
INFO[0021] [hostagent] The essential requirement 1 of 5 is satisfied 
INFO[0021] [hostagent] Waiting for the essential requirement 2 of 5: "user session is ready for ssh" 
INFO[0021] [hostagent] The essential requirement 2 of 5 is satisfied 
INFO[0021] [hostagent] Waiting for the essential requirement 3 of 5: "sshfs binary to be installed" 
INFO[0021] [hostagent] The essential requirement 3 of 5 is satisfied 
INFO[0022] [hostagent] Waiting for the essential requirement 4 of 5: "/etc/fuse.conf to contain \"user_allow_other\"" 
INFO[0022] [hostagent] The essential requirement 4 of 5 is satisfied 
INFO[0022] [hostagent] Waiting for the essential requirement 5 of 5: "the guest agent to be running" 
INFO[0022] [hostagent] The essential requirement 5 of 5 is satisfied 
INFO[0022] [hostagent] Mounting "/home/anders"          
INFO[0022] [hostagent] Mounting "/tmp/lima"             
INFO[0022] [hostagent] Waiting for the optional requirement 1 of 1: "user probe 1/1" 
INFO[0022] [hostagent] Forwarding "/run/user/1000/docker.sock" (guest) to "/home/anders/.lima/docker/sock/docker.sock" (host) 
INFO[0022] [hostagent] Forwarding "/run/lima-guestagent.sock" (guest) to "/home/anders/.lima/docker/ga.sock" (host) 
INFO[0022] [hostagent] Not forwarding TCP 127.0.0.53:53 
INFO[0022] [hostagent] The optional requirement 1 of 1 is satisfied 
INFO[0022] [hostagent] Not forwarding TCP 0.0.0.0:22    
INFO[0022] [hostagent] Not forwarding TCP [::]:22       
INFO[0022] [hostagent] Waiting for the final requirement 1 of 1: "boot scripts must have finished" 
INFO[0025] [hostagent] The final requirement 1 of 1 is satisfied 
INFO[0025] READY. Run `limactl shell docker` to open the shell. 
INFO[0025] Message from the instance "docker":          
$ export DOCKER_HOST=unix:///home/anders/.lima/docker/sock/docker.sock
$ docker ...
```

Something like `limactl message` to show the message again, perhaps ? I could just filter out the INFO, otherwise.

afbjorklund:
Closes #614

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -11,6 +11,7 @@ import (
 	"text/template"
 
 	"github.com/docker/go-units"
+	"github.com/lima-vm/lima/pkg/start"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -45,6 +46,7 @@ func newListCommand() *cobra.Command {
 
 	listCommand.Flags().StringP("format", "f", "", "Format the output using the given Go template")
 	listCommand.Flags().Bool("list-fields", false, "List fields available for format")
+	listCommand.Flags().Bool("message", false, "Only show names and messages")
 	listCommand.Flags().Bool("json", false, "JSONify output")
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
 
@@ -63,6 +65,10 @@ func instanceMatches(arg string, instances []string) []string {
 
 func listAction(cmd *cobra.Command, args []string) error {
 	quiet, err := cmd.Flags().GetBool("quiet")
+	if err != nil {
+		return err
+	}
+	message, err := cmd.Flags().GetBool("message")
 	if err != nil {
 		return err
 	}
@@ -94,6 +100,9 @@ func listAction(cmd *cobra.Command, args []string) error {
 	if quiet && jsonFormat {
 		return errors.New("option --quiet conflicts with --json")
 	}
+	if message && jsonFormat {
+		return errors.New("option --message conflicts with --json")
+	}
 	if goFormat != "" && jsonFormat {
 		return errors.New("option --format conflicts with --json")
 	}
@@ -120,6 +129,29 @@ func listAction(cmd *cobra.Command, args []string) error {
 	if quiet {
 		for _, instName := range instances {
 			fmt.Fprintln(cmd.OutOrStdout(), instName)
+		}
+		return nil
+	}
+
+	if message {
+		for _, instName := range instances {
+			fmt.Fprintln(cmd.OutOrStdout(), instName)
+			inst, err := store.Inspect(instName)
+			if err != nil {
+				logrus.WithError(err).Errorf("instance %q does not exist?", instName)
+				continue
+			}
+			msg, err := start.GetMessage(inst)
+			if err != nil {
+				logrus.WithError(err).Errorf("instance %q get message error", instName)
+				continue
+			}
+			if msg == nil {
+				msg = []string{"<no message>"}
+			}
+			for _, line := range msg {
+				fmt.Fprintln(cmd.OutOrStdout(), fmt.Sprintf("  %s", line))
+			}
 		}
 		return nil
 	}
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -244,30 +244,42 @@ func LimactlShellCmd(instName string) string {
 	return shellCmd
 }
 
-func ShowMessage(inst *store.Instance) error {
+func GetMessage(inst *store.Instance) ([]string, error) {
 	if inst.Message == "" {
-		return nil
+		return nil, nil
 	}
 	t, err := template.New("message").Parse(inst.Message)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	data, err := store.AddGlobalFields(inst)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	var b bytes.Buffer
 	if err := t.Execute(&b, data); err != nil {
-		return err
+		return nil, err
 	}
 	scanner := bufio.NewScanner(&b)
-	logrus.Infof("Message from the instance %q:", inst.Name)
+	lines := []string{}
 	for scanner.Scan() {
 		// Avoid prepending logrus "INFO" header, for ease of copypasting
-		fmt.Println(scanner.Text())
+		lines = append(lines, scanner.Text())
 	}
 	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+	return lines, nil
+}
+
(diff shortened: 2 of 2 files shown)


AkihiroSuda:
Maybe we should rather have `limactl describe` or `limactl inspect`?

cmd/limactl/list.go
```diff
@@ -94,6 +100,9 @@ func listAction(cmd *cobra.Command, args []string) error {
 	if quiet && jsonFormat {
 		return errors.New("option --quiet conflicts with --json")
 	}
+	if message && jsonFormat {
+		return errors.New("option --message conflicts with --json")
+	}
```
jandubois, line 105: `--message` also does conflict with `--quiet` (as does `--format`, so we missed that as well before).

Not sure if it is worth checking for all those combination though; we could just say that `--quiet` takes precedence over the others... 😄

jandubois (approved):
Thanks, LGTM

jandubois:
> Maybe we should have `limactl describe` or `limactl inspect` rather than overloading `limactl list`?

I'm fine having this as an alternative output format for `list` (as implemented by this PR), but maybe I'm biased as that was my original suggestion. But really, it is just formatting a subset of the information in a slightly different way.

cmd/limactl/list.go
afbjorklund: And --quiet is somewhat weird anyway, it is usually just the opposite of --verbose. Here it is more like --name-only

jandubois: Yes, `--format`, `--json`, `--quiet`, and `--message` are all alternative output formats, and you obviously can only choose one of them. So `--quiet` is indeed a weird name, and maybe we should introduce an alias (`--name`?), and mark `--quiet` as "hidden"?

AkihiroSuda:
> > Maybe we should have `limactl describe` or `limactl inspect` rather than overloading `limactl list`?
> 
> I'm fine having this as an alternative output format for `list` (as implemented by this PR), but maybe I'm biased as that was my original suggestion. But really, it is just formatting a subset of the information in a slightly different way.

I'm feeling that this PR is too much overloading the `list` command.

afbjorklund:
Okay, let's do another name.

jandubois:
> Okay, let's do another name.

Well, I disagree with creating another command, only to present the data in a slightly different format.

In that case I would rather not include the functionality at all and maybe just show in the docs how users can easily build variations of this themselves:

```console
$ cat message
#!/bin/sh
for INSTANCE in $(limactl ls --quiet); do
    MESSAGE="<no message>"
    FORMAT=$(limactl ls --format '{{.Message}}' $INSTANCE)
    if [ -n "$FORMAT" ]; then
        MESSAGE=$(limactl ls --format "$FORMAT" $INSTANCE)
    fi
    echo "$INSTANCE"
    echo "$MESSAGE" | sed 's/^/  /'
done

$ ./message
default
  <no message>
docker
  ------
  docker context create lima --docker "host=unix:///Users/jan/.lima/docker/sock/docker.sock"
  docker context use lima
  docker run hello-world
  ------
riscv64
  <no message>
```

FWIW I still disagree that this is overloading the `list` command. To me this is not much different than having `--label-columns`, `--output`, and `--template` option in `kubectl get`. They just allow you to format the output different ways. But then, maybe this functionality isn't really that important.

afbjorklund:
Seems to me like we are back to hacking scripts and reading comments hidden in the yaml.

I just wanted a simple command to show the message/README, with paths expanded etc.

jandubois:
> I just wanted a simple command to show the message/README, with paths expanded etc.

I think this PR does provide exactly that. I don't see how adding top-level commands for slight variations really help usability. Anyways, I've stated my opinion; in the end I will go with whatever @AkihiroSuda decides.

afbjorklund:
If it helps, I can split the cmd and the pkg.

jandubois:
> Then again, I also wanted the client command installed and the configuration set up for you.

That's essentially how [colima](https://github.com/abiosoft/colima/blob/v0.1.0/colima) was started.

afbjorklund:
Seems a bit much to fork it just for that.

jandubois:
> But I thought the minikube kubectl cmd was quite handy, even if I just "borrowed" it from k3s.

You lost me here. How is this related to `limactl ls --message`?

AkihiroSuda:
Maybe we should have `limactl ls --format '{{.MessageExpanded}}' docker` to expand variables in the message?

jandubois:
> Maybe we should have `limactl ls --format '{{.MessageExpanded}}' docker` to expand variables in the message?
> 
> Or add a template function like `limactl ls --format '{{expand .Message}}' docker`

[As I wrote before](https://github.com/lima-vm/lima/issues/614#issuecomment-1025229004) I would also be fine with just storing only the expanded message; why would you ever need the original template again?

I understand that this would be an incompatible change, but I honestly can see **anybody** relying on the current behaviour. Adding `.MessageExpanded` is probably the prudent choice though.

afbjorklund:
> > But I thought the minikube kubectl cmd was quite handy, even if I just "borrowed" it from k3s.
> 
> You lost me here. How is this related to `limactl ls --message`?

The end goal of the current message, is to show the user what command to run and how to set up the config.

Commands like `nerdctl.lima` or `minikube kubectl` instead run it for you...

Other than that, no relation

afbjorklund:
Also the feature is not so important, just trying to mimic some of the old `docker-machine` and `podman-machine` behavior.

AkihiroSuda:
> I would also be fine with just storing only the expanded message;

SGTM too

jandubois:
So with #903 merged, what is the plan for this PR?

@afbjorklund should it be closed now? Since @AkihiroSuda objects to adding more special-purpose output formats to `list`, I don't see a way forward, and with #903 merged, the need is also less (although not completely gone).

afbjorklund:
We can close it

PR #843, opened 2022-05-04
Unmount mount points under /tmp before moving to persistent volume

jandubois:
With 9p and cloud-init, the mounts happen before the provisioning scripts execute. So by the time we move `/tmp` to `/mnt/data/tmp` the `/tmp/lima` directory is already mounted, blocking the move.


AkihiroSuda (approved):
Thanks

Issue #515, opened 2022-01-01
Add hints on where to download missing host clients


afbjorklund:
Currently we mention that they are to be installed on the host: (just "assumed")

examples/docker.yaml:  To run `docker` on the host (assumes docker-cli is installed):
examples/podman.yaml:  To run `podman` on the host (assumes podman-remote is installed):
examples/k3s.yaml:  To run `kubectl` on the host (assumes kubectl is installed):
examples/k8s.yaml:  To run `kubectl` on the host (assumes kubectl is installed):

We could do a LookPath, and show a URL on where to install it from (if it is missing) ?

https://docs.docker.com/engine/install/binaries/
https://podman.io/getting-started/installation
https://kubernetes.io/docs/tasks/tools/#kubectl
https://kubernetes.io/docs/tasks/tools/#kubectl

The cmd and url would be given in the YAML file.

~~Alternatively it could be downloaded automatically...~~

----

No host clients available for nerdctl, so run with lima:

* `lima nerdctl`

* `lima ctr`

* `lima buildctl`

Theoretically everyone uses Homebrew, makes it easy.

* `brew install docker`

* `brew install podman`

* `brew install kubectl`

afbjorklund:
Should make sure to link to Docker Engine and Podman Remote, rather than to Docker Desktop and Podman Machine.

i.e. just need the client binary (a small statically linked go program), don't need the whole server and bundled VM solution.

jandubois:
> Should make sure to link to Docker Engine

You mean "Docker CLI", not "Docker Engine", I think.

Also note that we don't need a host-client for `nerdctl`; it only runs inside the VM. We already have the `nerdctl.lima` wrapper script.

I'm somewhat opposed to extend lima with knowledge about particular client applications (but documentation is obviously fine). We should not try to replace platform package managers.

afbjorklund:
> You mean "Docker CLI", not "Docker Engine", I think.

I actually meant the _product_ pages, since linking vendor:

~~https://docs.docker.com/desktop/~~

https://docs.docker.com/engine/

When it comes to deb packages, you are right. (`docker-cli`)
In brew it is only called `docker`, so I guess the names vary ?
The system package is called `docker.io` (a naming conflict)

Typically it looks something like this, throw away the `dockerd`:
https://download.docker.com/linux/static/stable/x86_64/docker-20.10.12.tgz

afbjorklund:
[...]

PR #855, opened 2022-05-11
Add client wrappers for docker and podman, similar to nerdctl wrapper

afbjorklund:
These will call the clients, without having to clobber the global configuration.

~~You only need to run the **first** command from the "message", not the second:~~

<details>
<pre>
INFO[0128] Message from the instance "docker":          
------
docker context create lima --docker "host=unix:///home/anders/.lima/docker/sock/docker.sock"
docker context use lima
docker run hello-world
------
</pre>

<pre>
INFO[0018] Message from the instance "podman":          
------
podman system connection add lima "unix:///home/anders/.lima/podman/sock/podman.sock"
podman system connection default lima
podman --remote run quay.io/podman/hello
------
</pre>
</details>

EDIT: Don't have to create the context/connection, using env variables instead.

They were mentioned as comments in the YAML files, if you looked close enough...

Sometimes the commands are missed, then it would be useful to view them:

* #614

Installing the clients _is_ painful, but you can use `brew` and it was scoped out:

* ~#515~

----

~~Only one VM of each kind is supported, since the name "lima" is hardcoded.~~

Note that you need `podman-remote` for [v3](https://github.com/containers/podman/releases/tag/v3.4.4), in order to talk to the podman VM:

```
Error: unable to connect to Podman socket: server API version is too old. Client "4.0.0" server "3.4.4"
```

If the binary clients are _missing_, it will default to SSH - just as for `nerdctl`...

`limactl shell docker docker`

`limactl shell podman podman`

### DIFF
--- a/cmd/docker.lima
+++ b/cmd/docker.lima
@@ -1,0 +1,18 @@
+#!/bin/sh
+set -eu
+: "${LIMA_INSTANCE:=docker}"
+: "${DOCKER:=docker}"
+
+if [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
+  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl start --name=$LIMA_INSTANCE template://docker\` to create a new instance" >&2
+  exit 1
+fi
+DOCKER=$(command -v "$DOCKER" || true)
+if [ -n "$DOCKER" ]; then
+  DOCKER_HOST=$(limactl list "$LIMA_INSTANCE" --format 'unix://{{.Dir}}/sock/docker.sock')
+  export DOCKER_HOST
+  exec "$DOCKER" "$@"
+else
+  export LIMA_INSTANCE
+  exec lima docker "$@"
+fi
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,8 @@ binaries: clean \
 	_output/bin/lima \
 	_output/bin/limactl \
 	_output/bin/nerdctl.lima \
+	_output/bin/docker.lima \
+	_output/bin/podman.lima \
 	_output/share/lima/lima-guestagent.Linux-x86_64 \
 	_output/share/lima/lima-guestagent.Linux-aarch64 \
 	_output/share/lima/lima-guestagent.Linux-riscv64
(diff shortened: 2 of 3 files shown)


cmd/docker.lima
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -eu
+: "${LIMA_INSTANCE:=docker}"
+: "${LIMA_DOCKER:=docker}"
+
+DOCKER=$(command -v "$LIMA_DOCKER" || true)
+if [ -n "$DOCKER" ]; then
+  context="lima" # TODO: multiple contexts
+  exec "$DOCKER" --context=$context "$@"
```
jandubois, line 9: Instead of using docker contexts you could specify `DOCKER_HOST` explicitly:

```shell

: "${LIMA_HOME:=$HOME/.lima}"

[...]

if [ -n "$DOCKER" ]; then
  export DOCKER_HOST="unix://$LIMA_HOME/$LIMA_INSTANCE/sock/docker.sock"
  exec "$DOCKER" "$@"
else

[...]
```

That way you can use multiple instances, and you don't need to create contexts at all; you just need to set `LIMA_INSTANCE`.

afbjorklund: It was like that, before...

jandubois: This was just a suggestion to your `# TODO: multiple contexts` comment, and I thought if you have to specify a context/instance anyways, then you don't really need to create a context.

This is just for `docker.lima`. I would still want to create a docker context because I normally don't use multiple docker sockets simultaneously, and I want to use `docker` instead of `docker.lima`.

Note that my suggestion above does **not** require setting any environment variables for the default case; it should work out-of-the-box like:

```console
limactl start template://docker
docker.lima run -it busybox
```

AkihiroSuda: Env var SGTM

cmd/docker.lima
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -eu
+: "${LIMA_INSTANCE:=docker}"
```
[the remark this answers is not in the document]
afbjorklund: Currently it just shows the default message, which is missing the yaml / template:

cmd/docker.lima
```diff
@@ -0,0 +1,13 @@
...
+: "${LIMA_DOCKER:=docker}"
```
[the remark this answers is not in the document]
afbjorklund: There is a risk that other programs are using that, but sure. Normally it used for the same purpose though, so should be fine...

AkihiroSuda:
Looks good but please squash commits

AkihiroSuda (approved):

jandubois (approved):
LGTM

I was briefly wondering about the case where the instance exists, but is not running, but I think the default error message from the docker CLI is good enough:

```console
$ docker.lima ps
Cannot connect to the Docker daemon at unix:///Users/jan/.lima/docker/sock/docker.sock. Is the docker daemon running?
```

Issue #342, opened 2021-10-18
Installing sshfs fails on CentOS


afbjorklund:
```
[   38.495119] cloud-init[1376]: + dnf install -y fuse-sshfs
[   39.845344] cloud-init[1376]: CentOS Linux 8 - AppStream                       13 MB/s | 9.3 MB     00:00
[   42.306361] cloud-init[1376]: CentOS Linux 8 - BaseOS                         9.6 MB/s | 7.5 MB     00:00
[   44.128504] cloud-init[1376]: CentOS Linux 8 - Extras                          21 kB/s |  10 kB     00:00
[   44.950542] cloud-init[1376]: No match for argument: fuse-sshfs
[   44.973079] cloud-init[1376]: Error: Unable to find a match: fuse-sshfs
```

Apparently you need to enable "PowerTools", in order to have `sshfs`...

https://centos.pkgs.org/8/centos-powertools-x86_64/fuse-sshfs-2.8-5.el8.x86_64.rpm.html

And probably should use "yum" rather than "dnf", on this distribution ?

```
# yum --enablerepo powertools install -y fuse-sshfs
CentOS Linux 8 - PowerTools                                                                                                              2.6 MB/s | 2.4 MB     00:00    
Last metadata expiration check: 0:00:01 ago on Mon 18 Oct 2021 03:43:59 PM UTC.
Dependencies resolved.
=========================================================================================================================================================================
 Package                                  Architecture                         Version                                    Repository                                Size
=========================================================================================================================================================================
Installing:
 fuse-sshfs                               x86_64                               2.8-5.el8                                  powertools                                56 k
Installing dependencies:
 fuse                                     x86_64                               2.9.7-12.el8                               baseos                                    82 k

Transaction Summary
=========================================================================================================================================================================
Install  2 Packages
```

afbjorklund:
Things got even worse in CentOS Stream 9, there it is not in EPEL and not in repo either. Can use binary from fc34...

Maybe lima should host a custom repo for these EL distributions, so that one can just install "fuse-sshfs" from there ?

jeffguorg:
Alma Linux 9 beta is already released. RHEL9 is released and Rocky Linux 9 is expected to be released in a few month.

I found EL's office repos does not contain sshfs any more. Instead there is a fuse-sshfs in  epel-release of every el version:
```
[root@lima-cs9 lima-cidata]# dnf info fuse-sshfs
[...]

PR #856, opened 2022-05-12
Install epel-release for EL distros

jeffguorg:
https://github.com/lima-vm/lima/issues/342#issuecomment-1124750969

I guess EL repos are putting sshfs into epel-release thus there is no sshfs in Base repos. So we should install epel-release first to support EL 9 or maybe later releases.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
@@ -67,7 +67,7 @@ elif command -v dnf >/dev/null 2>&1; then
 			dnf_install_flags="${dnf_install_flags} --repo ol8_baseos_latest --repo ol8_codeready_builder"
 		elif grep -q "release 8" /etc/system-release; then
 			dnf_install_flags="${dnf_install_flags} --enablerepo powertools"
-		else
+		elif grep -q 'ID_LIKE=".*rhel.*"' /etc/os-release; then
 			# Workaround for disappeared sshfs in Base repos
 			# shellcheck disable=SC2086
 			dnf install ${dnf_install_flags} epel-release
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
@@ -67,6 +67,11 @@ elif command -v dnf >/dev/null 2>&1; then
 			dnf_install_flags="${dnf_install_flags} --repo ol8_baseos_latest --repo ol8_codeready_builder"
 		elif grep -q "release 8" /etc/system-release; then
 			dnf_install_flags="${dnf_install_flags} --enablerepo powertools"
+		else
+			# Workaround for disappeared sshfs in Base repos
+			# shellcheck disable=SC2086
+			dnf install ${dnf_install_flags} epel-release
+			dnf_install_flags="${dnf_install_flags} --enablerepo epel"
 		fi
 		# shellcheck disable=SC2086
 		dnf install ${dnf_install_flags} ${pkgs}


pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -67,6 +67,11 @@ elif command -v dnf >/dev/null 2>&1; then
 			dnf_install_flags="${dnf_install_flags} --repo ol8_baseos_latest --repo ol8_codeready_builder"
 		elif grep -q "release 8" /etc/system-release; then
 			dnf_install_flags="${dnf_install_flags} --enablerepo powertools"
+		else
+			# Workaround for disappeared sshfs in Base repos
+			# shellcheck disable=SC2086
+			dnf install ${dnf_install_flags} epel-release
+			dnf_install_flags="${dnf_install_flags} --enablerepo epel"
```
AkihiroSuda, line 74: Unneeded on Fedora and its non-EL derivatives (how to detect?)

afbjorklund: The systemd os-release isn't very helpful:

```
ID="centos"
ID_LIKE="rhel fedora"
```

i.e. it has _both_ of rhel (epel) and fedora

afbjorklund: But probably looking for "rhel" in either ID or ID_LIKE field of /etc/os-release, should be enough for "is EL" (needs EPEL)

afbjorklund:
> I guess EL repos are putting sshfs into epel-release thus there is no sshfs in Base repos. So we should install epel-release first to support EL 9 or maybe later releases.

I think it worked the other way around, since "fuse-sshfs" was in PowerTools for EL8 it was removed from EPEL (where it exists for EL7). Then it was removed from PowerTools again in EL9, so then it had to be added back to EPEL (for EL9).

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
jeffguorg: @AkihiroSuda @afbjorklund Sorry I'm absent for some days.

I didn't realize that this need to be restricted to rhel releases. So i pushed a new commit on the branch. I tested `grep 'ID_LIKE=".*rhel.*"' /etc/os-releases` on theses containers and this should work fine.

Is there any other problem? It looks good to me now.

afbjorklund:
A directory in the packages server, like a yum repo

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
afbjorklund: I think enabling EPEL should be up to the system administrator, and not modified by lima.

So if you enable it just for installing "sshfs", it should be disabled again after the installation.

jeffguorg: @afbjorklund good point. So what about this?

- install epel-release package and thus has it enabled by default
- disable it like `dnf config-manager --set-disabled epel-release` or something similar
- install sshfs with epel-release temporarily enabled with `--enablerepo=epel-release`

afbjorklund: Yes, this was my idea. Somewhat similar to what was already being done for CentOS 7.

```diff
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
@@ -67,6 +67,10 @@ elif command -v dnf >/dev/null 2>&1; then
 			dnf_install_flags="${dnf_install_flags} --repo ol8_baseos_latest --repo ol8_codeready_builder"
 		elif grep -q "release 8" /etc/system-release; then
 			dnf_install_flags="${dnf_install_flags} --enablerepo powertools"
+		elif grep -q "release 9" /etc/system-release; then
+			dnf install ${dnf_install_flags} epel-release
+			dnf config-manager --disable epel >/dev/null 2>&1
+			dnf_install_flags="${dnf_install_flags} --enablerepo epel"
 		fi
 		# shellcheck disable=SC2086
 		dnf install ${dnf_install_flags} ${pkgs}

```

afbjorklund:
I made a PR to clarify between CentOS (Linux) and CentOS Stream.

Including examples for stream8 and stream9, willl submit separately...

CentOS | Linux | Stream
--- | --- | ---
7 | X [deprecated] | -
8 | ~[EOL](https://www.centos.org/centos-linux-eol/)~ | X
9 | - | X [experimental]

Technically one _can_ install CentOS 8 from vault, but not a good idea.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
afbjorklund: https://github.com/lima-vm/lima/pull/901/commits/5fff4534b5eed0af7322ade6ae36469ee98326d0

afbjorklund:
@jeffguorg all EL (7/8/9) distros should be working now

PR #863, opened 2022-05-20
Set required Containerd env vars in model zshrc - fixes #862

### LINKED ISSUE #862: nerdctl run fails on Debian VM with "failed to extract layer sha256:[...]: failed to mount /var/lib/containerd/tmpmounts/[...]: operation not permitted: unknown" when zsh is the default shell
### Description

limactl version: 0.10.0
ctr github.com/containerd/containerd v1.6.2

I created a Debian VM with `limactl start --name=default template://debian`.

Starting a container with `nerdctl run -p localhost:5432:5432 postgres -e POSTGRES_PASSWORD=postgres -d` fails with the following error:

```
FATA[0015] failed to extract layer sha256:95d804481634f13fc7cc6ad18cfea900988c8290bcdd6d8948bc56256db5f68a: failed to mount /var/lib/contain
erd/tmpmounts/containerd-mount2254351966: operation not permitted: unknown
```

It's unclear if this is an upstream issue or a problem with how containerd is provisioned with the Debian template in Lima.

davebarkerxyz:
I've added .zshrc to the list of shell rc files where the default `CONTAINERD_SNAPSHOTTER` gets set. I've tested this and it fixes #862. This is my first PR in the project, so apologies if I've missed any steps.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/20-rootless-base.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/20-rootless-base.sh
@@ -5,7 +5,7 @@ set -eux
 command -v systemctl >/dev/null 2>&1 || exit 0
 
 # Set up env
-for f in .profile .bashrc; do
+for f in .profile .bashrc .zshrc; do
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #867, opened 2022-05-23
Support configuring mount locations within guests

### LINKED ISSUE #659: Allow setting of destination for mounts
As seen here, the mounts `location` is used as both the source and destination: https://github.com/lima-vm/lima/blob/ae9733c7f1d27c59334663dbe481f02141f15319/pkg/hostagent/mount.go#L55-L58

I want to be able to mount my `~/.aws/credentials` in order to use the [Amazon ECR Credential Helper](https://github.com/awslabs/amazon-ecr-credential-helper), however as `~` on my mac expands to `/Users/myusername`, the `.aws` folder is mounted there, but `$HOME` inside the VM is `/home/myusername.linux`, so the cred helper doesn't find the creds.

There's probably other ways to work around this, but would being able to set the destination for a mount be desirable?

chancez:
Closes #659

Tested with both sshfs and 9p and confirmed that the mounts were configured in the location specified by `mounts[*].mountPoint`. 

eg:
```
mounts:
- location: "~/Movies"
  mountPoint: "/movies"
```

### DIFF
--- a/pkg/hostagent/mount.go
+++ b/pkg/hostagent/mount.go
@@ -34,11 +34,16 @@ func (a *HostAgent) setupMounts(ctx context.Context) ([]*mount, error) {
 }
 
 func (a *HostAgent) setupMount(ctx context.Context, m limayaml.Mount) (*mount, error) {
-	expanded, err := localpathutil.Expand(m.Location)
+	location, err := localpathutil.Expand(m.Location)
 	if err != nil {
 		return nil, err
 	}
-	if err := os.MkdirAll(expanded, 0755); err != nil {
+
+	mountPoint, err := localpathutil.Expand(m.MountPoint)
+	if err != nil {
+		return nil, err
+	}
+	if err := os.MkdirAll(location, 0755); err != nil {
 		return nil, err
 	}
 	// NOTE: allow_other requires "user_allow_other" in /etc/fuse.conf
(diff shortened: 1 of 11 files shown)


jandubois:
Needs documentation in `examples/default.yaml`.

Replacing the empty `Guest` with the `Location` value should happen in `limayaml.FillDefaults`. This includes properly merging the value from `defaults.yaml` and `override.yaml` there as well.

pkg/limayaml/limayaml.go
```diff
@@ -64,6 +64,7 @@ type Image struct {
 
 type Mount struct {
 	Location string `yaml:"location" json:"location"` // REQUIRED
+	Guest    string `yaml:"guest,omitempty" json:"guest,omitempty"`
```
AkihiroSuda, line 67: I think this should be called GuestLocation or something else.

jandubois: > I think this should be called GuestLocation or something else.

I agree. I would suggest `mountPoint`, as that is what it is called by `mount`...

chancez: Yeah I just took the example @jandubois wrote in a related GH issue and just used that and figured we would decide on a better name in the PR. Sounds good. I'll use `mountPoint`.

chancez: Updated

pkg/hostagent/mount.go
```diff
@@ -49,34 +54,35 @@ func (a *HostAgent) setupMount(ctx context.Context, m limayaml.Mount) (*mount, e
 	if *m.SSHFS.FollowSymlinks {
 		sshfsOptions = sshfsOptions + ",follow_symlinks"
 	}
-	logrus.Infof("Mounting %q", expanded)
+	logrus.Infof("Mounting host %q to guest %q", location, mountPoint)
```
jandubois, line 57: The meaning here is clear, but if we adopt the "$LOCATION on $MOUNTPOINT" style below, then I would use it here as well:

```suggestion
	logrus.Infof("Mounting %q on %q", location, mountPoint)
```

jandubois (approved):
Thanks, LGTM!

I have some nit-picky comments though. Feel free to address, or maybe @AkihiroSuda is willing to merge as-is; I don't feel too strongly about them.

jandubois:
I also just noticed that there is another field `Mount.Target`:

https://github.com/lima-vm/lima/blob/d6aa0d46dd373772ba7834325657985beb2837a9/pkg/cidata/template.go#L41

For consistency it would be nice to rename that as well to `Mount.MountPoint` I think (only if you still make other changes anyways).

jandubois (commented):
I don't know why this comment is still pending; Github is getting confused.

pkg/hostagent/mount.go
```diff
@@ -49,34 +54,35 @@ func (a *HostAgent) setupMount(ctx context.Context, m limayaml.Mount) (*mount, e
...
 		Host:                "127.0.0.1",
 		Port:                a.sshLocalPort,
-		RemotePath:          expanded,
+		RemotePath:          mountPoint,
 		Readonly:            !(*m.Writable),
 		SSHFSAdditionalArgs: []string{"-o", sshfsOptions},
 	}
 	if err := rsf.Prepare(); err != nil {
-		return nil, fmt.Errorf("failed to prepare reverse sshfs for %q: %w", expanded, err)
+		return nil, fmt.Errorf("failed to prepare reverse sshfs for %q->%q: %w", location, mountPoint, err)
```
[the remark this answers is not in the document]
chancez: Updated

jandubois (approved):
Thanks, LGTM.

Ok to merge @AkihiroSuda?

AkihiroSuda (approved):
Thanks

PR #869, opened 2022-05-25
Make sure macOS version is in dotted-tri format

### LINKED ISSUE #868: hostagent fails to parse macOS version 12.4
### Description

I just noticed this warning with lima 0.11.0 on MacOS 12.4

```
WARN[0000] [hostagent] failed to parse macOS version "12.4": 12.4 is not in dotted-tri format
```

jandubois:
Fixes #868

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -284,6 +284,10 @@ func getMacOSProductVersion() (*semver.Version, error) {
 		return nil, fmt.Errorf("failed to execute %v: %w", cmd.Args, err)
 	}
 	verTrimmed := strings.TrimSpace(string(b))
+	// macOS 12.4 returns just "12.4\n"
+	for strings.Count(verTrimmed, ".") < 2 {
+		verTrimmed += ".0"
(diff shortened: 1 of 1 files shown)


pkg/qemu/qemu.go
```diff
@@ -284,6 +284,10 @@ func getMacOSProductVersion() (*semver.Version, error) {
 		return nil, fmt.Errorf("failed to execute %v: %w", cmd.Args, err)
 	}
 	verTrimmed := strings.TrimSpace(string(b))
+	// macOS 12.4 returns just "12.4\n"
+	for strings.Count(verTrimmed, ".") < 2 {
+		verTrimmed += ".0"
```
AkihiroSuda, line 289: += ".0.0" when count == 0

jandubois: That's why it is in a `for` loop, so it will loop twice in that case: https://go.dev/play/p/j_0P1mZWRtj

AkihiroSuda (approved):
Thanks

PR #870, opened 2022-05-25
Add LIMA_CIDATA_HOSTHOME_MOUNTPOINT

jandubois:
This can be used by provisioning scripts to locate the mount point of the user's home directory. We can't just use `/Users/$LIMA_CIDATA_USER` because we don't know if the host is Linux or macOS, so it could be `/home/$LIMA_CIDATA_USER` or something completely custom. Also `$LIMA_CIDATA_USER` is the username inside the guest and might be set to `lima` if the username on the macOS host is not a valid Linux username.

Example usage to configure client certs for the docker daemon:

```shell
ln -s $LIMA_CIDATA_HOME_MOUNTPOINT/.docker/certs.d /etc/docker/certs.d
```

### DIFF
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -118,6 +118,7 @@ The volume label is "cidata", as defined by [cloud-init NoCloud](https://cloudin
 - `LIMA_CIDATA_MNT`: the mount point of the disk. `/mnt/lima-cidata`.
 - `LIMA_CIDATA_USER`: the user name string
 - `LIMA_CIDATA_UID`: the numeric UID
+- `LIMA_CIDATA_HOSTHOME_MOUNTPOINT`: the mount point of the host home directory, or empty if not mounted
 - `LIMA_CIDATA_MOUNTS`: the number of the Lima mounts
 - `LIMA_CIDATA_MOUNTS_%d_MOUNTPOINT`: the N-th mount point of Lima mounts (N=0, 1, ...)
 - `LIMA_CIDATA_MOUNTTYPE`: the type of the Lima mounts ("reverse-sshfs", "9p", ...)
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -128,6 +128,10 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 	case limayaml.NINEP:
 		fstype = "9p"
 	}
+	hostHome, err := localpathutil.Expand("~")
+	if err != nil {
+		return err
+	}
 	for i, f := range y.Mounts {
 		tag := fmt.Sprintf("mount%d", i)
 		location, err := localpathutil.Expand(f.Location)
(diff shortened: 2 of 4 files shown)


jandubois:
This PR is in draft mode because it will have merge conflicts with #867.

Please provide feedback if there is a less clunky way to do this!

docs/internal.md
```diff
@@ -118,6 +118,7 @@ The volume label is "cidata", as defined by [cloud-init NoCloud](https://cloudin
 - `LIMA_CIDATA_MNT`: the mount point of the disk. `/mnt/lima-cidata`.
 - `LIMA_CIDATA_USER`: the user name string
 - `LIMA_CIDATA_UID`: the numeric UID
+- `LIMA_CIDATA_HOST_HOME`: the mount point of `~` or empty if not mounted
```
jandubois, line 121: Maybe `LIMA_CIDATA_HOME_MOUNTPOINT` is a better name?

chancez: Seems reasonable.

chancez:
This seems useful, especially with custom mounts.

AkihiroSuda:
Still draft?

docs/internal.md
```diff
@@ -118,6 +118,7 @@ The volume label is "cidata", as defined by [cloud-init NoCloud](https://cloudin
 - `LIMA_CIDATA_MNT`: the mount point of the disk. `/mnt/lima-cidata`.
 - `LIMA_CIDATA_USER`: the user name string
 - `LIMA_CIDATA_UID`: the numeric UID
+- `LIMA_CIDATA_HOME_MOUNTPOINT`: the mount point of `~`, or empty if not mounted
```
AkihiroSuda, line 121:
```suggestion
- `LIMA_CIDATA_HOSTHOME_MOUNTPOINT`: the mount point of the host home directory, or empty if not mounted
```

pkg/cidata/cidata.go
```diff
@@ -156,6 +156,9 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 			options += ",nofail"
 		}
 		args.Mounts = append(args.Mounts, Mount{Tag: tag, MountPoint: mountPoint, Type: fstype, Options: options})
+		if f.Location == "~" {
```
AkihiroSuda, line 159:
```suggestion
		if Expand(f.Location) == Expand("~") {
```
(pseudo code, needs to check err)

pkg/cidata/template.go
```diff
@@ -61,6 +61,7 @@ type TemplateArgs struct {
 	Env             map[string]string
 	DNSAddresses    []string
 	CACerts         CACerts
+	HomeMountPoint  string
```
AkihiroSuda, line 64:
```suggestion
	HostHomeMountPoint  string
```

docs/internal.md
jandubois: I thought it would be obvious that this is the host home, because the guest home is either `/home/${LIMA_CIDATA_USER}.linux` when you are `root`, or just `$HOME` when you are not.

But it doesn't hurt to be explicit, so I made the change.

AkihiroSuda (approved):
Thanks

PR #875, opened 2022-06-02
Forward X11 to VM, if available

jordemort:
This allows people who have XQuartz installed to easily install and run X11 applications inside lima


jordemort:
After sleeping on it, I don't think I actually want this merged as-is - I'd like to go back and make it an option under `ssh:` in the template, similar to how `forwardAgent:` is currently implemented.

jordemort:
Closing in favor of https://github.com/lima-vm/lima/pull/877

PR #876, opened 2022-06-03
Use Fedora 36 to get podman 4

dmage:
Ubuntu 22.04 provides podman 3.4.4 which is not compatible with podman 4 client that is provided by brew on macOS.

### DIFF
--- a/examples/podman.yaml
+++ b/examples/podman.yaml
@@ -12,19 +12,12 @@
 
 # This example requires Lima v0.8.0 or later
 images:
-# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Would it be possible to just install Podman 4 onto Ubuntu? (Are there debs?)

dmage:
I'm not aware of podman 4 builds for Ubuntu, maybe there is some PPA or it'll be in 22.10.

AkihiroSuda (approved):

AkihiroSuda:
Cc @afbjorklund

afbjorklund:
There is no podman 4 for ubuntu, at least not until a much later version (it is in Debian experimental)

afbjorklund:
Current workaround is `PODMAN=podman3`

https://github.com/containers/podman/releases/tag/v3.4.4

afbjorklund (approved):

PR #877, opened 2022-06-03
Allow forwarding X11 to VM

jordemort:
This allows people who have XQuartz installed to easily install and run X11 applications inside lima.

This is a nicer replacement for https://github.com/lima-vm/lima/pull/875 - it adds two new settings to the `ssh` section of the template - `forwardX11` and `forwardX11Trusted` correspond to the similarly-named ssh options, like `forwardAgent` does.

### DIFF
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -108,7 +108,7 @@ func copyAction(cmd *cobra.Command, args []string) error {
 		// arguments such as ControlPath.  This is preferred as we can multiplex
 		// sessions without re-authenticating (MaxSessions permitting).
 		for _, instDir := range instDirs {
-			sshOpts, err = sshutil.SSHOpts(instDir, false, false)
+			sshOpts, err = sshutil.SSHOpts(instDir, false, false, false, false)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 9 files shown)


jordemort:
It's unclear to me why the alpine test is failing here

jordemort:
Looks like the alpine test cleared itself up somehow? I did not intervene.

AkihiroSuda:
> Looks like the alpine test cleared itself up somehow? I did not intervene.

I restarted the CI and it became green

jandubois:
I'm not keen on adding every SSH option individually to `lima.yaml`, but I guess having a generic mechanism to provide "any" additional options might be error-prone.

So, I'm not sure what I'm saying. I guess I'm fine with adding these 2 options, but if the list continues to grow, then this might become unmaintainable.

I'm going to approve this PR based on the code changes, but will leave the merging to @AkihiroSuda

jandubois (approved):

jordemort:
@jandubois Yeah it could get ugly if there are too many more of these. I could whip up an "arbitrary SSH option" PR if you'd rather; I think it could be done at least semi-safely, by filtering out the ones `limactl` cares about for its own use.

On the other hand, for at least my personal use case, I can't see wanting much else beyond this. The UNIX socket forwarding stuff that was added covers the rest of the weird stuff I want to do (GPG agent, PulseAudio, etc)

AkihiroSuda (approved):
Thanks

AkihiroSuda:
>  "arbitrary SSH option"

SGTM, but this PR also makes sense

PR #887, opened 2022-06-07
Add example for apptainer running on rocky

afbjorklund:
The new version is not yet available for fedora or centos,
but it will be available from EPEL after the 1.1.0 release.

https://github.com/apptainer/apptainer/releases/tag/v1.0.2

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -36,6 +36,7 @@ Others:
 - [`deprecated/centos-7.yaml`](./deprecated/centos-7.yaml): [deprecated] CentOS 7
 - [`experimental/9p.yaml`](experimental/9p.yaml): [experimental] use 9p mount type
 - [`experimental/riscv64.yaml`](experimental/riscv64.yaml): [experimental] RISC-V
+- [`experimental/apptainer.yaml`](./experimental/apptainer.yaml): [experimental] [Apptainer](https://apptainer.org/)
 
 ## Tier 1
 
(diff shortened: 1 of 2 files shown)


examples/apptainer.yaml
```diff
@@ -0,0 +1,45 @@
...
+  system: false
+  user: false
+provision:
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    command -v apptainer >/dev/null 2>&1 && exit 0
+    url="https://github.com/apptainer/apptainer"
+    tag="$(basename $(curl -fs -o/dev/null -w %{redirect_url} $url/releases/latest))"
```
AkihiroSuda, line 35: Can we pin a specific version?

examples/README.md
```diff
@@ -20,6 +20,7 @@ Container engines:
 - [`docker.yaml`](./docker.yaml): Docker
 - [`podman.yaml`](./podman.yaml): Podman
 - [`singularity.yaml`](./singularity.yaml): Singularity
```
AkihiroSuda, line 22: Can we now remove singularity? Or maybe move to the `deprecated` directory

examples/apptainer.yaml
afbjorklund: Sure, current is `1.0.2`

examples/README.md
afbjorklund: This new "apptainer" package is not yet available in any distributions, though.

afbjorklund: Requested a package for aarch64, currently the _only_ architecture is x86_64

afbjorklund: I moved it to `experimental`, and then removed the arm64 arch from list (for now)

AkihiroSuda (approved):
Thanks

PR #890, opened 2022-06-08
templates: update openSUSE Leap to 15.4

AkihiroSuda:


AkihiroSuda:
```
INFO[0141] [hostagent] fusermount3: option allow_other only allowed if 'user_allow_other' is set in /etc/fuse3.conf
...
DEBU[0173] [hostagent] executed script "wait-for-remote-ready", stdout="", stderr="sshfs does not seem to be mounted on /Users/suda\n", err=failed to execute script "wait-for-remote-ready": stdout="", stderr="sshfs does not seem to be mounted on /Users/suda\n": exit status 1 
WARN[0173] [hostagent] failed to confirm whether /Users/suda [remote] is successfully mounted
```

pkg/hostagent/requirements.go
```diff
@@ -104,15 +104,15 @@ A possible workaround is to run "apt-get install sshfs" in the guest.
 `,
 		})
 		req = append(req, requirement{
-			description: "/etc/fuse.conf to contain \"user_allow_other\"",
+			description: "/etc/fuse(3).conf to contain \"user_allow_other\"",
 			script: `#!/bin/bash
 set -eux -o pipefail
-if ! timeout 30s bash -c "until grep -q ^user_allow_other /etc/fuse.conf; do sleep 3; done"; then
-	echo >&2 "/etc/fuse.conf is not updated to contain \"user_allow_other\""
+if ! timeout 30s bash -c "until grep -q ^user_allow_other /etc/fuse.conf || grep -q ^user_allow_other /etc/fuse3.conf; do sleep 3; done"; then
```
jandubois, line 110: Can be written as:

```suggestion
if ! timeout 30s bash -c "until grep -q ^user_allow_other /etc/fuse.conf /etc/fuse3.conf; do sleep 3; done"; then
```

I was concerned about what happens when `/etc/fuse3.conf` doesn't exist, but it turns out that `grep -q` will silence that error:

```console
lima-alpine:~# grep ^user_allow_other /etc/fuse.conf /etc/fuse3.conf; echo $?
/etc/fuse.conf:user_allow_other
grep: /etc/fuse3.conf: No such file or directory
2
lima-alpine:~# grep -q ^user_allow_other /etc/fuse.conf /etc/fuse3.conf; echo $?
0
lima-alpine:~# grep -q ^not_found /etc/fuse.conf; echo $?
1
```

This feels a little too magical, so I would also be fine with this:

```suggestion
if ! timeout 30s bash -c "until grep -q ^user_allow_other /etc/fuse*.conf; do sleep 3; done"; then
```

jandubois (approved):
PR works fine, thanks!

Not sure if you still want to modify the "requirements" condition or not.

I'm wondering about one more scenario: you start with a distro that only has `/etc/fuse.conf`, which will be updated to include `user_allow_other`. Then this distro is updated (maybe manually) to also have `/etc/fuse3.conf`. In that case the `requirement` will pass before `fuse3` is updated because `fuse` already contains the magic string, even though it is no longer being used.

I'm willing to say that this is obscure enough that is doesn't need special handling; wdyt?

pkg/hostagent/requirements.go
AkihiroSuda: Thanks, updated

AkihiroSuda:
Yes, setting up fuse is complicated... Hope we can switch default to 9p soon (after release of QEMU 7.1 or maybe 7.2)

PR #891, opened 2022-06-08
Check for /sbin/openrc-run instead of /sbin/openrc-init

jandubois:
Alpine doesn't use openrc-init (it uses busybox init) and has deleted it from the package in 3.16:

https://gitlab.alpinelinux.org/alpine/aports/-/commit/7a4d6d22cdb4696e93b3a0fce91e851c4db5366e


AkihiroSuda (approved):
Thanks

Issue #892, opened 2022-06-08
FreeBSD Support


heywoodlh:
### Description

FreeBSD lacks support for containerization generally but it does support QEMU. I was able to get Lima running on FreeBSD 13.0 but it doesn't work out of the box. I suspect that it wouldn't be difficult to fine-tune to get it up and running. And I feel that Lima could bring Docker support to a platform that currently lacks it which would be awesome.

I wrote-up a bit more detail on how I got Lima running on FreeBSD 13.0 here: https://the-empire.systems/freebsd-docker-2022

It's extremely slow but it works! I'm sure that I could have added something to QEMU's settings to make it perform better but I'm ignorant to the features that would help in this situation.

## Installing Lima on FreeBSD 13.0:
1. Compile+install Lima from source:

```
pkg install gmake go git
git clone https://github.com/lima-vm/lima /opt/lima
cd /opt/lima
```

For some reason, compiling Lima fails with: `pkg/sshutil/sshutil_others.go:14:19: undefined: err`

Adding the following to the `detectAESAcceleration()` function in `pkg/sshutil/sshutil_others.go` fixes it:

```
var err error
```

Once `err` is defined, Lima compiles just fine with `gmake`:

```
gmake && gmake install
```

Not sure why that fails just on FreeBSD but nothing else.

2. Compile+install QEMU 7.0 from source:

```
pkg install ninja pkgconf glib pixman

git clone --depth=1 git://git.qemu-project.org/qemu.git /opt/qemu
mkdir -p /opt/qemu/build
cd /opt/qemu/build

../configure
gmake ## add -j to speed up the compilation
gmake install
```

Not sure how necessary this actually is but the version of QEMU packaged with the FreeBSD repos is 6.2.0 so upgrading to 7.0 seemed like a good idea.

3. Set the CPU to emulate:

FreeBSD doesn't have `-cpu host` (unlike Linux and MacOS) available so I had to manually specify the processor on my machine (I was using an Intel VPS with Vultr):

```
export QEMU_SYSTEM_X86_64="qemu-system-x86_64 -cpu Cascadelake-Server"
```

## Problems needed to be solved before it's usable on FreeBSD:

1. Slowness (I am not to sure where to start there but I suspect that `-cpu host` would really help with this) 
2. Minor: Fix that `undefined: err` in `pkg/sshutil/sshutil_others.go` for FreeBSD (could be as simple as defining that variable if the OS is FreeBSD)
[...]

PR #894, opened 2022-06-11
Only use cpuType=host for OS with accel available

afbjorklund:
This should be matching the qemu getAccel() values

Specifically, no acceleration on FreeBSD right now

Issue #892

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -105,7 +105,7 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 		RISCV64: "rv64", // FIXME: what is the right choice for riscv64?
 	}
 	for arch := range cpuType {
-		if IsNativeArch(arch) {
+		if IsNativeArch(arch) && IsAccelOS() {
 			cpuType[arch] = "host"
 		}
 	}
@@ -610,6 +610,16 @@ func ResolveArch(s *string) Arch {
 		return NewArch(runtime.GOARCH)
 	}
 	return *s
+}
+
+func IsAccelOS() bool {
+	switch runtime.GOOS {
+	case "darwin", "linux", "netbsd", "windows":
+		// Accelerator
+		return true
+	}
+	// Using TCG
+	return false
 }
 
 func IsNativeArch(arch Arch) bool {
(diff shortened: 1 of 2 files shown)


pkg/limayaml/defaults.go
```diff
@@ -612,6 +612,14 @@ func ResolveArch(s *string) Arch {
 	return *s
 }
 
+func IsAccelOS() bool {
+	accelMacOS := runtime.GOOS == "darwin"
+	accelLinux := runtime.GOOS == "linux"
+	accelNetBSD := runtime.GOOS == "netbsd"
+	accelWindows := runtime.GOOS == "windows"
+	return accelMacOS || accelLinux || accelNetBSD || accelWindows
```
AkihiroSuda, line 620:
```
switch runtime.GOOS {
  case "darwin", "linux", "netbsd":
    return true
  default:
    return false
}
```

AkihiroSuda (approved):
Thanks

iamleot:
I will share some information because I hit that on NetBSD too but feel free to ignore such notes (I think that for NetBSD `-cpu host` can be added easily when NVMM accelerator is used):

I think that's a bit more complicated then that, unfortunately.

I.e. if I'm not missing something (quite probable!) only certain accelerators supports `-cpu host` for `x86` I think only KVM and HVF:

<https://git.qemu.org/?p=qemu.git;a=blob;f=target/i386/cpu.c;h=6a57ef13af8695f210b02f4adb1aeedb7aff869a;hb=HEAD#l1487>

So I guess that's Linux and macOS are probably good in that case but at least for NetBSD not and I have no idea regarding Windows.

The other possible problem is that also trying to possibly list that via `qemu -cpu help` can points out that there is a `-cpu host` but that could work only if a certain accelerator is selected (but maybe that's just a theoretical issue that is not triggered by lima :)).

afbjorklund:
Currently it just assumed that _all_ accelerators support `-cpu host`, if that is not true it should probably be fine tuned...

It also assumes that it is always available and enabled, for instance enabled in BIOS etc and kernel module loaded.

afbjorklund:
It sounds like a good idea to have the flags return `-cpu max` on platforms where `-cpu host` does not work...

Care to make a PR ?

afbjorklund:
Just checked, and there is no `-cpu host` on Windows either.

iamleot:
Thank you @afbjorklund!

PR #895, opened 2022-06-11
Remove nonexisting err from AES detection warning

afbjorklund:
Fixes build on platforms other than darwin and linux.

For instance: `GOOS=netbsd`, `GOOS=freebsd`

Issue #892


AkihiroSuda (approved):
Thanks

PR #899, opened 2022-06-12
examples: use the geo.mirror.pkgbuild.com mirror for Arch Linux

nl6720:
geo.mirror.pkgbuild.com is a GeoDNS mirror that points to sponsored mirrors.

This avoids the usage of Arch's infrastructure (mirror.pkgbuild.com) for large downloads.
See https://gitlab.archlinux.org/archlinux/infrastructure/-/merge_requests/522 for details.

### DIFF
--- a/examples/archlinux.yaml
+++ b/examples/archlinux.yaml
@@ -1,7 +1,7 @@
 # This example requires Lima v0.7.0 or later
 images:
 # Try to use yyyyMMdd.REV image if available. Note that yyyyMMdd.REV will be removed after several months.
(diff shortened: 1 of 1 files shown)


nl6720:
"test / Lints & Unit tests" currently fails because https://gitlab.freedesktop.org/ is down.

AkihiroSuda:
> Signed-off-by: nl6720 

Could you consider using real name?

nl6720:
Sorry, I'm not willing to do that. If that's an obstacle, I'll open an issue instead.

AkihiroSuda:
The change is subtle, so I'm merging this.
Thanks

PR #900, opened 2022-06-12
Not all OS with -accel has -cpu host option

afbjorklund:
Instead use -cpu max, for everything available.

Affects NetBSD (NVMM) and Windows (WHPX)

Follow-up from #894

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -106,7 +106,9 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 	}
 	for arch := range cpuType {
 		if IsNativeArch(arch) && IsAccelOS() {
-			cpuType[arch] = "host"
+			if HasHostCPU() {
+				cpuType[arch] = "host"
+			}
 		}
 	}
 	for k, v := range d.CPUType {
@@ -619,6 +621,17 @@ func IsAccelOS() bool {
 		return true
 	}
 	// Using TCG
+	return false
+}
+
+func HasHostCPU() bool {
+	switch runtime.GOOS {
+	case "darwin", "linux":
+		return true
+	case "netbsd", "windows":
+		return false
+	}
+	// Not reached
 	return false
 }
 
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -84,7 +84,9 @@ func TestFillDefault(t *testing.T) {
 		},
 	}
 	if IsAccelOS() {
-		builtin.CPUType[arch] = "host"
+		if HasHostCPU() {
+			builtin.CPUType[arch] = "host"
+		}
 	}
 
 	defaultPortForward := PortForward{


pkg/limayaml/defaults.go
```diff
@@ -106,7 +106,11 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 	}
 	for arch := range cpuType {
 		if IsNativeArch(arch) && IsAccelOS() {
-			cpuType[arch] = "host"
+			if HasHostCPU() {
+				cpuType[arch] = "host"
+			} else {
+				cpuType[arch] = "max"
+			}
```
jandubois, line 113: Maybe this should be a function, as the logic is duplicated in the test?

pkg/limayaml/defaults.go
```diff
@@ -622,6 +626,16 @@ func IsAccelOS() bool {
 	return false
 }
 
+func HasHostCPU() bool {
+	switch runtime.GOOS {
+	case "darwin", "linux":
+		return true
+	case "netbsd", "windows":
+		return false
```
jandubois, line 634: Why not let this be handled by the default?

afbjorklund: It was an effort at documentation, possibly not working

pkg/limayaml/defaults.go
afbjorklund: Yeah, the flags and functions are a bit spread out between qemu and limayaml

pkg/limayaml/defaults.go
afbjorklund: OS without any specific `-accel` (beyond "tcg"), will never call this

jandubois: > It was an effort at documentation, possibly not very successful :-)

If it is documentation, then I'm not getting it. :)

It feels like something you will only understand if you already know it, so it doesn't make sense to have a special case for it, unless you also add a comment to explain why these are treated specially.

afbjorklund: The only idea was that we should have the same OS listed in all three places.

```go
                case "darwin":
                        return "hvf"
                case "linux":
                        return "kvm"
                case "netbsd":
                        return "nvmm" // untested
                case "windows":
                        return "whpx" // untested
```

But it wasn't obvious from the context, and it was actually tested.

pkg/limayaml/defaults.go
```diff
@@ -106,7 +106,11 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 	}
 	for arch := range cpuType {
 		if IsNativeArch(arch) && IsAccelOS() {
-			cpuType[arch] = "host"
+			if HasHostCPU() {
+				cpuType[arch] = "host"
+			} else {
+				cpuType[arch] = "max"
```
AkihiroSuda, line 112: Probably better to leave empty (QEMU default)

afbjorklund: You mean remove `-cpu` argument altogether ?

pkg/limayaml/defaults.go
afbjorklund: Added comment

pkg/limayaml/defaults.go
AkihiroSuda: Yes

pkg/limayaml/defaults.go
```diff
@@ -106,7 +106,9 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 	}
 	for arch := range cpuType {
 		if IsNativeArch(arch) && IsAccelOS() {
-			cpuType[arch] = "host"
+			if HasHostCPU() {
```
AkihiroSuda, line 109: You can merge this `if` to the `if` in the previous line

pkg/limayaml/defaults_test.go
```diff
@@ -84,7 +84,9 @@ func TestFillDefault(t *testing.T) {
 		},
 	}
 	if IsAccelOS() {
-		builtin.CPUType[arch] = "host"
+		if HasHostCPU() {
```
AkihiroSuda, line 87: Same, you can merge this `if` to the `if` in the previous line

PR #903, opened 2022-06-15
Always expand inst.Message field in store.Inspect

jandubois:
There is no reason to let the user deal with template expansion when we can do this automatically after loading. This is especially useful in `limactl list` output:

```console
$ limactl ls -f '{{.Message}}' docker
------
docker context create lima --docker "host=unix:///Users/jan/.lima/docker/sock/docker.sock"
docker context use lima
docker run hello-world
------
```

See also https://github.com/lima-vm/lima/pull/838#issuecomment-1122020716

### DIFF
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 	"syscall"
+	"text/template"
 	"time"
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #907, opened 2022-06-17
Protect quotes in show-ssh output

jandubois:
Some ssh option are quoted because their arguments may contain spaces. These quotes must be passed to ssh and not removed by the shell, so this command can work:

```console
sh -c $(limactl show-ssh $INSTANCE)
```

Example output:

```console
ssh -o 'IdentityFile="/Users/jan/Library/Application Support/rancher-desktop/lima/_config/user"' ...
```

Without the extra single quotes, the `sh -c` command above fails with:

```
command-line line 0: garbage at end of line; "Support/rancher-desktop/lima/_config/user".
```


AkihiroSuda (approved):
Thanks

PR #908, opened 2022-06-18
Force sync after moving persistent directories to data volume

jandubois:
Otherwise the data configured by cloud-init/lima-init may get lost if the instance is forcibly stopped (e.g. via `limactl stop -f`).

Example: `$HOME/.ssh/authorized_hosts` is lost (size 0):

Even though cloud-init/lima-init will re-create the file on next boot, the new version will be shadowed by the `/mnt/data/home/$USER.linux/.ssh` directory that will be mounted on top and the instance will remain
inaccessible.


jandubois:
I didn't fully realize before that the `authorized_keys` for an Alpine instance could only be configured during the initial boot, as any later changes would only be applied to the hidden directory. But that is an independent issue from the data needing to be committed before the instance may get terminated non-gracefully.

I'm also not sure if this is not an acceptable limitation anyways, as you can always update the `authorized_keys` file manually and not via cloud-init/lima-init.

AkihiroSuda (approved):

PR #921, opened 2022-06-20
Allow dns.go to be consumed by external projects 5/5

Nino-K:
This change introduces additional debug lines which makes the debugging process slightly better.

### DIFF
--- a/pkg/hostagent/dns/dns.go
+++ b/pkg/hostagent/dns/dns.go
@@ -16,6 +16,21 @@ import (
 // https://github.com/lima-vm/lima/issues/380
 const truncateSize = 512
 
+var defaultFallbackIPs = []string{"8.8.8.8", "1.1.1.1"}
+
+type HandlerOptions struct {
+	IPv6            bool
+	StaticHosts     map[string]string
+	UpstreamServers []string
+}
+
+type ServerOptions struct {
+	HandlerOptions
+	Address string
+	TCPPort int
+	UDPPort int
+}
+
 type Handler struct {
 	clientConfig *dns.ClientConfig
 	clients      []*dns.Client
@@ -38,21 +53,20 @@ func (s *Server) Shutdown() {
 	}
 }
 
-func newStaticClientConfig(ips []net.IP) (*dns.ClientConfig, error) {
+func newStaticClientConfig(ips []string) (*dns.ClientConfig, error) {
 	s := ``
 	for _, ip := range ips {
-		s += fmt.Sprintf("nameserver %s\n", ip.String())
+		s += fmt.Sprintf("nameserver %s\n", ip)
 	}
 	r := strings.NewReader(s)
 	return dns.ClientConfigFromReader(r)
 }
 
-func newHandler(IPv6 bool, hosts map[string]string) (dns.Handler, error) {
+func NewHandler(opts HandlerOptions) (dns.Handler, error) {
 	cc, err := dns.ClientConfigFromFile("/etc/resolv.conf")
 	if err != nil {
-		fallbackIPs := []net.IP{net.ParseIP("8.8.8.8"), net.ParseIP("1.1.1.1")}
-		logrus.WithError(err).Warnf("failed to detect system DNS, falling back to %v", fallbackIPs)
-		cc, err = newStaticClientConfig(fallbackIPs)
+		logrus.WithError(err).Warnf("failed to detect system DNS, falling back to %v", defaultFallbackIPs)
+		cc, err = newStaticClientConfig(defaultFallbackIPs)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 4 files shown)


pkg/hostagent/dns/dns.go
```diff
@@ -57,6 +57,7 @@ func (s *Server) Shutdown() {
 }
 
 func newStaticClientConfig(ips []string) (*dns.ClientConfig, error) {
+	logrus.Infof("newStaticClientConfig creating config for the the following IPs: %v", ips)
```
AkihiroSuda, line 60: Debugf

pkg/hostagent/dns/dns.go
```diff
@@ -63,12 +64,31 @@ func newStaticClientConfig(ips []string) (*dns.ClientConfig, error) {
...
+		} else {
+			// For windows, the only fallback addresses are defaultFallbackIPs
+			// since there is no /etc/resolv.conf
+			cc, err = newStaticClientConfig(defaultFallbackIPs)
+			if err != nil {
+				return nil, err
+			}
+		}
+	} else {
+		if cc, err = newStaticClientConfig(opts.UpstreamServers); err != nil {
```
jandubois, line 91: This should log an error if the configured upstream servers are being replaced with the default fallback.

pkg/hostagent/dns/dns.go
```diff
@@ -99,14 +99,14 @@ func NewHandler(opts HandlerOptions) (dns.Handler, error) {
...
+		cnameToHost:  make(map[string]string),
+		hostToIP:     make(map[string]net.IP),
 	}
 	for host, address := range opts.StaticHosts {
 		if ip := net.ParseIP(address); ip != nil {
-			h.ip[host] = ip
+			h.hostToIP[host] = ip
 		} else {
-			h.cname[host] = limayaml.Cname(address)
+			h.cnameToHost[host] = limayaml.Cname(address)
```
jandubois, line 112: I think I remember that this one reference to `limayaml` pulls in a whole bunch of other packages into the downstream project's `go.mod`. Maybe it doesn't matter, as the unused code won't be included in the binary, but I felt that maybe moving `Cname` into the `dns` package would be cleaner.

This can be done in a followup PR, if this should be done at all. Opinions?

pkg/hostagent/dns/dns.go
```diff
@@ -140,26 +140,60 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
...
 			}
 			fallthrough
-		case dns.TypeCNAME, dns.TypeA:
+		case dns.TypeA:
+			var err error
+			var addrs []net.IP
+			if _, ok := h.hostToIP[q.Name]; ok {
+				addrs = []net.IP{h.hostToIP[q.Name]}
+			} else {
+				addrs, err = net.LookupIP(q.Name)
```
jandubois, line 153: This code will do a CNAME lookup using upstream DNS, but will not resolve names from `h.cnameToHost` here, like the old code did. Why did you change this?

Maybe it doesn't matter, as an `A` record lookup normally doesn't look up `CNAME` records either, but we do this here anyways because there is no `net.LookupA` but only `net.LookupIP`, and handling internal and external CNAMEs differently feels wrong to me.

pkg/hostagent/dns/dns.go
```diff
@@ -268,8 +270,12 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
 		}
 	}
 	if handled {
-		reply.Truncate(truncateSize)
-		_ = w.WriteMsg(&reply)
+		if h.truncate {
+			reply.Truncate(truncateSize)
+		}
+		if err := w.WriteMsg(&reply); err != nil {
+			logrus.Debugf("handleQuery failed writing DNS reply: %v", err)
```
jandubois, line 287: Please use `%w` when formatting `error` values. This applies to a lot of changes below as well.

jandubois: Or in these cases, use `logrus.WithError(err).Debugf(...)`.

pkg/hostagent/dns/dns.go
```diff
@@ -319,7 +340,13 @@ func Start(opts ServerOptions) (*Server, error) {
 		}()
```
jandubois, line 353: The code for tcp and udp is nearly identical; the only differences being the `TruncateReply` setting and the `dns.Server.Net` field. So maybe this code should be turned into a function:

```
listenAndServe("udp", opts)
listenAndServe("tcp", opts)
```

jandubois (changes requested):
I would like to see the one missing error message, and consistent use of `logrus.WithError(err)...` thoughout the changes.

And I don't understand the reasoning behind the `A` vs. `CNAME` lookup; is there more discussion about this in the downstream repo?

I'm indifferent about creating the `listenAndServe` function. It would remove some duplicated code, but might not save that much overall.

pkg/hostagent/dns/dns.go
Nino-K: Sure, that makes sense to put it under the `dns` package.

jandubois (approved):
Thanks, LGTM

pkg/hostagent/dns/dns.go
jandubois: I understand this has not been done because just moving the function over would create an import loop. Not really important anyways; we can always changes this later.

jandubois:
@AkihiroSuda not merging yet in case you want to do a review as well

AkihiroSuda (approved):
Thanks

PR #926, opened 2022-06-22
Add exe extension to the limactl binary on windows

afbjorklund:
It needs to be called `limactl.exe`, or it will silently fail exec.

Also replaced the examples symlink with yet another copy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ GO ?= go
 TAR ?= tar
 PLANTUML ?= plantuml # may also be "java -jar plantuml.jar" if installed elsewhere
 
-GOOS ?= $(GO) env GOOS
+GOOS ?= $(shell $(GO) env GOOS)
 ifeq ($(GOOS),windows)
 exe = .exe
 endif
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -88,7 +99,7 @@ uninstall:
 	@test -f "$(DEST)/bin/lima" || echo "lima not found in $(DEST) prefix"
 	rm -rf \
 		"$(DEST)/bin/lima" \
-		"$(DEST)/bin/limactl" \
+		"$(DEST)/bin/limactl$(exe)" \
 		"$(DEST)/bin/nerdctl.lima" \
 		"$(DEST)/bin/docker.lima" \
 		"$(DEST)/bin/podman.lima" \
```
AkihiroSuda, line 105: Probably we do not need this the wrapper scripts for Windows.

afbjorklund: I'm not sure it is related, but it is possible to replace both "lima" and "nerdctl.lima" with programs - if wanted.

Currently it is assumed that user has installed `sh.exe` (bash.exe) just as they have installed `ssh.exe` and so on...

Otherwise _all_ scripts would need to be rewritten, which seems unnecessary ?

Something like "Git for Windows" is enough to install, to get the requirements.

PR #931, opened 2022-06-29
whpx: Don't use -cpu max, add kernel-irqchip=off

afbjorklund:
Doesn't work now, after adding 017a56ca10f0a867c73d2b06547302787ffddabc

Also needs to avoid falling back on tcg

----

Tested on Windows 10 (Pro), x64 (Intel)

Using Alpine Linux, and Alpine-Lima


AkihiroSuda (approved):
Thanks

PR #937, opened 2022-07-02
podman: Use different connections for different instances

afbjorklund:
Allows having more than one instance using podman.

Similar to change applied for the docker context.

Compare:

* https://github.com/lima-vm/lima/pull/935

Trying to keep docker and podman reasonably similar...


jandubois (approved):
Thanks!

Issue #936, opened 2022-07-01
The MinGW and MSYS2 terminals are not supported on Windows


afbjorklund:
### Description

The TTY works on Posix and it works on Windows command prompt (DOS),
but it does not work when using a Unix terminal such as [mintty](https://mintty.github.io/) on Windows.

* MinGW (like in Git for Windows)

* MSYS2

The detection uses `IsCygwinTerminal` rather than the usual `IsTerminal`,
and the `survey` module is hardcoded to use either/or (based on the GOOS)

* https://github.com/mattn/go-isatty

* https://github.com/AlecAivazis/survey

----

This breaks the template selection ("Incorrect function")

But it also breaks terminal handling when running shell

Other side effects was JSON logging to the terminal too

And different output when it comes to the downloader

afbjorklund:
The code in logrus also failed to detect these terminals, overrriding with ForceColors

````
vendor/github.com/sirupsen/logrus/text_formatter.go:	isColored := f.ForceColors || (f.isTerminal && (runtime.GOOS != "windows"))
```

afbjorklund:
The workaround for the template selection, is either to use `template://` or `winpty.exe`:

```
winpty limactl start
```

Cactusinhand:
> The workaround for the template selection, is either to use template:// or winpty.exe:

@afbjorklund 
Sorry, what does `template://` mean here?

PR #938, opened 2022-07-02
Add partial support for cygwin terminal on windows

afbjorklund:
This enables terminal in shell and colored logs/download.
The interactive template selection is not implemented yet.

Issue #936


AkihiroSuda (approved):
Thanks

PR #943, opened 2022-07-05
Add artifacts for windows (x86_64) target

afbjorklund:
Not building these by default, until it is more stable.

The zip is for windows-without-msys, i.e. no tar...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ DEST := $(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's:///*:/:g; s://*$$::')
 
 GO ?= go
 TAR ?= tar
+ZIP ?= zip
 PLANTUML ?= plantuml # may also be "java -jar plantuml.jar" if installed elsewhere
 
 GOOS ?= $(shell $(GO) env GOOS)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -133,6 +134,13 @@ artifacts-linux:
 	GOOS=linux GOARCH=arm64 CC=aarch64-linux-gnu-gcc make clean binaries
 	$(TAR) -C _output/ -czvf _artifacts/lima-$(VERSION_TRIMMED)-Linux-aarch64.tar.gz ./
 
+.PHONY: artifacts-windows
+artifacts-windows:
+	mkdir -p _artifacts
+	GOOS=windows GOARCH=amd64 CC=x86_64-w64-mingw32-gcc make clean binaries
+	$(TAR) -C _output/ -czvf _artifacts/lima-$(VERSION_TRIMMED)-Windows-x86_64.tar.gz ./
```
AkihiroSuda, line 141: Probably no need to create tar.gz for Windows

afbjorklund: The tarballs are the default, on all platforms (same everywhere). They work fine on Windows too, when having a development environment installed (e.g. Git for Windows)

The .zip was just a fallback, for a user setup with minimal tools

AkihiroSuda (approved):

PR #944, opened 2022-07-05
Add lima wrapper for windows without bash

afbjorklund:
Advanced users would like to install a "msys" environment
with bash, but for casual lima usage this should be enough.

lima-dos

Apparently `C:\Windows\System32\OpenSSH\ssh.exe` is included.
(User does have to add the `C:\Program Files\qemu` [setup](https://qemu.weilnetz.de/), though)

----

PS. I added the "ver" (windows version) to match the "uname" (unix name)
```console
$ uname
Darwin

$ lima uname
Linux
```
Running under MINGW64 or MSYS does have both "uname" and "bash"...


Makefile
```diff
@@ -26,6 +27,7 @@ exe: _output/bin/limactl$(exe)
 .PHONY: binaries
 binaries: clean \
 	_output/bin/lima \
```
AkihiroSuda, line 29: I guess this line can be removed now

afbjorklund: I want to have both of them, the shell variants are preferred - the other is just a fallback

I _don't_ want to spend time porting anything like nerdctl.lima to it, but made an exception for "lima"

AkihiroSuda (approved):

PR #948, opened 2022-07-08
add usb hub to each vm

aluedeke:
I am working on a project where we need to forward devices into the qemu guest OS. ARM guests already get a USB hub, unfortunately, x86 doesn't. It would be great to add to each x86 vm an xhci usb hub as well. (All other usb related work can then happen through qmp/hmp (at some point it might be cool to have an easy way to find the path of the colima qmp socket, but for now we can just build the path on our own).


AkihiroSuda (approved):

djhardrich:
@aluedeke Do you have any information on how to use usb forwarding in a qemu guest in lima (qemu 7.0)? Apologies to all if this isn't the best place to ask, and thanks!

aluedeke:
@djhardrich checkout https://gist.github.com/aluedeke/37a74364700296d4540d51d937f1b3d9 and let me know if that helps

PR #951, opened 2022-07-12
Add no arch option

brorbw:
Adding a no arch option allow forcing the `tcg` accelerator. This
allows lima to run on AMD based hackintosh machines and other macos
machines that does not support hvf.


brorbw:
This is essentially like adding `cpu.x86_64.qemu64` to the configuration. Thus closing this

PR #952, opened 2022-07-12
Add option to force tcg accelerator

brorbw:
More a less a different implementation of
https://github.com/lima-vm/lima/pull/951

`useAccel` allow forcing the `tcg` accelerator. This
allows lima to run on AMD based hackintosh machines and other macos
machines that does not support hvf.

Right now it is only possible with `cpu.x86_64:qemu64` which seems to bee a little convoluted

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -347,6 +347,10 @@ hostResolver:
 # - 1.1.1.1
 # - 1.0.0.1
 
+# Use Accel: "true", "false". Uses hvf, kvm, nvmm, whpx if true or tcg if false.
+# Builtin default: true
+useAccel: null
+
 # ===================================================================== #
 # GLOBAL DEFAULTS AND OVERRIDES
 # ===================================================================== #
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -75,6 +75,16 @@ func MACAddress(uniqueID string) string {
 // - DNS are picked from the highest priority where DNS is not empty.
 // - CACertificates Files and Certs are uniquely appended
 func FillDefault(y, d, o *LimaYAML, filePath string) {
+	if y.UseAccel == nil {
+		y.UseAccel = d.UseAccel
+	}
+	if o.UseAccel != nil {
+		y.UseAccel = o.UseAccel
+	}
+	if y.UseAccel == nil {
+		y.UseAccel = pointer.Bool(true)
+	}
+
 	if y.Arch == nil {
 		y.Arch = d.Arch
 	}
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -82,6 +82,7 @@ func TestFillDefault(t *testing.T) {
 		CACertificates: CACertificates{
 			RemoveDefaults: pointer.Bool(false),
 		},
+		UseAccel: pointer.Bool(true),
 	}
 	if IsAccelOS() {
 		if HasHostCPU() {
@@ -305,6 +306,7 @@ func TestFillDefault(t *testing.T) {
 				"-----BEGIN CERTIFICATE-----\nYOUR-ORGS-TRUSTED-CA-CERT\n-----END CERTIFICATE-----\n",
 			},
 		},
+		UseAccel: pointer.Bool(false),
 	}
 
 	expect = d
@@ -463,6 +465,7 @@ func TestFillDefault(t *testing.T) {
 		CACertificates: CACertificates{
 			RemoveDefaults: pointer.Bool(true),
 		},
+		UseAccel: pointer.Bool(false),
 	}
 
 	y = filledDefaults
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -29,6 +29,7 @@ type LimaYAML struct {
 	DNS               []net.IP          `yaml:"dns,omitempty" json:"dns,omitempty"`
 	HostResolver      HostResolver      `yaml:"hostResolver,omitempty" json:"hostResolver,omitempty"`
 	UseHostResolver   *bool             `yaml:"useHostResolver,omitempty" json:"useHostResolver,omitempty"` // DEPRECATED, use `HostResolver.Enabled` instead
+	UseAccel          *bool             `yaml:"useAccel,omitempty" json:"useAccel,omitempty"`
 	PropagateProxyEnv *bool             `yaml:"propagateProxyEnv,omitempty" json:"propagateProxyEnv,omitempty"`
 	CACertificates    CACertificates    `yaml:"caCerts,omitempty" json:"caCerts,omitempty"`
 }
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -350,7 +350,7 @@ func Cmdline(cfg Config) (string, []string, error) {
 	}
 
 	// Architecture
-	accel := getAccel(*y.Arch)
+	accel := getAccel(*y.Arch, *y.UseAccel)
 	if !strings.Contains(string(features.AccelHelp), accel) {
 		return "", nil, fmt.Errorf("accelerator %q is not supported by %s", accel, exe)
 	}
@@ -602,8 +602,8 @@ func getExe(arch limayaml.Arch) (string, []string, error) {
 	return exe, args, nil
 }
 
-func getAccel(arch limayaml.Arch) string {
-	if limayaml.IsNativeArch(arch) {
+func getAccel(arch limayaml.Arch, useAccel bool) string {
+	if limayaml.IsNativeArch(arch) && useAccel {
 		switch runtime.GOOS {
 		case "darwin":
 			return "hvf"


AkihiroSuda:
> This allows lima to run on AMD based hackintosh machines and other macos
machines that does not support hvf.

Sorry, these are out of the scope of the project

oxycoder:
@brorbw I've tested your PR and it doesn't seem to work for me on ryzen hackintosh.

brorbw:
@oxycoder which part? Using `tcg` or the PR?

oxycoder:
@brorbw enable `tcg` by using flag useAccel doesn't work for me.
VM boot up with weird error

```[2J[01;01H[=3h[2J[01;01HBdsDxe: loading Boot0008 "ubuntu" from HD(15,GPT,BFE2241A-474C-471C-BD73-6A38E9731D0D,0x2800,0x35000)/\EFI\ubuntu\shimx64.efi
BdsDxe: starting Boot0008 "ubuntu" from HD(15,GPT,BFE2241A-474C-471C-BD73-6A38E9731D0D,0x2800,0x35000)/\EFI\ubuntu\shimx64.efi
error: cannot allocate kernel buffer.

error: you need to load the kernel first.


Press any key to continue...


  Failed to boot both default and fallback entries.


` ``

lima.yml
```
# ===================================================================== #
# BASIC CONFIGURATION
# ===================================================================== #

# Default values in this YAML file are specified by `null` instead of Lima's "builtin default" values,
# so they can be overridden by the $LIMA_HOME/_config/default.yaml mechanism documented at the end of this file.

# Arch: "default", "x86_64", "aarch64".
arch: x86_64

# OpenStack-compatible disk image.
# 🟢 Builtin default: null (must be specified)
# 🔵 This file: Ubuntu 22.04 Jammy Jellyfish images
images:
- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20220420/ubuntu-22.04-server-cloudimg-amd64.img"
  arch: "x86_64"
  digest: "sha256:de5e632e17b8965f2baf4ea6d2b824788e154d9a65df4fd419ec4019898e15cd"
- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20220420/ubuntu-22.04-server-cloudimg-arm64.img"
  arch: "aarch64"
  digest: "sha256:66224c7fed99ff5a5539eda406c87bbfefe8af6ff6b47d92df3187832b5b5d4f"
- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"
  arch: "x86_64"
- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img"
  arch: "aarch64"

# CPUs: if you see performance issues, try limiting cpus to 1.
# 🟢 Builtin default: 4
cpus: 8

# Memory size
# 🟢 Builtin default: "4GiB"
memory: 16

# Disk size
# 🟢 Builtin default: "100GiB"
disk: null

# 🟢 Builtin default: null (Mount nothing)
# 🔵 This file: Mount the home as read-only, /tmp/lima as writable
mounts:
- location: "~"
  # Configure the mountPoint inside the guest.
  # 🟢 Builtin default: value of location
  mountPoint: null
  # 🟢 Builtin default: false
  writable: null
  sshfs:
    # Enabling the SSHFS cache will increase performance of the mounted filesystem, at
    # the cost of potentially not reflecting changes made on the host in a timely manner.
    # Warning: It looks like PHP filesystem access does not work correctly when
    # the cache is disabled.
    # 🟢 Builtin default: true
    cache: null
    # contents of the symlink. As a result, symlinked files & folders on the Host
    # system will look and feel like regular files directories in the Guest OS.
    # 🟢 Builtin default: false
    followSymlinks: null
    # SFTP driver, "builtin" or "openssh-sftp-server". "openssh-sftp-server" is recommended.
    # 🟢 Builtin default: "openssh-sftp-server" if OpenSSH SFTP Server binary is found, otherwise "builtin"
    sftpDriver: null
  9p:
    # Supported security models are "passthrough", "mapped-xattr", "mapped-file" and "none".
    # 🟢 Builtin default: "mapped-xattr"
    securityModel: null
    # Select 9P protocol version. Valid options are: "9p2000" (legacy), "9p2000.u", "9p2000.L".
    # 🟢 Builtin default: "9p2000.L"
    protocolVersion: null
    # The number of bytes to use for 9p packet payload, where 4KiB is the absolute minimum.
    # 🟢 Builtin default: "128KiB"
    msize: null
    # Specifies a caching policy. Valid options are: "none", "loose", "fscache" and "mmap".
    # Try choosing "mmap" or "none" if you see a stability issue with the default "fscache".
    # See https://www.kernel.org/doc/Documentation/filesystems/9p.txt
    # 🟢 Builtin default: "fscache" for non-writable mounts, "mmap" for writable mounts
    cache: null
- location: "/tmp/lima"
  # 🟢 Builtin default: false
  # 🔵 This file: true (only for "/tmp/lima")
  writable: true

# Mount type for above mounts, such as "reverse-sshfs" (from sshocker) or "9p" (EXPERIMENTAL, from QEMU’s virtio-9p-pci, aka virtfs)
# 🟢 Builtin default: "reverse-sshfs"
mountType: null

ssh:
  # A localhost port of the host. Forwarded to port 22 of the guest.
  # 🟢 Builtin default: 0 (automatically assigned to a free port)
  # NOTE: when the instance name is "default", the builtin default value is set to
  # 60022 for backward compatibility.
  localPort: 0
  # Load ~/.ssh/*.pub in addition to $LIMA_HOME/_config/user.pub .
  # This option is useful when you want to use other SSH-based
  # applications such as rsync with the Lima instance.
  # If you have an insecure key under ~/.ssh, do not use this option.
  # 🟢 Builtin default: true
  loadDotSSHPubKeys: null
  # Forward ssh agent into the instance.
  # 🟢 Builtin default: false
  forwardAgent: true
  # Forward X11 into the instance
  # 🟢 Builtin default: false
  forwardX11: null
  # Trust forwarded X11 clients
  # 🟢 Builtin default: false
  forwardX11Trusted: null

# ===================================================================== #
# ADVANCED CONFIGURATION
# ===================================================================== #

caCerts:
  # If set to `true`, this will remove all the default trusted CA certificates that
  # are normally shipped with the OS.
  # 🟢 Builtin default: false
  removeDefaults: null

  # A list of trusted CA certificate files. The files will be read and passed to cloud-init.
  files:
  # - examples/hello.crt

  # A list of trusted CA certificates. These are directly passed to cloud-init.
  certs:
  # - |
  #   -----BEGIN CERTIFICATE-----
  #   -----END CERTIFICATE-----
  # - |
  #   -----BEGIN CERTIFICATE-----
  #   -----END CERTIFICATE-----

containerd:
  # Enable system-wide (aka rootful)  containerd and its dependencies (BuildKit, Stargz Snapshotter)
  # 🟢 Builtin default: false
  system: null
  # Enable user-scoped (aka rootless) containerd and its dependencies
  # 🟢 Builtin default: true
  user: null
#  # Override containerd archive
#  # 🟢 Builtin default: hard-coded URL with hard-coded digest (see the output of `limactl info | jq .defaultTemplate.containerd.archives`)
#  archives:
#  - location: "~/Downloads/nerdctl-full-X.Y.Z-linux-amd64.tar.gz"
#    arch: "x86_64"
#    digest: "sha256:..."

# Provisioning scripts need to be idempotent because they might be called
# multiple times, e.g. when the host VM is being restarted.
# 🟢 Builtin default: null
# provision:
# # `system` is executed with the root privilege
# - mode: system
#   script: |
#     #!/bin/bash
#     set -eux -o pipefail
#     export DEBIAN_FRONTEND=noninteractive
#     apt-get install -y vim
# # `user` is executed without the root privilege
# - mode: user
#   script: |
#     #!/bin/bash
#     set -eux -o pipefail
#     cat <<EOF > ~/.vimrc
#     set number
#     EOF

# Probe scripts to check readiness.
# 🟢 Builtin default: null
# probes:
# # Only `readiness` probes are supported right now.
# - mode: readiness
#   description: vim to be installed
#   script: |
#      #!/bin/bash
#      set -eux -o pipefail
#      if ! timeout 30s bash -c "until command -v vim; do sleep 3; done"; then
#        echo >&2 "vim is not installed yet"
#        exit 1
#      fi
#   hint: |
#     vim was not installed in the guest. Make sure the package system is working correctly.
#     Also see "/var/log/cloud-init-output.log" in the guest.

# ===================================================================== #
# FURTHER ADVANCED CONFIGURATION
# ===================================================================== #

# Specify desired QEMU CPU type for each arch.
cpuType:
  # 🟢 Builtin default: "cortex-a72" (or "host" when running on aarch64 host)
  aarch64: null
  # 🟢 Builtin default: "qemu64" (or "host" when running on x86_64 host)
  x86_64: qemu64

firmware:
  # Use legacy BIOS instead of UEFI. Ignored for aarch64.
  # 🟢 Builtin default: false
  legacyBIOS: false

video:
  # QEMU display, e.g., "none", "cocoa", "sdl", "gtk".
  # As of QEMU v6.2, enabling this is known to have negative impact
  # 🟢 Builtin default: "none"
  display: null

# The instance can get routable IP addresses from the vmnet framework using
# https://github.com/lima-vm/vde_vmnet.
# 🟢 Builtin default: null
networks:
# Lima can manage daemons for networks defined in $LIMA_HOME/_config/networks.yaml
# automatically. Both vde_switch and vde_vmnet binaries must be installed into
# secure locations only alterable by the "root" user.
# - lima: shared
#   # MAC address of the instance; lima will pick one based on the instance name,
#   # so DHCP assigned ip addresses should remain constant over instance restarts.
#   macAddress: ""
#   # Interface name, defaults to "lima0", "lima1", etc.
#   interface: ""
#
# Lima can also connect to "unmanaged" vde networks addressed by "vnl". This
# means that the daemons will not be controlled by Lima, but must be started
# before the instance.  The interface type (host, shared, or bridged) is
# configured in vde_vmnet and not in lima.
# vnl (virtual network locator) points to the vde_switch socket directory,
# optionally with vde:// prefix
# - vnl: "vde:///var/run/vde.ctl"
#   # VDE Switch port number (not TCP/UDP port number). Set to 65535 for PTP mode.
#   # Builtin default: 0
#   switchPort: 0
#   # MAC address of the instance; lima will pick one based on the instance name,
#   # so DHCP assigned ip addresses should remain constant over instance restarts.
#   macAddress: ""
#   # Interface name, defaults to "lima0", "lima1", etc.
#   interface: ""

# Port forwarding rules. Forwarding between ports 22 and ssh.localPort cannot be overridden.
# Rules are checked sequentially until the first one matches.
# portForwards:
# - guestPort: 443
#   hostIP: "0.0.0.0" # overrides the default value "127.0.0.1"; allows privileged port forwarding
# # default: hostPort: 443 (same as guestPort)
# # default: guestIP: "127.0.0.1" (also matches bind addresses "0.0.0.0", "::", and "::1")
# # default: proto: "tcp" (only valid value right now)
#
# - guestPortRange: [4000, 4999]
#   hostIP:  "0.0.0.0" # overrides the default value "127.0.0.1"
# # default: hostPortRange: [4000, 4999] (must specify same number of ports as guestPortRange)
#
# - guestPort: 80
#   hostPort: 8080 # overrides the default value 80
#
# - guestIP: "127.0.0.2" # overrides the default value "127.0.0.1"
#   hostIP: "127.0.0.2" # overrides the default value "127.0.0.1"
# # default: guestPortRange: [1, 65535]
#
# - guestPort: 8888
#   ignore: true (don't forward this port)
#
# - guestPort: 7443
#   guestIP: "0.0.0.0"       # Will match *any* interface
#   guestIPMustBeZero: true  # Restrict matching to 0.0.0.0 binds only
#   hostIP: "0.0.0.0"        # Forwards to 0.0.0.0, exposing it externally
#
# - guestSocket: "/run/user/{{.UID}}/my.sock"
#   hostSocket: mysocket
# # default: reverse: false
# # "guestSocket" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
# # "hostSocket" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
# # "reverse" can only be used for unix sockets right now, not for tcp sockets.
# # Put sockets into "{{.Dir}}/sock" to avoid collision with Lima internal sockets!
# # Sockets can also be forwarded to ports and vice versa, but not to/from a range of ports.
# # Forwarding requires the lima user to have rw access to the "guestsocket",
# # and the local user rwx access to the directory of the "hostsocket".
#
# # Lima internally appends this fallback rule at the end:
# - guestIP: "127.0.0.1"
#   guestPortRange: [1, 65535]
#   hostIP: "127.0.0.1"
#   hostPortRange: [1, 65535]
# # Any port still not matched by a rule will not be forwarded (ignored)

# Message. Information to be shown to the user, given as a Go template for the instance.
# The same template variables as for listing instances can be used, for example {{.Dir}}.
# You can view the complete list of variables using `limactl list --list-fields` command.
# It also includes {{.HostOS}} and {{.HostArch}} vars, for the runtime GOOS and GOARCH.
# 🟢 Builtin default: null
# message: |
#   This will be shown to the user.

# Extra environment variables that will be loaded into the VM at start up.
# These variables are consumed by internal init scripts, and also added
# to /etc/environment.
# If you set any of "ftp_proxy", "http_proxy", "https_proxy", or "no_proxy", then
# Lima will automatically set an uppercase variant to the same value as well.
# 🟢 Builtin default: null
# env:
#   KEY: value

# Lima will override the proxy environment variables with values from the current process
# environment (the environment in effect when you run `limactl start`). It will automatically
# replace the strings "localhost" and "127.0.0.1" with the host gateway address from inside
# the VM, so it stays routable. Use of the process environment can be disabled by setting
# propagateProxyEnv to false.
# 🟢 Builtn default: true
propagateProxyEnv: null

# The host agent implements a DNS server that looks up host names on the host
# using the local system resolver. This means changing VPN and network settings
# are reflected automatically into the guest, including conditional forward,
# and mDNS lookup. By default only IPv4 addresses will be returned. IPv6 addresses
# can only work when using a vmnet network interface and the host has working
# IPv6 configured as well.
hostResolver:
  # 🟢 Builtin default: true
  enabled: null
  # 🟢 Builtin default: false
  ipv6: null
  # Static names can be defined here as an alternative to adding them to the hosts /etc/hosts.
  # Values can be either other hostnames, or IP addresses. The host.lima.internal name is
  # 🟢 Builtin default: null
  hosts:
    # guest.name: 127.1.1.1
    # host.name: host.lima.internal

# If useHostResolver is false, then the following rules apply for configuring dns:
# Explicitly set DNS addresses for qemu user-mode networking. By default qemu picks *one*
# nameserver from the host config and forwards all queries to this server. On macOS
# Lima adds the nameservers configured for the first host interface in service order,
# that has an IPv4 address, to the list. In case this still doesn't work (e.g. VPN
# setups), the servers can be specified here explicitly. If nameservers are specified
# here, then the configuration from network preferences will be ignored.
# 🟢 Builtin default: null
# dns:
# - 1.1.1.1
# - 1.0.0.1

# ===================================================================== #
# GLOBAL DEFAULTS AND OVERRIDES
# ===================================================================== #
# Use Accel: "true", "false". Uses hvf, kvm, nvmm, whpx if true or tcg if false.
# Builtin default: true
useAccel: false

# The builtin defaults can be changed globally by creating a $LIMA_HOME/_config/default.yaml
# file. It will be used by ALL instances under the same $LIMA_HOME, and it
# will be applied on each `limactl start`, so can affect instance restarts.

# A similar mechanism is $LIMA_HOME/_config/override.yaml, which will take
# precedence even over the settings in an instances lima.yaml file.
# It too applies to ALL instances under the same $LIMA_HOME, and is applied
# on each restart. It can be used to globally override settings, e.g. make
# the mount of the home directory writable.

# On each instance start the config settings are determined: If a value is
# not set in `lima.yaml`, then the `default.yaml` is used. If that file
# doesn't exist, or the value is not defined in the file, then the builtin
# default is used. If `override.yaml` exists and defines the value, then
# it overrides whatever has been chosen so far.

# For slices (e.g. `mounts`, `provision`) and maps (`env`) the entries are
# combined instead of replacing each other. Slices are produced from override
# settings, followed by lima.yaml, followed by defaults.yaml (but NOT from
# builtin defaults). Maps are produced starting with defaults.yaml values,
# overwriting with lima.yaml ones, overwriting with override.yaml.

# Exceptions:
# - `dns` will use the list from the highest priority file; they are not
#   combined. If override.yaml defines a list of `dns` entries, then the
#   settings in default.yaml and lima.yaml are ignored.
#
# - `mounts` will update the `writable` setting when 2 entries have the
#   same `location` value. For this reason they are processed in the opposite
#   order: starting with default, followed by lima, and then override.
#
# -`networks` will replace lower priority entries with the same `interface`
#   name with higher priority definitions. This does not apply if the
#  `interface` field is empty. `networks` are therefore also processed
#  in lowest to highest priority order.

# ===================================================================== #
# END OF TEMPLATE
# ===================================================================== #

```

Seem like the images provided by default doesn't alloc correct memory

brorbw:
@oxycoder does it work if you set `cpu.x86_64:qemu64` instead?

PR #953, opened 2022-07-12
Use forward slash for iso9660.Create on windows

afbjorklund:
The "go-embed" library has hardcoded forward slashes (`/`)
and doesn't like when mixing them with backslashes (`\`).


AkihiroSuda (approved):
Thanks

PR #954, opened 2022-07-12
Expand tilde using the current home path

afbjorklund:
For instance it might be `C:\Users\anders` on the host,
but translated to `/c/Users/anders` path in the guest.

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -134,11 +134,11 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 	}
 	for i, f := range y.Mounts {
 		tag := fmt.Sprintf("mount%d", i)
-		location, err := localpathutil.Expand(f.Location)
+		location, err := localpathutil.ExpandHome(f.Location, hostHome)
 		if err != nil {
 			return err
 		}
-		mountPoint, err := localpathutil.Expand(f.MountPoint)
+		mountPoint, err := localpathutil.ExpandHome(f.MountPoint, u.HomeDir)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


afbjorklund:
This would need some revisiting, if we want the default mounts to work:

```
mounts:
- location: "~"
- location: "/tmp/lima"
  writable: true
```

This needs to be expanded to the (converted) host path, not the guest path.

Should `/home/anders` be mounted under `/home/anders.linux`, it breaks...

----

The refactoring points are valid, it is not really "localpath" util anymore.

And it seems like a fair point, to not override the real HomeDir of the user.

jandubois:
> This would need some revisiting, if we want the default mounts to work:

I'm not asking for the mounts to change. I was just recommending to create a separate function to return the mount point inside the VM of the home directory on the host. This is a different thing than the user's home directory inside the VM and has no relation with what `osutil.LimaUser()` should return as the `HomeDir`.

I'm arguing that it was a bug that `HomeDir` was set to the path of the home directory on the host before, and you just extended it to the Windows case as well. I'm arguing that we should fix the bug and then use `u.HomeDir` in all the places where we currently do `fmt.Sprintf("/home/%s.linux", u.Username` or similar.

And then create a separate function to return the `LIMA_CIDATA_HOSTHOME_MOUNTPOINT` and use that in all the places that need that value.

jandubois:
Note that the user can specify the mountpoint now, so the simplistic translation is not necessarily correct:

```yaml
mounts:
- location: "~"
  mountPoint: /var/hosthome
```

So the function will need to check the definitions from `lima.yaml` as well.

afbjorklund:
Sounds good to return the meaning of home dir (to match that of the user), and adjust the code.

PR #955, opened 2022-07-12
Use short path for LIMA_HOME on windows

afbjorklund:
Other commands such as `qemu-system-x86_64.exe` don't work with
`C:\Users\AndersBjörklund`, they need to use `C:\Users\ANDERS~1`

In theory they _could_ have supported using UNC paths such as
`\\?\C:\Users\AndersBjörklund`, but in practice they do not...

https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getshortpathnamew

> A long file name is considered to be any file name that exceeds the short MS-DOS (also called 8.3) style naming convention.

### GOOS=linux (host)

```
=== RUN   TestShortNameShort
    shortname_test.go:20: /tmp/foo => /tmp/foo
--- PASS: TestShortNameShort (0.00s)
=== RUN   TestShortNameLong
    shortname_test.go:31: /tmp/baaaaaaaaaar => /tmp/baaaaaaaaaar
--- PASS: TestShortNameLong (0.00s)
PASS
```

### GOOS=windows (wine)

```
=== RUN   TestShortNameShort
    shortname_test.go:20: C:\users\anders\Temp\foo => C:\users\anders\Temp\foo
--- PASS: TestShortNameShort (0.00s)
=== RUN   TestShortNameLong
    shortname_test.go:31: C:\users\anders\Temp\baaaaaaaaaar => C:\users\anders\Temp\BAAA~MWI
--- PASS: TestShortNameLong (0.00s)
PASS
```

### DIFF
--- a/pkg/store/dirnames/dirnames.go
+++ b/pkg/store/dirnames/dirnames.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 
 	"github.com/lima-vm/lima/pkg/store/filenames"
 )
@@ -17,6 +18,10 @@ const DotLima = ".lima"
 // NOTE: We do not use `~/Library/Application Support/Lima` on macOS.
 // We use `~/.lima` so that we can have enough space for the length of the socket path,
 // which can be only 104 characters on macOS.
+//
+// NOTE: There are some issues when using "long names" on Windows.
+// We use "short names" here, so that it works with user names containing unicode etc.
+// They normally have 8+3 characters, with suffix.
 func LimaDir() (string, error) {
 	dir := os.Getenv("LIMA_HOME")
 	if dir == "" {
@@ -24,9 +29,20 @@ func LimaDir() (string, error) {
 		if err != nil {
 			return "", err
 		}
+		// on windows, 8.3 paths are needed by some tools like QEMU
+		if runtime.GOOS == "windows" {
+			homeDir, err = ShortPathName(homeDir)
+			if err != nil {
+				return "", err
+			}
+		}
 		dir = filepath.Join(homeDir, DotLima)
 	}
 	if _, err := os.Stat(dir); errors.Is(err, os.ErrNotExist) {
+		return dir, nil
+	}
+	// on windows, EvalSymlinks translates short paths back to long again
+	if runtime.GOOS == "windows" {
 		return dir, nil
 	}
 	realdir, err := filepath.EvalSymlinks(dir)
--- a/pkg/store/dirnames/shortname_windows.go
+++ b/pkg/store/dirnames/shortname_windows.go
@@ -1,0 +1,23 @@
+package dirnames
+
+import (
+	"syscall"
+)
+
+// ShortPathName return the short path name, when given a long path.
+func ShortPathName(path string) (string, error) {
+	p := syscall.StringToUTF16(path)
+	b := p // GetShortPathName says we can reuse buffer
+	n, err := syscall.GetShortPathName(&p[0], &b[0], uint32(len(b)))
+	if err != nil {
+		return "", err
+	}
+	if n > uint32(len(b)) {
+		b = make([]uint16, n)
+		n, err = syscall.GetShortPathName(&p[0], &b[0], uint32(len(b)))
+		if err != nil {
+			return "", err
+		}
+	}
+	return syscall.UTF16ToString(b), nil
+}
--- a/pkg/store/dirnames/shortname_others.go
+++ b/pkg/store/dirnames/shortname_others.go
@@ -1,0 +1,9 @@
+//go:build !windows
+// +build !windows
+
+package dirnames
+
+// ShortPathName just returns the provided path.
+func ShortPathName(path string) (string, error) {
+	return path, nil
+}
--- a/pkg/store/dirnames/shortname_test.go
+++ b/pkg/store/dirnames/shortname_test.go
@@ -1,0 +1,37 @@
+package dirnames
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+// Note: can't use t.TempDir(), because it is _always_ long... <sigh>
+//       instead use os.TempDir(), something like `C:\users\anders\Temp`
+
+func TestShortPathNameShort(t *testing.T) {
+	d := os.TempDir()
+	l := filepath.Join(d, "foo")
+	err := os.Mkdir(l, 0755)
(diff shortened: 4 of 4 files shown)


pkg/store/dirnames/dirnames.go
```diff
@@ -24,11 +25,22 @@ func LimaDir() (string, error) {
 		if err != nil {
 			return "", err
 		}
+		// on windows, DOS paths are needed by some tools like QEMU
```
AkihiroSuda, line 28: Could you explain DOS paths in the comment?

pkg/store/dirnames/shortname_windows.go
```diff
@@ -0,0 +1,22 @@
+package dirnames
+
+import (
+	"syscall"
+)
+
+func ShortName(path string) (string, error) {
```
AkihiroSuda, line 7: Could you add godoc?

AkihiroSuda:
> C:\Users\AndersBjörklund, they need to use C:\Users\ANDERS~1

What if your home directory is like `C:\Users\Björklund` or `C:\Users\漢字` ?

jandubois:
> What if your home directory is like `C:\Users\Björklund` or `C:\Users\漢字` ?

`GetShortPathName` will automatically translate them to 8.3 format (if the file exists; there is no way to generate a short name before creating the file, as the name depends on what other files already exist in the directory). `dir /x` shows both the short and long names:

```console
C:\Users\Jan\Names>dir /x
 Volume in drive C has no label.
 Volume Serial Number is 5C37-CD74

 Directory of C:\Users\Jan\Names

07/12/2022  10:10 PM    <DIR>                       .
07/12/2022  10:10 PM    <DIR>                       ..
07/12/2022  10:08 PM    <DIR>          BJRKLU~1     Björklund
07/12/2022  10:08 PM    <DIR>          ED60~1       漢字
               0 File(s)              0 bytes
               4 Dir(s)  41,248,669,696 bytes free

C:\Users\Jan\Names>perl -MWin32 -E "say Win32::GetShortPathName('Björklund')"
BJRKLU~1
```

jandubois:
> Other commands such as `qemu-system-x86_64.exe` don't work with `C:\Users\AndersBjörklund`, they need to use `C:\Users\ANDERS~1`

BTW, are you sure the problem is the long pathname and not some code page problem? It doesn't really matter though, as the short name algorithm always returns ASCII compatible strings, so you don't have to worry about encodings (unless people enable `HKLM\System\CurrentControlSet\Control\FileSystem\NtfsAllowExtendedCharacterIn8dot3Name` in which case they can only blame themselves).

afbjorklund:
Actually I think the problem is related to how Go handles paths in the `exec.Command`, when running from a prompt it seems to work. Unfortunately each program is required to handle parsing in Windows, and it seems to fail with the long names.

pkg/store/dirnames/dirnames.go
afbjorklund: I think the technical name would be 8.3 names

jandubois: `8dot3name`, e.g.

```console
C:\Users\Jan\Names>fsutil 8dot3name query
The registry state is: 2 (Per volume setting - the default).
```

afbjorklund: I hadn't expected to see them again, this century :-)

afbjorklund:
There is some confusion between $HOME and os.UserHomeDir, but the plan is to go with the windows default  %HOMEPATH%

jandubois:
> There is some confusion between $HOME and os.UserHomeDir, but the plan is to go with the windows default %HOMEPATH%

Note that `%HOMEPATH%` does not include the drive; you would need `%HOMEDRIVE%%HOMEPATH%`. However, I think you want to use `%USERPROFILE%` here, which is where application settings are stored. Homedrive/path often point to the same location, but don't have to.

On second thoughts, `%LOCALAPPDATA%\Lima` is where LIMA_HOME should live by default. Do **not** use `%APPDATA%`, which is the location of the roaming profile; it will be copied back and forth to the domain controller on login/logout, which will take forever to copy the full VM; it is only meant for small config files, like editor settings.

afbjorklund:
I was suspecting that "appdata" and friends had the same concern on Windows as on macOS, so didn't use.

Didn't expect home to fail too, but here we are.

jandubois:
> Actually I think the problem is related to how Go handles paths in the `exec.Command`, when running from a prompt it seems to work.

This seems to confirm that the issue is an encoding problem, and not the long filenames.

I found this possible workaround, but haven't tested it:

```go
cmd := exec.Command("powershell.exe", "-c", "chcp", "65001", ">", "$null", ";", command)
```

Maybe even more gross than using short names. 😄

afbjorklund:
For myself, I only care about MSYS2 and MinGW64...

It is nice if the old school Command Prompt works too (if it isn't _too_ much hassle, like a simple .bat file or something)

But I will not bother with Powershell, if the user is interested in Linux containers then I assume they have a decent shell.

Something like this: <https://dilbert.com/strip/1995-06-24>

afbjorklund:
> However, I think you want to use `%USERPROFILE%` here, which is where application settings are stored.

Actually, that **is** what is being done: https://pkg.go.dev/os#UserHomeDir

> On Unix, including macOS, it returns the `$HOME` environment variable. On Windows, it returns `%USERPROFILE%`. On Plan 9, it returns the `$home` environment variable.

The rest was just my own guesses, based on variables (which I botched)

jandubois:
> But I will not bother with Powershell

This is not about the user using Powershell; it is just a way to switch the codepage before running the command. You can probably do the same thing with `cmd.exe`. Anyways, I don't really care; was just trying to provide some additional background info. I'm out of here now. 😄

jandubois:
> > However, I think you want to use `%USERPROFILE%` here, which is where application settings are stored.
> 
> Actually, that **is** what is being done: https://pkg.go.dev/os#UserHomeDir

I still think for this particular case `%LOCALAPPDATA%\Lima` is the right choice (which will typically be just `%USERPROFILE%\AppData\Local\Lima`. You don't want `%USERPROFILE%\.lima` because dot files are not a thing on Windows (they are not hidden by default). OTOH, if you only care about MSYS and MinGW64, then maybe dot files are the way to go... 🤷

pkg/store/dirnames/shortname_windows.go
afbjorklund: Added docs, renamed to ShortPathName

afbjorklund:
We can have a poll later, but it is something nice about using the same path on all platforms ?

PR #957, opened 2022-07-13
Disable starting an instance with default yaml if existing-instance f…

azhouwd:
Hi. Today if we start a VM with limactl in a workflow with tty=false without specifing template or yaml, it will create a new instance with default yaml if we start an instance that doesn't exist. For example "limactl start newInstance --tty=false" will create a new instance with name "newInstance" from default.yaml file. We want to provide a way for users to start only existing instances without creating a new instance from default.yaml file unintentionlly. For backward capability, we added new flag --existing-instance to start command so that when user specifies the flag to be true, lima will throw an error instead of creating a new instance if the instance doesn't exist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,6 +183,11 @@ To see the template list:
 $ limactl start --list-templates
 ```
 
+To specify using an existing instance
+```console
+$ limactl start <INSTANCE> --existing-instance
+```
+
 To create an instance "default" from a local file:
 ```console
 $ limactl start --name=default /usr/local/share/lima/examples/fedora.yaml
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -58,6 +58,7 @@ $ limactl start --name=default https://raw.githubusercontent.com/lima-vm/lima/ma
 	startCommand.Flags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "enable TUI interactions such as opening an editor, defaults to true when stdout is a terminal")
 	startCommand.Flags().String("name", "", "override the instance name")
 	startCommand.Flags().Bool("list-templates", false, "list available templates and exit")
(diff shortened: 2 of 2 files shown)


azhouwd:
> I'm not sure that this works well; if you forget to specify the new `--existing-instance` option (which is hard to type), then you will still create a new instance.
> 
> I would have preferred if we had separate `create` and `start` commands (with `create` auto-starting by default, and a `--no-start` option to prevent it), but that ship has sailed, as they say.
> 
> Assuming you are wrapping the start command with a script anyways, you could just add a check to the script to make sure the instance exists, e.g.
> 
> ```shell
> if [ -z "$(limactl ls -f '{{.Status}}' $INSTANCE 2>/dev/null)" ]; then
>     echo Instance $INSTANCE does not exist
>     exit 1
> fi
> ```
> 
> @AkihiroSuda I wonder why `limactl ls $INSTANCE` just shows a warning and still has an exit code of `0` when the instance doesn't exist; that feels wrong to me. Should we change that?

Thanks for the feedback @jandubois. I'm using a cobra application to wrap the limactl command, is it possible to do something like the shell script you have mentioned in the cobra app?

jandubois:
> I'm using a cobra application to wrap the limactl command, is it possible to do something like the shell script you have mentioned in the cobra app?

I'm not sure what "cobra app" means; to me `cobra` is just a Go library to parsing commandline arguments.

If this is a Go program, then yes, you can do this; either by running `limactl ls` as a subprocess, or just by checking if the directory `$LIMA_HOME/$INSTANCE` exists (defaults to `$HOME/.lima/$INSTANCE` when `LIMA_HOME` is not set).

AkihiroSuda:
>  I wonder why limactl ls $INSTANCE just shows a warning and still has an exit code of 0 when the instance doesn't exist; that feels wrong to me. Should we change that?

Yes

PR #967, opened 2022-07-19
Adopt CNCF code of conduct

jandubois:


jandubois:
Just learned about [default community health files](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file), to supply a file to each repo that doesn't define it itself.

Let's use that mechanism for `CODE_OF_CONDUCT.md` etc

jandubois:
I've created https://github.com/lima-vm/.github/blob/main/CODE_OF_CONDUCT.md and it does show up in the sidebar of all lima-vm projects now.

PR #968, opened 2022-07-19
SECURITY.md: move to lima-vm/.github repo

AkihiroSuda:
Moved to https://github.com/lima-vm/.github/blob/main/SECURITY.md

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,7 @@ binaries: clean \
 	cp -aL examples _output/share/lima
 	mkdir -p _output/share/doc/lima
 	cp -aL *.md LICENSE docs _output/share/doc/lima
+	echo "Moved to https://github.com/lima-vm/.github/blob/main/SECURITY.md" >_output/share/doc/lima/SECURITY.md
 ifneq ($(GOOS),windows)
 	ln -sf ../../lima/examples _output/share/doc/lima
 else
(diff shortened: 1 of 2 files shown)


jandubois:
Wouldn't it be better to just delete the file? That way the links in "issue reporting" will point directly to the shared file instead of to a file that simply says "moved".

AkihiroSuda:
> Wouldn't it be better to just delete the file? That way the links in "issue reporting" will point directly to the shared file instead of to a file that simply says "moved".
> 
> That's what I did in [lima-vm/alpine-lima#78](https://github.com/lima-vm/alpine-lima/pull/78).

done

jandubois (dismissed):

jandubois (approved):

Makefile
```diff
@@ -39,6 +39,7 @@ binaries: clean \
 	cp -aL examples _output/share/lima
 	mkdir -p _output/share/doc/lima
 	cp -aL *.md LICENSE docs _output/share/doc/lima
+	echo "Moved to https://github.com/lima-vm/.github/blob/main/SECURITY.md" >_output/share/doc/lima/SECURITY.md
```
AkihiroSuda, line 42: Updated PR to add this for binary distributions

PR #977, opened 2022-07-22
Add AES detection for Windows using cpuid

afbjorklund:
Avoids annoying warning about detection failed

### DIFF
--- a/pkg/sshutil/sshutil_windows_amd64.go
+++ b/pkg/sshutil/sshutil_windows_amd64.go
@@ -1,0 +1,9 @@
+package sshutil
+
+import (
+	"github.com/intel-go/cpuid"
+)
+
+func detectAESAcceleration() bool {
+	return cpuid.HasFeature(cpuid.AES)
+}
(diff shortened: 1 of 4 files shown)


AkihiroSuda (approved):
Thanks

pkg/sshutil/sshutil_windows_amd64.go
```diff
@@ -0,0 +1,9 @@
+package sshutil
+
+import (
+	"github.com/intel-go/cpuid"
+)
+
+func detectAESAcceleration() bool {
+	return cpuid.HasFeature(cpuid.AES)
```
AkihiroSuda, line 8: I guess we can use this on macOS and Linux too?

afbjorklund: Possibly, not sure if it needs to check for OS "support" too ?

It might also be available in go: https://pkg.go.dev/internal/cpu

afbjorklund: Also I don't know how things work on Windows arm, so left it.

PR #978, opened 2022-07-22
Avoid extra quotes in the ssh command parameters

afbjorklund:
This causes "invalid quotes" on systems without shells.

The parameters are sent to the ssh exec.Command, anyway.

----

This was causing the `lima` shell wrapper to fail under MSYS2.

It failed to recognize the extra quotes within the option parameters.

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -142,7 +142,7 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	if _, present := os.LookupEnv("COLORTERM"); present {
 		// SendEnv config is cumulative, with already existing options in ssh_config
-		sshArgs = append(sshArgs, "-o", "SendEnv=\"COLORTERM\"")
+		sshArgs = append(sshArgs, "-o", "SendEnv=COLORTERM")
 	}
 	sshArgs = append(sshArgs, []string{
 		"-q",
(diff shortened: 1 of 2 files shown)


jandubois:
> This causes "invalid quotes" on systems without shells.

I'm pretty sure that ssh itself is parsing the quotes, so not sure why this breaks without a shell. It sounds rather like a broken implementation of ssh.

> The parameters are sent to the ssh exec.Command, anyway.

And this tells you that on Unix the quotes are processed by ssh because there is no shell involved in calling the command. And on Windows the quotes are required if any argument contains whitespace because everything is joined together into a single commandline.

So I don't understand the failure mode you are observing.

I think this PR will also break #907, which goes the extra mile to protect these quotes from being swallowed by the shell because not having the quotes also breaks on Unix if arguments (e.g. filenames) contain whitespace.

afbjorklund:
I think the quotes will get added back (if needed) by exec.Command, it does all kinds of weird tricks.

The ssh command works fine if you run it from the shell, so the failure is when run by `limactl shell`

Most likely this willl need more testing and better code, if it should work with both windows and spaces...

Not sure there is a valid reason for spaces in filenames, but if it is requirement to have them then OK.

afbjorklund:
I was hoping to avoid this, but maybe it can't be avoided...

> On Windows, processes receive the whole command line as a single string and do their own parsing. Command combines and quotes Args into a command line string with an algorithm compatible with applications using CommandLineToArgvW (which is the most common way). Notable exceptions are msiexec.exe and cmd.exe (and thus, all batch files), which have a different unquoting algorithm. **In these or other similar cases, you can do the quoting yourself and provide the full command line in SysProcAttr.CmdLine, leaving Args empty.**

PR #979, opened 2022-07-22
Avoid translating paths in MSYS2 shell wrapper

afbjorklund:
The paths in the command are for the Linux virtual machine,
so we want remote unix paths instead of local windows paths.

Otherwise you get strange results from `lima`, such as:
`cat: 'C:/msys64/etc/os-release': No such file or directory`

----

See https://www.msys2.org/docs/filesystem-paths/#automatic-unix-windows-path-conversion

### DIFF
--- a/cmd/lima
+++ b/cmd/lima
@@ -30,4 +30,7 @@ fi
 if [ -n "${LIMA_WORKDIR}" ]; then
   set - --workdir "$LIMA_WORKDIR" "$@"
 fi
+# Avoid converting paths with MSYS2
+MSYS2_ARG_CONV_EXCL="*"
+export MSYS2_ARG_CONV_EXCL
 exec "$LIMACTL" shell "$@"


cmd/lima
```diff
@@ -30,4 +30,5 @@ fi
 if [ -n "${LIMA_WORKDIR}" ]; then
   set - --workdir "$LIMA_WORKDIR" "$@"
 fi
+export MSYS2_ARG_CONV_EXCL="*"
```
AkihiroSuda, line 33: This is bash extension, please use the classic sh form, or just replace the shebang to `/bin/bash`
```suggestion
# (please add a comment line here to explain this variable)
MSYS2_ARG_CONV_EXCL="*"
export MSYS2_ARG_CONV_EXCL
```

afbjorklund: Sorry, only tried with `dash` - maybe it has extensions over traditional bourne shell

afbjorklund: Oops, missed the comment line. Will update again.

AkihiroSuda (approved):
Thanks

PR #980, opened 2022-07-22
Allow running agents with tcp as well as unix

afbjorklund:
Use TCP sockets on Windows, instead of the Unix sockets.

**Will need to add TLS as well, before general availability.**


AkihiroSuda:
`Enable unix socket support on Windows` is now merged in the master (QEMU 7.2) https://github.com/qemu/qemu/commit/d409373b9d0482bbce312539d02a648cbb1c790a

https://github.com/lima-vm/lima/pull/981#issuecomment-1236873143

afbjorklund:
I would prefer if someone from the community could take over and complete the windows port, leaving this PR as-is for now.

PR #981, opened 2022-07-22
Allow using named pipes for qemu communication

afbjorklund:
Normally using unix sockets, but can also used named pipes (fifo)

**Currently "pipe" doesn't work properly, so workaround with "null"**


AkihiroSuda:
`Enable unix socket support on Windows` is now merged in the master (QEMU 7.2) https://github.com/qemu/qemu/commit/d409373b9d0482bbce312539d02a648cbb1c790a

afbjorklund:
I would prefer if someone from the community could take over and complete the windows port, leaving this PR as-is for now.

PR #982, opened 2022-07-22
Allow finding uefi firmware for qemu windows

afbjorklund:
The files are located in C:\Program Files\qemu\share on Windows.

For some reason*, using pflash doesn't work - only old bios flag.

----

I'm not sure the default files work, had provide a custom `OVMF_CODE.fd`

\* Issue being tracked at: https://gitlab.com/qemu-project/qemu/-/issues/513

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -416,7 +416,11 @@ func Cmdline(cfg Config) (string, []string, error) {
 		if err != nil {
 			return "", nil, err
 		}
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Is this still draft?

afbjorklund:
I left all the Windows patches in "draft" for early review, until they can be completed "properly"

The problem with this `-bios` workaround was that it required a custom firmware file - not the edk2 bundled with qemu...

PR #990, opened 2022-07-24
Document the default option for video.display

afbjorklund:
Useful for writing cross-platform yaml files:

Platform | DisplayType
--- | ---
Darwin | cocoa
Linux | gtk
Windows | sdl

But is not a new display type, or anything...

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -208,7 +208,9 @@ firmware:
   legacyBIOS: null
 
 video:
-  # QEMU display, e.g., "none", "cocoa", "sdl", "gtk".
+  # QEMU display, e.g., "none", "cocoa", "sdl", "gtk", "default".
+  # Choosing "none" will hide the video output, and not show any window.
+  # Choosing "default" will pick the first available of: gtk, sdl, cocoa.
   # As of QEMU v6.2, enabling this is known to have negative impact
   # on performance on macOS hosts: https://gitlab.com/qemu-project/qemu/-/issues/334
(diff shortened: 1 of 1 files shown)


examples/default.yaml
```diff
@@ -208,9 +208,11 @@ firmware:
   legacyBIOS: null
 
 video:
-  # QEMU display, e.g., "none", "cocoa", "sdl", "gtk".
+  # QEMU display, e.g., "none", "cocoa", "sdl", "gtk", "default".
   # As of QEMU v6.2, enabling this is known to have negative impact
   # on performance on macOS hosts: https://gitlab.com/qemu-project/qemu/-/issues/334
+  # Choosing "none" will hide the video output, and not show any window.
+  # Choosing "default" will pick the first available of: gtk, sdl, cocoa.
```
AkihiroSuda, line 215: It should be noted that this is the default value of QEMU, not Lima

afbjorklund: Hmm, that was supposed to be on the line directly below that one.
```
  # 🟢 Builtin default: "none"
```

AkihiroSuda: I guess that are confusing to new comers and need explicit explanation

afbjorklund: Maybe it would be less confusing, if all five alternatives were explained ?

afbjorklund: Perhaps if the text is not close to the other one, it would be better ? Flipped them...

AkihiroSuda (approved):

PR #999, opened 2022-07-29
Add github test for windows using wine

afbjorklund:
This will only run the unit tests, not any integration tests.

For that, a real Windows machine (with Hyper-V) is needed...

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,6 +51,27 @@ jobs:
     - name: Uninstall
       run: sudo make uninstall
 
+  cross-windows:
+    name: "Windows tests (cross)"
+    runs-on: ubuntu-20.04
+    timeout-minutes: 30
+    steps:
+    - uses: actions/setup-go@v3
+      with:
+        go-version: 1.18.x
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -51,6 +51,28 @@ jobs:
     - name: Uninstall
       run: sudo make uninstall
 
+  windows:
```
AkihiroSuda, line 54:
```suggestion
  wine:
```

or something like “cross-windows”

AkihiroSuda:
Does cirrus work?
https://github.com/containerd/nerdctl/blob/6354356c5c6224e98d7bd78e04742180503fe5f2/.cirrus.yml#L21

afbjorklund:
Apparently the .NET support (mono-runtime) kicked in, for the .exe files

afbjorklund:
Don't know anything about Cirrus, sorry... There was some other system, next to Travis when that was in play.

Something like Apptainer: "Appveyor" ? But I have no idea how it worked, just remembered seeing a file for it.

AkihiroSuda (approved):

PR #1004, opened 2022-07-31
Add VNC video display including password

### LINKED ISSUE #1003: Add support for VNC, in addition to the current QEMU displays
### Description

The qemu display might slow down the host, and the window be in the way...

But it is also possible to connect to the qemu virtual machine desktop using [vnc](https://en.wikipedia.org/wiki/Virtual_Network_Computing)

The parameters can be somewhat tricky, so give it a separate configuration.

This also clearly separates a display window with a network connection.

---

The recommended configuration only binds to localhost, **not** to the network.

It is also possible to add a password (8 char, clear text) to the VNC connection.

In addition, it is also possible to run the whole interface in a web browser with [novnc](https://novnc.com/info.html).

It uses a websocket connection to port 5700+d, instead of the regular port 5900+d.

afbjorklund:
The other display options open a window always, while the
vnc is more "on demand" by using a separate vnc viewer.
    
Add localhost and password support for some minimal security.
The password is generated, and is stored as an instance file.

Closes #1003

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -245,13 +245,21 @@ firmware:
   legacyBIOS: null
 
 video:
-  # QEMU display, e.g., "none", "cocoa", "sdl", "gtk", "default".
+  # QEMU display, e.g., "none", "cocoa", "sdl", "gtk", "vnc", "default".
   # Choosing "none" will hide the video output, and not show any window.
+  # Choosing "vnc" will use a network server, and not show any window.
   # Choosing "default" will pick the first available of: gtk, sdl, cocoa.
-  # As of QEMU v6.2, enabling this is known to have negative impact
+  # As of QEMU v6.2, enabling anything but none or vnc is known to have negative impact
   # on performance on macOS hosts: https://gitlab.com/qemu-project/qemu/-/issues/334
   # 🟢 Builtin default: "none"
   display: null
+  # VNC (Virtual Network Computing) is a platform-independent graphical
+  # desktop-sharing system that uses the Remote Frame Buffer protocol (RFB)
+  vnc:
+    # VNC display, e.g.,"to=L", "host:d", "unix:path", "none"
+    # By convention the TCP port is 5900+d, connections from any host.
+    # 🟢 Builtin default: "127.0.0.1:0,to=9"
+    display: null
 
 # The instance can get routable IP addresses from the vmnet framework using
 # https://github.com/lima-vm/socket_vmnet.
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net"
 	"os"
 	"os/exec"
@@ -251,6 +252,15 @@ func (a *HostAgent) emitEvent(_ context.Context, ev events.Event) {
 	}
 }
 
+func generatePassword(length int) (string, error) {
+	passwd := ""
+	rand.Seed(time.Now().UnixNano())
+	for i := 0; i < length; i++ {
+		passwd += strconv.Itoa(rand.Intn(10))
+	}
+	return passwd, nil
+}
+
 func (a *HostAgent) Run(ctx context.Context) error {
 	defer func() {
 		exitingEv := events.Event{
@@ -283,6 +293,50 @@ func (a *HostAgent) Run(ctx context.Context) error {
 	errCh, err := a.driver.Start(ctx)
 	if err != nil {
 		return err
+	}
+
+	if *a.y.Video.Display == "vnc" {
+		vncdisplay, vncoptions, _ := strings.Cut(*a.y.Video.VNC.Display, ",")
+		vnchost, vncnum, err := net.SplitHostPort(vncdisplay)
+		if err != nil {
+			return err
+		}
+		n, err := strconv.Atoi(vncnum)
+		if err != nil {
+			return err
+		}
+		vncport := strconv.Itoa(5900 + n)
+		vncpwdfile := filepath.Join(a.instDir, filenames.VNCPasswordFile)
+		vncpasswd, err := generatePassword(8)
+		if err != nil {
+			return err
+		}
+		if err := a.driver.ChangeDisplayPassword(ctx, vncpasswd); err != nil {
+			return err
+		}
+		if err := os.WriteFile(vncpwdfile, []byte(vncpasswd), 0600); err != nil {
+			return err
+		}
+		if strings.Contains(vncoptions, "to=") {
+			vncport, err = a.driver.GetDisplayConnection(ctx)
+			if err != nil {
+				return err
+			}
+			p, err := strconv.Atoi(vncport)
+			if err != nil {
+				return err
+			}
+			vncnum = strconv.Itoa(p - 5900)
+			vncdisplay = net.JoinHostPort(vnchost, vncnum)
+		}
+		vncfile := filepath.Join(a.instDir, filenames.VNCDisplayFile)
+		if err := os.WriteFile(vncfile, []byte(vncdisplay), 0600); err != nil {
+			return err
+		}
+		vncurl := "vnc://:" + vncpasswd + "@" + net.JoinHostPort(vnchost, vncport)
+		logrus.Infof("VNC server running at <%s>", vncurl)
+		logrus.Infof("VNC Display: \"%s\" `%s`", vncdisplay, vncfile)
+		logrus.Infof("VNC Password: \"%s\" `%s`", vncpasswd, vncpwdfile)
 	}
 
 	stBase := events.Status{
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math/rand"
 	"net"
 	"os"
 	"os/exec"
@@ -33,6 +32,7 @@ import (
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/lima-vm/sshocker/pkg/ssh"
+	"github.com/sethvargo/go-password/password"
 	"github.com/sirupsen/logrus"
 )
 
@@ -253,12 +253,8 @@ func (a *HostAgent) emitEvent(_ context.Context, ev events.Event) {
 }
 
 func generatePassword(length int) (string, error) {
-	passwd := ""
-	rand.Seed(time.Now().UnixNano())
-	for i := 0; i < length; i++ {
-		passwd += strconv.Itoa(rand.Intn(10))
-	}
-	return passwd, nil
+	// avoid any special symbols, to make it easier to copy/paste
+	return password.Generate(length, length/4, 0, false, false)
 }
 
 func (a *HostAgent) Run(ctx context.Context) error {
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -329,10 +329,10 @@ func (a *HostAgent) Run(ctx context.Context) error {
 		if err := os.WriteFile(vncfile, []byte(vncdisplay), 0600); err != nil {
 			return err
 		}
-		vncurl := "vnc://:" + vncpasswd + "@" + net.JoinHostPort(vnchost, vncport)
-		logrus.Infof("VNC server running at <%s>", vncurl)
-		logrus.Infof("VNC Display: \"%s\" `%s`", vncdisplay, vncfile)
-		logrus.Infof("VNC Password: \"%s\" `%s`", vncpasswd, vncpwdfile)
+		vncurl := "vnc://" + net.JoinHostPort(vnchost, vncport)
+		logrus.Infof("VNC server running at %s <%s>", vncdisplay, vncurl)
+		logrus.Infof("VNC Display: `%s`", vncfile)
+		logrus.Infof("VNC Password: `%s`", vncpwdfile)
 	}
 
 	stBase := events.Status{
--- a/pkg/qemu/qemu_driver.go
+++ b/pkg/qemu/qemu_driver.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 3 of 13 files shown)


examples/default.yaml
```diff
@@ -215,6 +215,18 @@ video:
...
   display: null
+  # VNC (Virtual Network Computing) is a platform-independent graphical
+  # desktop-sharing system that uses the Remote Frame Buffer protocol (RFB)
+  vnc:
+    # VNC display, e.g., "127.0.0.1:0", "host:d", "unix:path", "none"
+    # By convention the TCP port is 5900+d, connections from any host.
+    display: null
+    # VNC websocket, e.g. "127.0.0.1:5700", "host:port"
+    # By convention the TCP port is 5700+d, connections from any host.
+    websocket: null
```
AkihiroSuda, line 226:
```suggestion
    webSocket: null
```

examples/default.yaml
```diff
@@ -215,6 +215,18 @@ video:
...
+    # VNC password, e.g. "on", "off"
+    # Use QMP to set password: `change-vnc-password password=secret`
+    password: null
```
AkihiroSuda, line 229: Password is necessary and should be autogenerated and saved as a text file under the instance dir

examples/default.yaml
```diff
@@ -215,6 +215,18 @@ video:
   # on performance on macOS hosts: https://gitlab.com/qemu-project/qemu/-/issues/334
```
AkihiroSuda, line 215: Do we ever need VNC if we can just get this issue fixed?

examples/default.yaml
afbjorklund: https://pkg.go.dev/github.com/digitalocean/go-qemu/qmp/raw#Monitor.ChangeVNCPassword

examples/default.yaml
afbjorklund: It is a separate feature, that bug seems specific to macOS ?

afbjorklund:
Probably should move the display of the password, or perhaps just show the filename ?

```
INFO[0000] [hostagent] Starting QEMU (hint: to watch the boot progress, see "/home/anders/.lima/default/serial.log") 
INFO[0001] [hostagent] VNC Password for 127.0.0.1:0 is: 17791850 
INFO[0001] SSH Local Port: 60022                        
INFO[0001] [hostagent] Waiting for the essential requirement 1 of 5: "ssh" 
```

Some viewers accept a passwordfile, rather than prompting for the (clear-text) password.

afbjorklund:
Oops, forgot to seed the random number generator so the "password" is always the same.

https://pkg.go.dev/math/rand

afbjorklund:
Should show a real URL, so that it is clickable in the console (maybe include the VNC password too?)

The current output works in vncviewer:

`127.0.0.1:0` <--> `vnc://127.0.0.1:5900`

https://tools.ietf.org/html/rfc7869

afbjorklund:
```
INFO[0000] [hostagent] VNC Password for 127.0.0.1:0 <vnc://127.0.0.1:5900> is: 
INFO[0000] [hostagent]     "46479782" | `/home/anders/.lima/default/vncpassword` 
```

pkg/hostagent/hostagent.go
```diff
@@ -243,6 +244,15 @@ func logPipeRoutine(r io.Reader, header string) {
 	}
 }
 
+func randomPassword(length int) string {
+	passwd := ""
+	rand.Seed(time.Now().UnixNano())
+	for i := 0; i < length; i++ {
+		passwd += strconv.Itoa(rand.Intn(10))
```
AkihiroSuda, line 251: numeric-only password is insecure

pkg/store/filenames/filenames.go
```diff
@@ -37,6 +37,7 @@ const (
 	SerialLog          = "serial.log"
 	SerialSock         = "serial.sock"
 	SSHSock            = "ssh.sock"
+	VNCPasswordFile    = "vncpassword"
```
AkihiroSuda, line 40: Please update `docs/internal.md` to explain this

pkg/hostagent/hostagent.go
afbjorklund: the password is sent in clear-text

jandubois:
> Note that the default user doesn't have any password, so still needs passwd for console.

So this works great to get the console, but how do you log in?

examples/default.yaml
```diff
@@ -215,6 +215,12 @@ video:
...
   # 🟢 Builtin default: "none"
   display: null
+  # VNC (Virtual Network Computing) is a platform-independent graphical
+  # desktop-sharing system that uses the Remote Frame Buffer protocol (RFB)
+  vnc:
+    # VNC display, e.g., "127.0.0.1:0", "host:d", "unix:path", "none"
```
jandubois, line 221: This should explain that if you don't know what these options mean, then you want `127.0.0.1:0`.

jandubois:
> > Note that the default user doesn't have any password, so still needs passwd for console.
> 
> So this works great to get the console, but how do you log in?

Sorry, was confused; I just realized that this is exactly the same as the regular console 😄

afbjorklund:
It also depends on the example, if you add a login manager you could also add more login methods ?

The VNC "password" is just to protect the display as such, it doesn't change anything about authentication

examples/default.yaml
afbjorklund: The only issue would be if you started two instances, then one of them would be `127.0.0.1:1`

afbjorklund:
It seems there is a nice feature to try next available port, maybe have that as a vnc default and require "display" ?

```yaml
  display: "vnc"
  vnc:
    display: "127.0.0.1:0,to=9"
```

Then it wouldn't use vnc.display, unless also setting display=vnc. And return an unused display 0-9 ? (port 5900-5909)

           to=L
               With this option, QEMU will try next available VNC displays, until the number L, if the origianlly defined "-vnc display" is not available, e.g. port 5900+display is already used by
               another application. By default, to=0.

           host:d
               TCP connections will only be allowed from host on display d.  By convention the TCP port is 5900+d. Optionally, host can be omitted in which case the server will accept connections
               from any host.

pkg/hostagent/hostagent.go
AkihiroSuda: Still we have to choose a secure password, [even for localhost](https://theevilbit.github.io/posts/vmware_fusion_11_guest_vm_rce_cve-2019-5514/)

afbjorklund:
I suppose QueryVNC would do the trick:

```
(QEMU) query-vnc
{"return": {"enabled": true, "auth": "vnc", "family": "ipv4", "clients": [], "service": "5900", "host": "127.0.0.1"}}
```

pkg/hostagent/hostagent.go
afbjorklund: okay, I will try to find a suitable library and use that instead

afbjorklund: Added https://github.com/sethvargo/go-password

examples/default.yaml
afbjorklund: Implemented the "127.0.0.1:0,to=9" now, to pick the first available display

afbjorklund:
Now the default vnc display will pick the first available one, from `127.0.0.1:0` `127.0.0.1:1` ... `127.0.0.1:9`

This means that all the user needs to to do is to change the display type from "none" to "vnc", and get the password.

It can either be seen in the log, or read from the instance file.

The log will also show the display number, if it wasn't fixed.

AkihiroSuda:
> Then you could just read the vncdisplay file, for the address ?

SGTM.
Maybe we can also have `limactl show-vnc` (akin to `limactl show-ssh`) to show the VNC info.


Btw, I guess the VNC feature should be marked as an experimental for now, at least until we can have integration tests (how?)?

AkihiroSuda:
CI failing

afbjorklund:
> Btw, I guess the VNC feature should be marked as an experimental for now, at least until we can have integration tests (how?)?

I couldn't really find any integration tests for the other display settings either, except for the default of `none` I guess.

afbjorklund:
Just because I said that, I actually managed to break "default" in the latest re-design (when moving to Video.Display)

```
"level":"fatal","msg":"Could not set password"
```

oops.

afbjorklund:
Broken by the driver introduction

afbjorklund:
Added some new functions to the BaseDriver, need to verify that it is QEMU before allowing VNC...

```go
      ChangeDisplayPassword(_ context.Context, password string) error

      GetDisplayConnection(_ context.Context) (string, error)

```

examples/default.yaml
```diff
@@ -229,13 +229,21 @@ firmware:
...
   # Choosing "default" will pick the first available of: gtk, sdl, cocoa.
-  # As of QEMU v6.2, enabling this is known to have negative impact
+  # As of QEMU v6.2, enabling anything but none or vnc is known to have negative impact
   # on performance on macOS hosts: https://gitlab.com/qemu-project/qemu/-/issues/334
   # 🟢 Builtin default: "none"
   display: null
+  # VNC (Virtual Network Computing) is a platform-independent graphical
+  # desktop-sharing system that uses the Remote Frame Buffer protocol (RFB)
+  vnc:
+    # VNC display, e.g.,"to=L", "host:d", "unix:path", "none"
```
AkihiroSuda, line 243: What does "to=L" mean?
And how is "none" expected to be used?

pkg/hostagent/hostagent.go
```diff
@@ -259,6 +265,50 @@ func (a *HostAgent) Run(ctx context.Context) error {
...
+			vncdisplay = net.JoinHostPort(vnchost, vncnum)
+		}
+		vncfile := filepath.Join(a.instDir, filenames.VNCDisplayFile)
+		if err := os.WriteFile(vncfile, []byte(vncdisplay), 0600); err != nil {
+			return err
+		}
+		vncurl := "vnc://:" + vncpasswd + "@" + net.JoinHostPort(vnchost, vncport)
+		logrus.Infof("VNC server running at <%s>", vncurl)
+		logrus.Infof("VNC Display: \"%s\" `%s`", vncdisplay, vncfile)
+		logrus.Infof("VNC Password: \"%s\" `%s`", vncpasswd, vncpwdfile)
```
AkihiroSuda, line 309: The raw password string shouldn't be printed. Just print the password file path would be ok.

pkg/hostagent/hostagent.go
```diff
@@ -259,6 +265,50 @@ func (a *HostAgent) Run(ctx context.Context) error {
...
+		if err != nil {
+			return err
+		}
+		n, err := strconv.Atoi(vncnum)
+		if err != nil {
+			return err
+		}
+		vncport := strconv.Itoa(5900 + n)
+		vncpwdfile := filepath.Join(a.instDir, filenames.VNCPasswordFile)
+		vncpasswd, err := generatePassword(8)
```
AkihiroSuda, line 280: 8 seems too short. What about 32?

examples/default.yaml
```diff
@@ -229,13 +229,21 @@ firmware:
...
   # Choosing "none" will hide the video output, and not show any window.
+  # Choosing "vnc" will use a network server, and not show any window.
   # Choosing "default" will pick the first available of: gtk, sdl, cocoa.
-  # As of QEMU v6.2, enabling this is known to have negative impact
+  # As of QEMU v6.2, enabling anything but none or vnc is known to have negative impact
   # on performance on macOS hosts: https://gitlab.com/qemu-project/qemu/-/issues/334
   # 🟢 Builtin default: "none"
   display: null
+  # VNC (Virtual Network Computing) is a platform-independent graphical
+  # desktop-sharing system that uses the Remote Frame Buffer protocol (RFB)
```
AkihiroSuda, line 241: Let's mark this experimental.

pkg/hostagent/hostagent.go
afbjorklund: It only supports 8

examples/default.yaml
afbjorklund: This was from the upstream documentation, could copy the detailed explanations

pkg/hostagent/hostagent.go
afbjorklund: https://www.rfc-editor.org/rfc/rfc6143.html#section-7.2.2

pkg/hostagent/hostagent.go
afbjorklund: Then the link won't work and it would be more hassle for the user (running the `cat`), but can do this.

examples/default.yaml
afbjorklund: to=L
               With this option, QEMU will try next available VNC displays,
               until the number L, if the origianlly defined "-vnc display"
               is not available, e.g. port 5900+display is already used by
               another application. By default, to=0.

           host:d
               TCP connections will only be allowed from host on display d.
               By convention the TCP port is 5900+d. Optionally, host can be
               omitted in which case the server will accept connections from
               any host.

           unix:path
               Connections will be allowed over UNIX domain sockets where
               path is the location of a unix socket to listen for
               connections on.

           none
               VNC is initialized but not started. The monitor "change"
               command can be used to later start the VNC server.

pkg/hostagent/hostagent.go
jandubois: Maybe we should still have a `limactl vnc INSTANCE` command after all, if we can't have clickable links (and I reluctantly agree that printing passwords is always bad).

I know I argued against the vnc subcommand before. I still don't think it should embed a VNC client, but just run `open vnc://...` for the user.

afbjorklund: Maybe call it `viewer` or something, in case there is some other protocol added in the future ?

afbjorklund: And it's not really a password, it's more like a PIN code or a little like a vnc token (although always granted)

afbjorklund: You can click the link (like before), just that you also need to retrieve the password from the file...

afbjorklund:
```
INFO[0001] [hostagent] VNC server running at 127.0.0.1:0 <vnc://127.0.0.1:5900> 
INFO[0001] [hostagent] VNC Display: `/home/anders/.lima/default/vncdisplay` 
INFO[0001] [hostagent] VNC Password: `/home/anders/.lima/default/vncpassword` 
```

examples/default.yaml
AkihiroSuda: @afbjorklund @jandubois WDYT?

AkihiroSuda:
Needs rebase

examples/default.yaml
AkihiroSuda: FYI: The experimental features are now tracked in https://github.com/lima-vm/lima/blob/master/docs/experimental.md

See also https://github.com/lima-vm/lima/search?q=experimental

afbjorklund:
For some versions of QEMU, the **only** display option is VNC.

pkg/qemu/qemu_driver.go
```diff
@@ -106,6 +107,80 @@ func (l *LimaQemuDriver) Stop(ctx context.Context) error {
...
+	}
+	return *info.Service, nil
+}
+
+func (l *LimaQemuDriver) removeVNCFiles() error {
+	vncfile := filepath.Join(l.Instance.Dir, filenames.VNCDisplayFile)
+	_ = os.RemoveAll(vncfile)
+	vncpwdfile := filepath.Join(l.Instance.Dir, filenames.VNCPasswordFile)
+	_ = os.RemoveAll(vncpwdfile)
+	return nil
```
AkihiroSuda, line 181: Why discard err?

afbjorklund: It was more like `rm -f`

afbjorklund: Can check for instance instead, perhaps ?

AkihiroSuda: `RemoveAll` returns nil err when the file does not exist

AkihiroSuda (approved):
Thanks

PR #1049, opened 2022-09-13
Return error when template expansion fails

jandubois:
Unfortunately I could not determine what triggered this error, but I've run into it while testing Rancher Desktop with socket_vmnet, and qemu failed because it had a `-netdev` option without `id` in the value (there was no value at all).

I've not been able to reproduce it again since I fixed the error reporting.


AkihiroSuda (approved):
Thanks

PR #1052, opened 2022-09-15
templates: add podman-rootful

afbjorklund:
For completeness, when compared with docker-rootful

There is no "podman" group, so use `sudo podman`

----

Note: podman is daemonless, so shell doesn't use sock

If you really wanted to, you could add a `--remote` flag...

### DIFF
--- a/examples/podman-rootful.yaml
+++ b/examples/podman-rootful.yaml
@@ -1,0 +1,70 @@
+# Example to use Podman instead of containerd & nerdctl
+# $ limactl start ./podman-rootful.yaml
+# $ limactl shell podman-rootful sudo podman run -it -v $HOME:$HOME --rm docker.io/library/alpine
+
+# To run `podman` on the host (assumes podman-remote is installed):
+# $ export CONTAINER_HOST=$(limactl list podman-rootful --format 'unix://{{.Dir}}/sock/podman.sock')
+# $ podman --remote ...
+
+# To run `docker` on the host (assumes docker-cli is installed):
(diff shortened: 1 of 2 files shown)


examples/podman-rootful.yaml
```diff
@@ -0,0 +1,70 @@
...
+    command -v podman >/dev/null 2>&1 && exit 0
+    if [ ! -e /etc/systemd/system/podman.socket.d/override.conf ]; then
+      mkdir -p /etc/systemd/system/podman.socket.d
+      cat <<-EOF >/etc/systemd/system/podman.socket.d/override.conf
+      [Socket]
+      SocketUser=${LIMA_CIDATA_USER}
+    EOF
+    fi
+    if [ ! -e /etc/tmpfiles.d/podman.conf ]; then
+      # D! : "If the exclamation mark is used, this line is only safe to execute during boot"
```
AkihiroSuda, line 43: What is `D!` ?

afbjorklund: It was the original line, the one we are overriding here. Maybe the old value or even a diff would have been more obvious...

Anyway, it only creates the directory at boot time - which is not the best for Lima since that would require an instance restart

AkihiroSuda: Your PR does not seem using `D!`, so it might be confusing to mention `D!` here without showing the context

afbjorklund: It is probably not even relevant, just noted it after seeing https://github.com/containers/podman/commit/2e0a9c453b03d2a372a3ab03b9720237e93a067c

AkihiroSuda (approved):
Thanks

PR #1054, opened 2022-09-15
Implement the backup and restore commands

### LINKED ISSUE #1051: Does lima support checkpoint backup and restore like normal virtual machine?
### Discussed in https://github.com/lima-vm/lima/discussions/1050

<div type='discussions-op-text'>

<sup>Originally posted by **f91og** September 13, 2022</sup>
Just want to quick restore to a certain status after tried several things that caused develop environment broken</div>

afbjorklund:
For making qcow2 snapshots of the running virtual machine.

It is possible to make multiple snapshots, in the diffdisk.

Closes #1051


afbjorklund:
This has similar problems to the pause commands, in that the ssh mux is not _too_ happy about the restore ?

* https://github.com/lima-vm/lima/pull/642

cmd/limactl/backup.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/lima-vm/lima/pkg/snapshot"
+	"github.com/lima-vm/lima/pkg/store"
+
+	"github.com/spf13/cobra"
+)
+
+func newBackupCommand() *cobra.Command {
+	var backupCmd = &cobra.Command{
+		Use:               "backup INSTANCE",
+		Short:             "Backup (save) an instance",
```
AkihiroSuda, line 15: I'd expect "backup" to back up the VM to an external storage.

Maybe it should be like `limactl snapshot (ls|create|rollback)` ?

afbjorklund: Might want to use the same words as qemu, in that case. (list/create/apply/delete)

I can make a go at implementing the "list", changing the cmd should be easier to do

afbjorklund: Possibly should also try to implement "off-line" versions, of the HMP commands ?

```
FATA[0000] expected status "Running", got "Stopped"     
```

       Parameters to snapshot subcommand:

       snapshot
           is the name of the snapshot to create, apply or delete

       -a  applies a snapshot (revert disk to saved state)

       -c  creates a snapshot

       -d  deletes a snapshot

       -l  lists all snapshots in the given image

So that one can perform those operations with `qemu-img`, when the VM is not running ?

afbjorklund: The list doesn't have any fancy output options, so probably just dump the output as-is ?

```
Snapshot list:
ID        TAG                     VM SIZE                DATE       VM CLOCK
1         tag                         0 B 2022-09-21 18:31:41   00:00:00.000
```

afbjorklund:
Adding basic implementation of "delvm" and "list", will rewrite as per comments above.

```
$ _output/bin/limactl backup --tag tag alpine
INFO[0000] Sending HMP savevm command                   
$ _output/bin/limactl restore --list alpine
Snapshot list:
ID        TAG                     VM SIZE                DATE       VM CLOCK
1         tag                     274 MiB 2022-09-21 19:34:50   00:02:39.937
$ _output/bin/limactl backup --tag tag --rm alpine
INFO[0000] Sending HMP delvm command                    
$ _output/bin/limactl restore --list alpine
$ 
```

pkg/qemu/qemu.go
```diff
@@ -128,6 +131,76 @@ func EnsureDisk(cfg Config) error {
...
+	}
+	defer func() { _ = qmpClient.Disconnect() }()
+	rawClient := raw.NewMonitor(qmpClient)
+	logrus.Infof("Sending HMP %s command", cmd)
+	hmc := fmt.Sprintf("%s %s", cmd, tag)
+	_, err = rawClient.HumanMonitorCommand(hmc, nil)
+	return err
+}
+
+func sendImgCommand(cfg Config, args ...string) (string, error) {
```
AkihiroSuda, line 159:
```suggestion
func execImgCommand(cfg Config, args ...string) (string, error) {
```

pkg/qemu/qemu.go
```diff
@@ -128,6 +131,76 @@ func EnsureDisk(cfg Config) error {
...
+func Load(cfg Config, run bool, tag string) error {
+	if run {
+		return sendHmpCommand(cfg, "loadvm", tag)
+	}
+	// -a  applies a snapshot
+	_, err := sendImgCommand(cfg, "snapshot", "-a", tag)
+	return err
+}
+
+func List(cfg Config) (string, error) {
```
AkihiroSuda, line 198: Perhaps add a comment to clarify that the result is not machine-parsable

cmd/limactl/snapshot.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	ctx := cmd.Context()
+	return snapshot.Load(ctx, inst, tag)
+}
+
+func newSnapshotListCommand() *cobra.Command {
+	var listCmd = &cobra.Command{
+		Use:               "list INSTANCE",
+		Short:             "List existing snapshots",
```
AkihiroSuda, line 133: Perhaps add `Long` message to clarify that the result is not machine-parsable

AkihiroSuda:
Thanks, the CLI design looks good, could you update the docs and integration tests?

pkg/qemu/qemu.go
afbjorklund: Good idea, I think the original API just promises a list of strings (lines).

cmd/limactl/snapshot.go
afbjorklund: Theoretically we _could_ parse it, it is a space separated list with a column header.

https://github.com/qemu/qemu/blob/v7.0.0/block/qapi.c#L664_L698

afbjorklund: I added a simple --quiet command, to only print the tags (but didn't bother with --json)

pkg/qemu/qemu.go
afbjorklund: I removed the initial heading, so now it at least _looks_ similar to `limactl ls`

```console
$ limactl snapshot create --tag snap alpine
$ limactl snapshot list alpine
ID        TAG                     VM SIZE                DATE       VM CLOCK
1         snap                        0 B 2022-10-16 11:37:23   00:00:00.000
```

afbjorklund:
Left to do is adding integration tests, for the `limactl snapshot` sub-commands.

afbjorklund:
Some help with the integration tests for snapshots would be appreciated. 🙏

AkihiroSuda:
Integration tests can be added like this
https://github.com/lima-vm/lima/blob/4a4cc62749c1ad5d781054981dd1da19de1202bf/hack/test-example.sh#L24
https://github.com/lima-vm/lima/blob/4a4cc62749c1ad5d781054981dd1da19de1202bf/hack/test-example.sh#L240-L267

afbjorklund:
Snapshot worked on default (and alpine, but failed on experimental/9p ?

```
TEST| [INFO] Testing online snapshots
time="2022-11-16T20:21:20Z" level=info msg="Sending HMP savevm command"
TEST| [INFO] snapshot list: expected=snap1 got=
Error: ERROR] snapshot list did not return expected value
```

Need to see if there is a better way to return errors, after running HMP/QMP...

afbjorklund:
Now it is at least logging the real problem:
`"output: Error: Migration is disabled when VirtFS export path '/tmp/lima' is mounted in the guest using mount_tag 'mount1'\r\n"`

So this means that snapshots won't work with 9p mounts, while the instance is still running.

cmd/limactl/snapshot.go
```diff
@@ -0,0 +1,181 @@
...
+
+	ctx := cmd.Context()
+	return snapshot.Del(ctx, inst, tag)
+}
+
+func newSnapshotApplyCommand() *cobra.Command {
+	var applyCmd = &cobra.Command{
+		Use:               "apply INSTANCE",
+		Aliases:           []string{"load"},
+		Short:             "Apply (load) a snapshot",
```
AkihiroSuda, line 100: Can we just call this "load"? "apply" might be confusing.

AkihiroSuda (approved):
Thanks

pkg/snapshot/snapshot.go
```diff
@@ -0,0 +1,57 @@
...
+func Save(ctx context.Context, inst *store.Instance, tag string) error {
+	y, err := inst.LoadYAML()
+	if err != nil {
+		return err
+	}
+	limaDriver := driverutil.CreateTargetDriverInstance(&driver.BaseDriver{
+		Instance: inst,
+		Yaml:     y,
+	})
+	return limaDriver.CreateSnapshot(ctx, inst.Status == store.StatusRunning, tag)
```
balajiv113, line 32: Can we do this inst.status check in the qemu driver itself (we do have access to instance in driver)??

This way we don't need to expose a bool arg in driver interface

afbjorklund: Sure, I will remove it from the "API"

AkihiroSuda:
Could you rebase to rerun the CI?

afbjorklund:
It would be nice, if the generated command help stayed out of the README....

pkg/driver/driver.go
```diff
@@ -39,3 +48,19 @@ func (d *BaseDriver) Start(_ context.Context) (chan error, error) {
...
+func (d *BaseDriver) ApplySnapshot(_ context.Context, _ string) error {
+	return fmt.Errorf("unimplemented")
+}
+
+func (d *BaseDriver) DeleteSnapshot(_ context.Context, _ string) error {
+	return fmt.Errorf("unimplemented")
+}
+
+func (d *BaseDriver) ListSnapshots(_ context.Context) (string, error) {
+	return "", nil
```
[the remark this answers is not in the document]
afbjorklund: Make sense, but would apply to all the commands in that case ?

Here it just returns "an empty list" (as string)

pkg/driver/driver.go
```diff
@@ -39,3 +48,19 @@ func (d *BaseDriver) Start(_ context.Context) (chan error, error) {
...
+
+func (d *BaseDriver) ApplySnapshot(_ context.Context, _ string) error {
+	return fmt.Errorf("unimplemented")
+}
+
+func (d *BaseDriver) DeleteSnapshot(_ context.Context, _ string) error {
+	return fmt.Errorf("unimplemented")
+}
+
+func (d *BaseDriver) ListSnapshots(_ context.Context) (string, error) {
```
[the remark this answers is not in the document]
afbjorklund: As noted above, the output is from the human monitor and not parseable by machines (but scrapable)

But since we already convert it in the command, I suppose we could do the same in the driver instead.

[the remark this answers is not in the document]
afbjorklund:
```go
        out, err := snapshot.List(ctx, inst)
        if err != nil {
                return err
        }
        if quiet {
                for i, line := range strings.Split(out, "\n") {
                        // "ID", "TAG", "VM SIZE", "DATE", "VM CLOCK", "ICOUNT"
                        fields := strings.Fields(line)
                        if i == 0 || line == "" {
                                // skip header and empty line after using split
                                continue
                        }
                        tag := fields[1]
                        fmt.Printf("%s\n", tag)
                }
                return nil
        }
        fmt.Print(out)

```

pkg/driver/driver.go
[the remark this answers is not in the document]
balajiv113: Other commands were already returning error if am not wrong.
I would prefer error even though returning an empty list as string is not wrong.

Mainly because any of snapshot command should not work for vz driver.

pkg/driver/driver.go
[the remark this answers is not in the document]
balajiv113: Yes exactly, we could scrap it to a know object in driver itself. 
So the commands will become independent of the driver output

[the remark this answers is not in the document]
afbjorklund: Rebased with `string` (including HEADER line), needs to return an array of... something ?

The only required field in it is Tag, since that (string) is required by the other commands.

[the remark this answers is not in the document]
balajiv113: I was referring to more of creating a struct itself with the following,
```go
type SnapshotInfo struct {
id string
tag string
vmSize string
date string
vmClock string
}
```
As part of QEMU driver we will do the exact same like we done to fetch tag, but we will do for all needed fields. After that we can return SnapshotInfo[] as part of the command.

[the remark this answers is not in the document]
afbjorklund: There really is no consistent output to parse, it is intended to be read by human.

Output while running:

```
INFO[0000] Sending HMP [info snapshots] command         
There is no snapshot available.

INFO[0000] Sending HMP [info snapshots] command         
List of snapshots present on all disks:
ID        TAG                     VM SIZE                DATE       VM CLOCK
--        snap1                  2.33 GiB 2023-02-26 18:16:36   00:06:18.553
```

Output while stopped:

```
DEBU[0000] Running qemu-img [snapshot -l /home/anders/.lima/default/diffdisk] command 

DEBU[0000] Running qemu-img [snapshot -l /home/anders/.lima/default/diffdisk] command 
ID        TAG                     VM SIZE                DATE       VM CLOCK
1         snap1                  2.33 GiB 2023-02-26 18:16:36   00:06:18.553
```

Will need to filter out the header bits, but it still seems a bit error prone.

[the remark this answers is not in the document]
balajiv113: Sure. Thanks. We can ignore this in that case

AkihiroSuda:
Sorry needs rebase again

AkihiroSuda (dismissed):
Thanks

AkihiroSuda:
CI failing
```
+ set +x
TEST| [INFO] Testing online snapshots
time="2023-05-09T05:21:18Z" level=info msg="Sending HMP savevm command"
time="2023-05-09T05:21:26Z" level=info msg="Sending HMP info command"
TEST| [INFO] snapshot list: expected=snap1 got=snap1
time="2023-05-09T05:21:26Z" level=info msg="Sending HMP loadvm command"

Timed out!
```

balajiv113 (dismissed):
LGTM :)

AkihiroSuda:
CI for `experimental/net-user-v2.yaml` seems consistenly failing

afbjorklund:
> CI for `experimental/net-user-v2.yaml` seems consistenly failing

Looks like the net-user does **not** survive an "online" VM restore.

So it needs to be excluded.

AkihiroSuda (approved):
thanks

PR #1065, opened 2022-09-23
Add `disk` command and allows multiple additional disks

sam-berning:
https://github.com/lima-vm/lima/issues/722

After looking through the discussion on https://github.com/lima-vm/lima/pull/759, I implemented a basic volume command. Volumes are stored under a `_volumes` directory under `$LIMA_HOME` and can be attached by name to an instance in its LimaYAML.

Also added a boot script that performs some setup for volumes. So far, it creates a filesystem labeled with the volume name if it's the first time the volume has been attached to an instance. In the future, we could expand it to bind-mount the volume to specific directories.


examples/default.yaml
```diff
@@ -353,6 +353,14 @@ hostResolver:
 # - 1.1.1.1
 # - 1.0.0.1
 
+# Lima volumes to attach to the instance. The volumes will be accessible from inside
+# the instance, labeled by name. (e.g. if the volume is named "data", it will be labeled
+# "lima-data" inside the instance).
+# 🟢 Builtin default: null
+volumes:
+# volumes should be a list of volume name strings, for example:
+# - "data"
```
AkihiroSuda, line 362: Perhaps this should be added near `mounts`?

pkg/store/filenames/filenames.go
```diff
@@ -47,6 +48,12 @@ const (
 	SocketDir = "sock"
 )
 
+// Filenames used under a volume directory
+
+const (
+	DataDisk = "datadisk"
```
AkihiroSuda, line 54: Please update `docs/internal.md` to explain the new files and directories
https://github.com/lima-vm/lima/blob/master/docs/internal.md#instance-directory-lima_homeinstance

cmd/limactl/volume.go
```diff
@@ -0,0 +1,184 @@
...
+		Use:   "volume",
+		Short: "Lima volume management",
+		Example: `  Create a volume:
+  $ limactl volume create data_volume
+
+  List existing volumes:
+  $ limactl volume ls
+
+  Delete a volume:
+  $ limactl volume delete data_volume`,
```
AkihiroSuda, line 28: Please update the command reference https://github.com/lima-vm/lima#command-reference

AkihiroSuda: And also needs CI
https://github.com/lima-vm/lima/blob/master/hack/test-example.sh

pkg/cidata/cidata.TEMPLATE.d/boot/05-lima-volumes.sh
```diff
@@ -0,0 +1,22 @@
...
+}
+
+for i in $(seq 0 $((LIMA_CIDATA_VOLUMES - 1))); do
+  VOLUME_NAME="$(get_volume_var $i "NAME")"
+  DEVICE_NAME="$(get_volume_var $i "DEVICE")"
+
+  # first time setup
+  if [[ ! -b "/dev/disk/by-label/lima-${VOLUME_NAME}" ]]; then
+    # TODO: skip if volume is tagged as "raw"
+    echo 'type=83' | sfdisk --label dos "/dev/${DEVICE_NAME}"
```
AkihiroSuda, line 19: Please add a comment line to explain this magic value

AkihiroSuda, line 19: Can we use GPT?

examples/default.yaml
```diff
@@ -353,6 +353,14 @@ hostResolver:
 # - 1.1.1.1
 # - 1.0.0.1
 
+# Lima volumes to attach to the instance. The volumes will be accessible from inside
+# the instance, labeled by name. (e.g. if the volume is named "data", it will be labeled
+# "lima-data" inside the instance).
```
AkihiroSuda, line 358: Where is the mount point of the volume?

pkg/qemu/qemu.go
```diff
@@ -437,6 +465,10 @@ func Cmdline(cfg Config) (string, []string, error) {
 	} else if !isBaseDiskCDROM {
 		args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio", baseDisk))
 	}
+	for _, volDisk := range volumeDisks {
+		args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio", volDisk))
```
AkihiroSuda, line 469: Does this lock the file so that the same volume can't be written by multiple instances simultaneously?

examples/default.yaml
sam-berning: Will add in the next commit. Planning on mounting them under `/mnt/lima-${VOLUME_NAME}`

pkg/qemu/qemu.go
sam-berning: No, it doesn't. Just tested attaching a volume to two instances at once and writing to it from each of the instances - they overwrote each other and corrupted the volume. I'll introduce some sort of locking mechanism

pkg/cidata/cidata.TEMPLATE.d/boot/05-lima-volumes.sh
sam-berning: I think I'll just replace the magic value with the `linux` alias as per the [manpage](https://man7.org/linux/man-pages/man8/sfdisk.8.html), any concerns?

jandubois:
Please fix linter errors (run `make lint` locally to verify).

cmd/limactl/volume.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+		if vol.Instance == "" {
+			logrus.Warnf("Ignoring unlocked volume %q", volName)
+			continue
+		}
+		inst, err := store.Inspect(vol.Instance)
+		if err != nil {
+			logrus.Warnf("Could not inspect instance %q using volume %q", vol.Instance, volName)
+			continue
+		}
```
jandubois, line 224: This is not correct. When `store.Inspect()` returns an `err`, it means the instance does not exist. In that case we need to skip the check if the instance is running and go straight to unlocking the volume.

This can happen e.g. when you `limactl delete -f INSTANCE`.

I would argue that force-deleting an instance should still unlock the volumes too, but the logic here should be fixed anyways, in case there are other ways to end up with a deleted instance that leaves the symlinks behind.

If there are any errors besides the instance not existing, then they are returned in `inst.Errors`. If that slice is non-empty, then the errors should be reported here, and the unlock should be skipped.

pkg/hostagent/hostagent.go
```diff
@@ -466,6 +466,22 @@ func (a *HostAgent) startHostAgentRoutines(ctx context.Context) error {
...
+	if len(a.y.Volumes) > 0 {
+		a.onClose = append(a.onClose, func() error {
+			var unlockMErr error
+			for _, v := range a.y.Volumes {
+				vol, inspectErr := store.InspectVolume(v)
+				if inspectErr != nil {
+					unlockMErr = multierror.Append(unlockMErr, inspectErr)
+					continue
+				}
+				if unlockErr := vol.Unlock(); unlockErr != nil {
```
jandubois, line 478: For symmetry with the mounts it would be nice to provide logging for mounting/unmounting volumes too.

```console
$ l start default
[...]
INFO[0022] [hostagent] Mounting "/Users/jan" on "/Users/jan"
INFO[0022] [hostagent] Mounting "/tmp/lima" on "/tmp/lima"

$ l stop default
[...]
INFO[0000] [hostagent] Unmounting "/Users/jan"
INFO[0000] [hostagent] Unmounting "/tmp/lima"
```
There is no corresponding logging for volumes right now.

For mounts the logging happens in `m.Close()`. I'm not sure if the logging should happen in `vol.Unlock()` or explicitly here in the hostagent code. The rest of `pkg/store` is not emitting any logs because it is low-level code.

cmd/limactl/volume.go
```diff
@@ -0,0 +1,235 @@
...
+		Short: "Create a Lima volume",
+		Args:  cobra.ExactArgs(1),
+		RunE:  volumeCreateAction,
+	}
+	volumeCreateCommand.Flags().String("size", "2G", "configure the volume size")
+	return volumeCreateCommand
+}
+
+func volumeCreateAction(cmd *cobra.Command, args []string) error {
+	_, err := createVolume(cmd, args)
```
jandubois, line 57: Why do you need a separate `createVolume()` function that just returns an extra value that is being ignored here.
The function isn't called from anywhere else, so why isn't it inlined here instead?

cmd/limactl/volume.go
```diff
@@ -0,0 +1,235 @@
...
+	diskSize, err := units.RAMInBytes(size)
+	if err != nil {
+		return nil, err
+	}
+
+	if err := qemu.CreateDataDisk(volDir, int(diskSize)); err != nil {
+		return nil, err
+	}
+
+	vol, err := store.InspectVolume(name)
```
jandubois, line 97: I don't think there is any need to call `InspectVolume` here; the volume has just been created, without any errors, so why should the `Inspect` fail already.

cmd/limactl/volume.go
```diff
@@ -0,0 +1,235 @@
...
+		Example: `
+To list existing volumes:
+$ limactl volume list
+`,
+		Short:   "List existing Lima volumes",
+		Aliases: []string{"ls"},
+		Args:    cobra.NoArgs,
+		RunE:    volumeListAction,
+	}
+	return volumeListCommand
```
jandubois, line 116: This can of course be added later, but it would be nice to have a `--json` flag, just like on the main `list` command:

```console
$ l ls --json | jq
{
  "name": "alpine",
  "status": "Stopped",
  "dir": "/Users/jan/.lima/alpine",
  "arch": "x86_64",
  "cpuType": "host",
  "cpus": 4,
  "memory": 4294967296,
  "disk": 107374182400
}
```

cmd/limactl/volume.go
```diff
@@ -0,0 +1,235 @@
...
+	for _, volName := range args {
+		vol, err := store.InspectVolume(volName)
+		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				logrus.Warnf("Ignoring non-existent volume %q", volName)
+				continue
+			}
+			return err
+		}
+		if err := deleteVolume(vol); err != nil {
```
jandubois, line 172: I think you should call `LockVolume` first to make sure you are not deleting a volume that is in use by a running instance.

cmd/limactl/volume.go
```diff
@@ -0,0 +1,235 @@
...
+		Aliases: []string{"remove", "rm"},
+		Short:   "Delete one or more Lima volumes",
+		Args:    cobra.MinimumNArgs(1),
+		RunE:    volumeDeleteAction,
+	}
+	return volumeDeleteCommand
+}
+
+func volumeDeleteAction(cmd *cobra.Command, args []string) error {
+	for _, volName := range args {
```
jandubois, line 163: I wonder if it makes sense to read the `lima.yaml` files of **all** instances, to see if a volume is still referenced by any stopped instance, and then ask for confirmation before deleting it (with `limactl volume rm -f` overriding the check).

The networking code does a similar thing to shut down managed network daemons when no running instance is using the specific network anymore, so there is precedence for considering all instances.

pkg/store/volume.go
```diff
@@ -0,0 +1,69 @@
...
+type Volume struct {
+	Name        string `json:"name"`
+	Size        int64  `json:"size"`
+	Dir         string `json:"dir"`
+	DiskPath    string `json:"diskPath"`
+	Instance    string `json:"instance"`
+	InstanceDir string `json:"instanceDir"`
+}
+
+func InspectVolume(volName string) (*Volume, error) {
```
jandubois, line 22: I wonder if it would make sense to follow the same API as the `store.Inspect()` function: to only return an error if the volume doesn't exists, and otherwise collecting errors in a `volume.Errors` slice.

It isn't really needed here, but it might be nice for consistency.

jandubois (commented):
Thank you, this is very useful.  I'm done with this round of reviewing, but have some feedback that should be addressed.

README.md
```diff
@@ -224,6 +224,14 @@ Use `<INSTANCE>:<FILENAME>` to specify a source or target inside an instance.
 #### `limactl edit`
 `limactl edit <INSTANCE>`: edit the instance
 
+#### `limactl volume`
+
```
AkihiroSuda, line 228: Needs an explanation of "volume". (filesystem, mount point, size, ...)

Probably it should be also noted that this is unrelated to `nerdctl volume` volumes.
(So, `limactl storage` or `limactl disk` might be less confusing name?)

docs/internal.md
```diff
@@ -59,6 +59,16 @@ Host agent:
...
 
+## Volume directory (`${LIMA_HOME}/_volumes/<VOLUME>`)
+
+A volume directory contains the following files:
+
+disk:
+- `datadisk`: the qcow2 disk that is attached to an instance when the volume is attached to an instance
+
+lock:
+- `in_use_by`: symlink to the instance directory that is using the volume
```
AkihiroSuda, line 70: relative symlink or absolute symlink?

pkg/store/volume.go
```diff
@@ -0,0 +1,69 @@
...
+		vol.Instance = filepath.Base(instDir)
+		vol.InstanceDir = instDir
+	}
+
+	return vol, nil
+}
+
+func (v *Volume) Lock(instanceDir string) error {
+	inUseBy := filepath.Join(v.Dir, filenames.InUseBy)
+	return os.Symlink(instanceDir, inUseBy)
```
AkihiroSuda, line 63: Shouldn't we use flock?

README.md
jandubois: > Needs an explanation of "volume". (filesystem, mount point, size, ...)

Yes, but not necessarily in the `README.md`. The other commands also include only the most common option and an abbreviated description. The question is where else it should go; ideally we would have some proper documentation, but who is going to write it?

> Probably it should be also noted that this is unrelated to nerdctl volume volumes.
> (So, limactl storage or limactl disk might be less confusing name?)

I don't know; "volume" was the first choice for this functionality for me too, because that's what it is.

But I understand the risk of confusion; just like people get confused between nerdctl namespaces and kubernetes namespaces.

I'm still fine with `volume` but would prefer `disk` over `storage`.

docs/internal.md
jandubois: It is an absolute symlink, which I think is good, but not sure if it matters.

pkg/store/volume.go
jandubois: We could. `flock` is just advisory, but since this is only checked by lima, that doesn't matter. The advantage of the symlink is that the `basename(symlink)` gives us the instance name, so we can see which instance holds the lock.

examples/default.yaml
jandubois: The mount point should be documented right here in `default.yaml`.

pkg/qemu/qemu.go
```diff
@@ -437,6 +475,10 @@ func Cmdline(cfg Config) (string, []string, error) {
...
```
jandubois, line 479: Needs to add `discard=on` option to enable `fstrim` support. See https://github.com/lima-vm/lima/pull/1102/files.

```suggestion
		args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio,discard=on", volDisk))
```

docs/internal.md
sam-berning: I agree that an absolute symlink is preferable. This would let us further separate volume storage from instance storage if we wanted to (e.g. a separate `$LIMA_VOLUME_HOME`)

README.md
sam-berning: I'd be ok with changing the naming to `disk`. I agree that `volume` would be better in isolation but I think it's likely this overlap will confuse a few people. I'll update in the next commit.

cmd/limactl/volume.go
jandubois: I just realized that `LockVolume` would require an instance here, so `vol.InspectVolume()` should be called instead and **must** return an error before you can call `deleteVolume()`.

docs/internal.md
jandubois: > I agree that an absolute symlink is preferable.

It is already an absolute symlink iirc.

> This would let us further separate volume storage from instance storage if we wanted to (e.g. a separate $LIMA_VOLUME_HOME)

I can't really see the use case for this. If we wanted to do this at all, I would rather be able to specify a full pathname for the volume in `limactl volume create`, and then put a symlink to that file into `$LIMA_HOME/_volumes/$VOLUME`. That way you have total flexibility on where the volumes are stored, and you can also easily move them manually yourself.

Doing the manual moving should already work right now with this PR without further code changes.

jandubois:
Please rebase the PR to resolve merge conflicts!

cmd/limactl/disk.go
```diff
@@ -0,0 +1,307 @@
...
+		Use: "create DISK",
+		Example: `
+To create a new disk:
+$ limactl disk create DISK
+`,
+		Short: "Create a Lima disk",
+		Args:  cobra.ExactArgs(1),
+		RunE:  diskCreateAction,
+	}
+	diskCreateCommand.Flags().String("size", "2G", "configure the disk size")
```
AkihiroSuda, line 53: Isn't this too small?

AkihiroSuda:
Please squash commits and make sure to keep the PR title is consistent with the actual code

cmd/limactl/volume.go
jandubois: > When store.Inspect() returns an err, it means the instance does not exist. In that case we need to skip the check if the instance is running and go straight to unlocking the volume.

Thank you for fixing this!

> I would argue that force-deleting an instance should still unlock the volumes too, but the logic here should be fixed anyways, in case there are other ways to end up with a deleted instance that leaves the symlinks behind.

Please edit `stopInstanceForcibly` in `cmd/limactl/stop.go` and unlock all disks locked by the instance being killed!

README.md
jandubois: I know we have already all agreed on `disk`, but I only now noticed that this means we now have both a `disk` and a `disks` key in `lima.yaml`, which is confusing.

if we are sticking with `disk` for the concept (I'm still fine with using `volume`), then maybe we should at least change the key in `lima.yaml` to `dataDisks`, to separate it from `disk` which really should have been `diskSize`.

@AkihiroSuda What do you think?

AkihiroSuda: What about `additionalDisks`?

cmd/limactl/disk.go
jandubois: > Isn't this too small?

I agree that it sounds small, but I don't know how to pick a reasonable default size.

Given that this is just the maximum size of a sparse file, there is not much harm in picking a larger value, but which size should be "large enough" for most use cases?

Maybe `10G`? But it still is pretty arbitrary.

sam-berning: Maybe half the default size of the primary disk? `50G`?

chancez: Every disk needs a size, why not make it empty and make it a required flag?

cmd/limactl/disk.go
```diff
@@ -0,0 +1,305 @@
...
+	)
+	return diskCommand
+}
+
+func newDiskCreateCommand() *cobra.Command {
+	var diskCreateCommand = &cobra.Command{
+		Use: "create DISK",
+		Example: `
+To create a new disk:
+$ limactl disk create DISK
```
AkihiroSuda, line 47: The `size` has to be now explicitly specified?

cmd/limactl/disk.go
```diff
@@ -0,0 +1,305 @@
...
+		return err
+	}
+
+	diskSize, err := units.RAMInBytes(size)
+	if err != nil {
+		return err
+	}
+
+	// only exactly one arg is allowed
+	name := args[0]
```
AkihiroSuda, line 70: Perhaps make sure that the `name` does not contain `os.PathSeparator`

examples/disk.yaml
```diff
@@ -0,0 +1,25 @@
+# This example requires Lima v0.7.0 or later.
```
AkihiroSuda, line 1: v0.14.0

examples/disk.yaml
```diff
@@ -0,0 +1,25 @@
...
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img"
+  arch: "aarch64"
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+
+# in order to use this example, you must first create the disk "data". run:
+# $ limactl disk create data
```
AkihiroSuda, line 23: Needs size

jandubois:
This PR also needs to be rebased on latest `master` so it can pass integration tests.

cmd/limactl/disk.go
sam-berning: This should be handled by `identifiers.Validate()` in `store.DiskDir()`

AkihiroSuda (dismissed):
Thanks, LGTM but needs rebase again.

The locking mechanism seems still racy, but can be addressed in another PR

AkihiroSuda (approved):

pkg/store/disk.go
```diff
@@ -0,0 +1,71 @@
...
+	disk.Size = info.VSize
+
+	instDir, err := os.Readlink(filepath.Join(diskDir, filenames.InUseBy))
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			disk.Instance = ""
+			disk.InstanceDir = ""
+		} else {
+			return nil, err
+		}
```
jandubois, line 52: The struct is already initialized with the null values, so the explicit assignment seems redundant:

```suggestion
		if !errors.Is(err, fs.ErrNotExist) {
			return nil, err
		}
```

cmd/limactl/disk.go
```diff
@@ -0,0 +1,305 @@
...
+	diskDir, err := store.DiskDir(name)
+	if err != nil {
+		return err
+	}
+
+	if _, err := os.Stat(diskDir); !errors.Is(err, fs.ErrNotExist) {
+		return fmt.Errorf("disk %q already exists (%q)", name, diskDir)
+	}
+
+	logrus.Infof("Creating a disk %q", name)
```
jandubois, line 81: Maybe include the disk size in the log, so show how the size was parsed:

```suggestion
	logrus.Infof("Creating disk %q size %s", name, units.BytesSize(float64(diskSize)))
```

cmd/limactl/disk.go
```diff
@@ -0,0 +1,305 @@
...
+func newDiskListCommand() *cobra.Command {
+	var diskListCommand = &cobra.Command{
+		Use: "list",
+		Example: `
+To list existing disks:
+$ limactl disk list
+`,
+		Short:   "List existing Lima disks",
+		Aliases: []string{"ls"},
+		Args:    cobra.NoArgs,
```
jandubois, line 103: Thought for a later PR (low priority):

I think it would be good to allow disk names to be passed as args (for symmetry with `limactl ls`, which allows instance names), and then only emit the data for those disk images.

cmd/limactl/disk.go
```diff
@@ -0,0 +1,305 @@
...
+
+To delete multiple disks:
+$ limactl disk delete DISK1 DISK2 ...
+`,
+		Aliases: []string{"remove", "rm"},
+		Short:   "Delete one or more Lima disks",
+		Args:    cobra.MinimumNArgs(1),
+		RunE:    diskDeleteAction,
+	}
+	diskDeleteCommand.Flags().Bool("force", false, "force delete")
```
jandubois, line 171: The option should allow `-f` shorthand, just like `limactl delete -f INSTANCE`:

```suggestion
	diskDeleteCommand.Flags().BoolP("force", "f", false, "force delete")
```

cmd/limactl/disk.go
```diff
@@ -0,0 +1,305 @@
...
+		disk, err := store.InspectDisk(diskName)
+		if err != nil {
+			if errors.Is(err, fs.ErrNotExist) {
+				logrus.Warnf("Ignoring non-existent disk %q", diskName)
+				continue
+			}
+			return err
+		}
+		if disk.Instance != "" {
+			return fmt.Errorf("cannot delete disk %q in use by instance %q", disk.Name, disk.Instance)
```
jandubois, line 208: It feels inconsistent to me that a locked disk will abort the command, but a non-existing disk is simply skipped:

```
limactl disk rm foo bar
```

If `foo` doesn't exist, `bar` will still be deleted. But if `foo` exists and is locked, then `bar` will not be deleted (and there is no message to the effect that additional disks have been skipped due to the error).

I think this the locked disk should be logged as an error, but additional disks should still be processed.

And then I wonder if failures if `deleteDisk` should be treated the same way too: log an error, but continue to attempt deleting the remaining disks?

Of course the command must exit with a non-zero status code if any disk couldn't be deleted. I'm not sure how non-existing disks should be treated: should they also cause a non-zero error?

@AkihiroSuda WDYT?

cmd/limactl/disk.go
```diff
@@ -0,0 +1,305 @@
...
+		}
+		instances, err := store.Instances()
+		if err != nil {
+			return err
+		}
```
jandubois, line 213: This should be moved outside the `for` loop:

```
		instances, err := store.Instances()
		if err != nil {
			return err
		}
		for _, diskName := range args {
			...
		}
```

Although the "expensive" operation is `store.Inspect(instName)`, so those would have to be cached outside the loop as well. Maybe not worth the extra code.

hack/test-example.sh
```diff
@@ -78,6 +82,14 @@ export ftp_proxy=http://localhost:2121
 
 INFO "Starting \"$NAME\" from \"$FILE\""
 defer "limactl delete -f \"$NAME\""
+
+if [[ -n ${CHECKS["disk"]} ]]; then
+	if ! limactl disk ls | grep -q "^data\s"; then
```
jandubois, line 87: This is an example where it would be useful to be able to pass disk names to the `ls` command:

```
if ! limactl disk ls data; then
   ...
fi
```

pkg/qemu/qemu.go
```diff
@@ -422,6 +438,30 @@ func Cmdline(cfg Config) (string, []string, error) {
...
+				logrus.Errorf("could not load disk %q: %q", diskName, err)
+				return "", nil, err
+			}
+
+			if d.Instance != "" {
+				logrus.Errorf("could not attach disk %q, in use by instance %q", diskName, d.Instance)
+				return "", nil, err
+			}
+			logrus.Infof("Mounting disk %q on %q", diskName, d.MountPoint)
+			err = d.Lock(cfg.InstanceDir)
```
jandubois, line 455: If there are any errors from here until `ha.Run()` is called, then we won't unlock the disk and the user will have to manually `limactl disk unlock DISK` before being able to use it again. It would be better to have some "unlock on error" handling somewhere in `hostagentAction`.

jandubois (approved):
I've "resolved" all existing discussions, as they all seem to have been implemented.

I've done one more round of reviews, and found a few minor issues. I'm fine with merging this PR as-is, and addressing the issues in a follow, because we have a bunch of large PRs piled up now, and we will get more and more merge conflicts until we start merging the big pieces.

@sam-berning So I'm going to merge this PR, but please look at all unresolved feedback, and either address in a follow-PR, or let me know, and I'll create an open issue for it!

cmd/limactl/disk.go
sam-berning: Can you open an issue for this? I think we should keep this behavior consistent between `limactl disk delete` and `limactl delete`, and I think this needs some more discussion.

PR #1069, opened 2022-09-27
 simplify `FillDefault()` using reflects

AkihiroSuda:
The type of ` CPUType  map[Arch]string` is now changed to `map[Arch]*string`, as `pkg/reflectutil.Merge()` is designed to distinguish "empty" from "nil".

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -299,18 +299,6 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 		// After defaults processing the singular HostPort and GuestPort values should not be used again.
 	}
 
-	// If both `useHostResolved` and `HostResolver.Enabled` are defined in the same config,
-	// then the deprecated `useHostResolved` setting is silently ignored.
-	if y.HostResolver.Enabled == nil {
-		y.HostResolver.Enabled = y.UseHostResolver
-	}
-	if d.HostResolver.Enabled == nil {
-		d.HostResolver.Enabled = d.UseHostResolver
-	}
-	if o.HostResolver.Enabled == nil {
-		o.HostResolver.Enabled = o.UseHostResolver
-	}
-
 	if y.HostResolver.Enabled == nil {
 		y.HostResolver.Enabled = d.HostResolver.Enabled
 	}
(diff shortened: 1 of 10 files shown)


pkg/reflectutil/reflectutil.go
```diff
@@ -0,0 +1,104 @@
...
+func MergeMany(vv ...interface{}) (interface{}, error) {
+	if l := len(vv); l < 2 {
+		return nil, fmt.Errorf("expected len(vv) >= 2, got %d", l)
+	}
+	x := vv[0]
+	for _, v := range vv[1:] {
+		var err error
+		x, err = Merge(x, v)
+		if err != nil {
+			return x, err
```
jandubois, line 30: Don't return incompletely merged result in case of error:

```suggestion
			return nil, err
```

pkg/reflectutil/reflectutil_test.go
```diff
@@ -0,0 +1,154 @@
...
+	logX(t, x, "x")
+	assert.DeepEqual(t, expected, x)
+}
+
+func logX(t testing.TB, x interface{}, format string, args ...interface{}) {
+	// Print in JSON for human readability
+	j, err := json.Marshal(x)
+	assert.NilError(t, err)
+	t.Logf(format+": %s", append(args, string(j))...)
+}
```
jandubois, line 154: `args` is never used; `format` is really just a name:

```suggestion
func logX(t testing.TB, x interface{}, name string) {
	// Print in JSON for human readability
	j, err := json.Marshal(x)
	assert.NilError(t, err)
	t.Logf("%s: %s", name, string(j))
}
```

Personally I would also swap the order of `x` and `name`, but it doesn't really matter.

pkg/limayaml/defaults.go
```diff
@@ -76,343 +125,23 @@ func MACAddress(uniqueID string) string {
...
-		if nw.MACAddress == "" {
-			// every interface in every limayaml file must get its own unique MAC address
-			nw.MACAddress = MACAddress(fmt.Sprintf("%s#%d", filePath, i))
-		}
-		if nw.Interface == "" {
-			nw.Interface = "lima" + strconv.Itoa(i)
-		}
-	}
+	// *EXCEPTION*: Mounts are appended in d, y, o order, but "merged" when the Location matches a previous entry;
+	//  the highest priority Writable setting wins.
```
jandubois, line 144: Just below is this block:

```go
	mounts := make([]Mount, 0, len(d.Mounts)+len(y.Mounts)+len(o.Mounts))
	location := make(map[string]int)
	for _, mount := range append(append(d.Mounts, y.Mounts...), o.Mounts...) {
```

It should have been updated to no longer use `d`, `y`, and `o`, but just `x`:

```go
	mounts := make([]Mount, 0, len(x.Mounts))
	location := make(map[string]int)
	for _, mount := range x.Mounts {
```

jandubois: Same thing again further down for `networks`.

pkg/limayaml/defaults.go
```diff
@@ -489,51 +218,144 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
 
-	if y.MountType == nil {
-		y.MountType = d.MountType
-	}
-	if o.MountType != nil {
-		y.MountType = o.MountType
+	// *EXCEPTION*: Networks are appended in d, y, o order
+	networks := make([]Network, 0, len(d.Networks)+len(y.Networks)+len(o.Networks))
+	iface := make(map[string]int)
+	for _, nw := range append(append(d.Networks, y.Networks...), o.Networks...) {
```
jandubois, line 224:
```suggestion
	networks := make([]Network, 0, len(x.Networks))
	iface := make(map[string]int)
	for _, nw := range x.Networks {
```

PR #1072, opened 2022-09-27
docs: why one fs driver over another

ghost:
I loved reading the mount documentation.

I found myself asking why should I care about one implementation over the other.

So here is that internal conversation!

### DIFF
--- a/docs/mount.md
+++ b/docs/mount.md
@@ -10,6 +10,13 @@ The default mount type is shown in the following table:
 | >= 0.10          | reverse-sshfs + OpenSSH SFTP server |
 | >= 1.0 (Planned) | 9p                                  |
 
+## Which mount type should I pick? Or why care?
+
+sshfs leverages mature networking tools to make a reliable filesystem mount.
+Leaning on the classics.
+9p implements innovative ideas from the experimental [Plan 9 operating system](https://9p.io/wiki/plan9/Overview/index.html).
+Exploring new horizons.
+
 ## Mount types
 
 ### reverse-sshfs


docs/mount.md
```diff
@@ -10,6 +10,13 @@ The default mount type is shown in the following table:
 | >= 0.10          | reverse-sshfs + OpenSSH SFTP server |
 | >= 1.0 (Planned) | 9p                                  |
 
+## Which mount type should I pick? Or why care?
+
+sshfs leverages mature networking tools to make a reliable filesystem mount.
+Leaning on the classics.
+9p implements innovative ideas from the experimental [Plan 9 operating system](https://9p.io/wiki/plan9/Overview/index.html).
+Exploring new horizons.
```
AkihiroSuda, line 18: 9p is a classic protocol and actually even much older than sshfs 
https://pdos.csail.mit.edu/6.828/2007/lec/l-plan9.html

docs/mount.md
```diff
@@ -10,6 +10,13 @@ The default mount type is shown in the following table:
 | >= 0.10          | reverse-sshfs + OpenSSH SFTP server |
 | >= 1.0 (Planned) | 9p                                  |
 
+## Which mount type should I pick? Or why care?
+
+sshfs leverages mature networking tools to make a reliable filesystem mount.
```
AkihiroSuda, line 15: What is “reliable”? “mature”?

AkihiroSuda (changes requested):
Sorry, disagree.

The advantage of 9p is that the mount can be kept even if the SSH connection was shutdown as long as the virtio-9p-pci device is functional.
(This is actually an advantage of virtio-9p-pci rather than 9p itself.)

The advantage of sshfs is that it works on RHEL-derived distros.

PR #1076, opened 2022-09-29
templates: update apptainer to 1.1.0

afbjorklund:
Now has support for rootless containers (user namespaces)

https://github.com/apptainer/apptainer/releases/tag/v1.1.0

Add support for aarch64

Currently only from EPEL

### DIFF
--- a/examples/experimental/apptainer.yaml
+++ b/examples/experimental/apptainer.yaml
@@ -8,6 +8,9 @@ images:
 - location: "https://dl.rockylinux.org/pub/rocky/8.6/images/Rocky-8-GenericCloud-8.6.20220702.0.x86_64.qcow2"
   arch: "x86_64"
   digest: "sha256:7b786a39eeb96e22dd85386377ff186737f6c1b9a5faa105b0a0a7a4895c29d0"
+- location: "https://dl.rockylinux.org/pub/rocky/8.6/images/Rocky-8-GenericCloud-8.6.20220702.0.aarch64.qcow2"
+  arch: "aarch64"
+  digest: "sha256:42da3cc0c10677d37e43f862c4a8f3ab3b72bcccc91ce0c0ef9f4100f4b950b4"
 mounts:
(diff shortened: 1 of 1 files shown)


afbjorklund:
The Rocky download is currently 2.5 GiB in size :-(

But it is from the same vendor, so seemed natural...

https://ciq.co/

Also, that problem is more for examples/rocky.yaml ?

afbjorklund:
Seems to work OK. Also tested without suid, and that looks ok as well.

```console
$ limactl shell apptainer apptainer run -u -B $HOME:$HOME docker://alpine
INFO:    Converting OCI blobs to SIF format
INFO:    Starting build...
Copying blob 213ec9aee27d done  
Copying config 29f453b10b done  
Storing signatures
2022/09/29 20:13:26  info unpack layer: sha256:213ec9aee27d8be045c6a92b7eac22c9a64b44558193775a1a7f626352392b49
INFO:    Creating SIF file...
Apptainer> more /etc/os-release 
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.16.2
PRETTY_NAME="Alpine Linux v3.16"
HOME_URL="https://alpinelinux.org/"
BUG_REPORT_URL="https://gitlab.alpinelinux.org/alpine/aports/-/issues"
Apptainer> exit
$ limactl shell apptainer apptainer --version
apptainer version 1.1.0-1.el8
```

AkihiroSuda:
Next time please consider squashing commits

afbjorklund:
I squashed the relayed commits (one version bump, one new arch)

afbjorklund:
This example will get replaced by the much simpler Fedora example:

* #1081

It has the same software, but is smaller download and shorter yaml...

Currently at 1.0.3, but 1.1.0 is in testing: [FEDORA-2022-038f1a2ce3](https://bodhi.fedoraproject.org/updates/FEDORA-2022-038f1a2ce3)

PR #1079, opened 2022-09-30
Fixing TruncateReply option passing

vmasutin:
Fixing TruncateReply option passing, fixes issues with oauth2.googleapis.com dns resolution

Signed-off-by: Valery Masiutsin <val.masutin@gmail.com>

### DIFF
--- a/pkg/hostagent/dns/dns.go
+++ b/pkg/hostagent/dns/dns.go
@@ -124,6 +124,7 @@ func NewHandler(opts HandlerOptions) (dns.Handler, error) {
 		{Net: "tcp"},
 	}
 	h := &Handler{
+		truncate:     opts.TruncateReply,
 		clientConfig: cc,
 		clients:      clients,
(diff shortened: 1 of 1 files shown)


pkg/hostagent/dns/dns.go
```diff
@@ -368,7 +369,7 @@ func listenAndServe(network Network, opts ServerOptions) (*dns.Server, error) {
 	var addr string
 	// always enable reply truncate for UDP
 	if network == UDP {
-		opts.TruncateReply = true
+		opts.HandlerOptions.TruncateReply = true
```
Nino-K, line 372: Nit: no need for `HandlerOptions` since it is embeded.

jandubois (approved):
Thank you for catching and fixing this!

pkg/hostagent/dns/dns.go
jandubois: It is not needed, but it doesn't hurt either.

PR #1081, opened 2022-10-01
The singularity package was renamed to apptainer

afbjorklund:
In fedora, singularity-3.8.7 was replaced with apptainer-1.0.3

https://singularity.hpcng.org/

https://apptainer.org/news/community-announcement-20211130

So the example should be renamed too, to reflect the reality.

----

SingularityCE has **not** been packaged as dnf, only from sylabs.io

https://sylabs.io/singularity/

https://docs.sylabs.io/guides/3.10/admin-guide/

If requested, it _could_ be resurrected as an example (like e.g. centos stream)


afbjorklund:
Once the fedora package has been updated to 1.1.0, this PR replaces the experimental/apptainer.yaml example.

I believe that using Fedora is a better match than using Rocky, because of 1) download size 2) uses EPEL anyway

afbjorklund:
Currently there are problems with using tmpfs as /tmp, but a simple workaround is `TMPDIR=/var/tmp`

I will do a separate PR for this, once renamed... https://fedoraproject.org/wiki/Features/tmp-on-tmpfs

AkihiroSuda:
> Once the fedora package has been updated to 1.1.0, this PR replaces the experimental/apptainer.yaml example.

How long will it take?
Shall we wait until that happens or shall we merge this PR now?

afbjorklund:
> Shall we wait until that happens or shall we merge this PR now?

I think the only difference was that `apptainer-suid` was split into a separate package:

```diff
-    dnf install --enablerepo=epel-testing -y apptainer apptainer-suid
+    dnf install -y apptainer
```

I tried enabling the updates-testing, but seems like the package only made it to fc38 ?

https://src.fedoraproject.org/rpms/apptainer

But you could merge it now, and we could tweak it later perhaps. It doesn't _change_ anything.

The previous example has the same end result, with a symlink `/usr/bin/singularity -> apptainer`

afbjorklund:
@AkihiroSuda note that this also a new feature, in the new release: i.e. rootless by default

https://apptainer.org/news/apptainer-1-1-0-20220927

It is still possible to use rootless, by using flags or by changing the apptainer.conf configuration:

`allow setuid = no`

AkihiroSuda:
If we are going to merge this now, shall we remove `experimental/apptainer.yaml`?

afbjorklund:
We need to wait until it has merged from "testing" into "stable", otherwise we need a similar workaround:

```
sudo dnf upgrade --enablerepo=updates-testing --refresh --advisory=FEDORA-2022-038f1a2ce3
```

https://bodhi.fedoraproject.org/updates/FEDORA-2022-038f1a2ce3

But after that (1.1.0), we could provide one apptainer.yaml and one apptainer-rootful.yaml, like the others ?

afbjorklund:
@AkihiroSuda apptainer 1.1.0 has now been accepted in both Fedora and EPEL

https://src.fedoraproject.org/rpms/apptainer



Release | Stable version | Version in testing
-- | -- | --
Fedora 36 | apptainer-1.1.0-1.fc36 |  
Fedora 35 | apptainer-1.1.0-1.fc35 |  
Fedora EPEL 9 | apptainer-1.1.0-1.el9 |  
Fedora EPEL 8 | apptainer-1.1.0-1.el8

afbjorklund:
The rootless example doesn't need squashfs anymore, since it uses squashfuse instead.

AkihiroSuda:
Added follow-up commits

AkihiroSuda (approved):
Thanks

Issue #268, opened 2021-09-27
Create icon for lima project


yangyang198703:
We'd love to see an official icon for lima project, for referencing it in the graphs, etc.

AkihiroSuda:
Help wanted.

Hints:
- "lima" means "five", or "hand" in a lot of Austronesian languages https://en.wiktionary.org/wiki/Reconstruction:Proto-Austronesian/lima
- "līma" means "file" (carpenter's tool) in Latin https://en.wiktionary.org/wiki/lima#Latin
- "lima" means "lime" in several languages such as Spanish, Portuguese, and Polish https://en.wiktionary.org/wiki/lima#Spanish

yangyang198703:
Should we focus on the  **Li**nux on **Ma**c concept?

AkihiroSuda:
> Should we focus on the **Li**nux on **Ma**c concept?

Not sure. At least we should make sure to avoid causing trademark issues.

jandubois:
As somebody who frequently mistypes `lima` as `lime` I'm partial to a Lime icon (I also like the color). Do a Google image search for [lime icon](https://www.google.com/search?q=lime+icon) to get some inspiration.

Just make sure it looks distinct from the LimeWire icon. The software has been dead for over a decade, but we don't want the association with p2p sharing software.

Theoretically I like "File" (because it is a tool), but files make for rather boring images...

"Five" sounds interesting, but not sure what to make with it.

toricls:
🦙 ([Lama](https://en.wikipedia.org/wiki/Lama_(genus))) ‪🤗‬

rothgar:
I threw this together as an option (or inspriation)

I liked the idea of a lime (circle) and thought the blue from the old mac OS 9 "face" logo would be interesting to mix in 

Lima also made me think of fitting a square peg in a round hole (but in this case linux is a round peg)

![lima](https://user-images.githubusercontent.com/371796/134860991-01731897-5a40-403c-8227-b367d7a72420.gif)

I also animated it which could be cool. I'm not sure why the "i" and "a" have clipping issues but you get the idea.

![lima (1)](https://user-images.githubusercontent.com/371796/134861308-e28daa15-01f1-4844-b9c2-957839c5a4d6.gif)

A green circle with blue square is really basic (I'm not a designer) but could be very simple and identifiable

It's also kinda nice that you can create the logo in unicode which also could be embedded in the terminal output :green_circle: :blue_square:

yangyang198703:
[...]

PR #1090, opened 2022-10-08
Don't try to patch the prompt on other systems

### LINKED ISSUE #1086: Fedora: 08-shell-prompt.sh: line 10: patch: command not found
### Description

limactl version 0.12.0

macOS 12.6

Fedora:

https://github.com/lima-vm/lima/blob/fc783ec455a91d70639f9a1d7f22e9890fe6b1cd/examples/fedora.yaml)

```bash
$ uname -a
Linux lima-default 5.17.5-300.fc36.x86_64 #1 SMP PREEMPT Thu Apr 28 15:51:30 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

### -1. If you came here for a solution

1. `dnf install patch -y`
2. goto step 4, use that `.bashrc` to replace your current one

### 0. systemd Failed Units: 1

```bash
$ lima
[systemd]
Failed Units: 1
  cloud-final.service
[vito@lima-default ~]$
```

### 1. patch: command not found

```bash
$ sudo journalctl  -u cloud-final
...
Oct 07 10:59:17 lima-default cloud-init[841]: LIMA| Executing /mnt/lima-cidata/boot/08-shell-prompt.sh
Oct 07 10:59:17 lima-default cloud-init[841]: + '[' default = default ']'
Oct 07 10:59:17 lima-default cloud-init[841]: + grep -q '^# Lima PS1' /home/vito.linux/.bashrc
Oct 07 10:59:17 lima-default cloud-init[841]: + patch --forward -r - /home/vito.linux/.bashrc
Oct 07 10:59:17 lima-default cloud-init[841]: /mnt/lima-cidata/boot/08-shell-prompt.sh: line 10: patch: command not found
Oct 07 10:59:17 lima-default cloud-init[841]: LIMA| WARNING: Failed to execute /mnt/lima-cidata/boot/08-shell-prompt.sh
...
```

then I installed patch with `sudo dnf install patch`

### 2. Hunk ... FAILED at ...

```bash
$ sudo journalctl  -u cloud-final
...
Oct 07 11:08:10 lima-default cloud-init[834]: LIMA| Executing /mnt/lima-cidata/boot/08-shell-prompt.sh
Oct 07 11:08:10 lima-default cloud-init[834]: + '[' default = default ']'
Oct 07 11:08:10 lima-default cloud-init[834]: + grep -q '^# Lima PS1' /home/vito.linux/.bashrc
Oct 07 11:08:10 lima-default cloud-init[834]: + patch --forward -r - /home/vito.linux/.bashrc
Oct 07 11:08:10 lima-default cloud-init[834]: patching file /home/vito.linux/.bashrc
Oct 07 11:08:10 lima-default cloud-init[834]: Hunk #1 FAILED at 37.
Oct 07 11:08:10 lima-default cloud-init[834]: Hunk #2 FAILED at 56.
Oct 07 11:08:10 lima-default cloud-init[834]: 2 out of 2 hunks FAILED
Oct 07 11:08:10 lima-default cloud-init[834]: LIMA| WARNING: Failed to execute /mnt/lima-cidata/boot/08-shell-prompt.sh
...
```

### 3. Check ~/.bashrc

```bash
$ cat ~/.bashrc
# .bashrc

# Source global definitions
if [ -f /etc/bashrc ]; then
        . /etc/bashrc
fi

# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]
then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH

# Uncomment the following line if you don't like systemctl's auto-paging feature:
# export SYSTEMD_PAGER=

# User specific aliases and functions
if [ -d ~/.bashrc.d ]; then
        for rc in ~/.bashrc.d/*; do
                if [ -f "$rc" ]; then
                        . "$rc"
                fi
        done
fi

unset rc
# Lima BEGIN
# Make sure iptables and mount.fuse3 are available
PATH="$PATH:/usr/sbin:/sbin"
export PATH
# Lima END
```

### 4. Guess: you guys only made this for Ubuntu

Then I started an Ubuntu and copied the `.bashrc` to the Fedora.

<details>
  <summary>.bashrc from Ubuntu</summary>

```bash
# ~/.bashrc: executed by bash(1) for non-login shells.
# see /usr/share/doc/bash/examples/startup-files (in the package bash-doc)
# for examples

# If not running interactively, don't do anything
case $- in
    *i*) ;;
      *) return;;
esac

# don't put duplicate lines or lines starting with space in the history.
# See bash(1) for more options
HISTCONTROL=ignoreboth

# append to the history file, don't overwrite it
shopt -s histappend

# for setting history length see HISTSIZE and HISTFILESIZE in bash(1)
HISTSIZE=1000
HISTFILESIZE=2000

# check the window size after each command and, if necessary,
# update the values of LINES and COLUMNS.
shopt -s checkwinsize

# If set, the pattern "**" used in a pathname expansion context will
# match all files and zero or more directories and subdirectories.
#shopt -s globstar

# make less more friendly for non-text input files, see lesspipe(1)
[ -x /usr/bin/lesspipe ] && eval "$(SHELL=/bin/sh lesspipe)"

# set variable identifying the chroot you work in (used in the prompt below)
if [ -z "${debian_chroot:-}" ] && [ -r /etc/debian_chroot ]; then
    debian_chroot=$(cat /etc/debian_chroot)
fi

# set a fancy prompt (non-color, unless we know we "want" color)
case "$TERM" in
    xterm-color|*-256color) color_prompt=yes;;
esac

# uncomment for a colored prompt, if the terminal has the capability; turned
# off by default to not distract the user: the focus in a terminal window
# should be on the output of commands, not on the prompt
#force_color_prompt=yes

if [ -n "$force_color_prompt" ]; then
    if [ -x /usr/bin/tput ] && tput setaf 1 >&/dev/null; then
	# We have color support; assume it's compliant with Ecma-48
	# (ISO/IEC-6429). (Lack of such support is extremely rare, and such
	# a case would tend to support setf rather than setaf.)
	color_prompt=yes
    else
	color_prompt=
    fi
fi

if [ "$color_prompt" = yes ]; then
    PS1='${debian_chroot:+($debian_chroot)}\[\033[01;32m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
else
    PS1='${debian_chroot:+($debian_chroot)}\u@\h:\w\$ '
fi
unset color_prompt force_color_prompt

# If this is an xterm set the title to user@host:dir
case "$TERM" in
xterm*|rxvt*)
    PS1="\[\e]0;${debian_chroot:+($debian_chroot)}\u@\h: \w\a\]$PS1"
    ;;
*)
    ;;
esac

# enable color support of ls and also add handy aliases
if [ -x /usr/bin/dircolors ]; then
    test -r ~/.dircolors && eval "$(dircolors -b ~/.dircolors)" || eval "$(dircolors -b)"
    alias ls='ls --color=auto'
    #alias dir='dir --color=auto'
    #alias vdir='vdir --color=auto'

    alias grep='grep --color=auto'
    alias fgrep='fgrep --color=auto'
    alias egrep='egrep --color=auto'
fi

# colored GCC warnings and errors
#export GCC_COLORS='error=01;31:warning=01;35:note=01;36:caret=01;32:locus=01:quote=01'

# some more ls aliases
alias ll='ls -alF'
alias la='ls -A'
alias l='ls -CF'

# Add an "alert" alias for long running commands.  Use like so:
#   sleep 10; alert
alias alert='notify-send --urgency=low -i "$([ $? = 0 ] && echo terminal || echo error)" "$(history|tail -n1|sed -e '\''s/^\s*[0-9]\+\s*//;s/[;&|]\s*alert$//'\'')"'

# Alias definitions.
# You may want to put all your additions into a separate file like
# ~/.bash_aliases, instead of adding them here directly.
# See /usr/share/doc/bash-doc/examples in the bash-doc package.

if [ -f ~/.bash_aliases ]; then
    . ~/.bash_aliases
fi

# enable programmable completion features (you don't need to enable
# this, if it's already enabled in /etc/bash.bashrc and /etc/profile
# sources /etc/bash.bashrc).
if ! shopt -oq posix; then
  if [ -f /usr/share/bash-completion/bash_completion ]; then
    . /usr/share/bash-completion/bash_completion
  elif [ -f /etc/bash_completion ]; then
    . /etc/bash_completion
  fi
fi
# Lima BEGIN
# Make sure iptables and mount.fuse3 are available
PATH="$PATH:/usr/sbin:/sbin"
export PATH
# Lima END
```

</details>

----

Problem solved.

afbjorklund:
The current script was assuming that "default" is Ubuntu LTS...

But now it is possible to mix and match names and templates.

Closes #1086

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/08-shell-prompt.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/08-shell-prompt.sh
@@ -1,7 +1,9 @@
 #!/bin/sh
 set -eux
 
-if [ "${LIMA_CIDATA_NAME}" = "default" ]; then
+# This script is only intended for the default.yaml image, which is based on Ubuntu LTS
+
+if [ "${LIMA_CIDATA_NAME}" = "default" ] && command -v patch >/dev/null 2>&1 && grep -q color_prompt "/home/${LIMA_CIDATA_USER}.linux/.bashrc"; then
 
 	! grep -q "^# Lima PS1" "/home/${LIMA_CIDATA_USER}.linux/.bashrc" || exit 0
 


pkg/cidata/cidata.TEMPLATE.d/boot/08-shell-prompt.sh
```diff
@@ -1,7 +1,9 @@
 #!/bin/sh
 set -eux
 
-if [ "${LIMA_CIDATA_NAME}" = "default" ]; then
+# This script is only intended for the default.yaml image, which is based on Ubuntu LTS
+
+if [ "${LIMA_CIDATA_NAME}" = "default" ] && grep -q color_prompt /etc/skel/.bashrc; then
```
jandubois, line 6: Why is this grepping `/etc/skel/.bashrc` and not `/home/${LIMA_CIDATA_USER}.linux/.bashrc`, the file we are actually going to patch?

My concerns about using `patch` at all from https://github.com/lima-vm/lima/pull/486#issuecomment-996532216 remain.

But going further, if this should only ever be applied to the "default" Ubuntu instance, why isn't this a provisioning script in `default.yaml` instead, making it much easier for the end user to tweak?

afbjorklund: It was relying on the knowledge of it being the template file.

afbjorklund: I was reluctant to add any scripts to the default ("empty" ?) template

The original goal was just to tweak the prompt when running `lima`:

* https://github.com/lima-vm/lima/issues/427

There are probably other ways to accomplish the same thing...

antoineco: I know this ship has probably sailed, but injecting the following snippet at the end of `/etc/bashrc` would have probably been more than enough to solve the "make the prompt distinguishable" problem:

```sh
[ "$PS1" ] && PS1="[lima]$PS1"
```

Fiddling with colors shouldn't be the responsibility of Lima IMO. It's just impossible to support all OS out there, let alone all their versions, with any kind of PS1 patching approach.

afbjorklund: As mentioned in other comments, there was never any intentation on patching any other system than default.yaml

afbjorklund:
This could probably need a regression test, to see that it actually fixes the problem:

`limactl start --name=default template://fedora`

Assuming that the original template is already covered by the other tests.

`limactl start`

pkg/cidata/cidata.TEMPLATE.d/boot/08-shell-prompt.sh
afbjorklund: Note that the colors are _already_ being fiddled with, the idea was just to make it specific to ubuntu-lima distribution ?

But the since the whole idea with "lime" is gone anyway, and we don't have a new one, I'm fully prepared to just drop it...

* https://github.com/lima-vm/lima/issues/268#issuecomment-1069351680

* https://github.com/lima-vm/lima/pull/486

* `#BFFF00` "lime" (154 GreenYellow)

* `#00bf8c` "cyan" (36 DarkCyan)

There was also support for `NO_COLOR`, if you just wanted to get away...

https://no-color.org/

But there's two "lima":

```
$ TERM=xterm lima
anders@lima-default:/home/anders$ [ "$PS1" ] && PS1="[lima]$PS1"
[lima]anders@lima-default:/home/anders$ 
```

antoineco: I don't have a strong opinion on what the prompt should look like. In fact, as a Lima user my preference would be to leave the guest OS' shell config alone entirely for two reasons:

- Probably 90% of people have a customized prompt of some sort on their host, they will notice a change while opening a `lima` terminal. (Good point regarding the Lima hostname in the prompt.) 
- Lima's provisioning scripts are powerful enough for users to decide whether or not they want to apply their own customizations. (Mount/symlink dotfiles, patch configs, ...)

Regarding this PR, it solves the problem with VMs consistently reporting a failed provisioning (via cloud-final unit) so I think it's a useful, low-hanging workaround. Thanks for addressing it!

afbjorklund: Yeah, that was probably a bug (or lack of paranoia)

AkihiroSuda: @jandubois LGTY?

AkihiroSuda (approved):

Issue #1093, opened 2022-10-10
Allow further customization of Lima’s boot workflow


pendo324:
### Description

Lima currently tries to have sane defaults per every OS it supports when installing system dependencies (such as iptables, fuse, sshfs, etc.). This works great most of the time, but sometimes, further customization could be useful. Some examples are when adding support for a new OS, or installing customized/patched versions of those system dependencies, or even adding a custom package repository (yum/apt/etc.) before trying to install packages.

Additionally, customizing the dependency installation workflow could also enable advanced users to remove the requirement of internet access for downloading these dependencies (for example, if a user set up a local yum repository and then installs their system dependencies from that, rather than the default repositories).

Another nice to have feature is allowing Lima users to customize cloud init’s cloud-init-local.service behavior via bootCmds, which run very early in the boot process ([more info here](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#run-commands-on-first-boot)). Users can use this feature to disable unnecessary services via systemctl (or other means, depending on the OS), or run other arbitrary commands for other reasons.

In order to support all of these use cases, I propose adding a few new features to Lima’s config:

1. Allow users to add cloud init bootCmds directly in their Lima configs
    1. This is a relatively straightforward change and would utilize the existing templating for CIDATA
2. Allow users to supply a new type of provisioning script of mode “dependency”
    1. These dependency scripts would allow users to bypass the default logic in [30-install-package.sh](https://github.com/lima-vm/lima/blob/master/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh) and replace it with their own logic. Alternatively, the scripts could be made to run in addition to the default logic (with another new flag added to suppress the default logic)
3. Allow users to supply 1 or n “extra” archives that get added into CIDATA
    1. This allows users to supply their own local packages or package repositories for use with a custom “dependency” script (since such a script would run before sshfs mounts are available)

If there is interest in these features, I plan on following up with several PRs to add the features mentioned in this issue. I can also split this issue into more granular issues, if that’s better.

PR #1094, opened 2022-10-11
add "boot" provisioning script mode

pendo324:
Allows users to customize their early boot workflow by specifying `bootCmds`. These commands are run by `cloud-init-local.service`, which runs before `cloud-init.service` (which only starts after the network is up), and allow users to disable services (as in the example) or run other commands early on in the boot process. [More info here](https://cloudinit.readthedocs.io/en/latest/topics/examples.html#run-commands-on-first-boot).

Resolves one of the proposals in #1093

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -171,6 +171,12 @@ containerd:
 #     cat <<EOF > ~/.vimrc
 #     set number
 #     EOF
+# # `boot` is executed directly by /bin/sh as part of cloud-init-local.service's early boot process,
+# # which is why there is no hash-bang specified in the example
+# # See cloud-init docs for more info https://cloudinit.readthedocs.io/en/latest/topics/examples.html#run-commands-on-first-boot
+# - mode: boot
+#   script: |
+#     systemctl disable NetworkManager-wait-online.service
 
 # Probe scripts to check readiness.
 # 🟢 Builtin default: null
--- a/pkg/cidata/cidata.TEMPLATE.d/user-data
+++ b/pkg/cidata/cidata.TEMPLATE.d/user-data
@@ -64,3 +64,13 @@ ca-certs:
     {{- end }}
   {{- end }}
 {{- end }}
+
+{{- if .BootCmds }}
+bootcmd:
+  {{- range $cmd := $.BootCmds }}
+- |
+    {{- range $line := $cmd.Lines }}
+  {{ $line }}
+    {{- end }}
+  {{- end }}
+{{- end }}
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -219,6 +219,8 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 		args.CACerts.Trusted = append(args.CACerts.Trusted, cert)
 	}
 
+	args.BootCmds = getBootCmds(y.Provision)
+
 	if err := ValidateTemplateArgs(args); err != nil {
 		return err
 	}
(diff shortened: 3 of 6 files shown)


AkihiroSuda:
https://github.com/apps/dco

examples/default.yaml
```diff
@@ -172,6 +172,11 @@ containerd:
 #     set number
 #     EOF
 
+# Adds commands to the cloud-init user-data config
+# See cloud-init docs for more info https://cloudinit.readthedocs.io/en/latest/topics/examples.html#run-commands-on-first-boot
+# bootCmds:
+#  - systemctl disable NetworkManager-wait-online.service
```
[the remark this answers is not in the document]
pendo324: Yeah, that makes sense. I'll push another revision to include it with the rest of the provisioning scripts

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Could you squash commits?

examples/default.yaml
```diff
@@ -171,6 +171,12 @@ containerd:
 #     cat <<EOF > ~/.vimrc
 #     set number
 #     EOF
+# # `boot` is executed as part of cloud-init-local.service's early boot process
+# # See cloud-init docs for more info https://cloudinit.readthedocs.io/en/latest/topics/examples.html#run-commands-on-first-boot
+# - mode: boot
+#   script: |
+#     #!/bin/bash
```
jandubois, line 178: Does this work to execute the commands using `bash`?

I have my doubts because in `bootcmd` you can mix strings and list forms, and the list forms are executed by `execve` whereas the strings are executed by `sh`. So I doubt any hash-bang line has any effect on subsequent commands. But I don't know, so could you please verify, and then add a corresponding comment?

pkg/cidata/cidata.TEMPLATE.d/user-data
```diff
@@ -64,3 +64,13 @@ ca-certs:
...
+
+{{- if .BootCmds }}
+bootcmd:
+  {{- range $cmd := $.BootCmds }}
+  - |
+    {{- range $line := $cmd.Lines }}
+    {{ $line }}
+    {{- end }}
+  {{- end }}
+{{- end }}
```
jandubois, line 76: Ok, this is nit-picking, but the list element is double-indented, it looks like this:

```yaml
bootcmd:
  - |
    #!/bin/bash
    ...
```

when it should look like this:

```yaml
bootcmd:
- |
  #!/bin/bash
  ...
```

```suggestion
{{- if .BootCmds }}
bootcmd:
  {{- range $cmd := $.BootCmds }}
- |
    {{- range $line := $cmd.Lines }}
  {{ $line }}
    {{- end }}
  {{- end }}
{{- end }}
```

examples/default.yaml
jandubois: Yeah, doesn't look like it. I used this boot script:

```yaml
- mode: boot
  script: |
    #!/bin/bash
    cat /proc/$$/cmdline | tr '\000' ' ' > /etc/bootcmdline
```

And it confirms that the script is launched via `/bin/sh`:

```console
$ lima cat /etc/bootcmdline
/bin/sh /run/cloud-init/tmp/tmp98wefsz7.sh
```

The hash-bang line will only be used when the file is loaded via the kernel; when passed directly to `/bin/sh` like this, it is just a comment.

So the hash-bang line should be removed from the sample script and the docs above should explain that the script will be executed by `/bin/sh`, and that you can't specify a hash-bang line.

jandubois (changes requested):
Thank you for the contribution! I've tested it, and it works well.

I would like to see the misleading hash-bang line issue addressed.

I wouldn't insist on fixing the YAML indentation, but if you edit the PR anyways, then please address that as well.

jandubois (approved):
Thank you, looks good!

jandubois:
I don't know why the Cirrus CI checks have not been running; I'm going to close this PR and then re-open it, to hopefully trigger them again.

jandubois:
Doing another cycle of Close+Reopen, as Cirrus-CI seems to be working fine in other repos right now; but still not triggering here.

PR #1101, opened 2022-10-13
Add wrapper for running apptainer on Mac

afbjorklund:
Now that Apptainer has been moved from experimental, and accepted as LF project, it might be time for a wrapper ?

There is no client for `apptainer`, similar to `nerdctl`.

So use a small shell wrapper, to run "lima apptainer".

This is so that the user can follow the documentation, without having to write an extra `lima ` prefix before every command:

https://apptainer.org/docs/user/main/quick_start.html

We might still have to document the read-only home.

```
FATAL:   While performing build: while creating SIF: while creating container: open /home/anders/lima/alpine.sif: read-only file system
```

---

This is different from the situation for docker and podman.

Since they already have client binaries, available for Mac.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ binaries: clean \
 	_output/bin/lima$(bat) \
 	_output/bin/limactl$(exe) \
 	_output/bin/nerdctl.lima \
+	_output/bin/apptainer.lima \
 	_output/bin/docker.lima \
 	_output/bin/podman.lima \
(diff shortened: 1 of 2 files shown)


afbjorklund:
It would be nice if this could automatically add the `-B $HOME:$HOME` mount, to `run`.

AkihiroSuda (approved):
Thanks

PR #1102, opened 2022-10-14
Enable trimming QEMU disks

### LINKED ISSUE #356: Enable trimming of the `diffdisk`
The `diffdisk` starts out small, but grows quickly as users create and delete container images. This can become an issue on laptops with limited free space.

I tried to enable trim support just for Alpine, for testing, but couldn't get it to work. Here is the patch I tried:

```diff
--- pkg/cidata/cidata.TEMPLATE.d/boot/05-persistent-data-volume.sh
+++ pkg/cidata/cidata.TEMPLATE.d/boot/05-persistent-data-volume.sh
@@ -14,7 +14,7 @@ DATADIRS="/etc /home /tmp /usr/local /var/lib"
 if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
        mkdir -p /mnt/data
        if [ -e /dev/disk/by-label/data-volume ]; then
-               mount -t ext4 /dev/disk/by-label/data-volume /mnt/data
+               mount -t ext4 -o discard /dev/disk/by-label/data-volume /mnt/data
        else
                # Find an unpartitioned disk and create data-volume
                DISKS=$(lsblk --list --noheadings --output name,type | awk '$2 == "disk" {print $1}')
@@ -32,7 +32,7 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
                                echo 'type=83' | sfdisk --label dos /dev/"${DISK}"
                                PART=$(lsblk --list /dev/"${DISK}" --noheadings --output name,type | awk '$2 == "part" {prin
                                mkfs.ext4 -L data-volume /dev/"${PART}"
-                               mount -t ext4 /dev/disk/by-label/data-volume /mnt/data
+                               mount -t ext4 -o discard /dev/disk/by-label/data-volume /mnt/data
                                for DIR in ${DATADIRS}; do
                                        DEST="/mnt/data$(dirname "${DIR}")"
                                        mkdir -p "${DIR}" "${DEST}"
--- pkg/qemu/qemu.go
+++ pkg/qemu/qemu.go
@@ -261,7 +261,7 @@ func Cmdline(cfg Config) (string, []string, error) {
                args = appendArgsIfNoConflict(args, "-boot", "order=c,splash-time=0,menu=on")
        }
        if diskSize, _ := units.RAMInBytes(cfg.LimaYAML.Disk); diskSize > 0 {
-               args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio", diffDisk))
+               args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio,discard=unmap", diffDisk))
        } else if !isBaseDiskCDROM {
                args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio", baseDisk))
        }
```

I've created some files via `dd if=/dev/urandom of=1.bin bs=64M count=64 iflag=fullblock` etc and verified the growth in `diffdisk` size.

I then deleted the `*.bin` files and ran:

```console
lima-alpine:~$ sudo fstrim -v /mnt/data
/mnt/data: 105074479104 bytes trimmed
lima-alpine:~$ sudo fstrim -v /mnt/data
/mnt/data: 0 bytes trimmed
```

But the size of the disk never shrinks.

I wonder if this is a macOS limitation, that qemu doesn't implement the sparse file logic for APFS.

Thoughts?

chrisx8:
This would enable trimming of QEMU virtual disks, by adding the `discard=on` argument to `-drive` flag for `basedisk` or `diffdisk`. As a result, running `fstrim -a` in the guest would reduce the disk space consumed by QEMU disks (`basedisk` or `diffdisk`), and reduce the disk usage of Lima guests.

Current behavior: there's no easy way to reduce the disk usage of `diffdisk`, and running `fstrim -a` in the guest has no effect.

Resolves #356

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -433,9 +433,9 @@ func Cmdline(cfg Config) (string, []string, error) {
 		args = appendArgsIfNoConflict(args, "-boot", "order=c,splash-time=0,menu=on")
 	}
 	if diskSize, _ := units.RAMInBytes(*cfg.LimaYAML.Disk); diskSize > 0 {
-		args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio", diffDisk))
+		args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio,discard=on", diffDisk))
 	} else if !isBaseDiskCDROM {
-		args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio", baseDisk))
+		args = append(args, "-drive", fmt.Sprintf("file=%s,if=virtio,discard=on", baseDisk))
 	}
 	// cloud-init
 	switch *y.Arch {


jandubois (approved):
Thank you!

I'm so happy this actually works! 😄

jandubois:
FYI: Alpine doesn't have `fstrim -a`, but requires mountpoint, e.g. `fstrim /tmp`. It still releases all unused space from the `diffdisk`.

CzBiX:
Does `basedisk` really need discard too?

jandubois:
> Looks like Lima only uses either basedisk or diffdisk, so I'd say yes.

We never write to `basedisk` though: it either is a read-only ISO, or it is the base for the `diffdisk`; all modifications should always go to the `diffdisk` unless I misunderstand how this works. So I don't think it is needed for the `basedisk`, but does it hurt? Is there any performance/memory cost associated with turning it on?

chrisx8:
https://github.com/lima-vm/lima/blob/020d7ee10b219822d4c59d72c3764f3e1ee3baa6/pkg/qemu/qemu.go#L435-L439

This code snippet suggests that only one of `basedisk` or `diffdisk` is used on the QEMU VM. If QEMU is set to boot from the `diffdisk`, it knows about the `basedisk`.

There shouldn't be any performance impact during normal use. However, `fstrim` may take slightly more time, although I can't really tell the difference when I was testing.

jandubois:
Ah yes, the only time we use the `basedisk` directly is when it is not an ISO and the `diffdisk` size is 0. So everything looks correct.

PR #1105, opened 2022-10-14
add "dependency" provisioning mode

pendo324:
Adds a new provisioning script mode that allows users to customize the workflow of `pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh`.

This can be useful in the case where a user wants to override the default behavior to (for example) add repositories before attempting to install system dependencies.

Another option is to add a flag to limayaml to allow running dependency provisioning scripts _in addition to_ the default logic (instead of bypassing the default logic), please comment if that's more preferable.

More details in #1093


jandubois:
> Another option is to add a flag to limayaml to allow running dependency provisioning scripts _in addition to_ the default logic (instead of bypassing the default logic), please comment if that's more preferable.

I may need to think about this some more, but my gut feeling would be that the `dependencies` scripts should always be run in addition to, and not instead of the default installation.

I don't know if there is a need to potentially disable the default package installation, but that would then also require a mechanism to disable the corresponding requirements checks (`pkg/hostagent/requirements.go`), doesn't it?

pendo324:
> I don't know if there is a need to potentially disable the default package installation, but that would then also require a mechanism to disable the corresponding requirements checks (`pkg/hostagent/requirements.go`), doesn't it?

I think that the essential requirements should still be the same, but I do see the benefit in having another flag in limayaml to disable default requirement checks.

The most customizable approach would probably be something like:
- dependency mode provisioning scripts to run in addition to the default `30-install-packages.sh`
  - these scripts actually don't have to be limited to the "dependency" scope, naming them something like "early" or "pre-installation" mode would also make sense
- new option like `skipDefaultDependencyInstallation` to skip the default dependency installation
- new option like `skipDefaultRequirements` to skip the default requirements in `pkg/hostagent/requirements.go`

`skipDefaultDependencyInstallation` could also be overloaded to skip requirements too, if we think that two new options would be too much

AkihiroSuda:
Please squash commits

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -23,150 +23,165 @@ if [ "${LIMA_CIDATA_UDP_DNS_LOCAL_PORT}" -ne 0 ] || [ "${LIMA_CIDATA_TCP_DNS_LOC
...
-	if [ "${INSTALL_IPTABLES}" = 1 ] && ! command -v iptables >/dev/null 2>&1; then
-		pkgs="${pkgs} iptables"
-	fi
-	if [ -n "${pkgs}" ]; then
-		apk update
-		# shellcheck disable=SC2086
-		apk add ${pkgs}
-	fi
+else
+	echo "LIMA_CIDATA_USE_DEFAULT_DEPENDENCY_RESOLUTION set, skipping regular dependency installation"
```
AkihiroSuda, line 184: Did you mean
```suggestion
	echo "LIMA_CIDATA_USE_DEFAULT_DEPENDENCY_RESOLUTION != 1, skipping regular dependency installation"
```

pkg/limayaml/defaults.go
```diff
@@ -535,6 +535,16 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
 	y.CACertificates.Certs = caCerts
+
+	if y.UseDefaultDependencyResolution == nil {
+		y.UseDefaultDependencyResolution = d.UseDefaultDependencyResolution
+	}
+	if o.UseDefaultDependencyResolution != nil {
+		y.UseDefaultDependencyResolution = o.UseDefaultDependencyResolution
+	}
+	if y.UseDefaultDependencyResolution == nil {
+		y.UseDefaultDependencyResolution = pointer.Bool(true)
```
AkihiroSuda, line 546: Wondering this should be rather inverted to `SkipDefaultDependency` with the default value `false`

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
pendo324: Yeah, good catch. I think changing it to "skip" also makes more sense, like your other comment suggested.

The default value of "false" is easier to reason about, since most users won't interact with this option. Just pushed an update to address both comments, and also update the default example yaml

AkihiroSuda (dismissed):
Thanks

AkihiroSuda:
@jandubois LGTY?

AkihiroSuda:
Sorry, needs another rebase

AkihiroSuda (dismissed):
Thanks

AkihiroSuda:
@lima-vm/maintainers Let me know if we can merge this.
Do we want to treat this as an experimental feature?

examples/default.yaml
```diff
@@ -199,6 +199,19 @@ containerd:
...
+#   script: |
+#     #!/bin/bash
+#     dnf config-manager --add-repo ...
+#     dnf install ...
+
+# `skipDefaultDependencyResolution` can be used with `dependency` mode provisioning scripts to skip
+# the regular dependency resolution workflow in pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
+# and provide a fully customizable replacement
+# 🟢 Builtin default: false
+# skipDefaultDependencyResolution: true
```
jandubois, line 214: This is quite a mouthful, and also feels somewhat disconnected here. I'm wondering if this would make more sense to make this an option on the `dependency` provisioning script.

I don't have strong opinions, it just feels a bit clunky to have this weirdly-named option at the top level, that only makes sense when you also have some other bits of configuration defined.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -23,149 +23,164 @@ if [ "${LIMA_CIDATA_UDP_DNS_LOCAL_PORT}" -ne 0 ] || [ "${LIMA_CIDATA_TCP_DNS_LOC
 fi
 
 # Install minimum dependencies
+# Run any user provided dependency scripts first
+if [ -d "${LIMA_CIDATA_MNT}"/provision.dependency ]; then
+	echo "Detected dependency provisioning scripts, running before default dependency installation"
+	for f in "${LIMA_CIDATA_MNT}"/provision.dependency/*; do
+		echo "Executing $f"
+		if ! "$f"; then
+			echo "Failed to execute $f"
```
jandubois, line 32: Is there a reason we don't want to `exit 1` here? The rest of the script runs under `set -e`, so any error return would immediately exit the script.

On the other hand, in `boot.sh`, we set `CODE=1` when a provisioning script fails, and then `exit $CODE` at the end.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -23,149 +23,164 @@ if [ "${LIMA_CIDATA_UDP_DNS_LOCAL_PORT}" -ne 0 ] || [ "${LIMA_CIDATA_TCP_DNS_LOC
...
 # to mimic certain options but doesn't offer full parameters compatibility
 # See : https://github.com/lima-vm/lima/pull/1014
-if hexdump -C -n 4 "$(command -v apt-get)" | grep -qF 'ELF' >/dev/null 2>&1; then
-	pkgs=""
-	if [ "${LIMA_CIDATA_MOUNTTYPE}" = "reverse-sshfs" ]; then
-		if [ "${LIMA_CIDATA_MOUNTS}" -gt 0 ] && ! command -v sshfs >/dev/null 2>&1; then
-			pkgs="${pkgs} sshfs"
+if [ "${LIMA_CIDATA_SKIP_DEFAULT_DEPENDENCY_RESOLUTION}" = 1 ]; then
+	echo "LIMA_CIDATA_SKIP_DEFAULT_DEPENDENCY_RESOLUTION is set, skipping regular dependency installation"
+else
```
jandubois, line 44: The conditionals are already nested too deeply in this file for my taste, so adding a 140 line `else` clause does not feel appealing.

I would rather do an early exit here, and then move the 2 actions at the bottom of the script (setting up DNS, and updating fuse_conf) into a separate script.

jandubois (commented):
I'm generally fine with this PR, but have some stylistic feedback.

I think marking it as experimental gives us some wiggle-room to still change how we configure it in the future.

examples/default.yaml
pendo324: Good point, addressed in latest revision. I added a new field to the Provisioning scripts in general, but invalidate it when used on non-dependency scripts.

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
pendo324: Makes sense. Changed it to collect all errors in a CODE var and exit if its 1 after all dependency scripts, to match boot.sh behavior

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
pendo324: I wasn't a big fan of the nesting either. Split the file into two and exit early now

AkihiroSuda (approved):
Thanks

pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh
```diff
@@ -23,11 +10,30 @@ if [ "${LIMA_CIDATA_UDP_DNS_LOCAL_PORT}" -ne 0 ] || [ "${LIMA_CIDATA_TCP_DNS_LOC
...
+# Run any user provided dependency scripts first
+if [ -d "${LIMA_CIDATA_MNT}"/provision.dependency ]; then
+	echo "Detected dependency provisioning scripts, running before default dependency installation"
+	CODE=0
+	for f in "${LIMA_CIDATA_MNT}"/provision.dependency/*; do
+		if ! "$f"; then
+			CODE=1
+		fi
+	done
+	if [ $CODE != 0 ]; then
```
jandubois, line 22: I think adding a log message that we are exiting early would be nice to have (as we do in `boot.sh`):

```suggestion
	if [ $CODE -ne 0 ]; then
	    echo "Exiting with code $CODE"
```

jandubois (approved):
Thanks, LGTM.

I think having the comment on the error exit would be nice, but I'm also fine with merging as-is.

Issue #1103, opened 2022-10-14
Automatically expand the filesystem on mounted volumes when they have grown


jandubois:
### Description

This issue applies to the `diffdisk` when the `basedisk` is an ISO (Alpine), and to additional volumes provided via #1065 

Is is possible to expand the max file size of the image on the host (see https://github.com/lima-vm/lima/issues/356#issuecomment-1278459691), but the additional space does not become available until the partition and filesystem have been extended to include the additional space.

This should be implemented in [05-persistent-data-volume.sh](https://github.com/lima-vm/lima/blob/master/pkg/cidata/cidata.TEMPLATE.d/boot/05-persistent-data-volume.sh) (renamed to `04-...` once #1065 is merged).

@chrisx8 Is this something you would like to work on?

jandubois:
> I can look into this, but I'm not too familiar with Alpine. Assuming it's on ext4, we can just run `e2fsck` then `resize2fs` in the script.

Yes, it is always `ext4`. I just checked, and our alpine-lima image has `e2fsck`, but not `resize2fs`. I will add `e2fsprogs-extra` to our images. We need to check if this is installed in the other distros, or if we need to install those dependencies there as well (which would require re-arranging some code in the boot sequence).

> Questions:
> 
> * How do we handle errors, in case `e2fsck` or `resize2fs` fails?

I think it will just be logged, and then we proceed without resizing the filesystem.

> * We probably should warn users if `e2fsck` fails, as this suggests the filesystem is unhealthy. Should we add this, if so, where?

This happens during boot via cloud-init. So I think the only thing we'll have are cloud-init logs.

Not sure what else we can do? Some warning via `/etc/profile` that this happened? I would ignore this for now; it can be added in a later PR if we feel it is necessary.

jandubois:
> I will add `e2fsprogs-extra` to our images.

Done in https://github.com/lima-vm/alpine-lima/pull/88 (pending review).

I'll create a new tag/release once approved/merged.

jandubois:
Alpine image with `resize2fs` has been released, and the `examples/alpine.yaml` has been updated in #1106 to use it.

chrisx8:
@jandubois Can we have the `cloud-utils-growpart` package in Alpine? I want to use `growpart` to expand the partition on the datadisk.

chrisx8:
[...]

PR #1106, opened 2022-10-15
Bump lima-alpine to 0.2.22 to include resize2fs

jandubois:
This will be needed to grow the filesystem when the size of the attached data volume is increased. See https://github.com/lima-vm/lima/issues/1103


AkihiroSuda (approved):
Thanks

Issue #971, opened 2022-07-19
v1.0 roadmap: change the default mount driver from reverse-sshfs to 9p (for QEMU) and virtiofs (for vz)


AkihiroSuda:
https://github.com/lima-vm/lima/blob/master/docs/mount.md

Lima v1.0 will change the default mount driver from reverse-sshfs to 9p for QEMU, virtiofs for vz. (The default vm driver will be vz)
RHEL-like templates will continue to use reverse-sshfs as their kernel lacks support for virtio-9p-pci.

afbjorklund:
Did changing the sftp server fix the odd bugs, like when you can't compile any code (due to umask) ?

* https://github.com/lima-vm/sshocker/issues/17

* https://github.com/lima-vm/lima/issues/602

Support for virtfs is also lacking on Windows and other host OS (like BSD), `requires Linux or macOS`

AkihiroSuda:
QEMU 7.1 is released and seems to have resolved issues like:
- https://github.com/lima-vm/lima/issues/831

@jandubois @afbjorklund Are we ready to change the driver (on Linux and macOS) and release v1.0 soon?

afbjorklund:
According to one user*, there might be 9p support also on Windows hosts soon: (haven't tried it myself, not included in 7.1.0)

\* https://github.com/lima-vm/lima/issues/909#issuecomment-1231547972

As far as I know, there should be no major issues with virtfs on Linux (if reverse-sshfs is still available as a fallback, that is) ?

AkihiroSuda:
Before changing the default driver probably we should add a simple CLI flag like `limactl start --mount-type=reverse-sshfs` to allow switching back to the reverse-sshfs driver.

The `--mount-type=...` flag value will be stored in the "local" `override.yaml` stored in the instance dir so as not to break comment lines and white spaces in the main `lima.yaml`.
The "global" `override.yaml` in the `_config` dir will still have higher precedence.
I'll work on this after getting https://github.com/lima-vm/lima/pull/1069 merged.

(Instead of messing up the overriding system we could just modifying `lima.yaml`, but modifying YAMLs without breaking comments lines and white spaces turned out to be hard: https://github.com/AkihiroSuda/yamlctl)

cc @jandubois

dnrce:
Per https://github.com/lima-vm/lima/discussions/454#discussioncomment-1749904, will a writable `~` mount also become the default as part of this change?

XVilka:
[...]

PR #1110, opened 2022-10-17
9p: change the default `securityModel` from `mapped-xattr` to `none`

AkihiroSuda:
The `mapped-xattr` model is known to be incompatible with symlinks: https://gitlab.com/qemu-project/qemu/-/issues/173

The `none` model is obviously incompatible with chown, but this is fine, as the reverse-sshfs does not support chown either.

See issue #971

### DIFF
--- a/docs/mount.md
+++ b/docs/mount.md
@@ -59,7 +59,8 @@ mounts:
 - location: "~"
   9p:
     # Supported security models are "passthrough", "mapped-xattr", "mapped-file" and "none".
-    # 🟢 Builtin default: "mapped-xattr"
+    # "mapped-xattr" and "mapped-file" are useful for persistent chown but incompatible with symlinks.
+    # 🟢 Builtin default: "none" (since Lima v0.13)
     securityModel: null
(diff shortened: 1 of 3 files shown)


docs/mount.md
```diff
@@ -59,7 +59,8 @@ mounts:
 - location: "~"
   9p:
     # Supported security models are "passthrough", "mapped-xattr", "mapped-file" and "none".
-    # 🟢 Builtin default: "mapped-xattr"
+    # "mapped-xattr" and "mapped-file" are useful for persistent chown but incompatible with symlinks.
+    # 🟢 Builtin default: "none" (since Lima v0.12.1)
```
AkihiroSuda, line 63: (Maybe this should be rather in v0.13.0)

jandubois: Yes, I feel that changing defaults merits at least a minor version bump.

AkihiroSuda: Changed to v0.13.0.

v0.12.1 will be cancelled.

jandubois (approved):
Thanks, sounds good!

Issue #1111, opened 2022-10-17
localhostregexp not support sub domain for the localhost


saltbo:
My http_proxy such as `http_proxy=http://docker.for.mac.localhost:6666`, lima replaced to `http_proxy=http://docker.for.mac.192.168.5.2:6666`

PR #1112, opened 2022-10-17
Fix the regexp for the localhost proxy match

saltbo:
Fix issue #1111

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -4,9 +4,10 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
+	"net/url"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -22,6 +23,16 @@ import (
 	"github.com/lima-vm/lima/pkg/usrlocalsharelima"
 	"github.com/sirupsen/logrus"
 )
+
+var netLookupIP = func(host string) []net.IP {
+	ips, err := net.LookupIP(host)
+	if err != nil {
+		logrus.Debugf("net.LookupIP %s: %s", host, err)
+		return nil
+	}
+
+	return ips
+}
 
 func setupEnv(y *limayaml.LimaYAML) (map[string]string, error) {
 	// Start with the proxy variables from the system settings.
@@ -50,16 +61,25 @@ func setupEnv(y *limayaml.LimaYAML) (map[string]string, error) {
 			}
 		}
 	}
-	// Replace "localhost" in proxy settings with the gateway address
-	localhostRegexes := []*regexp.Regexp{
-		regexp.MustCompile(`\blocalhost\b`),
-		regexp.MustCompile(`\b127.0.0.1\b`),
-	}
+	// Replace IP that IsLoopback in proxy settings with the gateway address
 	for _, name := range append(lowerVars, upperVars...) {
 		value, ok := env[name]
 		if ok && !strings.EqualFold(name, "no_proxy") {
-			for _, re := range localhostRegexes {
-				value = re.ReplaceAllString(value, qemu.SlirpGateway)
+			u, err := url.Parse(value)
+			if err != nil {
+				logrus.Warnf("Ignoring invalid proxy %q=%v: %s", name, value, err)
+				continue
+			}
+
+			for _, ip := range netLookupIP(u.Hostname()) {
+				if ip.IsLoopback() {
+					newHost := qemu.SlirpGateway
+					if u.Port() != "" {
+						newHost = net.JoinHostPort(newHost, u.Port())
+					}
+					u.Host = newHost
+					value = u.String()
+				}
 			}
 			if value != env[name] {
 				logrus.Infof("Replacing %q value %q with %q", name, env[name], value)
--- a/pkg/cidata/cidata_test.go
+++ b/pkg/cidata/cidata_test.go
@@ -1,0 +1,63 @@
+package cidata
+
+import (
+	"net"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/lima-vm/lima/pkg/limayaml"
+	qemu "github.com/lima-vm/lima/pkg/qemu/const"
+	"github.com/xorcare/pointer"
+	"gotest.tools/v3/assert"
+)
+
+func fakeLookupIP(host string) []net.IP {
+	return []net.IP{net.IPv4(127, 0, 0, 0)}
+}
+
+func TestSetupEnv(t *testing.T) {
+	netLookupIP = fakeLookupIP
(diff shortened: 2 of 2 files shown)


jandubois:
> Please also look at the lint failures (you can run `make lint` locally before pushing the commits),

Sorry, just realized the lint/test failures are unrelated to your PR and have been fixed by #1116. Please rebase your PR (or create a new one, as it will be pretty much all different code anyways) based on the latest `master`.

pkg/cidata/cidata.go
```diff
@@ -50,17 +51,23 @@ func setupEnv(y *limayaml.LimaYAML) (map[string]string, error) {
...
 		if ok && !strings.EqualFold(name, "no_proxy") {
-			for _, re := range localhostRegexes {
-				value = re.ReplaceAllString(value, qemu.SlirpGateway)
+			u, err := url.Parse(value)
+			if err != nil {
+				logrus.Warnf("Ignoring invalid proxy %q=%v: %s", name, value, err)
+				continue
+			}
+
+			ips, _ := net.LookupIP(u.Hostname())
```
jandubois, line 64: Why did you not include error checking here?

saltbo: ips = nil if error != nil, then it will skip the for logic.

expect how to handle the error? warn log? exit? I think not need to handle

jandubois: > ips = nil if error != nil, then it will skip the for logic.

Yes, practically speaking this is correct (looking at the source of `net.LookupIP`), and but from a best practices point of view, you should always check `err`, and ignore all other return values if the value isn't `nil`, as there is no guarantee that the other return values are proper null values or not.

Yes, it is a bit pedantic in this case, but I would prefer to consistently check errors.

> expect how to handle the error? warn log? exit? I think not need to handle

Just don't access the `ips` variable when `err != nil`, e.g.

```go
			if ips, err := net.LookupIP(u.Hostname()); err == nil {
				for _, ip := range ips {
					if ip.IsLoopback() {
						u.Host = net.JoinHostPort(qemu.SlirpGateway, u.Port())
					}
				}
				value = u.String()
			}
```

pkg/cidata/cidata_test.go
```diff
@@ -0,0 +1,56 @@
...
+		urlMustParse("sock4://127.0.0.1:8080"),
+		urlMustParse("sock5://127.0.0.1:8080"),
+		urlMustParse("http://127.0.0.1:8080/"),
+		urlMustParse("http://127.0.0.1:8080/path"),
+		urlMustParse("http://localhost:8080"),
+		urlMustParse("http://localhost:8080/"),
+		urlMustParse("http://localhost:8080/path"),
+		urlMustParse("http://docker.for.mac.localhost:8080"),
+		urlMustParse("http://docker.for.mac.localhost:8080/"),
+		urlMustParse("http://docker.for.mac.localhost:8080/path"),
```
jandubois, line 36: `docker.for.mac.localhost` must be in your hosts `/etc/hosts` file; it does not resolve otherwise:

```
--- FAIL: TestSetupEnv (0.13s)
    --- FAIL: TestSetupEnv/docker.for.mac.localhost:8080 (0.00s)
        cidata_test.go:45: assertion failed: http://docker.for.mac.localhost:8080 (envs[envKey] string) != http://192.168.5.2:8080 (string)
    --- FAIL: TestSetupEnv/docker.for.mac.localhost:8080#01 (0.00s)
        cidata_test.go:45: assertion failed: http://docker.for.mac.localhost:8080/ (envs[envKey] string) != http://192.168.5.2:8080/ (string)
    --- FAIL: TestSetupEnv/docker.for.mac.localhost:8080#02 (0.00s)
        cidata_test.go:45: assertion failed: http://docker.for.mac.localhost:8080/path (envs[envKey] string) != http://192.168.5.2:8080/path (string)
```

pkg/cidata/cidata.go
```diff
@@ -50,17 +51,23 @@ func setupEnv(y *limayaml.LimaYAML) (map[string]string, error) {
...
+			for _, ip := range ips {
+				if ip.IsLoopback() {
+					u.Host = net.JoinHostPort(qemu.SlirpGateway, u.Port())
```
jandubois, line 67: Looks like `net.JoinHostPort("127.0.0.1", "")` is buggy and returns `127.0.0.1:` instead of `127.0.0.1`.

So you need to check if `u.Port()` returns the empty string.

The test cases should include at least one host without a port.

pkg/cidata/cidata_test.go
saltbo: use a fakeLookupIP to resolved

pkg/cidata/cidata.go
saltbo: there refactored by the fakeLookupIP, but if not, I would to write
```go
ips, err := net.LookupIP(u.Hostname())
if err != nil {
    logrus.Debugf("net.LookupIP %s: %s", host, err)
    continue
}

for _, ip := range ips {
    newHost := qemu.SlirpGateway
    if u.Port() != "" {
	    newHost = net.JoinHostPort(newHost, u.Port())
    }
    if ip.IsLoopback() {
	    u.Host = newHost
    }
}
```
because we should handle error cases/special conditions first and return early or continue the loop
https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

pkg/cidata/cidata.go
```diff
@@ -50,17 +61,26 @@ func setupEnv(y *limayaml.LimaYAML) (map[string]string, error) {
...
+			}
+
+			for _, ip := range netLookupIP(u.Hostname()) {
+				newHost := qemu.SlirpGateway
+				if u.Port() != "" {
+					newHost = net.JoinHostPort(newHost, u.Port())
+				}
+				if ip.IsLoopback() {
+					u.Host = newHost
+				}
```
jandubois, line 81: It doesn't really matter, but all this code could be inside the `IsLoopback` conditional:

```suggestion
				if ip.IsLoopback() {
					newHost := qemu.SlirpGateway
					if u.Port() != "" {
						newHost = net.JoinHostPort(newHost, u.Port())
					}
					u.Host = newHost
				}
```

jandubois (commented):
Thank you! Looks good now, but please squash commits and apply DCO signature to the commit: https://github.com/lima-vm/lima/pull/1112/checks?check_run_id=8951827839

I had one more comment; normally I wouldn't go through another round of review for it, but since you need to squash commits and sign anyways, please consider moving the `newhost` code inside the `IsLoopback` conditional.

jandubois (approved):
Thank you!

saltbo:
> Thank you! Looks good now, but please squash commits and apply DCO signature to the commit: https://github.com/lima-vm/lima/pull/1112/checks?check_run_id=8951827839

Thank you for the knowledge for the DCO

Issue #841, opened 2022-05-02
Failed to start any RHEL 8-based Linux on M1 Pro mac (stuck on the GRUB menu)


echupriyanov:
### Description

I'm trying to start any arm64 RHEL 8 based Linux using lima with provided examples.
But it even can't boot and stuck on the GRUB menu:
<img width="889" alt="Снимок экрана 2022-05-03 в 00 03 41" src="https://user-images.githubusercontent.com/2070616/166327481-2f9e21b9-5146-4918-a749-13fe4210ef4f.png">

x86-64 VMs are working fine
Ubuntu arm64 is working fine too.
The only issue is with RHEL8 based distros. What can be wrong here?

sdadh01:
I have the same issue (M1 Pro Mac, latest versions of everything). Before the grub menu appears in the serial.log this appears until grub times out:

```
EFI stub: ERROR: This 64 KB granular kernel is not supported by your CPU

  Failed to boot both default and fallback entries.

Press any key to continue...
```
Subtly different error to [issue 712](https://github.com/lima-vm/lima/issues/712) but maybe the same cause?

blissdev:
I'm experiencing this issue as well on an M1 Max MBP.

AkihiroSuda:
According to https://forums.rockylinux.org/t/installation-of-rocky-not-proceeding-in-macbook-pro-m1-silicon-chip/4734 :

> It’s unfortunately not possible right now. The aarch64 kernel assumes a 64k page size that is NOT supported on M1 macs. M1 page size is 4k or 16k.
> 
> EL9 should work on M1.

AkihiroSuda:
Update: Oracle Linux 8 seems to work on M1 (thanks to @plexilcrf)

- https://github.com/lima-vm/lima/discussions/3522

PR #1120, opened 2022-10-18
templates: Promote EL9 from experimental

### LINKED ISSUE #1099: Promote `{rocky, almalinux, oracle, centos-stream}-9` from experimental
EL8 distros do not work on M1:
- https://github.com/lima-vm/lima/issues/841

So I'm planning to promote `{rocky, almalinux, oracle, centos-stream}-9` from experimental:
e.g., `experimental/rocky-9.yaml` -> `rocky.yaml`

The old `rocky.yaml` will be renamed to `rocky-8.yaml`, or `deprecated/rocky-8.yaml` if we want to deprecate EL8 now.

AkihiroSuda:
- Moved `rocky.yaml` to `rocky-8.yaml`
- Moved `experimental/rocky-9.yaml` to `rocky-9.yaml`
- Symlinked `rocky-9.yaml` to `rocky.yaml`

Same for almalinux, oraclelinux, and centos-stream too.

EL8 are now "soft" deprecated as they are known not to work on M1 (#841). If the issue is not going to be fixed in EL 8.7, maybe we should officially deprecate EL8.
    
Fix #1099


jandubois (dismissed):
LGTM

I'm wondering though if each section in `examples/README.md` shouldn't be sorted alphabetically, just for easier browsability.

At first I was irritated that the `foo-9` distros are sorted before `foo-8`, but then noticed that a couple others are also in non-alpha-order. I assume it is meant to show more important ones first, but then `ubuntu` is listed all the way at the bottom in Distros (and `vmnet` is all the way at the top for Others). I don't care too much, but personally would prefer all sections sorted by name.

AkihiroSuda:
Updated to sort the templates

jandubois (approved):
Thanks!

PR #1121, opened 2022-10-18
Add singularity and run-singularity wrappers

afbjorklund:
This is for compatibility with older versions of apptainer when it was called still singularity, and to support the .sif file format.

These self-contained (pun intended) files are supposed to be executable, so that you can "run" them by just executing them:

```console
$ apptainer build /tmp/lima/debian.sif docker://debian
INFO:    Starting build...
Copying blob f606d8928ed3 skipped: already exists  
Copying config 0311b76201 done  
Storing signatures
2022/10/18 19:53:52  info unpack layer: sha256:f606d8928ed378229f2460b94b504cca239fb906efc57acbdf9340bd298d5ddf
INFO:    Creating SIF file...
INFO:    Build complete: /tmp/lima/debian.sif
$ singularity sif list /tmp/lima/debian.sif
------------------------------------------------------------------------------
ID   |GROUP   |LINK    |SIF POSITION (start-end)  |TYPE
------------------------------------------------------------------------------
1    |1       |NONE    |32176-32208               |Def.FILE
2    |1       |NONE    |32208-36135               |JSON.Generic
3    |1       |NONE    |36135-36227               |JSON.Generic
4    |1       |NONE    |36864-51163136            |FS (Squashfs/*System/amd64)
$ file /tmp/lima/debian.sif 
/tmp/lima/debian.sif: a /usr/bin/env run-singularity script executable (binary data)
$ /tmp/lima/debian.sif
Apptainer> 
```

All three programs are included in the apptainer rpm:

```console
[anders@lima-apptainer lima]$ rpm -ql apptainer | grep /usr/bin
/usr/bin/apptainer
/usr/bin/run-singularity
/usr/bin/singularity
```

So make them (all three) available, also on the Mac ?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,6 +124,9 @@ diagrams: docs/lima-sequence-diagram.png
 docs/lima-sequence-diagram.png: docs/images/lima-sequence-diagram.puml
 	$(PLANTUML) ./docs/images/lima-sequence-diagram.puml
 
+# Simple wrapper for 'apptainer run' (.sif)
+RUN_SINGULARITY=./cmd/run-singularity.lima
+
 .PHONY: install
 install: uninstall
 	mkdir -p "$(DEST)"
@@ -131,6 +134,8 @@ install: uninstall
 	( cd _output && tar c * | tar Cxv "$(DEST)" )
 	if [ "$(shell uname -s )" != "Linux" -a ! -e "$(DEST)/bin/nerdctl" ]; then ln -sf nerdctl.lima "$(DEST)/bin/nerdctl"; fi
 	if [ "$(shell uname -s )" != "Linux" -a ! -e "$(DEST)/bin/apptainer" ]; then ln -sf apptainer.lima "$(DEST)/bin/apptainer"; fi
+	if [ "$(shell uname -s )" != "Linux" -a ! -e "$(DEST)/bin/singularity" ]; then ln -sf apptainer.lima "$(DEST)/bin/singularity"; fi
+	if [ "$(shell uname -s )" != "Linux" -a ! -e "$(DEST)/bin/run-singularity" ]; then cp -a $(RUN_SINGULARITY) "$(DEST)/bin/run-singularity"; fi
 
 .PHONY: uninstall
 uninstall:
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
I'm not sure about this.
I'd prefer to reserve `singularity` for Singularity CE (if somebody wants to contribute)

afbjorklund:
> I'd prefer to reserve `singularity` for Singularity CE (if somebody wants to contribute)

Valid point, but _could_ have "run-singularity" call "singularity" (which is program or symlink?)

You get a warning if you have both installed, but that's otherwise about it - as far as I know.

```
WARNING: /usr/local/etc/singularity/ exists, migration to apptainer by system administrator is not complete
```

afbjorklund:
As usual, you can run "apptainer" to get the local version on Linux and "apptainer.lima" to get the remote version (from VM)

Same as with nerdctl/docker/podman ?

If "singularity.lima" was added as a singularity-ce.yaml example, that would also work the same way...

But it seems unlikely, there is already Singularity Desktop for Mac.

DrDaveD:
@AkihiroSuda The singularity package name really belongs with the apptainer project.  If somebody wants to contribute a SingularityCE version, they should use a new package named singularity-ce, because they [forked the project](https://sylabs.io/2021/05/singularity-community-edition/).  As Anders said, the apptainer package still includes `singularity` and `run-singularity` binaries and it has to for compatibility.  In any case, his latest update here should work with either an `apptainer` or a future `singularity-ce` package.

afbjorklund:
This improvement can come later, it is not _necessary_ to get started with apptainer (but a nice feature)

It should also include the support for SINGULARITY_HOME and SINGULARITY_BINDPATH variables.

AkihiroSuda:
I don't know who is the current owner of (the unregistered? trademark of)  "Singularity", but Apptainer users should just switch to use the `apptainer` command, so I'm not inclined to merge this PR. Sorry.

afbjorklund:
No worries, it is needed to support the "self-executable" feature of the format (the shebang), but it is still possible to `run` them with the program

PR #1125, opened 2022-10-20
Use APPTAINER_BINDPATH instead of APPTAINER_HOME

afbjorklund:
Not all software was happy with a read-only $HOME directory.

Using APPTAINER_HOME changed the $HOME in the container.

----

So use the regular VM $HOME, but do bind-mount host $HOME.

Using APPTAINER_BINDPATH is the same as using the `-B` flag.

### DIFF
--- a/cmd/apptainer.lima
+++ b/cmd/apptainer.lima
@@ -1,10 +1,15 @@
 #!/bin/sh
 set -eu
 : "${LIMA_INSTANCE:=apptainer}"
+: "${APPTAINER_BINDPATH:=}"
 
 if [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
   echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl create --name=$LIMA_INSTANCE template://apptainer\` to create a new instance" >&2
   exit 1
 fi
 export LIMA_INSTANCE
-exec lima APPTAINER_HOME="$HOME" apptainer "$@"
+if [ -n "$APPTAINER_BINDPATH" ]; then
+  APPTAINER_BINDPATH="$APPTAINER_BINDPATH,"
+fi
+APPTAINER_BINDPATH="$APPTAINER_BINDPATH$HOME"
+exec lima APPTAINER_BINDPATH="$APPTAINER_BINDPATH" apptainer "$@"


cmd/apptainer.lima
```diff
@@ -7,4 +7,4 @@ if [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
   exit 1
 fi
 export LIMA_INSTANCE
-exec lima APPTAINER_HOME="$HOME" apptainer "$@"
+exec lima APPTAINER_BINDPATH="$HOME" apptainer "$@"
```
DrDaveD, line 10: It would be better to not assume that APPTAINER_BINDPATH isn't already set in the environment.  Extend it if it is already set, with a comma separator.

Even worse, there's the possibility that SINGULARITY_BINDPATH is set as well, which apptainer would honor (with a deprecation warning) if you hadn't set APPTAINER_BINDPATH here, but setting only APPTAINER_BINDPATH here overrides that (with another warning).  So the safest thing to do is something like this:
```
if [ -n "$APPTAINER_BINDPATH" ]; then
    SINGULARITY_BINDPATH="$APPTAINER_BINDPATH"
fi
if [ -n "$SINGULARITY_BINDPATH" ]; then
  SINGULARITY_BINDPATH="$SINGULARITY_BINDPATH,"
fi
SINGULARITY_BINDPATH="$SINGULARITY_BINDPATH$HOME"
exec lima APPTAINER_BINDPATH="$SINGULARITY_BINDPATH" apptainer "$@"
```
I deliberately didn't export SINGULARITY_BINDPATH because it isn't necessary.  On the other hand if it was previously exported, it won't hurt -- apptainer will see they're both the same and let it go.  I am assuming that lima passes all of its environment to apptainer except for anything overridden on the command line..

DrDaveD (changes requested):

cmd/apptainer.lima
afbjorklund: The environment is being filtered through the ssh command, which has `ssh_config` (`SendEnv`) and `sshd_config` (`AcceptEnv`) setting. In this wrapper script, we are _explicitly_ setting it in the wrapper script (i.e. for bash) being invoked...

`DEBU[0000] executing ssh (may take a long)): [/usr/bin/ssh -F /dev/null -o IdentityFile="/home/anders/.lima/_config/user" -o IdentityFile="/home/anders/.ssh/id_rsa" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o NoHostAuthenticationForLocalhost=yes -o GSSAPIAuthentication=no -o PreferredAuthentications=publickey -o Compression=no -o BatchMode=yes -o IdentitiesOnly=yes -o Ciphers="^aes128-gcm@openssh.com,aes256-gcm@openssh.com" -o User=anders -o ControlMaster=auto -o ControlPath="/home/anders/.lima/apptainer/ssh.sock" -o ControlPersist=5m -t -o SendEnv="COLORTERM" -q -p 33473 127.0.0.1 -- cd /home/anders/lima || cd /home/anders ; exec "$SHELL" --login -c 'APPTAINER_BINDPATH=/home/anders apptainer version']`

If we think that it is going to be a common thing to install Singularity (Desktop?) on the Mac, then we can make more efforts in apptainer.lima and run-singularity scripts to allow for `singularity` and `SINGULARITY_BINDPATH` being present ?

* #1121

afbjorklund: The code snippet didn't work, partly due to unset variables (`-u`) - but it had some other issues as well.

`WARNING: While bind mounting '/home/anders:/home/anders': destination is already in the mount point list`

afbjorklund: Fixed APPTAINER_BINDPATH, can add singularity support in the other PR (i.e. SINGULARITY_BINDPATH)

afbjorklund:
@AkihiroSuda : it would be nice to have this bugfix in 0.13.0, if it is possible

* #1101

cmd/apptainer.lima
```diff
@@ -1,10 +1,17 @@
...
 : "${LIMA_INSTANCE:=apptainer}"
+: "${APPTAINER_BINDPATH:=}"
 
 if [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
   echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl start --name=$LIMA_INSTANCE template://apptainer\` to create a new instance" >&2
   exit 1
 fi
 export LIMA_INSTANCE
-exec lima APPTAINER_HOME="$HOME" apptainer "$@"
+if [ "$APPTAINER_BINDPATH" != "$HOME" ]; then
```
DrDaveD, line 11: I think this 'if' is unnecessary, but it doesn't hurt.

DrDaveD (approved):

cmd/apptainer.lima
afbjorklund: I can remove it, it was for the corner case of alreadying being set

afbjorklund: `WARNING: While bind mounting '/home/anders:/home/anders': destination is already in the mount point list`

DrDaveD: But won't apptainer add that if $HOME is /home/anders regardless of whether or not it is in APPTAINER_BINDPATH?  By default it binds the home directory.  Who would have put it in APPTAINER_BINDPATH?

afbjorklund: That was if you did it explicitly, but since it was only a warning not an error (my bad) there is no need to watch out for it...

DrDaveD (approved):

afbjorklund:
Rebased on top of `v0.13.0` release, will recommend `limactl shell apptainer` until it is fixed.

https://apptainer.org/docs/admin/main/installation.html#id3

AkihiroSuda (approved):

AkihiroSuda:
Sorry, I should have approved this much earlier

PR #1132, opened 2022-10-25
feat(shell): allow configuring ssh alias

### LINKED ISSUE #1131: `limactl shell` does not consider potential `ssh` shell aliases
### Description

_This issue is somewhere between a bug report and a feature request._

`lima shell` always invokes the `ssh` executable instead of whichever `ssh` command the user may have defined in their own shell (e.g. via a shell alias):

https://github.com/lima-vm/lima/blob/8cc56933cef9808e2b62981f231adb8dd56a5246/cmd/limactl/shell.go#L129

My concern with this approach is that the potential of terminal emulators with advanced shell integrations over SSH, such as [kitty](https://sw.kovidgoyal.net/kitty/shell-integration/#shell-integration-over-ssh), cannot be fully utilized with `limactl`.

For example, I alias `ssh` to `kitty +kitten ssh`.
To be able to benefit from SSH integrations provided by my terminal emulator, I currently need to export the SSH config of my VMs using `limactl show-ssh` then use plain SSH commands.

I can think about three ways to approach this:

- Run `ssh` commands through the active shell instead of directly (e.g. `/bin/bash -c ssh`).
    Potential problem: we need to use the actual user's shell instead of a hardcoded default value.

- If `LIMA_SHELL` is set, use that value.
    Potential problem: the user might legitimately want to use different shells inside and outside the VM.

- Introduce an optional `LIMASHELL_SSH` setting, and use that if defined. (_my favourite_)

antoineco:
If the `SSH` environment variable is set, `limactl shell` will consider this as an alias for the `ssh` command to execute, instead of looking up a `ssh` executable in the PATH.

The provided command can be anything that is CLI-compatible with the OpenSSH client, such as a shell script that passes all provided arguments to `ssh`.

Closes #1131 

---

Demo

```sh
export SSH='kitty +kitten ssh'
```

```sh
go run ./cmd/limactl/ shell my-vm
```

```console
$ ps ww
16004 s000  S+     0:00.07 kitty +kitten ssh -F /dev/null -o IdentityFile="/Users/acotten/.lima/_config/user" -o IdentityFile="/Users/acotten/.ssh/id_ed25519" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o NoHostAuthenticationForLocalhost=yes -o GSSAPIAuthentication=no -o PreferredAuthentications=publickey -o Compression=no -o BatchMode=yes -o IdentitiesOnly=yes -o Ciphers="^aes128-gcm@openssh.com,aes256-gcm@openssh.com" -o User=acotten -o ControlMaster=auto -o ControlPath="/Users/acotten/.lima/my-vm/ssh.sock" -o ControlPersist=5m -t -o SendEnv="COLORTERM" -q -p 51808 127.0.0.1 -- cd /Users/acotten/git/antoineco/lima-vm-lima || cd /Users/acotten ; exec "$SHELL" --login
```

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -12,13 +12,21 @@ import (
 	"github.com/lima-vm/lima/pkg/sshutil"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/mattn/go-isatty"
+	"github.com/mattn/go-shellwords"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
+// Environment variable that allows configuring the command (alias) to execute
+// in place of the 'ssh' executable.
+const envShellSSH = "SSH"
+
 var shellHelp = `Execute shell in Lima
 
 lima command is provided as an alias for limactl shell $LIMA_INSTANCE. $LIMA_INSTANCE defaults to "` + DefaultInstanceName + `".
+
+By default, the first 'ssh' executable found in the host's PATH is used to connect to the Lima instance.
+A custom ssh alias can be used instead by setting the $` + envShellSSH + ` environment variable.
 
 Hint: try --debug to show the detailed logs, if it seems hanging (mostly due to some SSH issue).
 `
@@ -126,9 +134,28 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		)
 	}
 
-	arg0, err := exec.LookPath("ssh")
-	if err != nil {
-		return err
+	var arg0 string
+	var arg0Args []string
+
+	if sshShell := os.Getenv(envShellSSH); sshShell != "" {
+		sshShellFields, err := shellwords.Parse(sshShell)
+		switch {
+		case err != nil:
+			logrus.WithError(err).Warnf("Failed to split %s variable into shell tokens. "+
+				"Falling back to 'ssh' command", envShellSSH)
+		case len(sshShellFields) > 0:
+			arg0 = sshShellFields[0]
+			if len(sshShellFields) > 1 {
+				arg0Args = sshShellFields[1:]
+			}
+		}
+	}
+
+	if arg0 == "" {
+		arg0, err = exec.LookPath("ssh")
+		if err != nil {
+			return err
+		}
 	}
 
 	sshOpts, err := sshutil.SSHOpts(inst.Dir, *y.SSH.LoadDotSSHPubKeys, *y.SSH.ForwardAgent, *y.SSH.ForwardX11, *y.SSH.ForwardX11Trusted)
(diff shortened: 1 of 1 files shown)


cmd/limactl/shell.go
```diff
@@ -126,9 +130,24 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
 	}
 
-	arg0, err := exec.LookPath("ssh")
-	if err != nil {
-		return err
+	var arg0 string
+	var arg0Args []string
+
+	if sshShell := os.Getenv(envShellSSH); sshShell != "" {
+		sshShellFields := strings.Fields(sshShell)
```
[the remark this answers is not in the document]
antoineco: I had simple use cases in mind, but you're raising an excellent point.

The feedback I received in the linked issue seems to indicate that the implementation is going to change to pass arguments to `$SHELL -c` instead of overriding the command in Lima. I'll leave this open for a bit in case somebody wants to chime in.

antoineco:
Closing in favor of a different implementation (TBD).
See PR comment and linked issue.

cmd/limactl/shell.go
```diff
@@ -41,6 +41,10 @@ func newShellCommand() *cobra.Command {
 	return shellCmd
 }
 
+// Environment variable that allows configuring the command (alias) to execute
+// in place of the 'ssh' executable.
+const envShellSSH = "LIMASHELL_SSH"
```
AkihiroSuda, line 46:
```suggestion
const envShellSSH = "SSH"
```

I think this is safe and probably compatible with other applications that support custom ssh command

cmd/limactl/shell.go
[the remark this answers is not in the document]
antoineco: Done in [b6645274..5ca81769](https://github.com/lima-vm/lima/compare/b66452748ceae0d89c7da0acdfeb5c1db52098b9..5ca81769c4da3a72edb29156ceea99464a745758).

PTAL

AkihiroSuda (dismissed):
Thanks

go.mod
```diff
@@ -17,6 +17,7 @@ require (
 	github.com/foxcpp/go-mockdns v1.0.0
 	github.com/goccy/go-yaml v1.9.6
 	github.com/google/go-cmp v0.5.9
+	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
```
AkihiroSuda, line 20: We have been using go-shellwords: https://github.com/lima-vm/lima/blob/020d7ee10b219822d4c59d72c3764f3e1ee3baa6/pkg/qemu/qemu.go#L616

Not sure which one of them is better though

antoineco: Thanks for the pointer, I didn't even see this. 

Done in [ce109f95..648c47d4](https://github.com/lima-vm/lima/compare/ce109f9548a3ce3381a49eed03011c112d953e5d..648c47d49f8e81ebb738530cd593aa16401d90a6).

AkihiroSuda (approved):
Thanks

jandubois (approved):
Thanks! LGTM and works as expected.

PR #1133, opened 2022-10-25
feat(shell): execute ssh command through shell

### LINKED ISSUE #1131: `limactl shell` does not consider potential `ssh` shell aliases

antoineco:
If the `SHELL` environment variable is set, `limactl shell` will execute the `ssh` command and its arguments through the user's shell instead of looking up a `ssh` executable in the PATH.

This allows supporting user-defined `ssh` commands such as shell aliases and functions.

Fixes #1131

---

Demo:

```sh
alias ssh='kitty +kitten ssh'
```

```console
$ go run ./cmd/limactl/ shell --debug my-vm
DEBU[0000] changeDirCmd="cd /Users/acotten/git/antoineco/lima-vm-lima || cd /Users/acotten" 
DEBU[0000] OpenSSH version 8.6.1 detected               
DEBU[0000] executing ssh (may take a long)): [/bin/zsh -i -l -c ssh -F /dev/null -o 'IdentityFile="/Users/acotten/.lima/_config/user"' -o 'IdentityFile="/Users/acotten/.ssh/id_ed25519"' -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o NoHostAuthenticationForLocalhost=yes -o GSSAPIAuthentication=no -o PreferredAuthentications=publickey -o Compression=no -o BatchMode=yes -o IdentitiesOnly=yes -o 'Ciphers="^aes128-gcm@openssh.com,aes256-gcm@openssh.com"' -o User=acotten -o ControlMaster=auto -o 'ControlPath="/Users/acotten/.lima/my-vm/ssh.sock"' -o ControlPersist=5m -t -o 'SendEnv="COLORTERM"' -q -p 49316 127.0.0.1 -- 'cd /Users/acotten/git/antoineco/lima-vm-lima || cd /Users/acotten ; exec "$SHELL" --login'] 
```

```console
$ ps ww
...
 6736 s000  S+     0:00.10 kitty +kitten ssh -F /dev/null -o IdentityFile="/Users/acotten/.lima/_config/user" ...
```

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -40,6 +41,10 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().String("workdir", "", "working directory")
 	return shellCmd
 }
+
+// envShell contains the user's shell, as specified by the password database.
+// http://man.openbsd.org/login.1#SHELL
+const envShell = "SHELL"
 
 func shellAction(cmd *cobra.Command, args []string) error {
 	// simulate the behavior of double dash
@@ -126,17 +131,13 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		)
 	}
 
-	arg0, err := exec.LookPath("ssh")
-	if err != nil {
-		return err
-	}
-
 	sshOpts, err := sshutil.SSHOpts(inst.Dir, *y.SSH.LoadDotSSHPubKeys, *y.SSH.ForwardAgent, *y.SSH.ForwardX11, *y.SSH.ForwardX11Trusted)
 	if err != nil {
 		return err
 	}
 	sshArgs := sshutil.SSHArgsFromOpts(sshOpts)
-	if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {
+	isTerm := isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd())
+	if isTerm {
 		// required for showing the shell prompt: https://stackoverflow.com/a/626574
 		sshArgs = append(sshArgs, "-t")
 	}
@@ -151,7 +152,24 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		"--",
 		script,
 	}...)
-	sshCmd := exec.Command(arg0, sshArgs...)
+
+	var sshCmd *exec.Cmd
+
+	if usrShell := os.Getenv(envShell); isTerm && usrShell != "" {
+		sshCmdStr := shellescape.QuoteCommand(append([]string{"ssh"}, sshArgs...))
+		shellArgs := []string{"-i", "-c", sshCmdStr}
+		if !isCsh(usrShell) {
+			shellArgs = append([]string{"-l"}, shellArgs...)
+		}
+		sshCmd = exec.Command(usrShell, shellArgs...)
+	} else {
+		arg0, err := exec.LookPath("ssh")
+		if err != nil {
+			return err
+		}
+		sshCmd = exec.Command(arg0, sshArgs...)
+	}
+
 	sshCmd.Stdin = os.Stdin
 	sshCmd.Stdout = os.Stdout
 	sshCmd.Stderr = os.Stderr
@@ -163,4 +181,18 @@ func shellAction(cmd *cobra.Command, args []string) error {
 
 func shellBashComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 	return bashCompleteInstanceNames(cmd)
+}
+
+const cshSuffix = "csh"
+
+// isCsh returns whether the given shell is the C shell (csh) or any of its variants.
+// The shell parameter can be either a shell name ("*sh") or the path to a
+// shell executable ("/bin/*sh").
+func isCsh(shell string) bool {
+	suffixLen := len(cshSuffix)
+	basename := filepath.Base(shell)
+	if len(basename) < suffixLen {
+		return false
+	}
+	return basename[len(basename)-suffixLen:] == cshSuffix
 }


cmd/limactl/shell.go
```diff
@@ -151,7 +150,20 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		"--",
 		script,
 	}...)
-	sshCmd := exec.Command(arg0, sshArgs...)
+
+	var sshCmd *exec.Cmd
+
+	if usrShell := os.Getenv(envShell); usrShell != "" {
+		sshCmdStr := shellescape.QuoteCommand(append([]string{"ssh"}, sshArgs...))
+		sshCmd = exec.Command(usrShell, "-i", "-l", "-c", sshCmdStr)
```
antoineco, line 158: Ideally, instead of always passing `-i` and `-l`, we should detect whether the current shell is actually a _login_ shell, and whether it is _interactive_.
Unfortunately, there doesn't seem to exist a portable way to do that: [How to check if a shell is login/interactive/batch](https://unix.stackexchange.com/q/26676/152409) (StackExchange - Unix & Linux).

antoineco:
The failing test on Debian seems possibly unrelated (?):

> ```time="2022-10-25T17:43:29Z" level=fatal msg="rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/"```

chancez:
Oh this sounds awesome. I have some `ssh` configs so I can SSH to my VMs to deal with this problem, this will simplify things!

cmd/limactl/shell.go
jandubois: Note that `csh` (is anybody still using that?) does not support the `-l` option unless it is the only option given.

antoineco: Right, good point. [csh(1)](https://manpages.org/csh) also mentions

> **-i**
> [...] Shells are interactive without this option if their inputs and outputs are terminals.

Which makes `-i` unnecessary. But passing it explicitly doesn't seem to cause any side effect.

If `csh` is the only exception I guess we can have a simple check for `basename == csh`?

cmd/limactl/shell.go
```diff
@@ -151,7 +151,20 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		"--",
 		script,
 	}...)
-	sshCmd := exec.Command(arg0, sshArgs...)
+
+	var sshCmd *exec.Cmd
+
+	if usrShell := os.Getenv(envShell); isTerm && usrShell != "" {
```
AkihiroSuda, line 157: Doesn't `$SHELL` work when `!isTerm` ?

cmd/limactl/shell.go
jandubois: > If csh is the only exception

`tcsh` works the same as `csh` 😄

cmd/limactl/shell.go
```diff
@@ -41,6 +41,10 @@ func newShellCommand() *cobra.Command {
 	return shellCmd
 }
 
+// envShell contains the user's shell, as specified by the password database.
+// http://man.openbsd.org/login.1#SHELL
```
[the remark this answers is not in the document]
antoineco: I linked the first man page I found that was documenting the SHELL env var.
I can also either link to manpages.org, or remove the link altogether.

cmd/limactl/shell.go
antoineco: It does, but what would be the benefit of shelling out if not in a terminal?

AkihiroSuda: I thought the aliases are expected to be honored regardless to presence of tty

cmd/limactl/shell.go
antoineco: Dang.
Let's introduce a trivial `isCsh()` helper then? It's hard to imagine covering all edge cases, but at least those ones should be fairly easy to get right.

AkihiroSuda: > `IsCsh()`

`isBash() || isZsh()`  might be safer

jandubois: `|| isFish()` 🐠

cmd/limactl/shell.go
antoineco: There are some ramifications to consider:

- `.<X>sh_profile` is only read if the shell is a _login_ shell
- `.<X>shrc` is only read if the shell is an _interactive_ shell

Omitting `-l` makes shelling out useless, since neither `.<X>sh_profile` nor `.<X>shrc` are considered in this case.
So we _always_ need to pass `-l` to make the feature useful.

Setting `-i` without a tty prints an error message, so we shouldn't set in in the absence of tty.

Are we OK with this?

AkihiroSuda: Sorry to go back and forth, but maybe just support custom ssh command via `$SSH` might be better?
An equivalent of the option 3 in the original issue but with a simpler env var name.

cmd/limactl/shell.go
antoineco: That sounds a bit restrictive to me to assume Lima users are limited to Bash, Zsh and Fish users.

Although I agree that might cover 95% of cases or more, simply checking for `sh|ash|bash|zsh|fish` doesn't cover the full range of options out there, whereas eliminating `/*csh` seems to eliminate the few exceptions that don't support `-il` well.

You decide.

antoineco: Current status: [60b58339..f4ebbc67](https://github.com/lima-vm/lima/compare/60b58339fe79cb88018a37b50b2a45b55560bc35..f4ebbc6795cdd26a35f70a5d22c8932e0aa63941)

Demo with tcsh (macOS), notice `-l` was omitted:

```
DEBU[0000] executing ssh (may take a long)): [/bin/csh -i -c ssh -F /dev/null -o 'Identi...
```

Happy to revisit.

cmd/limactl/shell.go
antoineco: (Ref. #1132 for completeness)

antoineco: I like both approaches equally really.
The env var offers control, the shell makes things work without extra config.

@jandubois what's your opinion?

cmd/limactl/shell.go
antoineco: Marking as resolved because the original comment was about whether or not to rely on `IsTerminal()`.

Let's open a new thread to discuss further concerns with shell type detection (if any).

cmd/limactl/shell.go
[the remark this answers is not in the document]
antoineco: Marking as resolved because OpenBSD's man page for `login(1)` seems to be the only publicly available documentation to have ever documented the `SHELL` environment variable.

If this is causing friction I can also remove the comment.

cmd/limactl/shell.go
chancez: Why an env-var and not just a config option in the lima.yaml? I could see either/both making sense.

jandubois: > The env var offers control, the shell makes things work without extra config.
> 
> @jandubois what's your opinion?

Given the issues we ran into making it work with different shells, login shells vs. interactive shells, and how they are treated differently on macOS vs. Linux, I think I prefer adding the `LIMA_SSH` variable.

I think I prefer `LIMA_SSH` over `SSH`, but probably can be convinced otherwise. 😄 

> Why an env-var and not just a config option in the lima.yaml? I could see either/both making sense.

I do not view this as an instance-specific setting, so I think `lima.yaml` is the wrong place for it. Why would you want different `ssh` commands for different instances?

antoineco: Agree about this not being instance specific.

I would prefer to avoid `SSH` as well, this is too generic, not in Lima's namespace.

I like `LIMASHELL_SSH` better than other options because it tells exactly what command it relates too, whereas `LIMA_` could have a very broad scope.

I'm re-opening #1132.

antoineco:
Reverting to #1132 based on maintainers' decision.

PR #1138, opened 2022-10-31
pkg/networks: wait for daemon to stop after killing

pendo324:
Currently, when running an instance with a `socket_vmnet` network configured, if you try the following:

```
limactl stop default && limactl start default
```

you may notice that the `limactl start default` command consistently fails. In my testing, this is because the `pkill` command is issued, but the `socket_vmnet` process hasn't actually stoped yet.

This is what the hostagent logs look like on my system:

```
{"level":"fatal","msg":"template: :1:21: executing \"\" at \u003cfd_connect \"/private/var/run/$NETWORK_NAME/socket_vmnet.$NETWORK_NAME\"\u003e: error calling fd_connect: fd_connect: dial unix /private/var/run/$NETWORK_NAME/socket_vmnet.$NETWORK_NAME: connect: connection refused","time":"2022-10-31T18:31:21-04:00"}
```

Either waiting a few seconds between the stop and start command, or removing this conditional check seems to resolve the issue, but adding another check (like `if daemon == networks.VDEVMNet || daemon == networks.SocketVMNet`) also works, I just wan't sure which one to go with.

### DIFF
--- a/pkg/networks/reconcile/reconcile.go
+++ b/pkg/networks/reconcile/reconcile.go
@@ -226,19 +226,18 @@ func stopNetwork(config *networks.YAML, name string) error {
 				return err
 			}
 		}
-		// wait for VMNet to terminate (up to 5s) before stopping Switch, otherwise the socket may not get deleted
-		if daemon == networks.VDEVMNet {
-			startWaiting := time.Now()
-			for {
-				if pid, _ := store.ReadPIDFile(config.PIDFile(name, daemon)); pid == 0 {
-					break
-				}
-				if time.Since(startWaiting) > 5*time.Second {
-					logrus.Infof("%q daemon for %q network still running after 5 seconds", daemon, name)
-					break
-				}
-				time.Sleep(500 * time.Millisecond)
+		// wait for daemons to terminate (up to 5s) before stopping, otherwise the sockets may not get deleted which
+		// will cause subsequent start commands to fail.
+		startWaiting := time.Now()
+		for {
+			if pid, _ := store.ReadPIDFile(config.PIDFile(name, daemon)); pid == 0 {
+				break
 			}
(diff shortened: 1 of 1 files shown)


jandubois:
I think this is fine by me, even though it does not solve the general problem of having a race condition in `limactl stop vm1 && limactl start vm2` if both VMs use the same shared network (with no other VMs using it at the same time).

I don't think solving this race condition in general is worth the effort.

Extending the waiting period so that you can stop and restart the same VM does make sense to me though; the only request would be to have more extensive documentation of why we do the waiting. I think the existing comment is no longer covering all the use cases.

@AkihiroSuda WDYT?

AkihiroSuda:
SGTM

pendo324:
Thanks for the feedback. Any suggestions for the comment? Also, do you think this PR should implement any smarter checks, or can that come in a later PR?

jandubois:
> Any suggestions for the comment?

Just explain that we are waiting for all daemons to stop in case the user is just restarting the VM. Otherwise the sockets might not be deleted in time, and the restart would fail.

> Also, do you think this PR should implement any smarter checks, or can that come in a later PR?

That can be in a later PR.

---

Please rebase your PR on latest `master` to fix the error with the integration tests on Github macOS runners!

pendo324:
> Just explain that we are waiting for all daemons to stop in case the user is just restarting the VM. Otherwise the sockets might not be deleted in time, and the restart would fail.

Got it, just pushed a new version with an updated comment, let me know what you think.
 
> Please rebase your PR on latest `master` to fix the error with the integration tests on Github macOS runners!

Ah, didn't realize that was why the test was failing, just rebased,

jandubois (approved):
Thanks, LGTM

AkihiroSuda (approved):
Thanks

PR #1143, opened 2022-11-02
feat: configurable start timeout

### LINKED ISSUE #418: Provide a way to pass in the wait timeout for emulation mode
Since the emulation mode like arm on intel or intel on arm will take a lot of time to create the vm, it would be nice to pass in the timeout. The default could be 10min as it is now. But in emulation mode one could override it by passing in as a parameter for both start and stop commands.

antoineco:
Allows configuring the duration to wait for the "running" event from the host agent before timing out.

Closes #418

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -53,6 +53,7 @@ $ limactl start --name=default https://raw.githubusercontent.com/lima-vm/lima/ma
 	startCommand.Flags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "enable TUI interactions such as opening an editor, defaults to true when stdout is a terminal")
 	startCommand.Flags().String("name", "", "override the instance name")
 	startCommand.Flags().Bool("list-templates", false, "list available templates and exit")
+	startCommand.Flags().Duration("timeout", start.DefaultWatchHostAgentEventsTimeout, "duration to wait for the instance to be running before timing out")
 	return startCommand
 }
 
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

cmd/limactl/start.go
```diff
@@ -53,6 +53,7 @@ $ limactl start --name=default https://raw.githubusercontent.com/lima-vm/lima/ma
 	startCommand.Flags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "enable TUI interactions such as opening an editor, defaults to true when stdout is a terminal")
 	startCommand.Flags().String("name", "", "override the instance name")
 	startCommand.Flags().Bool("list-templates", false, "list available templates and exit")
+	startCommand.Flags().Duration("timeout", start.DefaultWatchHostAgentEventsTimeout, "duration to wait for the instance to be running before timing out")
```
jandubois, line 56: I would find it helpful if the help text would show the default timeout value for this setting. It should also mention if the duration is to be specified in seconds or minutes, although may this would be implied when you include the default value, i.e. either "10 minutes" or "600 seconds".

jandubois (changes requested):

cmd/limactl/start.go
antoineco: @jandubois I'm not sure I understand. The flags library used in this project _does_ print the default value in the command's help, at the end of each flag's description, in the format of a duration string.

Right now it should be showing:

> `duration to wait for the instance to be running before timing out (default: 10m0s)`

when you invoke the command with `--help` or provide an erroneous flag.

Are you suggesting we add _"formatted as a duration string"_ somewhere in the description? I can do that, but IMO this is implied by the default value mentioned above and should be pretty self explanatory.

jandubois: I'm sorry, I was confused. The code is good as-is.

AkihiroSuda:
> @[jandubois](https://github.com/jandubois) jandubois dismissed their [stale review](https://github.com/lima-vm/lima/pull/1143#pullrequestreview-1165640509) [19 minutes ago](https://github.com/lima-vm/lima/pull/1143#event-7727819124)

LGTY?

Issue #1145, opened 2022-11-05
limactl: start: introduce `--set <stringArray>` flag to pin static config values instead of creating custom config file


Dentrax:
### Description

Since limactl already [supports](https://github.com/lima-vm/lima/blob/master/examples/default.yaml) passing custom configuration files, sometimes it decreases the overall UX. It enforces us to create custom config files each time even we want to set just a single variable, such as `localPort`. Here is an example use-case:

So that we can also override the values in a file by using the '--set' flag and pass configuration from the command line as follows:

```
$ limactl start foo -f my_defauly.yaml --set ssh.localPort=15555 --set arch=x86_64
```

We can specify the `--set` flag multiple times. The priority will be given to the last (right-most) set specified.

`--set stringArray`:
```
set values on the command line (can specify multiple or separate values with commas: key1=val1,key2=val2)
```

/cc @developer-guy

AkihiroSuda:
This has been on plan, and will be implemented after merging https://github.com/lima-vm/lima/pull/1069

AkihiroSuda:
See also https://github.com/lima-vm/lima/issues/971#issuecomment-1258833707

AkihiroSuda:
Implemented in 2023:
- #1359

e.g.,
```
limactl start --set='.cpus = 2 | .memory = "2GiB"'
```

PR #1147, opened 2022-11-07
Support for Virtualization.Framework for macOS 13

balajiv113:
This PR provides support for using Virtualization.Framework as a optional driver.

The following are the changes done related to using drivers,
- [x] Driver interface
- [x] Migrate current QEMU implementation to driver
- [x] New driver vz for Virtualization.Framework

The following are the features of lima, the vz driver should provide support for below,
- [x] Running VM (Using both disk img and iso)*
- [x] Slirp network for guest to host communication (uses gvisor-tap-vsock)
- [x] Host to guest network (uses vz NAT gateway)*
- [x] Directory sharing, supports reverse-sshfs and virtiofs (newly added)
- [x] Port forwarding
- [x] Host DNS resolver
- [ ] Display (Deferring it for now as it requires runtime.LockOsThread() to be called in the beginning of hostagent start cmd)

**Notes**
- Vz driver internally converts qcow to raw image using (qemu-img convert). This is because vz only supports raw disk
- Need to provide yaml configuration for NAT

**Know Issues**
- Serial log doesn't contain system boot logs
- Gvisor-tap-vsock will throw some errors like `[e.connection](tcpproxy:)` during start-up (This is because of port 22 Forwards being tried before VM is ready) [Not a failure/blocker just info message thrown, but we can look into fixing in a follow-up to call ssh forward manually]
- When vz vm stops, it doesn't trigger hostagent stop [Fixed]

**Testing**
Tested the following templates with `driver: vz`. All test are done on macOS 13 intel as of now. 

- [x] almalinux.yaml
- [x] alpine.yaml
- [x] apptainer.yaml
- [ ] archlinux.yaml
- [x] buildkit.yaml
- [ ] centos-stream.yaml
- [x] debian.yaml
- [x] docker.yaml
- [x] faasd.yaml
- [x] fedora.yaml
- [x] k3s.yaml
- [x] k8s.yaml
- [x] nomad.yaml
- [ ] opensuse.yaml
- [ ] oraclelinux.yaml
- [x] podman.yaml
- [ ] rocky.yaml
- [x] ubuntu.yaml
- [x] vmnet.yaml


examples/docker.yaml
```diff
@@ -7,6 +7,7 @@
 # $ docker ...
 
 # This example requires Lima v0.8.0 or later
+driver: "vz"
```
AkihiroSuda, line 10: The default must remain qemu

AkihiroSuda: You can create `examples/vz.yaml` for the VZ example. (Maybe under `experimental` dir?)

balajiv113: Yes yes that make sense. Will put it under experimental directory.

pkg/store/filenames/filenames.go
```diff
@@ -42,6 +42,8 @@ const (
 	HostAgentSock      = "ha.sock"
 	HostAgentStdoutLog = "ha.stdout.log"
 	HostAgentStderrLog = "ha.stderr.log"
+	Identifier         = "identifier"
```
AkihiroSuda, line 45: ?

pkg/store/filenames/filenames.go
```diff
@@ -42,6 +42,8 @@ const (
...
+	EFI                = "efi"
```
AkihiroSuda, line 46: ?

pkg/store/filenames/filenames.go
AkihiroSuda: Please describe in https://github.com/lima-vm/lima/blob/master/docs/internal.md

Also, perhaps the filename should be something like `vz-identifier`?

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+
+	fileConn, err := net.FileConn(server)
+	if err != nil {
+		return nil, err
+	}
+
+	err = machine.Start()
+
+	networks.StartGVisor(ctx, &networks.GVisorOpts{
```
AkihiroSuda, line 40: This looks a misnomer.
This starts the network stack that originates from gvisor, but does not start gvisor (sandboxed kernel) actually

pkg/limayaml/limayaml.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 type LimaYAML struct {
+	Driver            *Driver           `yaml:"driver,omitempty" json:"driver,omitempty"`
```
AkihiroSuda, line 10: Should this be called like `VMType` for consistency with `MountType`?

AkihiroSuda: Also please create a document like `docs/vmtype.md` or `docs/vz.md` to explain the usage, requirements, and limitations of vz

pkg/limayaml/limayaml.go
```diff
@@ -43,6 +45,10 @@ const (
 
 	REVSSHFS MountType = "reverse-sshfs"
 	NINEP    MountType = "9p"
+	VIRTIOFS MountType = "virtiofs"
```
AkihiroSuda, line 48: Please explain this in https://github.com/lima-vm/lima/blob/master/docs/mount.md

pkg/networks/gvisor.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+			return vn.AcceptQemu(ctx, opts.Conn)
+		})
+	}
+
+	return nil
+}
+
+func searchDomains() []string {
+	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
```
AkihiroSuda, line 94: Perhaps just `runtime.GOOS != "windows"`

pkg/qemu/imgutil/imgutil.go
```diff
@@ -14,6 +14,18 @@ type Info struct {
 	Format string `json:"format,omitempty"` // since QEMU 1.3
 }
 
+func QcowToRaw(source string, dest string) error {
+	var stdout, stderr bytes.Buffer
+	cmd := exec.Command("qemu-img", "convert", source, dest)
```
AkihiroSuda, line 19: Needs `-O raw`?

pkg/vz/vm_darwin.go
AkihiroSuda: Maybe `StartGVisorNetstack`? `StartNetstack`? `StartGvisorNet`?

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,493 @@
...
+		for {
+			select {
+			case <-ctx.Done():
+				logrus.Println("Context closed, stopping vm")
+				if machine.CanStop() {
+					_, err := machine.RequestStop()
+					logrus.Errorf("Error while stopping the VM %q", err)
+				}
+			case newState := <-machine.StateChangedNotify():
+				if newState == vz.VirtualMachineStateRunning {
```
AkihiroSuda, line 64: Can we use `switch{}` for `newState` comparisons?

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,493 @@
...
+	}
+	consoleConfig, err := vz.NewVirtioConsoleDeviceSerialPortConfiguration(serialPortAttachment)
+	config.SetSerialPortsVirtualMachineConfiguration([]*vz.VirtioConsoleDeviceSerialPortConfiguration{
+		consoleConfig,
+	})
+	return err
+}
+
+func attachNetwork(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfiguration, networkConn *os.File) error {
+	//Slirp network
```
AkihiroSuda, line 189: Did you mean gvisor network?

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,493 @@
...
+
+func getEFI(driver *driver.BaseDriver) (*vz.EFIVariableStore, error) {
+	efi := filepath.Join(driver.Instance.Dir, filenames.EFI)
+	if _, err := os.Open(efi); os.IsNotExist(err) {
+		return vz.NewEFIVariableStore(efi, vz.WithCreatingEFIVariableStore())
+	}
+	return vz.NewEFIVariableStore(efi)
+}
+
+func createFileAndWriteTo(data []byte, path string) error {
```
AkihiroSuda, line 461: `os.WriteFile` may suffice

pkg/limayaml/limayaml.go
balajiv113: Isn't VMType kind of gives a wrong meaning (Type of VM like linux etc).

How about DriverType ?

AkihiroSuda: > Type of VM like linux etc

We only support Linux.
Eventually we may support FreeBSD (if somebody contributes), but that will be probably called "type of guest", not "type of VM"

pkg/vz/vm_darwin.go
balajiv113: Done

pkg/vz/vm_darwin.go
balajiv113: Yes, updated the comments

pkg/vz/vm_darwin.go
balajiv113: Done

pkg/qemu/imgutil/imgutil.go
balajiv113: Default output format is raw, so didn't add it explicitly. 
Let me know if needed, i can i add as well

pkg/limayaml/limayaml.go
balajiv113: Done the VMType change only for yaml config related places. 
Do let me know if you want to change the word driver everywhere (like driver interface, implementations etc).

Will add doc related things as next step.

pkg/vz/vm_darwin.go
balajiv113: Done

examples/docker.yaml
balajiv113: Done.

examples/experimental/vz.yaml
```diff
@@ -0,0 +1,71 @@
...
+  script: |
+    #!/bin/sh
+    sed -i 's/host.lima.internal.*/host.lima.internal host.docker.internal/' /etc/hosts
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    command -v docker >/dev/null 2>&1 && exit 0
+    export DEBIAN_FRONTEND=noninteractive
+    curl -fsSL https://get.docker.com | sh
```
AkihiroSuda, line 33: `vz.yaml` should not have Docker.
You may create `docker-vz.yaml` separately though

pkg/limayaml/limayaml.go
AkihiroSuda: Thanks, "driver" in the impl is ok

examples/experimental/vz.yaml
AkihiroSuda: But anyway no need to keep `docker-vz.yaml` separately with 
- https://github.com/lima-vm/lima/issues/1145

balajiv113: Sure. Then i will keep a default example alone in vz.yaml

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+	networkConfig.SetMACAddress(address)
+
+	configurations := []*vz.VirtioNetworkDeviceConfiguration{
+		networkConfig,
+	}
+	//Other networks
+	for _, network := range driver.Yaml.Networks {
+		//TODO - Handle NAT network type
+		if network.Socket != "" {
```
balajiv113, line 219: @AkihiroSuda 
For NAT network type, shall i go ahead with `network.NAT: true/false` or `network.socket: NAT` ?

AkihiroSuda: What do you mean by "NAT"?

AkihiroSuda, line 219: The `network.Socket` value does not seem consumed

AkihiroSuda: Anyway probably it should be a separate PR

balajiv113: For host to guest network, the below code uses https://developer.apple.com/documentation/virtualization/vznatnetworkdeviceattachment provided from virtualization.framework itself.

This doesn't require sudo access.

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,485 @@
...
+					if err := os.WriteFile(pidFile, []byte(strconv.Itoa(os.Getpid())+"\n"), 0644); err != nil {
+						logrus.Errorf("error writing to pid fil %q", pidFile)
+						cancelHa()
+					}
+					defer os.RemoveAll(pidFile)
+					logrus.Println("[VZ] - VM state change: running")
+				case vz.VirtualMachineStateStopped:
+					logrus.Println("[VZ] - VM state change: stopped")
+					cancelHa()
+				}
```
AkihiroSuda, line 81: Every `switch{}` should have `default:`

pkg/vz/vm_darwin.go
AkihiroSuda: Would it be possible to  just use vznat by default and drop gvisor netstack ?

If vznat has to be optional, probably it should be like `network.VZNAT: true/false`

balajiv113: Yes that should work but with vzNAT there will be other issues like VPN won't be working. 

Also i faced issues with calls to host dns resolver via UDP didn't work

pkg/vz/vz_driver_darwin.go
```diff
@@ -0,0 +1,62 @@
...
+	return &LimaVzDriver{
+		BaseDriver: driver,
+	}
+}
+
+func (l *LimaVzDriver) Validate() error {
+	if *l.Yaml.MountType == limayaml.NINEP {
+		return fmt.Errorf("field `mountType` must be %q or %q for VZ driver , got %q", limayaml.REVSSHFS, limayaml.VIRTIOFS, *l.Yaml.MountType)
+	}
+	return nil
```
AkihiroSuda, line 32: Warnings should be printed for unsupported YAML propertries: https://github.com/containerd/nerdctl/blob/c1a4159035b77a4a739a503b98139a17ac3440b8/pkg/composer/serviceparser/serviceparser.go#L38

Can be another PR though

pkg/vz/vm_darwin.go
AkihiroSuda: Thanks, let's keep using gvisor netstack by default, and revisit VZNAT in a separate PR

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,485 @@
...
+			return nil, err
+		}
+		return machineIdentifier, nil
+	}
+	return vz.NewGenericMachineIdentifierWithDataPath(identifier)
+}
+
+func getEFI(driver *driver.BaseDriver) (*vz.EFIVariableStore, error) {
+	efi := filepath.Join(driver.Instance.Dir, filenames.VzEfi)
+	if _, err := os.Open(efi); os.IsNotExist(err) {
```
Code-Hex, line 458: I think [Stat](https://pkg.go.dev/os#Stat) is better because if the file exists, it will not be closed until it is GC'd.

pkg/vz/vz_driver_darwin.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return nil
+}
+
+func (l *LimaVzDriver) Start(ctx context.Context) error {
+	logrus.Infof("Starting VZ (hint: to watch the boot progress, see %q)", filepath.Join(l.Instance.Dir, filenames.SerialLog))
+	vm, err := startVM(ctx, l.BaseDriver)
+	if err != nil {
+		return err
+	}
```
Code-Hex, line 48: I think error handling would be required to check version compatibility. Given that lima-vm will be built with the latest OS, handling of `vz.ErrBuildTargetOSVersion` is unnecessary. (maybe)
https://github.com/Code-Hex/vz#version-compatibility-check

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,485 @@
...
+			return err
+		}
+		config.SetDirectoryShare(share)
+		mounts[i] = config
+	}
+	vmConfig.SetDirectorySharingDevicesVirtualMachineConfiguration(mounts)
+	return nil
+}
+
+func attachOtherDevices(_ *driver.BaseDriver, vmConfig *vz.VirtualMachineConfiguration) error {
```
shyim, line 366: does it make sense to register there also `NewLinuxRosettaDirectoryShare` 

So we have Rosetta inside VZ? https://github.com/Code-Hex/vz/blob/c436fc367107393c6eda419ff0f69903f674c118/shared_directory_arm64.go#L63

AkihiroSuda: Can be another PR

pkg/vz/vz_driver_darwin.go
Code-Hex: @balajiv113 Right. However, will be returned error might be not helpful for user side. So I meant to return an error with message which is helpful that this driver cannot use below macOS 13.

pkg/store/filenames/filenames.go
balajiv113: Done

pkg/store/filenames/filenames.go
balajiv113: Done

pkg/vz/vm_darwin.go
balajiv113: Done. By default doing a debug log.

pkg/vz/vm_darwin.go
balajiv113: Done

pkg/vz/vz_driver_darwin.go
balajiv113: Done

pkg/vz/vm_darwin.go
AkihiroSuda: I expected it could be supported via VZFileHandleNetworkDeviceAttachment , but seems incompatible with QEMU:
https://developer.apple.com/documentation/virtualization/vzfilehandlenetworkdeviceattachment?language=objc

So we have to error out when `network.Socket` is non-empty

AkihiroSuda:
No need to support legacy BIOS (in this PR, at least).
Could you document that restriction in a documentation like `docs/vmtype.md` ?

Also please consider squashing commits

Code-Hex:
@balajiv113 I believe it's correct behavior that It's started up and it does not appear kernel early messages on the serial attachment because the console is provided via `virtio-pci` (not UART). This is undocumented officially.

> Note that Virtualization.framework provides all IO as virtio-pci, including the console (i.e. not a UART).

https://github.com/evansm7/vftool#kernelsnotes

However, as noted on the wiki, some messages are displayed.

> It's not possible to capture early boot messages with it. It starts working around the time /dev is populated. (by using [mdev](https://wiki.gentoo.org/wiki/Udev), [sdev](https://wiki.gentoo.org/wiki/Mdev), etc.)

https://github.com/Code-Hex/vz/wiki#linux-kernel-requirements

pkg/vz/vm_darwin.go
balajiv113: I think we can support it, maybe some extra support might be needed from socker_vmnet as well. I will try to support this in a different PR.

pkg/limayaml/limayaml.go
balajiv113: Done

pkg/vz/vz_driver_darwin.go
balajiv113: Added errors for not supported one like legacyBios. As vm will not even start with this as true. 
For other things will take it up in a follow-up PR like networks, mount as 9p etc.

balajiv113:
@AkihiroSuda 
Have addressed the requested changes and also added docs for vmtype and mount. Please let me know if anything further needed.

docs/vmtype.md
```diff
@@ -0,0 +1,34 @@
...
+```yaml
+# Example to run ubuntu using vmType: vz instead of qemu (Default)
+vmType: "vz"
+images:
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img"
+  arch: "aarch64"
+mounts:
+  - location: "~"
```
AkihiroSuda, line 24: Can we specify virtiofs here?
We can also consider making it default for vz, but it can be another PR if controversial.

docs/vmtype.md
```diff
@@ -0,0 +1,34 @@
...
+```
+
+### Caveats
+- "vz" option is only supported on macOS 13 or above
+- Virtualization.Framework is a virtualization software, so it doesn't support running "intel guest on arm" and vice versa
```
AkihiroSuda, line 29:
```suggestion
- Virtualization.framework doesn't support running "intel guest on arm" and vice versa
```

docs/vmtype.md
balajiv113: We can will add now.

pkg/vz/vz_driver_darwin.go
```diff
@@ -0,0 +1,70 @@
...
+		BaseDriver: driver,
+	}
+}
+
+func (l *LimaVzDriver) Validate() error {
+	if *l.Yaml.MountType == limayaml.NINEP {
+		return fmt.Errorf("field `mountType` must be %q or %q for VZ driver , got %q", limayaml.REVSSHFS, limayaml.VIRTIOFS, *l.Yaml.MountType)
+	}
+	if *l.Yaml.Firmware.LegacyBIOS {
+		return fmt.Errorf("`firmwate.legacyBIOS` configuration is not support for VZ driver")
```
AkihiroSuda, line 35: typo: firmwate

AkihiroSuda:
Thanks, tested on macOS 13 Intel, and saw a few trivial issues. Can be fixed in follow-up PRs

- The `vz` tempalte didn't start up without mkdiring `/tmp/lima`:
```
{"level":"fatal","msg":"stat /tmp/lima: no such file or directory","time":"2022-11-12T00:19:02+09:00"}
```

- `ERRO[0033] [hostagent] r.CreateEndpoint() = connection was refused ` errors in the limactl log
```
$ limactl start vz
INFO[0000] Using the existing instance "vz"             
INFO[0000] Attempting to download the nerdctl archive from "https://github.com/containerd/nerdctl/releases/download/v1.0.0/nerdctl-full-1.0.0-linux-amd64.tar.gz"  digest="sha256:b7f76a3bf1b8161eb94ebe885945feb2887dfc0d274f9da908a17bc0ef853eb9"
INFO[0000] Using cache "/Users/suda/Library/Caches/lima/download/by-url-sha256/86e8280c3d639367efe7a50660ecfc4eade10b1696a9deeba27fdbf086d11098/data" 
INFO[0001] [hostagent] Starting VZ (hint: to watch the boot progress, see "/Users/suda/.lima/vz/serial.log") 
INFO[0001] SSH Local Port: 57795                        
INFO[0001] [hostagent] new connection from  to          
INFO[0001] [hostagent] Waiting for the essential requirement 1 of 3: "ssh" 
INFO[0001] [hostagent] [VZ] - vm state change: running  
INFO[0004] [hostagent] 2022/11/12 00:20:41 tcpproxy: for incoming conn 127.0.0.1:57797, error dialing "192.168.5.15:22": connect tcp 192.168.5.15:22: no route 
INFO[0015] [hostagent] Waiting for the essential requirement 1 of 3: "ssh" 
INFO[0015] [hostagent] 2022/11/12 00:20:52 tcpproxy: for incoming conn 127.0.0.1:57804, error dialing "192.168.5.15:22": connect tcp 192.168.5.15:22: connection was refused 
INFO[0025] [hostagent] Waiting for the essential requirement 1 of 3: "ssh" 
INFO[0026] [hostagent] The essential requirement 1 of 3 is satisfied 
INFO[0026] [hostagent] Waiting for the essential requirement 2 of 3: "user session is ready for ssh" 
ERRO[0033] [hostagent] r.CreateEndpoint() = connection was refused 
ERRO[0033] [hostagent] r.CreateEndpoint() = connection was refused 
INFO[0041] [hostagent] Waiting for the essential requirement 2 of 3: "user session is ready for ssh" 
INFO[0041] [hostagent] The essential requirement 2 of 3 is satisfied 
INFO[0041] [hostagent] Waiting for the essential requirement 3 of 3: "the guest agent to be running" 
INFO[0041] [hostagent] The essential requirement 3 of 3 is satisfied 
INFO[0041] [hostagent] Waiting for the optional requirement 1 of 2: "systemd must be available" 
INFO[0042] [hostagent] Forwarding "/run/lima-guestagent.sock" (guest) to "/Users/suda/.lima/vz/ga.sock" (host) 
INFO[0042] [hostagent] The optional requirement 1 of 2 is satisfied 
INFO[0042] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed" 
INFO[0042] [hostagent] Not forwarding TCP [::]:22       
INFO[0042] [hostagent] Not forwarding TCP 0.0.0.0:22    
INFO[0042] [hostagent] Not forwarding TCP 127.0.0.53:53 
ERRO[0049] [hostagent] r.CreateEndpoint() = connection was refused 
INFO[0063] [hostagent] The optional requirement 2 of 2 is satisfied 
INFO[0063] [hostagent] Waiting for the final requirement 1 of 1: "boot scripts must have finished" 
INFO[0075] [hostagent] The final requirement 1 of 1 is satisfied 
INFO[0075] READY. Run `limactl shell vz` to open the shell. 
```


- Writing a file on the (read-only) home fails with a cryptic error. I expect this to fail with `EROFS`
```console
$ touch /Users/suda/foo
touch: setting times of '/Users/suda/foo': No such file or directory
```

docs/vmtype.md
AkihiroSuda: Thanks, yes, changing the default can be another PR

balajiv113:
`The vz tempalte didn't start up without mkdiring /tmp/lima`
vz expects all mounting folders to be present, i will for now remove /tmp/lima from the template. In a follow-up we may look to ignore mounts with warning if not present

`ERRO[0033] [hostagent] r.CreateEndpoint() = connection was refused errors in the limactl log`
This is actually from gvisor-tap-vsock (https://github.com/containers/gvisor-tap-vsock/blob/main/pkg/services/forwarder/tcp.go#L44). As far as i checked it is thrown as error but doesn't break network connection.

`Writing a file on the (read-only) home fails with a cryptic error. I expect this to fail with EROFS`
Will look into it. Maybe missing some mount option. Will fix it (If able to) in a separate PR.

AkihiroSuda:
> In a follow-up we may look to ignore mounts with warning if not present

For consistency with QEMU, it should be automatically mkdired

docs/vmtype.md
balajiv113: Done

pkg/vz/vz_driver_darwin.go
balajiv113: Done

mprimeaux: typo: should be “not supported”.

balajiv113: Thanks. Done the change :)

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,463 @@
...
+	return nil
+}
+
+func attachFolderMounts(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfiguration) error {
+	mounts := make([]vz.DirectorySharingDeviceConfiguration, len(driver.Yaml.Mounts))
+	for i, mount := range driver.Yaml.Mounts {
+		expandedPath, err := localpathutil.Expand(mount.Location)
+		if err != nil {
+			return err
+		}
```
AkihiroSuda, line 322: Seems to need mkdir here, LGTM other than that, thanks

AkihiroSuda:
@jandubois 
Could you take a look?
This is being added as an experimental feature, so the implementation can be revised later at any time, but I'd like to make sure that we are fine with the YAML property (Just `vmType: "vz"`)

pkg/vz/vm_darwin.go
balajiv113: Done the changes and updated the vz template to include /tmp/lima

AkihiroSuda (approved):
Thanks

chancez:
To be clear, this doesn't enable a Rosetta directory share yet right? Ie: https://github.com/Code-Hex/vz/blob/9de1c1041d5e560f46f9f94fc499a1cc72fd92e5/example/gui-linux/rosseta_directory_share_arm64.go#LL14C15-L14C15

chancez:
Minor nit, logs mention QEMU when doing a force delete:
```
(⎈|kind-kind:default) ~/p/lima ❯❯❯ ./_output/bin/limactl delete --force vz                                                                                                                                 ✘ 1 driver
INFO[0000] Sending SIGKILL to the QEMU process 62640    
INFO[0000] Sending SIGKILL to the host agent process 62640 
```

balajiv113:
> To be clear, this doesn't enable a Rosetta directory share yet right? Ie: https://github.com/Code-Hex/vz/blob/9de1c1041d5e560f46f9f94fc499a1cc72fd92e5/example/gui-linux/rosseta_directory_share_arm64.go#LL14C15-L14C15

Yes exactly. The current changes in this PR doesn't support rosseta. This will be done in a new PR

pkg/vz/vm_darwin.go
```diff
@@ -0,0 +1,470 @@
...
+					if _, err := os.Stat(pidFile); !errors.Is(err, os.ErrNotExist) {
+						logrus.Errorf("pidfile %q already exists", pidFile)
+						errCh <- err
+					}
+					if err := os.WriteFile(pidFile, []byte(strconv.Itoa(os.Getpid())+"\n"), 0644); err != nil {
+						logrus.Errorf("error writing to pid fil %q", pidFile)
+						errCh <- err
+					}
+					defer os.RemoveAll(pidFile)
+					logrus.Println("[VZ] - vm state change: running")
```
chancez, line 77: Why are these `logrus.Println` vs `logrus.Info` or something else?

chancez:
So I went ahead and implemented support for Rosetta because I was excited. I can open a PR targeting your branch or I can open a PR targeting master and just rebase later, or you could just copy the code/reimplement it yourself.

https://github.com/balajiv113/lima/compare/driver...chancez:lima:pr/chancez/vz_rosetta?expand=1

Testing the functionality works for running amd64 binaries:
```
chancezibolski@lima-vz:/tmp$ uname -a
Linux lima-vz 5.15.0-27-generic #28-Ubuntu SMP Thu Apr 14 12:56:31 UTC 2022 aarch64 aarch64 aarch64 GNU/Linux
chancezibolski@lima-vz:/tmp$ curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
100   138  100   138    0     0   1234      0 --:--:-- --:--:-- --:--:--  1243
100 42.9M  100 42.9M    0     0  57.6M      0 --:--:-- --:--:-- --:--:-- 85.1M
chancezibolski@lima-vz:/tmp$ chmod +x ./kubectl
chancezibolski@lima-vz:/tmp$ file ./kubectl
./kubectl: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, Go BuildID=dY7TJ07x2P2mabT5iSQq/f6EqwRdp0KI77M6ZHcly/i8XNAtvopP0rhCwp9WKx/9a1QsYr3peGkHbpROXlp, stripped
chancezibolski@lima-vz:/tmp$ ./kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"25", GitVersion:"v1.25.4", GitCommit:"872a965c6c6526caa949f0c6ac028ef7aff3fb78", GitTreeState:"clean", BuildDate:"2022-11-09T13:36:36Z", GoVersion:"go1.19.3", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v4.5.7
The connection to the server localhost:8080 was refused - did you specify the right host or port?
chancezibolski@lima-vz:/tmp$ 
```

And comparing build times of an product I work on with Native, Qemu, and Rosetta:

For the Go project I work on, which is ~3 go binaries compiled in a docker image, the `RUN` command takes roughly `21s` natively for arm64, regardless of if the VM is created via qemu or hypervisor.framework (this is expected).

Building the binaries/image for amd64, requiring emulation takes:
* 224s with Qemu for emulation. 10x slower than native.
* 87s with Rosetta emulation. 4x slower than native. 2.5x faster than Qemu.

balajiv113:
@chancez 
Awesome. Thanks for the quick try out.

About the changes,
I would prefer to raise a new PR (feel free to raise one). Reason being wanted to keep it lite with basic required stuffs supporting current features.

Just some points maybe, 
- see if we need to provide this support as configurable (in lima yaml like rosseta: true/something similar 
- if it is via configuration, maybe we can directly install without prompts
- The mount commands also can be managed internally via cidata based on this config

chancez:
@balajiv113 yeah it was just a hack/POC to share with others, so I agree with your points. Using provisioning scripts for mounting is non-persistent so it doesn't work after reboot either.

chancez:
One consideration, if we're going add VM-type specific configuration (options specific to VZ or Qemu), perhaps `vmType` should be made into a object with values.

Eg, renaming `vmType` to `driver.type`:
```
driver:
  type: "qemu|vz"
  vz:
    rosetta:
      enabled: true
  qemu:
    someQemuOnlyOption: true
```

This came up when I was thinking about where to put rosetta configuration into the `LimaYAML`. Thoughts?

AkihiroSuda:
@chancez Thanks, could you open your rosetta PR after merging this base PR?

> One consideration, if we're going add VM-type specific configuration (options specific to VZ or Qemu), perhaps `vmType` should be made into a object with values.

We already have VM-specific configs in the top-level struct, so I guess we are too late to make that change 😞 
https://github.com/lima-vm/lima/blob/06a4b3302f32d893bd9d19d59530c94b05ca2071/examples/default.yaml#L212-L219
https://github.com/lima-vm/lima/blob/06a4b3302f32d893bd9d19d59530c94b05ca2071/examples/default.yaml#L226-L233

chancez:
> @chancez Thanks, could you open your rosetta PR after merging this base PR?
Yeah I'll make a PR with some adjustments to make it closer to "ready" and we can discuss the rest a bit further in that PR.

AkihiroSuda:
Sorry, needs another rebase

balajiv113:
Done the rebase.
Also addressed @chancez  comments about QEMU log during delete instance and also updated printf with infof

pkg/vz/vm_darwin.go
balajiv113: Done the updates

AkihiroSuda (dismissed):

pkg/limayaml/defaults.go
```diff
@@ -630,6 +637,25 @@ func NewArch(arch string) Arch {
...
+		logrus.Warnf("Unknown driver: %s", driver)
+		return driver
+	}
+}
+
+func ResolveVMType(s *string) VMType {
+	if s == nil || *s == "" || *s == "default" {
+		return NewVMType("qemu")
+	}
+	return *s
```
jandubois, line 656: Nitpicking (you only need to call `NewVMType` when you potentially have an unknown driver name):

```suggestion
	if s == nil || *s == "" || *s == "default" {
		return QEMU
	}
	return NewVMType(*s)
```

pkg/qemu/imgutil/imgutil.go
```diff
@@ -15,6 +15,18 @@ type Info struct {
 	VSize  int64  `json:"virtual-size,omitempty"`
 }
 
+func QcowToRaw(source string, dest string) error {
```
jandubois, line 18: Given that QCOW is an acronym (QEMU Copy On Write), shouldn't this be `QCOWToRaw`?

cmd/limactl/stop.go
```diff
@@ -105,12 +105,12 @@ func waitForHostAgentTermination(ctx context.Context, inst *store.Instance, begi
 
 func stopInstanceForcibly(inst *store.Instance) {
 	if inst.QemuPID > 0 {
-		logrus.Infof("Sending SIGKILL to the QEMU process %d", inst.QemuPID)
+		logrus.Infof("Sending SIGKILL to the %s driver process %d", inst.VMType, inst.QemuPID)
```
jandubois, line 108: Should `QemuPID` be renamed to `DriverPID`?

Also `filenames.QemuPID` should then be renamed as well, and use a different filename. It is confusing to have a `qemu.pid` file, but no qemu process.

AkihiroSuda: QEMU driver should continue using `qemu.pid`, somebody might be potentially depending on it

docs/mount.md
```diff
@@ -79,3 +79,20 @@ The "9p" mount type requires Lima v0.10.0 or later.
...
+vmType: "vz"
+mountType: "virtiofs"
+mounts:
+- location: "~"
+```
+
+The "vz" mount type requires Lima v0.14.0 or later.
+
+#### Caveats
+- The "virtiofs" mount type is supported only on mocOS 13 or above with `vmType: vz` config
```
jandubois, line 98:
```suggestion
- The "virtiofs" mount type is supported only on macOS 13 or above with `vmType: vz` config.
```

cmd/limactl/stop.go
balajiv113: Exactly, i wanted to change this to driver.pid. But later realised that this might have impact on other current running instances as well (Stop may not work well, unless we support both files). So thought of doing it in a follow-up.

There are 2 ways i could see to solve this,
- Maybe introduce a way to get driver specific pid file (qemu.pid/vz.pid). The hostagent can use this driver specific method in places like stop, identify status. 
- deprecate qemu.pid and for qemu driver start creating both driver.pid and qemu.pid. We can clean qemu.pid once we decide when to remove (Maybe 1.0)

Am happy to do the required changes as part of this PR as well.

docs/mount.md
balajiv113: My bad !! Will add these changes

cmd/limactl/stop.go
AkihiroSuda: > driver specific pid file (qemu.pid/vz.pid)

SGTM

jandubois: Good point! Followup PR is fine; this PR is big enough already! 😄

> Maybe introduce a way to get driver specific pid file (qemu.pid/vz.pid). The hostagent can use this driver specific method in places like stop, identify status.

I think I would prefer this way to "solve" the issue. Maybe have a method in the driver package that provides the pid filename?

balajiv113: Yes exactly, driver can give out their own pid name. This will be used through out hostagent code-base.

I will take this up in a follow-up PR 👍

pkg/limayaml/defaults.go
balajiv113: Done

pkg/qemu/imgutil/imgutil.go
balajiv113: Done

docs/mount.md
balajiv113: Done

AkihiroSuda (approved):

docs/vmtype.md
```diff
@@ -0,0 +1,35 @@
...
+mountType: "virtiofs"
+```
+
+### Caveats
+- "vz" option is only supported on macOS 13 or above
+- Virtualization.framework doesn't support running "intel guest on arm" and vice versa
+
+### Known Issues
+- "vz" doesn't support `legacyBoot: true` option, so guest machine like centos-stream, archlinux, oraclelinux will not work
+- Host to guest networking (`networks` section in lima yaml) is not supported
```
abiosoft, line 34: Is this outrightly not supported by the driver, and with no workarounds?

AkihiroSuda: Expected to be added as "vznat" network in a separate PR

jandubois:
I've been testing this PR today on an M1 machine running Ventura, and I saw this error when starting `template://experimental/vz`:

```
ERRO[0296] [hostagent] r.CreateEndpoint() = connection was refused
```

Which seems to originate from https://github.com/containers/gvisor-tap-vsock/blob/main/pkg/services/forwarder/tcp.go#L45.

Is this expected? There was no additional information in `ha.stderr.log`.

jandubois:
> This is being added as an experimental feature, so the implementation can be revised later at any time, but I'd like to make sure that we are fine with the YAML property (Just vmType: "vz")

I'm fine with the `vmType` property, and fine with merging it now, and refining it later.

I've not had time to review the code in detail, but so far it all looks good, and most importantly: works. I've only tested on M1 so far, but should be able to get remote access to an Intel Ventura machine for testing as well.

PR #1155, opened 2022-11-15
Support Rosetta with VZ driver

chancez:
Useful reading: https://developer.apple.com/documentation/virtualization/running_intel_binaries_in_linux_vms_with_rosetta

Implementation based off the sample here: https://github.com/Code-Hex/vz/tree/9de1c1041d5e560f46f9f94fc499a1cc72fd92e5/example/gui-linux


examples/experimental/vz.yaml
```diff
@@ -11,3 +16,13 @@ mounts:
...
+
+provision:
+- mode: system
+  # configure binfmt support for Rosetta
+  script: |
+    apt install binfmt-support
+    /usr/sbin/update-binfmts --install rosetta /mnt/lima-rosetta/rosetta \
+    --magic "\x7fELF\x02\x01\x01\x00\x00\x00\x00\x00\x00\x00\x00\x00\x02\x00\x3e\x00" \
+    --mask "\xff\xff\xff\xff\xff\xfe\xfe\x00\xff\xff\xff\xff\xff\xff\xff\xff\xfe\xff\xff\xff" \
+    --credentials yes --preserve no --fix-binary yes
```
AkihiroSuda, line 28: - No need to install binfmt-support if you write to `/proc/sys/fs/binfmt_misc` by yourself
- I'd expect this to be moved to a YAML property like `rosetta.binfmt: true`

pkg/cidata/cidata.TEMPLATE.d/boot/05-rosetta-volume.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
```
AkihiroSuda, line 2: Every shell script must have `set -eu`

pkg/cidata/cidata.TEMPLATE.d/boot/05-rosetta-volume.sh
```diff
@@ -0,0 +1,9 @@
...
+if [ ! -d /tmp/rosetta-share ]; then
```
AkihiroSuda, line 3: What's /tmp/rosetta-share ?

jandubois: It is the the "Rosetta Directory Share" (see https://developer.apple.com/documentation/virtualization/running_intel_binaries_in_linux_vms_with_rosetta).

The sample code mounts it under `/tmp`, but I would prefer to mount it as `/mnt/rosetta` instead.

cedws:
I don't have any comments on the changes other than nice work and LGTM, but do you have any suggested reading material on how this Rosetta acceleration works?

Thanks.

pkg/cidata/cidata.TEMPLATE.d/boot/05-rosetta-volume.sh
chancez: Oops, I just forgot to update the path in the conditional to match the path where I mounted it.

pkg/cidata/cidata.TEMPLATE.d/boot/05-rosetta-volume.sh
```diff
@@ -0,0 +1,9 @@
...
+  mkdir -p /mnt/lima-rosetta
+  # Cannot find a way to check that the virtiofs "device" is present, so just
+  # Try mounting it and hope for the best. If it fails, then it probably isn't
+  # a mac, or the virtio volume isn't being provided to the guest.
+  mount -t virtiofs vz-rosetta /mnt/lima-rosetta || true
```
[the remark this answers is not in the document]
chancez: Good thought, I'll see how difficult that is. I'm not sure if mounts works for anything besides bind-mounts, but if it does, then that would work.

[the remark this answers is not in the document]
balajiv113: Ah !! It won't work i think.
I missed a case, it will not work when we have mix of mountType. Using mountType: reversessh-fs and wanting to use rosetta.
This will not work as all the y.mounts are done based on mountType.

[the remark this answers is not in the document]
chancez: We could add another mountType for this potentially. Not sure if it would be useful outside of this use-case yet though.

examples/experimental/vz.yaml
chancez: This was promising, but I'm hitting permissions issues as root. I suspect it's something to do with apparmor, and that the binfmt-support package probably adds a profile to allow this. Any ideas how I could work around this?

chancez: Ah nvm I see, I have to use `/proc/sys/fs/binfmt_misc/register`

chancez:
Okay, PR is updated based on feedback and I like it much better than the initial MVP. Let me know what you all think. I also rebased and moved it from draft->ready to review.

pkg/vz/rosetta_directory_share.go
```diff
@@ -0,0 +1,14 @@
...
+package vz
+
+import (
+	"fmt"
+
+	"github.com/Code-Hex/vz/v3"
+)
+
+func createRosettaDirectoryShareConfiguration(install bool) (*vz.VirtioFileSystemDeviceConfiguration, error) {
+	return nil, fmt.Errorf("Rosetta is unsupported on non-ARM64 hosts")
```
balajiv113, line 13: We should do a warning instead of error.

This is to make vz template runnable on both arm & x64

pkg/vz/vm_darwin.go
```diff
@@ -344,6 +344,16 @@ func attachFolderMounts(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineCo
 		config.SetDirectoryShare(share)
 		mounts[i] = config
 	}
+
+	if driver.Yaml.Rosetta.Enabled {
+		logrus.Info("Setting up Rosetta share")
+		directorySharingDeviceConfig, err := createRosettaDirectoryShareConfiguration(driver.Yaml.Rosetta.Install)
```
balajiv113, line 350: Just a thought,

Do we needed Rosetta.Install ? 
Can we directly install since Rosetta.Enabled is true already ?

chancez:
@cedws 
> I don't have any comments on the changes other than nice work and LGTM, but do you have any suggested reading material on how this Rosetta acceleration works?
> 
> Thanks.

For how it works with Linux: https://developer.apple.com/documentation/virtualization/running_intel_binaries_in_linux_vms_with_rosetta

Rosetta in general: https://developer.apple.com/documentation/apple-silicon/about-the-rosetta-translation-environment

pkg/cidata/cidata.TEMPLATE.d/boot/05-rosetta-volume.sh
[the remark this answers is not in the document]
chancez: I ended up continuing to use cidata, but we can revisit the approach later if needed.

pkg/vz/rosetta_directory_share.go
chancez: Couldn't users just disable Rosetta on x86 in that case?

pkg/vz/vm_darwin.go
chancez: I'm not sure. I figured it would make sense to make it configurable in case users want to install Rosetta themselves. I already have it installed, so I'm not sure what it actually looks like if you try without it.

pkg/vz/rosetta_directory_share.go
chancez: I also think in that case, we probably can still return an error here, but we can turn it into a named error and check using `errors.Is` in the caller.

chancez:
There was also some discussion about whether or not this should be enabled by default in the other PR. What do people think?

pkg/vz/rosetta_directory_share.go
chancez: Updated

pkg/limayaml/limayaml.go
```diff
@@ -99,6 +107,10 @@ type NineP struct {
 	Cache           *string `yaml:"cache,omitempty" json:"cache,omitempty"`
 }
 
+type VirtioFS struct {
+	Tag *string `yaml:"tag,omitempty" json:"tag,omitempty"`
```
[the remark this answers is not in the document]
chancez: Fixed

AkihiroSuda:
> There was also some discussion about whether or not this should be enabled by default in the other PR. What do people think?

The mount can be enabled by default, but binfmt should be probably opt-in for now.
The `vz.yaml` template can have `rosetta.binfmt: true` by default.

chancez:
> > There was also some discussion about whether or not this should be enabled by default in the other PR. What do people think?
> 
> The mount can be enabled by default, but binfmt should be probably opt-in for now. The `vz.yaml` template can have `rosetta.binfmt: true` by default.

Is that best done via `FillDefault` or should we change the field to `Disable` so the zero value is that Rosetta is enabled? (Disable: false)

pkg/vz/vm_darwin.go
balajiv113: During installation nothing will pop-up or happen. It just like installation via command line. So i think it should be fine. 

But let's also hear others opinion on this.

AkihiroSuda:
> > > There was also some discussion about whether or not this should be enabled by default in the other PR. What do people think?
> > 
> > 
> > The mount can be enabled by default, but binfmt should be probably opt-in for now. The `vz.yaml` template can have `rosetta.binfmt: true` by default.
> 
> Is that best done via `FillDefault` or should we change the field to `Disable` so the zero value is that Rosetta is enabled? (Disable: false)

Lima itself should leave it disabled by default. The `vz.yaml` template file may have `rosetta.binfmt: true` line to override the built-in default value.

chancez:
@AkihiroSuda Sorry I meant for `rosetta.enabled`. 

Also what are your thoughts about the Rosetta installation behavior?

AkihiroSuda:
Looks like installing rosetta requires the root of the macOS host, so it should be disabled by default.

https://developer.apple.com/documentation/virtualization/running_intel_binaries_in_linux_vms_with_rosetta?language=objc
> Installing Rosetta is a one-time process per computer that requires the user to grant permission for the system to install Rosetta. If Rosetta is already installed in macOS, the system activates it for use under the Virtualization framework; if Rosetta isn’t installed, the framework downloads the installer from the network and performs the installation. The installer interactively prompts the user for authorization, and your app should handle any of the possible error conditions that could occur during the authorization, download, and installation process.

chancez:
@AkihiroSuda then it sounds like the PR as it is should be correct with the current defaults and `vz.yaml`. I've rebased.

chancez:
Getting a lint error on my `errRosettaUnsupported ` variable from golangci-lint, presumably because it's running without the build tags. I can add the needed build tags, but this probably means `golangci-lint` is not running on the new VZ code. We should update the GitHub action to specify the necessary build tags so it's getting properly linted.

pkg/vz/vm_darwin.go
AkihiroSuda: > Can we directly install since Rosetta.Enabled is true already ?

:+1:

AkihiroSuda: > Can we directly install since Rosetta.Enabled is true already ?

:+1: , assuming that `Enabled = true` doesn't show the macOS password popup as long as Rosetta is already installed on the macOS host.

balajiv113: @AkihiroSuda Your assumption is true. Only for installation the password pop-up will be shown. Not for every run of rosetta.

chancez: PR updated

AkihiroSuda (approved):
Thanks

pkg/vz/rosetta_directory_share.go
```diff
@@ -0,0 +1,12 @@
+//go:build darwin && !arm64
+// +build darwin,!arm64
```
balajiv113, line 2: I believe we need to use !no_vz buildtag in both rosetta files to avoid this issue 
https://github.com/lima-vm/lima/issues/1176

pkg/vz/errors.go
```diff
@@ -0,0 +1,8 @@
+//go:build darwin
+// +build darwin
```
balajiv113, line 2: Do we need this tags ?? 
errors.go can be generic irrespective of platform (We can move this https://github.com/lima-vm/lima/blob/master/pkg/vz/vz_driver_others.go#L13 as well here)

If we need this tag, then naming the file as errors_darwin.go should be sufficient and consistent across i believe

pkg/vz/vm_darwin.go
```diff
@@ -345,6 +345,19 @@ func attachFolderMounts(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineCo
...
 		mounts[i] = config
 	}
+
+	if driver.Yaml.Rosetta.Enabled {
+		logrus.Info("Setting up Rosetta share")
+		directorySharingDeviceConfig, err := createRosettaDirectoryShareConfiguration()
+		if errors.Is(err, errRosettaUnsupported) {
+			logrus.Warnf("Unable to configure Rosetta: %s", err)
+		} else if err != nil {
+			return err
```
AkihiroSuda, line 355: Probably, any error from `createRosettaDirectoryShareConfiguration` can be a warning.
e.g., failure of installation of Rosetta

pkg/vz/errors.go
chancez: Without build tags the variable is detected as unused in non-darwin builds.

pkg/vz/vm_darwin.go
chancez: I'm still unsure of this. Is there other example cases where Lima fails to do setup and continues on anyways? It seems like continuing on would lead to unexpected results for users who may not notice the warnings, etc.

AkihiroSuda: > Is there other example cases where Lima fails to do setup and continues on anyways? 

Yes, in a bunch of places (`git grep -F .Warn`)

e.g.,

https://github.com/lima-vm/lima/blob/32baa6216f89ce4aaacbc254d13f31535bffeb2d/pkg/start/start.go#L213

chancez: Gotcha. I've updated it to warn instead.

pkg/vz/errors.go
chancez: Renamed the file, and I updated build tags as more were required due to your previous comment: https://github.com/lima-vm/lima/pull/1155#discussion_r1029696381

pkg/vz/rosetta_directory_share.go
chancez: Done

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Released beta.1 with Rosetta
https://github.com/lima-vm/lima/releases/tag/v0.14.0-beta.1

PR #1168, opened 2022-11-16
templates: alpine: remove legacyBIOS=true

AkihiroSuda:


jandubois (approved):
LGTM, and legacy BIOS doesn't seem to be needed.

Do you remember why it was originally added?

AkihiroSuda:
> Do you remember why it was originally added?

IIRC the alpine image at that time didn't support UEFI boot ?

PR #1170, opened 2022-11-16
fix alpine data-volume issue

balajiv113:
Fixes the issue mentioned here
https://github.com/lima-vm/alpine-lima/issues/94

The reason is, 04-persistent-data-volume.sh bind the first available disk as data volume. Since cidata was first available and it was already mounted. the script was failing. Due to disk name mismatch (lsblk say vda for cidata, but /proc/mounts give disk/by-label/lima-cidata as name)

Keeping cidata as last disk similar to qemu fixes this issue.


abiosoft (approved):
I can confirm that this fixes the issue.

AkihiroSuda (approved):
Thanks

PR #1171, opened 2022-11-16
support for driver specific pid files

balajiv113:
This PR resolves this discussion
https://github.com/lima-vm/lima/pull/1147#discussion_r1022460455

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -622,7 +622,7 @@ func Cmdline(cfg Config) (string, []string, error) {
 
 	// QEMU process
 	args = append(args, "-name", "lima-"+cfg.Name)
-	args = append(args, "-pidfile", filepath.Join(cfg.InstanceDir, filenames.QemuPID))
+	args = append(args, "-pidfile", filepath.Join(cfg.InstanceDir, filenames.PIDFile(*y.VMType)))
 
 	return exe, args, nil
 }
--- a/cmd/limactl/stop.go
+++ b/cmd/limactl/stop.go
@@ -104,9 +104,9 @@ func waitForHostAgentTermination(ctx context.Context, inst *store.Instance, begi
 }
 
 func stopInstanceForcibly(inst *store.Instance) {
-	if inst.QemuPID > 0 {
-		logrus.Infof("Sending SIGKILL to the %s driver process %d", inst.VMType, inst.QemuPID)
-		if err := osutil.SysKill(inst.QemuPID, osutil.SigKill); err != nil {
+	if inst.DriverPID > 0 {
+		logrus.Infof("Sending SIGKILL to the %s driver process %d", inst.VMType, inst.DriverPID)
+		if err := osutil.SysKill(inst.DriverPID, osutil.SigKill); err != nil {
 			logrus.Error(err)
 		}
 	} else {
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -48,6 +48,7 @@ QEMU:
 - `serial.sock`: QEMU serial socket, for debugging (Usage: `socat -,echo=0,icanon=0 unix-connect:serial.sock`)
 
 VZ:
+- `vz.pid`: VZ PID
 - `vz-identifier`: Unique machine identifier file for a VM
 - `vz-efi`: EFIVariable store file for a VM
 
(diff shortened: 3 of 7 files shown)


balajiv113:
Info,
Have add DriverPIDName to info() API of hostagent.

This made a lot of sense as hostagent is the one starting and managing the driver. And store is not required to know about it. Even though it can.

pkg/driver/driver.go
```diff
@@ -12,6 +12,8 @@ type Driver interface {
 
 	CreateDisk() error
 
+	GetPIDName() string
```
AkihiroSuda, line 15:
```suggestion
	Name() string
```

The pid file name can be just `Name() + ".pid"`

balajiv113: Ah nice !! This looks more clean.

Done the changes

pkg/qemu/qemu.go
```diff
@@ -31,6 +31,7 @@ type Config struct {
 	Name         string
 	InstanceDir  string
 	LimaYAML     *limayaml.LimaYAML
+	Driver       *LimaQemuDriver
```
jandubois, line 34: With this change the whole `Config` struct becomes redundant because all other fields are just copies from the `Driver` struct.

Would it make sense to get rid of `Config` and just pass around `Driver` instead?

pkg/hostagent/api/api.go
```diff
@@ -1,5 +1,6 @@
 package api
 
 type Info struct {
-	SSHLocalPort int `json:"sshLocalPort,omitempty"`
+	SSHLocalPort int    `json:"sshLocalPort,omitempty"`
+	DriverName   string `json:"driverName,omitempty"`
```
AkihiroSuda, line 5: - Can we eliminate this from the `Info`?  Just like CPU nums, arch, memory, etc.
- If we really can't remove this, this should be named like `vmType` or `vmDriverName`, as we also have a bunch of other "drivers" such as mount driver, network driver, etc.
- We can also consider marshaling the entire YAML as a JSON here if we are going to copy other YAML properties too here

pkg/qemu/qemu.go
jandubois: I have not properly thought this through; it just occurred to me that there was no additional information in `Config` beyond what is already in `Driver`. In general I would try to avoid keeping the same data in multiple places because then it can get out of sync, and hard to debug...

Maybe wait for feedback from @AkihiroSuda before making this change?

AkihiroSuda: If this `Driver` addition is just for getting the driver name, can we just get the driver name from the yaml?

jandubois: We could, but what is the reason for having a separate `Config` struct? The base `Driver` is just pointers to instance data and YAML:

```go
type BaseDriver struct {
	Instance *store.Instance
	Yaml     *limayaml.LimaYAML
}
```

pkg/hostagent/api/api.go
balajiv113: - As drivers are already relying on store, the moment we rely on driver it will have this cyclic import problem. We can look into resolving it as well.
- If am not able to resolve the above issue i will look into moving other properties to rely on this info endpoint itself by YAML to JSON conversion in info endopoint, so that store will simply depend on hostagent info and not assume anything directly from configuration

BalajiVijayakumar1195: @AkihiroSuda 
One way i could think of solving it is, our pid names can always be equal to vmType. This way, in instance.Inspect() we can simply rely on vmType to get according pid.

But we could also argue that with this assumption we don't even need Name() in driver interface :)

Let me know if this is fine. In that case, i will maybe get rid of Name itself fully from driver and go with vmType itself.

balajiv113: @AkihiroSuda 
One way i could think of solving it is, our pid names can always be equal to vmType. This way, in instance.Inspect() we can simply rely on vmType to get according pid.

But we could also argue that with this assumption we don't even need Name() in driver interface :)

Let me know if this is fine. In that case, i will maybe get rid of Name itself fully from driver and go with vmType itself.

AkihiroSuda: > go with vmType itself

:+1:

balajiv113: Done the changes

AkihiroSuda (approved):
Thanks

PR #1173, opened 2022-11-17
fix host network resolver issue with alpine

### LINKED ISSUE #1172: vz: host resolver not working
### Description

I noticed that the host resolver is not working with the vz driver https://github.com/lima-vm/lima/pull/1147, and it is not documented among the known issues.

balajiv113:
fixes #1172

### DIFF
--- a/pkg/networks/gvisor.go
+++ b/pkg/networks/gvisor.go
@@ -39,11 +39,17 @@ func StartGVisorNetstack(ctx context.Context, gVisorOpts *GVisorNetstackOpts) {
 		protocol = types.QemuProtocol
 	}
 
+	// The way gvisor-tap-vsock implemented slirp is different from tradition SLIRP,
+	// - GatewayIP handling all request, also answers DNS queries
+	// - based on NAT configuration, gateway forwards and translates calls to host
+	// Comparing this with QEMU SLIRP,
+	// - DNS is equivalent to GatewayIP
+	// - GatewayIP is equivalent to NAT configuration
 	config := types.Configuration{
 		Debug:             false,
 		MTU:               opts.MTU,
 		Subnet:            SlirpNetwork,
-		GatewayIP:         SlirpGateway,
+		GatewayIP:         SlirpDNS,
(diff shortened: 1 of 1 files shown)


pkg/networks/gvisor.go
```diff
@@ -43,7 +43,7 @@ func StartGVisorNetstack(ctx context.Context, gVisorOpts *GVisorNetstackOpts) {
 		Debug:             false,
 		MTU:               opts.MTU,
 		Subnet:            SlirpNetwork,
-		GatewayIP:         SlirpGateway,
+		GatewayIP:         SlirpDNS,
```
AkihiroSuda, line 46: Could you add a code comment to explain this

balajiv113: Done. Hope this is clear now.

abiosoft (dismissed):
This works 🎉 

Thanks once again @balajiv113

balajiv113:
Modified comments with some more info :) To keep it informative

abiosoft (dismissed):

pkg/networks/gvisor.go
abiosoft: The convention in Go for code comments is to have a space between the double slash `//` and the text.

jandubois: I have to admit that even with the comment, I still don't understand **why** this works. Why is it called a gateway if it isn't actually a gateway?

abiosoft: I think the constants specific to vz driver should be re-defined to avoid confusion. Type aliases can be used.

AkihiroSuda: Yes, thanks, LGTM when the above is added as code comment lines

BalajiVijayakumar1195: Done

balajiv113: Done

AkihiroSuda (approved):
Thanks

PR #1179, opened 2022-11-18
Increase inotify limits by default

### LINKED ISSUE #1178: "too many open files" error upon creating multiple Kind clusters on Lima VMs.
### Lima Version

```
$: lima --version
limactl version 0.12.0
```

### Host Vesrion

```
$: lima cat /etc/os-release
PRETTY_NAME="Ubuntu 22.04.1 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.1 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```

### Problem Description

Starting multiple Kind clusters on Lima VMs created from the `docker-rootful` template causes control plane nodes to fail with EMFILE errors.

Increasing inotify limits per [Kind's documentation](https://kind.sigs.k8s.io/docs/user/known-issues/#pod-errors-due-to-too-many-open-files) resolves this.

The PR linked to this issue adds this change to the `provision` block for this template so that others don't run into this issue.

carlosonunez-vmw:
This resolves #1178.


AkihiroSuda:
https://github.com/apps/dco

examples/docker-rootful.yaml
```diff
@@ -54,6 +54,14 @@ provision:
     fi
     export DEBIAN_FRONTEND=noninteractive
     curl -fsSL https://get.docker.com | sh
+- mode: system
+  script: |
+    #!/bin/bash
+    # Increase inotify limits to prevent nested Kubernetes control planes
```
AkihiroSuda, line 60: Every bash script should have `set -eu -o pipefail`

examples/docker-rootful.yaml
```diff
@@ -54,6 +54,14 @@ provision:
...
+    # from crash looping.
+    echo 'fs.inotify.max_user_watches = 524288' >> /etc/sysctl.conf
+    echo 'fs.inotify.max_user_instances = 512' >> /etc/sysctl.conf
+    sysctl --system
```
AkihiroSuda, line 64: Can we replicate this to docker.yaml, podman*.yaml, k8s.yaml, k3s.yaml too?

carlosonunez-vmw:
sure; i'll try to make these changes some time between tomorrow and friday-- CarlosOn Nov 19, 2022, at 19:04, Akihiro Suda ***@***.***> wrote:﻿
@AkihiroSuda commented on this pull request.



In examples/docker-rootful.yaml:
> @@ -54,6 +54,14 @@ provision:
     fi
+- mode: system
+  script: |
+    #!/bin/bash
+    # Increase inotify limits to prevent nested Kubernetes control planes
+    # from crash looping.
+    echo 'fs.inotify.max_user_watches = 524288' >> /etc/sysctl.conf
+    echo 'fs.inotify.max_user_instances = 512' >> /etc/sysctl.conf
+    sysctl --system

Can we replicate this to docker.yaml, podman*.yaml, k8s.yaml, k3s.yaml too?

afbjorklund:
Seems needlessly complicated for the k3s and k8s examples, since they would have VMs as nodes (not containers) ?

AkihiroSuda:
- Please sign off the commit for DCO: https://github.com/apps/dco
- Please squash commits
- Please consider doing the same for `podman*.yaml`

carlosonunez:
✅  Please sign off the commit for DCO: https://github.com/apps/dco
✅  Please squash commits
⚠️ Please consider doing the same for podman*.yaml

I'm not sure if Podman needs this treatment, as it uses `crun` instead of `runc` which handles nested cgroup mounting differently. This would require additional testing.

Can that be a separate pull request, given that this behavior is known for containerd-based engines?

carlosonunez-vmw:
✅  Please sign off the commit for DCO: https://github.com/apps/dco
✅  Please squash commits
⚠️ Please consider doing the same for podman*.yaml

I'm not sure if Podman needs this treatment, as it uses `crun` instead of `runc` which handles nested cgroup mounting differently. This would require additional testing.

Can that be a separate pull request, given that this behavior is known for containerd-based engines?

examples/k3s.yaml
```diff
@@ -41,7 +41,15 @@ provision:
   script: |
     #!/bin/sh
     curl -sfL https://get.k3s.io | sh -
-
+- mode: system
+  script: |
+    #!/bin/bash
+    # Increase inotify limits to prevent nested Kubernetes control planes
+    # from crash looping.
```
AkihiroSuda, line 48: Is this needed for k3s? If so, it should be needed for `k8s.yaml` too?

afbjorklund: As far as I know, it is only needed for k3d and kind - not for k3s and k8s

BenTheElder: Not necessarily, it's used anytime you're using a lot of inotify, which can happen with k3s as well, anything using configmaps will need one per configmap, user workloads of other kinds may also run into this.

BenTheElder: kind usage is a common way to encounter it, because you often start multiple kubelets on the same kernel, but that's only one way to run up usage. a single kubelet with many configmaps could hit the same limit.

BenTheElder:
> I'm not sure if Podman needs this treatment, as it uses crun instead of runc which handles nested cgroup mounting differently. This would require additional testing.

inotify isn't namespaced in the kernel, if you start another VM / kernel you'll have separate limits but otherwise this applies to anything using inotify (consider also things like the inotify command line tool, IDEs, etc)

crun/runc/... shouldn't change that. Increasing the inotify limits is probably a good idea on all the templates, at the cost of increasing the template complexity and some kmem

carlosonunez:
Hello! I apologize for not signing the DCO with my `vmware` account. I've since moved on and no longer have access to this GitHub handle.

@AkihiroSuda , should I re-raise this PR with my personal GitHub account (i.e. this one) and sign-off with that? I still increase my `inotify` limits in my Docker template. (I have not tested whether podman + kind needs this treatment, though I likely will soon.)

jandubois:
> should I re-raise this PR with my personal GitHub account (i.e. this one) and sign-off with that?

Yes, please! We are unable to merge any commits without a valid DCO; it is a CNCF requirement.

jandubois:
> Increasing the inotify limits is probably a good idea on all the templates, at the cost of increasing the template complexity and some kmem

We could do this in the internal provisioning scripts, if we can agree that it should be done for all distros. That way it would not complicate the templates. We would also have to decide what the new limits are supposed to be, and if they can be the same for each distro.

Is there any downside to increasing the limits? If not, why are the default limits so low?

@AkihiroSuda any opinion on this?

An alternative in the future (when we have composable templates) would be to implement this in a mix-in template, and users could add it with `base: template://options/inotify` or whatever we want to call it. It could even be parameterized with `param` settings. But we are not there yet until later this year.

BenTheElder:
> Is there any downside to increasing the limits? If not, why are the default limits so low?

Setting low limits caps the memory used by the kernel for this purpose (tracking inotify watches)

https://man7.org/linux/man-pages/man7/inotify.7.html#:~:text=/proc%20interfaces%0A%20%20%20%20%20%20%20The%20following%20interfaces%20can%20be%20used%20to%20limit%20the%20amount%20of%20kernel%0A%20%20%20%20%20%20%20memory%20consumed%20by%20inotify%3A

So there's definitely a downside to high limits. Since it's kernel memory it can't be swapped.

I don't remember where 524288 but that's probably .... excessive

BenTheElder:
AIUI that memory is only consumed if the user actually creates that many watches though.

The kernel default is pretty low


https://www.monodevelop.com/documentation/inotify-watches-limit/#:~:text=Managed%20file%20watching%20is%20less,The%20default%20is%208192.

https://fleet-support.jetbrains.com/hc/en-us/articles/8084899752722-Inotify-Watches-Limit-Linux

https://www.suse.com/support/kb/doc/?id=000020048

524288 max_user_watches seems pretty common, but I think a lower value would be fine.

There's some discussion in https://patchwork.kernel.org/project/linux-fsdevel/patch/20201026204418.23197-1-longman@redhat.com/#23713335

carlosonunez:
i agree that upping inotify limits only makes sense when many containerized kubernetes clusters will run on the VM (IIRC I ran into this limit when i was doing cluster-api work using the capd provider which creates lots of clusters).given this, another thing I could do is add a bool field in the Config struct, like "enableHighFSWatcherCount", that adds the appropriate sysctl commands automatically to reduce complexity within template YAMLs.thoughts?On Mar 1, 2025, at 00:25, Benjamin Elder ***@***.***> wrote:﻿
  AIUI that memory is only consumed if the user actually creates that many watches though.
The kernel default is pretty low
https://www.monodevelop.com/documentation/inotify-watches-limit/#:~:text=Managed%20file%20watching%20is%20less,The%20default%20is%208192.
https://fleet-support.jetbrains.com/hc/en-us/articles/8084899752722-Inotify-Watches-Limit-Linux
https://www.suse.com/support/kb/doc/?id=000020048
524288 max_user_watches seems pretty common, but I think a lower value would be fine.
There's some discussion in ***@***.***/#23713335—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you commented.Message ID: ***@***.***>



BenTheElder left a comment (lima-vm/lima#1179)
AIUI that memory is only consumed if the user actually creates that many watches though.
The kernel default is pretty low
https://www.monodevelop.com/documentation/inotify-watches-limit/#:~:text=Managed%20file%20watching%20is%20less,The%20default%20is%208192.
https://fleet-support.jetbrains.com/hc/en-us/articles/8084899752722-Inotify-Watches-Limit-Linux
https://www.suse.com/support/kb/doc/?id=000020048
524288 max_user_watches seems pretty common, but I think a lower value would be fine.
There's some discussion in ***@***.***/#23713335

BenTheElder:
A workload like a remote IDE or one cluster with as many lighter workloads as the multiple clusters together can also do it. We tend to notice it with multiple clusters because there are a minimum number of system workloads, but just adding workloads to one cluster can hit this instead, not to mention workloads that themselves use inotify.

Bumping inotify limits is common for other developer tools that use it, see the links above.

It also shouldn't cost anything if unused. It just caps the number of entries (~Inodes) that are permitted, which consume kmem.

one thing to note: these limits are per user, so they only sort of work as a defensive limit, if you run workloads as lots of users that will multiply the maximum anyhow ...

https://watchexec.github.io/docs/inotify-limits.html

AkihiroSuda:
> Increasing the inotify limits is probably a good idea on all the templates, at the cost of increasing the template complexity and some kmem

Yes, this should be probably added to <https://github.com/lima-vm/lima/tree/master/pkg/cidata/cidata.TEMPLATE.d/boot>.

examples/k3s.yaml
BenTheElder: BTW on Ubuntu defaults kubernetes's e2e tests created enough pods to exceed it running a Kubernetes worker node on the host (not kind, and not a single node cluster), in particular max_user_instances default seems to be pretty low (128)

https://github.com/kubernetes/kubernetes/pull/130990

I setup my fork the other day and have been meaning to work up a new PR, that didn't happen yet, but leaving this breadcrumb in the meantime. There's also some pointers in the linked issue with example tuning in other cluster tools in the project.

Issue #682, opened 2022-02-28
Allow setting up "socket" network for VM-to-VM communication


afbjorklund:
### Description

Currently lima sets up a "user" network, so that each VM gets host and internet access.

It would be possible to set up a second "socket" network, so that they could reach eachother.

https://wiki.qemu.org/Documentation/Networking

I'm not sure what would be the best configuration, but made a small hack PoC to set it up.

Basically I added one boolean to say if it is the listen or the connect, and also the port number.

```diff
diff --git a/pkg/limayaml/limayaml.go b/pkg/limayaml/limayaml.go
index 47ac646..d91f002 100644
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -22,6 +22,8 @@ type LimaYAML struct {
        Probes            []Probe           `yaml:"probes,omitempty" json:"probes,omitempty"`
        PortForwards      []PortForward     `yaml:"portForwards,omitempty" json:"portForwards,omitempty"`
        Message           string            `yaml:"message,omitempty" json:"message,omitempty"`
+       SocketPort        *int              `yaml:"socketPort,omitempty" json:"socketPort,omitempty"`
+       SocketMain        *bool             `yaml:"socketMain,omitempty" json:"socketMain,omitempty"`
        Networks          []Network         `yaml:"networks,omitempty" json:"networks,omitempty"`
        Network           NetworkDeprecated `yaml:"network,omitempty" json:"network,omitempty"` // DEPRECATED, use `networks` instead
        Env               map[string]string `yaml:"env,omitempty" json:"env,omitempty"`
diff --git a/pkg/qemu/qemu.go b/pkg/qemu/qemu.go
index d681400..8f582a7 100644
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -277,6 +277,15 @@ func Cmdline(cfg Config) (string, []string, error) {
        args = append(args, "-netdev", fmt.Sprintf("user,id=net0,net=%s,dhcpstart=%s,hostfwd=tcp:127.0.0.1:%d-:22",
                qemu.SlirpNetwork, qemu.SlirpIPAddress, cfg.SSHLocalPort))
        args = append(args, "-device", "virtio-net-pci,netdev=net0,mac="+limayaml.MACAddress(cfg.InstanceDir))
+       if y.SocketPort != nil && *y.SocketPort != 0 {
+               port := *y.SocketPort
[...]

PR #1183, opened 2022-11-20
Add feature to install kubernetes worker nodes

afbjorklund:
If your instances are connected through a proper network,
then it is possible to join additional workers to the control plane.

Beyond the URL of the apiserver (running on port 6443),
you will also need the secret "token" emitted by the initial installation.

---

The two (or more) Kubernetes nodes will need to be able reach eachother, though.

The default network ("user") will **not** allow this, but it is possible to use "socket":

* https://github.com/lima-vm/lima/issues/682

Probably still need the original NAT network to pull images, so needs eth0/eth1 ?

This is the setup done in minikube, originally by using the VirtualBox networks...

* https://github.com/kubernetes/minikube/issues/4938

### DIFF
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -1,7 +1,7 @@
 # Deploy kubernetes via kubeadm.
 # $ limactl start ./k8s.yaml
 # $ limactl shell k8s sudo kubectl
-
+#
 # It can be accessed from the host by exporting the kubeconfig file;
 # the ports are already forwarded automatically by lima:
 #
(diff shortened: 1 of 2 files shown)


afbjorklund:
Based on question in: https://github.com/lima-vm/lima/discussions/1181

examples/k8s.yaml
```diff
@@ -96,6 +96,16 @@ provision:
   script: |
     #!/bin/bash
     set -eux -o pipefail
+    # To install a worker node, fill in and uncomment this line:
```
AkihiroSuda, line 99: Needs more detailed usage
- How to start the control plane node, with vmnet
- How to get the `K8S_URL`, `K8S_TOKEN`, and `K8S_HASH` values

afbjorklund: I don't think vmnet works with Linux, for instance

afbjorklund:
Choosing the apiserver address and the flannel interface, with k3s and k8s:

k3s:
`--node-ip 10.0.2.100`
`--flannel-iface=eth1`

k8s:
`kubeadm --apiserver-advertise-address 10.0.2.100` (localAPIEndpoint: advertiseAddress: in yaml)
`/opt/bin/flanneld --iface=eth1` (in kube-flannel.yml)

When having nodes with multiple interfaces, and `eth0` is the host/slirp interface.

PR #1185, opened 2022-11-20
Proof-of-concept implementation of socket network

afbjorklund:
Revisited the old feature, and added the possibility to configure the (static) IP address automatically.

This allows setting up a network between two instances, without admin privileges or extra software.

The YAML could probably be structured better...

Issue #682

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -45,6 +45,8 @@ Optional feature enablers:
 - [`vmnet.yaml`](./vmnet.yaml): ⭐enable [`vmnet.framework`](../docs/network.md)
 - [`experimental/9p.yaml`](experimental/9p.yaml): [experimental] use 9p mount type
 - [`experimental/riscv64.yaml`](experimental/riscv64.yaml): [experimental] RISC-V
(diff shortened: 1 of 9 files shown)


AkihiroSuda:
I'd prefer this to be implemented in an external program that supports `.[]networks.socket`
https://github.com/lima-vm/lima/blob/cd504a41491faa3e05ae82fc9bade406c0e3684e/pkg/limayaml/limayaml.go#L179-L180

Maybe we can modify `socket_vmnet` to support that vmnet-less/rootless/dhcp-less mode.

afbjorklund:
It should be possible to do some tun/tap implemention for Linux, if wanting a real network ?

But that would require root access and extra programs. This was "builtin", like the virtfs for 9p.

AkihiroSuda:
Sorry, I'm not planning to merge this PR, as we are going to support multi-VM rootless networking in https://github.com/lima-vm/lima/pull/1383 , and bare socket networks can be still supported via `.[]networks.socket`.

afbjorklund:
No worries, I just rebased the merge conflict since it (the patch) came up during discussion

PR #1203, opened 2022-11-25
limactl list: add "VMTYPE" column

AkihiroSuda:
e.g,
```console
$ limactl ls
NAME       STATUS     SSH            VMTYPE    ARCH      CPUS    MEMORY    DISK      DIR
default    Stopped    127.0.0.1:0    qemu       x86_64    4       4GiB      100GiB    /Users/suda/.lima/default
vz         Stopped    127.0.0.1:0    vz         x86_64    4       4GiB      100GiB    /Users/suda/.lima/vz
```

### DIFF
--- a/cmd/limactl/disk.go
+++ b/cmd/limactl/disk.go
@@ -135,7 +135,7 @@ func diskListAction(cmd *cobra.Command, args []string) error {
 	}
 
 	w := tabwriter.NewWriter(cmd.OutOrStdout(), 4, 8, 4, ' ', 0)
-	fmt.Fprintln(w, "NAME\tSIZE\tDIR\tIN USE BY")
+	fmt.Fprintln(w, "NAME\tSIZE\tDIR\tIN-USE-BY")
 
 	if len(allDisks) == 0 {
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
cc @balajiv113

jandubois:
I would prefer that column labels do not include spaces (so would use `VMTYPE` instead of `VM TYPE`).

This will simplify things if we ever add an option like `--columns=NAME,STATUS` to customize the output columns.

AkihiroSuda:
Sure, but anyway we already have `limactl ls --format '{{.Name}},{{.Status}},{{.VMType}}'`

balajiv113 (approved):

AkihiroSuda:
Also added a new commit `limactl disk list: s/IN USE BY/IN-USE-BY/g`

jandubois (approved):

jandubois:
> we already have `limactl ls --format '{{.Name}},{{.Status}},{{.VMType}}'`

I know, but that produces essentially a CSV file and not a table.

PR #1206, opened 2022-11-25
vz: support "lima: shared" networking using VZNATNetworkDeviceAttachment

### LINKED ISSUE #1161: vz: support "vznat" network

AkihiroSuda:
See `examples/experimental/vz.yaml` .

Unlike QEMU, VZ does not need sudoers for enabling vmnet.

Fix #1161 

- - -
No support for `bridged` mode, as `VZBridgedNetworkDeviceAttachment` requires the `com.apple.vm.networking`entitlement which needs to contact an Apple representative.
 - https://developer.apple.com/documentation/virtualization/vzbridgednetworkdeviceattachment?language=objc
 - https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_vm_networking?language=objc

Probably we need to use `socket_vmnet` too for supporting `bridged` mode with VZ.
(The socket protocol is different, so we need to modify `socket_vmnet`)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,8 +46,6 @@ jobs:
       run: make
     - name: Install
       run: sudo make install
-    - name: Validate examples (except vmnet.yaml)
-      run: find examples -name '*.yaml' | grep -v 'vmnet.yaml' | xargs limactl validate
     - name: Uninstall
       run: sudo make uninstall
 
@@ -89,6 +87,8 @@ jobs:
       run: make
     - name: Install
       run: make install
+    - name: Validate examples
+      run: find examples -name '*.yaml' | xargs limactl validate
     - name: Install test dependencies
       # QEMU:      required by Lima itself
       # bash:      required by test-example.sh (OS version of bash is too old)
--- a/docs/network.md
+++ b/docs/network.md
@@ -44,12 +44,17 @@ If `useHostResolver` is false, then DNS servers can be configured manually in `l
 
 ## Managed VMNet networks (192.168.105.0/24)
 
+### QEMU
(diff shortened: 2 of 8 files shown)


AkihiroSuda:
cc @balajiv113 @jandubois

balajiv113:
@AkihiroSuda 
My suggestion here would be to provide a new config like `networks.vzNAT: true` (Like we discussed before). And this will be supported only for vz driver

Reasons for this would be, If we use `lima: shared` for vzNAT, even with this we cannot communicate across vm's (qemu vm talking to vz vm). But from config it clearly state `lima: shared` not driver specific

Once we are able to provide support for vz in socket_vmnet (Surely doable) then it will provide support for cross driver communication as well. Using `lima: shared` for vzNAT now would cause confusions at that time.

**What is needed to provide support for vz in socket_vmnet ?**
This is from my initial analysis, i might be missing something as well,
1. vz network packets doesn't contain header packet of 4bytes (contains length of original packet in qemu). So some logics have to be modified
2. dgram_socket based connectivity (Optional, as we can pipe dgram packets to unix socket as well)

AkihiroSuda:
> we cannot communicate across vm's (qemu vm talking to vz vm)

Why can't?

AkihiroSuda:
Opened a new PR:
- #1207

balajiv113:
> > we cannot communicate across vm's (qemu vm talking to vz vm)
> 
> Why can't?

I thought since the vmnet interfaces will be different for vzNAT and the socket_vmnet. So they maynot support cross communication

AkihiroSuda:
Looks like VZ uses `vmnet_enable_isolation_key` and it is not configurable 😞 
(`/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/vmnet.framework/Versions/Current/Headers/vmnet.h`)

PR #1207, opened 2022-11-25
vz: support `.[]networks.vzNAT` networking

### LINKED ISSUE #1161: vz: support "vznat" network

AkihiroSuda:
Introduces a new YAML property `.[]networks.vzNAT` for enabling `VZNATNetworkDeviceAttachment`. No root privilege is needed.

`.[]networks.lima` is reserved for future support of `socket_vmnet` for VZ.

Fix #1161
Replaces #1206

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,8 +46,6 @@ jobs:
       run: make
     - name: Install
       run: sudo make install
-    - name: Validate examples (except vmnet.yaml)
-      run: find examples -name '*.yaml' | grep -v 'vmnet.yaml' | xargs limactl validate
     - name: Uninstall
       run: sudo make uninstall
 
@@ -89,6 +87,8 @@ jobs:
       run: make
     - name: Install
       run: make install
+    - name: Validate examples
+      run: find examples -name '*.yaml' | xargs limactl validate
     - name: Install test dependencies
       # QEMU:      required by Lima itself
       # bash:      required by test-example.sh (OS version of bash is too old)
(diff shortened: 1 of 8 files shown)


AkihiroSuda:
> Is the IP address configurable?

No. I don't see the API for that.
https://developer.apple.com/documentation/virtualization/vznatnetworkdeviceattachment?language=objc

abiosoft:
> No. I don't see the API for that. https://developer.apple.com/documentation/virtualization/vznatnetworkdeviceattachment?language=objc

I was surprised to get an IP address in the same IP range used for socket_vmnet (on Colima) without any extra configuration i.e. ` 192.168.106.x`.

balajiv113:
> Looks like the IP is reachable from the host but not from other VZ instances 😞
> 
> ```
> $ curl 192.168.105.28
> curl: (7) Failed to connect to 192.168.105.28 port 80 after 3077 ms: No route to host
> ```
> 
> This issue does not happen with QEMU + `socket_vmnet` 🤔

That's odd, maybe gvisor-tap-vsock routing it via their interface but even then the host network should be able to access the other guest ip.

AkihiroSuda:
> other guest IP

It looks like VZNAT enables `vmnet_enable_isolation_key` by default and it is not configurable: https://github.com/cirruslabs/softnet/tree/main#working-model

balajiv113:
Support for it is there in private API
`0x001e7412c18 -[VZNATNetworkDeviceAttachment _setInterfaceIsolationEnabled:]`

Reference - https://github.com/Code-Hex/vz/wiki/Private-API-on-macOS-13

Apple might expose it in upcoming versions

balajiv113:
I was able to call using this private API, but interestingly VM crashes when i set it to false.

balajiv113 (commented):
Apart from the restrictions of vzNAT the changes looks good to me.

Just one point,
Should we also provide a way to get VM IP from limactl ?? I think its not present as of now as well or maybe document ??

AkihiroSuda:
> Should we also provide a way to get VM IP from limactl ?? I think its not present as of now as well or maybe document ??

`limactl shell INSTANCE ip` might suffice?

balajiv113:
> > Should we also provide a way to get VM IP from limactl ?? I think its not present as of now as well or maybe document ??
> 
> `limactl shell INSTANCE ip` might suffice?

Ah yes, that should be sufficient 👍

AkihiroSuda:
I assume this is mergable

PR #1210, opened 2022-11-27
Move hostagent command to separate binary

### LINKED ISSUE #1209: [v0.14.0-beta.0 regression] `limactl ls` takes 67ms even when there is no instance (v0.13 only took 11ms)
```console
$ git checkout v0.13.0 && make && ntimes -n 100 ./_output/bin/limactl ls
...
real average: 11.479705ms, max: 199.824274ms, min: 8.735928ms, std dev: 19.041288ms
```

```console
$ git checkout v0.14.0-beta.0 && make && ntimes -n 100 ./_output/bin/limactl ls
...
real average: 66.930248ms, max: 320.575082ms, min: 60.808298ms, std dev: 25.759197ms
```

afbjorklund:
Allows linking only the hostagent with the driver libraries, while keeping the limactl smaller and faster to start up.

Probably needs to live in "libexec" or something like that ? And maybe the subcommand should be renamed, too.

Closes #1209


afbjorklund:
This will break the other pending PRs, since they will need to add hostagent API - and not call the driver directly...

But that is probably a _good_ thing, even if it does make the hostagent slightly more complicated (more API calls) ?

AkihiroSuda:
Can't we just fix the issue on Code-Hex/vz?

afbjorklund:
> Can't we just fix the issue on Code-Hex/vz?

Sure, it depends on what the root cause is...

PR #1212, opened 2022-11-29
limactl: allow passing disks as args to `disk ls`

sam-berning:
Addresses some follow-up comments from https://github.com/lima-vm/lima/pull/1065.

Signed-off-by: Sam Berning <bernings@amazon.com>


AkihiroSuda (approved):
Thanks

PR #1216, opened 2022-11-29
support for validating unknown fields

### LINKED ISSUE #1162: vz: warn ignored YAML fields
>  Warnings should be printed for ignored YAML propertries: https://github.com/containerd/nerdctl/blob/c1a4159035b77a4a739a503b98139a17ac3440b8/pkg/composer/serviceparser/serviceparser.go#L38
>
> Can be another PR though

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/1147#discussion_r1016138374_

balajiv113:
Signed-off-by: Balaji Vijayakumar <kuttibalaji.v6@gmail.com>

Fixes #1162

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -121,22 +121,28 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 			}
 		}
 	}
+	var overrideCPUType bool
 	for k, v := range d.CPUType {
 		if len(v) > 0 {
+			overrideCPUType = true
 			cpuType[k] = v
 		}
 	}
 	for k, v := range y.CPUType {
 		if len(v) > 0 {
+			overrideCPUType = true
 			cpuType[k] = v
 		}
 	}
 	for k, v := range o.CPUType {
 		if len(v) > 0 {
+			overrideCPUType = true
 			cpuType[k] = v
 		}
 	}
-	y.CPUType = cpuType
+	if *y.VMType == QEMU || overrideCPUType {
+		y.CPUType = cpuType
+	}
 
 	if y.CPUs == nil {
 		y.CPUs = d.CPUs
@@ -176,7 +182,7 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 	if o.Video.Display != nil {
 		y.Video.Display = o.Video.Display
 	}
-	if y.Video.Display == nil || *y.Video.Display == "" {
+	if (y.Video.Display == nil || *y.Video.Display == "") && *y.VMType == QEMU {
 		y.Video.Display = pointer.String("none")
 	}
 
--- a/pkg/reflectutil/reflectutil.go
+++ b/pkg/reflectutil/reflectutil.go
@@ -1,0 +1,61 @@
+// This file has been adapted from https://github.com/containerd/nerdctl/blob/v1.0.0/pkg/reflectutil/reflectutil.go
+/*
+   Copyright The containerd Authors.
+   Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


balajiv113:
With this one problem i see is we are getting some unintended warnings like,
```
INFO[0000] Using the existing instance "vz"             
WARN[0000] Ignoring: vmType vz: [CPUType Firmware Video] 
WARN[0000] Ignoring: vmType vz: mounts[0]: [SSHFS NineP] 
WARN[0000] Ignoring: vmType vz: mounts[1]: [SSHFS NineP] 
```
Is this okay to have like this ?? Any pointers ??

One way i could think of is to use pointer and handle null checks accordingly.

pkg/reflectutil/reflectutil.go
```diff
@@ -0,0 +1,49 @@
+// This file has been adapted from https://github.com/containerd/nerdctl/blob/v1.0.0/pkg/reflectutil/reflectutil.go
```
AkihiroSuda, line 1: Please retain the copyright header

pkg/reflectutil/reflectutil.go
```diff
@@ -0,0 +1,49 @@
...
+func isEmpty(v reflect.Value) bool {
+	// NOTE: IsZero returns false for zero-length map and slice
+	if v.IsZero() {
+		return true
+	}
+	switch v.Kind() {
+	case reflect.Map, reflect.Slice:
+		return v.Len() == 0
+	}
+	return false
```
AkihiroSuda, line 48: Can we check empty structs like this? https://stackoverflow.com/a/52562726

balajiv113: The current code is already able to ignore empty structs. 

**For Eg:** 
In below, Firmware is not returned as unknown field
```
Yaml: {Firmware: {}}
```
The issue is more related to how we do in FillDefaults.
https://github.com/lima-vm/lima/blob/master/pkg/limayaml/defaults.go#L189-L191

From our defaults, we are getting
```
Yaml: {Firmware: {pointer.Bool(false)}}
```

We do similar for Mount SSHFS,NinePS setting default if null

AkihiroSuda: `Mount.NineP` can be just ignored without any warning (as long as the `MounType` is verified)

The `FillDefault()` function should leave `Video.Display` and `CPUType` empty when the driver is VZ, and the VZ driver should print warnings when those properties are non-empty.

pkg/reflectutil/reflectutil.go
balajiv113: Done

pkg/reflectutil/reflectutil.go
balajiv113: Done as per your suggestion

pkg/limayaml/defaults.go
```diff
@@ -121,22 +121,28 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 			}
 		}
 	}
+	yamlOverride := false
```
AkihiroSuda, line 124: Is this variable needed?

balajiv113: Yes yes, for cpuType we are setting defaults first and overriding based on yaml property. 

I went with this variable to avoid doing append(y, d, o) and get length from there.

AkihiroSuda, line 124:
```suggestion
	var overrideCPUType bool
```

balajiv113: Done 👍

AkihiroSuda (approved):
Thanks

PR #1220, opened 2022-12-01
vmnet: support detecting Homebrew's socket_vmnet path

AkihiroSuda:
On Homebrew environments, socket_vmnet is installed on:
- `/usr/local/opt/socket_vmnet/Cellar/<VERSION>/bin/socket_vmnet` (Intel)
- `/opt/homebrew/opt/socket_vmnet/Cellar/<VERSION>/bin/socket_vmnet` (ARM)

The binary is usually owned by a non-root admin user. (i.e., the homebrew user).
The group is typically set to "admin".

https://github.com/Homebrew/homebrew-core/commit/636ac0700e011b23c3e41cc4176b30e01a3af5ae

### DIFF
--- a/docs/network.md
+++ b/docs/network.md
@@ -53,11 +53,23 @@ The configuration steps are different across QEMU and VZ:
 ### QEMU
 #### Managed (192.168.105.0/24)
 
-Either [`socket_vmnet`](https://github.com/lima-vm/socket_vmnet) (since Lima v0.12) or [`vde_vmnet`](https://github.com/lima-vm/vde_vmnet) (Deprecated)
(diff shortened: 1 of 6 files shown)


chancez:
Shouldn't these already be in a users PATH, if so, why does Lima need to look for them itself?

AkihiroSuda:
> Shouldn't these already be in a users PATH

No, purposely
https://github.com/Homebrew/homebrew-core/blob/636ac0700e011b23c3e41cc4176b30e01a3af5ae/Formula/socket_vmnet.rb#L18

AkihiroSuda:
I'm planning to merge this on Monday

PR #1243, opened 2022-12-14
Check if growpart is installed before calling it

### LINKED ISSUE #1242: [v0.14.0 regression] On colima, docker data become invisible  (but not removed) ; `/var/lib/docker` is unmounted
On colima, docker data such as containers and images in `/var/lib/docker` become invisible after upgrading Lima from v0.13.0 to v0.14.0
```
colima start
docker pull hello-world
colima stop

brew upgrade
colima start
docker images
```

The data are not removed though, and can be rescued by downgrading Lima to v0.13.0
```
brew uninstall lima
git clone https://github.com/lima-vm/lima.git
cd lima
git checkout v0.13.0
make
make install
```

Apparently the issue seems specific to colima, and not reproducible with the vanilla Lima templates.
Tested the `default` template (`nerdctl pull hello-world`) and `alpine` template (`sudo touch /var/lib/foo`, as the vanilla `alpine` template does not contain container runtime).

- - -

Originally reported in https://github.com/abiosoft/colima/issues/506
Thanks to @rpkoller and @rfay for reporting.

jandubois:
Both `cloud-utils-growpart` (for `growpart`) and `e2fsprogs-extra` (for `resize2fs`) have only been recently added to alpine-lima.

Running an existing VM based on an older Alpine ISO will then fail the boot script calling utilities that have not been installed.

Untested, but I believe this
fixes #1242

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
@@ -16,11 +16,14 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
 	if [ -e /dev/disk/by-label/data-volume ]; then
 		# Find which disk is data volume on
 		DATA_DISK=$(blkid | grep "data-volume" | awk '{split($0,s,":"); sub(/\d$/, "", s[1]); print s[1]};')
-		# Automatically expand the data volume filesystem
-		growpart "$DATA_DISK" 1 || true
-		# Only resize when filesystem is in a healthy state
-		if e2fsck -f -p /dev/disk/by-label/data-volume; then
-			resize2fs /dev/disk/by-label/data-volume
+		# growpart command may be missing in older VMs
+		if command -v growpart >/dev/null 2>&1 && command -v resize2fs >/dev/null 2>&1; then
+			# Automatically expand the data volume filesystem
+			growpart "$DATA_DISK" 1 || true
+			# Only resize when filesystem is in a healthy state
+			if e2fsck -f -p /dev/disk/by-label/data-volume; then
+				resize2fs /dev/disk/by-label/data-volume || true
+			fi
 		fi
 		# Mount data volume
 		mount -t ext4 /dev/disk/by-label/data-volume /mnt/data


pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -16,11 +16,14 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
...
-		# Only resize when filesystem is in a healthy state
-		if e2fsck -f -p /dev/disk/by-label/data-volume; then
-			resize2fs /dev/disk/by-label/data-volume
+		# growpart command may be missing in older VMs
+		if command -v growpart >/dev/null 2>&1; then
+			# Automatically expand the data volume filesystem
+			growpart "$DATA_DISK" 1 || true
+			# Only resize when filesystem is in a healthy state
+			if e2fsck -f -p /dev/disk/by-label/data-volume; then
+				resize2fs /dev/disk/by-label/data-volume
```
AkihiroSuda, line 25: Even if resize2fs failed, the following `mount` has to be executed

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -16,11 +16,14 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
...
 		DATA_DISK=$(blkid | grep "data-volume" | awk '{split($0,s,":"); sub(/\d$/, "", s[1]); print s[1]};')
-		# Automatically expand the data volume filesystem
-		growpart "$DATA_DISK" 1 || true
-		# Only resize when filesystem is in a healthy state
-		if e2fsck -f -p /dev/disk/by-label/data-volume; then
-			resize2fs /dev/disk/by-label/data-volume
+		# growpart command may be missing in older VMs
+		if command -v growpart >/dev/null 2>&1; then
+			# Automatically expand the data volume filesystem
+			growpart "$DATA_DISK" 1 || true
```
AkihiroSuda, line 22: Not new in this PR, but the final exit code of the script should be set to non-zero when growpart fails

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
jandubois: Sure, added `|| true`

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
jandubois: Why does it make a difference; won't the failure of `growpart` (or `resize2fs`) show up in the cloud-init log anyways?

Please file an issue if you think this is really necessary; I won't have time for it until after the holidays, and will probably forget by then...

AkihiroSuda (approved):
Thanks

PR #1249, opened 2022-12-14
add inspect CLI option

### LINKED ISSUE #1246: Information about a guest
### Description

How do I get detailed information about a guest? I assume I just missed it, but nothing in `limactl info` or `limactl ls` (including `--json`) provides it.

I want to be able to inspect a VM (running or stopped) to find out information similar to `docker inspect`:

* what is the VM image?
* what are the mounts, r/w and r/o?
* resources (cpu, memory)
* containerd
* etc

All of the above might be reasonable handled if I can see the template that was used to launch it and its contents.

In theory, I can run `limactl start --list-templates`, but that doesn't give me the template contents, or which template (if any of those) was used to create a particular VM.

deitch:
Fixes #1246

### DIFF
--- a/cmd/limactl/inspect.go
+++ b/cmd/limactl/inspect.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func newInspectCommand() *cobra.Command {
+	inspectCommand := &cobra.Command{
+		Use:   "inspect [flags] [INSTANCE]...",
+		Short: "Inspect instances of Lima.",
+		Long: `Inspect instances of Lima.
+		Includes both the current state, as well as the combined configuration used to create the instance.
+		The output can be presented in one of several formats, using the --format <format> flag.
+		
+		--format json  - output in json format
+		--format yaml  - output in yaml format
+		--format table - output in table format
+		--format '{{ <go template> }}' - if the format begins and ends with '{{ }}', then it is used as a go template.
+	  `,
+		Args:              cobra.ArbitraryArgs,
+		RunE:              inspectAction,
+		ValidArgsFunction: inspectBashComplete,
+	}
+
+	inspectCommand.Flags().StringP("format", "f", "json", "output format, one of: 'json', 'yaml', a go-template")
+
+	return inspectCommand
+}
+
+func inspectAction(cmd *cobra.Command, args []string) error {
+	format, err := cmd.Flags().GetString("format")
+	if err != nil {
+		return err
+	}
+
+	allinstances, err := store.Instances()
+	if err != nil {
+		return err
+	}
+
+	instanceNames := []string{}
+	if len(args) > 0 {
+		for _, arg := range args {
+			matches := instanceMatches(arg, allinstances)
+			if len(matches) > 0 {
+				instanceNames = append(instanceNames, matches...)
+			} else {
+				logrus.Warnf("No instance matching %v found.", arg)
+			}
+		}
+	} else {
+		instanceNames = allinstances
+	}
+
+	// get the state and config for all the requested instances
+	var instances []*store.Instance
+	for _, instanceName := range instanceNames {
+		instance, err := store.Inspect(instanceName)
+		if err != nil {
+			return fmt.Errorf("unable to load instance %s: %w", instanceName, err)
+		}
+		instances = append(instances, instance)
+	}
+
+	return store.PrintInstances(cmd.OutOrStdout(), instances, format)
+}
+
+func inspectBashComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+	return bashCompleteInstanceNames(cmd)
+}
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -2,18 +2,23 @@ package store
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
+	"os/user"
 	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
 	"syscall"
+	"text/tabwriter"
 	"text/template"
 	"time"
 
 	"github.com/docker/go-units"
+	"github.com/goccy/go-yaml"
 	hostagentclient "github.com/lima-vm/lima/pkg/hostagent/api/client"
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/lima-vm/lima/pkg/store/dirnames"
@@ -46,6 +51,7 @@ type Instance struct {
 	HostAgentPID    int                `json:"hostAgentPID,omitempty"`
 	DriverPID       int                `json:"driverPID,omitempty"`
 	Errors          []error            `json:"errors,omitempty"`
+	Config          *limayaml.LimaYAML `json:"config,omitempty"`
 }
 
 func (inst *Instance) LoadYAML() (*limayaml.LimaYAML, error) {
@@ -79,6 +85,7 @@ func Inspect(instName string) (*Instance, error) {
 		inst.Errors = append(inst.Errors, err)
 		return inst, nil
 	}
+	inst.Config = y
 	inst.Arch = *y.Arch
 	inst.VMType = *y.VMType
 	inst.CPUType = y.CPUType[*y.Arch]
@@ -224,4 +231,65 @@ func AddGlobalFields(inst *Instance) (FormatData, error) {
 		return FormatData{}, err
 	}
 	return data, nil
+}
+
+// PrintInstances prints instances in a requested format to a given io.Writer.
+// Supported formats are "json", "yaml", "table", or a go template
+func PrintInstances(w io.Writer, instances []*Instance, format string) error {
+	switch format {
+	case "json":
+		b, err := json.Marshal(instances)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(w, string(b))
+	case "yaml":
+		b, err := yaml.Marshal(instances)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(w, string(b))
+	case "table":
+		w := tabwriter.NewWriter(w, 4, 8, 4, ' ', 0)
+		fmt.Fprintln(w, "NAME\tSTATUS\tSSH\tVMTYPE\tARCH\tCPUS\tMEMORY\tDISK\tDIR")
+
+		u, err := user.Current()
+		if err != nil {
+			return err
+		}
+		homeDir := u.HomeDir
+
+		for _, instance := range instances {
+			dir := instance.Dir
+			if strings.HasPrefix(dir, homeDir) {
+				dir = strings.Replace(dir, homeDir, "~", 1)
+			}
+			fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\t%d\t%s\t%s\t%s\n",
+				instance.Name,
+				instance.Status,
+				fmt.Sprintf("127.0.0.1:%d", instance.SSHLocalPort),
+				instance.VMType,
+				instance.Arch,
+				instance.CPUs,
+				units.BytesSize(float64(instance.Memory)),
+				units.BytesSize(float64(instance.Disk)),
+				dir,
+			)
+		}
+		return w.Flush()
+	default:
+		tmpl, err := template.New("format").Parse(format)
+		if err != nil {
+			return fmt.Errorf("invalid go template: %w", err)
+		}
+		for _, instance := range instances {
+			err = tmpl.Execute(w, instance)
+			if err != nil {
+				return err
+			}
+			fmt.Fprintln(w)
+		}
+		return nil
+	}
+	return nil
 }
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -1,17 +1,12 @@
 package main
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
-	"os/user"
 	"reflect"
 	"sort"
 	"strings"
-	"text/tabwriter"
-	"text/template"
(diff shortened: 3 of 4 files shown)


cmd/limactl/inspect.go
```diff
@@ -0,0 +1,90 @@
...
+func newInspectCommand() *cobra.Command {
+	inspectCommand := &cobra.Command{
+		Use:               "inspect [flags] [INSTANCE]...",
+		Short:             "Inspect instances of Lima.",
+		Args:              cobra.ArbitraryArgs,
+		RunE:              inspectAction,
+		ValidArgsFunction: inspectBashComplete,
+	}
+
+	inspectCommand.Flags().String("format", "yaml", "output format, one of: json, yaml")
```
AkihiroSuda, line 23: I'd prefer the default to be json for consistency with other commands (`limactl info`)

jandubois:
I don't know how useful it is to inspect multiple instances in one run, given that the `limayaml` struct doesn't even include the name of the instance. This is especially a problem when you run `limactl inspect` and get the data for all VMs, but there is no way to map the results back to individual instances.

So I wonder if you would want to include the `Instance` info together with the `limayaml` data, even though there is some redundancy, as some of it is copied into the instance store.

AkihiroSuda:
> I don't know how useful it is to inspect multiple instances in one run, given that the `limayaml` struct doesn't even include the name of the instance. This is especially a problem when you run `limactl inspect` and get the data for all VMs, but there is no way to map the results back to individual instances.
> 
> So I wonder if you would want to include the `Instance` info together with the `limayaml` data, even though there is some redundancy, as some of it is copied into the instance store.

Maybe just support only one instance for simplicity?
I'd also expect the output of `limactl inspect --format=yaml` can be directly passed to `limactl start`.

jandubois:
> Maybe just support only one instance for simplicity?

That was my first reaction too. But then I saw that you can pass multiple container names to `docker inspect`, so I thought maybe there is precedence to allowing it.

If we allow only one instance, then it also shouldn't be wrapped in an array. Right now you have to write:

```console
$ limactl inspect alpine --format json | jq .[0].hostResolver.enabled
true
```

When limited to a single instance you should be able to write:

```console
$ limactl inspect alpine --format json | jq .hostResolver.enabled
true
```

jandubois:
> I'd also expect the output of `limactl inspect --format=yaml` can be directly passed to `limactl start`.

You don't need the `--format yaml`, as any valid JSON document can be parsed by a YAML parser as well (JSON is a proper subset of YAML).

cmd/limactl/inspect.go
deitch: No problem, I will change it.

deitch:
> So I wonder if you would want to include the Instance info together with the limayaml data, even though there is some redundancy, as some of it is copied into the instance store.

I thought about that, but didn't follow through. If we have agreement on that, I can change it.

deitch:
> That was my first reaction too. But then I saw that you can pass multiple container names to docker inspect, so I thought maybe there is precedence to allowing it.

There was that, but my primary precedent was `limactl list`, which allows for multiple. In that case, though, it is separate json, i.e. not in an array. I put it in an array here so it is easy to consume.

I was trying for consistency.

AkihiroSuda:
Probably no need to include the instance metadata if the command supports only one argument.

Supporting multiple instances can be probably discussed later, if there is a usecase.

deitch:
> Probably no need to include the instance metadata if the command supports only one argument.

The more I think about it, the more useful it appears to me. `inspect` implies "tell me all about this", which is more than "give me the config yaml for this".

cmd/limactl/inspect.go
deitch: Changed to json

deitch:
> Supporting multiple instances can be probably discussed later, if there is a usecase.

For the json, it is easier to decide upfront. If we support just one instance now, and not in an array, then it is a `{..}` object. Then it becomes messy when we add multiple. Is it an array, so then sometimes it is an array and sometimes not? Or is it multiple individual objects, which is not valid json and makes parsing harder?

Choosing to support it now makes it consistently a json array, whether one or many. Or choosing not to do so now makes us much less likely to do so in the future.

deitch:
Ah, shoot. I started an experiment to see what an `Instance` with its config embedded would look like, but `store.Instance` does not _have_ a property for its config.

What do you think, should we add one?

jandubois:
> There was that, but my primary precedent was limactl list, which allows for multiple. In that case, though, it is separate json, i.e. not in an array. I put it in an array here so it is easy to consume.

I think the use-cases are different. For `limactl ls` you want to get a list of all VMs and their properties, whereas `limactl inspect` seems more targeted at drilling down in the minute details of a specific VM.

> I was trying for consistency.

As for consistency: we use e.g. `limactl ls --format '{{.Name}}'` to specify a Go template expression, and `limactl ls --json` to have JSON output.

This is at odds with using `limactl inspect --format json` to specify the format; I think we should have a `--yaml` option instead to switch to a different output format.

jandubois:
> Ah, shoot. I started an experiment to see what an `Instance` with its config embedded would look like, but `store.Instance` does not _have_ a property for its config.
> 
> What do you think, should we add one?

Not sure. The `Instance` has lots of fields copied over from the config, so if you were to embed the config directly, then I think the duplicated fields have to go.

This appeals to me because generally having redundant copies of data can be a problem when things get out of sync (we don't have this problem right now though). But I cannot remember why we didn't start out this way; maybe there is a good reason not to do it. @AkihiroSuda ?

deitch:
Trying to summarize what we have to date. It sounds like we want:

* `inspect` to work on one instance and one only at a time
* `--format {json|yaml}` to be replaced with `--json`  and `--yaml` with `--json` the default
* undecided about putting the entire `Instance` with the embedded config as of yet

I will say that I never have much liked having different CLI flags for different formats. Now we need to figure out if someone passed (`--format` and `--yaml`) or (`--yaml` and `--json`), any of which would conflict. Having just `--format` (or even `--output-format` or similar) forces one to pick a single one.

I get what you are saying about using go templates for `--format`, but we do not have that for inspect, and if we wanted to, I think that would fit better as another `--format` option:

* `limactl inspect --format json foo`
* `limactl inspect --format yaml foo`
* `limactl inspect --format template:'{{.Name}}' foo` or `limactl inspect --format template --template '{{.Name}}' foo` (or lots of others I can think of)

jandubois:
> I will say that I never have much liked having different CLI flags for different formats. Now we need to figure out if someone passed (`--format` and `--yaml`) or (`--yaml` and `--json`), any of which would conflict. Having just `--format` (or even `--output-format` or similar) forces one to pick a single one.

I agree, but we already have precedence in `limactl ls`. So if we want to change it, it should be done in a way that `limactl ls` can adopt too, while still supporting the old options to avoid breaking backwards compatibility.

Personally I don't mind the way `kubectl get` does it, which accepts e.g. `-o jsonpath="{.spec...}"`.

PS: I'm soon going offline for the rest of the year, so will not be providing much more feedback; you will have to hash it out with @AkihiroSuda 😄

AkihiroSuda:
> limactl inspect --format json foo

SGTM, but we will have to update `limactl ls --format` too for consistency

deitch:
I can add that to `limactl ls`.

deitch:
I have been thinking about this, and banging my head a little bit, not quite sure where to take it.

Current state of this PR:

* `limactl list` - list state of VM with 3 fields from config (vmtype, arch, disk) - template, json, table formats
* `limactl inspect` - show config of VM - json, yaml formats

Potential alternate:

* `limactl inspect` - show state of VM along with config - json, yaml formats

The problems with both of those are:

* current: `inspect` doesn't quite sound like "give me the configs", but it could be. It just seems a bit odd to tell people "`list` to see state, `inspect` to see config". It could work, but is a bit strange.
* proposed: it is odd to use `list` to see state, and `inspect` to see state _and_ config. If we did that, maybe we would be better off just expanding `list` command so that in non-table formats, it can show everything?

AkihiroSuda:
SGTM, but probably we will have to have `limactl inspect --format '{{yaml .Config}'` for config-only mode

deitch:
I don't understand @AkihiroSuda ; which option are you recommending?

AkihiroSuda:
Just your idea with an optional extension for `limactl inspect --format '{{yaml .Config}}'`.

`limactl inspect` may print both state and config by default, as you proposed

deitch:
@AkihiroSuda have another look. 

* `limactl inspect` now returns the state and the config as embedded
* `limactl inspect` now supports 3 formats: json, yaml, and a go-template
* `pkg/store.Inspect(instName string) (*Instance, error)` now keeps the yaml that it is loading anyways, which simplifies a lot

The one thing I did not address is the duplication that @jandubois raised [here](https://github.com/jandubois). There are 4 fields on `Instance` which now can be accessed via `Instance.Config`, so should we remove the duplication? I think so, but first wanted to check with this group.

Addressing the `list --format` now.

deitch:
Added the `--format` options to `limactl list`. Which is starting to look a _lot_ like `limactl inspect`, except for its option to just list names, list fields, and table output. But go ahead and have a look.

cmd/limactl/inspect.go
```diff
@@ -0,0 +1,109 @@
...
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(cmd.OutOrStdout(), string(b))
+	case format == "yaml":
+		b, err := yaml.Marshal(instances)
+		if err != nil {
+			return err
+		}
+		fmt.Fprintln(cmd.OutOrStdout(), string(b))
```
AkihiroSuda, line 85:
```go
switch format {
  case "json":
    format = "{{json .}}" // use template.FuncMap to implement the "json" func
  case "yaml":
    format = "{{yaml .}}"
}
```

cmd/limactl/inspect.go
```diff
@@ -0,0 +1,109 @@
...
+	inspectCommand := &cobra.Command{
+		Use:   "inspect [flags] [INSTANCE]...",
+		Short: "Inspect instances of Lima.",
+		Long: `Inspect instances of Lima.
+		Includes both the current state, as well as the combined configuration used to create the instance.
+		The output can be presented in one of several formats, using the --format <format> flag.
+		
+		  --format json - output in json format
+		  --format yaml - output in yaml format
+		  --format '{{ <go template> }}' - if the format begins and ends with '{{ }}', then it is used as a go template.
```
AkihiroSuda, line 25: Not robust, `{{` and `}}` can appear anywhere in the string

AkihiroSuda:
>  There are 4 fields on Instance which now can be accessed via Instance.Config, so should we remove the duplication?

For compatibility sake these fields have to be preserved

cmd/limactl/inspect.go
AkihiroSuda: e.g.,
https://github.com/lima-vm/lima/blob/f8f3a7b5664b877421d89745ca74f5891e7061ba/examples/docker.yaml#L6

cmd/limactl/inspect.go
AkihiroSuda: Also, the logic can be moved to `pkg/textutil` so that it can be shared across multiple commands

deitch:
> For compatibility sake these fields have to be preserved

No skin off my back if we leave them.

cmd/limactl/inspect.go
deitch: Actually, I think the simplest solution is, "if a reserved work like 'json' or 'yaml' or 'table', then use those, else just treat it as a go template".

cmd/limactl/inspect.go
deitch: Not quite, it creates a loop:

```
package github.com/lima-vm/lima/cmd/lima-guestagent
        imports github.com/lima-vm/lima/pkg/textutil
        imports github.com/lima-vm/lima/pkg/store
        imports github.com/lima-vm/lima/pkg/limayaml
        imports github.com/lima-vm/lima/pkg/networks
        imports github.com/lima-vm/lima/pkg/textutil: import cycle not allowed
```

I can put it in `pkg/store`

deitch: I do not see the advantage of doing the `switch` that way and defining the json func, vs the way it is now with a switch?

deitch:
Have another look

pkg/store/instance.go
```diff
@@ -225,3 +232,64 @@ func AddGlobalFields(inst *Instance) (FormatData, error) {
 	}
 	return data, nil
 }
+
+// PrintInstances prints instances in a requested format to a given io.Writer.
+// Supported formats are "json", "yaml", "table", or a go template
+func PrintInstances(w io.Writer, instances []*Instance, format string) error {
+	switch {
```
AkihiroSuda, line 239:
```suggestion
	switch format {
```

pkg/store/instance.go
```diff
@@ -225,3 +232,64 @@ func AddGlobalFields(inst *Instance) (FormatData, error) {
...
+				units.BytesSize(float64(instance.Disk)),
+				dir,
+			)
+		}
+		return w.Flush()
+	default:
+		tmpl, err := template.New("format").Parse(format)
+		if err != nil {
+			return fmt.Errorf("invalid go template: %w", err)
+		}
```
AkihiroSuda, line 284: I still expect this to support `{{json .Foo.Bar}}`, but it can be done later

AkihiroSuda:
Could you rebase with the current master to make the CI green?

pkg/store/instance.go
deitch: Good point. Before, when I was doing regex, it needed to switch on cases; now, with just the format name, it can do a regular `switch format {...}`. Updated.

pkg/store/instance.go
deitch: Is that what you meant? To explicitly support it, so it can build the filtering in? OK, I get it, but, yeah, let's get this in, then do that later.

deitch:
Rebased.

AkihiroSuda (approved):
Thanks

PR #1250, opened 2022-12-14
add support for reading template from stdin

### LINKED ISSUE #1247: ability to provide template via stdin?
### Description

In addition to the supported `template://`, `http://`, `https://`, `file://`, it would be helpful to be able to do:

```sh
$ cat template.yaml | limactl start -
# or maybe `limactl start stdin://`? Doesn't matter really.
```

deitch:
Fixes #1247 

Some comments:

* if `--tty` is not provided and the template is read from stdin, then it defaults to `--tty=false`
* if `--tty=true` as a CLI flag and the template is read from stdin, return an error
* if `--name` as a CLI flag is not provided and the template is read from stdin, return an error
* uses `-` as the indicator for stdin, inline with what most Unix-style programs do

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #1266, opened 2022-12-27
Hide columns in the list command

### LINKED ISSUE #1204: `limactl list`: dynamically reduces the column, depending on the terminal width
- https://github.com/lima-vm/lima/pull/1203

We are going to have too many columns:
```console
$ limactl ls
NAME       STATUS     SSH            VM TYPE    ARCH      CPUS    MEMORY    DISK      DIR
default    Stopped    127.0.0.1:0    qemu       x86_64    4       4GiB      100GiB    /Users/suda/.lima/default
vz         Stopped    127.0.0.1:0    vz         x86_64    4       4GiB      100GiB    /Users/suda/.lima/vz
```

When the terminal is small, some columns (especially "DIR") can be just hidden.

afbjorklund:
* Hide identical columns in the list command
    
    If all the instances have the same type or arch,
    then don't display the column - unless requested.

* Hide the directory column on narrow terminal
    
    If the terminal width does not fit all the list
    columns, then hide "dir" column - unless requested.

before
```
NAME    STATUS     SSH            VMTYPE    ARCH      CPUS    MEMORY    DISK   
   DIR
k8s     Stopped    127.0.0.1:0    qemu      x86_64    4       4GiB      100GiB  
  ~/.lima/k8s
```

after
```
NAME    STATUS     SSH            CPUS    MEMORY    DISK      DIR
k8s     Stopped    127.0.0.1:0    4       4GiB      100GiB    ~/.lima/k8s
```

```
NAME    STATUS     SSH            VMTYPE    ARCH      CPUS    MEMORY    DISK
k8s     Stopped    127.0.0.1:0    qemu      x86_64    4       4GiB      100GiB
```

One can use the `--all-fields` flag, to get the old behaviour back again.

On a wide enough terminal, the table output is the same as before.

Closes #1204

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -3,11 +3,14 @@ package main
 import (
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 	"sort"
 	"strings"
 
+	"github.com/cheggaaa/pb/v3/termutil"
 	"github.com/lima-vm/lima/pkg/store"
+	"github.com/mattn/go-isatty"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
@@ -163,6 +166,13 @@ func listAction(cmd *cobra.Command, args []string) error {
 
 	options := store.PrintOptions{AllFields: allFields}
 	out := cmd.OutOrStdout()
+	if out == os.Stdout {
+		if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {
+			if w, err := termutil.TerminalWidth(); err == nil {
+				options.TerminalWidth = w
+			}
+		}
+	}
 	return store.PrintInstances(out, instances, format, &options)
 }
 
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -54,6 +54,7 @@ func newListCommand() *cobra.Command {
 	listCommand.Flags().Bool("list-fields", false, "List fields available for format")
 	listCommand.Flags().Bool("json", false, "JSONify output")
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
+	listCommand.Flags().Bool("all-fields", false, "Show all fields")
 
 	return listCommand
 }
@@ -155,7 +156,14 @@ func listAction(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	return store.PrintInstances(cmd.OutOrStdout(), instances, format)
+	allFields, err := cmd.Flags().GetBool("all-fields")
+	if err != nil {
+		return err
+	}
+
+	options := store.PrintOptions{AllFields: allFields}
+	out := cmd.OutOrStdout()
+	return store.PrintInstances(out, instances, format, &options)
 }
 
 func listBashComplete(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -233,7 +233,8 @@ func AddGlobalFields(inst *Instance) (FormatData, error) {
 }
 
 type PrintOptions struct {
-	AllFields bool
+	AllFields     bool
+	TerminalWidth int
 }
 
 // PrintInstances prints instances in a requested format to a given io.Writer.
@@ -252,14 +253,45 @@ func PrintInstances(w io.Writer, instances []*Instance, format string, options *
 			archs[instance.Arch]++
 		}
 		all := options != nil && options.AllFields
+		width := 0
+		if options != nil {
+			width = options.TerminalWidth
+		}
+		columnWidth := 8
 		hideType := false
 		hideArch := false
 		hideDir := false
 
-		hideType = len(types) == 1 && !all
+		columns := 1 // NAME
+		columns += 2 // STATUS
+		columns += 2 // SSH
+		// can we still fit the remaining columns (7)
+		if width == 0 || (columns+7)*columnWidth > width && !all {
+			hideType = len(types) == 1
+		}
+		if !hideType {
+			columns++ // VMTYPE
+		}
 		// only hide arch if it is the same as the host arch
 		goarch := limayaml.NewArch(runtime.GOARCH)
-		hideArch = len(archs) == 1 && instances[0].Arch == goarch && !all
+		// can we still fit the remaining columns (6)
+		if width == 0 || (columns+6)*columnWidth > width && !all {
+			hideArch = len(archs) == 1 && instances[0].Arch == goarch
+		}
+		if !hideArch {
+			columns++ // ARCH
+		}
+		columns++ // CPUS
+		columns++ // MEMORY
+		columns++ // DISK
+		// can we still fit the remaining columns (2)
+		if width != 0 && (columns+2)*columnWidth > width && !all {
+			hideDir = true
+		}
+		if !hideDir {
+			columns += 2 // DIR
+		}
+		_ = columns
 
 		w := tabwriter.NewWriter(w, 4, 8, 4, ' ', 0)
 		fmt.Fprint(w, "NAME\tSTATUS\tSSH")
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -232,17 +232,48 @@ func AddGlobalFields(inst *Instance) (FormatData, error) {
 	return data, nil
 }
 
+type PrintOptions struct {
+	AllFields bool
+}
+
 // PrintInstances prints instances in a requested format to a given io.Writer.
 // Supported formats are "json", "yaml", "table", or a go template
-func PrintInstances(w io.Writer, instances []*Instance, format string) error {
+func PrintInstances(w io.Writer, instances []*Instance, format string, options *PrintOptions) error {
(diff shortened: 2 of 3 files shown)


cmd/limactl/list.go
```diff
@@ -48,6 +52,7 @@ func newListCommand() *cobra.Command {
 	listCommand.Flags().Bool("list-fields", false, "List fields available for format")
 	listCommand.Flags().Bool("json", false, "JSONify output")
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
+	listCommand.Flags().Bool("show", false, "Show all fields")
```
AkihiroSuda, line 55: Maybe

```suggestion
	listCommand.Flags().Bool("all-fields", false, "Show all fields")
```

cmd/limactl/list.go
```diff
@@ -165,8 +182,46 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+	show, err := cmd.Flags().GetBool("show")
+	if err != nil {
+		return err
+	}
+
+	out := cmd.OutOrStdout()
+	w := tabwriter.NewWriter(out, 4, 8, 4, ' ', 0)
+
+	fmt.Fprint(w, "NAME\tSTATUS\tSSH")
+	cols := 5 // two for status, ssh
```
AkihiroSuda, line 205: and three for what?

AkihiroSuda:
Thanks, but needs rebase

cmd/limactl/list.go
jandubois: I think a `col` here means 8 characters, so 1 for `name`, 2 for `status` and 2 for `ssh` add up to 5. See line 218 on how it is used; that seems to be the only place the value is actually read and not updated.

afbjorklund: Maybe it would make more sense to count characters

afbjorklund: The "col" was a supposed to be a column in the TabWriter, defined as 8

```go
func NewWriter(output io.Writer, minwidth, tabwidth, padding int, padchar byte, flags uint) *Writer
```

AkihiroSuda:
Found a regression: The `ARCH` field is no longer printed even when the terminal width is enough.

LGTM other than that.

AkihiroSuda:
I'd also like to see unit tests

afbjorklund:
> Found a regression: The `ARCH` field is no longer printed even when the terminal width is enough.

No regression, it was separate features. But we can make the arch/type dedup depend on width too ?

jandubois:
> Found a regression: The `ARCH` field is no longer printed even when the terminal width is enough.

I think I'm find with the `ARCH` field being removed, even when there is enough space, as long as all machines use the native arch:

```
hideArch = len(archs) == 1 && archs[0] == runtime.GOARCH
```

I don't think it adds any value in that situation, which I expect is going to be rather common.

AkihiroSuda:
Lints are failing

afbjorklund:
> Lints are failing

Needed to upgrade my `golangci-lint`

afbjorklund:
> I don't think it adds any value in that situation

I never found any value from the SSH column, but maybe it has some ?

Especially the `127.0.0.1:0` is weird, but also others.

You need to use  `show-ssh` anyway, to get  something actually usable.

jandubois:
> Especially the `127.0.0.1:0` is weird, but also others.

I agree, but maybe we should leave this for future discussion / another PR, so we can get this change wrapped up and merged?

afbjorklund:
Ok, so let's autohide again and see how that goes.

afbjorklund:
Just that it always shows the vm type, even if there is only one vm type (on Linux)

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Sorry needs rebase

cmd/limactl/list.go
```diff
@@ -155,7 +159,22 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+	allFields, err := cmd.Flags().GetBool("all-fields")
+	if err != nil {
+		return err
+	}
+
+	options := store.PrintOptions{AllFields: allFields}
+	out := cmd.OutOrStdout()
+	if out == os.Stdout {
+		if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {
+			options.IsTerminal = true
```
jandubois, line 171: The `options.IsTerminal` field is not really used for anything (because it being `false` is treated the same way as `TerminalWidth` being `0`), so I would drop it.

pkg/store/instance.go
```diff
@@ -250,31 +264,96 @@ func PrintInstances(w io.Writer, instances []*Instance, format string) error {
...
+		archs := map[string]int{}
+		for _, instance := range instances {
+			types[instance.VMType]++
+			archs[instance.Arch]++
+		}
+		all := options != nil && options.AllFields
+		width := 0
+		if options != nil && options.IsTerminal {
+			width = options.TerminalWidth
+		}
```
jandubois, line 277:
```suggestion
		if options != nil {
			width = options.TerminalWidth
		}
```

pkg/store/instance.go
```diff
@@ -250,31 +264,96 @@ func PrintInstances(w io.Writer, instances []*Instance, format string) error {
...
+		hideType := false
+		hideArch := false
+		hideDir := false
+
+		columns := 1 // NAME
+		columns += 2 // STATUS
+		columns += 2 // SSH
+		if width != 0 && (columns+7)*8 > width && !all {
```
jandubois, line 285: Can we have a comment explaining where the magic `7` comes from? Maybe define a constant to give it a name.

Maybe also use a `columnWidth = 8` constant everywhere?

pkg/store/instance.go
```diff
@@ -250,31 +264,96 @@ func PrintInstances(w io.Writer, instances []*Instance, format string) error {
...
+			hideType = len(types) == 1
+		}
+		if !hideType {
+			columns++ // VMTYPE
+		}
+		hideArch = len(archs) == 1 && !all
```
jandubois, line 291: We only want to hide the ARCH column if all instances are using the host arch

```suggestion
		hideArch = len(archs) == 1 && instances[0].Arch == runtime.GOARCH && !all
```

pkg/store/instance.go
```diff
@@ -250,31 +264,96 @@ func PrintInstances(w io.Writer, instances []*Instance, format string) error {
...
+			fmt.Fprint(w, "\tDIR")
+		}
+		fmt.Fprintln(w)
 
-		u, err := user.Current()
+		homeDir, err := homeDir()
 		if err != nil {
 			return err
 		}
-		homeDir := u.HomeDir
```
jandubois, line 260: I don't understand the point of this change. The `homeDir()` wrapper seems to add complexity for no benefit.

jandubois (changes requested):
ARCH should only be hidden when all instances use the host arch.

pkg/store/instance.go
afbjorklund: It was needed for the unit test, but I can duplicate it there instead.

pkg/store/instance.go
afbjorklund: The width of the remaining columns is somewhat inaccurate, because of the padding.

It probably has to calculate the width of all possibly combinations, and fit the biggest ?

All of these funky auto-hiding columns should probably have a template func or a subclass ?

```go
w := tabwriter.NewWriter(w, 4, 8, 4, ' ', 0)
```

pkg/store/instance.go
afbjorklund: I think strictly speaking that the runtime.GOARCH is the binary arch and not the host arch

At least we have had some qemu/rosetta incidents in the path, so maybe it needs a function ?

afbjorklund:
Did the rebase and the simple edits, others are left todo.

pkg/store/instance.go
jandubois: > I think strictly speaking that the runtime.GOARCH is the binary arch and not the host arch

Yes, but we use it everywhere else as the host arch (e.g. in the `lima.yaml` defaults), so it would still be consistent.

> At least we have had some qemu/rosetta incidents in the path, so maybe it needs a function ?

Does this really happen? If you use the `x86_64` version of `limactl` on `aarch64` with Rosetta2, it would also try to call the `qemu-system-x86_64` binary, which most likely won't be installed.

I think the code I suggested is good enough, and matches what we do in other places to determine the host arch.

pkg/store/instance.go
jandubois: > The width of the remaining columns is somewhat inaccurate, because of the padding.
>
> It probably has to calculate the width of all possibly combinations, and fit the biggest ?

I would be fine with leaving the heuristics as-is, and maybe revisit this if or when we have examples where they fail dramatically.

I would just like the code to be a bit more self-explanatory. The column counting code is somewhat at odds with the hard-coded constants in the conditions.

AkihiroSuda:
> Did the rebase and the simple edits, others are left todo.

Should this PR be marked as a draft?

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@jandubois LGTY?

jandubois (approved):
Thanks, LGTM!

PR #1267, opened 2022-12-27
Add template functions for list/inspect message

### LINKED ISSUE #614: Add command to show the start message again

afbjorklund:
Makes it possible to show messages using a format string like:

    {{.Name}}{{printf "\n"}}{{.Message | indent 2 | trim "\n"}}

Without having to add a command like `message` or a flag like `--message`.

Sample output:

```console
$ limactl ls -f '{{.Name}}{{printf "\n"}}{{.Message | indent 2 | trim "\n"}}'
k3d
  To run `kubectl` on the host (assumes kubectl is installed), run the following commands:
  ------
  mkdir -p "/home/anders/.lima/k3d/conf"
  export KUBECONFIG="/home/anders/.lima/k3d/conf/kubeconfig.yaml"
  limactl shell k3d sh -c 'cat ~/.kube/config' >$KUBECONFIG
  kubectl ...
  ------
  ------
  export DOCKER_HOST="unix:///home/anders/.lima/k3d/sock/docker.sock"
  docker ...
  ------
kind
  To run `kubectl` on the host (assumes kubectl is installed), run the following commands:
  ------
  mkdir -p "/home/anders/.lima/kind/conf"
  export KUBECONFIG="/home/anders/.lima/kind/conf/kubeconfig.yaml"
  limactl shell kind sh -c 'cat ~/.kube/config' >$KUBECONFIG
  kubectl ...
  ------
  ------
  export DOCKER_HOST="unix:///home/anders/.lima/kind/sock/docker.sock"
  docker ...
  ------
```

Closes #614

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -41,7 +41,8 @@ func newListCommand() *cobra.Command {
 		  --format yaml  - output in yaml format
 		  --format table - output in table format
 		  --format '{{ <go template> }}' - if the format begins and ends with '{{ }}', then it is used as a go template.
-
+` + store.FormatHelp +
+			`
 		The following legacy flags continue to function:
 		  --json - equal to '--format json'
 		`,
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -285,6 +285,7 @@ func PrintInstances(w io.Writer, instances []*Instance, format string) error {
 		if err != nil {
 			return err
 		}
+		data.Message = strings.TrimSuffix(instance.Message, "\n")
 		err = tmpl.Execute(w, data)
 		if err != nil {
 			return err
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -211,6 +211,11 @@ type FormatData struct {
 	IdentityFile string
 }
 
+var FormatHelp = "\n" + textutil.PrefixString("\t\t",
+	"These functions are available to go templates:\n\n"+
+		textutil.IndentString(2,
+			strings.Join(textutil.FuncHelp, "\n")+"\n"))
+
 func AddGlobalFields(inst *Instance) (FormatData, error) {
 	var data FormatData
 	data.Instance = *inst
--- a/pkg/textutil/textutil.go
+++ b/pkg/textutil/textutil.go
@@ -3,6 +3,7 @@ package textutil
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"text/template"
@@ -74,8 +75,46 @@ var TemplateFuncMap = template.FuncMap{
 		}
 		return "---\n" + strings.TrimSuffix(b.String(), "\n")
 	},
-	"indent":  IndentString,
-	"missing": MissingString,
+	"indent": func(a ...interface{}) (string, error) {
+		if len(a) == 0 {
+			return "", errors.New("function takes at at least one string argument")
+		}
+		if len(a) > 2 {
+			return "", errors.New("function takes at most 2 arguments")
+		}
+		var ok bool
+		size := 2
+		if len(a) > 1 {
+			if size, ok = a[0].(int); !ok {
+				return "", errors.New("optional first argument must be an integer")
+			}
+		}
+		text := ""
+		if text, ok = a[len(a)-1].(string); !ok {
+			return "", errors.New("last argument must be a string")
+		}
+		return IndentString(size, text), nil
+	},
+	"missing": func(a ...interface{}) (string, error) {
+		if len(a) == 0 {
+			return "", errors.New("function takes at at least one string argument")
+		}
+		if len(a) > 2 {
+			return "", errors.New("function takes at most 2 arguments")
+		}
+		var ok bool
+		message := "<missing>"
+		if len(a) > 1 {
+			if message, ok = a[0].(string); !ok {
+				return "", errors.New("optional first argument must be a string")
+			}
+		}
+		text := ""
+		if text, ok = a[len(a)-1].(string); !ok {
+			return "", errors.New("last argument must be a string")
+		}
+		return MissingString(message, text), nil
+	},
 }
 
 // TemplateFuncHelp is help for TemplateFuncMap.
--- a/pkg/textutil/textutil.go
+++ b/pkg/textutil/textutil.go
@@ -25,14 +25,15 @@ func ExecuteTemplate(tmpl string, args interface{}) ([]byte, error) {
 
 // PrefixString adds prefix to beginning of each line
 func PrefixString(prefix string, text string) string {
-	result := ""
+	result := []string{}
 	for _, line := range strings.Split(text, "\n") {
 		if line == "" {
+			result = append(result, "")
 			continue
 		}
-		result += prefix + line + "\n"
+		result = append(result, prefix+line)
 	}
-	return result
+	return strings.Join(result, "\n")
 }
 
 // IndentString add spaces to beginning of each line
@@ -74,13 +75,11 @@ var TemplateFuncMap = template.FuncMap{
 		return "---\n" + strings.TrimSuffix(b.String(), "\n")
 	},
 	"indent":  IndentString,
-	"trim":    TrimString,
 	"missing": MissingString,
 }
 
 // TemplateFuncHelp is help for TemplateFuncMap.
 var FuncHelp = []string{
 	"indent <size>: add spaces to beginning of each line",
-	"trim <cutset>: remove characters from beginning and end",
 	"missing <message>: return message if the text is empty",
 }
--- a/pkg/textutil/textutil.go
+++ b/pkg/textutil/textutil.go
@@ -23,6 +23,37 @@ func ExecuteTemplate(tmpl string, args interface{}) ([]byte, error) {
 	return b.Bytes(), nil
 }
 
+// PrefixString adds prefix to beginning of each line
+func PrefixString(prefix string, text string) string {
+	result := ""
+	for _, line := range strings.Split(text, "\n") {
+		if line == "" {
+			continue
+		}
+		result += prefix + line + "\n"
+	}
+	return result
+}
+
+// IndentString add spaces to beginning of each line
+func IndentString(size int, text string) string {
+	prefix := strings.Repeat(" ", size)
+	return PrefixString(prefix, text)
+}
+
+// TrimString removes characters from beginning and end
+func TrimString(cutset string, text string) string {
+	return strings.Trim(text, cutset)
+}
+
+// MissingString returns message if the text is empty
+func MissingString(message string, text string) string {
+	if text == "" {
+		return message
+	}
+	return text
+}
+
 // TemplateFuncMap is a text/template FuncMap.
 var TemplateFuncMap = template.FuncMap{
 	"json": func(v interface{}) string {
@@ -42,4 +73,14 @@ var TemplateFuncMap = template.FuncMap{
 		}
 		return "---\n" + strings.TrimSuffix(b.String(), "\n")
 	},
+	"indent":  IndentString,
+	"trim":    TrimString,
+	"missing": MissingString,
+}
+
+// TemplateFuncHelp is help for TemplateFuncMap.
+var FuncHelp = []string{
+	"indent <size>: add spaces to beginning of each line",
+	"trim <cutset>: remove characters from beginning and end",
+	"missing <message>: return message if the text is empty",
 }
--- a/pkg/textutil/textutil_test.go
+++ b/pkg/textutil/textutil_test.go
@@ -9,11 +9,11 @@ import (
 )
 
 func TestPrefixString(t *testing.T) {
-	assert.Equal(t, "- foo\n", PrefixString("- ", "foo"))
+	assert.Equal(t, "- foo", PrefixString("- ", "foo"))
 	assert.Equal(t, "- foo\n- bar\n", PrefixString("- ", "foo\nbar\n"))
 }
 func TestIndentString(t *testing.T) {
-	assert.Equal(t, "  foo\n", IndentString(2, "foo"))
+	assert.Equal(t, "  foo", IndentString(2, "foo"))
 	assert.Equal(t, "  foo\n  bar\n", IndentString(2, "foo\nbar\n"))
 }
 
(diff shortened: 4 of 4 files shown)


afbjorklund:
And another template function, to be able to decorate instances without a message (as show in example)

`limactl ls -f '{{.Name}}{{printf "\n"}}{{.Message | missing "<no message>" | indent 2 | trim "\n"}}'`

cmd/limactl/list.go
```diff
@@ -126,7 +149,12 @@ func listAction(cmd *cobra.Command, args []string) error {
 	}
 
 	if goFormat != "" {
-		tmpl, err := template.New("format").Parse(goFormat)
+		funcMap := template.FuncMap{
+			"indent":  indentString,
+			"trim":    trimString,
+			"missing": missingString,
+		}
```
AkihiroSuda, line 156: Can we move this to under `pkg/`, and use this for `limactl inspect` too?

AkihiroSuda: I'd also like to see unit tests

jandubois: Also documentation; not sure where to put it though

jandubois:
I wonder if it is possible to provide default values for the 2nd argument, so you could write

```
limactl ls -f '{{.Name}}{{"\n"}}{{.Message | missing | indent | trim }}'
```

instead of[^1]

```
limactl ls -f '{{.Name}}{{printf "\n"}}{{.Message | missing "<missing>" | indent 2 | trim "\n"}}'
```

Maybe using `i ...interface{}` for the second argument can make it work?

Of course we then need a round of bikeshedding to agree on the right default values. E.g. I think `trim` should trim all whitespace and not just newlines.

[^1]: You can use `{{"\n"}}` instead of `{{printf "\n"}}`.

afbjorklund:
The simple approach does **not** work, at least...

`template: format:1:10: executing "format" at <indent>: wrong number of args for indent: want 2 got 1`

> Of course we then need a round of bikeshedding to agree on the right default values.

Like there is no argument about indent sizes... ;-)

>  think trim should trim all whitespace and not just newlines.

Trimming on all whitespace, removes the indent ?

cmd/limactl/list.go
afbjorklund: > Also documentation; not sure where to put it though

Currently the documentation is duplicated in the cmds

`--format '{{ <go template> }}' - if the format begins and ends with '{{ }}', then it is used as a go template.`

afbjorklund: Maybe something like `--list-fields`, like `--list-funcs` ?

Which by the way seems to be missing from `inspect'

jandubois:
> The simple approach does **not** work, at least...
> 
> `template: format:1:10: executing "format" at <indent>: wrong number of args for indent: want 2 got 1`

It worked for me with:

```golang
// IndentString add spaces to beginning of each line
func IndentString(i ...interface{}) string {
	size := 2
	if len(i) > 1 {
		size = i[0].(int)
	}
	text := i[len(i)-1].(string)
	prefix := strings.Repeat(" ", size)
	return PrefixString(prefix, text)
}
```

Of course needs proper type checking to avoid it from panicking.

```console
$ limactl ls -f '{{.Name}}{{"\n"}}{{.Dir | indent | trim "\n"}}'
alpine
  /Users/jan/.lima/alpine
default
  /Users/jan/.lima/default

$ limactl ls -f '{{.Name}}{{"\n"}}{{.Dir | indent 4 | trim "\n"}}'
alpine
    /Users/jan/.lima/alpine
default
    /Users/jan/.lima/default
```

jandubois:
The need for the `trim` function seems to come from the fact that `PrefixString` will append a trailing newline, even when the input string didn't have one. I think with this change, we can get rid of the `trim` function:

```diff
--- pkg/templateutil/templateutil.go
+++ pkg/templateutil/templateutil.go
@@ -8,17 +8,32 @@ import (
 // PrefixString adds prefix to beginning of each line
 func PrefixString(prefix string, text string) string {
        result := ""
+       trailingNewline := strings.HasSuffix(text, "\n")
        for _, line := range strings.Split(text, "\n") {
                if line == "" {
                        continue
                }
                result += prefix + line + "\n"
        }
-       return result
+       if trailingNewline {
+               return result
+       }
+       return strings.TrimSuffix(result, "\n")
 }
```

jandubois:
> > Of course we then need a round of bikeshedding to agree on the right default values.
> Like there is no argument about indent sizes... ;-)

Suggestions for defaults:

* `indent`: `2`

* `missing`: `<missing>`

* `prefix`: `> `

An alternative for `missing` would be `<no value>` which the template package uses to render `nil`.

For `prefix` my first though would be the empty string, but that makes it a no-op. Or 2 spaces, but that makes it an alias for `indent`. So `> ` is kind of a fallback. I guess I don't really see the need for `prefix`...

jandubois:
> Of course needs proper type checking to avoid it from panicking.

Doesn't actually panic:

```console
$ limactl ls -f '{{.Name}}{{"\n"}}{{2 | indent 2}}'
alpine
FATA[0000] template: format:1:23: executing "format" at <indent 2>: error calling indent: interface conversion: interface {} is int, not string

$ limactl ls -f '{{.Name}}{{"\n"}}{{.Dir | indent "x"}}'
alpine
FATA[0000] template: format:1:26: executing "format" at <indent "x">: error calling indent: interface conversion: interface {} is string, not int
```

So maybe the code is good enough without additional type checking.

jandubois:
But maybe it is more user-friendly to provide custom errors:

```golang
// IndentString add spaces to beginning of each line
func IndentString(i ...interface{}) (string, error) {
	if len(i) == 0 {
		return "", errors.New("function takes at at least one string argument")
	}
	if len(i) > 2 {
		return "", errors.New("function takes at most 2 arguments")
	}
	var ok bool
	size := 2
	if len(i) > 1 {
		if size, ok = i[0].(int); !ok {
			return "", errors.New("optional first argument must be an integer")
		}
	}
	text := ""
	if text, ok = i[len(i)-1].(string); !ok {
		return "", errors.New("last argument must be a string")
	}
	prefix := strings.Repeat(" ", size)
	return PrefixString(prefix, text), nil
}
```

Resulting in:

```console
$ limactl ls -f '{{indent}}'
FATA[0000] template: format:1:2: executing "format" at <indent>: error calling indent: function takes at at least one string argument

$ limactl ls -f '{{2 | indent}}'
FATA[0000] template: format:1:6: executing "format" at <indent>: error calling indent: last argument must be a string

$ limactl ls -f '{{.Dir | indent "x"}}'
FATA[0000] template: format:1:9: executing "format" at <indent "x">: error calling indent: optional first argument must be an integer

$ limactl ls -f '{{.Dir | indent 4 8}}'
FATA[0000] template: format:1:9: executing "format" at <indent 4 8>: error calling indent: function takes at most 2 arguments
```

afbjorklund:
> The need for the `trim` function seems to come from the fact that `PrefixString` will append a trailing newline, even when the input string didn't have one.

The message had a trailing newline, and then the --format added another newline. So that's why it was trimmed.

Otherwise there was an empty space, between instances.

```
foo
  ...

bar
  ...

baz
  <no message>
```

jandubois:
> The message had a trailing newline, and then the --format added another newline. So that's why it was trimmed.

I see. I don't think the trailing newline on the `message` is ever useful when evaluated in the `--format` context, so I would prefer to always trim it before evaluating the template:

```diff
@@ -289,6 +286,7 @@ func PrintInstances(w io.Writer, instances []*Instance, format string) error {
                        return fmt.Errorf("invalid go template: %w", err)
                }
                for _, instance := range instances {
+                       instance.Message = strings.TrimSuffix(instance.Message, "\n")
                        err = tmpl.Execute(w, instance)
                        if err != nil {
                                return err
```

My goal is to have the templates do what the user want automatically, while keeping the templates as concise as possible.

```console
$ limactl ls -f '{{.Name}}{{printf "\n"}}{{.Message | missing | indent }}'
alpine
  <missing>
default
  <missing>
docker
  ------
  docker context create lima-docker --docker "host=unix:///Users/jan/.lima/docker/sock/docker.sock"
  docker run hello-world
  ------
podman
  ------
  podman system connection add lima-podman "unix:///Users/jan/.lima/podman/sock/podman.sock"
  podman system connection default lima-podman
  podman run quay.io/podman/hello
  ------
```

Can you think of a use-case where you would want to keep the trailing newline?

afbjorklund:
Apparently it is a YAML syntax thing, if we wanted the message to **not** have a newline it should use `|-`.

```
# Message. Information to be shown to the user, given as a Go template for the instance.
# The same template variables as for listing instances can be used, for example {{.Dir}}.
# You can view the complete list of variables using `limactl list --list-fields` command.
# It also includes {{.HostOS}} and {{.HostArch}} vars, for the runtime GOOS and GOARCH.
# 🟢 Builtin default: null
# message: |
#   This will be shown to the user.
```

Now we use the template also to show json and yaml (since 820b78761946dd4b3902536e98ec949c876e0fc3) so otherwise it would be lying ?

afbjorklund:
> For `prefix` my first though would be the empty string, but that makes it a no-op. Or 2 spaces, but that makes it an alias for `indent`. So `> ` is kind of a fallback. I guess I don't really see the need for `prefix`...

Removed the func map, it was more of an implementation detail anyway.

afbjorklund:
Did the rebase, but probably want to merge "textutil" (no tests?) and "templateutil"

jandubois:
> Now we use the template also to show json and yaml (since [820b787](https://github.com/lima-vm/lima/commit/820b78761946dd4b3902536e98ec949c876e0fc3)) so otherwise it would be lying ?

I would not consider it "lying"; it is just that we can define `.Message` to be the expanded template, with any trailing newlines stripped. I believe there never is any reason to keep the trailing newlines, but I may be wrong. Please show any example where they would be useful.

Note the you still have `.Config.Message` containing the original template. I am **not** suggesting that we change this value, only the already processed `.Message` one.

pkg/templateutil/templateutil.go
```diff
@@ -0,0 +1,49 @@
+package templateutil
```
AkihiroSuda, line 1: This should be merged to `pkg/textutil`.

AkihiroSuda: ("templateutil" is reserved for `template://TEMPLATE` utils)

jandubois:
Conversation here seems to have stalled. From my point of view, I would like to see the `indent` and `missing` function to support default arguments (`2`, and `"missing"`, respectively).

I also think that `PrefixString` should not append an extra newline if the input value didn't have one, and that the `.Message` field should automatically be stripped of trailing newlines. I think this would completely eliminate the need for the `trim` function, which I would like to remove unless there is some demonstrated use-case.

With those changes I would be willing to approve this PR.

Let me know if you need help, and I can propose a commit that implements the changes I would like to see.

Or continue the conversation, if you think my suggestions are problematic.

afbjorklund:
I put it on the backlog, since the feature wasn't that important...

* #614

But will move to the new "util" directory, and add the new arguments.

We can strip the newline in Execute, but it did feel like a bit of a hack.

afbjorklund:
Added the default arguments, with the custom argument checker and errors.

The existing functions were a bit weird, since they call `panic` instead of errors.

Also seemed like they had hardcoded some stuff for list, so less generic ?

e.g. the JSON adds a trailing newline and  the YAML adds a document marker...

jandubois (approved):
I'm fine with merging this PR as-is, and addressing any further feedback in a follow-up PR.

### Feedback:

I expected the default `missing` argument to be `<missing>`, as `<no message>` only makes sense when the final argument is `.Message`. However, I expect this to be the primary use case of `missing`, so maybe this is fine.

It just occurred to me that we could simply set `.Message` to `<no message>` if it is empty, eliminating the need to use the `missing` function in this context. Not sure if we can agree on this though.

I also just noticed that the `limactl ls --help` output is badly formatted (separate issue, not affected by this PR):

```console
$ limactl ls --help
List instances of Lima.
		The output can be presented in one of several formats, using the --format <format> flag.

		  --format json  - output in json format
		  --format yaml  - output in yaml format
		  --format table - output in table format
		  --format '{{ <go template> }}' - if the format begins and ends with '{{ }}', then it is used as a go template.

		These functions are available to go templates:

		  indent <size>: add spaces to beginning of each line
		  missing <message>: return message if the text is empty

		The following legacy flags continue to function:
		  --json - equal to '--format json'

Usage:
  limactl list [flags] [INSTANCE]...

Aliases:
  list, ls

Flags:
  -f, --format string   output format, one of: json, yaml, table, go-template (default "table")
  -h, --help            help for list
      --json            JSONify output
      --list-fields     List fields available for format
  -q, --quiet           Only show names

Global Flags:
      --debug   debug mode
```

The first block of text (before the actual "Usage:") is indented by 16 spaces (2 tabs), whereas it should be indented just 2 spaces.

afbjorklund:
Changed the default missing text (to `missing`), I just wanted to show a single message anyway.

Should remove the two tabs then, so that the generated func help lines up with the static text.

afbjorklund:
We can do it separately, just that this PR adds two tabs - that would also need to be changed, accordingly

jandubois:
> We can do it separately, just that this PR adds two tabs - that would also need to be changed, accordingly

Ok, let's do it separately, so that we can draw down the big backlog of in-flight PRs.

PR #1277, opened 2023-01-04
Support for VBox driver

### LINKED ISSUE #1287: VirtualBox driver, for legacy compatibility and portability
### Description

Once upon a time, there was something called Docker Machine and it was running on VirtualBox.

It did have some other drivers, like [qemu](https://github.com/machine-drivers/docker-machine-driver-qemu), but the virtualbox driver was the portable and default...

Now that Lima has a driver framework, it could be useful to have a "vbox" driver next to the "qemu" ?

It does support features like host networking and shared folders, that QEMU is still struggling with.

* https://www.virtualbox.org/manual/ch06.html#network_hostonly

* https://www.virtualbox.org/manual/ch04.html#sharedfolders

-----

Pros: might run on more platforms, such as legacy releases or other operating systems

* https://www.virtualbox.org/manual/ch01.html#hostossupport

* https://docs.freebsd.org/en/books/handbook/virtualization/#virtualization-host-virtualbox

Cons: only supports Intel images, no support for COW images (resulting in larger files)

But the biggest advantage is having another driver that also works on non-Mac hosts.

And it can help to improve the driver API, if testing with more different implementations ?

afbjorklund:
This is a legacy driver for Oracle VirtualBox, x86_64 only.

It also has support for shared folders and hostonly networks,
but they are not implemented just yet (sshfs and nat only)

Closes #1287

* #1287

### DIFF
--- a/pkg/vbox/vbox_driver.go
+++ b/pkg/vbox/vbox_driver.go
@@ -224,3 +224,18 @@ func (l *LimaVBoxDriver) Stop(ctx context.Context) error {
 	err := qCmd.Run()
 	return err
 }
+
+func (l *LimaVBoxDriver) Register(ctx context.Context) error {
+	name := "lima-" + l.Instance.Name
+	args := []string{"registervm", filepath.Join(l.Instance.Dir, name)}
+	qCmd := exec.CommandContext(ctx, "VBoxManage", args...)
+	err := qCmd.Run()
+	return err
+}
+
+func (l *LimaVBoxDriver) Unregister(ctx context.Context) error {
+	args := []string{"unregistervm", "lima-" + l.Instance.Name}
+	qCmd := exec.CommandContext(ctx, "VBoxManage", args...)
+	err := qCmd.Run()
+	return err
+}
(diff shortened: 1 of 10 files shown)


afbjorklund:
There is a fair deal of copy/paste, between pkg/vz/disk and pkg/vbox/disk

And there is some similar cleanup to do, between pkg/qemu and pkg/vbox...

pkg/vbox/vbox_driver.go
```diff
@@ -0,0 +1,230 @@
...
+		"--medium", filepath.Join(l.Instance.Dir, filenames.CIDataISO)).Run(); err != nil {
+		logrus.Debugf("cidata %v", err)
+		return err
+	}
+
+	logrus.Debugf("network")
+
+	slirpMACAddress := limayaml.MACAddress(l.Instance.Dir)
+	if out, err := exec.CommandContext(ctx, qExe, "modifyvm", name,
+		"--nic1", "nat",
```
balajiv113, line 150: Just putting it out.

Is vpn connection from vm is working??
Also please check if our hostResolver works as well.

In vz, with nat we had issues with these, i believe same should happen here as well. 
This was the main reason why we used gvisor-tap-vsock to setup slirp network

afbjorklund: It does not, there are separate flags for host resolver afaik.

Early days yet, this was more of a hack to see if it worked...

afbjorklund:
To be moved to an external driver, once framework is available.

PR #1281, opened 2023-01-06
Make sure `limactl list --json` to print line-delimited JSON

### LINKED ISSUE #1280: `limactl list --json` now returns an array instead of newline delimited JSON
### Description

Originally reported in https://github.com/abiosoft/colima/issues/572

I don't think this was intentional, and should be fixed before the next release, as it breaks backwards compatibility.

And I think `limactl inspect` should be changed accordingly for consistency.

AkihiroSuda:
Fix #1280
A regression in 75594fa33bd73a2cc80a389993c7573b4bf3ec58 (#1249)

### DIFF
--- a/pkg/textutil/textutil.go
+++ b/pkg/textutil/textutil.go
@@ -2,7 +2,12 @@ package textutil
 
 import (
 	"bytes"
+	"encoding/json"
+	"fmt"
+	"strings"
 	"text/template"
+
+	"github.com/goccy/go-yaml"
 )
 
 // ExecuteTemplate executes a text/template template.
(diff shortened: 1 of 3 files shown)


pkg/textutil/textutil.go
```diff
@@ -17,3 +22,24 @@ func ExecuteTemplate(tmpl string, args interface{}) ([]byte, error) {
...
+// TemplateFuncMap is a text/template FuncMap.
+var TemplateFuncMap = template.FuncMap{
+	"json": func(v interface{}) string {
+		var b bytes.Buffer
+		enc := json.NewEncoder(&b)
+		enc.SetEscapeHTML(false)
+		if err := enc.Encode(v); err != nil {
+			panic(fmt.Errorf("failed to marshal as JSON: %+v: %w", v, err))
+		}
+		return strings.TrimSpace(b.String())
```
jandubois, line 35: Why is the `TrimSpace` needed?

AkihiroSuda: Because otherwise it produces "\n\n"-delimited JSON lines

jandubois: Wouldn't it be clearer to show the intent with this:

```suggestion
		return strings.TrimSuffix(b.String(), "\n")
```

jandubois (approved):
LGTM

pkg/textutil/textutil.go
AkihiroSuda: done

jandubois (approved):
Thanks, LGTM

PR #1289, opened 2023-01-10
Return NODATA for IPv6 AAAA question types

Nino-K:
Previously we were replying with A response for AAAA queries which was causing an error in some clients e.g nginx. The new approach waits for 1 second and then replies with NODATA. This approach allows for more semantic reply even when a client uses the same Transaction ID for both A and AAAA questions.

This PR is meant to address the issue that was reported on Rancher Desktop here: https://github.com/rancher-sandbox/rancher-desktop/issues/3447

Signed-off-by: Nino Kodabande <nkodabande@suse.com>

### DIFF
--- a/pkg/hostagent/dns/dns.go
+++ b/pkg/hostagent/dns/dns.go
@@ -7,14 +7,18 @@ import (
 	"net"
 	"runtime"
 	"strings"
+	"time"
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

Issue #1285, opened 2023-01-06
Lima get stuck when certain frequency of DNS resolution occurs


matsukaz:
### Description

**Actual Behavior**

When I continue to execute nslookup/ping/dig or any other DNS name resolution operations within Lima, Lima get stuck after a certain number of times (lasting about 4 minutes).
This issue also occurs inside Docker container when using Rancher Desktop, so this is critial issue for any system that uses container linking, like Apache+PHP and MySQL.

**Steps to Reproduce**

1. Login to Lima and keep running nslookup.
```
$ rdctl shell
lima-rancher-desktop:/Users/xxx$ while true; do nslookup www.google.co.jp; done
```
2. While the above process is running, show a list of UDP open files that qemu-system-aarch64 handles on host OS.
```
$ lsof -p $(pgrep qemu-system-aarch64) | grep "UDP"
...
qemu-syst 6788 xxxx  119u  IPv4 0x2c6ecf140850ff5f         0t0                 UDP *:63544
qemu-syst 6788 xxxx  120u  IPv4 0x2c6ecf140851762f         0t0                 UDP *:63398
```
3. A number of UDP open files keep increasing and after it reaches to FD=1024u, Lima get stuck.
```
$ lsof -p $(pgrep qemu-system-aarch64) | grep "UDP"
...
qemu-syst 6788 xxxx  1023u  IPv4 0x2c6ecf14085191bf         0t0                 UDP *:54486
qemu-syst 6788 xxxx  1024u  IPv4 0x2c6ecf140852088f         0t0                 UDP *:62934
```
4. If you wait about 4 minutes, all UDP open files get released and Lima starts running again.

**Workaround**

If you set `hostResolver.enabled=false` , this issue doesn't occur.

```
hostResolver:
  enabled: false
```

**Versions**

- Lima 0.14.2
- Rancher Desktop 1.7.0

**Operating System / Build Version / CPU**

- MacOS Monterey 12.6 (M1 2020)
- MacOS Ventura 13.0.1 (Intel Core i5, 2019)

jandubois:
I know we looked into this before; @Nino-K do you remember the current status?

Nino-K:
@matsukaz Thanks for reporting the issue, I managed to repro this, however, I'm not 100% convinced that the hostresolver is the root cause of this issue. I still see the problem occurs even when I disable the hostresolver. 

[...]

PR #1294, opened 2023-01-11
Closes the responseWriter after serving the Answer

Nino-K:
Currently, we are leaving the response writer open for both connected sockets (TCP) and non-connected sockets (UPD), this change closes the response writer accordingly.

This change is meant to address the user-reported issue on Lima: https://github.com/lima-vm/lima/issues/1285

Signed-off-by: Nino Kodabande <nkodabande@suse.com>


AkihiroSuda (approved):
Thanks

Issue #1300, opened 2023-01-14
One-line configuration of the kubernetes client


afbjorklund:
### Description

This is a re-iteration of an earlier feature request, that was deemed redundant:

* https://github.com/lima-vm/lima/issues/514

I would like to make the config more similar to the config of Docker / Podman.

This by letting lima do the steps of creating the config directory and the file...

----

That way, we don't have to carry the shell scripts and config locations around.

Making it more similar, to how we tunnel the docker.sock and podman.sock ?

`export DOCKER_HOST=$(limactl list docker --format 'unix://{{.Dir}}/sock/docker.sock')`

`export CONTAINER_HOST=$(limactl list podman --format 'unix://{{.Dir}}/sock/podman.sock')`

So the request is to make ConfigFile into a configuration item, like PortForward:

```diff
diff --git a/examples/k3s.yaml b/examples/k3s.yaml
index c3d0769..438b018 100644
--- a/examples/k3s.yaml
+++ b/examples/k3s.yaml
@@ -5,8 +5,7 @@
 # It can be accessed from the host by exporting the kubeconfig file;
 # the ports are already forwarded automatically by lima:
 #
-# $ export KUBECONFIG=$PWD/kubeconfig.yaml
-# $ limactl shell k3s sudo cat /etc/rancher/k3s/k3s.yaml >$KUBECONFIG
+# $ export KUBECONFIG=$(limactl list k3s --format 'unix://{{.Dir}}/conf/kubeconfig.yaml')
 # $ kubectl get no
 # NAME       STATUS   ROLES                  AGE   VERSION
 # lima-k3s   Ready    control-plane,master   69s   v1.21.1+k3s1
@@ -54,11 +53,12 @@ probes:
     The k3s kubeconfig file has not yet been created.
     Run "limactl shell k3s sudo journalctl -u k3s" to check the log.
     If that is still empty, check the bottom of the log at "/var/log/cloud-init-output.log".
+configFiles:
+- guestFile: "/etc/rancher/k3s/k3s.yaml"
+  hostFile: "{{.Dir}}/conf/kubeconfig.yaml"
 message: |
   To run `kubectl` on the host (assumes kubectl is installed), run the following commands:
   ------
-  mkdir -p "{{.Dir}}/conf"
   export KUBECONFIG="{{.Dir}}/conf/kubeconfig.yaml"
-  limactl shell {{.Name}} sudo cat /etc/rancher/k3s/k3s.yaml >$KUBECONFIG
[...]

PR #1301, opened 2023-01-14
Add possibility to copy ConfigFile to host

afbjorklund:
* Refactor PortForward guest/host processing
* Add possibility to copy ConfigFile to host

Issue #1300

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -336,6 +336,13 @@ networks:
 #   hostPortRange: [1, 65535]
 # # Any port still not matched by a rule will not be forwarded (ignored)
 
+# Copy files from the guest to the host. Copied after provisioning scripts have been completed.
+# copyToHost:
+# - guest: "/etc/myconfig.cfg"
+#   host: "{{.Dir}}/copied-from-guest/myconfig"
+# # "guest" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
+# # "host" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
+
 # Message. Information to be shown to the user, given as a Go template for the instance.
 # The same template variables as for listing instances can be used, for example {{.Dir}}.
 # You can view the complete list of variables using `limactl list --list-fields` command.
--- a/examples/k3s.yaml
+++ b/examples/k3s.yaml
@@ -5,8 +5,7 @@
 # It can be accessed from the host by exporting the kubeconfig file;
 # the ports are already forwarded automatically by lima:
 #
-# $ export KUBECONFIG=$PWD/kubeconfig.yaml
-# $ limactl shell k3s sudo cat /etc/rancher/k3s/k3s.yaml >$KUBECONFIG
+# $ export KUBECONFIG=$(limactl list k3s --format 'unix://{{.Dir}}/copied-from-guest/kubeconfig.yaml')
 # $ kubectl get no
 # NAME       STATUS   ROLES                  AGE   VERSION
 # lima-k3s   Ready    control-plane,master   69s   v1.21.1+k3s1
@@ -54,11 +53,12 @@ probes:
     The k3s kubeconfig file has not yet been created.
     Run "limactl shell k3s sudo journalctl -u k3s" to check the log.
     If that is still empty, check the bottom of the log at "/var/log/cloud-init-output.log".
+copyToHost:
+- guest: "/etc/rancher/k3s/k3s.yaml"
+  host: "{{.Dir}}/copied-from-guest/kubeconfig.yaml"
 message: |
   To run `kubectl` on the host (assumes kubectl is installed), run the following commands:
   ------
-  mkdir -p "{{.Dir}}/conf"
-  export KUBECONFIG="{{.Dir}}/conf/kubeconfig.yaml"
-  limactl shell {{.Name}} sudo cat /etc/rancher/k3s/k3s.yaml >$KUBECONFIG
+  export KUBECONFIG="{{.Dir}}/copied-from-guest/kubeconfig.yaml"
   kubectl ...
   ------
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -527,6 +527,47 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 	y.CACertificates.Certs = caCerts
 }
 
+func executeGuestTemplate(format string) (bytes.Buffer, error) {
+	tmpl, err := template.New("").Parse(format)
+	if err == nil {
+		user, _ := osutil.LimaUser(false)
+		data := map[string]string{
+			"Home": fmt.Sprintf("/home/%s.linux", user.Username),
+			"UID":  user.Uid,
+			"User": user.Username,
+		}
+		var out bytes.Buffer
+		if err := tmpl.Execute(&out, data); err == nil {
+			return out, nil
+		}
+	}
+	return bytes.Buffer{}, err
+}
+
+func executeHostTemplate(format string, instDir string) (bytes.Buffer, error) {
+	tmpl, err := template.New("").Parse(format)
+	if err == nil {
+		user, _ := osutil.LimaUser(false)
+		home, _ := os.UserHomeDir()
+		limaHome, _ := dirnames.LimaDir()
+		data := map[string]string{
+			"Dir":  instDir,
+			"Home": home,
+			"Name": filepath.Base(instDir),
+			"UID":  user.Uid,
+			"User": user.Username,
+
+			"Instance": filepath.Base(instDir), // DEPRECATED, use `{{.Name}}`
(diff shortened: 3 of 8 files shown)


examples/k3s.yaml
```diff
@@ -54,11 +53,12 @@ probes:
     The k3s kubeconfig file has not yet been created.
     Run "limactl shell k3s sudo journalctl -u k3s" to check the log.
     If that is still empty, check the bottom of the log at "/var/log/cloud-init-output.log".
+configFiles:
```
AkihiroSuda, line 56:
```suggestion
copyToHost:
```

examples/default.yaml
```diff
@@ -336,6 +336,16 @@ networks:
...
 # # Any port still not matched by a rule will not be forwarded (ignored)
 
+# Config files. Copied after provisioning scripts have been completed.
+# copyToHost:
+# - guest: "/etc/myconfig.cfg"
+#   host: myconfig
+# # default: mode: system
+# # "guest" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
+# # "host" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
+# # "mode" can be "system" (with root) or "user" (without root), similar to provision
```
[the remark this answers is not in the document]
afbjorklund: Just seemed wrong to hardcode sudo/root, for those cases where it is not needed.

pkg/limayaml/defaults.go
```diff
@@ -527,6 +532,47 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 	y.CACertificates.Certs = caCerts
 }
 
+func processGuest(format string) (bytes.Buffer, error) {
```
jandubois, line 535: I find `process` kind of meaningless, as it doesn't really tell me what is happening.

Maybe `expandGuestTemplate` (and `expandHostTemplate`) are better names?

jandubois (commented):
Except for the file permissions this seems to be ready for merging, I think, so maybe remove the Draft status?

pkg/limayaml/defaults.go
afbjorklund: Those names sound better, I was struggling a bit

afbjorklund: Or even executeGuestTemplate, like template.Execute

jandubois (approved):
Thanks, LGTM.

@AkihiroSuda Please merge if you have no objections!

pkg/store/filenames/filenames.go
```diff
@@ -45,8 +45,11 @@ const (
 	VzIdentifier       = "vz-identifier"
 	VzEfi              = "vz-efi"
 
-	// SocketDir is the default location for forwarded sockets with a relative paths in HostSocket
-	SocketDir = "sock"
+	// InstanceSocketDir is the default location for forwarded sockets with a relative path in HostSocket
+	InstanceSocketDir = "sock"
+
+	// InstanceConfigDir is the default location for config files with a relative path in HostFile
+	InstanceConfigDir = "conf"
```
AkihiroSuda, line 52: I still think this name is confusing.

What about "copied-from-guest"

AkihiroSuda (changes requested):
I still think "conf" ("config") is a confusing name

AkihiroSuda:
Also please consider squashing comnits

afbjorklund:
> I still think "conf" ("config") is a confusing name'

> Also please consider squashing commits

I can squash them when the new names are finalized, I was deliberately trying to scope it down from the last time...

Now it looks like a more generic functionality, but the requirement was only to do a `mkdir -p` and ssh a `sudo cat`.

I don't have a much better name, for a directory for $KUBECONFIG files.

It was supposed to be an internal concept, similar to the "sock" directory

Could call it "yaml", based on file type ?

Or maybe even "kube", to specialize it.

jandubois:
> > I still think "conf" ("config") is a confusing name'
> 
> I don't have a much better name, for a directory for $KUBECONFIG files.

The use for `$KUBECONFIG` files is your immediate use case, but the feature is more general, and called `copyToHost`. I think @AkihiroSuda wants to quite literally call it `$INSTANCE_DIR/copied-from-guest`, which is pretty descriptive.

I admit that personally I find it too long, but I can see the value for users not yet familiar with the concept.

I too don't have a better suggestion.

> It was supposed to be an internal concept, similar to the "sock" directory

Anything documented in the `default.yaml` is not an internal concept, but a supported API.

That said, I thought `conf` was fine, but I see now that the concept is bigger than that. However, I'm not sure what else you would want to exfiltrate from the guest after provisioning, but config files.

> Could call it "yaml", based on file type ?

No, that doesn't work for me. Do you actually means to create different directories, based on the file extension of the copied file? What about files without an extension?
 
> Or maybe even "kube", to specialize it.

I think the idea is to generalize the name beyond this specific use case, not to specialize it even further.

Note that this is also just the default location for relative host names, so mostly useful for shared templates. For local instances I would probably use absolute host names to point to a convenient location instead.

afbjorklund:
"yaml" was extrapolating from "sock" (i.e. based on file suffix)

"kube" was based on the default location, $HOME/.kube/config

If I understand the code correctly, using relative paths would also work.

```yaml
portForwards:
- guestSocket: "/var/run/docker.sock"
  hostSocket: "docker.sock"
```

```yaml
copyToHost:
- guest: "/etc/kubernetes/admin.conf"
  host: "kubeconfig.yaml"
```

It wouldn't help for printing out the location, though (probably opposite)

At least not without more variables, for the subdirectories of `{{.Dir}}`

Something like "client" perhaps, but it wouldn't make sense without "config"

Can go with `copied-from-guest`

Maybe OK, with the new wrapper.

pkg/store/filenames/filenames.go
afbjorklund: Done

jandubois:
> "yaml" was extrapolating from "sock" (i.e. based on file suffix)

But "sock" was not based on file suffix (except they accidentally match). It is based on the fact the the only thing you can forward are sockets. And the name was shortened to `sock` to make it more likely to stay under the maximum socket name length of 103 characters on macOS.

> Can go with copied-from-guest

Yeah, I can go with that if @AkihiroSuda feels strongly about not using `conf`.

> I don't really see the difference, I guess the sockets would be forwarded-from-guest

No, as explained above, there can only be sockets inside `sock`, but there can be any kind if files in `copied-from-guest`. And the name of the socket directory needs to be short. There is no such limitation on the "copied from guest" directory name.

> And think it looks rather long and ugly

I agree, but I can see the point of it being more self-explanatory.

Alternatively, maybe this could be solved by more extensive explanation in the `default.yaml` comments?

afbjorklund:
I removed the global constants, so now it is up to each example to name the local files to whatever suits best...

The default.yaml shows the suggested `{{.Dir}}/copied-from-guest/myconfig` location, as an example host path.

AkihiroSuda:
Thanks, but the commit does not seem updated

jandubois:
> Thanks, but the commit does not seem updated

Seems updated to me. The current state of the PR now requires an absolute path for both the guest and the host filename, and no longer provides a default location.

It is not clear to me what you are requesting: should the default location be restored and set to `.../copied-from-guest`?

examples/default.yaml
```diff
@@ -336,6 +336,13 @@ networks:
 #   hostPortRange: [1, 65535]
 # # Any port still not matched by a rule will not be forwarded (ignored)
 
+# Config files. Copied after provisioning scripts have been completed.
```
AkihiroSuda, line 339:
```suggestion
# Copy files from the guest to the host. Copied after provisioning scripts have been completed.
```

AkihiroSuda (approved):
Thanks, LGTM but with a single nit

afbjorklund:
This also opens for choosing a more suitable name for each example, matching the files that are copied.

jandubois:
> This also opens for choosing a more suitable name for each example, matching the files that are copied.
> 
> Like `{{.Dir}}/kube/config.yaml`

This is actually a reason why I preferred to have a default location because this now opens the instance directory up to users creating arbitrarily named directories, which could conflict with files or directories Lima wan't to create in the future.

This was the reason we specified `sock/` as the default location for forwarded sockets instead of creating them directly in the instance dir.

afbjorklund:
Actually they can copy them anywhere on the host, since we require it to be an absolute path and not just a filename...

jandubois:
> Actually they can copy them anywhere on the host, since we require it to be an absolute path and not just a filename...

Yes, but users copy from the provided examples, and if our examples create random directories inside the instance directories, then that means that we implicitly endorse doing so.

AkihiroSuda:
Any idea to move this forward?  (for v0.16 ?)

jandubois:
> Any idea to move this forward? (for v0.16 ?)

Does this need anything more than the single line comment change you requested above, and that @afbjorklund gave the 👍 for, but hasn't pushed yet?

afbjorklund:
Missed to update this, didn't know if the names were settled.

jandubois (approved):
Thanks, LGTM

PR #1302, opened 2023-01-16
Add kubectl kubernetes client wrapper

afbjorklund:
Will connect to a running "k3s" or "k8s" instance.

Similar to other wrappers, but uses conf not sock.

Issue #1300

Needs PR #1301 in order to create the "conf" files

### DIFF
--- a/cmd/kubectl.lima
+++ b/cmd/kubectl.lima
@@ -1,0 +1,30 @@
+#!/bin/sh
+set -eu
+: "${LIMA_INSTANCE:=}"
+: "${KUBECTL:=kubectl}"
+
+if [ -z "$LIMA_INSTANCE" ]; then
+  if [ "$(limactl ls -f '{{.Status}}' k3s 2>/dev/null)" = "Running" ]; then
+    LIMA_INSTANCE=k3s
+  elif [ "$(limactl ls -f '{{.Status}}' k8s 2>/dev/null)" = "Running" ]; then
+    LIMA_INSTANCE=k8s
+  else
+    echo "No k3s or k8s running instances found. Either start one with" >&2
+    echo "limactl start --name=k3s template://k3s" >&2
+    echo "limactl start --name=k8s template://k8s" >&2
+    echo "or set LIMA_INSTANCE to the name of your Kubernetes instance" >&2
+    exit 1
+  fi
+elif [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
+  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl start --name=$LIMA_INSTANCE\` to create a new instance" >&2
+  exit 1
+fi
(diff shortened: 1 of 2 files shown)


cmd/kubectl.lima
```diff
@@ -0,0 +1,27 @@
...
+: "${LIMA_INSTANCE:=}"
+: "${KUBECTL:=kubectl}"
+
+if [ -z "$LIMA_INSTANCE" ]; then
+  if [ "$(limactl ls -f '{{.Status}}' k3s 2>/dev/null)" = "Running" ]; then
+    LIMA_INSTANCE=k3s
+  elif [ "$(limactl ls -f '{{.Status}}' k8s 2>/dev/null)" = "Running" ]; then
+    LIMA_INSTANCE=k8s
+  else
+    echo "no running instance found, run \`limactl start\` to start an instance" >&2
```
jandubois, line 12: Maybe include a hint that we are only looking for instance names `k3s` and `k8s`, and tell the user to set `LIMA_INSTANCE` if they use any other?

jandubois (commented):
Thanks, LGTM, but maybe the error message can be a bit more helpful?

afbjorklund:
It at least outputs the name now, but it doesn't mention which template to use...

```console
$ ./cmd/docker.lima 
instance "docker" does not exist, run `limactl start --name=docker template://docker` to create a new instance
$ ./cmd/podman.lima 
instance "podman" does not exist, run `limactl start --name=podman template://podman` to create a new instance
$ ./cmd/kubectl.lima 
no running instance found, run `limactl start` to start an instance
looked for instance names "k3s" and "k8s", maybe set LIMA_INSTANCE?
```

Maybe should add similar output here, and check if it exists or if it is just stopped...

jandubois:
> Maybe should add similar output here, and check if it exists or if it is just stopped...

Yeah, probably should.

> No `k3s` or `k8s` running instances found. Either start one with
>   limactl start --name=k3s template://k3s
>   limactl start --name=k8s template://k8s
> or set `LIMA_INSTANCE` to the name of your Kubernetes instance.

It quickly gets complicated when you want to check for stopped instances, and then recommand to start those instead of creating a new one, but maybe try to see if you can write it concisely?

jandubois:
Note that the other scripts don't catch it if an instance exists, but hasn't been started. But as you wrote earlier, I think the error message you get in that case should be sufficient already.

jandubois:
I haven't tested again right now, but LGTM. But it requires #1301 to be final before we can approve/merge this, as the name of the `conf` directory may still change.

jandubois (approved):
Thanks, LGTM

Requires #1301 to be merged

AkihiroSuda:
Next time please set the milestone before merging

jandubois:
> Next time please set the milestone before merging

Sorry about that, I normally do, but I guess I missed it because most of the current PRs already had a milestone assigned, so I forgot to double-check.

PR #1305, opened 2023-01-18
Add VM labels

### LINKED ISSUE #1304: Add VM labels support
### Description

It'd be nice to have VM labels implemented. This way one can filter VM list by labels. So far I'd say it's enough to:
1. Add `labels` list to YAML schema.
2. Attach those labels to a VM internal structure.
3. Display those labels comma-separated in the "LABELS" field of the `limactl list` output (can even be hidden by default, so one can use `--format {{.Labels}}` to display them.

Later some filters can be implemented but not really necessary I'd say.. as long as those fields are available as an array in the JSON output one can use it instead.

I'd personally want this to be able to group my VMs without polluting the VM name with group name. Then I can improve my Puppet Bolt inventory plugin so it can auto-assign VMs to groups.

jay7x:
Implement ability to add custom user-defined labels to a VM (by editing the VM's `lima.yaml` at the moment). Labels are only available in `limactl list --format json` currently.

Closes #1304

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -45,6 +45,10 @@ memory: null
 # 🟢 Builtin default: "100GiB"
 disk: null
 
+# List of VM user-defined labels.
+# 🟢 Builtin default: []
+labels: null
+
 # Expose host directories to the guest, the mount point might be accessible from all UIDs in the guest
 # 🟢 Builtin default: null (Mount nothing)
 # 🔵 This file: Mount the home as read-only, /tmp/lima as writable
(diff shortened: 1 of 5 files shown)


jandubois (changes requested):
So far I find the justification for adding these changes not sufficient, so I don't think this should be merged: https://github.com/lima-vm/lima/issues/1304#issuecomment-1387692935

jandubois:
Closing this PR because there are no plans currently to merge it.

Further discussion should take place in #1309.

PR #1308, opened 2023-01-20
Enable silent log-level for hiding INFO messages

afbjorklund:
For a normal run, with downloads already cached, this will only output the message (if any)

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -55,8 +55,17 @@ func newApp() *cobra.Command {
 		SilenceUsage:  true,
 		SilenceErrors: true,
 	}
+	rootCmd.PersistentFlags().String("log-level", "", "Set the logging level [trace, debug, info, warn, error]")
 	rootCmd.PersistentFlags().Bool("debug", false, "debug mode")
 	rootCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
+		l, _ := cmd.Flags().GetString("log-level")
+		if l != "" {
+			lvl, err := logrus.ParseLevel(l)
+			if err != nil {
+				return err
+			}
+			logrus.SetLevel(lvl)
+		}
 		debug, _ := cmd.Flags().GetBool("debug")
 		if debug {
 			logrus.SetLevel(logrus.DebugLevel)


jandubois:
Thanks, LGTM, but can we discuss if "silent" shouldn't be the default, and `--info` would be an option to switch the info-level output back on?

@AkihiroSuda WDYT?

cmd/limactl/main.go
```diff
@@ -54,8 +54,13 @@ func newApp() *cobra.Command {
 		SilenceUsage:  true,
 		SilenceErrors: true,
 	}
+	rootCmd.PersistentFlags().Bool("silent", false, "silent mode")
```
AkihiroSuda, line 57: What about `--log-level  (debug|warn|info)`

afbjorklund: That would also work, as an alternative to `--debug` and `--silent` ?

Or maybe `--quiet` is a more common, but this one was _very_ quiet.

afbjorklund:
minikube saves all logging output to a file, _unless_ you give an explicit option: `--alsologtostderr`

For the regular progress output, it goes all in on emojis (unless you disable them). It divides the crowd.

cmd/limactl/main.go
jandubois: I'm fine with `--log-level`, and I would prefer `--log-level warn` to be the default instead of `info`.

jandubois:
> minikube saves all logging output to a file, _unless_ you give an explicit option: `--alsologtostderr`

I always found this a bit unconventional; but I can see the advantage for something that runs a little longer (like `limactl start`). It allows you to get the full log after the command failed without having to run it again.
  
> For the regular progress output, it goes all in on emojis (unless you disable them). It divides the crowd.

Personally I like the minikube emoji, but I understand that some people feel differently. I do find it challenging to pick meaningful emoji for some status messages though.

cmd/limactl/main.go
AkihiroSuda: Changing the default might be hard (and will result in losing the `message` in the YAML)

jandubois: > Changing the default might be hard

Why? Do you consider this an incompatible change, or would we actually be breaking something?

> (and will result in losing the message in the YAML)

No, it doesn't, just losing the "Message from the instance $INSTANCE" header, and we could write that directly to STDOUT too, if we want to:

```golang
	logrus.Infof("Message from the instance %q:", inst.Name)
	for scanner.Scan() {
		// Avoid prepending logrus "INFO" header, for ease of copypasting
		fmt.Println(scanner.Text())
	}
```

afbjorklund:
As noted elsewhere, it also needs a fix for the URL that it is downloading (by default **only** showing progress bar)

```
pkg/fileutils/download.go:      logrus.WithField("digest", f.Digest).Infof("Attempting to download %s from %q", description, f.Location)
pkg/fileutils/download.go-      res, err := downloader.Download(dest, f.Location,
pkg/fileutils/download.go-              downloader.WithCache(),
pkg/fileutils/download.go-              downloader.WithExpectedDigest(f.Digest),
pkg/fileutils/download.go-      )
--
pkg/start/start.go:             logrus.WithField("digest", f.Digest).Infof("Attempting to download the nerdctl archive from %q", f.Location)
pkg/start/start.go-             res, err := downloader.Download("", f.Location, downloader.WithCache(), downloader.WithExpectedDigest(f.Digest))
```

The suggestion is to just print the URL on a separate line, maybe log as debug (so that is not too redundant).

cmd/limactl/main.go
afbjorklund: I don't think we do (want to repeat the instance name again)

jandubois: Fine with me; I only included it to show how we can replicate the current behaviour:

```console
INFO[0090] READY. Run `limactl shell docker` to open the shell.
INFO[0090] Message from the instance "docker":
------
docker context create lima-docker --docker "host=unix:///Users/jan/.lima/docker/sock/docker.sock"
docker run hello-world
------
```

We probably still want to have some kind of header line before the actual message; but I don't care too much how it looks like. Maybe

```console
INFO[0090] Message from the instance template:
```

afbjorklund: Just meant that it might not be needed.

```
? Creating an instance "default" Open an editor to review or modify the current configuration
Starting "default"

READY.
This will be shown to the user.

```

jandubois: > Just meant that it might not be needed.

Ok

> And now that we have the command wrappers, you don't even need to configure anything anyway.

We should still have a message that tells the user about the wrapper.

And I think it is still useful to tell them about the docker context option as well because if they use other tools to issue `docker` commands (VS Code plugins?), they will not be aware of the `docker.lima` wrapper, but might do the right thing when the context is set.

afbjorklund:
Added `--log-level`

cmd/limactl/main.go
```diff
@@ -55,8 +55,22 @@ func newApp() *cobra.Command {
...
+		if l != "" {
+			lvl, err := logrus.ParseLevel(l)
+			if err != nil {
+				return err
+			}
+			logrus.SetLevel(lvl)
+		}
+		silent, _ := cmd.Flags().GetBool("silent")
+		if silent {
+			logrus.SetLevel(logrus.WarnLevel)
```
jbergstroem, line 72: In many unix utilities `silent` implies no output at all. Curl for instance wouldn't give you anything other than the output from the command.

cmd/limactl/main.go
jbergstroem: In many unix utilities `silent` implies no output at all. Curl for instance wouldn't give you anything other than the output from the command.

afbjorklund: I guess `--quiet` would have been an option as well

cmd/limactl/main.go
```diff
@@ -55,8 +55,22 @@ func newApp() *cobra.Command {
...
+		}
```
jandubois, line 73: I agree with @jbergstroem that `silent` should mean "no output", so the log level should be "Error" and not "Warn".

But if this option is just an alias for `--log-level error`, then I would rather not have it at all. If we have it as a distinct option, then it also needs to suppress of progress bars, and the message output.

If it was just an alias, then I would rather have separate `--debug`, `--info`, `--warn`, and `--error` flags.

jandubois (changes requested):

cmd/limactl/main.go
afbjorklund: It was just an alias, like the --debug.  I guess that --quiet would have been better.

afbjorklund:
Now `--quiet`

cmd/limactl/main.go
jandubois: > I guess that --quiet would have been better.

To me the issue is not (just) the name, but also that it is redundant. And it isn't even clear what it does: is `--quiet` the same as `--log-level warn` or `--log-level error`?

Or does it something else, and should be combined with `--log-level` because it really quiets down stuff that is not a log entry, like the download progress bars, or the message from `lima.yaml`?

> It was just an alias, like the --debug.

I would have objected to adding `--debug` if we already had `--log-level`, but now we want to keep it for backwards compatibility.

As I mentioned before, we should have either `--trace`, `--debug`, `--info`, `--warn`, and `--error` options, or `--log-level`, but not both. But even if we wanted to have both, why would the shortcut be called `--quiet` or `--silent` and not `--warn` or `--error` to make it clear what they do?

Just for the record, I prefer to have `--log-level` over individual options, if for no other reason than to keep the `--help` output tidy.

Having redundant options has a cost, especially to new users: they wonder what the differences are, and then have to look at the documentation. It also clutters up the `--help` output:

```Global Flags:
      --debug              debug mode
      --log-level string   Set the logging level [trace, debug, info, warn, error, fatal, panic]
      --silent             silent mode
```

We now have 3 options just to set a log level, when 1 would be sufficient.

So unless `--silent`/`--quiet` provides additional functionality, I'm opposed to adding it.

Maybe this is a separate issue, but I would not expect this output:

```console
$ limactl start --silent --tty=false template://docker > lima.log

31.66 MiB / 647.50 MiB (4.89%) ? p/s
68.80 MiB / 647.50 MiB (10.62%) 7.43 MiB/s
101.78 MiB / 647.50 MiB (15.72%) 7.37 MiB/s
139.12 MiB / 647.50 MiB (21.49%) 7.38 MiB/s
173.44 MiB / 647.50 MiB (26.79%) 7.35 MiB/s
209.23 MiB / 647.50 MiB (32.31%) 7.34 MiB/s
246.92 MiB / 647.50 MiB (38.13%) 7.35 MiB/s
282.62 MiB / 647.50 MiB (43.65%) 7.33 MiB/s
317.83 MiB / 647.50 MiB (49.09%) 7.32 MiB/s
355.36 MiB / 647.50 MiB (54.88%) 7.33 MiB/s
390.67 MiB / 647.50 MiB (60.34%) 7.31 MiB/s
426.27 MiB / 647.50 MiB (65.83%) 7.30 MiB/s
464.20 MiB / 647.50 MiB (71.69%) 7.32 MiB/s
499.06 MiB / 647.50 MiB (77.08%) 7.29 MiB/s
533.50 MiB / 647.50 MiB (82.39%) 7.27 MiB/s
570.41 MiB / 647.50 MiB (88.09%) 7.28 MiB/s
605.97 MiB / 647.50 MiB (93.59%) 7.27 MiB/s
643.31 MiB / 647.50 MiB (99.35%) 7.28 MiB/s
647.50 MiB / 647.50 MiB (100.00%) 7.56 MiB/s
```

Or without redirecting STDOUT:

```console
$ limactl start --silent --tty=false template://docker
647.50 MiB / 647.50 MiB [-----------------------------------] 100.00% 7.16 MiB/s
------
docker context create lima-docker --docker "host=unix:///Users/jan/.lima/docker/sock/docker.sock"
docker run hello-world
------
```

afbjorklund:
I removed the short forms, so now one needs to use the long way of hiding INFO messages:

`--log-level WARN`

The old debug flag remains, it could (potentially) do more things than just change log level ?

`--log-level DEBUG`

cmd/limactl/main.go
afbjorklund: Now there is only a single (new) option, hopefully it should be less surprising ?

```
      --log-level string   Set the logging level [trace, debug, info, warn, error]
```

AkihiroSuda (approved):

jandubois (approved):
Thanks, LGTM

PR #1315, opened 2023-01-22
fix unclean stop for vz driver

### LINKED ISSUE #1313: Running 'apt upgrade' breaks the VM
### Description

I've tried running the `docker-rootful` example with a VZ VM, but I've noticed that a simple `apt update` and `apt upgrade` completely breaks the VM: the boot process gets stuck at `Waiting for the essential requirement 1 of 3: "ssh"` with the CPU usage pinned at 100% (`ssh` probably broke during the upgrade).

If this is the expected behaviour, I think the user should be at least warned or, even better, such an action should be blocked.

balajiv113:
fixes #1313

**Why this change ?**
lima hostagent is terminating before the vm is fully stopped. Due to this there were disk corruptions.

This PR adds support for waiting for the state change to complete before closing hostagent

### DIFF
--- a/pkg/vz/vz_driver_darwin.go
+++ b/pkg/vz/vz_driver_darwin.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"time"
 
 	"github.com/lima-vm/lima/pkg/reflectutil"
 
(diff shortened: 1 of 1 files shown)


pkg/vz/vz_driver_darwin.go
```diff
@@ -121,10 +122,27 @@ func (l *LimaVzDriver) Start(ctx context.Context) (chan error, error) {
...
+			case <-tick:
+				if l.machine.State() == vz.VirtualMachineStateStopped {
+					return nil
+				}
+			}
+		}
 	}
 
-	return nil
+	return errors.New("VZ CanRequestStop is not supported")
```
AkihiroSuda, line 147:
```suggestion
	return errors.New("vz: CanRequestStop is not supported")
```

AkihiroSuda (dismissed):
Thank you!

I guess we should make the next release very soon

pkg/vz/vz_driver_darwin.go
balajiv113: Done

balajiv113:
> next release very soon

yes, i believe. This might even be the cause for factory-reset required for some cases.

AkihiroSuda (approved):
Thanks

PR #1317, opened 2023-01-22
[POC] USB sharing with host

sheerun:
I managed to flash USB from lima, but I don't feel this is ready to ship. I don't have time to handle this anymore, but I hope it will help. It only works if instance is started with `sudo limactl start` (otherwise qemu is not working with admin privileges which it needs for device forwarding), and lima is started with `sudo lima` (otherwise you get `Bad port '0'` error). USB attaching / unattaching when physical device is plugged / unplugged seems to be working and I feel it will continue when buses and ports are uses instead of productid / vendorid (see below).

Main issues to solve:
- USB needs to be inserted when VM is started. This should be solved by binding all buses/ports instead using productid/vendorid of devices, like explained here: https://unix.stackexchange.com/questions/452934/can-i-pass-through-a-usb-port-via-qemu-command-line
- QEMU vm needs to be started with sudo which is problematic because currently  background agent is running with user privileges. It can be solved by asking for password and increasing privileges, but I'm not sure lima is willing to make such drastic changes to current workflow. I guess it would help if machine was started from foreground process, not background process because then just `sudo -E` can be prepended to qemu process (maybe with -u and -g flags set to current user).

Mind issue is with second point and it would require either managing everithing without agent process or making agent process running as root, or having 2 agent processes.

Hope it helps!

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -66,9 +65,9 @@ func newApp() *cobra.Command {
 			formatter.ForceColors = true
 			logrus.StandardLogger().SetFormatter(formatter)
 		}
-		if os.Geteuid() == 0 {
-			return errors.New("must not run as the root")
-		}
+		// if os.Geteuid() == 0 {
+		// 	return errors.New("must not run as the root")
+		// }
 		// Make sure either $HOME or $LIMA_HOME is defined, so we don't need
 		// to check for errors later
 		if _, err := dirnames.LimaDir(); err != nil {
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -392,6 +392,10 @@ hostResolver:
 # - 1.1.1.1
 # - 1.0.0.1
 
+# Privileged mode enables some features that are not availble in normal mode, like USB sharing
+# in such mode "limectl start" will ask for root password
+privileged: false
+
 # ===================================================================== #
 # GLOBAL DEFAULTS AND OVERRIDES
 # ===================================================================== #
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/coreos/go-semver/semver"
 	"github.com/docker/go-units"
+	"github.com/google/gousb"
 	"github.com/lima-vm/lima/pkg/fileutils"
 	"github.com/lima-vm/lima/pkg/iso9660util"
 	"github.com/lima-vm/lima/pkg/limayaml"
@@ -468,6 +469,16 @@ func Cmdline(cfg Config) (string, []string, error) {
 		// TODO: consider using virtio cdrom for all the architectures
 		args = append(args, "-cdrom", filepath.Join(cfg.InstanceDir, filenames.CIDataISO))
 	}
+
+	// USB
+	ctx := gousb.NewContext()
+	defer ctx.Close()
+
+	ctx.OpenDevices(func(desc *gousb.DeviceDesc) bool {
+		args = append(args, "-device", fmt.Sprintf("nec-usb-xhci,id=usb-%s-%s", desc.Vendor, desc.Product))
+		args = append(args, "-device", fmt.Sprintf("usb-host,bus=usb-%s-%s.0,vendorid=0x%s,productid=0x%s", desc.Vendor, desc.Product, desc.Vendor, desc.Product))
+		return false
+	})
 
 	// Kernel
 	kernel := filepath.Join(cfg.InstanceDir, filenames.Kernel)
--- a/go.mod
+++ b/go.mod
@@ -48,6 +48,7 @@ require (
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/google/btree v1.0.1 // indirect
 	github.com/google/gopacket v1.1.19 // indirect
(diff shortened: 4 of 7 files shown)


AkihiroSuda:
Is this for Linux host, or does this work on macOS too? 👀

sheerun:
I don't know if it works on linux. I've tested it on Mac M1

AkihiroSuda:
> Mind issue is with second point and it would require either managing everithing without agent process or making agent process running as root, or having 2 agent processes, one running as root handling root vms and one in user mode handling user mode qms



The latter mode is preferable. We have been doing the same for `socket_vmnet`.

dan-corneanu:
Is adding USB pass through more likely to happen with the new `--vm-type=vz`?

AkihiroSuda:
> Is adding USB pass through more likely to happen with the new `--vm-type=vz`?

Unlikely.
vz doesn't seem to have an API for USB pass through: https://developer.apple.com/documentation/virtualization?language=objc

dan-corneanu:
I have tested a linux guest VM running with UTM (https://mac.getutm.app/) on M1, using qemu. UTM's UI has a button to select what USB devices you want to be passed through to the VM. It works as expected. I can select/unselect USB devices and they correctly show up or get removed from the guest VM while the VM is running.

Any idea how is this implemented?

retpolanne:
@dan-corneanu I guess it's implemented like this [1][2]

[1] https://qemu-project.gitlab.io/qemu/system/devices/usb.html
[2] https://github.com/utmapp/UTM/blob/62bd84c78c045870c580df34c1ba233d75e02995/Configuration/UTMQemuConfiguration%2BArguments.swift#L661

marcofranssen:
I would love to get this feature working. My usecase would be to use it with the ESPHome device builder running in a container to flash my ESP32 devices. 

What are the bottlenecks that have to be tackled to get this PR in?

P.S. I'm on a Macbook M3 Pro… Using Colima latest released version on Homebrew.

pkg/qemu/qemu.go
```diff
@@ -469,6 +470,16 @@ func Cmdline(cfg Config) (string, []string, error) {
 		args = append(args, "-cdrom", filepath.Join(cfg.InstanceDir, filenames.CIDataISO))
 	}
 
+	// USB
```
AkihiroSuda, line 473: The YAML config should have a property for USB
```yaml
usb:
  enabled: true
```

examples/default.yaml
```diff
@@ -392,6 +392,10 @@ hostResolver:
 # - 1.1.1.1
 # - 1.0.0.1
 
+# Privileged mode enables some features that are not availble in normal mode, like USB sharing
+# in such mode "limectl start" will ask for root password
+privileged: false
```
AkihiroSuda, line 397: ` "limectl start" will ask for root password` doesn't seem implemented

examples/default.yaml
```diff
@@ -392,6 +392,10 @@ hostResolver:
 # - 1.1.1.1
 # - 1.0.0.1
 
+# Privileged mode enables some features that are not availble in normal mode, like USB sharing
+# in such mode "limectl start" will ask for root password
```
AkihiroSuda, line 396: How will this affect the permission of the mounts (9p, reverse-sshfs) ?

cmd/limactl/main.go
```diff
@@ -66,9 +65,9 @@ func newApp() *cobra.Command {
 			formatter.ForceColors = true
 			logrus.StandardLogger().SetFormatter(formatter)
 		}
-		if os.Geteuid() == 0 {
-			return errors.New("must not run as the root")
-		}
+		// if os.Geteuid() == 0 {
+		// 	return errors.New("must not run as the root")
+		// }
```
AkihiroSuda, line 70: By default, root should be still prohibited

examples/default.yaml
```diff
@@ -392,6 +392,10 @@ hostResolver:
 # - 1.1.1.1
 # - 1.0.0.1
 
+# Privileged mode enables some features that are not availble in normal mode, like USB sharing
```
AkihiroSuda, line 395: Should be marked as experimental https://lima-vm.io/docs/releases/experimental/

cmd/limactl/main.go
AkihiroSuda, line 1: Needs rebase (not just this file)

AkihiroSuda:
Can we use this?
- https://github.com/lima-vm/lima/issues/2224#issuecomment-2110411523

```
It does run the `usbredirect` daemon as root, one per device and this kinda works.
```

cc @mschoenlaub

AkihiroSuda:
> What are the bottlenecks that have to be tackled to get this PR in?

The main bottleneck is that this contradicts Lima's fundamental design that prohibits running VM as the root

CamJN:
> > Is adding USB pass through more likely to happen with the new `--vm-type=vz`?
> 
> Unlikely. vz doesn't seem to have an API for USB pass through: https://developer.apple.com/documentation/virtualization?language=objc

Seems this has been added:
https://developer.apple.com/documentation/virtualization/usb-devices?language=objc

AkihiroSuda:
Saw this project implements USB passthrough for Lima by sending QMP commands to `~/.lima/<INSTANCE>/qmp.sock`
https://github.com/PentHertz/RF-Swift/commit/4d06fb9b026fd84c83115dc45b4a4a7d95ec4381

https://github.com/PentHertz/RF-Swift/blob/v2.2.0/go/rfswift/rfutils/hostcli_mac.go#L221-L248

PR #1326, opened 2023-01-24
Write `~/.lima/<INSTANCE>/ssh.config`

AkihiroSuda:
```console
$ limactl ls --format='{{.SSHConfigFile}}' default
/Users/example/.lima/default/ssh.config
```

### DIFF
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -54,6 +54,7 @@ VZ:
 
 SSH:
 - `ssh.sock`: SSH control master socket
+- `ssh.config`: SSH config file for `ssh -F`. Not consumed by Lima itself.
 
 Guest agent:
 - `ga.sock`: Forwarded to `/run/lima-guestagent.sock` in the guest, via SSH
--- a/README.md
+++ b/README.md
@@ -225,6 +225,20 @@ Use `<INSTANCE>:<FILENAME>` to specify a source or target inside an instance.
 
 `limactl disk list`: list all existing disks
 
+#### `limactl show-ssh`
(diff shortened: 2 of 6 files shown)


docs/internal.md
```diff
@@ -54,6 +54,7 @@ VZ:
 
 SSH:
 - `ssh.sock`: SSH control master socket
+- `ssh.config`: SSH config file, printed by `limactl show-ssh --format=config-file` . Not consumed by Lima itself.
```
jandubois, line 57: I was confused by this on first reading: it sounds like the file is written only when you run `limactl show-ssh --format=config-file`. But it is actually written ("printed") by the host agent during startup, and this command just prints the name of the config file.

jandubois:
I understand why you added this to the `show-ssh` command, but it feels somewhat off to me. All other format options show the actual options, whereas this one shows a filename.

Also, to show the name of the identity file you run[^1]:

[^1]: This is actually broken right now in `master`, but was still working in 0.14.2; will file a bug for it after looking some more (e.g. `LimaHome` is also broken; probably all "global" fields).

```console
$ limactl ls --format '{{.IdentityFile}}' default
/Users/jan/.lima/_config/user
```

But to get the name of the config file you run:

```console
$ limactl show-ssh --format=config-file default
/Users/jan/.lima/default/ssh.config
```

This seems rather inconsistent to me. But I don't have a suggestion how to improve this right now.

AkihiroSuda:
Updated
```console
$ limactl ls --format='{{.SSHConfigFile}}' default
/Users/example/.lima/default/ssh.config
```

jandubois (approved):
Thanks, LGTM

PR #1335, opened 2023-01-28
improvement(hostagent): link ssh agent socket to predetermined location

ryancurrah:
This PR resolves the issue https://github.com/rancher-sandbox/rancher-desktop/issues/3042 and relates to the comment https://github.com/rancher-sandbox/rancher-desktop/issues/3488. 

This change ensures the ssh agent socket can be mounted into containers without having to determine the location first. The location will be in a static location /run/host-services/ssh-auth.sock. This is the same location as Docker Desktop uses.

Signed-off-by: Ryan Currah <ryan@currah.ca>

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -121,6 +121,9 @@ ssh:
   # 🟢 Builtin default: true
   loadDotSSHPubKeys: null
   # Forward ssh agent into the instance.
+  # The ssh agent socket can be mounted in a container at the path `/run/host-services/ssh-auth.sock`.
+  # Set the environment variable `SSH_AUTH_SOCK` value to the path above.
+  # The socket is accessible by the non-root user inside the Lima instance.
   # 🟢 Builtin default: false
   forwardAgent: null
   # Forward X11 into the instance
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -317,6 +317,19 @@ func (a *HostAgent) startHostAgentRoutines(ctx context.Context) error {
 	if err := a.waitForRequirements(ctx, "essential", a.essentialRequirements()); err != nil {
 		mErr = multierror.Append(mErr, err)
 	}
+	if *a.y.SSH.ForwardAgent {
+		faScript := `#!/bin/bash
+set -eux -o pipefail
+sudo mkdir -p -m 700 /run/host-services
+sudo ln -sf "${SSH_AUTH_SOCK}" /run/host-services/ssh-auth.sock
+sudo chown -R "${USER}" /run/host-services`
+		faDesc := "linking ssh auth socket to static location /run/host-services/ssh-auth.sock"
+		stdout, stderr, err := ssh.ExecuteScript("127.0.0.1", a.sshLocalPort, a.sshConfig, faScript, faDesc)
+		logrus.Debugf("stdout=%q, stderr=%q, err=%v", stdout, stderr, err)
+		if err != nil {
+			mErr = multierror.Append(mErr, fmt.Errorf("stdout=%q, stderr=%q: %w", stdout, stderr, err))
+		}
+	}
 	if *a.y.MountType == limayaml.REVSSHFS {
 		mounts, err := a.setupMounts(ctx)
 		if err != nil {


jandubois:
I'm wondering about permission. I suspect it doesn't make a difference in the Lima context, but in general the permissions of the `/tmp/ssh-XXXXnnnn` directory are `700`, so no other user has access to the socket within.

I'm not sure if changing the permissions of the socket are a good idea, so maybe we should also create a directory with matching permissions, and then create the link in there?

I also don't think we need to put `lima` in the name, as this is just an alias for the regular ssh agent socket. My suggestions for the bike shed:

```
/tmp/SSH_AUTH_SOCK/agent.sock
/tmp/ssh-auth/agent.sock
```

jandubois:
> We could do this or auto-mount the agent socket into containers when `forwardAgent` is `true`.

I'm not sure how you would actually do this, but even if we could, I think it is a bad idea; the user should control which containers get access to the auth socket, as this is a security issue. Only trusted containers that need access should have it mounted.

ryancurrah:
Good call out I was reading about ssh sock a little more after I created this pull request and the solution as I have it now could be a security issue for the same reasons you mentioned. I'll make the recommended changes and also see if there is anything else we need to be cautious of.

ryancurrah:
Done and here are the permissions for comparison. Also the man page mentions the sockets should be removed when the agent exits. I assume our socket link won't work once the SSH session exits but maybe we could remove it when the VM is shutdown, might be overkill though.

```bash
╰─ _output/bin/limactl shell default

ryancurrah@lima-default:/Users/ryancurrah/git/github.com/ryancurrah/lima$ ls -l /tmp/
total 24
drwx------ 2 ryancurrah ryancurrah 4096 Jan 29 03:35 ssh-XXXX7nR2Xe
drwx------ 2 ryancurrah ryancurrah 4096 Jan 29 03:35 ssh-auth

ryancurrah@lima-default:/Users/ryancurrah/git/github.com/ryancurrah/lima$ ls -l /tmp/ssh-XXXX7nR2Xe/
srwxrwxr-x 2 ryancurrah ryancurrah 0 Jan 29 03:35 agent.2322

ryancurrah@lima-default:/Users/ryancurrah/git/github.com/ryancurrah/lima$ ls -l /tmp/ssh-auth/
srwxrwxr-x 2 ryancurrah ryancurrah 0 Jan 29 03:35 agent.sock
```

And it works inside a container...

```bash
╰─ lima nerdctl run -it -v /tmp/ssh-auth/agent.sock:/tmp/ssh-auth/agent.sock -e SSH_AUTH_SOCK=/tmp/ssh-auth/agent.sock golang:latest bash
root@5bd9ef65b5c9:/go# ssh-add -l
4096 SHA256:kx6jAhYN4RGheucjhe259FsJ2C7Mtgcj1BrJKiy0ivA ryan@currah.ca (RSA)
```

`ssh-agent` man page

```
ENVIRONMENT
     SSH_AUTH_SOCK  When ssh-agent starts, it creates a UNIX-domain socket and stores its pathname in this variable.  It is accessible only to the
                    current user, but is easily abused by root or another instance of the same user.

FILES
     $TMPDIR/ssh-XXXXXXXXXX/agent.<ppid>
             UNIX-domain sockets used to contain the connection to the authentication agent.  These sockets should only be readable by the owner.
             The sockets should get automatically removed when the agent exits.
```

pkg/hostagent/requirements.go
```diff
@@ -132,6 +132,20 @@ Also see "/var/log/cloud-init-output.log" in the guest.
 A possible workaround is to run "lima-guestagent install-systemd" in the guest.
 `,
 	})
+	if *a.y.SSH.ForwardAgent {
+		req = append(req, requirement{
+			description: "linking ssh auth socket to /tmp/ssh-auth/agent.sock",
+			script: `#!/bin/bash
+set -eux -o pipefail
+if ! timeout 30s bash -c "mkdir -p -m 700 /tmp/ssh-auth && ln -f $SSH_AUTH_SOCK /tmp/ssh-auth/agent.sock"; then
```
jandubois, line 140: I haven't tested it, but I think you need to remove the old socket before linking it again, in case `/tmp` has not been cleared by the reboot:

```suggestion
if ! timeout 30s bash -c "mkdir -p -m 700 /tmp/ssh-auth && rm -f /tmp/ssh-auth/agent.sock && ln -f $SSH_AUTH_SOCK /tmp/ssh-auth/agent.sock"; then
```

Otherwise I would expect you to get a `ln: /tmp/ssh-auth/agent.sock: File exists` error.

ryancurrah: The `-f` argument removes any existing files.

```
-f, --force
              remove existing destination files
```

pkg/hostagent/requirements.go
```diff
@@ -132,6 +132,20 @@ Also see "/var/log/cloud-init-output.log" in the guest.
...
+	echo >&2 "unable to link $SSH_AUTH_SOCK to /tmp/ssh-auth/agent.sock"
+	exit 1
+fi
+`,
+			debugHint: `Linking $SSH_AUTH_SOCK to /tmp/ssh-auth/agent.sock failed.
+Ensure the SSH_AUTH_SOCK environment variable exists and is not empty.`,
```
jandubois, line 146: This is not really helpful, as the user wouldn't know how to create the variable; it is provided by `ssh`, and if it isn't created, I'm not sure how the user can fix it.

In practice I think this will never happen; when the ssh connection succeeds, and agent forwarding has been enabled, then ssh will create the variable.

pkg/hostagent/requirements.go
ryancurrah: I've tried stopping the VM and and starting it again and it works.

pkg/hostagent/requirements.go
ryancurrah: Yeah I'm not really sure what helpful advice to put here other than it failed

jandubois: > Yeah I'm not really sure what helpful advice to put here other than it failed

I think that is enough. As I said above, I expect this to be essentially unreachable code.

jandubois (approved):
I'm pretty much fine with the PR; not sure why it is still in Draft mode. I'll wait for feedback from other maintainers before merging though.

pkg/hostagent/requirements.go
```diff
@@ -132,6 +132,19 @@ Also see "/var/log/cloud-init-output.log" in the guest.
 A possible workaround is to run "lima-guestagent install-systemd" in the guest.
 `,
 	})
+	if *a.y.SSH.ForwardAgent {
```
ryancurrah, line 135: I'm not sure if this should be an `essentialRequirements` or an `optionalRequirements `.

jandubois:
Please squash commits before marking it as ready for review!

pkg/hostagent/requirements.go
ryancurrah: Ok removed the unhelpful messaging.

ryancurrah:
> I'm pretty much fine with the PR; not sure why it is still in Draft mode. I'll wait for feedback from other maintainers before merging though.

It's in draft mode because I was not 100% sure this was going to be the right solution. I'll put it in Ready For Review.

pkg/hostagent/requirements.go
jandubois: Ah, good point. I think this should go into `optionalRequirements` instead.

The essential requirements are things that are needed for Lima itself to work properly (e.g. `sshfs` may need to be installed for the mounts to work). Everything that is only needed for user operations should be in "optional".

ryancurrah: Moved to `optionalRequirements`.

jandubois (approved):
Thanks, LGTM!

Waiting for feedback from other @lima-vm/maintainers before merging.

pkg/hostagent/requirements.go
```diff
@@ -170,6 +170,19 @@ Also see "/var/log/cloud-init-output.log" in the guest.
...
+			script: `#!/bin/bash
+set -eux -o pipefail
+if ! timeout 30s bash -c "mkdir -p -m 700 /tmp/ssh-auth && ln -f $SSH_AUTH_SOCK /tmp/ssh-auth/agent.sock"; then
+	echo >&2 "unable to link $SSH_AUTH_SOCK to /tmp/ssh-auth/agent.sock"
+	exit 1
+fi
+`,
+			debugHint: `Linking $SSH_AUTH_SOCK to /tmp/ssh-auth/agent.sock failed.`,
+		})
+	}
```
AkihiroSuda, line 185: I think this should be in `$XDG_RUNTIME_DIR/lima`

jandubois: The whole point of this PR is to have a predictable address, that you can use in scripts, so you don't have to run `$(lima printenv SSH_AUTH_SOCK)` to get the filename.

Putting the link into another location that can only be determined by shelling into the VM makes the whole PR pointless IMO.

jandubois: However, I just noticed that this PR should include documentation (assuming we can agree on a fixed location).

Just for reference, ssh also puts the agent socket into `/tmp`:

```console
$ lima printenv SSH_AUTH_SOCK
/tmp/ssh-XXXXv2u4Qd/agent.2043
```

So linking it to `/tmp/ssh-auth/agent.sock` doesn't feel wrong to me.

AkihiroSuda: $XDG_RUNTIME_DIR/lima (/run/user/$UID/lima) is still predictable when you can get the UID

AkihiroSuda: Generally, creating a dir under /tmp with a predictable name is known as a TOCTOU attack vector for multi-user system.

Lima isn't really expected to be used for multi-user VM, but I'm little a bit on the fence to adopt that vulnerable-ish pattern.

AkihiroSuda: Using /run/lima is ok for me though

jandubois: `XDG_RUNTIME_DIR` being the same as `/run/user/$UID` seems based on common usage and does not seem to be guaranteed by the spec.

`/run/lima` is likely on a different filesystem, so we would have to use a symlink instead of a hard link (probably not a problem though).

However, I don't see any TOCTOU attack that doesn't already apply to the original socket location. If `/tmp/ssh-XXXXv2u4Qd/agent.2043` is secure, then `/tmp/ssh-auth/agent.sock` should be as well.

As for the `/run/lima` name, I would prefer a name that doesn't include `lima` because I may want to use the same location in the future with WSL2, where using `/run/lima` would be confusing.

jandubois: So theoretically I see a vulnerability because we don't check the ownership of the directory before creating the link in there. The same vulnerability would however apply to any other file system location since we are using a predictable name.

But practically speaking, given that we allow password-less `sudo`, none of this is preventing a targeted attack.

We are just trying to avoid accidental exposure to workloads users are running inside the VM, ideally in containers.

jandubois: Avoiding TOCTOU:

```console
jan@lima-default:/tmp$ ln -s -f $(dirname $SSH_AUTH_SOCK) /tmp/ssh-auth
jan@lima-default:/tmp$ ln -s -f $SSH_AUTH_SOCK /tmp/ssh-auth/agent.sock
jan@lima-default:/tmp$ ls -l /tmp/ssh-auth/agent.sock
lrwxrwxrwx 1 jan jan 30 Jan 29 06:10 /tmp/ssh-auth/agent.sock -> /tmp/ssh-XXXXv2u4Qd/agent.2043
```

But in this case having the additional directory level is pointless, so we could go back to:

```console
jan@lima-default:/tmp$ ln -s -f $SSH_AUTH_SOCK /tmp/SSH_AUTH_SOCK
jan@lima-default:/tmp$ ls -l /tmp/SSH_AUTH_SOCK
lrwxrwxrwx 1 jan jan 30 Jan 29 06:12 /tmp/SSH_AUTH_SOCK -> /tmp/ssh-XXXXv2u4Qd/agent.2043
```

jandubois: FWIW, using a symlink was my original suggestion in https://github.com/rancher-sandbox/rancher-desktop/issues/3488#issuecomment-1406936969. It avoids all the permissions problems because the original permissions of the target still apply.

pkg/hostagent/requirements.go
```diff
@@ -170,6 +170,19 @@ Also see "/var/log/cloud-init-output.log" in the guest.
 `,
 			})
 	}
+	if *a.y.SSH.ForwardAgent {
+		req = append(req, requirement{
```
AkihiroSuda, line 174: The `requirement` script are designed for validation, not for mutation of the VM.
So this should be rather moved to the (custom) provision script.
(And then we don't need to care about the file path, as it is customizable in the script?)

ryancurrah: I'm not sure where that is but I will look. I'm hoping to make this improvement for all users of forwardAgent so they don't have to do the hack mentioned in the issue. Similar to Docker Desktop where the agent sock was in a static location. Was one issue we ran into when migrating from Docker Desktop.

jandubois: > So this should be rather moved to the (custom) provision script.

Yes, but it would require something like #287 to be implemented first to run the script after the requirements scripts are done. The current provisioning scripts run before the host agent has established the ssh connection, so `SSH_AUTH_SOCK` is not yet available.

jandubois: > Similar to Docker Desktop where the agent sock was in a static location.

I didn't realize Docker Desktop did this. It uses `/run/host-services/ssh-auth`, so I would propose to use the same location for lima, if we are going to agree on a way to implement this.

AkihiroSuda: Using `/run/host-services/ssh-auth` is ok for me, but this has to be moved to the provision script or the boot script, as the `requirements` are designed for validation scripts.

jandubois: I understand, but it means we need to implement an additional type of provisioning script that will be executed by the host agent once the SSH connection has been setup (which is essentially #287, except that one also asks for the sshfs mounts to be done).

I'll leave it to @ryancurrah to decide if he wants to help with implementing this. Otherwise I'm fine with implementing it just for Rancher Desktop for now (simply by running `lima ln ...` once the VM has started, which will be a trivial change).

jandubois: > this has to be moved to the provision script or the boot script

@AkihiroSuda Are you opposed to implementing is directly in the hostagent code, either before or after setting up the sshfs mounts? That would be a lot simpler, as it doesn't need to implement a new type of provisioning script first.

AkihiroSuda: Hostagent is ok, but not in requirements.go

ryancurrah: Ack I will look at doing this in hostagent code.

ryancurrah: Started looking at this. So if we are trying to be compatible with Docker ssh-agent this is how it is laid out on the Docker VM.

```bash
╰─ docker run -it --rm --privileged --pid=host justincormack/nsenter1
bash-5.0# ls -l /run | grep host-
drwxr-xr-x    2 root     root           300 Jan 30 20:00 host-services

bash-5.0# ls -l /run/host-services/
srwxr-xr-x    1 root     root             0 Jan 30 20:00 backend.native.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 backend.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 dns.internal.grpc.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 dns.system.grpc.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 docker.proxy.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 extension-manager.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 filesystem.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 hubproxy.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 ntp.udp.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 socks.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 ssh-auth.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 vpnkit-data.sock
srwxr-xr-x    1 root     root             0 Jan 30 20:00 vpnkit.sock
```

ryancurrah: To create a folder and link in `/run` I will need to use sudo.

ryancurrah: So my changes work when tested locally, but not sure if this is ideal. 

Three concerns from me right now:
- Having to use `sudo` to create the directory.
- The openness of the permissions to read from the socket.
- The area within `hostagent.go` where I put the code.

The verification steps are below.

Verify socket got created on Lima VM.

```
╰─ _output/bin/limactl shell default

ryancurrah@lima-default:/Users/ryancurrah/git/github.com/ryancurrah/lima$ ls -l /run/ | grep host-
drwxr-xr-x  2 root root   60 Jan 31 02:28 host-services

ryancurrah@lima-default:/Users/ryancurrah/git/github.com/ryancurrah/lima$ ls -l /run/host-services/
lrwxrwxrwx 1 root root 30 Jan 31 02:28 ssh-auth.sock -> /tmp/ssh-XXXX5CLqMY/agent.2334
```

Verify ssh-agent works in a container.

```
╰─ lima nerdctl run -it -v /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock -e SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock golang:latest bash

root@03723ef10743:/go# ssh-add -l
4096 SHA256:......KOd0ivA ryan@currah.ca (RSA)
```

jandubois: > * Having to use `sudo` to create the directory and link.

I think this is fine.

> * The openness of the permissions to read from the socket.

These are the permissions on the symlink. The user still needs to satisfy the permissions of the symlink target. So I think this is fine too.

> * The area within `hostagent.go` where I put the code.

I think it should go directly before setting up the `sshfs` mounts.

jandubois: > These are the permissions on the symlink. 

Well, they will be in our case. For Docker Desktop these seem to be the sockets directly. But they probably don't care because unlike Lima they don't expect users to run **anything** directly on the VM, but only in the containers.

ryancurrah: Moved the code and confirmed it still works. I think I moved it to the right place.

pkg/hostagent/hostagent.go
```diff
@@ -317,6 +317,18 @@ func (a *HostAgent) startHostAgentRoutines(ctx context.Context) error {
 	if err := a.waitForRequirements(ctx, "essential", a.essentialRequirements()); err != nil {
 		mErr = multierror.Append(mErr, err)
 	}
+	if *a.y.SSH.ForwardAgent {
+		faScript := `#!/bin/bash
+set -eux -o pipefail
```
AkihiroSuda, line 322: Does this work for Alpine?

AkihiroSuda (approved):
Thanks

pkg/hostagent/hostagent.go
ryancurrah: Yes, it appears to work with this example.

```
# This example requires Lima v0.7.0 or later.
images:
- location: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.26/alpine-lima-std-3.17.0-x86_64.iso"
  arch: "x86_64"
  digest: "sha512:93bf8d52ad2b3a6ef401a1a565f67e4f06f8022ec8963cad36cac4a25253f056ac0b755d1ed56b8993b0261440e0d5d38bad15b271864ed33cd3b02272738672"
- location: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.26/alpine-lima-std-3.17.0-aarch64.iso"
  arch: "aarch64"
  digest: "sha512:aacd3a9e7a3231553bcee879c6770b379afbfad3dcfac7f274a80d6c490faf33f3c23b6130efa881f57c0c5235ced2877e2f45e3cdf4fefff11531d6023ae214"

mounts:
- location: "~"
- location: "/tmp/lima"
  writable: true

# The built-in containerd installer does not support Alpine currently.
# Hint: use the "rd" ISO instead of the "std" ISO to enable containerd: https://github.com/lima-vm/alpine-lima/releases/
containerd:
  system: false
  user: false

ssh:
  forwardAgent: true
```

Validation.

```
╰─ _output/bin/limactl shell default

lima-default:/Users/ryancurrah/git/github.com/ryancurrah/lima$ ls -l /run/host-services/ssh-auth.sock
lrwxrwxrwx    1 root     root            30 Feb  2 01:53 /run/host-services/ssh-auth.sock -> /tmp/ssh-XXXXpKBFdf/agent.2615

lima-default:/Users/ryancurrah/git/github.com/ryancurrah/lima$ ssh-add -l
4096 SHA256:....Od0ivA ryan@currah.ca (RSA)
```

ryancurrah:
> Thanks

You're welcome. Is there any documentation I should add? If so, where.

pkg/hostagent/hostagent.go
jandubois: > Does this work for Alpine?

Yes, it does; we use it in the boot scripts too, e.g. https://github.com/lima-vm/lima/blob/f7e7addab557da560da7146a829c32c59e939a33/pkg/cidata/cidata.TEMPLATE.d/boot/05-lima-disks.sh#L3

jandubois:
> Is there any documentation I should add? If so, where.

I think this should at least be mentioned in `default.yaml` where the agent forwarding key is defined: https://github.com/lima-vm/lima/blob/f7e7addab557da560da7146a829c32c59e939a33/examples/default.yaml#L123-L125

ryancurrah:
> I think this should at least be mentioned in `default.yaml` where the agent forwarding key is defined:

Done.

examples/default.yaml
```diff
@@ -121,6 +121,8 @@ ssh:
   # 🟢 Builtin default: true
   loadDotSSHPubKeys: null
   # Forward ssh agent into the instance.
+  # The ssh agent socket can be mounted in a container at the path `/run/host-services/ssh-auth.sock`.
```
AkihiroSuda, line 124: Permission should be explained too

pkg/hostagent/hostagent.go
```diff
@@ -317,6 +317,18 @@ func (a *HostAgent) startHostAgentRoutines(ctx context.Context) error {
...
+sudo mkdir -p -m 755 /run/host-services
```
AkihiroSuda, line 323: Should this be chowned to the main user with chmod 0700 ?

examples/default.yaml
ryancurrah: What should I mention about the permission?

pkg/hostagent/hostagent.go
jandubois: I don't know; as long as the content are just symlinks, the real access control is done by the permission of the symlink targets. The symlinks themselves are just labels.

ryancurrah: We could...

```
sudo mkdir -p -m 700 /run/host-services
sudo ln -sf "${SSH_AUTH_SOCK}" /run/host-services/ssh-auth.sock
sudo chown -R "${USER}" /run/host-services
```

jandubois: We could indeed. I don't see the point, but it doesn't hurt either, so 🤷

ryancurrah: Done.

examples/default.yaml
AkihiroSuda: "The socket is accessible by the non-root user" might be ok

ryancurrah: I'm not sure that matters in the container.

```
╰─ lima nerdctl run -it -v /run/host-services/ssh-auth.sock:/run/host-services/ssh-auth.sock -e SSH_AUTH_SOCK=/run/host-services/ssh-auth.sock golang:latest bash
root@7730a73568de:/go#
root@7730a73568de:/go# ls -l /run/host-services/ssh-auth.sock
srwxrwxr-x 1 root root 0 Feb  2 02:31 /run/host-services/ssh-auth.sock
```

ryancurrah: Oh, I guess you mean from inside the VM.

ryancurrah: Done.

AkihiroSuda (approved):
Thanks

jandubois (approved):
Thanks!

Still feels weird to me to have a single entry under `/run` not owned by `root`, but I guess it doesn't really matter.

cyberw:
Sorry to hijack this ticket, but is there any documentation on how to use this? 

A list of the steps needed to give access from MacOS ssh-agent to a container run using rancher-desktop I mean...

PR #1339, opened 2023-01-29
Add global fields to instance data before executing template

jandubois:
This functionality was lost in 75594fa:

```console
$ limactl ls -f '{{.LimaHome}}' default
FATA[0000] template: format:1:2: executing "format" at <.LimaHome>: can't evaluate field LimaHome in type *store.Instance
```


AkihiroSuda (approved):
Thanks

PR #1348, opened 2023-02-02
CI: update Go (1.20)

AkihiroSuda:
Depends on
- https://github.com/golangci/golangci-lint/pull/3414

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
         xcrun --show-sdk-version
     - uses: actions/setup-go@v3
       with:
-        go-version: 1.19.x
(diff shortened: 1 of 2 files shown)


balajiv113:
Should we update go.mod as well ??

afbjorklund:
> Should we update go.mod as well ??

The go.mod is the _minimum_ version

balajiv113 (approved):
LGTM 👍

AkihiroSuda:
Cirrus CI is down and can't be restarted 😞

jandubois:
> Cirrus CI is down and can't be restarted 😞

Should be coming back shortly; they moved to a different GCE region because the old one was running out of available resources.

Issue #893, opened 2022-06-10
Suggestion: allow using "." in mount location and\or scripts


centur:
### Description

Hi, first - thanks for the great project, lima is a first thing that worked properly from the start on M1 Mac for me, 💯 👍 .

I'm trying to build dev environment for the project using lima-vm and found that I need somehow map dirs which are relative to lima yaml file. The scenario is to have a VM with installed tools (provisioning scripts work great there) and configuration for tools from a repository. To achieve this I need some form of "relative" path to be mapped into vm on creation. There is an example of such path - `~` is expanded to user's home directory when used in `mounts.location`. Can we have something like `.` and `..` or some other path expansion tokens, which would be expanded to current template file directory, so during VM creation we can use something like:
```
mounts:
  - location: "./../config"
     mountPoint: /opt/hashicorp/config # will be supported in next version
     writable: false
```
to map certain folders from host to guest. Technically it can be achieved today by specifying a location path relative to home, e.g.`  - location: "~/myprojects/project-name/devenv/config"` but it's less portable as different developers may have different structure in their home dir.
Mount point is probably an ideal case, as it will be in sync all the time and can be updated from host with changes reflected in guest.
Alternative options may be  (1) to allow location to be  dynamically specified in `limactl start` command or (2) have an original template directory mounted in provisioning scripts. In the latter case I can simply copy latest config files in a provisioning script but have to restart an underlying VM. 

Alternatives are not ideal but would work better anyway, than asking developer to edit template and point to his project path before creating a vm from template.

jandubois:
I really don't understand what the relative filenames would be relative **to**, and how this wouldn't still be somehow depending on how a developer sets up their environment. I've added some comments in https://github.com/lima-vm/lima/pull/1351#issuecomment-1426872212 about the difficulty of figuring out what base to use to resolve relative pathnames in `lima.yaml`.

Why not tell the user to set an environment variable (inside `lima.yaml`) to point to the config file they want to use, e.g.

```yaml
env:
  MY_CONFIG: ~/.my_config

message: |
  Using config settings from {{.Config.Env.MY_CONFIG}}
  Change the config file by stopping the VM and then run `limactl edit {{.Name}}`
```

It will be displayed by `limactl start`:

```
limactl start myapp.yaml
[...]
INFO[0021] [hostagent] The final requirement 1 of 1 is satisfied
INFO[0021] READY. Run `limactl shell myapp` to open the shell.
INFO[0021] Message from the instance "myapp":
[...]

PR #1351, opened 2023-02-05
Add possibility to copy files from host to guest

afbjorklund:
Adding `copyToGuest` script, basically the reverse of `copyToHost` config

Added features of `{{.Pwd}}`, and copying user/group/executable bit.

Feature:

* #893

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -339,6 +339,13 @@ networks:
 #   hostPortRange: [1, 65535]
 # # Any port still not matched by a rule will not be forwarded (ignored)
 
+# Copy files from the host to the guest. Copied before provisioning scripts have been started.
+# copyToGuest:
+# - host: "{{.Pwd}}/myscript"
+#   guest: "/tmp/myscript.sh"
+# # "guest" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
+# # "host" can include {{.Pwd}}, {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
+
 # Copy files from the guest to the host. Copied after provisioning scripts have been completed.
 # copyToHost:
 # - guest: "/etc/myconfig.cfg"
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -388,6 +388,12 @@ sudo chown -R "${USER}" /run/host-services`
 			return unlockMErr
 		})
 	}
+	// Copy all script files _before_ the requirements are run
+	for _, rule := range a.y.CopyToGuest {
+		if err := copyToGuest(ctx, a.sshConfig, a.sshLocalPort, rule.HostFile, rule.GuestFile); err != nil {
+			mErr = multierror.Append(mErr, err)
+		}
+	}
 	go a.watchGuestAgentEvents(ctx)
 	if err := a.waitForRequirements(ctx, "optional", a.optionalRequirements()); err != nil {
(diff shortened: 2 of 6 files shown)


jandubois:
Could you create an issue to discuss the feature before creating a PR?

On first thought this seems redundant, as we already mount the host filesystem into the guest, so why do you need a way to copy files?

So I would like to understand the requirements first before discussing the implementation details.

afbjorklund:
The feature was mostly discussed on Slack, but _is_ described in the ticket...

I think the main problem was one of "current directory", where script is run ?

Just thought it could use the same construct, so that's why I made the PR.

jandubois:
> The feature was mostly discussed on Slack, but _is_ described in the ticket... (#893)

Thanks, the discussion on Slack seems to be about a different feature though, and those requirements are not actually met by this PR.

If we started with an issue, we could have this design discussion first:

### Feature Request

For a multi-vm setup the user wants to share provisioning scripts between the different `lima.yaml` files.

### Suggested implementation (implemented in this PR)

Instead of specifying the script inline in the template, it should be possible to refer to it by a relative filename.

### Problems with the proposed solution

This PR copies files from the host into the guest. This is redundant with the fact that the directories are already mounted anyways. But it also doesn't help because the copying happens in the guest agent, and at this time the provisioning scripts have already finished.

The concept of a "relative directory" is also ill-defined:

1. I don't think this can be relative to `$PWD` when the instance is started; that would artificially limit how `limactl` can be called. More importantly, this information is not recorded, so what value do we use when the instance is stopped and later restarted via `limactl start`, but from within a different directory?

2. We could use the location of the template used to create the instance. But what happens if the instance was created via an https URL, e.g. to the raw Github file? We would have to implement fetching additional files with modified URLs. Which seems tedious, but fine, until you think about restarting the instance while offline, so the script source might not be available. Do we need to introduce some caching mechanism.

I think it quickly becomes obvious that this mechanism is too special-purpose and too much work to implement generally.

### Alternative: run provisioning scripts after mounts are done

This has been a long-standing feature request (#287). If this feature existed, there would be no need to copy files, as they are already mounted. This feature is also considerably more generic, so can be re-used by others for different purposes (e.g. the recent #1335 feature could have been done by the end-user).

### But how do we deal with relative paths

For the purpose of this PR, we can define additional environment variables that are set before the provisioning scripts run, so anything that can be expanded in a template can also be passed in via the environment.

Since the whole reference to additional files is now the responsibility of the provisioning script, it becomes their responsibility to fetch files from the web if they don't exist locally or whatever else they want to do.

afbjorklund:
Okay, seems like reusing the code for copying will not be needed here. Closing PR.

afbjorklund:
> This is redundant with the fact that the directories are already mounted anyways.

Mounting disks is optional, though ? So not all examples / templates actually have the default mounts, like $HOME.

> But it also doesn't help because the copying happens in the guest agent, and at this time the provisioning scripts have already finished.

I'm pretty sure that the copying happened _after_ the essential steps and disks were mounted, but before the optional/final were run ?

https://github.com/afbjorklund/lima/blob/f340933eedf19a39f5ac946e921729ebaa32c4d7/pkg/hostagent/hostagent.go#L391

But it is true that the scripts were not available at boot time.

jandubois:
> I'm pretty sure that the copying happened _after_ the essential steps and disks were mounted, but before the optional/final were run ?

Yes, that's what I'm saying: the provisioning scripts run _before_ disks are mounted (for reverse-sshfs).

PR #1355, opened 2023-02-09
feat: recognize Kubernetes NodePorts and LoadBalancer

### LINKED ISSUE #1021: pkg/guestagent/iptables: recognize Kubernetes NodePorts
> I guess we need to adjust the regexp here to support kube NodePorts ?
> 
> 
> https://github.com/lima-vm/lima/blob/4edd35381c48e33721f54868d75408a4e6b41c8c/pkg/guestagent/iptables/iptables.go#L20-L34

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/discussions/1018#discussioncomment-3405760_

lobshunter:
This PR closes #1021, recognizes kubernetes NodePorts and LoadBalancer for port-forwarding. Based on the idea of https://github.com/lima-vm/lima/issues/1021#issuecomment-1230898249


pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -0,0 +1,130 @@
...
+
+		s.serviceInformer = serviceInformer
+		return true, nil
+	})
+}
+
+func tryGetKubeClient() (kubernetes.Interface, error) {
+	candidateKubeConfigs := []string{
+		"/etc/rancher/k3s/k3s.yaml",
+		"/root/.kube/config",
```
lobshunter, line 65: A more reliable way is to add kubernetes related config into lima YAML, but I feel it a bit exaggerated.

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -0,0 +1,130 @@
...
+const (
+	// UDP/SCTP when lima port forwarding works on those protocols
+	TCP Protocol = "TCP"
+)
+
+type Entry struct {
+	Protocol Protocol
+	IP       net.IP
+	Port     uint16
+}
```
lobshunter, line 32: I feel the need of unifying port detecting packages(prochttp, iptables, kubernetesservice) if this PR is merged. These packages should use the same `Entry` struct for instance.

lobshunter (commented):
I will fill up unit test if this implementation looks reasonable.

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -0,0 +1,130 @@
...
+		}
+
+		return kubeClient, nil
+	}
+
+	return nil, errors.New("no valid kubeconfig found")
+}
+
+func (s *ServiceWatcher) GetPorts() []Entry {
+	if s.serviceInformer == nil {
```
AkihiroSuda, line 103: Seems to need mutex

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -0,0 +1,130 @@
...
+			continue
+		}
+
+		for _, port := range service.Spec.Ports {
+			if port.Protocol != corev1.ProtocolTCP {
+				// currently only TCP port can be forwarded
+				continue
+			}
+
+			entries = append(entries, Entry{
```
Nino-K, line 121: I believe you may need a different port mapping logic for `ServiceTypeNodePort` and `ServiceTypeLoadBalancer`. 
This is something we came across in Rancher Desktop, for `ServiceTypeLoadBalancer` you probably want the `port.Port` as opposed to `port.NodePort`. Here is the [commit](https://github.com/rancher-sandbox/rancher-desktop-agent/pull/22/files#diff-062c5a2c47251be5d8a16a9d929c98ee7f201b0a791464d7b8a9a59f71ac037aR165) on Rancher Desktop that addressed the issue.

pkg/guestagent/kubernetesservice/kubernetesservice.go
lobshunter: Yes, I use the assumption that pointer type access is atomic. But that's not great, fixed in a76ecebc212b50d43ae3441086c030920e769ed8.

lobshunter:
I will squash commits  after all reviews are done, for reviewers' convenience.

pkg/guestagent/kubernetesservice/kubernetesservice.go
lobshunter: Great catch, fixed in e7e98f784122b8efcf132ea9503ae6251c6d4185 and manually tested with MetalLB.

AkihiroSuda (approved):
Thanks, could you squash the commits?

jandubois:
I think this is fine as-is, but I would like to review this on Monday together with @Nino-K to make sure any remaining additional code we have in the Rancher Desktop version is only required for WSL support on Windows, and not relevant to Lima.

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -0,0 +1,152 @@
...
+		}
+		u, err := url.Parse(restConfig.Host)
+		if err != nil {
+			return nil, fmt.Errorf("parse kubeconfig host %s failed: %w", restConfig.Host, err)
+		}
+		if u.Hostname() != "127.0.0.1" { // might need to support IPv6
+			// ensures the kubeconfig points to local k8s
+			continue
+		}
+
```
Nino-K, line 104: What you have here seems fine, but, want to let you know that you can also leverage the library to do more work for you, something like:
```
func getClientConfig(configPath string) (*restclient.Config, error) {
	loadingRules := clientcmd.ClientConfigLoadingRules{
		ExplicitPath: configPath,
	}
	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(&loadingRules, nil)
	config, err := clientConfig.ClientConfig()
	if err != nil {
		return nil, fmt.Errorf("could not load Kubernetes client config from %s: %w", configPath, err)
	}

	return config, nil
}
```

Nino-K (commented):
LGTM

pkg/guestagent/kubernetesservice/kubernetesservice.go
lobshunter: Nice one, I wasn't aware of this approach. Will update the code later today.

lobshunter: Hmmm, after some trying I would like to stay at current implementation. Because

using `Precedence` makes error message kind of confusing
```go
	loadingRules := clientcmd.ClientConfigLoadingRules{
		Precedence: []string{
			"/etc/rancher/k3s/k3s.yaml",
			"/root/.kube/config",
		},
	}
```

using `ExplicitPath` needs the same amount of code, and it would rely on client-go to return os.ErrNotExist if `configPath` is missing.
```go
	loadingRules := clientcmd.ClientConfigLoadingRules{
		ExplicitPath: configPath,
	}
```

AkihiroSuda (approved):
Thanks

PR #1359, opened 2023-02-12
Add builtin yq support to lima start command

afbjorklund:
Allows modifying the yaml template "inplace" for a single instance,
if you don't want to change the default or override for all instances.

This is similar to running `yq` on the template, from the command-line.
Except it does it on a temporary file, and for all different kinds of templates.

```console
$ yq < examples/default.yaml > default.yaml
$
```

Currently it removes empty lines, some quirk with the yaml parser used.
It also seems to change the indentation on some comments (only), but OK:

```console
$ diff -B -w examples/default.yaml default.yaml
$
```

This can be used as a general fallback, similar to `jq`, for modifying templates.
Instead of adding the usual `--cpus` `--memory` `--arch`, or other parameters ?

----

See <https://github.com/mikefarah/yq> for details

* #545

`limactl start config.yaml --yq '.arch = "aarch64"'`

It is not YAML syntax, but it is rather straightforward.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,6 +176,11 @@ $ limactl start --name=default template://docker
 > NOTE: `limactl start template://TEMPLATE` requires Lima v0.9.0 or later.
 > Older releases require `limactl start /usr/local/share/doc/lima/examples/TEMPLATE.yaml` instead.
 
+To create an instance "default" with modified parameters:
+```console
+$ limactl start --set='.cpus = 2 | .memory = "2GiB"'
+```
+
 To see the template list:
 ```console
 $ limactl start --list-templates
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -21,6 +21,7 @@ import (
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/lima-vm/lima/pkg/templatestore"
+	"github.com/lima-vm/lima/pkg/yqutil"
 	"github.com/mattn/go-isatty"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -35,6 +36,9 @@ $ limactl start
 
 To create an instance "default" from a template "docker":
 $ limactl start --name=default template://docker
+
+To create an instance "default" with modified parameters:
+$ limactl start --set='.cpus = 2 | .memory = "2GiB"'
 
 To see the template list:
 $ limactl start --list-templates
@@ -56,6 +60,7 @@ $ cat template.yaml | limactl start --name=local -
 	// TODO: "survey" does not support using cygwin terminal on windows yet
 	startCommand.Flags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "enable TUI interactions such as opening an editor, defaults to true when stdout is a terminal")
 	startCommand.Flags().String("name", "", "override the instance name")
+	startCommand.Flags().String("set", "", "modify the template inplace, using yq syntax")
 	startCommand.Flags().Bool("list-templates", false, "list available templates and exit")
 	startCommand.Flags().Duration("timeout", start.DefaultWatchHostAgentEventsTimeout, "duration to wait for the instance to be running before timing out")
 	return startCommand
@@ -79,6 +84,10 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string) (*store.Instance, e
 	}
 
 	st.instName, err = cmd.Flags().GetString("name")
+	if err != nil {
+		return nil, err
+	}
+	st.yq, err = cmd.Flags().GetString("set")
 	if err != nil {
 		return nil, err
 	}
@@ -206,6 +215,9 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string) (*store.Instance, e
 		}
 	} else {
 		logrus.Info("Terminal is not available, proceeding without opening an editor")
+		if err := modifyInPlace(st); err != nil {
+			return nil, err
+		}
 	}
 	saveBrokenEditorBuffer := tty
 	return createInstance(st, saveBrokenEditorBuffer)
@@ -261,10 +273,27 @@ func createInstance(st *creatorState, saveBrokenEditorBuffer bool) (*store.Insta
 type creatorState struct {
 	instName string // instance name
 	yBytes   []byte // yaml bytes
+	yq       string // yq expression
+}
+
+func modifyInPlace(st *creatorState) error {
+	if st.yq == "" {
+		return nil
+	}
+	out, err := yqutil.EvaluateExpression(st.yq, st.yBytes)
+	if err != nil {
+		return err
+	}
+	st.yBytes = out
+	return nil
 }
 
 func chooseNextCreatorState(st *creatorState) (*creatorState, error) {
 	for {
+		if err := modifyInPlace(st); err != nil {
+			logrus.WithError(err).Warn("Failed to evaluate yq expression")
+			return st, err
+		}
 		var ans string
 		prompt := &survey.Select{
 			Message: fmt.Sprintf("Creating an instance %q", st.instName),
--- a/pkg/yqutil/yqutil.go
+++ b/pkg/yqutil/yqutil.go
@@ -1,0 +1,66 @@
+package yqutil
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+
+	"github.com/mikefarah/yq/v4/pkg/yqlib"
+	"github.com/sirupsen/logrus"
+	logging "gopkg.in/op/go-logging.v1"
+)
+
+// EvaluateExpression evaluates the yq expression, and returns the modified yaml.
+func EvaluateExpression(expression string, content []byte) ([]byte, error) {
+	tmpYAMLFile, err := os.CreateTemp("", "lima-yq-*.yaml")
+	if err != nil {
+		return nil, err
+	}
+	tmpYAMLPath := tmpYAMLFile.Name()
+	defer os.RemoveAll(tmpYAMLPath)
+	err = os.WriteFile(tmpYAMLPath, content, 0o600)
+	if err != nil {
+		return nil, err
+	}
+
+	memory := logging.NewMemoryBackend(0)
+	backend := logging.AddModuleLevel(memory)
+	logging.SetBackend(backend)
+	yqlib.InitExpressionParser()
+
+	indent := 2
+	encoder := yqlib.NewYamlEncoder(indent, false, yqlib.ConfiguredYamlPreferences)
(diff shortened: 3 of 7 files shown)


pkg/yqutil/yqutil.go
```diff
@@ -0,0 +1,65 @@
...
+			prefix := fmt.Sprintf("[%s] ", node.Record.Module)
+			message := prefix + node.Record.Message()
+			switch node.Record.Level {
+			case logging.CRITICAL:
+				entry.Fatal(message)
+			case logging.ERROR:
+				entry.Error(message)
+			case logging.WARNING:
+				entry.Warn(message)
+			case logging.NOTICE:
```
AkihiroSuda, line 53: fallthrough

pkg/yqutil/yqutil_test.go
```diff
@@ -0,0 +1,28 @@
...
+func TestEvaluateExpression(t *testing.T) {
+	expression := `.cpus = 2 | .memory = "2GiB"`
+	content := `
+# CPUs
+cpus: null
+
+# Memory size
+memory: null
+`
+	// Note: yq currently removes empty lines, but not comments
```
AkihiroSuda, line 18: Nice, is the empty line issue tracked somewhere in the upstream?

pkg/yqutil/yqutil_test.go
```diff
@@ -0,0 +1,28 @@
+package yqutil
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestEvaluateExpression(t *testing.T) {
+	expression := `.cpus = 2 | .memory = "2GiB"`
```
AkihiroSuda, line 10: Can we also test a complex expression like mounts?

cmd/limactl/start.go
```diff
@@ -56,6 +60,7 @@ $ cat template.yaml | limactl start --name=local -
 	// TODO: "survey" does not support using cygwin terminal on windows yet
 	startCommand.Flags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "enable TUI interactions such as opening an editor, defaults to true when stdout is a terminal")
 	startCommand.Flags().String("name", "", "override the instance name")
+	startCommand.Flags().String("yq", "", "modify the template inplace")
```
AkihiroSuda, line 63: Maybe:
```suggestion
	startCommand.Flags().String("set", "", "modify the template inplace, using yq syntax")
```

AkihiroSuda: - I guess we have to prohibit modifying images?
- Maybe we want to treat this as an experimental?
- Please update `README.md` too

pkg/yqutil/yqutil_test.go
afbjorklund: * https://github.com/mikefarah/yq/issues/515

pkg/yqutil/yqutil_test.go
afbjorklund: Sure, probably should do an array like the template tests...

pkg/yqutil/yqutil_test.go
AkihiroSuda: Could you link that issue in the `// Note:` line?

cmd/limactl/start.go
afbjorklund: Why is it not allowed to modified images, from the templates ?

AkihiroSuda: Might be ok for new instances, but not allowable for existing instances, obviously

afbjorklund: Right. But this new tool is not very different, from doing it yourself in the editor...

AkihiroSuda (approved):
Thanks

AkihiroSuda:
> Instead of adding the usual --cpus --memory --arch, or other parameters ?

We should add them too, for typechecking, shell completion, etc. (In separate PRs)

README.md
```diff
@@ -176,6 +176,11 @@ $ limactl start --name=default template://docker
 > NOTE: `limactl start template://TEMPLATE` requires Lima v0.9.0 or later.
 > Older releases require `limactl start /usr/local/share/doc/lima/examples/TEMPLATE.yaml` instead.
 
+To create an instance "default" with modified parameters:
+```console
+$ limactl start --set='.cpus = 2 | .memory = "2GiB"'
+```
+
```
jandubois, line 183: I wonder if we need to include a reference to the `yq` documentation, so people can learn to construct more powerful expressions for e.g. scripting:

```console
$ limactl start --set '.env.HOSTPATH=strenv(PATH)' template://alpine
```

jandubois: I expected the same syntax to be available for the `limactl edit` command as well.

pkg/yqutil/yqutil_test.go
```diff
@@ -0,0 +1,65 @@
...
+# Expose host directories to the guest, the mount point might be accessible from all UIDs in the guest
+# 🟢 Builtin default: null (Mount nothing)
+# 🔵 This file: Mount the home as read-only, /tmp/lima as writable
+mounts:
+- location: "~"
+  # Configure the mountPoint inside the guest.
+  # 🟢 Builtin default: value of location
+  mountPoint: null
+`
+	// Note: yq will use canonical yaml, with indented sequences
```
jandubois, line 42: Not sure what "canonical YAML" means, but I would not consider indented sequences to be "canonical"; they are just the result of simplistic generators... 😛

jandubois (approved):
Thanks, LGTM.

I wish the option would also be added to `limactl edit`, but that could be done in a followup PR.

It also makes me a little sad that this little feature increases the size of the `limactl` binary by 20% (20MB → 24MB), but I guess that can't be helped, and doesn't really matter when you compare it to the size of QEMU. 😄

README.md
jandubois: I guess I should have specified: I would expect `limactl edit --set ...` to just edit the config, but not to invoke an editor.

It would just be a mechanism to modify the config programmatically without starting the instance.

afbjorklund:
> It also makes me a little sad that this little feature increases the size of the `limactl` binary by 20% (20MB → 24MB), but I guess that can't be helped, and doesn't really matter when you compare it to the size of QEMU. smile

I noticed that as well.

It would be possible to make a more conservative option, either some special-case for cpus/memory/arch/etc.
Or doing the `--set` approach, that would create a special override.yaml just for the particular instance.

The feature as such has a lot of "overkill" in it, like using `jq` (or even go templates) to change the list output...

pkg/yqutil/yqutil_test.go
afbjorklund: Yeah, in this case I think the issues listed (blank lines, indents) are quirks of the particular yaml library used.

But there were no good config or workaround for it, people were resorting to diff/patch and other such hacks

README.md
afbjorklund: We can use something like `--eval`, since that is the name of the `yq` command being invoked ?

```
Available Commands:
  completion       Generate the autocompletion script for the specified shell
  eval             (default) Apply the expression to each document in each yaml file in sequence
  eval-all         Loads _all_ yaml documents of _all_ yaml files and runs expression once
  help             Help about any command
  shell-completion Generate completion script
```

I agree that `--set` is somewhat misleading, and that `--yq` might be a bit cryptic for lima users...

Adding it to `edit` should be straight-forward.

AkihiroSuda: >  I would expect limactl edit --set ... to just edit the config, but not to invoke an editor.

:+1:

`--eval` sounds much more misleading

AkihiroSuda:
The binary footprint seems acceptable for me

README.md
AkihiroSuda: `limactl edit --set` can be another PR if complicated

pkg/yqutil/yqutil_test.go
jandubois: I think it is ok; just having comments preserved is already great.

README.md
afbjorklund: The error handling can be a bit tricky to do, so I'd like that.

jandubois:
> It would be possible to make a more conservative option,

No, I think I would prefer to have the expressiveness of `yq` expressions. I was just commenting that I noticed the size increase, but over time bloat is inevitable, and I think most users won't care (or even notice).

Idle curiosity: `yq` is 30× the size of `jq`

```console
$ ls -lh /usr/local/Cellar/jq/1.6/bin/jq
-r-xr-xr-x  1 jan  admin   296K  2 Jan 18:54 /usr/local/Cellar/jq/1.6/bin/jq
$ ls -lh /usr/local/Cellar/yq/4.30.8/bin/yq
-r-xr-xr-x  1 jan  admin   9.0M 14 Jan 16:37 /usr/local/Cellar/yq/4.30.8/bin/yq
```

afbjorklund:
It also does XML and JSON, in addition to just YAML. But I don't think it had any build flags ?

jandubois:
> It also does XML and JSON, in addition to just YAML. But I don't think it had any build flags ?

Of course it does JSON, as JSON is just a subset of YAML; every valid JSON file is also a valid YAML file. I didn't know `yq` also did XML though.

afbjorklund:
I meant that it links with both encoding/json and encoding/xml libraries:

```
1813	gopkg.in/yaml.v3
1478	syscall
1083	encoding/json
994	encoding/xml
917	time
830	os
773	fmt
677	regexp
650	github.com/magiconair/properties
552	math
524	github.com/alecthomas/participle/v2/lexer
503	unicode
497	strconv
487	strings
430	gopkg.in/op/go-logging.v1
386	bytes
368	github.com/goccy/go-json
343	bufio
315	io
302	net/url
292	sort
289	github.com/jinzhu/copier
273	io/fs
269	github.com/a8m/envsubst/parse
250	github.com/fatih/color
205	math/rand
151	github.com/goccy/go-yaml/printer
135	encoding/csv
122	encoding/base64
94	container/list
84	github.com/elliotchance/orderedmap
81	golang.org/x/net/html/charset
55	github.com/dimchansky/utfbom
44	errors
14	github.com/goccy/go-yaml/lexer
```

Sizes in kiB (`build-size`).

AkihiroSuda:
Needs rebase.

Are we ready to merge this after rebasing?

jandubois:
> Are we ready to merge this after rebasing?

I'm fine with merging.

afbjorklund:
I ran `go mod tidy` on it, so I hope those dependencies (`go sum`) are correct...
The logging is a direct dependency, due to the "translation" over to logrus.

For what it is worth, I made a PR to yq in order to make json and xml optional.
So that should cut the library size in half, if merged. `-tags yq_nojson,yq_noxml`

jandubois (approved):

AkihiroSuda (approved):
Thanks

PR #1368, opened 2023-02-14
Don't log DNS queries by default

### LINKED ISSUE #1367: DNS logging should be opt-in
Right now the host agent always logs at the DEBUG level, and at that level the hostresolver writes every single lookup to the log files.

This is problematic for some users who are not allowed to share internal server names, so they can't share log files.

It seems like HA logging is intentionally locked at the DEBUG level instead of inheriting the `--debug` setting from the `limactl start` command, so not sure what the right solution would be?

Should we move DNS logging to TRACE level, and switch from DEBUG to TRACE level logging when the instance was started by `limactl --debug start ...`?

Or just move it to TRACE level, and let the developer create a debugging binary where the hostagent logs at TRACE level, if we ever need to debug the hostresolver again? Maybe that will never happen anyways if we switch to gvisor for the user level networking.

Thoughts?

jandubois:
Local server names/addresses are often considered confidential and should not by default be logged. Their loglevel has been lowered to TRACE.

The host agent code now switches to TRACE level logging when the instance is started via `limactl --debug start ...`.

Fixes #1367

### DIFF
--- a/cmd/limactl/hostagent.go
+++ b/cmd/limactl/hostagent.go
@@ -122,5 +122,10 @@ func initLogrus(stderr io.Writer) {
 	logrus.SetOutput(stderr)
 	// JSON logs are parsed in pkg/hostagent/events.Watcher()
 	logrus.SetFormatter(new(logrus.JSONFormatter))
-	logrus.SetLevel(logrus.DebugLevel)
+	// HostAgent logging is one level more verbose than the start command itself
+	if logrus.GetLevel() == logrus.DebugLevel {
+		logrus.SetLevel(logrus.TraceLevel)
+	} else {
+		logrus.SetLevel(logrus.DebugLevel)
+	}
 }
--- a/pkg/hostagent/dns/dns.go
+++ b/pkg/hostagent/dns/dns.go
@@ -67,7 +67,7 @@ func (s *Server) Shutdown() {
 }
 
 func newStaticClientConfig(ips []string) (*dns.ClientConfig, error) {
-	logrus.Debugf("newStaticClientConfig creating config for the the following IPs: %v", ips)
+	logrus.Tracef("newStaticClientConfig creating config for the the following IPs: %v", ips)
 	s := ``
 	for _, ip := range ips {
 		s += fmt.Sprintf("nameserver %s\n", ip)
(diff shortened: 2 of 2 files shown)


jandubois:
@Nino-K Can you take a look if this makes sense to you?

Nino-K (commented):
Just a small comment, otherwise LGTM

cmd/limactl/hostagent.go
```diff
@@ -122,5 +122,10 @@ func initLogrus(stderr io.Writer) {
 	logrus.SetOutput(stderr)
 	// JSON logs are parsed in pkg/hostagent/events.Watcher()
 	logrus.SetFormatter(new(logrus.JSONFormatter))
-	logrus.SetLevel(logrus.DebugLevel)
+	// HostAgent logging is one level more verbose than the start command itself
```
AkihiroSuda, line 125: Isn't this a bug?

jandubois: Idk, I thought it was intentionally set to always log at DEBUG because the output goes to `ha.stderr.log` and doesn't pollute the terminal output: e88241ec8fa2f82f3b8322759c3c627f141eb141.

Which made sense to me: I don't normally want the output of `limactl --debug start` in the terminal, but I only look at `ha.stderr.log` when something is wrong, so always getting DEBUG output there is useful, so you don't have to restart your VM with `--debug` to get full logs.

So assuming this was intentional, I extended it to include TRACE level output when you run with `--debug`, as there is no way to specify the logging level for HA separately from the main logging level.

AkihiroSuda: IIRC that was just my silly workaround to avoid bother implementing `--log-level` propagation 🤦

jandubois: > IIRC that was just my silly workaround to avoid bother implementing --log-level propagation

Log-level propagation is implemented (and I tested that it works when you remove the explicit override the DEBUG in the HA code): https://github.com/lima-vm/lima/blob/5b9df0e0d7b0b471fb5e72d48625a378dc579ea4/pkg/start/start.go#L127-L129

I continue to think that always having debug logging active in `ha.stderr.log` is useful, even though it feels inconsistent. I just don't want to include DNS lookups in it by default.

So on a practical level this PR works for me. If we want to regularize the settings, I think we need a separate log-level for HA that should default to DEBUG. But this is just adding complexity for little benefit.

jandubois: > Log-level propagation is implemented 

I just checked, and the propagation was already implemented by the time the override was added. So it still feels intentional to me. 😄

AkihiroSuda: Propagation of custom log level is not implemented

AkihiroSuda: Sorry, you are right, I've forgotten my own coding design 😞

AkihiroSuda (approved):

PR #1369, opened 2023-02-15
Bump alpine-lima from 0.2.26 → 0.2.27

jandubois:
Updates binfmt handler to qemu 7.0 and fixes a bug that was preventing mounts of path names containing spaces.


AkihiroSuda (approved):
Thanks

PR #1373, opened 2023-02-19
fix bug with multiple extra qemu disks

BlaineEXE:
When starting a lima vm with multiple extra disks on the qemu backend, qemu would be unable to start the vm due to a misconfiguration, showing the error below:

DEBU[0000] [hostagent] qemu[stderr]: qemu-system-aarch64: -cdrom
/Users/blaine/.lima/lima-rook/cidata.iso: drive with bus=0, unit=2 (index=2) exists

When creating extra disks, use the 'index' parameter with a very high offset to ensure extra disks won't collide with any existing virtio devices. e.g., firmware, CDROM, cloud config, data disk, network device. This also allows more devices to be added in the future, even a reasonably arbitrary number of them, without much risk of colliding with extra disks.

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -34,6 +34,8 @@ type Config struct {
 	SSHLocalPort int
 }
 
+const extra_disk_offset_index = 100
+
 // EnsureDisk also ensures the kernel and the initrd
 func EnsureDisk(cfg Config) error {
 	diffDisk := filepath.Join(cfg.InstanceDir, filenames.DiffDisk)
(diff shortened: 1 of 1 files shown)


balajiv113:
@BlaineEXE 
Have you tried with lima head ??
We have already fixed one similar issue as part of https://github.com/lima-vm/lima/issues/1328

BlaineEXE:
> @BlaineEXE Have you tried with lima head ?? We have already fixed one similar issue as part of #1328

I haven't. I'll try that.

BlaineEXE:
> @BlaineEXE Have you tried with lima head ?? We have already fixed one similar issue as part of #1328

Fixed in master.

PR #1378, opened 2023-02-21
Document `rdctl shell` command for Rancher Desktop

jandubois:
And discourage us of `limactl` with the Rancher Desktop VM.


AkihiroSuda (approved):
Thanks

PR #1383, opened 2023-02-24
Support for lima usernet network

### LINKED ISSUE #1222: Enable guest-to-guest networking by default, with (gvisor-based) usermode networking
Currently the guests cannot talk to each other because each of them has the dedicated libslirp stack.

Lima should launch a (gvisor-based) usermode networking daemon and connect the guests to it by default so as to provide guest-to-guest networking.
The usermode networking daemon has to provide DHCP for the IP range that can be configured via `~/.lima/_config/networks.yaml`.

In addition, the daemon may provide virtual HTTP/SOCKS proxy (and/or other VPN-ish stuff) to allow connecting to the guest IP from the host. ( Similar to https://norouter.io/docs/getting-started/vpn/ )

We can even consider connecting the network to IaaS instances. That will be a lot of fun.

Alternative to
- #1185

balajiv113:
fixes #1222 

This PR adds support for the following,
- [x] Lima Usernet (Provides host -> vm, vm <-> vm communication) 
- [x] Usernet as daemon and support for reconciler similar to socket_vmnet 
- [x] New property added in yaml usernet.enabled, defaults to false. If its set to true, lima usernet will replace the slirp network 

Todo items
- [x] Handle timeout in usernet client 
- [x] Example template
- [x] Test for Performance issues
- [x] Writing integration test (Atleast for QEMU)
- [x] Documentation


pkg/limayaml/limayaml.go
```diff
@@ -31,6 +31,7 @@ type LimaYAML struct {
 	Env          map[string]string `yaml:"env,omitempty" json:"env,omitempty"`
 	DNS          []net.IP          `yaml:"dns,omitempty" json:"dns,omitempty"`
 	HostResolver HostResolver      `yaml:"hostResolver,omitempty" json:"hostResolver,omitempty"`
+	UserNet      UserNet           `yaml:"userNet,omitempty" json:"userNet,omitempty"`
```
AkihiroSuda, line 34: Shouldn't this be in `Networks`?

AkihiroSuda, line 34: >  Seems confusing to have two network options both called "user"...
>
> https://wiki.qemu.org/Documentation/Networking#User_Networking_(SLIRP)

_Originally posted by @afbjorklund in https://github.com/lima-vm/lima/issues/1222#issuecomment-1443778603_

balajiv113: Didn't add under networks because as of now Networks are additional config for us, 
but this will replace the behaviour of base network itself. So went ahead with a separate one.

AkihiroSuda: Bikeshedding:

- `gvisor` (confusing because it does not use gvisor's process sandbox)
- `gvisor-tap-vsock` (confusing because it is not gvisor, nor tap, nor vsock)
- `netstack` (confusing because any networking can be called "net(work) stack"
- `user-v2` (confusing because there was no `user-v1`)
- `modern` (confusing because it can be eventually legacy)

afbjorklund: Actually it would be OK to both of them "user", and then toggle the slirp vs gvisor later ?

Just was afraid that we would have one called "user" and a different one called "usernet"...

AkihiroSuda: I'd expect the default networking to be eventually configurable in `networks.yaml` 
```yaml
networks:
  default:
    mode: user-v2
    gateway: 192.168.5.1
    dhcpEnd: 192.168.5.254
    netmask: 255.255.255.0
  shared:
    mode: shared
    gateway: 192.168.105.1
    dhcpEnd: 192.168.105.254
    netmask: 255.255.255.0
```

When it is out of experimental, the `default` network will be automatically prepended to the network list to replace the legacy slirp network

balajiv113: Ah it makes sense. I made usernet as object mainly to provide this configuration later. But missed the part of configuring via networks.yaml 

Will move it under networks

balajiv113: Am bit confused on this "two network options both called user"

As of now we don't have any configuration for slirp right?
And the one we provide now will be more a toggle, if enabled will use the latest or revert to the old modal ?

afbjorklund: It is called user in qemu, is what I meant. If there is an option called usernet in lima too, it is confusing.

balajiv113: Done the changes. 
Added new property under Networks called UserNet to point to mode value networks.yml. Kept the mode as user-v2 for now.

Based on below discussion we can finalise a name and change it here as well accordingly

balajiv113: We could reuse Network.Lima property also. But we will end up needing to resolve networks as well. To identify if the Lima property is for user-v2 or socker_vmnet. 

Added UserNet to make this simple and keep this separate. If you think this will cause confusion happy to change code to reuse Lima property

AkihiroSuda: What about calling it "builtin"

AkihiroSuda: > But we will end up needing to resolve networks as well. To identify if the Lima property is for user-v2 or socker_vmnet.

Yes, is this problematic?

AkihiroSuda: Or "multivm", "multi-vm", "vm-to-vm", "vm2vm", "v2v"

balajiv113: `Yes, is this problematic?`
I thought it would be, but it is not actually. Done the changes with Lima property itself.

balajiv113: To me the following suggestions looks good,
- user-v2 as this give support for giving further versions as well like user-v3 etc.
- builtin since usernet command is within limactl it kind of makes sense as well

balajiv113:
Marking PR as ready for review as the code looks lot better compared to the initial version.

There are some todo items but i think it can be done incrementally as well. Should not affect the current code much.
Will start with documentation as a last item once naming are finalised

balajiv113:
Testing for performance are done as well. 

I tested in M1 with some long running instance (for more than a day with sleep). Haven't faced any issues (There was slight degrade when did iperf3, but on restarting iperf3 -s on host the performance was back to more meaningful value)

Also performance of QEMU is improved with this PR https://github.com/containers/gvisor-tap-vsock/pull/188

Performance report
**VZ <-> Host**
```
$ iperf3 -c host.lima.internal
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  2.64 GBytes  2.27 Gbits/sec    0             sender
[  5]   0.00-10.02  sec  2.64 GBytes  2.26 Gbits/sec                  receiver

iperf Done.
$ iperf3 -c host.lima.internal -R
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  2.02 GBytes  1.73 Gbits/sec    0             sender
[  5]   0.00-10.00  sec  2.02 GBytes  1.73 Gbits/sec                  receiver
```

**QEMU <-> Host**
```
$ iperf3 -c host.lima.internal
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  1.84 GBytes  1.58 Gbits/sec    0             sender
[  5]   0.00-10.00  sec  1.84 GBytes  1.58 Gbits/sec                  receiver

iperf Done.
$ iperf3 -c host.lima.internal -R
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  1.21 GBytes  1.04 Gbits/sec    0             sender
[  5]   0.00-10.00  sec  1.21 GBytes  1.04 Gbits/sec                  receiver

iperf Done.
```

**VZ <-> QEMU**
```
$ iperf3 -c 192.168.5.1
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.00  sec  1.44 GBytes  1.24 Gbits/sec    0             sender
[  5]   0.00-10.07  sec  1.44 GBytes  1.23 Gbits/sec                  receiver

iperf Done.
$ iperf3 -c 192.168.5.1 -R
[ ID] Interval           Transfer     Bitrate         Retr
[  5]   0.00-10.04  sec  1.83 GBytes  1.57 Gbits/sec    0             sender
[  5]   0.00-10.00  sec  1.83 GBytes  1.57 Gbits/sec                  receiver

iperf Done.

```

examples/experimental/usernet.yaml
```diff
@@ -0,0 +1,15 @@
...
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img"
+  arch: "aarch64"
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+mountType: "virtiofs"
+networks:
+- lima: default
```
AkihiroSuda, line 15: If the new network is stable enough to make it default, the `default` network should be prepended to the network list by default.
If not, the network shouldn't be called "default" at this moment.

examples/experimental/usernet.yaml
```diff
@@ -0,0 +1,15 @@
+# Example to run vz instance with lima usernet enabled
```
AkihiroSuda, line 1: > `usernet.yaml`

`usernet-v2.yaml`, or maybe `net-user-v2.yaml`

pkg/limayaml/limayaml.go
AkihiroSuda: @afbjorklund @jandubois
WDYT?

examples/experimental/usernet.yaml
balajiv113: On my tests, the network looked stable only. But would like to see others opinions as well mainly with to high load and sleep and wake scenarios.

shall i rename it to `experimental` ?

examples/experimental/usernet.yaml
balajiv113: Will rename this once the name is decided in above comment

examples/experimental/usernet.yaml
AkihiroSuda: `user-v2` or `user-v2-experimental`
(should correspond to the network driver name)

pkg/limayaml/limayaml.go
afbjorklund: I don't know what happened with "user-v1", is that the normal one ? But "user-v2" seems complicated to me, as a yaml default. But maybe it needs versioning (as per above, "user-v3", "user-v4") otherwise "builtin" is simpler.

afbjorklund:
Here is what I got:

```
? Creating an instance "usernet" Proceed with the current configuration
FATA[0001] the YAML is invalid, saved the buffer as "lima.REJECTED.yaml": field `networks[0].lima` is only supported on macOS right now 
```

balajiv113:
Missed support for linux with networks.lima property.
Will fix it. Thanks 👍

balajiv113:
@afbjorklund 
Done changes to support linux, can you check if it works now :)

pkg/limayaml/limayaml.go
abiosoft: > What about calling it "builtin"

My vote is for `builtin`

jandubois:
> @afbjorklund @jandubois
> WDYT?

To be honest, I kind of like the name "usernet". I'm also not confusing it with QEMU networking, as I always think of that one as SLIRP. "User mode networking" is just the general category, and SLIRP is a specific instance of it.

Now, you could argue that we should not use the category name ("user mode networking") and usurp it for the specific Lima implementation, but again personally I'm not bothered by it because it is disambiguated by context (it is the Lima usernet). I think descriptive names are good.

And if we ever need another name, and still can't come up with anything better, then we can still use `usernet-v2`. But maybe (hopefully) we will just have some different configuration parameter to tweak the implementation instead of having to replace it wholesale again.

AkihiroSuda:
> To be honest, I kind of like the name "usernet". I'm also not confusing it with QEMU networking, as I always think of that one as SLIRP. "User mode networking" is just the general category, and SLIRP is a specific instance of it.

"slirp" has become a general category too.
Docker calls VPNKit as "slirp", although its implementation is different from the original slirp.

examples/experimental/usernet.yaml
AkihiroSuda: ping 🙏

balajiv113: @AkihiroSuda 
Shall I go with user-v2 as most people are okay with that naming ?

Also there is still issues related to network freeze on high load. I have been trying to debug it but still in progress. We can merge this as experimental and analyse that on the side as well.

AkihiroSuda: Yes, user-v2 SGTM.
We can rename it later if we want, as long as we keep it experimental.

balajiv113: Done

examples/experimental/usernet.yaml
balajiv113: Done

balajiv113:
@AkihiroSuda
Added documentation as well. For testing, the newly added template is already been tested via integration test. I hope that should be sufficient for now.

I will do a follow-up story to support configurable IP range for user-v2. Before this change we need some improvement / cleanup with DNS resolution.

examples/experimental/net-user-v2.yaml
```diff
@@ -0,0 +1,15 @@
+# Example to run vz instance with experimental user-v2 network enabled
```
AkihiroSuda, line 1: Can we test this in the CI?
Probably we can just use Linux instances on Cirrus (faster than GHA).

docs/network.md
```diff
@@ -193,3 +193,33 @@ networks:
...
+
+```yaml
+networks:
+   - lima: example-user-v2
+```
+
+_Note_
+
+- Enabling this network will disable the [default user-mode network](#user-mode-network--1921685024-)
+- Subnet used for this network is 192.168.5.0/24 with 192.168.5.2 used for host connection and 192.168.5.3 used for DNS resolution
```
AkihiroSuda, line 224: Is this configurable or not?

docs/network.md
```diff
@@ -193,3 +193,33 @@ networks:
...
+Instances can then reference these networks from their `lima.yaml` file:
+
+```yaml
+networks:
+   - lima: example-user-v2
+```
+
+_Note_
+
+- Enabling this network will disable the [default user-mode network](#user-mode-network--1921685024-)
```
AkihiroSuda, line 223: What are drawbacks?

docs/experimental.md
```diff
@@ -6,6 +6,7 @@ The following features are experimental and subject to change:
 - `vmType: vz` and relevant configurations (`mountType: virtiofs`, `rosetta`, `[]networks.vzNAT`)
 - `arch: riscv64`
 - `video.display: vnc` and relevant configuration (`video.vnc.display`)
+- `mode: user-v2` in networks.yml and relevant lima configuration in lima.yaml
```
AkihiroSuda, line 9: nit
```suggestion
- `mode: user-v2` in `networks.yml` and relevant configurations in `lima.yaml `
```

pkg/fd/fd.go
```diff
@@ -0,0 +1,109 @@
+//go:build !windows
+// +build !windows
```
AkihiroSuda, line 2: You can drop these lines if you rename the files to `*_unix.go`

pkg/fd/fd.go
```diff
@@ -0,0 +1,109 @@
...
+
+// This file has been adapted from https://github.com/ftrvxmtrx/fd/blob/master/fd.go
```
AkihiroSuda, line 4:
```suggestion
// This file has been adapted from https://github.com/ftrvxmtrx/fd/blob/c6d800382fff6dc1412f34269f71b7f83bd059ad/fd.go

```

Also please follow https://github.com/ftrvxmtrx/fd/blob/c6d800382fff6dc1412f34269f71b7f83bd059ad/LICENSE.MIT

AkihiroSuda: Maybe you can just fork https://github.com/ftrvxmtrx/fd into your own repo, and just import it here

pkg/fd/fd_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+import (
+	"errors"
+	"net"
+	"os"
+	"sync"
+	"testing"
+)
+
+var SockFilename = "/tmp/sendfd.sock"
```
AkihiroSuda, line 16: `t.TempDir()`

pkg/fd/fd_test.go
```diff
@@ -0,0 +1,87 @@
+//go:build linux && darwin
+// +build linux,darwin
+
+// This file has been adapted from https://github.com/ftrvxmtrx/fd/blob/master/fd_test.go
```
AkihiroSuda, line 4: https://github.com/lima-vm/lima/pull/1383/files#r1151127463

pkg/limayaml/defaults.go
```diff
@@ -49,6 +51,17 @@ func defaultContainerdArchives() []File {
 	}
 }
 
+func GetFirstUsernetIndex(l *LimaYAML) int {
```
AkihiroSuda, line 54:
```suggestion
// Please add Godoc here
func FirstUsernetIndex(l *LimaYAML) int {
```

https://go.dev/doc/effective_go#Getters

pkg/networks/networks.TEMPLATE.yaml
```diff
@@ -22,6 +22,8 @@ paths:
 group: everyone
 
 networks:
+  user-v2:
```
AkihiroSuda, line 25: Can we have comment lines here too?

pkg/networks/usernet/recoincile.go
```diff
@@ -0,0 +1,128 @@
...
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/lima-vm/lima/pkg/store/dirnames"
+	"github.com/sirupsen/logrus"
+)
+
+func StartUsernet(ctx context.Context, name string) error {
```
AkihiroSuda, line 18: Seems to need flock

AkihiroSuda, line 18:
```suggestion
func Start(ctx context.Context, name string) error {
```

cmd/limactl/usernet.go
```diff
@@ -0,0 +1,71 @@
...
+		Use:    "usernet",
+		Short:  "run usernet",
+		Args:   cobra.ExactArgs(0),
+		RunE:   usernetAction,
+		Hidden: true,
+	}
+	hostagentCommand.Flags().StringP("pidfile", "p", "", "write pid to file")
+	hostagentCommand.Flags().StringP("endpoint", "e", "", "exposes usernet api(s) on this endpoint")
+	hostagentCommand.Flags().String("listen-qemu", "", "listen for qemu connections")
+	hostagentCommand.Flags().String("listen-fd", "", "listen for fd connections")
```
AkihiroSuda, line 24:
```suggestion
	hostagentCommand.Flags().String("listen", "", "listen on a Unix socket and receive Bess-compatible FDs as SCM_RIGHTS messages")
```

Becase `listen-fd` sounds like to take an integer value that represents an FD

pkg/networks/usernet/gvproxy.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+		defer f.Close()
+		sc := bufio.NewScanner(f)
+		searchPrefix := "search "
+		for sc.Scan() {
+			if strings.HasPrefix(sc.Text(), searchPrefix) {
+				searchDomains := strings.Split(strings.TrimPrefix(sc.Text(), searchPrefix), " ")
+				logrus.Debugf("Using search domains: %v", searchDomains)
+				return searchDomains
+			}
```
AkihiroSuda, line 221: Can we have a unit test for this function?

docs/network.md
balajiv113: Not configurable as of now. I will provide this support in a follow-up PR. 
This needs some changes related to DNS resolving as well

docs/network.md
balajiv113: Only drawback i see is,
Running high number of vm's under this network might add bottleneck from gvisor-tap-vsock.

pkg/fd/fd.go
balajiv113: Will do this.

docs/experimental.md
balajiv113: Done

pkg/fd/fd.go
balajiv113: Moved to fork repo

pkg/fd/fd_test.go
balajiv113: Moved to fork repo

cmd/limactl/usernet.go
balajiv113: Done

pkg/networks/usernet/gvproxy.go
balajiv113: Done

pkg/limayaml/defaults.go
balajiv113: Done

pkg/networks/networks.TEMPLATE.yaml
balajiv113: Done

pkg/networks/usernet/recoincile.go
balajiv113: Done. 
Moved the storage location of usernet files to <LIMA_HOME>/_networks/<name>/files

balajiv113: Done

pkg/networks/usernet/usernet-client.go
```diff
@@ -0,0 +1,96 @@
...
+		case <-tick:
+			leases, err := c.leases()
+			if err != nil {
+				return err
+			}
+
+			for ipAddr, leaseAddr := range leases {
+				if vmMacAddr == leaseAddr {
+					err = c.delegate.Expose(&types.ExposeRequest{
+						Local:    fmt.Sprintf(":%d", sshPort),
```
AkihiroSuda, line 47:
```suggestion
						Local:    fmt.Sprintf("127.0.0.1:%d", sshPort),
```

pkg/networks/usernet/usernet-client.go
AkihiroSuda, line 1: s/-/_/ in the filename

pkg/networks/usernet/usernet-client.go
balajiv113: Done

pkg/networks/usernet/usernet-client.go
balajiv113: Done

examples/experimental/net-user-v2.yaml
balajiv113: Done, added as part of current it test script itself. 

Missed some condition changes. Will update this and make the pipeline green in linux.

examples/experimental/net-user-v2.yaml
```diff
@@ -0,0 +1,15 @@
...
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img"
+  arch: "aarch64"
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+mountType: "virtiofs"
```
AkihiroSuda, line 13: Shouldn't be needed

examples/experimental/net-user-v2.yaml
```diff
@@ -0,0 +1,15 @@
+# Example to run vz instance with experimental user-v2 network enabled
+vmType: "vz"
```
AkihiroSuda, line 2: Shouldn't be needed

examples/experimental/net-user-v2.yaml
AkihiroSuda: Thanks, I'd like to see VM-to-VM communication test, but that can be a separate PR

examples/experimental/net-user-v2.yaml
balajiv113: Done

examples/experimental/net-user-v2.yaml
balajiv113: Done

examples/experimental/net-user-v2.yaml
balajiv113: This is done as well. For now its one way ping from one vm to another. But can improve this later.

AkihiroSuda:
Sorry needs another rebase

balajiv113:
@AkihiroSuda - Done :)

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@jandubois @afbjorklund May I merge this, or do you want to take the final look before merging?

jandubois:
> May I merge this, or do you want to take the final look before merging?

I did want to review it, but I continue to be out of time. Feel free to merge if you are ok with it. At least it gets it into more people's hands for testing.

AkihiroSuda:
> Any chance of releasing this? It seems it would fix some issues I'm having.

Yes, probably next week

PR #1384, opened 2023-02-24
k8s: Remove redundant crictl.yaml config line

afbjorklund:
The default for image-endpoint is to use runtime-endpoint.

So it is not needed, to set both of them to the same value.

----

```
   --image-endpoint value, -i value    Endpoint of CRI image manager service (default: uses 'runtime-endpoint' setting) [$IMAGE_SERVICE_ENDPOINT]
   --runtime-endpoint value, -r value  Endpoint of CRI container runtime service (default: uses in order the first successful one of [unix:///var/run/dockershim.sock unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock unix:///var/run/cri-dockerd.sock]). Default is now deprecated and the endpoint should be set instead. [$CONTAINER_RUNTIME_ENDPOINT]
```

### DIFF
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -62,7 +62,6 @@ provision:
     apt-get install -y cri-tools
     cat  <<EOF | sudo tee /etc/crictl.yaml
     runtime-endpoint: unix:///run/containerd/containerd.sock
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #1397, opened 2023-03-01
Fix broken state on QEMU ACPI shutdown

### LINKED ISSUE #1390: ACPI shutdown shouldn't cause "Broken" status (`errors="[host agent is running but driver is not]"`)
```console
$ lima sudo poweroff

$ limactl ls
WARN[0000] instance "default" has errors                 errors="[host agent is running but driver is not]"
NAME         STATUS    SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
default      Broken    127.0.0.1:60022    qemu      x86_64     2       4GiB      100GiB    ~/.lima/default
```

Workaround: `limactl stop -f default`

balajiv113:
Fixes #1390


AkihiroSuda (approved):
Thanks

AkihiroSuda:
The status can be still `Broken` for a few seconds during shutting down, but I guess it is not a big deal.

PR #1398, opened 2023-03-01
Run qemu-img info in shared mode for GetInfo

### LINKED ISSUE #1392: Stopping an instance does not detach the disk
### Description

It seems like when you stop an instance, the disk is still flagged as being "in use" (by itself).

This means that you can't start it again, without first running `stop -f` to detach the disk...

afbjorklund:
Avoids errors about the disk being in use by the running qemu.

Older versions of qemu default to read-write mode for images.

Closes #1392


jandubois (approved):
Thanks, LGTM

AkihiroSuda (approved):
Thanks

PR #1405, opened 2023-03-02
feat: mount additional disks when using vz

pendo324:
This PR adds support for using the additional disks feature with Virtualization.framework. I noticed it was missing, so I figured I'd add it.

Considerations:
- [Virtualization.framework can only mount RAW disk images](https://developer.apple.com/documentation/virtualization/vzdiskimagestoragedeviceattachment). The additional disk feature [currently creates QCOW2 images](https://github.com/lima-vm/lima/blob/5a9bca3d09481ed7109b14f8d3f0074816731f43/pkg/qemu/qemu.go#L115). Since QEMU can support RAW images, is it ok to change the additional disk feature to create RAW images by default?
  - This change already included in this diff, but I can remove it
- Should Lima automatically convert QCOW2 disks to RAW before trying to mount a QCOW2 additional with `vmType: vz`? There's already [precedent for runtime conversion of disks in imgutil](https://github.com/lima-vm/lima/blob/5a9bca3d09481ed7109b14f8d3f0074816731f43/pkg/qemu/imgutil/imgutil.go#L18-L28), just wanted to make sure we want to do it for additional disks before adding
  - Auto conversion is already included in this diff, but I can remove it
  - I was having issues doing an in place conversion (I saw in place conversion in other uses of `imgutil.QCOWToRaw`), causing data loss, which is why there's an intermediate file right now. Not sure if there's a better way to do this?

Test:

```shell
$ qemu-img info --output=json ~/.lima/_disks/test/datadisk
{
    "virtual-size": 53687091200,
    "filename": "/Users/alvajus/.lima/_disks/test/datadisk",
    "cluster-size": 65536,
    "format": "qcow2",
    "actual-size": 14221312,
    "format-specific": {
        "type": "qcow2",
        "data": {
            "compat": "1.1",
            "compression-type": "zlib",
            "lazy-refcounts": false,
            "refcount-bits": 16,
            "corrupt": false,
            "extended-l2": false
        }
    },
    "dirty-flag": false
}

# Create an instance using the default fedora.yaml template with addition of additionalDisks: - test (note: test was created as QCOW2 as seen above)
$ ./_output/bin/limactl start --name=fedora template://fedora
...
INFO[0129] READY. Run `limactl shell fedora` to open the shell.

$ ./_output/bin/limactl shell fedora
[alvajus@lima-fedora lima]$
[alvajus@lima-fedora lima]$ ls /mnt/lima-test/
lost+found
[alvajus@lima-fedora lima]$ echo "THIS IS MY TEST FILE" | sudo tee /mnt/lima-test/testfile
THIS IS MY TEST FILE
[alvajus@lima-fedora lima]$ cat /mnt/lima-test/testfile
THIS IS MY TEST FILE
[alvajus@lima-fedora lima]$ exit
logout

$ ./_output/bin/limactl stop fedora
...
INFO[0002] [hostagent] QEMU has exited

$ ./_output/bin/limactl remove fedora
...
INFO[0000] Deleted "fedora" ("/Users/alvajus/.lima/fedora")

# Modify fedora.yaml to still use the same disk, but also include vmType: "vz" and recreate the VM
$ ./_output/bin/limactl start --name=fedora template://fedora
? Creating an instance "fedora" Proceed with the current configuration
WARN[0000] `vmType: vz` is experimental
...
INFO[0146] READY. Run `limactl shell fedora` to open the shell.

$ ./_output/bin/limactl shell fedora
[alvajus@lima-fedora lima]$ ls /mnt/lima-test/
lost+found  testfile
[alvajus@lima-fedora lima]$ cat /mnt/lima-test/testfile
THIS IS MY TEST FILE
[alvajus@lima-fedora lima]$ exit
logout

$ qemu-img info --output=json /Users/alvajus/.lima/_disks/test/datadisk
{
    "virtual-size": 53687091200,
    "filename": "/Users/alvajus/.lima/_disks/test/datadisk",
    "format": "raw",
    "actual-size": 6107136,
    "dirty-flag": false
}
```

### DIFF
--- a/pkg/vz/vm_darwin.go
+++ b/pkg/vz/vm_darwin.go
@@ -21,6 +21,7 @@ import (
 	"github.com/lima-vm/lima/pkg/localpathutil"
 	"github.com/lima-vm/lima/pkg/networks"
 	"github.com/lima-vm/lima/pkg/qemu/imgutil"
+	"github.com/lima-vm/lima/pkg/store"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/sirupsen/logrus"
 )
@@ -333,6 +334,58 @@ func attachDisks(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfigura
 		return err
 	}
 	configurations = append(configurations, diffDisk)
+
+	for _, diskName := range driver.Yaml.AdditionalDisks {
+		d, err := store.InspectDisk(diskName)
+		if err != nil {
+			return fmt.Errorf("failed to run load disk %q: %q", diskName, err)
+		}
+
+		if d.Instance != "" {
+			return fmt.Errorf("failed to run attach disk %q, in use by instance %q", diskName, d.Instance)
+		}
+		logrus.Infof("Mounting disk %q on %q", diskName, d.MountPoint)
+		err = d.Lock(driver.Instance.Dir)
+		if err != nil {
+			return fmt.Errorf("failed to run lock disk %q: %q", diskName, err)
+		}
+		extraDiskPath := filepath.Join(d.Dir, filenames.DataDisk)
+
+		extraDiskFormat, err := imgutil.DetectFormat(extraDiskPath)
+		if err != nil {
+			return fmt.Errorf("failed to run detect disk format %q: %q", diskName, err)
+		}
+		if extraDiskFormat != "raw" {
+			rawPath := fmt.Sprintf("%s.raw", extraDiskPath)
+			qcow2Path := fmt.Sprintf("%s.qcow2", extraDiskPath)
+			if err = imgutil.QCOWToRaw(extraDiskPath, rawPath); err != nil {
+				return fmt.Errorf("failed to convert qcow2 disk %q to raw for vz driver: %w", diskName, err)
+			}
+			if err = os.Rename(extraDiskPath, qcow2Path); err != nil {
+				return fmt.Errorf("failed to rename additional disk for vz driver: %w", err)
+			}
+			if err = os.Rename(rawPath, extraDiskPath); err != nil {
+				return fmt.Errorf("failed to rename additional disk for vz driver: %w", err)
+			}
+			if err = os.Remove(qcow2Path); err != nil {
+				logrus.Errorf("Failed to delete unused qcow2 additional disk %q."+
+					"Disk is no longer needed by Lima and it can be removed manually.", qcow2Path)
+			}
+		}
+
+		if err = validateDiskFormat(extraDiskPath); err != nil {
+			return fmt.Errorf("failed to validate extra disk %q: %w", extraDiskPath, err)
+		}
+		extraDiskPathAttachment, err := vz.NewDiskImageStorageDeviceAttachmentWithCacheAndSync(extraDiskPath, false, vz.DiskImageCachingModeAutomatic, vz.DiskImageSynchronizationModeFsync)
+		if err != nil {
+			return fmt.Errorf("failed to create disk attachment for extra disk %q: %w", extraDiskPath, err)
+		}
+		extraDisk, err := vz.NewVirtioBlockDeviceConfiguration(extraDiskPathAttachment)
+		if err != nil {
+			return fmt.Errorf("failed to create new virtio block device config for extra disk %q: %w", extraDiskPath, err)
+		}
+		configurations = append(configurations, extraDisk)
+	}
 
 	if err = validateDiskFormat(ciDataPath); err != nil {
 		return err
--- a/pkg/vz/vz_driver_darwin.go
+++ b/pkg/vz/vz_driver_darwin.go
@@ -64,6 +64,7 @@ func (l *LimaVzDriver) Validate() error {
 		"PropagateProxyEnv",
 		"CACertificates",
 		"Rosetta",
+		"AdditionalDisks",
 	); len(unknown) > 0 {
 		logrus.Warnf("Ignoring: vmType %s: %+v", *l.Yaml.VMType, unknown)
 	}
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ Use `<INSTANCE>:<FILENAME>` to specify a source or target inside an instance.
 
 #### `limactl disk`
 
-`limactl disk create <DISK> --size <SIZE>`: create a new external disk to attach to an instance
+`limactl disk create <DISK> --size <SIZE> [--format qcow2]`: create a new external disk to attach to an instance
 
 `limactl disk delete <DISK>`: delete an existing disk
 
--- a/cmd/limactl/disk.go
+++ b/cmd/limactl/disk.go
@@ -20,7 +20,7 @@ func newDiskCommand() *cobra.Command {
 		Use:   "disk",
 		Short: "Lima disk management",
 		Example: `  Create a disk:
-  $ limactl disk create DISK --size SIZE
+  $ limactl disk create DISK --size SIZE [--format qcow2]
 
   List existing disks:
   $ limactl disk ls
@@ -44,7 +44,7 @@ func newDiskCreateCommand() *cobra.Command {
 		Use: "create DISK",
 		Example: `
 To create a new disk:
-$ limactl disk create DISK --size SIZE
+$ limactl disk create DISK --size SIZE [--format qcow2]
 `,
 		Short: "Create a Lima disk",
 		Args:  WrapArgsError(cobra.ExactArgs(1)),
@@ -52,6 +52,7 @@ $ limactl disk create DISK --size SIZE
 	}
 	diskCreateCommand.Flags().String("size", "", "configure the disk size")
 	diskCreateCommand.MarkFlagRequired("size")
+	diskCreateCommand.Flags().String("format", "qcow2", "specify the disk format")
 	return diskCreateCommand
 }
 
@@ -61,9 +62,20 @@ func diskCreateAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	format, err := cmd.Flags().GetString("format")
+	if err != nil {
+		return err
+	}
+
 	diskSize, err := units.RAMInBytes(size)
 	if err != nil {
 		return err
+	}
+
+	switch format {
+	case "qcow2", "raw":
+	default:
+		return fmt.Errorf(`disk format %q not supported, use "qcow2" or "raw" instead`, format)
 	}
 
 	// only exactly one arg is allowed
@@ -78,13 +90,13 @@ func diskCreateAction(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("disk %q already exists (%q)", name, diskDir)
 	}
 
-	logrus.Infof("Creating disk %q with size %s", name, units.BytesSize(float64(diskSize)))
+	logrus.Infof("Creating %s disk %q with size %s", format, name, units.BytesSize(float64(diskSize)))
 
 	if err := os.MkdirAll(diskDir, 0700); err != nil {
 		return err
 	}
 
-	if err := qemu.CreateDataDisk(diskDir, int(diskSize)); err != nil {
+	if err := qemu.CreateDataDisk(diskDir, format, int(diskSize)); err != nil {
 		return err
 	}
 
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -74,10 +74,14 @@ Host agent:
 A disk directory contains the following files:
 
 data disk:
-- `datadisk`: the qcow2 disk that is attached to an instance
+- `datadisk`: the qcow2 or raw disk that is attached to an instance
 
 lock:
 - `in_use_by`: symlink to the instance directory that is using the disk
+
+When using `vmType: vz` (Virtualization.framework), on boot, any qcow2 (default) formatted disks that are specified in `additionalDisks` will be converted to RAW since [Virtualization.framework only supports mounting RAW disks](https://developer.apple.com/documentation/virtualization/vzdiskimagestoragedeviceattachment). This conversion enables additional disks to work with both Virtualization.framework and QEMU, but it has some consequences when it comes to interacting with the disks. Most importantly, a regular macOS default `cp` command will copy the _entire_ virtual disk size, instead of just the _used/allocated_ portion. The easiest way to copy only the used data is by adding the `-c` option to cp: `cp -c old_path new_path`. `cp -c` uses clonefile(2) to create a copy-on-write clone of the disk, and should return instantly.
+
+`ls` will also only show the full/virtual size of the disks. To see the allocated space, `du -h disk_path` or `qemu-img info disk_path` can be used instead. See [#1405](https://github.com/lima-vm/lima/pull/1405) for more details.
 
 ## Lima cache directory (`~/Library/Caches/lima`)
 
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -104,14 +104,14 @@ func EnsureDisk(cfg Config) error {
 	return nil
 }
 
-func CreateDataDisk(dir string, size int) error {
+func CreateDataDisk(dir, format string, size int) error {
 	dataDisk := filepath.Join(dir, filenames.DataDisk)
 	if _, err := os.Stat(dataDisk); err == nil || !errors.Is(err, fs.ErrNotExist) {
 		// datadisk already exists
 		return err
 	}
 
-	args := []string{"create", "-f", "qcow2", dataDisk, strconv.Itoa(size)}
+	args := []string{"create", "-f", format, dataDisk, strconv.Itoa(size)}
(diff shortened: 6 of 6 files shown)


pkg/vz/vz_driver_darwin.go
```diff
@@ -106,6 +108,15 @@ func (l *LimaVzDriver) CreateDisk() error {
 	return nil
 }
 
+func (l *LimaVzDriver) InspectDisk(diskName string) (*store.Disk, error) {
+	disk, err := l.BaseDriver.InspectDisk(diskName)
```
balajiv113, line 112: can return directly

pkg/vz/vm_darwin.go
```diff
@@ -334,6 +335,58 @@ func attachDisks(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfigura
 	}
 	configurations = append(configurations, diffDisk)
 
+	if len(driver.Yaml.AdditionalDisks) > 0 {
```
balajiv113, line 338: This condition is not needed. In case of length 0 for loop should take care i believe

balajiv113:
> Should Lima automatically convert QCOW2 disks to RAW before trying to mount

IMO The goal of additionalDisks are to share disk between instances doing conversion from QCOW2 - RAW defeats that purpose

> is it ok to change the additional disk feature to create RAW images by default

My take on this would be to provide support for creation of additionalDisks with formatType. 

`limactl disk create --type raw`

If raw is used, it can be shared across drivers.

About default, my take would be to keep qcow2 as default as they have optimal storage.

balajiv113 (changes requested):

pendo324:
Thanks for the review, @balajiv113, I pushed a new revision to incorporate your comments.

> > Should Lima automatically convert QCOW2 disks to RAW before trying to mount
> 
> IMO The goal of additionalDisks are to share disk between instances doing conversion from QCOW2 - RAW defeats that purpose
> 
> > is it ok to change the additional disk feature to create RAW images by default
> 
> My take on this would be to provide support for creation of additionalDisks with formatType.
> 
> `limactl disk create --type raw`
> 
> If raw is used, it can be shared across drivers.
> 
> About default, my take would be to keep qcow2 as default as they have optimal storage.

While QCOW2 is better optimized for storage, I wonder if RAW as a default is still viable given that RAW images can also be sparse?

RAW images are a little harder to deal with, in that if you (for example) `cp` them, the _entire_ disk will be copied, even if its not all in use. This can be avoided by using something like `rsync -S`, but that's not exactly obvious.

The only reason I'm proposing this is for interoperability between QEMU and Virtualization.framework VMs, but its possible that Lima can just leave this for individual users to decide too.

Instead of automatic runtime conversion, I can:
- remove the runtime conversion code
- add the `--format raw` option to the `disk create` command
- add notes on how to do the conversion manually, if you want to share data between the two VM types

Let me know if that sounds better, thanks!

jandubois:
> IMO The goal of additionalDisks are to share disk between instances doing conversion from QCOW2 - RAW defeats that purpose

How does it defeat the purpose? The images would still work with QEMU.

I just tested this by switching an Alpine instance from QEMU to VZ, which converted the `diffdisk` to RAW format. I made some changes to the image, stopped the instance and switched it back to QEMU. The disk remained in RAW format, but the modified data was still there and everything seemed to work normally.

What is the functionality we would lose by switching to RAW format?

balajiv113:
@jandubois 

> doing conversion from QCOW2 - RAW defeats that purpose

With automatic conversion we will endup with 2 different disks (one with qcow2 format and other will be raw).

From vz, whatever we change that won't be reflected to qcow2 disk. 

As you said, we can mount qemu also with raw format in that case it is fine.  
Considering this, it would be better to decide on the disk format on disk create itself `limactl disk create --format raw` instead of doing auto conversion to raw on the fly

jandubois:
> With automatic conversion we will endup with 2 different disks (one with qcow2 format and other will be raw).

Oh, I didn't catch this. To me "automatic conversion" meant: the first time the disk is mounted to a VZ instance it will be **replaced** with the converted RAW disk, and from then on both QEMU and VZ instances would only use the RAW disk going forward. This should be transparent to the QEMU instances and guarantee data consistency because there is always only a single copy of the data.

I didn't look at the code in the PR yet, so I didn't realize this is not how it was implemented.

My main concern is about what happens if there is an error during the conversion (e.g. you run out of disk space). The old QCOW2 disk must remain in working state until the conversion has completed successfully.

### Storage size

I don't have any knowledge about what advantages QCOW2 would have for storage size. Given that both formats have support for sparse files, how is QCOW2 optimized for storage? Do you have any references I can read up on this? Any comparison on storage efficiency and access speed (e.g. if QCOW2 would store data in compressed format, I would expect the disk to be slower than RAW).

balajiv113:
> Oh, I didn't catch this

My bad, Am the culprit here. I overlooked and missed that the PR already takes care of moving QEMU disk to a new extension .qcow2. So it is as per you said only.

> formats have support for sparse files

Again my bad :D I was looking into `ls` which was giving 100G for raw disks and qcow was showing only the used space and this was misleading. But checked the disk utility now it is fine. 

@jandubois 
Thanks for the details :) Its mostly things missed from my end. The current approach of using raw format looks good to me

jandubois:
> I was looking into `ls`

This is a good point though: If you are only using QEMU, and you are not excluding the Lima VMs and disks from your backups, then you may end up backing up the full sparse file for a RAW disk if you have overallocated it significantly, and much of the space remains permanently unused.

So I think it makes sense to allocate additional disks as QCOW2 by default (but provide a `--raw` option to override), and auto-convert to RAW the first time it is mounted to a VZ instance. That way everyone automatically gets the best setup:

* Users only using QEMU will still use QCOW2
* Users using VZ (or both) will convert to RAW and have interoperability

So this would be my recommendation, as long as the auto-conversion to RAW is done in a way that any failure keeps the QCOW2 copy unharmed.

<hr>

PS: Thinking about this some more, I think we are already currently at risk of damaging the VM disk if we ever run out of host disk space while trying to commit additional sparse data blocks. I wonder if `limactl start` should display a warning when you start an instance and  the remaining free space is not enough to completely allocate the sparse space for the disk (plus an additional buffer of e.g. at least 1GB)?

pendo324:
Thank you for the great discussion. QCOW2 disks are definitely more user friendly, in that users can just use default/familiar tools like `cp` and `ls` to interact with them. Even if they can be sparse, that might not be obvious on first glance.

I updated the code to allow a new `--format` option for `disk create`, it defaults to `qcow2`. Right now we only really need qcow2 or RAW, but I figured it might be good to make it more extensible now rather than changing it later if needed. Let me know what you think.

I also added a bunch of information on how to best deal with sparse RAW files to the docs. Let me know if it looks good, or if I should move that info to another location.

jandubois:
@pendo324 On macOS you can use `clonefile (2)` (via `cp -c`) to copy a sparse file. No need to use GNU tools or `rsync`. It even shares the data blocks between the source and destination until either of them is modified (copy-on-write), and of course preserves the sparse-ness as well.

pendo324:
> @pendo324 On macOS you can use `clonefile (2)` (via `cp -c`) to copy a sparse file. No need to use GNU tools or `rsync`. It even shares the data blocks between the source and destination until either of them is modified (copy-on-write), and of course preserves the sparse-ness as well.

Oh wow, I couldn't find anything like that on macOS when I looked through the docs for `cp`. This is much better for users. Thanks for the info! I just updated the docs.

pkg/driver/driver.go
```diff
@@ -12,6 +12,8 @@ type Driver interface {
 
 	CreateDisk() error
 
+	InspectDisk(diskName string) (*store.Disk, error)
```
balajiv113, line 15: I don't think we need InspectDisk in driver.

As this is not doing anything driver specific. In the usage point we could directly call store.InspectDisk

pkg/vz/vm_darwin.go
```diff
@@ -334,6 +335,55 @@ func attachDisks(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfigura
...
+		extraDiskFormat, err := imgutil.DetectFormat(extraDiskPath)
+		if err != nil {
+			return fmt.Errorf("failed to run detect disk format %q: %q", diskName, err)
+		}
+		if extraDiskFormat != "raw" {
+			rawPath := fmt.Sprintf("%s.raw", extraDiskPath)
+			if err = imgutil.QCOWToRaw(extraDiskPath, rawPath); err != nil {
+				return fmt.Errorf("failed to convert qcow2 disk %q to raw for vz driver: %w", diskName, err)
+			}
+			cmd := exec.Command("mv", extraDiskPath, fmt.Sprintf("%s.qcow2", extraDiskPath))
```
balajiv113, line 363: We could use os Rename() instead of exec mv commands

pendo324: Good call, updated in latest revision

pkg/driver/driver.go
pendo324: Makes sense, I just updated it in the latest revision

pendo324:
> > I updated the code to allow a new `--format` option for `disk create`, it defaults to `qcow2`. Right now we only really need qcow2 or RAW, but I figured it might be good to make it more extensible now rather than changing it later if needed. Let me know what you think.
> 
> I used .vdi format for the VirtualBox driver, it should support raw as well - as long as you add a .img suffix to the file...

I can add vdi to the list of supported formats, or add all of the formats from [this list](https://qemu-project.gitlab.io/qemu/system/images.html#disk-image-file-formats). Might be more suitable in a separate PR though?

balajiv113 (approved):
LGTM 👍

balajiv113:
> I can add vdi to the list of supported formats

I don't think it is required as of now. 
Maybe once we support virtualbox driver then we can add the required format later in a separate PR

AkihiroSuda (dismissed):
Thanks

pendo324:
Hey @jandubois, please let me know if there's any further changes you think should be made to get this merged when you get a chance. Thanks!

jandubois:
@pendo324 I've been extremely busy the last 2 weeks, so I barely looked at any outstanding code reviews; sorry about that!

I've looked at the PR now once more, and there is one thing I don't get: After converting the qcow2 disk to raw format, you rename the old disk to `datadisk.qcow2`:

```console
$ ls -l ~/.lima/_disks/foo/
total 9360
-rw-r--r--@ 1 jan  staff  10737418240 17 Mar 21:36 datadisk
-rw-r--r--@ 1 jan  staff       196768 17 Mar 21:35 datadisk.qcow2
lrwxr-xr-x@ 1 jan  staff           24 17 Mar 21:35 in_use_by -> /Users/jan/.lima/default
```

What is the point of keeping the old disk around. It will never be accessed by Lima again, and there is no way delete it with any lima commands, except when you delete the raw disk with `limactl disk delete ...`. Why don't you want to delete it right away, to save the additional disk space (in case it was converted after data was already copied into it)?

<hr>

Probably unrelated to this PR: the first time I stopped the VM and tried to delete the disk, the `in_use_by` symlink was still in place. I don't understand how that could have happened, and also couldn't reproduce after force-deleting and re-creating the disk, and then starting/stopping the VM once more. Just in case anybody else has any ideas:

```console
$ l stop
INFO[0000] Sending SIGINT to hostagent process 205
INFO[0000] [hostagent] 2023/03/17 21:17:17 tcpproxy: for incoming conn 127.0.0.1:62994, error dialing "192.168.5.15:22": connect tcp 192.168.5.15:22: no route to host
INFO[0000] [hostagent] Stopping forwarding "/run/lima-guestagent.sock" (guest) to "/Users/jan/.lima/default/ga.sock" (host)
INFO[0000] [hostagent] Unmounting disk "foo"
INFO[0000] [hostagent] Unmounting "/Users/jan"
INFO[0000] [hostagent] Unmounting "/tmp/lima"
ERRO[0000] [hostagent] dhcp: unhandled message type: RELEASE
INFO[0000] [hostagent] [VZ] - vm state change: stopped

$ l disk delete foo
WARN[0000] Skipping deleting disk "foo", disk is referenced by one or more non-running instances: ["default"]
WARN[0000] To delete anyway, run "limactl disk delete --force foo"
```

pendo324:
> @pendo324 I've been extremely busy the last 2 weeks, so I barely looked at any outstanding code reviews; sorry about that!
> 

No problem, thanks for taking a look at it again 👍.

> I've looked at the PR now once more, and there is one thing I don't get: After converting the qcow2 disk to raw format, you rename the old disk to `datadisk.qcow2`:
> 
> ```
> $ ls -l ~/.lima/_disks/foo/
> total 9360
> -rw-r--r--@ 1 jan  staff  10737418240 17 Mar 21:36 datadisk
> -rw-r--r--@ 1 jan  staff       196768 17 Mar 21:35 datadisk.qcow2
> lrwxr-xr-x@ 1 jan  staff           24 17 Mar 21:35 in_use_by -> /Users/jan/.lima/default
> ```
> 
> What is the point of keeping the old disk around. It will never be accessed by Lima again, and there is no way delete it with any lima commands, except when you delete the raw disk with `limactl disk delete ...`. Why don't you want to delete it right away, to save the additional disk space (in case it was converted after data was already copied into it)?
> 

Good point, I just updated the PR to delete the old qcow2 disk after swapping the RAW disk to be named `datadisk`. I think only removing the qcow2 disk after the RAW disk is in a usable state makes it more of a transaction, and ensures that at least one disk is usable (or at least recoverable) if the rename / deletion somehow fails. Let me know what you think. The alternative is to just delete the qcow2 disk as soon as the RAW disk is created, I wouldn't be opposed to that either.

----

> Probably unrelated to this PR: the first time I stopped the VM and tried to delete the disk, the `in_use_by` symlink was still in place. I don't understand how that could have happened, and also couldn't reproduce after force-deleting and re-creating the disk, and then starting/stopping the VM once more. Just in case anybody else has any ideas:
> 
> ```
> $ l stop
> INFO[0000] Sending SIGINT to hostagent process 205
> INFO[0000] Waiting for the host agent and the driver processes to shut down
> INFO[0000] [hostagent] 2023/03/17 21:17:17 tcpproxy: for incoming conn 127.0.0.1:62994, error dialing "192.168.5.15:22": connect tcp 192.168.5.15:22: no route to host
> INFO[0000] [hostagent] Received SIGINT, shutting down the host agent
> INFO[0000] [hostagent] Shutting down the host agent
> INFO[0000] [hostagent] Stopping forwarding "/run/lima-guestagent.sock" (guest) to "/Users/jan/.lima/default/ga.sock" (host)
> INFO[0000] [hostagent] Unmounting disk "foo"
> INFO[0000] [hostagent] Unmounting "/Users/jan"
> INFO[0000] [hostagent] Unmounting "/tmp/lima"
> INFO[0000] [hostagent] Shutting down VZ
> ERRO[0000] [hostagent] dhcp: unhandled message type: RELEASE
> INFO[0000] [hostagent] [VZ] - vm state change: stopped
> 
> $ l disk delete foo
> WARN[0000] Skipping deleting disk "foo", disk is referenced by one or more non-running instances: ["default"]
> WARN[0000] To delete anyway, run "limactl disk delete --force foo"
> ```

Not sure why this would happen, I've never seen that before, but I'll try to look into it.

pkg/vz/vm_darwin.go
```diff
@@ -334,6 +335,58 @@ func attachDisks(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfigura
...
+		extraDiskFormat, err := imgutil.DetectFormat(extraDiskPath)
+		if err != nil {
+			return fmt.Errorf("failed to run detect disk format %q: %q", diskName, err)
+		}
+		if extraDiskFormat != "raw" {
+			rawPath := fmt.Sprintf("%s.raw", extraDiskPath)
+			qcow2Path := fmt.Sprintf("%s.qcow2", extraDiskPath)
+			if err = imgutil.QCOWToRaw(extraDiskPath, rawPath); err != nil {
+				return fmt.Errorf("failed to convert qcow2 disk %q to raw for vz driver: %w", diskName, err)
+			}
```
jandubois, line 363: Technically we don't know that the current format is qcow2, except for the fact that `limactl disk create` currently only creates either `raw` or `qcow2` formatted volumes. So this code could be kept more generic, like this:

```suggestion
			oldFormatPath := fmt.Sprintf("%s.%s", extraDiskPath, extraDiskFormat)
			if err = imgutil.ConvertToRaw(extraDiskPath, rawPath); err != nil {
				return fmt.Errorf("failed to convert %s disk %q to raw for vz driver: %w", extraDiskFormat, diskName, err)
			}
			...
```

jandubois (approved):
Thanks, LGTM.

I have one comment, but I'm also fine with merging the PR as-is.

pkg/vz/vm_darwin.go
pendo324: Good point, create a new issue to track this, I'll implement it after we get this merged if that's alright https://github.com/lima-vm/lima/issues/1438

PR #1409, opened 2023-03-08
Make cache pruning more flexible

jay7x:
This command deletes only those cached downloads which are not referenced by any existing VM configs.

### DIFF
--- a/cmd/limactl/prune.go
+++ b/cmd/limactl/prune.go
@@ -21,12 +21,30 @@ func newPruneCommand() *cobra.Command {
 		ValidArgsFunction: cobra.NoFileCompletions,
 	}
 
+	pruneCommand.Flags().Bool("dry-run", false, "Show which images would be deleted, but don't really delete them")
 	pruneCommand.Flags().Bool("no-digest-only", false, "Only prune images without a digest specified (\"fallback\" images usually)")
 	pruneCommand.Flags().Bool("unreferenced-only", false, "Only prune downloads not referenced in any VM")
 	return pruneCommand
 }
 
+func deletePathMaybe(path string, isDryRun bool) error {
+	if isDryRun {
+		return nil
+	}
+	err := os.RemoveAll(path)
+	if err != nil {
+		logrus.WithError(err).WithFields(logrus.Fields{
+			"path": path,
+		}).Errorf("Cannot delete directory. Skipping...")
+	}
+	return err
+}
+
 func pruneAction(cmd *cobra.Command, args []string) error {
+	pruneDryRun, err := cmd.Flags().GetBool("dry-run")
+	if err != nil {
+		return err
+	}
 	pruneWithoutDigest, err := cmd.Flags().GetBool("no-digest-only")
 	if err != nil {
 		return err
@@ -55,34 +73,28 @@ func pruneAction(cmd *cobra.Command, args []string) error {
 
 			logrus.WithFields(entryFields).Debug("cache entry")
 
+			if pruneDryRun {
+				entryFields["is_dry_run"] = pruneDryRun
+			}
+
 			// check if the cache entry is referenced
 			if refFile, refFound := files[entry.ID]; refFound {
 				if refFile.Location != entry.Location { // sanity check
-					logrus.WithFields(logrus.Fields{
-						"id":                  entry.ID,
-						"location":            entry.Location,
-						"referenced_location": refFile.Location,
-					}).Warnf("Sanity check failed! URL mismatch")
+					ef := entryFields
+					ef["referenced_location"] = refFile.Location
+					logrus.WithFields(ef).Warnf("Sanity check failed! URL mismatch")
 				}
 
 				if pruneWithoutDigest && refFile.Digest == "" {
 					// delete the fallback image entry (entry w/o digest) even if referenced
 					logrus.WithFields(entryFields).Infof("Deleting fallback entry")
-					if err := os.RemoveAll(entry.Path); err != nil {
-						logrus.WithError(err).WithFields(logrus.Fields{
-							"path": entry.Path,
-						}).Errorf("Cannot delete directory. Skipping...")
-					}
+					deletePathMaybe(entry.Path, pruneDryRun)
 				}
 			} else {
 				if pruneUnreferenced {
 					// delete the unreferenced cached entry
 					logrus.WithFields(entryFields).Infof("Deleting unreferenced entry")
-					if err := os.RemoveAll(entry.Path); err != nil {
-						logrus.WithError(err).WithFields(logrus.Fields{
-							"path": entry.Path,
-						}).Errorf("Cannot delete directory. Skipping...")
-					}
+					deletePathMaybe(entry.Path, pruneDryRun)
 				}
 			}
 		}
@@ -95,8 +107,11 @@ func pruneAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	cacheDir := filepath.Join(ucd, "lima")
-	logrus.Infof("Pruning everything in %q", cacheDir)
-	return os.RemoveAll(cacheDir)
+	logrus.WithFields(logrus.Fields{
+		"is_dry_run": pruneDryRun,
+		"location":   cacheDir,
+	}).Infof("Pruning entire cache")
+	return deletePathMaybe(cacheDir, pruneDryRun)
 }
 
 // Collect all downloads referenced in VM definitions and templates
--- a/cmd/limactl/prune.go
+++ b/cmd/limactl/prune.go
@@ -4,6 +4,10 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/lima-vm/lima/pkg/downloader"
+	"github.com/lima-vm/lima/pkg/limayaml"
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/opencontainers/go-digest"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
@@ -16,15 +20,101 @@ func newPruneCommand() *cobra.Command {
 		RunE:              pruneAction,
 		ValidArgsFunction: cobra.NoFileCompletions,
 	}
+
+	pruneCommand.Flags().Bool("no-digest-only", false, "Only prune images without a digest specified (\"fallback\" images usually)")
+	pruneCommand.Flags().Bool("unreferenced-only", false, "Only prune downloads not referenced in any VM")
 	return pruneCommand
 }
 
 func pruneAction(cmd *cobra.Command, args []string) error {
+	pruneWithoutDigest, err := cmd.Flags().GetBool("no-digest-only")
+	if err != nil {
+		return err
+	}
+	pruneUnreferenced, err := cmd.Flags().GetBool("unreferenced-only")
+	if err != nil {
+		return err
+	}
+
+	if pruneWithoutDigest || pruneUnreferenced {
+		files, err := getReferencedDownloads()
+		if err != nil {
+			return err
+		}
+
+		cacheEntries, err := downloader.CachedDownloads(downloader.WithCache())
+		if err != nil {
+			return err
+		}
+
+		for _, entry := range cacheEntries {
+			entryFields := logrus.Fields{
+				"id":       entry.ID,
+				"location": entry.Location,
+			}
+
+			logrus.WithFields(entryFields).Debug("cache entry")
+
+			// check if the cache entry is referenced
+			if refFile, refFound := files[entry.ID]; refFound {
+				if refFile.Location != entry.Location { // sanity check
+					logrus.WithFields(logrus.Fields{
+						"id":                  entry.ID,
+						"location":            entry.Location,
+						"referenced_location": refFile.Location,
+					}).Warnf("Sanity check failed! URL mismatch")
+				}
+
+				if pruneWithoutDigest && refFile.Digest == "" {
+					// delete the fallback image entry (entry w/o digest) even if referenced
+					logrus.WithFields(entryFields).Infof("Deleting fallback entry")
+					if err := os.RemoveAll(entry.Path); err != nil {
+						logrus.WithError(err).WithFields(logrus.Fields{
+							"path": entry.Path,
+						}).Errorf("Cannot delete directory. Skipping...")
+					}
+				}
+			} else {
+				if pruneUnreferenced {
+					// delete the unreferenced cached entry
+					logrus.WithFields(entryFields).Infof("Deleting unreferenced entry")
+					if err := os.RemoveAll(entry.Path); err != nil {
+						logrus.WithError(err).WithFields(logrus.Fields{
+							"path": entry.Path,
+						}).Errorf("Cannot delete directory. Skipping...")
+					}
+				}
+			}
+		}
+		return nil
+	}
+
+	// prune everything if no options specified
 	ucd, err := os.UserCacheDir()
 	if err != nil {
 		return err
 	}
 	cacheDir := filepath.Join(ucd, "lima")
-	logrus.Infof("Pruning %q", cacheDir)
+	logrus.Infof("Pruning everything in %q", cacheDir)
 	return os.RemoveAll(cacheDir)
+}
+
+// Collect all downloads referenced in VM definitions and templates
+func getReferencedDownloads() (map[string]limayaml.File, error) {
+	digests := make(map[string]limayaml.File)
+
+	vmRefs, err := store.Downloads()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, f := range vmRefs {
+		d := digest.SHA256.FromString(f.Location).Encoded()
+		logrus.WithFields(logrus.Fields{
+			"id":       d,
+			"location": f.Location,
+		}).Debugf("referenced file")
+		digests[d] = f
+	}
+	return digests, nil
 }
--- a/pkg/downloader/downloader.go
+++ b/pkg/downloader/downloader.go
@@ -39,6 +39,13 @@ type Result struct {
 	ValidatedDigest bool
 }
 
+type CacheEntry struct {
+	ID       string        // sha256 of the Location
+	Location string        // source URL of the entry
+	Path     string        // Absolute path of the entry
+	Digest   digest.Digest // Checksum of the entry's data (a file downloaded)
+}
+
 type options struct {
 	cacheDir       string // default: empty (disables caching)
 	decompress     bool   // default: false (keep compression)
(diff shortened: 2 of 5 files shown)


afbjorklund:
Should we maybe keep also the images that are included in the current template versions ?

jay7x:
> Should we maybe keep also the images that are included in the current template versions, not only instances ?

I have no strong opinion on this.. my case is to cleanup images I've downloaded for some tests and which are not used (doesn't exists) anymore..

afbjorklund:
I added an issue for the discussion, if we want to keep this PR for the implementation review

cmd/limactl/prune.go
```diff
@@ -1,30 +1,172 @@
...
+	logrus.Infof("Pruning everything in %q", cacheDir)
 	return os.RemoveAll(cacheDir)
 }
+
+func newPruneDownloadsCommand() *cobra.Command {
+	pruneDownloadsCommand := &cobra.Command{
+		Use:   "downloads",
+		Short: "Prune unreferenced downloads",
+		RunE:  pruneDownloadsAction,
+	}
```
AkihiroSuda, line 54: How is "download" different from "cache"?

cmd/limactl/prune.go
```diff
@@ -1,30 +1,172 @@
...
+		}
+		for _, archive := range instance.Config.Containerd.Archives {
+			digests[makeDigest(archive.Location)] = archive.Location
+		}
+	}
+	return digests, nil
+}
+
+// Collect all cached downloads
+func getCachedDigests(downloadsDir string) (map[string]string, error) {
```
AkihiroSuda, line 146: Should be in `pkg/downloader`, with unit tests

cmd/limactl/prune.go
```diff
@@ -1,30 +1,172 @@
...
+	}
+	return filepath.Join(ucd, "lima"), nil
+}
+
+func makeDigest(s string) string {
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(s)))
+}
+
+// Collect all downloads referenced in VM definitions
+func getReferencedDigests() (map[string]string, error) {
```
AkihiroSuda, line 118: Should be in `pkg/store`

cmd/limactl/prune.go
```diff
@@ -1,30 +1,172 @@
...
-	cacheDir := filepath.Join(ucd, "lima")
-	logrus.Infof("Pruning %q", cacheDir)
+	logrus.Infof("Pruning everything in %q", cacheDir)
 	return os.RemoveAll(cacheDir)
 }
+
+func newPruneDownloadsCommand() *cobra.Command {
+	pruneDownloadsCommand := &cobra.Command{
+		Use:   "downloads",
+		Short: "Prune unreferenced downloads",
```
AkihiroSuda, line 52: I'm not sure about the usecase of this.
"Referenced" blobs are copied to `~/.lima/<INSTANCE>/basedisk` and can be removed safely.

cmd/limactl/prune.go
jay7x: We don't have it YET :) but why do we have "download" directory under the cache then? ;)

cmd/limactl/prune.go
jay7x: it's about download speed mostly. I decided to implement this while staying in Medan city in Indonesia because every time I need to spawn another image I should sit and wait for ~30-60min to download 700-1500Mb. So pruning the whole cache after set of experiments was not really good idea.

Though I agree we can move this one CLI level up. I.e. change the default `lima prune` behavior if downloads are the only cached objects we have.

AkihiroSuda: Wondering something like `limactl prune --keep="3 days"` may make more sense.

Images that have been used within 3 days will be kept in the cache, regardless to whether it is actually used by any active instance, so that you can avoid downloading when you are going to create an instance from it.

jay7x: Ah, right.. I was thinking about TTL yesterday too.. let's bring this to discussion in #1410

jay7x:
I've simplified things and removed the `prune` subcommands. New CLI switches added instead:
```
  -F, --fallback       Only prune images without a digest specified (fallback images usually)
  -U, --unreferenced   Only prune downloads not referenced in any VM or template
```

cmd/limactl/prune.go
```diff
@@ -16,15 +22,97 @@ func newPruneCommand() *cobra.Command {
 		RunE:              pruneAction,
 		ValidArgsFunction: cobra.NoFileCompletions,
 	}
+
+	pruneCommand.Flags().BoolP("fallback", "F", false, "Only prune images without a digest specified (fallback images usually)")
+	pruneCommand.Flags().BoolP("unreferenced", "U", false, "Only prune downloads not referenced in any VM or template")
```
AkihiroSuda, line 27:
```suggestion
	pruneCommand.Flags().Bool("unreferenced-only", false, "Only prune downloads not referenced in any VM")
```
Blobs referred by templates are not worth keeping

cmd/limactl/prune.go
```diff
@@ -16,15 +22,97 @@ func newPruneCommand() *cobra.Command {
 		RunE:              pruneAction,
 		ValidArgsFunction: cobra.NoFileCompletions,
 	}
+
+	pruneCommand.Flags().BoolP("fallback", "F", false, "Only prune images without a digest specified (fallback images usually)")
```
AkihiroSuda, line 26:
```suggestion
	pruneCommand.Flags().Bool("no-digest-only", false, "Only prune images without a digest specified (\"fallback\" images usually)")
```

cmd/limactl/prune.go
```diff
@@ -16,15 +22,97 @@ func newPruneCommand() *cobra.Command {
...
 	}
 	cacheDir := filepath.Join(ucd, "lima")
-	logrus.Infof("Pruning %q", cacheDir)
+	logrus.Infof("Pruning everything in %q", cacheDir)
 	return os.RemoveAll(cacheDir)
 }
+
+func makeDigest(s string) string {
+	return fmt.Sprintf("%x", sha256.Sum256([]byte(s)))
+}
```
AkihiroSuda, line 95: go-digest might be better

pkg/downloader/downloader.go
```diff
@@ -385,3 +392,86 @@ func downloadHTTP(localPath, url string, expectedDigest digest.Digest) error {
...
+
+// Read the file and return its contents as a string
+// Never use it with large files
+func readSmallTextFile(filename string) (string, error) {
+	data, err := os.ReadFile(filename)
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
+}
```
AkihiroSuda, line 404: Not worth defining as a new function

pkg/downloader/downloader.go
```diff
@@ -385,3 +392,86 @@ func downloadHTTP(localPath, url string, expectedDigest digest.Digest) error {
...
+
+// Read first *.digest file found and return the Digest object from its contents
+func guessDigestFromDir(dir string) (digest.Digest, error) {
+	dirList, err := os.ReadDir(dir)
+	if err != nil {
+		return "", err
+	}
+
+	for _, f := range dirList {
+		if strings.HasSuffix(f.Name(), ".digest") && f.Type().IsRegular() {
```
AkihiroSuda, line 414: Why only regular?

pkg/downloader/downloader.go
```diff
@@ -36,6 +36,13 @@ type Result struct {
 	ValidatedDigest bool
 }
 
+type CacheEntry struct {
+	ID       string        // sha256 of the Location
+	Location string        // source URL of the entry
+	Path     string        // Absolute path of the entry
+	Digest   digest.Digest // Checksum of the entry's data
```
AkihiroSuda, line 43: What data?

pkg/store/store.go
```diff
@@ -102,3 +102,32 @@ func LoadYAMLByFilePath(filePath string) (*limayaml.LimaYAML, error) {
 	}
 	return y, nil
 }
+
+func Downloads() ([]limayaml.File, error) {
```
AkihiroSuda, line 106: Probably `limayaml.YAML` object should have `ReferredFiles() ([]limayaml.File, error)` method

cmd/limactl/prune.go
jay7x: I agree.. but then this line should be changed too: https://github.com/lima-vm/lima/blob/master/pkg/downloader/downloader.go#L167

jay7x: Not sure about this one: https://github.com/lima-vm/lima/blob/master/pkg/limayaml/defaults.go#L53

pkg/downloader/downloader.go
jay7x: But it's called from 3 places in my code.. Could you please suggest what is the solution then?

pkg/downloader/downloader.go
jay7x: Well.. mainly because caching code stores it as a regular file. I can replace with `!( f.IsDir() )` if you'd prefer though.

pkg/downloader/downloader.go
jay7x: This is about the cached data which is stored under the `data` file here: https://github.com/lima-vm/lima/blob/master/pkg/downloader/downloader.go#L168
Could you suggest a better comment maybe?

pkg/store/store.go
jay7x: Ok, let me move it there, np.

cmd/limactl/prune.go
jay7x: That was my opinion too.. but there are ppl who wants this: https://github.com/lima-vm/lima/issues/1410#issuecomment-1461496682

jandubois: Why would you not want to keep images referenced in templates. You will only have images in the cache that you have used before, so it is not unreasonable that you want to use them again.

If you stop using a specific template, eventually the template will be updated to refer to a newer version of the image, and the old cached version will then become unreferenced.

So just because I currently don't have a fedora instance doesn't mean I want to prune the current fedora image. I only want to prune the old versions of fedora that are not being used by templates anymore, as it is now unlikely that I will ever use them again.

pkg/downloader/downloader.go
jay7x: Ah.. you're right actually.. there are just 2 invocations left and I think it should be easy to deal with

cmd/limactl/prune.go
jay7x: I dropped the `makeDigest()` function as it was used just in single place after the refactoring. Also `sha256.sha256()` is replaced with the `digest` module functions chain. Though I'm not sure if downloader.go:167 should be replaced too.

cmd/limactl/prune.go
jay7x: As I understand there is no easy way to calculate the age of an image w/o taking special actions..

jay7x: Sorry, I meant something other actually.. it's easy to get the age of an image for sure.. but I'd say users would like to keep recenlty *used* images

AkihiroSuda: > to keep recenlty used images..

`st_atime` isn't enough?

jay7x: Well.. on MacOS it might work but it's almost always disabled on Linux systems. Not sure how many Linux users we have 🤔

jay7x: @AkihiroSuda @jandubois How about removing the template references part and merge this PR after? We can discuss another improvements separately then.

cmd/limactl/prune.go
jay7x: @AkihiroSuda please let me know if I should update the https://github.com/lima-vm/lima/blob/master/pkg/downloader/downloader.go#L180 to use `digest` too.. I'm a bit worrying about inconsistency possible..

jay7x:
Updates:
1. Downloads referenced by a template are dropped from the PR scope
2. Changes are rebased on the current master branch
3. Commits are squashed
4. Few small improvements (better logging, use `digest` instead of sprintf()'ed sha256)

cmd/limactl/prune.go
jay7x: Downloads referenced by a template are dropped from the PR scope.

pkg/downloader/downloader.go
jay7x: Comment is improved

cmd/limactl/prune.go
jay7x: New code is using `digest` instead of sprintf()'ed sha256. Old code in downloader.go still uses the sha256 call.

AkihiroSuda (dismissed):
Thanks

jandubois:
I think this PR is good, but have run out of time right now (will finish tomorrow).

I found it useful to add a `--dry-run` option for testing, and I think it may be generally useful, so please consider adding it (but can also be done later in another PR):

```diff
--- cmd/limactl/prune.go
+++ cmd/limactl/prune.go
@@ -21,12 +21,17 @@ func newPruneCommand() *cobra.Command {
                ValidArgsFunction: cobra.NoFileCompletions,
        }

+       pruneCommand.Flags().Bool("dry-run", false, "Show which images would be deleted, but don't really delete them")
        pruneCommand.Flags().Bool("no-digest-only", false, "Only prune images without a digest specified (\"fallback\" images usually)")
        pruneCommand.Flags().Bool("unreferenced-only", false, "Only prune downloads not referenced in any VM")
        return pruneCommand
 }

 func pruneAction(cmd *cobra.Command, args []string) error {
+       dryRun, err := cmd.Flags().GetBool("dry-run")
+       if err != nil {
+               return err
+       }
        pruneWithoutDigest, err := cmd.Flags().GetBool("no-digest-only")
        if err != nil {
                return err
@@ -68,20 +73,24 @@ func pruneAction(cmd *cobra.Command, args []string) error {
                                if pruneWithoutDigest && refFile.Digest == "" {
                                        // delete the fallback image entry (entry w/o digest) even if referenced
                                        logrus.WithFields(entryFields).Infof("Deleting fallback entry")
-                                       if err := os.RemoveAll(entry.Path); err != nil {
-                                               logrus.WithError(err).WithFields(logrus.Fields{
-                                                       "path": entry.Path,
-                                               }).Errorf("Cannot delete directory. Skipping...")
+                                       if !dryRun {
+                                               if err := os.RemoveAll(entry.Path); err != nil {
+                                                       logrus.WithError(err).WithFields(logrus.Fields{
+                                                               "path": entry.Path,
+                                                       }).Errorf("Cannot delete directory. Skipping...")
+                                               }
                                        }
                                }
                        } else {
                                if pruneUnreferenced {
                                        // delete the unreferenced cached entry
                                        logrus.WithFields(entryFields).Infof("Deleting unreferenced entry")
-                                       if err := os.RemoveAll(entry.Path); err != nil {
-                                               logrus.WithError(err).WithFields(logrus.Fields{
-                                                       "path": entry.Path,
-                                               }).Errorf("Cannot delete directory. Skipping...")
+                                       if !dryRun {
+                                               if err := os.RemoveAll(entry.Path); err != nil {
+                                                       logrus.WithError(err).WithFields(logrus.Fields{
+                                                               "path": entry.Path,
+                                                       }).Errorf("Cannot delete directory. Skipping...")
+                                               }
                                        }
                                }
                        }
@@ -96,6 +105,9 @@ func pruneAction(cmd *cobra.Command, args []string) error {
        }
        cacheDir := filepath.Join(ucd, "lima")
        logrus.Infof("Pruning everything in %q", cacheDir)
+       if dryRun {
+               return nil
+       }
        return os.RemoveAll(cacheDir)
 }
```

jay7x:
If there is no release expected in 2 next days I'd like to add dry run because I like the idea. I'll add it tomorrow.

jandubois (changes requested):
> * Downloads referenced by a template are dropped from the PR scope

I somehow missed this. What is the reason for that?

I just noticed that when I deleted all my current instances, then `limactl prune --unreferenced-only` will delete **everything** in the cache (including the current `nerdctl-full` archives).

That seems like a major breakage.

If this is the functionality you want, then you should add a different option name for it. Like `--all-except-currently-in-use` (but a better name).

jandubois:
Just for illustration, assuming my cache looks like this:

```
653M	https://cloud-images.ubuntu.com/releases/22.04/release-20230302/ubuntu-22.04-server-cloudimg-amd64.img
654M	https://cloud-images.ubuntu.com/releases/22.04/release-20230518/ubuntu-22.04-server-cloudimg-amd64.img
727M	https://cloud-images.ubuntu.com/releases/22.10/release-20230413/ubuntu-22.10-server-cloudimg-amd64.img
779M	https://cloud-images.ubuntu.com/releases/23.04/release-20230502/ubuntu-23.04-server-cloudimg-amd64.img
233M	https://github.com/containerd/nerdctl/releases/download/v1.3.0/nerdctl-full-1.3.0-linux-amd64.tar.gz
233M	https://github.com/containerd/nerdctl/releases/download/v1.3.1/nerdctl-full-1.3.1-linux-amd64.tar.gz
238M	https://github.com/containerd/nerdctl/releases/download/v1.4.0/nerdctl-full-1.4.0-linux-amd64.tar.gz
 66M	https://github.com/lima-vm/alpine-lima/releases/download/v0.2.28/alpine-lima-std-3.18.0-x86_64.iso
 66M	https://github.com/lima-vm/alpine-lima/releases/download/v0.2.30/alpine-lima-std-3.18.0-x86_64.iso
```

I would expect `limactl prune --unreferenced-only` to leave these files behind, even if I delete all my current instances:

```
727M	https://cloud-images.ubuntu.com/releases/22.10/release-20230413/ubuntu-22.10-server-cloudimg-amd64.img
238M	https://github.com/containerd/nerdctl/releases/download/v1.4.0/nerdctl-full-1.4.0-linux-amd64.tar.gz
 66M	https://github.com/lima-vm/alpine-lima/releases/download/v0.2.30/alpine-lima-std-3.18.0-x86_64.iso
```

I guess for backwards compatibility we need to keep the existing behaviour, but I feel like we should require an `--all` option to wipe the whole cache, and have something safer, like `--unreferenced-only` the default.

jandubois:
I'm running out of time again, but I also think the `--no-digest-only` implementation doesn't work:

```
$ ls -l ~/Library/Caches/lima/download/by-url-sha256/791040918a29bb23fd0bc4bd074e32891299ffcdc0d0d20cd76a4af9593b0fa4
total 1503112
-rw-r--r--  1 jan  staff  769589248 25 May 22:59 data
-rw-r--r--  1 jan  staff         93 25 May 22:57 url
$ limactl ls
WARN[0000] No instance found. Run `limactl start` to create an instance.
$ limactl prune --dry-run --no-digest-only
[no output at all]
```

Maybe I made a mistake, will need to take a look again on the weekend.

AkihiroSuda:
Let me move this to v0.17

jay7x:
> > * Downloads referenced by a template are dropped from the PR scope
> 
> I somehow missed this. What is the reason for that?

The reason was to remove the point of the conflict and merge things which everyone is ok with (except @AkihiroSuda maybe 😄 )

> I just noticed that when I deleted all my current instances, then `limactl prune --unreferenced-only` will delete **everything** in the cache (including the current `nerdctl-full` archives).

This is expected as there is nothing referencing the downloads. Which is the default behavior of `lima prune` before this PR.

jay7x:
> I'm running out of time again, but I also think the `--no-digest-only` implementation doesn't work:

Well.. you're right :( I just found it's broken.. I've over-optimized things a bit.. let me fix it

AkihiroSuda:
> > I'm running out of time again, but I also think the `--no-digest-only` implementation doesn't work:
> 
> Well.. you're right :( I just found it's broken.. I've over-optimized things a bit.. let me fix it

Do you plan to update this PR?

jay7x:
> > > I'm running out of time again, but I also think the `--no-digest-only` implementation doesn't work:
> > 
> > 
> > Well.. you're right :( I just found it's broken.. I've over-optimized things a bit.. let me fix it
> 
> Do you plan to update this PR?

Yes, I'll continue to work on this later. @jandubois created pretty good definition of the process in #1409. My plan is to follow it. Though this requires some thinking.. Let me change the PR to draft again.

jay7x:
I found that I'm not needed this anymore (I can use my local set of images if needed). Closing this.

PR #1412, opened 2023-03-08
Add builtin yq support to limactl edit command

afbjorklund:
Also add` --tty` parameter, similar to the start command, for avoiding user interaction asking whether to start it.

```console
$ limactl edit --tty=false --set ".cpus = 2" default
INFO[0000] Instance "default" configuration edited      
$ limactl edit --tty=false --set ".cpus = 2" default
INFO[0000] Aborting, no changes made to the instance  
```

Follow-up to:
* #1359

### DIFF
--- a/cmd/limactl/edit.go
+++ b/cmd/limactl/edit.go
@@ -14,6 +14,8 @@ import (
 	"github.com/lima-vm/lima/pkg/start"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/lima-vm/lima/pkg/store/filenames"
+	"github.com/lima-vm/lima/pkg/yqutil"
+	"github.com/mattn/go-isatty"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
@@ -26,6 +28,9 @@ func newEditCommand() *cobra.Command {
 		RunE:              editAction,
 		ValidArgsFunction: editBashComplete,
 	}
+	// TODO: "survey" does not support using cygwin terminal on windows yet
+	editCommand.Flags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "enable TUI interactions such as opening an editor, defaults to true when stdout is a terminal")
+	editCommand.Flags().String("set", "", "modify the template inplace, using yq syntax")
 	return editCommand
 }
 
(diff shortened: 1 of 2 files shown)


cmd/limactl/edit.go
```diff
@@ -53,14 +58,30 @@ func editAction(cmd *cobra.Command, args []string) error {
...
+		}
+	} else {
+		hdr := fmt.Sprintf("# Please edit the following configuration for Lima instance %q\n", instName)
+		hdr += "# and an empty file will abort the edit.\n"
+		hdr += "\n"
+		hdr += editutil.GenerateEditorWarningHeader()
+		yBytes, err = editutil.OpenEditor(instName, yContent, hdr)
+		if err != nil {
+			return err
+		}
```
AkihiroSuda, line 83: Needs to check `tty` before opening an editor

afbjorklund: Not necessarily, that would depend on the editor ?

afbjorklund: Unfortunately you can't give a command (like yq) as EDITOR, but you do get an error message if it doesn't work

```
$ _output/bin/limactl --debug edit default </dev/null
DEBU[0000] opening editor "/usr/bin/editor" for a file "/tmp/lima-editor-996832437" 
Too many errors from stdin
FATA[0000] could not execute editor "/usr/bin/editor" for a file "/tmp/lima-editor-996832437": exit status 1 
```

AkihiroSuda: When `--tty` is true but the tty is not available, it should fail with a human-readable error

AkihiroSuda: @afbjorklund WDYT?

afbjorklund: I will take a look at what `start` is doing, it has the same strange connection between tty and editor

```
      --tty                enable TUI interactions such as opening an editor, defaults to true when stdout is a terminal (default true)
```

afbjorklund: Now it will only open the editor if a tty is present, but will not ask to start without a tty.

The user can run `limactl start` after `limactl edit` explicitly, if the tty is not available.

cmd/limactl/edit.go
```diff
@@ -53,17 +58,37 @@ func editAction(cmd *cobra.Command, args []string) error {
...
-	if len(yBytes) == 0 {
-		logrus.Info("Aborting, as requested by saving the file with empty content")
-		return nil
+	yq, err := cmd.Flags().GetString("set")
+	if err != nil {
+		return err
+	}
+	var yBytes []byte
+	if yq != "" {
+		logrus.Warn("`--set` is experimental")
```
AkihiroSuda, line 71: https://github.com/lima-vm/lima/blob/master/docs/experimental.md has to be updated too

jandubois:
I think `limactl edit --set ...` should not invoke the editor. I can't think of any situation where I want to set some settings via `--set`, but also want to use the editor as well; I would simply not bother writing the `--set` options in the first place.

If you ever wanted to do both (which I believe would be extremely uncommon, but maybe from a script), you could still call the command twice:

```
limactl edit --tty=false --set ".cpus = 3" default
limactl edit default
```

Even when `--set` is specified, we should use the `--tty` setting to determine if we should ask to start the instance or not.

afbjorklund:
Hmm, for `start` we open the editor after --set ? That was the behaviour that was copied here.

jandubois:
> for `start` we open the editor after --set ?

No, we don't. We **ask** if the user wants to edit the YAML file. But I see the point that maybe that shouldn't happen either. But that should be a separate issue.

For `edit` we always open the editor right away (because that is the point of the command), so it is more jarring that I have to close the editor again even if I didn't want to open it (because I used `--set`). And if I specify `--tty=false` to prevent it, then it won't ask me to start the VM, and I have to start it manually as well.

cmd/limactl/edit.go
```diff
@@ -53,14 +58,31 @@ func editAction(cmd *cobra.Command, args []string) error {
...
+		yBytes, err = yqutil.EvaluateExpression(yq, yContent)
+		if err != nil {
+			return err
+		}
+	} else if tty {
+		hdr := fmt.Sprintf("# Please edit the following configuration for Lima instance %q\n", instName)
+		hdr += "# and an empty file will abort the edit.\n"
+		hdr += "\n"
+		hdr += editutil.GenerateEditorWarningHeader()
+		yBytes, err = editutil.OpenEditor(instName, yBytes, hdr)
```
jandubois, line 81: `yBytes` is empty; we need to pass `yContent`:

```suggestion
		yBytes, err = editutil.OpenEditor(instName, yContent, hdr)
```

jandubois (changes requested):
Works as expected now (doesn't invoke editor when using `--set`), but is not showing the `lima.yaml` content in the editor anymore.

cmd/limactl/edit.go
afbjorklund: My bad, it got lost when I redid/restructured. Previously the --set still worked with the editor

jandubois (approved):
Thanks, LGTM

PR #1413, opened 2023-03-10
feat: allow rosetta fields to be overridden

pendo324:
Allow Rosetta fields to be overridden by making them pointers and setting default values based the same way they are set for other config options.


jandubois (approved):
Thanks, LGTM

Issue #1422, opened 2023-03-17
Support offline creation of machine instance


afbjorklund:
### Description

Currently the baseimage and the nerdctl-full can be cached, so that they don't have to be downloaded.

But then there are still some packages missing, which means that the machine needs Internet access.

It would be nice if there was a way to cache also these files, perhaps using something like `repro-get` ?

So that one could start up an instance without external network, after making sure everything is cached.

-----

These are cached:

```
INFO[0001] Attempting to download the image from "https://cloud-images.ubuntu.com/releases/22.10/release-20230215/ubuntu-22.10-server-cloudimg-amd64.img"  digest="sha256:5e5c68cb12002111032d3239ade3763ce25639f1287a59d2509a1603c2b1f7e6"
INFO[0001] Using cache "/home/anders/.cache/lima/download/by-url-sha256/7db6503d427a306824b1f62ce99d3a19a4c83f3b69af39022477aecca5c69b33/data" 
INFO[0001] Attempting to download the nerdctl archive from "https://github.com/containerd/nerdctl/releases/download/v1.2.1/nerdctl-full-1.2.1-linux-amd64.tar.gz"  digest="sha256:e8a3e40d442c566ee494375a4c563121da69d7c7837f50f4a3a171742757b36c"
INFO[0002] Using cache "/home/anders/.cache/lima/download/by-url-sha256/3ca338b148379fea1376fb450380cb21a2b26d8b1618da6eb1c5c5e352b03ea7/data" 
```

But these are not:

```
[   26.224771] cloud-init[2091]: + apt-get install -y --no-upgrade --no-install-recommends -q sshfs uidmap fuse3 dbus-user-session
[   26.261518] cloud-init[2091]: Reading package lists...
[   26.416159] cloud-init[2091]: Building dependency tree...
[   26.417192] cloud-init[2091]: Reading state information...
[   26.648233] cloud-init[2091]: Skipping fuse3, it is already installed and upgrade is not set.
[   26.648657] cloud-init[2091]: fuse3 set to manually installed.
[   26.648816] cloud-init[2091]: Skipping dbus-user-session, it is already installed and upgrade is not set.
[   26.648940] cloud-init[2091]: dbus-user-session set to manually installed.
[   26.649051] cloud-init[2091]: The following additional packages will be installed:
[   26.651759] cloud-init[2091]:   libsubid4
[   26.673312] cloud-init[2091]: The following NEW packages will be installed:
[...]

PR #1423, opened 2023-03-19
Generate hash file for all installed packages

afbjorklund:
Try to generate the manifest (unidentifiably called "SHA256SUMS"), for the installed packages.

This is a proof-of-concept, for later using `repro-get` to install the packages from the cache...

Currently it comes with some **heavy** dependencies, for dockerconfig and for compression.

Maybe one can work with repro-get upstream, to get those included more on a selective basis ?

Issue:
* #1422


afbjorklund:
ubuntu:
118M	/var/lib/apt
82M	/var/cache/apt

fedora:
4.2M	/var/lib/dnf
253M	/var/cache/dnf

cmd/lima-guestagent/distro_linux.go
```diff
@@ -0,0 +1,68 @@
...
+	"sort"
+
+	"github.com/reproducible-containers/repro-get/pkg/distro"
+	"github.com/reproducible-containers/repro-get/pkg/distro/alpine"
+	"github.com/reproducible-containers/repro-get/pkg/distro/arch"
+	"github.com/reproducible-containers/repro-get/pkg/distro/debian"
+	"github.com/reproducible-containers/repro-get/pkg/distro/distroutil/detect"
+	"github.com/reproducible-containers/repro-get/pkg/distro/fedora"
+	"github.com/reproducible-containers/repro-get/pkg/distro/none"
+	"github.com/reproducible-containers/repro-get/pkg/distro/ubuntu"
```
AkihiroSuda, line 14: I'd prefer to just os/exec it

go.mod
```diff
@@ -45,20 +46,26 @@ require (
...
 	github.com/a8m/envsubst v1.4.2 // indirect
 	github.com/alecthomas/participle/v2 v2.0.0-beta.5 // indirect
 	github.com/apparentlymart/go-cidr v1.1.0 // indirect
+	github.com/containerd/nerdctl v0.23.1-0.20221008120401-b4c01094b581 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/digitalocean/go-libvirt v0.0.0-20201209184759-e2a69bcd5bd1 // indirect
 	github.com/dimchansky/utfbom v1.1.1 // indirect
+	github.com/docker/cli v20.10.18+incompatible // indirect
+	github.com/docker/docker v20.10.18+incompatible // indirect
+	github.com/docker/docker-credential-helpers v0.7.0 // indirect
```
AkihiroSuda, line 61: Too many deps. Let's just os/exec it.

afbjorklund:
I guess that means that we have to install the `repro-get` binary (11M), somehow.

cmd/lima-guestagent/distro_linux.go
afbjorklund: Maybe this particular function should be moved from cmd to pkg, upstream ? (in repro-get)

afbjorklund:
And it needs to handle errors, for those distributions that are not yet supported.

afbjorklund:
This will **not** bundle partial commands of `github.com/reproducible-containers/repro-get` in Lima, so will close PR.

If it is implemented, it will probably install and use `repro-get` as a program instead. So only the shell part remains...

PR #1424, opened 2023-03-19
Adjust table test for longer architectures

afbjorklund:
Turns out that "aarch64" is one character longer than "x86_64".

In the future, we might want to have a special tablewriter mod.

----

Fixes tests breaking:

GOARCH=amd64 gotestsum ./...
DONE 77 tests, 4 skipped in 0.645s
GOARCH=arm64 gotestsum ./...
DONE 77 tests, 4 skipped, 2 failures in 1.233s

Might want to add a cross-platform test target, to catch these ?

### DIFF
--- a/pkg/store/instance_test.go
+++ b/pkg/store/instance_test.go
@@ -5,6 +5,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/lima-vm/lima/pkg/limayaml"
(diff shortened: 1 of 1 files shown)


afbjorklund:
It also failed for "riscv64", along with some other test (in a separate issue)

DONE 77 tests, 4 skipped, 3 failures in 1.443s

=== FAIL: pkg/limayaml TestFillDefault (0.07s)

Something about assuming that all non-amd64 architectures are arm64...

afbjorklund:
* #1425

AkihiroSuda (commented):
Thanks

AkihiroSuda (approved):

PR #1425, opened 2023-03-19
Tests fail on RISCV64 due to hardcoded AARCH64

afbjorklund:
The tests were assuming that all non-X8664 is RISCV64

Also shows a lack of testing, on the RISC-V platform.


afbjorklund:
GOARCH=riscv64 gotestsum ./...
DONE 77 tests, 4 skipped, 3 failures in 1.415s

AkihiroSuda (approved):
Thanks

PR #1428, opened 2023-03-19
Allow setting LIMA_CACHE for the cache directory

### LINKED ISSUE #1427: Allow changing the location of the lima cache
### Description

Currently you can easily change the location of the lima files, using $LIMA_HOME...

It would be nice if it is was just as simple to change the cache location, $LIMA_CACHE.

Both default to using $HOME directories by default, which might not always be suitable.

Should also document the default location on Linux, which is under $XDG_CACHE_HOME.

#### Related

* https://pkg.go.dev/os#UserCacheDir
* https://pkg.go.dev/os#UserHomeDir
* https://pkg.go.dev/os#UserConfigDir

afbjorklund:
Closes #1427

### DIFF
--- a/cmd/limactl/prune.go
+++ b/cmd/limactl/prune.go
@@ -2,8 +2,8 @@ package main
 
 import (
 	"os"
-	"path/filepath"
 
+	"github.com/lima-vm/lima/pkg/store/dirnames"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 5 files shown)


AkihiroSuda:
Why not just use `XDG_CACHE_HOME`?

AkihiroSuda (approved):

afbjorklund:
> Why not just use `XDG_CACHE_HOME`?

It would affect other programs as well, similar to changing `HOME` in order to move LIMA_HOME ?

jandubois:
@AkihiroSuda Can you give some feedback on the discussion in #1427? Should the cache directory be linked to `LIMA_HOME` or be set independently via `LIMA_CACHE`?

AkihiroSuda:
> @AkihiroSuda Can you give some feedback on the discussion in #1427? Should the cache directory be linked to `LIMA_HOME` or be set independently via `LIMA_CACHE`?

The former one SGTM

jandubois (changes requested):
Please modify the PR as suggested in https://github.com/lima-vm/lima/issues/1427#issuecomment-1475597273 to use `$LIMA_HOME/_cache` as the cache directory (if it already exists).

afbjorklund:
Sounds like a very different change, the cache is already under $HOME (on Linux) so then it wouldn't make any difference

PR #1444, opened 2023-03-30
limactl shell flags MUST come before the instance name

jandubois:
Everything after the instance name is sent to the shell.

Updated usage:

```
  $ limactl shell --help
  Execute shell in Lima
  ...
  Usage:
    limactl shell [flags] INSTANCE [COMMAND...]
```

Ref: https://github.com/lima-vm/lima/discussions/586#discussioncomment-5477505


AkihiroSuda (approved):
Thanks

Issue #1438, opened 2023-03-27
Make disk conversion more generic when using vmType: vz


pendo324:
### Description

Tracking followup to https://github.com/lima-vm/lima/pull/1405#discussion_r1148315941.

Technically, the disk is just "not RAW", but the code assumes its qcow2. This is fine for now since `limactl` only creates QCOW2 disks with the `disk create` command, but this is not a safe assumption in general.

PR #1463, opened 2023-04-06
refactor: vz disk conversion

pendo324:
Closes: #1438 

Refactored disk conversion functions to be more explicit and used new `imgutil.ConvertToRaw` function in VZ driver, like suggested in https://github.com/lima-vm/lima/pull/1405#discussion_r1148315941.

### DIFF
--- a/pkg/qemu/imgutil/imgutil.go
+++ b/pkg/qemu/imgutil/imgutil.go
@@ -15,9 +15,9 @@ type Info struct {
 	VSize  int64  `json:"virtual-size,omitempty"`
 }
 
-func QCOWToRaw(source string, dest string) error {
+func ConvertToRaw(source string, dest string) error {
 	var stdout, stderr bytes.Buffer
-	cmd := exec.Command("qemu-img", "convert", source, dest)
+	cmd := exec.Command("qemu-img", "convert", "-O", "raw", source, dest)
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 	if err := cmd.Run(); err != nil {
--- a/pkg/vz/vm_darwin.go
+++ b/pkg/vz/vm_darwin.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/Code-Hex/vz/v3"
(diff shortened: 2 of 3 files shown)


pkg/qemu/imgutil/imgutil.go
```diff
@@ -17,7 +17,19 @@ type Info struct {
...
-	cmd := exec.Command("qemu-img", "convert", source, dest)
+	cmd := exec.Command("qemu-img", "convert", "-f", "qcow2", "-O", "raw", source, dest)
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("failed to run %v: stdout=%q, stderr=%q: %w",
+			cmd.Args, stdout.String(), stderr.String(), err)
+	}
+	return nil
+}
```
AkihiroSuda, line 28: Why do we still need this function?

pkg/vz/vm_darwin.go
```diff
@@ -357,19 +357,19 @@ func attachDisks(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfigura
 		}
 		if extraDiskFormat != "raw" {
 			rawPath := fmt.Sprintf("%s.raw", extraDiskPath)
-			qcow2Path := fmt.Sprintf("%s.qcow2", extraDiskPath)
-			if err = imgutil.QCOWToRaw(extraDiskPath, rawPath); err != nil {
-				return fmt.Errorf("failed to convert qcow2 disk %q to raw for vz driver: %w", diskName, err)
+			oldFormatPath := fmt.Sprintf("%s.%s", extraDiskPath, extraDiskFormat)
+			if err = imgutil.ConvertToRaw(extraDiskPath, rawPath); err != nil {
```
AkihiroSuda, line 361: Make sure that `extraDiskFormat` does not contain `os.PathSeparator`

pkg/qemu/imgutil/imgutil.go
pendo324: Now that you mention it, we probably don't. It's only useful if we want to explicitly specify that the input is qcow2, but qemu-img should be able to detect that on its own. Updated in latest rev.

pkg/vz/vm_darwin.go
pendo324: Thanks, updated in latest rev

AkihiroSuda (approved):
Thanks

Issue #1448, opened 2023-04-01
Changing the disk size in lima.yaml should resize the disk


jandubois:
Just running `limactl edit` does not change the actual disk image, I guess one has to run `qemu-img` manually (for now)

```
qemu-img resize ~/.lima/almalinux-8/diffdisk 30G

qemu-img info ~/.lima/almalinux-8/diffdisk
```

After that, it seems to have done the rest of the work on next start (changed it in the lima.yaml, to `disk: "30GiB"`).

_Originally posted by @afbjorklund in https://github.com/lima-vm/lima/discussions/1446#discussioncomment-5487504_

Of course it should only allow increasing the disk size, as reducing it will corrupt the existing partition.

PR #1468, opened 2023-04-08
limactl: add `--cpus`, `--memory`, `--mount-type`, `--vm-type`, ...

AkihiroSuda:
`limactl create`, `limactl start`, `limactl edit`:
- `--cpus=INT` (Similar to `colima start --cpu`, but the flag name is slightly different)
- `--dns=IP` (Similar to `colima start --dns`)
- `--memory=FLOAT` (Similar to `colima start --memory`)
- `--mount=DIR[:w]` (Similar to `colima start --mount`)
- `--mount-type=(reverse-sshfs|9p|virtiofs)` (Similar to `colima start --mount-type`)
- `--mount-writable=BOOL`
- `--network=(lima:shared|lima:bridged|lima:host|lima:user-v2|vzNAT)`
- `--rosetta=BOOL`

`limactl create` and `limactl start` only:
- `--arch=(x86_64|aarch64|riscv64)` (Similar to `colima start --arch`)
- `--containerd=(user|system|user+system|none)`
- `--disk=FLOAT` (Similar to `colima start --disk`)
- `--vm-type=(qemu|vz)` (Similar to `colima start --vm-type`)

### DIFF
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -1,0 +1,252 @@
+package editflags
+
+import (
+	"fmt"
+	"math/bits"
+	"runtime"
+	"strconv"
+	"strings"
+
+	"github.com/pbnjay/memory"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+)
+
+// RegisterEdit registers flags related to in-place YAML modification, for `limactl edit`.
+func RegisterEdit(cmd *cobra.Command) {
+	registerEdit(cmd, "")
+}
+
+func registerEdit(cmd *cobra.Command, commentPrefix string) {
+	flags := cmd.Flags()
+
+	flags.Int("cpus", 0, commentPrefix+"number of CPUs") // Similar to colima's --cpu, but the flag name is slightly different (cpu vs cpus)
+	_ = cmd.RegisterFlagCompletionFunc("cpus", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		var res []string
+		for _, f := range completeCPUs(runtime.NumCPU()) {
+			res = append(res, strconv.Itoa(f))
+		}
+		return res, cobra.ShellCompDirectiveNoFileComp
+	})
+
+	flags.IPSlice("dns", nil, commentPrefix+"specify custom DNS (disable host resolver)") // colima-compatible
+
+	flags.Float32("memory", 0, commentPrefix+"memory in GiB") // colima-compatible
+	_ = cmd.RegisterFlagCompletionFunc("memory", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		var res []string
+		for _, f := range completeMemoryGiB(memory.TotalMemory()) {
+			res = append(res, fmt.Sprintf("%.1f", f))
+		}
+		return res, cobra.ShellCompDirectiveNoFileComp
+	})
+
+	flags.StringSlice("mount", nil, commentPrefix+"directories to mount, suffix ':w' for writable (Do not specify directories that overlap with the existing mounts)") // colima-compatible
+
+	flags.String("mount-type", "", commentPrefix+"mount type (reverse-sshfs, 9p, virtiofs)") // Similar to colima's --mount-type=(sshfs|9p|virtiofs), but "reverse-sshfs" is Lima is called "sshfs" in colima
+	_ = cmd.RegisterFlagCompletionFunc("mount-type", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"reverse-sshfs", "9p", "virtiofs"}, cobra.ShellCompDirectiveNoFileComp
+	})
+
+	flags.Bool("mount-writable", false, commentPrefix+"make all mounts writable")
+
+	flags.StringSlice("network", nil, commentPrefix+"additional networks, e.g., \"vzNAT\" or \"lima:shared\" to assign vmnet IP")
+	_ = cmd.RegisterFlagCompletionFunc("network", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		// TODO: retrieve the lima:* network list from networks.yaml
+		return []string{"lima:shared", "lima:bridged", "lima:host", "lima:user-v2", "vzNAT"}, cobra.ShellCompDirectiveNoFileComp
+	})
+
+	flags.Bool("rosetta", false, commentPrefix+"enable Rosetta (for vz instances)")
+
+	flags.String("set", "", commentPrefix+"modify the template inplace, using yq syntax")
+}
+
+// RegisterCreate registers flags related to in-place YAML modification, for `limactl create`.
+func RegisterCreate(cmd *cobra.Command, commentPrefix string) {
+	registerEdit(cmd, commentPrefix)
+	flags := cmd.Flags()
+
+	flags.String("arch", "", commentPrefix+"machine architecture (x86_64, aarch64, riscv64)") // colima-compatible
+	_ = cmd.RegisterFlagCompletionFunc("arch", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"x86_64", "aarch64", "riscv64"}, cobra.ShellCompDirectiveNoFileComp
+	})
+
+	flags.String("containerd", "", commentPrefix+"containerd mode (user, system, user+system, none)")
+	_ = cmd.RegisterFlagCompletionFunc("vm-type", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"user", "system", "user+system", "none"}, cobra.ShellCompDirectiveNoFileComp
+	})
+
+	flags.Float32("disk", 0, commentPrefix+"disk size in GiB") // colima-compatible
+	_ = cmd.RegisterFlagCompletionFunc("memory", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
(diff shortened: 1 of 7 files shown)


afbjorklund:
In the legacy systems, they used MiB instead (probably so you could do things like 0.5GiB, when using integers?)

```
   --virtualbox-cpu-count "1"										number of CPUs for the machine (-1 to use the number of CPUs available) [$VIRTUALBOX_CPU_COUNT]
   --virtualbox-memory "1024"										Size of memory for host in MB [$VIRTUALBOX_MEMORY_SIZE]
   --virtualbox-disk-size "20000"									Size of disk for host in MB [$VIRTUALBOX_DISK_SIZE]

   --qemu-cpu-count "1"											Number of CPUs
   --qemu-memory "1024"											Size of memory for host in MB
   --qemu-disk-size "20000"										Size of disk for host in MB
```

For a modern system, it makes much more sense to use GiB. It also looks better, next to the vCPU count. (e.g. 4 / 4)

```
      --cpus int            number of CPUs
      --memory int          memory in GiB
```

cmd/limactl/editflags/editflags.go
```diff
@@ -0,0 +1,60 @@
...
+	flags := cmd.Flags()
+
+	flags.Int("cpus", 0, "number of CPUs") // Similar to colima's --cpu
+	_ = cmd.RegisterFlagCompletionFunc("cpus", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"1", "2", "4", "8"}, cobra.ShellCompDirectiveNoFileComp
+	})
+	flags.Int("memory", 0, "memory in GiB") // colima-compatible
+	_ = cmd.RegisterFlagCompletionFunc("memory", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"1", "2", "4", "8", "16", "32"}, cobra.ShellCompDirectiveNoFileComp
+	})
```
afbjorklund, line 22: Should also add "disk" here, ("disk in GiB")

afbjorklund:
It should probably repeat the defaults ?

```
      --cpus int            number of CPUs (default 4)
      --memory int          memory in GiB (default 4)
```

Or would that information be misleading, if it is different in the template or in the override ?

jandubois:
> For a modern system, it makes much more sense to use GiB. It also looks better, next to the vCPU count. (e.g. 4 / 4)
> 
> ```
>       --cpus int            number of CPUs
>       --memory int          memory in GiB
> ```

I don't understand why "memory in GiB" should be an integer. Using rational numbers works just fine:

```console
$ l edit alpine --set '.memory = "4.5GiB"'
WARN[0000] `--set` is experimental
INFO[0000] Instance "alpine" configuration edited
[...]
$ ps -ef | grep qemu
  501 26304 26294   0 12:30pm ttys000    0:09.16 /usr/local/bin/qemu-system-x86_64 -m 4608 ...
```

cmd/limactl/editflags/editflags.go
jandubois: I have some concern here, because disk size can only ever be increased. The only way to shrink disk size is by destroying and re-creating the instance, losing all your data.

So this is not something that you should casually do. Or it should require a confirmation prompt for the user to acknowledge (disabled by `--tty=false`).

jandubois:
> It should probably repeat the defaults ?
> 
> ```
>       --cpus int            number of CPUs (default 4)
>       --memory int          memory in GiB (default 4)
> ```
> 
> Or would that information be misleading, if it is different in the template or in the override ?

I think this is fine; it shows the built-in defaults, not the current values.

jandubois:
Just for the record, I did have some concerns about this being yet another redundant way to do something that is already possible *with the `--set` option). But after consideration I think the `yq` syntax is maybe too obscure for the casual user, so having a shortcut for the most common settings is justified.

cmd/limactl/editflags/editflags.go
AkihiroSuda: For new instances it can be casually done

jandubois: > For new instances it can be casually done

Yes, my concern is purely about `limectl edit default --disk 500GiB`, and then being unable to ever shrink it again.

afbjorklund: Could disable (or prompt) the feature to edit the disk image, until grow and shrink has been implemented ?

* https://github.com/lima-vm/lima/issues/1448

jandubois: Yes, `--disk` should not be implemented until Lima can grow the disk automatically.

FWIW, I don't expect us to ever be able to shrink disks; is this even possible?

AkihiroSuda:
> I don't understand why "memory in GiB" should be an integer.

I’m trying to follow colima flags (when possible)

AkihiroSuda:
> Why should the user not be allowed to write --memory 2.5?

Updated to allow this

jandubois:
> > Why should the user not be allowed to write --memory 2.5?
> 
> Updated to allow this

Thanks! Now we just have to make it accept `--memory 2⅜` as a synonym for `--memory 2.875` 😄 

j/k

AkihiroSuda:
Will rebase after merging:
- https://github.com/lima-vm/lima/pull/1643

AkihiroSuda:
Rebased

jandubois (approved):
Thanks, LGTM; only cosmetics feedback. Making completion dynamic maybe not worth the effort.

Personally i wouldn't bother marking `--network`, `--set`, and `--rosetta` as experimental, just to make a `log.Warn` call when they are used. Do you expect to change them in a backward-incompatible way?

cmd/limactl/editflags/editflags.go
```diff
@@ -0,0 +1,210 @@
...
+func RegisterEdit(cmd *cobra.Command) {
+	registerEdit(cmd, "")
+}
+
+func registerEdit(cmd *cobra.Command, commentPrefix string) {
+	flags := cmd.Flags()
+
+	flags.Int("cpus", 0, commentPrefix+"number of CPUs") // Similar to colima's --cpu, but the flag name is slightly different (cpu vs cpus)
+	_ = cmd.RegisterFlagCompletionFunc("cpus", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"1", "2", "4", "8"}, cobra.ShellCompDirectiveNoFileComp
```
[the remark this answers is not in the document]
AkihiroSuda: The number of the guest CPUs can be same as the host CPUs AFAICS

cmd/limactl/editflags/editflags.go
```diff
@@ -0,0 +1,210 @@
...
+	})
+
+	flags.IPSlice("dns", nil, commentPrefix+"specify custom DNS (disable host resolver)") // colima-compatible
+
+	flags.Float32("memory", 0, commentPrefix+"memory in GiB") // colima-compatible
+	_ = cmd.RegisterFlagCompletionFunc("memory", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		return []string{"1", "2", "4", "8", "16", "32"}, cobra.ShellCompDirectiveNoFileComp
```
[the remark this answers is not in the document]
AkihiroSuda: Done

cmd/limactl/editflags/editflags.go
```diff
@@ -0,0 +1,210 @@
...
+
+	flags.Bool("mount-writable", false, commentPrefix+"make all mounts writable")
+
+	flags.StringSlice("network", nil, commentPrefix+"additional networks, e.g., \"vzNAT\" or \"lima:shared\" to assign vmnet IP")
+	_ = cmd.RegisterFlagCompletionFunc("network", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		// TODO: retrieve the lima:* network list from networks.yaml
+		return []string{"lima:shared", "lima:bridged", "lima:host", "lima:user-v2", "vzNAT"}, cobra.ShellCompDirectiveNoFileComp
+	})
+
+	flags.Bool("rosetta", false, commentPrefix+"enable Rosetta (For vz instances)")
```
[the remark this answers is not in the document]
AkihiroSuda: Done

AkihiroSuda:
> Personally i wouldn't bother marking --network, --set, and --rosetta as experimental, just to make a log.Warn call when they are used. Do you expect to change them in a backward-incompatible way?

We can revisit this when we make vz driver and user-v2 network default

cmd/limactl/editflags/editflags.go
```diff
@@ -0,0 +1,222 @@
...
+
+func registerEdit(cmd *cobra.Command, commentPrefix string) {
+	flags := cmd.Flags()
+
+	flags.Int("cpus", 0, commentPrefix+"number of CPUs") // Similar to colima's --cpu, but the flag name is slightly different (cpu vs cpus)
+	_ = cmd.RegisterFlagCompletionFunc("cpus", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		var res []string
+		for i := runtime.NumCPU(); i >= 1; i /= 2 {
+			res = append(res, strconv.Itoa(i))
+		}
```
jandubois, line 28: This doesn't work well when the number of CPUs is not a power of 2. On a machine with 20 CPUs I get this:

```
jan@studio ~ % limactl create --cpus
1   10  2   20  5
```

I would have preferred to get `1 2 4 8 16 20` as choices instead of `1 2 5 10 20`.

I guess there is nothing we can do to sort completions numerically instead of alphabetically, or is there?

jandubois (approved):
I'm fine with merging this as is, and improving tab completion later.

cmd/limactl/editflags/editflags.go
AkihiroSuda: Fixed, but not sortable

jandubois: The same will also be true for some memory configurations; there are M2 chips with 24GB, 96GB, and 192GB of RAM (and others that are proper powers of 2).

jandubois (approved):

PR #1471, opened 2023-04-09
Setup apk cache on persistent volume

jandubois:
Added packages must be reinstalled on each boot because Alpine runs from an ISO. Setting up the package cache avoid having to re-download the packages each time.


AkihiroSuda (approved):
Thanks

PR #1480, opened 2023-04-13
templates: openSUSE: fall back to provo-mirror.opensuse.org

AkihiroSuda:
download.opensuse.org is inaccessible from Japan (500 Internal Server Error) https://bugzilla.opensuse.org/show_bug.cgi?id=1210240


jandubois (approved):
Thanks, LGTM.

I do however have problems creating an opensuse-tumbleweed VM with this PR right now.

On `x86_64` I cannot proceed beyond

```
INFO[0035] [hostagent] Waiting for the essential requirement 1 of 5: "ssh"
```

and on `aarch64` the startup is hanging at

```
INFO[0323] [hostagent] Waiting for the final requirement 1 of 1: "boot scripts must have finished"
```

Given that the `x86_64` hasn't change, I don't think it is related to this PR, so approving it anyways. But will look some more tomorrow.

AkihiroSuda:
> I do however have problems creating an opensuse-tumbleweed VM 

Could you open an issue

Issue #1485, opened 2023-04-14
`[hostagent] Driver stopped due to error: "signal: abort trap" ` (since QEMU 8.0.0-rc1)


AkihiroSuda:
Lima doesn't work with the current HEAD of qemu (v8.0.0-rc4): 

```console
$ qemu-system-x86_64 -version
QEMU emulator version 7.2.94 (v7.0.0-8004-g7dbd6f8a27)
Copyright (c) 2003-2022 Fabrice Bellard and the QEMU Project developers

$ limactl -v
limactl version 0.15.0-84-ge8f4701

$ limactl start default 
INFO[0000] Using the existing instance "default"        
INFO[0000] Attempting to download the nerdctl archive    arch=x86_64 digest="sha256:ce4615eb027265d2aaa61f185388738096b652541f7fc7c607bc3f829dd5713d" location="https://github.com/containerd/nerdctl/releases/download/v1.3.0/nerdctl-full-1.3.0-linux-amd64.tar.gz"
INFO[0000] Using cache "/Users/suda/Library/Caches/lima/download/by-url-sha256/e22c9d3619acd34258d3d43e2b685bf6d4b421aa5c83f4241d96b1ee6b6fe141/data" 
INFO[0001] [hostagent] Starting QEMU (hint: to watch the boot progress, see "/Users/suda/.lima/default/serial.log") 
INFO[0001] SSH Local Port: 60022                        
INFO[0001] [hostagent] Waiting for the essential requirement 1 of 5: "ssh" 
INFO[0004] [hostagent] Driver stopped due to error: "signal: abort trap" 
INFO[0005] [hostagent] Shutting down the host agent     
WARN[0005] [hostagent] failed to exit SSH master        
INFO[0005] [hostagent] Shutting down QEMU with ACPI     
WARN[0005] [hostagent] failed to open the QMP socket "/Users/suda/.lima/default/qmp.sock", forcibly killing QEMU 
INFO[0005] [hostagent] QEMU has already exited          
FATA[0005] exiting, status={Running:false Degraded:false Exiting:true Errors:[] SSHLocalPort:0} (hint: see "/Users/suda/.lima/default/ha.stderr.log") 
```

`ha.stderr.log`:
```
...
{"level":"info","msg":"Starting QEMU (hint: to watch the boot progress, see \"/Users/suda/.lima/default/serial.log\")","time":"2023-04-14T14:58:06+09:00"}
[...]

PR #1487, opened 2023-04-14
qemu: disable pdpe1gb on Intel Mac

AkihiroSuda:
https://stackoverflow.com/a/72863744/5167443

Workaround for #1485

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -120,6 +120,15 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 				cpuType[arch] = "max"
 			}
 		}
+		if arch == X8664 && runtime.GOOS == "darwin" {
+			switch cpuType[arch] {
+			case "host", "max":
+				// Disable pdpe1gb on Intel Mac
+				// https://github.com/lima-vm/lima/issues/1485
+				// https://stackoverflow.com/a/72863744/5167443
+				cpuType[arch] += ",-pdpe1gb"
+			}
+		}
 	}
 	var overrideCPUType bool
 	for k, v := range d.CPUType {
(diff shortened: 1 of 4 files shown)


jandubois (approved):
I haven't tested it, but patch looks good.

I'm just confused why this is not required with earlier versions of qemu; don't they enable large page sizes by default?

The SO entry seems to describe a different error/symptom; just the workaround seems to be the same.

But if prevents the crash, then this should be good.

pkg/limayaml/defaults.go
```diff
@@ -120,6 +120,15 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
 			}
 		}
+		if arch == X8664 && runtime.GOOS == "darwin" {
+			switch cpuType[arch] {
+			case "host", "max":
+				// Disable pdpe1gb on Intel Mac
+				// https://github.com/lima-vm/lima/issues/1485
+				// https://stackoverflow.com/a/72863744/5167443
+				cpuType[arch] += ",-pdpe1gb"
+			}
```
jandubois, line 130: I can't make up my mind if this should be mentioned in `defaults.yaml`:

```yaml
cpuType:
  # 🟢 Builtin default: "qemu64" (or "host" when running on x86_64 host)
  x86_64: null
```

Maybe the runtime warning is good enough? idk

AkihiroSuda: Added

AkihiroSuda:
> I'm just confused why this is not required with earlier versions of qemu; don't they enable large page sizes by default?

QEMU v8.0.0-rc1 bumped up the EDK2 binary (https://gitlab.com/qemu-project/qemu/-/commit/86305e864191123dcf87c3af639fddfc59352ac6), that might be the reason.

jandubois (approved):

PR #1505, opened 2023-04-20
Make sure that yaml is valid with gopkg.in/yaml.v3 as well

### LINKED ISSUE #1503: Invalid YAML syntax not caught by validate and silently failing
### Description

So this was confusing me a fair bit almost thinking there was a race condition with the provisioning or maybe the image I was using. Suffice it to say it's neither. Consider the following snippet which prepares an instance with [distrobox](https://github.com/89luca89/distrobox) and a default container:

```
images:
- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"
  arch: "x86_64"
- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img"
  arch: "aarch64"

provision:
- mode: system
  script: |
    #!/bin/sh
    command -v distrobox >/dev/null 2>&1 && exit 0
    DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends podman flatpak
- mode: system
    #!/bin/sh
    curl -s https://raw.githubusercontent.com/89luca89/distrobox/main/install | sh
- mode: user
  script: |
    #!/bin/sh
    distrobox create --yes
probes:
- script: |
    #!/bin/bash
    set -eux -o pipefail
    timeout 30s bash -c "until command -v podman >/dev/null 2>&1; do sleep 3; done"
  description: podman to be installed
- script: |
    #!/bin/bash
    set -eux -o pipefail
    timeout 30s bash -c "until command -v distrobox >/dev/null 2>&1; do sleep 3; done"
  description: distrobox to be installed
```

Apparently the missing `script:` key is never caught, and it leads to the instance booting fine but with only `podman` installed.

`limactl validate` doesn't seem to detect this either.

afbjorklund:
The goccy library doesn't catch all the errors, that other libraries such as the one used by "yq" does.
So check the yaml with both of them.

Now catches things such as missing `script:` line:

`FATA[0000] failed to load YAML file "examples/distrobox.yaml": failed to unmarshal YAML (main file "/home/anders/lima/examples/distrobox.yaml"): yaml: line 12: did not find expected key`

Closes #1503

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	golang.org/x/sync v0.1.0
 	golang.org/x/sys v0.7.0
 	gopkg.in/op/go-logging.v1 v1.0.0-20160211212156-b2cb9fa56473
+	gopkg.in/yaml.v3 v3.0.1
 	gotest.tools/v3 v3.4.0
 	inet.af/tcpproxy v0.0.0-20220326234310-be3ee21c9fa0
 	k8s.io/api v0.26.3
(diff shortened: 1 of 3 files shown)


afbjorklund:
Could need some unit tests, to highlight some problematic pieces that are escaping goccy/go-yaml.

It would also be nice with a "save" function, for troubleshooting (with and without defaults/overrides) ?

AkihiroSuda (approved):
Thanks

Issue #1508, opened 2023-04-22
Support for FreeBSD guests ?


afbjorklund:
### Description

There was some question in Slack, about supporting also FreeBSD guests.

I guess it would be "freema" or something, and not just "lima", but it is doable...

There are qcow2 images for all architectures, but it does involve adding OS.

I was just doing some experiments, but thought I would raise the question first.

afbjorklund:
Example:

```yaml
os: "freebsd"

images:
- location: https://download.freebsd.org/ftp/snapshots/VM-IMAGES/14.0-CURRENT/amd64/Latest/FreeBSD-14.0-CURRENT-amd64.qcow2.xz
  arch: "x86_64"
- location: https://download.freebsd.org/ftp/snapshots/VM-IMAGES/14.0-CURRENT/aarch64/Latest/FreeBSD-14.0-CURRENT-arm64-aarch64.qcow2.xz
  arch: "aarch64"

video:
  display: "default" 
```

afbjorklund:
It is supposed to have support for OCI runtime:

https://github.com/samuelkarp/runj

jay7x:
Hi! I was the original requestor on Slack. My need is to run a Puppet module acceptance tests on a FreeBSD. Lima is really attractive here because it's headless and scriptable. People are using VirtualBox (with or w/o vagrant) to do this usually. But after I got the M1-based Macbook I lost the ability to do this. In my case there is no real urgency but it'd be really nice to have. I'll try it on a coming week and will share the experience here.

afbjorklund:
The support for linux containers (on freebsd) seems to have merged as well, in 1.7.

afbjorklund:
Short list of portability issues:

* There is no `/run`, only `/var/run` (added variable)
* There is no `/bin/bash`, only `/usr/local/bin/bash`. (added variable)

And of course: s/linux/freebsd/

The LocalPorts support for guestagent can be implemented using `sockstat`.
Note: the output of the sockstat command is subtly different, between the OS.

The `sshfs`  options for "fusefs" does not have `allow_other`, unfortunately:
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=215797

Not sure if `-o slave` works, but there is some `-o passive` flag instead ?
Also the kernel module is **not** loaded by default, needs /boot/loader.conf

But otherwise, it seems to be working:

```
   sudo mkdir -p /mnt/lima-cidata
[...]

PR #1509, opened 2023-04-22
Investigating guestagent on FreeBSD, for containerd with runj

afbjorklund:
* https://samuel.karp.dev/blog/2021/05/running-freebsd-jails-with-containerd-1-5/

* https://productionwithscissors.run/2022/09/04/containerd-linux-on-freebsd/

There is no systemd on BSD, so it needs to use `/var/run`.

GNU bash is in `/usr/local/bin/bash`, default is `/bin/csh`.

~~Eventually it should use BSM Audit events, but empty here.~~
Using a simple `sockstat` call, instead of trying to get kernel events

Added "os"

```yaml
# OS: "Linux", "FreeBSD".
# 🟢 Builtin default: "Linux"
os: null

```

Issue #1508

Related #1735


pkg/hostagent/requirements.go
```diff
@@ -116,11 +117,15 @@ fi
 		})
 
 	}
+	run := "/run"
+	if runtime.GOOS == "freebsd" {
```
afbjorklund, line 121: this should use the guest OS, not the host OS (my bad)

pkg/hostagent/hostagent.go
```diff
@@ -480,6 +481,9 @@ func (a *HostAgent) watchGuestAgentEvents(ctx context.Context) {
 
 	localUnix := filepath.Join(a.instDir, filenames.GuestAgentSock)
 	remoteUnix := "/run/lima-guestagent.sock"
+	if runtime.GOOS == "freebsd" {
```
afbjorklund, line 484: should use guest OS

Makefile
```diff
@@ -49,7 +49,10 @@ binaries: clean \
 	_output/bin/kubectl.lima \
 	_output/share/lima/lima-guestagent.Linux-x86_64 \
 	_output/share/lima/lima-guestagent.Linux-aarch64 \
-	_output/share/lima/lima-guestagent.Linux-riscv64
+	_output/share/lima/lima-guestagent.Linux-riscv64 \
+	_output/share/lima/lima-guestagent.FreeBSD-x86_64 \
+	_output/share/lima/lima-guestagent.FreeBSD-aarch64 \
```
AkihiroSuda, line 54: The binary name should be `FreeBSD-arm64`, if `uname -m` returns `arm64`

afbjorklund: Then we would need special code for that, to make arch depend on os.

afbjorklund:
I found some old code that I hadn't pushed, it needs to be rebased and cleaned up

Makefile
afbjorklund: Currently we just use the enum values directly, as in FREEBSD "-" AARCH64

afbjorklund:
Note: there is no use case for FreeBSD yet, except for "it exists" and in order to test `runj`

Makefile
AkihiroSuda: On the second thought `aarch64` is probably ok, as it is hard coded for the `.arch` field in the YAML and the `qemu-system-<ARCH>` binary name.

examples/experimental/freebsd.yaml
```diff
@@ -5,6 +5,12 @@ images:
   arch: "x86_64"
 - location: https://download.freebsd.org/ftp/snapshots/VM-IMAGES/14.0-CURRENT/aarch64/Latest/FreeBSD-14.0-CURRENT-arm64-aarch64.qcow2.xz
   arch: "aarch64"
+- location: https://download.freebsd.org/ftp/snapshots/VM-IMAGES/14.0-CURRENT/riscv64/Latest/FreeBSD-14.0-CURRENT-riscv-riscv64.qcow2.xz
+  arch: "riscv64"
+  kernel:
```
afbjorklund, line 10: this u-boot image was mostly a hack to get past the validate (for amd64/arm64), but it _is_ the real image:

https://wiki.freebsd.org/riscv/QEMU

We might want to import it to https://github.com/lima-vm/u-boot-qemu-mirror/ proper, so it can be used here.

examples/experimental/freebsd.yaml
```diff
@@ -0,0 +1,28 @@
...
+video:
+  display: "default"
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+
+additionalDisks:
+- "zfs"
```
afbjorklund, line 24: The ZFS disk is actually only needed if wanting to run containers, it will boot fine without it.

examples/experimental/freebsd.yaml
AkihiroSuda: Why can't we use Debian's u-boot binary for FreeBSD?

Makefile
```diff
@@ -58,7 +58,10 @@ binaries: clean \
 	_output/share/lima/lima-guestagent.Linux-x86_64 \
 	_output/share/lima/lima-guestagent.Linux-aarch64 \
 	_output/share/lima/lima-guestagent.Linux-armv7l \
-	_output/share/lima/lima-guestagent.Linux-riscv64
+	_output/share/lima/lima-guestagent.Linux-riscv64 \
+	_output/share/lima/lima-guestagent.FreeBSD-x86_64 \
+	_output/share/lima/lima-guestagent.FreeBSD-aarch64 \
+	_output/share/lima/lima-guestagent.FreeBSD-riscv64
```
AkihiroSuda, line 64: Eventually we may want to have `configure` or `make menuconfig` to specify build targets

cmd/lima-guestagent/daemon_freebsd.go
```diff
@@ -0,0 +1,69 @@
...
+	l, err := net.Listen("unix", socket)
+	if err != nil {
+		return err
+	}
+	if err := os.Chmod(socket, 0777); err != nil {
+		return err
+	}
+	logrus.Infof("serving the guest agent on %q", socket)
+	return srv.Serve(l)
+}
```
AkihiroSuda, line 69: The srv code can be shared with Linux

cmd/lima-guestagent/main_freebsd.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/spf13/cobra"
+)
+
+func main() {
+	if err := newApp().Execute(); err != nil {
+		logrus.Fatal(err)
+	}
+}
+
+func newApp() *cobra.Command {
```
AkihiroSuda, line 17: Can be (mostly) shared with Linux

pkg/hostagent/hostagent.go
```diff
@@ -118,6 +119,10 @@ func New(instName string, stdout io.Writer, sigintCh chan os.Signal, opts ...Opt
 	sshConfig := &ssh.SSHConfig{
 		AdditionalArgs: sshutil.SSHArgsFromOpts(sshOpts),
 	}
+	sshShell := "/bin/bash"
```
AkihiroSuda, line 122:
```suggestion
	bash := "/bin/bash"
```

examples/experimental/freebsd.yaml
afbjorklund: I think that should be fine as well, just haven't tested it.

Makefile
afbjorklund: * https://github.com/lima-vm/lima/pull/1790

dfr (commented):
A common idiom is to use:
```
#! /usr/bin/env bash
```

pkg/hostagent/hostagent.go
afbjorklund: Used `/usr/bin/env bash`

afbjorklund:
Note: the build configuration for guest agent defaults to FreeBSD disabled (only Linux enabled), so you want to make sure to enable this before building and installing...

`CONFIG_GUESTAGENT_OS_FREEBSD=y`

cmd/lima-guestagent/daemon_freebsd.go
afbjorklund: Added comment, about differences

cmd/lima-guestagent/main_freebsd.go
afbjorklund: Added comment, about differences

AkihiroSuda:
Needs rebase.

Does cloud-init work with FreeBSD?
No change is needed for the cidata stuff?

afbjorklund:
> Does cloud-init work with FreeBSD?

I think it could work (in theory at least), but that it is not included by default in the ISO image.

It seems to be quite new: https://www.freebsd.org/status/report-2022-10-2022-12/cloud-init/

PR #1516, opened 2023-04-28
Allow leaving additional disks unformatted

### LINKED ISSUE #1393: Add functionality to add disk devices without formatting them
### Description

Make it an option to create disks, without adding the GPT partitions or the ext4 filesystem.

This probably means that the structure needs to become a map, but it could support both ?

```yaml
additionalDisks:
- "data"
```

```yaml
additionalDisks:
- name: "data"
  format: false
```

I was using it to try out [Ceph](https://ceph.io/), but I suppose there could be other uses for not formatting...

----

The workaround was to remove the filesystem (with `wipefs`) and remove the signature (with `gdisk`)

Then I was able to use `cephadm`, to start up the cluster using containers... Worked nicely, with [dashboard](https://docs.ceph.com/en/quincy/mgr/dashboard/).

afbjorklund:
Default to a simple string with name, just like before.

```yaml
additionalDisks:
- "disk"
```

But also allow skipping the partitioning and formatting.

```yaml
additionalDisks:
- name: "disk"
  format: true
  fsType: "ext4"
```

----

Example 1)

Like when running something like [ceph](https://ceph.io), which uses LVM.

It requires the entire disk device to be "raw", unformatted:

```yaml
additionalDisks:
- name: "disk"
  format: false
```

https://docs.ceph.com/en/latest/cephadm/install/

> A storage device is considered available if all of the following conditions are met:
>
> * The device must have no partitions.
> * The device must not have any LVM state.
> * The device must not be mounted.
> * The device must not contain a file system.
> * The device must not contain a Ceph BlueStore OSD.
> * The device must be larger than 5 GB.

Example 2)

Also allow using something else than "ext4", like "xfs".

Or for adding custom arguments, such as with [glusterfs](https://www.gluster.org/):

```yaml
additionalDisks:
- name: "disk"
  format: true
  fsType: "xfs"
  fsArgs: ["-i","size=512"]
```

https://docs.gluster.org/en/latest/Quick-Start-Guide/Quickstart/

> The inode size is set to 512 bytes to accommodate for the extended attributes used by GlusterFS.

-----

Closes #1393

### DIFF
--- a/cmd/limactl/disk.go
+++ b/cmd/limactl/disk.go
@@ -245,7 +245,7 @@ func diskDeleteAction(cmd *cobra.Command, args []string) error {
 			for _, inst := range instances {
 				if len(inst.AdditionalDisks) > 0 {
 					for _, d := range inst.AdditionalDisks {
-						if d == diskName {
+						if d.Name == diskName {
 							refInstances = append(refInstances, inst.Name)
 						}
 					}
--- a/cmd/limactl/stop.go
+++ b/cmd/limactl/stop.go
@@ -113,7 +113,8 @@ func stopInstanceForcibly(inst *store.Instance) {
 		logrus.Infof("The %s driver process seems already stopped", inst.VMType)
 	}
 
-	for _, diskName := range inst.AdditionalDisks {
+	for _, d := range inst.AdditionalDisks {
+		diskName := d.Name
 		disk, err := store.InspectDisk(diskName)
 		if err != nil {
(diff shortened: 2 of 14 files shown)


AkihiroSuda:
Is this still a draft?

afbjorklund:
I think I put it in a draft, mostly in order to not interfer with the release at that time

AkihiroSuda (approved):
Thanks

balajiv113 (approved):

PR #1532, opened 2023-05-11
Support audio device for vz driver

### LINKED ISSUE #1531: Support audio device for vz driver
### Description

This is a follow up of #1526 for vz driver

balajiv113:
Fixes #1531 

Had to do some setup to configure audio. Steps are as follows,
- Install `linux-image-extra-virtual` to enable virtio-snd
- Restart the VM
- Run `aplay -l` to verify if device is shown
- Install `alsa-utils`
- Run `usermod -a -G audio <user>`
- Restart the vm
- Run `nano ~/.asoundrc and add the following
```
pcm.!default {
        type hw
        card 1
}

ctl.!default {
        type hw
        card 1
}
```
- Run `speaker-test -t wav` and verify audio working

_Note: These are the steps worked for me, there might be some simplified steps to configure as well._

### DIFF
--- a/pkg/vz/vm_darwin.go
+++ b/pkg/vz/vm_darwin.go
@@ -158,6 +158,10 @@ func createVM(driver *driver.BaseDriver) (*vz.VirtualMachine, error) {
 		return nil, err
 	}
 
+	if err = attachAudio(driver, vmConfig); err != nil {
+		return nil, err
+	}
+
 	if err = attachOtherDevices(driver, vmConfig); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


afbjorklund:
Does vz also have a dummy driver, that presents an audio device but doesn't do any sound ?

If so, it could be connected to the "none" value

pkg/vz/vm_darwin.go
```diff
@@ -556,6 +560,28 @@ func attachFolderMounts(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineCo
 	return nil
 }
 
+func attachAudio(driver *driver.BaseDriver, config *vz.VirtualMachineConfiguration) error {
+	if *driver.Yaml.Audio.Device == "vz" {
+		outputStream, err := vz.NewVirtioSoundDeviceHostOutputStreamConfiguration()
+		if err != nil {
+			return err
+		}
+		inputStream, err := vz.NewVirtioSoundDeviceHostInputStreamConfiguration()
```
afbjorklund, line 569: didn't configure any input stream for qemu, but that is probably a good idea... maybe a boolean ? or not

afbjorklund (approved):

AkihiroSuda (approved):
Thanks

balajiv113:
> Does vz also have a dummy driver, that presents an audio device but doesn't do any sound ?
> 
> If so, it could be connected to the "none" value

No vz doesn't have a dummy driver

pkg/vz/vm_darwin.go
balajiv113: I think it should be fine as QEMU directly has a device that support both. But i haven't tested microphone case.

I think it needs some permission (NSMicrophoneUsageDescription) in Info.plist. Checking on how to provide that.

balajiv113: Just verified Microphone as well, it is working without any extra Info.plist. Not sure how but working :D

balajiv113: Ah this makes a lot of sense.
Thanks for the clarification 😃

PR #1536, opened 2023-05-11
Add audio input to qemu in order to match vz

afbjorklund:
PR #1532 

Follow-up:
* #1527

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -600,7 +600,7 @@ func Cmdline(cfg Config) (string, []string, error) {
 		// audio controller
 		args = append(args, "-device", "ich9-intel-hda")
 		// audio codec
-		args = append(args, "-device", fmt.Sprintf("hda-output,audiodev=%s", id))
+		args = append(args, "-device", fmt.Sprintf("hda-duplex,audiodev=%s", id))
 	}
 	// Graphics
 	if *y.Video.Display != "" {


pkg/qemu/qemu.go
```diff
@@ -600,7 +600,7 @@ func Cmdline(cfg Config) (string, []string, error) {
 		// audio controller
 		args = append(args, "-device", "ich9-intel-hda")
 		// audio codec
-		args = append(args, "-device", fmt.Sprintf("hda-output,audiodev=%s", id))
+		args = append(args, "-device", fmt.Sprintf("hda-duplex,audiodev=%s", id))
```
AkihiroSuda, line 603: Could you update the comment in `default.yaml`?

Also wondering if we can have a YAML property to disable the audio input

AkihiroSuda: @afbjorklund WDYT?

afbjorklund: A boolean could be added, not sure how to implement it for vz though

AkihiroSuda: VZ implementation can be another PR

AkihiroSuda: What should we do with this for v0.16.0?

balajiv113: My take would be we can support both for now as by default (as we have support for this in both drivers).

Later if we get any usecase for disabling input we will be able to easily provide at that point of time

AkihiroSuda: For security I'd prefer to disable input by default

balajiv113: Ah yes, i missed the security point. It makes sense.

I will create a PR for vz to support the same.

afbjorklund: ok, so then we are good for qemu ?

PR #1538, opened 2023-05-14
fix: Allow port forwards to also bind IPv6 addresses

heyvito:
Hey folks! 

Yesterday I hit a weird case in which an application managed to bind to the same port Lima should be using. With the kind help from folks on CNCF's `#lima` Slack channel, I understood that Lima was binding to IPv4, and the other application to IPv6.

I confirmed those changes fixed the behaviour on my machines (M1/Intel both running the latest stable macOS version). Also `go test` passes, and `make lint` complains of code unrelated to this PR. However, I'm not really sure how to test those changes further, and couldn't find instructions on how to do so. That said, maintainers, could you kindly take a look and provide further information on how to correctly test those changes?

I'll keep this as a draft until we have tests. Thanks!

### DIFF
--- a/hack/test-example.sh
+++ b/hack/test-example.sh
@@ -86,6 +86,7 @@ function diagnose() {
 export ftp_proxy=http://localhost:2121
 
 INFO "Starting \"$NAME\" from \"$FILE\""
+defer "cat \"$HOME/.lima/${NAME}\"/*.log"
 defer "limactl delete -f \"$NAME\""
 
 if [[ -n ${CHECKS["disk"]} ]]; then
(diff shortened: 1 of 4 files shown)


jandubois:
I do not understand what you are trying to achieve with this change.

The code looks like you are trying to add a second default rule to forward all connections from `127.0.0.1` inside the guest to both the `127.0.0.1` and `::` on the host.

That is not how the portforwarding mechanism works: it stops at the first rule that matches the guest portion, and then forwards to the requested host port.

There is currently no mechanism for forwarding a single guest port to multiple host ports. Can you please file an issue first, so we can discuss the requirements first?

If we want to implement dualstack support, then is should be generalized so it can be applied to any forwarding rule, and not just the fallback default rule.

heyvito:
Hey @jandubois, thank you for taking some time to review this. I apologise if I made any kind of confusion here; I'll close this PR and file an issue instead.

PR #1543, opened 2023-05-17
Fix unstable vz driver network issue

balajiv113:
This should hopefully fix the network freeze when using vz driver

https://github.com/lima-vm/lima/issues/1200#issuecomment-1327740698

I didn't add fix issue id's mainly to try and see if this really fixes or not.

**What is happening, why this change ?**
When we use file.FD(), whenever the file gets closed or garbage collected (finaliser closing file) the FD give before becomes invalid.

In our case, we use file.FD() for both client and server connections. Since these are DGRAM connection, if we try to read on a closed FD we won't get any error instead it get stuck indefinitely. This could be the reason for the freeze of vz network 

Before this change, i verified that that file getting garbage collected in ~2mins. So logically FD would become invalid at some point.

### DIFF
--- a/pkg/networks/usernet/UDPFileConn.go
+++ b/pkg/networks/usernet/UDPFileConn.go
@@ -1,0 +1,21 @@
+package usernet
+
+import (
+	"errors"
+	"net"
+	"time"
+)
+
+type UDPFileConn struct {
+	net.Conn
+}
+
+func (conn *UDPFileConn) Read(b []byte) (n int, err error) {
+	// Check if the connection has been closed
+	if err := conn.SetReadDeadline(time.Time{}); err != nil {
+		if opErr, ok := err.(*net.OpError); ok && opErr.Err.Error() == "use of closed network connection" {
+			return 0, errors.New("UDPFileConn connection closed")
+		}
+	}
+	return conn.Conn.Read(b)
+}
--- a/pkg/networks/usernet/gvproxy.go
+++ b/pkg/networks/usernet/gvproxy.go
@@ -163,11 +163,11 @@ func listenFD(ctx context.Context, vn *virtualnetwork.VirtualNetwork) error {
 			files[0].Close()
 
 			go func() {
-				err = vn.AcceptBess(ctx, fileConn)
+				err = vn.AcceptBess(ctx, &UDPFileConn{Conn: fileConn})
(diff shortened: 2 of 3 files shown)


AkihiroSuda (dismissed):
Thanks

balajiv113:
Lets hold the merge.

Unfortunately this doesn't fix it :( I tried a long running instance in M1 and the freeze still happened

balajiv113:
@AkihiroSuda 
I have done a additional change to wrap gvisor-tap-vsock connection with UDPFileConn.go

In this we are calling setReadDeadline before every read. This will make sure if the UDP connection is not closed. Before trying to read. Reading on closed UDP connection will end up getting stuck in that read

We can merge these current set of changes if it looks good mainly to get more info when error happens again

balajiv113:
I am testing in my M1 from yesterday, with both long running VM and as well as load testing network with multiple torrent download. Unfortunately freeze didn't till now.

AkihiroSuda:
Thanks

> Unfortunately freeze didn't till now.

s/Unfortunately/Fortunately/ ? 🙂

balajiv113:
> s/Unfortunately/Fortunately/ ?

I was expecting it to freeze so unfortunate :D

pkg/networks/usernet/UDPFileConn.go
```diff
@@ -0,0 +1,21 @@
...
+
+type UDPFileConn struct {
+	net.Conn
+}
+
+func (conn *UDPFileConn) Read(b []byte) (n int, err error) {
+	// Check if the connection has been closed
+	if err := conn.SetReadDeadline(time.Time{}); err != nil {
+		if opErr, ok := err.(*net.OpError); ok && opErr.Err.Error() == "use of closed network connection" {
+			return 0, errors.New("UDPFileConn connection closed")
```
AkihiroSuda, line 17: Can we wrap the original err?
```suggestion
			return 0,  fmt.Errorf"the UDPFileConn connection closed: %w", err)
```

balajiv113: We could but since we are already checking if error is equal to `use of closed network connection` 

I thought we won't be getting anything new out of error object

AkihiroSuda (approved):

PR #1545, opened 2023-05-17
vz: validate required version before hostagent start

### LINKED ISSUE #1533: vz: fail before running hostagent when macOS is too old
https://github.com/lima-vm/lima/blob/23a8a7f014a30c15c79e0724bebfade8b65c79c2/pkg/vz/vz_driver_darwin.go#LL110-L122

Currently the `vz driver requires macOS 13 or higher to run` error is only visible in `~/.lima/vz/ha.stderr.log`.
This should be printed before launching the hostagent

balajiv113:
fixes #1533


AkihiroSuda (approved):
Thanks

PR #1548, opened 2023-05-18
CI: replace Wine with Windows

AkihiroSuda:


afbjorklund:
As long as it is running in VMs instead of containers, I suppose it might as well run on Windows

afbjorklund (approved):

PR #1562, opened 2023-05-22
CI: add colima compatibility test

AkihiroSuda:


hack/test-colima.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -eux -o pipefail
+
+colima start
+
+docker run --rm hello-world
+
+colima stop
```
AkihiroSuda, line 8: cc @abiosoft
Feel free to open a follow-up PR to add more tests if you like

jandubois (dismissed):
LGTM

I don't know why Cirrus shows a warning though:

```
8 | task_template: &task_template
  | you've probably meant template_task
```

Maybe rename the node to get rid of the warning, even though Cirrus shouldn't care about this.

PR #1564, opened 2023-05-22
Alpine 3.18 on VZ requires macOS 13.3+

jandubois:
(Alpine 3.16 works on 13.1+; Alpine 3.17 also requires 13.3+)


AkihiroSuda (approved):
Thanks

PR #1580, opened 2023-05-25
Bump alpine to 0.2.30 and fix Rosetta support

jandubois:
The alpine-lima changes make sure that the `binfmt` filesystem is mounted before the bootscripts are executed, so Rosetta can be registered.

Similarly, they ensure that the `qemu-binfmt` service is started before `lima-init`, so the qemu handler can be uninstalled when Rosetta is enabled.


AkihiroSuda (dismissed):
Thanks

AkihiroSuda (approved):

PR #1589, opened 2023-05-27
vz: remove audio input

AkihiroSuda:
Follow-up to PR #1532

Can be added again, with an additional knob in the YAML


balajiv113 (approved):
Thanks :) LGTM

Issue #1507, opened 2023-04-21
vz: remove dependency on `qemu-img` binary for converting qcow2 to raw


AkihiroSuda:
Maybe we can use https://github.com/dleonard0/read-qcow2 (CC0)

AkihiroSuda:
Tried to use https://github.com/dleonard0/read-qcow2, but it turned out to fail for most distro images (`openSUSE-Leap-15.4-JeOS.x86_64-OpenStack-Cloud.qcow2`, `CentOS-Stream-GenericCloud-8-20230404.0.x86_64.qcow2`, ...) due to `ENOTSUP` here :disappointed: :

https://github.com/dleonard0/read-qcow2/blob/189fbe18d2fabdfed20a9a7221460570222dc30b/qcow2.c#L224

AkihiroSuda:
Wrote by myself https://github.com/lima-vm/go-qcow2reader

AkihiroSuda:
PR:
- https://github.com/lima-vm/lima/pull/1596

PR #1596, opened 2023-05-31
vz: drop dependency on qemu-img

AkihiroSuda:
It is still recommended to install qemu, as vz driver is still experimental

Fix https://github.com/lima-vm/lima/issues/1507


pkg/nativeimgutil/nativeimgutil.go
```diff
@@ -0,0 +1,156 @@
...
+		rN, rErr := r.Read(buf)
+		if rErr != nil {
+			eof = errors.Is(rErr, io.EOF)
+			if !eof {
+				return n, fmt.Errorf("failed to read: %w", rErr)
+			}
+		}
+		// TODO: qcow2reader should have a method to notify whether buf is zero
+		if bytes.Equal(buf, zeroBuf) {
+			if _, sErr := w.Seek(int64(rN), os.SEEK_CUR); sErr != nil {
```
balajiv113, line 130: os.SEEK_CUR is deprecated. 

We could use the io.SeekCurrent

pkg/nativeimgutil/nativeimgutil.go
```diff
@@ -0,0 +1,156 @@
...
+			if wN != rN {
+				return n, fmt.Errorf("read %d, but wrote %d bytes", rN, wN)
+			}
+		}
+	}
+	return n, nil
+}
+
+func MakeSparse(f *os.File, n int64) error {
+	if _, err := f.Seek(n, os.SEEK_SET); err != nil {
```
balajiv113, line 152: os.SEEK_SET is deprecated.

We could use the io.SeekStart

pkg/nativeimgutil/nativeimgutil.go
```diff
@@ -0,0 +1,156 @@
...
+		return fmt.Errorf("failed to call copySparse(), bufSize=%d, copied=%d: %w", bufSize, copied, err)
+	}
+
+	// Resize
+	if size != nil {
+		if err = MakeSparse(destTmpF, *size); err != nil {
+			return err
+		}
+	}
+	if err = destTmpF.Close(); err != nil {
```
balajiv113, line 81: In case of vm starting stopped during conversion of disk, the temp files are not cleared

AkihiroSuda: The line 65 has `os.RemoveAll(destTmp)`

pkg/nativeimgutil/nativeimgutil.go
AkihiroSuda: Fixed

pkg/nativeimgutil/nativeimgutil.go
AkihiroSuda: Fixed

pkg/nativeimgutil/nativeimgutil.go
balajiv113: Ya i noticed but on testing this usecase, i could see stale temp diffdisk files.

Steps i did,
- limactl  start template://experimental/vz
- ctrl+c when we get that converting log
- Noticed the diffdisk-*.temp in directory

AkihiroSuda: `limactl stop` now remove `*.tmp`

balajiv113 (approved):
LGTM :)

One thing i noticed is, the conversion takes more time than before. But we can improve on that incrementally as well if possible.

PR #1597, opened 2023-06-01
Support for display with vz driver

### LINKED ISSUE #1165: vz: support `video.display`
https://github.com/lima-vm/lima/blob/871a320f609456d609b504feb6cc0a9945ad2d54/examples/default.yaml#L226-L233

I guess we can add a new value "vz" here

balajiv113:
fixes #1165

### DIFF
--- a/cmd/limactl/hostagent.go
+++ b/cmd/limactl/hostagent.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"runtime"
 	"strconv"
 
 	"github.com/gorilla/mux"
@@ -27,6 +28,7 @@ func newHostagentCommand() *cobra.Command {
 	}
 	hostagentCommand.Flags().StringP("pidfile", "p", "", "write pid to file")
 	hostagentCommand.Flags().String("socket", "", "hostagent socket")
+	hostagentCommand.Flags().Bool("run-gui", false, "run gui synchronously within hostagent")
 	hostagentCommand.Flags().String("nerdctl-archive", "", "local file path (not URL) of nerdctl-full-VERSION-linux-GOARCH.tar.gz")
 	return hostagentCommand
 }
@@ -54,6 +56,16 @@ func hostagentAction(cmd *cobra.Command, args []string) error {
 	}
 
 	instName := args[0]
+
+	runGUI, err := cmd.Flags().GetBool("run-gui")
+	if err != nil {
+		return err
+	}
+	if runGUI {
+		//Without this the call to vz.RunGUI fails. Adding it here, as this has to be called before the vz cgo loads.
+		runtime.LockOSThread()
+		defer runtime.UnlockOSThread()
+	}
 
 	sigintCh := make(chan os.Signal, 1)
 	signal.Notify(sigintCh, os.Interrupt)
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -8,13 +8,31 @@ import (
 	"github.com/lima-vm/lima/pkg/store"
 )
 
+// Driver interface is used by hostagent for managing vm.
+//
+// This interface is extended by BaseDriver which provides default implementation.
+// All other driver definition must extend BaseDriver
 type Driver interface {
+	// Validate returns error if the current driver isn't support for given config
 	Validate() error
 
+	// CreateDisk returns error if the current driver fails in creating disk
 	CreateDisk() error
 
+	// Start is used for booting the vm using driver instance
+	// It returns a chan error on successful boot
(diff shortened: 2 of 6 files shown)


pkg/vz/vz_driver_darwin.go
```diff
@@ -29,6 +30,11 @@ type LimaVzDriver struct {
 	machine *virtualMachineWrapper
 }
 
+func init() {
+	//Without this the call to StartGraphicApplication fails. This has to be called before the vz cgo loads
+	runtime.LockOSThread()
```
balajiv113, line 35: Though this is done in vz driver init() it will be applied even for hostagent of qemu instance.

Not sure if this will have any impact. I have done some basic testing all looked fine only. Will try to do some perf test.

pkg/hostagent/hostagent.go
```diff
@@ -335,12 +335,22 @@ func (a *HostAgent) Run(ctx context.Context) error {
 		logrus.Infof("VNC Password: `%s`", vncpwdfile)
 	}
 
+	if startGui {
+		go func() {
+			err = a.startRoutinesAndWait(ctx, errCh)
+			logrus.Error(err)
```
AkihiroSuda, line 341: Please check nil err

pkg/driver/driver.go
```diff
@@ -13,7 +13,9 @@ type Driver interface {
 
 	CreateDisk() error
 
-	Start(_ context.Context) (chan error, error)
+	Start(_ context.Context) (bool, chan error, error)
```
AkihiroSuda, line 16: Please add a comment to explain bool

pkg/vz/vz_driver_darwin.go
balajiv113: I don't think so, reverse-ssfs will be via separate process right ??

From my understanding, this LockOSThread should mostly affect go routines within hostagent.

pkg/hostagent/hostagent.go
balajiv113: Done

pkg/driver/driver.go
balajiv113: Done, Added comments for start and stop methods in driver. Do let me know if you have some suggestions on those comments

pkg/vz/vz_driver_darwin.go
balajiv113: This is also done

cmd/limactl/hostagent.go
```diff
@@ -32,6 +33,10 @@ func newHostagentCommand() *cobra.Command {
 }
 
 func hostagentAction(cmd *cobra.Command, args []string) error {
+	//Without this the call to vz.StartGUI fails. Adding it here, as this has to be called before the vz cgo loads.
+	runtime.LockOSThread()
+	defer runtime.UnlockOSThread()
```
AkihiroSuda, line 38: Can we do this only for vz mode?

pkg/driver/driver.go
```diff
@@ -8,13 +8,29 @@ import (
...
 	CreateDisk() error
 
-	Start(_ context.Context) (chan error, error)
+	// Start is used for booting the vm using driver instance
+	// It returns a bool as first return value, indicating whether StartGUI need to be called
+	// The second argument will be a chan error returned on successful boot
+	// The third argument may contain error occurred while starting driver
+	Start(_ context.Context) (bool, chan error, error)
 
+	// StartGUI is used for starting GUI synchronously by hostagent. This method should be returned only after vm terminates
```
AkihiroSuda, line 28:
```suggestion
	// RunGUI is used for starting GUI synchronously by hostagent. This method should be returned only after vm terminates
```
(Similar to os/exec.Command.Run)

cmd/limactl/hostagent.go
balajiv113: I tried options, but unfortunately could find a way to make it work. 

If i move this inside hostagent it will start to panic, because once within hostagent the vz cgo and all were loaded. So only if i add before cgo loads it works else ending up in panic.

I will try to add a limayaml check directly here and see if that works

balajiv113: Done the changes. Now it should impact only vz driver with display="vz"
Sending if hostagent need to run a sync ui as a command arg. In start.go deciding if we need to run a gui based on driver. 

Do check it out and let me know if you have any feedback :)

pkg/driver/driver.go
balajiv113: Done

AkihiroSuda (dismissed):
Thanks

AkihiroSuda (approved):

PR #1603, opened 2023-06-03
pkg/limayaml: adjust the default CPUs and memory size

AkihiroSuda:
CPUs: 4 → min(4, host CPU cores)
Memory: 4GiB → min(4 GiB, host memory / 4)


balajiv113 (dismissed):

jandubois (approved):

jandubois:
The change looks fine, but I'm not sure where the ¼ of total memory default comes from.

It won't make any difference for machines with 16GiB or more, but for Github runners we will move from 4GiB to 3GiB. I guess this is fine, but it is close to the minimum you need if you want to run Kubernetes inside your VM and actually run any workloads as well.

As for running Lima on an 8GiB laptop, I guess this is possible, but not very practical anyways, so using 2GiB of memory may be fine.

AkihiroSuda:
> The change looks fine, but I'm not sure where the ¼ of total memory default comes from.

Just a random number. Let me know if 1/2 is better

afbjorklund:
> we will move from 4GiB to 3GiB. I guess this is fine, but it is close to the minimum 

Historically we used 1GiB for Docker Machine and 2GiB for Kubernetes (Minikube)...

I guess expectations about what the "minimum" needs, has increased since then :-)

AkihiroSuda:
Changed from 1/4 to 1/2

PR #1615, opened 2023-06-06
Use debian genericcloud instead of generic images

afbjorklund:
This saves download time and disk space, by not including
drivers for physical hardware but only for virtual cloud.

https://cloud.debian.org/images/cloud/

> _generic_: Should run in any environment using cloud-init, for e.g. OpenStack, DigitalOcean and also on bare metal.
> _genericcloud_: Similar to generic. Should run in any virtualised environment. Is smaller than `generic` by excluding drivers for physical hardware.

Add fallback URLs, similar to the others.


AkihiroSuda (approved):
Thanks

Issue #312, opened 2021-10-09
Some rootless base setup is needed even without systemd


afbjorklund:
Two are still needed:

* ~~Enable cgroup delegation~~

* Set up sysctl

* **Set up subuid**

*  ~~Start systemd session~~

----

`[rootlesskit:parent] error: failed to setup UID/GID map: failed to compute uid/gid map: open /etc/subuid: no such file or directory`

The scripts look like they should be OK, without `systemctl`.

Issue #311, opened 2021-10-09
Ability to choose container daemon in nerdctl.lima


afbjorklund:
The nerdctl wrapper assumes rootless, maybe it should allow root ?

So that if the user configured it to, it would automatically prepend `sudo`

```
$ nerdctl.lima version
WARN[0000] environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
FATA[0000] rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
$ lima sudo nerdctl version
Client:
 Version:	0.12.0
 Git commit:	<unknown>

Server:
 containerd:
  Version:	1.4.9
  GitCommit:	
```

This would of course be an _option_, similar to $LIMA_INSTANCE

Just to make the invocation slightly more "pretty" (shorter, that is)

----

In Podman, they made up a whole "connection" concept for this.

```console
$ podman system connection ls
Name                         Identity                                  URI
podman-machine-default*      /home/anders/.ssh/podman-machine-default  ssh://core@localhost:44365/run/user/1000/podman/podman.sock
podman-machine-default-root  /home/anders/.ssh/podman-machine-default  ssh://root@localhost:44365/run/podman/podman.sock
```

But I think that is overkill, it (rootless) is more or less a boolean ?

afbjorklund:
A shell alias works as well, but is "harder" to configure **for** the user.

Oh well, maybe I will make a PR that adds a variable and you can decide whether it is "in the way" or not :-)
Otherwise it would be more like `sudo podman` everywhere, as opposed to traditional `docker` (root group)

PR #1619, opened 2023-06-11
Add Alpine examples with the container engines

afbjorklund:
These are (1G) smaller than the usual Ubuntu or Fedora based

Uses the regular Alpine system packages, for the installation

----

Not sure if we want to add this to the core distribution, though.

It adds to the maintenance, and was _rather_ straightforward?

```
254M	alpine-docker
1,5G	docker
197M	alpine-podman
1,3G	podman
```

I was just comparing it with the custom alpine-lima ISO images.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -37,6 +37,11 @@ Container engines:
 - [`podman-rootful`](./podman-rootful.yaml): Podman (rootful)
 - LXD is installed in the default Ubuntu template, so there is no `lxd.yaml`
 
+Distro and engine examples:
+- [`experimental/alpine-nerdctl`](./experimental/alpine-nerdctl.yaml): Alpine Linux, with containerd/nerdctl
+- [`experimental/alpine-docker`](./experimental/alpine-docker.yaml): Alpine Linux, with Docker (rootful)
+- [`experimental/alpine-podman`](./experimental/alpine-podman.yaml): Alpine Linux, with Podman (rootless)
+
 Container image builders:
 - [`buildkit`](./buildkit.yaml): BuildKit
 
--- a/examples/experimental/alpine-docker.yaml
+++ b/examples/experimental/alpine-docker.yaml
@@ -1,0 +1,62 @@
+# This template requires Lima v0.7.0 or later.
+# Using the Alpine 3.18 aarch64 image with vmType=vz requires macOS Ventura 13.3 or later.
+
+images:
+- location: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.32/alpine-lima-std-3.18.0-x86_64.iso"
+  arch: "x86_64"
+  digest: "sha512:7b00fff78736a27a24e4a7de5f28037e9c7cf0fc539a33ec551c6ac619eb54237b5f25bfa35512fa7233cf23396dc249592710ef9150f619afa15267f9c8cbd4"
+- location: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.32/alpine-lima-std-3.18.0-aarch64.iso"
+  arch: "aarch64"
+  digest: "sha512:bf23a22e05854670eef74d9bfad056caa249832f22d5594eb6bb02fa9aae109d33c764242f862d48de5b6715c4792a3ee29c19888a0711fb27113ba5cf1ccf21"
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+# containerd is managed by Docker, not by Lima, so the values are set to false here.
+containerd:
+  system: false
+  user: false
+provision:
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    command -v docker >/dev/null 2>&1 && exit 0
+    apk add docker
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    if ! grep -q -- --group /etc/conf.d/docker; then
+      # Alternatively we could just add the user to the "docker" group, but that requires restarting the user session
+      echo "DOCKER_OPTS=\"--group ${LIMA_CIDATA_USER}\"" >> /etc/conf.d/docker
+    fi
+    rc-update add docker boot
+    service docker start
+probes:
+- description: "docker to be installed"
+  script: |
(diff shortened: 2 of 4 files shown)


AkihiroSuda (approved):

jandubois:
I don't have strong opinions, but since we already build our own ISO, we could just create an edition that includes both docker and podman, and then use that in these templates.

I like the ISO to have **everything**. That way the VM starts faster. And you have the additional packages in your cache because you have the whole ISO in the cache. So deleting and recreating a machine is also faster.

Maybe we should call it the `all` edition? And include `nerdctl` etc too, given that the regular mechanism to add nerdctl with Lima doesn't work with Alpine.

afbjorklund:
> I like the ISO to have **everything**.

Right, that was one of the goals.

But it is not necessary for testing it out, so I put it up as a separate (alpine-lima) issue.
Most of it would be the same, just the image and the `apk` installation would change.

> And include nerdctl etc too

The nerdctl doesn't really run without systemd, so didn't make an isolated example.
One is still available with the nerdctl daemon, but it doesn't use set-uid for `nerdctl`

https://github.com/afbjorklund/nerdctld/tree/main/lima

---

I think that adding everything would go against the purpose, but it could be done.
Just do something like the `rd` edition, and enable everything for an `all` build ?

The regular install won't work, both because the lack of systemd and the lack of glibc.
There are some things left to fix also in Lima, like setting up `/etc/subuid` etc.

* https://github.com/lima-vm/lima/issues/312

```console
lima-alpine:/home/anders$ sudo apk add nerdctl
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/main/x86_64/APKINDEX.tar.gz
fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/community/x86_64/APKINDEX.tar.gz
(1/6) Installing libseccomp (2.5.4-r2)
(2/6) Installing runc (1.1.7-r1)
(3/6) Installing containerd (1.7.2-r0)
(4/6) Installing containerd-openrc (1.7.2-r0)
(5/6) Installing cni-plugins (1.3.0-r0)
(6/6) Installing nerdctl (1.3.1-r1)
Executing busybox-1.36.0-r9.trigger
OK: 225 MiB in 92 packages
lima-alpine:/home/anders$ nerdctl version
WARN[0000] environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
WARN[0000] environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
WARN[0000] environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
WARN[0000] environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
FATA[0000] rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/ 
lima-alpine:/home/anders$ containerd-rootless-setuptool.sh install
/bin/ash: containerd-rootless-setuptool.sh: not found
```

The podman installation works, though. The docker installation does not include the script.

```
WARN[0000] "/" is not a shared mount, this could cause issues or missing mounts with rootless containers 
ERRO[0000] cannot find UID/GID for user anders: no subuid ranges found for user "anders" in /etc/subuid - check rootless mode in man pages. 
WARN[0000] Using rootless single mapping into the namespace. This might break some images. Check /etc/subuid and /etc/subgid for adding sub*ids if not using a network user 
```

```/bin/ash: dockerd-rootless-setuptool.sh: not found
```

Seems like it would be possible, just that it needs some special wrapping without systemd.

```
[INFO] systemd not detected, dockerd-rootless.sh needs to be started manually:

PATH=/usr/local/bin:/sbin:/usr/sbin:$PATH dockerd-rootless.sh 

[INFO] Creating CLI context "rootless"
Successfully created context "rootless"
[INFO] Using CLI context "rootless"
Current context is now "rootless"
```

`WARNING: Running in rootless-mode without cgroups. To enable cgroups in rootless-mode, you need to boot the system in cgroup v2 mode.`

afbjorklund:
Basically the Alpine packages just assume that you are running things with `sudo`.

While this was the standard with Docker, I'm not sure if it is "good enough" for Lima?

But it seems to work, as long as you do the extra steps to set up rootless containers.

It is possible that support for doing so without systemd will go away in the future.

afbjorklund:
Updated version, alpine-lima 0.2.32


```
172M	/home/anders/.lima/alpine
201M	/home/anders/.lima/alpine-podman
258M	/home/anders/.lima/alpine-docker
278M	/home/anders/.lima/alpine-nerdctl

Alpine Linux v3.18
podman version 4.5.1
Docker version 23.0.6, build ef23cbc4315ae76c744e02d687c09548ede461bd
nerdctl version 1.3.1  # containerd v1.7.2, buildkitd 0.11.6
```


Added help text to the nerdctl template.

* https://github.com/lima-vm/lima/issues/311

```
To run `nerdctl`, use `lima sudo nerdctl` (not `nerdctl.lima`):
------
export LIMA_INSTANCE=alpine-nerdctl
lima sudo nerdctl run ghcr.io/afbjorklund/hello-nerd
------
```

And "hello world" image: https://github.com/afbjorklund/hello-nerd

examples/README.md
```diff
@@ -37,6 +37,11 @@ Container engines:
 - [`podman-rootful`](./podman-rootful.yaml): Podman (rootful)
 - LXD is installed in the default Ubuntu template, so there is no `lxd.yaml`
 
+Distro and engine examples:
+- [`alpine-nerdctl`](./alpine-nerdctl.yaml): Alpine Linux, with containerd/nerdctl
+- [`alpine-docker`](./alpine-docker.yaml): Alpine Linux, with Docker (rootful)
+- [`alpine-podman`](./alpine-podman.yaml): Alpine Linux, with Podman (rootless)
```
AkihiroSuda, line 43: Can we put them to experimental/ ?

afbjorklund:
Will leave these downstream, there are too many templates already.

PR #1628, opened 2023-06-15
qemu: Add virtiofs support on Linux

refi64:
This adds support for using virtiofs to mount filesystems on Linux hosts, via QEMU's vhost-user-fs-pci device + the Rust implementation of virtiofsd. In a simple "benchmark" running sha256sum on a copy of the Windows 11 ARM64 VHDK (because it's a large file I randomly had lying around):

- reverse-sshfs took ~21s
- 9p took ~13-15s
- virtiofs took ~6-7s

(For comparison, running it directly on the host system took ~5s.)

This is marked as "experimental" because it has undergone testing by...me and relies on additional tools installed other than just QEMU.

Unfortunately, this does *not* include support for DAX, because that's not merged into upstream QEMU yet, making it rather difficult to test.

Ref. #20.

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -2,6 +2,7 @@ package qemu
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -478,6 +479,12 @@ func Cmdline(cfg Config) (string, []string, error) {
 	memBytes = adjustMemBytesDarwinARM64HVF(memBytes, accel, features)
 	args = appendArgsIfNoConflict(args, "-m", strconv.Itoa(int(memBytes>>20)))
 
+	if *y.MountType == limayaml.VIRTIOFS {
+		args = appendArgsIfNoConflict(args, "-object",
+			fmt.Sprintf("memory-backend-file,id=virtiofs-shm,size=%s,mem-path=/dev/shm,share=on", strconv.Itoa(int(memBytes))))
+		args = appendArgsIfNoConflict(args, "-numa", "node,memdev=virtiofs-shm")
+	}
+
 	// CPU
 	cpu := y.CPUType[*y.Arch]
 	if runtime.GOOS == "darwin" && runtime.GOARCH == "amd64" {
@@ -775,7 +782,7 @@ func Cmdline(cfg Config) (string, []string, error) {
 
 	// We also want to enable vsock here, but QEMU does not support vsock for macOS hosts
 
-	if *y.MountType == limayaml.NINEP {
+	if *y.MountType == limayaml.NINEP || *y.MountType == limayaml.VIRTIOFS {
 		for i, f := range y.Mounts {
 			tag := fmt.Sprintf("mount%d", i)
 			location, err := localpathutil.Expand(f.Location)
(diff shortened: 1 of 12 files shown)


refi64:
Huh so Cirrus CI is failing...it would seem the Go version being installed there is *older* than Lima's minimum? Lima's `go.mod` has Go 1.19, but [Ubuntu 22.04 has Go 1.18](https://packages.ubuntu.com/jammy/golang). Should I update the CI to a newer Ubuntu, install Go 1.19 or newer onto 22.04, or...?

pkg/qemu/qemu_driver.go
```diff
@@ -191,6 +281,17 @@ func (l *LimaQemuDriver) removeVNCFiles() error {
...
 
+func (l *LimaQemuDriver) killVhosts() error {
+	var errs []error
+	for i, vhost := range l.vhostCmds {
+		if err := vhost.Process.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) {
+			errs = append(errs, fmt.Errorf("Failed to kill virtiofsd instance #%d: %w", i, err))
+		}
+	}
+
+	return errors.Join(errs...)
```
AkihiroSuda, line 292: We have been using go-multierror: https://github.com/lima-vm/lima/blob/658ef436ee7d7bf7f1c48499058fb46aed3777e4/go.mod#L25

Eventually we can switch away from `errors.Join` though.

pkg/qemu/qemu.go
```diff
@@ -812,6 +835,75 @@ func Cmdline(cfg Config) (string, []string, error) {
...
+	const vhostConfigsDir = "/usr/share/qemu/vhost-user"
+
+	configEntries, err := os.ReadDir(vhostConfigsDir)
+	if err != nil {
+		return "", err
+	}
+
+	for _, configEntry := range configEntries {
+		logrus.Debugf("checking vhost config %s", configEntry.Name())
+		if !strings.HasSuffix(configEntry.Name(), ".json") || !configEntry.Type().IsRegular() {
```
AkihiroSuda, line 853: No support for symlinks?

pkg/qemu/qemu.go
```diff
@@ -812,6 +835,75 @@ func Cmdline(cfg Config) (string, []string, error) {
 	return exe, args, nil
 }
 
+func FindVirtiofsd() (string, error) {
+	type vhostUserBackend struct {
+		BackendType string `json:"type"`
+		Binary      string `json:"binary"`
+	}
+
+	const vhostConfigsDir = "/usr/share/qemu/vhost-user"
```
AkihiroSuda, line 844: Should be relative to qemu binary

https://github.com/lima-vm/lima/blob/658ef436ee7d7bf7f1c48499058fb46aed3777e4/pkg/qemu/qemu.go#L877-L926

AkihiroSuda:
> Should I update the CI to a newer Ubuntu, install Go 1.19 or newer onto 22.04, or...?

I guess replacing `errors.Join` with `go-multierror` is enough to get CI green.
We can upgrade CI to Go 1.19 and switch away from `go-multierror` to `errors.Join`, but I expect it to happen in a separate PR.

pkg/qemu/qemu.go
refi64: ngl I'm not sure why I put that, removed it now

refi64:
@AkihiroSuda oops thanks for the info, switched to go-multierror now

AkihiroSuda (approved):
Thanks

Issue #1622, opened 2023-06-12
Add armv7 (qemu-system-arm) emulation support


c-herz:
### Description

I often need to debug and configure programs that run on embedded Linux systems based on the armv7 architecture (NXP i.MX, Broadcom, etc.). I use a MacBook Pro 14" 2021 model as my main machine, and I find Lima very useful for development that specifically requires a Linux system. However, I would love to be able to set "arch=armv7" or something along those lines and emulate an entirely 32-bit ARM system with a 32-bit kernel, since this is the native architecture that most of these embedded processors are built on. To do this all while benefiting from the convenient filesystem integration that Lima provides, without having to deal with the headache of spinning up and manually configuring a qemu VM, would greatly simplify my workflow and make it much easier for me to analyze and run embedded ARM binaries.

afbjorklund:
Did this a lot for the older Raspberry Pi models, but I think that most are giving up on building arm32 images.

First the `armv6` binaries disappeared, and now the `armv7` seem to be slowly fading - for `armv8` and arm64...

Maybe support it as an architecture for limactl ?

But without the containers and without the images.

XVilka:
@afbjorklund there are always Linux distributions tailored for older hardware, these could be used. Debian Ports also tend to carry architecture support for a looong time, e.g. HPPA and Alpha are still supported unofficially:
- https://www.alpinelinux.org/downloads/
- https://www.debian.org/ports/
- https://www.ports.debian.org/

afbjorklund:
Here are the test results, of the PR.

$ limactl --debug start ./examples/experimental/armv7l.yaml
$ limactl shell armv7l

anders@lima-armv7l:~$ arch
armv7l

Running the "Raspberry Pi 2" CPU:

```console
anders@lima-armv7l:~$ grep PRETTY /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.2 LTS"
anders@lima-armv7l:~$ uname -a
Linux lima-armv7l 5.15.0-75-generic-lpae #82-Ubuntu SMP Wed Jun 7 00:16:35 UTC 2023 armv7l armv7l armv7l GNU/Linux
anders@lima-armv7l:~$ lscpu 
Architecture:           armv7l
  Byte Order:           Little Endian
CPU(s):                 4
  On-line CPU(s) list:  0-3
Vendor ID:              ARM
  Model name:           Cortex-A7
    Model:              5
    Thread(s) per core: 1
    Core(s) per socket: 4
[...]

PR #1633, opened 2023-06-16
Add support for armv7l architecture

afbjorklund:
Closes: #1622

The name was selected to match `uname` output, like with all the other Lima architectures.

Note that armv5 and armv6 are **not** supported, only armv7 (container architecture: `linux/arm/v7`)
This architecture is expected to be emulated, there is no arm32 hardware with CPU virt support - afaik.

### DIFF
--- a/examples/experimental/armv7l.yaml
+++ b/examples/experimental/armv7l.yaml
@@ -1,0 +1,22 @@
+# This example requires Lima v0.7.0 or later.
+
+arch: "armv7l"
+images:
+# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20230518/ubuntu-22.04-server-cloudimg-armhf.img"
+  arch: "armv7l"
+  digest: "sha256:f9baf349e9f4bb4acc6cbd14dc82aa2aa514fd76e06c74678e5453f0e6717fd2"
+# Fallback to the latest release image.
+# Hint: run `limactl prune` to invalidate the cache
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-armhf.img"
+  arch: "armv7l"
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+
+# We do not have arm-v7 binaries of containerd
+containerd:
+  system: false
+  user: false
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -32,6 +32,14 @@ func TestFillDefault(t *testing.T) {
 		arch = X8664
 	case "arm64":
 		arch = AARCH64
+	case "arm":
+		if runtime.GOOS != "linux" {
+			t.Skipf("unsupported GOOS: %s", runtime.GOOS)
+		}
+		if arm := goarm(); arm < 7 {
+			t.Skipf("unsupported GOARM: %d", arm)
+		}
+		arch = ARMV7L
 	case "riscv64":
 		arch = RISCV64
 	default:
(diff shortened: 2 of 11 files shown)


examples/experimental/armv7l.yaml
```diff
@@ -0,0 +1,20 @@
+# This example requires Lima v0.7.0 or later.
+
+arch: "armv7l"
+images:
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img"
+  arch: "aarch64"
```
AkihiroSuda, line 8: Doesn't make sense to add them to armv7l

pkg/limayaml/defaults_test.go
```diff
@@ -32,6 +32,8 @@ func TestFillDefault(t *testing.T) {
 		arch = X8664
 	case "arm64":
 		arch = AARCH64
+	case "arm":
+		arch = ARMV7L // assume runtime.GOARM == 7
```
AkihiroSuda, line 36: Can we panic if runtime.GOARM != 7 ?

pkg/qemu/qemu.go
```diff
@@ -441,14 +458,28 @@ func adjustMemBytesDarwinARM64HVF(memBytes int64, accel string, features *featur
...
+	switch arch {
+	case limayaml.X8664:
+		return "q35"
+	case limayaml.AARCH64:
+		return "virt"
+	case limayaml.ARMV7L:
+		return "virt"
+	case limayaml.RISCV64:
+		return "virt"
+	}
```
AkihiroSuda, line 471:
```go
switch arch {
  case limayaml.X8664:
    return "q35"
  case limayaml.AARCH64, limayaml.ARMV7L, limayaml.RISCV64:
    return "virt"
  default: // what?
    return "none"
}
```

pkg/qemu/qemu.go
```diff
@@ -812,8 +856,15 @@ func Cmdline(cfg Config) (string, []string, error) {
 	return exe, args, nil
 }
 
+func getArch(arch limayaml.Arch) string {
```
AkihiroSuda, line 859:
```suggestion
// getQemuSystemArch returns the arch string used by `qemu-system-${...}` and `edk2-${...}-code.fd` .
func getQemuSystemArch(arch limayaml.Arch) string {
```

pkg/limayaml/defaults_test.go
afbjorklund: The thing is that there _is_ no runtime.GOARM, there is only a $GOARM. So we would have to reimplement:

> goarm 6 requires: cpu.ARM.HasVFP
> goarm 7 requires: cpu.ARM.HasVFPv3

But this is also a host check (I think), so it seems unlikely that it would even be tried on `armv5` or `armv6`?

This whole arm32 target is likely to be emulation-only, I think Raspberry Pi 3* was the first one with KVM...

\* i.e. GOARCH=arm64 and `armv8`

The use case was for testing programs intended for embedded devices, that required a 32-bit kernel too.

Otherwise you would probably just run a `--platform=linux/arm` container, with qemu-user, instead.

examples/experimental/armv7l.yaml
afbjorklund: It was just showing the similarity, if it was to be merged. But can of course be removed, since arch was "hardcoded"

pkg/qemu/qemu.go
afbjorklund: I'm not sure why there was a compiler warning when testing all possibilities of an enum?

But maybe it was because using a magic string instead, either way it "shouldn't happen"

AkihiroSuda:
Needs rebase

AkihiroSuda (approved):
Thanks

PR #1637, opened 2023-06-17
Add rootful flag to start for rootful template

### LINKED ISSUE #1636: Add start flag for choosing the rootful template or containerd instead
### Description

When using `podman machine init`, they have a `--rootful` flag to choose a rootful connection instead of a rootless one.

It would be useful to have a similar flag in Lima, which would make it easier to change (similar to --cpus --memory etc)

* https://github.com/lima-vm/lima/pull/1468

When using a template, it would choose the one named "-rootful" and when using default it would use a `yq` statement.

The name of the instance would not change, so `limactl start --rootful template://podman` would be named: podman

* https://github.com/containers/podman-desktop/issues/1081#issuecomment-1563296492

----

See https://docs.podman.io/en/latest/markdown/podman-machine-init.1.html for the other flags

Most of them don't apply, when not using CoreOS and Ignition, but some of them are similar to Lima's

afbjorklund:
Closes #1636

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -61,6 +61,7 @@ $ cat template.yaml | limactl start --name=local -
 	startCommand.Flags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "enable TUI interactions such as opening an editor, defaults to true when stdout is a terminal")
 	startCommand.Flags().String("name", "", "override the instance name")
 	startCommand.Flags().String("set", "", "modify the template inplace, using yq syntax")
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
This is being covered by `limactl start --containerd=system`
- https://github.com/lima-vm/lima/pull/1468

afbjorklund:
Also for podman and docker?

AkihiroSuda:
> Also for podman and docker?

`limactl start --name=podman template://podman-rootful`

afbjorklund:
It was an attempt at making the syntax more similar, and compatible

Still needs some more CI code, to make `nerdctl` accessible as a user.

PR #1642, opened 2023-06-22
Silence "field mounts[%d].virtiofs.queueSize is only supported on Linux" when it is not specified

AkihiroSuda:


pkg/limayaml/defaults.go
```diff
@@ -552,7 +567,7 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 		if mount.NineP.Msize == nil {
 			mounts[i].NineP.Msize = pointer.String(Default9pMsize)
 		}
-		if mount.Virtiofs.QueueSize == nil {
+		if mount.Virtiofs.QueueSize == nil && *y.VMType == QEMU && *y.MountType == VIRTIOFS {
```
jandubois, line 570: Still needs to check `GOOS`:

```suggestion
		if runtime.GOOS == "linux" && mount.Virtiofs.QueueSize == nil && *y.VMType == QEMU && *y.MountType == VIRTIOFS {
```

AkihiroSuda: `QEMU && VIRTIOFS` on non-Linux is checked in another place

jandubois: Yeah, but sometimes too late:

```console
$ l edit alpine --set '.mountType="virtiofs"'
WARN[0000] `--set` is experimental
WARN[0000] field mounts[0].virtiofs.queueSize is only supported on Linux
WARN[0000] field mounts[1].virtiofs.queueSize is only supported on Linux
WARN[0000] field mounts[2].virtiofs.queueSize is only supported on Linux
INFO[0000] Instance "alpine" configuration edited
FATA[0001] interrupt
```

But I agree that checking for `GOOS` here is probably not right.

jandubois (approved):
LGTM

PR #1643, opened 2023-06-23
limactl: add `limactl create` command

AkihiroSuda:
`limactl create` is similar to `limactl start` but it does not start the created instance, and it does not accept an existing instance name as an argument.

Now it is discouraged (not deprecated) to use `limactl start` for creating new instances.

Discouraged form:
```
limactl start --name=foo template://docker
```

Recommended form:
```
limactl create --name=foo template://docker
limactl start foo
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,6 +77,7 @@ jobs:
 
         ## Usage
         \`\`\`console
+        [macOS]$ limactl create
         [macOS]$ limactl start
         ...
         INFO[0029] READY. Run \`lima\` to open the shell.
--- a/cmd/apptainer.lima
+++ b/cmd/apptainer.lima
@@ -3,7 +3,7 @@ set -eu
 : "${LIMA_INSTANCE:=apptainer}"
 
 if [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
-  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl start --name=$LIMA_INSTANCE template://apptainer\` to create a new instance" >&2
+  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl create --name=$LIMA_INSTANCE template://apptainer\` to create a new instance" >&2
   exit 1
 fi
 export LIMA_INSTANCE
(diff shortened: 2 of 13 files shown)


afbjorklund:
Sounds good to me, more similar to `docker-machine`. For podman they ended up with both `init` and `--now`.

You might end up having to keep the discouraged form forever, or by adding a `--start` flag to the create command.

jandubois:
`limactl start` should be included in the command reference in `README.md`.

There are still lots of instances of `run "limactl start" to create an instance`:

```
$ ag "start.*create.*inst"
cmd/apptainer.lima
6:  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl start --name=$LIMA_INSTANCE template://apptainer\` to create a new instance" >&2

cmd/docker.lima
7:  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl start --name=$LIMA_INSTANCE template://docker\` to create a new instance" >&2

cmd/limactl/copy.go
81:					return fmt.Errorf("instance %q does not exist, run `limactl start %s` to create a new instance", instName, instName)

cmd/limactl/list.go
122:		logrus.Warn("No instance found. Run `limactl start` to create an instance.")

cmd/limactl/show_ssh.go
71:			return fmt.Errorf("instance %q does not exist, run `limactl start %s` to create a new instance", instName, instName)

cmd/podman.lima
7:  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl start --name=$LIMA_INSTANCE template://podman\` to create a new instance" >&2

cmd/kubectl.lima
19:  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl start --name=$LIMA_INSTANCE\` to create a new instance" >&2

cmd/limactl/shell.go
73:			return fmt.Errorf("instance %q does not exist, run `limactl start %s` to create a new instance", instName, instName)

examples/README.md
70:Run `limactl start fedora.yaml` to create a Lima instance named "fedora".
```

AkihiroSuda:
> `limactl start` should be included in the command reference in `README.md`.

I guess you meant `limactl create`?
I'll update them right after making a release, to avoid "I followed README.md but it doesn't work" reports

jandubois:
> I guess you meant `limactl create`?

Yes, sorry, muscle memory... 😄 

> I'll update them right after making a release, to avoid "I followed README.md but it doesn't work" reports

Ok, that makes sense.

jandubois (approved):
Thanks, LGTM

PR #1650, opened 2023-06-30
Add example config to deploy kubernetes via RKE2

Hirozy:
[RKE2 V1.27.3](https://github.com/rancher/rke2/releases/tag/v1.27.3%2Brke2r1) has added support for the arm64 architecture, so there is one more way to deploy K8S on Apple Silicon. The `INSTALL_RKE2_CHANNEL=latest` indicates that the latest `v1.27.x` is installed instead of the stable `v1.25.x`. And I think `k8s-rke2.yaml` should be in `examples/experimental` before `v1.27.x` becomes a stable channel.

```shell
env | grep "http_proxy\|https_proxy\|no_proxy" >> /etc/default/rke2-server
sed -i "s/http_proxy/CONTAINERD_HTTP_PROXY/g" /etc/default/rke2-server
sed -i "s/https_proxy/CONTAINERD_HTTPS_PROXY/g" /etc/default/rke2-server
sed -i "s/no_proxy/CONTAINERD_NO_PROXY/g" /etc/default/rke2-server
```
To avoid service unavailability due to `no_proxy`, use the variable with the prefix `CONTAINERD_` according to the [RKE2 documentation](https://docs.rke2.io/advanced#configuring-an-http-proxy).

### DIFF
--- a/examples/experimental/rke2.yaml
+++ b/examples/experimental/rke2.yaml
@@ -1,0 +1,75 @@
+# Deploy kubernetes via rke2 (which installs a bundled containerd).
+# $ limactl start ./rke2.yaml
+# $ limactl shell rke2 sudo kubectl
+#
+# It can be accessed from the host by exporting the kubeconfig file;
+# the ports are already forwarded automatically by lima:
+#
+# $ export KUBECONFIG=$(limactl list rke2 --format 'unix://{{.Dir}}/copied-from-guest/kubeconfig.yaml')
+# $ kubectl get no
+# NAME            STATUS   ROLES                       AGE   VERSION
+# lima-rke2       Ready    control-plane,etcd,master   68s   v1.27.3+rke2r1
+#
+# For more details of RKE2, please refer to https://docs.rke2.io/
+#
+# This example requires Lima v0.7.0 or later.
+
+
+images:
+# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20230518/ubuntu-22.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
+  digest: "sha256:afb820a9260217fd4c5c5aacfbca74aa7cd2418e830dc64ca2e0642b94aab161"
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20230518/ubuntu-22.04-server-cloudimg-arm64.img"
+  arch: "aarch64"
+  digest: "sha256:b47f8be40b5f91c37874817c3324a72cea1982a5fdad031d9b648c9623c3b4e2"
+# Fallback to the latest release image.
+# Hint: run `limactl prune` to invalidate the cache
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img"
+  arch: "aarch64"
+
+# Mounts are disabled in this example, but can be enabled optionally.
+mounts: []
+
+# containerd is managed by rke2, not by Lima, so the values are set to false here.
+containerd:
+  system: false
+  user: false
+
+provision:
+- mode: system
+  script: |
+    #!/bin/sh
+    if [ ! -d /var/lib/rancher/rke2 ]; then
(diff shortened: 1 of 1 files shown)


Hirozy:
@jandubois Please kindly review and let me know if you have any comments

examples/experimental/k8s-rke2.yaml
```diff
@@ -0,0 +1,70 @@
...
+# containerd is managed by rke2, not by Lima, so the values are set to false here.
+containerd:
+  system: false
+  user: false
+
+provision:
+- mode: system
+  script: |
+    #!/bin/sh
+    curl -sfL https://get.rke2.io | INSTALL_RKE2_CHANNEL=v1.27 sh -
```
jandubois, line 42: Reviewing this PR made me realize that we download the install script for k3s on each boot, whereas it is only needed on the first boot, so I created #1684.

I think a similar check should be added here too:

```suggestion
    if [ ! -d /var/lib/rancher/rke2 ]; then
            curl -sfL https://get.rke2.io | INSTALL_RKE2_CHANNEL=v1.27 sh -
            ...
```

jandubois (changes requested):
I've tried the template, and it works, but took about 15min for the instance to get ready because several images ended up in an ImagePullBackup state. I haven't tried to figure out why that happens.

I'm not sure about the naming; I would just call ist `experimental/rke2.yaml`. Otherwise we should rename `k8s` → `k8s-kubeadm` and `k3s` → `k8s-k3s` because they are all different ways to install `k8s`. @AkihiroSuda what do you think?

jandubois:
Maybe the template should include a link to https://docs.rke2.io for references (and add a similar link to https://docs.k3s.io` to `templates/k3s.yaml` as well)?

examples/experimental/k8s-rke2.yaml
Hirozy: Thanks for the comment.
I changed the filename to "experimental/rke2.yaml". And added that check.

examples/experimental/rke2.yaml
```diff
@@ -0,0 +1,75 @@
...
+#
+# It can be accessed from the host by exporting the kubeconfig file;
+# the ports are already forwarded automatically by lima:
+#
+# $ export KUBECONFIG=$(limactl list rke2 --format 'unix://{{.Dir}}/copied-from-guest/kubeconfig.yaml')
+# $ kubectl get no
+# NAME            STATUS   ROLES                       AGE   VERSION
+# lima-rke2       Ready    control-plane,etcd,master   68s   v1.27.3+rke2r1
+#
+# For more details of RKE2, please refer to https://docs.rke2.io/
```
Hirozy, line 13: I have put the description of "refer to https://docs.rke2.io/" here.

jandubois (approved):
Thanks, LGTM. Leaving it to @AkihiroSuda to merge if he is ok with the filename.

And FWIW, v1.27.4+rke2r1 was just released today:

```shell
$ KUBECONFIG="/Users/jan/.lima/rke2/copied-from-guest/kubeconfig.yaml" k get no
NAME        STATUS   ROLES                       AGE   VERSION
lima-rke2   Ready    control-plane,etcd,master   10m   v1.27.4+rke2r1
```

afbjorklund:
> Maybe the template should include a link to https://docs.rke2.io for references (and add a similar link to [https://docs.k3s.io`](https://docs.k3s.io%60) to `templates/k3s.yaml` as well)?

Feel free to add https://docs.k8s.io/ to templates/k8s.yaml, in that case. There's some step-by-step links now.

examples/experimental/rke2.yaml
```diff
@@ -0,0 +1,75 @@
...
+containerd:
+  system: false
+  user: false
+
+provision:
+- mode: system
+  script: |
+    #!/bin/sh
+    if [ ! -d /var/lib/rancher/rke2 ]; then
+      curl -sfL https://get.rke2.io | INSTALL_RKE2_CHANNEL=v1.27 sh -
```
afbjorklund, line 46: does the version have to be hardcoded, or is there something like "stable" that could be used instead ?

Hirozy: The current stable minor version of RKE2 is `v1.25`, which does not support aarch64.
The hard core here is to consider that keep minor version on `v1.27` until `v1.27` becomes stable. After that the hard core will be removed and the template file will be renamed from `examples/experimental/rke2.yaml` to `examples/rke2.yaml`.
The RKE2 version channels are here [https://github.com/rancher/rke2/blob/master/channels.yaml](https://github.com/rancher/rke2/blob/master/channels.yaml)

AkihiroSuda (approved):
Thanks

PR #1651, opened 2023-07-02
Add Podman Desktop to the list of Lima adopters

afbjorklund:
It is possible to use Lima, instead of Podman Machine, to run Podman, or Docker, or Kubernetes (for running with containerd).

How is documented in the Podman Desktop documentation, but as long as the instance is running it will appear automatically.

https://podman-desktop.io/docs/Installation/creating-a-lima-instance-with-podman-desktop

I don't want to link directly to the sub-page, since they have a tendency to break their URLs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Container environments:
 - [Rancher Desktop](https://rancherdesktop.io/): Kubernetes and container management to the desktop
 - [Colima](https://github.com/abiosoft/colima): Docker (and Kubernetes) on macOS with minimal setup
 - [Finch](https://github.com/runfinch/finch): Finch is a command line client for local container development
+- [Podman Desktop](https://podman-desktop.io/): Podman Desktop GUI has a plug-in for Lima virtual machines
 
 GUI:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -47,6 +47,7 @@ Container environments:
 - [Rancher Desktop](https://rancherdesktop.io/): Kubernetes and container management to the desktop
 - [Colima](https://github.com/abiosoft/colima): Docker (and Kubernetes) on macOS with minimal setup
 - [Finch](https://github.com/runfinch/finch): Finch is a command line client for local container development
+- [Podman Desktop](https://podman-desktop.io/): Containers and Kubernetes for application developers
```
AkihiroSuda, line 50: Maybe it should be noted that Lima is supported as an optional extension

afbjorklund: Yes, that would be fair to mention - even if the extension is included by default

afbjorklund: In the marketing they separate between "Docker Desktop extension" and "Podman Desktop plug-in", so I guess it should be the same here - even if it is still packaged as an extension or "container runtime provider" or whatever.

https://podman-desktop.io/extend

AkihiroSuda (approved):
Thanks

jandubois (approved):

PR #1652, opened 2023-07-03
Add fallback for the centos-stream-8 image too

afbjorklund:
Follow-up to:
* #1593


AkihiroSuda (approved):
Thanks

afbjorklund:
They seem to be deleting these rather fast

Issue #1406, opened 2023-03-04
Any plan to support Container Linux, such as CoreOS, Flatcar?


ronnin:
### Description

I don't know how to assign the ignition file/content to vm during its first-time boot.

Generally, how to apply cloud-init to vm?

Thank you. 

Great job!

afbjorklund:
Mentioned some of the issues in:
* #1372

Basically it doesn't fully support the standard cloud-init, but uses a custom system called "ignition"

AkihiroSuda:
Let me close this issue as a duplicate, and continue the discussion in #1372

Issue #1372, opened 2023-02-19
Flatcar Container Linux support


kullervo610:
### Description

Flatcar Container Linux [(https://www.flatcar.org/)](https://www.flatcar.org/) is an interesting immutable Linux OS which suits well to run e.g. Kubernetes. Any change support for Flatcar Linux can be added to Lima as it does not seem to work "out-of-the-box" based on quick testing done.

afbjorklund:
I think that Flatcar (and CoreOS) use ignition rather than cloud-init, so I'm not sure it is compatible.

afbjorklund:
Note that you can run Podman Desktop instead of Rancher Desktop, if you want Fedora CoreOS support...

It is "interesting immutable", and has a common ancestor with Container Linux (that would be: CoreOS)

afbjorklund:
Are you able to provide a YAML file, for the "stable" image ?

afbjorklund:
Currently there is no support for compressed images in Lima, so the .bz2 needs to be decompressed.

afbjorklund:
```yaml
images:
#- location: "https://stable.release.flatcar-linux.net/amd64-usr/3374.2.5/flatcar_production_qemu_image.img.bz2"
#  arch: "x86_64"
#  digest: "sha256:cb8a30d891f1a9c4dd3fe34508c2f00aa2f18956bbe48f649848201292ffd1a9"
- location: "flatcar_production_qemu_image.img"
  arch: "x86_64"
  digest: "sha256:28f570a120f040523972a7cae948eb6ee41cf444f3c7c92f1bc721bf95ee5891"

video:
  display: "default" 
```

For some reason, Flatcar produces every checksum - except for SHA256...


```
# MD5 HASH
0a8b08387670cbc62c2d5086622a0956  flatcar_production_qemu_image.img.bz2
# SHA1 HASH
52edb61aacf14e19f23716d20311197fe47a9fa9  flatcar_production_qemu_image.img.bz2
# SHA512 HASH
7aa371a6e7013d02f6e3d6031c155420ce617dfe920d4d744950f87faa6b6f83c5e938525fcba5e8d1fb885e647af20fb412c3b56dbaff1a2d3dfcf7815a3145  flatcar_production_qemu_image.img.bz2
```

afbjorklund:
The issue was not the OEM, it was that the labels and paths for `user-cloudinit` are hardcoded...

* The label of the CD is "config-2", not "cidata"

[...]

PR #1653, opened 2023-07-03
Add example for Fedora CoreOS, running ignition not cloud-init

afbjorklund:
The default image comes with podman and docker and containerd installed.

It is read-only (for the most part), so the regular scripts wouldn't work anyway.

Apparently the uid conflicts, and setting the home directory doesn't work.

But otherwise it seems to be working - as long you have installed `butane`.

----

https://docs.fedoraproject.org/en-US/fedora-coreos/provisioning-qemu/

https://docs.fedoraproject.org/en-US/fedora-coreos/producing-ign/

* https://github.com/lima-vm/lima/issues/1372

* https://github.com/lima-vm/lima/issues/1406

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -26,6 +26,7 @@ Distro:
 - [`ubuntu`](./ubuntu.yaml): Ubuntu (same as `default.yaml` but without extra YAML lines)
 - [`ubuntu-lts`](./ubuntu-lts.yaml): Ubuntu LTS (same as `ubuntu.yaml` but pinned to an LTS version)
 - [`deprecated/centos-7`](./deprecated/centos-7.yaml): [deprecated] CentOS Linux 7
+- [`experimental/fedora-coreos`](./experimental/fedora-coreos.yaml): [experimental] Fedora CoreOS
 - [`experimental/opensuse-tumbleweed`](./experimental/opensuse-tumbleweed.yaml): [experimental] openSUSE Tumbleweed
 
 Container engines:
--- a/.yamllint
+++ b/.yamllint
@@ -2,6 +2,10 @@
 
 extends: default
 
+ignore: |
+  # this is a yaml template, needs to executed
+  pkg/cidata/cidata.TEMPLATE.d/ignition.yaml
+
 rules:
   indentation:
     indent-sequences: false
--- a/examples/experimental/fedora-coreos.yaml
+++ b/examples/experimental/fedora-coreos.yaml
@@ -1,0 +1,20 @@
+# This example requires Lima v0.15.1 or later.
+# <https://fedoraproject.org/coreos/download/>
+
+images:
+- location: "https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/38.20230918.3.0/x86_64/fedora-coreos-38.20230918.3.0-qemu.x86_64.qcow2.xz"
+  arch: "x86_64"
+  digest: "sha256:52d98121563a61fdab6924a8e3190bc1cbf7775942d8a8046f9ef2a689ed7aa0"
+- location: "https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/38.20230918.3.0/aarch64/fedora-coreos-38.20230918.3.0-qemu.aarch64.qcow2.xz"
+  arch: "aarch64"
+  digest: "sha256:432b5af49f56fcf6d2ed9046ee7b3654adb6345203d69fa14fe8d37b72b79eb9"
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+
+# The built-in containerd installer does not support CoreOS currently.
+containerd:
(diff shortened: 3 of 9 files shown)


afbjorklund:
Now with `nerdctl` installer, at least DIY prepared:

```console
$ export LIMA_INSTANCE=fedora-coreos
$ nerdctl.lima version
FATA[0000] rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): stat /run/user/1001/containerd-rootless: no such file or directory 
$ lima containerd-rootless-setuptool.sh install
[INFO] Checking RootlessKit functionality
[INFO] Checking cgroup v2
[INFO] Checking overlayfs
[INFO] Requirements are satisfied
[INFO] Creating "/var/home/anders/.config/systemd/user/containerd.service"
[INFO] Starting systemd unit "containerd.service"
...
[INFO] Installed "containerd.service" successfully.
[INFO] To control "containerd.service", run: `systemctl --user (start|stop|restart) containerd.service`
[INFO] To run "containerd.service" on system startup automatically, run: `sudo loginctl enable-linger anders`
[INFO] ------------------------------------------------------------------------------------------
[INFO] Use `nerdctl` to connect to the rootless containerd.
[INFO] You do NOT need to specify $CONTAINERD_ADDRESS explicitly.
$ nerdctl.lima version
Client:
 Version:	v1.4.0
 OS/Arch:	linux/amd64
 Git commit:	7e8114a82da342cdbec9a518c5c6a1cce58105e9
 buildctl:
  Version:	v0.11.6
  GitCommit:	2951a28cd7085eb18979b1f710678623d94ed578

Server:
 containerd:
  Version:	v1.7.1
  GitCommit:	1677a17964311325ed1c31e2c0a3589ce6d5c30d
 runc:
  Version:	1.1.7
  GitCommit:	v1.1.7-0-g860f061b

```

examples/README.md
```diff
@@ -14,6 +14,7 @@ Distro:
 - [`debian-11.yaml`](./debian-11.yaml): Debian GNU/Linux 11(bullseye)
 - [`debian-12.yaml`](./debian-12.yaml), `debian.yaml`: ⭐Debian GNU/Linux 12(bookworm)
 - [`fedora.yaml`](./fedora.yaml): ⭐Fedora
+- [`fedora-coreos.yaml`](./fedora-coreos.yaml): Fedora CoreOS
```
AkihiroSuda, line 17: Probably this should be in `experimental/`

examples/fedora-coreos.yaml
```diff
@@ -0,0 +1,15 @@
+# This example requires Lima v0.7.0 or later.
```
AkihiroSuda, line 1: Not v0.7

examples/fedora-coreos.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+images:
+- location: "https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/38.20230609.3.0/x86_64/fedora-coreos-38.20230609.3.0-qemu.x86_64.qcow2.xz"
+  arch: "x86_64"
+  digest: "sha256:2c039e778505f608ff78b8ba02d497780ee95182f477dbd939ddd335ece29326"
+- location: "https://builds.coreos.fedoraproject.org/prod/streams/stable/builds/38.20230609.3.0/aarch64/fedora-coreos-38.20230609.3.0-qemu.aarch64.qcow2.xz"
+  arch: "aarch64"
+  digest: "sha256:0410362c212df040096933738607e56027005deb1c4a3eae417079180aaadc65"
+
+# The guest home directory can not be changed with CoreOS currently.
```
AkihiroSuda, line 12: The host home directory can be still mounted?

examples/fedora-coreos.yaml
AkihiroSuda, line 1: Can we test this in CI?

examples/fedora-coreos.yaml
afbjorklund: My problem was they both use `/home/anders`... baby there is a "homeDir" field in ignition - just that it looks unimplemented at the moment, and gives errors.

Another issue was that the "core" user now claims UID 1000, previously it used 500

examples/fedora-coreos.yaml
afbjorklund: This program is required to be installed **on the host**, as `butane`:

* https://github.com/coreos/butane/releases/download/v0.18.0/butane-x86_64-unknown-linux-gnu

* https://github.com/coreos/butane/releases/download/v0.18.0/butane-aarch64-unknown-linux-gnu

It translates the `ignition.yaml` into `config.ign`, which is required.

fwilhe2:
Very cool, seems to work as expected for me.

afbjorklund:
* https://github.com/lima-vm/lima/issues/1372#issuecomment-1891390458

PR #1655, opened 2023-07-04
Allow changing prefix for the installations

### LINKED ISSUE #1654: Add optional prefix parameter to the template
### Description

Currently the prefix is hardcoded to `/usr/local`, for installing guestagent and containerd.

But that doesn't work, if `/usr` is read-only (like on Container Linux, which uses `/opt`).

So add an optional parameter.

```yaml
# Prefix to use for installing guest agent, and containerd with dependencies (if configured)
prefix: null

```

afbjorklund:
If /usr/local is read-only, you can override the prefix to use something like /opt instead. But it is assumed that /etc is always read-write...

Closes #1654

----

For some reason /opt/bin was not in the PATH of the check, even if it is the PATH of the actual user later on.

Workaround was to check with `test -x` in addition to `command -v`, to see if `nerdctl` has been installed yet.

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -430,6 +430,10 @@ hostResolver:
 # - 1.1.1.1
 # - 1.0.0.1
 
+# Prefix to use for installing guest agent, and containerd with dependencies (if configured)
+# 🟢 Builtin default: /usr/local
+guestInstallPrefix: null
+
 # ===================================================================== #
 # GLOBAL DEFAULTS AND OVERRIDES
 # ===================================================================== #
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
@@ -15,7 +15,7 @@ if [ "${LIMA_CIDATA_MOUNTTYPE}" = "reverse-sshfs" ]; then
 fi
 
 # Install or update the guestagent binary
-install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent /usr/local/bin/lima-guestagent
+install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/lima-guestagent
 
 # Launch the guestagent service
 if [ -f /sbin/openrc-run ]; then
@@ -27,7 +27,7 @@ supervisor=supervise-daemon
 name="lima-guestagent"
 description="Forward ports to the lima-hostagent"
 
-command=/usr/local/bin/lima-guestagent
+command=${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent
 command_args="daemon"
 command_background=true
 pidfile="/run/lima-guestagent.pid"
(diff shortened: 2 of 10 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -2,6 +2,8 @@
 
 set -eux
 
+: "${LIMA_CIDATA_PREFIX:=/usr/local}"
+
```
jandubois, line 6: Why is this needed (also in `40-install-containerd.sh`)?

Lima already provides the default value, so `LIME_CIDATA_PREFIX` should never be empty.

examples/default.yaml
```diff
@@ -158,6 +158,10 @@ caCerts:
   #   YOUR-ORGS-TRUSTED-CA-CERT-HERE
   #   -----END CERTIFICATE-----
 
+# Prefix to use for installing guest agent, and containerd with dependencies (if configured)
+# 🟢 Builtin default: /usr/local
+prefix: null
```
jandubois, line 163: `prefix` feels a bit too generic to me in the context of `lima.yaml`. Maybe `installPrefix` or even `guestInstallPrefix` would be better?

AkihiroSuda: What shout "guestAgentInstallPrefix"

afbjorklund: It is also used for nerdctl-full install prefix

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
afbjorklund: Maybe it was not needed, but there were some weird test failures installing to "bin/nerdctl"

afbjorklund: Will revert

examples/default.yaml
afbjorklund: Trying with `guestInstallPrefix`

examples/default.yaml
```diff
@@ -158,6 +158,10 @@ caCerts:
   #   YOUR-ORGS-TRUSTED-CA-CERT-HERE
   #   -----END CERTIFICATE-----
 
+# Prefix to use for installing guest agent, and containerd with dependencies (if configured)
+# 🟢 Builtin default: /usr/local
+guestInstallPrefix: null
```
AkihiroSuda, line 163: This should be probably in the "FURTHER ADVANCED CONFIGURATION" section

pkg/limayaml/limayaml.go
```diff
@@ -23,6 +23,7 @@ type LimaYAML struct {
 	Video           Video           `yaml:"video,omitempty" json:"video,omitempty"`
 	Provision       []Provision     `yaml:"provision,omitempty" json:"provision,omitempty"`
 	Containerd      Containerd      `yaml:"containerd,omitempty" json:"containerd,omitempty"`
+	Prefix          *string         `yaml:"guestInstallPrefix,omitempty" json:"guestInstallPrefix,omitempty"`
```
AkihiroSuda, line 26: Please rename the go variable too

pkg/cidata/template.go
```diff
@@ -58,6 +58,7 @@ type TemplateArgs struct {
 	Mounts                          []Mount
 	MountType                       string
 	Disks                           []Disk
+	Prefix                          string
```
AkihiroSuda, line 61: This needs to be renamed too

AkihiroSuda (approved):

PR #1658, opened 2023-07-06
Update DNS set up script to use a dedicated chain 

a-palchikov:
This PR updates the behavior of the host DNS configuration script to avoid hiccups on VM restarts (or rather stop/start attempts).
The issue is that when hostagent starts up, it picks a random free port for UDP/TCP which is then reflected in the guest's nat table in append mode. If there was a previous rule to route the DNS traffic on a (now potentially invalid) port, the DNS requests will time out.

This changeset moves the host DNS rules to a dedicated chain to be able to update them seamless on restart and when hostagent ports change.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/09-host-dns-setup.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/09-host-dns-setup.sh
@@ -1,24 +1,25 @@
 #!/bin/sh
 set -eux
 
+readonly chain=LIMADNS
+
+chain_exists() {
+	iptables --table nat -n --list "${chain}" >/dev/null 2>&1
+}
+
 # Wait until iptables has been installed; 35-configure-packages.sh will call this script again
 if command -v iptables >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,6 +6,8 @@ GO ?= go
 TAR ?= tar
 ZIP ?= zip
 PLANTUML ?= plantuml # may also be "java -jar plantuml.jar" if installed elsewhere
+YAML_SRCS := $(shell git ls-files --exclude-standard . | grep '\.yaml$$')
+SCRIPT_SRCS := $(shell git ls-files --exclude-standard . | grep '\.sh$$')
```
AkihiroSuda, line 10: This should be probably a separate PR

a-palchikov: Removed from this PR.

AkihiroSuda:
Please squash commits

a-palchikov:
> Please squash commits

Done in [f243a75](https://github.com/lima-vm/lima/pull/1658/commits/f243a75407a72716d21bd3a4a89d90d74bf930ab)

PR #1673, opened 2023-07-21
Makefile: add `make minimal`

AkihiroSuda:


afbjorklund (approved):
Should it include the `lima` wrapper as well ?

AkihiroSuda:
> Should it include the `lima` wrapper as well ?

No, it should be just minimal

afbjorklund:
I consider it to be (minimal), but I don't know what this will be used for

AkihiroSuda:
> I consider it to be (minimal), but I don't know what this will be used for

For local testing I don't want to compile guestagents for non-native archs

afbjorklund:
I found doing something similar for qemu to be useful, for a new project/repository.

jandubois:
> I found doing something similar for qemu to be useful (for a new project).

Do you have that code in a repo somewhere? It might be useful for the `lima-and-qemu` repo used by Rancher Desktop (and possibly by finch; they do run a fork of it).

Bonus question, since this just came up on `lima-and-qemu`: is it possible to cross-compile qemu for aarch64 on amd64 (for Linux)? (because Github still don't have aarch64 runners).

afbjorklund:
> Do you have that code in a repo somewhere?

It was more of a configure line for the qemu build, and I didn't do any work of packaging the result up either...

Just thinking it could be alternative to the default build in for instance brew, which weighs in at a gigabyte or so

But yeah, adding a "qemu-lima" repo could be done

afbjorklund:
> s it possible to cross-compile qemu for aarch64 on amd64 (for Linux)?

I think it should be enough to set the compilers, i.e. ` CC` and `CXX`, with the arch

Can't see any special other mentions on : https://cblfs.clfs.org/index.php/QEMU

PR #1676, opened 2023-07-22
CI: mostly disable snapshot tests

AkihiroSuda:
The snapshot tests are too flaky, especially with the archlinux template


afbjorklund (approved):
At least the "online" tests, since they seem to lose the ssh connection sometimes ?

I'm not sure the "offline" tests are as unstable, they mostly work on the image...

AkihiroSuda:
Offline tests seem stable, but it is less specific to the guest distro, so just testing it with `test-misc` should be fine anyway IMHO

PR #1694, opened 2023-07-30
CI: test `template://docker` (and promote `template://docker` to "Tier 1")

AkihiroSuda:


hack/test-templates.sh
```diff
@@ -62,6 +64,9 @@ case "$NAME" in
 	CHECKS["port-forwards"]=""
 	CHECKS["user-v2"]=1
 	;;
+"docker")
+	CONTAINER_ENGINE="docker"
```
jandubois, line 68: I don't have time to do a proper test of this PR, but just looking at the changes it seems fine.

Minor nit-pick: To me `CONTAINER_ENGINE` should be `containerd` or either `dockerd` or `moby`, so I would pick a different name for the CLI (`nerdctl` vs `docker`), e.g. `CONTAINER_CTL`.

AkihiroSuda: I'd say containerd is a high-level runtime but not an "engine". There is no proper definition of "engine" though.

jandubois: I think of `runc` as the runtime, and `containerd` the engine, but I agree that there doesn't seem to be a generally accepted definition of the terms.

Anyways, just minor nit-picking; if you are happy with the current name, then I'm fine with it too.

PR #1721, opened 2023-08-08
feat: add WSL2 driver

pendo324:
This PR users WSL2 to add a new VM Driver for Windows. This addresses the longstanding GitHub issue asking for Windows support (https://github.com/lima-vm/lima/discussions/391).

The traditional thinking (as seen in https://github.com/lima-vm/lima/issues/909) is to make QEMU work on Windows, and the addition of a WSL2 driver should not inhibit that effort, just provide another option (I'd also like to see a direct HyperV option 👀 ).

Considerations:
- The implementation in this PR currently does not forward the VM's SSH address to `127.0.0.1:<Instance.SSHLocalPort>`, although this is possible through the use of netsh (I couldn't get ssh port forwarding to work). Currently, the PR just uses the WSL2 instance's IP address and port 22. I can add some netsh calling code to get this working relatively easily, I just wanted to know if that's required
- The implementation requires an `ssh` binary on Windows, which isn't common. My recommendation is to have users install [Git for Windows](https://github.com/git-for-windows/git/) (specifically the MinGit release), which comes with some other helpful utilities like `cygpath`. Bundling this with Lima or just requiring users to run `winget install -e --id Git.MinGit` ([winget](https://github.com/microsoft/winget-cli) is now built into Windows) and adding it to path are both possible. The version of OpenSSH installable as a Windows optional feature is pretty old. Open to suggestions incase I missed something.
- The implementation uses WSL2's built-in network sharing and ssh forwarding of ports. It should technically be possible to use gvisor or similar, but I think that can be future work
- The implementation keeps the WSL2 instance running until stopped by executing an extremely long `sleep` command. Open to suggestions here, as I'm sure there are better ways to do this
- There might be some final debug code left in this diff. I will remove it in revisions over the coming days
- There are a lot of tests left to be added, which I'll be working in the coming days


cmd/limactl/delete.go
```diff
@@ -53,9 +55,16 @@ func deleteInstance(inst *store.Instance, force bool) error {
 
 	stopInstanceForcibly(inst)
 
+	if inst.VMType == limayaml.WSL {
+		if _, err := executil.RunUTF16leCommand([]string{"wsl", "--unregister", inst.DistroName}); err != nil {
```
afbjorklund, line 59: This could be a good place to use the Unregister interface, that was introduced with Virtualbox.

* https://github.com/lima-vm/lima/pull/1277

Maybe we should add that to Lima first (as a no-op), independent of any driver that might use it ?

```diff
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -35,6 +35,12 @@ type Driver interface {
 	// It returns error if there are any errors during Stop
 	Stop(_ context.Context) error
 
+	// Register adds the instance to the registry.
+	Register(_ context.Context) error
+
+	// Unregister removes the instance from the registry.
+	Unregister(_ context.Context) error
+
 	ChangeDisplayPassword(_ context.Context, password string) error
 
 	GetDisplayConnection(_ context.Context) (string, error)

```

pkg/store/instance.go
```diff
@@ -17,20 +19,24 @@ import (
...
 
 type Status = string
 
 const (
-	StatusUnknown Status = ""
-	StatusBroken  Status = "Broken"
-	StatusStopped Status = "Stopped"
-	StatusRunning Status = "Running"
+	StatusUnknown       Status = ""
+	StatusUnititialized Status = "Unititialized"
```
afbjorklund, line 35: typo: Uninitialized

pkg/ioutilx/ioutilx.go
```diff
@@ -20,3 +26,28 @@ func ReadAtMaximum(r io.Reader, n int64) ([]byte, error) {
...
+func FromUTF16leToString(r io.Reader) (string, error) {
+	out, err := io.ReadAll(FromUTF16le(r))
+	if err != nil {
+		return "", err
+	}
+
+	return string(out), nil
+}
+
+func CannonicalWindowsPath(orig string) string {
```
afbjorklund, line 44: typo: CanonicalWindowsPath

afbjorklund:
@pendo324 : thank you for your contribution, it would be have been good to open an issue for discussion about the feature itself and keep the PR for the implementation - but that can be done afterwards too, and reference this PR

Running Lima in a container is slightly different from all the other drivers, so it might need some thought (Lima itself has some similaries to WSL2). I agree that a more traditional HyperV driver would also be a nice addition, like "VZ"...

cmd/lima-guestagent/daemon_linux.go
```diff
@@ -21,6 +21,8 @@ func newDaemonCommand() *cobra.Command {
 		RunE:  daemonAction,
 	}
 	daemonCommand.Flags().Duration("tick", 3*time.Second, "tick for polling events")
+	daemonCommand.Flags().Bool("audit", true, "use audit features")
+	daemonCommand.Flags().Bool("tcp", false, "use tcp server instead a UNIX socket")
```
AkihiroSuda, line 25: I heard modern Windows supports Unix sockets

AkihiroSuda: Also wondering if we can use virtio-serial or vsock

cmd/lima-guestagent/daemon_linux.go
```diff
@@ -21,6 +21,8 @@ func newDaemonCommand() *cobra.Command {
 		RunE:  daemonAction,
 	}
 	daemonCommand.Flags().Duration("tick", 3*time.Second, "tick for polling events")
+	daemonCommand.Flags().Bool("audit", true, "use audit features")
```
AkihiroSuda, line 24: Doesn’t WSL2 support audit?

AkihiroSuda: Can we remove this flag and auto detect the availability?

cmd/limactl/delete.go
AkihiroSuda, line 59: Can we just limit it to ASCII?

examples/experimental/wsl.yaml
```diff
@@ -0,0 +1,27 @@
...
+# ===================================================================== #
+
+# Default values in this YAML file are specified by `null` instead of Lima's "builtin default" values,
+# so they can be overridden by the $LIMA_HOME/_config/default.yaml mechanism documented at the end of this file.
+
+# VM type: "qemu" or "vz" (on macOS 13 and later).
+# The vmType can be specified only on creating the instance.
+# The vmType of existing instances cannot be changed.
+# 🟢 Builtin default: "qemu"
+vmType: wsl
```
AkihiroSuda, line 12:
```suggestion
vmType: wsl2
```

pkg/driverutil/instance.go
```diff
@@ -5,12 +5,16 @@ import (
...
+	"github.com/lima-vm/lima/pkg/wsl"
 )
 
 func CreateTargetDriverInstance(base *driver.BaseDriver) driver.Driver {
 	limaDriver := base.Yaml.VMType
 	if *limaDriver == limayaml.VZ {
 		return vz.New(base)
 	}
+	if *limaDriver == limayaml.WSL {
+		return wsl.New(base)
```
AkihiroSuda, line 17: s/WSL/WSL2/ ?

cmd/lima-guestagent/daemon_linux.go
pendo324: Windows does support Unix sockets (AF_UNIX), and I tried getting them to work, but I couldn't. I could create the socket on either side, but I couldn't get Windows to recognize a socket created by the guestagent (or the other way around). Would love to know if anyone else has experience or guidance here.

I think vscock could work though, I'm looking into it.

cmd/lima-guestagent/daemon_linux.go
pendo324: > Doesn’t WSL2 support audit?

Seems like the kernel does have support:
```
$ modprobe configs ; gunzip -dc /proc/config.gz | grep AUDIT
modprobe: FATAL: Module configs not found in directory /lib/modules/5.15.90.1-microsoft-standard-WSL2
# CONFIG_AUDIT is not set
CONFIG_HAVE_ARCH_AUDITSYSCALL=y
CONFIG_AUDIT_ARCH=y
# CONFIG_KVM_MMU_AUDIT is not set
```

but the WSL2 instances are created without the necessary capability to use audit:
```
$ sudo auditctl -s
Error - audit support not in kernel
Cannot open netlink audit socket
```

> Can we remove this flag and auto detect the availability?

Sounds good, I'll do this in the next revision unless there's some other ideas on enabling audit in a WSL2 instance.

cmd/limactl/delete.go
pendo324: Good idea. I'll port your Unregister changes to a separate PR and comment back when it's ready 👍

pendo324:
> @pendo324 : thank you for your contribution, it would be have been good to open an issue for discussion about the feature itself and keep the PR for the implementation - but that can be done afterwards too, and reference this PR
>
> Running Lima in a container is slightly different from all the other drivers, so it might need some thought (Lima itself has some similaries to WSL2). I agree that a more traditional HyperV driver would also be a nice addition, like "VZ"...

I'll create an issue to track Windows support in general, and weigh the different options (WSL2, HyperV, qemu, etc.) as I do think there are reasons you might want to use one over the other. I wanted to come to the project with some code instead of just an issue, but it ballooned into a lot of code before I knew it 😅

cmd/lima-guestagent/daemon_linux.go
pendo324: Adde vsock support, turned out to be pretty interesting @AkihiroSuda

cmd/lima-guestagent/daemon_linux.go
pendo324: Removed the flag in latest revision

examples/default.yaml
```diff
@@ -442,6 +442,11 @@ hostResolver:
 # 🟢 Builtin default: /usr/local
 guestInstallPrefix: null
 
+# Adjust options for the guestAgent which runs inside of the Linux VM
+guestAgent:
+  # 🟢 Builtin default: UNIX (macOS, Linux), VSOCK (Windows/WSL2)
+  protocol: null
+
```
AkihiroSuda, line 449: No need to make this customizable, at least for now?

examples/experimental/wsl.yaml
```diff
@@ -0,0 +1,27 @@
...
+# BASIC CONFIGURATION
+# ===================================================================== #
+
+# Default values in this YAML file are specified by `null` instead of Lima's "builtin default" values,
+# so they can be overridden by the $LIMA_HOME/_config/default.yaml mechanism documented at the end of this file.
+
+# VM type: "qemu" or "vz" (on macOS 13 and later).
+# The vmType can be specified only on creating the instance.
+# The vmType of existing instances cannot be changed.
+# 🟢 Builtin default: "qemu"
```
AkihiroSuda, line 11: These comment lines can be removed

examples/experimental/wsl.yaml
```diff
@@ -0,0 +1,27 @@
...
+vmType: wsl2
+
+# OpenStack-compatible disk image.
+# 🟢 Builtin default: null (must be specified)
+# 🔵 This file: Ubuntu 23.04 Lunar Lobster images
+images:
+# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
+- location: "file://C:/Users/Administrator/Downloads/fedora-38-cloud-rootfs.tar"
```
AkihiroSuda, line 19: Can we download this from https?

examples/experimental/wsl.yaml
AkihiroSuda, line 1: The file name should be “wsl2.yaml” if we are going to call the VM driver “wsl2”

examples/experimental/wsl.yaml
```diff
@@ -0,0 +1,27 @@
...
+  arch: "x86_64"
+  digest: "sha256:112400224d801c3781c1e301408685920ba980024bfb20877628a0bd82fdc574"
+
+mountType: wsl
```
AkihiroSuda, line 23: Should be “wsl2”?

examples/experimental/wsl.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+containerd:
+  system: true
+  user: false
```
AkihiroSuda, line 27: Can be removed, as it corresponds to the default values

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -72,7 +72,7 @@ fi
 
 # Signal that provisioning is done. The instance-id in the meta-data file changes on every boot,
 # so any copy from a previous boot cycle will have different content.
-cp "${LIMA_CIDATA_MNT}"/meta-data /run/lima-boot-done
+sudo cp "${LIMA_CIDATA_MNT}"/meta-data /run/lima-boot-done
```
AkihiroSuda, line 75: ?

pkg/downloader/downloader_unix.go
```diff
@@ -0,0 +1,22 @@
+//go:build !windows
+// +build !windows
```
AkihiroSuda, line 2: These lines can be removed

pkg/downloader/downloader_windows.go
```diff
@@ -0,0 +1,22 @@
+//go:build windows
+// +build windows
```
AkihiroSuda, line 2: These lines can be removed

pkg/httpclientutil/httpclientutil_others.go
```diff
@@ -0,0 +1,52 @@
...
+
+	"github.com/mdlayher/vsock"
+)
+
+// NewHTTPClientWithSocketPath creates a client.
+// socketPath is a path to the UNIX socket, without unix:// prefix.
+func NewHTTPClientWithSocketPath(socketPath string) (*http.Client, error) {
+	// https://github.com/adrg/xdg/pull/14
+	// TODO: move to separate file for compile time check instead of runtime
+	if runtime.GOOS == "windows" {
```
AkihiroSuda, line 21: Unreachable code

pkg/limayaml/limayaml.go
```diff
@@ -56,9 +57,11 @@ const (
 	REVSSHFS MountType = "reverse-sshfs"
 	NINEP    MountType = "9p"
 	VIRTIOFS MountType = "virtiofs"
+	WSLMount MountType = "wsl"
```
AkihiroSuda, line 60: Should be “wsl2”?

pkg/limayaml/limayaml.go
```diff
@@ -243,6 +246,18 @@ type CACertificates struct {
 	Certs          []string `yaml:"certs,omitempty" json:"certs,omitempty"`
 }
 
+type GuestAgentProto = string
+
+const (
+	GuestAgentUNIXProto  GuestAgentProto = "UNIX"
+	GuestAgentVSockProto GuestAgentProto = "VSOCK"
+	GuestAgentTCPProto   GuestAgentProto = "TCP"
```
AkihiroSuda, line 254: These should be in lower characters we are going to support customizing the protocol, but probably we can just make this uncustomizable

pkg/sshutil/sshutil.go
```diff
@@ -193,10 +200,18 @@ func CommonOpts(useDotSSH bool) ([]string, error) {
 		// We prioritize AES algorithms when AES accelerator is available.
 		if sshInfo.aesAccelerated {
 			logrus.Debugf("AES accelerator seems available, prioritizing aes128-gcm@openssh.com and aes256-gcm@openssh.com")
-			opts = append(opts, "Ciphers=\"^aes128-gcm@openssh.com,aes256-gcm@openssh.com\"")
+			if runtime.GOOS == "windows" {
+				opts = append(opts, "Ciphers=^aes128-gcm@openssh.com,aes256-gcm@openssh.com")
+			} else {
+				opts = append(opts, "Ciphers=\"^aes128-gcm@openssh.com,aes256-gcm@openssh.com\"")
+			}
```
AkihiroSuda, line 207: ?

pkg/sshutil/sshutil.go
```diff
@@ -216,11 +231,16 @@ func SSHOpts(instDir string, useDotSSH, forwardAgent bool, forwardX11 bool, forw
...
+	controlPath := fmt.Sprintf("ControlPath=\"%s\"", controlSock)
+	if runtime.GOOS == "windows" {
+		controlSock = ioutilx.CanonicalWindowsPath(controlSock)
+		controlPath = fmt.Sprintf("ControlPath=%s", controlSock)
+	}
 	opts = append(opts,
 		fmt.Sprintf("User=%s", u.Username), // guest and host have the same username, but we should specify the username explicitly (#85)
 		"ControlMaster=auto",
-		fmt.Sprintf("ControlPath=\"%s\"", controlSock),
-		"ControlPersist=5m",
```
AkihiroSuda, line 223: ?

pkg/store/filenames/filenames.go
```diff
@@ -52,6 +53,8 @@ const (
 	HostAgentStderrLog = "ha.stderr.log"
 	VzIdentifier       = "vz-identifier"
 	VzEfi              = "vz-efi"
+	WslRootFs          = "wsl-rootfs"
+	WslRootFsDir       = "wsl-root"
```
AkihiroSuda, line 57: Please update docs/internals.md too

AkihiroSuda: Also names seem confusing

pkg/store/instance.go
```diff
@@ -27,10 +30,12 @@ import (
 type Status = string
 
 const (
-	StatusUnknown Status = ""
-	StatusBroken  Status = "Broken"
-	StatusStopped Status = "Stopped"
-	StatusRunning Status = "Running"
+	StatusUnknown       Status = ""
+	StatusUninitialized Status = "Uninitialized"
+	StatusInstalling    Status = "Installing"
```
AkihiroSuda, line 35: Installing what?

pkg/store/instance.go
```diff
@@ -52,6 +57,9 @@ type Instance struct {
 	DriverPID       int                `json:"driverPID,omitempty"`
 	Errors          []error            `json:"errors,omitempty"`
 	Config          *limayaml.LimaYAML `json:"config,omitempty"`
+	RootFsPath      string             `json:"rootfs,omitempty"`
+	DistroName      string             `json:"distroName,omitempty"`
```
AkihiroSuda, line 61: Why do we need these values here?

pkg/wsl2/wsl_driver_others.go
```diff
@@ -0,0 +1,41 @@
...
+package wsl2
+
+import (
+	"context"
+	"errors"
+
+	"github.com/lima-vm/lima/pkg/driver"
+)
+
+var ErrUnsupported = errors.New("vm driver 'wsl' requires Windows 10 build 19041 or later (Hint: try recompiling Lima if you are seeing this error on Windows 10+)")
```
AkihiroSuda, line 13:
```suggestion
var ErrUnsupported = errors.New("vm driver 'wsl2' requires Windows 10 build 19041 or later (Hint: try recompiling Lima if you are seeing this error on Windows 10+)")
```

pkg/store/instance.go
pendo324: While the hostagent is running `wsl2.initVM` (the `wsl.exe --import ...` command), the status reported by wsl.exe --list -v will be "Installing". I can change the name or just map "Installing" to "Uninitialized" if that makes more sense

image

pkg/store/filenames/filenames.go
pendo324: Now that I'm looking at this again, I can probably just use BaseDisk for WslRootFsDir (that directory just holds the a `.vhdx` virtual harddrive).

WslRootFs is the downloaded root file system that is imported by `wsl.exe --import` and transformed into the `.vhdx`.

examples/experimental/wsl.yaml
pendo324: Fixed in latest revision

examples/experimental/wsl.yaml
pendo324: Fixed in latest revision

examples/experimental/wsl.yaml
pendo324: Updated comments in latest revision

examples/experimental/wsl.yaml
pendo324: Fixed in latest revision

pkg/sshutil/sshutil.go
pendo324: Same answer as above comment https://github.com/lima-vm/lima/pull/1721/files/4101c78bf8ca54b6448823a27cf3205558608876#r1295258558

pkg/store/instance.go
pendo324: We don't, removed in latest revision. It's leftover from prototyping

pkg/store/instance.go
```diff
@@ -375,3 +408,58 @@ func PrintInstances(w io.Writer, instances []*Instance, format string, options *
 	}
 	return nil
 }
+
+func GetWslStatus(instName, distroName string) (string, error) {
```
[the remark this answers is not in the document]
pendo324: Still thinking about the best way to move this to a `_windows.go` file. Might have to just duplicate `Inspect` (and  maybe`PrinitInstances`)

pendo324 (commented):
Addressed most of the feedback in latest revision, please re-review when available.

pkg/sshutil/sshutil.go
AkihiroSuda: @afbjorklund WDYT?

pkg/store/instance.go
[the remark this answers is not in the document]
pendo324: Refactored in latest revision, let me know what you think

pkg/store/filenames/filenames.go
pendo324: Fixed in latest revision

examples/experimental/wsl.yaml
pendo324: I think default is `user: true`, not `system: true`?

AkihiroSuda:
Needs rebase

cmd/lima-guestagent/daemon_linux.go
```diff
@@ -21,6 +23,9 @@ func newDaemonCommand() *cobra.Command {
 		RunE:  daemonAction,
 	}
 	daemonCommand.Flags().Duration("tick", 3*time.Second, "tick for polling events")
+	daemonCommand.Flags().Int("tcp-port", 0, "use tcp server instead a UNIX socket")
```
AkihiroSuda, line 26: Do we still need this?
For security reason I don't like listening on TCP. (Even though we don't leak critical info from the guest agent now)

examples/experimental/wsl2.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+images:
+# Source: https://github.com/runfinch/finch-core/blob/main/Dockerfile
+- location: "https://deps.runfinch.com/common/x86-64/finch-rootfs-production-amd64-1690920103.tar.zst"
+  arch: "x86_64"
+  digest: "sha256:53f2e329b8da0f6a25e025d1f6cc262ae228402ba615ad095739b2f0ec6babc9"
+
+mountType: wsl2 
+
+# Use system because of an error when setting up RootlessKit (see )
```
AkihiroSuda, line 13: see ?

AkihiroSuda (dismissed):
Thanks, overall looks good.
I still don't like the quoting around ssh commands, but it can be improved later

cmd/lima-guestagent/daemon_linux.go
pendo324: Not required, I'll remove it

pkg/sshutil/sshutil.go
afbjorklund: I also ran into all kinds of weird quoting issues, but not an expert sorry (I would rather run bash, on windows)

afbjorklund:
It is still weird to run Lima in a container, so I am a bit skeptic about it (similar to running Lima in Docker...)

pendo324:
> That is why I preferred having a Hyper-V driver.

While I don't disagree that HyperV would be good to add to Lima, I don't think WSL2 is a bad thing to have as an option even after HyperV support is added. It seems like Microsoft is committed to making incremental improvements (for example, automatic allocation/reclamation of resources) to WSL2 that could be useful for users and we'd get "for free" (or with a modification of `/etc/wsl.conf`).

> Calling a WSL "distribution" a VM is not really _true_, and confuses the user. They are all using the same kernel?

Looking at the landscape: Podman, Rancher, and Docker Desktop on Windows all use WSL2 in some way or another so I don't think it breaks user expectation entirely. I can add more docs in Lima explaining how WSL2 works and how its different from other options if that helps.

Even if there was a HyperV option, I think some users might prefer WSL2 because they can more directly interact with the distro using e.g. `wsl -d lima-$instance-name` and other easy to use commands, vs the more intimidating HyperV interfaces.

AkihiroSuda:
> I think some users might prefer WSL2 

I guess the most common reason to prefer WSL2 over bare Hyper-V would be to share files using 9p?

afbjorklund:
It would probably be a good idea for containerd either way, e.g. Docker and Podman already use WSL2

And as far as I know, you can already use QEMU (with WHPX) so Hyper-V isn't strictly necessary either?

examples/experimental/wsl2.yaml
pendo324: Forgot link, fixed in latest revision. Also added docs

pendo324:
Addressed comments and rebased. Also added documentation. Would be good to merge #1726 first since it's just extracted from @afbjorklund PR. I can just squash it into this PR if that's preferable though.

jandubois:
> I guess the most common reason to prefer WSL2 over bare Hyper-V would be to share files using 9p?

Another one is that WSL2 is supported in Windows Home whereas Hyper-V requires at least Windows Pro. This is purely a licensing issue as WSL2 is running on top of Hyper-V.

pendo324:
Not sure why the tests are failing, but it doesn't seem related. Should I keep retrying @AkihiroSuda?

AkihiroSuda:
> Not sure why the tests are failing, but it doesn't seem related. Should I keep retrying @AkihiroSuda?

CI for Colima seems consistently failing after restarting several times. Maybe a regression in this PR?

pendo324:
> > Not sure why the tests are failing, but it doesn't seem related. Should I keep retrying @AkihiroSuda?
> 
> CI for Colima seems consistently failing after restarting several times. Maybe a regression in this PR?

Found the bug (typo) and fixed it (tests passed locally). Should be good to go after CI passes

docs/vmtype.md
```diff
@@ -45,3 +45,36 @@ mountType: "virtiofs"
   kernel v6.3 and later should boot, as long as it is booted via GRUB.
   https://github.com/lima-vm/lima/issues/1577#issuecomment-1565625668
   The issue is fixed in macOS 13.5.
+
+## WSL2
+> **Warning**
+> "wsl2" mode is experimental
+
+| :zap: Requirement | Lima >= 0.14, macOS >= 13.0 |
```
AkihiroSuda, line 53: Seems a copy paste error

AkihiroSuda:
Thanks, almost looks good, but needs rebase

docs/vmtype.md
pendo324: Thanks, fixed in latest revision

pendo324:
> Thanks, almost looks good, but needs rebase

Thanks. Just rebased and addressed your new comment

AkihiroSuda (approved):
Thanks

pendo324:
Couldn't repro test error locally:

```
$ go test -timeout 30s -v -run ^TestGetPorts github.com/lima-vm/lima/pkg/guestagent/kubernetesservice

=== RUN   TestGetPorts
=== RUN   TestGetPorts/nodePort_serivce
=== RUN   TestGetPorts/loadBalancer_service
=== RUN   TestGetPorts/clusterIP_service
--- PASS: TestGetPorts (0.00s)
    --- PASS: TestGetPorts/nodePort_serivce (0.00s)
    --- PASS: TestGetPorts/loadBalancer_service (0.00s)
    --- PASS: TestGetPorts/clusterIP_service (0.00s)
PASS
ok      github.com/lima-vm/lima/pkg/guestagent/kubernetesservice        0.440s
```

Is it possible to re-run the step?

AkihiroSuda:
Restarted CI

AkihiroSuda:
PTAL:
- https://github.com/lima-vm/lima/issues/1781

PR #1725, opened 2023-08-10
Don't delete local socket when reverse

### LINKED ISSUE #1724: Reverse unix socket forwarding in Mac deletes the source unix socket.
### Description

limactl version 0.16.0

I'm trying to run Docker Desktop on my M2 Mac, and connecting to its unix socket from inside the lima vm. So I created this configuration:

```
portForwards:
  hostSocket: "/Users/rangelreale/.docker/run/docker.sock"
  reverse: true
```

it seems to work, but the unix socket stops work both in the host and in the guest
host:
```
$ docker container ls
Cannot connect to the Docker daemon at unix:///Users/rangelreale/.docker/run/docker.sock. Is the docker daemon running?
```

guest:
```
docker container ls
error during connect: Get "http://%2Frun%2Fuser%2F501%2Fdocker.sock/v1.24/containers/json": read unix @->/run/user/501/docker.sock: read: connection reset by peer
```

It seems that when using `reverse: true`, the host socket is being deleted, according to @afbjorklund in slack.

afbjorklund:
When we are forwarding a local socket, we should not delete it.

Theoretically we could delete the remote socket, if it exists.

Closes #1724

* https://github.com/lima-vm/lima/pull/836

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -597,9 +597,9 @@ func forwardSSH(ctx context.Context, sshConfig *ssh.SSHConfig, port int, local, 
 				logrus.Infof("Forwarding %q (host) to %q (guest)", local, remote)
 			} else {
 				logrus.Infof("Forwarding %q (guest) to %q (host)", remote, local)
-			}
-			if err := os.RemoveAll(local); err != nil {
-				logrus.WithError(err).Warnf("Failed to clean up %q (host) before setting up forwarding", local)
+				if err := os.RemoveAll(local); err != nil {
+					logrus.WithError(err).Warnf("Failed to clean up %q (host) before setting up forwarding", local)
+				}
 			}
 			if err := os.MkdirAll(filepath.Dir(local), 0750); err != nil {
 				return fmt.Errorf("can't create directory for local socket %q: %w", local, err)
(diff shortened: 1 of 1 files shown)


afbjorklund:
Actually we _should_ delete the remote socket, otherwise restart fails.

WARN[0031] [hostagent] Failed to set up forward from "/var/tmp/docker.sock" (guest) to "/var/run/docker.sock" (host)

pkg/hostagent/hostagent.go
```diff
@@ -568,6 +568,21 @@ const (
...
+		"127.0.0.1",
+		"--",
+	)
+	args = append(args, command...)
+	cmd := exec.CommandContext(ctx, sshConfig.Binary(), args...)
+	if out, err := cmd.Output(); err != nil {
+		return fmt.Errorf("failed to run %v: %q: %w", cmd.Args, string(out), err)
+	}
+	return nil
+}
```
AkihiroSuda, line 584: Shouldn't we use `sshutil.SSHOpts` ?

afbjorklund: We are, it is getting passed in sshConfig as far as I know?

```go
        sshConfig := &ssh.SSHConfig{
                AdditionalArgs: sshutil.SSHArgsFromOpts(sshOpts),
        }
```

AkihiroSuda (approved):
Thankss

AkihiroSuda:
Rebasing may fix the CI failure

AkihiroSuda (approved):
Thanks, but we should have a test for this

PR #1726, opened 2023-08-10
feat: add Register/Unregister to Driver interface

pendo324:
Addresses comment on https://github.com/lima-vm/lima/pull/1721#discussion_r1286619203 regarding reusing the Register/Unregister interface for both the new VBox and WSL2 drivers.

Refactored the implementation out of @afbjorklund's PR https://github.com/lima-vm/lima/pull/1277. I didn't see a call to `Register`, so I just added it before `Start` in the `pkg/hostagent`. Also added a nil instantiation of the `BaseDriver` in `pkg/driver` to help with typechecking/noticing if there are unimplemented members of the interface (can remove).

Added @afbjorklund as a co-author since I mostly just refactored. Let me know if that makes sense to you, happy to change it.

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -35,6 +35,14 @@ type Driver interface {
 	// It returns error if there are any errors during Stop
 	Stop(_ context.Context) error
 
+	// Register will add an instance to a registry.
+	// It returns error if there are any errors during Register
+	Register(_ context.Context) error
+
+	// Unregister will perform any cleanup related to the vm instance.
+	// It returns error if there are any errors during Unregister
+	Unregister(_ context.Context) error
+
 	ChangeDisplayPassword(_ context.Context, password string) error
 
 	GetDisplayConnection(_ context.Context) (string, error)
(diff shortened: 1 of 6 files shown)


pkg/driver/driver.go
```diff
@@ -35,6 +35,14 @@ type Driver interface {
 	// It returns error if there are any errors during Stop
 	Stop(_ context.Context) error
 
+	// Register will add an instance to a registry.
+	// It returns error if there are any errors during Register
+	Register(_ context.Context) error
```
balajiv113, line 40: Does create and delete instead of Register and Unregister makes more sense ?

We call unregister/delete -> during delete of instance
We call register/create -> during either create/start of instance

pendo324: Good point, just moved it to `createInstance`

afbjorklund: The VM was already created or deleted, the API call just removed it from the global definition (in VirtualBox)

If you didn't do this, it would still be listed in the GUI - just gave a "file not found" when you tried to access it.

We actually never called `Register`, it was easier to use the `-register` parameter.

The main reason to add it to the API, was so that the driver could decide what to do.

afbjorklund: For most drivers these would be no-op. But the "hook" was needed, for the vbox driver.

balajiv113 (approved):

AkihiroSuda (approved):
Thanks

PR #1728, opened 2023-08-11
templates: remove nomad

AkihiroSuda:
Nomad is no longer free software:
https://github.com/hashicorp/nomad/commit/b3e30b1dfa185d9437a25830522da47b91f78816


afbjorklund:
It _could_ be moved to "deprecated", since Nomad 1.6.1 is still under MPL 2.0 and supported until next year:

https://www.hashicorp.com/license-faq#What-products-will-be-covered-by-BSL-1.1-in-their-next-release

But I don't think very many are using this template, and they could resurrect it from the git history if so?

Another issue is that Roblox doesn't seem to want to main [the nomad containerd plugin](https://github.com/Roblox/nomad-driver-containerd) anymore, either.

AkihiroSuda:
> But I don't think very many are using this template, and they could resurrect it from the git history if so?

👍

balajiv113 (dismissed):

Issue #1499, opened 2023-04-15
Web site (https://lima-vm.io/): migrate to docsy, mkdocs, gitbook, or something else


AkihiroSuda:
The https://lima-vm.io/ website just (poorly) renders markdown on the client side.
We should consider migrating the website to docsy, mkdocs, gitbook, or something else.

## Requirements
- The site SHOULD be multi-versioned (v0.15.x, v0.16.x, master, ...)
- The site MAY support multiple languages, but its maintenance cost will be probably high. Just linking to a machine translation site might be enough.

- The site MUST have the `limactl` CLI reference
- The site MUST have the built-in template list. Just linking to https://github.com/lima-vm/lima/blob/master/examples/README.md might be enough.
- The site SHOULD have the YAML reference

## Resources
Netlify base directory: https://github.com/lima-vm/lima/tree/master/website

afbjorklund:
Would it be a good idea to keep the web site out of the lima code repository ?

The monorepo can quickly get out of hand, like it did for minikube (with hugo)

AkihiroSuda:
> Would it be a good idea to keep the web site out of the lima code repository ?

I'd prefer monorepo

balajiv113:
I will work on this .. Will try to provide some basic website within this week

afbjorklund:
The Podman Desktop team added a "white" version of the logo, for use on black backgrounds (for the "dark mode")

black (light)
![](https://raw.githubusercontent.com/containers/podman-desktop/ab3434da2f0ab70eb9679e16d4e210e66512dee3/extensions/lima/logo-light.png)

white (dark)
![](https://raw.githubusercontent.com/containers/podman-desktop/ab3434da2f0ab70eb9679e16d4e210e66512dee3/extensions/lima/logo-dark.png)

We probably want to add at least the SVG version of it to the repository, for use in:

https://github.com/cncf/artwork/blob/master/examples/sandbox.md#lima-logos

So far there is no movement on the icon situation, but then again it is probably not needed for the website either...

* https://github.com/lima-vm/lima/issues/268

PR #1731, opened 2023-08-12
README.md: remove "(on macOS, in most cases)"

AkihiroSuda:
Lima is also useful on Linux hosts and can be an alternative to Vagrant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,13 +5,9 @@ Translations: [日本語(Japanese)](README.ja.md) [简体中文（Simplified Chi
 
 ![Lima logo](./docs/images/lima-logo-01.svg)
 
-# Lima: Linux virtual machines (on macOS, in most cases)
+# Lima: Linux Machines
 
-Lima launches Linux virtual machines with automatic file sharing and port forwarding (similar to WSL2), and [containerd](https://containerd.io).
-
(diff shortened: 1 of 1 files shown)


afbjorklund (approved):

afbjorklund:
We might even want to mention provisioning in the "bullet point" list:

* Automatic file sharing

* Automatic port forwarding

* Built-in support for containerd (Other container engines can be used too)

* Built-in support for provisioning (Allow automatic customization of instance)

* Intel on Intel

* ARM on Intel

* ARM on ARM

* Intel on ARM

If we want to put some more emphasis on the "Vagrant replacement".

AkihiroSuda:
> We might even want to mention provisioning in the "bullet point" list:

I guess we can revisit this when we build a "real" website:
- https://github.com/lima-vm/lima/issues/1499

fwilhe:
Awesome. I've been using vagrant-libvirt before, but lima actually seems to work more consistently, both on Linux and on macos. I'm very happy with it so far. Thanks for working on it 👍

afbjorklund:
If you want to continue the Vagrant discussion, you can use https://github.com/lima-vm/lima/discussions

PR #1735, opened 2023-08-12
Add OS to lima.yaml for future guests

afbjorklund:
Avoid hardcoding "linux" everywhere, use OS instead (like Arch).

~~TODO: Add proper cidata template parameter for LIMA_CIDATA_HOME~~

There are some other hardcoded Linux or Systemd specific things too:

`/bin/bash`  or `/run`, those are not addressed here and are left as is.

----

Doing the OS refactoring in a separate PR, makes it easier to add OS

* #1509


afbjorklund:
Home directory parameter here:
* https://github.com/lima-vm/lima/pull/1736

AkihiroSuda (approved):
Thanks

PR #1736, opened 2023-08-12
Add variable for the guest home directory

afbjorklund:
Just a refactoring, to avoid hardcoding `/home` and `linux` everywhere in cidata.

### DIFF
--- a/docs/internal.md
+++ b/docs/internal.md
@@ -151,6 +151,7 @@ The volume label is "cidata", as defined by [cloud-init NoCloud](https://cloudin
 - `LIMA_CIDATA_MNT`: the mount point of the disk. `/mnt/lima-cidata`.
 - `LIMA_CIDATA_USER`: the user name string
 - `LIMA_CIDATA_UID`: the numeric UID
+- `LIMA_CIDATA_HOME`: the guest home directory
 - `LIMA_CIDATA_HOSTHOME_MOUNTPOINT`: the mount point of the host home directory, or empty if not mounted
 - `LIMA_CIDATA_MOUNTS`: the number of the Lima mounts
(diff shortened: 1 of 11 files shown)


docs/internal.md
```diff
@@ -151,6 +151,7 @@ The volume label is "cidata", as defined by [cloud-init NoCloud](https://cloudin
 - `LIMA_CIDATA_MNT`: the mount point of the disk. `/mnt/lima-cidata`.
 - `LIMA_CIDATA_USER`: the user name string
 - `LIMA_CIDATA_UID`: the numeric UID
+- `LIMA_CIDATA_HOME`: the guest home directory
```
AkihiroSuda, line 154: Maybe:
```suggestion
- `LIMA_CIDATA_USER_HOME`: the guest home directory
```

Because we may potentially use `LIMA_CIDATA_HOME` for something like `LIMA_HOME` which is not a user home directory?

afbjorklund: It was supposed to be simple wrappers for $USER and $HOME, we can give them long and strange names like USER_NAME and USER_HOME but I'm not sure it will help? Better do something like LIMA_CIDATA_HOSTHOME (or LIMA_CIDATA_HOST_LIMA_HOME, or whatever it would end up looking like) when needed

afbjorklund:
It still duplicates the `.linux` in a lot of places, so there probably should be a wrapper method for that construct.

`fmt.Sprintf("/home/%s.linux", u.Username)`

Maybe it should even set Homedir in the `LimaUser`

docs/internal.md
AkihiroSuda: This is not visible to end users anyway, so I don't have a strong opinion

AkihiroSuda (approved):
Thanks

PR #1738, opened 2023-08-12
Delete the conf file when instance is stopped

### LINKED ISSUE #1737: Remove conf file again, when the instance shuts down
### Description

Currently when we forward a socket, it is created when the instance is started and removed when the instance is stopped.

e.g. docker, podman

But when we copy client configuration, it is created when the instance is started - but _never_ removed, even if it is invalid...

e.g. k8s, k3s

It would be good to have an optional flag, to remove the file again when the instance is stopped:

```yaml
# Copy files from the guest to the host. Copied after provisioning scripts have been completed.
# copyToHost:
# - guest: "/etc/myconfig.cfg"
#   host: "{{.Dir}}/copied-from-guest/myconfig"
# # "guest" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
# # "host" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
# # default: delete: false
copyToHost:
- guest: "/etc/kubernetes/admin.conf"
  host: "{{.Dir}}/copied-from-guest/kubeconfig.yaml"
  delete: true
```

So when `delete: true` is set, the local file will be deleted again when the instance is stopped.

This will avoid trying to contact the local cluster, only to be met with a connection error or timeout.

----

* https://github.com/lima-vm/lima/issues/1300

afbjorklund:
Closes #1737

start
```
INFO[0027] [hostagent] The final requirement 1 of 1 is satisfied 
INFO[0027] [hostagent] Copying config from /etc/rancher/k3s/k3s.yaml to /home/anders/.lima/k3s/copied-from-guest/kubeconfig.yaml 
```

stop
```
INFO[0000] [hostagent] Deleting /home/anders/.lima/k3s/copied-from-guest/kubeconfig.yaml 
```

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -379,8 +379,10 @@ networks:
 # copyToHost:
 # - guest: "/etc/myconfig.cfg"
 #   host: "{{.Dir}}/copied-from-guest/myconfig"
+# # deleteOnStop: false
 # # "guest" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
 # # "host" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
+# # "deleteOnStop" will delete the file from the host when the instance is stopped.
 
 # Message. Information to be shown to the user, given as a Go template for the instance.
 # The same template variables as for listing instances can be used, for example {{.Dir}}.
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -156,6 +156,7 @@ probes:
 copyToHost:
 - guest: "/etc/kubernetes/admin.conf"
   host: "{{.Dir}}/copied-from-guest/kubeconfig.yaml"
+  deleteOnStop: true
 message: |
   To run `kubectl` on the host (assumes kubectl is installed), run the following commands:
   ------
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -210,8 +210,9 @@ type PortForward struct {
 }
 
 type CopyToHost struct {
-	GuestFile string `yaml:"guest,omitempty" json:"guest,omitempty"`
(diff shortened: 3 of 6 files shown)


pkg/limayaml/limayaml.go
```diff
@@ -208,6 +208,7 @@ type PortForward struct {
 type CopyToHost struct {
 	GuestFile string `yaml:"guest,omitempty" json:"guest,omitempty"`
 	HostFile  string `yaml:"host,omitempty" json:"host,omitempty"`
+	Delete    bool   `yaml:"delete,omitempty" json:"delete,omitempty"`
```
AkihiroSuda, line 211: Could you explain this in:
https://github.com/lima-vm/lima/blob/fa053a3d7c439b5f363eb49df0ea123d628bd33e/examples/default.yaml#L378-L383

examples/k8s.yaml
```diff
@@ -156,6 +156,7 @@ probes:
 copyToHost:
 - guest: "/etc/kubernetes/admin.conf"
   host: "{{.Dir}}/copied-from-guest/kubeconfig.yaml"
+  delete: true
```
AkihiroSuda, line 159: `rke2.yaml` has to be updated?

pkg/limayaml/limayaml.go
AkihiroSuda: `DeleteOnStop` might be more descriptive

examples/default.yaml
```diff
@@ -379,8 +379,10 @@ networks:
 # copyToHost:
 # - guest: "/etc/myconfig.cfg"
 #   host: "{{.Dir}}/copied-from-guest/myconfig"
+# # default: deleteOnStop: false
```
AkihiroSuda, line 382:
```suggestion
#   deleteOnStop: false
```

AkihiroSuda (approved):
Thanks

Issue #1742, opened 2023-08-12
⚠️ QEMU (homebrew) is broken on Intel: `[hostagent] Driver stopped due to error: "signal: abort trap"` (or `"exit status 255"`) ... `QEMU has already exited`


AkihiroSuda:
**Update (Aug 27, 2023)**: The issue is solved again in `8.1.0-1` (https://github.com/Homebrew/brew/pull/15903 , https://github.com/Homebrew/homebrew-core/pull/140596). Run `brew reinstall -f --force-bottle qemu` to install the updated v8.1.0 bottle.

**Update (Aug 23, 2023)**: This seems to be broken again in v8.1.0  😞 (https://github.com/Homebrew/homebrew-core/issues/140244) . See the `Workarounds` below.

**Update (Aug 14, 2023)**: The issue is now solved in https://github.com/Homebrew/homebrew-core/pull/139492 . 
Run `brew reinstall -f --force-bottle qemu` to install the updated v8.0.4 bottle.

- - -
Homebrew bottle of QEMU v8.0.4 (Intel) ~is~ was broken due to a signing issue: https://github.com/Homebrew/homebrew-core/pull/139409
```console
$ limactl start
...
[hostagent] Driver stopped due to error: "signal: abort trap" 
...
[hostagent] QEMU has already exited
...

$ qemu-system-x86_64 -accel hvf
qemu-system-x86_64: -accel hvf: Unknown Error
Abort trap: 6

$ codesign --verify /usr/local/Cellar/qemu/8.0.4/bin/qemu-system-x86_64 
/usr/local/Cellar/qemu/8.0.4/bin/qemu-system-x86_64: invalid signature (code or signature have been modified)
In architecture: x86_64
```
(The error message can be also `[hostagent] Driver stopped due to error: "exit status 255"`)

### Option 1: Downgrade QEMU to v8.0.3

```bash
brew uninstall qemu
curl -OSL https://raw.githubusercontent.com/Homebrew/homebrew-core/dc0669eca9479e9eeb495397ba3a7480aaa45c2e/Formula/qemu.rb
brew install ./qemu.rb
```

### Option 2: Install QEMU from the source

```bash
brew uninstall qemu
brew install --build-from-source qemu
```

### Option 3: Sign the QEMU binary locally
Lima v0.17.2 shows a prompt to suggest applying this workaround.

```bash
cat >entitlements.xml <<EOF
<?xml version="1.0" encoding="UTF-8"?>
[...]

PR #1743, opened 2023-08-13
qemu: ask to sign QEMU binary when the binary is not properly signed

AkihiroSuda:
```console
$ limactl start
WARN[0000] QEMU binary "/usr/local/bin/qemu-system-x86_64" is not properly signed with the "com.apple.security.hypervisor" entitlement  error="failed to run [codesign --verify /usr/local/bin/qemu-system-x86_64]: exit status 1 (out=\"/usr/local/bin/qemu-system-x86_64: invalid signature (code or signature have been modified)\\nIn architecture: x86_64\\n\")"
? Try to sign "/usr/local/bin/qemu-system-x86_64" with the "com.apple.security.hypervisor" entitlement? Yes
...
```

- - -

Workaround for:
- #1742

This workaround is needed because Homebrew's QEMU binary is not properly signed since v8.0.4. https://github.com/Homebrew/homebrew-core/pull/139409#issuecomment-1676165225

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -9,11 +9,14 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"text/template"
 	"time"
 
 	"github.com/lima-vm/lima/pkg/driver"
 	"github.com/lima-vm/lima/pkg/driverutil"
+	"github.com/lima-vm/lima/pkg/qemu"
+	"github.com/lima-vm/lima/pkg/qemu/entitlementutil"
 
 	"github.com/lima-vm/lima/pkg/downloader"
 	"github.com/lima-vm/lima/pkg/fileutils"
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
cc @lima-vm/maintainers 

I'll release v0.17.2 after merging this workaround.

afbjorklund:
Wouldn't this sign any random qemu binary in the path ?

AkihiroSuda:
> Wouldn't this sign any random qemu binary in the path ?

Sorry what do you mean?

afbjorklund:
How do we know that the current qemu is the authentic homebrew ?

AkihiroSuda:
> How do we know that the current qemu is the authentic homebrew ?

We may check if the binary path contains "/Cellar/" ?

AkihiroSuda:
Modified to open a `Y/n` prompt before signing.
NOP if tty is not available.

pkg/start/start.go
```diff
@@ -9,11 +9,14 @@ import (
...
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"text/template"
 	"time"
 
 	"github.com/lima-vm/lima/pkg/driver"
 	"github.com/lima-vm/lima/pkg/driverutil"
+	"github.com/lima-vm/lima/pkg/qemu"
+	"github.com/lima-vm/lima/pkg/qemu/entitlementutil"
```
AkihiroSuda, line 19: (I don't want to let `pkg/start` depend on a specific driver, but I'm going to merge this right now anyway, as the Homebrew issue https://github.com/lima-vm/lima/issues/1742 is critical. We can refactor this later.)

mggluscevic59:
On Mac Big Sur entitlement check exits on: `codesign --display --entitlements - --xml /usr/local/bin/qemu-system-x86_64
` because that command is a bit different - it does **not** have `--xml`

Is there a way to turn off entitlements check all together?

AkihiroSuda:
> Big Sur

Addressed in:
- https://github.com/lima-vm/lima/pull/1981

Will be in v0.19

PR #1745, opened 2023-08-13
Validate arch against image not default

### LINKED ISSUE #1744: Bug when checking arch of kernel and initrd optional images
### Description

Currently we check the architecture against a hardcoded global.

But that doesn't work, when the template allows multiple arch:
(i.e. when the arch is no longer hardcoded in the template itself)

`FATA[0000] failed to load YAML file "examples/experimental/riscv64.yaml": images[0].kernel has unexpected architecture "riscv64", must be "x86_64" `

So the kernel architecture should be checked against the image.
And we should not skip setting the arch (as usual), for each image.

Otherwise this stops merging the riscv64.yaml into the ubuntu.yaml.

afbjorklund:
Closes #1744


AkihiroSuda (approved):
Thanks

PR #1750, opened 2023-08-15
fix(lint): update homebrew formula links

pendo324:
Noticed that link linting started failing on another PR (https://github.com/lima-vm/lima/pull/1726) because of the new sharding of Homebrew formulas (see https://github.com/Homebrew/homebrew-core/pull/139592).


AkihiroSuda (approved):
Thanks

PR #1763, opened 2023-08-25
Use pkgs.k8s.io instead of apt.k8s.io for k8s

afbjorklund:
The package hosting is moving from Google to OBS.

https://kubernetes.io/blog/2023/08/15/pkgs-k8s-io-introduction/

There is now one repository per release version.

It is the exact _same_ binary content, as is available on [dl.k8s.io](https://www.downloadkubernetes.com/)

----

No building, just packaging:

```
12ea68bfef0377ccedc1a7c98a05ea76907decbcf1e1ec858a60a7b9b73211bb  kubeadm_1.28.0-00_amd64/usr/bin/kubeadm
12ea68bfef0377ccedc1a7c98a05ea76907decbcf1e1ec858a60a7b9b73211bb  kubeadm_1.28.0-1.1_amd64/usr/bin/kubeadm
```

https://dl.k8s.io/v1.28.0/bin/linux/amd64/kubeadm.sha256

PR #1775, opened 2023-09-04
Update k8s gpg key and kubernetes repository url

waku893:
The k8s gpg key and repository URL are old, so I updated them to new URLs.


afbjorklund (changes requested):
* https://github.com/lima-vm/lima/pull/1763

PR #1754, opened 2023-08-17
Revamp lima website using docsy

balajiv113:
Fixes #1499 

Creating as draft to keep track of the progress

- [X] Home page
- [x] Documentation - Installation
- [x] ~~Documentation - Getting Started~~
- [x] Documentation - Templates
- [x] ~~Documentation - Architecture and other informations present under docs folder~~
- [x] Documentation - Cli reference
- [x] Contribution
- [x] FAQ

PR #1778, opened 2023-09-05
Support for generating cli reference with different output formats

balajiv113:
Like man, docsy

Reference: https://github.com/lima-vm/lima/pull/1754#issuecomment-1705219198

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,13 @@ _output/share/lima/lima-guestagent.Linux-riscv64:
 .PHONY: manpages
 manpages: _output/bin/limactl$(exe)
 	@mkdir -p _output/share/man/man1
-	$< generate-man _output/share/man/man1 \
+	$< generate-doc _output/share/man/man1 \
+		--output _output --prefix $(PREFIX)
+
+.PHONY: docsy
+docsy: _output/bin/limactl$(exe)
+	@mkdir -p website/_output/docsy
+	$< generate-doc --type docsy website/_output/docsy \
 		--output _output --prefix $(PREFIX)
 
 .PHONY: diagrams
--- a/cmd/limactl/gendoc.go
+++ b/cmd/limactl/gendoc.go
@@ -2,9 +2,11 @@ package main
 
 import (
 	"bytes"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/cpuguy83/go-md2man/v2/md2man"
 	"github.com/sirupsen/logrus"
@@ -12,20 +14,21 @@ import (
 	"github.com/spf13/cobra/doc"
 )
 
-func newGenManCommand() *cobra.Command {
+func newGenDocCommand() *cobra.Command {
 	genmanCommand := &cobra.Command{
-		Use:    "generate-man DIR",
-		Short:  "Generate manual pages",
+		Use:    "generate-doc DIR",
+		Short:  "Generate cli-reference pages",
 		Args:   WrapArgsError(cobra.MinimumNArgs(1)),
-		RunE:   genmanAction,
+		RunE:   gendocAction,
 		Hidden: true,
 	}
+	genmanCommand.Flags().String("type", "man", "Output type  (man, docsy)")
 	genmanCommand.Flags().String("output", "", "Output directory")
 	genmanCommand.Flags().String("prefix", "", "Install prefix")
 	return genmanCommand
 }
 
-func genmanAction(cmd *cobra.Command, args []string) error {
+func gendocAction(cmd *cobra.Command, args []string) error {
 	output, err := cmd.Flags().GetString("output")
 	if err != nil {
 		return err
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -137,7 +137,13 @@ _output/share/lima/lima-guestagent.Linux-riscv64:
...
 	@mkdir -p _output/share/man/man1
-	$< generate-man _output/share/man/man1 \
+	$< generate-doc _output/share/man/man1 \
+		--output _output --prefix $(PREFIX)
+
+.PHONY: docsy
+docsy: _output/bin/limactl$(exe)
+	@mkdir -p _output/share/docsy
+	$< generate-doc --type docsy _output/share/docsy \
 		--output _output --prefix $(PREFIX)
```
AkihiroSuda, line 147: The docsy files are not expected to be installed on the host, so it shouldn't be generated under `_output/share`.

Probably you want to mkdir something like `_output.website` or `website/_output`.

cmd/limactl/gendoc.go
```diff
@@ -2,30 +2,33 @@ package main
...
 	"github.com/spf13/cobra/doc"
 )
 
-func newGenManCommand() *cobra.Command {
+func newGenDocCommand() *cobra.Command {
 	genmanCommand := &cobra.Command{
-		Use:    "generate-man DIR",
-		Short:  "Generate manual pages",
+		Use:    "generate-doc DIR",
+		Short:  "Generate cli-reference pages",
```
AkihiroSuda, line 20: It might not be necessary to rename the command.
Docsy manuals are still manuals, but in a different format.

Makefile
afbjorklund: This was the main reason for why I never added the markdown or html (or text) rendering to the code.

I did not want redundant copies of man pages installed on the drive, better to use xman (or similar) ?

cmd/limactl/gendoc.go
afbjorklund: The command is internal anyway, so it is not important.

Those were just the command names that I used, for them...

afbjorklund:
I don't think docsy has anything to do with this ? The generated pages seem to be markdown

cmd/limactl/gendoc.go
```diff
@@ -38,11 +41,33 @@ func genmanAction(cmd *cobra.Command, args []string) error {
...
 	if err != nil {
 		return err
 	}
 	dir := args[0]
+	switch outputType {
+	case "man":
+		if err := genMan(cmd, dir); err != nil {
+			return err
+		}
+	case "docsy":
```
afbjorklund, line 58: I think this should be "markdown", and genMarkdown

balajiv113: The generated contents are md. But in the generated md we customise it for using it with docsy. 

That's why i kept it as docsy. So that later if we need to render markdown as is it will be fine

cmd/limactl/gendoc.go
```diff
@@ -69,14 +94,25 @@ and $LIMA_WORKDIR.
...
+		name = strings.ReplaceAll(name, "limactl_", "")
+		name = strings.ReplaceAll(name, "_", " ")
+		name = strings.TrimSuffix(name, filepath.Ext(name))
+		return fmt.Sprintf(`---
+title: %s
+weight: 3
+---
+`, name)
+	}, func(s string) string {
+		//Use ../ for move one folder up for docsy
```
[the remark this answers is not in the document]
balajiv113: GenMarkdownTree renders `see also` section.

This section contains links to other commands. This is not in sync on how docsy renders. That's why using this to patch those links to work with docsy

Makefile
balajiv113: Fixed. Moved to _output/website

cmd/limactl/gendoc.go
balajiv113: Yes exactly. I didn't want to have that confusion with man. 
So just renamed it to more generic one so its clear.

Makefile
AkihiroSuda: >  Moved to _output/website

Doesn't seem correct, as this makes the `website` directory  to be installed onto `/usr/local/website`

https://github.com/lima-vm/lima/blob/de1b3ee69561953f71d03cf8fb4b11d72482a350/Makefile#L148-L154

AkihiroSuda (changes requested):
https://github.com/lima-vm/lima/pull/1778/files#r1316745325

Makefile
balajiv113: My bad missed the install flow. Moved it to website/_output.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@afbjorklund LGTY?

afbjorklund:
I'm OK with having it custom for the website, and only install the man pages on the local machine.

afbjorklund (approved):

PR #1790, opened 2023-09-13
Add basic Kconfig support for lima-guestagent

afbjorklund:
Make it possible to choose which lima-guestagent binaries to build...

This makes it possible to have optional binaries, such as FreeBSD etc.

There is support for "make config":

```
*
* Lima
*
guestagent OS: Linux (GUESTAGENT_OS_LINUX) [Y/n/?] 
guestagent Arch: x86_64 (GUESTAGENT_ARCH_X8664) [Y/n/?] 
guestagent Arch: aarch64 (GUESTAGENT_ARCH_AARCH64) [Y/n/?] 
guestagent Arch: armv7l (GUESTAGENT_ARCH_ARMV7L) [Y/n/?] 
guestagent Arch: riscv64 (GUESTAGENT_ARCH_RISCV64) [Y/n/?] 
#
# No change to .config
#
```

And also for "make menuconfig":

lima-kconfig-menuconfig

It is possible to just edit the `.config`file (marked `DO NOT EDIT`).

Regenerating it requires one of the standalone Kconfig programs:

* https://github.com/WangNan0/kbuild-standalone

* https://packages.ubuntu.com/kconfig-frontends

* https://pypi.org/project/kconfiglib

See also: https://www.kernel.org/doc/html/next/kbuild/kconfig-language.html


Makefile
```diff
@@ -7,6 +7,9 @@ TAR ?= tar
 ZIP ?= zip
 PLANTUML ?= plantuml # may also be "java -jar plantuml.jar" if installed elsewhere
 
+KCONFIG_CONF ?= conf
+KCONFIG_MCONF ?= mconf
```
AkihiroSuda, line 11: These commands are rarely available on macOS, so I don't think we should depend on this

afbjorklund: The alternative would be compile them from source in the project, which I would like to avoid.
There is a good rationale in https://github.com/WangNan0/kbuild-standalone (or use python?)

It is also only needed, if making major changes to the Kconfig file itself. Otherwise, edit `.config`.
The format should be quite straight-forward, even if one needs to comment out - not set as 'n'

AkihiroSuda: I guess we should just adopt `vi .config` or `./configure --enable-XXX`

afbjorklund:
We could remove the scary "DO NOT EDIT" comment, for instance python's `alldefconfig` et al does not add it.

```diff
@@ -1,9 +1,5 @@
-#
-# Automatically generated file; DO NOT EDIT.
-# Lima
-#
 CONFIG_GUESTAGENT_OS_LINUX=y
 CONFIG_GUESTAGENT_ARCH_X8664=y
 CONFIG_GUESTAGENT_ARCH_AARCH64=y
 CONFIG_GUESTAGENT_ARCH_ARMV7L=y
 CONFIG_GUESTAGENT_ARCH_RISCV64=y
```

It also has an alternative `guiconfig` (in Tkinter), which might be cross-platform enough (unlike `gconf` and `qconf`)

lima-kconfig-guiconfig

Makefile
afbjorklund: Adding support for autoconf would require more infrastructure and generate more boilerplate, imho

afbjorklund:
Making a homebrew "formula" for the kbuild-standalone programs could be a good idea?

https://github.com/WangNan0/kbuild-standalone

* `conf`

* `mconf`

Makefile
```diff
@@ -7,6 +7,9 @@ TAR ?= tar
 ZIP ?= zip
 PLANTUML ?= plantuml # may also be "java -jar plantuml.jar" if installed elsewhere
 
+KCONFIG_CONF ?= $(shell which kconfig-conf || which kbuild-conf || echo oldconfig)
```
AkihiroSuda, line 10: `which` is often missing on modern distros

AkihiroSuda: Probably better to explain the KCONFIG stuff is not required for installation with the default config.

Makefile
```diff
@@ -45,6 +48,11 @@ GO_BUILD := $(GO) build -ldflags="-s -w -X $(PACKAGE)/pkg/version.Version=$(VERS
...
 all: binaries manpages
 
+.PHONY: help
+help:
+	@echo  '  binaries        - Build all binaries'
+	@echo  '  manpages        - Build manual pages'
+
 exe: _output/bin/limactl$(exe)
 
 .PHONY: minimal
```
AkihiroSuda, line 58: `make minimal` should now generate a new `.config` and run `make binary` ?

.config
AkihiroSuda, line 1: Maybe this should be named `defconfig` and automatically copied to `.config` on `make`?

`.config` will have to be gitignored.

afbjorklund: It doesn't _have_ to be ignored, but you are right that it will be flagged as modified.

The `defconfig` is actually empty by default, it is generated using the `savedefconfig`

Makefile
afbjorklund: Not everything is configurable, and there is no mechanism for "host arch"

afbjorklund:
I added some (very) basic help text, mostly because it was mentioned in the output and interface.

`export KCONFIG_CONF=kconfig-conf`
`export KCONFIG_MCONF=kconfig-mconf`

```
*** End of the configuration.
*** Execute 'make' to start the build or try 'make help'.
```

README.md
```diff
@@ -129,6 +129,8 @@ curl -fsSL "https://github.com/lima-vm/lima/releases/download/${VERSION}/lima-${
 
 - To install Lima from the source, run `make && make install`.
 
+- To change configuration, run `make config` or edit `.config`.
```
AkihiroSuda, line 132:
```suggestion
- To change compilation configuration, run `make config` or edit `.config`. You may have to install XXX from XXX.
```

config.default
AkihiroSuda, line 1: I fear this file to be misinterpreted as the default lima YAML config.

Can we just generate the default config with heredoc?

AkihiroSuda: Or put this file to  somewhere like `Makefile.d/` or `Kconfig.d`/

afbjorklund:
It is also a problem that the python library does not implement `oldaskconfig`, the default mode.

config.default
afbjorklund: Renamed it to config.mk, hopefully there will be no confusion.

It is possible to edit this file (if not using kconfig tools).
The main difference is that it is checked into the repository.
To make user-specific config, it is better to edit `.config`.
The default make target will copy config.mk, if it is missing.

afbjorklund:
I updated the kconfig-frontends from v4.11 to v4.19, just in case. The [old repo](https://web.archive.org/web/20190901055132/ymorin.is-a-geek.org/projects/kconfig-frontends) seemed missing?

And the debian packaging only had an import of the generated files, not the original git sources.

https://github.com/afbjorklund/kconfig-frontends

~~http://ymorin.is-a-geek.org/git/kconfig-frontends~~

> 2017-05-03: 4.11.0.1 has been released
> 2013-11-09: 3.12.0.0 has been released 
> 2012-03-19: first release ever: 3.3.0-1 has been released
> 2012-01-25: this project announced

AkihiroSuda (dismissed):

README.md
```diff
@@ -129,9 +129,20 @@ curl -fsSL "https://github.com/lima-vm/lima/releases/download/${VERSION}/lima-${
...
 > Lima is not regularly tested on ARM Mac (due to lack of CI).
 
+##### Kconfig tools
+
+The tools are available as either "kconfig-frontends" or "kbuild-standalone".
+There is one `conf` for the text, and one `mconf` for the menu interface.
+
+A python implementation is available at <https://pypi.org/project/kconfiglib>.
+It can be installed with `pip install --user kconfiglib`, including `guiconfig`.
+
```
AkihiroSuda, line 145: This should be now in https://lima-vm.io/docs/installation/

afbjorklund: I think you mean `website/content/en/docs/Installation/`

afbjorklund: Done

website/content/en/docs/Installation/_index.md
```diff
@@ -56,5 +56,18 @@ cd lima
 make
 make install
 ```
+
+To change the build configuration, run `make config` or `make menuconfig`.
+
+This requires kconfig tools installed, it is also possible to edit `.config`.
+The default configuration can be found in the file `config.mk` (make syntax).
+
```
AkihiroSuda, line 64: This paragraphs should be inside `## Kconfig tools`

website/content/en/docs/Installation/_index.md
```diff
@@ -56,5 +56,18 @@ cd lima
...
+## Kconfig tools
```
AkihiroSuda, line 65:
```suggestion
## Kconfig tools (optional)
```

website/content/en/docs/Installation/_index.md
afbjorklund: It was for the user that doesn't have any tools, but edits the config as text

AkihiroSuda (approved):

PR #1803, opened 2023-09-18
examples/README.md: clean up

AkihiroSuda:
See https://lima-vm.io/docs/templates


balajiv113:
This file was added as a linking to website. Before removing we need to duplicate this content there

https://github.com/lima-vm/lima/blob/master/website/content/en/docs/Templates/_index.md?plain=1#L6

https://github.com/lima-vm/lima/blob/master/website/hugo.toml#L205

AkihiroSuda:
Ah, didn't notice 😔

PR #1805, opened 2023-09-18
mv docs/talks.md https://lima-vm.io/docs/talks/

AkihiroSuda:
The commits are intentionally unsquashed to keep the `git mv` history


jandubois (approved):
Thanks, LGTM

PR #1811, opened 2023-09-19
limactl: deprecate `limactl show-ssh` (Use `ssh -F ...` instead)

AkihiroSuda:
`limactl show-ssh` is now deprecated.
Instead, use `ssh -F ~/.lima/default/ssh.config lima-default` .

### DIFF
--- a/cmd/limactl/show_ssh.go
+++ b/cmd/limactl/show_ssh.go
@@ -4,10 +4,14 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/lima-vm/lima/pkg/sshutil"
 	"github.com/lima-vm/lima/pkg/store"
+	"github.com/lima-vm/lima/pkg/store/dirnames"
+	"github.com/lima-vm/lima/pkg/store/filenames"
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
cc @jandubois @abiosoft @pendo324  FYI

afbjorklund:
So you should use `cat ~/.lima/default/ssh.config`, if you need the config for some other program ?

AkihiroSuda:
> So you should use `cat ~/.lima/default/ssh.config`, if you need the config for some other program ?

Yes.

cmd/limactl/show_ssh.go
```diff
@@ -42,8 +45,13 @@ const showSSHExample = `
...
 func newShowSSHCommand() *cobra.Command {
 	var shellCmd = &cobra.Command{
-		Use:               "show-ssh [flags] INSTANCE",
-		Short:             "Show the ssh command line",
+		Use:   "show-ssh [flags] INSTANCE",
+		Short: "Show the ssh command line (DEPRECATED; use `ssh -F` instead)",
+		Long: `Show the ssh command line (DEPRECATED)
+
+WARNING: 'limactl show-ssh' is deprecated.
+Instead, use 'ssh -F ~/.lima/default/ssh.config lima-default' .
```
afbjorklund, line 53: Should probably not hardcode LIMA_HOME here

AkihiroSuda: Fixed

afbjorklund:
I sort of liked the old command, and think this is something of a regression...

With docker-machine and podman machine, the ssh was always hidden from the user.

The Lima approach made it simple to get the config in various formats, and set up ssh.

Now we are back to knowing location of files, and so on. Just my initial thoughts.

AkihiroSuda:
> Now we are back to knowing location of files, and so on. 


`limactl ls --format='{{.SSHConfigFile}}'` can be still used

afbjorklund:
Yeah well that just leads back to the `inspect` discussion so I think I will just abstain from "voting".

But adding the lima hosts to ~/.ssh/config was still a neat thing, no matter how it is accomplished.

Then you could `ssh` or `rsync` or what have you, using only the `lima-default`.

I thought that was a pretty neat party trick. For most other things, there's `lima`.

AkihiroSuda:
> But adding the lima hosts to ~/.ssh/config was still a neat thing, no matter how it is accomplished.

`limactl show-ssh` doesn't do this automatically.

You can add `Include ~/.lima/default/ssh.config` in `~/.ssh/config` manually though.

AkihiroSuda:
Can we merge?

PR #1817, opened 2023-09-19
Enable display device on the aarch64 architecture

afbjorklund:
The display had been disabled on ARM virtual machines.

So it only showed the TianoCore boot menu for UEFI...

The `virtio-vga` device is not available, but `virtio-gpu` is.

However, there are still some stability issues on QEMU 4.2.1


AkihiroSuda (approved):
Thanks

PR #1825, opened 2023-09-20
fix: windows directory path containing spaces

pendo324:
Without this change, Lima can't be used to create a VM in a path that contains a space. For example, if the InstanceDir is `C:\Users\My Name\.lima\data\`, all VM creations would fail.

Tested on a Windows machine.


AkihiroSuda (approved):
Thanks

PR #1828, opened 2023-09-21
fix: allow non-admin users to run limactl start

pendo324:
Remove write access for the IsVSockPortFree and GetRandomFreeVSockPort functions because Write access to those registry keys requires Administrator privileges, which would otherwise _not_ be required to run limactl start.


AkihiroSuda (approved):
Thanks

PR #1833, opened 2023-09-22
Prompt the user to also start the created instance

afbjorklund:
Now that lima doesn't start the instances by default after creation, we need to suggest the user to do so.

----

```console
$ docker.lima version
instance "docker" does not exist, run `limactl create --name=docker template://docker` to create a new instance
$ limactl create --name=docker template://docker
? Creating an instance "docker" Proceed with the current configuration
INFO[0000] Attempting to download the image              arch=x86_64 digest="sha256:d5b419272e01cd69bfc15cbbbc5700d2196242478a54b9f19746da3a1269b7c8" location="https://cloud-images.ubuntu.com/releases/22.04/release-20230729/ubuntu-22.04-server-cloudimg-amd64.img"
INFO[0001] Using cache "/home/anders/.cache/lima/download/by-url-sha256/625afb7f7a4fbe30a232043af243cca76e06866624604da5ba5ec130d56c4244/data" 
INFO[0001] Run `limactl start docker` to start the instance. 
$ docker.lima version
instance "docker" is not running, run `limactl start docker` to start the existing instance
```

Otherwise it will just show error.

```
Cannot connect to the Docker daemon at unix:///home/anders/.lima/docker/sock/docker.sock. Is the docker daemon running?
```

### DIFF
--- a/cmd/apptainer.lima
+++ b/cmd/apptainer.lima
@@ -6,6 +6,9 @@ set -eu
 if [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
   echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl create --name=$LIMA_INSTANCE template://apptainer\` to create a new instance" >&2
   exit 1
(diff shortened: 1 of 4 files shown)


AkihiroSuda (approved):
Thanks

PR #1834, opened 2023-09-22
Ignore system proxy settings by overriding with empty string

### LINKED ISSUE #1829: How to unset https_proxy with /etc/environment
### Description
Every time a new virtual machine is started, the proxy environment variable will be set. How to cancel it?

jandubois:
Setting proxy variables to the empty string doesn't make sense, so just remove them completely. This allows the user to set `http_proxy` and friends to the empty string, either in lima.yaml or in the process environment before running `limactl start`.

Fixes #1829

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -65,9 +65,12 @@ func setupEnv(y *limayaml.LimaYAML, args TemplateArgs) (map[string]string, error
 		}
 	}
 	// Replace IP that IsLoopback in proxy settings with the gateway address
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

jandubois:
It says `vz` test was cancelled, but I cannot tell why, or by whom. What does this mean?

AkihiroSuda:
> It says `vz` test was cancelled, but I cannot tell why, or by whom. What does this mean?

macos-13 CI has been just flaky, and we can't see the logs 😞

Issue #1739, opened 2023-08-12
`limactl create`: add `--plain` flag to disable mounts, port forwarding, containerd, etc.


AkihiroSuda:
The `--plain` flag will disable:
- mounts
- port forwarding (except the SSH port)
- containerd
- (maybe) guest agent process
- aptgetting sshfs packages, etc.

fwilhe2:
Question: Could this also support systems that don't have a package manager? I'm asking because I'm currently looking into image-based, immutable systems. There are various approaches to implement this, but they typically have in common that no 'normal' package manager is available. I think it would be great if lima was usable with such systems, and I see that depending on what the image provides some features of lima might not be available, but getting a basic shell via ssh would already be pretty great imo.

afbjorklund:
The downside is that those read-only systems, will **not** support all features of a regular operating system. You can install some things in `/var` (like nerdctl-full), but if for instance fuse and sshfs is missing from the image you will need to make do without it (and use e.g. virtfs). They typically also have bigger disk images, both for including the "kitchen sink" (since there is no package manager) and for doing dual-booting (with an active/passive partitoning, or with snapshots)

But those two systems are examples on what _can_ be done with the current split. The biggest issue maintaining it is that all setup scripts need to be duplicated, since it doesn't support the regular shell scripts but requires systemd units... And of course that all cloud-init configuration needs to be converted to ignition, since it doesn't support the standard. I think the current number of scripts and configs makes it tolerable, but it could stand in the way of bigger future changes?

afbjorklund:
@fwilhe2 Note that typically CoreOS did not install any software whatsoever on the host. You started a `toolbox` container, and then used the package manager inside that... Either "live", or in your Dockerfile

Similar to what WSL2 is doing with the "distributions" <sic>

fwilhe2:
Sure, I was mentioning it because a package manager is described as a requirement in the Lima readme

Bean looks interesting 👀

afbjorklund:
The shell scripts do require a package manager

jandubois:
> The shell scripts do require a package manager

Technically they don't if your base image already includes all the required prerequisites.

afbjorklund:
I found that it was possible to use the pre-installed software on CoreOS, provided that one mimicked the other setup from the boot scripts (that are not getting called, when not using cloud-init*).

[...]

PR #1840, opened 2023-09-26
Add "plain" mode (disables mounts, port forwarding, containerd, etc.)

AkihiroSuda:
```bash
limactl start --plain
```

When the "plain" mode is enabled:
- the YAML properties for mounts, port forwarding, containerd, etc. will be ignored
- guest agent will not be running.
- dependency packages like sshfs will not be installed into the VM

User-specified provisioning scripts will be still executed.

Fixes:
- #1739


pkg/start/start.go
```diff
@@ -240,7 +240,11 @@ func watchHostAgentEvents(ctx context.Context, inst *store.Instance, haStdoutPat
 				return true
 			}
 
-			logrus.Infof("READY. Run `%s` to open the shell.", LimactlShellCmd(inst.Name))
+			if *inst.Config.Plain {
+				logrus.Infof("READY. Run `ssh -F %q lima-%s` to open the shell.", inst.SSHConfigFile, inst.Name)
```
balajiv113, line 244: Y not use LimactlShellCmd(inst.Name) here ?? Ssh connection stays the same right evn for plan mode ?

AkihiroSuda: Because plain mode users are less likely to prefer lima-ist commands.
The limactl shell command continue to work.

balajiv113 (approved):

pkg/start/start.go
afbjorklund: I still think the `lima` wrapper (but only that script, not the others) belongs to the "minimal" footprint.

* https://github.com/lima-vm/lima/pull/1673

But I suppose `ssh` is even more "plain".

afbjorklund:
Here was the suggestion for a "minimal" version of the regular `lima` command, wrapping `ssh` directly instead.

```sh
#!/bin/sh
exec ssh -F "${LIMA_HOME:-$HOME/.lima}/$(echo $1 | sed s/lima-//)/ssh.config" $LIMA_INSTANCE "$@"
```

What do you think ? It is just to make the invocation more "pretty", by hiding the path to the ssh.config location...

```
INFO[0020] READY. Run `ssh -F "/home/anders/.lima/default/ssh.config" lima-default` to open the shell. 
```

----

READY. Run `lima-ssh lima-default`

PR #1841, opened 2023-09-26
CI: release: use ubuntu-20.04 for Linux binaries

AkihiroSuda:
Use an old release of Ubuntu for better glibc compatibility


afbjorklund (approved):
Sounds good

PR #1871, opened 2023-10-01
k8s: Stop using LIMA_CIDATA_SLIRP_IP_ADDRESS

afbjorklund:
When using user-v2 networks instead of slirp, the address is no longer static but dynamic.

Fortunately, we don't need to know the previous address in order to replace it with localhost...

```diff
-    server: https://192.168.5.15:6443
+    server: https://127.0.0.1:6443
```

So just remove the usage of this variable, from the k8s template.

### DIFF
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -123,7 +123,8 @@ provision:
     kubectl apply -f https://github.com/flannel-io/flannel/releases/download/v0.22.1/kube-flannel.yml
     # Control plane node isolation
     kubectl taint nodes --all node-role.kubernetes.io/control-plane-
-    sed -e "s/${LIMA_CIDATA_SLIRP_IP_ADDRESS:-192.168.5.15}/127.0.0.1/" -i $KUBECONFIG
+    # Replace the server address with localhost, so that it works also from the host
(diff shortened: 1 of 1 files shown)


examples/k8s.yaml
```diff
@@ -123,7 +123,7 @@ provision:
     kubectl apply -f https://github.com/flannel-io/flannel/releases/download/v0.22.1/kube-flannel.yml
     # Control plane node isolation
     kubectl taint nodes --all node-role.kubernetes.io/control-plane-
-    sed -e "s/${LIMA_CIDATA_SLIRP_IP_ADDRESS:-192.168.5.15}/127.0.0.1/" -i $KUBECONFIG
+    sed -e "/server:/ s|https://.*:\([0-9]*\)$|https://127.0.0.1:\1|" -i $KUBECONFIG
```
AkihiroSuda, line 126: nit: maybe we need a comment line to explain this regexp

AkihiroSuda (approved):
Thanks

AkihiroSuda (approved):

examples/k8s.yaml
afbjorklund: Added a line to explain what we are replacing here.

Once we implement multi-node, we might need to be more careful so that the other nodes can still talk to the apiserver even if we edit the config to work with the tunneled port from the host side. Same goes for k3s, then.

PR #1873, opened 2023-10-01
Add `limactl protect <INSTANCE>` to prohibit accidental removal

### LINKED ISSUE #1595: Add `limactl protect <INSTANCE>` to prohibit accidental removal
Should be implemented with (an equivalent of) `chmod +a "everyone deny delete,delete_child,file_inherit,directory_inherit"`.

We may want to add `limactl prune-instances` that will remove all unprotected instances

AkihiroSuda:
Fix #1595

### DIFF
--- a/cmd/limactl/protect.go
+++ b/cmd/limactl/protect.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func newProtectCommand() *cobra.Command {
+	var protectCommand = &cobra.Command{
+		Use:   "protect INSTANCE [INSTANCE, ...]",
+		Short: "Protect an instance to prohibit accidental removal",
+		Long: `Protect an instance to prohibit accidental removal via the 'limactl delete' command.
+The instance is not being protected against removal via '/bin/rm', Finder, etc.`,
+		Args:              WrapArgsError(cobra.MinimumNArgs(1)),
+		RunE:              protectAction,
+		ValidArgsFunction: protectBashComplete,
+	}
+	return protectCommand
+}
+
+func protectAction(_ *cobra.Command, args []string) error {
+	var errs []error
+	for _, instName := range args {
+		inst, err := store.Inspect(instName)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to inspect instance %q: %w", instName, err))
+			continue
+		}
+		if inst.Protected {
+			logrus.Warnf("Instance %q is already protected. Skipping.", instName)
+			continue
+		}
+		if err := inst.Protect(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to protect instance %q: %w", instName, err))
+			continue
+		}
+		logrus.Infof("Protected %q", instName)
+	}
+	return errors.Join(errs...)
+}
+
+func protectBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	return bashCompleteInstanceNames(cmd)
+}
--- a/cmd/limactl/unprotect.go
+++ b/cmd/limactl/unprotect.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func newUnprotectCommand() *cobra.Command {
+	var unprotectCommand = &cobra.Command{
+		Use:               "unprotect INSTANCE [INSTANCE, ...]",
+		Short:             "Unprotect an instance",
+		Args:              WrapArgsError(cobra.MinimumNArgs(1)),
+		RunE:              unprotectAction,
+		ValidArgsFunction: unprotectBashComplete,
+	}
+	return unprotectCommand
+}
+
+func unprotectAction(_ *cobra.Command, args []string) error {
+	var errs []error
+	for _, instName := range args {
+		inst, err := store.Inspect(instName)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to inspect instance %q: %w", instName, err))
+			continue
+		}
+		if !inst.Protected {
+			logrus.Warnf("Instance %q isn't protected. Skipping.", instName)
+			continue
+		}
+		if err := inst.Unprotect(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to unprotect instance %q: %w", instName, err))
+			continue
+		}
+		logrus.Infof("Unprotected %q", instName)
+	}
+	return errors.Join(errs...)
+}
+
+func unprotectBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	return bashCompleteInstanceNames(cmd)
+}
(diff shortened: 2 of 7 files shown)


cmd/limactl/protect.go
```diff
@@ -0,0 +1,40 @@
...
+func protectAction(_ *cobra.Command, args []string) error {
+	for _, instName := range args {
+		inst, err := store.Inspect(instName)
+		if err != nil {
+			return fmt.Errorf("failed to inspect instance %q: %w", instName, err)
+		}
+		if err := inst.Protect(); err != nil {
+			return fmt.Errorf("failed to protect instance %q: %w", instName, err)
+		}
+		logrus.Infof("Protectd %q", instName)
```
jandubois, line 33:
```suggestion
		logrus.Infof("Protected %q", instName)
```

cmd/limactl/unprotect.go
```diff
@@ -0,0 +1,38 @@
...
+func unprotectAction(_ *cobra.Command, args []string) error {
+	for _, instName := range args {
+		inst, err := store.Inspect(instName)
+		if err != nil {
+			return fmt.Errorf("failed to inspect instance %q: %w", instName, err)
+		}
+		if err := inst.Unprotect(); err != nil {
+			return fmt.Errorf("failed to unprotect instance %q: %w", instName, err)
+		}
+		logrus.Infof("Unprotectd %q", instName)
```
jandubois, line 31:
```suggestion
		logrus.Infof("Unprotected %q", instName)
```

cmd/limactl/protect.go
AkihiroSuda: fixed

cmd/limactl/unprotect.go
AkihiroSuda: fixed

cmd/limactl/protect.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func newProtectCommand() *cobra.Command {
+	var protectCommand = &cobra.Command{
+		Use:   "protect INSTANCE [INSTANCE, ...]",
+		Short: "Protect an instance to prohibit accidental removal",
+		Long: `Protect an instance to prohibit accidental removal via the 'limactl delete' command.
+An instance can be still removable via '/bin/rm', Finder, etc.`,
```
jandubois, line 16: Sounds weird to me; I would say "can still be removed" or "is still removable".

Or maybe even: "The instance is not being protected against removal via ...".

cmd/limactl/protect.go
```diff
@@ -0,0 +1,40 @@
...
+
+func protectAction(_ *cobra.Command, args []string) error {
+	for _, instName := range args {
+		inst, err := store.Inspect(instName)
+		if err != nil {
+			return fmt.Errorf("failed to inspect instance %q: %w", instName, err)
+		}
+		if err := inst.Protect(); err != nil {
+			return fmt.Errorf("failed to protect instance %q: %w", instName, err)
+		}
```
jandubois, line 32: Why does processing stop on the first error? So `limactl protect foo missing bar` will protect `foo` but not `bar` because `missing` did not exist?

I think it would be better to collect all errors and report them at the end.

cmd/limactl/unprotect.go
```diff
@@ -0,0 +1,38 @@
...
+		Args:              WrapArgsError(cobra.MinimumNArgs(1)),
+		RunE:              unprotectAction,
+		ValidArgsFunction: unprotectBashComplete,
+	}
+	return unprotectCommand
+}
+
+func unprotectAction(_ *cobra.Command, args []string) error {
+	for _, instName := range args {
+		inst, err := store.Inspect(instName)
```
jandubois, line 24: Shouldn't we show a warning (not an error) if the instance isn't protected?

```go
if !inst.Protected {
    logurs.warn("Instance %q isn't protected", instName)
    continue
}
```

This might alert the user that they may have specified the wrong instance name, because they clearly expected it to be protected.

jandubois (commented):
Some more feedback; have to stop for a meeting now... 😄

cmd/limactl/protect.go
AkihiroSuda: Fixed

cmd/limactl/protect.go
AkihiroSuda: Fixed

cmd/limactl/unprotect.go
AkihiroSuda: Fixed

cmd/limactl/unprotect.go
```diff
@@ -0,0 +1,46 @@
...
+	var errs []error
+	for _, instName := range args {
+		inst, err := store.Inspect(instName)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to inspect instance %q: %w", instName, err))
+			continue
+		}
+		if !inst.Protected {
+			logrus.Warn("Instance %q isn't protected", instName)
+			// try Unprotect() anyway; this may help working around pontential bugs
```
AkihiroSuda, line 33: @jandubois FYI

cmd/limactl/protect.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func newProtectCommand() *cobra.Command {
+	var protectCommand = &cobra.Command{
+		Use:   "protect INSTANCE [INSTANCE, ...]",
+		Short: "Protect an instance to prohibit accidental removal",
+		Long: `Protect an instance to prohibit accidental removal via the 'limactl delete' command.
+The instance is not being protected agains removal via '/bin/rm', Finder, etc.`,
```
jandubois, line 17:
```suggestion
The instance is not being protected against removal via '/bin/rm', Finder, etc.`,
```

cmd/limactl/unprotect.go
jandubois: I don't think this is necessary. And I find it confusing to see both messages:

```
Instance "foo" isn't protected
Unprotected "foo"
```

cmd/limactl/protect.go
```diff
@@ -0,0 +1,44 @@
...
+
+func protectAction(_ *cobra.Command, args []string) error {
+	var errs []error
+	for _, instName := range args {
+		inst, err := store.Inspect(instName)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to inspect instance %q: %w", instName, err))
+			continue
+		}
+		if err := inst.Protect(); err != nil {
```
jandubois, line 33: I think we should check if the instance is already protected, and print a friendly message like `Instance %q is already protected` instead of this error:

```
$ l protect alpine
FATA[0000] failed to protect instance "alpine": open /Users/jan/.lima/alpine/protected: permission denied
```

jandubois (approved):
I found some more nits, but otherwise I think it is good.

I also checked how `limactl delete INSTANCE1 INSTANCE2` handles errors and saw that it ignores instances that don't exist, but otherwise aborts processing additional instances on any other error.

I prefer the way this is now implemented in this PR and think that the `delete` command should be updated to match (collect errors, and report them at the end). LMK if I should file an issue (or file one yourself).

cmd/limactl/protect.go
AkihiroSuda: Fixed

cmd/limactl/unprotect.go
AkihiroSuda: Fixed

cmd/limactl/protect.go
AkihiroSuda: Fixed

PR #1876, opened 2023-10-02
Make icons corresponding to links clickable

Victor-Chris:
Updated _index.html under main content to make the icons corresponding to the necessary links clickable.

### DIFF
--- a/website/content/en/_index.html
+++ b/website/content/en/_index.html
@@ -22,20 +22,8 @@ linkTitle = "Lima"
 </p>
 {{% /blocks/lead %}}
 {{< blocks/section color="dark" type="row" >}}
-{{% blocks/feature icon="fa-solid fa-code" title="Install Lima" %}}
-  <a class="mr-3 mb-4 text-secondary" href="{{< relref "/docs/installation" >}}">
-Get Lima Here <i class="fas fa-arrow-alt-circle-right ml-2"></i></a>
-{{% /blocks/feature %}}
 
-{{% blocks/feature icon="fa-solid fa-chalkboard" title="Try Lima" %}}
-<a class="mr-3 mb-4 text-secondary" href="{{< relref "/docs/usage" >}}">
-Get started with Lima <i class="fas fa-arrow-alt-circle-right ml-2"></i></a>
-{{% /blocks/feature %}}
-
-{{% blocks/feature icon="fa-brands fa-github" title="Contributions welcome" %}}
-<a class="mr-3 mb-4 text-secondary" href="{{< relref "/community" >}}">
-Join the Lima community <i class="fas fa-arrow-alt-circle-right ml-2"></i></a>
-{{% /blocks/feature %}}
+  {{< blocks/helpfullinks >}}
 
 {{< /blocks/section >}}
 
--- a/website/data/helpfullinks.yaml
+++ b/website/data/helpfullinks.yaml
@@ -1,0 +1,12 @@
+- header: Install Lima
+  icon: fa-solid fa-code
+  link: /docs/installation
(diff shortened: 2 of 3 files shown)


website/content/en/_index.html
```diff
@@ -22,20 +22,43 @@
...
-<a class="mr-3 mb-4 text-secondary" href="{{< relref "/docs/usage" >}}">
-Get started with Lima <i class="fas fa-arrow-alt-circle-right ml-2"></i></a>
-{{% /blocks/feature %}}
-
-{{% blocks/feature icon="fa-brands fa-github" title="Contributions welcome" %}}
-<a class="mr-3 mb-4 text-secondary" href="{{< relref "/community" >}}">
-Join the Lima community <i class="fas fa-arrow-alt-circle-right ml-2"></i></a>
-{{% /blocks/feature %}}
+
+  <div class="col-lg-4 mb-5 mb-lg-0 text-center">
```
balajiv113, line 26: This has lot of duplicates now. 
Can you try adding url attribute to `{{% blocks/feature `. See if that becomes clickable 

Source https://www.docsy.dev/docs/adding-content/shortcodes/#blocksfeature

balajiv113 (changes requested):

website/content/en/_index.html
Victor-Chris: I agree about the duplicates as well. I tried a number of options to do away with them, but with no success. I'm considering exploring the option of creating a custom block to minimise the duplication.

The option suggested with the link you've shared seems to just add a "Read more..." link to the **{{% blocks/feature** 
 without making the block/icons clickable.

Screenshot 2023-10-03 at 08 53 54

balajiv113: Ah okay. In that case, do create a custom shortcode to minimise the duplication

We already have some custom defined shortcode here 
https://github.com/lima-vm/lima/blob/master/website/layouts/shortcodes/blocks

Victor-Chris: Sure thing, I'd already taken a look through some of them. 
Let me try a custom shortcode and I'll revert afterwards.

Victor-Chris:
Have you had a chance to check out the latest commit @balajiv113 ?

balajiv113:
@Victor-Chris 
Kindly squash your commits and also fix the yamllint issue in the new yaml file

website/data/helpfullinks.yaml
```diff
@@ -0,0 +1,12 @@
+- link: /docs/installation
```
balajiv113, line 1: Naming conventions in this file could be better. Maybe something like
```
header: Install Lima
icon: fa-solid fa-code
link: /docs/installation
label/linkLabel/linkText: Get Lima Here
```

website/layouts/shortcodes/blocks/helpfullinks.html
```diff
@@ -0,0 +1,15 @@
+{{ $usedBy := site.Data.helpfullinks -}}
+
+{{ range $usedBy -}}
+  <div class="col-lg-4 mb-5 mb-lg-0 text-center">
+    <a class="text-white" href="{{ .link }}">
```
balajiv113, line 5:
```suggestion
    <a class="text-primary" href="{{ .link }}">
```

balajiv113: This way, when we want to change the primary color, these won't be affected

balajiv113 (changes requested):

balajiv113 (approved):

AkihiroSuda (approved):
Thanks

PR #1877, opened 2023-10-02
Don't download files that are already cached

### LINKED ISSUE #1875: Cut down on the amount of logging when nerdctl-full cache already exist
### Description

Currently we skip the logging of the "image" (basedisk), when starting up an instance.

But we still log the "nerdctl archive" (nerdctl-full) every time, even if the file already exists.

I think we should check the cache, and skip the log if the file is already cached for this version.

Especially since we are already caching the file on create, so "normally" it _will_ be in the cache...

----

limactl create

```
INFO[0000] Attempting to download the image              arch=x86_64 digest="sha256:1e35473cea5e1b827b91ad6ebb43b605a00d506c11f66c75076c424ae5372440" location="https://cloud-images.ubuntu.com/releases/23.04/release-20230926/ubuntu-23.04-server-cloudimg-amd64.img"
INFO[0000] Using cache "/home/anders/.cache/lima/download/by-url-sha256/c97220a456808578ed26e569621cdec0baae5996af78eb6ff7fd3668dd31526b/data" 
```

```
INFO[0000] Attempting to download the nerdctl archive    arch=x86_64 digest="sha256:2c5c43a8b77ed62090241027361baa62d8fb70a759bc9e7a82c637135598701f" location="https://github.com/containerd/nerdctl/releases/download/v1.6.0/nerdctl-full-1.6.0-linux-amd64.tar.gz"
INFO[0000] Using cache "/home/anders/.cache/lima/download/by-url-sha256/addb72512fd2fef72b08d8648173d7aaec3b4d5ae0566ac0bfd03c5adeeafc78/data" 
```
limactl start

```
INFO[0000] Attempting to download the nerdctl archive    arch=x86_64 digest="sha256:2c5c43a8b77ed62090241027361baa62d8fb70a759bc9e7a82c637135598701f" location="https://github.com/containerd/nerdctl/releases/download/v1.6.0/nerdctl-full-1.6.0-linux-amd64.tar.gz"
INFO[0000] Using cache "/home/anders/.cache/lima/download/by-url-sha256/addb72512fd2fef72b08d8648173d7aaec3b4d5ae0566ac0bfd03c5adeeafc78/data" 
```

It is probably overkill to make a copy (or even symlink) of it, like we do with the `basedisk`:

```go
        if _, err := os.Stat(baseDisk); errors.Is(err, os.ErrNotExist) {

                        if _, err := fileutils.DownloadFile(baseDisk, f.File, true, "the image", *cfg.LimaYAML.Arch); err != nil {
```

But if needed, we could write some kind of placeholder cache reference to the lima directory

```go
        if _, err := os.Stat(cacheFile); errors.Is(err, os.ErrNotExist) {

                path, err := fileutils.DownloadFile("", f, false, "the nerdctl archive", *y.Arch)
```

My idea was to just look and see if the **cache** file already exists, and if so skip the download.

This by adding a new function to fileutils, either return the cache path or just an "exists" boolean.

afbjorklund:
The nerdctl(-full) archive has already been downloaded to the cache (during `create`),
so don't log downloading it again (during `start`) - unless file is missing or digest fails.

Don't use files for different arch, or check for local files that are not available in the cache.
Refactor common methods out from the longer functions, so that they can be reused.

Closes #1875

Edit: it wasn't _actually_ downloading the file again, just spamming the INFO log about it

### DIFF
--- a/pkg/downloader/downloader.go
+++ b/pkg/downloader/downloader.go
@@ -97,8 +97,8 @@ func WithDecompress(decompress bool) Opt {
 // - The digest was not specified.
 // - The file already exists in the local target path.
 //
-// When the `data` file exists in the cache dir with `digest.<ALGO>` file,
-// the digest is verified by comparing the content of `digest.<ALGO>` with the expected
+// When the `data` file exists in the cache dir with `<ALGO>.digest` file,
+// the digest is verified by comparing the content of `<ALGO>.digest` with the expected
 // digest string. So, the actual digest of the `data` file is not computed.
 func WithExpectedDigest(expectedDigest digest.Digest) Opt {
 	return func(o *options) error {
@@ -183,15 +183,11 @@ func Download(local, remote string, opts ...Opt) (*Result, error) {
 		return res, nil
 	}
 
-	shad := filepath.Join(o.cacheDir, "download", "by-url-sha256", fmt.Sprintf("%x", sha256.Sum256([]byte(remote))))
+	shad := cacheDirectoryPath(o.cacheDir, remote)
 	shadData := filepath.Join(shad, "data")
-	shadDigest := ""
-	if o.expectedDigest != "" {
-		algo := o.expectedDigest.Algorithm().String()
-		if strings.Contains(algo, "/") || strings.Contains(algo, "\\") {
-			return nil, fmt.Errorf("invalid digest algorithm %q", algo)
-		}
-		shadDigest = filepath.Join(shad, algo+".digest")
+	shadDigest, err := cacheDigestPath(shad, o.expectedDigest)
+	if err != nil {
+		return nil, err
 	}
 	if _, err := os.Stat(shadData); err == nil {
 		logrus.Debugf("file %q is cached as %q", localPath, shadData)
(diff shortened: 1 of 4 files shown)


pkg/downloader/downloader.go
```diff
@@ -242,6 +242,38 @@ func Download(local, remote string, opts ...Opt) (*Result, error) {
 	return res, nil
 }
 
+func Cached(remote string, opts ...Opt) (string, error) {
```
AkihiroSuda, line 245: Could you add godoc comment to explain the arg string and the returned string

pkg/fileutils/download.go
```diff
@@ -41,6 +41,17 @@ func DownloadFile(dest string, f limayaml.File, decompress bool, description str
 	return res.CachePath, nil
 }
 
+// CachedFile returns path in cache.
```
AkihiroSuda, line 44: > path in cache

Is this relative? abosolute?

pkg/start/start.go
```diff
@@ -34,14 +34,21 @@ const DefaultWatchHostAgentEventsTimeout = 10 * time.Minute
...
 		return "", nil
 	}
 
 	errs := make([]error, len(y.Containerd.Archives))
 	for i, f := range y.Containerd.Archives {
+		// Skip downloading again if the file is already in the cache
+		if start && f.Arch == *y.Arch && !downloader.IsLocal(f.Location) {
+			path, err := fileutils.CachedFile(f)
+			if err == nil {
+				return path, nil
```
AkihiroSuda, line 49: "path" conflicts with the stdlib `path` package

AkihiroSuda:
Thanks, looks good but a couple of nits

pkg/fileutils/download.go
afbjorklund: Changed to use the same Result, with the same (absolute) CachePath as already being used in Download

AkihiroSuda (approved):
Thanks

PR #1826, opened 2023-09-21
CI: see if  wsl2 works

AkihiroSuda:
- #1781

PR #1883, opened 2023-10-04
ci: add WSL2 CI

### LINKED ISSUE #1781: CI: test WSL2
Follow-up to:
- https://github.com/lima-vm/lima/pull/1721

`runs-on: windows-2022-8-cores` ([larger runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners)) is known to support Hyper-V, but not sure if it supports WSL2.
Probably it should work: https://techcommunity.microsoft.com/t5/itops-talk-blog/wsl2-now-available-on-windows-server-2022/ba-p/3447570

pendo324:
Picking up where #1826 left off

According to [these docs](https://github.com/actions/runner-images/blob/win22/20230903.2/images/win/Windows2022-Readme.md), Windows runners should have Git for Windows pre-installed, meaning `tar.exe` and `gzip.exe` should be in the path already.

Let's see if just changing the rootfs from zstd to gz compression gets us any further, and iterate from there.

Closes #1781.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -73,9 +73,35 @@ jobs:
 
   windows:
     name: "Windows tests"
-    runs-on: windows-2022
+    runs-on: windows-2022-8-cores
     timeout-minutes: 30
     steps:
+    - name: Enable WSL2
+      run: |
+        wsl --set-default-version 2
+        # Manually install the latest kernel from MSI
+        Invoke-WebRequest -Uri "https://wslstorestorage.blob.core.windows.net/wslblob/wsl_update_x64.msi" -OutFile "wsl_update_x64.msi"
+        $pwd = (pwd).Path
+        Start-Process msiexec.exe -Wait -ArgumentList "/I $pwd\wsl_update_x64.msi /quiet"
+        wsl --update
+        wsl --status
+        wsl --list --online
+    - name: Install WSL2 distro
+      timeout-minutes: 3
+      run: |
+        # FIXME: At least one distro has to be installed here,
+        # otherwise `wsl --list --verbose` (called from Lima) fails:
+        # https://github.com/lima-vm/lima/pull/1826#issuecomment-1729993334
+        # The distro image itself is not consumed by Lima.
+        # ------------------------------------------------------------------
+        # Ubuntu-22.04:    gets stuck in some infinite loop during adduser
+        # OracleLinux_9_1: almostly silently fails, and just prints "Usage: adduser [options] LOGIN"
+        wsl --install -d openSUSE-Leap-15.5
+        wsl --list --verbose
+    - name: Set gitconfig
+      run: |
+        git config --global core.autocrlf false
+        git config --global core.eol lf
     - uses: actions/checkout@v4
       with:
         fetch-depth: 1
@@ -86,6 +112,19 @@ jobs:
       run: go test -v ./...
     - name: Make
       run: make
+    - name: Smoke test
+      # Make sure the path is set properly and then run limactl
+      run: |
+        $env:Path = 'C:\Program Files\Git\usr\bin;' + $env:Path
+        Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH -Value $env:Path
+        .\_output\bin\limactl.exe start template://experimental/wsl2
+      # TODO: run the full integration tests
+    - name: Debug
+      if: always()
+      run: type C:\Users\runneradmin\.lima\wsl2\ha.stdout.log
+    - name: Debug
+      if: always()
+      run: type C:\Users\runneradmin\.lima\wsl2\ha.stderr.log
 
   integration:
     name: Integration tests
--- a/examples/experimental/wsl2.yaml
+++ b/examples/experimental/wsl2.yaml
@@ -4,9 +4,9 @@ vmType: wsl2
 
 images:
 # Source: https://github.com/runfinch/finch-core/blob/main/Dockerfile
-- location: "https://deps.runfinch.com/common/x86-64/finch-rootfs-production-amd64-1690920103.tar.zst"
+- location: "https://deps.runfinch.com/common/x86-64/finch-rootfs-production-amd64-1694791577.tar.gz"
   arch: "x86_64"
-  digest: "sha256:53f2e329b8da0f6a25e025d1f6cc262ae228402ba615ad095739b2f0ec6babc9"
+  digest: "sha256:2d4d2e7386450899c6d0587fd0db21afadb31d974fa744aa9365c883935c5341"
 
 mountType: wsl2
 
--- a/pkg/store/instance_windows.go
+++ b/pkg/store/instance_windows.go
@@ -35,14 +35,33 @@ func inspectStatus(instDir string, inst *Instance, y *limayaml.LimaYAML) {
 	}
 }
 
-// GetWslStatus runs `wsl --list --verbose` and parses its output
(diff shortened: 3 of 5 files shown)


pendo324:
See a weird `Forbidden (403).` error, from `wsl --update` I think? Going to retry it in a few minutes to see if its a network issue

pendo324:
Well, I found out why the error wasn't being logged, fixed in https://github.com/lima-vm/lima/pull/1883/commits/88ce26871f22a3f651841d61ad96c2436a1cb05d.

The wsl command (or subcommand wslpath command) was returning an error that implemented `ExitCode()`, which caused the final [logrus.Fatal](https://github.com/lima-vm/lima/blob/master/cmd/limactl/main.go#L26-L27) to never get executed.

I also added some quotes to wrap the input of the temp file path when calling wslpath.

Now it seems like the output of wslpath is just... wrong for some reason, outputting:
```
time="2023-10-04T23:45:50Z" level=info msg="[hostagent] limaBootFilePathOnLinux: \"C:UsersRUNNER~1AppDataLocalTemplima-wsl2-boot-36860512[26](https://github.com/lima-vm/lima/actions/runs/6412733028/job/17410508491?pr=1883#step:8:27).sh\"\n"
```

From my local testing, double backslashes seem to work. Adding a change to convert \ to \\

balajiv113:
In WSL2, all folders are shared under C drive ??

I could see script is under `C:/Users/RUNNER~1/AppData/Local/Temp/` and userhome under `C:\\\\Users\\\\runneradmin\\\\.lima\\\\wsl2`

pendo324:
> In WSL2, all folders are shared under C drive ??
> 
> I could see script is under `C:/Users/RUNNER~1/AppData/Local/Temp/` and userhome under `C:\\\\Users\\\\runneradmin\\\\.lima\\\\wsl2`

Yeah, the entire C drive is shared by default. I'm pretty sure the files exist, but the wslpath call isn't returning the right output. I'm going to try to debug it on my local windows machine as to not waste resources. Just running the expected command (`wsl.exe -d <distro> wslpath -u "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\lima-wsl2-boot-2835403810.sh"`) does work and outputs a valid WSL path for me on my machine in a PowerShell terminal, so I'm not sure why its not working here.

balajiv113:
If this still doesn't work, we can compute the wslpath ourself (/mnt prefix and slashes changed accordingly)
There might already be a go library as well to do this.

I hope this should surely work without these glitches

jandubois:
I don't understand why this is necessary:

```go
bootFileWSLPath := strconv.Quote(limaBootFilePathOnWindows)
```

Isn't `exec.Command()` already taking care of the quoting? So you would now pass in a double quoted string?

> we can compute the wslpath ourself.

You can't really, in general, as even the mount prefix is configurable.

pendo324:
> I don't understand why this is necessary:
> 
> ```go
> bootFileWSLPath := strconv.Quote(limaBootFilePathOnWindows)
> ```
> 
> Isn't `exec.Command()` already taking care of the quoting? So you would now pass in a double quoted string?
> 

I used it to convert the slashes from \\ to \\\\ and add the quote around the string. It seems like `wslpath` expects \\\\ path separators, at least on certain versions of WSL.

From my testing, the command only works as intended when going through `bash -c`. I'd be lying if I said I understood exactly why this is the case 🤷‍♂️, but it seems to at least output a reasonable path now, so yay (if anyone has an explanation, I'm _very_ interested).

Seems like WSL still might have an issue with the `~` in the path from the looks of it. Seems like that's from an ["8.3" or "short" filename](https://en.wikipedia.org/wiki/8.3_filename), looking in to how to convert that to the "long" filename to see if that helps

jandubois:
> looking in to how to convert that to the "long" filename

I think the short names should just work, even from WSL.

You can convert them with PowerShell, but that has a high startup cost. But if you only do it once, then it doesn't matter:

```console
C:\>powershell "(Get-Item -LiteralPath 'C:\PROGRA~1').FullName"
C:\Program Files
```

jandubois:
> I'd be lying if I said I understood exactly why this is the case

The problem is that Windows doesn't have the concept of passing `*argv[]` to a new process; all you get is a single `char*`. So every app (or runtime) implements their own commandline parsing and argument quoting rules. It is all a big mess.

`command.exec()` has to smash the arguments together on its own, quoting them in a way that most programs will be able to parse again. But it is all based on heuristics.

pendo324:
We have liftoff :)

pendo324:
> > looking in to how to convert that to the "long" filename
> 
> I think the short names should just work, even from WSL.
> 
> You can convert them with PowerShell, but that has a high startup cost. But if you only do it once, then it doesn't matter:
> 
> ```
> C:\>powershell "(Get-Item -LiteralPath 'C:\PROGRA~1').FullName"
> C:\Program Files
> ```

Yeah you're right, the short path actually does work with `wslpath`. I added the code to convert to a regular "long path", using syscalls instead of PowerShell, but I'll probably remove it before merging unless there's a good reason to keep it

AkihiroSuda:
Seems green now 💚 

Please rebase, add back macOS CI, and ping me when this is ready to merge

pendo324:
@AkihiroSuda, I think it's good now. Let me know if you want me to make any more changes

.github/workflows/test.yml
```diff
@@ -73,9 +73,35 @@ jobs:
...
+        wsl --status
+        wsl --list --online
+    - name: Install WSL2 distro
+      timeout-minutes: 3
+      run: |
+        # FIXME: At least one distro has to be installed here,
+        # otherwise `wsl --list --verbose` (called from Lima) fails:
+        # https://github.com/lima-vm/lima/pull/1826#issuecomment-1729993334
+        # The distro image itself is not consumed by Lima.
+        # ------------------------------------------------------------------
```
AkihiroSuda, line 96: I’d like to see this resolved. Can be another PR though.

pkg/hostagent/hostagent.go
```diff
@@ -378,6 +380,7 @@ func (a *HostAgent) Run(ctx context.Context) error {
 		}()
 		return a.driver.RunGUI()
 	}
+	logrus.Infof("Running startRoutinesAndWait")
```
AkihiroSuda, line 383: Should be Debugf, or just removed

.github/workflows/test.yml
pendo324: I'll keep looking into this separately. Maybe there's something we can do with conditionally calling `wsl --import` / `wsl --install` based on the status returned from `wsl --list -v`

pkg/hostagent/hostagent.go
```diff
@@ -316,9 +316,11 @@ func (a *HostAgent) Run(ctx context.Context) error {
 		return err
 	}
 
+	logrus.Infof("After driver.Start()")
 	// WSL instance SSH address isn't known until after VM start
 	if *a.y.VMType == limayaml.WSL2 {
 		sshAddr, err := store.GetSSHAddress(a.instName)
+		logrus.Infof("sshAddress: %s", sshAddr)
```
AkihiroSuda, line 323: Debugf or remove

pkg/hostagent/hostagent.go
```diff
@@ -316,9 +316,11 @@ func (a *HostAgent) Run(ctx context.Context) error {
 		return err
 	}
 
+	logrus.Infof("After driver.Start()")
```
AkihiroSuda, line 319: Debugf or remove

AkihiroSuda (approved):
Thanks 💚

Issue #1872, opened 2023-10-01
RFC: should we make HOME writable in v1.0?


AkihiroSuda:
v1.0 will change the default mount driver from reverse-sshfs to virtiofs (for vz) and 9p (for qemu),
so mounting will be more stable:
- https://github.com/lima-vm/lima/issues/971

v1.0 is also a good milestone to have breaking changes, although we should try to avoid having breaking changes.

In v1.0, should we make the host home mount (`/Users/foo`) writable or not?

My current position is no, as I still want to minimize having breaking changes, and the mount can be easily configured to be writable with `limactl start --mount-writable` .
But I'm happy to hear opinions.

afbjorklund:
My thoughts that it was a security concern, that Docker Machine would export all of the home directory to the VM*. Both the privacy part, and the risk of doing damage to the host when doing something wrong in the VM or privileged containers...

\* Actually it only _worked_ in VirtualBox and not in qemu/kvm, but that is another story (missing the 9p support).

So I made it **not** mount the home directory by default in Podman Machine, even though it was available as an option if the user accepted the risks involved. The users complained, since it was "not Docker", and the decision was reverted* by Red Hat for [v4.1.0](https://github.com/containers/podman/releases/tag/v4.1.0).

\* It is now `$HOME:$HOME` (read-write): https://docs.podman.io/en/latest/markdown/podman-machine-init.1.html

---

For Lima, the privacy concerns are still there. But the read-only mount made it harder to do damage to the host...

There are also some performance implications, with encouraging users to work on a remote file system by default.

So I also think "no" (i.e. as-is/opt-in)

But also know it will confuse some.

jandubois:
I never considered Lima instances to be a "secure sandbox" to run untrusted code, so I never though of mounting directories read-only as a security concern. I assumed we did this to minimize the risk that we would accidentally overwrite some host data.

With Lima (just like Microsoft with WSL) we are trying to make the boundary between the host and the VM disappear as much as possible. When you run containers, we try to make it look like they run on the host and not in the VM. Not being able to write to the mounted directories feels like it contradicts his goal; it re-introduces friction when the purpose of the tool is to remove friction.

Personally I've always been using an `override.yaml` file to make the home directory (and my `/Volumes`) writable, for my personal convenience.

I know that many users expect this as well, which is why Rancher Desktop mounts all directories as writable as well.

[...]

PR #1884, opened 2023-10-05
Add MountWritable for setting writable default

afbjorklund:
Previously setting --mount-writable=false would turn off the writable for "/tmp/lima", which was undocumented.

Make it a global config, so that it is easy to change in the future (and easy to revert in local config, too)

Issue #1872

Note: this does not _change_ the default (it is still false), it just makes it _possible_ to change (in one place)


AkihiroSuda:
> Previously setting --mount-writable=false would turn off the writable for "/tmp/lima", which was undocumented.

Can we just document this?

afbjorklund:
It was documented in the flags help, and when used in docs it was always `true`

Issue #1892, opened 2023-10-07
RFC: Remove the color prompt overrides from default instance and shell environment for v1.0?


afbjorklund:
### Description

We had an old issue, to make the `lima` prompt distinguishable from the host prompt.

The color was playing on the "lime" word game, and used a bright yellow-green ([Lime](https://en.wikipedia.org/wiki/Lime_(color))) 

<details>
Here the color(s) shown in Ubuntu terminal:

![lima-shell-color-prompt](https://user-images.githubusercontent.com/10364051/142941057-6ef7fdae-32da-4242-a298-c857e2e62c11.png)
</details>

----

But it requires 24-bit color support to look best, as an addition to 256 (or 16) colors support:

* https://github.com/lima-vm/lima/pull/433

* https://github.com/lima-vm/lima/pull/435

* `COLORTERM=truecolor`

And it doesn't really _work_ on light backgrounds, so requires support for the terminal "theme":

* https://github.com/lima-vm/lima/issues/485

* https://github.com/lima-vm/lima/pull/486

* `TERMTHEME=Dark`

----

So I think the best would be to remove those changes, and go back to the original issue:

* https://github.com/lima-vm/lima/issues/427

Go back to the original Ubuntu prompt:

```diff
-if [ "$color_prompt" = yes ]; then
-    PS1='${debian_chroot:+($debian_chroot)}\[\033[01;32m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
+# Lima PS1: set color to lime
+if [ "$color_prompt" = true ]; then
+    PS1='${debian_chroot:+($debian_chroot)}\[\033[38;2;192;255;0m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
+elif [ "$color_prompt" = 256 ]; then
+    PS1='${debian_chroot:+($debian_chroot)}\[\033[38;5;154m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
+elif [ "$color_prompt" = yes ]; then
+    PS1='${debian_chroot:+($debian_chroot)}\[\033[01;92m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
 else
     PS1='${debian_chroot:+($debian_chroot)}\u@\h:\w\$ '
 fi
```

Note that the default prompt **does** have the host name in it, which does include "lima":

```console
[...]

PR #1893, opened 2023-10-07
Change prompt color from lime to lime green

### LINKED ISSUE #485: Don't set the prompt color when running in light terminal theme

afbjorklund:
This makes it more legible with light backgrounds, and changes the fallback color from "light green" back to regular "green".

The previous "lime" color had issues on light background, where "lime green" works on dark backgrounds and OK on light backgrounds.

Closes #485

Issue #1892


AkihiroSuda (approved):
Thanks

PR #1898, opened 2023-10-09
Fix DNS issue with alpine after upgrade

### LINKED ISSUE #1783: user-v2 regression in master: DNS wouldn't work with existing VMs
> @balajiv113: I found that after upgrading from lima to 0.17.2 to HEAD-de1b3ee, DNS wouldn't work with existing VMs. I had to delete and recreate the VM for DNS to work at all. Is that expected?

_Originally posted by @aaronlehmann in https://github.com/lima-vm/lima/issues/1333#issuecomment-1707404240_

balajiv113:
Fixes #1783

### DIFF
--- a/pkg/networks/usernet/gvproxy.go
+++ b/pkg/networks/usernet/gvproxy.go
@@ -37,6 +37,8 @@ var (
 	opts *GVisorNetstackOpts
 )
 
+const gatewayMacAddr = "5a:94:ef:e4:0c:dd"
+
 func StartGVisorNetstack(ctx context.Context, gVisorOpts *GVisorNetstackOpts) error {
 	opts = gVisorOpts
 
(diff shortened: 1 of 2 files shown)


pkg/networks/usernet/gvproxy.go
```diff
@@ -54,7 +54,7 @@ func StartGVisorNetstack(ctx context.Context, gVisorOpts *GVisorNetstackOpts) er
 			}
 		}
 	}
-	leases[gatewayIP] = "5a:94:ef:e4:0c:df"
+	leases[gatewayIP] = "5a:94:ef:e4:0c:dd"
```
AkihiroSuda, line 57: Could you explain this?

balajiv113: Its just a code smell change to match gateway mac addr to the one defined here
https://github.com/lima-vm/lima/blob/d722c35452e21b95fd51ce2092a615ab6c93bdd7/pkg/networks/usernet/gvproxy.go#L70

AkihiroSuda: Can we use a const?

AkihiroSuda (approved):
Thanks

PR #1899, opened 2023-10-09
Use tablet instead of mouse when using vnc display

afbjorklund:
There is the occasional offset between the software and hardware cursor when using mouse with a VNC display...

By using a "tablet" instead of a "mouse", this is avoided. It sends all positions as absolute instead of relative.


AkihiroSuda (approved):
Thanks

afbjorklund:
Support added to the Lima GUI here:

https://github.com/afbjorklund/lima-gui/commit/367b745e009d107043f4a5b0ed8f584c2490171f

PR #1909, opened 2023-10-11
Add ebpf lima example

iogbole:
This template installs the necessary packages you need to get started with writing and testing eBPF programs on Lima VMs.


AkihiroSuda:
Sorry, for template PRs basically we only accept base images and container engines

PR #1913, opened 2023-10-13
Support for inotify in mounted directories

### LINKED ISSUE #615: File system change events (inotify) support for mounted directories
### Description

From https://github.com/abiosoft/colima/issues/151:

We're using Colima to run Docker in a VM on our macOS laptops. Colima uses Lima under the hood to run this VM.

Are there any plans to support file system change events for mounted directories through inotify?

This is a key feature for us as we use containers in scenarios like auto-reloading dev servers and re-running unit tests on changes.

If there are no plans to support it at the moment, what kind of effort would be involved in adding this? Depending on the effort required, we might be open to putting together a PR to add support for this.

balajiv113:
Fixes #615 

### Tasks
- [x] Basic testing with vz (reverse-sshfs, virtiofs)
- [x] Basic testing with qemu macOS (reverse-sshfs, 9p)
- [ ] Basic testing with qemu linux (Need help!!)
- [ ] Basic testing with wsl windows (Need help!!, Need to see if this is needed/wsl mount already supports this)
- [x] Explore alternate for chmod

#### chtimes/touch (Currently used)
- No recursion issue / vim events are also triggered
- ~~But have issues with IDE (intellij keeps telling load file from system).~~ Fixed by passing through host time

#### chmod
- No recursion issue / vim events are also triggered
- IDE don't have any problem
- This deals with permissions of files in host (It would be better if we can find a alternate)

Do suggest if there are someother approach for this.


go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/lima-vm/lima
 
-go 1.20
+
+go 1.21
```
AkihiroSuda, line 4: Go >= 1.21 needs the `go 1.21.3` form.

.github/workflows/test.yml
```diff
@@ -260,7 +260,7 @@ jobs:
         fetch-depth: 1
     - uses: actions/setup-go@v4
       with:
-        go-version: 1.20.x
+        go-version: 1.21.x
```
AkihiroSuda, line 263: Is this necessary?

pkg/guestagent/api/server/server.go
```diff
@@ -76,8 +76,33 @@ func (b *Backend) GetEvents(w http.ResponseWriter, r *http.Request) {
...
+	w.Header().Set("Content-Type", "application/x-ndjson")
+	w.WriteHeader(http.StatusOK)
+	flusher.Flush()
+}
+
 func AddRoutes(r *mux.Router, b *Backend) {
 	v1 := r.PathPrefix("/v1").Subrouter()
 	v1.Path("/info").Methods("GET").HandlerFunc(b.GetInfo)
 	v1.Path("/events").Methods("GET").HandlerFunc(b.GetEvents)
+	v1.Path("/inotify").Methods("POST").HandlerFunc(b.ReceiveInotify)
```
AkihiroSuda, line 107:
```suggestion
	v1.Path("/inotify").Methods("POST").HandlerFunc(b.PostInotify)
```

pkg/guestagent/api/server/server.go
```diff
@@ -76,8 +76,33 @@ func (b *Backend) GetEvents(w http.ResponseWriter, r *http.Request) {
...
+	_, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	inotifyEvent := api.InotifyEvent{}
+	decoder := json.NewDecoder(r.Body)
+	if err := decoder.Decode(&inotifyEvent); err != nil {
+		logrus.Warn(err)
+		return
+	}
+	b.Agent.HandleInotify(inotifyEvent)
```
AkihiroSuda, line 91: This should be in a goroutine to avoid potential blocking?

pkg/hostagent/events/test.txt
AkihiroSuda, line 1: Remove?

pkg/guestagent/api/api.go
```diff
@@ -34,3 +35,8 @@ type Event struct {
 	LocalPortsRemoved []IPPort `json:"localPortsRemoved,omitempty"`
 	Errors            []string `json:"errors,omitempty"`
 }
+
+type InotifyEvent struct {
+	Location string      `json:"location,omitempty"`
```
AkihiroSuda, line 40:
```suggestion
	Path string      `json:"path,omitempty"` // Path in the guest
```

pkg/guestagent/guestagent_linux.go
```diff
@@ -333,3 +334,13 @@ func (a *agent) fixSystemTimeSkew() {
 		ticker.Stop()
 	}
 }
+
+func (a *agent) HandleInotify(event api.InotifyEvent) {
+	location := event.Location
+	if _, err := os.Stat(location); err == nil {
+		err := os.Chmod(location, event.Mode)
```
AkihiroSuda, line 341: Shouldn't we just use the mode statted in L340?

pkg/guestagent/api/api.go
```diff
@@ -34,3 +35,8 @@ type Event struct {
...
+	Mode     os.FileMode `json:"mode,omitempty"`
```
AkihiroSuda, line 41: Doesn't seem necessary

pkg/hostagent/hostagent.go
```diff
@@ -568,13 +568,19 @@ func (a *HostAgent) watchGuestAgentEvents(ctx context.Context) {
 		guestSocketAddr = fmt.Sprintf("0.0.0.0:%d", a.vSockPort)
 	}
 
+	go func() {
+		err := a.startInotify(ctx, guestSocketAddr, localUnix, remoteUnix)
```
AkihiroSuda, line 572: This should be enabled only with `limactl start --mount-inotify`, and it should be marked experimental.

pkg/guestagent/guestagent_linux.go
balajiv113: I tried this, and at some point it actually corrupted my host system with file permissions as --------

AkihiroSuda: chmod looks dangerous anyway, let's just use `touch` and call it a day, even if it doesn't work with IntelliJ.

AkihiroSuda: For a long-term solution we have to submit a kernel patch to allow injecting inotify events from userspace.
Likely to take multiple years though.

balajiv113: > chmod looks dangerous anyway

Exactly. That's why looking for possible alternates

> even if it doesn't work with IntelliJ

My fear is this will just be one-of usecase, this may even block the remote development possibilities from lima

pkg/guestagent/api/server/server.go
balajiv113: Done

pkg/hostagent/events/test.txt
balajiv113: Done

pkg/guestagent/api/api.go
balajiv113: Removed

pkg/guestagent/guestagent_linux.go
balajiv113: Changed to touch and also fixed the intellij issue. 
Previously events were getting dropped but now it look stable in macOS

.github/workflows/test.yml
balajiv113: Was using clear() which is present from 1.21. I could replace it with Maps.clear as well if we need to keep 1.20

aeifn:
One drawback I noticed: sometimes — say twice an hour — VM (`vz`,`virtfs`) crashes silently while intensive `haskell` build inside VM in mounted directory.

balajiv113:
@aeifn Yes yes i am working on optimising it and address these crashes on high load

pkg/hostagent/hostagent.go
balajiv113: Should we expose this via yaml or just via cmd ?

balajiv113:
@aeifn - Can you give a try now. 
Hopefully the silent crash should not happen now

.github/workflows/test.yml
balajiv113: Changed code to support 1.20

go.mod
balajiv113: Using 1.20

pkg/hostagent/hostagent.go
AkihiroSuda: Yaml, just like other mount options

balajiv113: Done

balajiv113:
@AkihiroSuda 
Marking this as ready for review. As all the required features are added.

I have tested this feature with reverse-sshfs, 9p, virtiofs on macOS. All these cases are working fine without issues. 
Also done long running and average load test, as of now i don't see any crashes.

But might some help to test on linux and windows.

.github/workflows/test.yml
AkihiroSuda: > Changed code to support 1.20

Could you revert the changes to GHA too?

examples/default.yaml
```diff
@@ -103,6 +103,11 @@ mounts:
 # 🟢 Builtin default: "reverse-sshfs" (for QEMU), "virtiofs" (for vz)
 mountType: null
 
+# Enable inotify support for mounted directories (EXPERIMENTAL)
+# 🟢 Builtin default: Disabled by default
+mountInotify: null
```
AkihiroSuda, line 108: Wondering if this should be configurable per `[]mounts` entry

website/content/en/docs/Config/Mount/_index.md
```diff
@@ -165,3 +165,32 @@ mountType: "wsl2"
 #### Caveats
 - WSL2 file permissions may not work exactly as expected when accessing files that are natively on the Windows disk ([more info](https://github.com/MicrosoftDocs/WSL/blob/mattw-wsl2-explainer/WSL/file-permissions.md))
 - WSL2's disk sharing system uses a 9P protocol server, making the performance similar to [Lima's 9p](#9p) mode ([more info](https://github.com/MicrosoftDocs/WSL/blob/mattw-wsl2-explainer/WSL/wsl2-architecture.md#wsl-2-architectural-flow))
+
+## Mount Inotify
+> **Warning**
+> "mountInotify" is experimental
```
AkihiroSuda, line 171: Please update https://lima-vm.io/docs/releases/experimental/ too

examples/default.yaml
jandubois: I think that would be good, but could also be added later.

Ideally we should have the global option, and then a way to override it at the individual mount level.

website/content/en/docs/Config/Mount/_index.md
balajiv113: Done

examples/default.yaml
balajiv113: >  but could also be added later

Exactly my thought as well. Based on need we could add it later as well

balajiv113:
Just a FYI,
In terms of CPU. With inotify enabled am seeing some slight increase in CPU during normal usage.

Am testing out GRPC for communication. I will create a follow-up PR to address this issue. I will keep performance out of scope for this PR.

.github/workflows/test.yml
balajiv113: Done

website/content/en/docs/Config/Mount/_index.md
```diff
@@ -165,3 +165,32 @@ mountType: "wsl2"
...
+
+| ⚡ Requirement | Lima >= 1.0+ |
```
AkihiroSuda, line 173:
```suggestion
| ⚡ Requirement | Lima >= 0.19 |
```

AkihiroSuda:
Thanks, LGTM but a single nit above

website/content/en/docs/Config/Mount/_index.md
```diff
@@ -165,3 +165,32 @@ mountType: "wsl2"
...
+{{< tabpane text=true >}}
+{{% tab header="CLI" %}}
+```bash
+limactl start --mount-inotify
+```
+{{% /tab %}}
+{{% tab header="YAML" %}}
+```yaml
+vmType: "vz"
+mountType: "virtiofs"
```
AkihiroSuda, line 192: Is virtiofs necessary?

balajiv113: Its not necessary, Since vz was used i just set virtiofs

website/content/en/docs/Config/Mount/_index.md
balajiv113: Done

website/content/en/docs/Config/Mount/_index.md
AkihiroSuda: Is vz necessary here?

website/content/en/docs/Config/Mount/_index.md
```diff
@@ -165,3 +165,32 @@ mountType: "wsl2"
...
+> "mountInotify" is experimental
+
+| ⚡ Requirement | Lima >= 0.19 |
+| ----------------- |--------------------------------------------------------|
+
+The `mountInotify` support enables inotify support for all different mountTypes like 9p, virtiofs etc.
+
+When mountInotify is enabled, hostagent will listen and send inotify events from host machine to guest. 
+
+This support will be enabled only for writable mounts
```
jandubois, line 180: I was briefly wondering why we don't set this up for read-only mounts, as the notification direction is from the host to the guest.

But then I realized that this is not a choice, but a limitation, because we need to modify the mounted directories/files inside the guest to re-create the inotify events, so we need write access for that.

So I think it would be useful to explain this here, so people don't start asking to support it on read-only mounts as well.

balajiv113: Good catch. Will update it as well

balajiv113:
On doing final testing i noticed QEMU and 9p not working.

Will try to debug and fix it. 

**Issue**
Inotify events not triggered for 9p mount point within linux vm itself. Because of that host events are not triggering change

AkihiroSuda:
I guess the support for qemu can be deferred

AkihiroSuda:
Needs another rebase

website/content/en/docs/Config/Mount/_index.md
```diff
@@ -165,3 +165,38 @@ mountType: "wsl2"
...
+- hostagent will listen and send inotify events from host machine to guest. 
+- Guest will modify the file to trigger inotify on guest side 
+
+This support will be enabled only for writable mounts because only for writable mount guest will be able to trigger inotify
+
+An example configuration:
+{{< tabpane text=true >}}
+{{% tab header="CLI" %}}
+```bash
+limactl start --mount-inotify
```
AkihiroSuda, line 188: If vz is necessary, this should have `--vm-type=vz`

balajiv113:
Rebased. Updated the document with more info and known issues as well.

To address the issues will create a PR to use GRPC for communication instead of REST. If we have time we can even target this for 0.19. So the feature will be complete

AkihiroSuda:
> To address the issues will create a PR to use GRPC for communication instead of REST. If we have time we can even target this for 0.19. So the feature will be complete


Thanks, but I'd prefer to release the vz fix ASAP:
- https://github.com/lima-vm/lima/pull/2026

Maybe we should just release v0.19 from the current main branch (with bumping up images, etc.), and then release v0.20 in a couple of weeks with the gRPC and inotify?

balajiv113:
Yes i would prefer that as well we can move inotify to 0.20 so it will be a complete feature.

AkihiroSuda:
What's the current status?

AkihiroSuda (dismissed):
Thanks

AkihiroSuda:
Needs rebase

afbjorklund:
@balajiv113 : tested with linux (qemu), and it seems to be working for basic usage

```console
anders@lima-default:/home/anders/lima$ inotifywait -m /tmp/lima
Setting up watches.
Watches established.
/tmp/lima/ ATTRIB foo
/tmp/lima/ ATTRIB bar
```

website/content/en/docs/Config/Mount/_index.md
afbjorklund: I missed this detail, as well.

Mostly because it was not documented in lima.yaml, but only in the documentation (and who reads that).

AkihiroSuda (approved):
Thanks

website/content/en/docs/Config/Mount/_index.md
jandubois: It is my understanding that neither `vmType` nor `mountType` are required; the PR descriptions says it works with QEMU as well, and for all mount types. In that case these 2 lines should be removed for clarity.

balajiv113: Will remove it. I Just added it as a example reference of where to use it

jandubois:
I tried to test it on macOS 12.7.3 (Monterey) on Intel, and it didn't work for me at all, both using Alpine and Ubuntu (default), with either reverse-sshfs or 9p:

```console
jan@lima-default:/Users/jan/suse/lima$ sudo -i
root@lima-default:~# inotifywait -m /tmp/lima
Setting up watches.
Watches established.
```

I then run `touch /tmp/lima/foo` on the host, and the file is visible inside the guest, but `inotifywait` doesn't report anything.

`limactl start` confirms that `mountInotify` has been set:

```console
$ l edit
Waiting for Emacs...
WARN[0021] `mountType: 9p` is experimental
WARN[0021] `mountInotify` is experimental
INFO[0021] Instance "default" configuration edited
? Do you want to start the instance now?  Yes
…
```

jandubois:
> I tried to test it on macOS 12.7.3 (Monterey) on Intel,

Same thing happens for me on macOS 14.3.1 (Sonoma) on M1: I don't get any inotify events (inside the VM) for files touched on the host, but I do get events for files touched inside the VM.

Do I misunderstand how this is supposed to work? I expected that touch a file on the host would cause the guestagent to touch the file inside the VM, triggering an inotify event that could be captured by applications supporting hot-reload.

I guess I can give it one more try using VZ, but I suspect I must be doing something else wrong.

jandubois:
> I guess I can give it one more try using VZ, but I suspect I must be doing something else wrong.

Same results with VZ.

Only interesting observation was seeing these 2 lines during startup:

```
ERRO[0055] [hostagent] r.CreateEndpoint() = connection was refused
ERRO[0056] [hostagent] r.CreateEndpoint() = connection was refused
```

I don't think I've seen them before; could they be related?

balajiv113:
@jandubois Do you see any logs like 
`enable inotify for writable mount <path>` in ha logs ??

jandubois:
> Do you see any logs like `enable inotify for writable mount <path>` in ha logs ??

@balajiv113 Yes, I do:

```console
$ ag inoti
lima.yaml
97:# Enable inotify support for mounted directories (EXPERIMENTAL)
99:mountInotify: true

ha.stderr.log
36:{"level":"info","msg":"enable inotify for writable mount: /tmp/lima","time":"2024-02-29T17:30:42-08:00"}
43:{"level":"info","msg":"enable inotify for writable mount: /Volumes/ThunderBlade","time":"2024-02-29T17:30:42-08:00"}
```

balajiv113:
Found the issue, /tmp/lima is a symlink to /private/tmp/lima. 
Due to this it is not able to find a similar file in guest

Thanks for the find, Will fix it. I was trying with home folder didn't encounter this :(

balajiv113:
Fixed the symlink issue.

@jandubois - Can you try actually editing the file and see if that triggers a event. Instead of just touch.

The problem is we are relying on `notify.Create|notify.Write` events. These are internally `FSEventsCreated|FSEventsModified`. 

When we do a touch, it triggers `FSEventsInodeMetaMod` and not `FSEventsModified`. If we rely on `FSEventsInodeMetaMod` we will be getting too many events (As we are doing touch in guest, which triggers a cycle of events).

For now, I believe `FSEventsModified` should be enough to triggers events on file change.

Note: Am not adding it to document, Because at some usecase, even touch will work. For instance, 
edit a file and save -> notice a event is triggered -> now do touch file -> notice a event is triggered

AkihiroSuda:
@jandubois PTAL 🙏

jandubois:
@balajiv113 I can confirm that the symlink issue seems to be fixed.

I do get different results for touch vs. edit though.

E.g. when I `touch foo` on the host, I get `/tmp/lima/ ATTRIB foo` from `inotifywait`.

But when I then write to the file with e.g. `cp ~/foo /tmp/lima/foo`, I don't get any event even though the file content changed. This looks wrong.

This only happens for empty files. If the file contains data, and it is overwritten, then I do get a notification.

Also it looks like `ATTRIB` is the only notification I get. When I modify files in the same shared directory from a second shell from inside the VM, then I get the usual ACCESS, OPEN, CREATE, MODIFY, CLOSE notifications as well. So I wonder if ATTRIB will be enough to trigger hot-reload with the various applications using inotify.

I'm sorry, I only got to try the PR, I haven't yet found a big enough chunk of time to review the code in one sitting... 😞

balajiv113:
> Also it looks like ATTRIB is the only notification I get.

This is intended. The reason is when we need to replicate the same pattern we will have complexity of handling the duplicate event in host. All we do is just touch the file in guest agent

> So I wonder if ATTRIB will be enough to trigger hot-reload with the various applications using inotify

I checked with some of the frontend hot reload like nextjs and it worked fine with attrib

> But when I then write to the file with e.g. cp ~/foo /tmp/lima/foo, I don't get any event

I will check this case

balajiv113:
@jandubois 
I have done changes to support `FSEventsInodeMetaMod`. All use-cases should work fine. Even touch with empty file should work now

Note: As said above, We will always receive ATTRIB notification only.

AkihiroSuda:
@jandubois PTAL

jandubois:
> Even touch with empty file should work now

I can confirm that it works now.

I wonder if removing a file should send a notification or not. Because for me it doesn't. And if I remove a file from a webapp, shouldn't it trigger a hot reload?

If this is hard/complex, then maybe documenting this as a limitation would be fine too.

balajiv113:
> I wonder if removing a file should send a notification or not

Our current model is to touch the file so that the guest consider something changed. 
With this approach we cannot support remove file logic. As file is removed and we cannot trigger a touch

>  maybe documenting this as a limitation would be fine too

Will document as a limitation

AkihiroSuda (approved):
Thanks, let's merge this

ishan-simplismart:
On OSX 15.5, and ubuntu 24.04 vm this only works when location and mountPoint are the same, when they are different for example, it doesn't work anymore.
Tested it is working when they are same
```
  - location: "/tmp/lima"
    mountPoint: "/tmp/host"
 ```

PR #1918, opened 2023-10-16
fix: Ctrl-C does not interrupt create prompt

### LINKED ISSUE #1887: Hitting <Ctrl>+C on the `limactl start` interactive prompt does not cancel the operation
### Steps to reproduce:

1. Create a new lima VM with `limactl start --name=nameofVMthatdoesntexistyet`
2. Wait for prompt to appear and then press `<Ctrl>+C`

### Expected result:

- The user is returned back to the shell prompt and the lima VM creation process does NOT initiate

### Actual result:

- The lima VM creation process continues as if `<Enter>` was pressed, accepting the `Proceed with current configuration` prompt

Details:
- OS: Archlinux
- lima version: 0.17.2

cwegener:
Fixes #1887

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/containerd/containerd/identifiers"
 	"github.com/lima-vm/lima/cmd/limactl/editflags"
 	"github.com/lima-vm/lima/cmd/limactl/guessarg"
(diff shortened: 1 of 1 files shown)


cmd/limactl/start.go
```diff
@@ -376,6 +377,9 @@ func chooseNextCreatorState(st *creatorState, yq string) (*creatorState, error)
 			},
 		}
 		if err := survey.AskOne(prompt, &ans); err != nil {
+			if err == terminal.InterruptErr {
+				logrus.Fatal("interrupted")
```
AkihiroSuda, line 381: Maybe
```suggestion
				logrus.Fatal("terminal was interrupted")
```

cwegener: I think that the "more correct" message probably would be:

`Interrupted by user`

WDYT?

AkihiroSuda: SGTM

AkihiroSuda (approved):
Thanks

Issue #1922, opened 2023-10-17
Loading and building images in Lima Kubernetes cluster


afbjorklund:
### Description

Podman Desktop is using Lima as an alternative for running local Kubernetes.

It works fine for this, but is missing two features from kind and from minikube:

`kind load docker-image`
`kind load image-archive`

```
Available Commands:
  docker-image  Loads docker images from host into nodes
  image-archive Loads docker image from archive into nodes
```

* https://podman-desktop.io/docs/kubernetes/kind

`minikube image load`
`minikube image build`

```
Available Commands:
  build         Build a container image in minikube
  load          Load an image into minikube
```

* https://podman-desktop.io/docs/kubernetes/minikube

It is possible to do this also with `nerdctl`, but it needs some help with copy.

The image archive and the build context needs transferring, to the lima node...

* https://podman-desktop.io/docs/kubernetes/lima

----

I was thinking of a wrapper script, that helps with these functions, and others too:

LIMA_INSTANCE=k8s lima sudo CONTAINERD_NAMESPACE=k8s.io nerdctl "$@"

`kubenerdctl load busybox:latest`

`kubenerdctl build -t myimage .` 

It will create a tar archive from the local image or the build context, and `copy` it.

Then the load will run from the temporary tar, or the build from the unpacked tar.

afbjorklund:
Proof of concept...

Containerfile:

```Dockerfile
FROM busybox
RUN true
```

```console
$ ../cmd/nördctl.lima load busybox:latest
tmp.FHIgIdyxU8.tar                                                                    100% 1434KB 328.0MB/s   00:00    
unpacking docker.io/library/busybox:latest (sha256:44bb20ae7cef787c15638fa2c0de3f67f1674c35463f32ec77eaae002c164d7f)...
Loaded image: busybox:latest
$ ../cmd/nördctl.lima build . -t myimage
tar: ./tmp.LiVFmPhYKM.tar: filen är själva arkivet; inte arkiverad
tmp.LiVFmPhYKM.tar                                                                    100%   10KB  11.8MB/s   00:00    
[+] Building 5.1s (6/6)                                                                                                 
 => [internal] load build definition from Containerfile                                                            0.0s
[+] Building 5.2s (6/6) FINISHED                                                                                        
 => [internal] load build definition from Containerfile                                                            0.0s
 => => transferring dockerfile: 62B                                                                                0.0s
 => [internal] load metadata for docker.io/library/busybox:latest                                                  5.0s
[...]

PR #1923, opened 2023-10-17
Proof of concept script for kubernetes images

afbjorklund:
Issue #1922

### DIFF
--- a/cmd/kubenerdctl.lima
+++ b/cmd/kubenerdctl.lima
@@ -1,0 +1,44 @@
+#!/bin/sh
+set -eu
+: "${LIMA_INSTANCE:=k8s}"
+: "${CONTAINERD_NAMESPACE:=k8s.io}"
+: "${DOCKER:=docker}"
+: "${TAR:=tar}"
+
+case "$1" in
+	load)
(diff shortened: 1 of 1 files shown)


jandubois:
> In case you are wondering, [nörd](https://sv.wikipedia.org/wiki/N%C3%B6rd) is the Swedish spelling. 🇸🇪

Cute, but a bit awkward to type even with tab completion (unless you use a scandinavian/german keyboard layout), e.g. on English keyboard: `n` `⌥u` `o`, `⇥`. Using tab completion on `n` alone gives too many alternatives...

cmd/nördctl.lima
AkihiroSuda, line 1: Probably this should be under the `contrib/`dir or a third party repo

afbjorklund:
I did not find a good name at the time, so it was a work in progress

cmd/nördctl.lima
afbjorklund: That probably just means that it will not be available to Podman Desktop

afbjorklund: You mean something like `kulima`, right?

PR #1931, opened 2023-10-20
feat: propagate host timezone to guest

### LINKED ISSUE #1921: Propagate timezone from the host by default (unless `--plain` is specified)

lobshunter:
This PR closes #1921, propagates host timezone to guest VM.

This doesn't work for template://alpine for it doesn't have pre-installed tzdata. And this PR doesn't use timedatectl because some distros might not be using systemd.

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
 	"time"
@@ -136,6 +137,7 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 		VMType:         *y.VMType,
 		VSockPort:      vsockPort,
 		Plain:          *y.Plain,
+		Timezone:       getHostTimezone(),
 	}
 
 	firstUsernetIndex := limayaml.FirstUsernetIndex(y)
@@ -368,6 +370,54 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 	}
 
 	return iso9660util.Write(filepath.Join(instDir, filenames.CIDataISO), "cidata", layout)
+}
+
+// getHostTimezone tries to get host timezone from env or specific files
+func getHostTimezone() string {
+	if runtime.GOOS == "windows" {
+		logrus.Debug("skip get host timezone on Windows")
+		return ""
+	}
+
+	// 1. try to get timezone from env.TZ
+	tzEnv := os.Getenv("TZ")
+	if tzEnv != "" &&
+		// skip if TZ is in "std offset dst offset, rule" format
+		!strings.Contains(tzEnv, " ") {
+		return tzEnv
+	}
+	logrus.Debugf("cannot get timezone from env TZ: [%s]", tzEnv)
+
+	// 2. try to get timezone from /etc/timezone
+	tzName, err := os.ReadFile("/etc/timezone")
+	if err == nil {
+		return string(tzName)
+	}
+	logrus.Debugf("cannot get timezone from /etc/timezone: %s", err)
+
+	// 3. try to get timezone from /etc/localtime soft link
+	// base on the assumption that <zoneinfo_basedir>/zone.tab file exists
+	// and /etc/localtime is a soft link to <zoneinfo_basedir>/<timezone>
+	tzFilePath, err := filepath.EvalSymlinks("/etc/localtime")
+	if err == nil {
+		for baseDir := filepath.Dir(tzFilePath); baseDir != "/"; baseDir = filepath.Dir(baseDir) {
+			anchorFile := filepath.Join(baseDir, "zone.tab")
+			_, err = os.Stat(anchorFile)
+			if err != nil {
+				continue
+			}
+
+			tz := strings.TrimPrefix(tzFilePath, baseDir+"/")
+			_, err = time.LoadLocation(tz)
+			if err == nil {
+				return tz
+			}
+		}
+	}
+	logrus.Debugf("cannot get host timezone from /etc/localtime: %s", err)
+
+	logrus.Warnf("cannot get host timezone")
+	return ""
 }
 
 func GuestAgentBinary(ostype limayaml.OS, arch limayaml.Arch) (io.ReadCloser, error) {
--- a/pkg/cidata/cidata.TEMPLATE.d/user-data
+++ b/pkg/cidata/cidata.TEMPLATE.d/user-data
@@ -1,6 +1,10 @@
 #cloud-config
 # vim:syntax=yaml
 
+{{- if .Timezone }}
+timezone: {{.Timezone}}
+{{- end }}
+
 growpart:
   mode: auto
   devices: ['/']
--- a/pkg/cidata/template.go
+++ b/pkg/cidata/template.go
@@ -82,6 +82,7 @@ type TemplateArgs struct {
 	VMType                          string
 	VSockPort                       int
 	Plain                           bool
+	Timezone                        string
 }
 
 func ValidateTemplateArgs(args TemplateArgs) error {


balajiv113:
@lobshunter Thanks for the PR.

Can you try out setting timezone via cloud-init (via pkg/cidata/)
https://cloudinit.readthedocs.io/en/latest/reference/modules.html#timezone

This way it will set for all distro supporting cloud-init. Also will be set early in boot

jandubois:
> This doesn't work for template://alpine for it doesn't have pre-installed tzdata.

We can add `tzdata` to the image; it is just 3MB or so.

If you change the timezone via `cloud-init`, then we also have to update `lima-init` to support it, as we currently don't use `cloud-init` in our Alpine image.

lobshunter:
> Can you try out setting timezone via cloud-init (via pkg/cidata/)

Hi, I updated the cloud-init version in the second commit. PTAL which one do you prefer.

> We can add tzdata to the image; it is just 3MB or so.

It's `3068 KiB` for alpine:3.18.3. Should I update it [here](https://github.com/lima-vm/alpine-lima/blob/main/Dockerfile#L7)?

> as we currently don't use cloud-init in our Alpine image

I am confused by this actually. Alpine doesn't have cloud-init installed, but it did run the new added `60-propagate-timezone.sh` script.

pkg/cidata/cidata.TEMPLATE.d/boot/60-propagate-timezone.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+set -eux
+
+tz=$LIMA_TIMEZONE
```
jandubois, line 4: Not sure why you need to create an alias for the variable; why not use the full name below?

pkg/cidata/cidata.TEMPLATE.d/boot/60-propagate-timezone.sh
```diff
@@ -0,0 +1,10 @@
...
+
+if [ -n "$tz" ]; then
+    test -f /usr/share/zoneinfo/$tz || exit 0 # incase tzdate is not installed, shouldn't block startup
```
jandubois, line 7: I don't like the `|| exit 0` part hidden in the middle of the line. It should either be a proper `if` block, or just fold it into the condition of the outer block:

```suggestion
if [ -n "$tz" ] && [ -f "/usr/share/zoneinfo/$tz" ]; then
```

pkg/cidata/cidata.TEMPLATE.d/lima.env
```diff
@@ -46,3 +46,4 @@ LIMA_CIDATA_PLAIN=1
 {{- else}}
 LIMA_CIDATA_PLAIN=
 {{- end}}
+LIMA_TIMEZONE={{ .Timezone }}
```
jandubois, line 49: For consistency with the other names it should use the `LIMA_CIDATA_` prefix:

```suggestion
LIMA_CIDATA_TIMEZONE={{ .Timezone }}
```

pkg/cidata/cidata.go
```diff
@@ -413,6 +414,24 @@ func getBootCmds(p []limayaml.Provision) []BootCmds {
 	return bootCmds
 }
 
+// getTimezone returns the timezone of host, to be used in the VM.
+func getTimezone() string {
+	_, GMTOffset := time.Now().Zone()
+	if GMTOffset%3600 != 0 {
+		logrus.Warnf("Host timezone offset is not a multiple of 3600 seconds: %d, skip timezone propagation", GMTOffset)
```
jandubois, line 421: The warning is not correct; you are just rounding the timezone offset to full hours, but it is still being applied.

pkg/cidata/cidata.go
```diff
@@ -413,6 +414,24 @@ func getBootCmds(p []limayaml.Provision) []BootCmds {
...
+	}
+
+	GMTOffsetHours := GMTOffset / 3600
+	var timezone string
+	if GMTOffsetHours > 0 {
+		timezone = fmt.Sprintf("Etc/GMT-%d", GMTOffsetHours)
+	} else {
+		timezone = fmt.Sprintf("Etc/GMT+%d", GMTOffsetHours)
```
jandubois, line 429: Don't you have to fix the sign of the offset?

```suggestion
		timezone = fmt.Sprintf("Etc/GMT+%d", -GMTOffsetHours)
```

jandubois (changes requested):
I'm not happy that this PR cannot support timezones with fractional-hours offsets from GMT. E.g. India is GMT+5:30, and there are plenty more.

I wonder if there isn't a way to get an exact match to the host timezone?

jandubois:
Please also address linter errors!

jandubois:
> I wonder if there isn't a way to get an exact match to the host timezone?

macOS sets `/etc/localtime` to a symlink too, so we could scrape it from there:

```shell
$ readlink /etc/localtime | sed 's#.*zoneinfo/##'
America/Vancouver
```

pkg/cidata/cidata.go
lobshunter: It's not a bug, perhaps I should have put a comment to it, but the sign is reversed [by design](https://stackoverflow.com/a/53089164/11004152)... I did test it locally with Alpine & Ubuntu VM BTW...

pkg/cidata/cidata.TEMPLATE.d/boot/60-propagate-timezone.sh
lobshunter: Updated in [97c113c](https://github.com/lima-vm/lima/pull/1931/commits/97c113cde70a6db042accad70143046000a44da2), mainly because I forgot polish some debugging code.

jandubois:
> Hi, I updated the cloud-init version in the second commit.

I think this is not what @balajiv113 meant. `cloud-init` has support for setting timezone built-in. You just have to add a line to `user-data`, e.g.

```yaml
timezone: US/Eastern
```

You would not need the `60-propagate-timezone.sh` script because cloud-init would already do the work for you.

Please ignore my earlier review and use the proper `cloud-init` mechanism instead.

> Alpine doesn't have cloud-init installed, but it did run the new added 60-propagate-timezone.sh script.

It uses `lima-init`, which implements just a subset of `cloud-init` using shell and `awk`. It doesn't have support for the `timezone` key in `user-data`, so that would have to be added. It does support running the boot script, which in turn runs the provisioning scripts.

pkg/cidata/cidata.TEMPLATE.d/boot/60-propagate-timezone.sh
lobshunter: Took suggestion in [97c113c](https://github.com/lima-vm/lima/pull/1931/commits/97c113cde70a6db042accad70143046000a44da2).

pkg/cidata/cidata.TEMPLATE.d/lima.env
lobshunter: Updated in [97c113c](https://github.com/lima-vm/lima/pull/1931/commits/97c113cde70a6db042accad70143046000a44da2), didn't notice the `_CIDATA` part.

pkg/cidata/cidata.go
lobshunter: Log updated in [97c113c](https://github.com/lima-vm/lima/pull/1931/commits/97c113cde70a6db042accad70143046000a44da2).

pkg/cidata/cidata.go
jandubois: I suspect you did not test it with a negative offset. You would end up with something like `Etc/GMT+-5`, which does not exist. I assume you would want `Etc/GMT+5`.

But I still think you should use the timezone name from the host instead of cooking up a name based on the GMT offset.

lobshunter:
> use the proper cloud-init mechanism instead.

Yes, will do.

> I'm not happy that this PR cannot support timezones with fractional-hours offsets from GMT

Me neither, but I couldn't find a proper way after digging for half an hour. This is the simplest one I got.

> I wonder if there isn't a way to get an exact match to the host timezone?

I did try, but the problem is Lima currently supports multiple host platforms (Linux/MacOS/Windows(that's WSL2 IIUC)). Especially for Windows, I didn't find a portable way to get the host timezone. Using Golang's built-in function would be the most portable one, but it doesn't provide the `get timezone in geographic location format`.

I noticed location timezone can be retrieved in `/etc/localtime` on MacOS and `/etc/timezone` on Linux(Ubuntu & Manjaro).  But didn't have a Windows setup for testing. Also using platform-specific approach reduces the testability of code. I'll do more study on this.

pkg/cidata/cidata.go
lobshunter: > I suspect you did not test it with a negative offset

That's true. I missed that part. Will try to find the host location timezone.

jandubois:
> Especially for Windows, I didn't find a portable way to get the host timezone.

Yeah, it is almost impossible to map Windows time zones to IANA (tzdata) time zone names. There is a Python app that seems to implement it.

However, I think WSL2 already sets this up by itself, so we may not need this. Well, I saw it set up correctly in Ubuntu, but not in Alpine. I wonder if this would work in Alpine if the tzdata package was installed, but my Windows machine is in a data centre in a different country and seems to have network issues connecting to the Alpine package CDN right now.

I don't have time to research this further right now, but if Windows doesn't need to be solved, then the solution might turn out to be simple.

jandubois:
> I wonder if this would work in Alpine if the tzdata package was installed

I gave it one more try, and it did indeed work. After `apk add tzdata` I shut down the distro. When I started it up again, the following 2 files appeared automatically:

```console
/ # cat /etc/timezone
America/Los_Angeles
/ # ls -l /etc/localtime
lrwxrwxrwx    1 root     0               39 Oct 20 22:19 /etc/localtime -> /usr/share/zoneinfo/America/Los_Angeles
```

Which means we don't need to propagate the host timezone to the guest on WSL2.

pkg/cidata/cidata.go
```diff
@@ -369,6 +370,49 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
...
+
+	// 2. try to get timezone from /etc/timezone
+	tzName, err := os.ReadFile("/etc/timezone")
+	if err == nil {
+		return string(tzName)
+	}
+	logrus.Debugf("cannot get timezone from /etc/timezone: %s", err)
+
+	// 3. try to get timezone from /etc/localtime soft link
+	// base on the assumption that <zoneinfo_basedir>/zone.tab file exists
```
lobshunter, line 392: Haven't found any official documents about zoneinfo data file hierarchy, just _assume_ `zone.tab` file exists.

jandubois:
@lobshunter I see you updated this PR, but it is still marked as a draft. Not sure if you are still working on it or not. When you think it is ready for another review, please move it out of draft mode and re-request a review!

pkg/cidata/cidata.go
```diff
@@ -370,6 +372,54 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
...
+		return ""
+	}
+
+	// 1. try to get timezone from env.TZ
+	tzEnv := os.Getenv("TZ")
+	if tzEnv != "" &&
+		// skip if TZ is in "std offset dst offset, rule" format
+		!strings.Contains(tzEnv, " ") {
+		return tzEnv
+	}
```
jandubois, line 388: I think trying to parse `TZ` is futile; there are just too many different possible formats, and different C runtime libraries implement them differently.

- [glib specification](https://www.gnu.org/software/libc/manual/html_node/TZ-Variable.html)
- [POSIX specification](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03)

Neither of them show any spaces in the "std offset dst offset" format, and I believe they are not even allowed optionally.

Furthermore, `cloud-init` only allows settings corresponding to zoneinfo entries:

> timezone: (string) The timezone to use as represented in /usr/share/zoneinfo

If we want to support importing the `TZ` environment variable, then I think it should be optional, and added to `/etc/environment`. There is some precent for doing this (we have logic for proxy settings), but this should be a separate issue, if we decide to do this at all. For this PR I think you should concentrate on synchronizing the `/etc/localtime` and `/etc/timezone` settings, and nothing else.

pkg/cidata/cidata.go
jandubois: I don't know either. Right now it seems like a safe assumption, but there are 2 comments in the file that might want us to look for safer alternatives:

```
# tzdb timezone descriptions (deprecated version)
[...]
# This file is intended as a backward-compatibility aid for older programs.
```

Maybe we should look for a directory named `Antarctica` (or a file named `Etc/UTC`) instead?

pkg/cidata/cidata.go
```diff
@@ -370,6 +372,54 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
...
+			_, err = os.Stat(anchorFile)
+			if err != nil {
+				continue
+			}
+
+			tz := strings.TrimPrefix(tzFilePath, baseDir+"/")
+			_, err = time.LoadLocation(tz)
+			if err == nil {
+				return tz
+			}
```
jandubois, line 414: I don't think the check if `time.LoadLocation` succeeds is useful. It is the setting that has been configured, so if the datafile somehow got corrupted, it is not our problem. And the real issue would be if the data file exists inside the VM, and we can't answer that anyways.

If we kept the check, then I would still argue that we need to `return ""` in the failure case, and not to keep looking in higher levels of the directory tree; that seems to be wrong either way.

```suggestion
			return strings.TrimPrefix(tzFilePath, baseDir+"/")
```

pkg/cidata/cidata.go
```diff
@@ -370,6 +372,54 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
...
+		}
+	}
+	logrus.Debugf("cannot get host timezone from /etc/localtime: %s", err)
+
+	logrus.Warnf("cannot get host timezone")
```
jandubois, line 419: We should have only one log message, not 2 at different levels. Since we no longer have the `err` value, we should drop the first one.

jandubois (changes requested):
I strongly feel that we don't want to look at `TZ` here, plus some additional comments.

AkihiroSuda:
@lobshunter Could you check comments from @jandubois ?

pkg/cidata/cidata.go
```diff
@@ -136,6 +137,7 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 		VMType:         *y.VMType,
 		VSockPort:      vsockPort,
 		Plain:          *y.Plain,
+		Timezone:       getHostTimezone(),
```
AkihiroSuda, line 140: Timezone shouldn't be set when `*y.Plain` is true

pkg/cidata/cidata.go
```diff
@@ -370,6 +372,54 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 	return iso9660util.Write(filepath.Join(instDir, filenames.CIDataISO), "cidata", layout)
 }
 
+// getHostTimezone tries to get host timezone from env or specific files
+func getHostTimezone() string {
```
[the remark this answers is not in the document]
AkihiroSuda: (Tried but it didn't work; `time.Now().Zone()` returns "JST" but it is missing in `/usr/share/zoneinfo/`)

PR #1937, opened 2023-10-22
Optimize the upgrade comparison by storing nerdctl version

afbjorklund:
Store the version in the containerd archives:

```json
[
  {
    "location": "https://github.com/containerd/nerdctl/releases/download/v1.6.2/nerdctl-full-1.6.2-linux-amd64.tar.gz",
    "arch": "x86_64",
    "digest": "sha256:37678f27ad341a7c568c5064f62bcbe90cddec56e65f5d684edf8ca955c3e6a4",
    "version": "1.6.2"
  },
  {
    "location": "https://github.com/containerd/nerdctl/releases/download/v1.6.2/nerdctl-full-1.6.2-linux-arm64.tar.gz",
    "arch": "aarch64",
    "digest": "sha256:ea30ab544c057e3a0457194ecd273ffbce58067de534bdfaffe4edf3a4da6357",
    "version": "1.6.2"
  }
]
```

Then we can compare it with the `nerdctl --version` later, to see if it has changed.

Pass it as a separate flag, since it is not apparent from the cache file name (sha2)...

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
@@ -8,18 +8,22 @@ fi
 # This script does not work unless systemd is available
 command -v systemctl >/dev/null 2>&1 || exit 0
 
-# Extract bin/nerdctl and compare whether it is newer than the current /usr/local/bin/nerdctl (if already exists).
-# Takes 4-5 seconds. (FIXME: optimize)
-tmp_extract_nerdctl="$(mktemp -d)"
-tar Cxzf "${tmp_extract_nerdctl}" "${LIMA_CIDATA_MNT}"/nerdctl-full.tgz bin/nerdctl
-
-if [ ! -f "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/nerdctl ] || [[ "${tmp_extract_nerdctl}"/bin/nerdctl -nt "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/nerdctl ]]; then
+# Check nerdctl version and compare whether it is different than the current /usr/local/bin/nerdctl (if already exists).
+if [ ! -f "${LIMA_CIDATA_MNT}"/nerdctl-version.txt ]; then
+	tmp_extract_nerdctl="$(mktemp -d)"
+	tar Cxzf "${tmp_extract_nerdctl}" "${LIMA_CIDATA_MNT}"/nerdctl-full.tgz bin/nerdctl
+	tmp_version_nerdctl="$("${tmp_extract_nerdctl}"/bin/nerdctl --version)"
+	rm -rf "${tmp_extract_nerdctl}"
+else
+	tmp_version_nerdctl="$(cat "${LIMA_CIDATA_MNT}"/nerdctl-version.txt)"
+fi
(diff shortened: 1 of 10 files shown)


pkg/cidata/cidata.go
```diff
@@ -357,6 +357,12 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 			Reader: nftgzR,
 		})
 	}
+	if nerdctlVersion != "" {
+		layout = append(layout, iso9660util.Entry{
+			Path:   "nerdctl-full.txt",
```
AkihiroSuda, line 362:
```suggestion
			Path:   "nerdctl-ver.txt",
```

pkg/limayaml/limayaml.go
```diff
@@ -75,6 +75,11 @@ type File struct {
 	Digest   digest.Digest `yaml:"digest,omitempty" json:"digest,omitempty"`
 }
 
+type Archive struct {
```
AkihiroSuda, line 78:
```suggestion
type NerdctlArchive struct {
```

pkg/limayaml/limayaml.go
```diff
@@ -75,6 +75,11 @@ type File struct {
...
+	File
+	Version string `yaml:"version,omitempty" json:"version,omitempty"`
```
AkihiroSuda, line 80:
```suggestion
	NerdctlVersion string `yaml:"nerdctlVersion,omitempty" json:"nerdctlVersion,omitempty"`
```

Also wondering if this can be just parsed from the URL

afbjorklund: It could, but then the URL was obfuscated in the cache:

`--nerdctl-archive /home/anders/.cache/lima/download/by-url-sha256/a23acacc818a76fadb89c79cc79548adbdac0942bf562bfc87463861561c824b/data`

pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
```diff
@@ -8,18 +8,22 @@ fi
...
+# Check nerdctl version and compare whether it is different than the current /usr/local/bin/nerdctl (if already exists).
+if [ ! -f "${LIMA_CIDATA_MNT}"/nerdctl-version.txt ]; then
+	tmp_extract_nerdctl="$(mktemp -d)"
+	tar Cxzf "${tmp_extract_nerdctl}" "${LIMA_CIDATA_MNT}"/nerdctl-full.tgz bin/nerdctl
+	tmp_version_nerdctl="$("${tmp_extract_nerdctl}"/bin/nerdctl --version)"
+	rm -rf "${tmp_extract_nerdctl}"
+else
+	tmp_version_nerdctl="$(cat "${LIMA_CIDATA_MNT}"/nerdctl-version.txt)"
+fi
+if [ ! -f "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/nerdctl ] || [ "$tmp_version_nerdctl" != "$("${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/nerdctl --version)" ]; then
```
AkihiroSuda, line 20: The user may install a custom version of nerdctl by themselves inside `limactl shell`, so we have to continue comparing the timestamp

afbjorklund: But if they can update the nerdctl binary _without_ changing the version, then I don't think this PR would work...

afbjorklund: It was assuming that the nerdctl version would be something like `0.18.0-74-g5e3dc32`, for custom versions

afbjorklund:
Maybe I was reading too much into the "Takes 4-5 seconds. (FIXME: optimize)"

PR #1951, opened 2023-10-26
 Use VZ by default for new instances on macOS >= 13.5

AkihiroSuda:
VZ is now the default type for new instances >= 13.5, unless the config is incompatible with vz (e.g., `firmware.legacyBIOS=true`, `mountType=9p`).
    
Existing instances will continue to use QEMU by default, unless `vz-identifier` is present in the instance directory.
    
Run `limactl start` with `--debug` to see how the vmType is resolved.

PR #1953, opened 2023-10-26
qemu: use 9p by default

### LINKED ISSUE #971: v1.0 roadmap: change the default mount driver from reverse-sshfs to 9p (for QEMU) and virtiofs (for vz)

AkihiroSuda:
## `qemu: 9p: graduate from experimental 
`template://experimental/9p` will be removed right before releasing Lima v1.0.

- - -

## `qemu: use 9p by default `
Templates for RHEL-like distros are updated to continue using reverse-sshfs, as 9p is not available on RHEL-like distros.

Close  #971

- - -
## `default.yaml: stop discouraging writable mounts (for 9p and virtiofs)`
Setting `writable: true` is still discouraged for reverse-sshfs due to
potential connectivity issues, but it should be fine for 9p (virtio-9p-pci)
and virtiofs.


examples/experimental/9p.yaml
```diff
@@ -1,3 +1,11 @@
+# ⚠️ ⚠️ ⚠️ `template://experimental/9p` will be removed in in Lima v1.0,
```
jandubois, line 1:
```suggestion
# ⚠️ ⚠️ ⚠️ `template://experimental/9p` will be removed in Lima v1.0,
```

examples/experimental/9p.yaml
```diff
@@ -1,3 +1,11 @@
...
+# as 9p will graduate from experimental and will be the default mount type for QEMU.
+#
+# Use the following command instead:
+# ```
+# limactl create --vm-type=qemu --mount-type=9p template://default
```
jandubois, line 6: It just said that `9p` is the default for `qemu`, so

```suggestion
# limactl create --vm-type=qemu template://default
```

jandubois (commented):
The PR looks fine, but I have concerns about how this seems to break existing instances.

I've created a `debian-12` instance with HEAD of `master`. I started and stopped the instance twice, and it worked fine.

I then switched to this PR branch and tried to start the instance. It hangs with `Waiting for the essential requirement 1 of 3: "ssh"`.

I stopped the instance and modified the mount type with

```console
$ limactl edit --mount-type reverse-sshfs debian-12
```

Trying to start the instance again, it still hangs with `Waiting for the essential requirement 1 of 5: "ssh"`. The fact that it now list "1 of 5" instead of "1 of 3" shows that the mount type has been set correctly.

I stop once more, and switch back to the `master` branch. The instance still does not start again.

I realize that this issue exists independently of this PR, but it seems like changing the default can result in permanent damage to an instance. I haven't investigated what might be causing this.

This was on macOS Monterey 12.7.1 on Intel.

I have some other idea about making either avoiding the automatic change to existing instances, or at least adding a warning when waiting for requirements, but maybe the failure to switch the debian image at all should be resolved first.

AkihiroSuda:
>  avoiding the automatic change to existing instances,

Yes, I wanted to do this, but couldn't figure out how we could implement this

jandubois:
> > avoiding the automatic change to existing instances,
> 
> Yes, I wanted to do this, but couldn't figure out how we could implement this

We can write a `lima.version` file in the instance directory during `limactl create`. We look for the file during `limactl start`. If the file is missing, we assume the version is `0.18.0`.

That way we can apply default settings based on the version of Lima used to create the instance.

We never update `lima.version`; it represents the version used to create the instance, not the version used to run it.

examples/experimental/9p.yaml
AkihiroSuda: fixed

examples/experimental/9p.yaml
AkihiroSuda: fixed

AkihiroSuda:
https://github.com/lima-vm/lima/actions/runs/9738847119/job/26873042581

Alpine is seeing `can't cd to /home/runner/work/lima/lima: No such file or directory` errors, but I can't repro this error locally

AkihiroSuda:
```
+ mount -t 9p -o 'ro,relatime,cache=8f,access=client,msize=131072,trans=virtio' mount0 /home/runner
mount: mounting mount0 on /home/runner failed: Invalid argument
LIMA 2024-07-01T07:25:44+00:00| WARNING: Failed to execute /mnt/lima-cidata/boot/04-persistent-data-volume.sh
```

jandubois:
Should this PR merge into a `release-1.0` integration branch instead of `main`? Otherwise it would no longer be possible to release the tentative 0.23 release if this PR got merged.

Alternatively this could be switched to Draft mode until 0.23 is released.

AkihiroSuda:
> Should this PR merge into a `release-1.0` integration branch instead of `main`? Otherwise it would no longer be possible to release the tentative 0.23 release if this PR got merged.
> 
> Alternatively this could be switched to Draft mode until 0.23 is released.

Can we just rename v0.23 to v1.0?

jandubois:
> Can we just rename v0.23 to v1.0?

Yes, I think that is preferable. Working on 2 releases in parallel just increases the risk of introducing regressions.

I just thought you wanted to do a 0.23 release first, as it has a separate milestone on GitHub.

AkihiroSuda:
> > Can we just rename v0.23 to v1.0?
> 
> Yes, I think that is preferable. Working on 2 releases in parallel just increases the risk of introducing regressions.
> 
> I just thought you wanted to do a 0.23 release first, as it has a separate milestone on GitHub.

Unified the milestones

examples/experimental/opensuse-tumbleweed.yaml
```diff
@@ -17,6 +17,8 @@ images:
 - location: "https://provo-mirror.opensuse.org/ports/aarch64/tumbleweed/appliances/openSUSE-Tumbleweed-ARM-JeOS-efi.aarch64.qcow2"
   arch: "aarch64"
 
+# The mount type is set to reverse-sshfs because 9p (default for QEMU) is not supported by the guest kernel. VZ users should specify `--mount-type=virtiofs` on `limactl create` for better performance.
+mountType: "reverse-sshfs"
```
jandubois, line 21: I wonder if we shouldn't have some builtin way to specify this, as an identical change is being made to about 10 templates. They will all run with `reverse-sshfs` when being used with `vz` (the new default in 1.0) unless the user explicitly overrides it.

I wish I could think about something better than this:

```yaml
# mountType will be reverse-sshfs when using QEMU or virtiofs when using VZ
mountType: not-9p
```

pkg/limayaml/defaults.go
```diff
@@ -551,9 +562,24 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 		y.MountType = o.MountType
 	}
 	if y.MountType == nil || *y.MountType == "" {
```
jandubois, line 564: Since we seem to support `"default"` for `vmType`, should we support it for `mountType` as well (would just be an alias to the empty string).

pkg/limayaml/defaults.go
```diff
@@ -551,9 +562,24 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
-		if *y.VMType == VZ {
+		switch *y.VMType {
+		case VZ:
 			y.MountType = ptr.Of(VIRTIOFS)
-		} else {
+		case QEMU:
+			y.MountType = ptr.Of(NINEP)
+			// Use REVSSHFS if the instance was created with Lima prior to v1.0
+			if _, err := os.Stat(filePath); err == nil { // existing instance
```
jandubois, line 571: There is a more complex `isExistingInstanceDir` function in #1951 that should probably be used here too.

pkg/limayaml/defaults.go
```diff
@@ -551,9 +562,24 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
+				switch {
+				case existingLimaVersion == "": // created with Lima < 0.20
+					y.MountType = ptr.Of(REVSSHFS)
```
jandubois, line 574: This case is identical to the default case, so you could just update the comment there.

pkg/limayaml/defaults.go
```diff
@@ -551,9 +562,24 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
...
+				case strings.TrimPrefix(existingLimaVersion, "v") == "1.0.0",
+					versionutil.GreaterThan(existingLimaVersion, "1.0.0"):
```
jandubois, line 576: I would prefer to add a `versionutil.GreaterEqual()` function. The `TrimPrefix` is an implementation detail that should not leak out of that package.

jandubois (approved):
Thanks, looks good, but I think I would prefer to see something like `mountType: not-9p` added (better name welcome).

pkg/limayaml/defaults.go
afbjorklund: Mount type was a string before, so old templates/instances used "" (not null)

examples/experimental/opensuse-tumbleweed.yaml
AkihiroSuda: What about
```yam
mountType: "default"
mountTypeUnsupported: ["9p"]
```

jandubois: Yes, that is better and more flexible. But since it takes a list it should be plural: 

```diff
  mountType: "default"
- mountTypeUnsupported: ["9p"]
+ mountTypesUnsupported: ["9p"]
```

AkihiroSuda:
Will rebase after merging:
- https://github.com/lima-vm/lima/pull/1951

examples/experimental/opensuse-tumbleweed.yaml
AkihiroSuda: done

pkg/limayaml/defaults.go
AkihiroSuda: done

pkg/limayaml/defaults.go
AkihiroSuda: done

pkg/limayaml/defaults.go
AkihiroSuda: done

AkihiroSuda:
Rebased and addressed the comments.

Sorry for taking too long for this.

pkg/limayaml/defaults.go
AkihiroSuda: added "default"

AkihiroSuda:
CI seems failing for archlinux
```
time="2024-10-04T06:46:43Z" level=fatal msg="failed to get unprivileged mount flags for \"/home/runner/lima-container-engine-test-tmp/random\": stat /home/runner/lima-container-engine-test-tmp/random: no such file or directory"
+ got=
```

AkihiroSuda:
Sorry, accidentally edited https://github.com/lima-vm/lima/pull/1953#issuecomment-2397530454 (now reverted)

AkihiroSuda:
We should eventually entirely switch to virtiofs, when QEMU supports virtiofs on macOS. (Has anybody tried to working on it?)

pkg/limayaml/defaults.go
```diff
@@ -171,6 +172,15 @@ func defaultGuestInstallPrefix() string {
 //   - CACertificates Files and Certs are uniquely appended in d, y, o order
 func FillDefault(y, d, o *LimaYAML, filePath string) {
 	instDir := filepath.Dir(filePath)
+
+	var existingLimaVersion string // empty if the instance was created with Lima prior to v0.20
```
jandubois, line 176: Maybe add a comment that it can also be empty when editing a template file without an instance, i.e. `limactl edit foo.yaml`.

That's why `existingLimaVersion` should always be used with `isExistingInstanceDir` below.

I just noticed that we could also check `filepath.Base(filePath) == "lima.yaml"`, but `isExistingInstanceDir` only receives the directory part, so it won't work. I think the current check is probably good enough.

pkg/limayaml/defaults.go
```diff
@@ -171,6 +172,15 @@ func defaultGuestInstallPrefix() string {
...
+	limaVersionFile := filepath.Join(instDir, filenames.LimaVersion)
+	if b, err := os.ReadFile(limaVersionFile); err == nil {
+		existingLimaVersion = strings.TrimSpace(string(b))
+	} else if !errors.Is(err, os.ErrNotExist) {
+		logrus.WithError(err).Warnf("Failed to read %q", limaVersionFile)
+	}
```
jandubois, line 182: Could this be factored out into a shared function in the `store` package?

pkg/version/versionutil/versionutil.go
```diff
@@ -38,3 +38,9 @@ func GreaterThan(limaVersion, oldVersion string) bool {
 	}
 	panic("unreachable")
 }
+
+// GreaterEqual return true if limaVersion >= oldVersion.
+func GreaterEqual(limaVersion, oldVersion string) bool {
+	return strings.TrimPrefix(limaVersion, "v") == strings.TrimPrefix(oldVersion, "v") ||
+		GreaterThan(limaVersion, oldVersion)
+}
```
jandubois, line 46: This function is subtly different from `GreaterThan` because it doesn't strip the git describe suffix from `limaVersion`.

I think it would be better to use something like this (untested):
```go
func compare(limaVersion, oldVersion string) int {
	if limaVersion == "" {
		return -1
	}
	version, err := Parse(limaVersion)
	if err != nil {
		return 1
	}
	cmp := version.Compare(*semver.New(oldVersion)) {
    if cmp == 0 && strings.Contains(limaVersion, "-") {
        cmp = 1
    }
    return cmp
}

func GreaterThan(limaVersion, oldVersion string) bool {
    return compare(limaVersion, oldVersion) > 0
}

func GreaterEqual(limaVersion, oldVersion string) bool {
    return compare(limaVersion, oldVersion) >= 0
}
```

website/content/en/docs/faq/_index.md
```diff
@@ -230,6 +231,25 @@ mounts:
...
+- Debian GNU/Linux
+- openSUSE
+
+A new instance of these OS still use `reverse-sshfs` by default.
+However, an existing instance created with a previous version of Lima may potentially need
+running the following command (usually not needed):
+
+```
+limactl edit --mount-type=reverse-sshfs <NAME>
+```
```
jandubois, line 249: When will this be needed? How could the instances have started if the mount type was not supported?

jandubois (changes requested):
Looks mostly fine, but I have some comments about the code.

Also haven't actually tested it yet, but will try to do so later today.

examples/default.yaml
```diff
@@ -112,9 +111,13 @@ mounts:
   # 🔵 This file: true (only for "/tmp/lima")
   writable: true
 
-# Mount type for above mounts, such as "reverse-sshfs" (from sshocker), "9p" (EXPERIMENTAL, from QEMU’s virtio-9p-pci, aka virtfs),
+# Deny list of mount types for resolving the "default" type.
+# 🟢 Builtin default: null
+mountTypesUnsupported: null
+
+# Mount type for above mounts, such as "reverse-sshfs" (from sshocker), "9p" (EXPERIMENTAL until Lima v1.0, from QEMU’s virtio-9p-pci, aka virtfs),
```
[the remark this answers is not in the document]
AkihiroSuda: I don't think vz supports 9p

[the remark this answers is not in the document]
jandubois: Yeah, I may have been confused by the fact that WSL2 is using it.

pkg/version/versionutil/versionutil.go
jandubois: Also please add tests for `GreaterEqual`.

examples/default.yaml
```diff
@@ -112,9 +111,13 @@ mounts:
   # 🔵 This file: true (only for "/tmp/lima")
   writable: true
 
-# Mount type for above mounts, such as "reverse-sshfs" (from sshocker), "9p" (EXPERIMENTAL, from QEMU’s virtio-9p-pci, aka virtfs),
+# Deny list of mount types for resolving the "default" type.
```
jandubois, line 114: This comment is slightly inaccurate, as the setting is not only used to resolve the default mount type, but it also fails template validation when you set `mountType` to an unsupported setting (which is good).

Maybe

```suggestion
# List of mount types not supported by the kernel of this distro.
# Also used to resolve the default mount type when not explicitly specified.
```

examples/default.yaml
```diff
@@ -71,8 +71,7 @@ mounts:
   # Configure the mountPoint inside the guest.
   # 🟢 Builtin default: value of location
   mountPoint: null
-  # CAUTION: `writable` SHOULD be false for the home directory.
-  # Setting `writable` to true is possible, but untested and dangerous.
+  # Setting `writable` to true is discouraged when mountType is set to "reverse-sshfs".
```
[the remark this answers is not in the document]
AkihiroSuda: > Is this really an issue? 

idk, but I just feel that Ethernet connection is less reliable compared to virtio PCI.


> What is the worst that can happen when you get an ssh disconnect?


Modifications can be lost

hack/test-templates/test-misc.yaml
```diff
@@ -1,5 +1,6 @@
 # The test template for testing misc configurations:
 # - disk
+# - snapshots
```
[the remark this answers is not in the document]
AkihiroSuda: Seems already added?

pkg/limayaml/defaults.go
AkihiroSuda: Can be done later, when the logic has to be duplicated in another place?

website/content/en/docs/faq/_index.md
AkihiroSuda: Practically never.
Just added a hint for potential bugs.

pkg/version/versionutil/versionutil.go
AkihiroSuda: done

examples/default.yaml
AkihiroSuda: done

pkg/limayaml/defaults.go
AkihiroSuda: done

.github/workflows/test.yml
```diff
@@ -256,6 +256,14 @@ jobs:
       if: matrix.template == '../hack/test-templates/test-misc.yaml'
     - name: "Show cache"
       run: ./hack/debug-cache.sh
+    - name: "Adjust LIMACTL_CREATE_ARGS"
+      # Change the mount type from 9p to reverse-sshfs, as a workaround for:
+      #
+      # > level=fatal msg="failed to get unprivileged mount flags for \"/home/runner/lima-container-engine-test-tmp/random\": stat /home/runner/lima-container-engine-test-tmp/random: no such file or directory"
```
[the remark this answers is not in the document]
AkihiroSuda: done
- https://github.com/lima-vm/lima/issues/2701

jandubois (approved):
Thanks, LGTM

PR #1956, opened 2023-10-27
docs: replace deprecated `useHostResolver` with new `hostResolver.enabled`

choznerol:
Update the doc according to: https://github.com/lima-vm/lima/blob/73200003a79cfe829caa2ee9cce7c68c99e33bcb/pkg/limayaml/limayaml.go#L37


jandubois (approved):
Thank you, LGTM

Issue #1940, opened 2023-10-23
`--set` with `.dns` doesn't always seem to work 


pmalek:
### Description

I use the following set of flags

```
'--set=.cpus = 6 | .memory = "10GiB" | .arch = "aarch64" | .vmType = "vz"  | .dns = []  | .dns += "192.168.1.250"'
```

Recently (with b5aa0dd) discovered that the dns setting doesn't always work as expected. I'm not sure what should be the effect of that setting but sometimes it does put `192.168.1.250` in `/etc/resolv.conf` and sometimes doesn't.

What should be the effect of that setting?

---

```
limactl -v
limactl version HEAD-b5aa0dd
```

balajiv113:
@pmalek Which template are you using ?? 
Is it Alpine or Ubuntu ??

neowu:
i encountered same behavior, after digging little into source code

it looks like, if vmType == VZ, then it's always using args.SlirpDNS (which is gatewayIp) to generate user-data
and bypass DNS settings in lima.yml, then output user-data like

```
manage_resolv_conf: true

resolv_conf:
  nameservers:
  - 192.168.5.2
```

not sure whether it's intended behavior.

from the source 
https://github.com/lima-vm/lima/blob/391133d0b37532d0ac94872a81c3754f92c2531c/pkg/cidata/cidata.go#L268
always go into first if condition branch
```
	if firstUsernetIndex != -1 || *y.VMType == limayaml.VZ {
		args.DNSAddresses = append(args.DNSAddresses, args.SlirpDNS)
	} else if *y.HostResolver.Enabled {
		args.UDPDNSLocalPort = udpDNSLocalPort
		args.TCPDNSLocalPort = tcpDNSLocalPort
		args.DNSAddresses = append(args.DNSAddresses, args.SlirpDNS)
	} else if len(y.DNS) > 0 {
		for _, addr := range y.DNS {
			args.DNSAddresses = append(args.DNSAddresses, addr.String())
		}
	} else {
		args.DNSAddresses, err = osutil.DNSAddresses()
		if err != nil {
			return err
		}
	}
```

balajiv113:
@neowu Yes that's the issue. Thanks for the analysis.
Willing to take it for fix ?? Or i can fix it

neowu:
@balajiv113 I am willing to submit PR, though I haven't worked on Go project before :)

just need to make sure the expected behavior, do we always want to use DNS settings if existed, something like
```
[...]

PR #1962, opened 2023-10-28
fix: cidata should respect DNS settings in lima.yaml

neowu:
fixes for lima-vm/lima#1940


neowu:
from the log, the "debian" integration test failed on phase 2 (test after a restarting), the first phase run successfully, 
it looks like caused by hostagent timeout, seems not relates to my change (otherwise phase 1 would fail)

anyway pls double check, thanks

```
time="2023-10-28T06:25:09Z" level=info msg="[hostagent] The optional requirement 2 of 2 is satisfied"
time="2023-10-28T06:25:09Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:25:49Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:26:29Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:27:14Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:28:48Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:29:44Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:30:24Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:31:05Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:31:50Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:32:31Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:33:14Z" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2023-10-28T06:33:37Z" level=fatal msg="did not receive an event with the \"running\" status"
time="2023-10-28T06:33:37Z" level=info msg="Sending SIGKILL to the qemu driver process 4939"```

AkihiroSuda:
https://github.com/apps/dco

neowu:
I have signed the commit, please review

balajiv113:
@neowu Changes looks good. Basically we are giving priority to dns first. Now am just thinking why not support dns as a secondary ones.

We always use slirp dns, additional to it we also use dns if its configured. This way slirp also will be there and custom DNS also can be configured.

@AkihiroSuda what do you think about this ?? I believe this was the intention to dns property itself.

jandubois:
> We always use slirp dns, additional to it we also use dns if its configured. This way slirp also will be there and custom DNS also can be configured.

No, this is not how DNS should be configured. You normally have multiple redundant DNS servers in case one of them is unavailable (e.g. because of maintenance). They could also be load-balancing. But they are not used as fallback if one DNS doesn't recognize a name.

So if e.g. one DNS has additional entries for a VPN, and the other one doesn't, then it is random if the VPN names will resolve. The resolver should pick one DNS at random, and only try another one if the first one doesn't respond within a timeout. It will not retry if it got a NOTFOUND response.

Different resolver implementations may implement things differently though, e.g. the musl implementation does send requests to multiple servers in parallel and then picks the first one to answer (which still could be NXDOMAIN, failing the lookup).

So in general it is a bad idea to configure multiple DNS servers unless they cover exactly the same domains.

jandubois (approved):
Thanks, LGTM

PR #1968, opened 2023-10-30
vz: trivial changes

AkihiroSuda:
Split from:
- https://github.com/lima-vm/lima/pull/1951

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -16,6 +16,14 @@ type Driver interface {
 	// Validate returns error if the current driver isn't support for given config
 	Validate() error
 
+	// Initialize is called on creating the instance for initialization.
+	// (e.g., creating "vz-identifier" file)
+	//
+	// Initialize MUST return nil when it is called against an existing instance.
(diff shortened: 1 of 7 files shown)


pkg/driver/driver.go
```diff
@@ -16,6 +16,14 @@ type Driver interface {
 	// Validate returns error if the current driver isn't support for given config
 	Validate() error
 
+	// Initialize is called on creating the instance for initialization.
+	// (e.g., creating "vz-identifier" file)
+	//
+	// Initialize must not return a non-nil error when it is called against an existing instance.
```
jandubois, line 22: Not a fan of double-negation

```suggestion
	// Initialize MUST return nil when it is called against an existing instance.
```

jandubois (dismissed):
LGTM, just nit-picking on a single comment

pkg/driver/driver.go
AkihiroSuda: done

balajiv113 (approved):

Issue #1964, opened 2023-10-30
vz A lima _disk converted from qcow2 > raw on startup truncates to 0 bytes and can't be used in the vm. 


mhio:
### Description

A lima disk created as the default qcow2 format, and added to a vz debian 12 vm, gets truncated to 0 bytes on `limactl start vm` and the device is not usable. 

Here are the steps:
```
% limactl disk create one --size 1G
INFO[0000] Creating qcow2 disk "one" with size 1GiB 
% limactl disk list
NAME           SIZE     DIR                                     IN-USE-BY
one            1GiB     /Users/me/.lima/_disks/one
```
lima.yml for `withone`
```
images:
  - location: 'https://cloud.debian.org/images/cloud/bookworm/20231013-1532/debian-12-genericcloud-amd64-20231013-1532.qcow2'
    arch: 'x86_64'
    digest: 'sha512:b2ddc01e8d13dabbcfde6661541aae92219be2d442653950f0e44613ddebaeb80dc7a83e0202c5509c5e72f4bd1f4edee4c83f35191f2562b3f31e20e9e87ec2'
  - location: 'https://cloud.debian.org/images/cloud/bookworm/latest/debian-12-genericcloud-amd64.qcow2'
    arch: 'x86_64'
vmType: 'vz'
mountType: 'virtiofs'
additionalDisks:
  - name: 'one'
    format: false
```
```
% limact start withone
INFO[0000] Using the existing instance "withone"        
INFO[0002] [hostagent] hostagent socket created at /Users/me/.lima/withone/ha.sock 
INFO[0002] [hostagent] Starting VZ (hint: to watch the boot progress, see "/Users/me/.lima/withone/serial*.log") 
INFO[0002] [hostagent] new connection from  to          
INFO[0002] [hostagent] Mounting disk "one" on "/mnt/lima-one" 
INFO[0002] [hostagent] Converting "/Users/me/.lima/_disks/one/datadisk" (qcow2) to a raw disk "/Users/me/.lima/_disks/one/datadisk" 
INFO[0002] [hostagent] 1.00 GiB / 1.00 GiB (100.00%) ? p/s 
INFO[0003] SSH Local Port: 52309                        
INFO[0003] [hostagent] Waiting for the essential requirement 1 of 3: "ssh" 
INFO[0003] [hostagent] [VZ] - vm state change: running  
INFO[0013] [hostagent] Waiting for the essential requirement 1 of 3: "ssh" 
[...]

PR #1972, opened 2023-10-30
Add disk convert command for qcow2 to raw

afbjorklund:
Mostly for reproducing #1964

`limactl disk create one --size 1G`
`limactl --debug disk convert one`

The `datadisk` is now 0 (zero) bytes.


jandubois:
I don't think we should burden the user with this. Lima just needs to fix the conversion to explicitly use a temp file, and then users never need to deal with it.

Issue #1053, opened 2022-09-15
Stop cleaning the vendor directory on make


afbjorklund:
### Description

It is fine that it is removed on "clean", but annoying that it is deleted on "make".

I sometimes use `go mod vendor` for reference, and then I have to run it again...

My workaround is `make exe` (also skips rebuilding the agents)

But I would prefer if make didn't touch any existing "vendor" dir ?

jandubois:
I don't have strong opinions, but would prefer that `make all` be equivalent to `make clean binaries` even if the default target doesn't invoke `clean` anymore.

afbjorklund:
But `all` _is_ the default target, so then it doesn't matter ?

At least lima doesn't **only** call clean, like machine did...

jandubois:
> But `all` _is_ the default target, so then it doesn't matter ?

No, I meant we could make `binaries` the default target, and remove the dependency on `clean`, like

```Makefile
binaries:
        ...

all: clean binaries
...
```

afbjorklund:
There is clearly some system to these makefiles, with the `_output` and the automatic clean and so on...
My assumptions are more based on autotools and friends (or maybe just make), and it had default targets.

The build targets are also "phony", which means that make is out of the loop anyway (rebuilds every time).
For some other projects I did try to "make depend" work, but it wasn't really worth it (similar to with javac)

Anyway, if `make all` starts cleaning that is also confusing 😀

So that is why I thought it would be better to just leave it as-is

jandubois:
> Anyway, if `make all` starts cleaning that is also confusing 😀

It is already cleaning because right now `binaries` is cleaning:

```Makefile

all: binaries

binaries: clean
```

I was just removing the `clean` from `binaries` and making `binaries` the default.

> So that is why I thought it would be better to just leave it as-is

I'm fine with leaving everything as-is; I thought you wanted `make` to not run `clean`.

afbjorklund:
> I thought you wanted `make` to not run `clean`.

Right, but "binaries" would still be cleaning anyway ?

Actually the issue was **only** about the (go mod) "vendor" directory...
(I don't really care if the binaries are deleted/rebuilt, because cache)

afbjorklund:
The "clean" target is also breaking parallel builds (currently disabled)

```Makefile
.NOTPARALLEL:
```

afbjorklund:
Unfortunately go does **not** support the `make` job server (yet?)

So removing the serial build would spawn up to 5 times as many
[...]

PR #1973, opened 2023-11-01
Change Make default target not to call `clean`

takanohi:
Picking this good first issue #1053 .

```
# Before
make == make all
make all == make binaries (with clean) manpages

# After
make == make binaries (without clean)
make all == make clean binaries manpages
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,31 +47,6 @@ GO_BUILD := $(GO) build -ldflags="-s -w -X $(PACKAGE)/pkg/version.Version=$(VERS
 
 .NOTPARALLEL:
 
-.PHONY: all
-all: binaries manpages
-
-.PHONY: help
-help:
-	@echo  '  binaries        - Build all binaries'
-	@echo  '  manpages        - Build manual pages'
-
-exe: _output/bin/limactl$(exe)
-
-.PHONY: minimal
-minimal: clean \
-	_output/bin/limactl$(exe) \
-	codesign \
-	_output/share/lima/lima-guestagent.Linux-$(shell uname -m | sed -e s/arm64/aarch64/)
-	mkdir -p _output/share/lima/templates
-	cp -aL examples/default.yaml _output/share/lima/templates/
-
-config: Kconfig
-	$(KCONFIG_CONF) $<
-
-menuconfig: Kconfig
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -75,7 +75,7 @@ jobs:
     - name: Unit tests
       run: go test -v ./...
     - name: Make
-      run: make
+      run: make all
```
AkihiroSuda, line 78: `make` has to continue to work

AkihiroSuda (changes requested):
`make` w/o target has to continue to work

afbjorklund:
It was also requested to continue to clean...

.github/workflows/test.yml
takanohi: fixed

afbjorklund:
I don't think this change is related to the issue that I opened, the question was if make could stop removing `vendor`

This is the code involved with that:

```Makefile
.PHONY: clean
clean:
	rm -rf _output vendor
```

If `make` (or `make all`, same thing) should call clean or not, is more of an optimization issue (over risk of stale build)...

afbjorklund (changes requested):
I don't think we should add clean only to the "all" target, and remove it from `make`

takanohi:
@afbjorklund 

You wrote a description in the issue.
`It is fine that it is removed on "clean", but annoying that it is deleted on "make".`

So changing default target will fix your problem immediately. (`make` doesn't remove vendor anymore.)
jandubois also had commented [here](https://github.com/lima-vm/lima/issues/1053#issuecomment-1250126025).

Do you want to remove `clean` process from the whole Makefile?
I think it will be our next action in another PR.

afbjorklund:
I think they are two different topics, but I am also willing to just do nothing at all (leave as-is)...

If my vendor directory gets wiped, I can do `go mod vendor` again. And avoid to run `make all`
The running of clean and build in parallel is another topic, but there is a workaround in place.
And there are limitations to `go build`, so I think that what we have is "good enough" for now.

afbjorklund:
@takanohi thanks for offering to help!

takanohi:
If leaving as-is is ok for you, this PR can be closed.
It's helpful if you close the issue again.

PR #1981, opened 2023-11-02
`codesign --xml` requires macOS Monterey or later

### LINKED ISSUE #1975: On startup while code sign --display command fails due to unrecognized ---xml option on Catalina
### Description

It seems that on older MacOS versions (**Catalina**), `the --xml option` is not supported by the command `codesign` and the check fails for the status `signedness`.

```
$ limactl start
WARN[0000] QEMU binary "/opt/local/bin/qemu-system-x86_64" does not seem properly signed with the "com.apple.security.hypervisor" entitlement  error="failed to run [codesign --display --entitlements - --xml /opt/local/bin/qemu-system-x86_64]: exit status 2 (out=\"codesign: unrecognized option `--xml'\\nUsage: codesign -s identity [-fv*] [-o flags] [-r reqs] [-i ident] path ... # sign\\n       codesign -v [-v*] [-R=<req string>|-R <req file path>] path|[+]pid ... # verify\\n       codesign -d [options] path ... # display contents\\n       codesign -h pid ... # display hosting paths\\n\")"
? Try to sign "/opt/local/bin/qemu-system-x86_64" with the "com.apple.security.hypervisor" entitlement? Yes
WARN[0001] Failed to sign "/opt/local/bin/qemu-system-x86_64"  error="failed to run [codesign --sign - --entitlements /var/folders/bj/58cccq_j2hbgglcw7xvfsl2h0000gn/T/lima-qemu-entitlements-612410575.xml --force /opt/local/bin/qemu-system-x86_64]: exit status 1 (out=\"/opt/local/bin/qemu-system-x86_64: replacing existing signature\\n/opt/local/bin/qemu-system-x86_64: resource fork, Finder information, or similar detritus not allowed\\n\")"
```

jandubois:
So just skip the check for missing entitlements on older macOS versions.

macOS 11 (Big Sur) is probably going to be end-of-lifed soon, but there is no reason not to add a simple affordance to keep Lima working, even if we no longer officially support the OS versions.

Fixes #1975

### DIFF
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -13,8 +13,10 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/lima-vm/lima/pkg/driver"
 	"github.com/lima-vm/lima/pkg/driverutil"
+	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/qemu"
 	"github.com/lima-vm/lima/pkg/qemu/entitlementutil"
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #1984, opened 2023-11-05
limactl create: don't override user-specified name

ZhongRuoyu:
Currently, when creating a new instance, the instance name specified by the user via `--name` gets discarded when a template is chosen later; the template name is used instead:

```console
$ ./_output/bin/limactl create --name test
? Creating an instance "test" Choose another template (docker, podman, archlinux, fedora, ...)
? Choose a template ubuntu-lts
? Creating an instance "ubuntu-lts" Proceed with the current configuration
[output truncated...]
INFO[0005] Run `limactl start ubuntu-lts` to start the instance.
```

Fix that by only setting the instance name to the template name when it's not already set. After this change:

```console
$ ./_output/bin/limactl create --name test
? Creating an instance "test" Choose another template (docker, podman, archlinux, fedora, ...)
? Choose a template ubuntu-lts
? Creating an instance "test" Proceed with the current configuration
[output truncated...]
INFO[0005] Run `limactl start test` to start the instance.
```

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -425,9 +425,11 @@ func chooseNextCreatorState(st *creatorState, yq string) (*creatorState, error) 
 				return st, fmt.Errorf("invalid answer %d for %d entries", ansEx, len(templates))
 			}
 			yamlPath := templates[ansEx].Location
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

PR #1991, opened 2023-11-06
Add Kubernetes examples with Docker socket

afbjorklund:
Some users are expecting to run both "Docker" and Kubernetes

Uses the regular Ubuntu system packages, for the installation

An alternative is to use Kubernetes-in-Docker, or even two VMs...

But then you can't share images between the two environments.

----

Experiment for Podman Desktop, possibly too bloated for upstream?

* https://github.com/containers/podman-desktop/pull/4539

It's a merge between k8s.yaml and docker-rootful/podman-rootful

Nobody uses the `podman-remote` client, so show `docker` instead.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -46,6 +46,11 @@ Container orchestration:
 - [`k3s`](./k3s.yaml): Kubernetes via k3s
 - [`k8s`](./k8s.yaml): Kubernetes via kubeadm
 
+Container runtimes:
+- [`k8s-nerdctl`](./experimental/k8s-nerdctl.yaml): Kubernetes via kubeadm, with containerd/buildkitd
+- [`k8s-docker`](./experimental/k8s-docker.yaml): Kubernetes via kubeadm, with Docker and CRI-Docker
+- [`k8s-podman`](./experimental/k8s-podman.yaml): Kubernetes via kubeadm, with Podman and CRI-O
+
 Optional feature enablers:
 - [`vmnet`](./vmnet.yaml): ⭐enable [`vmnet.framework`](../docs/network.md)
 - [`experimental/9p`](./experimental/9p.yaml): [experimental] use 9p mount type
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
We have too many yamls now, maybe we should consider to support parameterizing?

afbjorklund:
If there was a working (non-GCE ) get.k8s.io script, that would probably have helped too...

The k3s template is much shorter.

jandubois:
> The k3s template is much shorter.

You can just copy the `k3s` provisioning script into the `docker` template and add the `--docker` option to `k3s` get what you want... 😸

```shell
curl -sfL https://get.k3s.io | INSTALL_K3S_EXEC="--docker" sh -
```

afbjorklund:
Yeah, but the script is deprecated and only supports GCE (it does not even use `kubeadm`).

https://get.k8s.io

And the project prefers the interactive readme, over any automated script (only downstream).

https://docs.k8s.io (as in https://kubernetes.io/docs/setup/production-environment/ et al)

So, just use Rancher Desktop  perhaps ? 😀

It is more optimized and integrated, anyway.

----

But it was mostly an experiment, to see how much modification was needed to add docker/podman?
And I think there wasn't much unexpected, mostly just to install the package and forward the socket.

It is unlikely that Podman Desktop will use something else than `kind` (kubernetes-in-podman), anyway...
Just that the kind project does not want to support cri-o, so that means they have to run with containerd.


It would be possible to add a CRI-O installation to CoreOS, but then it competes with [OpenShift Local](https://developers.redhat.com/products/openshift-local/overview)...
So that is also unlikely to happen. And now their main focus is on Compose, not actually on Kubernetes.

Most likely outcome is that I will just close these, and run with the regular Lima containerd-only clusters.
It _would_ have been possible to use the CLI, but not if you only support the legacy API (not even `ssh://`).

The feature will be more of a "DIY" / `edit`.

If there is a unix socket, the GUI can use it?

afbjorklund:
We will leave these as downstream (from Lima), as some example of DYI clusters (adding on k8s.yaml).

The initial use case was Podman Desktop, which divides workflow between "containers" and "Kubernetes".

But it will pursue some path of sharing storage between the guest and the kubernetes-in-container node:

https://github.com/kubernetes/minikube/issues/17415#issuecomment-1817826884

----

If you don't need access to the Legacy Docker socket, then you _could_ just run `nerdctl.lima` out-of-the-box...

Like is done in Rancher Desktop, or Finch.

PR #1998, opened 2023-11-14
Support for native driver specific guest agent connection

balajiv113:
Implementation for #1925 

Advantages
- Relies on native driver way of host <-> guest communication
- Works even without network
- No extra round trip like in ssh forwarding
- Less load on network

Tasks
- [x] Vz support
- [x] qemu support
- [x] windows vsock support
- [x] Testing and clean-up


cmd/lima-guestagent/daemon_linux.go
```diff
@@ -26,16 +27,24 @@ func newDaemonCommand() *cobra.Command {
 	return daemonCommand
 }
 
+var (
+	vSockPort = 0
+
+	qemuFile = "/dev/virtio-ports/lima.guest_agent.0"
```
AkihiroSuda, line 33: Could you document these virtio ports in https://lima-vm.io/docs/dev/internals/ ?

AkihiroSuda: Also, "qemuFile" sounds confusing.
Maybe this should be called `virtioPort` ?

AkihiroSuda:
```suggestion
	virtioPort = "/dev/virtio-ports/io.lima-vm.guest_agent.0"
```

pkg/qemu/qemu.go
```diff
@@ -43,7 +43,10 @@ type Config struct {
 }
 
 // MinimumQemuVersion is the minimum supported QEMU version
-const MinimumQemuVersion = "4.0.0"
+const (
+	MinimumQemuVersion = "4.0.0"
+	LimaGuestSock      = "lima-guest.sock"
```
AkihiroSuda, line 48: Can this be called `ga.sock` to match `ha.sock` ?

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -40,9 +40,5 @@ else
 	# Remove legacy systemd service
 	rm -f "${LIMA_CIDATA_HOME}/.config/systemd/user/lima-guestagent.service"
 
-	if [ "$LIMA_CIDATA_VMTYPE" = "wsl2" ]; then
-		sudo "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/lima-guestagent install-systemd --vsock-port "${LIMA_CIDATA_VSOCK_PORT}"
-	else
-		sudo "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/lima-guestagent install-systemd
-	fi
+	sudo "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/lima-guestagent install-systemd --vsock-port "${LIMA_CIDATA_VSOCK_PORT}"
```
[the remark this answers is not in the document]
balajiv113: True but by default the port will be 0 for qemu so either way it will be ignored

pkg/qemu/qemu.go
balajiv113: This was un-used, removed it

cmd/lima-guestagent/install_systemd_linux.go
```diff
@@ -49,6 +49,7 @@ func installSystemdAction(cmd *cobra.Command, _ []string) error {
 	args := [][]string{
 		{"daemon-reload"},
 		{"enable", "--now", "lima-guestagent.service"},
+		{"try-restart", "lima-guestagent.service"},
```
balajiv113, line 52: This was needed to properly upgrade on first start.

Without this, the below case was failing
lima 0.15 create and start -> stop -> lima 1 start -> guest is updated but not restarted to pick up latest one -> This will be taken only after second restart

website/content/en/docs/dev/Internals/_index.md
```diff
@@ -71,7 +71,11 @@ VNC:
 - `vncpassword`: VNC display password
 
 Guest agent:
-- `ga.sock`: Forwarded to `/run/lima-guestagent.sock` in the guest, via SSH
+
+Each drivers use their own mode of communication 
+- `qemu`: uses virtio-port `io.lima-vm.guest_agent.0`
+- `vz`: uses vsock port 2222
+- `wsl2`: uses free random vsock port
```
AkihiroSuda, line 78: Can't we use a random port for vz too?

balajiv113: Vz create port on per vm basis so there is no way to get it ahead of time. 

Since we are bootstrapping the vm i think we can be sure that it will be free

AkihiroSuda (approved):
Thanks

afbjorklund:
This seems to break connectivity on non-native and experimental drivers, due to missing directories.

AkihiroSuda:
> This seems to break connectivity on non-native and experimental drivers, due to missing directories.

Does this only affect #2000?
Instances created with an existing release are not affected IIUC?

afbjorklund:
As long as you have a working cidata/cloud-init and a regular driver, it should work (and was tested)

But I was able to restore the functionality, and also removed the ugly hardcoded paths in lima-guestagent

PR #2000, opened 2023-11-15
Generic/External internal driver for all OS

afbjorklund:
Similar to the docker-machine "generic" driver, bring your own virtual machine (or physical server)

Not so useful in itself, but not so bad when made into a real driver or wrapped with helper scripts...

```
NAME          STATUS     SSH                   VMTYPE    ARCH      CPUS    MEMORY    DISK      DIR
beaglebone    Running    192.168.7.2:22        ext       armv7l    1       512MiB    4GiB      ~/.lima/beaglebone
core          Stopped    127.0.0.1:0           qemu      x86_64    1       1GiB      100GiB    ~/.lima/core
```

```yaml
vmType: ext

arch: "armv7l"
cpus: 1
memory: 512MiB
disk: 4GiB

# We do not have arm-v7 binaries of containerd
containerd:
  system: false
  user: false

ssh:
  address: 192.168.7.2
```

Requires:

* https://github.com/lima-vm/lima/pull/2006

----

Installed lima-guestagent, and nerdctl from tarballs/binaries.

```
$ _output/bin/limactl shell beaglebone nerdctl version
Client:
 Version:	v1.7.0
 OS/Arch:	linux/arm
 Git commit:	e674fe7ba6e49f12e88cd9c6c442e7ea5232502c
 buildctl:
  Version:	v0.12.3
  GitCommit:	438f47256f0decd64cc96084e22d3357da494c27

Server:
 containerd:
  Version:	v1.7.6
  GitCommit:	091922f03c2762540fd057fba91260237ff86acb
 runc:
  Version:	1.1.9
  GitCommit:	v1.1.9-0-gccaecfc
```

`limactl guest-install beaglebone`

* #2028
----

Hardware

https://www.beagleboard.org/boards/beaglebone-black

You could also use a Raspberry Pi Zero*, or a cloud droplet.

\* need the Zero 2, for arm-v7 (previous model was arm-v6)

https://www.raspberrypi.com/products/raspberry-pi-zero-2-w/

Discussion

* https://github.com/lima-vm/lima/discussions/2029

### DIFF
--- a/cmd/limactl/main_ext.go
+++ b/cmd/limactl/main_ext.go
@@ -1,0 +1,7 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+// Import ext driver to register it in the registry on all platforms.
+import _ "github.com/lima-vm/lima/v2/pkg/driver/ext"
--- a/pkg/driver/ext/ext_driver.go
+++ b/pkg/driver/ext/ext_driver.go
@@ -1,0 +1,176 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ext
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"slices"
+
+	"github.com/lima-vm/lima/v2/pkg/driver"
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/lima-vm/lima/v2/pkg/ptr"
+)
+
+type LimaExtDriver struct {
+	Instance *limatype.Instance
+}
+
+var _ driver.Driver = (*LimaExtDriver)(nil)
+
+func New() *LimaExtDriver {
+	return &LimaExtDriver{}
+}
+
+func (l *LimaExtDriver) Configure(inst *limatype.Instance) *driver.ConfiguredDriver {
+	l.Instance = inst
+
+	return &driver.ConfiguredDriver{
+		Driver: l,
+	}
+}
+
+func (l *LimaExtDriver) Validate(ctx context.Context) error {
+	return validateConfig(ctx, l.Instance.Config)
+}
+
+func validateConfig(_ context.Context, cfg *limatype.LimaYAML) error {
+	if cfg == nil {
+		return errors.New("configuration is nil")
+	}
+	if err := validateMountType(cfg); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// Helper method for mount type validation.
+func validateMountType(cfg *limatype.LimaYAML) error {
+	if cfg.MountTypesUnsupported != nil && cfg.MountType != nil && slices.Contains(cfg.MountTypesUnsupported, *cfg.MountType) {
+		return fmt.Errorf("mount type %q is explicitly unsupported", *cfg.MountType)
+	}
+
+	return nil
+}
+
+func (l *LimaExtDriver) FillConfig(_ context.Context, cfg *limatype.LimaYAML, _ string) error {
+	if cfg.VMType == nil {
+		cfg.VMType = ptr.Of(limatype.EXT)
+	}
+
+	if cfg.MountType == nil || *cfg.MountType == "" || *cfg.MountType == "default" {
+		cfg.MountType = ptr.Of(limatype.REVSSHFS)
+	}
+
+	mountTypesUnsupported := make(map[string]struct{})
+	mountTypesUnsupported[limatype.NINEP] = struct{}{}
+	mountTypesUnsupported[limatype.VIRTIOFS] = struct{}{}
+
+	if _, ok := mountTypesUnsupported[*cfg.MountType]; ok {
+		return fmt.Errorf("mount type %q is explicitly unsupported", *cfg.MountType)
+	}
+
+	return nil
+}
+
+func (l *LimaExtDriver) Start(_ context.Context) (chan error, error) {
+	errCh := make(chan error)
+
+	return errCh, nil
+}
+
+func (l *LimaExtDriver) BootScripts() (map[string][]byte, error) {
+	return nil, nil
+}
+
+func (l *LimaExtDriver) RunGUI() error {
+	return fmt.Errorf("RunGUI is not supported for the given driver '%s' and display '%s'", "ext", *l.Instance.Config.Video.Display)
+}
+
+func (l *LimaExtDriver) Stop(_ context.Context) error {
+	return nil
+}
+
+func (l *LimaExtDriver) GuestAgentConn(_ context.Context) (net.Conn, string, error) {
+	return nil, "", nil
+}
+
+func (l *LimaExtDriver) Info() driver.Info {
+	var info driver.Info
+	info.Name = "ext"
+	if l.Instance != nil && l.Instance.Dir != "" {
+		info.InstanceDir = l.Instance.Dir
+	}
+
+	info.Features = driver.DriverFeatures{
+		DynamicSSHAddress:    false,
+		SkipSocketForwarding: false,
+		CanRunGUI:            false,
+	}
+	return info
+}
+
+func (l *LimaExtDriver) SSHAddress(_ context.Context) (string, error) {
+	return l.Instance.SSHAddress, nil
+}
+
+func (l *LimaExtDriver) InspectStatus(_ context.Context, _ *limatype.Instance) string {
+	return ""
+}
+
+func (l *LimaExtDriver) Create(_ context.Context) error {
+	return nil
+}
+
+func (l *LimaExtDriver) Delete(_ context.Context) error {
+	return nil
+}
+
+func (l *LimaExtDriver) CreateDisk(_ context.Context) error {
+	return nil
+}
+
+func (l *LimaExtDriver) Register(_ context.Context) error {
+	return nil
+}
+
+func (l *LimaExtDriver) Unregister(_ context.Context) error {
+	return nil
+}
+
+func (l *LimaExtDriver) ChangeDisplayPassword(_ context.Context, _ string) error {
+	return nil
+}
+
+func (l *LimaExtDriver) DisplayConnection(_ context.Context) (string, error) {
+	return "", nil
+}
+
+func (l *LimaExtDriver) CreateSnapshot(_ context.Context, _ string) error {
+	return errUnimplemented
+}
+
+func (l *LimaExtDriver) ApplySnapshot(_ context.Context, _ string) error {
+	return errUnimplemented
+}
+
+func (l *LimaExtDriver) DeleteSnapshot(_ context.Context, _ string) error {
+	return errUnimplemented
+}
+
+func (l *LimaExtDriver) ListSnapshots(_ context.Context) (string, error) {
+	return "", errUnimplemented
+}
+
+func (l *LimaExtDriver) ForwardGuestAgent() bool {
+	// If driver is not providing, use host agent
+	return true
+}
+
+func (l *LimaExtDriver) AdditionalSetupForSSH(_ context.Context) error {
+	return nil
+}
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -176,8 +176,10 @@ func New(ctx context.Context, instName string, stdout io.Writer, signalCh chan o
 	if err := cidata.GenerateCloudConfig(ctx, inst.Dir, instName, inst.Config); err != nil {
 		return nil, err
 	}
-	if err := cidata.GenerateISO9660(ctx, limaDriver, inst.Dir, instName, inst.Config, udpDNSLocalPort, tcpDNSLocalPort, o.guestAgentBinary, o.nerdctlArchive, vSockPort, virtioPort, noCloudInit, rosettaEnabled, rosettaBinFmt); err != nil {
-		return nil, err
+	if *inst.Config.VMType != limatype.EXT {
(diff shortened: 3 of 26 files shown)


afbjorklund:
The probes are somewhat annoying when not using cidata, but that was the same story on FreeBSD and others.

Maybe there should be some fallback implementation, at least for ssh-ready/guestagent install/boot-done ?

```
sudo diff -q /run/lima-ssh-ready /mnt/lima-cidata/meta-data

install -m 755 /mnt/lima-cidata/lima-guestagent /usr/local/bin/lima-guestagent
sudo /usr/local/bin/lima-guestagent install-systemd

sudo diff -q /run/lima-boot-done /mnt/lima-cidata/meta-data
```

Like copying the lima-guestagent over the ssh connection (my workaround)

afbjorklund:
Added support for hostnames, so that you can use nice features like `avahi-daemon`

```
NAME           STATUS     SSH                     VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
raspberrypi    Running    raspberrypi.local:22    ext       aarch64    4       512MiB    32GiB     ~/.lima/raspberrypi
```

Previously I was assuming IP. Now, to have a nice way to install the `lima-guestagent`

afbjorklund:
Added host key checking:

```
The authenticity of host 'raspberrypi.local (192.168.0.113)' can't be established.
ECDSA key fingerprint is SHA256:tKIRfOeWP1HeCFLpM0UT30CUWSDXpC7gxPsKHUnS+h4.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
```

afbjorklund:
Added provision scripts.

Using `sudo` for system.

pkg/sshutil/sshutil.go
```diff
@@ -231,7 +237,7 @@ func SSHOpts(instDir string, useDotSSH, forwardAgent, forwardX11, forwardX11Trus
 	if err != nil {
 		return nil, err
 	}
-	opts, err := CommonOpts(useDotSSH)
+	opts, err := CommonOpts(useDotSSH, hostAddress == "127.0.0.1")
```
AkihiroSuda, line 240: What about `::1`?

pkg/limayaml/limayaml.go
```diff
@@ -66,6 +66,7 @@ const (
 	QEMU VMType = "qemu"
 	VZ   VMType = "vz"
 	WSL2 VMType = "wsl2"
+	EXT  VMType = "ext"
```
AkihiroSuda, line 69: Maybe this should be called "external" to avoid confusion with "extended"

examples/default.yaml
```diff
@@ -117,6 +117,9 @@ additionalDisks:
 #   fsType: "ext4"
 
 ssh:
+  # Address for the host.
+  # 🟢 Builtin default: "127.0.0.1" (localhost)
+  address: null
```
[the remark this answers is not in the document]
afbjorklund: It used as "port", but the old name is kept for compatibility.

If the port is zero, then for non-localhost it will be set to 22.
For localhost (127.0.0.1), it is still assigned to a random port.

Normally it is **not** set, and only the IP address is used for SSH.

pkg/sshutil/sshutil.go
afbjorklund: It was only meant to handle the default string, not any overrides like "localhost" or "::1" (ipv6)

There is probably a utility `net` function that one can call, instead of hardcoding strings like this.

pkg/limayaml/limayaml.go
afbjorklund: In minikube we ended up calling the driver "ssh", but I don't think it's great

afbjorklund:
Rebased to lima v0.21.0

```console
$ _output/bin/limactl shell raspberrypi nerdctl version
Client:
 Version:	v1.7.5
 OS/Arch:	linux/arm64
 Git commit:	cffed372371dcbea3dc9a646ce5a913fc1c09513
 buildctl:
  Version:	v0.12.5
  GitCommit:	bac3f2b673f3f9d33e79046008e7a38e856b3dc6

Server:
 containerd:
  Version:	v1.7.14
  GitCommit:	dcf2847247e18caba8dce86522029642f60fe96b
 runc:
  Version:	1.1.12
  GitCommit:	v1.1.12-0-g51d5e946

```

AkihiroSuda:
Would it be possible to test this on CI?

afbjorklund:
> Would it be possible to test this on CI?

As long as it is possible to supply a VM, with access through host keys and authorized keys, that should be possible.

I should detail the required steps (with example log), especially now with the addition of the cloud-config generation.

pkg/hostagent/hostagent.go
```diff
@@ -221,6 +223,9 @@ func determineSSHLocalPort(y *limayaml.LimaYAML, instName string) (int, error) {
 	if *y.SSH.LocalPort < 0 {
 		return 0, fmt.Errorf("invalid ssh local port %d", y.SSH.LocalPort)
 	}
+	if *y.SSH.LocalPort == 0 && *y.SSH.Address != "127.0.0.1" {
```
AkihiroSuda, line 226: https://pkg.go.dev/net#IP.IsLoopback should be used

pkg/limayaml/limayaml.go
AkihiroSuda: ping @lima-vm/maintainers @lima-vm/reviewers RFC

I still feel "ext" is confusing.
Sounds like some sort of ext2/ext3/ext4 stuff.

pkg/sshutil/sshutil.go
AkihiroSuda: https://pkg.go.dev/net#IP.IsLoopback should be used

pkg/limayaml/limayaml.go
```diff
@@ -69,6 +69,7 @@ const (
...
```
AkihiroSuda, line 72: Needs docs and integration tests.
Probably this should be also marked experimental

AkihiroSuda: Probably Incus (or LXD) can be used for emulating an "external" host

pkg/hostagent/hostagent.go
afbjorklund: Here we are only checking if the YAML has been set, to something other than the default value.

pkg/limayaml/limayaml.go
afbjorklund: Experimenting with LXC/LXD, after all we have it in the default install...

afbjorklund:
Probably not a _great_ idea to have the validation fail, when the host is offline (or not available)

` field `ssh.address` must be IP: lookup raspberrypi.local: i/o timeout"`

pkg/limayaml/limayaml.go
jandubois: I think `external` is fine. `unmanaged` or `raw` could be alternatives, but I prefer `external`.

afbjorklund:
LXD timed out in 30 minutes, not sure if it would work to run the "VM" with it...

https://documentation.ubuntu.com/lxd/en/latest/tutorial/first_steps/

afbjorklund:
Maybe this is a better way to do it, i.e. using GitHub Docker instead of Canonical LXD:

https://docs.github.com/en/actions/using-containerized-services/about-service-containers

Still needs a custom image, with sshd?

And setting up keys, possibly cloud-init

AkihiroSuda:
Maybe we can just use `limactl create --plane` with QEMU and re-register the instance with the generic driver

afbjorklund:
> Maybe we can just use `limactl create --plane` with QEMU and re-register the instance with the generic driver

I removed all the old LXD code, and there are still concerns about testing containerd and reverse-sshfs with this...

i.e. when trying to run a fake machine in a container, that leads to problems not happening on "real" machines

Whether those are actually real (hardware) machines, or if they are virtual machines emulating real hardware.

pkg/limayaml/limayaml.go
afbjorklund: Removed the integration tests, and the documentation does not have details.

pkg/limayaml/limayaml.go
AkihiroSuda: ping @afbjorklund WDYT?

AkihiroSuda:
Thanks for working on this, but I still think "ext"is confusing and should be called like "external" or something.
This also needs an integration test.

afbjorklund:
I will give it some more thought, what the new name of the "generic" driver should be (naming things is hard...)

Integration tests with (non-hosted) GitHub Actions seems to be very limited? Tried faking it with LXD, but it failed.

afbjorklund:
> I will give it some more thought, what the new name of the "generic" driver should be (naming things is hard...)

The new name of the "external" driver will be the "static" driver, since it does static allocation rather than dynamic.

afbjorklund:
Obviously this driver will need a new name, now that it is "internal" in the new framework...

afbjorklund:
Rebased to v2.0.0-alpha.0, only thing left to do is to find a driver name that is acceptable.

afbjorklund:
These patches should be converted to driver features:
* warn about experimental driver
* skip creating the "basedisk" file
* skip creating the "cidata.iso" file
* run provisioning scripts over ssh
* skip running certain requirements
* allow instance without "images"
* validate ssh.address (and required)
* allow PID status check for HA only

website/content/en/docs/config/vmtype/ext.md
```diff
@@ -0,0 +1,11 @@
+---
+title: EXT
+weight: 3
+---
+
+> **Warning**
+> "ext" mode is experimental
+
+"ext" option makes use of an external machine, either a virtual machine or a physical machine.
```
AkihiroSuda, line 9: The idea of "external machines" can be easily confused with "external drivers", so this driver should not be called "ext" ("external").

Let's call this either "generic" or "ssh"

templates/experimental/raspberrypi.yaml
```diff
@@ -0,0 +1,14 @@
...
+memory: 512MiB
+disk: 32GiB
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+
+ssh:
+  address: raspberrypi.local
```
AkihiroSuda, line 14: If you want to put raspberrypi-specific config here, the file name should be probably "raspberrypi.yaml"

cmd/limactl/main_ext.go
```diff
@@ -0,0 +1,7 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+// Import ext driver to register it in the registry on all platforms.
+import _ "github.com/lima-vm/lima/v2/pkg/driver/ext"
```
AkihiroSuda, line 7: Should this be compiled as an external driver? (w/ docs)

pkg/driver/ext/ext_driver.go
```diff
@@ -0,0 +1,172 @@
...
+		cfg.MountType = ptr.Of(limatype.REVSSHFS)
+	}
+
+	mountTypesUnsupported := make(map[string]struct{})
+	mountTypesUnsupported[limatype.NINEP] = struct{}{}
+	mountTypesUnsupported[limatype.VIRTIOFS] = struct{}{}
+
+	if _, ok := mountTypesUnsupported[*cfg.MountType]; ok {
+		return fmt.Errorf("mount type %q is explicitly unsupported", *cfg.MountType)
+	}
```
AkihiroSuda, line 75: Any mount type other than REVSSHFS should result in an error

cmd/limactl/main_ext.go
AkihiroSuda, line 1: Can we have a smoke test ?
Maybe using an LXD/Incus instance on GHA

afbjorklund: I tried that, but it didn't work out... Help needed

https://github.com/lima-vm/lima/pull/2000#issuecomment-2323265120

cmd/limactl/main_ext.go
afbjorklund: Currently that is not possible, but might be an option (like the other built-in ones)

afbjorklund:
I think we will throw this one out of 2.0 as well (like with 1.0). Maybe it will make a comeback in 3.0

cmd/limactl/main_ext.go
AkihiroSuda: What was the error?

cmd/limactl/main_ext.go
AkihiroSuda: Why not possible?

cmd/limactl/main_ext.go
AkihiroSuda: https://github.com/canonical/setup-lxd may work

I guess there is an equivalent of this for Incus too

cmd/limactl/main_ext.go
afbjorklund: https://github.com/lima-vm/lima/pull/2000#issuecomment-3262065114

website/content/en/docs/config/vmtype/ext.md
afbjorklund: I tried the name "ssh", but it looks weird in things like limatype.SSH and so on. So it is not a favorite.

AkihiroSuda:

afbjorklund:
It was a candidate for a hackathon project in Nov, but there were no takers. I'll put it back on the shelf for now.
There seemed to be _slightly_ more interest in the other lima drivers, for cloud and container, but that was it?

I'll keep the Raspberry Pi Kit "prize" for the spring, maybe there will work better then (and I have more time)
It does work equally well with a bigger computer (only needs ssh), but harder to carry around in the backpack :-)

PR #2006, opened 2023-11-17
Restore guest agent unix socket functionality

afbjorklund:
Make both vsock and virtio explicit, instead of hardcoding internal filenames in the agents. Fallback to to the unix.

* #1998

If neither vsock nor virtio is being used, then the driver silently falls back to forwarding the unix socket over ssh.

This means that new drivers don't _have_ to do anything, unless overriding.


balajiv113:
@afbjorklund Thanks for addressing this usecase 

How about we expose a util to forward unix sock ?? Drivers lile vbox can use this and follow the same mechanism of dial unix sock

In guest side, we will have last default as unix sock without any additional options.

This way, the guest communication will fully be taken care by drivers only and host will not hardcode any functionality for guest connection

afbjorklund:
> How about we expose a util to forward unix sock ?? Drivers lile vbox can use this and follow the same mechanism of dial unix sock

It is there now, but you don't have to call it explicitly - it is just the fallback function if you _don't_ implement the callback...

```go
        GuestAgentConn(_ context.Context) (net.Conn, error)
```

This means you can use the base driver, without getting an exception. Just like it was, before the new features were added.

```diff
+func (d *BaseDriver) ForwardUnixGuestAgent() bool {
+       return d.VSockPort == 0 && d.VirtioPort == ""
+}
+
 func (d *BaseDriver) GuestAgentConn(_ context.Context) (net.Conn, error) {
-       return nil, fmt.Errorf("unimplemented")
+       return nil, nil
 }
```

balajiv113:
Yes i agree on bringing the forwarded sock. But decision should be done from each driver and not on host agent.

The reason is, we are using ForwardUnixGuestAgent just for comparability reasons. Which doesn't give much value for a driver interface and also because of this hostagent has to take care of forwarding unix socket

The advantage of driver deciding is the separation of guest agent connection from host agent and keeping the driver interface clean of other. 

I agree some changes needed on vbox for this but it will be very minimal. We could do that just to keep the driver interface clear

afbjorklund:
I agree, keeping it in the interface was more of a workaround (due to vsock/virtio ports and conn being exposed already)

We _could_ add more callbacks (in the driver) to set up and cancel the sockets, and do the implementation in the basedriver.

afbjorklund:
Kept the implementation in the hostagent (since it was there before, and had all the functions needed)

We could move it (3 functions) to the BaseDriver with some effort, but I am not sure if it is worth spending

afbjorklund:
Ah, it doesn't inherit BaseDriver but just keeps a pointer to it. Oops.

afbjorklund:
Added some documentation about the previous/fallback functionality, best I could come up with.

cmd/lima-guestagent/daemon_linux.go
```diff
@@ -72,10 +68,14 @@ func daemonAction(cmd *cobra.Command, _ []string) error {
...
 	srv := &http.Server{Handler: r}
+	err = os.RemoveAll(socket)
+	if err != nil {
+		return err
+	}
 
 	var l net.Listener
-	if _, err := os.Stat(virtioPort); err == nil {
-		qemuL, err := serialport.Listen(virtioPort)
+	if virtioPort != "" {
```
balajiv113, line 77: Instead of using virtioPort as a arg, can we simply do os.Stat and fallback to unix sock ?

pkg/driver/driver.go
```diff
@@ -137,6 +142,12 @@ func (d *BaseDriver) ListSnapshots(_ context.Context) (string, error) {
 	return "", fmt.Errorf("unimplemented")
 }
 
+func (d *BaseDriver) ForwardGuestAgent() bool {
```
balajiv113, line 145: My concern is still here. 

My point is we don't need this. Instead of treating unix sock as fallback am saying lets treat it as a way drivers can use. But drivers need to decide on the communication medium. 

For eg: vz will always work on vsock, virtbox will always work on unix sock (This individual drivers should decide). Hostagent should not have that code

afbjorklund: I wanted to **keep** the hostagent tunneling as a fallback. It was handy when experimenting with external VMs.

For the VBox driver, it can probably use a serial console. Will do a rebase of it someday, it is not important...

cmd/lima-guestagent/daemon_linux.go
afbjorklund: Having it as an explicit argument was a feature, but it should probably be named serialSock

pkg/driver/driver.go
balajiv113: When we say fallback i wanted this to be a universal one not just for un configured driver.

Something like this  https://github.com/lima-vm/lima/commit/4640684a0e55ce0b72afce87133343ffbe3a7830. This is just a idea how we can do, based on the error we can fallback to this universal unix sock

cmd/lima-guestagent/daemon_linux.go
balajiv113: The communication medium is internal to us. I don't think there is a need to make it a argument. We can hardcode it like we do for unix sock just to simplify stuff

pkg/driver/driver.go
afbjorklund: As long as it is done in the hostagent, and not copy/paste (or explicit) in the driver, that would work as well

cmd/lima-guestagent/daemon_linux.go
afbjorklund: If the filename is indeed fixed we can hardcode it, but then it might not even need the constant

AkihiroSuda:
Needs rebase

afbjorklund:
Needs rewrite, without the changes to the interface (as suggested)

afbjorklund:
Rebase done.

pkg/driver/driver.go
afbjorklund: Not sure where to cancel the forwarding process, but maybe it would just die when the instance does?

AkihiroSuda:
CI is failing

AkihiroSuda (approved):
Thanks

PR #2015, opened 2023-11-20
hostagent: increase nofile limit

vasileknik76:
Default nofile limit in the standard terminal in macOS is 256. You can see it with `ulimit -n`.

So when I tried to open over 240 connections to 127.0.0.1, the ssh process crashed. 

Default limit is too low for a vm that can run many applications. 

I suggest to use the same approach as in golang - setting an increased limit for the lima needs.

PR relates on all types of networks, but only for the macOS.

I understand that users can increase the limit themselves using `ulimit -n <N>`. But the regular user expects to get a working environment without exploring such low-level features.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -288,6 +288,7 @@ func (a *HostAgent) Run(ctx context.Context) error {
 		}
 		a.emitEvent(ctx, exitingEv)
 	}()
+	adjustNofileRlimit()
 
 	firstUsernetIndex := limayaml.FirstUsernetIndex(a.y)
 	if firstUsernetIndex == -1 && *a.y.HostResolver.Enabled {
--- a/pkg/hostagent/hostagent_unix.go
+++ b/pkg/hostagent/hostagent_unix.go
@@ -1,0 +1,28 @@
+//go:build !windows
+
+package hostagent
+
+import (
+	"syscall"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Default nofile limit is too low on some system.
+// For example in the macOS standard terminal is 256.
+// It means that there are only ~240 connections available from the host to the vm.
+// That function increases the nofile limit for child processes, especially the ssh process
+//
+// More about limits in go: https://go.dev/issue/46279
+func adjustNofileRlimit() {
+	var limit syscall.Rlimit
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &limit); err != nil {
+		logrus.WithError(err).Debug("failed to get RLIMIT_NOFILE")
+	} else if limit.Cur != limit.Max {
+		limit.Cur = limit.Max
+		err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &limit)
+		if err != nil {
+			logrus.WithError(err).Debugf("failed to set RLIMIT_NOFILE (%+v)", limit)
+		}
+	}
+}
--- a/pkg/hostagent/hostagent_windows.go
+++ b/pkg/hostagent/hostagent_windows.go
@@ -1,0 +1,5 @@
+//go:build windows
+
+package hostagent
+
+func adjustNofileRlimit() {}


vasileknik76:
Go 1.19+ increase the limit only for itself. But in [exec](https://github.com/golang/go/blob/go1.21.4/src/syscall/exec_unix.go#L281-L284) limit returns to original value. It means that any `(*exec.Cmd).Run` will be runned with the old limit.

An explicit call `syscall.Setrlimit` modifies atomic [origRlimitNofile](https://github.com/golang/go/blob/go1.21.4/src/syscall/rlimit.go#L47) and the above condition does not affect the new exec's

jandubois:
> Go 1.19+ increase the limit only for itself.

It does? That seems to contradict the 1.19 release notes, which state:

> One impact of this change is that Go programs that in turn execute very old C programs in child processes may run those programs with too high a limit.

So if this change is indeed still necessary, then we obviously should raise the limit ourselves too. I'm generally prejudiced against using `init` functions for this because it makes error reporting/logging harder, but maybe we can assume that this call will never fail... 😄

vasileknik76:
This is actual for both Go 1.19 and Go 1.21.

A simple program for demonstration of limits:
```
package main

import (
	"fmt"
	"log"
	"os/exec"
	"syscall"
)

func printLimit() {
	cmd := exec.Command("ulimit", "-n")
	if out, err := cmd.CombinedOutput(); err != nil {
		log.Fatal(err)
	} else {
		fmt.Println(string(out))
	}
}

func main() {
	var limit syscall.Rlimit
	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &limit); err == nil && limit.Cur != limit.Max {
		limit.Cur = limit.Max

		fmt.Println("ulimit before setrlimit")
		printLimit()
		_ = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &limit)
		fmt.Println("ulimit after setrlimit")
		printLimit()

	}
}
```

On my M1 with go 1.19 it prints:
```
ulimit before setrlimit
256

ulimit after setrlimit
unlimited
```

In the worst case, during initialization there will be no increase in limits and everything will remain as before.

I have already applied the same patch to our Lima-based devenv and it showed good results. It seems that Lima will also benefit from this.

jandubois (dismissed):
Thanks, LGTM

jandubois:
Should this also be applied on Linux then? I realize the default will be higher, but is there any reason not to raise to the max limit there as well?

vasileknik76:
> Should this also be applied on Linux then? I realize the default will be higher, but is there any reason not to raise to the max limit there as well?

I think this can be applied to Linux as well. But different Linux distros may have different defaults. And since I'm not sure about Linux, I decided to make a patch for darwin first.

BTW: thanks for the quick review.

pkg/hostagent/port_darwin.go
```diff
@@ -8,13 +8,27 @@ import (
...
+// Default nofile limit in the standard terminal in macOS is 256.
+// It means that there are only ~240 connections available from the host to the vm.
+// That function increases the nofile limit for child processes, especially the ssh process
+//
+// More about limits in go: https://go.dev/issue/46279
+func init() {
+	var limit syscall.Rlimit
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &limit); err == nil && limit.Cur != limit.Max {
+		limit.Cur = limit.Max
+		_ = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &limit)
```
AkihiroSuda, line 28: Maybe the error should be logged via `logrus.Debug`?

AkihiroSuda: I also wonder this should be moved to https://github.com/lima-vm/lima/blob/master/cmd/limactl/hostagent.go , as it affects reverse-sshfs too IIUC.

AkihiroSuda: (It covers non-Darwin too)

vasileknik76: Yes, it affects reverse-sshfs too.

I moved the code to `hostagent.go` and make the fuction comment more wide.
Also I added debug logging if an error occurs

pkg/hostagent/hostagent.go
```diff
@@ -65,6 +66,25 @@ type options struct {
 
 type Opt func(*options) error
 
+// Default nofile limit is too low on some system.
+// For example in the macOS standard terminal is 256.
+// It means that there are only ~240 connections available from the host to the vm.
+// That function increases the nofile limit for child processes, especially the ssh process
+//
+// More about limits in go: https://go.dev/issue/46279
+func init() {
```
AkihiroSuda, line 75: Not a huge fun of `init()`.

This can be just a regular function and called from `cmd/limactl/hostagent.go`

pkg/hostagent/hostagent.go
```diff
@@ -65,6 +66,25 @@ type options struct {
...
+	var limit syscall.Rlimit
```
AkihiroSuda, line 76: CI is failing for Windows, so this should be moved `*_unix.go`

pkg/hostagent/hostagent.go
```diff
@@ -65,6 +66,25 @@ type options struct {
...
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &limit); err != nil {
+		logrus.Debugf("failed to get rlimit: %v", err)
+	} else if limit.Cur != limit.Max {
+		limit.Cur = limit.Max
+		err = syscall.Setrlimit(syscall.RLIMIT_NOFILE, &limit)
+		if err != nil {
+			logrus.Debugf("failed to set rlimit: %v", err)
```
AkihiroSuda, line 83:
```suggestion
			logrus.WithError(err).Debugf("failed to set RLIMIT_NOFILE (%+v)", limit)
```

AkihiroSuda (approved):
Thanks

pkg/hostagent/hostagent_windows.go
```diff
@@ -0,0 +1,5 @@
+//go:build windows
```
AkihiroSuda, line 1: This line is not needed

pkg/hostagent/hostagent_unix.go
```diff
@@ -0,0 +1,28 @@
+//go:build !windows
```
AkihiroSuda, line 1: This line is probably not needed for recent Go

vasileknik76: I tried building with go 1.21 and got the error `adjustNofileRlimit redeclared in this block`, so I added a limit

pkg/hostagent/hostagent_windows.go
vasileknik76: Without the constraint i got an error:
```
$ GOOS=windows go build ./cmd/limactl
# github.com/lima-vm/lima/pkg/hostagent
pkg/hostagent/hostagent_windows.go:3:6: adjustNofileRlimit redeclared in this block
        pkg/hostagent/hostagent_unix.go:15:6: other declaration of adjustNofileRlimit
pkg/hostagent/hostagent_unix.go:16:20: undefined: syscall.Rlimit
pkg/hostagent/hostagent_unix.go:17:20: undefined: syscall.Getrlimit
pkg/hostagent/hostagent_unix.go:17:38: undefined: syscall.RLIMIT_NOFILE
pkg/hostagent/hostagent_unix.go:21:17: undefined: syscall.Setrlimit
pkg/hostagent/hostagent_unix.go:21:35: undefined: syscall.RLIMIT_NOFILE

```

pkg/hostagent/hostagent_unix.go
omnilight: I think, there should be

```
//go:build unix
```

because file suffix `_unix` is not working in Go:
https://github.com/golang/go/issues/51572

still, for file `hostagent_windows` build tag is not required, as suffix `_windows` works well

PR #2019, opened 2023-11-22
website: use svg for colima logo

afbjorklund:
Instead of the rendered png

`colima-logo-horizontal-2.svg`

From the original colima PR

* https://github.com/abiosoft/colima/issues/781#issuecomment-1768532407


AkihiroSuda (approved):
Thanks

PR #2023, opened 2023-11-23
Fix for default network for user-v2

### LINKED ISSUE #1933: net-user-v2 example does not work (`invalid CIDR address: <nil>/0`)
### Description

I was trying network v2 today to see if helped with some dns issues we have been getting reports of from our users, but the example does not work. Do you have an guidance on how to resolve this error?

```
❯ limactl --version
limactl version 0.18.0
```

```
❯ limactl start https://raw.githubusercontent.com/lima-vm/lima/master/examples/experimental/net-user-v2.yaml
? Creating an instance "net-user-v2" Proceed with the current configuration
FATA[0001] failed to start usernet "user-v2": invalid CIDR address: <nil>/0
```

balajiv113:
Fixes #1933

### DIFF
--- a/pkg/networks/config.go
+++ b/pkg/networks/config.go
@@ -57,7 +57,7 @@ func fillDefaults(nwYaml YAML) (YAML, error) {
 		nwYaml.Networks = make(map[string]Network)
 	}
 	for nw := range nwYaml.Networks {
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #2024, opened 2023-11-23
templates: Update Rocky Linux to 8.9 and 9.3

tcooper:
Update Rocky Linux templates to point at the recently updated images and provides fallback to unversioned image which should always be the most recent image available.


AkihiroSuda (approved):
Thanks

Issue #1957, opened 2023-10-27
VM disk corruption with Apple Silicon


EdwardMoyse:
> [!TIP]
>
> **EDIT** by @AkihiroSuda 
>
>  For `--vm-type=vz`, this issue seems to have been solved in Lima v0.19 (https://github.com/lima-vm/lima/pull/2026)

 - - -
### Description

**Lima version**: 0.18.0
**macOS**: 14.0 (23A344)
**VM**: Almalinux9

I was trying to do a big compile, using a VM with the attached configuration (vz)

```
NAME           STATUS     SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
myalma9        Running    127.0.0.1:49434    vz        aarch64    4       16GiB     100GiB    ~/.lima/myalma9
```

The build aborted with:
```
from /Volumes/Lima/build/build/AthenaExternals/src/Geant4/source/processes/hadronic/models/lend/src/[xDataTOM_LegendreSeries.cc](http://xdatatom_legendreseries.cc/):7:
/usr/include/bits/types.h:142:10: fatal error: /usr/include/bits/time64.h: Input/output error
```

And afterwards, even in a different terminal, I see:
```
[emoyse@lima-myalma9 emoyse]$ ls
bash: /usr/bin/ls: Input/output error
```

I was also logged into a display, and there I saw e.g. 

<img width="981" alt="Screenshot 2023-10-26 at 17 44 45" src="https://github.com/lima-vm/lima/assets/6764617/f8ca7340-5690-4c42-ac55-8a245353b4f0">

If I try to log in again with:
```
limactl shell myalma9
```

each time  I see something like the following appear in the display window:
```
[56247.6427031] Core dump to l/usr/lib/systemd/systemd-coredump pipe failed
```

Edit: there has been a lot of discussion below, and the corruption can happen with both `vz` and `qemu`, and on external (to the VM) and internal disks. Some permutations seem more likely to provoke a corruption than others. I have reproduced my experiments in the table in the following [comment](https://github.com/lima-vm/lima/issues/1957#issuecomment-1812631707) below.

EdwardMoyse:
[...]

PR #2026, opened 2023-11-24
vz: use DiskImageCachingModeCached (rumored to fix disk corruption on ARM)

AkihiroSuda:
Switch away from `DiskImageCachingModeAutomatic` to `DiskImageCachingModeCached`, as this is rumored to fix disk corruption on ARM
- https://github.com/utmapp/UTM/issues/4840#issuecomment-1824340975
- https://github.com/utmapp/UTM/issues/4840#issuecomment-1824542732

Expected to fix issue #1957 (for vz)

### DIFF
--- a/pkg/vz/vm_darwin.go
+++ b/pkg/vz/vm_darwin.go
@@ -30,6 +30,13 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// diskImageCachingMode is set to DiskImageCachingModeCached so as to avoid disk corruption on ARM:
+// - https://github.com/utmapp/UTM/issues/4840#issuecomment-1824340975
+// - https://github.com/utmapp/UTM/issues/4840#issuecomment-1824542732
+//
+// Eventually we may bring this back to DiskImageCachingModeAutomatic when the corruption issue is properly fixed.
+const diskImageCachingMode = vz.DiskImageCachingModeCached
+
 type virtualMachineWrapper struct {
 	*vz.VirtualMachine
 	mu      sync.Mutex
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Merging, for ease of testing

hasan4791:
Seems to be working good so far. Will monitor for a week atleast. 👍🏽

rpkoller:
I've got notice of this issue in there https://github.com/lima-vm/lima/issues/1957 . I am using Colima and since the release of Colima 0.6.0 i ran into VMs that got corrupted all of a sudden. At first my suspicion was the root cause was in the changes for 0.6.0 but yesterday i'Ve started to test in plain Lima and today I ran into the same issue. After finding this issue i've updated to HEAD as well. I am already testing for a few hours now and no issues so far. Will also keep testing the coming days. But this looks promising so far. Cuz under Lima the error sometimes happened after starting a single project in a completely newly generated instance. But since the upgrade to head it looks good. I am using the template "Docker". Thank you for spotting this!

EdwardMoyse:
As noted on #1957 this seems to be helping me - no corruptions w/ vz VM since I moved to HEAD (and it was happening hourly before).

balajiv113:
@AkihiroSuda I believe this change will have disk io performance issues. 

We will see a reduced disk io performance

balajiv113:
Should we think of providing a option to switch over to older model for those who prefers performance over stability ?

AkihiroSuda:
I guess nobody wants disk corruption?

balajiv113:
If am not wrong this corruption is happening only on ARM. This change will now degrade performance for intel users as well.

balajiv113:
We can go ahead. If someone complaints on performance we can re-evaluate at that point.

balajiv113:
@rpkoller Thanks for verifying this.

We should be good in that case 👍

PR #2032, opened 2023-11-25
Support for Virt driver

afbjorklund:
This is a driver based on [libvirt.org,](https://libvirt.org) for remote hypervisors.

The libvirt functions are in a plugin, to minimize dependencies.

----

Currently only shows version:

```go
var VERSION uint32 = libvirt.VERSION_NUMBER

func Version() (uint32, error) { return libvirt.GetVersion() }
```

To prove that it can access the API.


Kconfig
```diff
@@ -30,3 +30,9 @@ config GUESTAGENT_ARCH_RISCV64
 	help
 		Build lima-guestagent for "riscv64" Arch
 	default y
+
+config PLUGIN_VIRT
+	bool "plugin for vmType virt"
```
AkihiroSuda, line 35: I think this should be called "libvirt"

AkihiroSuda: I also wonder if we can split out this to a separate plugin binary.

Same applies to the vbox driver, cloud drivers, etc.

afbjorklund: Sure, it just requires more work. And it would be needed, for cloud.
* https://github.com/lima-vm/lima/discussions/2007

afbjorklund:
Apparently this experiment does not work, since putting libvirt in go.mod requires libvirt-dev:

```
 No package 'libvirt' found
  Package libvirt was not found in the pkg-config search path.
  Perhaps you should add the directory containing `libvirt.pc'
  to the PKG_CONFIG_PATH environment variable
```

afbjorklund:
1) Running "session" qemu with libvirt doesn't add any except complexity (until there is proper "system" qemu support)
2) Trying to load libvirt from a plugin doesn't work due to go modules, needs a separate binary like docker-machine-kvm

afbjorklund:
All that you needed to do was to set the build tag `libvirt_dlopen`, then it doesn't need pkgconfig or the libraries...

afbjorklund:
Hmm, libvirt has dozens of connection points (function wrappers) so maybe it should be a plugin after all.

AkihiroSuda:
Would it be possible to just exec `virsh`?

afbjorklund:
Yes, that should work. Something like `virsh create {{.Dir}}/domain.xml`

PR #2033, opened 2023-11-26
Update /mnt/data/etc/network/interfaces on each boot

jandubois:
Otherwise it always reverts back to the first-boot settings once 04-persistent-data-volume.sh is done, and any further restarts of the networking service will revert to the old interfaces.

This problem was exposed by #1898. There are probably additional settings that might change `/etc` files on a reboot and need to be persisted, but I wanted to keep this PR small.

Similarly `/etc/resolv.conf` is generated by `udhcpc` (via `ifup`), but reverts back to the first-boot version after the bind-mount of the persistent data volume.

Restarting the networking service was just an expensive and indirect way to update `/etc/resolv.conf` by re-running `udhcpc` again.

Fixes https://github.com/rancher-sandbox/rancher-desktop/issues/6051

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
@@ -27,6 +27,10 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
 		fi
 		# Mount data volume
 		mount -t ext4 /dev/disk/by-label/data-volume /mnt/data
+		# Update /etc files that might have changed during this boot
+		cp /etc/network/interfaces /mnt/data/etc/network/
+		cp /etc/resolv.conf /mnt/data/etc/
+		# TODO there are probably others that should be updated as well
 	else
 		# Find an unpartitioned disk and create data-volume
 		DISKS=$(lsblk --list --noheadings --output name,type | awk '$2 == "disk" {print $1}')
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/50-alpine-fix.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/50-alpine-fix.sh
@@ -1,8 +1,0 @@
-#!/bin/sh
-set -eux
-
-# This script prepares Alpine for lima; there is nothing in here for other distros
-test -f /etc/alpine-release || exit 0
-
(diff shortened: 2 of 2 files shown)


AkihiroSuda (approved):
Thanks

balajiv113:
@jandubois 
The issue that i noticed was related to alpine within lima as well.

When the dns ip got updated (with vz) in Alpine it was not working. The core issue that i noticed was dns ip and all was updated correctly but the resolver didn't take the latest value. That's why included a restart to pick up the latest value

jandubois:
> When the dns ip got updated (with vz) in Alpine it was not working. The core issue that i noticed was dns ip and all was updated correctly but the resolver didn't take the latest value. That's why included a restart to pick up the latest value

Can you give me step-by-step instructions on how it got updated, and how you verified it? E.g. I don't see how restarting the networking service would update `/etc/resolv.conf`, so what am I looking for? Also lima-init is not updating resolv.conf, but changing the dhcp configuration.

So I would like to compare the effects of using cloud-init vs. lima-init, and check if there is any difference. But for that I need to know the exact steps on how to repro and verify the problem.

balajiv113:
As far as i remember the steps are as follows,
- Start a Alpine instance in lima 0.17.2 and check /etc/resolv.conf
- Stop the instance 
- Upgrade lima to latest, and start the instance 
- Note that the /etc/resolve.conf has a new nameservers but DNS calls will fail
- Run restart command it will work normally

jandubois:
> As far as i remember the steps are as follows,

Thank you, I could reproduce it (with VZ only). The root cause turns out to be the same as for `/etc/network/interfaces`: `/etc/resolv.conf` is updated during boot, but when we bind-mount the persistent volume on top, it reverts back to the first-boot value:

```console
lima-alpine:~# mkdir /bootfs
lima-alpine:~# mount --bind / /bootfs
lima-alpine:~# cat /bootfs/etc/resolv.conf
nameserver 192.168.5.2
lima-alpine:~# cat /etc/resolv.conf
nameserver 192.168.5.3
```

So the right fix IMO is to also copy the new version of the file to the persistent storage again on each boot. I've updated my PR, but left the TODO comment in there, as there might be still other settings that would benefit from the same treatment.

With this fix there is no longer a need to restart the networking service, assuming it was just done to force updating of `/etc/resolv.conf` via `ifup` and `udhcpd`.

balajiv113 (approved):
LGTM 👍

PR #2034, opened 2023-11-26
Disk resize command

vasileknik76:
The size of the created disk cannot be changed now. Even if the user resizes the disk using `qemu-img resize` he needs to run some commands in VM to grow the partition and resize the FS.

I added command `limactl disk resize` which can resize disk of a stopped vm using qemu-img.

Also I added parsing a format of an existing disk and render it in `limactl ls`

In cidata was added commands to resize the partition and the FS across all attached disks.

Shrinking of a disk is a danger operation, so i explicitly disallow it

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/05-lima-disks.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/05-lima-disks.sh
@@ -30,4 +30,17 @@ for i in $(seq 0 $((LIMA_CIDATA_DISKS - 1))); do
 
 	mkdir -p "/mnt/lima-${DISK_NAME}"
 	mount -t $FORMAT_FSTYPE "/dev/${DEVICE_NAME}1" "/mnt/lima-${DISK_NAME}"
+	if command -v growpart >/dev/null 2>&1 && command -v resize2fs >/dev/null 2>&1; then
+		growpart "/dev/${DEVICE_NAME}" 1 || true
+		# Only resize when filesystem is in a healthy state
+		if command -v "fsck.$FORMAT_FSTYPE" -f -p "/dev/disk/by-label/lima-${DISK_NAME}"; then
+			if [[ $FORMAT_FSTYPE == "ext2" || $FORMAT_FSTYPE == "ext3" || $FORMAT_FSTYPE == "ext4" ]]; then
+				resize2fs "/dev/disk/by-label/lima-${DISK_NAME}" || true
+			elif [ "$FORMAT_FSTYPE" == "xfs" ]; then
+				xfs_growfs "/dev/disk/by-label/lima-${DISK_NAME}" || true
+			else
+				echo >&2 "WARNING: unknown fs '$FORMAT_FSTYPE'. FS will not be grew up automatically"
+			fi
+		fi
+	fi
 done
--- a/pkg/store/disk.go
+++ b/pkg/store/disk.go
@@ -15,6 +15,7 @@ import (
 type Disk struct {
 	Name        string `json:"name"`
 	Size        int64  `json:"size"`
+	Format      string `json:"format"`
 	Dir         string `json:"dir"`
 	Instance    string `json:"instance"`
 	InstanceDir string `json:"instanceDir"`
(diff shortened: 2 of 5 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot/05-lima-disks.sh
```diff
@@ -28,6 +28,13 @@ for i in $(seq 0 $((LIMA_CIDATA_DISKS - 1))); do
 		fi
 	fi
 
+  if command -v growpart >/dev/null 2>&1 && command -v resize2fs >/dev/null 2>&1; then
+    growpart "/dev/${DEVICE_NAME}" 1 || true
+    # Only resize when filesystem is in a healthy state
+    if e2fsck -f -p "/dev/disk/by-label/lima-${DISK_NAME}"; then
+      resize2fs "/dev/disk/by-label/lima-${DISK_NAME}" || true
```
afbjorklund, line 35: Should probably check for an ext2 (et al) filesystem, perhaps also support xfs ?

afbjorklund: Apparently `-d` grows to the biggest.

vasileknik76: `xfs_repair` repairs FS but `e2fsck` just only checks it. I think need to use `fsck.$FSTYPE`

vasileknik76: I added `xfs_growfs` and tested `ext2/3/4` and `xfs`

afbjorklund: I thought you wanted to do a repair, since you did `e2fsck -p`

pkg/store/disk.go
```diff
@@ -57,32 +58,33 @@ func InspectDisk(diskName string) (*Disk, error) {
 	return disk, nil
 }
 
-// inspectDiskSize attempts to inspect the disk size by itself,
-// and falls back to inspectDiskSizeWithQemuImg on an error.
-func inspectDiskSize(fName string) (int64, error) {
+// inspectDiskSizeAndFormat attempts to inspect the disk size and format by itself,
+// and falls back to inspectDiskSizeAndFormatWithQemuImg on an error.
```
AkihiroSuda, line 62: Not WithQemuImg

pkg/store/disk.go
```diff
@@ -57,32 +58,33 @@ func InspectDisk(diskName string) (*Disk, error) {
 	return disk, nil
 }
 
-// inspectDiskSize attempts to inspect the disk size by itself,
-// and falls back to inspectDiskSizeWithQemuImg on an error.
-func inspectDiskSize(fName string) (int64, error) {
+// inspectDiskSizeAndFormat attempts to inspect the disk size and format by itself,
```
AkihiroSuda, line 61:
```suggestion
// inspectDisk attempts to inspect the disk size and format by itself,
```

pkg/store/disk.go
vasileknik76: Do you mean to use `WithImgUtil` instead of `WithQemuImg`?

pkg/cidata/cidata.TEMPLATE.d/boot/05-lima-disks.sh
vasileknik76: I did not plan to repair the file system, but only to check the functionality before increasing the size. Same code can be found at [04-persistent-data-volume.sh](https://github.com/lima-vm/lima/blob/bfd74711f565fe4cbd9cf923b96e6a11e4c785ca/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh#L24)

pkg/store/disk.go
AkihiroSuda: Sorry, I was misreading the comment line

AkihiroSuda (approved):
Thanks

AkihiroSuda:
CI not working?

Issue #1941, opened 2023-10-24
shell connections are bounced ~30 seconds into debian boot cycle subsequent to first boot/provision


mhio:
### Description

After the first boot/provision process has completed on a debian12 vm. Creating a new lima shell early in any subsequent boot process, the shell is bounced after about the same time as the initial `limactl start` creation command takes. 

limactl version 0.18.0

Watching the systemd journal before the shell session is bounced, it always exits here
```
$ lima sudo journalctl -f
<snip>
Oct 24 00:51:23 debvm cloud-init[1583]: Cloud-init v. 22.4.2 running 'modules:final' at Tue, 24 Oct 2023 00:51:23 +0000. Up 36.11 seconds.
```
Looking through the journal around then, I believe the `boot.sh` script output isn't flushed before termination so it's a bit later the session is terminated:
```
Oct 24 00:51:23 debvm cloud-init[1583]: + command -v loginctl
Oct 24 00:51:23 debvm cloud-init[1583]: + loginctl terminate-user mhio
Oct 24 00:51:23 debvm systemd[1]: Stopping session-2.scope - Session 2 of User mhio...
Oct 24 00:51:23 debvm sshd[1271]: pam_unix(sshd:session): session closed for user mhio
Oct 24 00:51:23 debvm sudo[1286]: pam_unix(sudo:session): session closed for user root
```
Which led me to 
https://github.com/lima-vm/lima/blob/b5aa0dd2f504a189382fb0ea6710a37151407abc/pkg/cidata/cidata.TEMPLATE.d/boot/07-etc-environment.sh#L17

I believe the purpose here is to inject environment before the lima provisioning starts. I assume some config changes might also need to be reflected here after first provision/boot? 

Could the [`login-ctl terminate-user`](https://github.com/lima-vm/lima/blob/b5aa0dd2f504a189382fb0ea6710a37151407abc/pkg/cidata/cidata.TEMPLATE.d/boot/07-etc-environment.sh#L7-L10) be guarded on the content of `/etc/environment` changing? Something like
```
if [ -e /etc/environment ]; then
  etc_environment_sum_pre="$(md5sum /etc/environment)"
  sed -i '/#LIMA-START/,/#LIMA-END/d' /etc/environment
fi
cat "${LIMA_CIDATA_MNT}/etc_environment" >>/etc/environment
etc_environment_sum_post="$(md5sum /etc/environment)"

[...]

PR #2048, opened 2023-12-02
pkg/cidata:  During boot, only terminate user connections when the environment changes 

mhio:
for #1941

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/07-etc-environment.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/07-etc-environment.sh
@@ -18,9 +18,6 @@ if command -v loginctl >/dev/null 2>&1 && [ "${orig}" != "$(cat /etc/environment
 	loginctl terminate-user "${LIMA_CIDATA_USER}" || true
 fi
 
-# Make sure the guestagent socket from a previous boot is removed before we open the "lima-ssh-ready" gate.
-rm -f /run/lima-guest-agent.sock
-
 # Signal that provisioning is done. The instance-id in the meta-data file changes on every boot,
 # so any copy from a previous boot cycle will have different content.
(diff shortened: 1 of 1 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot/07-etc-environment.sh
```diff
@@ -4,22 +4,22 @@ set -eux
...
 fi
 cat "${LIMA_CIDATA_MNT}/etc_environment" >>/etc/environment
+etc_environment_checksum_post="$(md5sum /etc/environment||printf 'forcepost')"
 
 # It is possible that a requirements script has started an ssh session before
 # /etc/environment was updated, so we need to kill it to make sure it will
 # restart with the updated environment before "linger" is being enabled.
 
-if command -v loginctl >/dev/null 2>&1; then
+if command -v loginctl >/dev/null 2>&1 && [ "${etc_environment_checksum_pre}" != "${etc_environment_checksum_post}" ]; then
```
jandubois, line 19: I'm not sure why we need to do the `md5sum` dance when we can just slurp the file into a variable. I also prefer somewhat shorter names.

Suggestion:

```
orig=$(test -f /etc/environment && cat /etc/environment)
if [ -f /etc/environment ]; then
	sed -i '/#LIMA-START/,/#LIMA-END/d' /etc/environment
fi
cat "${LIMA_CIDATA_MNT}/etc_environment" >>/etc/environment

# ...

if command -v loginctl >/dev/null 2>&1 && [ "${orig}" != "$(cat /etc/environment)" ]; then
```

jandubois:
Can you please squash commits? At least the first and last commit should be combined into a single commit.

AkihiroSuda (approved):
Thanks

Issue #2031, opened 2023-11-25
Libvirt driver, for hypervisors on remote machines 


afbjorklund:
### Description

You can use libvirt for talking to the local libvirtd, but it can also open a remote connection:

https://libvirt.org/remote.html

This makes it interesting for starting the QEMU machine on a local server, instead of localhost.

There is also the `qemu://session`, but it doesn't really add anything over running qemu-system?

----

Thinking of adding a new  "virt" driver, for using libvirt. The goal is to run qemu **remotely**.

* `qemu:///system` connects to a system mode daemon. (like rootful)

* `qemu:///session` connects to a session mode daemon. (like rootless)

The libvirt libraries bring a lot of dependencies though, so hide them in an optional .so plugin...

* https://github.com/lima-vm/lima/pull/2000

* https://github.com/lima-vm/lima/discussions/2007

----

### Alternatives

The alternative (to `qemu+ssh://`) is to `ssh` to the remote host, and run `limactl` from there...

https://libvirt.org/uri.html

The libvirt driver would be more interesting when coupled with [oVirt](https://ovirt.org/), to allocate a VM ?

But that would probably be better as a separate "cloud" driver, using the [go-ovirt-client](https://github.com/oVirt/go-ovirt-client)

afbjorklund:
Initial PR:
* #2032

Mostly to test the plugin system, it can successfully call the libvirt API _without_ adding to limactl.

```
INFO[0000] Starting the instance "virt" with VM driver "virt" 
INFO[0000] VERSION: 6000000                             
INFO[0000] Version: 6000000              
```

```
29M	_output/bin/limactl
4,7M	_output/lib/lima/plugin/virt.so
```

<details>
<pre>
$ ldd _output/bin/limactl 
	linux-vdso.so.1 (0x00007fff965a5000)
	libresolv.so.2 => /lib/x86_64-linux-gnu/libresolv.so.2 (0x00007f9fc6270000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f9fc624d000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f9fc6247000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9fc6055000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9fc62b3000)
$ ldd _output/lib/lima/plugin/virt.so 
[...]

PR #2054, opened 2023-12-05
Support for Virt driver

afbjorklund:


AkihiroSuda:
This has to be an external plugin:
- https://github.com/lima-vm/lima/discussions/2007

afbjorklund:
* https://github.com/lima-vm/lima/issues/2031

> The alternative (to `qemu+ssh://`) is to ssh to the remote host, and run limactl from there...

Issue #2055, opened 2023-12-05
`template://podman-rootful`: provision scripts should not depend on `LIMA_CIDATA_*`


AkihiroSuda:
https://github.com/lima-vm/lima/blob/b453dedb757e5fc1e249f352d0812d008e66fa3f/examples/podman-rootful.yaml#L44

The provision scripts should probably use template like `{{.User}}`
https://github.com/lima-vm/lima/pull/2052#issuecomment-1839114367

afbjorklund:
Also used in docker-rootful, I think

jandubois:
@AkihiroSuda Can you explain why template variables are preferable to the env variables?

It makes no big difference, but env variables are slightly more flexible if you want to work on the scripts in a running instance because the variables are not yet filled in (you can make changes to the script and copy it back into `lima.yaml`).

Another thing is that `lima.env` can be used to look up what information is available. But I guess we can document all the template variables in `default.yaml` as well.

Anyways, no strong feelings either way.

AkihiroSuda:
> Can you explain why template variables are preferable to the env variables?

Because these variables are implementation details of the init scripts, and can be removed even in a patch release.

jandubois:
> Because these variables are implementation details of the init scripts, and can be removed even in a patch release.

Oh, so the template would only make a subset of the information available? I thought you wanted to apply the cidata templating function to the scripts, which is functionally equivalent to exposing the shell variables; `lima.env` is essentially the mapping function.

So what are the template variables that should be exposed via the template?

AkihiroSuda:
> So what are the template variables that should be exposed via the template?

Just same as the template for the sockets?
https://github.com/lima-vm/lima/blob/b453dedb757e5fc1e249f352d0812d008e66fa3f/examples/default.yaml#L364

jandubois:
> Just same as the template for the sockets?

Ok, that works for me.

Would the `LIMA_*` variables still remain available for the system provisioning scripts (for backwards compatibility), or would you want to block them?

I guess we could do a string match for `LIMA_CIDATA` on the expanded script template and generate a warning?

AkihiroSuda:
> I guess we could do a string match for LIMA_CIDATA on the expanded script template and generate a warning?

👍

PR #2056, opened 2023-12-06
Export lima.env variables to user provisioning scripts

### LINKED ISSUE #2051: LIMA_CIDATA_* variables aren't accessible from user mode provision script
### Description

[Lima's environment vars](https://github.com/lima-vm/lima/blob/master/pkg/cidata/cidata.TEMPLATE.d/lima.env) are not accessible from a user mode provision script like this:

```yaml
provision:
- mode: user
  script: |
    #!/bin/sh
    ln -s $LIMA_CIDATA_HOSTHOME_MOUNTPOINT/.conan ~/.conan
```

The above script creates a wrong link since the `LIMA_CIDATA_HOSTHOME_MOUNTPOINT` variable isn't visible from the script:
```
# ls -la ~
total 20
drwxr-xr-x    4 zzzzz    zzzzz         4096 Dec  4 15:01 .
drwxr-xr-x    3 root     root          4096 Dec  2 00:02 ..
-rw-------    1 zzzzz    zzzzz          234 Dec  4 11:58 .ash_history
lrwxrwxrwx    1 zzzzz    zzzzz            7 Dec  4 15:01 .conan -> /.conan
drwxr-xr-x    3 zzzzz    zzzzz         4096 Dec  4 11:54 .config
drwx------    2 zzzzz    zzzzz         4096 Dec  2 00:02 .ssh
```

jandubois:
This is an alternative to #2052.

```console
$ yq .provision ~/.lima/alpine/lima.yaml
- mode: user
  script: "env | grep LIMA | sort"

$ limactl shell --workdir /var/log alpine sudo grep -A3 provision.user cloud-init-output.log
LIMA| Executing /mnt/lima-cidata/provision.user/00000000 (as user jan)
LIMA_CIDATA_CONTAINERD_SYSTEM=
LIMA_CIDATA_CONTAINERD_USER=
LIMA_CIDATA_DISKS=0
```

Fixes #2051

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.sh
@@ -56,8 +56,18 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.system ]; then
 	done
 fi
 
-USER_SCRIPT="${LIMA_CIDATA_HOME}/.lima-user-script"
 if [ -d "${LIMA_CIDATA_MNT}"/provision.user ]; then
+	USER_SCRIPT="${LIMA_CIDATA_HOME}/.lima-user-script"
+	USER_SCRIPT_LAUNCHER="${LIMA_CIDATA_HOME}/.lima-user-script-launcher"
+
+	# create wrapper script that exports all of lima.env before execing the user script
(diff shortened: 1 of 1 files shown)


jandubois:
We don't want to expose the variables to provisioning scripts (see #2055).

PR #2066, opened 2023-12-14
dependabot: enable groups

AkihiroSuda:
https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,6 +5,13 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+  groups:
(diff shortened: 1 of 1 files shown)


jandubois (approved):
LGTM, but I'm not sure how exactly the grouping is going to work. Let's try and see! 😄

AkihiroSuda:
> LGTM, but I'm not sure how exactly the grouping is going to work. Let's try and see! 😄

Like this:
- https://github.com/lima-vm/lima/pull/2072

(Closed because it is no longer compilable with Go 1.20)

Issue #1881, opened 2023-10-04
User interface library "survey" has been archived


afbjorklund:
### Description

> This repository has been archived by the owner on Sep 24, 2023. It is now read-only.

https://github.com/AlecAivazis/survey/commit/160123ed7175f77bc8625b59c44b8dc03f5333e3

Might want to look for a replacement, or use a different / maintained fork of the library.

https://github.com/AlecAivazis/survey/network

afbjorklund:
New library in 2023: https://github.com/charmbracelet/huh

AkihiroSuda:
tview looks attractive in the modest number of the dependencies
https://github.com/rivo/tview/blob/v0.42.0/go.mod

afbjorklund:
The code is nice, the main problem is that it is fullscreen and not inline.

<img width="996" height="390" alt="Image" src="https://github.com/user-attachments/assets/b3c45d1b-2f19-417f-b63c-2fb2995cb404" />

It seems to be a limitation in the underlying https://github.com/gdamore/tcell

```go
// Confirm is a regular text input that accept yes/no answers.
func Confirm(message string, defaultParam bool) (bool, error) {
	app := tview.NewApplication()
	var ans bool
	modal := tview.NewModal()
	modal.SetText(message)
	modal.AddButtons([]string{"Yes", "No"})
	if defaultParam {
		modal.SetFocus(0)
	} else {
		modal.SetFocus(1)
	}
	modal.SetDoneFunc(func(buttonIndex int, _ string) {
		ans = buttonIndex == 0
		app.Stop()
	})
	if err := app.SetRoot(modal, false).Run(); err != nil {
		return false, err
	}
	return ans, nil
}

// Select is a prompt that presents a list of various options
// to the user for them to select using the arrow keys and enter.
func Select(message string, options []string) (int, error) {
	app := tview.NewApplication()
	var ans int
	list := tview.NewList()
	list.ShowSecondaryText(false)
	for _, option := range options {
		list.AddItem(option, "", 0, nil)
	}
	list.SetSelectedFunc(func(listIndex int, _, _ string, _ rune) {
		ans = listIndex
		app.Stop()
	})
	if err := app.SetRoot(list, false).Run(); err != nil {
		return -1, err
	}
	return ans, nil
}
```

AkihiroSuda:
Thanks for investigation.
I had once thought that Bubbletea/Huh were overkill, but it might be fine, as we will probably eventually have a full-featured TUI menu.

[...]

PR #2078, opened 2023-12-16
Move all usage of survey into an internal pkg

afbjorklund:
This makes it easier to replace the deprecated survey module with another Text User Interface implementation, later on.

* #1881

* https://github.com/AlecAivazis/survey

None of the currently available alternatives is a good replacement, they take over the entire window and add a lot of deps.

* https://github.com/charmbracelet/bubbletea

* https://github.com/rivo/tview

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -10,8 +10,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/AlecAivazis/survey/v2"
-	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/containerd/containerd/identifiers"
 	"github.com/lima-vm/lima/cmd/limactl/editflags"
 	"github.com/lima-vm/lima/cmd/limactl/guessarg"
@@ -24,6 +22,7 @@ import (
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/lima-vm/lima/pkg/templatestore"
+	"github.com/lima-vm/lima/pkg/uiutil"
 	"github.com/lima-vm/lima/pkg/yqutil"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


afbjorklund:
Another alternative would be to fork github.com/lima-vm/survey, and keep the API.

pkg/interfaceutil/interfaceutil.go
```diff
@@ -0,0 +1,32 @@
+package interfaceutil
+
```
AkihiroSuda, line 2: Maybe "tuiutil" or just "uiutil" to avoid confusion with Go interfaces

pkg/interfaceutil/interfaceutil.go
```diff
@@ -0,0 +1,32 @@
...
+		Message: message,
+		Default: true,
+	}
+	if err := survey.AskOne(prompt, &ans); err != nil {
+		return false, err
+	}
+	return ans, nil
+}
+
+func Select(message string, options []string) (int, error) {
```
AkihiroSuda, line 22: Maybe "Choose" to avoid confusion with selecting chans or fds

cmd/limactl/start.go
```diff
@@ -360,14 +359,19 @@ func modifyInPlace(st *creatorState, yq string) error {
 	return nil
 }
 
+type Prompt struct {
```
AkihiroSuda, line 362: Needs godoc

AkihiroSuda: Maybe this can be rather moved to the interfaceutil pkg

AkihiroSuda:
Is this still a draft?

afbjorklund:
I was mostly using it as a base to experiment with bubbletea. Not so impressed with the result...

lima-bubbletea

* https://github.com/charmbracelet/bubbletea/tree/master/examples/list-simple

But we could add it as a refactoring, if the new  (minimal) "API" is enough to build on.

AkihiroSuda (approved):
Thanks

Issue #2064, opened 2023-12-13
Guest agent sometimes doesn't start (or is unreachable from host agent)


jandubois:
Well, at least not initially.

[Update: it seems to start, but hostagent cannot connect to it]

We've observed this multiple times with Rancher Desktop after switching to Lima 0.19.0 (https://github.com/rancher-sandbox/rancher-desktop/issues/6151).

AFAIK, this was always with QEMU and not VZ, and always when restarting an existing instance, and not when creating a new one.

Rancher Desktop is using Alpine; I have not seen it with other distros, but that doesn't mean much.

I don't think guest-agent logs are written to file on Alpine; I will need to look into fixing that.

I've run into it twice today on different machines, but there don't seem to be any steps to reliably reproduce the problem.

* macOS Sonoma 14.2 on M1 with QEMU 8.0.0

```
    'time="2023-12-13T10:04:13-08:00" level=info msg="[hostagent] Waiting for the guest agent to be running"\n' +
    'time="2023-12-13T10:04:13-08:00" level=info msg="[hostagent] Forwarding \\"/var/run/docker.sock\\" (guest) to \\"/Users/jan/.rd/docker.sock\\" (host)"\n' +
    'time="2023-12-13T10:05:13-08:00" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \\"boot scripts must have finished\\""\n' +
    'time="2023-12-13T10:05:13-08:00" level=info msg="[hostagent] The final requirement 1 of 1 is satisfied"\n' +
    'time="2023-12-13T10:05:13-08:00" level=error msg="[guest agent does not seem to be running; port forwards will not work]"\n' +
    'time="2023-12-13T10:05:13-08:00" level=warning msg="DEGRADED. The VM seems running, but file sharing and port forwarding may not work. (hint: see \\"/Users/jan/Library/Application Support/rancher-desktop/lima/0/ha.stderr.log\\")"\n' +
    'time="2023-12-13T10:05:13-08:00" level=fatal msg="degraded, status={Running:true Degraded:true Exiting:false Errors:[guest agent does not seem to be running; port forwards will not work] SSHLocalPort:63025}"\n',
```

* macOS Monterey 12.7.2 on Intel with QEMU 7.1.0

```
    'time="2023-12-13T10:35:24-08:00" level=info msg="[hostagent] Waiting for the guest agent to be running"\n' +
[...]

PR #2090, opened 2023-12-19
Add timestamps to LIMA boot.sh messages

jandubois:
to make it easier to correlate them to entries in ha.std*.log files.

Related to debugging #2064.

Messages look like this:

```
LIMA 2023-12-19T06:56:36+00:00| Executing /mnt/lima-cidata/boot/25-guestagent-base.sh
```

At first I appended them to the end:

```
LIMA| Executing /mnt/lima-cidata/boot/25-guestagent-base.sh at 2023-12-19T06:56:36+00:00
```

but I found that makes them hard to locate in the log file.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.sh
@@ -2,11 +2,11 @@
 set -eu
 
 INFO() {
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #2097, opened 2023-12-22
Set guest timezone from host setting

### LINKED ISSUE #1921: Propagate timezone from the host by default (unless `--plain` is specified)

jandubois:
Allow explicit override via lima.yaml.

Fixes #1921 

Replacement for abandoned #1931

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -259,6 +259,11 @@ rosetta:
   # 🟢 Builtin default: false
   binfmt: null
 
+# Specify the timezone name (as used by the zoneinfo database). Specify the empty string
+# to not set a timezone in the instance.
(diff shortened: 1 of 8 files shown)


jandubois:
Note that this PR will modify the timezone setting in existing instances, when they are restarted. I assume this is acceptable, as the timezone should just be a presentation setting.

If this is not acceptable then this PR should wait until we record the Lima version used to create the instance, so only new instances would get new default settings (see https://github.com/lima-vm/lima/pull/1953#issuecomment-1784546687 for how I would implement this).

AkihiroSuda (approved):
Thanks

AkihiroSuda:
vz mode now prints `WARN[0001] vmType vz: ignoring [TimeZone] ` on every start up

PR #2102, opened 2023-12-28
k8s: stop using sudo kubectl for kubeconfig

afbjorklund:
Applies change from 0aed87173d47327049d3006dfd1b564e0d17f20e

```
The connection to the server localhost:8080 was refused - did you specify the right host or port?
```

Reverts the previous a5427ddca0f87cf60235400feeac514782f57b80

Still need to run `sudo crictl`

### DIFF
--- a/examples/k8s.yaml
+++ b/examples/k8s.yaml
@@ -1,6 +1,6 @@
 # Deploy kubernetes via kubeadm.
 # $ limactl start ./k8s.yaml
-# $ limactl shell k8s sudo kubectl
+# $ limactl shell k8s kubectl
 
 # It can be accessed from the host by exporting the kubeconfig file;
 # the ports are already forwarded automatically by lima:
@@ -126,6 +126,14 @@ provision:
     # Replace the server address with localhost, so that it works also from the host
     sed -e "/server:/ s|https://.*:\([0-9]*\)$|https://127.0.0.1:\1|" -i $KUBECONFIG
     mkdir -p ${HOME:-/root}/.kube && cp -f $KUBECONFIG ${HOME:-/root}/.kube/config
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    export KUBECONFIG=/etc/kubernetes/admin.conf
+    mkdir -p ${LIMA_CIDATA_HOME}/.kube
+    cp -f $KUBECONFIG ${LIMA_CIDATA_HOME}/.kube/config
+    chown -R ${LIMA_CIDATA_USER} ${LIMA_CIDATA_HOME}/.kube
 probes:
 - description: "kubeadm to be installed"
   script: |
@@ -151,7 +159,7 @@ probes:
   script: |
     #!/bin/bash
     set -eux -o pipefail
-    if ! timeout 300s bash -c "until sudo kubectl version >/dev/null 2>&1; do sleep 3; done"; then
+    if ! timeout 300s bash -c "until kubectl version >/dev/null 2>&1; do sleep 3; done"; then
       echo >&2 "kubernetes cluster is not up and running yet"
       exit 1
     fi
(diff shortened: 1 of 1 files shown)


examples/k8s.yaml
```diff
@@ -125,7 +125,9 @@ provision:
     kubectl taint nodes --all node-role.kubernetes.io/control-plane-
     # Replace the server address with localhost, so that it works also from the host
     sed -e "/server:/ s|https://.*:\([0-9]*\)$|https://127.0.0.1:\1|" -i $KUBECONFIG
-    mkdir -p ${HOME:-/root}/.kube && cp -f $KUBECONFIG ${HOME:-/root}/.kube/config
```
AkihiroSuda, line 128: This should be retained to keep `sudo kubectl` functional?

afbjorklund: If needed, it is still possible to use something like:

`sudo KUBECONFIG=/etc/kubernetes/admin.conf kubectl`

https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/create-cluster-kubeadm/

AkihiroSuda: Yes, but I wonder if somebody might have been using a script that depends on the old behavior

AkihiroSuda: (Same discussion applies to k3s.yaml)

afbjorklund: I think for k3s it just used the same config file, and just relaxed the file permissions on it. So "should work"

cmd/kubectl.lima
```diff
@@ -33,5 +33,5 @@ if [ -n "$KUBECTL" ]; then
   exec "$KUBECTL" "$@"
 else
   export LIMA_INSTANCE
-  exec lima sudo kubectl "$@"
+  exec lima kubectl "$@"
```
jandubois, line 36: Is there a reason to remove the `sudo`? This will make existing instances created by previous Lima releases incompatible with the new script.

Since the kubeconfig still exists in both places, the existing code would continue to work, no?

afbjorklund: It should work now, similar to using `sudo docker` even if you are already part of the root group

jandubois: Yes, that's why you should leave the `sudo` in the script it, to avoid breaking it for older instances:

* User has Lima 0.19 and creates a `k8s` instance, which works fine with `kubectl.lima`.
* User upgrades to Lima 0.20 but keeps existing `k8s` instance.
* `kubectl.lima` no longer works because `${LIMA_CIDATA_HOME}/.kube/config` does not exist

The provisioning script is guarded with

```shell
test -e /etc/kubernetes/admin.conf && exit 0
```

So it won't be executed when you restart the existing instance with 0.20 and the user-mode kubeadmin will never get created.

Alll this can be avoided by leaving the `sudo` in `kubectl.lima`. Or if there is some advantage to removing it, then the provisioning script needs to be changed so that `${LIMA_CIDATA_HOME}/.kube/config` is created on each boot to migrate existing instances over.

jandubois: > Or if there is some advantage to removing it, then the provisioning script needs to be changed so that ${LIMA_CIDATA_HOME}/.kube/config is created on each boot to migrate existing instances over.

I'm sorry; that was a thinko on my part. The old instance will never see the new provisioning scripts; so `kubectl.lima` needs to continue to work with the old one if we don't want to break backwards compatibility (and I think we shouldn't).

So I think leaving the `sudo` in the `kubectl.lima` script is the right way to deal with this.

Sorry for the confusion!

AkihiroSuda:
What's current status?

afbjorklund:
I restored the old code, for backwards functionality with existing instances, so now it should work for both.

jandubois (approved):
Sorry, I missed again that the code had been updated.

Everything looks good to me and can be merged.

PR #2107, opened 2024-01-01
Store Lima version in the instance directory

jandubois:
That way future Lima releases can make decisions about default settings based on the Lima version that created the instance instead of the current instance, and settings for existing instances will not change just because the user updated Lima.

This implements the mechanism I described in https://github.com/lima-vm/lima/pull/1953#issuecomment-1784546687. I'm leaving it in draft mode for now because I'm not sure if the `store.MinVersion` function is the right way to use the data; I originally wanted to make it an instance method, but then noticed that most location that would need to call it don't have a copy of the instance.

So this PR probably needs to wait until we have a companion PR that makes use of this new mechanism. Or I could drop the `MinVersion` function for now and leave it for the first PR that will take advantage of the data.

```console
$ limactl ls --format '{{.Name}} - {{.LimaVersion}}'
alpine - v0.19.1
default - v0.19.1-16-gf3dc6ed.m
```

Either way, feedback welcome!

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -23,6 +23,7 @@ import (
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/lima-vm/lima/pkg/templatestore"
 	"github.com/lima-vm/lima/pkg/uiutil"
+	"github.com/lima-vm/lima/pkg/version"
 	"github.com/lima-vm/lima/pkg/yqutil"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -330,6 +331,9 @@ func createInstance(ctx context.Context, st *creatorState, saveBrokenEditorBuffe
 		return nil, err
 	}
 	if err := os.WriteFile(filePath, st.yBytes, 0o644); err != nil {
+		return nil, err
+	}
+	if err := os.WriteFile(filepath.Join(instDir, filenames.LimaVersion), []byte(version.Version), 0o444); err != nil {
 		return nil, err
 	}
 
--- a/pkg/store/filenames/filenames.go
+++ b/pkg/store/filenames/filenames.go
@@ -27,6 +27,7 @@ const (
 
 const (
 	LimaYAML           = "lima.yaml"
+	LimaVersion        = "lima-version" // Lima version used to create instance
 	CIDataISO          = "cidata.iso"
 	CIDataISODir       = "cidata"
 	BaseDisk           = "basedisk"
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -16,12 +16,14 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/docker/go-units"
 	hostagentclient "github.com/lima-vm/lima/pkg/hostagent/api/client"
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/lima-vm/lima/pkg/store/dirnames"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/lima-vm/lima/pkg/textutil"
+	"github.com/sirupsen/logrus"
 )
 
 type Status = string
@@ -56,6 +58,7 @@ type Instance struct {
 	Config          *limayaml.LimaYAML `json:"config,omitempty"`
 	SSHAddress      string             `json:"sshAddress,omitempty"`
 	Protected       bool               `json:"protected"`
+	LimaVersion     string             `json:"limaVersion"`
 }
 
 func (inst *Instance) LoadYAML() (*limayaml.LimaYAML, error) {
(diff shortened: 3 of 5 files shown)


pkg/store/instance.go
```diff
@@ -35,6 +37,8 @@ const (
 	StatusRunning       Status = "Running"
 )
 
+const DefaultLimaVersion = "v0.19.1"
```
AkihiroSuda, line 40: This shouldn't be hard-coded

pkg/store/instance.go
```diff
@@ -423,3 +436,22 @@ func (inst *Instance) Unprotect() error {
 	inst.Protected = false
 	return nil
 }
+
+// MinVersion checks if the instance has been created by the specified minimum version of Lima.
+// Prerelease versions are treated as the previous version, not the next, so "0.19.1-16-gf3dc6ed.m"
+// is considered the same as "0.19.1" and not "0.20.0".
+func MinVersion(instanceVersion, minVersion string) bool {
```
AkihiroSuda, line 443: Probably this should have a unit test

pkg/store/instance.go
```diff
@@ -167,6 +172,14 @@ func Inspect(instName string) (*Instance, error) {
 			}
 		}
 	}
+	if version, err := os.ReadFile(filepath.Join(instDir, filenames.LimaVersion)); err == nil {
+		inst.LimaVersion = strings.TrimSpace(string(version))
+	} else {
+		if !errors.Is(err, os.ErrNotExist) {
+			inst.Errors = append(inst.Errors, err)
+		}
+		inst.LimaVersion = DefaultLimaVersion
```
AkihiroSuda, line 181: Maybe just leave it empty?

pkg/store/filenames/filenames.go
```diff
@@ -27,6 +27,7 @@ const (
 
 const (
 	LimaYAML           = "lima.yaml"
+	LimaVersion        = "lima-version" // Lima version used to create instance
```
AkihiroSuda, line 30: Please update https://lima-vm.io/docs/dev/internals/#instance-directory-lima_homeinstance too

cmd/limactl/start.go
```diff
@@ -332,6 +333,9 @@ func createInstance(ctx context.Context, st *creatorState, saveBrokenEditorBuffe
 	if err := os.WriteFile(filePath, st.yBytes, 0o644); err != nil {
 		return nil, err
 	}
+	if err := os.WriteFile(filepath.Join(instDir, filenames.LimaVersion), []byte(version.Version), 0o644); err != nil {
```
AkihiroSuda, line 336:
```suggestion
	if err := os.WriteFile(filepath.Join(instDir, filenames.LimaVersion), []byte(version.Version), 0o444); err != nil {
```

As this file should never be updated?

jandubois:
@AkihiroSuda Thank you for your feedback; I believe I've addressed all your comments.

I switched to defining a `LimaVersionGreaterThan` function, so can select default changes based on the last version that still has the old functionality instead of having to guess the next release version number that will have the new default. Still needs to be validated by an actual implementation that uses it though.

```
if LimaVersionGreaterThan(inst.LimaVersion, "0.19.1") {
   // set new default
}
```

This code would use the new defaults with any current dev build because `v0.19.1-xxx` will be treated as greater than `0.19.1`.

This should also make it easier to test the new functionality in CI without having to bump the version first.

It does mean though that all instances created by any dev build of Lima after `0.19.1` will be treated as if it was created by the latest build. I think this makes sense, as Lima developers should understand what is going to happen. Let me know what you think!

jandubois:
Current failure happens because the CI version has no tags, so the version cannot be parsed from the SHA1:

```
time="2024-01-02T05:59:07Z" level=fatal msg="errors inspecting instance: [cannot parse lima version \"ca7752d\" from \"/Users/runner/.lima/default/lima-version\": ca7752d is not in dotted-tri format]"
```

Not sure if we should pull tags in CI, so we can then test version-specific changes?

AkihiroSuda:
> Not sure if we should pull tags in CI, so we can then test version-specific changes?

We may pull the tags, but anyway the commit hash string shouldn't cause a fatal error.

jandubois:
> We may pull the tags, but anyway the commit hash string shouldn't cause a fatal error.

The problem is that we won't know the version, so all the version check will return `false` then. Instead of the newest defaults, we get the oldest ones. Therefore I kind of prefer that this fails in CI instead of silently testing the old values.

We can make this configurable, but it feels like overkill: why do we need to support building outside of a git checkout?

AkihiroSuda:
> The problem is that we won't know the version, so all the version check will return false then. Instead of the newest defaults, we get the oldest ones. Therefore I kind of prefer that this fails in CI instead of silently testing the old values.

I think the version without semver prefix can be just assumed to be "latest", with a warning message.
For the CI, we can pull the tags to attach the semver prefix.

> why do we need to support building outside of a git checkout?

Because somebody might be using a package built in that way?

jandubois:
> I think the version without semver prefix can be just assumed to be "latest", with a warning message.

Ok, I think that would work for all practical scenarios; updated!

AkihiroSuda (approved):
Thanks, is this still draft?

PR #2112, opened 2024-01-03
Revert virtserialport back to unix socket for QEMU guest agent communication

jandubois:
The serial port sometimes doesn't seem to work: #2064.

The vsock implementation used for VZ doesn't seem to have this problem.

It makes sense to revert to the old mechanism for QEMU until the serialport has been fixed (if possible).

Just commenting out the virtioPort will fall back to the old socket method thanks to #2006.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -128,7 +128,8 @@ func New(instName string, stdout io.Writer, sigintCh chan os.Signal, opts ...Opt
 		}
 		vSockPort = port
 	} else if *y.VMType == limayaml.QEMU {
-		virtioPort = filenames.VirtioPort
+		// virtserialport doesn't seem to work reliably: https://github.com/lima-vm/lima/issues/2064
+		virtioPort = "" // filenames.VirtioPort
 	}
 
 	if err := cidata.GenerateISO9660(inst.Dir, instName, y, udpDNSLocalPort, tcpDNSLocalPort, o.nerdctlArchive, vSockPort, virtioPort); err != nil {
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
@@ -35,7 +35,7 @@ name="lima-guestagent"
 description="Forward ports to the lima-hostagent"
 
 command=${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent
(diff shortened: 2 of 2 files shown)


pkg/hostagent/hostagent.go
```diff
@@ -128,7 +128,8 @@ func New(instName string, stdout io.Writer, sigintCh chan os.Signal, opts ...Opt
 		}
 		vSockPort = port
 	} else if *y.VMType == limayaml.QEMU {
-		virtioPort = filenames.VirtioPort
+		// virtserialport doesn't seem to work reliably: https://github.com/lima-vm/lima/issues/2064
```
AkihiroSuda, line 131: Could you report this to https://gitlab.com/qemu-project/qemu/-/issues ?

jandubois: Not sure what exactly I would report; I have no more information beyond that the code in Lima sometimes fails. It doesn't even fail reliably, which is why the bisect script restarts the VM up to 100 times before accepting that the code works properly.

I haven't checked if the problem happens on the sending or the receiving side, as we are not getting any errors.

So I feel this would need some more investigation before reporting to upstream, but unfortunately I'm out of time right now. 😞

AkihiroSuda (dismissed):
Thanks

AkihiroSuda:
Shall we release a new version after merging this?
(v0.19.2 ? v0.20 ?)

jandubois:
Converted to draft because it only works properly with Ubuntu, but not with Alpine. Will investigate...

jandubois:
> it only works properly with Ubuntu, but not with Alpine

I think I fixed it now (just some quoting issues in `/etc/init.d/lima-guestagent`), but will leave it as draft until I have done more testing.

jandubois:
> will leave it as draft until I have done more testing

Seems to work fine with Alpine as well now.

> Shall we release a new version after merging this?

I think so, as running into the flakiness is annoying, and not immediately obvious if it is your fault or not. Don't really care about the version 0.19.2 vs. 0.20.0; either one works for me.

AkihiroSuda (approved):

pkg/hostagent/hostagent.go
```diff
@@ -128,7 +128,8 @@ func New(instName string, stdout io.Writer, sigintCh chan os.Signal, opts ...Opt
...
+		virtioPort = "" // filenames.VirtioPort
```
jandubois, line 132: Re-assigning the empty string here to satisfy the linter; couldn't get it to shut up blathering about empty branches and empty blocks if I left everything commented out...

AkihiroSuda:
@balajiv113 LGTY?

balajiv113 (approved):
LGTM 👍

PR #2118, opened 2024-01-08
Fix GuestAgentConn vsock support

pendo324:
When how Lima connects to the guest agent was refactored in #1998, https://github.com/mdlayher/vsock was used to support vsock connections on Windows, but the library doesn't support Windows. The following error log is from a WSL Lima instance on v0.19.1:

```
{"error":"dial vsock host(2):994831394: vsock: not implemented on windows","level":"warning","msg":"connection to the guest agent was closed unexpectedlydial vsock host(2):994831394: vsock: not implemented on windows","time":"2024-01-08T19:19:49Z"}
```

The corresponding log from the library is [here](https://github.com/mdlayher/vsock/blob/068aff8a5e29cc13661b0f5a16cc347614c2bc7d/vsock_others.go#L17).

With this change to revert to using [winio's Dial](https://github.com/microsoft/go-winio/blob/bc421d9108ade8ae2f76c8ef3e90cd4ee690e2c0/hvsock.go#L315), guestagent communication works as it did previously.

More information on the how the HyperV/Windows vsock support works (can add this to the comment in wsl_driver too):
* https://github.com/torvalds/linux/blob/master/net/vmw_vsock/hyperv_transport.c#L112-L156
* https://learn.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/make-integration-service#register-a-new-application

Also un-exported `MagicVSOCKSuffix` since it's now only used in the windows package.

### DIFF
--- a/pkg/windows/registry_windows.go
+++ b/pkg/windows/registry_windows.go
@@ -13,7 +13,7 @@ import (
 
 const (
 	guestCommunicationsPrefix = `SOFTWARE\Microsoft\Windows NT\CurrentVersion\Virtualization\GuestCommunicationServices`
-	MagicVSOCKSuffix          = "-facb-11e6-bd58-64006a7986d3"
+	magicVSOCKSuffix          = "-facb-11e6-bd58-64006a7986d3"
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #2121, opened 2024-01-09
Do guest template expansion for provisioning scripts

### LINKED ISSUE #2055: `template://podman-rootful`: provision scripts should not depend on `LIMA_CIDATA_*`

jandubois:
The LIMA_CIDATA_* variables are not a supported API and can change without warning.

Fixes #2055

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -184,6 +184,7 @@ containerd:
 
 # Provisioning scripts need to be idempotent because they might be called
 # multiple times, e.g. when the host VM is being restarted.
+# The scripts can use the following template variables: {{.Home}}, {{.UID}}, and {{.User}}
 # 🟢 Builtin default: null
 # provision:
 # # `system` is executed with the root privilege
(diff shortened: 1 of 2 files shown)


examples/docker-rootful.yaml
```diff
@@ -6,7 +6,7 @@
 # $ export DOCKER_HOST=$(limactl list docker-rootful --format 'unix://{{.Dir}}/sock/docker.sock')
 # $ docker ...
 
-# This template requires Lima v0.8.0 or later
+# This template requires Lima v0.20.0 or later
```
AkihiroSuda, line 9: This should happen after releasing v0.20, as some people may use the YAMLs in the master branch without installing Lima from the master branch.

jandubois: Split off template changes into #2123

afbjorklund:
A nice thing about this change, is that we can do provisioning without implying cloud-init (cidata).

AkihiroSuda (approved):
Thanks

PR #2123, opened 2024-01-09
Use template variables instead of LIMA_CIDATA_* bash variables

jandubois:
Should only be merged once #2121 has been included in a release.

### DIFF
--- a/examples/docker-rootful.yaml
+++ b/examples/docker-rootful.yaml
@@ -6,7 +6,7 @@
 # $ export DOCKER_HOST=$(limactl list docker-rootful --format 'unix://{{.Dir}}/sock/docker.sock')
 # $ docker ...
 
-# This template requires Lima v0.8.0 or later
+# This template requires Lima v0.20.0 or later
 images:
 # Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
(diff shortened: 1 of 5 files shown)


jandubois:
Updated to include `examples/k8s.yaml` changes from #2102.

Also added a warning when `LIMA_CIDATA` appears in a provisioning script, as agreed to in https://github.com/lima-vm/lima/issues/2055#issuecomment-1842165767.

I first included the script in the warning text as well, but then decided that would be too annoying for users still running older instances:

```console
$ l ls
WARN[0000] provisioning scripts should not reference the LIMA_CIDATA variables: "#!/bin/bash\nset -eux -o pipefail\nexport KUBECONFIG=/etc/kubernetes/admin.conf\nmkdir -p ${LIMA_CIDATA_HOME}/.kube\ncp -f $KUBECONFIG ${LIMA_CIDATA_HOME}/.kube/config\nchown -R ${LIMA_CIDATA_USER} ${LIMA_CIDATA_HOME}/.kube\n"
NAME    STATUS     SSH                VMTYPE    ARCH      CPUS    MEMORY    DISK      DIR
k8s     Running    127.0.0.1:52532    qemu      x86_64    4       4GiB      100GiB    ~/.lima/k8s
```

AkihiroSuda (approved):
Thanks

PR #2136, opened 2024-01-15
Bump to Alpine 3.19

jandubois:
Also adds support for timezone syncing:

```console
$ ls -l /etc/localtime | sed 's/.*[0-9] //'
/etc/localtime -> /var/db/timezone/zoneinfo/America/Vancouver

$ limactl shell alpine ls -l /etc/localtime | sed 's/.*[0-9] //'
/etc/localtime -> /etc/zoneinfo/America/Vancouver
```

### DIFF
--- a/examples/alpine.yaml
+++ b/examples/alpine.yaml
@@ -1,13 +1,13 @@
 # This template requires Lima v0.7.0 or later.
-# Using the Alpine 3.18 aarch64 image with vmType=vz requires macOS Ventura 13.3 or later.
+# Using the Alpine 3.19 aarch64 image with vmType=vz requires macOS Ventura 13.3 or later.
 
 images:
-- location: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.32/alpine-lima-std-3.18.0-x86_64.iso"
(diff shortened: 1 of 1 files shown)


jandubois:
I just noticed that OpenRC in Alpine 3.19 has changed the default of `rc_cgroup_mode` from `hybrid` to `unified`. Not sure if we should make a note about this in the template?

AkihiroSuda:
> I just noticed that OpenRC in Alpine 3.19 has changed the default of `rc_cgroup_mode` from `hybrid` to `unified`. Not sure if we should make a note about this in the template?

Yes, the release note has to mention it

AkihiroSuda (approved):
Thanks

jandubois:
> Yes, the release note has to mention it

Release note only, or also the template?

AkihiroSuda:
> > Yes, the release note has to mention it
> 
> Release note only, or also the template?

I think it doesn't need to be specifically mentioned in the template, as long as it corresponds to the upstream default.
Other templates do not specifically mention the cgroup hierarchy either.

PR #2140, opened 2024-01-15
hostagent: Handle SIGTERM signal for better compatibility with macOS launchd

norio-nomura:
This PR modifies the hostagent to handle not only the SIGINT (Interrupt) signal, but also the SIGTERM signal. This allows the hostagent to gracefully shut down when it receives a SIGTERM signal, which is a common way to request a process to terminate. This change improves the compatibility with macOS launchd, which uses SIGTERM to stop services.

The following are the steps to set up the Docker instance of Lima to start at login using launchd on macOS:
```bash
# Create the docker instance
limactl start template://docker --vm-type vz --rosetta --network vzNAT --tty=false
# Stop the docker instance
limactl stop docker
# Create the launchd plist
echo '<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
	<key>Label</key>
	<string>io.lima-vm.docker</string>
	<key>ProgramArguments</key>
	<array>
		<string>'"$(which limactl)"'</string>
		<string>hostagent</string>
		<string>--pidfile</string>
		<string>ha.pid</string>
		<string>--socket</string>
		<string>ha.sock</string>
		<string>docker</string>
	</array>
	<key>RunAtLoad</key>
	<true/>
	<key>StandardErrorPath</key>
	<string>ha.stderr.log</string>
	<key>StandardOutPath</key>
	<string>ha.stdout.log</string>
	<key>WorkingDirectory</key>
	<string>'"$(limactl list docker --format '{{.Dir}}')"'</string>
</dict>
</plist>' > ~/Library/LaunchAgents/io.lima-vm.docker.plist
# Load the launchd plist and launchd will start the docker instance
launchctl load ~/Library/LaunchAgents/io.lima-vm.docker.plist
```

rel: #1027

### DIFF
--- a/cmd/limactl/hostagent.go
+++ b/cmd/limactl/hostagent.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"runtime"
 	"strconv"
+	"syscall"
 
 	"github.com/gorilla/mux"
 	"github.com/lima-vm/lima/pkg/hostagent"
@@ -67,8 +68,8 @@ func hostagentAction(cmd *cobra.Command, args []string) error {
 		defer runtime.UnlockOSThread()
 	}
 
-	sigintCh := make(chan os.Signal, 1)
-	signal.Notify(sigintCh, os.Interrupt)
+	signalCh := make(chan os.Signal, 1)
+	signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM)
 
 	stdout := &syncWriter{w: cmd.OutOrStdout()}
 	stderr := &syncWriter{w: cmd.ErrOrStderr()}
@@ -82,7 +83,7 @@ func hostagentAction(cmd *cobra.Command, args []string) error {
 	if nerdctlArchive != "" {
 		opts = append(opts, hostagent.WithNerdctlArchive(nerdctlArchive))
 	}
-	ha, err := hostagent.New(instName, stdout, sigintCh, opts...)
+	ha, err := hostagent.New(instName, stdout, signalCh, opts...)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


jandubois (commented):
Change looks good to me, but I haven't had a chance to test it yet.

AkihiroSuda:
The hostagent should be invoked via limactl.
Isn't it currently possible to launch limactl start via launched?

norio-nomura:
The method of invoking `limactl start docker` with `launchd` does not work.
If `limactl start` starts `limactl hostagent` and terminates itself, `launchd` will detect the termination and terminate `limactl hostagent` and other processes in the same process group.
Using `<key>AbandonProcessGroup</key><true/>` will prevent `limactl hostagent` from terminating by `launchd`, but will also prevent it from terminating on user logout.
Therefore, I submitted this change because I thought it would be better to allow `launchd` to start and stop `limactl hostagent` directly.

jandubois:
> Isn't it currently possible to launch limactl start via launched?

Not easily; we would need to add an option so that instead of daemonizing a separate hostagent process we would run the hostagent as part of `limactl start`.

From [Creating Launch Daemons and Agents](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPSystemStartup/Chapters/CreatingLaunchdJobs.html):

> You must not daemonize your process. This includes calling the [daemon](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/daemon.3.html#//apple_ref/doc/man/3/daemon) function, calling fork followed by [exec](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/exec.3.html#//apple_ref/doc/man/3/exec), or calling fork followed by [exit](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/exit.3.html#//apple_ref/doc/man/3/exit). If you do, launchd thinks your process has died. Depending on your property list key settings, launchd will either keep trying to relaunch your process until it gives up (with a “respawning too fast” error message) or will be unable to restart it if it really does die.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
https://github.com/apps/dco

norio-nomura:
Thanks! 🙏🏻

PR #2141, opened 2024-01-16
Reinstall apk packages on reboot

jandubois:
Because we are running from a RAM disk, any packages not installed from the ISO are removed after a reboot. They are still registered in `/etc/apk/world` (which is on `/mnt/data/etc/apk/world`), and the packages themselves are cached on `/mnt/data/apk/cache`, so reinstalling is quick and easy.

This worked fine in my testing (`apk fix` seems to reinstall the software, but not the config files in `/etc`), but I haven't found any documentation that guarantees that the config will not be touched).


AkihiroSuda (approved):
LGTM, but I think we should consider to just create a proper harddisk image and ditch the ISO

PR #2143, opened 2024-01-16
Fixup location of boot cdrom repository

jandubois:
The boot iso is sometimes mounted as /media/cdrom but more often as /media/sr0.

I haven't been able to figure out why.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
@@ -90,6 +90,14 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
 	done
 	# Make sure to re-mount any mount points under /tmp
 	mount -a
+	# The boot iso is sometimes mounted as /media/cdrom but more often as /media/sr0.
+	# This can change on each reboot.
+	for MEDIA in cdrom sr0; do
+		if [ -d "/media/${MEDIA}/apks" ]; then
+			sed -i "s#^/media/.*/apks#/media/$MEDIA/apks#" /etc/apk/repositories
+			break
+		fi
+	done
 	# Reinstall packages from /mnt/data/apk/cache into the RAM disk
 	apk fix --no-network
 fi


jandubois:
Note that this PR is mostly cosmetic, to get rid of this kind of warning:

```
WARNING: opening /media/cdrom/apks: No such file or directory
```

All the packages from the boot ISO have already been installed during the boot process, so are not really needed for normal use-cases. I still wanted to silence the warning though.

AkihiroSuda (approved):
Thanks, but I wonder why this can't be deterministic

jandubois:
> I wonder why this can't be deterministic

Me too, but I couldn't figure it out. It is probably some kind of race condition. It sometimes (maybe 1 out of 4) happens on the first boot:

```console
lima-alpine:~# mount | grep media
/dev/sr0 on /media/cdrom type iso9660 (ro,relatime,nojoliet,check=s,map=n,blocksize=2048,iocharset=utf8)
```

But on reboot it then looks like

```console
lima-alpine:~# mount | grep media
/dev/sr0 on /media/sr0 type iso9660 (ro,relatime,nojoliet,check=s,map=n,blocksize=2048,iocharset=utf8)
```

I don't think I've seen `/media/cdrom` being used after a reboot, only sometime on first boot. But I didn't keep proper notes to know for sure.

`/media/cdrom` is listed in `/etc/fstab`:

```console
lima-alpine:~# grep cdrom /etc/fstab
/dev/cdrom	/media/cdrom	iso9660	noauto,ro 0 0
```

I don't know how `/media/sr0` is being mounted, but if that happens first, then the mount of `/media/cdrom` will fail because they both mount the same device (`/dev/cdrom` is just a symlink to `/dev/sr0`):

```console
lima-alpine:~# ls -l /dev/cdrom
lrwxrwxrwx    1 root     root             3 Jan 15 20:25 /dev/cdrom -> sr0
```

We could try to remove the `/dev/cdrom` line from `/etc/fstab` in the ISO image. I suspect that we would then consistently see `/media/sr0` all the time, but I have no energy now to run further experiments. I can do this tomorrow, if you prefer that change to the ISO over adding this PR. Just let me know!

AkihiroSuda:
> We could try to remove the /dev/cdrom line from /etc/fstab in the ISO image. I suspect that we would then consistently see /media/sr0 all the time, but I have no energy now to run further experiments. I can do this tomorrow, if you prefer that change to the ISO over adding this PR. Just let me know!

Thanks, 👍 on consistent `/media/sr0`

jandubois:
I've confirmed with the below script that with an empty `/etc/fstab` I only ever see `/media/sr0`:

```shell
#!/usr/bin/env bash
set -xeu
count=0
while [[ "$count" -lt 100 ]]; do
    printf "\e[0;1;32mcount=%d\e[0m\n" "${count}"
    limactl delete -f alpine || true
    limactl start --tty=false template://alpine
    mp=$(limactl shell alpine cat /proc/mounts | awk '$1 == "/dev/sr0" {print $2}')
    if [[ $mp != /media/sr0 ]]; then
        echo Mountpoint is $mp
        exit 1
    fi
    count=$((count+1))
done
```

I was worried that the `/etc/fstab` would only be installed from the apkovl file after `/media/cdrom` has potentially already been mounted, but that doesn't seem to be the case (or it is unmounted again).

I've also run a modified version of that script that creates the VM only once, and then just stops and restarts it. It confirmed that `/media/cdrom` was only ever seen on first-boot.

PR #2145, opened 2024-01-17
Bump Alpine image to make sure /etc/fstab is empty

jandubois:
This avoids a race condition between mounting `/dev/sr0` to both `/media/cdrom` and `/media/sr0`.

See also https://github.com/lima-vm/lima/pull/2143#issuecomment-1893062016


AkihiroSuda (approved):
Thanks

PR #2146, opened 2024-01-17
limactl start: add `--foreground` flag

norio-nomura:
This change adds the `--foreground` flag to `limactl start`.

This flag:

- Launches `limactl hostagent` with `syscall.Exec`, which allows `limactl start --foreground` to be used in macOS's `launchd.plist`, making it possible to hide `limactl hostagent`.
- When launched from the terminal, it uses the terminal for standard output and standard error instead of the default `ha.std{out,err}.log` files. In this case, a message indicating that the default log files are not used is written to the log files.
- When launched from something other than the terminal, such as `launchd`, it uses the `ha.std{out,err}.log` files as before.
- Cannot be used on Windows.

The following are the updated steps to set up the Docker instance of Lima to start at login using launchd on macOS:
```bash
# Create the docker instance
limactl create template://docker --vm-type vz --rosetta --network vzNAT --tty=false

# Create the launchd plist
echo '<?xml version="1.0" encoding="UTF-8"?>
<plist version="1.0">
<dict>
	<key>Label</key>
	<string>io.lima-vm.docker</string>
	<key>ProgramArguments</key>
	<array>
		<string>'"$(which limactl)"'</string>
		<string>start</string>
		<string>--foreground</string>
		<string>docker</string>
	</array>
	<key>RunAtLoad</key>
	<true/>
	<key>StandardErrorPath</key>
	<string>launchd.stderr.log</string>
	<key>StandardOutPath</key>
	<string>launchd.stdout.log</string>
	<key>WorkingDirectory</key>
	<string>'"$(limactl list docker --format '{{.Dir}}')"'</string>
</dict>
</plist>' > ~/Library/LaunchAgents/io.lima-vm.docker.plist

# Load the launchd plist and launchd will start the docker instance
launchctl load ~/Library/LaunchAgents/io.lima-vm.docker.plist
```
that changed from #2140:

- Changed to `limactl create` because it's no longer necessary to run `limactl start` before loading into `launchd`.
- Changed from `limactl hostagent` to `limactl start --foreground`.
- Changed `Standard{Error,Out}Path` to `launchd.std{err,out}.log`, so that only the output of `limactl start` is written.

### DIFF
--- a/cmd/limactl/edit.go
+++ b/cmd/limactl/edit.go
@@ -122,7 +122,7 @@ func editAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	return start.Start(ctx, inst)
+	return start.Start(ctx, inst, false)
 }
 
 func askWhetherToStart() (bool, error) {
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/containerd/containerd/identifiers"
@@ -92,6 +93,9 @@ See the examples in 'limactl create --help'.
 		RunE:              startAction,
 	}
 	registerCreateFlags(startCommand, "[limactl create] ")
+	if runtime.GOOS != "windows" {
+		startCommand.Flags().Bool("foreground", false, "run the hostagent in the foreground")
+	}
 	startCommand.Flags().Duration("timeout", start.DefaultWatchHostAgentEventsTimeout, "duration to wait for the instance to be running before timing out")
 	return startCommand
 }
@@ -510,6 +514,14 @@ func startAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	launchHostAgentForeground := false
+	if runtime.GOOS != "windows" {
+		foreground, err := cmd.Flags().GetBool("foreground")
+		if err != nil {
+			return err
+		}
+		launchHostAgentForeground = foreground
+	}
 	timeout, err := cmd.Flags().GetDuration("timeout")
 	if err != nil {
 		return err
@@ -518,7 +530,7 @@ func startAction(cmd *cobra.Command, args []string) error {
 		ctx = start.WithWatchHostAgentTimeout(ctx, timeout)
 	}
 
-	return start.Start(ctx, inst)
+	return start.Start(ctx, inst, launchHostAgentForeground)
 }
 
 func createBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"syscall"
 	"text/template"
 	"time"
 
@@ -19,6 +20,7 @@ import (
 	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/qemu"
 	"github.com/lima-vm/lima/pkg/qemu/entitlementutil"
+	"github.com/mattn/go-isatty"
 
 	"github.com/lima-vm/lima/pkg/downloader"
 	"github.com/lima-vm/lima/pkg/fileutils"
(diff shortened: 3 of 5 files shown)


cmd/limactl/start.go
```diff
@@ -510,6 +514,15 @@ func startAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	if runtime.GOOS != "windows" {
+		foreground, err := cmd.Flags().GetBool("foreground")
+		if err != nil {
+			return err
+		}
+		if foreground {
+			ctx = start.WithLaunchHostAgentForeground(ctx, foreground)
```
AkihiroSuda, line 523: Is there any reason to use a context for this?
Wondering if we can just add an argument to `start.Start()`

cmd/limactl/start.go
```diff
@@ -92,6 +93,9 @@ See the examples in 'limactl create --help'.
 		RunE:              startAction,
 	}
 	registerCreateFlags(startCommand, "[limactl create] ")
+	if runtime.GOOS != "windows" {
+		startCommand.Flags().Bool("foreground", false, "run the instance in the foreground")
```
AkihiroSuda, line 97:
```suggestion
		startCommand.Flags().Bool("foreground", false, "run the hostagent in the foreground")
```

AkihiroSuda:
Please squash the commits

cmd/limactl/start.go
norio-nomura: There's no particular reason. I just mimicked the way the timeout flag is written.

AkihiroSuda: The timeout is contextualized because it has to be passed to many functions.
This is different from the foreground flag that is only used once.

norio-nomura: Understood, I'll try changing it.

cmd/limactl/edit.go
```diff
@@ -122,7 +122,7 @@ func editAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	return start.Start(ctx, inst)
+	return start.Start(ctx, inst, start.DefaultLaunchHostAgentForeground)
```
norio-nomura, line 125: Would it be better if the `--foreground` flag was also available for `limactl edit`?

pkg/start/start.go
```diff
@@ -33,6 +35,10 @@ import (
 // to be running before timing out.
 const DefaultWatchHostAgentEventsTimeout = 10 * time.Minute
 
+// DefaultLaunchHostAgentForeground is the default value of the boolean to use for
+// Start()
+const DefaultLaunchHostAgentForeground = false
```
AkihiroSuda, line 40: No need to define this const

AkihiroSuda (approved):
Thanks

pkg/start/start.go
```diff
@@ -194,7 +196,29 @@ func Start(ctx context.Context, inst *store.Instance) error {
 
 	begin := time.Now() // used for logrus propagation
 
-	if err := haCmd.Start(); err != nil {
+	if launchHostAgentForeground {
+		logrus.Info("Running the host agent in the foreground")
+		if isatty.IsTerminal(os.Stdin.Fd()) || isatty.IsCygwinTerminal(os.Stdin.Fd()) {
```
jandubois, line 201: I wonder if it is time to create a wrapper for this condition:

```console
$ ag IsCygwin
cmd/limactl/main.go
82:		if runtime.GOOS == "windows" && isatty.IsCygwinTerminal(os.Stdout.Fd()) {

cmd/limactl/list.go
171:		if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {

cmd/limactl/shell.go
176:	if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {

pkg/progressbar/progressbar.go
15:	if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {

pkg/qemu/entitlementutil/entitlementutil.go
88:		if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {
```

Not sure what the right package is, maybe `uiutil.isatty()`?

jandubois (approved):
Thanks, LGTM

pkg/start/start.go
AkihiroSuda: Yes, but it can be another PR

AkihiroSuda (approved):

norio-nomura:
Thanks! 🙏🏻

PR #2151, opened 2024-01-18
CLI flag to generate autostart files

### LINKED ISSUE #2142: Add CLI command to generate launchd units for macOS hosts (systemd units for Linux hosts)
It looks like launching `limactl start` via launchd is difficult and we have to let launchd start `limactl hostagent` directly.

The `limactl hostagent` command is still internal to Lima and has to be wrapped by a "public" CLI command.

The public command for generating the unit files can be like `limactl (create|edit|start) --autostart=<BOOL>`  .

- - -

ref:

> The method of invoking `limactl start docker` with `launchd` does not work.
> If `limactl start` starts `limactl hostagent` and terminates itself, `launchd` will detect the termination and terminate `limactl hostagent` and other processes in the same process group.
> Using `<key>AbandonProcessGroup</key><true/>` will prevent `limactl hostagent` from terminating by `launchd`, but will also prevent it from terminating on user logout.
> Therefore, I submitted this change because I thought it would be better to allow `launchd` to start and stop `limactl hostagent` directly.
>
>_Originally posted by @norio-nomura in https://github.com/lima-vm/lima/issues/2140#issuecomment-1892909533_

roman-kiselenko:
Added a new public CLI command: `limactl start-at-login INSTANCE --enabled`.
This command facilitates the generation of unit files for `launchd/systemd`, providing users with a straightforward way to control `limactl` autostart behavior.

Simplified Integration:

Adjusted `launchd/systemd` integration to start the `limactl hostagent` directly, simplifying the launch process.

Fix #2142

Partialy based on #2140

### DIFF
--- a/cmd/limactl/delete.go
+++ b/cmd/limactl/delete.go
@@ -5,7 +5,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"runtime"
 
+	"github.com/lima-vm/lima/pkg/autostart"
 	networks "github.com/lima-vm/lima/pkg/networks/reconcile"
 	"github.com/lima-vm/lima/pkg/stop"
 	"github.com/lima-vm/lima/pkg/store"
@@ -43,6 +45,14 @@ func deleteAction(cmd *cobra.Command, args []string) error {
 		}
 		if err := deleteInstance(cmd.Context(), inst, force); err != nil {
 			return fmt.Errorf("failed to delete instance %q: %w", instName, err)
+		}
+		if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+			deleted, err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName)
+			if err != nil && !errors.Is(err, os.ErrNotExist) {
+				logrus.WithError(err).Warnf("The autostart file for instance %q does not exist", instName)
+			} else if deleted {
+				logrus.Infof("The autostart file %q has been deleted", autostart.GetFilePath(runtime.GOOS, instName))
+			}
 		}
 		logrus.Infof("Deleted %q (%q)", instName, inst.Dir)
 	}
--- a/cmd/limactl/start-at-login.go
+++ b/cmd/limactl/start-at-login.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"errors"
+	"os"
+	"runtime"
+
+	"github.com/lima-vm/lima/pkg/autostart"
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func startAtLoginCommand() *cobra.Command {
+	startAtLoginCommand := &cobra.Command{
+		Use:               "start-at-login INSTANCE",
+		Short:             "Register/Unregister an autostart file for the instance",
+		Args:              WrapArgsError(cobra.MaximumNArgs(1)),
+		RunE:              startAtLoginAction,
+		ValidArgsFunction: startAtLoginComplete,
+		GroupID:           advancedCommand,
+	}
+
+	startAtLoginCommand.Flags().Bool(
+		"enabled", true,
+		"Automatically start the instance when the user logs in",
+	)
+
+	return startAtLoginCommand
+}
+
+func startAtLoginAction(cmd *cobra.Command, args []string) error {
+	instName := DefaultInstanceName
+	if len(args) > 0 {
+		instName = args[0]
+	}
+
+	inst, err := store.Inspect(instName)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			logrus.Infof("Instance %q not found", instName)
+			return nil
+		}
+		return err
+	}
+
+	flags := cmd.Flags()
+	startAtLogin, err := flags.GetBool("enabled")
+	if err != nil {
+		return err
+	}
+	if startAtLogin {
+		if err := autostart.CreateStartAtLoginEntry(runtime.GOOS, inst.Name, inst.Dir); err != nil {
+			logrus.WithError(err).Warnf("Can't create an autostart file for instance %q", inst.Name)
+		} else {
+			logrus.Infof("The autostart file %q has been created or updated", autostart.GetFilePath(runtime.GOOS, inst.Name))
+		}
+	} else {
+		deleted, err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName)
+		if err != nil {
+			logrus.WithError(err).Warnf("The autostart file %q could not be deleted", instName)
+		} else if deleted {
+			logrus.Infof("The autostart file %q has been deleted", autostart.GetFilePath(runtime.GOOS, instName))
+		}
+	}
+
+	return nil
+}
+
+func startAtLoginComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	return bashCompleteInstanceNames(cmd)
+}
--- a/pkg/autostart/autostart.go
+++ b/pkg/autostart/autostart.go
@@ -1,0 +1,119 @@
+// Package autostart manage start at login unit files for darwin/linux
+package autostart
+
+import (
+	_ "embed"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"path"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/lima-vm/lima/pkg/textutil"
+)
+
+//go:embed lima-vm@INSTANCE.service
+var systemdTemplate string
+
+//go:embed io.lima-vm.autostart.INSTANCE.plist
+var launchdTemplate string
+
+// CreateStartAtLoginEntry respect host OS arch and create unit file
+func CreateStartAtLoginEntry(hostOS, instName, workDir string) error {
+	unitPath := GetFilePath(hostOS, instName)
+	if _, err := os.Stat(unitPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+	tmpl, err := renderTemplate(hostOS, instName, workDir, os.Executable)
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(filepath.Dir(unitPath), os.ModePerm); err != nil {
+		return err
+	}
+	if err := os.WriteFile(unitPath, tmpl, 0o644); err != nil {
+		return err
+	}
+	return enableDisableService("enable", hostOS, GetFilePath(hostOS, instName))
+}
+
+// DeleteStartAtLoginEntry respect host OS arch and delete unit file
+// return true, nil if unit file has been deleted
+func DeleteStartAtLoginEntry(hostOS, instName string) (bool, error) {
+	unitPath := GetFilePath(hostOS, instName)
+	if _, err := os.Stat(unitPath); err != nil {
+		return false, err
+	}
+	if err := enableDisableService("disable", hostOS, GetFilePath(hostOS, instName)); err != nil {
+		return false, err
+	}
+	if err := os.Remove(unitPath); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// GetFilePath returns the path to autostart file with respect of host
+func GetFilePath(hostOS, instName string) string {
+	var fileTmpl string
+	if hostOS == "darwin" { // launchd plist
+		fileTmpl = fmt.Sprintf("%s/Library/LaunchAgents/io.lima-vm.autostart.%s.plist", os.Getenv("HOME"), instName)
+	}
+	if hostOS == "linux" { // systemd service
+		// Use instance name as argument to systemd service
+		// Instance name available in unit file as %i
+		xdgConfigHome := os.Getenv("XDG_CONFIG_HOME")
+		if xdgConfigHome == "" {
+			xdgConfigHome = filepath.Join(os.Getenv("HOME"), ".config")
+		}
+		fileTmpl = fmt.Sprintf("%s/systemd/user/lima-vm@%s.service", xdgConfigHome, instName)
+	}
+	return fileTmpl
+}
+
+func enableDisableService(action, hostOS, serviceWithPath string) error {
+	// Get filename without extension
+	filename := strings.TrimSuffix(path.Base(serviceWithPath), filepath.Ext(path.Base(serviceWithPath)))
+
+	var args []string
+	if hostOS == "darwin" {
+		// man launchctl
+		args = append(args, []string{
+			"launchctl",
+			action,
+			fmt.Sprintf("gui/%s/%s", strconv.Itoa(os.Getuid()), filename),
+		}...)
+	} else {
+		args = append(args, []string{
+			"systemctl",
+			"--user",
+			action,
+			filename,
+		}...)
+	}
+	cmd := exec.Command(args[0], args[1:]...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	return cmd.Run()
+}
+
+func renderTemplate(hostOS, instName, workDir string, getExecutable func() (string, error)) ([]byte, error) {
+	selfExeAbs, err := getExecutable()
+	if err != nil {
+		return nil, err
+	}
+	tmpToExecute := systemdTemplate
+	if hostOS == "darwin" {
+		tmpToExecute = launchdTemplate
+	}
+	return textutil.ExecuteTemplate(
+		tmpToExecute,
+		map[string]string{
+			"Binary":   selfExeAbs,
+			"Instance": instName,
+			"WorkDir":  workDir,
+		})
+}
--- a/pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
+++ b/pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
@@ -1,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Label</key>
+	<string>io.lima-vm.autostart.{{ .Instance }}</string>
+	<key>ProgramArguments</key>
+	<array>
+		<string>{{ .Binary }}</string>
+		<string>start</string>
+		<string>{{ .Instance }}</string>
+		<string>--foreground</string>
+	</array>
+	<key>RunAtLoad</key>
+	<true/>
+	<key>StandardErrorPath</key>
+	<string>launchd.stderr.log</string>
+	<key>StandardOutPath</key>
+	<string>launchd.stdout.log</string>
+	<key>WorkingDirectory</key>
+	<string>{{ .WorkDir }}</string>
+	<key>ProcessType</key>
+	<string>Background</string>
+</dict>
+</plist>
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -123,6 +123,10 @@ func newApp() *cobra.Command {
 		newProtectCommand(),
 		newUnprotectCommand(),
 	)
+	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+		rootCmd.AddCommand(startAtLoginCommand())
+	}
+
 	return rootCmd
 }
 
--- a/pkg/autostart/autostart_test.go
+++ b/pkg/autostart/autostart_test.go
@@ -1,0 +1,124 @@
+package autostart
+
+import (
+	"runtime"
+	"strings"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestRenderTemplate(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("skipping testing on windows host")
+	}
+	tests := []struct {
+		Name          string
+		InstanceName  string
+		HostOS        string
+		Expected      string
+		WorkDir       string
+		GetExecutable func() (string, error)
+	}{
+		{
+			Name:         "render darwin launchd plist",
+			InstanceName: "default",
+			HostOS:       "darwin",
+			Expected: `<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Label</key>
+	<string>io.lima-vm.autostart.default</string>
+	<key>ProgramArguments</key>
+	<array>
+		<string>/limactl</string>
+		<string>start</string>
+		<string>default</string>
+		<string>--foreground</string>
+	</array>
+	<key>RunAtLoad</key>
+	<true/>
+	<key>StandardErrorPath</key>
+	<string>launchd.stderr.log</string>
+	<key>StandardOutPath</key>
+	<string>launchd.stdout.log</string>
+	<key>WorkingDirectory</key>
+	<string>/some/path</string>
+	<key>ProcessType</key>
+	<string>Background</string>
+</dict>
+</plist>`,
+			GetExecutable: func() (string, error) {
+				return "/limactl", nil
+			},
+			WorkDir: "/some/path",
+		},
+		{
+			Name:         "render linux systemd service",
+			InstanceName: "default",
+			HostOS:       "linux",
+			Expected: `[Unit]
+Description=Lima - Linux virtual machines, with a focus on running containers.
+Documentation=man:lima(1)
+
+[Service]
+ExecStart=/limactl start %i --foreground
+WorkingDirectory=%h
+Type=simple
+TimeoutSec=10
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target`,
+			GetExecutable: func() (string, error) {
+				return "/limactl", nil
+			},
+			WorkDir: "/some/path",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.Name, func(t *testing.T) {
+			tmpl, err := renderTemplate(tt.HostOS, tt.InstanceName, tt.WorkDir, tt.GetExecutable)
+			assert.NilError(t, err)
+			assert.Equal(t, string(tmpl), tt.Expected)
+		})
+	}
+}
+
+func TestGetFilePath(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("skipping testing on windows host")
+	}
+	tests := []struct {
+		Name         string
+		HostOS       string
+		InstanceName string
+		HomeEnv      string
+		Expected     string
+	}{
+		{
+			Name:         "darwin with docker instance name",
+			HostOS:       "darwin",
+			InstanceName: "docker",
+			Expected:     "Library/LaunchAgents/io.lima-vm.autostart.docker.plist",
+		},
+		{
+			Name:         "linux with docker instance name",
+			HostOS:       "linux",
+			InstanceName: "docker",
+			Expected:     ".config/systemd/user/lima-vm@docker.service",
+		},
+		{
+			Name:         "empty with empty instance name",
+			HostOS:       "",
+			InstanceName: "",
+			Expected:     "",
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.Name, func(t *testing.T) {
+			assert.Check(t, strings.HasSuffix(GetFilePath(tt.HostOS, tt.InstanceName), tt.Expected))
+		})
+	}
+}
--- a/pkg/autostart/lima-vm@INSTANCE.service
+++ b/pkg/autostart/lima-vm@INSTANCE.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Lima - Linux virtual machines, with a focus on running containers.
+Documentation=man:lima(1)
+
+[Service]
+ExecStart={{.Binary}} start %i --foreground
+WorkingDirectory=%h
+Type=simple
+TimeoutSec=10
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target


pkg/start/start.go
```diff
@@ -296,6 +297,11 @@ func watchHostAgentEvents(ctx context.Context, inst *store.Instance, haStdoutPat
 			} else {
 				logrus.Infof("READY. Run `%s` to open the shell.", LimactlShellCmd(inst.Name))
 			}
+			if inst.StartAtLogin {
+				if err := autostart.CreateFile(runtime.GOOS); err != nil {
```
AkihiroSuda, line 301: This should happen on creating an instance, not on starting it

pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Label</key>
+	<string>io.lima-vm.{{ .INSTANCE }}</string>
+	<key>ProgramArguments</key>
+	<array>
+		<string>'"$(which limactl)"'</string>
```
AkihiroSuda, line 9: Does `$()` really work here?

pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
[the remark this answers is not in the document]
roman-kiselenko: OK, thanks for the tip 🙇

pkg/start/start.go
roman-kiselenko: So if VM wont start would be fine to create autostart file?

AkihiroSuda: Yes, should be fine

pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
jandubois: I don't think so. Everything should be filled in via Go templating; there should be no shell interpolation going on.

pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Label</key>
+	<string>io.lima-vm.{{ .INSTANCE }}</string>
```
jandubois, line 6: I would like to add an additional subdomain here:

```suggestion
	<string>io.lima-vm.autostart.{{ .INSTANCE }}</string>
```

Just to reserve part of the namespace if we want to create launch agents for other purposes in the future. I know we reserved instance names starting with `_`, so have some reserved space, but adding the `autostart` "subdomain" feels cleaner.

pkg/start/start.go
jandubois: How would I turn on/off autostart for an existing image. We could do this via `limactl edit`, but it will require the user to shut down the instance to enable/disable autostart, which feels clumsy.

roman-kiselenko: Ok, I'll move it to create action 👍

pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
```diff
@@ -0,0 +1,25 @@
...
+	<string>io.lima-vm.autostart.{{ .Instance }}</string>
+	<key>ProgramArguments</key>
+	<array>
+		<string>{{ .Binary }}</string>
+		<string>start</string>
+		<string>--foreground</string>
+		<string>--pidfile</string>
+		<string>ha.pid</string>
+		<string>--socket</string>
+		<string>ha.sock</string>
```
norio-nomura, line 15: The `limactl start` command does not have the `--pidfile` and `--socket` options.
```suggestion
```

pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
```diff
@@ -0,0 +1,25 @@
...
+		<string>{{ .Instance }}</string>
+	</array>
+	<key>RunAtLoad</key>
+	<true/>
+	<key>StandardErrorPath</key>
+	<string>ha.stderr.log</string>
+	<key>StandardOutPath</key>
+	<string>ha.stdout.log</string>
```
norio-nomura, line 23: The outputs of `limactl start` are lost when `limactl start` and `limactl hostagent` output to the same files, so they need to be separated.
```suggestion
	<key>StandardErrorPath</key>
	<string>launchd.stderr.log</string>
	<key>StandardOutPath</key>
	<string>launchd.stdout.log</string>
```

roman-kiselenko: Thank you for the tip 🙏

pkg/autostart/io.lima-vm.autostart.INSTANCE.plist
roman-kiselenko: Thank you for the tip 🙏

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,112 @@
...
+	filename := strings.TrimSuffix(filenameExt, filepath.Ext(filenameExt))
+
+	var args []string
+	if hostOS == "darwin" {
+		// man launchctl
+		serviceTarget := fmt.Sprintf("gui/%s/%s", strconv.Itoa(os.Getuid()), filename)
+		if action == "start" {
+			args = append(args, []string{"launchctl", "enable", serviceTarget}...)
+		} else {
+			args = append(args, []string{"launchctl", "disable", serviceTarget}...)
```
norio-nomura, line 81: `enable` / `disable` only enable/disable startup at the next login, they do not start/stop. To start, you need to `enable` and then `bootstrap`. To stop, you need to `bootout` and then `disable`, or use `load -w` / `unload -w`.

If it is disabled, `bootstrap` will fail, so `enable` it first.
```bash
launchctl enable gui/501/io.lima-vm.autostart.docker
launchctl bootstrap gui/501 ~/Library/LaunchAgents/io.lima-vm.autostart.docker.plist
```
The order of `bootout` and `disable` is irrelevant, as they do not interfere.
```bash
launchctl bootout gui/501 ~/Library/LaunchAgents/io.lima-vm.autostart.docker.plist
launchctl disable gui/501/io.lima-vm.autostart.docker
```

Are these combinations used differently in `limactl create`, `limactl start`, `limactl stop` with each `--autostart` flag? It seems to get complicated.
My modest hope is that it will start via `launchctl` when `limactl start` if it is registered in `launchd`.

roman-kiselenko: > Are these combinations used differently in limactl create, limactl start, limactl stop with each --autostart flag? 

[The autostart file creation](https://github.com/lima-vm/lima/pull/2151#discussion_r1458576167) happens only at the `limactl create` action and doesn't start right away. The `--autostart` flag works only at the create.

/cc @AkihiroSuda @jandubois wdyt?

AkihiroSuda: Yes, `limactl create --autostart` will create the instance and the systemd/launchd unit file but will leave the instance stopped.
`limactl start --autostart` should equate to `limactl create --autostart && limactl start`.

AkihiroSuda: If `--autostart` is confusing it can be renamed to something else

norio-nomura: What are the steps to register an already created instance as autostart?

AkihiroSuda: limactl edit --autostart

norio-nomura: Does `limactl edit --autostart=false` unregister it?

norio-nomura: Thanks! 🙏🏻

roman-kiselenko:
I guess the note in the documentation would be good since the behaviour of `--start-at-login` is quite complex. 🤔

roman-kiselenko:
Review request @lima-vm/maintainers

cmd/limactl/edit.go
```diff
@@ -51,12 +58,29 @@ func editAction(cmd *cobra.Command, args []string) error {
 		return errors.New("Cannot edit a running instance")
 	}
 
+	flags := cmd.Flags()
+
+	if flags.Lookup("start-at-login").Changed {
+		if startAtLogin, _ := flags.GetBool("start-at-login"); startAtLogin {
```
AkihiroSuda, line 64: Must check err

cmd/limactl/delete.go
```diff
@@ -44,6 +46,13 @@ func deleteAction(cmd *cobra.Command, args []string) error {
 		if err := deleteInstance(cmd.Context(), inst, force); err != nil {
 			return fmt.Errorf("failed to delete instance %q: %w", instName, err)
 		}
+		if runtime.GOOS != "windows" {
+			if err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName); err != nil {
+				logrus.Errorf("Cant delete autostart file error: %s", err)
```
AkihiroSuda, line 51: You may use `logrus.WithError(err)`

cmd/limactl/start.go
```diff
@@ -72,6 +73,9 @@ $ cat template.yaml | limactl create --name=local -
 		GroupID:           basicCommand,
 	}
 	registerCreateFlags(createCommand, "")
+	if runtime.GOOS != "windows" {
+		createCommand.Flags().Bool("start-at-login", false, "create the systemd/launchd unit file but will leave the instance stopped")
```
AkihiroSuda, line 77: This should be automatically inherited via `registerCreateFlags`

cmd/limactl/edit.go
```diff
@@ -30,6 +32,11 @@ func newEditCommand() *cobra.Command {
 		GroupID:           basicCommand,
 	}
 	editflags.RegisterEdit(editCommand)
+
+	if runtime.GOOS != "windows" {
+		editCommand.Flags().Bool("start-at-login", false, "register(--start-at-login=true)/unregister(--start-at-login=false) the systemd/launchd unit file for instance")
```
AkihiroSuda, line 37: This should be in `pkg/editutil`

cmd/limactl/start.go
```diff
@@ -536,6 +548,14 @@ func startAction(cmd *cobra.Command, args []string) error {
 		ctx = start.WithWatchHostAgentTimeout(ctx, timeout)
 	}
 
+	if ok, _ := cmd.Flags().GetBool("start-at-login"); ok {
+		if err := autostart.CreateStartAtLoginEntry(runtime.GOOS, inst.Name, inst.Dir); err != nil {
+			logrus.Errorf("Cant create autostart file error: %s", err)
+		} else {
+			logrus.Infof("The autostart file (%q) has been created", autostart.GetStartAtLoginEntryPath(runtime.GOOS, inst.Name))
+		}
+	}
```
AkihiroSuda, line 557: Probably this can be moved to `loadOrCreateInstance()`

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,129 @@
...
+// GetFilePath returns the path to autostart file with respect of host
+func GetStartAtLoginEntryPath(hostOS, instName string) string {
+	var fileTmp string
+	if hostOS == "darwin" { // launchd plist
+		fileTmp = fmt.Sprintf("Library/LaunchAgents/io.lima-vm.autostart.%s.plist", instName)
+	}
+	if hostOS == "linux" { // systemd service
+		// Use instance name as argument to systemd service
+		// Instance name available in unit file as %i
+		fileTmp = fmt.Sprintf(".config/systemd/user/lima-vm@%s.service", instName)
```
AkihiroSuda, line 82: `$XDG_CONFIG_HOME` should be respected

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,129 @@
...
+		return err
+	}
+	return os.Remove(unitPath)
+}
+
+// GetFilePath returns the path to autostart file with respect of host
+func GetStartAtLoginEntryPath(hostOS, instName string) string {
+	var fileTmp string
+	if hostOS == "darwin" { // launchd plist
+		fileTmp = fmt.Sprintf("Library/LaunchAgents/io.lima-vm.autostart.%s.plist", instName)
```
AkihiroSuda, line 77: What should we do when an instance name contains a period?

jandubois: I think that is fine; it should just work.

roman-kiselenko: The instance name is validated here:

https://github.com/lima-vm/lima/blob/752afc08689e06d00293b3261817d5be70cb0148/pkg/store/store.go#L62

```go
const (
	maxLength  = 76
	alphanum   = `[A-Za-z0-9]+`
	separators = `[._-]`
)
```

cmd/limactl/edit.go
AkihiroSuda: Sorry, meant editflags

https://github.com/lima-vm/lima/blob/master/cmd/limactl/editflags/editflags.go

roman-kiselenko:
Review request @lima-vm/maintainers

.github/workflows/test.yml
```diff
@@ -398,7 +398,7 @@ jobs:
     - name: Install test dependencies
       run: brew install qemu bash coreutils
     - name: Test
-      uses: nick-invision/retry@v2
+      uses: nick-invision/retry@v3
```
AkihiroSuda, line 401: Doesn't seem relevant?

examples/apptainer-rootful.yaml
```diff
@@ -4,12 +4,12 @@
 
 images:
 # Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
-- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20231211/ubuntu-22.04-server-cloudimg-amd64.img"
+- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20240126/ubuntu-22.04-server-cloudimg-amd64.img"
```
AkihiroSuda, line 7: Doesn't seem relevant?

AkihiroSuda:
The code looks good, but the commit seems to contain irrelevant changes.
`git rebase master` will probably fix the issue.

roman-kiselenko:
> The code looks good, but the commit seems to contain irrelevant changes. `git rebase master` will probably fix the issue.

yep, forgot to rebase 🙇‍♂️

AkihiroSuda (dismissed):
Thanks

cmd/limactl/delete.go
```diff
@@ -44,6 +46,13 @@ func deleteAction(cmd *cobra.Command, args []string) error {
 		if err := deleteInstance(cmd.Context(), inst, force); err != nil {
 			return fmt.Errorf("failed to delete instance %q: %w", instName, err)
 		}
+		if runtime.GOOS != "windows" {
+			if err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName); err != nil {
+				logrus.WithError(err)
+			} else {
+				logrus.Infof("The autostart file (%q) has been deleted", autostart.GetStartAtLoginEntryPath(runtime.GOOS, instName))
```
jandubois, line 53: I find it confusing that `limactl delete` will claim it deleted the autostart file even if it didn't exist to begin with.

I think `autostart.DeleteStartAtLoginEntry` should return `os.ErrNotExist` in that case, so you can skip any logging at all here.

cmd/limactl/edit.go
```diff
@@ -51,12 +54,33 @@ func editAction(cmd *cobra.Command, args []string) error {
 		return errors.New("Cannot edit a running instance")
 	}
 
+	flags := cmd.Flags()
+
+	if runtime.GOOS != "windows" && flags.Lookup("start-at-login").Changed {
+		startAtLogin, err := flags.GetBool("start-at-login")
```
jandubois, line 60: How is this to be used?

```console
$ l edit --start-at-login=true
INFO[0000] The autostart file ("/Users/jan/Library/LaunchAgents/io.lima-vm.autostart.default.plist") has been created
Waiting for Emacs...
INFO[0004] Aborting, no changes made to the instance
```

So it asks me to edit `lima.yaml`, and if I don't make some unrelated change, it aborts the operation.

Compare with

```console
$ l edit --cpus 3
INFO[0000] Instance "default" configuration edited
FATA[0004] interrupt
```

Somehow the `--start-at-login` option doesn't really fit in with the `edit` subcommand because it doesn't change any `lima.yaml` settings.

Also `edit` doesn't work on running instance, which is not a sensible limitation for enabling/disabling autostart:

```console
$ l edit --start-at-login
FATA[0000] Cannot edit a running instance
```

So maybe this should be a separate subcommand instead of being shoehorned into `edit`?

cmd/limactl/editflags/editflags.go
```diff
@@ -60,6 +60,10 @@ func registerEdit(cmd *cobra.Command, commentPrefix string) {
 
 	flags.String("set", "", commentPrefix+"modify the template inplace, using yq syntax")
 
+	if runtime.GOOS != "windows" {
+		flags.Bool("start-at-login", false, commentPrefix+"register(--start-at-login=true)/unregister(--start-at-login=false) the systemd/launchd unit file for instance")
```
jandubois, line 64: We should print just `systemd` on Linux and just `launchd` on macOS. This message is already too long, and adding words that don't apply to this host is not helpful for unexperienced users.

cmd/limactl/delete.go
roman-kiselenko: `DeleteStartAtLoginEntry` [returns `nil`](https://github.com/lima-vm/lima/pull/2151/files#diff-7c73aad62568bcacc05068a3c8cdf050206daaf75d7d1f2416c8faa19d385b32R62) if the file is absent.

cmd/limactl/edit.go
roman-kiselenko: So the `--start-at-login` flag might be present in the `lima.yaml`? How do you see this? I'm a bit confused with the usage flow.

cmd/limactl/delete.go
jandubois: I know, that is the problem. When I delete the instance, I get

```
$ l delete default
INFO[0000] The qemu driver process seems already stopped
INFO[0000] Removing /Users/jan/.lima/default under "*.pid *.sock *.tmp"
INFO[0000] Removing "/Users/jan/.lima/default/ha.sock"
INFO[0000] The autostart file ("/Users/jan/Library/LaunchAgents/io.lima-vm.autostart.default.plist") has been deleted
INFO[0000] Deleted "default" ("/Users/jan/.lima/default")
```

It claims to have deleted the autostart file, even though the file didn't exist. So the log entry is plain wrong.

That's why it needs to report an error back to the caller when the file doesn't exist, so the caller doesn't create a bogus log entry.

roman-kiselenko: Fair enough, I'll fix it, thanks 🙏

cmd/limactl/edit.go
jandubois: Not sure; needs more discussion and feedback from other @lima-vm/maintainers.

Adding a `startAtLogin` setting to `lima.yaml` would fix part of the problem, but still wouldn't allow you to change the setting without stopping the instance first, which feels like an unnatural limitation.

`start-at-login` also doesn't feel like it is a configuration setting for the instance: it is not used at all for starting or configuring the instance; it is a configuration for the host on which the instance exists. Maybe that is too subtle a difference, but I think that is the reason why the semantics are different from other `lima.yaml` settings, and why it may not belong in there.

E.g. I can create a clone of the default instance with

```console
$ l create --name clone ~/.lima/default/lima.yaml
```

Should this copy the `--start-at-login` setting from the default instance too? I would think not, which is another sign that maybe the setting doesn't belong into `lima.yaml`.

Sorry, I don't have a constructive suggestion on what we should do; I need to think about it some more. But I would love to hear other people's opinions!

cmd/limactl/delete.go
roman-kiselenko: The [`DeleteStartAtLoginEntry`](DeleteStartAtLoginEntry) now return an additional variable if file has been deleted. WDYT?

cmd/limactl/edit.go
jandubois: This is just a suggestion for discussion; let's get consensus before you make any code changes:

A new command will create/delete the start-at-login files:

```
limactl start-at-login [INSTANCE] [--enabled]
```

This will create the file unless `--enabled=false` is specified, in which case it will be deleted. `INSTANCE` defaults to `default`, as usual.

The command can be executed as long as `INSTANCE` exists; it doesn't matter what state it is in.

`limactl create ... --start-at-login` could still be supported as a convenience, but it is unnecessary complexity[^1] for extremely minor benefit. I would prefer to not have it (I would want to keep it for compatibility if we had already released it, but we didn't).

[^1]: I'm talking about the user experience complexity, not the implementation complexity. The more options each command gets, the more overwhelming it becomes for new users to figure out what is important, and what isn't.

`limactl delete` will still delete the unit file, as it does right now.

What do you think?

roman-kiselenko: @jandubois 

Thank you for your thoughtful suggestion and for initiating this discussion.

Introducing a new command to manage `start-at-login` files could offer clarity and consistency in our workflow. Your proposed command structure seems intuitive and aligns well with our existing conventions.

The current implementation seems to have a drift with `lima.yaml` config file. In my view, the `create\start\edit` works primarily with the config file, and an additional flag should be included in `lima.yaml` in order to follow the internal flow.

jandubois:
> Introducing a new command to manage start-at-login files could offer clarity and consistency in our workflow. Your proposed command structure seems intuitive and aligns well with our existing conventions.


I was hoping to get some opinions from the other @lima-vm/maintainers; pinging them once more.

If they are too busy, then I would suggest you go with the suggested additional command and remove the `--start-at-login` option from the create/start/edit commands, as discussed above.

> The current implementation seems to have a drift with lima.yaml config file. In my view, the create\start\edit works primarily with the config file, and an additional flag should be included in lima.yaml in order to follow the internal flow.

I don't see why this would be needed/useful at all if you switch to a separate command for configuring start-at-login. The model becomes very simple then:

You create the file with `limactl start-at-login`, and you delete it with `limactl start-at-login --disable=true` (or implicitly via `limactl delete`), and that's it.

There is no need to store an extra setting anywhere at all; you just create or delete a file on demand.

This has the added advantage that you don't need to stop the instance to create or delete the file; a limitation of the current edit mechanism that makes no sense here.

balajiv113:
> Introducing a new command to manage start-at-login files could offer clarity and consistency in our workflow. Your proposed command structure seems intuitive and aligns well with our existing conventions.

I too agree towards maintaining a separate command like (start-at-login / autostart), this makes things clean and meaningful.

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,137 @@
...
+	return true, nil
+}
+
+// GetFilePath returns the path to autostart file with respect of host
+func GetStartAtLoginEntryPath(hostOS, instName string) string {
+	var fileTmp string
+	if hostOS == "darwin" { // launchd plist
+		fileTmp = fmt.Sprintf("%s/Library/LaunchAgents/io.lima-vm.autostart.%s.plist", os.Getenv("HOME"), instName)
+	}
+	if hostOS == "linux" { // systemd service
```
balajiv113, line 83: Can we move these if conditions to separate os specific files.

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return err
+	}
+	if err := os.WriteFile(unitPath, tmpl, 0o644); err != nil {
+		// Early return if error is unexpected
+		if !errors.Is(err, os.ErrExist) && !errors.Is(err, os.ErrNotExist) {
+			return err
+		}
+		// Unit file exists, return nil
+		if errors.Is(err, os.ErrExist) {
```
balajiv113, line 40: Shouldn't we overwrite ??

If for some reason, there is a template data change, we might need to overwrite and reinstall the service right ?

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,137 @@
...
+			return nil
+		}
+		// .config/systemd folder might be unavailable
+		if errors.Is(err, os.ErrNotExist) && hostOS == "linux" {
+			if err := os.MkdirAll(filepath.Dir(unitPath), os.ModePerm); err != nil {
```
balajiv113, line 45: Can we do this create all directories on os.stat failure itself. This way a single writeFile should be fine

pkg/autostart/autostart.go
roman-kiselenko: I dont get, could you explain?

pkg/autostart/autostart.go
roman-kiselenko: I don't think it make code more readable, there is not so much code to split in separate files.  Code in one file is much easy to maintain.

pkg/autostart/autostart.go
jandubois: I think @balajiv113 means that we should simply overwrite the file, even if it already exists.

But I believe `os.WriteFile` will already do that, so the only way we could get `os.ErrExist` would be if the file exists, but we don't have permission to write.

In which case I think an error should be reported because the content may now not be what we expect it to be.

Although this case should be rather obscure: (1) someone must have intentionally removed the write permission, and (2) something like the Lima install location must have changed, so the previous content would no longer be valid.

I would just remove all the special logic for `os.ErrExist` and return it like any other error.

pkg/autostart/autostart.go
jandubois: I would unconditionally call `MkdirAll` before `WriteFile`, just to simplify the logic. The call is cheap, and we only do it once. so no need trying to be clever.

roman-kiselenko:
Move to separate command and follow comments, I guess now it's ready.

@lima-vm/maintainers

cmd/limactl/start-at-login.go
```diff
@@ -0,0 +1,72 @@
...
+		Short:             "Register/Unregister the unit file for instance",
+		Args:              WrapArgsError(cobra.MaximumNArgs(1)),
+		RunE:              startAtLoginAction,
+		ValidArgsFunction: startAtLoginComplete,
+		GroupID:           advancedCommand,
+	}
+
+	resetCommand.Flags().Bool(
+		"enabled", true,
+		"--enabled=true/false the "+map[string]string{"darwin": "launchd", "linux": "systemd"}[runtime.GOOS]+" unit file for instance",
```
AkihiroSuda, line 26:
```suggestion
		"the "+map[string]string{"darwin": "launchd", "linux": "systemd"}[runtime.GOOS]+" unit file for instance",
```

AkihiroSuda: The flag shouldn't be added for Windows, NetBSD, etc

roman-kiselenko: Flag will be [added only on linux or darwin](https://github.com/lima-vm/lima/pull/2151/commits/d66089e94c84446a04b0d7498139c801123ec1aa#diff-95475f7f5560ba438d76bce7e63eb2f2d63b59438023d2af973e941510cdcf9cR126)

AkihiroSuda (approved):
Thanks

cmd/limactl/start-at-login.go
```diff
@@ -0,0 +1,72 @@
...
+	flags := cmd.Flags()
+	startAtLogin, err := flags.GetBool("enabled")
+	if err != nil {
+		return err
+	}
+	if startAtLogin {
+		if err := autostart.CreateStartAtLoginEntry(runtime.GOOS, inst.Name, inst.Dir); err != nil {
+			logrus.WithError(err)
+		} else {
+			logrus.Infof("The autostart file (%q) has been created", autostart.GetStartAtLoginEntryPath(runtime.GOOS, inst.Name))
```
jandubois, line 56: I don't think we need a separate message, but include the possibility that the file already existed, and has just been updated (which doesn't happen right now; see below).

```suggestion
			logrus.Infof("The autostart file (%q) has been created or updated", autostart.GetStartAtLoginEntryPath(runtime.GOOS, inst.Name))
```

cmd/limactl/start-at-login.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+	} else {
+		deleted, err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName)
+		if err != nil {
+			logrus.WithError(err)
+		} else if deleted {
+			logrus.Infof("The autostart file (%q) has been deleted", autostart.GetStartAtLoginEntryPath(runtime.GOOS, instName))
+		}
```
jandubois, line 64: I think we should have another `else` clause here and log that the autostart file didn't exist, so the user doesn't worry about why there is no feedback:

```console
$ l start-at-login alpine
INFO[0000] The autostart file ("/Users/jan/Library/LaunchAgents/io.lima-vm.autostart.alpine.plist") has been created
$ l start-at-login alpine --enabled=false
INFO[0000] The autostart file ("/Users/jan/Library/LaunchAgents/io.lima-vm.autostart.alpine.plist") has been deleted
$ l start-at-login alpine --enabled=false
```

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,125 @@
...
+
+//go:embed io.lima-vm.autostart.INSTANCE.plist
+var launchdTemplate string
+
+// CreateStartAtLoginEntry respect host OS arch and create unit file
+func CreateStartAtLoginEntry(hostOS, instName, workDir string) error {
+	unitPath := GetStartAtLoginEntryPath(hostOS, instName)
+	if _, err := os.Stat(unitPath); !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
```
jandubois, line 29: I don't think we should simply return if the file already exists: Maybe the user has re-installed Lima to a different location and wants to update the autostart file so it continues to work.

```suggestion
	if _, err := os.Stat(unitPath); err != nil && !errors.Is(err, os.ErrNotExist) {
		return err
	}
```

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,125 @@
...
+	tmpl, err := renderTemplate(hostOS, workDir, instName, os.Executable)
+	if err != nil {
+		return err
+	}
+	if err := os.MkdirAll(filepath.Dir(unitPath), os.ModePerm); err != nil {
+		return err
+	}
+	if err := os.WriteFile(unitPath, tmpl, 0o644); err != nil {
+		if !errors.Is(err, os.ErrExist) && !errors.Is(err, os.ErrNotExist) {
```
jandubois, line 38: Why are these 2 errors excluded? I don't think either of them would normally occur in this situation, but if they do, don't we want to report them?

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,125 @@
...
+	filename := strings.TrimSuffix(path.Base(serviceWithPath), filepath.Ext(path.Base(serviceWithPath)))
+
+	var args []string
+	if hostOS == "darwin" {
+		// man launchctl
+		args = append(args, []string{
+			"launchctl",
+			action,
+			fmt.Sprintf("gui/%s/%s", strconv.Itoa(os.Getuid()), filename),
+		}...)
```
jandubois, line 94: I don't think this code is necessary; the LaunchAgent will be enabled by default when you create the plist, and disabled when you delete it.

I believe the explicit enable/disable is only needed when you want to disable an agent without deleting the plist.

On the other hand, having the code shouldn't hurt.

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,125 @@
...
+//go:embed io.lima-vm.autostart.INSTANCE.plist
+var launchdTemplate string
+
+// CreateStartAtLoginEntry respect host OS arch and create unit file
+func CreateStartAtLoginEntry(hostOS, instName, workDir string) error {
+	unitPath := GetStartAtLoginEntryPath(hostOS, instName)
+	if _, err := os.Stat(unitPath); !errors.Is(err, os.ErrNotExist) {
+		return err
+	}
+	tmpl, err := renderTemplate(hostOS, workDir, instName, os.Executable)
```
jandubois, line 30: The arguments are in the wrong order:

```suggestion
	tmpl, err := renderTemplate(hostOS, instName, workDir, os.Executable)
```

This results in a broken plist file:

```xml
<dict>
        <key>Label</key>
        <string>io.lima-vm.autostart./Users/jan/.lima/alpine</string>
        …
        <key>WorkingDirectory</key>
        <string>alpine</string>
        …
</dict>
```

jandubois (changes requested):
Most of the feedback is minor, but the incorrect order of the template rendering arguments is a blocker.

pkg/autostart/autostart.go
roman-kiselenko: I vote to keep this code because it clarifies exactly what happens.

roman-kiselenko:
> Most of the feedback is minor, but the incorrect order of the template rendering arguments is a blocker.

Thanks for your review 🙏  
Found some bugs

AkihiroSuda (approved):
Thanks

cmd/limactl/start-at-login.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	flags := cmd.Flags()
+	startAtLogin, err := flags.GetBool("enabled")
+	if err != nil {
+		return err
+	}
+	if startAtLogin {
+		if err := autostart.CreateStartAtLoginEntry(runtime.GOOS, inst.Name, inst.Dir); err != nil {
+			logrus.WithError(err).Warnf("can't create a unit file for instance (%q)", inst.Name)
```
jandubois, line 54:
```suggestion
			logrus.WithError(err).Warnf("Can't create an autostart file for instance %q", inst.Name)
```

cmd/limactl/start-at-login.go
```diff
@@ -0,0 +1,72 @@
...
+		} else {
+			logrus.Infof("The autostart file (%q) has been created or updated", autostart.GetStartAtLoginEntryPath(runtime.GOOS, inst.Name))
+		}
+	} else {
+		deleted, err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName)
+		if err != nil {
+			logrus.WithError(err).Warnf("file for instance (%q) not exists", instName)
```
jandubois, line 61:
```suggestion
			logrus.WithError(err).Warnf("The autostart file for instance %q does not exist", instName)
```

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,119 @@
...
+//go:embed lima-vm@INSTANCE.service
+var systemdTemplate string
+
+//go:embed io.lima-vm.autostart.INSTANCE.plist
+var launchdTemplate string
+
+// CreateStartAtLoginEntry respect host OS arch and create unit file
+func CreateStartAtLoginEntry(hostOS, instName, workDir string) error {
+	unitPath := GetStartAtLoginEntryPath(hostOS, instName)
+	if _, err := os.Stat(unitPath); err != nil && !errors.Is(err, os.ErrNotExist) && !errors.Is(err, os.ErrExist) {
```
jandubois, line 27: I don't think `os.Stat` will ever return `os.ErrExist`. If the file at the given path exists, it will return `nil`.

`os.Mkdir` is the only function I can think of right now that will return `os.ErrExist` if the directory already exists. Are there any others?

```suggestion
	if _, err := os.Stat(unitPath); err != nil && !errors.Is(err, os.ErrNotExist) {
```

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,119 @@
...
+	if err := enableDisableService("disable", hostOS, GetStartAtLoginEntryPath(hostOS, instName)); err != nil {
+		return false, err
+	}
+	if err := os.Remove(unitPath); err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// GetFilePath returns the path to autostart file with respect of host
```
jandubois, line 59:
```suggestion
// GetStartAtLoginEntryPath returns the path to autostart file with respect of host
```

Given that we call this the "autostart file", maybe this function should be called `GetStartAtLoginFileName`? I'm not sure what `EntryPath` means.

pkg/autostart/autostart.go
```diff
@@ -0,0 +1,119 @@
...
+func GetStartAtLoginEntryPath(hostOS, instName string) string {
+	var fileTmp string
```
jandubois, line 61: Why does the variable name have a `Tmp` suffix? In which way is the file temporary? Should this be just `file` or `fileName`?

jandubois (changes requested):
I think the PR is almost ready for merging.

The only functional change I would like to see is the removal of the error message when an instance is deleted and no autostart file exists. That is not an error, but expected.

Otherwise I found some conditionals that seem to be always true/false and therefore can be eliminated.

The rest of the feedback is just nitpicking about wording/formatting of the log messages and help text.

pkg/autostart/autostart.go
roman-kiselenko: `Tmp` here is a file with template, not a temporary file.

pkg/autostart/autostart.go
roman-kiselenko: > I'm not sure what EntryPath means.

It means an entry of some file. `GetFilePath` doesn't say what kind of file path this function returns.

pkg/autostart/autostart.go
AkihiroSuda: It should be called Tmpl then?

roman-kiselenko: or `fileTemple` 😄

cmd/limactl/delete.go
```diff
@@ -44,6 +46,14 @@ func deleteAction(cmd *cobra.Command, args []string) error {
 		if err := deleteInstance(cmd.Context(), inst, force); err != nil {
 			return fmt.Errorf("failed to delete instance %q: %w", instName, err)
 		}
+		if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
+			deleted, err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName)
+			if err != nil && !errors.Is(err, os.ErrNotExist) {
+				logrus.WithError(err).Warnf("The autostart file for instance %q does not exist", instName)
+			} else if deleted {
+				logrus.Infof("The autostart file for instance %q could not be deleted", autostart.GetFilePath(runtime.GOOS, instName))
```
jandubois, line 54: The log message is backwards, it is supposed to confirm that the file has been deleted:

```suggestion
				logrus.Infof("The autostart file %q has been deleted", autostart.GetFilePath(runtime.GOOS, instName))
```

cmd/limactl/start-at-login.go
```diff
@@ -0,0 +1,72 @@
...
+	flags := cmd.Flags()
+	startAtLogin, err := flags.GetBool("enabled")
+	if err != nil {
+		return err
+	}
+	if startAtLogin {
+		if err := autostart.CreateStartAtLoginEntry(runtime.GOOS, inst.Name, inst.Dir); err != nil {
+			logrus.WithError(err).Warnf("Can't create an autostart file for instance %q", inst.Name)
+		} else {
+			logrus.Infof("The autostart file (%q) has been created or updated", autostart.GetFilePath(runtime.GOOS, inst.Name))
```
jandubois, line 56:
```suggestion
			logrus.Infof("The autostart file %q has been created or updated", autostart.GetFilePath(runtime.GOOS, inst.Name))
```

cmd/limactl/start-at-login.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+	} else {
+		deleted, err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName)
+		if err != nil {
+			logrus.WithError(err).Warnf("The autostart file for instance %q does not exist", instName)
+		} else if deleted {
+			logrus.Infof("The autostart file (%q) has been deleted", autostart.GetFilePath(runtime.GOOS, instName))
```
jandubois, line 63:
```suggestion
			logrus.Infof("The autostart file %q has been deleted", autostart.GetFilePath(runtime.GOOS, instName))
```

cmd/limactl/start-at-login.go
```diff
@@ -0,0 +1,72 @@
...
+	if startAtLogin {
+		if err := autostart.CreateStartAtLoginEntry(runtime.GOOS, inst.Name, inst.Dir); err != nil {
+			logrus.WithError(err).Warnf("Can't create an autostart file for instance %q", inst.Name)
+		} else {
+			logrus.Infof("The autostart file (%q) has been created or updated", autostart.GetFilePath(runtime.GOOS, inst.Name))
+		}
+	} else {
+		deleted, err := autostart.DeleteStartAtLoginEntry(runtime.GOOS, instName)
+		if err != nil {
+			logrus.WithError(err).Warnf("The autostart file for instance %q does not exist", instName)
```
jandubois, line 61:
```suggestion
			logrus.WithError(err).Warnf("The autostart file %q could not be deleted: %w", autostart.GetFilePath(runtime.GOOS, instName), err)
```

jandubois (changes requested):
I think 2 log messages are not correct yet.

The other 2 comments are just removing the parens in `(%q)` for consistency.

Otherwise it looks good to go.

roman-kiselenko:
> I think 2 log messages are not correct yet.
> 
> The other 2 comments are just removing the parens in `(%q)` for consistency.
> 
> Otherwise it looks good to go.

@jandubois Thank you for your careful review 🙏

jandubois (approved):
Thank you, looks great!

AkihiroSuda (approved):

PR #2155, opened 2024-01-19
CI: use standard GHA runners

AkihiroSuda:
"Larger" runners are no longer required for nested virt with Linux


jandubois:
> "Larger" runners are no longer required for nested virt with Linux

Is this change documented anywhere by Github?

AkihiroSuda:
> > "Larger" runners are no longer required for nested virt with Linux
> 
> Is this change documented anywhere by Github?

No, but relates to https://github.blog/2024-01-17-github-hosted-runners-double-the-power-for-open-source/

PR #2159, opened 2024-01-23
Add `excludeFromBackup` on macOS

norio-nomura:
```yml
# Exclude from backup: "all", "disks" or "none"
# Specify which file in the instance's configuration to exclude from backup.
# Only available on macOS.
# 🟢 Builtin default: "disks"
excludeFromBackup: null
```

Sets the [`NSURLIsExcludedFromBackupKey`](https://developer.apple.com/documentation/foundation/nsurlisexcludedfrombackupkey) attribute to:
- "all": the instance directory
- "disks": `basedisk` and `diffdisk` files
- "none": none

### DIFF
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -303,6 +303,16 @@ func Validate(y LimaYAML, warn bool) error {
 	if err := validateNetwork(y, warn); err != nil {
 		return err
 	}
+
+	switch *y.ExcludeFromBackup {
+	case ExcludeFromBackupAll, ExcludeFromBackupDisks, ExcludeFromBackupNone:
+	default:
+		return fmt.Errorf("field `excludeFromBackup` must be %q, %q, or %q; got %q", ExcludeFromBackupAll, ExcludeFromBackupDisks, ExcludeFromBackupNone, *y.ExcludeFromBackup)
+	}
+	if warn && runtime.GOOS != "darwin" && *y.ExcludeFromBackup != "none" {
+		logrus.Warn("field `excludeFromBackup` is only supported on macOS")
+	}
+
 	if warn {
 		warnExperimental(y)
 	}
--- a/pkg/start/start.go
+++ b/pkg/start/start.go
@@ -70,6 +70,26 @@ func ensureNerdctlArchiveCache(y *limayaml.LimaYAML, created bool) (string, erro
 	return "", fileutils.Errors(errs)
 }
 
+// applyExcludeFromBackup sets the NSURLIsExcludedFromBackupKey attribute of the instance directory.
+func applyExcludeFromBackup(instDir string, excludeFromBackup limayaml.ExcludeFromBackup) {
+	baseDisk := filepath.Join(instDir, filenames.BaseDisk)
+	diffDisk := filepath.Join(instDir, filenames.DiffDisk)
+	var targets map[string]bool
+	switch excludeFromBackup {
+	case limayaml.ExcludeFromBackupAll:
+		// Exclude the instance directory from Time Machine backups
+		targets = map[string]bool{instDir: true, baseDisk: false, diffDisk: false}
+	case limayaml.ExcludeFromBackupDisks:
+		// Exclude disk files from Time Machine backup.
+		targets = map[string]bool{instDir: false, baseDisk: true, diffDisk: true}
+	default: // limayaml.ExcludeFromBackupNone
+		targets = map[string]bool{instDir: false, baseDisk: false, diffDisk: false}
+	}
+	for target, exclude := range targets {
+		osutil.SetExcludeFromBackup(target, exclude)
+	}
+}
+
 type Prepared struct {
 	Driver              driver.Driver
 	NerdctlArchiveCache string
@@ -108,6 +128,7 @@ func Prepare(ctx context.Context, inst *store.Instance) (*Prepared, error) {
 	if err != nil {
 		return nil, err
 	}
+	applyExcludeFromBackup(inst.Dir, *y.ExcludeFromBackup)
 
 	return &Prepared{
 		Driver:              limaDriver,
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -265,6 +265,12 @@ rosetta:
   # 🟢 Builtin default: false
   binfmt: null
 
+# Exclude from backup: "all", "disks" or "none"
+# Specify which file in the instance's configuration to exclude from backup.
+# Only available on macOS.
+# 🟢 Builtin default: "disks"
+excludeFromBackup: null
+
 # Specify the timezone name (as used by the zoneinfo database). Specify the empty string
 # to not set a timezone in the instance.
 # 🟢 Builtin default: use name from /etc/timezone or deduce from symlink target of /etc/localtime
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -753,6 +753,20 @@ func FillDefault(y, d, o *LimaYAML, filePath string) {
 		y.Rosetta.BinFmt = ptr.Of(false)
 	}
 
+	if runtime.GOOS == "darwin" {
+		if y.ExcludeFromBackup == nil {
+			y.ExcludeFromBackup = d.ExcludeFromBackup
+		}
+		if o.ExcludeFromBackup != nil {
+			y.ExcludeFromBackup = o.ExcludeFromBackup
+		}
+		if y.ExcludeFromBackup == nil {
+			y.ExcludeFromBackup = ptr.Of(ExcludeFromBackupDisks)
+		}
+	} else {
+		y.ExcludeFromBackup = ptr.Of(ExcludeFromBackupNone)
+	}
+
 	if y.Plain == nil {
 		y.Plain = d.Plain
 	}
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -122,6 +122,11 @@ func TestFillDefault(t *testing.T) {
 			}
 		}
 	}
+	if runtime.GOOS == "darwin" {
+		builtin.ExcludeFromBackup = ptr.Of(ExcludeFromBackupDisks)
+	} else {
+		builtin.ExcludeFromBackup = ptr.Of(ExcludeFromBackupNone)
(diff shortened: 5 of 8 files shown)


jandubois:
I'm curious what this setting does. I assumed `NSURLIsExcludedFromBackupKey` would only apply to files under `~/Documents` and would prevent them from being backed up to iCloud. This normally does not apply to Lima VMs, which are stored under `~/.lima` by default.

Is this flag also used by Time Machine? I know that the Time Machine exclusion list is stored in a plist file somewhere and can be edited with the Time Machine preference panel, our via the `tmutil` commandline tool. Would files with the `NSURLIsExcludedFromBackupKey` flag also show up in that exclusion list?

Just trying to understand the motivation for this PR, and the actual scope.

norio-nomura:
You can check the exclusion by `tmutil isexcluded`:
```bash
$ tmutil isexcluded ~/.lima/docker{,/*}
[Included]    /System/Volumes/Data/Users/norio/.lima/docker
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/basedisk
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/cidata.iso
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/default_ep.sock
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/default_fd.sock
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/diffdisk
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/ha.pid
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/ha.sock
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/ha.stderr.log
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/ha.stdout.log
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/launchd.stderr.log
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/launchd.stdout.log
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/lima.yaml
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/serialv.log
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/sock
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/ssh.config
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/ssh.sock
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/vz-efi
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/vz-identifier
[Included]    /System/Volumes/Data/Users/norio/.lima/docker/vz.pid
```

norio-nomura:
With `excludeFromBackup: "all"`
```bash
$ tmutil isexcluded ~/.lima/docker{,/*}
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/basedisk
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/cidata.iso
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/default_ep.sock
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/default_fd.sock
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/diffdisk
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/ha.pid
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/ha.sock
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/ha.stderr.log
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/ha.stdout.log
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/launchd.stderr.log
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/launchd.stdout.log
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/lima.yaml
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/serialv.log
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/sock
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/ssh.config
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/ssh.sock
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/vz-efi
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/vz-identifier
[Excluded]    /System/Volumes/Data/Users/norio/.lima/docker/vz.pid
```

afbjorklund:
> Just trying to understand the motivation for this PR, and the actual scope.

Probably a good idea to open an issue, before adding the PR (guilty of that myself)

For instance, on Linux it might need something like `.deja-dup-ignore` created?

pkg/start/start.go
```diff
@@ -70,6 +70,26 @@ func ensureNerdctlArchiveCache(y *limayaml.LimaYAML, created bool) (string, erro
...
+		// Exclude the instance directory from Time Machine backups
+		targets = map[string]bool{instDir: true, baseDisk: false, diffDisk: false}
+	case limayaml.ExcludeFromBackupDisks:
+		// Exclude disk files from Time Machine backup.
+		targets = map[string]bool{instDir: false, baseDisk: true, diffDisk: true}
+	default: // limayaml.ExcludeFromBackupNone
+		targets = map[string]bool{instDir: false, baseDisk: false, diffDisk: false}
+	}
+	for target, exlucde := range targets {
+		osutil.SetExcludeFromBackup(target, exlucde)
```
jandubois, line 89:
```suggestion
	for target, exclude := range targets {
		osutil.SetExcludeFromBackup(target, exclude)
```

pkg/limayaml/validate.go
```diff
@@ -303,6 +303,16 @@ func Validate(y LimaYAML, warn bool) error {
 	if err := validateNetwork(y, warn); err != nil {
 		return err
 	}
+
+	switch *y.ExcludeFromBackup {
+	case ExcludeFromBackupAll, ExcludeFromBackupDisks, ExcludeFromBackupNone:
+	default:
+		return fmt.Errorf("field `excludeFromBackup` must be %q, %q or %q; got %q", ExcludeFromBackupAll, ExcludeFromBackupDisks, ExcludeFromBackupNone, *y.ExcludeFromBackup)
```
jandubois, line 310: I know it isn't necessary, but personally I prefer to always use the oxford comma:

```suggestion
		return fmt.Errorf("field `excludeFromBackup` must be %q, %q, or %q; got %q", ExcludeFromBackupAll, ExcludeFromBackupDisks, ExcludeFromBackupNone, *y.ExcludeFromBackup)
```

jandubois (commented):
I just did a quick look, but didn't actually test it. Just some spelling nits.

jandubois:
> * "all": the instance directory
> * "disks": `basedisk` and `diffdisk` files
> * "none": none

I'm not sure if we need the `disks` setting; I think it could be just a boolean. The only file besides the disk worth backing up is `lima.yaml`; everything else is generated data.

So I think we still could do a boolean for `excludeFromBackup`:

* `true`: only `lima.yaml` will be backed up
* `false`: everything will be backed up

It feels simpler to me, but curious what others think.

afbjorklund:
Am I right in understanding that this only affects installations where you have changed the location, and only in the case where $LIMA_HOME has not been added to the backup exclude list? (I think it _will_ be backed up on Linux, though.)

norio-nomura:
> `true`: only `lima.yaml` will be backed up

Excluding files other than `lima.yaml` from backup requires setting `NSURLIsExcludedFromBackupKey` for each file. I feel the balance is not good between the amount of code changes required and the backup reduction effect achieved by this. 🤔

norio-nomura:
> I agree because I suspect it would also pollute the exclusion list displayed in the Time Machine preference pane with too many entries.

Files excluded from backup with `NSURLIsExcludedFromBackupKey` do not appear in the Time Machine preference.

jandubois:
I wonder if the backup status of instances should be included in `limactl ls` output. If yes, what would be a compact name for the column?

I have 2 more questions, that are related, but outside the scope of this PR:

1. Should there be a way to exclude all of `$LIMA_HOME` from backups instead of setting it on each instance individually?

   This can be achieved with `override.yaml` and `defaults.yaml`, or better manually via `tmutil` or the preference dialog. So I think this just needs documentation, and no code.

   I suspect excluding all of `$LIMA_HOME` means that the `excludeFromBackups` settings of individual instances will be ignored, but don't actually know. If they are ignore, should we display a warning?

2. Should there be a way to exclude data disks from backups? What should be the mechanism to specify that?

    I think we could add another subcommand `limactl disk exclude-from-backups $INSTANCE` or similar. The status should then be included in `limactl disk ls` output.

jandubois:
> Files excluded from backup with `NSURLIsExcludedFromBackupKey` do not appear in the Time Machine preference.

So what happens if you set `NSURLIsExcludedFromBackupKey` on the directory to `true`, but on `lima.yaml` to `false`. Will `lima.yaml` be excluded because the directory is excluded, or will the setting on the file override the setting on the directory?

norio-nomura:
> So what happens if you set `NSURLIsExcludedFromBackupKey` on the directory to `true`, but on `lima.yaml` to `false`.

As indicated in https://github.com/lima-vm/lima/pull/2159#issuecomment-1905398427 , the `lima.yaml` is excluded because the directory is excluded.

norio-nomura:
The individual `NSURLIsExcludedFromBackupKey` settings can be checked with `xattr`.
With `excludeFromBackup: "disks"`:
```bash
$ xattr -r ~/.lima/docker 2>/dev/null|grep exclude
/Users/norio/.lima/docker/basedisk: com.apple.metadata:com_apple_backup_excludeItem
/Users/norio/.lima/docker/diffdisk: com.apple.metadata:com_apple_backup_excludeItem
```
With `excludeFromBackup: "all"`:
```bash
$ xattr -r ~/.lima/docker 2>/dev/null|grep exclude
/Users/norio/.lima/docker: com.apple.metadata:com_apple_backup_excludeItem
```

norio-nomura:
I've come up with a better idea, so I'm closing this PR. I might reopen this if that doesn't go well.

PR #2163, opened 2024-01-24
add completion for limactl disk

vasileknik76:
Added disk name completion for the following commands:
`limactl disk delete`
`limactl disk unlock`
`limactl disk resize`


AkihiroSuda (approved):
Thanks

PR #2179, opened 2024-02-03
limayaml: allow `{{.Name}}` in provision scripts

AkihiroSuda:
Instance name (`{{.Name}}`) has been allowed in host templates but missing in guest templates


jandubois (approved):
Thanks, LGTM

PR #2181, opened 2024-02-03
Use `(*os.File).Write()` instead of `os.WriteFile()`

norio-nomura:
When `os.WriteFile()` writes to an existing file, it does not apply the permissions passed as parameter. The desired permissions have already been set by `os.CreateTemp()`, so to avoid confusion, it should be changed to a writing method that does not specify permissions.


jandubois (approved):
Thanks, LGTM.

PR #2202, opened 2024-02-14
fix: COLORTERM quoting on Windows

pendo324:
On Windows, when `COLORTERM` is set, you'll get an error like this when trying to run any Lima command. exit status `255` with stderr:

```
command-line line 0: invalid quotes
```

This is an issue because some programs (like VSCode's integrated terminal) set `COLORTERM` by default.

I tracked this down to the quoting of `COLORTERM` in the `SendEnv` SSH option. The input to `SendEnv` doesn't even technically need to be quoted, but following the conventions already in the file, I just put it in single quotes instead, and it still works as expected.

---

Also, out of scope for this PR, but [the way we're adding COLORTERM to the sshd's `AcceptEnv`](https://github.com/lima-vm/lima/blob/88c89165273d87b33753a593af867b20c1d1c67d/pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh#L15) is broken on some modern distro's (like Fedora) because they don't ship an `/etc/ssh/sshd_config` with `AcceptEnv` in it at all.

I can make an issue or PR for this separately.

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -180,7 +180,7 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	if _, present := os.LookupEnv("COLORTERM"); present {
 		// SendEnv config is cumulative, with already existing options in ssh_config
-		sshArgs = append(sshArgs, "-o", "SendEnv=\"COLORTERM\"")
+		sshArgs = append(sshArgs, "-o", "SendEnv=COLORTERM")
 	}
 	sshArgs = append(sshArgs, []string{
 		"-q",


jandubois:
> The input to `SendEnv` doesn't even technically need to be quoted, but following the conventions already in the file, I just put it in single quotes instead, and it still works as expected.

If single quotes work everywhere, then I wonder why we have separate quotes for Windows and Linux in https://github.com/lima-vm/lima/blob/88c89165273d87b33753a593af867b20c1d1c67d/pkg/sshutil/sshutil.go#L139-L144

However, when I tried to test if either form works, I noticed that single, double, or no quotes, none would actually import the `COLORTERM` variable into the instance:

```console
$ lima sudo grep AcceptEnv /etc/ssh/sshd_config
AcceptEnv COLORTERM LANG LC_*
$ COLORTERM=truecolor lima env | grep COLOR
```

Any idea what is going wrong?

> some modern distro's (like Fedora) because they don't ship an `/etc/ssh/sshd_config` with `AcceptEnv` in it at all.

This is also true for Alpine 3.19.

I would suggest to write the `AcceptEnv` to a file in `/etc/ssh/sshd_config.d` when possible, and only patch `/etc/ssh/sshd_config` as a backup when the directory doesn't exist.

afbjorklund (approved):

afbjorklund:
> none would actually import the `COLORTERM` variable into the instance

did you restart the initial ssh session, such as limactl restart or rm ssh.sock?

(I know I know, you are _supposed_ to issue a `ssh -O exit` with the ssh config)

I didn't actually test the proposed change.

pendo324:
> > The input to `SendEnv` doesn't even technically need to be quoted, but following the conventions already in the file, I just put it in single quotes instead, and it still works as expected.
> 
> If single quotes work everywhere, then I wonder why we have separate quotes for Windows and Linux in
> 
> https://github.com/lima-vm/lima/blob/88c89165273d87b33753a593af867b20c1d1c67d/pkg/sshutil/sshutil.go#L139-L144
> 
> However, when I tried to test if either form works, I noticed that single, double, or no quotes, none would actually import the `COLORTERM` variable into the instance:
> 
> ```
> $ lima sudo grep AcceptEnv /etc/ssh/sshd_config
> AcceptEnv COLORTERM LANG LC_*
> $ COLORTERM=truecolor lima env | grep COLOR
> ```
> 
> Any idea what is going wrong?
> 
> > some modern distro's (like Fedora) because they don't ship an `/etc/ssh/sshd_config` with `AcceptEnv` in it at all.
> 
> This is also true for Alpine 3.19.
> 
> I would suggest to write the `AcceptEnv` to a file in `/etc/ssh/sshd_config.d` when possible, and only patch `/etc/ssh/sshd_config` as a backup when the directory doesn't exist.

I was able to verify the COLORTERM change on both Windows and Linux. I haven't tested if it works for the other special-cased options in `sshutil` yet. When I initially special-cased those quotes, I just left them as double quotes since I wasn't sure if there was a weird edge-case I wasn't aware of, but I can go back and test switching those to single quotes, at least on macOS.

As for why you weren't able to get COLORTERM to work, I just tried it on Fedora and followed @afbjorklund's steps:

1. Make sure `AcceptEnv` is configured properly
    ```bash
    limactl shell <instance>
    cat <<EOF >>/etc/ssh/sshd_config.d/50-lima-cidata.conf
    AcceptEnv=COLORTERM LANG LC_*
    
    EOF
    ```
1. Make sure the file has the right permissions
    ```bash
    sudo chown root:root /etc/ssh/sshd_config.d/50-lima-cidata.conf
    sudo chmod og-rwx /etc/ssh/sshd_config.d/50-lima-cidata.conf
    ```
1. Restart sshd (`sudo systemctl restart sshd`)
1. Exit Lima shell and delete `ssh.sock` (`rm $LIMA_HOME/<instance>/ssh.sock`)
1. Make sure COLORTERM is set on your host and then check if its propagated
    ```bash
    echo $COLORTERM
    truecolor
    limactl <instance> shell echo $COLORTERM
    truecolor
    ```

You can see that my command is using singe quotes in the `ps aux | grep COLORTERM output` (truncated for readability):
```
 ... /usr/bin/ssh ... -o SendEnv='COLORTERM' -q -p 50646 127.0.0.1 ...
```

jandubois:
> did you restart the initial ssh session, such as limactl restart or rm ssh.sock?

No, shouldn't this happen automatically via https://github.com/lima-vm/lima/blob/9a964c0fa9c012d9b2ca94495666220f0abdf0e3/pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh#L14-L24

Also `limactl restart` doesn't seem to be implemented 😄

jandubois (dismissed):
The change looks fine; I would prefer to just drop the quotes completely.

jandubois:
> 4\. Exit Lima shell and delete `ssh.sock` (`rm $LIMA_HOME/<instance>/ssh.sock`)

I see that it works with these steps with Fedora, and I think the PR is fine as-is, so I approved it.

But there is something wrong with the implementation when it doesn't work out-of-the-box for the `default` instance. Even restarting the instance doesn't make it work for me. But that is unrelated to your PR, so I'll open a separate issue (plus one for Fedora not having an `AccentEnv` line in `/etc/ssh/sshd_config`).

pendo324:
> The change looks fine; I would prefer to just drop the quotes completely.

Yeah, I think dropping the quotes entirely makes sense too. I'll make a revision.



> > 4. Exit Lima shell and delete `ssh.sock` (`rm $LIMA_HOME/<instance>/ssh.sock`)
> 
> I see that it works with these steps with Fedora, and I think the PR is fine as-is, so I approved it.
> 
> But there is something wrong with the implementation when it doesn't work out-of-the-box for the `default` instance. Even restarting the instance doesn't make it work for me. But that is unrelated to your PR, so I'll open a separate issue (plus one for Fedora not having an `AccentEnv` line in `/etc/ssh/sshd_config`).

Thanks for making these issues for tracking. I'll try to make some followup PRs to fix this for the different distros we test (especially default).

As a followup: Does Lima aim to support any distros that ship different sshd's with different requirements for these variables, or just different variables entirely?

jandubois:
> As a followup: Does Lima aim to support any distros that ship different sshd's with different requirements for these variables, or just different variables entirely?

I don't understand what you are asking. Do you want to be able to add additional variables to `AcceptEnv`, or are you asking about something else?

pendo324:
> > As a followup: Does Lima aim to support any distros that ship different sshd's with different requirements for these variables, or just different variables entirely?
> 
> I don't understand what you are asking. Do you want to be able to add additional variables to `AcceptEnv`, or are you asking about something else?

Just to clarify, this is unrelated to this PR and more for followup fixes to the issues you created.

OpenSSH specifies AcceptEnv variables one way, are there any other distros / sshds (like dropbear or something) that Lima supports or aims to support, and that we know do thing differently? It doesn't matter that much right now if everything is expected to use OpenSSH.

jandubois:
> if everything is expected to use OpenSSH.

I think this is the current assumption. We can think about generalizing things when there is an actual use-case?

jandubois:
> I'll make a revision.

Just to make sure: I haven't merged because you said you would still want to remove the quotes, so I'm waiting for that change. Otherwise let me know if you want to merge as-is.

pendo324:
> > I'll make a revision.
> 
> Just to make sure: I haven't merged because you said you would still want to remove the quotes, so I'm waiting for that change. Otherwise let me know if you want to merge as-is.

Was just triple checking without quoting, should be good now

jandubois (approved):
Thanks, LGTM

PR #2220, opened 2024-02-19
Support /etc/ssh/sshd_config.d for AcceptEnv

### LINKED ISSUE #2207: Make setting AcceptEnv more robust
> [the way we're adding COLORTERM to the sshd's AcceptEnv](https://github.com/lima-vm/lima/blob/88c89165273d87b33753a593af867b20c1d1c67d/pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh#L15) is broken on some modern distro's (like Fedora) because they don't ship an /etc/ssh/sshd_config with AcceptEnv in it at all.

_originally posted by @pendo324 in #2202_

I would suggest to write the AcceptEnv to a file in /etc/ssh/sshd_config.d when possible, and only patch /etc/ssh/sshd_config as a backup when the directory doesn't exist (see https://github.com/lima-vm/lima/pull/2202#issuecomment-1946708624).

afbjorklund:
Closes #2207

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh
@@ -1,18 +1,24 @@
 #!/bin/sh
 set -eux
 
-# openSUSE Tumbleweed uses /etc/ssh/sshd_config.d, not /etc/ssh/sshd_config
-# TODO: support /etc/ssh/sshd_config.d
-if [ ! -e /etc/ssh/sshd_config ]; then
+if [ -d /etc/ssh/sshd_config.d ]; then
+	if [ -e /etc/ssh/sshd_config.d/10-acceptenv-colorterm.conf ]; then
+		exit 0
+	fi
+
+	# accept any incoming COLORTERM environment variable
+	echo "AcceptEnv COLORTERM" >/etc/ssh/sshd_config.d/10-acceptenv-colorterm.conf
+elif [ -e /etc/ssh/sshd_config ]; then
+	if grep -q "COLORTERM" /etc/ssh/sshd_config; then
+		exit 0
+	fi
+
+	# accept any incoming COLORTERM environment variable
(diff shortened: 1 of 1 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot/11-colorterm-environment.sh
```diff
@@ -1,18 +1,24 @@
...
+		exit 0
+	fi
+
+	# accept any incoming COLORTERM environment variable
+	echo "AcceptEnv COLORTERM" >/etc/ssh/sshd_config.d/10-acceptenv-colorterm.conf
 fi
 
-if grep -q "COLORTERM" /etc/ssh/sshd_config; then
-	exit 0
+if [ -e /etc/ssh/sshd_config ]; then
```
jandubois, line 13: Should this be an `elif`? Otherwise you are creating the separate file **and** also patching the config. I guess it doesn't hurt, but wanted to make sure it is intentional.

```suggestion
elif [ -e /etc/ssh/sshd_config ]; then
```

afbjorklund: I was thinking that it might be overridden in the default sshd_config, after including the files...

Currently it seems to be either/or in the distributions, but could be a possibility in future debian?

jandubois: You can have multiple `AcceptEnv` settings, and they are cumulative:

```
AcceptEnv FOO
AcceptEnv BAR
```

is the same as

```
AcceptEnv FOO BAR
```

afbjorklund: Ah, OK. I think I will change the code to use `elif` then - and to exit if it doesn't find anything (shouldn't happen)

jandubois (approved):
Thanks, LGTM

PR #2228, opened 2024-03-04
Update Guest Agent error message

DennisRasey:
Updated Guest Agent error message to include a colon & space before the actual error message is written out to make it more readable.

**Before**
> WARN[0073] [hostagent] connection to the guest agent was closed **unexpectedlydial** unix /home/drasey/.lima/fedora-rawhide/ga.sock: connect: no such file or directory

**After**
> WARN[0073] [hostagent] connection to the guest agent was closed unexpectedly: dial unix /home/drasey/.lima/fedora-rawhide/ga.sock: connect: no such file or directory

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -598,12 +598,12 @@ func (a *HostAgent) watchGuestAgentEvents(ctx context.Context) {
 		if err == nil {
 			if err := a.processGuestAgentEvents(ctx, client); err != nil {
 				if !errors.Is(err, context.Canceled) {
-					logrus.WithError(err).Warn("guest agent events closed unexpectedly", err)
+					logrus.WithError(err).Warn("guest agent events closed unexpectedly: ", err)
 				}
 			}
 		} else {
 			if !errors.Is(err, context.Canceled) {
-				logrus.WithError(err).Warn("connection to the guest agent was closed unexpectedly", err)
+				logrus.WithError(err).Warn("connection to the guest agent was closed unexpectedly: ", err)
 			}
 		}
 		select {


pkg/hostagent/hostagent.go
```diff
@@ -598,12 +598,12 @@ func (a *HostAgent) watchGuestAgentEvents(ctx context.Context) {
 		if err == nil {
 			if err := a.processGuestAgentEvents(ctx, client); err != nil {
 				if !errors.Is(err, context.Canceled) {
-					logrus.WithError(err).Warn("guest agent events closed unexpectedly", err)
+					logrus.WithError(err).Warn("guest agent events closed unexpectedly: ", err)
```
jandubois, line 601: Isn't the additional `err` argument redundant with `WithError(err)`?

DennisRasey: I don't think so (though my Golang skills are practically non-existent)!

Removing `err` results in the actual being suppressed:

`logrus.WithError(err).Warn("guest agent events closed unexpectedly: ")`

Results in this:
`WARN[0044] [hostagent] guest agent events closed unexpectedly:  `

Keeping it in place prints the actual "error":
`WARN[0044] [hostagent] guest agent events closed unexpectedly: rpc error: code = Unavailable desc = error reading from server: EOF`

jandubois: I see. The problem is that the errors are send via `logrus.JSONFormatter`, but when they are unmarshalled in `logrusutil.PropagateJSON` it ignores any fields except `level`, `msg`, and `time`.

@AkihiroSuda How is this supposed to work? Should additional fields, or at least `logrus.ErrorKey` be imported by `PropagateJSON`?

It feels silly to include the error message twice in `ha.stderr.log` with

```golang
logrus.WithError(err).Warnf("Oops: %w", err)
```

If we don't want to unmarshal the error field, then we shouldn't use `WithError`. But I would prefer to keep the error data structured instead of inlining it all inside the message.

jandubois: Looking at the hostagent code, the only other field I see being emitted (and then ignored) is the `event` field:

```
pkg/hostagent/events/watcher.go
57:			logrus.WithField("event", ev).Debugf("received an event")

pkg/hostagent/hostagent.go
288:		logrus.WithField("event", ev).WithError(err).Error("failed to emit an event")
```

So if `error` should be unmarshalled, should we unmarshal `event` too?

jandubois: Here are some other log messages that include the error twice:

```console
$ ag WithError.*\\\).*,\ *err
pkg/hostagent/hostagent.go
604:				logrus.WithError(err).Warn("failed to start inotify", err)
619:					logrus.WithError(err).Warn("guest agent events closed unexpectedly", err)
624:				logrus.WithError(err).Warn("connection to the guest agent was closed unexpectedly", err)

pkg/hostagent/inotify.go
52:				logrus.WithError(err).Warn("failed to send inotify", err)
```

AkihiroSuda: > How is this supposed to work? Should additional fields, or at least logrus.ErrorKey be imported by PropagateJSON?

Yes, all the fields should be propagated

jandubois: > Yes, all the fields should be propagated

I've implemented that in #2233. If we merge that PR, then this PR can be closed.

jandubois:
@DennisRasey Thank your for your PR! As I mentioned above, it showed that we have a bug in how `WithError()` and `WithField()` data in the hostagent are reported back by the `limactl start` command. I've fixed this in #2233, and therefore this PR is no longer needed.

So thanks again for your contribution; without it we wouldn't have noticed that there was a bug in the error propagation logic!

PR #2233, opened 2024-03-06
Propagate all logging fields to the tty logger

jandubois:
The fields are serialized in JSON by the hostagent and then printed by the `limactl start` command to the console, so need to be explicitly added back to the logger entry.

This PR is in response to https://github.com/lima-vm/lima/pull/2228#discussion_r1513910877.

Note that nested structures (like `events.Event`) will be flattened by the standard logger, e.g.

```golang
       err := errors.New("oops")
       ev := events.Event{
               Status: events.Status{
                       Running: true,
                       Errors:  []string{"Error one", "Error two"},
               },
       }
       logrus.WithError(err).WithField("events", ev).Warn("Log message with error")
```

will result in

```console
WARN[0020] [hostagent] Log message with error            error=oops events="map[status:map[errors:[Error one Error two] running:true] time:0001-01-01T00:00:00Z]"
```
CleanShot 2024-03-06 at 00 26 57@2x

The change to `qemu_driver.go` is needed to avoid the ugly `<nil>` error in

```
INFO[0004] [hostagent] QEMU has exited                   error="<nil>"
```

@balajiv113 Your inotify PR has additional logging entries with duplicate `err` which should be changed if this PR is accepted/merged:

```
pkg/hostagent/hostagent.go
604:				logrus.WithError(err).Warn("failed to start inotify", err)

pkg/hostagent/inotify.go
52:				logrus.WithError(err).Warn("failed to send inotify", err)
```

### DIFF
--- a/pkg/logrusutil/logrusutil.go
+++ b/pkg/logrusutil/logrusutil.go
@@ -19,10 +19,14 @@ func PropagateJSON(logger *logrus.Logger, jsonLine []byte, header string, begin 
 	}
 
 	var (
-		lv  logrus.Level
-		j   JSON
-		err error
+		entry  *logrus.Entry
+		fields logrus.Fields
+		lv     logrus.Level
+		j      JSON
+		err    error
 	)
+	entry = logrus.NewEntry(logger)
+
 	if err := json.Unmarshal(jsonLine, &j); err != nil {
 		goto fallback
 	}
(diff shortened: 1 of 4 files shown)


pkg/logrusutil/logrusutil.go
```diff
@@ -33,24 +37,30 @@ func PropagateJSON(logger *logrus.Logger, jsonLine []byte, header string, begin
 	if err != nil {
 		goto fallback
 	}
+	if err := json.Unmarshal(jsonLine, &fields); err == nil {
+		delete(fields, "level")
+		delete(fields, "msg")
+		delete(fields, "time")
```
AkihiroSuda, line 43: Could you add comments to explain these deletes

jandubois: I added a comment, and also added a call to set the log timestamp to the original time:

```golang
	entry = entry.WithTime(j.Time)
```

jandubois: I've now also renamed the `logger` variable in the `qemu_driver.go` change to `entry` because its type is `logrus.Entry` and not `logrus.Logger`.

jandubois:
I made one more change to replace the `switch lv` block with this logic, which feels clearer to me:

```golang
if !logger.IsLevelEnabled(lv) {
	return
}
// …
if lv <= logrus.FatalLevel {
	entry = entry.WithField("level", lv)
	lv = logrus.ErrorLevel
}
entry.Log(lv, header+j.Msg)
```

jandubois:
@AkihiroSuda Speaking of comments, I think this line could use some explanation:

```
if !j.Time.IsZero() && !begin.IsZero() && begin.After(j.Time.Add(epsilon)) {
	return
}
```

It seems to be added by 21b7cb0e to control logging for `limactl stop` iiuc, and seems to ignore log messages that are more than one second before the `begin` value.

It is not clear to me why this is needed; I assume if there is still a backlog of unprocessed log messages from before the `stop` command, that you don't want to dump them out to the terminal?

jandubois:
I just realized that there are tests in `logrusutil_test.go` that should be updated too. I won't have time for this for the next few days though, so will turn this into draft for now.

I don't plan on changing anything else but the tests though, so feel free to review the draft code in case you have any feedback!

jandubois:
> I won't have time for this for the next few days though, so will turn this into draft for now.

I added the tests and took it out of draft; I think it is good now.

AkihiroSuda (approved):
Thanks

PR #2234, opened 2024-03-08
alpine: avoid wiping out writable host mounts under /home, etc.

### LINKED ISSUE #2221: Lost files on host when canceling limactl start
### Description

Canceling `limactl start debug` deleted files that were mounted from the host.

I don't have the exact config anymore but it was similar to this. The mounted ~/Files directory was deleted.
```yaml
images:
  - location: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.35/alpine-lima-std-3.19.0-aarch64.iso"
    arch: "aarch64"
    digest: "sha512:13e50601ee65af5d7a6dfd30bb41fd89f8bf806ecdb516c61fe238c3cf3b57cf67469418a99f329bb4c343e3387e6e0fd4fe20501cfd501f031f7244adc67215"

containerd:
  system: false
  user: false
  archives:
    - location: "https://github.com/containerd/nerdctl/releases/download/v2.0.0-beta.0/nerdctl-full-2.0.0-beta.0-linux-arm64.tar.gz"
      arch: "aarch64"
      digest: "sha256:29c58dc39ebca1642f14827915f67d909ec49b1d7458938401cf835edbc5b8d0"

os: "Linux"
arch: "aarch64"
vmType: "vz"
cpus: 2
memory: "2GiB"
disk: "2GiB"

mountType: "virtiofs"
mounts:    
  - location: "~/Files"                                                                                                                      
    mountPoint: "/Files"    
    writable: true  

audio:
  device: "vz"
video:
  display: "none"
networks:
  - vzNAT: true
ssh:
  localPort: 54201
  loadDotSSHPubKeys: false
  forwardAgent: false
  forwardX11: false
  forwardX11Trusted: false
```

AkihiroSuda:
A host directory could be wiped out when all the following conditions are met:
- The directory is mounted to Lima via virtiofs or 9p (reverse-sshfs is not affected)
- The mount is writable
- The mount point in the guest is under one of: /etc /home /root /usr/local /var/lib
- The guest OS is Alpine Linux

Fix #2221
Fix rancher-sandbox/rancher-desktop#6582

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
@@ -9,6 +9,36 @@ test -f /etc/alpine-release || exit 0
 # Data directories that should be persisted across reboots
 DATADIRS="/etc /home /root /tmp /usr/local /var/lib"
 
+# Prepare mnt.sh (used for restoring mounts later)
+echo "#!/bin/sh" >/mnt.sh
+echo "set -eux" >>/mnt.sh
+for DIR in ${DATADIRS}; do
+	while IFS= read -r LINE; do
+		[ -z "$LINE" ] && continue
+		MNTDEV="$(echo "${LINE}" | awk '{print $1}')"
+		# unmangle " \t\n\\#"
+		# https://github.com/torvalds/linux/blob/v6.6/fs/proc_namespace.c#L89
+		MNTPNT="$(echo "${LINE}" | awk '{print $2}' | sed -e 's/\\040/ /g; s/\\011/\t/g; s/\\012/\n/g; s/\\134/\\/g; s/\\043/#/g')"
+		# Ignore if MNTPNT is neither DIR nor a parent directory of DIR.
+		# It is not a parent if MNTPNT doesn't start with DIR, or the first
+		# character after DIR isn't a slash.
+		WITHOUT_DIR="${MNTPNT#"$DIR"}"
+		# shellcheck disable=SC2166
+		[ "$MNTPNT" != "$DIR" ] && [ "$MNTPNT" == "$WITHOUT_DIR" -o "${WITHOUT_DIR::1}" != "/" ] && continue
+		MNTTYPE="$(echo "${LINE}" | awk '{print $3}')"
+		[ "${MNTTYPE}" = "ext4" ] && continue
+		[ "${MNTTYPE}" = "tmpfs" ] && continue
+		MNTOPTS="$(echo "${LINE}" | awk '{print $4}')"
+		MNTPNT=${MNTPNT//\\/\\\\}
+		MNTPNT=${MNTPNT//\"/\\\"}
+		echo "mount -t \"${MNTTYPE}\" -o \"${MNTOPTS}\" \"${MNTDEV}\" \"${MNTPNT}\"" >>/mnt.sh
+		# Before mv, unmount filesystems (virtiofs, 9p, etc.) below "${DIR}", otherwise host mounts will be wiped out
+		# https://github.com/rancher-sandbox/rancher-desktop/issues/6582
+		umount "${MNTPNT}" || exit 1
+	done </proc/mounts
+done
+chmod +x /mnt.sh
+
 # When running from RAM try to move persistent data to data-volume
 # FIXME: the test for tmpfs mounts is probably Alpine-specific
 if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
@@ -61,11 +91,6 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
 				PART=$(lsblk --list /dev/"${DISK}" --noheadings --output name,type | awk '$2 == "part" {print $1}')
 				mkfs.ext4 -L data-volume /dev/"${PART}"
 				mount -t ext4 /dev/disk/by-label/data-volume /mnt/data
-				# Unmount all mount points under /tmp so we can move it to the data volume:
-				# "mount1 on /tmp/lima type 9p (rw,dirsync,relatime,mmap,access=client,trans=virtio)"
-				for MP in $(mount | awk '$3 ~ /^\/tmp\// {print $3}'); do
-					umount "${MP}"
-				done
 				# setup apk package cache
 				mkdir -p /mnt/data/apk/cache
 				mkdir -p /etc/apk
@@ -88,8 +113,8 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
 			mount --bind /mnt/data"${DIR}" "${DIR}"
 		fi
 	done
-	# Make sure to re-mount any mount points under /tmp
-	mount -a
+	# Remount submounts on top of the new ${DIR}
+	/mnt.sh
 	# Reinstall packages from /mnt/data/apk/cache into the RAM disk
 	apk fix --no-network
 fi
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -225,6 +225,7 @@ jobs:
         - opensuse.yaml
         - experimental/net-user-v2.yaml
         - docker.yaml
+        - ../hack/test-templates/alpine-9p-writable.yaml
     steps:
     - uses: actions/checkout@v4
       with:
--- a/hack/test-port-forwarding.pl
+++ b/hack/test-port-forwarding.pl
@@ -126,7 +126,7 @@ my $ha_log_size = -s $ha_log or die;
 foreach my $id (0..@test-1) {
     my $test = $test[$id];
     my $nc = "nc -l $test->{guest_ip} $test->{guest_port}";
-    if ($instance eq "alpine") {
+    if ($instance =~ /^alpine/) {
         $nc = "nc -l -s $test->{guest_ip} -p $test->{guest_port}";
     }
 
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -34,13 +34,15 @@ declare -A CHECKS=(
 	["vmnet"]=""
 	["disk"]=""
 	["user-v2"]=""
+	["mount-path-with-spaces"]=""
 )
 
 case "$NAME" in
-"alpine")
+"alpine"*)
 	WARNING "Alpine does not support systemd"
 	CHECKS["systemd"]=
 	CHECKS["container-engine"]=
+	[ "$NAME" = "alpine-9p-writable" ] && CHECKS["mount-path-with-spaces"]="1"
 	;;
 "k3s")
 	ERROR "File \"$FILE\" is not testable with this script"
@@ -59,6 +61,7 @@ case "$NAME" in
 	CHECKS["disk"]=1
 	CHECKS["snapshot-online"]="1"
 	CHECKS["snapshot-offline"]="1"
+	CHECKS["mount-path-with-spaces"]="1"
 	;;
 "net-user-v2")
 	CHECKS["port-forwards"]=""
@@ -113,6 +116,11 @@ set -x
 "${LIMACTL_CREATE[@]}" ${LIMACTL_CREATE_ARGS} "$FILE"
 set +x
 
+if [[ -n ${CHECKS["mount-path-with-spaces"]} ]]; then
+	mkdir -p "/tmp/lima test dir with spaces"
+	echo "test file content" >"/tmp/lima test dir with spaces/test file"
+fi
+
 INFO "Starting \"$NAME\""
 set -x
 if ! limactl start "$NAME"; then
@@ -125,6 +133,15 @@ limactl shell "$NAME" uname -a
 
 limactl shell "$NAME" cat /etc/os-release
 set +x
+
+INFO "Testing that host home is not wiped out"
+[ -e "$HOME/.lima" ]
+
+if [[ -n ${CHECKS["mount-path-with-spaces"]} ]]; then
+	INFO 'Testing that "/tmp/lima test dir with spaces" is not wiped out'
+	[ "$(cat "/tmp/lima test dir with spaces/test file")" = "test file content" ]
+	[ "$(limactl shell "$NAME" cat "/tmp/lima test dir with spaces/test file")" = "test file content" ]
+fi
 
 INFO "Testing proxy settings are imported"
 got=$(limactl shell "$NAME" env | grep FTP_PROXY)
@@ -409,3 +426,7 @@ sleep 3
 
 INFO "Deleting \"$NAME\""
 limactl delete "$NAME"
+
+if [[ -n ${CHECKS["mount-path-with-spaces"]} ]]; then
+	rm -rf "/tmp/lima test dir with spaces"
+fi
--- a/hack/test-templates/alpine-9p-writable.yaml
+++ b/hack/test-templates/alpine-9p-writable.yaml
@@ -1,0 +1,22 @@
+# Background: https://github.com/lima-vm/lima/pull/2234
+# Should be tested on a Linux host
+images:
+- location: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.37/alpine-lima-std-3.19.0-x86_64.iso"
+  arch: "x86_64"
+  digest: "sha512:568852df405e6b9858e678171a9894c058f483df0b0570c22cf33fc75f349ba6cc5bb3d50188180d8c31faaf53400fe884ca3e5f949961b03b2bf53e65de88d7"
+- location: "https://github.com/lima-vm/alpine-lima/releases/download/v0.2.37/alpine-lima-std-3.19.0-aarch64.iso"
+  arch: "aarch64"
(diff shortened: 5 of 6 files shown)


jandubois:
> * The directory is mounted to Lima via virtiofs

I have not tested it, but I suspect the same will be true for `9p`. The important factor is that the directory is mounted before `cloud-init` is running, which is true for `9p` and `virtiofs`, but not for `reverse-sshfs`.

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -6,6 +6,9 @@ set -eux -o pipefail
 # Restrict the rest of this script to Alpine until it has been tested with other distros
 test -f /etc/alpine-release || exit 0
 
+# Workaround https://github.com/lima-vm/lima/issues/2235
+sed -i 's/^timezone:.*//g' /etc/fstab
```
jandubois, line 10: This would be better fixed by adding a blank line in the template before the `timezone:` line:

```diff
--- pkg/cidata/cidata.TEMPLATE.d/user-data
+++ pkg/cidata/cidata.TEMPLATE.d/user-data
@@ -21,6 +21,7 @@ mounts:
 {{- end }}

 {{- if .TimeZone }}
+
 timezone: {{.TimeZone}}
 {{- end }}
```

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -74,6 +77,19 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
 				for DIR in ${DATADIRS}; do
 					DEST="/mnt/data$(dirname "${DIR}")"
 					mkdir -p "${DIR}" "${DEST}"
+					# Unmount filesystems (virtiofs, 9p, etc.) below "${DIR}", otherwise host mounts will be wiped out
```
jandubois, line 80: With this change the 5 lines above that start with `# Unmount all mount points under /tmp so we can move it to the data volume:` are redundant and can be removed.

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -74,6 +77,19 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
...
+					# https://github.com/rancher-sandbox/rancher-desktop/issues/6582
+					#
+					# set is for avoiding a failure on empty grep
+					set +e +o pipefail
+					awk '$3 != "ext4" && $3 != "tmpfs" {print $2}' /proc/mounts | grep "^${DIR}" | while IFS= read -r line; do
+						[ -z "$line" ] && continue
+						# unmangle " \t\n\\#"
+						# https://github.com/torvalds/linux/blob/v6.6/fs/proc_namespace.c#L89
+						unmangled="$(echo "$line" | sed -e 's/\\040/ /g; s/\\011/\t/g; s/\\012/\n/g; s/\\134/\\/g; s/\\043/#/g')"
+						umount "$unmangled" || exit 1
```
jandubois, line 90: Should this be a `break` instead of `exit 1`?

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -86,6 +102,14 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
 		if [ -d /mnt/data"${DIR}" ]; then
 			mkdir -p "${DIR}"
 			mount --bind /mnt/data"${DIR}" "${DIR}"
+			# Remount submounts on top of the new ${DIR}
```
jandubois, line 105: I was wondering why this is needed. Is it just because `mount -a` does not deal with mountpoints with spaces, or is there another reason too?

Either way, the `mount -a` should no longer be needed now.

jandubois (changes requested):
There is some code that has become redundant with this change and should be removed.

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
jandubois: The previous lines will also break if the mount point includes spaces (which currently doesn't work anyways, but will be fixed by https://github.com/lima-vm/alpine-lima/pull/124).

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
AkihiroSuda: exit 1 might be safer?

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
AkihiroSuda: The script bind-mounts `/home` after mounting `/home/USER` and makes `/home/USER` invisible,  but `mount -a` still thinks `/home/USER` is properly mounted, so `mount -a` just skips `/home/USER`.

jandubois: Where is `/home/$USER` being remounted. You unmount it above, and then you bind-mount `/home` here. Why does `mount -a` think that `/home/$USER` is still mounted when it has been explicitly unmounted?

Note that I'm just asking to understand. This is obviously the right way to do it because `mount -a` does not handle the mount points with space.

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
jandubois: Maybe. I was thinking if one directory cannot be unmounted because the device is busy, then we could just ignore moving this one directory, but still move the others.

But either way, it is surprising behaviour unless you actually check the logs.

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
AkihiroSuda: > You unmount it above,

Only for the first boot.

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
AkihiroSuda: It doesn't seem correct that we have to care about white spaces 😅

AkihiroSuda: Anyway, this is being fixed in:
- https://github.com/lima-vm/alpine-lima/pull/124
:thanks:

jandubois: > It doesn't seem correct that we have to care about white spaces 😅

It's because we are parsing YAML with `awk` and not an actual parser...

> Anyway, this is being fixed in:
>
> https://github.com/lima-vm/alpine-lima/pull/124

I don't think it is. It is just not being triggered in Rancher Desktop because it is still using Lima 0.19, which doesn't include the `timezone` code.

We obviously can make the `awk` code better, but adding the blank line will fix things for the existing Alpine image, and looks better in the generated `user-data` anyways.

I noticed it too late that the empty line in the template was being removed by the `{{- … }}` expression and didn't want to create a PR for what I thought was just aesthetics.

jandubois: > We obviously can make the awk code better, 

I will update the script in https://github.com/lima-vm/alpine-lima/pull/124 to handle this better

jandubois: Updated in https://github.com/lima-vm/alpine-lima/pull/126

AkihiroSuda: Removed the commit

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
AkihiroSuda: Removed the redundant unmount for /tmp

AkihiroSuda:
DNM, since fstab is gone, the PR has to be worked out again from scratch

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -61,19 +61,34 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
...
-				# "mount1 on /tmp/lima type 9p (rw,dirsync,relatime,mmap,access=client,trans=virtio)"
-				for MP in $(mount | awk '$3 ~ /^\/tmp\// {print $3}'); do
-					umount "${MP}"
-				done
 				# setup apk package cache
 				mkdir -p /mnt/data/apk/cache
 				mkdir -p /etc/apk
 				ln -s /mnt/data/apk/cache /etc/apk/cache
+				# Prepare /mnt.tmp.sh (used for restoring mounts later)
+				echo "#!/bin/sh" >/mnt.tmp.sh
```
jandubois, line 69: Don't you have to put the mount script on `/mnt/data` so it is still there after a reboot?

On the other hand, I don't remember why the remounting has to be done after a reboot; it should only be needed on first boot when mounts have been unmounted.

jandubois: > I don't remember why the remounting has to be done after a reboot;

Oh, it has to be done after the persistent directories have been bind-mounted, so the host directories are not hidden behind.

AkihiroSuda: This is a tmp script and removed at the end

jandubois: It is currently outside the first-boot branch:

```shell
if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
  if [ -e /dev/disk/by-label/data-volume ]; then
    …
  else
    # create /mnt.tmp.sh
  fi
  # bind-mount all persistent directories
  /mnt.tmp.sh
fi
```

The whole outer conditional is somewhat pointless, as it will always be true as long as we boot from an ISO. But the code after the inner conditional is executed on every boot.

AkihiroSuda: fixed to retain it under /mnt/data

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -61,21 +61,37 @@ if [ "$(awk '$2 == "/" {print $3}' /proc/mounts)" == "tmpfs" ]; then
...
 					DEST="/mnt/data$(dirname "${DIR}")"
 					mkdir -p "${DIR}" "${DEST}"
+					while IFS= read -r LINE; do
+						[ -z "$LINE" ] && continue
+						MNTDEV="$(echo "${LINE}" | awk '{print $1}')"
+						# unmangle " \t\n\\#"
+						# https://github.com/torvalds/linux/blob/v6.6/fs/proc_namespace.c#L89
+						MNTPNT="$(echo "${LINE}" | awk '{print $2}' | sed -e 's/\\040/ /g; s/\\011/\t/g; s/\\012/\n/g; s/\\134/\\/g; s/\\043/#/g')"
+						# This comparison needs [[, not [
+						[[ ${MNTPNT} != "${DIR}*" ]] && continue
```
jandubois, line 82: I think the `*` must be outside the quotes to be treated as a wildcard:

```suggestion
						[[ ${MNTPNT} != "${DIR}"* ]] && continue
```

But I don't think this works in `ash` (our `bash` is really just a wrapper for busybox `sh`, which is `ash` with some `bash` compatibility compiled in, but not everything works).

I would try something like this:

```shell
# Ignore if MNTPNT is neither DIR nor a parent directory of DIR.
# It is not a parent if MNTPNT doesn't start with DIR, or the first
# character after DIR isn't a slash.
WITHOUT_DIR=${MNTPNT#$DIR}
[ "$MNTPNT" != "$DIR" ] && [ "$MNTPNT" == "$WITHOUT_DIR" -o "${WITHOUT_DIR::1}" != "/" ] && continue
```

This also handles the case correctly when MNTPNT is `/tmpdir`, which would match `/tmp*`, but is not a subdirectory of it. Well, this is untested, so please double-check the condition!

AkihiroSuda: Thanks, tested and it seems to work.

AkihiroSuda:
Confirmed that the issue affects 9p as well

AkihiroSuda:
Added a test

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -9,6 +9,34 @@ test -f /etc/alpine-release || exit 0
...
+		# It is not a parent if MNTPNT doesn't start with DIR, or the first
+		# character after DIR isn't a slash.
+		WITHOUT_DIR="${MNTPNT#"$DIR"}"
+		# shellcheck disable=SC2166
+		[ "$MNTPNT" != "$DIR" ] && [ "$MNTPNT" == "$WITHOUT_DIR" -o "${WITHOUT_DIR::1}" != "/" ] && continue
+		MNTTYPE="$(echo "${LINE}" | awk '{print $3}')"
+		[ "${MNTTYPE}" = "ext4" ] && continue
+		[ "${MNTTYPE}" = "tmpfs" ] && continue
+		MNTOPTS="$(echo "${LINE}" | awk '{print $4}')"
+		echo "mount -t ${MNTTYPE} -o ${MNTOPTS} ${MNTDEV} ${MNTPNT}" >>/mnt.sh
```
jandubois, line 32: `$MNTPNT` must be quoted because it may contain spaces:

```suggestion
		echo "mount -t ${MNTTYPE} -o ${MNTOPTS} ${MNTDEV} \"${MNTPNT}\"" >>/mnt.sh
```

Ideally we would be using `printf "%q"` for quoting, but the busybox version doesn't support it. 😞

Maybe add a mountpoint with spaces to the tests as well?

AkihiroSuda: fixed, and added a test

jandubois (dismissed):
Thanks, LGTM and worked in my local tests.

I don't think the mounting code will work with mount points that include backslashes, due to the way we have to do the quoting, but I think that is acceptable; who is using backslashes in directory names?

pkg/cidata/cidata.TEMPLATE.d/boot/04-persistent-data-volume.sh
```diff
@@ -9,6 +9,34 @@ test -f /etc/alpine-release || exit 0
...
+		# It is not a parent if MNTPNT doesn't start with DIR, or the first
+		# character after DIR isn't a slash.
+		WITHOUT_DIR="${MNTPNT#"$DIR"}"
+		# shellcheck disable=SC2166
+		[ "$MNTPNT" != "$DIR" ] && [ "$MNTPNT" == "$WITHOUT_DIR" -o "${WITHOUT_DIR::1}" != "/" ] && continue
+		MNTTYPE="$(echo "${LINE}" | awk '{print $3}')"
+		[ "${MNTTYPE}" = "ext4" ] && continue
+		[ "${MNTTYPE}" = "tmpfs" ] && continue
+		MNTOPTS="$(echo "${LINE}" | awk '{print $4}')"
+		echo "mount -t \"${MNTTYPE}\" -o \"${MNTOPTS}\" \"${MNTDEV}\" \"${MNTPNT}\"" >>/mnt.sh
```
jandubois, line 32: I think this is fine to merge as-is. But if you want to support double-quotes and backslashes in mountpoints, then you could to

```suggestion
		MNTPNT=${MNTPNT//\\/\\\\}
		MNTPNT=${MNTPNT//\"/\\\"}
		echo "mount -t \"${MNTTYPE}\" -o \"${MNTOPTS}\" \"${MNTDEV}\" \"${MNTPNT}\"" >>/mnt.sh
```

I've tested it with

```console
$ MNTPNT=$'foo\\bar"baz\nfoo\\bar"baz'
$ echo "$MNTPNT"
foo\bar"baz
foo\bar"baz
$ MNTPNT=${MNTPNT//\\/\\\\}
$ MNTPNT=${MNTPNT//\"/\\\"}
$ echo "$MNTPNT"
foo\\bar\"baz
foo\\bar\"baz
```

AkihiroSuda: Thanks, applied

AkihiroSuda:
Cherry-picked to v0.20.2
- #2241

PR #2236, opened 2024-03-08
alpine: demote from Tier 1 to Tier 2

AkihiroSuda:
The alpine image is quite different from other ones. We have to admit that it is not as stable as other ones.


jandubois (approved):
LGTM

PR #2238, opened 2024-03-09
Bump alpine-lima image from 0.2.35 → 0.2.37

jandubois:
* Fix parsing ssh keys as block string
* Create a mount script instead of editing /etc/fstab
* Make lima-init.sh yaml parsing more robust


AkihiroSuda (approved):
Thanks

PR #2241, opened 2024-03-11
[release/0.20] alpine: avoid wiping out writable host mounts under /home, etc

AkihiroSuda:
Cherry-pick (clean):
- #2238

> * Fix parsing ssh keys as block string
> * Create a mount script instead of editing /etc/fstab
> * Make lima-init.sh yaml parsing more robust
> 

- #2234

> A host directory could be wiped out when all the following conditions are met:
> - The directory is mounted to Lima via virtiofs or 9p (reverse-sshfs is not affected)
> - The mount is writable
> - The mount point in the guest is under one of: /etc /home /root /usr/local /var/lib
> - The guest OS is Alpine Linux
> 
> Fix #2221
> Fix rancher-sandbox/rancher-desktop#6582

- #2242
> The escaping is needed to print the string with quotes, but would break the umount command.

PR #2242, opened 2024-03-11
Unmount MNTPNT before escaping quotes/backslashes

jandubois:
The escaping is needed to print the string with quotes, but would break the umount command.


AkihiroSuda (approved):
Thanks

AkihiroSuda:
Cherry-picked to v0.20.2
- https://github.com/lima-vm/lima/pull/2241

PR #2247, opened 2024-03-13
Remove redundant err from logrus that already uses WithError(err)

jandubois:
Sorry for the trivial PR; just something I noticed while cleaning up.


AkihiroSuda (approved):
Thanks

Issue #2045, opened 2023-12-01
connection to the guest agent was closed unexpectedlyunexpected EOF


terev:
### Description

```
colima version && limactl --version && sw_vers
colima version 0.6.6
git commit: 9ed7f4337861931b4d0192ca5409683a4b7d1cdc

runtime: docker
arch: aarch64

limactl version HEAD-cac97ed

ProductName:		macOS
ProductVersion:		14.1.1
```

I've recently been getting an issue where docker commands start failing because the connection through the forwarded socket is getting closed. It appears that the ssh process dies for some reason. While the socket is working lsof shows the ssh process has it open. But when the socket becomes unresponsive, lsof no longer shows this and the process id no longer exists.

from ~/.colima/_lima/colima/ha.stderr.log :
```
{"error":"unexpected EOF","level":"warning","msg":"connection to the guest agent was closed unexpectedlyunexpected EOF","time":"2023-11-30T23:21:26-05:00"}
```

AkihiroSuda:
With qemu or vz ?

balajiv113:
@terev Will you be able to attach the log for this ?

We recently updated guest agent communication. Technically the connection should be re-established and this should not have any issues with ssh / port forwarding

terev:
@balajiv113 sure. just the host agent log?

terev:
Just reproduced again. But this time the guest agent isn't responding at all:
```
{"error":"unable to connect to guest agent via vsock port 2222","level":"warning","msg":"connection to the guest agent was closed unexpectedlyunable to connect to guest agent via vsock port 2222","time":"2023-12-01T01:51:03-05:00"}
```

balajiv113:
@terev Are you able to ssh into guest ?? If so, try to collect logs from
`sudo journalctl -u lima-guestagent.service -f`

terev:
@balajiv113  Oh thanks! Yeah I was just able to get a log by tailing:
```
sudo journalctl -u lima-guestagent.service -f -n 100
Dec 01 07:22:57 colima systemd[1]: Started lima-guestagent.service - lima-guestagent.
Dec 01 07:22:57 colima lima-guestagent[1126]: time="2023-12-01T07:22:57Z" level=info msg="event tick: 3s"
Dec 01 07:22:58 colima lima-guestagent[1126]: time="2023-12-01T07:22:58Z" level=info msg="serving the guest agent on vsock port: 2222"
[...]

PR #2249, opened 2024-03-15
Adjust Guestagent OOM Parameters

terev:
I've been having issues with the guest agent being OOM killed in my vm. Changing these OOM parameters has helped with stability and I haven't seen this issue so far. Though, as a disclaimer I don't have experience with systemd so I don't know if this is the right thing to do. I've basically copied this pattern from what the docker daemon does.

Related to #2045


AkihiroSuda (approved):
Thanks

PR #2258, opened 2024-03-25
CI: drop Go 1.20

AkihiroSuda:
1.20 has already reached EOL

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -232,7 +232,7 @@ jobs:
         fetch-depth: 1
     - uses: actions/setup-go@v5
       with:
-        go-version: 1.20.x
+        go-version: 1.22.x
     - uses: actions/cache@v4
       with:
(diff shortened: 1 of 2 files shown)


vasileknik76:
Should we also update `1.20` to `1.21` in [website/go.mod](https://github.com/lima-vm/lima/blob/master/website/go.mod#L3)?

afbjorklund:
The new go toolchain has some crazy auto-updating feature, if you bump go.mod

AkihiroSuda:
> Should we also update `1.20` to `1.21` in [website/go.mod](https://github.com/lima-vm/lima/blob/master/website/go.mod#L3)?

I don't know how to update that file.
`go mod tidy` erases all the modules in the file.

vasileknik76:
> I don't know how to update that file. `go mod tidy` erases all the modules in the file.

It seems `go.mod` is not used in `website` directory, so it can be removed.

PR #2260, opened 2024-03-29
Revert "templates: update Arch Linux to 20240315.221711" to avoid xz 5.6.1-1

AkihiroSuda:
This reverts commit 5416270f8b175734cbc69d5ded09700f7c12b676.

The source code archive (not the git repo) of xz v5.6.0 and v5.6.1 turned out to be compromised:
- https://www.openwall.com/lists/oss-security/2024/03/29/4

While ArchLinux's xz package is very *unlikely* to be affected (see the `== Affected Systems ==` section in the email), it is better to rollback the ArchLinux image to use an older version of xz.

- ArchLinux v20240315.221711: xz v5.6.1-1
- ArchLinux v20240115.207158: xz v5.4.5-1

After running `pacman -Syu` on ArchLinux v20240115.207158, a user will get xz v5.6.1-2. This package revision does not use the compromised v5.6.1 source archive:
- https://gitlab.archlinux.org/archlinux/packaging/packages/xz/-/issues/2
- https://gitlab.archlinux.org/archlinux/packaging/packages/xz/-/commit/881385757abdc39d3cfea1c3e34ec09f637424ad

### DIFF
--- a/examples/archlinux.yaml
+++ b/examples/archlinux.yaml
@@ -1,9 +1,9 @@
 # This template requires Lima v0.7.0 or later
 images:
 # Try to use yyyyMMdd.REV image if available. Note that yyyyMMdd.REV will be removed after several months.
-- location: "https://geo.mirror.pkgbuild.com/images/v20240315.221711/Arch-Linux-x86_64-cloudimg-20240315.221711.qcow2"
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Looks like they retracted v20240315.221711 and released [v20240329.225538/](https://geo.mirror.pkgbuild.com/images/v20240329.225538/)

AkihiroSuda:
Probably we should just update the image to the latest revision

Issue #2265, opened 2024-04-04
Generated `user-data` is failing cloud-config schema validation


afbjorklund:
### Description

There are some reports in the log, about the user-data failing validation.

`[WARNING]: Invalid cloud-config provided: Please run 'sudo cloud-init schema --system' to see the schema errors.`

We should probably run yaml and jsonschema validation, on the `user-data`?

But at least fix the syntax errors, and try to conform to the cloud-init schema...

afbjorklund:
Note that changing to "ssh_authorized_keys" will actually **fail** some alternative cloud-init implementations...

Such as lima-init, or tinycloudinit. They have "ssh-authorized-keys" hardcoded, in their alternate parsers (awk)

jandubois:
> Such as lima-init

We can easily fix `lima-init`, but don't know about `tinycloudinit`. I did the same already for `ca-certs` and `ca_certs`: https://github.com/lima-vm/alpine-lima/blob/5a09f68c35124ba6049ec15fd972991c445856f7/lima-init.sh#L129-L130

```
awk -f- "${LIMA_CIDATA_MNT}"/user-data <<'EOF' > ${LIMA_CA_CERTS}
# Lima currently uses "ca-certs", which is deprecated and should be "ca_certs"
/^ca.certs:/ {
…
```

afbjorklund:
The tinycloudinit was just a hack that I made when playing with tinycorelinux, it is a similar awk script (with similar fix)

afbjorklund:
https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json

Example validation with `jv`

```
$ jv https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json cloud-config.yaml 
[I#] [S#] doesn't validate with https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json#
  [I#] [S#/allOf/8] allOf failed
    [I#] [S#/allOf/8/$ref] doesn't validate with '/$defs/cc_ca_certs'
      [I#/ca_certs] [S#/$defs/cc_ca_certs/properties/ca_certs/$ref] doesn't validate with '/$defs/ca_certs.properties'
        [I#/ca_certs/trusted] [S#/$defs/ca_certs.properties/properties/trusted/type] expected array, but got null
[...]

PR #2267, opened 2024-04-04
Validate generated user-data yaml with jsonschema

afbjorklund:
Currently there are 3 validation errors being reported:

`jsonschema: '/ca_certs/remove_defaults' does not validate with https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json#/allOf/8/$ref/properties/ca_certs/$ref/properties/remove_defaults/type: expected boolean, but got string`

`jsonschema: '/ca_certs/trusted' does not validate with https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json#/allOf/8/$ref/properties/ca_certs/$ref/properties/trusted/type: expected array, but got null`

`jsonschema: '/users/0' does not validate with https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json#/allOf/49/$ref/properties/users/items/oneOf/0/type: expected string, but got object`

And 1 deprecation warning, not yet being shown here.

```json
        "uid": {
          "description": "The user's ID. Default value [system default]",
          "oneOf": [
            {
              "type": "integer"
            },
            {
              "type": "string",
              "changed": true,
              "changed_description": "The use of ``string`` type is deprecated. Use an ``integer`` instead.",
              "changed_version": "22.3"
            }
          ]
        }

```

https://github.com/canonical/cloud-init/blob/main/cloudinit/config/schemas/schema-cloud-config-v1.json

----

Issue:

* #2265

The first two are fixed by

* #2266

But "ssh-authorized-keys" is a problem.

```json
        "ssh_authorized_keys": {
          "description": "List of SSH keys to add to user's authkeys file. Can not be combined with ``ssh_redirect_user``",
          "type": "array",
          "items": {
            "type": "string"
          },
          "minItems": 1
        },
```

https://git.launchpad.net/cloud-init/commit/?id=b27f713a

### DIFF
--- a/pkg/cidata/schema.go
+++ b/pkg/cidata/schema.go
@@ -1,7 +1,9 @@
 package cidata
 
 import (
+	"bytes"
 	_ "embed"
+	"fmt"
 	"strings"
 
 	"github.com/santhosh-tekuri/jsonschema/v5"
@@ -15,6 +17,9 @@ const schemaURL = "https://raw.githubusercontent.com/canonical/cloud-init/main/c
 var schemaText string
 
 func validateCloudConfig(userData []byte) error {
+	if !bytes.HasPrefix(userData, []byte("#cloud-config")) {
+		return fmt.Errorf("missing #cloud-config")
+	}
 	var m interface{}
 	err := yaml.Unmarshal(userData, &m)
 	if err != nil {
--- a/pkg/cidata/schema.go
+++ b/pkg/cidata/schema.go
@@ -26,7 +26,6 @@ func validateCloudConfig(userData []byte) error {
 		return err
 	}
 	compiler := jsonschema.NewCompiler()
-	compiler.ExtractAnnotations = true
 	if err := compiler.AddResource(schemaURL, strings.NewReader(schemaText)); err != nil {
 		return err
 	}
--- a/pkg/cidata/schema.go
+++ b/pkg/cidata/schema.go
@@ -1,9 +1,7 @@
 package cidata
 
 import (
-	"bytes"
 	_ "embed"
-	"fmt"
 	"strings"
 
 	"github.com/santhosh-tekuri/jsonschema/v5"
@@ -16,10 +14,7 @@ const schemaURL = "https://raw.githubusercontent.com/canonical/cloud-init/main/c
 //go:embed schemas/schema-cloud-config-v1.json
 var schemaText string
 
-func validateCloudConfig(userData []byte) error {
-	if !bytes.HasPrefix(userData, []byte("#cloud-config")) {
-		return fmt.Errorf("missing #cloud-config")
-	}
+func ValidateCloudConfig(userData []byte) error {
 	var m interface{}
 	err := yaml.Unmarshal(userData, &m)
 	if err != nil {
--- a/pkg/cidata/schema.go
+++ b/pkg/cidata/schema.go
@@ -1,0 +1,36 @@
+package cidata
+
+import (
+	_ "embed"
+	"strings"
+
+	"github.com/santhosh-tekuri/jsonschema/v5"
+	"gopkg.in/yaml.v3"
+)
+
+// schemaURL is the identifier, not the context
+const schemaURL = "https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json"
+
+//go:embed schemas/schema-cloud-config-v1.json
+var schemaText string
+
+func validateCloudConfig(userData []byte) error {
+	var m interface{}
+	err := yaml.Unmarshal(userData, &m)
+	if err != nil {
+		return err
+	}
+	compiler := jsonschema.NewCompiler()
+	compiler.ExtractAnnotations = true
+	if err := compiler.AddResource(schemaURL, strings.NewReader(schemaText)); err != nil {
+		return err
+	}
+	schema, err := compiler.Compile(schemaURL)
+	if err != nil {
+		return err
+	}
+	if err := schema.Validate(m); err != nil {
+		return err
+	}
+	return err
+}
--- a/cmd/limactl/validate.go
+++ b/cmd/limactl/validate.go
@@ -1,9 +1,13 @@
 package main
 
 import (
+	"bufio"
 	"fmt"
+	"io"
+	"os"
 
 	"github.com/lima-vm/lima/cmd/limactl/guessarg"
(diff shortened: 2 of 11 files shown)


afbjorklund:
As an example there was a warning in place for ca-certs, something that affected lima earlier (326bbaaeaae30ad6459634ecf6e1a4eb052a8084)

```json
    "cc_ca_certs": {
      "type": "object",
      "properties": {
        "ca_certs": {
          "$ref": "#/$defs/ca_certs.properties"
        },
        "ca-certs": {
          "allOf": [
            {
              "$ref": "#/$defs/ca_certs.properties"
            },
            {
              "deprecated": true,
              "deprecated_version": "22.3",
              "deprecated_description": "Use ``ca_certs`` instead."
            }
          ]
        }
      }
    },
```

But no such deprecation for ssh-authorized-keys.

afbjorklund:
The cidata tests are supposed to be failing, until the schema is valid (i.e. `ssh_authorized_keys`)

jandubois:
> The cidata tests are _supposed_ to be failing, until the schema is valid (i.e. `ssh_authorized_keys`)

I don't think validation should fail for using deprecated expressions. We cannot assume that all supported distros will have updated to a new `cloud-init` immediately, so Lima needs to continue to use the deprecated form until the new variant is supported everywhere.

This is kind of the point of deprecation instead of outright removal.

pkg/cidata/schema.go
```diff
@@ -0,0 +1,40 @@
...
+	"strings"
+
+	"github.com/santhosh-tekuri/jsonschema/v5"
+	"gopkg.in/yaml.v3"
+)
+
+const schemaURL = "https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json"
+
+//go:embed schema-cloud-config-v1.json
+var schemaText string
```
jandubois, line 16: I don't have time for a full review right now, but it bugs me that the URL points to `HEAD` of `main` in a repo, but we vendor a specific version. Shouldn't the URL match the version we embed?

I'm also not a big fan of manually vendoring files from other repos; I would rather create a git submodule for the other repo to create a fixed dependency to a specific commit.

I haven't tested it, but I assume we could make `schema-cloud-config-v1.json` a symlink into the submodule, a Go would still embed it correctly.

I know some people hate using submodules, so not sure if we want to go that route. Any opinions @AkihiroSuda ?

AkihiroSuda: I don't feel that using submodules is the best approach just for a single JSON file.

Probably we can just copy the JSON file into the repo?

pkg/cidata/schema.go
```diff
@@ -0,0 +1,40 @@
...
+	"bytes"
+	_ "embed"
+	"fmt"
+	"strings"
+
+	"github.com/santhosh-tekuri/jsonschema/v5"
+	"gopkg.in/yaml.v3"
+)
+
+const schemaURL = "https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json"
```
AkihiroSuda, line 13: What's the license of this file?

pkg/cidata/schema.go
jandubois: > Probably we can just copy the JSON file into the repo?

How do we even know which version of `cloud-init` this file comes from? It is just `HEAD` of `main` at some random point in time.

The name `schema-cloud-config-v1.json` is misleading, as it seems to imply it is a fixed version, but it it being updated all the time.

Eventually we will want to update it. What is the procedure to do it?

pkg/cidata/schema.go
jandubois: According to https://github.com/canonical/cloud-init/blob/main/LICENSE the repo is under your choice of GPL-3 or Apache-2.

pkg/cidata/schema.go
AkihiroSuda: `go generate` should wget it and validate the sha256sum?

afbjorklund: I can add a specific version of this file, using "main" was just what https://www.schemastore.org/json/ said

afbjorklund:
> I don't think validation should fail for using deprecated expressions.

It is not deprecated in the jsonschema, it is outright missing (and failing)

pkg/cidata/schema.go
afbjorklund: Will add a license statement for the file.

pkg/cidata/schema.go
afbjorklund: https://raw.githubusercontent.com/canonical/cloud-init/24.1/cloudinit/config/schemas/schema-cloud-config-v1.json

afbjorklund:
I added cloud-init jsonschema "latest" documentation version 24.1.3

Then I added my cloud-init patch as well: https://github.com/canonical/cloud-init/pull/5162

Note: the latest version on github is 24.1.4

Note: the jsonschema on `main` is different

```diff
index ff61dcaa6..a71cec996 100644
--- a/cloudinit/config/schemas/schema-cloud-config-v1.json
+++ b/cloudinit/config/schemas/schema-cloud-config-v1.json
@@ -835,6 +835,13 @@
                 "vault_password_file": {
                   "type": "string"
                 },
+                "verify_commit": {
+                  "type": "boolean",
+                  "default": false
+                },
+                "inventory": {
+                  "type": "string"
+                },
                 "module_name": {
                   "type": "string"
                 },
```

https://github.com/canonical/cloud-init/commit/b1bfa594b6e0679db6ed00ca83a9ddbecd5610b4

afbjorklund:
The patch for deprecating `ssh-authorized-keys` has now been accepted upstream, and is on `main`

afbjorklund:
Now when cloud-config.yaml is available in the instance directory, we can use `limactl validate` with it too.

* https://github.com/lima-vm/lima/pull/2271

afbjorklund:
It is possible to use the go version through cli, so we don't need to included it with lima...

```
go install github.com/santhosh-tekuri/jsonschema/cmd/jv@latest
```

And then it (`jv`) can be invoked with the schema file, similar to `check-jsonschema` above:

```
jv https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/schema-cloud-config-v1.json ~/.lima/default/cloud-config.yaml
```

PR #2028, opened 2023-11-24
Add hidden command to install the guest components

afbjorklund:
This will install the `lima-guestagent` in the instance.
    
It will install `nerdctl-full.tgz`, if it has been enabled.

Normally this is done by cidata, during the first boot...

Here it is just using `scp` and `ssh`, for convenience.

---

Workaround for:

* #2000

Something like:

```
INFO[0000] Copying "/home/anders/lima/_output/share/lima/lima-guestagent.Linux-aarch64" to lima-raspberrypi 
lima-guestagent.Linux-aarch64              100%   36MB   2.2MB/s   00:16    
INFO[0016] Installing /tmp/lima-guestagent to /usr/local/bin/lima-guestagent 
INFO[0017] Attempting to download the nerdctl archive    arch=aarch64 digest="sha256:ea30ab544c057e3a0457194ecd273ffbce58067de534bdfaffe4edf3a4da6357" location="https://github.com/containerd/nerdctl/releases/download/v1.6.2/nerdctl-full-1.6.2-linux-arm64.tar.gz"
INFO[0017] Using cache "/home/anders/.cache/lima/download/by-url-sha256/21921213c0426e38e2f26d2844446767e2a19c0b6655c774f9e4adbb5c368d8f/data" 
INFO[0017] Copying "nerdctl-full-1.6.2-linux-arm64.tar.gz" to lima-raspberrypi 
data                                       100%  205MB   2.7MB/s   01:16    
INFO[0094] Installing /tmp/nerdctl-full.tgz in /usr/local 
```

PR #2271, opened 2024-04-05
Generate cloud-config outside of cidata.iso too

### LINKED ISSUE #2268: Generate separate cloud-config.yaml for when not using cidata.iso
### Description

When looking at alternative implementations, such as Raspberry Pi or cloud instances, it could be useful to look at the generated basic cloud-init / cloud-config information. Without having it embedded as `user-data`, in the cidata.iso

Maybe we could also just create a stand-alone yaml file, like was done for the ignition yaml file. It doesn't include the boot scripts or the lima-guestagent/nerdctl-full, but does include basic stuff like the user and keys and so on (same args).

afbjorklund:
This does not include any mounts, networks, nor boot scripts.
    It is assumed that "reverse-sshfs" is being used, for mounts.
    
    It also does not include lima-guestagent, nerdctl-full.tgz,
    or any of the provisioning scripts that are in the cidata...

* #2268
* #2028

----
Closes #2268


pkg/cidata/cloud-config.yaml
AkihiroSuda, line 1: `pkg/cidata/cidata.TEMPLATE.d/user-data` should be now removed?
It is quite hard to maintain two (almost?) identical files

afbjorklund: There are some differences, but we could try to join and parametrize.

afbjorklund: Now they are using the same template file, with different parameters.

afbjorklund:
Note for testing, it is possible to extract the user-data of a running instance:

```
isoinfo -R -i ~/.lima/default/cidata.iso -x /user-data
```

pkg/cidata/cloud-config.yaml
```diff
@@ -0,0 +1 @@
+cidata.TEMPLATE.d/user-data
```
AkihiroSuda, line 1: Do we still need this symlink?
You can just mv user-data cloud-config.yaml

website/content/en/docs/dev/Internals/_index.md
```diff
@@ -37,6 +37,7 @@ Metadata:
 - `protected`: empty file, used by `limactl protect`
 
 cloud-init:
+- `cloud-config.yaml`: cloud-init configuration.
```
AkihiroSuda, line 40:
```suggestion
- `cloud-config.yaml`: cloud-init configuration. Modification to this file will be just discarded.
```

pkg/cidata/cloud-config.yaml
afbjorklund: No, it was not possible to embed a symlink

website/content/en/docs/dev/Internals/_index.md
afbjorklund: Made the file readonly (444) and added a comment

afbjorklund:
Seems like I managed to lock myself out of the home directory in CI, by thinko with "mode"

>    If the
>    file does not exist, WriteFile creates it with permissions perm (before
>    umask); otherwise WriteFile truncates it before writing, without changing
>    permissions.

So need to clean up the home directory, or add some workaround to remove the old files.

website/content/en/docs/dev/internals/_index.md
```diff
@@ -37,6 +37,7 @@ Metadata:
 - `protected`: empty file, used by `limactl protect`
 
 cloud-init:
+- `cloud-config.yaml`: cloud-init configuration, for reference only.
```
AkihiroSuda, line 40: Factory-reset should remove this yaml?

afbjorklund: It was created with the instance (`create`)

Maybe we should make sure to update it, like if the global config has changed.
But since we don't update lima.yaml either, we could also just leave it as-is...

AkihiroSuda: Shouldn't it be (re-)created on start?

afbjorklund: It is updated on start now, d22b200a33f1b4ad49ddd1f7078deedcac7702ce

afbjorklund: The main usage of this file was like:
1) `limactl create` the instance, using the "ext" driver
2) use the `cloud-config.yaml` to create an external VM
3) `limactl start` the instance, connect to existing VM

So it needs to be created before the start command runs.

AkihiroSuda: What about factory-reset?

afbjorklund: I'm not entirely sure what it _means_ to "factory reset" an instance, and it's not actually documented in the command... But usually it just means to restore settings to the defaults that it arrived with. But we don't reset the lima.yaml with the template, so I did not reset the generated cloud-config either. But we could add a line to regenerate it, maybe it is useful if you change the global '_config` or something? (Like loadDotSSHPubKeys)

cmd/limactl/factory-reset.go
```diff
@@ -63,6 +64,10 @@ func factoryResetAction(_ *cobra.Command, args []string) error {
 			}
 		}
 	}
+	if err := cidata.GenerateCloudConfig(inst.Dir, instName, inst.Config); err != nil {
```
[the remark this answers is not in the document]
AkihiroSuda: Could you add that context as comment lines in the Go code?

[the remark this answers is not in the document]
afbjorklund: Will do, as soon as I understand myself what a factory-reset means.

There is also new corner-case of running `limactl edit`, so maybe rethink...
The cidata.iso will be recreated on start, and we don't allow editing running.
Maybe the cloud-config.yaml should just be generated after every lima.yaml

website/content/en/docs/dev/internals/_index.md
AkihiroSuda: Yes, it should be either removed or regenerated

AkihiroSuda (approved):
Thanks

PR #2274, opened 2024-04-07
Use integer instead of string for cidata uid

afbjorklund:
`sudo cloud-init schema --system`

Cloud config schema deprecations: users.0.uid:  Changed in version 22.3.
The use of ``string`` type is deprecated. Use an ``integer`` instead.

* #2265

It was already an integer, in the template data and in LIMA_CIDATA_UID

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/user-data
+++ b/pkg/cidata/cidata.TEMPLATE.d/user-data
@@ -26,7 +26,7 @@ timezone: {{.TimeZone}}
 
 users:
   - name: "{{.User}}"
-    uid: "{{.UID}}"
+    uid: {{.UID}}
     homedir: "{{.Home}}"
     shell: /bin/bash
     sudo: ALL=(ALL) NOPASSWD:ALL


vasileknik76 (approved):

jandubois:
> Cloud config schema deprecations: users.0.uid: Changed in version 22.3.

22.3 was less than 2 years ago.

Did you verify that the "old" distros we support, like `almalinux-8`, `centos-stream-8`, etc. all ship with `cloud-init` 22.3 or later?

afbjorklund:
> Did you verify that the "old" distros we support, like `almalinux-8`, `centos-stream-8`, etc. all ship with `cloud-init` 22.3 or later?

I did not.

And apparently that was a bad thing, the previous behaviour (before 22.1) was to _ignore_ all integer UIDs...

https://github.com/canonical/cloud-init/commit/2837b835f101d81704f018a4f872b1d660eb6f3e in 22.1

https://github.com/canonical/cloud-init/commit/45b517a532477c31f3996d3cea5fbfd0249f2e2f in 22.2

Issue #1139, opened 2022-11-01
`limactl ls $INSTANCE` should exit with non-0 status if $INSTANCE does not exist


jandubois:
I wonder why `limactl ls $INSTANCE` just shows a warning and still has an exit code of `0` when the instance doesn't exist.

_Originally posted by @jandubois in https://github.com/lima-vm/lima/issues/957#issuecomment-1183514039_

sdemura:
I work around this with ```limactl list $INSTANCE --json | jq -res 'if . == [] then null else .[] | .status end'```

BajuMcBites:
Hello, I was wondering if I could be assigned this issue?

PR #2295, opened 2024-04-25
limactl ls $INSTANCE should exit with non-0 status if $INSTANCE does not exist #1139

BajuMcBites:
- switched the warning message into an error message

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -130,7 +130,7 @@ func listAction(cmd *cobra.Command, args []string) error {
 			if len(matches) > 0 {
 				instanceNames = append(instanceNames, matches...)
 			} else {
-				logrus.Warnf("No instance matching %v found.", arg)
+				return fmt.Errorf("no instance matching %v found", arg)
 			}
 		}
 	} else {


AkihiroSuda (approved):
Thanks

jandubois:
I realize this PR has already been merged, but I disagree with the implementation, as it introduces a regression:

I have a single `alpine` instance running. Before this PR I get the following output:


```console
$ limactl ls foo alpine bar
WARN[0000] No instance matching foo found.
WARN[0000] No instance matching bar found.
NAME      STATUS     SSH            VMTYPE    ARCH      CPUS    MEMORY    DISK      DIR
alpine    Stopped    127.0.0.1:0    qemu      x86_64    4       4GiB      100GiB    ~/.lima/alpine
```

With this PR the command stops at the first instance that is not found:

```console
$ limactl ls foo alpine bar
FATA[0000] no instance matching foo found
```

So I think we need to record the fact that at least one requested instance was not found, but only exit with `os.Exit(1)` after the information about the existing instances (and warnings about additional missing instances) have been printed.

I would like to see this fixed in a follow-up PR.

BajuMcBites:
Sorry about this, I did not think to account for multiple instances. I can work on another PR to fix this issue. Should it still log the "No instance matching <name> found." as warnings and then exit with 1, or is there a different message type that is preferred?

jandubois:
> I can work on another PR to fix this issue. Should it still log the "No instance matching found." as warnings and then exit with 1

Yes, I think the output should remain unchanged, just the exit code should be non-0 when at least one instance could not be found.

The easiest way to exit with a non-0 code would be to call `os.Exit(1)`.

Note that there are multiple output formats for this command, with some early return statements. Make sure that all of them are being covered!

PR #2299, opened 2024-04-26
Bump alpine image to 0.2.38

jandubois:
* Look for ssh_authorized_keys without dashes
* Bump QEMU version in tonistiigi/binfmt from 7.0.0 → 8.1.5


AkihiroSuda (approved):
Thanks

PR #2301, opened 2024-04-26
limactl ls $INSTANCE should exit with non-0 status if $INSTANCE does not exist #1139 #2295

BajuMcBites:
- I believe this should fix the issue where nothing is printed on an unmatched list error caused by PR #2295
- Prints the matching instances, and then exits with 1
- Please let me know if I missed anything

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -124,13 +124,15 @@ func listAction(cmd *cobra.Command, args []string) error {
 	}
 
 	instanceNames := []string{}
+	unmatchedInstances := false
 	if len(args) > 0 {
 		for _, arg := range args {
 			matches := instanceMatches(arg, allinstances)
 			if len(matches) > 0 {
 				instanceNames = append(instanceNames, matches...)
 			} else {
-				return fmt.Errorf("no instance matching %v found", arg)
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Please squash commits

cmd/limactl/list.go
```diff
@@ -174,6 +179,17 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+		if len(instanceNames) > 0 {
+			err := store.PrintInstances(out, instances, format, &options)
+			if err != nil {
+				return err
+			}
+		}
+		os.Exit(1)
+	}
+
 	return store.PrintInstances(out, instances, format, &options)
```
jandubois, line 193: This looks simpler (and avoids code duplication):

```suggestion
	err := store.PrintInstances(out, instances, format, &options)
	if err == nil && unmatchedInstance {
		os.Exit(1)
	}
	return err
```

I have also a slight preference to using the plural form `unmatchedInstances` over the singular, to match the "one or more" part.

jandubois (approved):
Thanks, LGTM. Any reason this is still in draft?

BajuMcBites:
No, I was just waiting for the tests to run before I marked it as ready for review. Thanks

PR #2306, opened 2024-04-30
Add command to generate jsonschema for limayaml

### LINKED ISSUE #2305: Generate jsonschema for validating the limayaml
### Description

There was some discussion about using json-schema.org for documentation, which doesn't go very well...

But it could still be used for syntax-checking the lima.yaml, similar to what is available for cloud-config.yaml

* https://github.com/lima-vm/lima/discussions/1489

* https://github.com/invopop/jsonschema

afbjorklund:
Closes #2305 

https://pypi.org/project/check-jsonschema/

Actually found two bugs, with the current code:

```
Schema validation errors were encountered.
  examples/default.yaml::$.vmType: None is not of type 'string'
  examples/default.yaml::$.os: None is not of type 'string'
  examples/default.yaml::$.arch: None is not of type 'string'
  examples/default.yaml::$.cpuType.armv7l: None is not of type 'string'
  examples/default.yaml::$.cpuType.aarch64: None is not of type 'string'
  examples/default.yaml::$.cpuType.x86_64: None is not of type 'string'
  examples/default.yaml::$.cpus: None is not of type 'integer'
  examples/default.yaml::$.memory: None is not of type 'string'
  examples/default.yaml::$.disk: None is not of type 'string'
```
```
Schema validation errors were encountered.
  examples/docker.yaml::$.probes[0]: Additional properties are not allowed ('hint', 'script' were unexpected)
  examples/docker.yaml::$.probes[0]: 'Mode' is a required property
  examples/docker.yaml::$.probes[0]: 'Description' is a required property
  examples/docker.yaml::$.probes[0]: 'Script' is a required property
  examples/docker.yaml::$.probes[0]: 'Hint' is a required property
```

But the rest of it is just about adding "nullable".

----

To test this feature: (hidden command: `limactl generate-jsonschema`)

`make schema-limayaml.json`

And use it to test: (https://check-jsonschema.readthedocs.io/)

`make check-jsonschema`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -422,6 +422,14 @@ ifeq ($(native_compiling),true)
 endif
 
 ################################################################################
+schema-limayaml.json: _output/bin/limactl$(exe)
+	$< generate-jsonschema >$@
+
+.PHONY: check-jsonschema
+check-jsonschema: schema-limayaml.json
+	check-jsonschema --schemafile $< examples/default.yaml
+
+################################################################################
 .PHONY: diagrams
 diagrams: docs/lima-sequence-diagram.png
 docs/lima-sequence-diagram.png: docs/images/lima-sequence-diagram.puml
--- a/cmd/limactl/genschema.go
+++ b/cmd/limactl/genschema.go
@@ -1,0 +1,31 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/invopop/jsonschema"
+	"github.com/lima-vm/lima/pkg/limayaml"
+	"github.com/spf13/cobra"
+)
+
+func newGenSchemaCommand() *cobra.Command {
+	genschemaCommand := &cobra.Command{
+		Use:    "generate-jsonschema",
+		Short:  "Generate json-schema document",
+		Args:   WrapArgsError(cobra.NoArgs),
+		RunE:   genschemaAction,
+		Hidden: true,
+	}
+	return genschemaCommand
+}
+
+func genschemaAction(cmd *cobra.Command, _ []string) error {
+	schema := jsonschema.Reflect(&limayaml.LimaYAML{})
+	j, err := json.MarshalIndent(schema, "", "    ")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(cmd.OutOrStdout(), string(j))
+	return err
+}
--- a/cmd/limactl/genschema.go
+++ b/cmd/limactl/genschema.go
@@ -22,6 +22,12 @@ func newGenSchemaCommand() *cobra.Command {
 
 func genschemaAction(cmd *cobra.Command, _ []string) error {
 	schema := jsonschema.Reflect(&limayaml.LimaYAML{})
+	// allow Disk to be either string (name) or object (struct)
+	schema.Definitions["Disk"].Type = "" // was: "object"
+	schema.Definitions["Disk"].OneOf = []*jsonschema.Schema{
+		{Type: "string"},
+		{Type: "object"},
+	}
 	j, err := json.MarshalIndent(schema, "", "    ")
 	if err != nil {
 		return err
--- a/cmd/limactl/genschema.go
+++ b/cmd/limactl/genschema.go
@@ -7,6 +7,7 @@ import (
 	"github.com/invopop/jsonschema"
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/spf13/cobra"
+	orderedmap "github.com/wk8/go-ordered-map/v2"
 )
 
 func newGenSchemaCommand() *cobra.Command {
@@ -20,6 +21,22 @@ func newGenSchemaCommand() *cobra.Command {
 	return genschemaCommand
 }
 
+func toAny(args []string) []any {
+	result := []any{nil}
+	for _, arg := range args {
+		result = append(result, arg)
+	}
+	return result
+}
+
+func getProp(props *orderedmap.OrderedMap[string, *jsonschema.Schema], key string) *jsonschema.Schema {
+	value, ok := props.Get(key)
+	if !ok {
+		return nil
+	}
+	return value
+}
+
 func genschemaAction(cmd *cobra.Command, _ []string) error {
 	schema := jsonschema.Reflect(&limayaml.LimaYAML{})
 	// allow Disk to be either string (name) or object (struct)
@@ -28,6 +45,11 @@ func genschemaAction(cmd *cobra.Command, _ []string) error {
 		{Type: "string"},
 		{Type: "object"},
 	}
+	properties := schema.Definitions["LimaYAML"].Properties
+	getProp(properties, "os").Enum = toAny(limayaml.OSTypes)
+	getProp(properties, "arch").Enum = toAny(limayaml.ArchTypes)
+	getProp(properties, "mountType").Enum = toAny(limayaml.MountTypes)
+	getProp(properties, "vmType").Enum = toAny(limayaml.VMTypes)
 	j, err := json.MarshalIndent(schema, "", "    ")
 	if err != nil {
 		return err
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -76,6 +76,13 @@ const (
 	WSL2 VMType = "wsl2"
 )
 
+var (
+	OSTypes    = []OS{LINUX}
+	ArchTypes  = []Arch{X8664, AARCH64, ARMV7L, RISCV64}
+	MountTypes = []MountType{REVSSHFS, NINEP, VIRTIOFS, WSLMount}
+	VMTypes    = []VMType{QEMU, VZ, WSL2}
+)
+
 type VMOpts struct {
 	QEMU QEMUOpts `yaml:"qemu,omitempty" json:"qemu,omitempty"`
 }
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -120,7 +120,7 @@ type Disk struct {
 
 type Mount struct {
 	Location   string   `yaml:"location" json:"location"` // REQUIRED
-	MountPoint string   `yaml:"mountPoint,omitempty" json:"mountPoint,omitempty"`
+	MountPoint *string  `yaml:"mountPoint,omitempty" json:"mountPoint,omitempty" jsonschema:"nullable"`
 	Writable   *bool    `yaml:"writable,omitempty" json:"writable,omitempty" jsonschema:"nullable"`
 	SSHFS      SSHFS    `yaml:"sshfs,omitempty" json:"sshfs,omitempty"`
 	NineP      NineP    `yaml:"9p,omitempty" json:"9p,omitempty"`
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -7,46 +7,46 @@ import (
 )
 
 type LimaYAML struct {
-	MinimumLimaVersion    *string       `yaml:"minimumLimaVersion,omitempty" json:"minimumLimaVersion,omitempty"`
-	VMType                *VMType       `yaml:"vmType,omitempty" json:"vmType,omitempty"`
+	MinimumLimaVersion    *string       `yaml:"minimumLimaVersion,omitempty" json:"minimumLimaVersion,omitempty" jsonschema:"nullable"`
+	VMType                *VMType       `yaml:"vmType,omitempty" json:"vmType,omitempty" jsonschema:"nullable"`
 	VMOpts                VMOpts        `yaml:"vmOpts,omitempty" json:"vmOpts,omitempty"`
-	OS                    *OS           `yaml:"os,omitempty" json:"os,omitempty"`
-	Arch                  *Arch         `yaml:"arch,omitempty" json:"arch,omitempty"`
+	OS                    *OS           `yaml:"os,omitempty" json:"os,omitempty" jsonschema:"nullable"`
+	Arch                  *Arch         `yaml:"arch,omitempty" json:"arch,omitempty" jsonschema:"nullable"`
 	Images                []Image       `yaml:"images" json:"images"` // REQUIRED
-	CPUType               CPUType       `yaml:"cpuType,omitempty" json:"cpuType,omitempty"`
-	CPUs                  *int          `yaml:"cpus,omitempty" json:"cpus,omitempty"`
-	Memory                *string       `yaml:"memory,omitempty" json:"memory,omitempty"` // go-units.RAMInBytes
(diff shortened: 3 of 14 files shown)


afbjorklund:
Remaining "issues".

```
Schema validation errors were encountered.
  examples/default.yaml::$.additionalDisks: None is not of type 'array'
  examples/default.yaml::$.mounts[0].mountPoint: None is not of type 'string'
  examples/default.yaml::$.mounts[0].writable: None is not of type 'boolean'
  examples/default.yaml::$.mounts[0].sshfs.cache: None is not of type 'boolean'
  examples/default.yaml::$.mounts[0].sshfs.followSymlinks: None is not of type 'boolean'
  examples/default.yaml::$.mounts[0].sshfs.sftpDriver: None is not of type 'string'
  examples/default.yaml::$.mounts[0].9p.securityModel: None is not of type 'string'
  examples/default.yaml::$.mounts[0].9p.protocolVersion: None is not of type 'string'
  examples/default.yaml::$.mounts[0].9p.msize: None is not of type 'string'
  examples/default.yaml::$.mounts[0].9p.cache: None is not of type 'string'
  examples/default.yaml::$.mountType: None is not of type 'string'
  examples/default.yaml::$.mountInotify: None is not of type 'boolean'
  examples/default.yaml::$.ssh.loadDotSSHPubKeys: None is not of type 'boolean'
  examples/default.yaml::$.ssh.forwardAgent: None is not of type 'boolean'
  examples/default.yaml::$.ssh.forwardX11: None is not of type 'boolean'
  examples/default.yaml::$.ssh.forwardX11Trusted: None is not of type 'boolean'
  examples/default.yaml::$.firmware.legacyBIOS: None is not of type 'boolean'
  examples/default.yaml::$.audio.device: None is not of type 'string'
  examples/default.yaml::$.video.display: None is not of type 'string'
  examples/default.yaml::$.video.vnc.display: None is not of type 'string'
  examples/default.yaml::$.upgradePackages: None is not of type 'boolean'
  examples/default.yaml::$.containerd.system: None is not of type 'boolean'
  examples/default.yaml::$.containerd.user: None is not of type 'boolean'
  examples/default.yaml::$.guestInstallPrefix: None is not of type 'string'
  examples/default.yaml::$.networks: None is not of type 'array'
  examples/default.yaml::$.hostResolver.enabled: None is not of type 'boolean'
  examples/default.yaml::$.hostResolver.ipv6: None is not of type 'boolean'
  examples/default.yaml::$.hostResolver.hosts: None is not of type 'object'
  examples/default.yaml::$.propagateProxyEnv: None is not of type 'boolean'
  examples/default.yaml::$.caCerts.removeDefaults: None is not of type 'boolean'
  examples/default.yaml::$.caCerts.files: None is not of type 'array'
  examples/default.yaml::$.caCerts.certs: None is not of type 'array'
  examples/default.yaml::$.rosetta.enabled: None is not of type 'boolean'
  examples/default.yaml::$.rosetta.binfmt: None is not of type 'boolean'
  examples/default.yaml::$.plain: None is not of type 'boolean'
  examples/default.yaml::$.timezone: None is not of type 'string'
```

mac2net:
Don't you think this universal rejection of “none” is more an opinionated code base rather than a validation error?

afbjorklund:
> Don't you think this universal rejection of “none” is more an opinionated code base rather than a validation error?

It's not a rejection (as such), there is a patch for it (to not have to decorate the go struct) but it has not been merged yet

* https://github.com/invopop/jsonschema/pull/110

With the patch, one could (perhaps) mark all string pointers and arrays as accepting `null` as a valid value (by default).

* https://github.com/invopop/jsonschema/pull/106

---

But it does catch some issues, with string being fed with null - which is valid in JavaScript but not really in Go...

Having to decorate the pointers and arrays I think is not so intrusive, so it seems like a valid workaround to me.

afbjorklund:
```console
$ check-jsonschema --schemafile schema-limayaml.json examples/default.yaml
ok -- validation done
```

These arrays have to allow `null`, since they are given that way in the default yaml:

```yaml
additionalDisks:
# disks should either be a list of disk name strings, for example:
# - "data"
# or a list of disk objects with extra parameters, for example:
# - name: "data"
#   format: true
#   fsType: "ext4"
```

```yaml
networks:
# Lima can manage daemons for networks defined in $LIMA_HOME/_config/networks.yaml
# automatically. The socket_vmnet binary must be installed into
# secure locations only alterable by the "root" user.
# The same applies to vde_switch and vde_vmnet for the deprecated VDE mode.
# - lima: shared
#   # MAC address of the instance; lima will pick one based on the instance name,
#   # so DHCP assigned ip addresses should remain constant over instance restarts.
#   macAddress: ""
#   # Interface name, defaults to "lima0", "lima1", etc.
#   interface: ""
#
# Lima can also connect to "unmanaged" networks addressed by "socket". This
# means that the daemons will not be controlled by Lima, but must be started
# before the instance.  The interface type (host, shared, or bridged) is
# configured in socket_vmnet and not in lima.
# - socket: "/var/run/socket_vmnet"

```

A better alternative would be to comment out the key as well, and leave them undefined?

The other keys are only allowing `null` values, when they **explicitly** are given that way:

```yaml
vmType: null
os: null
arch: null
cpus: null
memory: null
disk: null
mounts:
- location: "~"
  mountPoint: ""
  writable: null
  sshfs:
    cache: null
    followSymlinks: null
    sftpDriver: null
  9p:
    securityModel: null
    protocolVersion: null
    msize: null
    cache: null
mountType: null
mountInotify: null
additionalDisks:
ssh:
  localPort: 0
  loadDotSSHPubKeys: null
  forwardAgent: null
  forwardX11: null
  forwardX11Trusted: null
caCerts:
  removeDefaults: null
  files:
  certs:
upgradePackages: null
containerd:
  system: null
  user: null
rosetta:
  enabled: null
  binfmt: null
timezone: null
firmware:
  legacyBIOS: null
audio:
  device: null
video:
  display: null
  vnc:
    display: null
propagateProxyEnv: null
networks:
hostResolver:
  enabled: null
  ipv6: null
  hosts:
guestInstallPrefix: null
plain: null
```

Currently the only **required** key is "images", but even that might be optional (for ext).

afbjorklund:
Here is the empty yaml:

```yaml
images: []
```

And here is default yaml:

```yaml
images:
- location: https://cloud-images.ubuntu.com/releases/24.04/release-20240423/ubuntu-24.04-server-cloudimg-amd64.img
  arch: x86_64
  digest: sha256:32a9d30d18803da72f5936cf2b7b9efcb4d0bb63c67933f17e3bdfd1751de3f3
- location: https://cloud-images.ubuntu.com/releases/24.04/release-20240423/ubuntu-24.04-server-cloudimg-arm64.img
  arch: aarch64
  digest: sha256:c841bac00925d3e6892d979798103a867931f255f28fefd9d5e07e3e22d0ef22
- location: https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-amd64.img
  arch: x86_64
- location: https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-arm64.img
  arch: aarch64
cpuType:
  aarch64: ""
  armv7l: ""
  riscv64: ""
  x86_64: ""
mounts:
- location: ~
- location: /tmp/lima
  writable: true
ssh:
  localPort: 0
```

Hopefully the images can be made more similar to the containerd archives, and available from the code instead.

Or maybe **not** the go code, but to keep the nerdctl-full and the distro images in some kind of secondary config ?

```yaml
---
"ubuntu-24.04":
  images:
    arch: "x86_64"
    arch: "aarch64"
  arch: "x86_64"
  arch: "aarch64"

---
"nerdctl-full:1.7.6":
  images:
  - location: "https://github.com/containerd/nerdctl/releases/download/1.7.6/nerdctl-full-1.7.6-linux-amd64.tar.gz"
    arch: "x86_64"
    digest: "sha256:2c841e097fcfb5a1760bd354b3778cb695b44cd01f9f271c17507dc4a0b25606"
  - location: "https://github.com/containerd/nerdctl/releases/download/1.7.6/nerdctl-full-1.7.6-linux-arm64.tar.gz"
    arch: "aarch64"
    digest: "sha256:77c747f09853ee3d229d77e8de0dd3c85622537d82be57433dc1fca4493bab95"
    # No arm-v7
    # No riscv64
```

And then expanded by lima.

afbjorklund:
The resulting JSON is supposed to be uploaded to some external URL under https://schemastore.org/

Perhaps hosted on the lima-vm.io site, or as a fallback loaded from the git repo directly (like [cloud-config](https://raw.githubusercontent.com/canonical/cloud-init/main/cloudinit/config/schemas/versions.schema.cloud-config.json))

Then it can be used while editing the lima.yaml, even though it still needs to pass `limactl validate`.

But the hidden command and the make target is mostly for developers / while changing default.yaml

AkihiroSuda:
Needs rebase

Makefile
```diff
@@ -400,6 +400,14 @@ ifeq ($(native_compiling),true)
 		--output _output --prefix $(PREFIX)
 endif
 
+################################################################################
+schema-limayaml.json: _output/bin/limactl$(exe)
+	$< generate-schema >$@
```
AkihiroSuda, line 405: Maybe
```suggestion
	$< generate-jsonschema >$@
```

afbjorklund:
Another `null` array added.

```yaml
mountTypesUnsupported: null
```

AkihiroSuda (dismissed):
Thanks

AkihiroSuda:
Sorry, needs rebase again

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Still needs another rebase

pkg/limayaml/limayaml.go
```diff
@@ -7,51 +7,53 @@ import (
...
-	CPUType               CPUType       `yaml:"cpuType,omitempty" json:"cpuType,omitempty"`
-	CPUs                  *int          `yaml:"cpus,omitempty" json:"cpus,omitempty"`
-	Memory                *string       `yaml:"memory,omitempty" json:"memory,omitempty"` // go-units.RAMInBytes
-	Disk                  *string       `yaml:"disk,omitempty" json:"disk,omitempty"`     // go-units.RAMInBytes
-	AdditionalDisks       []Disk        `yaml:"additionalDisks,omitempty" json:"additionalDisks,omitempty"`
+	CPUType               CPUType       `yaml:"cpuType,omitempty" json:"cpuType,omitempty" jsonschema:"nullable"`
+	CPUs                  *int          `yaml:"cpus,omitempty" json:"cpus,omitempty" jsonschema:"nullable"`
+	Memory                *string       `yaml:"memory,omitempty" json:"memory,omitempty" jsonschema:"nullable"`                   // go-units.RAMInBytes
+	Disk                  *string       `yaml:"disk,omitempty" json:"disk,omitempty" jsonschema:"nullable"`                       // go-units.RAMInBytes
+	AdditionalDisks       []Disk        `yaml:"additionalDisks,omitempty" json:"additionalDisks,omitempty" jsonschema:"nullable"` // commented out
```
jandubois, line 20: What does "commented out" mean?

pkg/limayaml/limayaml.go
```diff
@@ -7,51 +7,53 @@ import (
...
+	Plain                *bool          `yaml:"plain,omitempty" json:"plain,omitempty" jsonschema:"nullable"`
+	TimeZone             *string        `yaml:"timezone,omitempty" json:"timezone,omitempty" jsonschema:"nullable"`
+	NestedVirtualization *bool          `yaml:"nestedVirtualization,omitempty" json:"nestedVirtualization,omitempty" jsonschema:"nullable"`
 }
 
 type (
 	OS        = string
+	OSType    = OS
 	Arch      = string
+	ArchType  = Arch
```
jandubois, line 56: Why are we adding `OSType` and `ArchType` aliases instead of renaming `OS` and `Arch`?

pkg/limayaml/limayaml.go
afbjorklund: I think it meant to say "the values were commented out" (to motivate the need for a `null` array)

But it could be read as "the key should be commented out", and then it would **not** need to be null

Anyway, needs to be removed or clarified.

Thanks!

pkg/limayaml/limayaml.go
afbjorklund: It was a much bigger change, but it (Types) should probably have gone in a separate PR (or two)

jandubois: It would just change another 34 lines, and avoid adding the 2 new aliases. I think you should include it in your PR, maybe just as a separate commit: https://github.com/jandubois/lima/commit/8e32e8a

jandubois: I guess it balloons a bit if you also rename local/struct variables from `arch` → `archType` etc, but I don't think you would have to do that.

afbjorklund: Renaming the types can be done separately, if at all. Especially if it would affect the field names or variables:

```
        VMType                *VMType       `yaml:"vmType,omitempty" json:"vmType,omitempty" jsonschema:"nullable"`
        OS                    *OS           `yaml:"os,omitempty" json:"os,omitempty" jsonschema:"nullable"`
        Arch                  *Arch         `yaml:"arch,omitempty" json:"arch,omitempty" jsonschema:"nullable"`
        CPUType               CPUType       `yaml:"cpuType,omitempty" json:"cpuType,omitempty" jsonschema:"nullable"`
}
```

AkihiroSuda:
Needs rebase

afbjorklund:
I left out the information about which arrays are `null` in the default template, I don't think anyone cares...

But left the "nullable" in, so that we **don't** get validation errors:

```
  examples/default.yaml::$.additionalDisks: None is not of type 'array'
  examples/default.yaml::$.mountTypesUnsupported: None is not of type 'array'
  examples/default.yaml::$.networks: None is not of type 'array'
  examples/default.yaml::$.caCerts.files: None is not of type 'array'
  examples/default.yaml::$.caCerts.certs: None is not of type 'array'
```

Otherwise they could commented out, and handled by the "omitempty"

cmd/limactl/genschema.go
```diff
@@ -0,0 +1,59 @@
...
+	getProp(properties, "arch").Enum = toAny(limayaml.ArchTypes)
+	getProp(properties, "mountType").Enum = toAny(limayaml.MountTypes)
+	getProp(properties, "vmType").Enum = toAny(limayaml.VMTypes)
+	j, err := json.MarshalIndent(schema, "", "    ")
+	if err != nil {
+		return err
+	}
+	_, err = fmt.Fprintln(cmd.OutOrStdout(), string(j))
+	return err
+}
```
AkihiroSuda, line 59: Can we have a test? What happens when this is out of the sync with the YAML struct definition?

AkihiroSuda (approved):

Issue #1621, opened 2023-06-12
VM fails to start on M1 CPU when CPU cores set to more than the host has


bowlofeggs:
### Description

Greetings,

I recently discovered why https://github.com/abiosoft/colima was taking so long to start on my M1, and recently refused to start at all. The symptom I had been observing was a seeming timeout. I would see lots of logs with `Waiting for the essential requirement 1 of 3: "ssh"` printed to the terminal, and then ultimately it would bail out saying that ssh never came up (I assume this is a timeout.)

It turned out that if you set the number of CPU cores that the guest uses to be higher than the number of cores the M1 has, you end up with devastating performance. I had assumed incorrectly that the M1 had SMT, and so I had set the guest to use 16 cores since the M1 I use has 8 physical cores (plus 2 low power cores).

I am unsure if it is slow because it starts to use the low power cores, or if it's just a bug in the hypervisor, or both.

I have a few suggestions as to how `lima` could help users like me who make such poor assumptions, and all of them involve detecting the number of threads supported by the physical hardware:
* Log a warning if the user configured more cores than the hardware supports.
* Exit with an error if the user configured more cores than the hardware supports.
* Silently correct to `min(num_cores_detected, num_cores_requested)`.

AkihiroSuda:
> Correct to min(num_cores_detected, num_cores_requested) and log a warning.

SGTM

NotPranav0:
Hello, my friend @BajuMcBites and I are students at UT Austin and after taking a virtualization class, we wanted to contribute to this project. Could we be assigned this issue? Thanks.

jandubois:
> Could we be assigned this issue?

I've assigned the issue to you; I couldn't add your friend; maybe they need to comment on the issue first? Anyways, it doesn't matter; just create a PR and we can discuss from there.

BajuMcBites:
Hello, I am @NotPranav0's partner. Thanks.

jandubois:
> Hello, I am @NotPranav0's partner.

Thanks, that seems to have enabled it. I've assigned you as well now.

NotPranav0:
Hey, we've managed to figure out the issue. Can we just make a new branch and submit a PR?

AkihiroSuda:
> Hey, we've managed to figure out the issue. Can we just make a new branch and submit a PR?

Yes, please, thanks

PR #2309, opened 2024-05-01
VM fails to start on M1 CPU when CPU cores set to more than the host has #1621

NotPranav0:


NotPranav0:
how do i run the linter?

pkg/limayaml/validate.go
```diff
@@ -108,6 +108,10 @@ func Validate(y LimaYAML, warn bool) error {
 		return errors.New("field `cpus` must be set")
 	}
 
+	if *y.CPUs > runtime.NumCPU() {
+		return fmt.Errorf("field `cpus` is set to %d, which is greater than the number of CPUs available (%d)", *y.CPUs, runtime.NumCPU())
+	}
```
AkihiroSuda, line 113: Wondering if the VM driver should automatically decrease the CPUs (with a warning)?

jandubois: We could (i have no strong opinion either way), but why do it in the driver and not in `FillDefault`? We set the default memory in there based on actual memory available, so why not implement the cap on the number of CPUs there as well?

jandubois: I just remembered: we may be executing `FillDefault` several times (we read the configurations of all instances during startup to look for things like shared network daemons, or shared additional disks). So any warnings generated while processing `lima.yaml` will be displayed multiple times, and we may be displaying warnings for other instances too.

So doing it in the driver might be preferable after all.

NotPranav0: Sorry, I have a question.
Am I correct in thinking this should be implemented in `vz_driver_darwin.go` and `qemu_driver.go`?
Thanks

NotPranav0: i'm able to generate a warning, but I'm not sure how to edit the yaml config to permanently set the number of cpu's.
setting `y.cpus` doesn't edit the yaml when viewed using `limactl edit`

PR #851, opened 2022-05-10
vmnet: Support socket_vmnet; deprecate vde_vmnet

AkihiroSuda:
socket_vmnet is similar to vde_vmnet but does not depend on VDE.

https://github.com/lima-vm/socket_vmnet

See `docs/network.md` for how to create `networks.yaml` with `socketVMNet`.
When both `socketVMNet` and `vdeVMNet` (deprecated) are present in the YAML, `socketVMNet` is chosen.

- - -
## iperf3 benchmark (host -> guest)

Mode                 | Shared (NAT) | Bridged
---------------|--------------|----------
socket_vmnet | 0.66 Gbps      | 1.23 Gbps
vde_vmnet      | 0.27 Gbps      | 0.31 Gbps

Tested on MacBook Pro 2020 (Intel), macOS 12
Lima commit 8db31e8087272da1c848d5d6d23f680004ad7d45 , socket_vmnet v1.0.0-alpha.0, vde_vmnet v0.6.0

Known issue: the throughput of the Shared (NAT) interface can be slower when both the Shared (NAT) and the Bridged interfaces are configured

PR #2311, opened 2024-05-01
Remove  the support for `vde_vmnet` (Deprecated since Sep 2022, in favor of `socket_vmnet`)

### LINKED ISSUE #2096: Remove  the support for `vde_vmnet` (Deprecated since Sep 2022, in favor of `socket_vmnet`)
`vde_vmnet` has been deprecated since Lima v0.12.0 (Sep 2022), in favor of `socket_vmnet`:
- https://github.com/lima-vm/lima/pull/851

We can remove the support for `vde_vmnet` as soon as Rancher Desktop stops using it:
- https://github.com/rancher-sandbox/rancher-desktop/issues/6248

AkihiroSuda:
`vde_vmnet` has been deprecated since Lima v0.12.0 (Sep 2022), in favor of `socket_vmnet`:
- https://github.com/lima-vm/lima/pull/851

- - -
Close #2096

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -319,7 +319,6 @@ networks:
 # Lima can manage daemons for networks defined in $LIMA_HOME/_config/networks.yaml
 # automatically. The socket_vmnet binary must be installed into
 # secure locations only alterable by the "root" user.
-# The same applies to vde_switch and vde_vmnet for the deprecated VDE mode.
 # - lima: shared
 #   # MAC address of the instance; lima will pick one based on the instance name,
 #   # so DHCP assigned ip addresses should remain constant over instance restarts.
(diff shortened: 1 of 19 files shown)


jandubois (approved):
The logic in `pkg/networks/*` should be rewritten to eliminate the `daemon` argument, now that there is only a single valid setting for it. Or do you expect that we will have another daemon choice again in the future?

AkihiroSuda:
> The logic in `pkg/networks/*` should be rewritten to eliminate the `daemon` argument, now that there is only a single valid setting for it. Or do you expect that we will have another daemon choice again in the future?

👍 on elimination

AkihiroSuda:
Regression
- https://github.com/lima-vm/lima/issues/2313

PR #2316, opened 2024-05-03
Refactor CPUType to add defaultCPUType

afbjorklund:
Currently there is no check that the arch is actually in the cpu type, specifically this test case:

```go
                Arch:   ptr.Of("unknown"),
                CPUType: map[Arch]string{
                        AARCH64: "arm64",
                        ARMV7L:  "armhf",
                        X8664:   "amd64",
                        RISCV64: "riscv64",
                },
```

But normally this should be handled in the validate, which will reject any arch not in the set.

```go
        case X8664, AARCH64, ARMV7L, RISCV64:
```

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -38,6 +38,35 @@ const (
 
 var IPv4loopback1 = net.IPv4(127, 0, 0, 1)
 
+func defaultCPUType() CPUType {
+	cpuType := map[Arch]string{
+		AARCH64: "cortex-a72",
+		ARMV7L:  "cortex-a7",
+		// Since https://github.com/lima-vm/lima/pull/494, we use qemu64 cpu for better emulation of x86_64.
+		X8664:   "qemu64",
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks

pkg/limayaml/defaults.go
```diff
@@ -38,6 +38,35 @@ const (
 
 var IPv4loopback1 = net.IPv4(127, 0, 0, 1)
 
+func defaultCPUType() CPUType {
+	cpuType := map[Arch]string{
+		AARCH64: "cortex-a72",
+		ARMV7L:  "cortex-a7",
+		// Since https://github.com/lima-vm/lima/pull/494, we use qemu64 cpu for better emulation of x86_64.
+		X8664:   "qemu64",
+		RISCV64: "rv64", // FIXME: what is the right choice for riscv64?
```
AkihiroSuda, line 47: nit: I guess the FIXME can be removed now

PR #2317, opened 2024-05-04
Fix "FATA[0002] networks.yaml field `paths.varRun` error: path "." is not an absolute path"

### LINKED ISSUE #2313: [master] `FATA[0002] networks.yaml field `paths.varRun` error: path "." is not an absolute path`
### Description

In the master branch of Lima (05ed15108b952f5cc7a0664eaee57f385360cec9), `limactl start template://vmnet` fails with an existing  `networks.yaml`

```console
$ limactl start template://vmnet 
? Creating an instance "vmnet" Proceed with the current configuration
FATA[0002] networks.yaml field `paths.varRun` error: path "." is not an absolute path
```

`~/.lima/_config/networks.yaml`:
```yaml
# Path to socket_vmnet executable. Because socket_vmnet is invoked via sudo it should be
# installed where only root can modify/replace it. This means also none of the
# parent directories should be writable by the user.
#
# The varRun directory also must not be writable by the user because it will
# include the socket_vmnet pid file. Those will be terminated via sudo, so replacing
# the pid file would allow killing of arbitrary privileged processes. varRun
# however MUST be writable by the daemon user.
#
# None of the paths segments may be symlinks, why it has to be /private/var
# instead of /var etc.
paths:
# socketVMNet requires Lima >= 0.12 .
# socketVMNet has precedence over vdeVMNet.
  socketVMNet: "/opt/socket_vmnet/bin/socket_vmnet"
# vdeSwitch and vdeVMNet are DEPRECATED.
  vdeSwitch: /opt/vde/bin/vde_switch
  vdeVMNet: /opt/vde/bin/vde_vmnet
  varRun: /private/var/run/lima
  sudoers: /private/etc/sudoers.d/lima

group: everyone

networks:
  user-v2:
    mode: user-v2
    gateway: 192.168.104.1
    netmask: 255.255.255.0
    # user-v2 network is experimental network mode which supports all functionalities of default usernet network and also allows vm -> vm communication.
    # Doesn't support configuration of custom gateway; hardcoded to 192.168.5.0/24
  shared:
    mode: shared
    gateway: 192.168.105.1
    dhcpEnd: 192.168.105.254
    netmask: 255.255.255.0
  bridged:
    mode: bridged
    interface: en0
    # bridged mode doesn't have a gateway; dhcp is managed by outside network
  host:
    mode: host
    gateway: 192.168.106.1
    dhcpEnd: 192.168.106.254
    netmask: 255.255.255.0
```

AkihiroSuda:
Non-nil `cache.err` was overwritten by nil

Fix #2313

### DIFF
--- a/pkg/networks/config.go
+++ b/pkg/networks/config.go
@@ -135,6 +135,7 @@ func loadCache() {
 		cache.err = yaml.UnmarshalWithOptions(b, &cache.config, yaml.Strict())
 		if cache.err != nil {
 			cache.err = fmt.Errorf("cannot parse %q: %w", configFile, cache.err)
+			return
 		}
 		cache.config, cache.err = fillDefaults(cache.config)
 		if cache.err != nil {
--- a/pkg/networks/config.go
+++ b/pkg/networks/config.go
@@ -132,10 +132,16 @@ func loadCache() {
 		if cache.err != nil {
 			return
 		}
-		cache.err = yaml.UnmarshalWithOptions(b, &cache.config, yaml.Strict())
+		cache.err = yaml.UnmarshalWithOptions(b, &cache.config, yaml.DisallowDuplicateKey())
 		if cache.err != nil {
 			cache.err = fmt.Errorf("cannot parse %q: %w", configFile, cache.err)
 			return
+		}
+		var strictConfig YAML
+		if strictErr := yaml.UnmarshalWithOptions(b, &strictConfig, yaml.Strict()); strictErr != nil {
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM.

I would have kept the 2 unused fields around indefinitely and kept the strict mode. It is mostly meant to catch typos, imo, so just not warning on these 2 old fields would have been fine. With this change you will still have to eventually break existing setups and force users to update their `network.yaml` file just for cleanup reasons. But I guess it doesn't really matter.

PR #2329, opened 2024-05-07
Issue an error message when trying to run on NFS

### LINKED ISSUE #2325: lima fails in a confusing way when the home directory is on NFS
### Description

I stumbled across the fact that a home directory on NFS does not work with lima, because qemu can't handle it.  I discussed it with @afbjorklund and he requested that I make an issue so something can be done to make it more clear why a failure is happening. 

The error I got looked like this:
```
$ limactl start
? Creating an instance "default" Choose another template (docker, podman, archlinux, fedora, ...)
? Choose a template ubuntu-lts
INFO[0014] Starting the instance "default" with VM driver "qemu"
INFO[0014] Attempting to download the image              arch=x86_64 digest="sha256:42dcf9757e75c3275486b397a752fb535c7cd8e5232ee5ee349554b7a55f1702" location="https://cloud-images.ubuntu.com/releases/22.04/release-20240308/ubuntu-22.04-server-cloudimg-amd64.img"
Downloading the image (ubuntu-22.04-server-cloudimg-amd64.img)
618.25 MiB / 618.25 MiB [----------------------------------] 100.00% 31.03 MiB/s
INFO[0038] Downloaded the image from "https://cloud-images.ubuntu.com/releases/22.04/release-20240308/ubuntu-22.04-server-cloudimg-amd64.img"
FATA[0038] failed to run [qemu-img create -f qcow2 -F qcow2 -b /nashome/d/dwd/.lima/default/basedisk /nashome/d/dwd/.lima/default/diffdisk 107374182400]: "Formatting '/nashome/d/dwd/.lima/default/diffdisk', fmt=qcow2 cluster_size=65536 extended_l2=off compression_type=zlib size=107374182400 backing_file=/nashome/d/dwd/.lima/default/basedisk backing_fmt=qcow2 lazy_refcounts=off refcount_bits=16\nqemu-img: /nashome/d/dwd/.lima/default/diffdisk: Image is not in qcow2 format\n": exit status 1
```

The workaround is to set LIMACTL_HOME to an empty directory on a local disk.  If you point it to a directory that has other subdirectories in it, lima unexpectedly dies looking for lima.yaml files in them.   I expected a variable name with `HOME` in it to look like $HOME where a .lima directory would be created; instead it behaves more like a `CONFIG` directory, so that name was also misleading.

afbjorklund:
Display an error message, when trying to put $LIMA_HOME on NFS network filesystem (instead of local filesystem)

Closes #2325

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -8,6 +8,7 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/lima-vm/lima/pkg/fsutil"
 	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/store/dirnames"
 	"github.com/lima-vm/lima/pkg/version"
@@ -92,8 +93,22 @@ func newApp() *cobra.Command {
 		}
 		// Make sure either $HOME or $LIMA_HOME is defined, so we don't need
 		// to check for errors later
-		if _, err := dirnames.LimaDir(); err != nil {
+		dir, err := dirnames.LimaDir()
+		if err != nil {
 			return err
+		}
+		// Make sure that directory is on a local filesystem, not on NFS
+		// if the directory does not yet exist, check the home directory
+		_, err = os.Stat(dir)
+		if errors.Is(err, os.ErrNotExist) {
+			dir = filepath.Dir(dir)
+		}
+		nfs, err := fsutil.IsNFS(dir)
+		if err != nil {
+			return err
+		}
+		if nfs {
+			return errors.New("must not run on NFS dir")
 		}
 		return nil
 	}
--- a/pkg/fsutil/fsutil_linux.go
+++ b/pkg/fsutil/fsutil_linux.go
@@ -1,0 +1,15 @@
+//go:build linux
+
+package fsutil
+
+import (
+	"golang.org/x/sys/unix"
+)
+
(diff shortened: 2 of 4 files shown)


cmd/limactl/main.go
```diff
@@ -92,9 +93,23 @@ func newApp() *cobra.Command {
...
+		_, err = os.Stat(dir)
+		if errors.Is(err, os.ErrNotExist) {
+			dir = filepath.Dir(dir)
+		}
+		nfs, err := fsutil.IsNFS(dir)
+		if err != nil {
+			return err
+		}
+		if nfs {
+			return errors.New("must not run on NFS dir")
```
AkihiroSuda, line 111: Probably this should be just a warning?

It still may work on NFS with a future potential version of QEMU, vz, or something else?

afbjorklund: Possibly, but I'm not sure if it is something that can be supported?

We can download/copy the "basedisk" from a remote location, such as ~/.cache, but keeping the "diffdisk" on a network file share could be a potential source of bugs and performance issues. Even if the bug here, regarding file locking with `qemu-img`, can be found and fixed? Something like iSCSI could work, but skeptical about NFS.

afbjorklund:
Apparently NFS_SUPER_MAGIC is only defined on Linux (not Unix).

AkihiroSuda (approved):
Thanks

PR #2333, opened 2024-05-10
Change cpuType default to null

afbjorklund:
Avoid issues with converting null strings in a map value

Change order to alphabetical, to match the "limactl info"

----

Helps with:

* https://github.com/lima-vm/lima/pull/1069

* https://github.com/lima-vm/lima/pull/2306

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -252,15 +252,12 @@ containerd:
 # Specify desired QEMU CPU type for each arch.
 # You can see what options are available for host emulation with: `qemu-system-$(arch) -cpu help`.
 # Setting of instructions is supported like this: "qemu64,+ssse3".
+# 🟢 Builtin default: hard-coded arch map with type (see the output of `limactl info | jq .defaultTemplate.cpuType`)
 cpuType:
-  # 🟢 Builtin default: "cortex-a72" (or "host" when running on aarch64 host)
-  aarch64: null
-  # 🟢 Builtin default: "cortex-a7" (or "host" when running on armv7l host)
-  armv7l: null
-  # 🟢 Builtin default: "qemu64" (or "host,-pdpe1gb" when running on x86_64 host)
-  x86_64: null
-  # 🟢 Builtin default: "rv64" (or "host" when running on riscv64 host)
-  riscv64: null
+  # aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
+  # armv7l: "cortex-a7" # (or "host" when running on armv7l host)
+  # riscv64: "rv64" # (or "host" when running on riscv64 host)
+  # x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
 
 rosetta:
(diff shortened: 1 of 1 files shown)


afbjorklund:
Note: the `-pdpe1gb`  is only true when running on a Darwin host, not on Linux (and looks a bit odd).

But I figured it is best to have details in the code, and avoid duplicating them here... Maybe remove it?

examples/default.yaml
```diff
@@ -252,15 +252,12 @@ containerd:
...
+# 🟢 Builtin default: hard-coded arch map with type (see the output of `limactl info | jq .defaultTemplate.cpuType`)
 cpuType:
-  # 🟢 Builtin default: "cortex-a72" (or "host" when running on aarch64 host)
-  aarch64: null
-  # 🟢 Builtin default: "cortex-a7" (or "host" when running on armv7l host)
-  armv7l: null
-  # 🟢 Builtin default: "qemu64" (or "host,-pdpe1gb" when running on x86_64 host)
-  x86_64: null
-  # 🟢 Builtin default: "rv64" (or "host" when running on riscv64 host)
-  riscv64: null
```
AkihiroSuda, line 263: Why remove these lines?

afbjorklund: There is a problem with the json, when it is trying to marshal "null" into a string value...

After changing CPUType to `map[Arch]string`, it is a problem to express as jsonschema.

Also, it means that we don't have to carry the list of architectures in the "empty" yaml?

Made some minor similar changes to ssh.localPort and Rosetta, but will post separately

afbjorklund:
```diff
 images: []
-cpuType:
-  aarch64: ""
-  armv7l: ""
-  riscv64: ""
-  x86_64: ""
-ssh:
-  localPort: 0
```

```json
{
   "audio" : {},
   "caCerts" : {},
   "containerd" : {},
   "cpuType" : {
      "aarch64" : "",
      "armv7l" : "",
      "riscv64" : "",
      "x86_64" : ""
   },
   "firmware" : {},
   "hostResolver" : {},
   "images" : null,
   "rosetta" : {
      "binfmt" : null,
      "enabled" : null
   },
   "ssh" : {
      "localPort" : 0
   },
   "video" : {
      "vnc" : {}
   }
}
```

afbjorklund: Also the (text of the) lines are **not** removed, they are just translated into comments instead.

When uncommented, the default is reconstructed (could use `""` or `null` for empty values)

```yaml
 cpuType:
  aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
  armv7l: "cortex-a7" # (or "host" when running on armv7l host)
  riscv64: "rv64" # (or "host" when running on riscv64 host)
  x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
```

AkihiroSuda (approved):
Thanks

Issue #1347, opened 2023-02-02
Automation for updating templates


AkihiroSuda:
It is really hard for me to create a PR like https://github.com/lima-vm/lima/pull/1236 to update the template image digests.

We have to have a tool for updating these templates automatically.
The tool must retain comment lines and indentation styles in the YAMLs.

afbjorklund:
Maybe also some means of sharing them, with some kind of FROM system ?

* https://github.com/lima-vm/lima/issues/824

AkihiroSuda:
> Maybe also some means of sharing them, with some kind of FROM system ?
> 
> * [Inherit system images and share them between examples #824](https://github.com/lima-vm/lima/issues/824)

Yes, but that is a separate issue

jlm0x017:
Do you already have the link to the updated image?  Or is that step needed as well?
I expect you grab the checksum provided at the source, rather than generating it yourself, but please do confirm.

AkihiroSuda:
> Do you already have the link to the updated image? 

No, e.g., we have to detect the latest version `20230124-1270` from https://cloud.debian.org/images/cloud/bullseye/  , but I'm not sure what is the robust way to do this.

`w3m | grep` might be enough, but seriously we should also consider adopting some machine learning stuff.

afbjorklund:
If I remember correctly, there was some standard metadata for scraping upstream for new releases to package...

Maybe something like that can be used here, to "describe" the various vendors and where they put their binaries ?

afbjorklund:
Example output:

https://qa.debian.org/cgi-bin/watch?pkg=containerd

Unfortunately, the others are **not** available as packages. But maybe something similar to this:

https://repology.org/project/nerdctl/versions

AkihiroSuda:
I'm now planning to use yq

balajiv113:
How about using libosinfo (`osinfo-db os`) ??

I could see their db has info of different os variants
https://gitlab.com/libosinfo/osinfo-db/-/tree/main/data/os

We might need to write a python wrapper on top of this library (the tool osinfo-db is not giving out info on image download URL's by architecture that is present in the xml file)

afbjorklund:
`yq` can read xml too

```
yq -p xml -P
```

afbjorklund:
> The library and database are licensed under the terms of the GNU LGPL version 2 or later.

AkihiroSuda:
> osinfo-db

[...]

PR #2336, opened 2024-05-11
Update almalinux 9 sample to 9.4

fwilhe2:
Updated images are available

### DIFF
--- a/examples/almalinux-9.yaml
+++ b/examples/almalinux-9.yaml
@@ -1,12 +1,12 @@
 # This template requires Lima v0.11.1 or later.
 
 images:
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Thanks, would you want to update rocky and oraclelinux too?

AkihiroSuda:
ping @fwilhe2

afbjorklund:
> Out of curiously: Is there an api to automate this process?

Not yet.

* https://github.com/lima-vm/lima/issues/1347

It is not super important, since once you update the image the result will be the same (and it can be done at boot).
But it saves bandwidth, not having to download an outdated image with outdated packages only to update it...

AkihiroSuda (approved):
Merging, other images can be updated by somebody else who is interested in

PR #2337, opened 2024-05-12
Convert limayaml functions to use pointer

afbjorklund:
Add a simple unit test for some coverage

### DIFF
--- a/cmd/limactl/edit.go
+++ b/cmd/limactl/edit.go
@@ -94,7 +94,7 @@ func editAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


afbjorklund:
Interesting failure, on Windows:

```
    validate_test.go:25: assertion failed: error is not nil: field `mounts[1].location` must be an absolute path, got "/tmp/lima"
```

Apparently using the "wrong" path.

i.e. it probably needs some TMPDIR

AkihiroSuda:
Could you explain the purpose of this?

AkihiroSuda (approved):
I'm merging this, but still want to know the background

PR #2338, opened 2024-05-12
Add save functionality to limayaml

afbjorklund:
Make it possible to fill in the defaults (null) with limactl validate, with a workaround for python yaml.

Was using a local patched version to test with the jsonschema and such, might as well be a flag ?

You can validate more than one yaml, then the output will be formatted with document separators.

`limactl validate examples/default.yaml examples/ubuntu.yaml`

---

Found an interesting python bug (libyaml difference?), when using `check-jsonschema`:

```
Schema validation errors were encountered.
  default.yaml::$.mounts[0].location: None is not of type 'string'
```
 
```yaml
mounts:
- location: ~
  mountPoint: ~
  writable: false
  sshfs:
    cache: true
    followSymlinks: false
    sftpDriver: ""
  9p:
    securityModel: none
    protocolVersion: 9p2000.L
    msize: 128KiB
    cache: fscache
- location: /tmp/lima
  mountPoint: /tmp/lima
  writable: true
  sshfs:
    cache: true
    followSymlinks: false
    sftpDriver: ""
  9p:
    securityModel: none
    protocolVersion: 9p2000.L
    msize: 128KiB
    cache: mmap
```

Added a workaround, not sure which language is "right" according to yaml spec ?

### DIFF
--- a/cmd/limactl/validate.go
+++ b/cmd/limactl/validate.go
@@ -18,12 +18,18 @@ func newValidateCommand() *cobra.Command {
 		RunE:    validateAction,
 		GroupID: advancedCommand,
 	}
+	validateCommand.Flags().Bool("fill", false, "fill defaults")
 	return validateCommand
 }
 
-func validateAction(_ *cobra.Command, args []string) error {
+func validateAction(cmd *cobra.Command, args []string) error {
+	fill, err := cmd.Flags().GetBool("fill")
+	if err != nil {
+		return err
+	}
+
 	for _, f := range args {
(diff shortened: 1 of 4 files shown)


afbjorklund:
According to yaml.org, `~` is the representation of the null value. It is not a valid string.

afbjorklund:
There are some minor differences between this output and the `limactl config` output (as JSON)

Due to the difference in sequence indentation, and the missing default for Virtiofs on Linux/9p.

AkihiroSuda (approved):
Thanks

PR #2342, opened 2024-05-12
Add firmware search paths for fedora rpms

afbjorklund:
Including systems based on Fedora, such as CentOS and AlmaLinux.

~~There doesn't seem to be any arm32 firmware available in Fedora.~~

The standard path (OVMF) was never backported to Legacy Fedora (EL):

https://src.fedoraproject.org/rpms/edk2/c/e1a8a9caa18b10ff4ea5ee5d131c734e61b379de?branch=rawhide

`yum provides "*/OVMF_CODE.fd"`

`yum provides "*/AAVMF_CODE.fd"`

* https://github.com/lima-vm/lima/discussions/1917

`export QEMU_SYSTEM_X86_64=/usr/libexec/qemu-kvm`

`export QEMU_SYSTEM_AARCH64=/usr/libexec/qemu-kvm`

---

The 32-bit firmware was there as well, but it was hidden in the RHEL builds

	
```specfile
%if %{defined rhel}
%define build_ovmf 0
%define build_aarch64 0
%ifarch x86_64
  %define build_ovmf 1
%endif
%ifarch aarch64
  %define build_aarch64 1
%endif
%define build_riscv64 0
%define build_loongarch64 0
%else
%define build_ovmf 1
%define build_aarch64 1
%define build_riscv64 1
%define build_loongarch64 1
%endif
```

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -1100,17 +1100,21 @@ func getFirmware(qemuExe string, arch limayaml.Arch) (string, error) {
 		// Debian package "ovmf"
 		candidates = append(candidates, "/usr/share/OVMF/OVMF_CODE.fd")
 		candidates = append(candidates, "/usr/share/OVMF/OVMF_CODE_4M.fd")
+		// Fedora package "edk2-ovmf"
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #2345, opened 2024-05-13
Add possibility to override /tmp

### LINKED ISSUE #2339: Add abstraction for the default `/tmp/lima` directory ?
### Description

It doesn't work so great in a multi-user environment, and doesn't work on Windows (compared with e.g. `C:\TEMP`)

i.e. when several users are sharing a server, they will see the files from the other users in the shared `/tmp/lima`

Hosts based on Fedora put `/tmp` [in RAM](https://fedoraproject.org/wiki/Features/tmp-on-tmpfs), so you need to use `/var/tmp` if you want to use a normal disk directory.

Should we have some placeholder value for the default.yaml, instead of the "/tmp/lima"? Similar to LIMA_HOME.

* https://github.com/lima-vm/lima/pull/2337#issuecomment-2106195608

afbjorklund:
Use `$TMPDIR` to change the location of `/tmp` on Unix.

Use `$TEMP` to change the location of `/tmp` on Windows.

https://pkg.go.dev/os#TempDir

Closes #2339

### DIFF
--- a/pkg/infoutil/infoutil.go
+++ b/pkg/infoutil/infoutil.go
@@ -13,6 +13,7 @@ type Info struct {
 	Templates       []templatestore.Template `json:"templates"`
 	DefaultTemplate *limayaml.LimaYAML       `json:"defaultTemplate"`
 	LimaHome        string                   `json:"limaHome"`
+	LimaTemp        string                   `json:"limaTemp"`
 	VMTypes         []string                 `json:"vmTypes"` // since Lima v0.14.2
 }
 
@@ -38,5 +39,6 @@ func GetInfo() (*Info, error) {
 	if err != nil {
 		return nil, err
 	}
+	info.LimaTemp = dirnames.LimaTmp()
 	return info, nil
 }
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -17,6 +17,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/cpu"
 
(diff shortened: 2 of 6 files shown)


afbjorklund:
On second thought, having `/var/tmp/lima` mount under `/tmp/lima` breaks the abstraction.

We probably still need to do something different on Windows, and not use **local**pathutil?

afbjorklund:
Reference to the `cygpath` utility, regarding filepath translation...

https://www.msys2.org/docs/filesystem-paths/ (skipped the /mnt)

jandubois:
> No idea where `C:\Users\UserName\AppData\Local\Temp\lima` is supposed to end up.
> 
> Possibly something like `/c/Users/UserName/AppData/Local/Temp/lima`, maybe `/mnt`

I have not really used Lima on Windows yet, but aren't we using WSL2 for the emulation? In that case we should be using WSL2 path names and not msys hacks.

Note that with WSL2 (depending on how it is configured) none of our mounting really matters as everything is already mounted anyways:

```console
suse@rd-w11-test-1:~$ wslpath -u c:\\user
/mnt/c/user
suse@rd-w11-test-1:~$ wslpath -w /tmp
\\wsl.localhost\Ubuntu-22.04\tmp
```

afbjorklund:
This was having the eye towards QEMU (or Hyper-V perhaps), but mostly it was just a unit test failure.

afbjorklund:
Then again, this feature is really on needed on Fedora or on Windows (i.e. without WSL)

It is not needed on Ubuntu or on macOS, so I think I will just leave it in "draft" for a while

Issue #2346, opened 2024-05-14
apk installs to tmpfs


moenodedev:
### Description

`apk` installs into `/usr` which is a tmpfs mount. If the disk runs out of space while installing packages ssh won't work after rebooting. Should `/usr` be in `/dev/disk/by-label/data-volume`?

```
$ df -h /usr
Filesystem                Size      Used Available Use% Mounted on
tmpfs                   955.7M     70.0M    885.7M   7% /
```

limactl version 0.22.0
macOS Sonoma 14.5
alpine-lima-std-3.19.0-aarch64.iso
mountType: "virtiofs"

jandubois:
I think if you want to move `/usr` to persistent storage, then you would be better off using a regular disk image instead of booting from an ISO.

It looks like Alpine is publishing cloudinit images now, so I would suggest trying to build a template using https://dl-cdn.alpinelinux.org/alpine/v3.19/releases/cloud/nocloud_alpine-3.19.1-x86_64-bios-cloudinit-r0.qcow2

We probably need to adjust some code in Lima that assumes Alpine always means ISO, but that shouldn't be a problem, once it is confirmed that the Alpine cloud images can work.

jandubois:
Or https://dl-cdn.alpinelinux.org/alpine/v3.19/releases/cloud/nocloud_alpine-3.19.1-aarch64-uefi-cloudinit-r0.qcow2 for `aarch64`.

afbjorklund:
Maybe apk should be disabled (or issue warnings?) in alpine-lima, and refer to making custom "editions" instead

jandubois:
> Maybe apk should be disabled (or issue warnings?) in [alpine-lima](https://github.com/lima-vm/alpine-lima),

Why? You can use the `apk add` in a provisioning script. The packages get cached, and will be reinstalled from the cache on reboot. From the user's point of view this should be indistinguishable from having `/usr` persisted as well, except for a tiny delay for reinstalling them.

Of course if you do massive installs and run out of RAM, then this whole mechanism of running of a LiveCD does not really work anymore.

afbjorklund:
OK, so then running out of RAM in alpine-lima should be equivalent to running out of /usr disk in alpine

afbjorklund:
> The packages get cached, and will be reinstalled from the cache on reboot

I added this feature to "tinycloud" as well, so it will use /mnt/data/tce for .tcz

[...]

PR #2360, opened 2024-05-25
Add alpine-image template

jandubois:
It uses the official Alpine cloud images instead of the alpine-lima ISO.

Ideally this template should be called `alpine` and the existing one `alpine-iso`, but that is going to break backwards compatibility.

@lima-vm/maintainers Should we leave things as-is, or should I make a copy of `alpine` to `alpine-iso`, and then add a `message` to the `alpine` template that tells users to switch to using the `alpine-iso` name because at some time in the future `alpine` may become a symlink to `alpine-image`?

I've only done cursory testing on this template, but it starts up fine on x86_64 and aarch64.

The PR should be reviewed while ignoring white-space changes; the first commit is just reducing indentation levels in `04-persistent-data-volume.sh`.

See also #2346 and #2359

### DIFF
--- a/examples/alpine-image.yaml
+++ b/examples/alpine-image.yaml
@@ -1,0 +1,20 @@
+images:
+- location: "https://dl-cdn.alpinelinux.org/alpine/v3.19/releases/cloud/nocloud_alpine-3.19.1-x86_64-bios-cloudinit-r0.qcow2"
+  arch: "x86_64"
+  digest: "sha512:f122b0e9e832cf90bda49ca73b800105ab4aa193bd340cba641e6acbda1da9aa5571de87870561380ac69bdb2dcd428590e7b4a898f7f62b559f5c12a4aefc01"
+- location: "https://dl-cdn.alpinelinux.org/alpine/v3.19/releases/cloud/nocloud_alpine-3.19.1-aarch64-uefi-cloudinit-r0.qcow2"
+  arch: "aarch64"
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
We can just have a breaking change when we release v1.0 with VZ as default

AkihiroSuda (approved):

examples/alpine-image.yaml
```diff
@@ -0,0 +1,20 @@
+images:
+- location: "https://dl-cdn.alpinelinux.org/alpine/v3.19/releases/cloud/nocloud_alpine-3.19.1-x86_64-bios-cloudinit-r0.qcow2"
```
afbjorklund, line 2: Can we change from "bios" to "uefi" here ?

examples/alpine-image.yaml
```diff
@@ -0,0 +1,20 @@
...
+  arch: "x86_64"
+  digest: "sha512:f122b0e9e832cf90bda49ca73b800105ab4aa193bd340cba641e6acbda1da9aa5571de87870561380ac69bdb2dcd428590e7b4a898f7f62b559f5c12a4aefc01"
+- location: "https://dl-cdn.alpinelinux.org/alpine/v3.19/releases/cloud/nocloud_alpine-3.19.1-aarch64-uefi-cloudinit-r0.qcow2"
+  arch: "aarch64"
+  digest: "sha512:d5e69cff0ecb0fd3850bd78f56f66131115934df27b2373c35a85a74b9def52822134dd43f90b4fabe239fdd4452026cb45f1c8f0b36a43af69335a26f0959b5"
+
+firmware:
+  legacyBIOS: true
```
afbjorklund, line 10: Then we don't need legacyBIOS, with UEFI.

PR #2364, opened 2024-05-25
Remove the embedded default user in alpine cloud

### LINKED ISSUE #2362: User creation errors, with the new alpine-image template
### Description

Getting errors from cloud-config, when trying to create the initial user:

```
2024-05-25 08:10:21,390 - schema.py[WARNING]: Invalid cloud-config provided: Please run 'sudo cloud-init schema --system' to see the schema errors.
2024-05-25 08:10:23,288 - util.py[WARNING]: Running module ca_certs (<module 'cloudinit.config.cc_ca_certs' from '/usr/lib/python3.11/site-packages/cloudinit/config/cc_ca_certs.py'>) failed
2024-05-25 08:10:23,300 - util.py[WARNING]: Failed to create user anders
2024-05-25 08:10:23,302 - util.py[WARNING]: Running module users_groups (<module 'cloudinit.config.cc_users_groups' from '/usr/lib/python3.11/site-packages/cloudinit/config/cc_users_groups.py'>) failed
```

Will try to access the image (no login through serial), to find the log file...

```
Welcome to Alpine Linux 3.19
Kernel 6.6.14-0-virt on an x86_64 (/dev/ttyS0)

lima-alpine-image login: alpine
alpine
Password: 
alpine
Login incorrect
```

----

* #2360

https://www.alpinelinux.org/cloud/

> NoCloud, ... images are considered beta quality

afbjorklund:
There should not be a non-system user in the image.

Let cloud-init handle the users and groups instead.

Closes #2362

Fixes the boot on Linux, and then you can add nerdctl.

----

`sudo apk add nerdctl containerd buildctl buildkit`

```
lima-alpine-image:/home/anders$ sudo nerdctl version
Client:
 Version:	1.7.6
 OS/Arch:	linux/amd64
 Git commit:	<unknown>
 buildctl:
  Version:	0.12.5
  GitCommit:	alpine

Server:
 containerd:
  Version:	v1.7.10
  GitCommit:	4e1fe7492b9df85914c389d1f15a3ceedbb280ac
 runc:
  Version:	1.1.12
  GitCommit:	51d5e94601ceffbbd85688df1c928ecccbfa4685
```

```
lima-alpine-image:/home/anders$ df -h /dev/vda1 /dev/vda2
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda1       486K  282K  204K  59% /boot/efi
/dev/vda2        94G  453M   90G   1% /
```

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/00-alpine-user-group.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/00-alpine-user-group.sh
@@ -1,0 +1,12 @@
+#!/bin/sh
+# Remove the user embedded in the image,
+# and use cloud-init for users and groups.
+test -f /etc/alpine-release || exit 0
+test "$LIMA_CIDATA_USER" != "alpine" || exit 0
+
+if [ "$(id -u alpine 2>&1)" = "1000" ]; then
+	userdel alpine
+	rmdir /home/alpine
+	cloud-init clean --logs
+	reboot
+fi


AkihiroSuda:
Can we test this in CI?

afbjorklund:
> Can we test this in CI?

Testing on Linux should be enough, I think that Mac "got away" with it because it uses UID 500 ?

pkg/cidata/cidata.TEMPLATE.d/boot/00-alpine-user-group.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+# Remove the user embedded in the image,
+# and use cloud-init for users and groups.
+test -f /etc/alpine-release || exit 0
+
```
jandubois, line 5: Although unlikely, should this have a safeguard for the case where the host user is also called `alpine`? Otherwise we would delete the actual user again on each reboot:

```bash
test [ "$LIMA_CIDATA_USER" != "alpine" ] || exit 0
```

jandubois (changes requested):
I'm fine with deleting the `alpine` user; it only exists so you can test the image without setting up a cloud-init profile.

But I think we should make an exception if the local user is also called `alpine`. Otherwise we would have to use the `lima` user name inside the VM.

Note that I haven't tested if using `alpine` as a host username would work otherwise; maybe it would still be broken for other reasons.

pkg/cidata/cidata.TEMPLATE.d/boot/00-alpine-user-group.sh
afbjorklund: Right, if your parents called you "alpine" as a kid you would end up with a boot loop

afbjorklund:
> I'm fine with deleting the `alpine` user; it only exists so you can test the image without setting up a cloud-init profile.

I think that cloud-init should handle this anyway, by using the "default_user" and the distro default /etc/could config?

jandubois (approved):
Thanks, LGTM

Issue #1617, opened 2023-06-08
Change the default value of `ssh.loadDotSSHPubKeys` from `true` to `false`


AkihiroSuda:
https://github.com/lima-vm/lima/blob/9d29da396c164fea261bc547b4537d470aa3a824/examples/default.yaml#L111-L122

This might be a breaking change for new instances, but does not affect existing instances.

PR #2365, opened 2024-05-26
Remove user@host from the lima ssh key

afbjorklund:
Don't expose the host information unless requested, i.e. when using loadDotSSHPubKeys=true for instance.

* https://github.com/lima-vm/lima/discussions/2256

* #1617


jandubois (approved):
Thanks, LGTM

PR #2367, opened 2024-05-27
PoC: Implement optional host proxy for lima cache

afbjorklund:
Implements both `http_proxy` and `https_proxy` (with mitm cert)

~~Does not actually cache anything (yet), just verbose INFO logging.~~

The use case is to use the regular $LIMA_CACHE also for apt/ctr.

But instead of doing custom mirrors, use a brute force http proxy.

```yaml
# The host proxy implements a HTTP and HTTPS proxy that can cache downloads on the host.
hostProxy:
  # 🟢 Builtin default: false
  enabled: null
```

TODO: Add some more options, for configuring apt and containerd

The target is to be able to cache the packages and images for k8s

----

Tested with curl, and with ctr by using a custom certs.d/docker.io

```toml
server = "https://docker.io"

[host."https://registry-1.docker.io"]
  capabilities = ["pull", "resolve"]
  ca = "proxy.crt"
```
And similar also for https://registry.k8s.io (and https://pkgs.k8s.io)

https://github.com/containerd/nerdctl/blob/main/docs/registry.md

/etc/containerd

`/etc/containerd/certs.d/_default/hosts.toml`

/etc/apt/apt.conf.d/50proxy.conf

`Acquire::https::CaInfo "/path/to/proxy.crt";`

---

* https://github.com/lima-vm/lima/issues/1422#issuecomment-2133180705
* https://github.com/lima-vm/lima/issues/397

PR #2368, opened 2024-05-27
Store time of last modified in cache dir

afbjorklund:
Was looking more at caching, and thought it would be a good idea to store the last modified time too.

```console
$ more ~/.cache/lima/download/by-url-sha256/2385cd7cc68589b1dab56df12415b9151d977ba9b9b668e913f02ebd35804ef7/url 
https://github.com/lima-vm/alpine-lima/releases/download/v0.2.38/alpine-lima-std-3.19.0-x86_64.iso
$ more ~/.cache/lima/download/by-url-sha256/2385cd7cc68589b1dab56df12415b9151d977ba9b9b668e913f02ebd35804ef7/time 
Fri, 26 Apr 2024 05:04:22 GMT
$ more ~/.cache/lima/download/by-url-sha256/2385cd7cc68589b1dab56df12415b9151d977ba9b9b668e913f02ebd35804ef7/type 
application/octet-stream
```

Then you can do things such as If-Modified-Since and helps when looking at implementing a simple proxy.

* #2367

When caching more things than OS images, it is also useful to record the content type (deb, tar, whatever)

`WARN[0000] reference for unknown type: application/octet-stream`

----

The information is available in the `Result`, when retreiving cached files (but the string can be empty)

~~It would be possible to add the _content_ instead, and use empty string for older cache without the files.~~

```diff
 type Result struct {
        Status          Status
        CachePath       string // "/Users/foo/Library/Caches/lima/download/by-url-sha256/<SHA256_OF_URL>/data"
+       LastModified    time.Time
+       ContentType     string
        ValidatedDigest bool
 }

```

### DIFF
--- a/pkg/downloader/downloader.go
+++ b/pkg/downloader/downloader.go
@@ -45,6 +45,7 @@ type Result struct {
 	Status          Status
 	CachePath       string // "/Users/foo/Library/Caches/lima/download/by-url-sha256/<SHA256_OF_URL>/data"
 	LastModified    string
+	ContentType     string
 	ValidatedDigest bool
 }
 
@@ -176,7 +177,7 @@ func Download(ctx context.Context, local, remote string, opts ...Opt) (*Result, 
 	}
 
 	if o.cacheDir == "" {
-		if err := downloadHTTP(ctx, localPath, "", remote, o.description, o.expectedDigest); err != nil {
+		if err := downloadHTTP(ctx, localPath, "", "", remote, o.description, o.expectedDigest); err != nil {
 			return nil, err
 		}
 		res := &Result{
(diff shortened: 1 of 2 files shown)


pkg/downloader/downloader.go
```diff
@@ -44,6 +44,8 @@ const (
 type Result struct {
 	Status          Status
 	CachePath       string // "/Users/foo/Library/Caches/lima/download/by-url-sha256/<SHA256_OF_URL>/data"
+	LastModified    string
```
AkihiroSuda, line 47: Shouldn't this be `time.Time` ?

pkg/downloader/downloader.go
```diff
@@ -118,6 +120,17 @@ func WithExpectedDigest(expectedDigest digest.Digest) Opt {
 	}
 }
 
+func readFile(path string) string {
```
AkihiroSuda, line 123: Why not just use `os.ReadFile`

afbjorklund: Sloppy error handling, only

afbjorklund:
Now the user (proxy) will need to use `time.IsZero` and `time.Format(time.RFC1123)`, for LastModified

afbjorklund:
Added some unit tests.

AkihiroSuda (approved):
Thanks

PR #2376, opened 2024-05-29
Never retain ownership during installation

danchr:
By default, tar will retain ownership when run as root. Even if built as a user, the install phase should not install binaries owned by that user. To suppress this, we pass `--no-same-owner` to tar.

While at it, I made it use the `TAR` environment variable during installation as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -209,7 +209,7 @@ docs/lima-sequence-diagram.png: docs/images/lima-sequence-diagram.puml
 install: uninstall
 	mkdir -p "$(DEST)"
 	# Use tar rather than cp, for better symlink handling
-	( cd _output && tar c * | tar Cxv "$(DEST)" )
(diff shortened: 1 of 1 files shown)


afbjorklund:
Another approach is to change the owner when creating the tarball - and archives?

That would mean that the user would be "root/root", instead of "runner/docker"

Like in nerdctl: (it also removes the leading `./` in all the entries or the tarball)

```Makefile
TAR_OWNER0_FLAGS=--owner=0 --group=0
TAR_FLATTEN_FLAGS=--transform 's/.*\///g'
```

But for the install, --no-same-owner works.

afbjorklund (approved):

AkihiroSuda (approved):
Thanks

AkihiroSuda:
@afbjorklund You may merge PRs 👍

AkihiroSuda:
https://github.com/apps/dco

danchr:
> Thanks, but please sign the commit for DCO https://github.com/apps/dco

Done!

Issue #2117, opened 2024-01-08
[User name: `admin`] Waiting for the essential requirement 1 of 4: "ssh"


fortunateLeopard:
### Description

I can't get seem to connect via ssh to lima with default using either drivers, output for `vz` below. My local username is `admin` in case that's relevant.

```bash
limactl --version    
limactl version 0.19.1

% sw_vers
ProductName:		macOS
ProductVersion:		14.2.1
BuildVersion:		23C71
```
```
limactl start             
INFO[0000] Using the existing instance "default"        
INFO[0000] Starting the instance "default" with VM driver "vz" 
INFO[0000] [hostagent] hostagent socket created at /Users/admin/.lima/default/ha.sock 
INFO[0000] [hostagent] Starting VZ (hint: to watch the boot progress, see "/Users/admin/.lima/default/serial*.log") 
INFO[0000] [hostagent] new connection from  to          
INFO[0001] SSH Local Port: 60022                        
INFO[0001] [hostagent] Waiting for the essential requirement 1 of 2: "ssh" 
INFO[0001] [hostagent] [VZ] - vm state change: running  
INFO[0011] [hostagent] Waiting for the essential requirement 1 of 2: "ssh" 
INFO[0021] [hostagent] Waiting for the essential requirement 1 of 2: "ssh" 
...
```

```bash
{"level":"info","msg":"Waiting for the essential requirement 1 of 2: \"ssh\"","time":"2024-01-08T06:40:33-05:00"}
{"level":"debug","msg":"executing script \"ssh\"","time":"2024-01-08T06:40:33-05:00"}
[...]

PR #2378, opened 2024-05-30
issue 2117: disallow admin username

bo17age:
disallow admin username; use fallback username.
Related issue: https://github.com/lima-vm/lima/issues/2117

### DIFF
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -48,10 +48,7 @@ func copyAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	u, err := osutil.LimaUser(false)
-	if err != nil {
-		return err
-	}
+
 	instDirs := make(map[string]string)
 	scpFlags := []string{}
 	scpArgs := []string{}
@@ -85,6 +82,10 @@ func copyAction(cmd *cobra.Command, args []string) error {
 			}
 			if inst.Status == store.StatusStopped {
 				return fmt.Errorf("instance %q is stopped, run `limactl start %s` to start the instance", instName, instName)
+			}
+			u, err := osutil.LimaUser(false, inst.LimaVersion)
+			if err != nil {
+				return err
 			}
 			if legacySSH {
 				scpFlags = append(scpFlags, "-P", fmt.Sprintf("%d", inst.SSHLocalPort))
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -25,6 +25,7 @@ import (
 	"github.com/lima-vm/lima/pkg/sshutil"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/lima-vm/lima/pkg/usrlocalsharelima"
+	"github.com/lima-vm/lima/pkg/version"
 	"github.com/sirupsen/logrus"
 )
 
@@ -115,7 +116,7 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
 	if err := limayaml.Validate(y, false); err != nil {
 		return err
 	}
-	u, err := osutil.LimaUser(true)
+	u, err := osutil.LimaUser(true, version.Version)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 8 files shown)


jandubois:
This is unfortunately not going to be as simple as changing the regex. The changed default setting must be made conditional on the version of Lima that creates the VM, otherwise you are going to break existing instances. #2107 introduces code that will make this possible.

I'm also wondering if we special-case `admin` (purely for the benefit of Ubuntu), if we should disallow other well-known Linux user and group names as well? Random list: `root`, `bin`, `adm`, `daemon`, `games`, `postmaster`, `guest`, `nobody`, `ntp`, but there are plenty more. And then there are lots of common group names too.

bo17age:
> This is unfortunately not going to be as simple as changing the regex. The changed default setting must be made conditional on the version of Lima that creates the VM, otherwise you are going to break existing instances. #2107 introduces code that will make this possible.
> 
> I'm also wondering if we special-case `admin` (purely for the benefit of Ubuntu), if we should disallow other well-known Linux user and group names as well? Random list: `root`, `bin`, `adm`, `daemon`, `games`, `postmaster`, `guest`, `nobody`, `ntp`, but there are plenty more. And then there are lots of common group names too.

Hi @jandubois ,
Thank you for your comment.
assume that the current user is admin, then they can not have any instance running.
For other Linux user and group, we can add up if there is an issue.

jandubois:
> assume that the current user is admin, then they can not have any instance running.

This is not correct; only Ubuntu has a predefined `admin` group that conflicts; other templates run fine with an `admin` user.

bo17age:
> > assume that the current user is admin, then they can not have any instance running.
> 
> This is not correct; only Ubuntu has a predefined `admin` group that conflicts; other templates run fine with an `admin` user.

Thank you, I understand now.

bo17age:
Hi @jandubois, please help to review this PR again.

jandubois:
Hi @bo17age,

My thoughts around how this should be handled have evolved since you originally opened this PR. I thought you had abandoned it, which is why I didn't write them down earlier.

I think we should make the username configurable in `lima.yaml`:

```yaml
user:
  name: null
```

And then `FillDefaults` should fill in the default user name according to our rules (set `user.name` to `"lima"` if the local user name doesn't look like a valid Linux user name). This is also where the check for the `admin` user would move.

If the `lima.yaml` already specifies a user name, then we accept it without further validation (so the user could force it to be `admin` because that would still work for non-Ubuntu images).

All other locations should use `y.User.Name` instead of calling `osutil.LimaUser()`.

The user could then also add it to their `override.yaml` and always use `lima` or whatever as the user name and not use the host user name ever.

I want to use a nested `user.name` instead of a top-level `username` because we may want to allow the user to override other settings in the future, like the gid/uid. This will allow us to disable more Lima-specific configurations beyond what cloud-init does by default. @afbjorklund knows more about these requirements.

Note that I have not yet verified if all the locations that currently call `osutil.LimaUser` have easy access to the `limayaml` structure, so maybe this requires some additional refactoring.

Sorry for moving the goal-post at this late stage, but let me know if you are up for modifying your PR along those lines.

Alternatively I could review this PR as is, and then make the changes I described above in a follow-up PR myself.

bo17age:
Hi @jandubois ,
Thank you for your suggestion. Yes, I'm willing to make the change and will start soon.

jandubois:
> Yes, I'm willing to make the change and will start soon.

Hi @bo17age,

Did you get a chance to work on this?

We are planning for a Lima 1.0 release later this month, or very early in November, and there will be some changes in default behaviour already. So it would be good to make the change for the `admin` user name at the same time.

Let me know if you think you cannot finish it by some time next week, and I can give it a try myself!

jandubois:
> Let me know if you think you cannot finish it by some time next week, and I can give it a try myself!

Sorry, I meant: let me know if you can finish it yourself by next week, or if that is going to be too tight for you. In that case I can try to do it myself.

jandubois:
> In that case I can try to do it myself.

I've implemented it in #2827. Turned out to be a bit more complex than originally anticipated.

Thank you @bo17age for spending the effort on the original PR, it is appreciated, even though it will not be merged.

PR #2389, opened 2024-06-04
Update Oracle Linux templates to OL8.10/OL9.4

AmedeeBulle:
Update sample templates for new Oracle Linux releases.

For aarch64, use the "cloud images" instead of the "full" ones as they are half the size. Only difference is that they don't include bare metal drivers.

Verified both OL8/OL9 on x86_64 and aarch64 platforms.


AkihiroSuda (approved):
Thanks

Issue #824, opened 2022-04-27
Inherit system images and share them between examples


afbjorklund:
### Description

For instance Vagrant had this concept of boxes, where you would hide complexity under a name.

https://www.vagrantup.com/docs/boxes

It has the same problems of central control and what not as the yaml files themselves do, though.

https://app.vagrantup.com/ubuntu/boxes/focal64

Side note: it was only for VirtualBox, the others were https://app.vagrantup.com/generic/boxes/ubuntu2004
But it doesn't matter for this discussion, it was supposed to just reference the official Ubuntu cloud image...

----

But the idea of reducing the OS definition to a single line, would still be a useful concept ?

Instead of:

```yaml
images:
# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20220420/ubuntu-22.04-server-cloudimg-amd64.img"
  arch: "x86_64"
  digest: "sha256:de5e632e17b8965f2baf4ea6d2b824788e154d9a65df4fd419ec4019898e15cd"
- location: "https://cloud-images.ubuntu.com/releases/22.04/release-20220420/ubuntu-22.04-server-cloudimg-arm64.img"
  arch: "aarch64"
  digest: "sha256:66224c7fed99ff5a5539eda406c87bbfefe8af6ff6b47d92df3187832b5b5d4f"
```

One would just say "ubuntu-22.04-20220420" (or something).

```yaml
images:
- location: "https://download.fedoraproject.org/pub/fedora/linux/releases/35/Cloud/x86_64/images/Fedora-Cloud-Base-35-1.2.x86_64.qcow2"
  arch: "x86_64"
  digest: "sha256:fe84502779b3477284a8d4c86731f642ca10dd3984d2b5eccdf82630a9ca2de6"
- location: "https://download.fedoraproject.org/pub/fedora/linux/releases/35/Cloud/aarch64/images/Fedora-Cloud-Base-35-1.2.aarch64.qcow2"
  arch: "aarch64"
  digest: "sha256:c71f2e6ce75b516d565e2c297ea9994c69b946cb3eaa0a4bbea400dbd6f59ae6"
```

This would be "fedora-35-1.2", and similar lookups for the rest.

Currently it is all copy/paste, between the different examples.

AkihiroSuda:
You can already use `limactl start template://ubuntu", "limactl start template://fedora", ...

[...]

PR #2404, opened 2024-06-08
PoC: Allow loading images by name from imagestore

### LINKED ISSUE #2418: Simplify syntax for selecting images, with locations and digests
### Description

Make the usage of Lima more similar to a `Vagrantfile` or a `Dockerfile`.

1)
```ruby
Vagrant.configure("2") do |config|
  config.vm.box = "cloud-image/ubuntu-24.04"
end
```
https://app.vagrantup.com/cloud-image

2)
```Dockerfile
FROM ubuntu:24.04
# or, with a fully qualified registry
FROM docker.io/library/ubuntu:24.04
```
https://hub.docker.com/_/ubuntu

This by separating the usage of the images with the definition of the images.

templates/[ubuntu.yaml](https://github.com/lima-vm/lima/blob/master/examples/ubuntu.yaml)

```yaml
images:
- ubuntu-24.04
```

images/ubuntu-24.04.yaml
```yaml
images:
  arch: "x86_64"
  arch: "aarch64"
  arch: "x86_64"
  arch: "aarch64"
```

The image can be shared with multiple templates, reducing the need for copy/paste of details.

### LINKED ISSUE #2422: Simplify syntax for default mounts
### Description

Currently this block is duplicated to most of the templates:

```yaml
mounts:
- location: "~"
- location: "/tmp/lima"
  writable: true
```

afbjorklund:
This allows you to have a template like:

```yaml
images:
- ubuntu-24.04
mounts:
- location: "~"
- location: "/tmp/lima"
  writable: true
```

Loading from another document like:

```yaml
images:
  arch: "x86_64"
  arch: "aarch64"
  arch: "x86_64"
  arch: "aarch64"
```

This makes it more similar to an OCI registry (or Vagrant), and allows for even using one (or IPFS) in the future.

* #824

Perhaps "images" isn't the best name, since unlike a vagrant box there are no images contained in the directory.

~~The current implementation is just using the yaml filename, rather than the "name" field in the document itself.~~

Closes #2418
Closes #2422

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -33,21 +33,9 @@ arch: null
 
 # OpenStack-compatible disk image.
 # 🟢 Builtin default: null (must be specified)
-# 🔵 This file: Ubuntu images
+# 🔵 This file: ["default"] (see the output of `limactl info | jq .defaultImage.images`)
 images:
-# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
-- location: "https://cloud-images.ubuntu.com/releases/24.04/release-20240821/ubuntu-24.04-server-cloudimg-amd64.img"
-  arch: "x86_64"
-  digest: "sha256:0e25ca6ee9f08ec5d4f9910054b66ae7163c6152e81a3e67689d89bd6e4dfa69"
-- location: "https://cloud-images.ubuntu.com/releases/24.04/release-20240821/ubuntu-24.04-server-cloudimg-arm64.img"
-  arch: "aarch64"
-  digest: "sha256:5ecac6447be66a164626744a87a27fd4e6c6606dc683e0a233870af63df4276a"
-# Fallback to the latest release image.
-# Hint: run `limactl prune` to invalidate the cache
-- location: "https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-amd64.img"
-  arch: "x86_64"
-- location: "https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-arm64.img"
-  arch: "aarch64"
+- default
 
 # CPUs
 # 🟢 Builtin default: min(4, host CPU cores)
--- a/images/ubuntu-24.04.yaml
+++ b/images/ubuntu-24.04.yaml
@@ -1,0 +1,14 @@
+images:
+# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
+- location: "https://cloud-images.ubuntu.com/releases/24.04/release-20240821/ubuntu-24.04-server-cloudimg-amd64.img"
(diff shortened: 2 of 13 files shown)


afbjorklund:
It would be possible to have both of them in the same yaml file, as separate documents. Like for "exporting".

examples/default.yaml
```diff
@@ -19,6 +19,9 @@ os: null
 # 🟢 Builtin default: "default" (corresponds to the host architecture)
 arch: null
 
+# Name of image to use.
+image: ""
```
AkihiroSuda, line 23: An alternative design to this PR would be to support extending YAML

e.g.
```yaml
extend: "template://ubuntu"
```

I think there was a GitHub issue, but I can't find it.

pkg/limayaml/limayaml.go
```diff
@@ -10,6 +10,7 @@ type LimaYAML struct {
 	VMType             *VMType       `yaml:"vmType,omitempty" json:"vmType,omitempty"`
 	OS                 *OS           `yaml:"os,omitempty" json:"os,omitempty"`
 	Arch               *Arch         `yaml:"arch,omitempty" json:"arch,omitempty"`
+	Image              string        `yaml:"image,omitempty" json:"image,omitempty"`
 	Images             []Image       `yaml:"images" json:"images"` // REQUIRED
```
AkihiroSuda, line 14: It is confusing that `Image` (field name) is not `Image` (struct name) while `Images` is `[]Image`

afbjorklund: It's not perfect, the current solution was just "matching" what you would find in a Kubernetes yaml

afbjorklund: Another approach would be to just re-use the images field, similar to what we are doing with disk

examples/default.yaml
afbjorklund: Another alternative is copy/paste, but they are both slightly more "technical"

You might also inherit more settings than just images, by including/extending?

afbjorklund:
The goal was to make the usage of Lima more similar to a Vagrantfile or a Dockerfile.

Should probably open a new ticket for discussion about that, old one was just "reuse"

images/ubuntu-24.04.yaml
```diff
@@ -0,0 +1,15 @@
+name: "ubuntu:24.04"
```
AkihiroSuda, line 1: The name should be probably just implied by the yaml file name.

The name probably shouldn't contain `:`, as it can be easily confused with Docker Hub images

afbjorklund:
Now allows using "default" instead of the default name (currently `ubuntu-22.04`)

Maybe something similar should be added for mounts, to get the default mounts?

examples/default.yaml
AkihiroSuda: > You might also inherit more settings than just images, by including/extending?

Yes

PR #2408, opened 2024-06-09
Add possibility to download IPFS images

### LINKED ISSUE #2407: IPFS for downloading images
### Description

It should be possible to upload the images to IPFS, and then use such an address to download them p2p:

`ipfs://QmUy3RRqbpsxXYQ7yp4h2koFHdGdVTCfRiqCBLkw1JobUK/ubuntu-24.04-server-cloudimg-amd64.img`

https://docs.ipfs.tech/how-to/kubo-basic-cli/#install-kubo

Implementation itself should be similar to ORAS: #2405

afbjorklund:
Closes #2407


afbjorklund:
The lima.yaml would look something like:

```
- location: "ipfs://QmUy3RRqbpsxXYQ7yp4h2koFHdGdVTCfRiqCBLkw1JobUK/ubuntu-24.04-server-cloudimg-amd64.img"
  arch: "x86_64"
```

afbjorklund:
Note: the ipfs tool will output v0 by default, unless using `--cid-version 1`

v0: QmUy3RRqbpsxXYQ7yp4h2koFHdGdVTCfRiqCBLkw1JobUK
v1: bafybeig5sch22ecfox7gq724rz7uivydwvnnpuqdcnjz72iwelgtrakzui



```console
$ ipfs add --cid-version 0 ubuntu-24.04-server-cloudimg-amd64.img 
added QmUy3RRqbpsxXYQ7yp4h2koFHdGdVTCfRiqCBLkw1JobUK ubuntu-24.04-server-cloudimg-amd64.img
 453.00 MiB / 453.00 MiB [=================================================================================================================] 100.00%
$ ipfs add --cid-version 1 ubuntu-24.04-server-cloudimg-amd64.img 
added bafybeig5sch22ecfox7gq724rz7uivydwvnnpuqdcnjz72iwelgtrakzui ubuntu-24.04-server-cloudimg-amd64.img
 453.00 MiB / 453.00 MiB [=================================================================================================================] 100.00%

```

afbjorklund:
Using `ipfs cat` to get our own progress, and `ipfs ls` to calculate the size.

```
Hash                                                        Size     Name
bafybeibnerap2c5tnmqvyftmyxftftjewvo52acwv2gg6thpipyw5zx7fe 45613056 
bafybeidrbml5blbqj2i67x5gujqbwluis6jy5daiso6ir5pakx4hseztum 45613056 
bafybeie56ruaziz43jj3e5iih5r3fyzudeurqbpts6vwigf7nbreoep5de 45613056 
bafybeifva7p55d4szahrrxqjdjqqqlcrcfxm7n2vnpmwucrdkvcexxngmu 45613056 
bafybeie5wduudyjkbcyjvcif7i6gliipapvl5cw4wia3fqoanxunud7dyi 45613056 
bafybeidezz5b2zng3etcwbn4vnu3tn4ao3wrh5qdwzoyo7ywdp7rbymgzq 45613056 
bafybeiaf7vcteys366fmjiclktnzwuqpygpn4u2ftqrl62juaq5sgjw4be 45613056 
bafybeiasvhz6mljylbqwygb7jvp6d6sohaw325yuan5awlo7shjrlslb7m 45613056 
bafybeieiuxw2i5nkelq5tihd2tgmtxjst2ln6rslt35jxslcz4w4mjsnha 45613056 
bafybeiest5wjnmjwyp7vgawfbityfnkgqpr27bfleysbhdrhldcwxw4yq4 45613056 
bafybeif7btzsv7657kfxu7qte6sh6zgpvpsrj7dnou7usj2nwv4zf77ft4 18874368 
```

afbjorklund:
At first I was thinking that calculating the digest was unnecessary, since it already has one included in the storage.

But we still want to compare the download with the digest we are expecting, to make sure it's the _same_ image...

AkihiroSuda:
Looks good, but needs a documentation

afbjorklund:
There is a design flaw with this approach. Currently it would look like:

```yaml
  arch: "x86_64"
- location: "ipfs://QmUy3RRqbpsxXYQ7yp4h2koFHdGdVTCfRiqCBLkw1JobUK/ubuntu-24.04-server-cloudimg-amd64.img"
  arch: "x86_64"
```

That means the checksum is duplicated, between the transports. Maybe:

```yaml
  arch: "x86_64"
  cid: QmUy3RRqbpsxXYQ7yp4h2koFHdGdVTCfRiqCBLkw1JobUK
``` 

----

Note:

The CID can be calculated, without adding the image to the disk store:

```console
$ sha256sum ubuntu-24.04-server-cloudimg-amd64.img 
32a9d30d18803da72f5936cf2b7b9efcb4d0bb63c67933f17e3bdfd1751de3f3  ubuntu-24.04-server-cloudimg-amd64.img
$ ipfs add --only-hash --quieter ubuntu-24.04-server-cloudimg-amd64.img 
QmUy3RRqbpsxXYQ7yp4h2koFHdGdVTCfRiqCBLkw1JobUK
```

afbjorklund:
Unfortunately, Content-Type and Last-Modified are **not** provided by IPFS...

Current gateway only use heuristics like file magic and relative freshness.

afbjorklund:
> Looks good, but needs a documentation

Currently it assumes that IPFS Kubo is set up.

i.e. that `ipfs add` and `ipfs get` is working

https://docs.ipfs.tech/how-to/kubo-basic-cli/

pkg/limayaml/containerd.yaml
```diff
@@ -2,8 +2,10 @@ archives:
 - location: https://github.com/containerd/nerdctl/releases/download/v1.7.6/nerdctl-full-1.7.6-linux-amd64.tar.gz
   arch: x86_64
   digest: sha256:2c841e097fcfb5a1760bd354b3778cb695b44cd01f9f271c17507dc4a0b25606
+  cid: bafybeieipdaxd3fzy3j7syzzxdaqxramk65j7ajzcqgmi6b5jyq4jgbwue
```
AkihiroSuda, line 5: I guess we don't want to use ipfs by default

afbjorklund: Maybe add it to the global configuration?

afbjorklund:
Could also add support for `IPFS_GATEWAY`, as a fallback?

https://blog.ipfs.tech/ipfs-uri-support-in-curl/

If set, it would rewrite any `ipfs:` into http/https instead...

`export IPFS_GATEWAY="http://127.0.0.1:8080"`

pkg/limayaml/containerd.yaml
afbjorklund: Instead of just `LookPath("ipfs")` et al

afbjorklund: But the Content-ID (CID) needs to be inline

afbjorklund: Added a boolean `ipfs` toggle, off by default.

examples/default.yaml
```diff
@@ -299,6 +299,10 @@ rosetta:
 # 🟢 Builtin default: use name from /etc/timezone or deduce from symlink target of /etc/localtime
 timezone: null
 
+# Allow using IPFS for downloading files with a provided CID.
+# 🟢 Builtin default: false
+ipfs: null
```
AkihiroSuda, line 304: Can we just use `location: cid://<CID>` ?

AkihiroSuda: Or `location: ipfs://<CID>` ?

afbjorklund: You _can_ use ipfs: instead of https:, but then it becomes a different URL.

https://github.com/lima-vm/lima/pull/2408#issuecomment-2180784234

afbjorklund:
Regarding addressing of ipfs objects, there are some more details here:
https://github.com/ipfs/in-web-browsers/blob/master/ADDRESSING.md

"The four stages of the upgrade path for path addressing."

1. Current: HTTP-to-IPFS gateway
e.g. https://ipfs.io/ipfs/$hash
2. Short term: URL
e.g. ipfs://$hash
3. Mid term: URI
e.g. dweb:/ipfs/$hash
4. Long term: NURI
e.g. /ipfs/$hash

So it is simpler to only provide the CID/hash, as additional information?
https://docs.ipfs.tech/concepts/content-addressing/#what-is-a-cid

Since it is a multihash, it doesn't need an additional prefix*.

\* it actually already has a couple of them, but in a string encoded form
(one can use `ipfs cid format`, or https://cid.ipfs.io/, to decipher them)

```
cid: bafybeieipdaxd3fzy3j7syzzxdaqxramk65j7ajzcqgmi6b5jyq4jgbwue
# base32-cidv1-dag-pb-(sha2-256:32:8878C171ECB9C6D3F96339B8C10BC40C57BA9F8139140CC4783D4E21C49836A1)
```

examples/default.yaml
afbjorklund: It is possible to do this with the current implementation, i.e. it supports both ways

You can change the location (url), or you can provide a cid for an existing location

afbjorklund: Since the fields are optional, they should not interfer so much with existing templates?

Probably want to have some more automation / sharing in place, before adding them.

afbjorklund:
More user-facing documentation (for the website) can go in a second PR. Maybe just use the links above?

Also needs documentation on how to update images, then again we don't have any docs for sha256 either...

afbjorklund:
"ipfssum" wrapper:

```sh
#!/bin/sh
ipfs add --only-hash --cid-version=1 --progress=false "$@"
```

Something like:
```console
$ sha256sum ubuntu-24.04-server-cloudimg-*.img
0e25ca6ee9f08ec5d4f9910054b66ae7163c6152e81a3e67689d89bd6e4dfa69  ubuntu-24.04-server-cloudimg-amd64.img
5ecac6447be66a164626744a87a27fd4e6c6606dc683e0a233870af63df4276a  ubuntu-24.04-server-cloudimg-arm64.img
$ ipfssum ubuntu-24.04-server-cloudimg-*.img
added bafybeievi2i673t7kgzx6vsxc6lod3bvagzc364e6kes43p6catfowgone ubuntu-24.04-server-cloudimg-amd64.img
added bafybeich5idqhqtdh3f4nxq6kv6l2fd2nsvynha3vdqek4a7mu7nawp6sm ubuntu-24.04-server-cloudimg-arm64.img

```

examples/default.yaml
```diff
@@ -39,9 +39,11 @@ images:
 - location: "https://cloud-images.ubuntu.com/releases/24.04/release-20240821/ubuntu-24.04-server-cloudimg-amd64.img"
   arch: "x86_64"
   digest: "sha256:0e25ca6ee9f08ec5d4f9910054b66ae7163c6152e81a3e67689d89bd6e4dfa69"
+  cid: "bafybeievi2i673t7kgzx6vsxc6lod3bvagzc364e6kes43p6catfowgone"
```
AkihiroSuda, line 42: I don't think we should add the cid to the built-in templates unless the distro publishes the cid by themselves, as computing the cid incurs downloading and reading the whole images.

The same applies to contained too.

You may still add experimental/ubuntu-ipfs.yaml with `location: ipfs://CID`.
I still think the locator object doesn't need a dedicated field for `cid`

AkihiroSuda: Also hack/update-template.sh should skip ubuntu-ipfs.yaml by default

afbjorklund: I can remove the CID feature, than the IPFS would be more similar to the ORAS.

afbjorklund:
We might want to refactor downloader separately, then it would be easier to add handlers...

i.e. ipfs for `ipfs:` and oras for `oras:`, and what other alternatives to plain `http:` and `https:`

PR #2410, opened 2024-06-10
Handle missing /etc/environment file in boot

### LINKED ISSUE #2409: Cannot Create OpenSUSE Tumbleweed Instance
### Description

Any time I use the existing template for OpenSUSE Tumbleweed, I get stuck in a loop of "user session is ready for ssh". 

No idea if this is known or not, but I figured I should submit an issue for it.

afbjorklund:
Closes #2409


jandubois (approved):
Thanks, LGTM

PR #2425, opened 2024-06-17
Use dummy WSL distro instead of OpenSUSE

jandubois:
The OpenSUSE installation hangs on a YaST dialog invoked via a first-run script.

Since we don't use the WSL distro anyways, just import an empty tarball instead, which is also faster because it doesn't need to download the distro image.

The purpose of this PR is to unblock the "test / Windows tests" check that seems currently be blocking.


AkihiroSuda (approved):
Thanks

PR #2427, opened 2024-06-18
Allow compressing lima-guestagent with zstd

### LINKED ISSUE #2426: Compress the lima-guestagent in the installation directory
### Description

There are a lot of architectures, and not all of them are used...

We could compress them in the install, and decompress on host.

The on-disk lima installation goes from 175M to 68M, with gzip.

Note: `lima-guestagent` is still uncompressed, in the `cidata.iso`

afbjorklund:
Closes #2426

Compressing using highest level 19 takes forever (30 seconds), so should use something like 3 during development...

Everything is handled in the code, so that user doesn't need to have `zstd` installed - neither on the host or in instance

### DIFF
--- a/Kconfig
+++ b/Kconfig
@@ -31,15 +31,35 @@ config GUESTAGENT_ARCH_RISCV64
 		Build lima-guestagent for "riscv64" Arch
 	default y
 
-config GUESTAGENT_COMPRESS_ZSTD
-	bool "guestagent compress zstd"
+config GUESTAGENT_COMPRESS
+	bool "guestagent compress"
 	help
-		Compress lima-guestagent using zstd
+		Compress lima-guestagent
 	default n
 
+choice GUESTAGENT_COMPRESS_PROGRAM
+	depends on GUESTAGENT_COMPRESS
+	bool "guestagent compress program"
+	default GUESTAGENT_COMPRESS_GZIP
+
+config GUESTAGENT_COMPRESS_GZIP
+	depends on GUESTAGENT_COMPRESS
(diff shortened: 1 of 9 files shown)


afbjorklund:
Example development config:

```Makefile
CONFIG_GUESTAGENT_OS_LINUX=y
# CONFIG_GUESTAGENT_ARCH_ARMV7L is not set
# CONFIG_GUESTAGENT_ARCH_RISCV64 is not set
CONFIG_GUESTAGENT_COMPRESS_ZSTD=y
CONFIG_GUESTAGENT_COMPRESS_LEVEL=3
```

Example release config:
```Makefile
CONFIG_GUESTAGENT_OS_LINUX=y
CONFIG_GUESTAGENT_COMPRESS_ZSTD=y
CONFIG_GUESTAGENT_COMPRESS_LEVEL=19
```

afbjorklund:
Might want to increase gzip compression from 6 to 9, as well ? `GZIP=-9`

Unfortunately someone has removed the option to choose compression level:
`gzip: warning: GZIP environment variable is deprecated; use an alias or script`

So might have to pass it explicitly to `tar`. (with `--use-compress-program`) :pensive:

AkihiroSuda:
Can we just use gzip, to avoid requiring an extra dependency for downstream packages

afbjorklund:
This PR is mostly for evaluation, to compare the sizes/speeds of the different options...

* none
* gzip (-6)
* gzip (-9)
* zstd (-3)
* zstd (-19)

Issue #2387, opened 2024-06-03
Old versions of fuse have read-only config file


afbjorklund:
### Description

This causes the probe to loop on startup, even though everything is fine.

```
INFO[0011] [hostagent] Waiting for the essential requirement 4 of 4: "/etc/fuse.conf (/etc/fuse3.conf) to contain \"user_allow_other\"" 
INFO[0051] [hostagent] Waiting for the essential requirement 4 of 4: "/etc/fuse.conf (/etc/fuse3.conf) to contain \"user_allow_other\"" 
INFO[0091] [hostagent] Waiting for the essential requirement 4 of 4: "/etc/fuse.conf (/etc/fuse3.conf) to contain \"user_allow_other\"" 
```

`-rw-r----- 1 root fuse 297 Jun  3 15:31 /etc/fuse.conf`

Seen with Ubuntu 15.04, but only when enabling the mounts (with sshfs)

---

The default was changed in Ubuntu 20.04 (fuse 2.9.3-5)

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=733312

The workaround was to disable the mounts (for k8s v1.0.0)

```yaml
mounts: []
```

afbjorklund:
i.e. suggested change is to use 'sudo grep`, when checking for the config

`pkg/cidata/cidata.TEMPLATE.d/boot/35-setup-packages.sh:		if ! grep -q "^user_allow_other" "${fuse_conf}"; then`

PR #2433, opened 2024-06-20
Adds latest still supported ubuntu lts templates

madalinignisca:
I wanted to add a few more ubuntu lts templates that I use often on local work I do and thought it would be useful for others to have them available.

### DIFF
--- a/examples/deprecated/ubuntu-18.04.yaml
+++ b/examples/deprecated/ubuntu-18.04.yaml
@@ -1,0 +1,20 @@
+# This template requires Lima v0.7.0 or later.
+images:
+# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
+- location: "https://cloud-images.ubuntu.com/releases/18.04/release-20230607/ubuntu-18.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
+  digest: "sha256:8dd2e6b5e5aad20c3f836123b300cba9861249408cbb07c359145a65d6bab6b6"
+- location: "https://cloud-images.ubuntu.com/releases/18.04/release-20230607/ubuntu-18.04-server-cloudimg-arm64.img"
+  arch: "aarch64"
+  digest: "sha256:9930a57427ad8c9e50a84a36132cd5ad81751b64ee678062dcc354a9b84654f7"
+# Fallback to the latest release image.
+# Hint: run `limactl prune` to invalidate the cache
+- location: "https://cloud-images.ubuntu.com/releases/18.04/release/ubuntu-18.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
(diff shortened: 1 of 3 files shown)


examples/ubuntu-22.04.yaml
AkihiroSuda, line 1: Should we now `mv ubuntu.yaml ubuntu-24.04.yaml; ln -s ubuntu-24.04.yaml ubuntu.yaml` for consistency?

examples/ubuntu-20.04.yaml
```diff
@@ -0,0 +1,20 @@
+# This template requires Lima v0.7.0 or later.
+images:
+# Try to use release-yyyyMMdd image if available. Note that release-yyyyMMdd will be removed after several months.
+- location: "https://cloud-images.ubuntu.com/releases/20.04/release-20240612/ubuntu-20.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
+  digest: "sha256:7fce49f6c9afd2cdd9c26618b056f8806e500dd4d6eff516c9c6b5fab12e7f79"
+- location: "https://cloud-images.ubuntu.com/releases/20.04/release-20240612/ubuntu-20.04-server-cloudimg-arm64.img"
```
AkihiroSuda, line 7: It is hard to keep every image up-to-date, so we may need to have an automation script before adding new templates
- https://github.com/lima-vm/lima/issues/1347

examples/ubuntu-22.04.yaml
madalinignisca: Would be ok

afbjorklund:
Ubuntu 16.04 and 18.04 are probably old enough to hide them from the general public (even if supported)

Having 20.04 and 22.04 should be old enough I think? And then we don't have to re-live all old bugs, like:

* #2387

Should probably delete centos-7.yaml (after tomorrow) as well, even if it is from 2014 rather than 2018...

madalinignisca:
Sorry, but my intention with this PR is only to introduce common ubuntu templates, not change anything not having a relation with them.

I agree automation to ensure latest releases would be great, but even in my case, an update after deployment is not such a pain, as usually even in public cloud environments, full upgrade is part of my cloud-init setup (even public cloud ain't having latest and greatest releases).

For next step, I'm willing when I get some time off for some automation for some of the templates that concern me, being Debian and Ubuntu, as I don't use any other distributions for server work.

AkihiroSuda:
We now have an updater for Ubuntu (https://github.com/lima-vm/lima/blob/master/hack/update-template-ubuntu.sh), so this PR can be revisited again after rebasing.

examples/deprecated/ubuntu-18.04.yaml
AkihiroSuda, line 1: This should be probably removed due to EOL

AkihiroSuda:
Opened an alternative PR:
- https://github.com/lima-vm/lima/pull/2795

PR #2438, opened 2024-06-23
Allow compressing lima-guestagent with gzip

### LINKED ISSUE #2426: Compress the lima-guestagent in the installation directory

afbjorklund:
Remove all compression configuration, and make it boolean.
```
         compressed        uncompressed  ratio uncompressed_name
           10829118            39059456  72.3% _output/share/lima/lima-guestagent.Linux-aarch64
           11190957            38797312  71.2% _output/share/lima/lima-guestagent.Linux-armv7l
           11431176            39387136  71.0% _output/share/lima/lima-guestagent.Linux-riscv64
           12076900            40189952  70.0% _output/share/lima/lima-guestagent.Linux-x86_64
           45528151           157433856  71.1% (totals)
```

Could become the default config (y), after some evaluation.

Replaces #2427
Closes #2426


AkihiroSuda (approved):
Thanks

jandubois (dismissed):
I feel like `usrlocalsharelima.Open` does not belong in that package because it has nothing to do with that location, but I don't have any other suggestion. Maybe `fileutils`?

Anyways, I'm fine with merging as-is.

afbjorklund:
I can move the code back to pkg/cidata again, now that it is a lot smaller.

afbjorklund:
> Maybe fileutils?

Something like `gzopen` would have been nice to have.

But I can't find the format detection (file magic) either?

pkg/cidata/cidata.go
```diff
@@ -341,14 +342,31 @@ func GenerateISO9660(instDir, name string, y *limayaml.LimaYAML, udpDNSLocalPort
...
+		reader, err = gzip.NewReader(guestAgent)
+		if err != nil {
+			return err
+		}
+	} else if err != nil {
+		return err
+	} else {
+		reader = guestAgent
+	}
+	defer guestAgent.Close()
```
jandubois, line 366: I find the logic slightly hard to follow and think this is simpler:

```golang
	var guestAgent io.ReadCloser
	guestAgent, err = os.Open(guestAgentBinary)
	if err != nil {
		if !errors.Is(err, os.ErrNotExist) {
			return err
		}
		compressedGuestAgent, err := os.Open(guestAgentBinary + ".gz")
		if err != nil {
			return err
		}
		logrus.Debugf("Decompressing %s.gz", guestAgentBinary)
		guestAgent, err = gzip.NewReader(compressedGuestAgent)
		if err != nil {
			return err
		}
	}
	defer guestAgent.Close()
	layout = append(layout, iso9660util.Entry{
		Path:   "lima-guestagent",
		Reader: guestAgent,
	})
```

jandubois (dismissed):
The logic seems correct, but I would prefer a simplified expression that keeps the error handling close to the location the error occurred.

jandubois (approved):
Thanks, LGTM

PR #2440, opened 2024-06-24
30-install-packages.sh: Use `head` instead of `hexdump`

norio-nomura:
The code changes in `30-install-packages.sh` modify the package installation script to use the `head` command instead of `hexdump` for checking the ELF header of the `apt-get` command.  
This change will allow the use of "Ubuntu Minimal Cloud Images" that do not include `hexdump`.  
https://cloud-images.ubuntu.com/minimal/releases/


AkihiroSuda (approved):
Thanks

PR #2442, opened 2024-06-24
Allow providing scripts as separate files

afbjorklund:
The "path" provided will be read, and used instead of the "script".

```yaml
provision:
- mode: user
  path: foo.sh

probes:
- mode: readiness
  path: bar.sh
```

* #2441

----

There are some unsolved issues with this, regarding relative paths.

* https://github.com/lima-vm/lima/issues/893

Currently it requires that `limactl` is always run from the top directory.

The paths PWD should probably be recorded in the instance directory?

### DIFF
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -198,6 +198,7 @@ type Provision struct {
 	Mode                            ProvisionMode `yaml:"mode" json:"mode"` // default: "system"
 	SkipDefaultDependencyResolution *bool         `yaml:"skipDefaultDependencyResolution,omitempty" json:"skipDefaultDependencyResolution,omitempty"`
 	Script                          string        `yaml:"script" json:"script"`
+	Path                            string        `yaml:"path,omitempty" json:"path,omitempty"`
 	Playbook                        string        `yaml:"playbook,omitempty" json:"playbook,omitempty"`
 }
 
@@ -217,6 +218,7 @@ type Probe struct {
 	Mode        ProbeMode // default: "readiness"
 	Description string
 	Script      string
+	Path        string `yaml:",omitempty" json:",omitempty"`
 	Hint        string
 }
 
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -208,6 +208,7 @@ containerd:
 # Provisioning scripts need to be idempotent because they might be called
 # multiple times, e.g. when the host VM is being restarted.
 # The scripts can use the following template variables: {{.Home}}, {{.UID}}, {{.User}}, and {{.Param.Key}}
+# They can be provided inline (as field `script`), or in external files (as field `path`).
 # 🟢 Builtin default: null
 # provision:
 # # `system` is executed with root privileges
@@ -251,6 +252,7 @@ containerd:
 # Probe scripts to check readiness.
 # The scripts run in user mode. They must start with a '#!' line.
 # The scripts can use the following template variables: {{.Home}}, {{.UID}}, {{.User}}, and {{.Param.Key}}
+# They can be provided inline (as field `script`), or in external files (as field `path`).
 # 🟢 Builtin default: null
 # probes:
 # # Only `readiness` probes are supported right now.
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -311,7 +311,10 @@ func GenerateISO9660(instDir, name string, instConfig *limayaml.LimaYAML, udpDNS
 		args.CACerts.Trusted = append(args.CACerts.Trusted, cert)
 	}
 
-	args.BootCmds = getBootCmds(instConfig.Provision)
+	args.BootCmds, err = getBootCmds(instConfig.Provision)
+	if err != nil {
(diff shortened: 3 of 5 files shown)


afbjorklund:
There are some issues with relative paths for the scripts, as has been discussed before.

afbjorklund:
The provision scripts are recorded in the cidata.iso, so they should be OK with changing directory.

But the probes are run from the hostagent, so they don't really record the "working directory" now.

afbjorklund:
As noted in the older issue, this becomes extra interesting if lima.yaml is loaded from an URL.

pkg/limayaml/limayaml.go
```diff
@@ -206,6 +207,7 @@ type Probe struct {
 	Mode        ProbeMode // default: "readiness"
 	Description string
 	Script      string
+	Path        string `yaml:",omitempty" json:",omitempty"`
```
AkihiroSuda, line 210: Shouldn't we consistently use the `File` object?

website/content/en/docs/dev/Internals/_index.md
```diff
@@ -33,6 +33,7 @@ An instance directory contains the following files:
 
 Metadata:
 - `lima-version`: the Lima version used to create this instance
+- `lima-workdir`: the Lima workdir used to create this instance
```
AkihiroSuda, line 36: What is workdir? Why does it need to be a dir?

afbjorklund: The workdir is the $PWD from when the limactl start was first run.

It is used to find relative paths, for the scripts and the playbooks...

Otherwise, you need to `cd` to the directory before any `limactl`.

Or make all paths absolute, making them static and unportable?

pkg/limayaml/limayaml.go
afbjorklund: The file objects also adds arch and digest, neither of which is applicable here...

We could rename `Path` to `Location`, to allow for running scripts from URLs?

website/content/en/docs/dev/Internals/_index.md
AkihiroSuda: `File` can be extended to support relative paths?

pkg/limayaml/limayaml.go
AkihiroSuda: at least digest seems to be applicable here

cmd/limactl/start.go
```diff
@@ -343,6 +343,13 @@ func createInstance(ctx context.Context, st *creatorState, saveBrokenEditorBuffe
 	if err := os.WriteFile(filePath, st.yBytes, 0o644); err != nil {
 		return nil, err
 	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return nil, err
+	}
+	if err := os.WriteFile(filepath.Join(instDir, filenames.LimaWorkDir), []byte(wd), 0o644); err != nil {
```
AkihiroSuda, line 350: This could be a symlink ?

pkg/limayaml/limayaml.go
afbjorklund: Hmm, probably needs more design then

afbjorklund: i.e. the wish was to "include" common scripts

If the digest changes every time you edit the script, you still need to copy/paste the sha256?

jandubois:
I think the proper way to handle local script files is to copy them during `limactl create` into a subdirectory of the instance directory, e.g. `~/.lima/debian/scripts/foo.sh` (with a mechanism to prevent collisions[^1]).

[^1]: E.g. a second `foo.sh` would become `foo-2.sh` etc.

Then `lima.yaml` is modified with updated relative pathnames:

```yaml
provision:
- mode: user
  path: scripts/foo.sh

probes:
- mode: readiness
  path: scripts/bar.sh
```

This makes sure that when the user in the future runs `limactl edit` and the `cidata.iso` is re-created, that the instance will continue to use the original scripts used to create the instance, and not a modified version that exists elsewhere on the filesystem. Or fail because the original script files have since been deleted.

I don't know how this would work for ansible playbooks, as they can in turn reference additional files, and parsing/rewriting the whole chain seems a bit too complex, but idk what else you can do.

I would still say that we copy the playbook itself, and if that in turn references additional files, then it is up to the user to keep track of them. We can try to display at least a warning when this happens.

I still don't know how external script files would be handled for templates loaded via `template://` URL, but whatever the mechanism, they also need to be copied into the `scripts/` subdirectory.

jandubois:
> For these reasons I think even external scripts referenced by absolute filename (or URL) should be copied into the instance directory and decoupled from the original source.

This is also a security concern because a hijacked domain could inject a malicious provisioning script into an existing instance.

afbjorklund:
> I think relative filenames should always be relative to the location of `lima.yaml` itself, and not depend on the cwd of the `limactl` process.

We can start with this change, and look at copying them from the template folder to the instance folder later.

jandubois:
@afbjorklund This functionality is now provided by #3072; I think this PR can be closed.

PR #2446, opened 2024-06-25
Add make target for generating current arch config

afbjorklund:
`make nativeconfig`

* https://github.com/lima-vm/lima/issues/2426#issuecomment-2187316192

```console
$ make config
cp config.mk .config
/usr/local/bin/kconfig-conf Kconfig
*
* Lima
*
guestagent OS: Linux (GUESTAGENT_OS_LINUX) [Y/n/?] 
guestagent Arch: x86_64 (GUESTAGENT_ARCH_X8664) [Y/n/?] 
guestagent Arch: aarch64 (GUESTAGENT_ARCH_AARCH64) [Y/n/?] 
guestagent Arch: armv7l (GUESTAGENT_ARCH_ARMV7L) [Y/n/?] 
guestagent Arch: riscv64 (GUESTAGENT_ARCH_RISCV64) [Y/n/?] 
#
# configuration written to .config
#
```

```console
$ make menuconfig
cp config.mk .config
MENUCONFIG_STYLE=aquatic \
/usr/local/bin/kconfig-mconf Kconfig


*** End of the configuration.
*** Execute 'make' to start the build or try 'make help'.
```

lima-menuconfig

The goal was to generate a config file, with only the current architecture.

---

Default config, empty `defconfig`:

```Makefile
#
# Automatically generated file; DO NOT EDIT.
# Lima
#
CONFIG_GUESTAGENT_OS_LINUX=y
```

Could have used files:

```
arch/x86_64/defconfig
arch/aarch64/defconfig
```

As in the tool default:

`ARCH=x86_64 kconfig-conf --defconfig Kconfig`
```
***
*** Can't find default configuration "arch/x86_64/defconfig"!
***
```

arch/x86_64/config.mk
```
CONFIG_GUESTAGENT_OS_LINUX=y
CONFIG_GUESTAGENT_ARCH_AARCH64=n
CONFIG_GUESTAGENT_ARCH_ARMV7L=n
CONFIG_GUESTAGENT_ARCH_RISCV64=n
```

alternatively, defconfig:
```Makefile
# CONFIG_GUESTAGENT_ARCH_X8664 is not set
# CONFIG_GUESTAGENT_ARCH_ARMV7L is not set
# CONFIG_GUESTAGENT_ARCH_RISCV64 is not set
```

`ARCH=aarch64 kconfig-conf --defconfig Kconfig`
```
***
*** Can't find default configuration "arch/aarch64/defconfig"!
***
```

arch/aarch64/config.mk
```
CONFIG_GUESTAGENT_OS_LINUX=y
CONFIG_GUESTAGENT_ARCH_X8664=n
CONFIG_GUESTAGENT_ARCH_ARMV7L=n
CONFIG_GUESTAGENT_ARCH_RISCV64=n
```

alternatively, defconfig:
```Makefile
# CONFIG_GUESTAGENT_ARCH_AARCH64 is not set
# CONFIG_GUESTAGENT_ARCH_ARMV7L is not set
# CONFIG_GUESTAGENT_ARCH_RISCV64 is not set
```

The framework is overkill here, compared with kernel.

https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,11 @@ menuconfig: Kconfig
 	MENUCONFIG_STYLE=aquatic \
 	$(KCONFIG_MCONF) $<
 
+# Generate config for the current architecture (only)
+# This is done to avoid a dependency on KCONFIG tools
+nativeconfig: .config
+	./hack/native-config.sh $<
+
 # Copy the default config, if not overridden locally
 # This is done to avoid a dependency on KCONFIG tools
 .config: config.mk
--- a/hack/native-config.sh
+++ b/hack/native-config.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+file="$1"
+test -e "$file" || exit 1
+script=""
+ARCH=${ARCH:-$(uname -m | sed -e s/arm64/aarch64/)}
+for arch in x86_64 aarch64 armv7l riscv64; do
+	if [ $arch = "$ARCH" ]; then
+		x=y
+	else
+		x=n
+	fi
+	: ${arch^^}
+	config="CONFIG_GUESTAGENT_ARCH_${_//_/}"
+	script="$script;s/$config=./$config=$x/"
+done
+sed -e "$script" "$file" >"$file.new" && mv "$file.new" "$file"


Makefile
```diff
@@ -72,6 +72,11 @@ menuconfig: Kconfig
 	MENUCONFIG_STYLE=aquatic \
 	$(KCONFIG_MCONF) $<
 
+# Generate config for the current architecture (only)
```
AkihiroSuda, line 75: Shouldn't we build arm64 and amd64 by default?

afbjorklund: Possibly. But we build for _everything_ by default. This was a request for something between "minimal" and "all"

> What I have wanted in the past is something like `make minimal templates install`, so I still get all the templates, but only the native guestagent. Because minimal only installs the default template.

Going with files (instead of a script) is probably easiest for discussion, even if config is somewhat duplicated.

afbjorklund:
We could use this target, when testing with LXD... Not sure why the cache isn't more effective.

norio-nomura:
I agree with the opinion that a way to install only the native guest agent is needed, but I don’t quite understand why that would require `kconfig` support. It seems like this could be achieved without using `kconfig`.

norio-nomura:
I don’t necessarily want to block this PR, but since I don't understand the reasons for its implementation, I don’t have the criteria to judge its merits and therefore cannot evaluate the PR.

afbjorklund:
> I agree with the opinion that a way to install only the native guest agent is needed, but I don’t quite understand why that would require `kconfig` support. It seems like this could be achieved without using `kconfig`.

The Kconfig is the configuration for the Makefile code, so enabling/disabling features is done there.
The output (`.config`) is not complicated: https://github.com/lima-vm/lima/blob/master/config.mk

A similar system is autoconf for automake, you might have seen the `./configure` script being used?
That system is still (rather) simple for the end user, but is much more complicated for the developer.

afbjorklund:
Another approach would be to add another boolean, that would override the other ones.

Makefile
```diff
@@ -72,6 +72,11 @@ menuconfig: Kconfig
...
+# This is done to avoid a dependency on KCONFIG tools
+defconfig:
```
norio-nomura, line 77: What does the name `defconfig` mean? If it stands for something like 'default config,' wouldn't `config.mk` already be the default? I believe the name should be something that users searching for the feature added by this PR can easily find.

afbjorklund: It's a kconfig thing, I think it stands for default config for this specific architecture or some-such?
It's unrelated to the "default config" (all options off / empty config file), in that it does have config.

The way it is created is that you select the options you want, and then run "savedefconfig" to defconfig.
The normal .config file that is generated has **all** available options, including ones you didn't pick.

afbjorklund: It is useful when you are building something linux or busybox, or an embedded system like buildroot.

It is totally overkill for something with a half dozen options (like Lima) - unless you are already used to it :-)

afbjorklund:
> I understand what this PR is doing, but I still don’t understand why it’s necessary.

The only requirement was to be able to build Lima for the host architecture only.

Building "minimal" does this, but it doesn't include _any_ templates or guestagents.

norio-nomura:
> The only requirement was to be able to build Lima for the host architecture only...

I’ve drafted a PR #2638 that refactors the Makefile, including a solution to that requirement.

Makefile
afbjorklund: The reason for doing a single script instead of multiple files, is that there would be duplication of other config.

afbjorklund:
In other news, I updated [kconfig-frontends](https://github.com/afbjorklund/kconfig-frontends) so that it builds without warnings (on Linux or Mac).

You do have to `brew install qt@5` if you want the Qt GUI, it did not yet support Qt6 in Linux 5.4.

afbjorklund:
An improvement would be to rename it from "defconfig" to "nativeconfig" or somesuch.

And to copy the non-arch fields from the `config.mk`, instead of duplicating them here.

norio-nomura (commented):
It might be preferable to use `go env GOHOSTARCH` instead of `uname -m`. This is because on ARM versions of Windows, `uname -m` returns `x86_64`, as only the x86_64 version is available, while Go correctly returns `arm64` since it has an ARM version.  
Additionally, while using `make GOOS=<GOOS> GOARCH=<GOARCH>` for cross-compilation, I'm unsure how this target should behave, which would help determine whether to use `GOARCH` or `go env GOHOSTARCH`.

afbjorklund:
I think I will just close this, since there are lot of subtargets now.

`make native-guestagent`

PR #2178, opened 2024-02-02
nerdctl: update to v2.0.0

AkihiroSuda:
https://github.com/containerd/nerdctl/releases/tag/v2.0.0

PR #2447, opened 2024-06-26
cidata: fix upgrading nerdctl

AkihiroSuda:
`containerd-rootless-setuptool.sh uninstall` was called without uninstalling BuildKit

- - -
Cherry-picked from:
- #2178


afbjorklund (approved):
Looks good to me, but haven't tested it

PR #1625, opened 2023-06-13
vz: Fix usernet sshLocalPort check failure for vz driver

fyuan1316:
fix #1620

PR #2455, opened 2024-06-30
Use environment var for ResolveIPAddressTimeout

### LINKED ISSUE #2454: [hostagent] Driver stopped due to error: \"usernet unable to resolve IP for SSH forwarding\"
### Description

**Description**
We are experiencing intermittent failures with GitHub Actions, which have become more frequent recently. Here are the action links for reference: [Link 1](https://github.com/OpenDevin/OpenDevin/actions/runs/9643397151/job/26593302550?pr=2449), [Link 2](https://github.com/OpenDevin/OpenDevin/actions/runs/9664771891/job/26660252033?pr=2628). The issue resolves upon retrying; see the action from the second attempt [here](https://github.com/OpenDevin/OpenDevin/actions/runs/9643397151/job/26594205397?pr=2449).

**Version**
lima was installed using the command: brew install --HEAD colima.

**Operating System**
 macOS Intel <= 13 (Ventura)
 
 **Other references**
#1620 https://github.com/abiosoft/colima/issues/1047

SmartManoj:
Like #1625; Resolves #2454 

---

Use the `CUSTOM_TIMEOUT` environment variable to configure the timeout for resolving IP addresses.

* Import the `strconv` package.
* Add a new variable `customTimeout` to read the `CUSTOM_TIMEOUT` environment variable.
* Parse the `CUSTOM_TIMEOUT` value and set the `timeout` variable accordingly.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/lima-vm/lima/pull/2455?shareId=438e1b28-7390-495a-8f35-eceb744637cc).

### DIFF
--- a/pkg/networks/usernet/client.go
+++ b/pkg/networks/usernet/client.go
@@ -7,6 +7,8 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"os"
+	"strconv"
 	"time"
 
 	gvproxyclient "github.com/containers/gvisor-tap-vsock/pkg/client"
@@ -74,11 +76,19 @@ func (c *Client) ResolveAndForwardSSH(ipAddr string, sshPort int) error {
 }
 
 func (c *Client) ResolveIPAddress(ctx context.Context, vmMacAddr string) (string, error) {
-	timeout := time.After(2 * time.Minute)
+	resolveIPAddressTimeout := 2 * time.Minute
+	resolveIPAddressTimeoutEnv := os.Getenv("LIMA_USERNET_RESOLVE_IP_ADDRESS_TIMEOUT")
+	if resolveIPAddressTimeoutEnv != "" {
+		if parsedTimeout, err := strconv.Atoi(resolveIPAddressTimeoutEnv); err == nil {
+			resolveIPAddressTimeout = time.Duration(parsedTimeout) * time.Minute
+		}
+	}
+	ctx, cancel := context.WithTimeout(ctx, resolveIPAddressTimeout)
+	defer cancel()
 	ticker := time.NewTicker(500 * time.Millisecond)
 	for {
 		select {
-		case <-timeout:
+		case <-ctx.Done():
 			return "", errors.New("usernet unable to resolve IP for SSH forwarding")
 		case <-ticker.C:
 			leases, err := c.Leases(ctx)


SmartManoj:
Will you follow [this option](https://community.openhab.org/t/dco-check-signing-off-with-github-web-editor-explanation/83330#:~:text=If%20you%20know%20you%20have%20a%20small%20changed%20introduced%20which%20doesn%E2%80%99t%20need%20a%20sign%2Doff%20statement%2C%0Athe%20reviewer%20(this%20is%20mostly%20myself%20at%20the%20time%20of%20writing%20this%20Post)%20will%20take%20care%20of%20it%20and%20override%20it%20if%20its%20reasonable.)?

----

How about adding a GitHub Actions workflow to sign off commits automatically if not done? This will help ensure compliance with the Developer Certificate of Origin (DCO) and streamline the contribution process.



Here is an example workflow that can be added to the repository:

```yaml
name: Auto Sign-Off

on:
  pull_request:
    types: [opened, synchronize]

jobs:
  signoff:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout code
        uses: actions/checkout@v2

      - name: Set up Git
        run: |
          git config --global user.name "github-actions[bot]"
          git config --global user.email "github-actions[bot]@users.noreply.github.com"

      - name: Sign off commits
        run: |
          for commit in $(git rev-list --reverse HEAD ^origin/main); do
            git cherry-pick $commit --strategy=recursive -X theirs --signoff
          done

      - name: Push changes
        run: |
          git push origin HEAD:refs/heads/${{ github.head_ref }} --force
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}

**Comment by AkihiroSuda** [comment]
> How about adding a GitHub Actions workflow to sign off commits automatically if not done? This will help ensure compliance with the Developer Certificate of Origin (DCO) and streamline the contribution process.

Nope, rather opposite, as automated signs are unlikely valid

File: pkg/networks/usernet/client.go [hunk 1]
```diff
@@ -74,7 +74,7 @@ func (c *Client) ResolveAndForwardSSH(ipAddr string, sshPort int) error {
 }
 
 func (c *Client) ResolveIPAddress(ctx context.Context, vmMacAddr string) (string, error) {
-	timeout := time.After(2 * time.Minute)
+	timeout := time.After(3 * time.Minute)
```
AkihiroSuda, line 77: Why does it take more than 2 minutes?

SmartManoj: During #1625, Why does it take more than 1 minute?

afbjorklund: Seems rather arbitrary, since it started out as 500ms

Seeing the same type of issues when working with mDNS - sometimes my machine is really slow to respond (like when waking up from idle) - but I also don't want to wait several seconds for when the machine really is gone.

jandubois: I find that the GitHub macOS runners are painfully slow, so I can see that you may need special accommodations for them. Maybe instead of relaxing timeouts generally, we should have some kind of setting, like `LIMA_SLOW_HOST_FACTOR=2` that would double all the timeouts in Lima to be more forgiving?

Feels quite kludgy, but still better than making the machine slow for the regular use case.

balajiv113: @SmartManoj Can you try setting `--cpus=1 --memory=1 ` see this helps ??

We are also using macOS runners. 2mins as timeout is working fine. I feel like even after the above 3mins change you will see some failures like this.

SmartManoj: How reducing the memory will help?

balajiv113: Problem with github macos runners is that if it takes more memory they will terminate the vm process / take more time to start basically slowers all other stuff.

In this memory, cpu setup we saw better success rate with macOS runners

SmartManoj: Still the same; https://github.com/OpenDevin/OpenDevin/actions/runs/9802679379/job/27067695657

pkg/networks/usernet/client.go
```diff
@@ -74,11 +75,14 @@ func (c *Client) ResolveAndForwardSSH(ipAddr string, sshPort int) error {
...
-	timeout := time.After(2 * time.Minute)
+	timeout := 2 * time.Minute
+	if os.Getenv("GITHUB_ACTIONS") != "" {
+		timeout = 3 * time.Minute
+	}
 	ticker := time.NewTicker(500 * time.Millisecond)
 	for {
 		select {
-		case <-timeout:
+		case <-time.After(timeout):
```
jandubois, line 85: Is this correct? Doesn't it create a new channel each time through the for loop? And since the ticker triggers every 500ms, this would effectively disable the timeout?

Or am I confused about how this would work?

SmartManoj: Sorry, you are right. 👍

jandubois (dismissed):
I'm fine with this change, so am approving it.

As it is kind of a hack though, I won't merge it unless it gets another approval from another @lima-vm/reviewers

afbjorklund:
> it is kind of a hack though, I won't merge it unless it gets another approval

It was already hack with 2 minutes, I don't see how 3 minutes make it worse...

* https://github.com/lima-vm/lima/pull/1625

The real hack is making GitHub Actions into a special case, is what I'm thinking

pkg/networks/usernet/client.go
```diff
@@ -74,11 +75,15 @@ func (c *Client) ResolveAndForwardSSH(ipAddr string, sshPort int) error {
 }
 
 func (c *Client) ResolveIPAddress(ctx context.Context, vmMacAddr string) (string, error) {
-	timeout := time.After(2 * time.Minute)
+	timeout := 2 * time.Minute
+	if os.Getenv("GITHUB_ACTIONS") != "" {
```
SmartManoj, line 79: @afbjorklund Did you mean this?

SmartManoj:
Why were the [tests ](https://github.com/lima-vm/lima/actions/runs/9803979726/job/27070987706?pr=2455)failed with `msg="did not receive an event with the \"running\" status"`?

pkg/networks/usernet/client.go
afbjorklund: Yeah, if it should be 3 (or 5) minutes everywhere ? And what is taking so long

SmartManoj: system resource contention?

jandubois:
> Why was the [test](https://github.com/lima-vm/lima/actions/runs/9803979726/job/27070987706?pr=2455) failed with `msg="did not receive an event with the \"running\" status"`?

We can re-run the test to see if it is just a flake, but you also need to squash your commits together and sign it with a DCO before it can be considered for merging (assuming there will otherwise be consensus that the change is desirable).

SmartManoj:
> Why was the [test](https://github.com/lima-vm/lima/actions/runs/9803979726/job/27070987706?pr=2455) failed with `msg="did not receive an event with the \"running\" status"`?

Due to this timeout. https://github.com/lima-vm/lima/blob/463ed828cb5fa0a2851a54f15941f85c189d8b82/pkg/start/start.go#L36

@jandubois Are there any objections to changing this in this PR?

jandubois:
> Due to this timeout. https://github.com/lima-vm/lima/blob/463ed828cb5fa0a2851a54f15941f85c189d8b82/pkg/start/start.go#L36
>
> @jandubois Are there any objections to changing this in this PR?

I would like to first see some evidence that a longer timeout "fixes" the issue.

I have a suspicion that something is just wedged, and it will never get ready until it is killed and restarted. In that case increasing the timeout will just mean it takes even longer before CI eventually fails.

So maybe use a much longer timeout in a branch, and re-run the tests a bunch of times, to see if they will then always pass.

Only then can we meaningfully discuss if we want to extend the timeout, and if it should again be gates by running inside GitHub actions.

SmartManoj:
> it will never get ready until it is killed and restarted.

first two attempts failed with `msg="[hostagent] QEMU did not exit in 3m0s, forcibly killing QEMU"`
third one with `msg="error starting docker: cannot restart, VM not previously started"`

here I didn't kill, stop, or delete it.

[Action link](https://github.com/OpenDevin/OpenDevin/actions/runs/9983889075/job/27592211025)

SmartManoj:
Use the `CUSTOM_TIMEOUT` environment variable to configure the timeout for resolving IP addresses.

* Import the `strconv` package.
* Add a new variable `customTimeout` to read the `CUSTOM_TIMEOUT` environment variable.
* Parse the `CUSTOM_TIMEOUT` value and set the `timeout` variable accordingly.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/lima-vm/lima/pull/2455?shareId=438e1b28-7390-495a-8f35-eceb744637cc).

pkg/networks/usernet/client.go
```diff
@@ -74,11 +76,18 @@ func (c *Client) ResolveAndForwardSSH(ipAddr string, sshPort int) error {
...
+		if parsedTimeout, err := strconv.Atoi(customTimeout); err == nil {
+			timeout = time.Duration(parsedTimeout) * time.Minute
+		}
+	}
+	timeoutChan := time.After(timeout)
 	ticker := time.NewTicker(500 * time.Millisecond)
 	for {
 		select {
-		case <-timeout:
+		case <-timeoutChan:
```
alexandear, line 90: Maybe it's better to control the timeout with the [context](https://pkg.go.dev/context#WithTimeout)?

```suggestion
		case <-ctx.Done():
```

https://go.dev/blog/context

pkg/networks/usernet/client.go
```diff
@@ -74,11 +76,18 @@ func (c *Client) ResolveAndForwardSSH(ipAddr string, sshPort int) error {
...
 
 func (c *Client) ResolveIPAddress(ctx context.Context, vmMacAddr string) (string, error) {
-	timeout := time.After(2 * time.Minute)
+	timeout := 2 * time.Minute
+	customTimeout := os.Getenv("CUSTOM_TIMEOUT")
+	if customTimeout != "" {
+		if parsedTimeout, err := strconv.Atoi(customTimeout); err == nil {
+			timeout = time.Duration(parsedTimeout) * time.Minute
+		}
+	}
```
alexandear, line 85: We can introduce the field `ResolveIPAddressTimeout` with the type `time.Duration` in the `Client` struct.

SmartManoj: Initialization will be done in this function again?

jandubois:
@lima-vm/reviewers Where are we with this PR? I think we can merge it now, but I think the environment variable should be documented "somewhere", I just can't figure out where. Any suggestions?

SmartManoj:
https://lima-vm.io/docs/usage/#customization
Here?

jandubois:
> https://lima-vm.io/docs/usage/#customization
> Here?

I think that page is supposed to be simple and have just the high-level examples.

Maybe we need an extra page under the "Configuration Guide"?

SmartManoj:
> Maybe we need an extra page under the "Configuration Guide"?

With the name "Environment Variables", documenting $LIMA_INSTANCE too?

pkg/networks/usernet/client.go
```diff
@@ -74,11 +76,19 @@ func (c *Client) ResolveAndForwardSSH(ipAddr string, sshPort int) error {
 }
 
 func (c *Client) ResolveIPAddress(ctx context.Context, vmMacAddr string) (string, error) {
-	timeout := time.After(2 * time.Minute)
+	ResolveIPAddressTimeout := 2 * time.Minute
+	ResolveIPAddressTimeoutEnv := os.Getenv("RESOLVE_IP_ADDRESS_TIMEOUT")
```
AkihiroSuda, line 80: Is this env var recognized by other programs too?
Otherwise this should have some prefix like `LIMA_USERNET_`

SmartManoj: How about `LIMA_CUSTOM_TIMEOUT` so that it can be a generic one?

AkihiroSuda: I'd it is more confusing, as this var only affects the timeout for resolving IP addresses with usernet.

jandubois: I agree with `LIMA_USERNET_TIMEOUT` to specify the approximate purpose.

> How about LIMA_CUSTOM_TIMEOUT so that it can be a generic one?

The word `CUSTOM` is redundant; every environment variable specifies a custom setting instead of using something hardcoded into the program.

And I think we don't need one "generic" timeout. If we need another timeout override, chances are that we might need a different value.

jandubois: I see you picked `LIMA_USERNET_RESOLVE_IP_ADDRESS_TIMEOUT`, which feels a bit long, but is fine with me.

SmartManoj: As `LIMA_TIMEOUT_FACTOR` with values `*2` or `+5`

AkihiroSuda: `LIMA_TIMEOUT_FACTOR` looks good too, but if we are going to adopt it, we have to make sure that this factor applies to (almost) all the timeout values across the entire code base

AkihiroSuda:
Please squash the commits

jandubois:
> > Maybe we need an extra page under the "Configuration Guide"?
> 
> With the name "Environment Variables", documenting $LIMA_INSTANCE too?

Sounds good to me.

In that case include all the other variables too (`LIMA_SHELL`, `LIMA_WORKDIR`, `LIMACTL`), and document where/when they are used.

`LIMA_INSTANCE` and `LIMACTL` are supported by `lima` and `lima.bat`, but `LIMA_SHELL` and `LIMA_WORKDIR` only by `lima`. `LIMA_INSTANCE` is also supported by `docker.lima` and `kubectl.lima`... Please check all the places!

pkg/networks/usernet/client.go
AkihiroSuda: > LIMA_USERNET_TIMEOUT

`LIMA_USERNET_RESOLVE_IP_ADDRESS_TIMEOUT` looks better, as it is more descriptive.
The length does not really matter, as this env var should be rarely used.

jandubois: I prefer absolute values.

A factor sounds like a good idea, but then you need to know the original value.

And if you want to extend a 10s timeout to 30s, it doesn't mean you want to extend a 5min timeout to 15min. Not all timeouts scale the same way.

SmartManoj:
[Copilot Workspace](https://copilot-workspace.githubnext.com/lima-vm/lima/pull/2455?shareId=af283a80-6fc2-4072-926a-2f1f0e37d9d4) for updating docs

It touches 8 files. First 2 is enough?

Should I update the docs in this PR itself?

jandubois (approved):
Thanks, LGTM

jandubois:
> Should I update the docs in this PR itself?

No, separate PR is fine since it will document more than just the single variable in this PR.

SmartManoj:
> Please squash the commits

image

Isn't squash and merge not enabled for this repo?

jandubois:
> Isn't squash and merge not enabled for this repo?

No, it is up to the PR author to combine commits and write a commit message that covers the squashed commits correctly.

Also (but rarely) there is justification for squashing commits into just a smaller number of commits instead of a single one. This should only be done to separate different areas of functionality, and not to maintain the individual steps that lead to the final change.

Typically this means that the 2 separate commits should have been 2 separate PRs.

AkihiroSuda:
> மனோஜ்குமார் பழனிச்சாமி

Next time please consider using Latin alphabets

PR #2460, opened 2024-07-02
boot/04-persistent-data-volume.sh: fix 9p cache options

AkihiroSuda:
`mount=fscache` is printed as `mount=8f` in /proc/mounts

https://github.com/torvalds/linux/blob/v6.6/fs/9p/v9fs.h#L61

--

Split from:
- #1953


jandubois (approved):
Thanks, LGTM.

It looks like we are not validating the 9p mount settings in `lima.yaml`; do you remember why, or was this an oversight?

Also the `default.yaml` template doesn't mention the `readahead` cache option. Maybe because it isn't mentioned in the https://www.kernel.org/doc/Documentation/filesystems/9p.txt?

AkihiroSuda:
> It looks like we are not validating the 9p mount settings in `lima.yaml`; do you remember why, or was this an oversight?

An oversight.

> Also the `default.yaml` template doesn't mention the `readahead` cache option. Maybe because it isn't mentioned in the https://www.kernel.org/doc/Documentation/filesystems/9p.txt?

Yes

PR #2466, opened 2024-07-04
Add sudo when trying to read fuse.conf

### LINKED ISSUE #2387: Old versions of fuse have read-only config file

afbjorklund:
On some distributions this configuration was only readable by members in the "fuse" group, not by the general public.

Closes #2387

### DIFF
--- a/pkg/hostagent/requirements.go
+++ b/pkg/hostagent/requirements.go
@@ -111,7 +111,7 @@ A possible workaround is to run "apt-get install sshfs" in the guest.
 			description: "fuse to \"allow_other\" as user",
 			script: `#!/bin/bash
 set -eux -o pipefail
-if ! timeout 30s bash -c "until grep -q ^user_allow_other /etc/fuse*.conf; do sleep 3; done"; then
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

PR #2479, opened 2024-07-10
add fuzz test for ConvertToRaw

AdamKorcz:
Adds a fuzz test that creates a `src.img` with pseudo random data and copies it by way of `ConvertToRaw`.

Can test it with `cd pkg/nativeimgutil && go test -fuzz=FuzzConvertToRaw`

### DIFF
--- a/pkg/nativeimgutil/fuzz_test.go
+++ b/pkg/nativeimgutil/fuzz_test.go
@@ -1,0 +1,19 @@
+package nativeimgutil
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func FuzzConvertToRaw(f *testing.F) {
+	f.Fuzz(func(t *testing.T, imgData []byte, withBacking bool, size int64) {
+		srcPath := filepath.Join(t.TempDir(), "src.img")
+		destPath := filepath.Join(t.TempDir(), "dest.img")
+		err := os.WriteFile(srcPath, imgData, 0o600)
+		if err != nil {
+			return
+		}
+		_ = ConvertToRaw(srcPath, destPath, &size, withBacking)
+	})
+}


pkg/nativeimgutil/fuzz_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"path/filepath"
+	"testing"
+)
+
+func FuzzConvertToRaw(f *testing.F) {
+	f.Fuzz(func(t *testing.T, imgData []byte, withBacking, specifySize bool) {
+		if len(imgData) < 3 {
+			return
+		}
+		imgDir, err := os.MkdirTemp("", "fuzz-images-")
```
afbjorklund, line 14: I think you want to use `testing.TempDir`, or defer the cleanup

AdamKorcz: https://github.com/lima-vm/lima/pull/2479/commits/e0723fd8a7c6660a89dc72e49fb2c5436a24c247

pkg/nativeimgutil/fuzz_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func FuzzConvertToRaw(f *testing.F) {
+	f.Fuzz(func(t *testing.T, imgData []byte, withBacking, specifySize bool) {
+		if len(imgData) < 3 {
+			return
+		}
```
jandubois, line 13: Why is this condition here? Shouldn't the fuzzer be allowed test arbitrary data sizes?

pkg/nativeimgutil/fuzz_test.go
```diff
@@ -0,0 +1,33 @@
...
+			f, err := os.Stat(srcPath)
+			if err != nil {
+				return
+			}
+			fileSize := f.Size()
+			//nolint:errcheck // The test doesn't check the return value
+			ConvertToRaw(srcPath, destPath, &fileSize, withBacking)
+		} else {
+			//nolint:errcheck // The test doesn't check the return value
+			ConvertToRaw(srcPath, destPath, nil, withBacking)
```
jandubois, line 30: If you edit the PR anyways, please use `_` instead of the `//nolint` directives.

pkg/nativeimgutil/fuzz_test.go
```diff
@@ -0,0 +1,28 @@
...
+		err := os.WriteFile(srcPath, imgData, 0o600)
+		if err != nil {
+			return
+		}
+		if specifySize {
+			f, err := os.Stat(srcPath)
+			if err != nil {
+				return
+			}
+			fileSize := f.Size()
```
jandubois, line 22: Just for my understanding, what is the reason to always pass the current filesize and not let the fuzzer modify it as well? Just to reduce the combinatorial space?

But it means it will never hit the error handling for a size too small, or trigger errors when the size is ridiculously large.

Also, won't `fileSize` always be the same as `len(imgData)`?

jandubois (approved):

PR #2483, opened 2024-07-12
add fuzz test for IsISO9660

AdamKorcz:
Adds a fuzz test that creates a file with random content and invokes `IsISO9660` with its path.


jandubois (approved):
Thanks, LGTM

PR #2484, opened 2024-07-15
add fuzz test for procnettcp parsing

AdamKorcz:
Adds a fuzz test for `pkg/guestagent/procnettcp.Parse`.


jandubois (approved):
Thanks, LGTM

PR #2488, opened 2024-07-17
Make sure /proc/sys/fs/binfmt_misc has been mounted (on Alpine)

jandubois:
Alpine cloud images don't mount the binfmt_misc filesystem by default. This is done by starting the `procfs` service.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/05-rosetta-volume.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/05-rosetta-volume.sh
@@ -7,6 +7,7 @@ if [ "$LIMA_CIDATA_ROSETTA_ENABLED" != "true" ]; then
 fi
 
 if [ -f /etc/alpine-release ]; then
+	rc-service procfs start --ifnotstarted
 	rc-service qemu-binfmt stop --ifstarted
 fi
 


AkihiroSuda (approved):
Thanks

afbjorklund:
Most likely more related to OpenRC than to the Alpine distro?

jandubois:
> Most likely more related to OpenRC than to the Alpine distro?

I don't understand what you mean.

The issue was that the Alpine cloud images don't load `procfs` automatically, so they have `/proc/sys/fs/binfmt_misc` but not `/proc/sys/fs/binfmt_misc/register` because the filesystem isn't mounted.

For our Alpine-Lima ISO image we explicitly add `procfs` to the `default` boot level with `rd_add procfs default`: [genapkovl-lima.sh](https://github.com/rancher-sandbox/alpine-lima/blob/41ead0beb9519d4a92483302f8d2b88a01b60d75/genapkovl-lima.sh#L175-L179), but we use upstream cloud images, so there we can't control the services loaded by default.

afbjorklund:
Wouldn't it be needed on other distros too?

jandubois:
> Wouldn't it be needed on other distros too?

I see what you mean now. Potentially yes.

I just checked `template://experimental/gentoo` and it already has `procfs` at the `boot` runlevel.

A quick search shows "Artix" and "Void" as distros that could potentially use Open-RC (but not by default). We don't have templates for them. Not sure if they publish cloud images that would work with Lima.

So not sure if this is worth worrying about.

What should the code look like?

```bash
if command -v rc-service >/dev/null 2>&1;; then
	rc-service procfs start --ifnotstarted
	rc-service qemu-binfmt stop --ifstarted
fi
```

afbjorklund:
We can address it, if/when it happens...

Just that other code had openrc checks

PR #2489, opened 2024-07-17
vz: configure Rosetta’s ahead of time (AOT) caching options using an abstract socket

norio-nomura:
This change allows starting the `/mnt/lima-rosetta/rosttad daemon` in the VM. Without this change, it cannot be started as shown below:
```console
$ sudo /mnt/lima-rosetta/rosettad daemon
rosetta error: Failed bind socket: 98
Trace/breakpoint trap
```

Since `rosettad` needs to run in the same namespace as `rosetta` and must be executed directly from `/mnt/lima-rosetta/rosettad` (copies of `rosettad` do not work), it is difficult to use during container image builds or similar processes.

In a container, `rosettad` can be run as follows:
```console
$ docker run -it --rm --platform linux/amd64 --name test-rosettad -v/mnt/lima-rosetta:/mnt/lima-rosetta ubuntu
root@45f887cb0eaa:/# ls -la ~/.cache/rosettad
/usr/bin/ls: cannot access '/root/.cache/rosettad': No such file or directory
root@45f887cb0eaa:/# /mnt/lima-rosetta/rosettad daemon&
[1] 10
root@45f887cb0eaa:/# ls -la ~/.cache/rosettad
total 2700
drwxr-xr-x 2 root root    4096 Jul 17 04:16 .
drwxr-xr-x 3 root root    4096 Jul 17 04:16 ..
-rwxr-xr-x 1 root root  186600 Jul 17 04:16 169a47096736cdeb5714fa040c30ef80426e439f02a3f4e4137733cbaae1fec5.aotcache
-rwxr-xr-x 1 root root  104848 Jul 17 04:16 1bea8094b78a3910345d80af3d182390fda07ae5788352651eb7773505dc39af.aotcache
-rwxr-xr-x 1 root root  121772 Jul 17 04:16 9fb501baf5ceec1ff4c9abb0473492ab4a66893203b932e28dfaa5c73c05e191.aotcache
-rwxr-xr-x 1 root root  244608 Jul 17 04:16 b7695da977e205398b33922d61493cc1e436cd2321bb2e54fbdc39b95fabeff7.aotcache
-rwxr-xr-x 1 root root 2092260 Jul 17 04:16 fc4c52f3910ed57a088d19ab86c671358f5e917cd4e95b21fd08e4fd922c0aa2.aotcache
```

Even if `rosettad` is not running, it does not affect the operation of `rosetta`, so the existing functionality remains unchanged.

There is also a configuration to use Unix Domain Sockets instead of Abstract Sockets. However, it appears that some preparation is required on the VM guest side before mounting, and I was unable to make it work. Therefore, I am using the configuration with Abstract Sockets.

SeeAlso: https://developer.apple.com/documentation/virtualization/running_intel_binaries_in_linux_vms_with_rosetta#4239539

### DIFF
--- a/pkg/vz/rosetta_directory_share_arm64.go
+++ b/pkg/vz/rosetta_directory_share_arm64.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 
 	"github.com/Code-Hex/vz/v3"
+	"github.com/coreos/go-semver/semver"
+	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/sirupsen/logrus"
 )
 
@@ -32,6 +34,17 @@ func createRosettaDirectoryShareConfiguration() (*vz.VirtioFileSystemDeviceConfi
 	rosettaShare, err := vz.NewLinuxRosettaDirectoryShare()
 	if err != nil {
 		return nil, fmt.Errorf("failed to create a new rosetta directory share: %w", err)
+	}
+	macOSProductVersion, err := osutil.ProductVersion()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get macOS product version: %w", err)
+	}
+	if !macOSProductVersion.LessThan(*semver.New("14.0.0")) {
+		cachingOption, err := vz.NewLinuxRosettaAbstractSocketCachingOptions("rosetta")
+		if err != nil {
+			return nil, fmt.Errorf("failed to create a new rosetta directory share caching option: %w", err)
+		}
+		rosettaShare.SetOptions(cachingOption)
 	}
 	config.SetDirectoryShare(rosettaShare)
 


jandubois (approved):
Thanks, the change looks fine to me.

I wonder how useful it is though to enable AOT compilation inside individual containers. Ideally you would want to share the rosettad cache between host and containers, but I wonder if that has potential race conditions.

Also the fact that you have to run `rosettad` inside each container makes it hard to use it with e.g. Kubernetes.

Anyways, tested successfully with Alpine and Ubuntu.

norio-nomura:
Thanks 🙏🏻 

I feel we should add tests on macOS 14 to the CI, but there are already many tests that fail randomly, which makes me hesitant.

PR #2491, opened 2024-07-18
add fuzz test for yqutil

AdamKorcz:
Adds a fuzz test for `EvaluateExpression`. Both of `EvaluateExpression` s parameters are provided by the fuzzer.


jandubois (approved):
Thanks, LGTM

PR #2493, opened 2024-07-19
limactl stop: extend the timeout for waiting for the hostAgent to stop

norio-nomura:
Currently, both the timeout for waiting for the hostAgent to stop in `limactl stop` and the timeout for `LimaQemuDriver.Stop()` calling `shutdownQEMU()` are set to 3 minutes. https://github.com/lima-vm/lima/blob/6fa914519f9412519fee64619755aa4a9b3774f9/pkg/qemu/qemu_driver.go#L204 As a result, the `limactl stop` might time out before QEMU is forcibly terminated, leading to the error:
> level=fatal msg="did not receive an event with the \"exiting\" status"

With this change, `limactl stop` should be able to receive the `"status":{"exiting":true}` event output by the hostAgent after it has forcibly terminated QEMU, preventing `limactl stop` from resulting in an error.

### DIFF
--- a/cmd/limactl/stop.go
+++ b/cmd/limactl/stop.go
@@ -75,7 +75,7 @@ func stopInstanceGracefully(inst *store.Instance) error {
 }
 
 func waitForHostAgentTermination(ctx context.Context, inst *store.Instance, begin time.Time) error {
-	ctx2, cancel := context.WithTimeout(ctx, 3*time.Minute)
+	ctx2, cancel := context.WithTimeout(ctx, 3*time.Minute+10*time.Second)
 	defer cancel()
 
 	var receivedExitingEvent bool


norio-nomura:
Based on the CI logs, it appears that the intended effect is being achieved.
https://github.com/lima-vm/lima/actions/runs/10001543825/job/27645343085?pr=2493#step:10:339
> time="2024-07-19T02:30:51Z" level=info msg="[hostagent] Shutting down QEMU with ACPI"
time="2024-07-19T02:30:51Z" level=info msg="[hostagent] Sending QMP system_powerdown command"
time="2024-07-19T02:33:51Z" level=warning msg="[hostagent] QEMU did not exit in 3m0s, forcibly killing QEMU"
time="2024-07-19T02:33:51Z" level=info msg="[hostagent] QEMU has exited, after killing forcibly" error="signal: killed"

jandubois (approved):
Thanks, LGTM

norio-nomura:
Thanks! 🙏🏻

PR #2498, opened 2024-07-22
limayaml: add a `param` field for defining variables used to customize scripts and other elements within `lima.yaml`.

norio-nomura:
The defined variables can be referenced within `lima.yaml` as `{{.Param.Key}}`.
The fields where this can be utilized are as follows:
- `provision[%d].script`
- `probes[%d].script`
- `copyToHost[%d].{guest,host}`
- `portForwards[%d].{guestSocket,hostSocket}`

It also changes:
- `limayaml`: add a check to verify whether the variables defined in `param` are actually used.

Following change has been merged as a separated PR: https://github.com/lima-vm/lima/pull/2501
- ~~`limactl start`: fix the issue where using `--set` would overwrite the existing instance's `lima.yaml` without validation.~~  

Following changes will be opened as a separated PR: https://github.com/lima-vm/lima/pull/2515
- ~~`docker.yaml`: add `.param.ContainerdImageStore`  
  By passing the `--set .param.ContainerdImageStore=true` option to `limactl {create,start,edit}`, the `.features."containerd-snapshotter"` option will be enabled in `docker/daemon.json` inside the VM.~~
- ~~`docker.yaml`: add `.param.Rootful`  
  By passing the `--set .param.Rootful=true` option to `limactl {create,start,edit}`, Docker inside the VM will run in rootful mode.~~
- ~~`docker-rootful.yaml`: make everything common except for setting `.param.Rootful=true` in `docker.yaml`.~~  

 
It might be better to split this into multiple parts, but for now, I’ll open it as a draft.

Thanks,

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -193,7 +193,7 @@ containerd:
 
 # Provisioning scripts need to be idempotent because they might be called
 # multiple times, e.g. when the host VM is being restarted.
-# The scripts can use the following template variables: {{.Home}}, {{.UID}}, and {{.User}}
+# The scripts can use the following template variables: {{.Home}}, {{.UID}}, {{.User}}, and {{.Param.Key}}
 # 🟢 Builtin default: null
 # provision:
 # # `system` is executed with root privileges
@@ -235,6 +235,7 @@ containerd:
 #   playbook: playbook.yaml
 
 # Probe scripts to check readiness.
+# The scripts can use the following template variables: {{.Home}}, {{.UID}}, {{.User}}, and {{.Param.Key}}
 # 🟢 Builtin default: null
 # probes:
 # # Only `readiness` probes are supported right now.
@@ -377,8 +378,8 @@ networks:
 # - guestSocket: "/run/user/{{.UID}}/my.sock"
 #   hostSocket: mysocket
 # # default: reverse: false
-# # "guestSocket" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
-# # "hostSocket" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
+# # "guestSocket" can include these template variables: {{.Home}}, {{.UID}}, {{.User}}, and {{.Param.Key}}.
+# # "hostSocket" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, {{.User}}, and {{.Param.Key}}.
 # # "reverse" can only be used for unix sockets right now, not for tcp sockets.
 # # Put sockets into "{{.Dir}}/sock" to avoid collision with Lima internal sockets!
 # # Sockets can also be forwarded to ports and vice versa, but not to/from a range of ports.
@@ -397,8 +398,8 @@ networks:
 # - guest: "/etc/myconfig.cfg"
 #   host: "{{.Dir}}/copied-from-guest/myconfig"
 # # deleteOnStop: false
-# # "guest" can include these template variables: {{.Home}}, {{.UID}}, and {{.User}}.
-# # "host" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, and {{.User}}.
+# # "guest" can include these template variables: {{.Home}}, {{.UID}}, {{.User}}, and {{.Param.Key}}.
+# # "host" can include {{.Home}}, {{.Dir}}, {{.Name}}, {{.UID}}, {{.User}}, and {{.Param.Key}}.
 # # "deleteOnStop" will delete the file from the host when the instance is stopped.
 
 # Message. Information to be shown to the user, given as a Go template for the instance.
@@ -417,6 +418,11 @@ networks:
 # 🟢 Builtin default: null
 # env:
 #   KEY: value
+
+# Defines variables used for customizing the functionality.
+# These variables can be referenced as {{.Param.Key}} in lima.yaml.
+# param:
+#   Key: value
 
 # Lima will override the proxy environment variables with values from the current process
 # environment (the environment in effect when you run `limactl start`). It will automatically
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
@@ -252,7 +253,7 @@ func Validate(y *LimaYAML, warn bool) error {
 		}
 		if rule.GuestSocket != "" {
 			if !path.IsAbs(rule.GuestSocket) {
-				return fmt.Errorf("field `%s.guestSocket` must be an absolute path", field)
+				return fmt.Errorf("field `%s.guestSocket` must be an absolute path, but is %q", field, rule.GuestSocket)
 			}
 			if rule.HostSocket == "" && rule.HostPortRange[1]-rule.HostPortRange[0] > 0 {
 				return fmt.Errorf("field `%s.guestSocket` can only be mapped to a single port or socket. not a range", field)
@@ -287,7 +288,7 @@ func Validate(y *LimaYAML, warn bool) error {
 		field := fmt.Sprintf("CopyToHost[%d]", i)
 		if rule.GuestFile != "" {
 			if !path.IsAbs(rule.GuestFile) {
-				return fmt.Errorf("field `%s.guest` must be an absolute path", field)
+				return fmt.Errorf("field `%s.guest` must be an absolute path, but is %q", field, rule.GuestFile)
 			}
 		}
 		if rule.HostFile != "" {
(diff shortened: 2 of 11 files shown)


norio-nomura:
I have separated the fix for `limactl start --set` into https://github.com/lima-vm/lima/pull/2501.

norio-nomura:
can not reproduce panic in Unit Test on local machine
https://github.com/lima-vm/lima/actions/runs/10050557033/job/27778685157?pr=2498#step:5:2584
> panic: test timed out after 10m0s
...
FAIL	github.com/lima-vm/lima/pkg/guestagent/kubernetesservice	600.059s

examples/docker-rootful.yaml
AkihiroSuda, line 1: "This template requires Lima v0.20.0 or later" header has to be updated

examples/default.yaml
```diff
@@ -418,6 +418,11 @@ networks:
 # env:
 #   KEY: value
 
+# Defines variables used for customizing the functionality.
+# These variables can be referenced as {{.Param.Key}} in lima.yaml.
+# param:
```
AkihiroSuda, line 423: bike shedding: param vs params

examples/docker-rootful.yaml
norio-nomura: I am not sure what the next version number will be. If this change is included, would it be 0.23.0?

examples/default.yaml
norio-nomura: I was also unsure which to choose, but I decided on `param` to mirror the usage of `env` and because it results in a shorter reference notation.

examples/docker-rootful.yaml
AkihiroSuda: Planned to be v1.0

norio-nomura: updated

norio-nomura:
I also want to include the following in `lima.yaml`:
```yaml
mounts:
- location: "~/.cache/ubuntu/{{.Name}}/var/cache/apt"
  mountPoint: "/var/cache/apt"
  writable: true
- location: "~/.cache/ubuntu/{{.Name}}/var/lib/apt"
  mountPoint: "/var/lib/apt"
  writable: true
```
Should this be a separate PR?

jandubois:
This PR has 2 warnings from `yamllint`:

```console
$ yamllint examples/docker.yaml
examples/docker.yaml
  58:14     warning  too few spaces before comment  (comments)

$ yamllint examples/docker-rootful.yaml
examples/docker-rootful.yaml
  58:14     warning  too few spaces before comment  (comments)
```

You can fix by adding another space before the comments, like

```diff
-- mode: user # configure docker under non-root user
+- mode: user  # configure docker under non-root user
```

Personally I don't really agree with the requirement for 2 spaces between content and comments, and would change `.yamllint` config instead:

```diff
--- .yamllint
+++ .yamllint
@@ -7,6 +7,8 @@ rules:
     indent-sequences: false
   truthy:
     allowed-values: ['true', 'false', 'on', 'off']
+  comments:
+    min-spaces-from-content: 1
   comments-indentation: disable
   document-start: disable
   line-length: disable
```

norio-nomura:
updated and rebased

examples/default.yaml
jandubois: >  I decided on `param` to mirror the usage of `env` and because it results in a shorter reference notation.

I'm fine with `param`, but if you want to go for the shortest notation, then there is also `arg`, with the same length as `env`. There is precedence for using it in `Dockerfile`. But for reasons I can't explain I find `param` more fitting.

examples/docker-rootful.yaml
```diff
@@ -44,16 +44,64 @@ provision:
...
-      [Socket]
-      SocketUser={{.User}}
+      cat <<EOF >"$override_conf"
+    [Socket]
+    SocketUser={{.User}}
     EOF
     fi
     export DEBIAN_FRONTEND=noninteractive
     curl -fsSL https://get.docker.com | sh
+- mode: user  # configure docker under non-root user
```
jandubois, line 58: I find the provisioning scripts in `examples/docker-rootful.yaml` and `examples/docker.yaml` to be a bit dense and would like to rework them to be a bit longer, but maybe easier to read/understand/maintain.

I understand they serve as an example on how the `Param` feature is going to be used, but I would still prefer to review them in a separate PR after the feature itself has been merged.

jandubois (commented):
Not sure if I get the time to review this today, but if I do, then I will skip the `examples/docker*` changes; I think they should go into a separate PR.

examples/docker-rootful.yaml
norio-nomura: Removed changes to `docker{,-rootful}.yaml` from this PR.

norio-nomura:
Updated

norio-nomura:
opened:
- https://github.com/lima-vm/lima/pull/2515

AkihiroSuda (dismissed):
Thanks

AkihiroSuda:
@lima-vm/maintainers 
Is this safe to merge in v0.23?

afbjorklund:
I am not a big fan of mixing in upper case, with the rest of the yaml that uses lower case...

Even if it means having to "translate" a param: key: value into a {{Param.Key}} so that they can each look their best

norio-nomura:
> Even if it means having to "translate" a param: key: value into a {{Param.Key}} so that they can each look their best

To reference `param: {key: value}`, use `{{.Param.key}}`. There is no functionality to convert `key` to `Key`.

jandubois:
> > Even if it means having to "translate" a param: key: value into a {{Param.Key}} so that they can each look their best
> 
> To reference `param: {key: value}`, use `{{.Param.key}}`. There is no functionality to convert `key` to `Key`.

I believe there is no way to reference non-exported fields from Go templates, so `{{.Param.key}}` cannot work.

I think what @afbjorklund meant was that you have this in your `lima.yaml`:

```yaml
param:
   containerdImageStore: false
   rootful: false
```

And we would automatically capitalize the first letter when putting it into the template data struct, so you could reference it as `{{.Param.Rootful}}`.

Now you could argue that we are already transforming `param` → `.Param`, so doing the same for the keys is consistent with that.

Personally I don't like this, and I think `param` and `rootful` are semantically different entities, and don't need to be treated the same. I would prefer that the parameters are spelled consistently across the whole `lima.yaml` file and not magically change their first letter to uppercase.

I would maybe go a completely different way, and use the same conventions I would use for environment variables for params:

```yaml
param:
   CONTAINERD_IMAGE_STORE: false
   ROOTFUL: false
```

And then reference them as `{{.Param.ROOTFUL}}`. While I don't think we should enforce this, it may be something we want to use in the bundled templates.

What do other people think?

Assuming we are not implementing @afbjorklund's suggestion to auto-capitalize param names (I would vote against it), we should catch them in validation.

norio-nomura:
> I believe there is no way to reference non-exported fields from Go templates, so {{.Param.key}} cannot work.


`{{.Param.key}}` works because `Param` is a [`map[string]string`](https://github.com/lima-vm/lima/pull/2498/files#diff-7cd9c1627051261f447947eaa346dd2480920529599fee8d2db3bd35f960239bR37).

In my local tests, #2515 actually works when I change it to:
```yaml
param:
  ContainerdImageStore: true
  rootful: true
```
and use `{{.Param.rootful}}` in combination.

> Now you could argue that we are already transforming `param` → `.Param`, so doing the same for the keys is consistent with that.

The reason it is `.Param` for the reference is not because `text/template` is converting it automatically, but because it is being passed to `text/template` as `Param` in [this line of the code](https://github.com/lima-vm/lima/pull/2498/files#diff-1e17bbd436fd831ebcaac290c75166c21aefc990f1be6125ddcea9832a1dd0b0R764).

jandubois:
> `{{.Param.key}}` works because `Param` is a [`map[string]string`]

Thanks, didn't think of that. In that case I'm even more against automatically capitalizing the first letter of params.

> The reason it is `.Param` for the reference is not because `text/template` is converting it automatically, but because it is being passed to `text/template` as `Param`

I understand. And the same way we could capitalize the first letter of `k` before we assign the values in the `param` list with `param[k] = v`.

I just think we shouldn't.

It is also possible that I misunderstand what @afbjorklund is asking for.

Just to be clear: I think the code should be left the way it is right now.

pkg/limayaml/validate.go
```diff
@@ -385,6 +386,46 @@ func validateNetwork(y *LimaYAML) error {
 	return nil
 }
 
+// ValidateParamIsUsed checks if the keys in the `param` field are used in any script, probe, copyToHost, or portForward.
+// It should be called before the `y` parameter is passed to FillDefault() that execute template.
+func ValidateParamIsUsed(y *LimaYAML) error {
+	for key := range y.Param {
+		re, err := regexp.Compile(`{{[^}]*\.Param\.` + key + `[^}]*}}`)
```
jandubois, line 393: This expression doesn't match when the parameter is only used in a template expression like this:

```go
"{{if eq .Param.Rootful \"true\"}}/var/run{{else}}/run/user/{{.UID}}{{end}}/docker.sock"
```

If would suggest to loosen up the regex, even if it could potentially get some false positives. That is still better than missing references and throwing an error:

```suggestion
		re, err := regexp.Compile(`\.Param\.` + key + `\b`)
```

Also add a test case for this scenario.

jandubois (changes requested):
Thanks, LGTM, except it needs a comment to show that probes now do template expansion, and the regex for checking if a param is used needs to loosen up.

jandubois:
> @lima-vm/maintainers Is this safe to merge in v0.23?

I think it is, and would like to see it included!

pkg/limayaml/validate.go
norio-nomura: I added a test using `"{{if eq .Param.rootful \"true\"}}/var/run{{else}}/run/user/{{.UID}}{{end}}/docker.sock"` to `TestValidateParamIsUsed`. It detected the pattern without needing to change the pattern passed to regexp, so that part remains unchanged.

examples/default.yaml
```diff
@@ -193,7 +193,7 @@ containerd:
 
 # Provisioning scripts need to be idempotent because they might be called
 # multiple times, e.g. when the host VM is being restarted.
-# The scripts can use the following template variables: {{.Home}}, {{.UID}}, and {{.User}}
+# The scripts can use the following template variables: {{.Home}}, {{.UID}}, {{.User}}, and {{.Param.Key}}
```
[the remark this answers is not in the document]
norio-nomura: added

jandubois (approved):
Thanks, looks good, but can you please squash commits?

norio-nomura:
Thanks! 🙏🏻

afbjorklund:
> I understand. And the same way we could capitalize the first letter of `k` before we assign the values in the `param` list with `param[k] = v`.

That was indeed my suggestion, and it was purely aesthetical (there was something "wrong" to me, with Key = value)

Now the params will work like environment variables, and I suppose there is some Lowest Common Denominator in that.

PR #2500, opened 2024-07-22
add fuzzer for Inspect

AdamKorcz:
Adds a fuzz test for `pkg/store.Inspect`. The fuzzer creates a file with the yaml in a temporary directory and then invokes `Inspect` with the name of that directory. It also creates a file from which Lima reads the version.

### DIFF
--- a/pkg/store/fuzz_test.go
+++ b/pkg/store/fuzz_test.go
@@ -4,6 +4,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/lima-vm/lima/pkg/store/filenames"
 )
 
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

I do wonder though if this isn't largely redundant with the `LoadYamlByFilePath` fuzzer: since you are feeding random data as `lima.yaml` you will always end up with failed YAML unmarshaling, and not fuzz any of the remaining `Inspect` code.

It feels like it would be useful to have a function that generates a valid (but otherwise random) `lima.yaml` file, to go deeper into `FillDefault` and `Inspect`.

Right now we are mostly fuzzing the YAML unmarshaling code, and not Lima.

PR #2504, opened 2024-07-23
Refactor stop/delete logic into pkg/instance

jandubois:
I'll move `pkg/start/*` into `pkg/instance` in another PR.


AkihiroSuda (approved):
Thanks

PR #2515, opened 2024-07-26
`docker{,-rootful}.yaml`: Use param in docker templates

norio-nomura:
based on #2498

- `docker.yaml`: add `.param.ContainerdImageStore`  
  By passing the `--set .param.ContainerdImageStore=true` option to `limactl {create,start,edit}`, the `.features."containerd-snapshotter"` option will be enabled in `docker/daemon.json` inside the VM.
- `docker.yaml`: add `.param.Rootful`  
  By passing the `--set .param.Rootful=true` option to `limactl {create,start,edit}`, Docker inside the VM will run in rootful mode.
- `docker-rootful.yaml`: make everything common except for setting `.param.Rootful=true` in `docker.yaml`.

Thanks,


norio-nomura:
Once #2498 is merged, I will rebase onto the main branch and then mark this as ready for review.

examples/docker-rootful.yaml
```diff
@@ -6,7 +6,7 @@
 # $ export DOCKER_HOST=$(limactl list docker-rootful --format 'unix://{{.Dir}}/sock/docker.sock')
 # $ docker ...
 
-# This template requires Lima v0.20.0 or later
+# This template requires Lima v1.0.0 or later
```
AkihiroSuda, line 9:
```suggestion
# This template requires Lima v0.23.0 or later
```

AkihiroSuda:
Is this still draft?

norio-nomura:
Could it be that the change to Ready for review will not be notified? I switched to Ready for review.

AkihiroSuda (approved):
Thanks

examples/docker-rootful.yaml
```diff
@@ -44,16 +44,64 @@ provision:
...
     EOF
     fi
     export DEBIAN_FRONTEND=noninteractive
     curl -fsSL https://get.docker.com | sh
+- mode: user  # configure docker under non-root user
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    command -v jq &>/dev/null || sudo apt-get install -y jq
+    readonly rootless_installed=$(systemctl --user list-unit-files docker.service &>/dev/null && echo true || echo false)
```
jandubois, line 63: I think our templates exist not only to implement the required functionality, but also serves as documentation and tutorials for others to adapt for their own needs.

Therefore I think we should not aim for the most compact representation, but instead write them in a simplistic style aimed at beginners, who may not be very experienced in shell scripting.

This means using long-form options whenever possible[^1], and using simplified `if … then … else … fi` control flow instead of trying to squeeze as much functionality as possible into the least amount of lines.

[^1]: I know we currently don't do that, but I would be in favour of expanding options on already existing code as well. E.g. `curl -fsSL` is non-obvious to many people.

So I would prefer something like this:

```suggestion
    set -o errexit -o nounset -o xtrace -o pipefail

    if ! command -v jq &>/dev/null; then
        sudo apt-get install --assume-yes jq
    fi
    if systemctl --user list-unit-files docker.service &>/dev/null; then
        readonly rootless_installed=true
    else
        readonly rootless_installed=true
    fi
```

examples/docker-rootful.yaml
```diff
@@ -44,16 +44,64 @@ provision:
...
+
+    if [ "{{.Param.Rootful}}" = "true" ]; then
+      readonly config_dir="/etc/docker"
+      readonly systemctl="sudo systemctl"
+      readonly tee="sudo tee"
+
+      [ "$rootless_installed" != "true" ] || systemctl --user disable --now docker
```
jandubois, line 70: To simplify the code I would especially try to avoid using boolean shortcut operator. Users unfamiliar with De Morgan's law find turning `if CONDITION; then ACTION; fi` into `! CONDITION || ACTION` unnatural, and it may be unobvious why they shouldn't write `CONDITION && ACTION` (it will cause the script to fail because of `set -o errexit`).

```suggestion
      if [ "$rootless_installed" == "true" ]; then
          systemctl --user disable --now docker
      fi
```

This feedback applies to other uses of `||` as well, but I won't repeat it on each instance.

examples/docker-rootful.yaml
```diff
@@ -44,16 +44,64 @@ provision:
...
     export DEBIAN_FRONTEND=noninteractive
     curl -fsSL https://get.docker.com | sh
+- mode: user  # configure docker under non-root user
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    command -v jq &>/dev/null || sudo apt-get install -y jq
+    readonly rootless_installed=$(systemctl --user list-unit-files docker.service &>/dev/null && echo true || echo false)
+
+    if [ "{{.Param.Rootful}}" = "true" ]; then
```
jandubois, line 65: This may be a personal preference, but I think using all uppercase letters for params makes them stand out more to show that this setting isn't purely internal but can be modified by the template user (similar to environment variables).

@lima-vm/reviewers What do you think?

```suggestion
    if [ "{{.Param.ROOTFUL}}" = "true" ]; then
```

examples/docker-rootful.yaml
```diff
@@ -44,16 +44,64 @@ provision:
     #!/bin/bash
     set -eux -o pipefail
     command -v docker >/dev/null 2>&1 && exit 0
-    if [ ! -e /etc/systemd/system/docker.socket.d/override.conf ]; then
-      mkdir -p /etc/systemd/system/docker.socket.d
+    readonly override_conf=/etc/systemd/system/docker.socket.d/override.conf
+    if [ ! -e "$override_conf" ]; then
+      mkdir -p $(dirname "$override_conf")
```
jandubois, line 49: You didn't have to quote `$override_conf` because it is a fixed readonly constant that doesn't include whitespace. But if you do, then you should also quote the result of the `dirname` call:

```suggestion
      mkdir -p "$(dirname "$override_conf")"
```

examples/docker-rootful.yaml
```diff
@@ -44,16 +44,64 @@ provision:
...
+      readonly systemctl="sudo systemctl"
+      readonly tee="sudo tee"
+
+      [ "$rootless_installed" != "true" ] || systemctl --user disable --now docker
+      docker context use default
+
+    else
+      readonly config_dir="$HOME/.config/docker"
+      readonly systemctl="systemctl --user"
+      readonly tee="tee"
```
jandubois, line 76: Instead of using variables I would create shell wrapper functions:

```shell
systemctl() {
    if [ "{{.Param.ROOTFUL}}" = "true" ]; then
        sudo command systemctl "$@"
    else
        command systemctl --user "$@"
    fi
}
```

and then call `systemctl` normally in the rest of the script. Same for `tee`.

examples/docker-rootful.yaml
```diff
@@ -44,16 +44,64 @@ provision:
...
+        $systemctl start dbus
+        [ ! -S /var/run/docker.sock ] || sudo rm /var/run/docker.sock
+        dockerd-rootless-setuptool.sh install
+      fi
+      docker context use rootless
+    fi
+    $systemctl enable --now docker
+
+    readonly config="$config_dir/daemon.json"
+    needs_restart=
```
jandubois, line 90: You use `true` and `false` for the `param.*` booleans, so I would suggest to do the same for other variables.

```suggestion
    needs_restart=false
```

examples/docker-rootful.yaml
```diff
@@ -44,16 +44,64 @@ provision:
...
+    function set_docker_daemon_json() {
+      function cat_config() { test -s "$config" && cat "$config" || echo "{}" ; }
```
jandubois, line 92: I don't like nested function definitions. They look like you have dynamically scoped naming, but `bash` doesn't do that; the function is still declared at the global scope.

examples/docker-rootful.yaml
```diff
@@ -62,8 +110,15 @@ probes:
       echo >&2 "docker is not installed yet"
       exit 1
     fi
-    if ! timeout 30s bash -c "until pgrep dockerd; do sleep 3; done"; then
-      echo >&2 "dockerd is not running"
+    if [ "{{.Param.Rootful}}" = "true" ]; then
+      target=dockerd
+      target_description="dockerd"
```
jandubois, line 115: Minor nit, but why is one value quoted and the other one isn't? Given that the `else` block needs quoting for the second value, maybe quote everything?

```suggestion
      target="dockerd"
      target_description="dockerd"
```

jandubois (changes requested):
I would like to see our provisioning scripts written in "baby-bash" to make them easier to understand for beginners.

This is obviously a personal opinion, so maybe wait for some confirmation from other @lima-vm/reviewers in case they don't agree with me. 😄

norio-nomura:
Updated to apply reviews

norio-nomura:
Updated PR description:
- `.param.ContainerdImageStore` -> `.param.CONTAINERD_IMAGE_STORE`
- `.param.Rootful` -> `.param.ROOTFUL`

afbjorklund:
Looks super-complicated to me.

AkihiroSuda:
> Looks super-complicated to me.

param.CONTAINERD_IMAGE_STORE is still acceptable?

norio-nomura:
For `param.CONTAINERD_IMAGE_STORE`, refer to [this Docker Desktop page](https://docs.docker.com/desktop/containerd/).  
Enabling this option in a Docker host started with Lima allows building multi-platform images without creating a builder instance using `docker buildx create`.  
The `param` feature was originally created to make it easy to toggle this option on and off.

norio-nomura:
> Looks super-complicated to me.

Since this option allows you to turn complicated settings on and off with a single option, it’s naturally going to seem complicated.  
If the way I've written it makes it look more complicated, I'm sorry, but I don't know how to improve it.

jandubois:
Sorry, I didn't get around to do another round of code review yet; I wrote #2520 instead to explain my general view/vision of where we should be going with this.

I would like to continue to discuss this PR to see if we can't get it into a shape where we agree that having a combined template is better than 2 separate ones. I don't know if we will get there, but I think we should try.

afbjorklund:
> If the way I've written it makes it look more complicated, I'm sorry, but I don't know how to improve it.

You don't have to be sorry, it was a hot day and it felt "too long" - I should have given it a proper review.

afbjorklund:
> param.CONTAINERD_IMAGE_STORE is still acceptable?

I still feel bad about param.ContainerdImageStore, but I was referring to the length of the new templates (doubled).

Switching from Bourne shell to Bash, also takes some getting used to. Simple things like `readonly` and `function`.

jandubois:
> I'm sorry, but I don't know how to improve it.

I just wanted to let you know that this issue is still on my radar, but I will be offline/travelling until the end of the month and won't have time to look at it again until I'm back.

PR #2533, opened 2024-08-06
Add a delay to make sure newly created file is visible inside the VM

jandubois:
The `hack/test-templates.sh` test seems to regularly fail for the `archlinux` template. I'm wondering that maybe the newly created file from the host is simply not yet visible inside the VM when `nerdctl` tries to `stat()` it before mounting it into a container.


jandubois:
This attempt did not address the root cause

PR #2535, opened 2024-08-07
Temporarily disable CI on archlinux

jandubois:


.github/workflows/test.yml
```diff
@@ -213,7 +213,7 @@ jobs:
         - alpine.yaml
         - debian.yaml
         - fedora.yaml
-        - archlinux.yaml
+        # - archlinux.yaml temporarily disabled because it breaks CI
```
AkihiroSuda, line 216: Could you link a GitHub issue?

jandubois: I've created #2538.

PR #2541, opened 2024-08-10
test.yml: use `inject-cmdline-to-template.sh` to append `no_timer_check` kernel command line option

norio-nomura:
This change aims to avoid kernel panics in the integration tests and the vmnet tests.
```console
==> /Users/runner/.lima/default/serial.log <==
[    0.015000]  setup_IO_APIC+0x2c3/0x370
[    0.015000]  ? enable_IO_APIC+0x1af/0x290
[    0.015000]  apic_intr_mode_init+0x61/0x130
[    0.015000]  x86_late_time_init+0x24/0x40
[    0.015000]  start_kernel+0x2be/0x450
[    0.015000]  x86_64_start_reservations+0x18/0x30
[    0.015000]  x86_64_start_kernel+0xbf/0x110
[    0.015000]  secondary_startup_64_no_verify+0x184/0x18b
[    0.015000]  </TASK>
[    0.015000] ---[ end Kernel panic - not syncing: IO-APIC + timer doesn't work!  Boot with apic=debug and send a report.  Then try booting with the 'noapic' option. ]---
```

The reason `no_timer_check` is used instead of `noapic`, as shown in the kernel panic log, is because I read the following page.
https://lore.kernel.org/all/18354be1-8dba-84f1-bdf5-6821a5013d78@oracle.com/T/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -184,6 +184,9 @@ jobs:
         time brew install qemu bash coreutils curl jq
     - name: "Show cache"
       run: ./hack/debug-cache.sh
+    - name: "Inject `no_timer_check` to kernel cmdline"
+      # workaround to https://github.com/lima-vm/lima/issues/84
+      run: ./hack/inject-cmdline-to-template.sh templates/default.yaml no_timer_check
     - name: "Test default.yaml"
       uses: nick-invision/retry@v3
       with:
@@ -356,6 +359,9 @@ jobs:
     - name: Unit test (pkg/networks) with socket_vmnet
       # Set -count=1 to disable cache
       run: go test -v -count=1 ./pkg/networks/...
+    - name: "Inject `no_timer_check` to kernel cmdline"
+      # workaround to https://github.com/lima-vm/lima/issues/84
+      run: ./hack/inject-cmdline-to-template.sh templates/vmnet.yaml no_timer_check
     - name: Test socket_vmnet
       uses: nick-invision/retry@v3
       with:
--- a/hack/inject-cmdline-to-template.sh
+++ b/hack/inject-cmdline-to-template.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+#
+# This script does
+# 1. detect arch from template if not provided
+# 2. extract location by parsing template using arch
+# 3. get the image location
+# 4. check the image location is supported
+# 5. build the kernel and initrd location, digest, and cmdline
+# 6. inject the kernel and initrd location, digest, and cmdline to the template
+# 7. output kernel_location, kernel_digest, cmdline, initrd_location, initrd_digest
+
+set -eu -o pipefail
+
+template="$1"
+appending_options="$2"
+# 1. detect arch from template if not provided
+arch="${3:-$(yq '.arch // ""' "${template}")}"
+arch="${arch:-$(uname -m)}"
+
+# normalize arch. amd64 -> x86_64, arm64 -> aarch64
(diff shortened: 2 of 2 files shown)


examples/default.yaml
```diff
@@ -27,9 +27,23 @@ images:
 - location: "https://cloud-images.ubuntu.com/releases/24.04/release-20240725/ubuntu-24.04-server-cloudimg-amd64.img"
   arch: "x86_64"
   digest: "sha256:d2377667ea95222330ca2287817403c85178dad397e9fed768a9b4aec79d2a7f"
+  kernel:
+    location: "https://cloud-images.ubuntu.com/releases/24.04/release-20240725/unpacked/ubuntu-24.04-server-cloudimg-amd64-vmlinuz-generic"
+    digest: "sha256:47bf2de2f093931bcb8e862b3390905038de301a9e9a8bc1b462666bb4011dcd"
+    cmdline: "root=LABEL=cloudimg-rootfs ro console=tty1 console=ttyAMA0 no_timer_check"
```
AkihiroSuda, line 33: I think this is fine for CI, however, this makes impossible to update the kernel and initrd via apt-get, so I guess we shouldn't do this outside the CI. (The panic issue doesn't seem to happen outside the CI AFAICS.)

AkihiroSuda: Ideally, Lima should be able to mount the qcow2 image and patch the grub config inside it, but its implementation cost is likely to be extremely high.

AkihiroSuda: I also wonder if the kernel cmdline could be supplied via an EFI var, fw_cfg, ACPI table, etc.

norio-nomura: > I also wonder if the kernel cmdline could be supplied via an EFI var, fw_cfg, ACPI table, etc.

I looked into the methods you suggested, but it seems quite complicated. Instead, how about adding an option in `lima.yaml` that only uses kernel-related settings during the initial boot, such as `.kernel.initial_boot_only`, `.initial_boot_kernel`, `.initial_boot_initrd`?

AkihiroSuda: Sounds overengineering.
On the CI, `no_timer_check` is needed on every boot of the instance, and it is not needed outside the CI.

AkihiroSuda: I guess CI should just use `yq` to inject the cmdline

norio-nomura: Updated to inject kernel related settings to template.

hack/inject-cmdline-to-template.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
```
AkihiroSuda, line 1: Please add comment lines to explain what this script does.
Also please squash commigs.

.github/workflows/test.yml
```diff
@@ -184,6 +184,8 @@ jobs:
         time brew install qemu bash coreutils curl jq
     - name: "Show cache"
       run: ./hack/debug-cache.sh
+    - name: "Inject `no_timer_check` to kernel cmdline"
```
AkihiroSuda, line 187: Please link a GitHub issue to explain why this is needed on the CI

hack/inject-cmdline-to-template.sh
norio-nomura: done

.github/workflows/test.yml
norio-nomura: done

AkihiroSuda (approved):
Thanks

PR #2557, opened 2024-08-20
limactl: run file manipulation subcommands under Rosetta

YorikSar:
This allows to run `completion`, `generate-doc` and `validate` subcommands when running under Rosetta.

This is useful when packaging lima or preparing configs for macOS on Apple Silicon for running on Intel arch.

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -77,8 +77,9 @@ func newApp() *cobra.Command {
 			logrus.SetLevel(logrus.DebugLevel)
 		}
 
-		if osutil.IsBeingRosettaTranslated() {
+		if osutil.IsBeingRosettaTranslated() && cmd.Parent().Name() != "completion" && cmd.Name() != "generate-doc" && cmd.Name() != "validate" {
 			// running under rosetta would provide inappropriate runtime.GOARCH info, see: https://github.com/lima-vm/lima/issues/543
+			// allow commands that are used for packaging to run under rosetta to allow cross-architecture builds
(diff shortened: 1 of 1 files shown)


cmd/limactl/main.go
```diff
@@ -77,7 +77,7 @@ func newApp() *cobra.Command {
 			logrus.SetLevel(logrus.DebugLevel)
 		}
 
-		if osutil.IsBeingRosettaTranslated() {
+		if osutil.IsBeingRosettaTranslated() && cmd.Parent().Name() != "completion" && cmd.Name() != "generate-doc" && cmd.Name() != "validate" {
```
AkihiroSuda, line 80: Maybe just check `name == "start"` ?

jandubois: I don't know; does it make sense to let the user run `create` but then fail on `start`? Same with most other commands; they are not really useful unless you can actually start an instance.

This PR is mostly just an affordance for developers/packagers to do cross-architecture builds, and not really useful for general users. So catching the arch mis-match for all other commands as soon as possible seems still desirable.

jandubois: If the change should be merged as-is, then I would like to see a comment above that explains why these 3 commands have been explicitly exempted from the Rosetta check (basically what the PR description says).

YorikSar: @AkihiroSuda I started looking through commands that are there and found some other that also don't depend on the architecture, but I agree with @jandubois that allowing only `start` to throw an error would worsen user experience.
I've added the comment as @jandubois suggested.

AkihiroSuda (approved):
Thanks

PR #2559, opened 2024-08-22
Use eBPF for identifying ports

balajiv113:
Locally worked on fedora and ubuntu wanted to make sure works for all combinations in CI


AkihiroSuda:
Why closed?

balajiv113:
Needs more work to make it work for all distros i believe.

Created the PR to get a idea on how it works in all distros. Once have things in working state will create a new PR

PR #2569, opened 2024-08-30
`limactl prune`: add `--keep-referred` option to keep objects that are referred by some instances or templates

norio-nomura:
~~Also adds `--all` option to `prune` behaves as same as previous function. (Remove all cacheDir)~~

### DIFF
--- a/cmd/limactl/prune.go
+++ b/cmd/limactl/prune.go
@@ -1,11 +1,9 @@
 package main
 
 import (
-	"crypto/sha256"
-	"fmt"
 	"os"
-	"path/filepath"
 
+	"github.com/lima-vm/lima/pkg/downloader"
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/lima-vm/lima/pkg/templatestore"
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Why ? An image cache used by an existing instance can be safely pruned

norio-nomura:
I often run `limactl factory-reset` because it's more convenient for me if the cache is preserved.

norio-nomura:
Cache entries that are not referenced by either the instance or the template are unlikely to be used again. I wanted to delete those.

AkihiroSuda:
Can we just retain the default behavior and  have a new flag like `--keep-referred=<BOOL>`?

AkihiroSuda (dismissed):
Thanks

cmd/limactl/prune.go
```diff
@@ -17,15 +22,115 @@ func newPruneCommand() *cobra.Command {
...
 	}
 	cacheDir := filepath.Join(ucd, "lima")
 	logrus.Infof("Pruning %q", cacheDir)
-	return os.RemoveAll(cacheDir)
+	if !keepReferred {
+		return os.RemoveAll(cacheDir)
+	}
+
+	// Prune downloads that are not used by any instances or templates
+	downloadDir := filepath.Join(cacheDir, "download", "by-url-sha256")
```
AkihiroSuda, line 45: This internal data structure should be handled in `pkg/downloader` rather than in the CLI

norio-nomura: Moved some codes to `pkg/downloader`.

AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

PR #2572, opened 2024-08-31
Add ls-lint to the list of linters

### LINKED ISSUE #2571: The ls-lint is missing from "make lint", but is only run in github CI
### Description

Preferrably, it should be possible to run the tests locally as well.

Especially when the command is hidden in a GitHub action step.

```yaml
    - name: Run file and directory name linter
      uses: ls-lint/action@v2.2.3
```

commit 35ccb6a299a26b6f9cab37ba7eee685e874e938b

----

https://github.com/ls-lint/action

https://ls-lint.org/

afbjorklund:
Closes #2571

Keep in sync with https://github.com/lima-vm/lima/blob/master/.github/workflows/test.yml#L21

We could consider using the same scripts/targets for both, and to mention where to get the tools

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -248,6 +248,7 @@ check-generated:
 lint: check-generated
 	golangci-lint run ./...
 	yamllint .
+	ls-lint
 	find . -name '*.sh' | xargs shellcheck
 	find . -name '*.sh' | xargs shfmt -s -d
 


AkihiroSuda (approved):
Thanks

PR #2582, opened 2024-09-02
Show progress only when logging to terminal

nirs:
Fix the terminal check so we use the right configuration when logging to file, and hide the progress in this case.

Fixes #2581

PR #2584, opened 2024-09-02
Add --log-format json option

### LINKED ISSUE #2583: Machine friendly logs
When limactl is run by another program, it may wan to show limactl logs in its
own logs. With the current text format, this show duplicate timetamps, log
level, and extra quoting in values (since the format is key="value".

Example usage from another program:

```console
% drenv start envs/vm.yaml -v
2024-09-01 21:49:16,582 INFO    [vm] Starting environment
2024-09-01 21:49:16,583 DEBUG   [main] Add executor profiles
2024-09-01 21:49:16,584 DEBUG   [cluster] Running ('/Users/nsoffer/src/lima/_output/bin/limactl', 'list', '--format', '{{.Name}}')
2024-09-01 21:49:16,603 INFO    [cluster] Starting lima cluster
2024-09-01 21:49:16,603 DEBUG   [cluster] Running ('/Users/nsoffer/src/lima/_output/bin/limactl', 'list', '--format', '{{.Name}}')
2024-09-01 21:49:16,633 DEBUG   [cluster] Running ('/Users/nsoffer/src/lima/_output/bin/limactl', 'create', '--name', 'cluster', '/var/folders/qs/_z10qgxn3gsfw8vmkr9b3s4w0000gn/T/drenv.lima.cluster.tmpe37ne8tn')
2024-09-01 21:49:16,647 DEBUG   [cluster] time="2024-09-01T21:49:16+03:00" level=info msg="Terminal is not available, proceeding without opening an editor"
2024-09-01 21:49:16,661 DEBUG   [cluster] time="2024-09-01T21:49:16+03:00" level=info msg="Attempting to download the image" arch=aarch64 digest= location="https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-arm64.img"
...

Can be fixed by adding a json log format:

```console
% _output/bin/limactl start --vm-type vz --plain --tty=false --log-format json
{"level":"info","msg":"Terminal is not available, proceeding without opening an editor","time":"2024-09-03T00:14:43+03:00"}
{"level":"info","msg":"Starting the instance \"default\" with VM driver \"vz\"","time":"2024-09-03T00:14:43+03:00"}
{"arch":"aarch64","digest":"sha256:5ecac6447be66a164626744a87a27fd4e6c6606dc683e0a233870af63df4276a","level":"info","location":"https://cloud-images.ubuntu.com/releases/24.04/release-20240821/ubuntu-24.04-server-cloudimg-arm64.img","msg":"Attempting to download the image","time":"2024-09-03T00:14:43+03:00"}
```

With this the program running limactl can:
- parse the log with standard parsers
- decide which messages to show based message properties
- change its own log level based on limactl message level
- pick which values to log from the json object

This may also help to process logs later for various purposes.

Discussed in in #2576

nirs:
When using --log-format json we use machine friendly flogs to make limactl easier to consume by other programs.

Since we don't have a json format for the progress bar, the progress is also hidden when using json log format.

Here is an example usage of the new option in drenv:
https://github.com/nirs/ramen/commit/b0001198f5c04b7eca97162f42e9dafceb8723b4

Fixes #2583

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -60,6 +60,7 @@ func newApp() *cobra.Command {
 		DisableAutoGenTag: true,
 	}
 	rootCmd.PersistentFlags().String("log-level", "", "Set the logging level [trace, debug, info, warn, error]")
+	rootCmd.PersistentFlags().String("log-format", "text", "Set the logging format [text, json]")
 	rootCmd.PersistentFlags().Bool("debug", false, "debug mode")
 	// TODO: "survey" does not support using cygwin terminal on windows yet
 	rootCmd.PersistentFlags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "Enable TUI interactions such as opening an editor. Defaults to true when stdout is a terminal. Set to false for automation.")
@@ -72,6 +73,24 @@ func newApp() *cobra.Command {
 			}
 			logrus.SetLevel(lvl)
 		}
+
+		logFormat, _ := cmd.Flags().GetString("log-format")
+		switch logFormat {
+		case "json":
+			formatter := new(logrus.JSONFormatter)
+			logrus.StandardLogger().SetFormatter(formatter)
+		case "text":
+			// logrus use text format by default.
+			if runtime.GOOS == "windows" && isatty.IsCygwinTerminal(os.Stderr.Fd()) {
+				formatter := new(logrus.TextFormatter)
+				// the default setting does not recognize cygwin on windows
+				formatter.ForceColors = true
+				logrus.StandardLogger().SetFormatter(formatter)
+			}
+		default:
+			return fmt.Errorf("unsupported log-format: %q", logFormat)
+		}
+
 		debug, _ := cmd.Flags().GetBool("debug")
 		if debug {
 			logrus.SetLevel(logrus.DebugLevel)
@@ -83,12 +102,6 @@ func newApp() *cobra.Command {
 			return errors.New("limactl is running under rosetta, please reinstall lima with native arch")
 		}
 
-		if runtime.GOOS == "windows" && isatty.IsCygwinTerminal(os.Stdout.Fd()) {
-			formatter := new(logrus.TextFormatter)
-			// the default setting does not recognize cygwin on windows
-			formatter.ForceColors = true
-			logrus.StandardLogger().SetFormatter(formatter)
-		}
 		if os.Geteuid() == 0 && cmd.Name() != "generate-doc" {
 			return errors.New("must not run as the root user")
 		}
--- a/pkg/progressbar/progressbar.go
+++ b/pkg/progressbar/progressbar.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/cheggaaa/pb/v3"
 	"github.com/mattn/go-isatty"
+	"github.com/sirupsen/logrus"
 )
 
 func New(size int64) (*pb.ProgressBar, error) {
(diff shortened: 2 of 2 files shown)


AkihiroSuda:
Merged #2582, please rebase

nirs:
One test failed - is this a flaky test?
https://github.com/lima-vm/lima/actions/runs/10683275352/job/29610982782

I cannot rerun the failing test, I hope some can rerun it...

cmd/limactl/main.go
```diff
@@ -72,6 +73,24 @@ func newApp() *cobra.Command {
 			}
 			logrus.SetLevel(lvl)
 		}
+
+		logFormat, _ := cmd.Flags().GetString("log-format")
+		switch logFormat {
+		case "json":
+			formatter := new(logrus.JSONFormatter)
+			logrus.StandardLogger().SetFormatter(formatter)
+		case "text":
```
jandubois, line 82: Can we change this to `default`. `text` sounds confusing because **all** logging output will be text.

FWIW, I'm not too happy with `default` either, but at least it makes it obvious that it is the same as not specifying a log format.

Better suggestions welcome!

jandubois (changes requested):
I think the failed test is a flake (I've triggered a re-run), but I would like to change the `text` option to something else.

cmd/limactl/main.go
nirs: "default" sounds good, waiting for for more feedback.

AkihiroSuda: I think "text" is fine, as it corresponds to logrus's terminology

AkihiroSuda: "default" should be an alias to another identifier

fatanugraha: What about `plain`?

https://docs.docker.com/reference/cli/docker/buildx/build/#progress

jandubois: > I think "text" is fine, as it corresponds to logrus's terminology

`text` allows a bunch of customization, like dropping timestamps etc, and I thought we might want to define shortcuts for some of them, and they would all use the TextFormatter.

But on second thoughts I think that might be overkill, and just supporting `text` and `json` with the default options may be all we need. Users can always transform JSON to whatever they want.

So I'm going to withdraw my objection to using `text`.

I don't think we need a `default` alias as long as we only have `text` and `json`.

jandubois (changes requested):
I no longer want to change the `--log-format=text` option, but this PR includes 2 commits that have already been merged to `master`. Please rebase.

I see failing `VZ` tests even on re-runs, but I don't think they are related to this PR, so just need more re-runs, but I'll wait until the commit history is fixed.

nirs:
> I'll wait until the commit history is fixed.

I already rebased on master, anything else to change?

cmd/limactl/main.go
nirs: I agree we don't need alias.

Some users may want to modify the "test" logs, but I does not seem nice to add many --log-xxx options. It could be better if this came from a default configuration file.

jandubois:
> I already rebased on master, anything else to change?

My apologies, I mistook "Show progress only when using text format" for a duplicate of "Show progress only when logging to terminal".

We do request that commits are squashed though, so could you please do that:

```console
$ g log --oneline -4
5cbac54 (HEAD -> log-format) Add json log format
547e45d Fix another terminal check
49f0925 Show progress only when using text format
01eb321 (upstream/master) Merge pull request #2582 from nirs/progress-to-file
```

nirs:
> We do request that commits are squashed though, so could you please do that:

Sorry I was not aware of that, pushing squashed version.

jandubois:
> Sorry I was not aware of that, pushing squashed version.

@nirs I don't see the squashed version on GitHub; are you sure you did a force-push?

nirs:
@jandubois should be ready now.

Latest push fixed a typo in a new comment.

jandubois (approved):
Thanks, LGTM.

PR #2562, opened 2024-08-28
VZ: Support Kernel Image

norio-nomura:
Similar to QEMU's kernel image support.

When using `VZLinuxBootLoader` (https://developer.apple.com/documentation/virtualization/vzlinuxbootloader), the `Virtualization.framework` does not provide EFI to the VM. As a result, the Linux kernel does not recognize a device available as `rtc0`, which means the correct time cannot be obtained during the kernel boot. Consequently, the VM's time is not corrected until `timesync.service` starts via `cloud-init`, which can cause issues like `apt-get` failures. If, for any reason, `timesync.service` fails to correct the time, `apt-get` in the provisioning script might fail, causing `limactl` to perceive the VM boot as failed.

I'm hesitant to add tests for this feature in CI because I'm unsure of the reasons why `timesync.service` might fail to correct the time.

I believe `inject-cmdline-to-template.sh` will be helpful when testing this feature.

PR #2593, opened 2024-09-05
Improve `limactl edit`

norio-nomura:
- yqutil: use yamlfmt to fix indentation of sequence
- `limactl edit`: support editing `lima.yaml` file  
  This allows `limactl edit --set "..." <limayaml>` to be used instead of `yq -i eval "..." <limayaml>` in scripts like `hack/inject-cmdline-to-template.sh`.

The changes made to `examples/ubuntu.yaml` using `limactl edit --set` in `hack/inject-cmdline-to-template.sh` are as follows:
```patch
diff --git forkSrcPrefix/examples/ubuntu.yaml forkDstPrefix/examples/ubuntu.yaml
index a7fb0b3246e71bd9f98df2e8849185ed1339c76b..9f19540104091b13e08cf255bd074831560af654 100644
--- forkSrcPrefix/examples/ubuntu.yaml
+++ forkDstPrefix/examples/ubuntu.yaml
@@ -7,13 +7,19 @@ images:
 - location: "https://cloud-images.ubuntu.com/releases/24.04/release-20240821/ubuntu-24.04-server-cloudimg-arm64.img"
   arch: "aarch64"
   digest: "sha256:5ecac6447be66a164626744a87a27fd4e6c6606dc683e0a233870af63df4276a"
+  kernel:
+    location: https://cloud-images.ubuntu.com/releases/24.04/release-20240821/unpacked/ubuntu-24.04-server-cloudimg-arm64-vmlinuz-generic
+    digest: sha256:69e66230d6acc503082f098fd072a672a804774007209110b353d25bb7007669
+    cmdline: root=LABEL=cloudimg-rootfs ro console=tty1 console=ttyAMA0 console=hvc0
+  initrd:
+    location: https://cloud-images.ubuntu.com/releases/24.04/release-20240821/unpacked/ubuntu-24.04-server-cloudimg-arm64-initrd-generic
+    digest: sha256:92acf9f2976b7a8b3b5e9e3f22211d17a6fdb3f2aed981467f91cd6dbfa16ae5
   arch: "x86_64"
   arch: "aarch64"
-
 mounts:
 - location: "~"
 - location: "/tmp/lima"
```

The issue with line breaks disappearing cannot be fixed with yamlfmt, so it still remains. However, the sequence is no longer indented, and I personally think it has become usable.

I'm considering making a separate PR to change `hack/inject-cmdline-to-template.sh` from `yq -i eval` to `limactl edit --set` after #2562, which is likely to have overlapping changes, is resolved.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,6 +23,7 @@ require (
 	github.com/foxcpp/go-mockdns v1.1.0
 	github.com/goccy/go-yaml v1.12.0
 	github.com/google/go-cmp v0.6.0
+	github.com/google/yamlfmt v0.13.0
 	github.com/lima-vm/go-qcow2reader v0.1.2
 	github.com/lima-vm/sshocker v0.3.4
 	github.com/mattn/go-isatty v0.0.20
@@ -58,6 +59,8 @@ require (
 	github.com/VividCortex/ewma v1.2.0 // indirect
 	github.com/a8m/envsubst v1.4.2 // indirect
 	github.com/alecthomas/participle/v2 v2.1.1 // indirect
+	github.com/bmatcuk/doublestar/v4 v4.6.0 // indirect
+	github.com/braydonk/yaml v0.7.0 // indirect
 	github.com/containerd/errdefs v0.1.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/digitalocean/go-libvirt v0.0.0-20220804181439-8648fbde413e // indirect
@@ -95,6 +98,7 @@ require (
 	github.com/mattn/go-runewidth v0.0.15 // indirect
 	github.com/mdlayher/socket v0.4.1 // indirect
 	github.com/mgutz/ansi v0.0.0-20170206155736-9520e82c474b // indirect
+	github.com/mitchellh/mapstructure v1.5.0 // indirect
 	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
 	github.com/modern-go/reflect2 v1.0.2 // indirect
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
@@ -104,6 +108,7 @@ require (
 	github.com/pkg/sftp v1.13.6 // indirect
 	github.com/rivo/uniseg v0.2.0 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
+	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06 // indirect
 	github.com/u-root/uio v0.0.0-20240224005618-d2acac8f3701 // indirect
 	github.com/x448/float16 v0.8.4 // indirect
 	github.com/yuin/gopher-lua v1.1.1 // indirect
--- a/pkg/yqutil/yqutil_test.go
+++ b/pkg/yqutil/yqutil_test.go
@@ -41,17 +41,18 @@ mounts:
 `
 	// Note: yq will use canonical yaml, with indented sequences
 	// Note: yq will not explicitly quote strings, when not needed
+	// Note: yamlfmt will fix indentation of sequences
 	expected := `
 # Expose host directories to the guest, the mount point might be accessible from all UIDs in the guest
 # 🟢 Builtin default: null (Mount nothing)
 # 🔵 This file: Mount the home as read-only, /tmp/lima as writable
 mounts:
-  - location: "~"
-    # Configure the mountPoint inside the guest.
-    # 🟢 Builtin default: value of location
-    mountPoint: null
-  - location: foo
-    mountPoint: bar
+- location: "~"
+  # Configure the mountPoint inside the guest.
+  # 🟢 Builtin default: value of location
+  mountPoint: null
+- location: foo
+  mountPoint: bar
 `
 	out, err := EvaluateExpression(expression, []byte(content))
 	assert.NilError(t, err)
--- a/cmd/limactl/edit.go
+++ b/cmd/limactl/edit.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 
 	"github.com/lima-vm/lima/cmd/limactl/editflags"
+	"github.com/lima-vm/lima/cmd/limactl/guessarg"
 	"github.com/lima-vm/lima/pkg/editutil"
 	"github.com/lima-vm/lima/pkg/instance"
 	"github.com/lima-vm/lima/pkg/limayaml"
@@ -22,8 +23,8 @@ import (
 
 func newEditCommand() *cobra.Command {
 	editCommand := &cobra.Command{
-		Use:               "edit INSTANCE",
-		Short:             "Edit an instance of Lima",
+		Use:               "edit INSTANCE|FILE.yaml",
+		Short:             "Edit an instance of Lima or a template",
 		Args:              WrapArgsError(cobra.MaximumNArgs(1)),
 		RunE:              editAction,
 		ValidArgsFunction: editBashComplete,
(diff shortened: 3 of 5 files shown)


go.mod
```diff
@@ -58,6 +59,8 @@ require (
 	github.com/VividCortex/ewma v1.2.0 // indirect
 	github.com/a8m/envsubst v1.4.2 // indirect
 	github.com/alecthomas/participle/v2 v2.1.1 // indirect
+	github.com/bmatcuk/doublestar/v4 v4.6.0 // indirect
+	github.com/braydonk/yaml v0.7.0 // indirect
```
AkihiroSuda, line 63: Not a huge fan of importing multiple YAML libraries

jandubois:
Tests are failing on Windows because it is adding `\r\n` line endings to the YAML file instead of sticking to plain newlines.

go.mod
norio-nomura: Is this the fifth YAML library?

norio-nomura: Oh, is this the first time a YAML library forked from https://github.com/go-yaml/yaml is being added?

afbjorklund: k8s yaml is a wrapper around go-yaml, as far as I know... <https://github.com/kubernetes-sigs/yaml>

norio-nomura: I want to express my gratitude to Go for allowing the use of both a forked library and the original library simultaneously without causing issues. 👏🏻

pkg/yqutil/yqutil_test.go
```diff
@@ -41,17 +41,18 @@ mounts:
 `
 	// Note: yq will use canonical yaml, with indented sequences
 	// Note: yq will not explicitly quote strings, when not needed
+	// Note: yamlfmt will fix indetation of sequences
```
jandubois, line 44:
```suggestion
	// Note: yamlfmt will fix indentation of sequences
```

jandubois (changes requested):
I agree with @afbjorklund that the new libraries provide sufficient value to justify adding one more YAML library to the project.

I think the fact that `limactl edit` can now edit templates in addition to instances should be noted in the command descriptions:

```diff
-		Use:               "edit INSTANCE",
-		Short:             "Edit an instance of Lima",
+		Use:               "edit INSTANCE|FILE.yaml",
+		Short:             "Edit an instance of Lima or a template",
```

jandubois:
Please squash commits!

jandubois (approved):
Thanks, LGTM

jandubois:
@AkihiroSuda Do you have any objections to merging this?

jandubois:
> The issue with line breaks disappearing cannot be fixed with yamlfmt, so it still remains.

Super-hacky idea: Replace all empty lines with `### this is an empty line ###` before editing the file, and then replace them back with empty lines after `yamlfmt` is done.

norio-nomura:
> Super-hacky idea: Replace all empty lines with `### this is an empty line ###` before editing the file, and then replace them back with empty lines after `yamlfmt` is done.

That's exactly the idea that `yamlfmt` adopts.  
https://github.com/google/yamlfmt/blob/main/internal/hotfix/retain_line_break.go

norio-nomura:
> That's exactly the idea that `yamlfmt` adopts

Oh, it seems that the following hack works well:
```patch
diff --git forkSrcPrefix/pkg/yqutil/yqutil.go forkDstPrefix/pkg/yqutil/yqutil.go
index 139d9fcbce3ff6c7ca3df38a04a1fb3ef36ccc1b..8787797f40b6e0a8fcaec1f06b203b25e8ba3c06 100644
--- forkSrcPrefix/pkg/yqutil/yqutil.go
+++ forkDstPrefix/pkg/yqutil/yqutil.go
@@ -1,6 +1,7 @@
 package yqutil
 
 import (
+	"bufio"
 	"bytes"
 	"fmt"
 	"os"
@@ -15,13 +16,17 @@ import (
 // EvaluateExpression evaluates the yq expression, and returns the modified yaml.
 func EvaluateExpression(expression string, content []byte) ([]byte, error) {
 	logrus.Debugf("Evaluating yq expression: %q", expression)
+	contentModified, err := replaceLineBreaksWithMagicString(content)
+	if err != nil {
+		return nil, err
+	}
 	tmpYAMLFile, err := os.CreateTemp("", "lima-yq-*.yaml")
 	if err != nil {
 		return nil, err
 	}
 	tmpYAMLPath := tmpYAMLFile.Name()
 	defer os.RemoveAll(tmpYAMLPath)
-	_, err = tmpYAMLFile.Write(content)
+	_, err = tmpYAMLFile.Write(contentModified)
 	if err != nil {
 		tmpYAMLFile.Close()
 		return nil, err
@@ -94,3 +99,41 @@ func yamlfmt(content []byte) ([]byte, error) {
 	}
 	return formatter.Format(content)
 }
+
+const yamlfmtLineBreakPlaceholder = "#magic___^_^___line"
+
+type paddinger struct {
+	strings.Builder
+}
+
+func (p *paddinger) adjust(txt string) {
+	var indentSize int
+	for i := 0; i < len(txt) && txt[i] == ' '; i++ { // yaml only allows space to indent.
+		indentSize++
+	}
+	// Grows if the given size is larger than us and always return the max padding.
+	for diff := indentSize - p.Len(); diff > 0; diff-- {
+		p.WriteByte(' ')
+	}
+}
+
+func replaceLineBreaksWithMagicString(content []byte) ([]byte, error) {
+	// hotfix: yq does not support line breaks in the middle of a string.
+	var buf bytes.Buffer
+	reader := bytes.NewReader(content)
+	scanner := bufio.NewScanner(reader)
+	var padding paddinger
+	for scanner.Scan() {
+		txt := scanner.Text()
+		padding.adjust(txt)
+		if strings.TrimSpace(txt) == "" { // line break or empty space line.
+			buf.WriteString(padding.String()) // prepend some padding incase literal multiline strings.
+			buf.WriteString(yamlfmtLineBreakPlaceholder)
+			buf.WriteString("\n")
+		} else {
+			buf.WriteString(txt)
+			buf.WriteString("\n")
+		}
+	}
+	return buf.Bytes(), scanner.Err()
+}
```

🤔

norio-nomura:
The changes after running `limactl edit templates/default.yaml --memory 8` using the above patch are as follows:
```patch
diff --git forkSrcPrefix/examples/default.yaml forkDstPrefix/examples/default.yaml
index 5b269ff7e25530c16ce7a9c4fd00251a35850b0d..c46886072da9ff064754da989698e618daa787e8 100644
--- forkSrcPrefix/examples/default.yaml
+++ forkDstPrefix/examples/default.yaml
@@ -43,7 +43,7 @@ cpus: null
 
 # Memory size
 # 🟢 Builtin default: min("4GiB", half of host memory)
-memory: null
+memory: 8GiB
 
 # Disk size
 # 🟢 Builtin default: "100GiB"
@@ -264,10 +264,10 @@ containerd:
 cpuType:
-  # aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
-  # armv7l: "cortex-a7" # (or "host" when running on armv7l host)
-  # riscv64: "rv64" # (or "host" when running on riscv64 host)
-  # x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
+# aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
+# armv7l: "cortex-a7" # (or "host" when running on armv7l host)
+# riscv64: "rv64" # (or "host" when running on riscv64 host)
+# x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
 
 rosetta:
   # Enable Rosetta for Linux (EXPERIMENTAL; will graduate from experimental in Lima v1.0).
@@ -456,8 +456,8 @@ hostResolver:
   # 🟢 Builtin default: null
   hosts:
-    # guest.name: 127.1.1.1
-    # host.name: host.lima.internal
+  # guest.name: 127.1.1.1
+  # host.name: host.lima.internal
 
 # If hostResolver.enabled is false, then the following rules apply for configuring dns:
 # Explicitly set DNS addresses for qemu user-mode networking. By default qemu picks *one*
```

jandubois:
> Oh, it seems that the following hack works well:

If you create a PR with that hack (which I would like to see once the current PR is merged), then I would suggest that we should also fix up the templates so that the comments round-trip correctly (all comments should be in front of the setting they apply to, and not after it).

Then we can add a check to CI to make sure they remain round-trippable. Something like

```shell
find examples -name '*.yaml' -exec limactl edit --set 'del(.nothing)' {} \;
git diff-index --quiet HEAD
```

jandubois:
@AkihiroSuda I'll take the fact that you assigned this PR to the `1.0` milestone as agreement to merge. I was waiting for your feedback because you originally commented:

> Not a huge fan of importing multiple YAML libraries

So I wasn't sure if you still had objections to merging it or not. But I guess you are ok with it now.

norio-nomura:
Thanks! 🙏🏻 

> Then we can add a check to CI to make sure they remain round-trippable.

Before doing that, I want to finalize #2532.

AkihiroSuda:
> @AkihiroSuda I'll take the fact that you assigned this PR to the `1.0` milestone as agreement to merge. I was waiting for your feedback because you originally commented:
> 
> > Not a huge fan of importing multiple YAML libraries
> 
> So I wasn't sure if you still had objections to merging it or not. But I guess you are ok with it now.

Yes, ok to merge, sorry for delay
🙇‍♂️

PR #2574, opened 2024-08-31
Run in a new process group when not using --foreground

nirs:
Previously the hostagenet process was running in the foreground even when not using the --foreground option. It was using the same pgid of limactl process. If the guest was running, but limactl was interrupted the hostagent received the signal and was killed, stopping the VM.

A simple way to fix this issue is to start the hostagent process in a new process group. This way it will be killed sending signals to the limactl process group.

Example run with this change:

    % ps -o pid,pgid,ppid,command
      PID  PGID  PPID COMMAND
    39442 39442 39440 -zsh
    63233 63233 39442 _output/bin/limactl start --vm-type vz --tty=false
    63299 63299 63233 /Users/nsoffer/src/lima/_output/bin/limactl hostagent ...

We can improve this later by adding an option to daemonize the hostagent process (like qemu --daemonize).

Fixes #2573

PR #2595, opened 2024-09-06
Run usernet process in background

nirs:
Similar to the how we run the hostagent process[1], we want to run the
usernet process in the background. Now a program using killpg to cleanup
child processes will not terminate the usernet process.

Example run with this change:

    % ps -o pid,pgid,ppid,command
      PID  PGID  PPID COMMAND
    55768 55768 55767 -zsh
    56126 56126 55768 limactl start userv2.yaml --tty=0
    56128 56128 56126 /Users/nsoffer/src/lima/_output/bin/limactl usernet ...
    56131 56131 56126 /Users/nsoffer/src/lima/_output/bin/limactl hostagent ...

    % ps -o pid,pgid,ppid,command
      PID  PGID  PPID COMMAND
    55768 55768 55767 -zsh
    56128 56128     1 /Users/nsoffer/src/lima/_output/bin/limactl usernet ...
    56131 56131     1 /Users/nsoffer/src/lima/_output/bin/limactl hostagent ...

[1] https://github.com/lima-vm/lima/pull/2574

### DIFF
--- a/pkg/executil/opts_others.go
+++ b/pkg/executil/opts_others.go
@@ -1,6 +1,6 @@
 //go:build !windows
 
-package instance
+package executil
 
 import (
 	"syscall"
--- a/pkg/executil/opts_windows.go
+++ b/pkg/executil/opts_windows.go
@@ -1,4 +1,4 @@
-package instance
+package executil
 
 import (
 	"syscall"
--- a/pkg/instance/start.go
+++ b/pkg/instance/start.go
@@ -17,6 +17,7 @@ import (
 	"github.com/coreos/go-semver/semver"
 	"github.com/lima-vm/lima/pkg/driver"
 	"github.com/lima-vm/lima/pkg/driverutil"
+	"github.com/lima-vm/lima/pkg/executil"
 	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/qemu"
 	"github.com/lima-vm/lima/pkg/qemu/entitlementutil"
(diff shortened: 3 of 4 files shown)


nirs:
Failed test on fedora seems unrelated:
https://github.com/lima-vm/lima/actions/runs/10746186646/job/29806633152?pr=2595

AkihiroSuda:
> Similar to the how we run the hostagent process, we want to run the usernet process in the background. Now a program using killpg to cleanup child processes will not terminate the usernet process.


Could you explain the motivation?

nirs:
> > Similar to the how we run the hostagent process, we want to run the usernet process in the background. Now a program using killpg to cleanup child processes will not terminate the usernet process.
> 
> Could you explain the motivation?

Sorry, I forgot to link the relevant PR, added in the current version.

The issue with the usernet process (that was fixed in the hostagent process) is that it runs in the same process group of the limactl. Since this is always a background process that must continue to run after limactl exit (otherwise user network breaks), it should at least run in a new process group. Ideally it would run as a proper daemon (like qemu --demonize), but this is a bigger change and running in the background is good enough.

Here is the relevant code in a program running limactl:
https://github.com/RamenDR/ramen/blob/b635a80ff528cd6af64b1f2eeea6df6af22113ed/test/drenv/__main__.py#L42

I can amend the commit message if needed.

AkihiroSuda (approved):
Thanks

balajiv113 (approved):

nirs:
@AkihiroSuda any reason to wait this change?

Issue #2577, opened 2024-09-01
Hostagent logs are too noisy


nirs:
### Description

INFO level logs from "[hostagent"] contain many messages that could be debug level or duplicate messages.

Example log:

```
msg="[hostagent] Waiting for the essential requirement 1 of 2: \"ssh\""
msg="[hostagent] [VZ] - vm state change: running"
msg="[hostagent] Waiting for the essential requirement 1 of 2: \"ssh\""
msg="[hostagent] The essential requirement 1 of 2 is satisfied"
msg="[hostagent] Waiting for the essential requirement 2 of 2: \"user session is ready for ssh\""
msg="[hostagent] The essential requirement 2 of 2 is satisfied"
msg="[hostagent] Waiting for the optional requirement 1 of 6: \"systemd must be available\""
msg="[hostagent] Guest agent is running"
msg="[hostagent] Not forwarding TCP 127.0.0.53:53"
msg="[hostagent] Not forwarding TCP 127.0.0.54:53"
msg="[hostagent] The optional requirement 1 of 6 is satisfied"
msg="[hostagent] Not forwarding TCP [::]:22"
msg="[hostagent] Waiting for the optional requirement 2 of 6: \"containerd binaries to be installed\""
msg="[hostagent] Not forwarding TCP 127.0.0.1:41899"
msg="[hostagent] The optional requirement 2 of 6 is satisfied"
msg="[hostagent] Waiting for the optional requirement 3 of 6: \"kubeadm to be installed\""
msg="[hostagent] The optional requirement 3 of 6 is satisfied"
msg="[hostagent] Waiting for the optional requirement 4 of 6: \"kubeadm to be completed\""
msg="[hostagent] Not forwarding TCP 127.0.0.1:45865"
msg="[hostagent] Not forwarding TCP 127.0.0.1:10248"
msg="[hostagent] Not forwarding TCP [::]:10250"
msg="[hostagent] The optional requirement 4 of 6 is satisfied"
msg="[hostagent] Waiting for the optional requirement 5 of 6: \"kubernetes cluster to be running\""
msg="[hostagent] Not forwarding TCP 192.168.105.23:2380"
msg="[hostagent] Not forwarding TCP 192.168.105.23:2379"
msg="[hostagent] Not forwarding TCP 127.0.0.1:2379"
msg="[hostagent] Not forwarding TCP 127.0.0.1:2381"
msg="[hostagent] Not forwarding TCP 127.0.0.1:10257"
[...]

PR #2596, opened 2024-09-08
Change "Not forwarding" messages to debug level

nirs:
Info level should be used only for important events or changes to the system, for example when we forward a port from the guest. Not forwarding a port can be helpful when debugging port forwarding, so keeping it as debug message.

Part-of: #2577

### DIFF
--- a/hack/test-port-forwarding.pl
+++ b/hack/test-port-forwarding.pl
@@ -6,7 +6,7 @@
 # running to perform the tests:
 #
 # ./hack/test-port-forwarding.pl templates/default.yaml
-# limactl --tty=false start templates/default.yaml
+# limactl --tty=false --log-level debug start templates/default.yaml
 # git restore templates/default.yaml
 # ./hack/test-port-forwarding.pl default
 #
--- a/pkg/hostagent/port.go
+++ b/pkg/hostagent/port.go
@@ -94,7 +94,7 @@ func (pf *portForwarder) OnEvent(ctx context.Context, ev *api.Event) {
 		}
 		local, remote := pf.forwardingAddresses(f)
 		if local == "" {
-			logrus.Infof("Not forwarding TCP %s", remote)
+			logrus.Debugf("Not forwarding TCP %s", remote)
 			continue
 		}
 		logrus.Infof("Forwarding TCP from %s to %s", remote, local)
--- a/pkg/portfwd/forward.go
+++ b/pkg/portfwd/forward.go
@@ -26,7 +26,7 @@ func (fw *Forwarder) OnEvent(ctx context.Context, client *guestagentclient.Guest
 	for _, f := range ev.LocalPortsAdded {
 		local, remote := fw.forwardingAddresses(f)
 		if local == "" {
-			logrus.Infof("Not forwarding %s %s", strings.ToUpper(f.Protocol), remote)
+			logrus.Debugf("Not forwarding %s %s", strings.ToUpper(f.Protocol), remote)
 			continue
 		}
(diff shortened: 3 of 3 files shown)


nirs:
I tried to run hack/test-port-forwarding.pl but is seems to be broken now:

```
% ./hack/test-port-forwarding.pl 
Use of uninitialized value $instance in -f at ./hack/test-port-forwarding.pl line 37.
Use of uninitialized value $instance in regexp compilation at ./hack/test-port-forwarding.pl line 64, <$ls> line 1.
Use of uninitialized value $instance in regexp compilation at ./hack/test-port-forwarding.pl line 64, <$ls> line 2.
Use of uninitialized value $instance in regexp compilation at ./hack/test-port-forwarding.pl line 64, <$ls> line 3.
Use of uninitialized value $instance in regexp compilation at ./hack/test-port-forwarding.pl line 64, <$ls> line 4.
Cannot determine sshLocalPort at ./hack/test-port-forwarding.pl line 68, <$ls> line 4.
```

Or maybe the required perl version is not documented?

AkihiroSuda:
I'd say "Not forwarding" message is important and should be printed by default.

Probably `limactl start` should have `--quiet` (`-q`) flag to silence it

nirs:
For the use case of disabling all port forwarding, the "not forwarding" message is not important.

Maybe we need a simple switch for port forwarding?

```yaml
portForwards:
  enabled: false
```

In this case the user made it clear that no forwarding should happen so there is no need to log anything.

If you enabled some port forwarding, it makes sense to show what was not forwarded in case you want to know if your rules are correct, although this is mainly for debugging.

Ranjandas:
👍 for having a simple switch to disable port forwarding.

jandubois:
> It already exists:
> 
> ```yaml
> portForwards:
> - ignore: true
> ```

Note that you cannot disable (or override) redirection of the SSH port (22) because without it Lima doesn't work.

nirs:
@jandubois thanks, but I think when I tried this:

```yaml
portForwards:
- guestPortRange: [1, 65535]
  ignore: true
```

it did not work, and some ports were forwarded from [::]. I had to use:

```yaml
portForwards:
- guestPortRange: [1, 65535]
  guestIP: "0.0.0.0"
  ignore: true
```

But assuming we fix the issue, can we detect this as "I don't want any port forwarding" and log the "not forwarding ..." message in debug level?

I think this is not good user or developer experience when you don't have a simple way to disable a feature, and you need to depend on a special configuration.

nirs:
#2600 seems a better way.

jandubois:
> I tried to run hack/test-port-forwarding.pl but is seems to be broken now:

It is not really something you can run by itself; it is used implicitly by `hack/test-templates.sh`.

You have to run it twice: once with a template filename to specify the template it should modify with the `portForwards` section needed for testing, and once you have started an instance with that modified template you run the script again with the instance name as the parameter.

Since you may also need to install the right version of `netcat` inside the instance, it is probably best to only run it via the `test-templates.sh` script.

nirs:
> > I tried to run hack/test-port-forwarding.pl but is seems to be broken now:
> 
> It is not really something you can run by itself; it is used implicitly by `hack/test-templates.sh`.

I ran it based on the documentation in the file (adding debug log level):

```
# ./hack/test-port-forwarding.pl templates/default.yaml
# limactl --tty=false --log-level debug start templates/default.yaml
# git restore templates/default.yaml
# ./hack/test-port-forwarding.pl default
```

Maybe the comment should mention the additional requirements, or just remove the comment and add a comment about the script that run it?

nirs:
> And, did it work, or show any errors?

If failed with the error I shared, but it looks like I run the last command without the argument.

Issue #2597, opened 2024-09-08
Avoid importing multiple YAML libraries


AkihiroSuda:
Currently, Lima imports 4 libraries of YAML 😓 

https://github.com/lima-vm/lima/blob/52f5ad31e4c1f2a58fc37d1191ceebfdd5150dbe/go.mod#L24

https://github.com/lima-vm/lima/blob/52f5ad31e4c1f2a58fc37d1191ceebfdd5150dbe/go.mod#L48

https://github.com/lima-vm/lima/blob/52f5ad31e4c1f2a58fc37d1191ceebfdd5150dbe/go.mod#L123

https://github.com/lima-vm/lima/blob/52f5ad31e4c1f2a58fc37d1191ceebfdd5150dbe/go.mod#L130

Wish Go had provided stdlib for YAML...

afbjorklund:
Some of these dependencies are indirect:

```sh
# github.com/goccy/go-yaml
github.com/lima-vm/lima/pkg/instance
github.com/goccy/go-yaml
# gopkg.in/yaml.v3
github.com/lima-vm/lima/pkg/limayaml
gopkg.in/yaml.v3
# gopkg.in/yaml.v2
github.com/lima-vm/lima/pkg/guestagent/kubernetesservice
k8s.io/api/core/v1
k8s.io/apimachinery/pkg/runtime
sigs.k8s.io/structured-merge-diff/v4/value
gopkg.in/yaml.v2
# sigs.k8s.io/yaml
github.com/lima-vm/lima/pkg/guestagent/kubernetesservice
k8s.io/client-go/tools/clientcmd
k8s.io/client-go/tools/clientcmd.test
sigs.k8s.io/yaml
```

But I guess you are not keen on a 5th?

* #2593

afbjorklund:
The "goccy" version of yqlib has this note:

` NOTE this is still a WIP - not yet ready.`

So I don't think it is ready to switch yet?

> Wish Go had provided stdlib for YAML...

The informal standard is/was [go-yaml](https://github.com/go-yaml/yaml) (v3)

But it has the same lack of progress as json.

So that's the problem of it being "built-in"

afbjorklund:
We could use yqlib for verification, instead of calling `gopkg.in/yaml.v3` directly...

This way, if/when yqlib changes their yaml library it would also be changed in lima?

It would not make a **huge** difference, but...

```diff
@@ -45,7 +45,6 @@ require (
[...]

PR #2601, opened 2024-09-09
Validate yaml with yqlib and not with go-yaml v3

afbjorklund:
Just in case that yqlib starts using another library...

Issue #2597


AkihiroSuda (approved):
Thanks

PR #2605, opened 2024-09-10
Allow separate rules for UDP port forwarding

jandubois:
Additionally a `Proto` value of `"any"` would match both TCP and UDP ports.

The default for `Proto` remains `tcp`, both for backwards compatibility, and because `any` rarely makes sense, except e.g. when you want to disable all port forwarding for any protocol.

We should also update `hack/test-portforwarding.pl` to cover UDP port forwarding as well. Maybe in a separate PR to allow for manual testing of UDP forwarding rules already.


jandubois:
UDP port forwarding tests should go into a separate PR.

This PR just makes it possible to have separate rules for TCP vs. UDP; without this PR the TCP rules were applied to UDP as well.

AkihiroSuda (approved):
Thanks

PR #2607, opened 2024-09-11
Makefile: Add architecture variations to the artifact targets

norio-nomura:


Makefile
```diff
@@ -265,28 +265,34 @@ install-tools:
...
-.PHONY: artifacts-darwin
-artifacts-darwin:
-	mkdir -p _artifacts
-	GOOS=darwin GOARCH=amd64 make clean binaries
-	$(TAR) -C _output/ -czvf _artifacts/lima-$(VERSION_TRIMMED)-Darwin-x86_64.tar.gz ./
-	GOOS=darwin GOARCH=arm64 make clean binaries
-	$(TAR) -C _output -czvf _artifacts/lima-$(VERSION_TRIMMED)-Darwin-arm64.tar.gz ./
+.PHONY: artifacts-darwin artifact-darwin-aarch64 artifact-darwin-arm64 artifact-darwin-x86_64
+artifacts-darwin: artifact-darwin-x86_64 artifact-darwin-arm64
+artifact-darwin-arm64: ENVS=GOOS=darwin GOARCH=arm64
```
jandubois, line 270: I was not aware that you can have target-specific environment variables in Makefiles, and kind of find it confusing how they are even applied to building prerequisites (and if multiple targets share a prerequisite, then it is kind of random which env variables are in effect).

But it looks like this isn't a problem in our case, so I guess it is fine. Still makes my head spin a little though.

jandubois (dismissed):
Everything seems to work, but I would prefer to remove the redundant `*-arm64` targets.

Makefile
```diff
@@ -265,28 +265,34 @@ install-tools:
...
+artifact-darwin-arm64: _artifacts/lima-$(VERSION_TRIMMED)-Darwin-arm64.tar.gz
+artifact-darwin-aarch64: artifact-darwin-arm64
```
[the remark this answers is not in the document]
norio-nomura: Since the artifact file names use the values returned by `uname -m` on each platform, it would be better to use the values returned by `uname -m` for the Makefile targets that are likely to run on each platform as well.
`arm64` on Darwin, `aarch64` on Linux.

Makefile
```diff
@@ -265,28 +265,34 @@ install-tools:
...
+artifact-darwin-x86_64: ENVS=GOOS=darwin GOARCH=amd64
+artifact-darwin-x86_64: _artifacts/lima-$(VERSION_TRIMMED)-Darwin-x86_64.tar.gz
+
+.PHONY: artifacts-linux artifact-linux-aarch64 artifact-linux-arm64 artifact-linux-x86_64
+artifacts-linux: artifact-linux-x86_64 artifact-linux-aarch64
+artifact-linux-arm64: artifact-linux-aarch64
```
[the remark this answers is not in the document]
norio-nomura: I'm starting to feel that it might be better not to provide an alias.

Makefile
norio-nomura: Even within the `artifacts-darwin` target before this change, there were inconsistencies in the way parameters were passed to `$(TAR)`, so I thought it would be best to unify them as much as possible.
```Makefile
.PHONY: artifacts-darwin
artifacts-darwin:
	mkdir -p _artifacts
	GOOS=darwin GOARCH=amd64 make clean binaries
	$(TAR) -C _output/ -czvf _artifacts/lima-$(VERSION_TRIMMED)-Darwin-x86_64.tar.gz ./
	GOOS=darwin GOARCH=arm64 make clean binaries
	$(TAR) -C _output -czvf _artifacts/lima-$(VERSION_TRIMMED)-Darwin-arm64.tar.gz ./
```

Makefile
```diff
@@ -265,28 +265,32 @@ install-tools:
...
 
-.PHONY: artifacts-darwin
-artifacts-darwin:
-	mkdir -p _artifacts
-	GOOS=darwin GOARCH=amd64 make clean binaries
-	$(TAR) -C _output/ -czvf _artifacts/lima-$(VERSION_TRIMMED)-Darwin-x86_64.tar.gz ./
-	GOOS=darwin GOARCH=arm64 make clean binaries
-	$(TAR) -C _output -czvf _artifacts/lima-$(VERSION_TRIMMED)-Darwin-arm64.tar.gz ./
+.PHONY: artifacts-darwin artifact-darwin-arm64 artifact-darwin-x86_64
+artifacts-darwin: artifact-darwin-x86_64 artifact-darwin-arm64
```
jandubois, line 269: I know this is a trivial thing, but it bothers me that the names are listed in different orders in these 2 lines:

```suggestion
.PHONY: artifacts-darwin artifact-darwin-arm64 artifact-darwin-x86_64
artifacts-darwin: artifact-darwin-arm64 artifact-darwin-x86_64
```

Same thing for the Linux artifacts.

jandubois (approved):
Thanks, LGTM

Makefile
norio-nomura: Changing that would alter the build order compared to before.
Additionally, personally, since I use the path to `_output/bin` in the working directory, I prefer having `arm64` built later. I don't need to rebuild after running `make artifacts-darwin`.😃

jandubois: You could also change the order in the `.PHONY` directive to match the build order. 😄 

But it isn't that important...

afbjorklund:
Did you test this? Looks like a no-op to me

	GOOS=linux GOARCH=amd64 CC=x86_64-linux-gnu-gcc make clean binaries

artifact-linux-aarch64: ENVS=GOOS=linux GOARCH=arm64 CC=aarch64-linux-gnu-gcc
artifact-linux-aarch64: _artifacts/lima-$(VERSION_TRIMMED)-Linux-aarch64.tar.gz

Or at least I thought the env is now on a separate line?

afbjorklund:
Was a bit late to the review, I also found the ENVS a bit confusing but seems to be very compact.

You can replace the phony target of "mkdir-artifacts" with a real target of "_artifacts", as a minor.

```Makefile
_artifacts/lima-%.tar.gz: _artifacts
	$(ENVS) make clean binaries
	$(TAR) -C _output/ -czvf $@ ./

_artifacts:
	mkdir -p $@
```

And probably just hide it from the output with a `@`

norio-nomura:
Thanks! 🙏🏻  

> You can replace the phony target of "mkdir-artifacts" with a real target of "_artifacts"

I considered that approach as well, but since the number of lines doesn't change, I decided to keep the current format, where the target names make it clear what they do.

PR #2608, opened 2024-09-11
downloader: re-download digest-less cached images if `last-modified` differs

norio-nomura:


pkg/downloader/downloader.go
```diff
@@ -239,18 +240,27 @@ func Download(ctx context.Context, local, remote string, opts ...Opt) (*Result,
...
-			if err := copyLocal(ctx, localPath, shadData, ext, o.decompress, o.description, o.expectedDigest); err != nil {
-				return nil, err
+			if match, err := matchLastModified(ctx, shadTime, remote); err != nil {
+				logrus.WithError(err).Info("Failed to retrieve last-modified for cached digest-less image; using cached image.")
+			} else if match {
+				if err := copyLocal(ctx, localPath, shadData, ext, o.decompress, o.description, o.expectedDigest); err != nil {
+					return nil, err
+				}
+			} else {
+				logrus.Info("Re-downloading digest-less image: last-modified mismatch detected.")
```
AkihiroSuda, line 250: Maybe print the last-modified value here

AkihiroSuda (approved):
Thanks

Eventually we should have a unit test for downloader

pkg/downloader/downloader.go
jandubois: I think if you print the last-modified time, then it would be nice to print both the time for the cached version, and for the remote version.

jandubois (approved):

pkg/downloader/downloader.go
norio-nomura: I rewrote `matchLastModified` in the process of including the actual `last-modified` timestamp in the logs, so I would appreciate it if you could review that part again.

jandubois (dismissed):
Thanks, LGTM

AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

Issue #2611, opened 2024-09-12
Update the openSUSE distributions


afbjorklund:
### Description

I noticed as I was watching an openSUSE presentations, that the distributions have been updated...

* https://events.opensuse.org/conferences/oSC24/program/proposals/4411

* https://media.ccc.de/v/4411-we-re-all-grown-up-opensuse-is-not-suse

So made some updates, to fix the "Leap" and the "Tumbleweed" distributions to their current state.

1. Renamed "opensuse" (symlink) to "opensuse-leap"

2. Added aarch64 version of "opensuse-tumbleweed"

afbjorklund:
Added the new logos to the Lima GUI, as well:

https://github.com/afbjorklund/lima-gui/commit/33ff5ba758da48b9ed8f9d02340cee97aa7484ce

https://en.opensuse.org/Logocontest

jandubois:
Unrelated to this issue, just FYI: Lima does not currently work on openSUSE as a host because qemu-ovmf has been broken: https://bugzilla.suse.com/show_bug.cgi?id=1230291

PR #2612, opened 2024-09-12
Rename opensuse to opensuse-leap

afbjorklund:
Issue #2611

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -19,7 +19,7 @@ Distro:
 - [`debian-11`](./debian-11.yaml): Debian GNU/Linux 11(bullseye)
 - [`debian-12`](./debian-12.yaml), `debian.yaml`: ⭐Debian GNU/Linux 12(bookworm)
 - [`fedora`](./fedora.yaml): ⭐Fedora
-- [`opensuse`](./opensuse.yaml): ⭐openSUSE Leap
+- [`opensuse-leap`](./opensuse-leap.yaml), `opensuse.yaml`: ⭐openSUSE Leap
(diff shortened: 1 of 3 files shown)


jandubois (dismissed):
Thanks, LGTM (only verified the x86_64 checksum though)

afbjorklund:
The current image is a rolling "symlink", so the checksums can only be done on the "Build"

```
95b20098fbb0e5456bf1529387c8b9513056405256549e9f2f9f35d367ac8c05 openSUSE-Leap-15.6-Minimal-VM.aarch64-Cloud.qcow2
95b20098fbb0e5456bf1529387c8b9513056405256549e9f2f9f35d367ac8c05  openSUSE-Leap-15.6-Minimal-VM.aarch64-15.6.0-Cloud-Build13.327.qcow2
```

And it might not be worth it, to update the URL for every build? (as in commit b9d6d6ab59183bffd15f2cbdbff66b8b8b4887ea)

jandubois:
> And it might not be worth it, to update the URL for every build?

Yes, unfortunately only the latest build is kept, so the URLs including the build number will break frequently. I believe this is due to concerns about the space requirement for the download mirrors.

So yes, dropping the checksum is probably the right solution.

jandubois (approved):

afbjorklund:
It still needs to be bumped for every OS release, like from 15.5 to 15.6 (the 42 was annoying)

jandubois:
> It still needs to be bumped for every OS release

Yeah, that is unavoidable. We should write a script that does this automatically...

> like from 15.5 to 15.6 (the 42 was annoying)

Yes, but still better than the Windows version "numbers": 95 → 98 → 4 → 2000 → XP → 7 … 😄

PR #2625, opened 2024-09-17
yqutil: fix to preserve line breaks

norio-nomura:
`yamlfmt` works by replacing line breaks with place holder strings to avoid their loss when using `gopkg.in/yaml.v3`. By replacing line breaks with place holder strings before `yqlib` uses `gopkg.in/yaml.v3`, we can ultimately prevent the loss of line breaks through `yamlfmt`.

https://github.com/lima-vm/lima/pull/2593#issuecomment-2351462571

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -264,10 +264,10 @@ containerd:
 # Setting of instructions is supported like this: "qemu64,+ssse3".
 # 🟢 Builtin default: hard-coded arch map with type (see the output of `limactl info | jq .defaultTemplate.cpuType`)
 cpuType:
-  # aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
-  # armv7l: "cortex-a7" # (or "host" when running on armv7l host)
-  # riscv64: "rv64" # (or "host" when running on riscv64 host)
-  # x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
+# aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
+# armv7l: "cortex-a7" # (or "host" when running on armv7l host)
+# riscv64: "rv64" # (or "host" when running on riscv64 host)
+# x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
 
 rosetta:
   # Enable Rosetta for Linux (EXPERIMENTAL; will graduate from experimental in Lima v1.0).
@@ -456,8 +456,8 @@ hostResolver:
   # predefined to specify the gateway address to the host.
   # 🟢 Builtin default: null
   hosts:
-    # guest.name: 127.1.1.1
-    # host.name: host.lima.internal
+  # guest.name: 127.1.1.1
+  # host.name: host.lima.internal
 
 # If hostResolver.enabled is false, then the following rules apply for configuring dns:
 # Explicitly set DNS addresses for qemu user-mode networking. By default qemu picks *one*
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -264,10 +264,10 @@ containerd:
 # Setting of instructions is supported like this: "qemu64,+ssse3".
 # 🟢 Builtin default: hard-coded arch map with type (see the output of `limactl info | jq .defaultTemplate.cpuType`)
 cpuType:
-# aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
-# armv7l: "cortex-a7" # (or "host" when running on armv7l host)
-# riscv64: "rv64" # (or "host" when running on riscv64 host)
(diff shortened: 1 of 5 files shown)


examples/default.yaml
```diff
@@ -456,8 +456,8 @@ hostResolver:
   # predefined to specify the gateway address to the host.
   # 🟢 Builtin default: null
   hosts:
-    # guest.name: 127.1.1.1
-    # host.name: host.lima.internal
+  # guest.name: 127.1.1.1
+  # host.name: host.lima.internal
```
jandubois, line 460: I think it looks better to maintain the indentation to show that the lines are commented out key/value mappings. Just move the `#` to the front:

```suggestion
  #   guest.name: 127.1.1.1
  #   host.name: host.lima.internal
```

examples/default.yaml
```diff
@@ -264,10 +264,10 @@ containerd:
...
 # 🟢 Builtin default: hard-coded arch map with type (see the output of `limactl info | jq .defaultTemplate.cpuType`)
 cpuType:
-  # aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
-  # armv7l: "cortex-a7" # (or "host" when running on armv7l host)
-  # riscv64: "rv64" # (or "host" when running on riscv64 host)
-  # x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
+# aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
+# armv7l: "cortex-a7" # (or "host" when running on armv7l host)
+# riscv64: "rv64" # (or "host" when running on riscv64 host)
+# x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
```
jandubois, line 270: I'm sorry, I should have mentioned that these comments should also remain indented:

```suggestion
#   aarch64: "cortex-a72" # (or "host" when running on aarch64 host)
#   armv7l: "cortex-a7" # (or "host" when running on armv7l host)
#   riscv64: "rv64" # (or "host" when running on riscv64 host)
#   x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
```

pkg/yqutil/yqutil.go
```diff
@@ -15,13 +16,17 @@ import (
...
+	contentModified, err := replaceLineBreaksWithMagicString(content)
+	if err != nil {
+		return nil, err
+	}
 	tmpYAMLFile, err := os.CreateTemp("", "lima-yq-*.yaml")
 	if err != nil {
 		return nil, err
 	}
 	tmpYAMLPath := tmpYAMLFile.Name()
 	defer os.RemoveAll(tmpYAMLPath)
```
jandubois, line 28: I know this is not part of this PR, but why is this calling `os.RemoveAll()` instead of `os.Remove()` given that `tmpYAMLPath` is a file and not a directory.

@afbjorklund Do you remember why you choose `RemoveAll`?

pkg/yqutil/yqutil.go
```diff
@@ -94,3 +99,41 @@ func yamlfmt(content []byte) ([]byte, error) {
 	}
 	return formatter.Format(content)
 }
+
+const yamlfmtLineBreakPlaceholder = "#magic___^_^___line"
```
jandubois, line 103: I was wondering how the magic string got replaced back by a newline until I realized that we are relying here on the internals of the `yamlfmt` hotfix.

We do have a test to warn us when upstream modifies their hotfix and breaks our piggyback code.

But I think this needs to be documented. And we need to acknowledge that the `adjust` and `replaceLineBreak…` functions are slightly modified versions of the `yamlfmt` code.

I'm also curious about the change in `replaceLineBreaksWithMagicString` relative to the upstream `replaceLineBreakFeature` function (the removal of the `inLineBreaks` logic). I assume this comment is supposed to explain it:

```go
// hotfix: yq does not support line breaks in the middle of a string.
```

It would help to clarify that this refers to the local change because we pass the YAML through yq before passing it to yamlfmt, and is not referring to the "hotfix" logic in yamlfmt.

See other cases where we modified upstream code:


```console
$ ag adapted
pkg/hostagent/dns/dns.go
1:// This file has been adapted from https://github.com/norouter/norouter/blob/v0.6.4/pkg/agent/dns/dns.go

pkg/reflectutil/reflectutil.go
1:// This file has been adapted from https://github.com/containerd/nerdctl/blob/v1.0.0/pkg/reflectutil/reflectutil.go
```

jandubois (changes requested):
Needs more comments and acknowledgement of the copied upstream code.

norio-nomura:
I rewrote it to avoid relying on `yamlfmt`'s internal code.

jandubois (approved):
Thanks, LGTM

norio-nomura:
Thanks! 🙏🏻

PR #2626, opened 2024-09-18
CI: see if macos-14-xlarge supports virt

AkihiroSuda:


jandubois:
I kind of doubt it because https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#limitations-for-macos-larger-runners still says:

> Nested-virtualization and Metal Performance Shaders (MPS) are not supported due to the limitation of Apple's Virtualization Framework.

But I guess it is worth trying

AkihiroSuda:
```
{"level":"fatal","msg":"Error Domain=VZErrorDomain Code=2 Description=\"Invalid virtual machine configuration. Virtualization is not available on this hardware.\" UserInfo={\n    NSLocalizedFailure = \"Invalid virtual machine configuration.\";\n    NSLocalizedFailureReason = \"Virtualization is not available on this hardware.\";\n}","time":"2024-09-18T04:41:24Z"}
```

jandubois:
Yeah, regardless of hardware (needs M3+) doesn't nested virtualization also require macOS Sequoia, which would be macos-15?

AkihiroSuda:
> Yeah, regardless of hardware (needs M3+) doesn't nested virtualization also require macOS Sequoia, which would be macos-15?

Right, unless they begin to host baremetal instances (like EC2)

Issue #2627, opened 2024-09-18
Rosetta register failing in Fedora distro - Selinux denial


hasan4791:
### Description

When Rosetta is being registered from init, there is an issue with selinux. We need to identify the correct target label for Rosetta or need to add custom selinux policy(only if required).

```
[chasan@lima-podman-rootful Work]$ sudo su -
[systemd]
Failed Units: 1
  systemd-binfmt.service
[root@lima-podman-rootful ~]# 
[root@lima-podman-rootful ~]# systemctl restart systemd-binfmt
type=AVC msg=audit(1726639503.952:487): avc:  denied  { execute } for  pid=2199 comm="systemd-binfmt" name="rosetta" dev="virtiofs" ino=2 scontext=system_u:system_r:init_t:s0 tcontext=system_u:object_r:nfs_t:s0 tclass=file permissive=0
type=SERVICE_START msg=audit(1726639503.953:488): pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=systemd-binfmt comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=failed'UID="root" AUID="unset"
Job for systemd-binfmt.service failed because the control process exited with error code.
See "systemctl status systemd-binfmt.service" and "journalctl -xeu systemd-binfmt.service" for details.
 [root@lima-podman-rootful ~]# 
 [root@lima-podman-rootful ~]#  journalctl -xeu systemd-binfmt.service 
░░ 
░░ The process' exit code is 'exited' and its exit status is 1.
Sep 18 11:44:22 lima-test systemd[1]: systemd-binfmt.service: Failed with result 'exit-code'.
░░ Subject: Unit failed
░░ Defined-By: systemd
░░ Support: https://lists.freedesktop.org/mailman/listinfo/systemd-devel
░░ 
░░ The unit systemd-binfmt.service has entered the 'failed' state with result 'exit-code'.
Sep 18 11:44:22 lima-test systemd[1]: Failed to start systemd-binfmt.service - Set Up Additional Binary Formats.
░░ Subject: A start job for unit systemd-binfmt.service has failed
░░ Defined-By: systemd
░░ Support: https://lists.freedesktop.org/mailman/listinfo/systemd-devel
░░ 
░░ A start job for unit systemd-binfmt.service has finished with a failure.
░░ 
░░ The job identifier is 1558 and the job result is failed.
[...]

PR #2630, opened 2024-09-18
Update selinux label for rosetta from nfs_t to bin_t

hasan4791:
With recent integration with systemd-binfmt,the existing nfs_t selinux label for Rosetta is causing issues while registering it. This behaviour needs to be fixed by setting the label as bin_t.

Fixes: https://github.com/lima-vm/lima/issues/2627

### DIFF
--- a/hack/test-selinux.sh
+++ b/hack/test-selinux.sh
@@ -20,8 +20,14 @@ NAME="$1"
 ## so keeping the label as nfs_t fits right. Package container-selinux by
 ## default adds rules for nfs_t context which allows container workloads to work as well.
 ## https://github.com/lima-vm/lima/pull/1965
+##
+## With integration[https://github.com/lima-vm/lima/pull/2474] with systemd-binfmt,
+## the existing "nfs_t" selinux label for Rosetta is causing issues while registering it.
+## This behaviour needs to be fixed by setting the label as "bin_t"
+## https://github.com/lima-vm/lima/pull/2630
 ##########################################################################################
-expected="context=system_u:object_r:nfs_t:s0"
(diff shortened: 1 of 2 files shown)


hack/test-selinux.sh
```diff
@@ -20,8 +20,14 @@ NAME="$1"
 ## so keeping the label as nfs_t fits right. Package container-selinux by
 ## default adds rules for nfs_t context which allows container workloads to work as well.
 ## https://github.com/lima-vm/lima/pull/1965
+##
+## With recent integration[https://github.com/lima-vm/lima/pull/2474] with systemd-binfmt,
```
AkihiroSuda, line 24: "recent" does not make sense as it is unclear to readers when this was written

hasan4791: I had attached the PR as well. Would that work?

AkihiroSuda: No, as people won't open all the links when they read the comment

hasan4791: Understood. So what do you think would suit here well, I will update according to that.

AkihiroSuda: Just remove "recent"

hasan4791: Done

hasan4791:
Wondering whats causing the test failure!!!
Locally this works but not in CI

hasan4791:
@AkihiroSuda couls you help me with rerunning those failed/cancelled tests again. Looks to be flaky!!

hasan4791:
Required test jobs are passed now but some job is failing which is irrelevant to the changes in this PR.

AkihiroSuda (approved):
Thanks

PR #2637, opened 2024-09-19
CI: try macos-14-large (Intel)

AkihiroSuda:
This might provide better (or worse) stability

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,7 +153,7 @@ jobs:
   integration:
     name: Integration tests
     # on macOS 12, the default vmType is QEMU
-    runs-on: macos-12
+    runs-on: macos-12  # Do not upgrade until GitHub removes macos-12
     timeout-minutes: 120
     steps:
     - uses: actions/checkout@v4
@@ -320,7 +320,7 @@ jobs:
 
   vmnet:
     name: "VMNet test"
-    runs-on: macos-12
+    runs-on: macos-14-large
     timeout-minutes: 120
     steps:
     - uses: actions/checkout@v4
@@ -370,7 +370,7 @@ jobs:
 
   upgrade:
     name: "Upgrade test"
-    runs-on: macos-12
+    runs-on: macos-14-large
     timeout-minutes: 120
     strategy:
       matrix:
@@ -400,8 +400,8 @@ jobs:
 
   vz:
     name: "vz"
-    # on macOS 13, the default vmType is VZ
-    runs-on: macos-13
+    # on macOS >= 13, the default vmType is VZ
+    runs-on: macos-14-large
     timeout-minutes: 120
     strategy:
       fail-fast: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -374,7 +374,7 @@ jobs:
     timeout-minutes: 120
     strategy:
       matrix:
-        oldver: ["v0.15.1"]
+        oldver: ["v0.23.2"]
     steps:
     - uses: actions/checkout@v4
       with:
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
`Upgrade test` is still killed without logs 😥

.github/workflows/test.yml
```diff
@@ -400,8 +400,8 @@ jobs:
 
   vz:
     name: "vz"
-    # on macOS 13, the default vmType is VZ
-    runs-on: macos-13
+    # on macOS >= 13, the default vmType is VZ
+    runs-on: macos-14-large
```
nirs, line 405: This is intel because you configured the repo with an intel runner?

jandubois: Note that the large runners are not available in free organizations; we can only use them because CNCF pays for it (or gets time donated by Microsoft, I don't know).

AkihiroSuda:
Can any of @lima-vm/committers analyze why the Upgrade test still fails without logs and is marked as “Skipped”?
🤔

jandubois:
> Can any of @lima-vm/committers analyze why the Upgrade test still fails without logs and is marked as “Skipped”?

I've been trying, but without success so far.

I've removed the retry action, removed the cache, tried to upgrade from 0.20.2 instead of 0.15.1, but the failure remains the same, the running in the upgrade test is hanging after the second ssh message:

```
time="2024-09-20T00:17:23Z" level=info msg="[hostagent] Starting QEMU (hint: to watch the boot progress, see \"/Users/runner/.lima/test-upgrade/serial.log\")"
[103](https://github.com/jandubois/lima/actions/runs/10950674588/job/30406341318?pr=1#step:6:104)
time="2024-09-20T00:17:23Z" level=info msg="SSH Local Port: 49209"
[104](https://github.com/jandubois/lima/actions/runs/10950674588/job/30406341318?pr=1#step:6:105)
time="2024-09-20T00:17:23Z" level=info msg="[hostagent] Waiting for the essential requirement 1 of 5: \"ssh\""
[105](https://github.com/jandubois/lima/actions/runs/10950674588/job/30406341318?pr=1#step:6:106)
time="2024-09-20T00:17:33Z" level=info msg="[hostagent] Waiting for the essential requirement 1 of 5: \"ssh\""
```

I think something is so wrong that the runner actually disconnects, and that somehow drops even the logs there were there until that point.

I'm trying a little bit longer, but the jobs are getting queued up, so progress is slow.

I wonder how this test ever passed before.

jandubois:
> I'm trying a little bit longer, but the jobs are getting queued up, so progress is slow.

I've tried a gazillion different things, and believe the issue is an incompatibility with the latest qemu version from brew and the GitHub runners. It is somewhat surprising, as that version seems to work fine for me locally, also on an Intel Monterey machine.

I've created #2640 and hope it fixes the upgrade test. It will build qemu 8.2.1 from source; if it works, then I will try tomorrow to download the bottle instead. I've been running into some issues with authorization when trying to fetch the bottle with `curl`, and I'm out of time for now.

BTW, brew no longer supports Monterey, so all formulas will start building from source once they get updated:

```
Warning: You are using macOS 12.
We (and Apple) do not provide support for this old version.
It is expected behaviour that some formulae will fail to build in this old version.
It is expected behaviour that Homebrew will be buggy and slow.
Do not create any issues about this on Homebrew's GitHub repositories.
Do not create any issues even if you think this message is unrelated.
Any opened issues will be immediately closed without response.
Do not ask for help from Homebrew or its maintainers on social media.
You may ask for help in Homebrew's discussions but are unlikely to receive a response.
Try to figure out the problem yourself and submit a fix as a pull request.
We will review it but may or may not accept it.
```

AkihiroSuda:
Thanks @jandubois 

Probably no need to use macos-14-large ?

PR #2639, opened 2024-09-19
nick-invision/retry has been transferred to nick-fields/retry

jandubois:
See https://github.com/nick-fields/retry/commit/e53cf64f1694d990deac53a56d2207af3fbedd6a

It seems like the repo was transferred, so GitHub has a redirect from the old name anyways, but it seems better to use the updated name.


AkihiroSuda (approved):
Thanks

PR #2642, opened 2024-09-21
Use script to pick the qemu formula for the desired version

jandubois:
This will automatically pick the right bottle for the OS of the runner, and we can easily change the QEMU version number without having to look up commit ids manually.

Based on discussion at https://github.com/lima-vm/lima/pull/2640#pullrequestreview-2317320994


norio-nomura (approved):
LGTM! 👍🏻

PR #969, opened 2022-07-19
Add Lima KubeCon 22 EU talk to repo

jandubois:
Signed-off-by: Jan Dubois <jan.dubois@suse.com>

PR #2649, opened 2024-09-23
Add fuzzing audit report

AdamKorcz:
cc @jandubois


jandubois:
@AkihiroSuda I know you mentioned previously that we don't want PDF files in the repo (https://github.com/lima-vm/lima/pull/969#discussion_r925021150).

However, it looks like CNCF is not publishing the reports themselves, but just blog posts that link to them.

So a good place would be to link it from the Lima website, but that would still end up in this repo.

Given that it is just 1MB, I don't think it matter that much, but what do you think?

AkihiroSuda (approved):

PR #2652, opened 2024-09-24
Get qemu bottle commit from git checkout

jandubois:
Running `gh search` sometimes runs into secondary rate limits.

See also https://github.com/cli/cli/issues/3292

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -379,6 +379,14 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
+    - name: Fetch homebrew-core commit messages
+      uses: actions/checkout@v4
+      with:
+        # needed by ./hack/brew-install-version.sh
+        repository: homebrew/homebrew-core
+        path: homebrew-core
+        fetch-depth: 0
+        filter: tree:0
     - uses: actions/setup-go@v5
       with:
         go-version: 1.23.x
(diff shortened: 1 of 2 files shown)


jandubois:
I see these messages in the action output, which look concerning, but don't seem to be errors, and also are not related to this change:

```
/usr/local/Homebrew/Library/Homebrew/list.sh: line 63: echo: write error: Broken pipe
(https://github.com/lima-vm/lima/actions/runs/11021605066/job/30609044009?pr=2652#step:6:74)
/usr/local/Homebrew/Library/Homebrew/list.sh: line 83: echo: write error: Broken pipe
(https://github.com/lima-vm/lima/actions/runs/11021605066/job/30609044009?pr=2652#step:6:75)
Uninstalling /usr/local/Cellar/qemu/9.1.0... (165 files, 626.1MB)
```

.github/workflows/test.yml
```diff
@@ -379,6 +379,13 @@ jobs:
     - uses: actions/checkout@v4
       with:
         fetch-depth: 0
+    - uses: actions/checkout@v4
```
norio-nomura, line 382: I think it will be complete once you add a `name` field to this `step` that indicates what is being checked out. 👍🏻

norio-nomura (approved):
If you could add a `name` field to the `checkout` step you added, then I think everything else looks good to me!

PR #2655, opened 2024-09-26
Make sure lima user fallback uses same validation as template

pvdvreede:
The regex currently being used is different from the identifier's validation from containerd. The fallback test does allow a starting `_` but the validation for the identifier does not.

This results in a bug where the a user that starts with an `_` will pass fallback validation (ie not be set to lima for the user), but will then fail the cidata validation here: https://github.com/lima-vm/lima/blob/master/pkg/cidata/template.go#L95.

Error log shows as:
` ERRO[0000] [hostagent] identifier "_nixbld1" must match ^[A-Za-z0-9]+(?:[._-](?:[A-Za-z0-9]+))*$: invalid argument  fields.level=fatal`

This PR sets the same validation check in both spots to fix this and make sure they stay in sync in the future.

### DIFF
--- a/pkg/osutil/user.go
+++ b/pkg/osutil/user.go
@@ -11,6 +11,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/containerd/containerd/identifiers"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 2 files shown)


jandubois:
> The regex currently being used is different from the identifier's validation from containerd.

I think using the `containerd` identifier validation for Linux hostnames is wrong. If anything we should be using the regex from `osutils/user`.

However, I think I would prefer to remove the validation in `cidata` completely and just rely on it being validated by `limayaml.FillDefaults` (see https://github.com/lima-vm/lima/pull/2378#issuecomment-2344372233 for my latest thoughts).

So I would like to either

1. close this PR without merging, or
2. change this PR to use the `osutils.user` regex in the `cidata` validation

I'm willing to consider (2) because it is not clear when a reworked version of #2378 will be available.

jandubois (changes requested):
Oops, sent my review as a regular comment.

So please either close the PR or modify it to use the `osutils.user` regex also in `cidata`.

pvdvreede:
Thanks @jandubois . I have created #2673 as per your suggestion, so closing this PR in favour of that one.

PR #2659, opened 2024-09-26
Add minimumLimaVersion and vmOpts.qemu.minimumVersion fields to lima.yaml

jandubois:
This will allow us to specify the requirements in the templates not just in a comment, but in a way that can be validated.

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -5,11 +5,23 @@
 # Default values in this YAML file are specified by `null` instead of Lima's "builtin default" values,
 # so they can be overridden by the $LIMA_HOME/_config/default.yaml mechanism documented at the end of this file.
 
+# A template should specify the minimum Lima version required to parse this template correctly.
+# It should not be set if the minimum version is less than 1.0.0
+# 🟢 Builtin default: not set
+minimumLimaVersion: null
+
 # VM type: "qemu", "vz" (on macOS 13 and later), or "default".
 # The vmType can be specified only on creating the instance.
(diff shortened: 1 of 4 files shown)


jandubois:
I noticed we use inconsistent capitalization of the QEMU name, sometimes even in the same type:

```
func (l *LimaQemuDriver) killQEMU(…)
```

I feel that completely capitalizing it as `QEMU` is the right way to use it, but we should do this consistently. If people prefer `Qemu`, then I can use this in the new variables too, but we should then rename all the old names to match as well.

examples/default.yaml
```diff
@@ -5,6 +5,16 @@
...
+# A template should specify the minimum Lima version required to parse this template correctly.
+# It should not be set if the minimum version is less than 1.0.0
+# 🟢 Builtin default: not set
+minimumLimaVersion: null
+
+# Minimum version of QEMU required to create an instance of this template.
+# Will be ignored if the vmType is not "qemu"
+# 🟢 Builtin default: not set
+minimumQEMUVersion: null
+
```
AkihiroSuda, line 17:
```suggestion
vmOpts:
  qemu:
    minimumVersion: null
```

jandubois: Feels a bit too deeply nested right now, but I guess we can put additional stuff (something like `QEMU_SYSTEM_X86_64`) under there as well in the future.

AkihiroSuda (approved):
Thanks

PR #2661, opened 2024-09-27
instConfig is redundant with inst.Config

### LINKED ISSUE #2636: The BaseDriver seems to have a redundant limayaml.LimaYAML field
https://github.com/lima-vm/lima/blob/df5d2e65bb0ac1d8050b1953743377093f1aa3a1/pkg/driver/driver.go#L75-L76

This seems to be redundant with the `store.Instance.Config` value that is accessible via the `BaseDriver.Instance` field.

If this is not redundant, then we need comments explaining the differences.

jandubois:
Fixes #2636

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -78,10 +78,6 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	if inst.Status == store.StatusStopped {
 		return fmt.Errorf("instance %q is stopped, run `limactl start %s` to start the instance", instName, instName)
 	}
-	y, err := inst.LoadYAML()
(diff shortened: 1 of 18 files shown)


afbjorklund (approved):
If there are two copies of the same struct, I'm fine with removing the redundant copy.

But I do think it will need to be refactored later, to introduce pointers for the used part.

e.g. `inst.Config.SSH` or `driver.Instance.Config`

jandubois:
> But I do think it will need to be refactored later, to introduce pointers for the used part.

There is more cleanup to be done for sure, but I wanted to keep this PR to be single-issue; it was already pretty big.

Issue #1437, opened 2023-03-27
[socketVMNet] Disable symlink detection for the socketVMNet directory


mritd:
### Description

Currently, when Lima is installed for the first time on macOS, it will automatically create a `~/.lima/_config/networks.yaml` configuration file, in which `socketVMNet` is automatically filled with a non-symlink directory.

But when we update `socketVMNet` through `brew` command, the old version directory may be deleted, which will cause Lima to fail to start.

If we set the `socketVMNet` directory to `/opt/homebrew/opt/socket_vmnet/bin/socket_vmnet`(this directory is a symlink maintained by brew.), it will also cause startup failure due to internal detection mechanism. 

Can we use some compatible code to allow the `socketVMNet` directory to use symlink?

avoidik:
do you think that the `socket_vmnet` verification should be really strict as it is now? could it be relaxed?

https://github.com/lima-vm/lima/pull/2662

jandubois:
> do you think that the `socket_vmnet` verification should be really strict as it is now? could it be relaxed?

It has already been relaxed in #1220 to make it easier for Homebrew users. I feel uneasy about that already because it is not different from allowing password-less sudo access for the admin user. Which maybe is acceptable, but it should be something the user explicitly configures, and not a side-effect of the configuration of socket_vmnet.

So if anything, I think the verification should be stricter.

You should be able to delete the `sudoers` line from `~/.lima/_config/networks.yaml` and allow password-less sudo, and the verification should be disabled already.

Just past something like `%admin ALL=(ALL) NOPASSWD:ALL` into `/etc/sudoers.d/nopasswd`.

avoidik:
as stated in the https://github.com/lima-vm/lima/pull/2662#issuecomment-2380644324 this will work if only my user is a member of the admin group, for example, if my primary group is "staff", it won't work even if my user name is in the sudoers file

jandubois:
I believe loosening the restrictions to allow ownership by `admin` instead of only `root` was a mistake, and will lobby to tighten them again.

There are ways to install/run the daemon securely as an unmanaged launchdaemon.

For convenience Lima allows you to create sudoers rules to run just the commands needed to start/stop the daemon as `root`. This is only secure if the files are owned by root. Otherwise this becomes a backdoor that allows an attacker to run **any** command on the system as root.

If this is not a concern, then you can enable password-less sudo and Lima will skip all file ownership checks. You don't even need the sudoers file. I've just tested it myself:

```console
[...]

PR #2662, opened 2024-09-27
add option to relax socket_vmnet validation

avoidik:
hello,

with this merge request I'd like to introduce an option to be able to relax socket_vmnet verification logic

regards

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -344,7 +344,7 @@ jobs:
       run: brew install qemu bash coreutils iperf3
     - name: Install socket_vmnet
       env:
-        SOCKET_VMNET_VERSION: v1.1.1
+        SOCKET_VMNET_VERSION: v1.1.4
       run: |
         (
           cd ~
--- a/pkg/networks/networks.TEMPLATE.yaml
+++ b/pkg/networks/networks.TEMPLATE.yaml
@@ -14,6 +14,7 @@ paths:
   socketVMNet: "{{.SocketVMNet}}"
   varRun: /private/var/run/lima
   sudoers: /private/etc/sudoers.d/lima
+  relaxedVerification: true
 
 group: everyone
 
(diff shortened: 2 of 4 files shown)


jandubois (changes requested):
I don't think this is needed; you can just delete the `sudoers` entry in the `network.yaml` file and enable password-less `sudo` on the machine. Please file an issue (or PR) if that isn't working.

Allowing relaxed verification is already opening the possibility of privilege escalation without a password prompt, so requiring the user to enable it makes the risk explicit.

Thanks for filing this PR though, it made me look at the code and realize that we should not even create the sudoers file unless all binaries are secure.

avoidik:
@jandubois it won't work if a user is not a member of the admin group, although is listed in the sudoers file and is an owner of the directory, here is a relevant error message (with the `sudoers` line removed in the `networks.yaml` file):

```terminal
$ limactl start --name docker-cri-aarch64 --arch aarch64 --cpus 2 --memory 4 --disk 20 --vm-type vz --tty=false template://vmnet
WARN[0000] `vmType: vz` is experimental
FATA[0000] networks.yaml field `paths.socketVMNet` error: dir "/opt/homebrew/Cellar/socket_vmnet/1.1.4" owner XXXis not an admin
$ ls -ld /opt/homebrew/Cellar/socket_vmnet/1.1.4
drwxr-xr-x@ 11 user.name  admin  352 Sep 26 21:42 /opt/homebrew/Cellar/socket_vmnet/1.1.4
$ echo $USER
user.name
$ id -u
XXX
```

this check is in this line:
https://github.com/lima-vm/lima/blob/619a4cad33af3c02dc71db82a437c4a3a908d434/pkg/networks/validate.go#L103

jandubois:
> however it then complains that the entire directories chain up to / should be owned by a root (or any user which is a member of the admin group)

Yes, this is necessary to keep this feature secure. I've shown in https://github.com/lima-vm/lima/issues/1437#issuecomment-2380858664 how you can enable password-less sudo and remove the configuration setting for the `sudoers` file. Then `limactl start` will no longer verify the security of the daemon executable because you allow execution of any binary by root anyways.

PR #2664, opened 2024-09-27
Don't create insecure sudoers file

### LINKED ISSUE #2663: limactl sudoers needs to verify the network config before writing the sudoers file
Right now the check for secure paths only happens at runtime.

But if the user generated a sudoers file to insecure binaries, then these commands could still be run by any other process that doesn't validate the binary location.

We need to protect the user from accidentally creating a sudoers file that would allow password-less sudo access to e.g. world-writable file.

jandubois:
We need to validate the paths before we generate a sudoers file.

Lima would not execute commands when they are insecure, but another app or user could once they are configured in `/etc/sudoers.d`. So it is best not to generate an insecure file at all.

Fixes #2663

### DIFF
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -46,12 +46,20 @@ func sudoersAction(cmd *cobra.Command, args []string) error {
 	if runtime.GOOS != "darwin" {
 		return errors.New("sudoers command is only supported on macOS right now")
 	}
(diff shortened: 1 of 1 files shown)


jandubois:
Sorry for the churn, but the initial PR was not correct.

I believe it is fixed now and ready for review.

AkihiroSuda (approved):
Thanks

PR #2666, opened 2024-09-29
Makefile: Improve rebuilding

### LINKED ISSUE #2682: Man pages are missing from the artifacts
### Issue: Man pages are missing from the artifacts

Man pages are not included in the release artifacts for Lima. For example, when downloading and inspecting the `lima-1.0.0-alpha.0-Darwin-arm64.tar.gz` release, there are no man pages present.

#### Steps to Reproduce:

```bash
# Download and inspect the release archive for missing man pages
$ curl -sL https://github.com/lima-vm/lima/releases/download/v1.0.0-alpha.0/lima-1.0.0-alpha.0-Darwin-arm64.tar.gz | tar -tzf - | grep -i "man"
./share/lima/templates/podman-rootful.yaml
./share/lima/templates/podman.yaml
./bin/podman.lima
```

#### Expected Behavior:
The release artifacts should include man pages for Lima commands.

#### Actual Behavior:
No man pages are found in the release artifacts.

norio-nomura:
The goal is to ensure proper rebuilding without using the `clean` target by setting appropriate dependencies for each target.
- [x] `codesign`: depends to `_output/bin/limactl$(exe)` ~~skip if `_output/bin/limactl$(exe)` has valid code signature with expected entitlements~~  
- [x] `_output/bin/limactl$(exe)`: depends to source codes, forece rebuild if build variables are differs to built binary. invoke codesign in receipe.
- [x] `_output/share/lima/lima-guestagent.Linux-%`: depends to source codes
- [x] `limactl`: stop depends on `codesign` target
- [ ] ~~`templates`:~~
- [x] symbolic links: add a symlink check to targets that create symlinks.
- [x] `manpages`, `docsy`: set a file dependency. only run when building a native `limactl`.
- [x] `_artifacts/lima-%.tar.gz`: rewrite `artifact*` targets to move archive creation to the called make subprocess. also fixes #2682
- [x] cross building: use `GOHOSTOS` on windows version check. remove `limactl.exe` if `GOOS` is not Windows. `artifact` target to prevent `guestagent` creation from being affected by `.config`. add `--no-xattrs` to `tar`. remove `clean` from `binaries`


AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

PR #2667, opened 2024-09-29
Refactor: network.YAML → network.Config

jandubois:
Because YAML is just the serialization format, and not what the data is used for.

### DIFF
--- a/pkg/vz/vm_darwin.go
+++ b/pkg/vz/vm_darwin.go
@@ -324,7 +324,7 @@ func attachNetwork(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfigu
 			}
 			configurations = append(configurations, networkConfig)
 		} else if nw.Lima != "" {
-			nwCfg, err := networks.Config()
+			nwCfg, err := networks.LoadConfig()
 			if err != nil {
 				return err
 			}
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -36,9 +36,9 @@ See %s for the usage.`, networksMD),
 		RunE:    sudoersAction,
 		GroupID: advancedCommand,
 	}
-	configFile, _ := networks.ConfigFile()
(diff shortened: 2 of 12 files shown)


pkg/vz/vm_darwin.go
```diff
@@ -353,12 +353,12 @@ func attachNetwork(driver *driver.BaseDriver, vmConfig *vz.VirtualMachineConfigu
 				if runtime.GOOS != "darwin" {
 					return fmt.Errorf("networks.yaml '%s' configuration is only supported on macOS right now", nw.Lima)
 				}
-				socketVMNetOk, err := nwCfg.IsDaemonInstalled(networks.SocketVMNet)
+				socketVMNetOk, err := config.IsDaemonInstalled(networks.SocketVMNet)
```
AkihiroSuda, line 356: nwCfg isn't fine?

jandubois: Yeah, I was unsure about it, but then decided to change it for consistency. Any time `networks.LoadConfig()` is called, the return value is stored in `config`. This made more sense to me than using different names in different places.

But I agree it is confusing inside `qemu`, which has its own `Config` type. How about calling it `config` inside the `networks` package and `nwConfig` everywhere else?

I'm also fine with using `cfg` and `nwCfg`, but again would want to use these names then consistently everywhere.

AkihiroSuda: nwCfg is more descriptive than config

jandubois: Ok, new attempt:

A `network.Config` variable is called:

* `c` when it is the receiver
* `cfg` inside the `networks` package
* `nwCfg` elsewhere

I adjusted some other "config" names in the qemu package, but not in vz, as that would modify over 100 lines of unrelated configs, which I will deal with once we agree on a naming scheme.

@AkihiroSuda PTAL

AkihiroSuda (approved):
Thanks

PR #2670, opened 2024-09-30
Rename instConfig → instCfg and inst.Config → inst.Cfg

jandubois:
Once this is merged I plan to also rename all `Instance` fields and variables to `Inst`.

Mostly for consistency, but I think the shorter field names also make the code easier to read.

### DIFF
--- a/cmd/limactl/prune.go
+++ b/cmd/limactl/prune.go
@@ -70,7 +70,7 @@ func knownLocations() (map[string]limayaml.File, error) {
 		if err != nil {
 			return nil, err
 		}
-		for k, v := range locationsFromLimaYAML(instance.Config) {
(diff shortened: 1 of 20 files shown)


AkihiroSuda:
For exported fields I prefer complete names (Config, Instance) over incomplete ones (Cfg, Inst)

jandubois:
> For exported fields I prefer complete names (Config, Instance) over incomplete ones (Cfg, Inst)

I was aiming for some consistency between local names and struct fields, and I also would prefer `drv.Inst.Cfg.Audio.Device` over `driver.Instance.Config.Audio.Device`.

But if we don't care about consistency, then there is probably no reason to rename existing variables either.

PR #2673, opened 2024-10-02
Make lima user fallback regex and cidata user validation consistent

pvdvreede:
This is a rework of #2655 as per the suggestion to use osutil user validation in cidata, rather than the other way around.


jandubois (approved):
Thanks, LGTM

Issue #2282, opened 2024-04-15
Additional provisioning formats beyond shell scripts


afbjorklund:
### Description

Currently we are using shell scripts, for the standard provisioning

It could also be useful to support other means, such as Ansible?

```yaml
provision:
- mode: ansible
  playbook: playbook.yaml
```

It would create an inventory, and then call `ansible-playbook`

* Host ("control node") requirements: Ansible

* Guest ("managed node") requirements: Python

See also:

* https://docs.ansible.com/ansible/latest/scenario_guides/guide_vagrant.html

* https://developer.hashicorp.com/vagrant/docs/provisioning/ansible ([v2.3.7](https://www.hashicorp.com/license-faq#products-covered-by-bsl))

```
╭─────────╮
⟨ ansible ⟩
╰─────────╯
        ╲   ^__^
         ╲  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||
```

PR #2676, opened 2024-10-02
Add example of ansible provisioning with playbook

afbjorklund:
Example for:

* #2282

This has the same problem with relative paths, as was fixed for the scripts:

* https://github.com/lima-vm/lima/pull/2442

```
INFO[0022] Waiting for ansible playbook "./ansible/playbook.yaml" 
ERROR! the playbook: ./ansible/playbook.yaml could not be found
```

commit 3b335496a46010a179578199c329f6e2dec42618

Currently it uses the working directory, so it only works from the git repo top.

```console
$ limactl start ./templates/ansible.yaml
```

----

With the awesome Unicode Unicow:
```
INFO[0067] Waiting for ansible playbook "./templates/ansible/playbook.yaml" 
╭──────────────────────╮
⟨ PLAY [My first play] ⟩
╰──────────────────────╯
        ╲   ^__^
         ╲  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

╭────────────────────────╮
⟨ TASK [Gathering Facts] ⟩
╰────────────────────────╯
        ╲   ^__^
         ╲  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

ok: [ansible]
╭───────────────────────────────────────────────╮
⟨ TASK [hostname : write hostname using jinja2] ⟩
╰───────────────────────────────────────────────╯
        ╲   ^__^
         ╲  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

ok: [ansible]
╭──────────────────────╮
⟨ TASK [Ping my hosts] ⟩
╰──────────────────────╯
        ╲   ^__^
         ╲  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

ok: [ansible]
╭──────────────────────╮
⟨ TASK [Print message] ⟩
╰──────────────────────╯
        ╲   ^__^
         ╲  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

ok: [ansible] => {
    "msg": "Hello world"
}
╭────────────╮
⟨ PLAY RECAP ⟩
╰────────────╯
        ╲   ^__^
         ╲  (oo)\_______
            (__)\       )\/\
                ||----w |
                ||     ||

ansible                    : ok=4    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   
```

* https://docs.ansible.com/ansible/latest/getting_started/get_started_playbook.html
* https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_templating.html

### DIFF
--- a/templates/ansible/playbook.yaml
+++ b/templates/ansible/playbook.yaml
@@ -1,0 +1,11 @@
+- name: My first play
+  hosts: all
+  tasks:
+  - name: Ping my hosts
+    ansible.builtin.ping:
+
+  - name: Print message
+    ansible.builtin.debug:
+      msg: Hello world
+  roles:
+  - hostname
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -122,7 +122,7 @@ jobs:
       run: sudo make install
     - name: Verify templates match `limactl edit` format
       run: |
-        find templates -name '*.yaml' -exec limactl edit --set 'del(.nothing)' {} \;
+        find templates -name '*.yaml' -not \( -path "templates/ansible/*" -prune \) -exec limactl edit --set 'del(.nothing)' {} \;
         git diff-index --exit-code HEAD
     - name: Uninstall
       run: sudo make uninstall
@@ -196,7 +196,7 @@ jobs:
     - name: Validate jsonschema
       run: make schema-limayaml.json
     - name: Validate templates
-      run: find -L templates -name '*.yaml' | xargs limactl validate
+      run: find -L templates -name '*.yaml' -not \( -path "templates/ansible/*" -prune \) | xargs limactl validate
     - name: Install test dependencies
       # QEMU:      required by Lima itself
       # bash:      required by test-templates.sh (OS version of bash is too old)
--- a/templates/README.md
+++ b/templates/README.md
@@ -32,6 +32,9 @@ Distro:
 - [`experimental/opensuse-tumbleweed`](./experimental/opensuse-tumbleweed.yaml): [experimental] openSUSE Tumbleweed
 - [`experimental/debian-sid`](./experimental/debian-sid.yaml): [experimental] Debian Sid
 
+Provisioning:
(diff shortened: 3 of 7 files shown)


jandubois:
Is there a way to reference environment variables from inside the playbook?

I realized that tests in #2570 don't verify that the `$PARAM_FOO` variables are visible to ansible provisioning (and probably they aren't). So it would be good if we could incorporate a test for that in the example?

afbjorklund:
> Is there a way to reference environment variables from inside the playbook?

I think you can do something like `{{ lookup('env','ENV_VAR') }}`

Then there is the `ansible_env`, but that it set rather than get

afbjorklund:
I will add a separate test for params, they are propagated but they fail to validate:

field `param` key "ANSIBLE" is not used in any provision, probe, copyToHost, or portForward

templates/ansible/playbook.yaml
AkihiroSuda, line 1: Not sure if we want to put non-lima yamls to the template store.
Maybe we should have a `examples/` directory?

jandubois: > Not sure if we want to put non-lima yamls to the template store.

We don't because they would be included in `limactl create --list-templates`, which would be confusing.

afbjorklund: It needs to be next to the template, so that it can be referenced.

```
provision:
- mode: ansible
  playbook: ./templates/ansible/playbook.yaml
```

It could probably use the new feature for scripts:

* #3072

afbjorklund:
Should be rewritten with the new feature to assemble templates.

jandubois:
> Should be rewritten with the new feature to assemble templates.

Yes, it should. Once created, instances should not rely on any external files.

Unfortunately there don't seem to be any tools to flatten playbooks; we would have to implement it ourselves. I'll create an issue for it, but I won't get around to it for quite a while.

PR #2678, opened 2024-10-03
Validate that the playbook path actually exists

afbjorklund:
Especially when using relative paths, and not absolute paths.

Also do some sanity checking on the mode and script fields.


AkihiroSuda (approved):
Thanks

Issue #2685, opened 2024-10-04
Compressing guestagent causes prompt during building


afbjorklund:
### Description

Now when not always cleaning, there is a prompt when building:

```
GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -ldflags="-s -w -X github.com/lima-vm/lima/pkg/version.Version=v1.0.0-alpha.0-180-g7e2738c9"  -o _output/share/lima/lima-guestagent.Linux-x86_64 ./cmd/lima-guestagent
chmod 644 _output/share/lima/lima-guestagent.Linux-x86_64
gzip _output/share/lima/lima-guestagent.Linux-x86_64
gzip: _output/share/lima/lima-guestagent.Linux-x86_64.gz already exists; do you wish to overwrite (y or n)? 
```

Need to use the force flag, for the gzip command (to overwrite)

* https://github.com/lima-vm/lima/pull/2666

afbjorklund:
Also the make optimization does not work, since it is always rebuilt.

afbjorklund:
The "rebuild always" is not worse now than what it was before (when make called clean)

And if it gets boring, you can disable compress during development and re-enable later

PR #2687, opened 2024-10-06
Makefile: improve compressed guestagent rebuilds

norio-nomura:
This change optimizes the rebuild of the compressed guest agent mentioned in https://github.com/lima-vm/lima/issues/2685#issuecomment-2393298548.
- Set compressed guestagents as dependencies instead of gzipping them within the guestagent recipe.
- Decompress the compressed guestagents on each rebuild to read the build configuration.


AkihiroSuda (approved):
Thanks

PR #2702, opened 2024-10-08
Add `hack/update-template-ubuntu.sh`

norio-nomura:
```console
$ hack/update-ubuntu-image.sh --help
update-ubuntu-image.sh: Update the Ubuntu image location in the specified templates

Usage:
  update-ubuntu-image.sh [--minimal|--server] [--version <version>] <template.yaml>...

Description:
  This script updates the Ubuntu image location in the specified templates.
  If the image location in the template contains a release date in the URL, the script replaces it with the latest available date.
  If no flags are specified, the script uses the flavor and version from the image location basename in the template.

  Image location basename format: ubuntu-<version>-<flavor>-cloudimg-<arch>.img

  Released Ubuntu image information is fetched from the following URLs:

    Server: https://cloud-images.ubuntu.com/releases/stream/v1/com.ubuntu.cloud:released:download.json
    Minimal: https://cloud-images.ubuntu.com/minimal/releases/stream/v1/com.ubuntu.cloud:released:download.json

  The downloaded JSON file will be cached in the Lima cache directory.

Examples:
  Update the Ubuntu image location in templates/**.yaml:
  $ update-ubuntu-image.sh templates/**.yaml

  Update the Ubuntu image location in ~/.lima/ubuntu/lima.yaml:
  $ update-ubuntu-image.sh ~/.lima/ubuntu/lima.yaml

  Update the Ubuntu image location to ubuntu-24.04-minimal-cloudimg-<arch>.img in ~/.lima/docker/lima.yaml:
  $ update-ubuntu-image.sh --minimal --version 24.04 ~/.lima/docker/lima.yaml

Flags:
  --server             Use the server flavor image
  --minimal            Use the minimal flavor image
  --version <version>  Use the specified version
  -h, --help           Print this help message
```

### DIFF
--- a/hack/cache-common-inc.sh
+++ b/hack/cache-common-inc.sh
@@ -218,13 +218,33 @@ function location_to_sha256() {
 
 # e.g.
 # ```console
+# $ cache_download_dir
+# .download # on GitHub Actions
+# /home/user/.cache/lima/download # on Linux
+# /Users/user/Library/Caches/lima/download # on macOS
+# /home/user/.cache/lima/download # on others
+# ```
+function cache_download_dir() {
+	if [[ ${GITHUB_ACTIONS:-false} == true ]]; then
+		echo ".download"
+	else
+		case "$(uname -s)" in
+		Linux) echo "${XDG_CACHE_HOME:-${HOME}/.cache}/lima/download" ;;
+		Darwin) echo "${HOME}/Library/Caches/lima/download" ;;
+		*) echo "${HOME}/.cache/lima/download" ;;
+		esac
+	fi
+}
+
+# e.g.
+# ```console
 # $ location_to_cache_path "https://cloud-images.ubuntu.com/releases/24.04/release-20240809/ubuntu-24.04-server-cloudimg-arm64.img"
 # .download/by-url-sha256/ae988d797c6de06b9c8a81a2b814904151135ccfd4616c22948057f6280477e8
 # ```
 function location_to_cache_path() {
 	local location=$1
 	[[ ${location} != "null" ]] || return
-	sha256=$(location_to_sha256 "${location}") && echo ".download/by-url-sha256/${sha256}"
+	sha256=$(location_to_sha256 "${location}") && download_dir=$(cache_download_dir) && echo "${download_dir}/by-url-sha256/${sha256}"
 }
 
 # e.g.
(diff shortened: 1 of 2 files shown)


hack/update-ubuntu-image.sh
```diff
@@ -0,0 +1,324 @@
...
+					[[ ${kernel_cmdline} != "null" ]] && image_entry=$(jq ".kernel.cmdline = \"${kernel_cmdline}\"" <<<"${image_entry}")
+					echo "${kernel_location}"
+				fi
+				if [[ -n ${initrd_location} ]]; then
+					image_entry=$(jq ". + {initrd: {location: \"${initrd_location}\"}}" <<<"${image_entry}")
+					echo "${initrd_location}"
+				fi
+			fi
+		fi
+		limactl edit --log-level error --set "
```
AkihiroSuda, line 318: Can we just use yq?

hack/update-ubuntu-image.sh
AkihiroSuda, line 1: Maybe the script should like `update-template.sh`?
I assume we can eventually add the support for non-Ubuntu templates too?

hack/update-ubuntu-image.sh
```diff
@@ -0,0 +1,324 @@
...
+
+  Update the Ubuntu image location in ~/.lima/ubuntu/lima.yaml:
+  $ $(basename "${BASH_SOURCE[0]}") ~/.lima/ubuntu/lima.yaml
+
+  Update the Ubuntu image location to ubuntu-24.04-minimal-cloudimg-<arch>.img in ~/.lima/docker/lima.yaml:
+  $ $(basename "${BASH_SOURCE[0]}") --minimal --version 24.04 ~/.lima/docker/lima.yaml
+
+Flags:
+  --server             Use the server flavor image
+  --minimal            Use the minimal flavor image
```
AkihiroSuda, line 36: This should be like `--flavor=(server|minimal)`)?
More "flavors" can be potentially added, especially when adding the support for non-Ubuntu distros.

Does `minimal` flavor really work?

hack/update-ubuntu-image.sh
jandubois: Why? Wouldn't it still essentially be the same command?

By using `limactl` we have a test case that `limactl edit` can edit template files directly, as well as instances.

norio-nomura: It's no exaggeration to say that #2593 was created for this script.

hack/update-ubuntu-image.sh
norio-nomura: > Does `minimal` flavor really work?

As far as I’ve tested, `docker.yaml` works fine even when switched to minimal. Any issues have already been fixed in the past (#2440).

hack/update-ubuntu-image.sh
norio-nomura: > I assume we can eventually add the support for non-Ubuntu templates too?

This script involves a lot of network access, so including code for other distributions would cause excessive network access during development testing. This would require flags to exclude non-development-target distributions, which would effectively be the same as splitting the script.

If you want to update everything in a single script, I think it would be better to create another script that calls all the individual update scripts for each distribution.

hack/update-ubuntu-image.sh
norio-nomura: > This should be like --flavor=(server|minimal))?

added `--flavor <flavor>` flag.

hack/update-ubuntu-image.sh
AkihiroSuda: Sure, but I guess those scripts should have some common prefix?
e.g., `update-template.ubuntu.sh`, `update-template.debian.sh`, ...

AkihiroSuda:
Could you squash the commits?

hack/update-ubuntu-image.sh
norio-nomura: changed to `update-template-ubuntu.sh` since `ls-lint` says:
```console
ls-lint
2024/10/10 08:43:05 hack/update-template.ubuntu.sh failed for rules: kebabcase
```

norio-nomura:
squashed and rebased on master.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
CI failure is irrelevant:
- https://github.com/lima-vm/lima/issues/2714

norio-nomura:
Thanks! 🙏🏻

PR #2704, opened 2024-10-09
templates: replace `alpine` with `alpine-image`

### LINKED ISSUE #2366: v1.0 roadmap: alpine: switch away from ISO to qcow2
- https://github.com/lima-vm/lima/pull/2360

In Lima v1.0, the current ISO-based `alpine` template should be ditched away, and the `alpine-image` (qcow2) template should take over the name of `alpine`

### LINKED ISSUE #2386: CI: test `template://alpine-image` (qcow2, non-ISO)
- https://github.com/lima-vm/lima/pull/2360

The new `alpine-image` template is currently not tested on the CI

AkihiroSuda:
The old ISO9660-based `alpine` template is still available as `alpine-iso`.

Fix #2366
Fix #2386


jandubois (changes requested):
Why is it moved into a `deprecated` directory; it was supposed to be called `alpine-iso.yaml` (see discussion in #2366).

AkihiroSuda:
Reverted the deprecation

AkihiroSuda:
CI ks failing around `[INFO] Testing proxy settings are imported`

examples/alpine-iso.yaml
```diff
@@ -0,0 +1,21 @@
+# This template requires Lima v0.7.0 or later.
+# Using the Alpine 3.20 aarch64 image with vmType=vz requires macOS Ventura 13.3 or later.
```
jandubois, line 2: We can probably delete this comment now.

pkg/cidata/cidata.TEMPLATE.d/boot/10-alpine-prep.sh
```diff
@@ -29,6 +29,8 @@ usermod -p '*' "${LIMA_CIDATA_USER}"
 
 # Alpine disables TCP forwarding, which is needed by the lima-guestagent
 sed -i 's/AllowTcpForwarding no/AllowTcpForwarding yes/g' /etc/ssh/sshd_config
+# Enable PAM so as to load /etc/environment via pam_env
+sed -i 's/#UsePAM no/UsePAM yes/g' /etc/ssh/sshd_config
```
jandubois, line 33: This looks unrelated to the PR (renaming the templates), but is fine because it is a tiny change and in a separate commit.

jandubois (approved):
Thanks, LGTM

examples/alpine-iso.yaml
AkihiroSuda: Why?

jandubois: I guess I was confused; I thought we required 13.3 for VZ on aarch64 for all images, but it looks like it is Alpine specific, so the comment should stay.

pkg/cidata/cidata.TEMPLATE.d/boot/10-alpine-prep.sh
jandubois: I understand. It is just something we should have had even before; it was just CI that discovered it was missing. Anyways, I already wrote that I have no problem with including it in this PR. It is ready to be merged as soon as CI is green.

AkihiroSuda:
CI failures are irrelevant. Merging,

PR #2706, opened 2024-10-09
Disable `ssh.loadDotSSHPubKeys` by default

### LINKED ISSUE #1617: Change the default value of `ssh.loadDotSSHPubKeys` from `true` to `false`

AkihiroSuda:
https://github.com/lima-vm/lima/blob/9d29da396c164fea261bc547b4537d470aa3a824/examples/default.yaml#L111-L122

This might be a breaking change for new instances, but does not affect existing instances.

Fix #1617

### DIFF
--- a/examples/default.yaml
+++ b/examples/default.yaml
@@ -148,7 +148,7 @@ ssh:
   # This option is useful when you want to use other SSH-based
   # applications such as rsync with the Lima instance.
   # If you have an insecure key under ~/.ssh, do not use this option.
-  # 🟢 Builtin default: true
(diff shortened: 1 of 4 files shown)


jandubois (approved):
Thanks, LGTM

I've confirmed that old keys on existing instances remain installed, both for `alpine` and `default` (`lima-init` and `cloud-init`).

PR #2710, opened 2024-10-09
limactl: add `tunnel` command (experimental)

AkihiroSuda:
```console
$ limactl tunnel --help
Create a tunnel for Lima

Create a SOCKS tunnel so that the host can join the guest network.

Usage:
  limactl tunnel [flags] INSTANCE

Flags:
  -h, --help             help for tunnel
      --socks-port int   SOCKS port, defaults to a random port
      --type string      Tunnel type, currently only "socks" is implemented (default "socks")

Global Flags:
      --debug               debug mode
      --log-format string   Set the logging format [text, json] (default "text")
      --log-level string    Set the logging level [trace, debug, info, warn, error]
      --tty                 Enable TUI interactions such as opening an editor. Defaults to true when stdout is a terminal. Set to false for automation. (default true)

$ limactl tunnel default
Open <System Settings> → <Network> → <Wi-Fi> (or whatever) → <Details> → <Proxies> → <SOCKS proxy>,
and specify the following configuration:
- Server: 127.0.0.1
- Port: 54940
The instance can be connected from the host as <http://lima-default.internal> via a web browser.

$ curl --proxy socks5h://127.0.0.1:54940 http://lima-default.internal
<!DOCTYPE html>
[...]
```

### DIFF
--- a/cmd/limactl/tunnel.go
+++ b/cmd/limactl/tunnel.go
@@ -1,0 +1,158 @@
+package main
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"runtime"
+	"strconv"
+
+	"github.com/lima-vm/lima/pkg/freeport"
+	"github.com/lima-vm/lima/pkg/sshutil"
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/mattn/go-shellwords"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+const tunnelHelp = `Create a tunnel for Lima
+
+Create a SOCKS tunnel so that the host can join the guest network.
+`
+
+func newTunnelCommand() *cobra.Command {
+	tunnelCmd := &cobra.Command{
+		Use:   "tunnel [flags] INSTANCE",
+		Short: "Create a tunnel for Lima",
+		PersistentPreRun: func(*cobra.Command, []string) {
+			logrus.Warn("`limactl tunnel` is experimental")
+		},
+		Long:              tunnelHelp,
+		Args:              WrapArgsError(cobra.ExactArgs(1)),
+		RunE:              tunnelAction,
+		ValidArgsFunction: tunnelBashComplete,
+		SilenceErrors:     true,
+		GroupID:           advancedCommand,
+	}
+
+	tunnelCmd.Flags().SetInterspersed(false)
+	// TODO: implement l2tp, ikev2, masque, ...
+	tunnelCmd.Flags().String("type", "socks", "Tunnel type, currently only \"socks\" is implemented")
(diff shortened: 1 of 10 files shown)


balajiv113:
Does this also allows connecting to other VM in case of user-v2 network ?

AkihiroSuda:
> Does this also allows connecting to other VM in case of user-v2 network ?

Yes

AkihiroSuda:
> The changes looks good for me for now.
> 
> I was thinking if providing 1 tunnel for every instance is confusing. Rather shouldn't we provide 1 for every network instance (like user-v2).

I wrote a POC, but seems much worse

- https://github.com/lima-vm/lima/pull/2782

> Usage:
> 
> ```
> curl \
>   --proxy socks5h://localhost/$HOME/.lima/_networks/user-v2/user-v2_socks.sock \
>   192.168.104.4
> ```
> 
> This is similar to the `limactl tunnel` proposal (PR #2710). While PR #2710 creates a proxy per an instance, this commit creates a proxy per a user-v2 network.
> 
> Remarks:
> 
> * Only works for user-v2 networks.
> * DNS lookup is not implemented yet in this POC. Could be taken from https://github.com/norouter/norouter/blob/v0.6.5/pkg/agent/socks/socks.go#L57-L75
> * https://github.com/cybozu-go/usocksd is a dependency hog (See the `go.mod` diff). Should be replaced with a fork or another library.
> 
> Overall, PR #2710 might be better than this commit, as PR #2710 works for any network driver, does not need an additional DNS resolver, and does not incur additional `go.mod` deps.

AkihiroSuda:
Maybe `limactl tunnel` may have `limactl tunnel --target=instance INSTANCE` (default?) and `limactl tunnel --target=network NETWORK` to support the both concepts

cmd/limactl/tunnel.go
```diff
@@ -0,0 +1,155 @@
...
+		fmt.Fprintf(stdout, "and specify the following configuration:\n")
+		fmt.Fprintf(stdout, "- Address: socks=127.0.0.1\n")
+		fmt.Fprintf(stdout, "- Port: %d\n", port)
+	default:
+		fmt.Fprintf(stdout, "Set `ALL_PROXY=socks5h://127.0.0.1:%d`, etc.\n", port)
+	}
+	fmt.Fprintf(stdout, "The instance can be connected from the host as <http://lima-%s.internal> via a web browser.\n", inst.Name)
+
+	// TODO: show the port in `limactl list --json` ?
+	// TODO: add `--stop` flag to shut down the tunnel
```
balajiv113, line 149: Are we planning to address these todo now or later release ??

AkihiroSuda: Updated the PR to mark the command experimental

balajiv113 (approved):
LGTM 👍

PR #2711, opened 2024-10-09
CI: vz: try macos-15-large (Intel)

AkihiroSuda:
This might be less flaky than the current macos-13?


jandubois (commented):
Looks fine to me; let's see if it helps. Might even just be that they use better hardware for the macos-15 runners than for macos-13.

jandubois (approved):

PR #2712, opened 2024-10-09
Setup RAM disk for vz tests

jandubois:
The GitHub runners have really slow disks, and the vz tests are very flaky. Maybe having better disk performance will make them less likely to fail.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -451,6 +451,11 @@ jobs:
       run: brew install bash coreutils jq
     - name: Uninstall qemu
       run: brew uninstall --ignore-dependencies --force qemu
+    - name: Setup RAM disk for ~/.lima
+      run: ./hack/setup-lima-home-ramdisk.sh
+    - name: "Adjust LIMACTL_CREATE_ARGS"
+      # use small disk that fits into the RAM disk because it will become non-sparse on conversion to raw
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
```
time="2024-10-09T22:10:59Z" level=fatal msg="failed to convert \"/Users/runner/.lima/fedora/basedisk\" to a raw disk \"/Users/runner/.lima/fedora/diffdisk\": no space left on device"
```
- #2713

jandubois:
Looks like the `default` vz test passed on the first attempt with an image size of `4` and a RAM disk size of `6`. Unfortunately this is not enough for the `fedora` image, so increasing to `5` and `7` GB respectively. Hopefully it leaves enough RAM for the rest of the system.

jandubois:
> this is not enough for the `fedora` image, so increasing to `5` and `7` GB respectively.

`default` vz test passed again, but `fedora` eventually disconnected and was skipped:

CleanShot 2024-10-09 at 17 24 03@2x

jandubois:
Trying to give the VM 2 CPUs and 2GB of RAM to see if that makes a difference. 1 CPU and 1GB instinctively feel a bit tight; maybe Fedora needs more than Ubuntu.

jandubois:
#2711 seems to have resolved the flakiness of the vz tests.

Since this PR is not obviously helpful (it works for `default` but not `fedora`) I don't think we should pursue this further right now.

PR #2715, opened 2024-10-10
nativeimgutil: fix making a sparse file

### LINKED ISSUE #2713: vz: "Expanding to 100GiB" phase makes a non-sparse 100GiB image
>  I just found out that the built-in conversion needs more diskspace than `qemu-img convert`. While the end-result is still a sparse disk, it seems to require the full 100GB disk space temporarily, so you cannot convert from QCOW2 to RAW on a device with limited free space.
> 
> ```console
> $ df -h ~/.lima3
> Filesystem    Size    Used   Avail Capacity iused ifree %iused  Mounted on
> /dev/disk5    50Gi   692Mi    49Gi     2%      11  4.3G    0%   /Users/jan/.lima3
> 
> $ l start --vm-type vz
> ? Creating an instance "default" Proceed with the current configuration
> INFO[0001] Starting the instance "default" with VM driver "vz"
> …
> INFO[0002] Converting "/Users/jan/.lima3/default/basedisk" (qcow2) to a raw disk "/Users/jan/.lima3/default/diffdisk"
> 3.50 GiB / 3.50 GiB [-------------------------------------] 100.00% 206.87 MiB/s
> INFO[0019] Expanding to 100GiB
> FATA[0020] failed to convert "/Users/jan/.lima3/default/basedisk" to a raw disk "/Users/jan/.lima3/default/diffdisk": no space left on device
> ```
> 
> 
> Using `qemu-img convert` seems to require little extra space beyond what the new sparse file actually occupies.
> 

_Originally posted by @jandubois in https://github.com/lima-vm/lima/issues/2579#issuecomment-2403566144_

> > While the end-result is still a sparse disk
> 
> Actually, it is not, with the builtin conversion. It turns into a fully allocated disk. So this is even worse. That also might explain why it takes so long: it possibly writes the full 100GB to disk.
> 

_Originally posted by @jandubois in https://github.com/lima-vm/lima/issues/2579#issuecomment-2403587553_

- - -
https://github.com/lima-vm/lima/blob/bc774ded0a0710acb6716580719ace284d9ec3d4/pkg/nativeimgutil/nativeimgutil.go#L85-L91
https://github.com/lima-vm/lima/blob/bc774ded0a0710acb6716580719ace284d9ec3d4/pkg/nativeimgutil/nativeimgutil.go#L170-L175

AkihiroSuda:
"Expanding to 100GiB" phase was making a non-sparse 100GiB image

FIx #2713

### DIFF
--- a/pkg/nativeimgutil/nativeimgutil.go
+++ b/pkg/nativeimgutil/nativeimgutil.go
@@ -14,7 +14,6 @@ import (
 	"github.com/lima-vm/go-qcow2reader"
 	"github.com/lima-vm/go-qcow2reader/image/qcow2"
 	"github.com/lima-vm/go-qcow2reader/image/raw"
-	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/progressbar"
 	"github.com/sirupsen/logrus"
 )
@@ -85,7 +84,7 @@ func ConvertToRaw(source, dest string, size *int64, allowSourceWithBackingFile b
 	// Resize
 	if size != nil {
 		logrus.Infof("Expanding to %s", units.BytesSize(float64(*size)))
-		if err = MakeSparse(destTmpF, *size); err != nil {
+		if err = destTmpF.Truncate(*size); err != nil {
 			return err
 		}
 	}
@@ -113,7 +112,7 @@ func convertRawToRaw(source, dest string, size *int64) error {
 		if err != nil {
 			return err
 		}
-		if err = MakeSparse(destF, *size); err != nil {
+		if err = destF.Truncate(*size); err != nil {
 			_ = destF.Close()
 			return err
 		}
@@ -162,14 +161,7 @@ func copySparse(w *os.File, r io.Reader, bufSize int64) (int64, error) {
 
 	// Ftruncate must be run if the file contains only zeros
 	if !hasWrites {
-		return n, MakeSparse(w, n)
+		return n, w.Truncate(n)
 	}
 	return n, nil
-}
-
-func MakeSparse(f *os.File, n int64) error {
-	if _, err := f.Seek(n, io.SeekStart); err != nil {
-		return err
-	}
-	return osutil.Ftruncate(int(f.Fd()), n)
 }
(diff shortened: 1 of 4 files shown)


pkg/nativeimgutil/nativeimgutil.go
```diff
@@ -82,16 +81,24 @@ func ConvertToRaw(source, dest string, size *int64, allowSourceWithBackingFile b
...
 	// Resize
 	if size != nil {
+		// Reopen the closed fd
+		destTmpF, err = os.OpenFile(destTmp, os.O_RDWR, 0o600)
+		if err != nil {
+			return err
+		}
 		logrus.Infof("Expanding to %s", units.BytesSize(float64(*size)))
-		if err = MakeSparse(destTmpF, *size); err != nil {
+		if err = destTmpF.Truncate(*size); err != nil {
```
jandubois, line 98: Don't you still have to close the file again?

```suggestion
		if err = destTmpF.Truncate(*size); err != nil {
			_ = destTmpF.Close()
			return err
		}
		if err = destTmpF.Close(); err != nil {
			return err
		}
```

jandubois (dismissed):
It looks good, except for the potential file handle leak.

I've tested it, and I could create a 5GB `fedora` VZ image in a 6GB RAM disk (on macOS aarch64).

pkg/nativeimgutil/nativeimgutil.go
AkihiroSuda: done

jandubois:
Sorry, I need to do more testing, but ran out of time for today. Will try again tomorrow.

It turns out that I was testing on a RAM disk formatted with HFS+, which does not support sparse files (thanks @nirs for pointing me at that possibility).

I thought I tested it again with this PR, and that it then worked, but I can no longer reproduce it. Need to retest all scenarios once more.

jandubois:
> I've tested it, and I could create a 5GB `fedora` VZ image in a 6GB RAM disk (on macOS aarch64).

Ok, I was obviously confused for that one, as that is not a valid test of creating a sparse image.

AkihiroSuda:
We may rather merge:
- #2718

Issue #2714, opened 2024-10-09
ArchLinux CI is failing since 2024-10-10 (due to the update of the `cloud-init` package)


AkihiroSuda:
**EDIT**: this is a regression in cloud-init 24.3.1-1
https://gitlab.archlinux.org/archlinux/packaging/packages/cloud-init/-/issues/3

- - -
https://github.com/lima-vm/lima/actions/runs/11263406944/job/31321212683

```
time="2024-10-09T22:01:04Z" level=info msg="[hostagent] Waiting for the essential requirement 2 of 4: \"user session is ready for ssh\""
time="2024-10-09T22:01:44Z" level=info msg="[hostagent] Waiting for the essential requirement 2 of 4: \"user session is ready for ssh\""
time="2024-10-09T22:02:24Z" level=info msg="[hostagent] Waiting for the essential requirement 2 of 4: \"user session is ready for ssh\""
time="2024-10-09T22:02:52Z" level=fatal msg="did not receive an event with the \"running\" status"
Error: ERROR] Failed to start "archlinux"
```

AkihiroSuda:
This might relate to the recent update of the cloud-init package: https://gitlab.archlinux.org/archlinux/packaging/packages/cloud-init/-/commit/08e804f7c41f7d742496aac6439ea8545d7202b6

https://github.com/canonical/cloud-init/compare/24.2...24.3.1

Potentially relevant:
- https://github.com/canonical/cloud-init/pull/5536

AkihiroSuda:
https://github.com/canonical/cloud-init/commit/143bc9e40f7f33695216db60f73e10a900d1d1cd#diff-eb94d50ed1a0a76032438b4bc5bccb389e9cc03d6766d7881463d8ee75f21f8fR25 introduced the dependency on  `nc.openbsd`, but this binary seems missing on ArchLinux

PR #2717, opened 2024-10-10
CI: temporarily remove ArchLinux (and demote it from Tier 1)

AkihiroSuda:
Workaround for:
- #2714

The "user session is ready for ssh" phase can never finishes with the cloud-init 24.3.1-1 package due to a regression that breaks execution of `/var/lib/cloud/scripts/per-boot/` scripts.

This commit also demotes ArchLinux from Tier 1; it is practically impossible to maintain good stability as the packages are too frequently updated.

After the regression is fixed we may bring ArchLinux to Tier 2, but probably never to Tier 1 again.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -221,7 +221,8 @@ jobs:
         - alpine.yaml
         - debian.yaml
         - fedora.yaml
-        - archlinux.yaml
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #2718, opened 2024-10-10
Fix nativeimgutil.ConvertToRaw()

nirs:
- Truncate before copying the data to eliminate the seeks during the
  copy. This also provides a hint to the file system that can minimize
  allocations and fragmentation of the file.

- Avoid unneeded seeks during copy using WriteAt. This does not improve
  performance since practically all time is spent on reading from the
  source image.

- Fix zero detection to handle short reads. Previously we would compare
  the entire buffer which can contain non-zero bytes from the previous
  read.

- Fix write to handle short reads. Previously we would write the entire
  buffer including data from previous read, corrupting the image.

- Fix error message for failed write. Looks like it was copied from the
  read branch.

### DIFF
--- a/pkg/nativeimgutil/nativeimgutil.go
+++ b/pkg/nativeimgutil/nativeimgutil.go
@@ -68,6 +68,13 @@ func ConvertToRaw(source, dest string, size *int64, allowSourceWithBackingFile b
 	defer os.RemoveAll(destTmp)
 	defer destTmpF.Close()
 
+	// Truncating before copy eliminates the seeks during copy and provide a
+	// hint to the file system that may minimize allocations and fragmanation
+	// of the file.
+	if err := MakeSparse(destTmpF, srcImg.Size()); err != nil {
+		return err
+	}
+
 	// Copy
 	srcImgR := io.NewSectionReader(srcImg, 0, srcImg.Size())
 	bar, err := progressbar.New(srcImg.Size())
@@ -124,8 +131,8 @@ func convertRawToRaw(source, dest string, size *int64) error {
 
 func copySparse(w *os.File, r io.Reader, bufSize int64) (int64, error) {
 	var (
-		n              int64
-		eof, hasWrites bool
+		n   int64
+		eof bool
 	)
 
 	zeroBuf := make([]byte, bufSize)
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Thanks, this seems to work, but the output is confusing
```
INFO[0000] Converting "/Users/suda/.lima/alpine/basedisk" (qcow2) to a raw disk "/Users/suda/.lima/alpine/diffdisk" 
INFO[0000] Expanding to 100GiB                          
212.00 MiB / 212.00 MiB [---------------------------------] 100.00% 115.06 MiB/s
```

It should be like
```
Expanding ...
Converting ...
[Progress bar:
```

nirs:
> Thanks, this seems to work, but the output is confusing
> 
> ```
> INFO[0000] Converting "/Users/suda/.lima/alpine/basedisk" (qcow2) to a raw disk "/Users/suda/.lima/alpine/diffdisk" 
> INFO[0000] Expanding to 100GiB                          
> 212.00 MiB / 212.00 MiB [---------------------------------] 100.00% 115.06 MiB/s
> ```
> 
> It should be like
> 
> ```
> Expanding ...
> Converting ...
> [Progress bar]
> ```

But this does not work for the raw-to-raw case, which must expand the image as a second step. I think we should keep the same flow since it is simple, easy to follow, and make it easy to verify that the convert was correct before expanding.

So I plan to do this:

1. truncate to source size
2. convert using WriteAt avoiding the seeks
3. expand

AkihiroSuda:
Do you plan to update this PR or open a follow-up PR separately ?

nirs:
> Do you plan to update this PR or open a follow-up PR separately ?

I can update it this weekend.

nirs:
@AkihiroSuda new version keeps the same user interface:

```
% limactl create --tty=false --plain 
INFO[0000] Attempting to download the image              arch=aarch64 digest="sha256:5ecac6447be66a164626744a87a27fd4e6c6606dc683e0a233870af63df4276a" location="https://cloud-images.ubuntu.com/releases/24.04/release-20240821/ubuntu-24.04-server-cloudimg-arm64.img"
Downloading the image (ubuntu-24.04-server-cloudimg-arm64.img)
551.31 MiB / 551.31 MiB [----------------------------------] 100.00% 57.27 MiB/s
INFO[0010] Downloaded the image from "https://cloud-images.ubuntu.com/releases/24.04/release-20240821/ubuntu-24.04-server-cloudimg-arm64.img" 
INFO[0010] Converting "/Users/nsoffer/.lima/default/basedisk" (qcow2) to a raw disk "/Users/nsoffer/.lima/default/diffdisk" 
3.50 GiB / 3.50 GiB [-------------------------------------] 100.00% 199.27 MiB/s
INFO[0028] Expanding to 100GiB                          
INFO[0028] Run `limactl start default` to start the instance.
```

AkihiroSuda (approved):
Thanks

PR #2723, opened 2024-10-11
Fix parallel downloads of the same image

### LINKED ISSUE #2722: Downloading images concurrently breaks in many ways
### Description

We use limactl to create multiple cluster in parallel, all using the same k8s based template. If a developer already used lima the ubuntu server cloud image is already downloaded, but if this is the first time you download the image, or maybe you delete the ~/Library/Cache/lima, the image will be downloaded in parallel.

Since all limactl processes are downloading the same to the same temporary directory
~/Library/Caches/lima/download/by-url-sha256/xxxyyy/data.tmp and then try to rename the same temporary file to the same target file, the process can fail in various ways:

- Directory not empty since other process already started the download
  ```
  2024-10-10 22:56:59,617 ERROR   [hub] failed to download "https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-arm64.img": unlinkat /Users/nir/Library/Caches/lima/download/by-url-sha256/002fbe468673695a2206b26723b1a077a71629001a5b94efd8ea1580e1c3dd06: directory not empty
  ```
- Temporary file renamed by another process
  ```
  2024-10-10 22:58:28,695 ERROR   [dr1] failed to download "https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-arm64.img": rename /Users/nir/Library/Caches/lima/download/by-url-sha256/002fbe468673695a2206b26723b1a077a71629001a5b94efd8ea1580e1c3dd06/data.tmp /Users/nir/Library/Caches/lima/download/by-url-sha256/002fbe468673695a2206b26723b1a077a71629001a5b94efd8ea1580e1c3dd06/data: no such file or directory
  ```
- Corrupted download (not sure it possible with current code)

An easy way to avoid the conflicts is to download to per-process temporary file:

    data.tmp.{pid}

When the download is finished, renaming to data is safe even with multiple processes since posix rename is atomic. If we have N processes renaming N identical downloads:

    data.tmp.123 -> data
    data.tmp.345 -> data
    data.tmp.678 -> data

All renames will succeed in unknown order, but since content is the same it does not matter.

A better but more complex way is to use a lockfile so the first process does the download, and the other processes wait on the lockfile. When the first process finish it unlock the lockfile and the next processes can grab it and find the downloaded file and continue.

Since this may happen at most once for every image, I would go with the simpler solution.

nirs:
When starting multiple vms in parallel using the same uncached yet image, we downloaded the data to the same temporary file (data.tmp) and wrote metadata to the same files (url, time, type, {digest}.digest). We could fail in many ways:

- One process removes the cache directory when another tries to rename a temporary file to the destination file.
- One process fails to remove the cache directory since another process started to download and the directory is not empty.
- Corrupting data when writing to the same file from multiple processes. Mostly likely when writing downloaded data.

Fixes:

- Do not remove the cache directory before the download. This avoids the failing to remove non-empty directory, and failure to rename a temporary file.
- Write all data to temporary files, replacing the destination file when the write is complete. This avoid corrupted data.
- Do not remove destination file when creating a new temporary file or renaming since creating a file truncate existing file and renaming replace it.

Issues:

- On windows os.Rename is not atomic, but it should work when using NTFS.
- On failures we may have lefover temporary files ({name}.tmp.{pid}).

Fixes #2722

Logs from creating 10 vms in parallel: 
[test-parallel-download.tar.gz](https://github.com/user-attachments/files/17336484/test-parallel-download.tar.gz)

### DIFF
--- a/pkg/downloader/downloader.go
+++ b/pkg/downloader/downloader.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"time"
 
@@ -262,14 +263,11 @@ func Download(ctx context.Context, local, remote string, opts ...Opt) (*Result, 
 			return res, nil
 		}
 	}
-	if err := os.RemoveAll(shad); err != nil {
-		return nil, err
-	}
 	if err := os.MkdirAll(shad, 0o700); err != nil {
 		return nil, err
 	}
 	shadURL := filepath.Join(shad, "url")
-	if err := os.WriteFile(shadURL, []byte(remote), 0o644); err != nil {
+	if err := atomicWrite(shadURL, []byte(remote), 0o644); err != nil {
 		return nil, err
 	}
 	if err := downloadHTTP(ctx, shadData, shadTime, shadType, remote, o.description, o.expectedDigest); err != nil {
@@ -280,7 +278,7 @@ func Download(ctx context.Context, local, remote string, opts ...Opt) (*Result, 
 		return nil, err
 	}
 	if shadDigest != "" && o.expectedDigest != "" {
-		if err := os.WriteFile(shadDigest, []byte(o.expectedDigest.String()), 0o644); err != nil {
+		if err := atomicWrite(shadDigest, []byte(o.expectedDigest.String()), 0o644); err != nil {
 			return nil, err
 		}
 	}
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks, can we have unit tests?

pkg/downloader/downloader.go
```diff
@@ -600,10 +598,8 @@ func downloadHTTP(ctx context.Context, localPath, lastModified, contentType, url
 		return fmt.Errorf("downloadHTTP: got empty localPath")
 	}
 	logrus.Debugf("downloading %q into %q", url, localPath)
-	localPathTmp := localPath + ".tmp"
-	if err := os.RemoveAll(localPathTmp); err != nil {
-		return err
-	}
+
+	localPathTmp := perProcessTempfile(localPath)
```
AkihiroSuda, line 602: Can we `defer os.RemoveAll(localPathTmp)` to minimize the possibility of leaking the tmp file?

AkihiroSuda: perProcessTempfile is a new code though

AkihiroSuda:
SGTM

nirs:
> Thanks, can we have unit tests?

Here: https://github.com/lima-vm/lima/pull/2728

PR #2726, opened 2024-10-11
Refactor limayaml.Save and store.SaveYAML → limayaml.Marshal

jandubois:
`store.SaveYAML` was the only caller of `limayaml.Save`, and was only called from `main.validateAction`.

The name `Save` implies writing to a file, so `Marshal` seems like a more logical choice for the name. And since it is marshalling `limayaml` structs it should be in the `limayaml` package and not in `store`.

Also move `limayaml.unmarshalYAML` from `load.go` to `marshal.go` for consistency.

<hr>

I was wondering if `load.go` and `marshal.go` should be combined into a single file, but kept them separate for now.

I also feel like `store.LoadYAMLByFilePath` should probably belong in the `limayaml` package, but am not sure.


AkihiroSuda (approved):
Thanks

PR #2727, opened 2024-10-11
Always delete downloader temporary data file

nirs:
If the download failed, the temporary data file (data.tmp.pid) was not deleted. Ensure it is delete in all cases.

Also move the creation of the temporary file right before we need it. The chance that creating a temporary file will fail is practically zero so there is no need to do this upfront. This makes the code easier to follow and maintain.


jandubois (approved):
Thanks, LGTM

PR #2729, opened 2024-10-13
Fix locktuil on windows

nirs:
Add the missing test, revealing that the windows implementation does not match the unix one. It allows multiple threads to acquire the lock at the same time.

Fix the lock flags to create a blocking exclusive lock, and improve the code to prevent future errors.

Tested using the unit tests.

@AkihiroSuda Looks like this code was copied from nerdctl and it needs the same test and fix:
https://github.com/containerd/nerdctl/tree/v0.13.0/pkg/lockutil

### DIFF
--- a/pkg/lockutil/lockutil_windows.go
+++ b/pkg/lockutil/lockutil_windows.go
@@ -33,20 +33,37 @@ var (
 	procUnlockFileEx = modkernel32.NewProc("UnlockFileEx")
 )
 
+const (
+	// see https://msdn.microsoft.com/en-us/library/windows/desktop/aa365203(v=vs.85).aspx
+	LOCKFILE_EXCLUSIVE_LOCK   = 0x00000002
+	LOCKFILE_FAIL_IMMEDIATELY = 0x00000001
+)
+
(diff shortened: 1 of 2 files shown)


pkg/lockutil/lockutil_windows.go
```diff
@@ -33,20 +33,37 @@ var (
 	procUnlockFileEx = modkernel32.NewProc("UnlockFileEx")
 )
 
+const (
+	// see https://msdn.microsoft.com/en-us/library/windows/desktop/aa365203(v=vs.85).aspx
+	LOCKFILE_EXCLUSIVE_LOCK   = 0x00000002
+	LOCKFILE_FAIL_IMMEDIATELY = 0x00000001
```
nirs, line 39: I used the same constant names used in windows docs to make it easy to find the flag in the docs or using search.

I see that the original code form bolt use nicer go style:
https://github.com/boltdb/bolt/blob/2f1ce7a837dcb8da3ec595b1dac9d0632f0f99e8/bolt_windows.go#L18

We can use this style.

AkihiroSuda (approved):
Thanks

PR #2730, opened 2024-10-13
templates: merge `experimental/{riscv64,armv7l}` into `default`

AkihiroSuda:
These archs still remain experimental but can be merged into the `default` template.

`.containerd.user` now defaults to `false` for these archs, as `nerdctl-full` archive is still missing for them.


cmd/limactl/start.go
```diff
@@ -131,6 +131,12 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string, createOnly bool) (*
 		// No need to use SecureJoin here. https://github.com/lima-vm/lima/pull/805#discussion_r853411702
 		templateName := filepath.Join(templateURL.Host, templateURL.Path)
 		logrus.Debugf("interpreting argument %q as a template name %q", arg, templateName)
+		switch templateName {
+		case "experimental/riscv64":
+			logrus.Warn("template://experimental/riscv64 was merged into the default template in Lima v1.0. Use `limactl create --arch=riscv64 template://default` instead.")
+		case "experimental/armv7l":
+			logrus.Warn("template://experimental/armv7l was merged into the default template in Lima v1.0. Use `limactl create --arch=armv7l template://default` instead.")
+		}
```
AkihiroSuda, line 139: I'll also remove `experimental/{9p,virtiofs-linux,vz}` in favor of the limactl CLI flags, in a separate PR

examples/default.yaml
```diff
@@ -20,7 +20,7 @@ vmOpts:
   qemu:
     # Minimum version of QEMU required to create an instance of this template.
     # Will be ignored if the vmType is not "qemu"
-    # 🟢 Builtin default: not set
+    # 🟢 Builtin default: not set (but RISC-V requires QEMU 9.1 or later)
     minimumVersion: null
```
AkihiroSuda, line 24: Maybe `minimumVersion` should be typed as `map[Arch] bool`? (in a separate PR)

AkihiroSuda: (Anyway riscv64 is experimental; I don't want to spend too much time for overengineering it)

AkihiroSuda: > I think it worked on earlier versions as well, as long as you supplied the firmware and kernel?

Yes, but s/and/or/

> Maybe the qemu implementation details can stay out of the default configuration comments.

Done

jandubois (approved):
Thanks, LGTM

Issue #2580, opened 2024-09-01
Unneeded convert and actual copy of basedisk for every instance when using qcow2 image


nirs:
### Description

With the current scheme on macOS, each time I create a vz vm from the same cached image, there is a slow convert from qcow2 to raw (#2579) to the instance basedisk. Then we create a fast copy-on-write copy for vz for the diffdisk.

So we end with:
- slow convert
- actual copy of the base disk for every instance

Instead we can convert the qcow2 image once after the download and. Then create basedisk as a fast copy-on-write copy of the cached image. 

Suggested layout:

- cache
  - ubuntu-xyz.img (converted from ubunti-xyz.qcow2 after download) 
- instance1 (vz)
  basedisk.img (fast copy of ../_templates/ubuntu-xyz.img)
  diffdisk.img (fast copy of basedisk.img)
- instance2 (qemu)
  basedisk.img (fast copy of ../_templates/ubuntu-xyz.img)
  diffdisk.img (fast qcow2 image using basedisk.img as backing file)

Of course this works only on file systems providing copy-on-write. When we don't have copy-on-write we can do actual copies.

For qemu we can use qcow2 images using a shared template as backing file, but is very hard to managed correctly. You must ensure that template images are never deleted or changed while an instance image used them as a backing file. If we fail, the instance image will be corrupted when you start the instance.

jandubois:
I think the simple solution would be to keep the converted image in the regular cache as well, so all instances use a copy-on-write copy of the cached image.

We do need to figure out how to deal with verifying checksums, but maybe we should just trust the conversion and only verify that the checksum of the original image matches.

nirs:
Great question!

For verifying checksums we can use blksum:
https://gitlab.com/nirs/blkhash

It requires libnbd and qemu-nbd for reading qcow2 images (or any other format), but if you have your own qcow2 reader you can use the blkhash library via cgo.

But if want an easy solution with existing tools you can do this:

```
curl ... -o xxx.qcow2
qemu-img convert -f qcow2 -O raw xxx.qcow2 xxx.raw
qemu-img compare xxx.qcow2 xxx.raw
```

The checksum published for the qcow2 image can be used to verify the download, but it is useless for verifying the converted image, since it is a checksum of the actual qcow2 image and not of the content of the image.

qemu-img compares image content seen by the guest, regardless of the image format (including compression). It skips unallocated areas so you read and compare only the actual data, and you don't pay for hash computation.

[...]

PR #2736, opened 2024-10-15
templates: remove `vz`, `vmnet`, etc. in favor of the limactl CLI args

AkihiroSuda:
- `experimental/vz`: Merged into the default template in Lima v1.0. See also <https://lima-vm.io/docs/config/vmtype/>.
- `vmnet`: Removed in Lima v1.0. Use `limactl create --network=lima:shared template://default` instead. See also <https://lima-vm.io/docs/config/network/>."
- `experimental/net-user-v2`: Removed in Lima v1.0. Use `limactl create --network=lima:user-v2 template://default` instead. See also <https://lima-vm.io/docs/config/network/>."
- `experimental/9p`: Removed in Lima v1.0. Use `limactl create --vm-type=qemu --mount-type=9p template://default` instead. See also <https://lima-vm.io/docs/config/mount/>."
- `experimental/virtiofs-linux`: Removed in Lima v1.0. Use `limactl create --mount-type=virtiofs template://default` instead. See also <https://lima-vm.io/docs/config/mount/>."


hack/test-templates/test-misc.yaml
```diff
@@ -58,3 +58,6 @@ probes:
 # $ limactl disk create data --size 10G
 additionalDisks:
 - "data"
+
+-networks:
```
jandubois, line 62:
```suggestion
networks:
```

AkihiroSuda: Thanks, fixed

jandubois:
@AkihiroSuda Did you intend to include all the `cidata` changes in the last force-push? They seem unrelated to this PR.

hack/test-templates.sh
```diff
@@ -81,6 +74,21 @@ if limactl ls -q | grep -q "$NAME"; then
 	exit 1
 fi
 
+# Create ${NAME}-tmp to inspect the enabled features.
+# TODO: skip downloading and converting the image here?
```
jandubois, line 78: Why? We are about to create an instance for real right afterwards anyways, so it's not like we don't need the image.

jandubois (dismissed):
Thanks, LGTM

hack/test-templates.sh
AkihiroSuda: Probably we should have `limactl create --dry-run-output=<YAML or JSON file path>`, but it is beyond the scope of this PR

jandubois: When/If I get around to implement the `basedOn` mechanism, I was planning to add a `limactl template` command that would output the combined template. It might make sense to add the `--set` and related commands to it as well. But that is now a hypothetical depending on a hypothetical... 😄 

The name was inspired by `helm template`.

jandubois: > This instance is removed in a second without starting it, so it doesn't need the image to be present here.

Yes, but you will create another instance based on the same image again afterwards. So all you would save is the local copy / conversion. I think it is not worth worrying about for CI.

AkihiroSuda: (The conversion is not cached though)
- https://github.com/lima-vm/lima/issues/2580

AkihiroSuda:
Weird failure
https://github.com/lima-vm/lima/actions/runs/11351964745/job/31577434611?pr=2736

```json
{"level":"info","msg":"Starting QEMU (hint: to watch the boot progress, see \"/home/runner/.lima/test-misc-1/serial*.log\")","time":"2024-10-15T19:42:20Z"}
{"level":"debug","msg":"qCmd.Args: []","time":"2024-10-15T19:42:20Z"}
{"level":"fatal","msg":"exec: no command","time":"2024-10-15T19:42:20Z"}
```

jandubois:
Failure:

```
ERROR: Ansible requires blocking IO on stdin/stdout/stderr. Non-blocking file handles detected: <stderr>
```

See also https://github.com/ansible/ansible/issues/80108

AkihiroSuda:
> Failure:
> 
> ```
> ERROR: Ansible requires blocking IO on stdin/stdout/stderr. Non-blocking file handles detected: <stderr>
> ```
> 
> See also [ansible/ansible#80108](https://github.com/ansible/ansible/issues/80108)

What should we do with this?
Is this relevant to merging net-user-v2.yaml into test-misc.yaml ?

jandubois:
> What should we do with this?
> Is this relevant to merging net-user-v2.yaml into test-misc.yaml ?

No, it is not. If re-running the test gets rid of the error, then let's merge this.

AkihiroSuda:
Green, but took 52 mins (usually 20 mins)

AkihiroSuda:
Decoupled `net-user-v2` from `test-misc` again

PR #2742, opened 2024-10-16
Add OSSF scorecard badge to README.md

jandubois:


AkihiroSuda (approved):
Thanks

AkihiroSuda:
Maybe we should improve the score before adding the badge to README

jandubois:
> Maybe we should improve the score before adding the badge to README

I think the score is already pretty awesome. Once we fix the token permissions the score should be over 9. Will look into that tomorrow, unless you beat me to it. 😉

PR #2748, opened 2024-10-16
Test workflow should not need any write permissions

jandubois:
Not sure why the StepSecurity bot is ignoring the test workflow.


AkihiroSuda (approved):
Thanks

PR #2758, opened 2024-10-17
Remove experimental notice for vz, 9p, user-v2 net, grpc forwarder

AkihiroSuda:


examples/default.yaml
```diff
@@ -126,8 +126,8 @@ mounts:
 # 🟢 Builtin default: null
 mountTypesUnsupported: null
 
-# Mount type for above mounts, such as "reverse-sshfs" (from sshocker), "9p" (EXPERIMENTAL until Lima v1.0, from QEMU’s virtio-9p-pci, aka virtfs),
-# or "virtiofs" (EXPERIMENTAL, needs `vmType: vz`; will graduate from experimental in Lima v1.0)
+# Mount type for above mounts, such as "reverse-sshfs" (from sshocker), "9p" (QEMU’s virtio-9p-pci, aka virtfs),
+# or "virtiofs".
```
jandubois, line 130: Don't we want to keep the comment that `virtiofs` needs `vmType: vz`?

examples/default.yaml
```diff
@@ -296,7 +296,7 @@ cpuType:
 #   x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
 
 rosetta:
-  # Enable Rosetta for Linux (EXPERIMENTAL; will graduate from experimental in Lima v1.0).
+  # Enable Rosetta for Linux
```
jandubois, line 299:
```suggestion
  # Enable Rosetta inside the VM; needs `vmType: vz`
```

website/content/en/docs/config/mount/_index.md
```diff
@@ -12,7 +12,7 @@ The default mount type is shown in the following table:
 | < 0.10           | reverse-sshfs + Builtin SFTP server                           |
 | >= 0.10          | reverse-sshfs + OpenSSH SFTP server                           |
 | >= 0.17          | reverse-sshfs + OpenSSH SFTP server for QEMU, virtiofs for VZ |
-| >= 1.0 (Planned) | 9p for QEMU, virtiofs for VZ                                  |
+| >= 1.0           | 9p for QEMU, virtiofs for VZ                                  |
```
jandubois, line 15: You could shrink the width of the whole column now by 4 characters, if you want to.

jandubois (changes requested):
LGTM, but I think we need to keep the hint that `virtiofs` requires `vz`.

examples/default.yaml
AkihiroSuda: No, on Linux

jandubois:
```suggestion
# or "virtiofs" (needs `vmType: vz` on macOS).
```

AkihiroSuda: Added "needs vz on macOS"

examples/default.yaml
AkihiroSuda: done

examples/default.yaml
```diff
@@ -126,8 +126,8 @@ mounts:
 # 🟢 Builtin default: null
 mountTypesUnsupported: null
 
-# Mount type for above mounts, such as "reverse-sshfs" (from sshocker), "9p" (EXPERIMENTAL until Lima v1.0, from QEMU’s virtio-9p-pci, aka virtfs),
-# or "virtiofs" (EXPERIMENTAL, needs `vmType: vz`; will graduate from experimental in Lima v1.0)
+# Mount type for above mounts, such as "reverse-sshfs" (from sshocker), "9p" (QEMU’s virtio-9p-pci, aka virtfs),
+# or "virtiofs" (needs `vmType: vz` on macOS).
```
jandubois, line 130: Just realized that `virtiofs` is still experimental on Linux; do you want to include that here?

```suggestion
# or "virtiofs" (experimental on Linux; needs `vmType: vz` on macOS).
```

jandubois (approved):
Thanks, LGTM

examples/default.yaml
AkihiroSuda: done

jandubois (approved):
Thanks, LGTM

PR #2761, opened 2024-10-17
Upgrade go-qcow2reader to v0.2.0

nirs:
This version is up to 36 times faster than previous version.


nirs:
I see that this is handle automatically by #2760 :-)

AkihiroSuda:
Thanks, but this is a duplicate of
- #2760

PR #2763, opened 2024-10-18
limactl shell: ask whether to start the instance if not running

AkihiroSuda:
Before
```console
$ lima
FATA[0000] instance "default" does not exist, run `limactl create default` to create a new instance
```

After
```console
$ lima
ERRO[0000] instance "default" does not exist, run `limactl create default` to create a new instance 
? Do you want to create and start the instance "default" using the default template now? Yes
[...]
INFO[0114] READY.
suda@lima-default:/Users/suda/gopath/src/github.com/lima-vm/lima$
```

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -51,6 +51,7 @@ func newShellCommand() *cobra.Command {
 }
 
 func shellAction(cmd *cobra.Command, args []string) error {
+	os.Setenv("_LIMACTL_SHELL_IN_ACTION", "")
 	// simulate the behavior of double dash
 	newArg := []string{}
 	if len(args) >= 2 && args[1] == "--" {
@@ -71,12 +72,24 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	inst, err := store.Inspect(instName)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return fmt.Errorf("instance %q does not exist, run `limactl create %s` to create a new instance", instName, instName)
+			_, err = startInteractive(cmd, instName, true, true)
+			if err != nil {
+				return err
+			}
+			inst, err = store.Inspect(instName)
 		}
-		return err
+		if err != nil {
+			return err
+		}
 	}
 	if inst.Status == store.StatusStopped {
-		return fmt.Errorf("instance %q is stopped, run `limactl start %s` to start the instance", instName, instName)
+		if _, err = startInteractive(cmd, instName, false, true); err != nil {
+			return err
+		}
+		inst, err = store.Inspect(instName)
+		if err != nil {
+			return err
+		}
 	}
 
 	// When workDir is explicitly set, the shell MUST have workDir as the cwd, or exit with an error.
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -529,3 +529,55 @@ func startBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobr
 	compTmpl, _ := bashCompleteTemplateNames(cmd)
 	return append(compInst, compTmpl...), cobra.ShellCompDirectiveDefault
 }
+
+// startInteractive starts the instance after a confirmation prompt.
+//
+// If newInst is set to true, this function creates the instance too.
+//
+// If mandatory is set to true, and the answer to the confirmation prompt is No,
+// the function returns an error that contains an instruction to start the instance manually.
+func startInteractive(cmd *cobra.Command, instName string, newInst, mandatory bool) (bool, error) {
+	flags := cmd.Flags()
+	tty, err := flags.GetBool("tty")
+	if err != nil {
+		return false, err
+	}
+	errS := fmt.Sprintf("instance %q is stopped, run `limactl start %s` to start the instance", instName, instName)
+	if newInst {
+		errS = fmt.Sprintf("instance %q does not exist, run `limactl create %s` to create a new instance", instName, instName)
+	}
+	if !tty {
+		if mandatory {
+			return false, errors.New(errS)
+		}
+		return false, nil
+	}
+	logrus.Error(errS)
+
+	message := fmt.Sprintf("Do you want to start the instance %q now?", instName)
+	if newInst {
+		message = fmt.Sprintf("Do you want to create and start the instance %q using the default template now?", instName)
+	}
+	ans, err := uiutil.Confirm(message, true)
+	if err != nil {
+		return false, err
+	}
+	if !ans {
+		if mandatory {
+			return ans, errors.New(errS)
+		}
+		return ans, nil
+	}
+	if newInst {
+		rootCmd := cmd.Root()
+		// The create command shows the template chooser UI, etc.
+		rootCmd.SetArgs([]string{"create", instName})
+		if err := rootCmd.Execute(); err != nil {
+			return ans, err
+		}
+	}
+	rootCmd := cmd.Root()
+	// The start command reconciliates the networks, etc.
+	rootCmd.SetArgs([]string{"start", instName})
+	return ans, rootCmd.Execute()
+}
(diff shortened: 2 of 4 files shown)


afbjorklund:
I think the improvement of the "lima" command is good, but haven't looked at the code yet...

We might need something similar for the shell wrappers, like docker.lima and podman.lima?

```
instance "docker" does not exist, run `limactl create --name=docker template://docker` to create a new instance
```

```
instance "podman" does not exist, run `limactl create --name=podman template://podman` to create a new instance
```

The nerdctl.lima wrapper should be good, since it only calls `lima` (which calls `limactl shell`)

It could need some work, to get the same "user interface" in the bash script (as with survey)

AkihiroSuda:
> We might need something similar for the shell wrappers, like docker.lima and podman.lima?

Yes, but it is beyond the scope of this PR

afbjorklund:
The current user experience is actually **two** steps, one to create and one to start:

```console
$ lima
FATA[0000] instance "default" does not exist, run `limactl create default` to create a new instance 
$ limactl create default
...
INFO[0003] Run `limactl start default` to start the instance. 
```

```console
$ lima
FATA[0000] instance "default" is stopped, run `limactl start default` to start the instance 
$ limactl start default
...
INFO[0068] READY. Run `lima` to open the shell.         
```

Or maybe three steps, considering that you still have to run the original `lima` command again?

So it's a good thing reducing 3 to 1.

afbjorklund:
> Yes, but it is beyond the scope of this PR

Didn't find any issue for the entire change.

cmd/limactl/start.go
```diff
@@ -529,3 +529,55 @@ func startBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobr
...
 	return append(compInst, compTmpl...), cobra.ShellCompDirectiveDefault
 }
+
+// startWithAsking starts the instance after a confirmation prompt.
+//
+// If newInst is set to true, this function creates the instance too.
+//
+// If mandatory is set to true, and the answer to the confirmation prompt is No,
+// the function returns an error that contains an instruction to start the instance manually.
+func startWithAsking(cmd *cobra.Command, instName string, newInst, mandatory bool) (bool, error) {
```
nirs, line 539: Maybe startInteractive?

cmd/limactl/start.go
```diff
@@ -529,3 +529,55 @@ func startBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobr
 	compTmpl, _ := bashCompleteTemplateNames(cmd)
 	return append(compInst, compTmpl...), cobra.ShellCompDirectiveDefault
 }
+
+// startWithAsking starts the instance after a confirmation prompt.
+//
+// If newInst is set to true, this function creates the instance too.
+//
+// If mandatory is set to true, and the answer to the confirmation prompt is No,
+// the function returns an error that contains an instruction to start the instance manually.
```
nirs, line 538: The mandatory flag feels to complex. If we return a sentinel error like ErrUserCanceled the caller can print a helpful message without complicating this function.

cmd/limactl/start.go
```diff
@@ -529,3 +529,55 @@ func startBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobr
...
+		if mandatory {
+			return false, errors.New(errS)
+		}
+		return false, nil
+	}
+	logrus.Error(errS)
+
+	message := fmt.Sprintf("Do you want to start the instance %q now?", instName)
+	if newInst {
+		message = fmt.Sprintf("Do you want to create and start the instance %q using the default template now?", instName)
```
nirs, line 559: `Using the default template` is relevant only when file or template:// are not specified, but this used here without considering the the arguments.

afbjorklund: If this was a parameter, the same approach could be used for docker and podman - and not only default

cmd/limactl/shell.go
```diff
@@ -51,6 +51,7 @@ func newShellCommand() *cobra.Command {
 }
 
 func shellAction(cmd *cobra.Command, args []string) error {
+	os.Setenv("_LIMACTL_SHELL_IN_ACTION", "")
```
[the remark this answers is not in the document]
AkihiroSuda: Because this is a "private" variable

cmd/limactl/start.go
AkihiroSuda: The code path is only reachable from `limactl shell`, and it does not take a template as an argument.

cmd/limactl/start.go
AkihiroSuda: renamed

cmd/limactl/start.go
AkihiroSuda: Tried but I felt it was even more complicated

AkihiroSuda:
What should we do with this?
@lima-vm/maintainers

jandubois:
> What should we do with this?

I'll try to take a look soon. I hope we can get this into 1.0.

cmd/limactl/start.go
```diff
@@ -529,3 +529,55 @@ func startBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobr
...
 	return append(compInst, compTmpl...), cobra.ShellCompDirectiveDefault
 }
+
+// startInteractive starts the instance after a confirmation prompt.
+//
+// If newInst is set to true, this function creates the instance too.
+//
+// If mandatory is set to true, and the answer to the confirmation prompt is No,
+// the function returns an error that contains an instruction to start the instance manually.
+func startInteractive(cmd *cobra.Command, instName string, newInst, mandatory bool) (bool, error) {
```
jandubois, line 539: It is not clear what the `bool` return value represents. It also seems to be completely unused; all 3 call-sites are `_, err = startInteractive(…)`.

cmd/limactl/start.go
```diff
@@ -529,3 +529,55 @@ func startBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobr
...
+		return false, err
+	}
+	if !ans {
+		if mandatory {
+			return ans, errors.New(errS)
+		}
+		return ans, nil
+	}
+	if newInst {
+		rootCmd := cmd.Root()
```
jandubois, line 572: `rootCmd` could be setup outside the conditional because it is needed again later.

cmd/limactl/start.go
```diff
@@ -529,3 +529,55 @@ func startBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobr
...
+		// The create command shows the template chooser UI, etc.
```
jandubois, line 573: I would have expected that the template would default to the instance name, if a template existed with that name. Otherwise it should show an error and only then default to `default`.

```console
$ l shell alpine
ERRO[0000] instance "alpine" does not exist, run `limactl create alpine` to create a new instance
? Do you want to create and start the instance "alpine" using the default template now?
```

This should say "using the alpine template".

jandubois: I agree that the `mandatory` mechanism is too complex/confusing. It also results in the same error being displayed twice, first as `ERRO` and then as `FATA`:

```console
$ l shell alpine
ERRO[0000] instance "alpine" does not exist, run `limactl create alpine` to create a new instance
? Do you want to create and start the instance "alpine" using the default template now? No
FATA[0002] instance "alpine" does not exist, run `limactl create alpine` to create a new instance
```

I think the `mandatory` argument can just be removed, and the function always just returns `nil` when the user elected not to create/start the instance.

The caller is already calling `store.Inspect()` afterwards, and can then simply exit if the instance isn't in running state.

jandubois (changes requested):
Sorry, this isn't a complete review, but I think the `mandatory` logic is not working correctly (it shows error messages multiple times).

I'm suggesting an alternative, but I'm not totally sure if it will work; I would need to try it myself. Let me know if you want me to make an attempt!

jandubois:
> Let me know if you want me to make an attempt!

Ok, I went ahead and created #2832, which I think is simpler to follow. LMK what you think!

Issue #2707, opened 2024-10-09
`hack/test-port-forwarding.pl` should print a human-readable error when `nc` is not installed on the host


AkihiroSuda:
Currently the script just prints `❌ Forwarding TCP ...` when `nc` is not installed on the host

PR #2764, opened 2024-10-18
[`test-port-forwarding.pl`] fix: add  human-readable error when nc is not installed on the host

Mr-Sunglasses:
fix: #2707

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -153,19 +153,22 @@ jobs:
       run: go test -v ./...
     - name: Make
       run: make
-    - name: Smoke test
-      # Make sure the path is set properly and then run limactl
-      run: |
-        $env:Path = 'C:\Program Files\Git\usr\bin;' + $env:Path
-        Set-ItemProperty -Path 'Registry::HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment' -Name PATH -Value $env:Path
-        .\_output\bin\limactl.exe start template://experimental/wsl2
-      # TODO: run the full integration tests
-    - name: Debug
-      if: always()
-      run: type C:\Users\runneradmin\.lima\wsl2\ha.stdout.log
-    - name: Debug
-      if: always()
-      run: type C:\Users\runneradmin\.lima\wsl2\ha.stderr.log
+# FIXME: Windows CI began to fail on Oct 21, 2024.
+# Something seems to have changed between win22/20241006.1 and win22/20241015.1.
+# https://github.com/lima-vm/lima/issues/2775
+#    - name: Smoke test
+#      # Make sure the path is set properly and then run limactl
+#      run: |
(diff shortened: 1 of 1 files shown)


hack/test-port-forwarding.pl
```diff
@@ -20,8 +20,24 @@
...
+            "  - On macOS: brew install netcat\n" .
+            "  - On Ubuntu/Debian: sudo apt-get install netcat\n" .
+            "  - On RHEL/CentOS: sudo yum install nmap-ncat\n";
+    }
+}
+
 my $instance = shift;
 
+# Check for netcat before proceeding
+check_netcat_installed() unless -f $instance;
```
jandubois, line 39: Why not move the check after the `if (-f $instance)` block below, so you don't need to repeat the check?

And given that the whole file is just one long script, I think the code should just be inlined and not put into a function.

hack/test-port-forwarding.pl
```diff
@@ -20,8 +20,24 @@
...
+# check for nc
+sub check_netcat_installed {
+    my $nc_path = `which nc 2>/dev/null`;
+    chomp $nc_path;
+    unless ($nc_path) {
+        die "Error: 'nc' (netcat) is not installed on the host system.\n" .
+            "Please install netcat to run this test script:\n" .
+            "  - On macOS: brew install netcat\n" .
+            "  - On Ubuntu/Debian: sudo apt-get install netcat\n" .
+            "  - On RHEL/CentOS: sudo yum install nmap-ncat\n";
```
jandubois, line 32: It doesn't really matter, but instead of all the explicit newlines and concatenation you could use a here-document:

```suggestion
        die <<"end_of_message";
Error: 'nc' (netcat) is not installed on the host system.
Please install netcat to run this test script:
- On macOS: brew install netcat
- On Ubuntu/Debian: sudo apt-get install netcat
- On RHEL/CentOS: sudo yum install nmap-ncat
end_of_message
```

jandubois (changes requested):
I don't really care either way about the here-document, but the repeated test for `-f $instance` should be eliminated.

hack/test-port-forwarding.pl
```diff
@@ -20,8 +20,24 @@
 use Socket qw(inet_ntoa);
 use Sys::Hostname qw(hostname);
 
+# check for nc
+sub check_netcat_installed {
+    my $nc_path = `which nc 2>/dev/null`;
```
AkihiroSuda, line 25: `which` is often missing. `command -v` should work.

Mr-Sunglasses: Done 😄

hack/test-port-forwarding.pl
Mr-Sunglasses: Done 😄

hack/test-port-forwarding.pl
Mr-Sunglasses: Done 😄

jandubois:
Please squash commits

PR #2769, opened 2024-10-20
Validate the jsonschema against default.yaml

### LINKED ISSUE #2768: jsonschema should have tests
> Can we have a test? What happens when this is out of the sync with the YAML struct definition?

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/2306#discussion_r1807391382_

afbjorklund:
We don't want to use the external location of the schema*, and prefer to not use the external python tool.

\* that doesn't even exist yet https://www.schemastore.org/json/

https://pypi.org/project/check-jsonschema/ `check-jsonschema`

So make it possible for the generator to validate that the generated schema matches the `default.yaml`.

Closes #2768

### DIFF
--- a/cmd/limactl/genschema.go
+++ b/cmd/limactl/genschema.go
@@ -3,21 +3,27 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 
+	"github.com/goccy/go-yaml"
 	"github.com/invopop/jsonschema"
 	"github.com/lima-vm/lima/pkg/limayaml"
+	jsonschema2 "github.com/santhosh-tekuri/jsonschema/v6"
 	"github.com/spf13/cobra"
 	orderedmap "github.com/wk8/go-ordered-map/v2"
+
+	"github.com/sirupsen/logrus"
 )
 
 func newGenSchemaCommand() *cobra.Command {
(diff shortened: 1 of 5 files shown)


cmd/limactl/genschema.go
```diff
@@ -3,21 +3,27 @@ package main
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 
+	"github.com/goccy/go-yaml"
 	"github.com/invopop/jsonschema"
 	"github.com/lima-vm/lima/pkg/limayaml"
+	jsonschema2 "github.com/santhosh-tekuri/jsonschema/v6"
```
AkihiroSuda, line 11: Would it be possible to avoid importing two jsonschema libs

afbjorklund: One is for generating, one is for validating. Their naming is unfortunate.

https://json-schema.org/tools?query=&languages=Go

afbjorklund: Their import names could be better than "1" and "2" though.

Perhaps if it was put in two different files, as helper functions?

AkihiroSuda (approved):
Thanks

PR #2778, opened 2024-10-21
Remove codesign target from Makefile

jandubois:
The regular target for `_output/bin/limactl$(exe)` already does code signing unconditionally, and the redundant `codesign` target doesn't use `-f`, so would always fail anyways: `_output/bin/limactl: is already signed`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -575,12 +575,6 @@ artifacts-misc: | _artifacts
 MKDIR_TARGETS += _artifacts
 
 ################################################################################
-.PHONY: codesign
-codesign: _output/bin/limactl
-ifeq ($(GOOS),darwin)
-	codesign --entitlements vz.entitlements -s - $<
-endif
-
 # This target must be placed after any changes to the `MKDIR_TARGETS` variable.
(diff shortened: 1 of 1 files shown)


jandubois:
Nope, doesn't make a difference

jandubois:
Oops, closed the wrong PR :(

norio-nomura (approved):
I personally don't need this target, but I left it in during the previous refactoring because I thought the reason for having a separate target might be to fulfill a requirement like "wanting to codesign standalone after building through some other method besides `make`." If @AkihiroSuda doesn't need it either, I think it's fine to remove it.

jandubois:
> "wanting to codesign standalone after building through some other method besides `make`."

Doesn't sound like a reasonable use-case. If you build without `make`, why won't you `codesign` without `make` as well?

In the past you had to use it, like `make clean exe codesign`, but now `exe` includes the code signing, and the extra target is broken unless you add an `-f`. That's how I found out, and why I created the PR. 😄

AkihiroSuda (approved):
Thanks

PR #2779, opened 2024-10-21
Use Rancher Desktop setup for WSL for GHA

jandubois:
Just an experiment to see if this WSL2 setup avoids the Windows test failures we are seeing since yesterday.


jandubois:
Nope, didn't make a difference.

PR #2782, opened 2024-10-22
[POC, DNM] Expose socks proxy (per user-v2 net, not per instance)

AkihiroSuda:
Usage:
```
curl \
  --proxy socks5h://localhost/$HOME/.lima/_networks/user-v2/user-v2_socks.sock \
  192.168.104.4
```

This is similar to the `limactl tunnel` proposal (PR #2710). While PR #2710 creates a proxy per an instance, this commit creates a proxy per a user-v2 network.

Remarks:
- Only works for user-v2 networks.
- DNS lookup is not implemented yet in this POC. Could be taken from https://github.com/norouter/norouter/blob/v0.6.5/pkg/agent/socks/socks.go#L57-L75
- https://github.com/cybozu-go/usocksd is a dependency hog (See the `go.mod` diff). Should be replaced with a fork or another library.

Overall, PR #2710 might be better than this commit, as PR #2710 works for any network driver, does not need an additional DNS resolver, and does not incur additional `go.mod` deps.


AkihiroSuda:
> Maybe `limactl tunnel` may have `limactl tunnel --target=instance INSTANCE` (default?) and `limactl tunnel --target=network NETWORK` to support the both concepts

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/issues/2710#issuecomment-2430015082_

go.mod
```diff
@@ -55,6 +56,30 @@ require (
 	k8s.io/client-go v0.31.1
 )
 
+require (
+	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/cespare/xxhash/v2 v2.3.0 // indirect
+	github.com/cybozu-go/log v1.6.1 // indirect
+	github.com/cybozu-go/netutil v1.4.2 // indirect
+	github.com/cybozu-go/well v1.11.0 // indirect
+	github.com/hashicorp/hcl v1.0.0 // indirect
```
AkihiroSuda, line 65: License check is failing
```
Not allowed license MPL-2.0 found for library github.com/hashicorp/hcl
```

jandubois: It is in the CNCF exception list (like all the other Hashicorp libraries used by Kubernetes), so using it would be fine:

```json
  {
    "package": "github.com/hashicorp/hcl",
    "license": "MPL-2.0",
    "comment": "not auto-allowlist because: Non-allowlist license(s); approved by GB exception 2019-03-11"
  },
```

From https://github.com/cncf/foundation/blob/1e80c35a4b43868d5070366d9e4501babafd8a52/license-exceptions/cncf-exceptions-2022-04-12.json#L57-L61

I have not been able to figure out[^1] how this makes sense from a legal point of view, that some MPL libraries are acceptable and others aren't.

[^1]: Yes, including asking a TOC member.

PR #2795, opened 2024-10-24
templates: add ubuntu-20.04, ubuntu-22.04

AkihiroSuda:
Replaces:
- https://github.com/lima-vm/lima/pull/2433


jandubois (approved):
Thanks, LGTM

I've tested both templates on Intel.

I have not tested the aarch64 URLs/checksums, nor the fallback URLs.

PR #2798, opened 2024-10-24
Replace copySparse with go-qcow2reader/convert

### LINKED ISSUE #2579: Converting qcow2 images to raw is too slow
### Description

Based on the logs, converting ubuntu server cloud image (xxx MiB) to raw format takes 17 seconds. The same operation using qemu-img convert takes 1.8 seconds.

Example log:

```
time="2024-09-02T01:43:25+03:00" level=info msg="Converting \"/Users/nsoffer/.lima/cluster/basedisk\" (qcow2) to a raw disk \"/Users/nsoffer/.lima/cluster/diffdisk\""
...
time="2024-09-02T01:43:42+03:00" level=info msg="Expanding to 20GiB"
```

Same with qemu-img
```
% time qemu-img convert -f qcow2 -O raw ~/.lima/cluster/basedisk diffdisk
qemu-img convert -f qcow2 -O raw ~/.lima/cluster/basedisk diffdisk  2.37s user 1.90s system 241% cpu 1.768 total
```

Lima has nice progress bar during the slow convert, but qemu-img is fast enough so no progress bar is needed. It has also a progress bar option that can be used to extract progress if needed.

Fix:
- use `qemu-img convert` if available
- use `-p` to show progress

nirs:
With this converting the default ubuntu 24.10 qcow2 compressed image to raw is 5.4 times faster:

Before:

    % limactl create --tty=false
    3.50 GiB / 3.50 GiB [-------------------------------------] 100.00% 317.58 MiB/s

After:

    % limactl create --tty=false
    3.50 GiB / 3.50 GiB [-------------------------------------] 100.00% 1.67 GiB/s

Fixes #2579


pkg/progressbar/progressbar.go
```diff
@@ -9,6 +10,17 @@ import (
...
 )
 
+type ProxyReaderAt struct {
+	io.ReaderAt
+	Bar *pb.ProgressBar
+}
+
+func (r *ProxyReaderAt) ReadAt(p []byte, off int64) (int, error) {
+	n, err := r.ReaderAt.ReadAt(p, off)
+	r.Bar.Add(n)
```
AkihiroSuda, line 20: This should check negative n?

nirs: This is documented to return positive value so I think we are safe.

> WriteAt writes len(p) bytes from p to the underlying data stream at offset off. It returns the number of bytes written from p (0 <= n <= len(p)) and any error encountered that caused the write to stop early. WriteAt must return a non-nil error if it returns n < len(p).

If we get negative value the convert will fail here:
https://github.com/lima-vm/go-qcow2reader/pull/36/commits/2cdb2331b6edce792b6d92c1fa35cb8e09736fe5#diff-b4093fac13dc66eefdab36a0bd2f4ee72c635a84a9c5687f8fc3e33185c6d9aeR131

AkihiroSuda: If err is non-nil, r.Bad.Add should not be called?

nirs: > If err is non-nil and n is negative

n is never negative

I pasted io.WriterAt docs before, but io.ReaderAt docs are the same:

> ReadAt reads len(p) bytes into p starting at offset off in the underlying input source. It returns the number of bytes read (0 <= n <= len(p)) and any error encountered.

> r.Bar.Add should not be called?

err may be io.EOF:

> If the n = len(p) bytes returned by ReadAt are at the end of the input source, ReadAt may return either err == EOF or err == nil.

So I think always adding n and let the caller handle errors in the right thing. This is also what pb.Reader is doing: https://github.com/cheggaaa/pb/blob/4ca3463f4bcf4446e1746efc6e484e795cfe7b12/reader.go#L15

We can handle n == 0, but I'm not sure it is helpful to add a check that fail for all good reads, for case that happens only when err != nil, maybe once when per convert.

nirs:
Updated to build with latest version.

go.mod
```diff
@@ -142,3 +142,6 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+// Temporary raplcement for testing https://github.com/lima-vm/go-qcow2reader/pull/36
+replace github.com/lima-vm/go-qcow2reader v0.2.1 => github.com/nirs/go-qcow2reader v0.0.0-20241025131821-46183d3fe6b2
```
AkihiroSuda, line 147: Released https://github.com/lima-vm/go-qcow2reader/releases/tag/v0.3.0

AkihiroSuda (approved):
LGTM after updating go.mod

go.mod
nirs: go.mod was not updated yet, should I updated it in this PR?

nirs: Add a commit to updated the dependency, can be removed if we go with the standard dependency update.

AkihiroSuda (approved):
Thanks

PR #2801, opened 2024-10-25
Avoid spamming the log when /dev/rtc does not exist

afbjorklund:
For instance the Raspberry Pi does not have any real-time clock, so rely on network time synchronization instead (systemd-timesyncd).

----

Currently you get a warning every 10 seconds in the guestagent log

```
2024-10-24T16:01:43.502852+00:00 keylimepi lima-guestagent[726]: time="2024-10-24T16:01:43Z" level=warning msg="fixSystemTimeSkew: lookup error: open /dev/rtc: no such file or directory"
2024-10-24T16:01:53.512171+00:00 keylimepi lima-guestagent[726]: time="2024-10-24T16:01:53Z" level=warning msg="fixSystemTimeSkew: lookup error: open /dev/rtc: no such file or directory"
```

But it only needs to be logged once, that there is no `/dev/rtc` file.

### DIFF
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -324,6 +324,11 @@ const deltaLimit = 2 * time.Second
 
 func (a *agent) fixSystemTimeSkew() {
 	for {
+		ok, err := timesync.HasRTC()
+		if !ok {
+			logrus.Warnf("fixSystemTimeSkew: error: %s", err.Error())
+			break
+		}
 		ticker := time.NewTicker(10 * time.Second)
 		for now := range ticker.C {
 			rtc, err := timesync.GetRTCTime()
--- a/pkg/guestagent/timesync/timesync_linux.go
+++ b/pkg/guestagent/timesync/timesync_linux.go
@@ -1,6 +1,7 @@
 package timesync
 
 import (
+	"errors"
 	"os"
 	"time"
 
@@ -8,6 +9,11 @@ import (
 )
 
 const rtc = "/dev/rtc"
+
+func HasRTC() (bool, error) {
+	_, err := os.Stat(rtc)
+	return !errors.Is(err, os.ErrNotExist), err
+}
 
 func GetRTCTime() (t time.Time, err error) {
 	f, err := os.Open(rtc)


pkg/guestagent/guestagent_linux.go
```diff
@@ -326,6 +326,10 @@ func (a *agent) fixSystemTimeSkew() {
 	for {
 		ticker := time.NewTicker(10 * time.Second)
 		for now := range ticker.C {
+			if !timesync.DeviceExists() {
+				logrus.Infof("fixSystemTimeSkew: device does not exist")
```
AkihiroSuda, line 330:
```suggestion
				logrus.Info("fixSystemTimeSkew: /dev/rtc does not exist")
```

AkihiroSuda (approved):
Thanks

AkihiroSuda:
> For instance the Raspberry Pi does not have any real-time clock, so rely on network time synchronization instead (systemd-timesyncd).

Are you trying to use Raspberry Pi as a "remote" non-virtual machine guest with your PR XXX?

afbjorklund:
It is a zero, so it doesn't have so many options. For the bigger Raspberry Pi, you can add* a battery RTC as well.

The Raspberry Pi 5 does feature a RTC by default, but then again the 4-5 (and 400) are more like a small ARM PC.

pkg/guestagent/guestagent_linux.go
```diff
@@ -326,6 +326,10 @@ func (a *agent) fixSystemTimeSkew() {
...
+				break
```
afbjorklund, line 331: Apparently the "break" is **not** enough to exit, it is like the range has queued a lot of events already?

pkg/guestagent/guestagent_linux.go
afbjorklund: now it says `level=warning msg="fixSystemTimeSkew: error: stat /dev/rtc: no such file or directory"` (once)

pkg/guestagent/guestagent_linux.go
```diff
@@ -324,6 +324,11 @@ const deltaLimit = 2 * time.Second
 
 func (a *agent) fixSystemTimeSkew() {
 	for {
+		ok, err := timesync.HasRTC()
+		if !ok {
+			logrus.Warnf("fixSystemTimeSkew: error: %s", err.Error())
```
AkihiroSuda, line 329: You can use logrus.WithError

AkihiroSuda: ping @afbjorklund

afbjorklund: I can add it, just didn't know if it was required (the other code is not using it yet)

AkihiroSuda (approved):

PR #1679, opened 2023-07-23
Rename "examples" to "templates"

AkihiroSuda:
The "examples" directory in the git repo is not renamed, for keeping existing hyperlinks alive

PR #2809, opened 2024-10-28
Rename `examples` to `templates` on the git repo

### LINKED ISSUE #2808: git repo: rename `examples` to `templates` (will break existing hyperlinks)
https://github.com/lima-vm/lima/tree/master/examples

This will break existing hyperlinks, but probably acceptable in v1.0

AkihiroSuda:
Also removes `/usr/local/share/lima/examples` (symlink to `templates`).

Follow-up to:
- #1679

Fix #2808


Makefile
AkihiroSuda, line 1:
```
$ make
Makefile:366: warning: overriding commands for target `_output/share/doc/lima/templates'
Makefile:357: warning: ignoring old commands for target `_output/share/doc/lima/templates'
cp config.mk .config
Makefile:366: warning: overriding commands for target `_output/share/doc/lima/templates'
Makefile:357: warning: ignoring old commands for target `_output/share/doc/lima/templates'
make: *** No rule to make target `pkg/limayaml/default.yaml', needed by `_output/bin/limactl'.  Stop.
```

jandubois (approved):
Thanks, LGTM

afbjorklund:
> Also removes /usr/local/share/lima/examples (symlink to templates).

This "broke" the Lima GUI, in that it will no longer change to the correct directory for you when using the Open... dialog

AkihiroSuda:
> This "broke" the Lima GUI, in that it will no longer change to the correct directory for you when using the Open... dialog

Do you want the symlink to be restored, or do you plan to just update your GUI?

afbjorklund:
I will update it, it was a bit hardcoded... Probably should look for both of them (for a while, at least) for old Lima.

AkihiroSuda:
You can just use ` getPrefix() + /share/doc/lima/templates`. This path has been available since https://github.com/lima-vm/lima/pull/1679 (v0.17.0, Aug 2023).

afbjorklund:
Right. Then again the class is called Example too :-) So it would probably need some rewriting anyway.

Issue #2807, opened 2024-10-25
Add support to install nerdctl and buildkit with the alpine-iso template


jandubois:
All 3 binaries are now available in Alpine 3.17, from `buildctl` and `buildkit` \<sic\> and `nerdctl` apks.

Will make a separate PR to make the setuid chmod optional, to require using `sudo` with `nerdctl`...

_Originally posted by @afbjorklund in https://github.com/lima-vm/alpine-lima/issues/77#issuecomment-1357393401_

jandubois:
I've only created this issue because I want to close https://github.com/lima-vm/alpine-lima/issues/77.

I'm not sure if we care about supporting `containerd.system: true` for Alpine (although the code would then also work for the regular non-ISO image.

I'm also fine with closing this issue and leaving it to the user to install additional packages via provisioning scripts.

afbjorklund:
> I'm also fine with closing this issue and leaving it to the user to install additional packages via provisioning scripts.

That is OK with me, the original issue was only to split the installation in the ISO into the same three components...

afbjorklund:
I think the nerdctl `containerd-rootless-setuptool.sh` script had some issues, when trying to run it outside of systemd.

* https://github.com/lima-vm/lima/issues/310

AkihiroSuda:
`containerd-rootless-setuptool.sh` is dedicated for systemd, but `containerd-rootless.sh` should work without it

PR #2812, opened 2024-10-28
Bump alpine-iso from 3.20.0  → 3.20.3

jandubois:
Replace the "hint" to use the "rd" edition with a generic note to install containerd, buildkit, and nerdctl via provisioning script. Advice may change again depending on https://github.com/lima-vm/lima/issues/2807.


AkihiroSuda (approved):
Thanks

PR #2814, opened 2024-10-28
templates: fix minimumLimaVersion for riscv64-enabled templates ;  default: move several properties to the (further) advanced section

AkihiroSuda:


jandubois (approved):
Thanks, LGTM

PR #2815, opened 2024-10-28
limactl: make sure to remove "." from instance names

### LINKED ISSUE #2813: `limactl start template://ubuntu-24.10` creates an instance named `ubuntu-24.10`, while `limactl start ./ubuntu-24.10.yaml` creates `ubuntu-24-10`
Probably `ubuntu-24-10` is the expected result?

AkihiroSuda:
Now `limactl create template://ubuntu-24.10` creates `ubuntu-24-10`.

Fix #2813

### DIFF
--- a/cmd/limactl/guessarg/guessarg.go
+++ b/cmd/limactl/guessarg/guessarg.go
@@ -45,6 +45,16 @@ func SeemsYAMLPath(arg string) bool {
 	return strings.HasSuffix(lower, ".yml") || strings.HasSuffix(lower, ".yaml")
 }
 
+func InstNameFromTemplateName(templateName string) (string, error) {
+	s := filepath.Base(templateName)
+	s = strings.ReplaceAll(s, ".", "-")
+	if err := identifiers.Validate(s); err != nil {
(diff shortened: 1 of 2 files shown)


jandubois (changes requested):
> Now limactl create template://ubuntu-24.10 creates ubuntu-24-10

I disagree with this. I think we should rename the template to `ubuntu-24-10.yaml` instead.

jandubois:
> I disagree with this. I think we should rename the template to `ubuntu-24-10.yaml` instead.

Alternatively, we _could_ call the instance `ubuntu-24.10` but the hostname inside the VM `lima-ubuntu-24-10` since we modify the instance name to get the hostname anyways.

I still have a slight preference to rename the template though, to keep all the names in sync.

AkihiroSuda:
> Alternatively, we could call the instance ubuntu-24.10 but the hostname inside the VM lima-ubuntu-24-10 since we modify the instance name to get the hostname anyways.

👍 

- https://github.com/lima-vm/lima/pull/2817

PR #2817, opened 2024-10-28
Allow "." and "_" in instance names, disallow them in hostnames

### LINKED ISSUE #2813: `limactl start template://ubuntu-24.10` creates an instance named `ubuntu-24.10`, while `limactl start ./ubuntu-24.10.yaml` creates `ubuntu-24-10`

AkihiroSuda:
Now `limactl create template://ubuntu-24.10` and `limactl create ./ubuntu-24.10.yaml` creates an instance named `ubuntu-24.10` with the hostname `lima-ubuntu-24-10`.

Eventually the hostname should be customizable via the YAML, but it is beyond the scope of this commit.

Fix #2813
Alternative to:
- #2815

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -22,6 +22,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/cpu"
 
+	"github.com/lima-vm/lima/pkg/identifierutil"
 	"github.com/lima-vm/lima/pkg/networks"
 	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/ptr"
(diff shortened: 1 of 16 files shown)


pkg/limayaml/defaults.go
```diff
@@ -814,12 +815,14 @@ func executeGuestTemplate(format, instDir string, param map[string]string) (byte
...
+		name := filepath.Base(instDir)
 		data := map[string]interface{}{
-			"Home":  fmt.Sprintf("/home/%s.linux", user.Username),
-			"Name":  filepath.Base(instDir),
-			"UID":   user.Uid,
-			"User":  user.Username,
-			"Param": param,
+			"Home":     fmt.Sprintf("/home/%s.linux", user.Username),
+			"Name":     name,
+			"Hostname": identifierutil.HostnameFromInstName(name), // TODO: support customization
```
jandubois, line 822: This new variable should be documented in `default.yaml`.

jandubois (approved):
Thanks, LGTM

I think the new Hostname field in the guest templates should be documented.

The whole logic is a bit convoluted, with the repeated recomputation of the hostname; this should all go away when `hostname` becomes a field in `lima.yaml`, and is only set once in `FillDefaults`.

pkg/limayaml/defaults.go
AkihiroSuda: done

jandubois (approved):
Thanks, LGTM

PR #2821, opened 2024-10-29
Update fedora template to f41

subpop:

Issue #2701, opened 2024-10-08
ArchLinux + 9p fails on GitHub Actions (but not on local MacBook): `failed to get unprivileged mount flags for \"/home/runner/lima-container-engine-test-tmp/random\": stat /home/runner/lima-container-engine-test-tmp/random: no such file or directory`


AkihiroSuda:
```
level=fatal msg="failed to get unprivileged mount flags for \"/home/runner/lima-container-engine-test-tmp/random\": stat /home/runner/lima-container-engine-test-tmp/random: no such file or directory"
```

https://github.com/lima-vm/lima/pull/1953/files/88cca7543122b694b7392e25816676779b9f38a3#r1790627760

AkihiroSuda:
> Looks like we are hitting the kernel bug introduced in v6.9-rc1, fixed in v6.12-rc5: `Revert "fs/9p: simplify iget to remove unnecessary paths"` https://github.com/torvalds/linux/commit/be2ca38

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/issues/2821#issuecomment-2447803822_

AkihiroSuda:
Carried as:
- https://github.com/lima-vm/lima/pull/2822

jandubois:
I think this issue is "fixed" by #2822.

PR #2822, opened 2024-10-30
templates: update fedora to 41; templates: disable 9p for Linux 6.9-6.11

AkihiroSuda:
9p is broken in Linux v6.9, v6.10, and v6.11 (see #2701 and https://github.com/lima-vm/lima/pull/2821#issuecomment-2444856353).
    
The issue was fixed in Linux v6.12-rc5 (https://github.com/torvalds/linux/commit/be2ca38).


jandubois (approved):
Thanks, LGTM

I was a bit surprised that images like `docker`, `faasd`, `k3s`, `k8s` are not based on `default` but on the LTS version. Maybe that should be documented?

PR #2825, opened 2024-10-30
Make sure the port forwarding protocol is always lowercase

jandubois:
The ServicePort.Protocol is always uppercase, e.g. "TCP", but the api.IPPort.protocol is always lowercase, i.e. "tcp".

Since UDP support was added in #2411 the hostagent filters on the protocol values.

cc @pendo324 @balajiv113 @Nino-K

### DIFF
--- a/pkg/guestagent/kubernetesservice/kubernetesservice.go
+++ b/pkg/guestagent/kubernetesservice/kubernetesservice.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"net/url"
 	"os"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jandubois:
Testing (fails without this PR, succeeds after it is applied):

```shell
limactl start --tty=false template://k3s
export KUBECONFIG="$HOME/.lima/k3s/copied-from-guest/kubeconfig.yaml"
kubectl create deployment nginx --image=nginx
kubectl create service nodeport nginx --tcp=80:80 --node-port=32000
curl http://localhost:32000
```

AkihiroSuda (approved):
Thanks

PR #2830, opened 2024-11-01
Copy the GECOS information for user

afbjorklund:
Normally this is the display name.

Mostly unused, these days. (finger?)

```yaml
users:
  - name: "anders"
    uid: "1000"
    gecos: "Anders"
    homedir: "/home/anders.linux"
    ....
```


pkg/cidata/cidata.TEMPLATE.d/lima.env
```diff
@@ -2,6 +2,7 @@ LIMA_CIDATA_DEBUG={{ .Debug }}
 LIMA_CIDATA_NAME={{ .Name }}
 LIMA_CIDATA_USER={{ .User }}
 LIMA_CIDATA_UID={{ .UID }}
+LIMA_CIDATA_GECOS={{ printf "%q" .GECOS }}
```
jandubois, line 5: There should be no quotes in `lima.env` unless they are part of the value. The file is not sourced, but processed in a read loop:

```console
$ export "GECOS=Jan Dubois"
$ echo $GECOS
Jan Dubois
```

```suggestion
LIMA_CIDATA_GECOS={{ .GECOS }}
```

jandubois (approved):
Thanks, LGTM

PR #2831, opened 2024-11-01
Bump alpine-iso to 0.2.41

jandubois:
This adds the ability to set the user's comment and home directory via lima-init instead of using hard-coded values (see #2827).


AkihiroSuda (approved):
Thanks

PR #2848, opened 2024-11-05
templates: update

AkihiroSuda:


jandubois (approved):
Thanks, LGTM

I've only tested the almalinux-8 template on Intel. I assume we just trust the update scripts to generate working URLs and checksums. Or do we download and verify that they are all functional?

AkihiroSuda:
Yes, we are just relying on the script

PR #2861, opened 2024-11-07
Make sure err isn't nil when returning failure

### LINKED ISSUE #2859: gRPC port forwarder seems to occasionally hang (with `template://docker`); also panics
The gRPC port forwarder introduced in Lima v1.0 seems to have a stability issue with `template://docker`:

https://github.com/lima-vm/lima/actions/runs/11638179464/job/32412581211

```
+ limactl shell docker docker pull --quiet ghcr.io/stargz-containers/nginx:1.19-alpine-org
time="2024-11-02T00:34:15Z" level=warning msg="treating lima version \"5329fd4\" from \"/home/runner/.lima/docker/lima-version\" as very latest release"
ghcr.io/stargz-containers/nginx:1.19-alpine-org
+ limactl shell docker docker run -d --name nginx -p 127.0.0.1:8080:80 ghcr.io/stargz-containers/nginx:1.19-alpine-org
time="2024-11-02T00:34:29Z" level=warning msg="treating lima version \"5329fd4\" from \"/home/runner/.lima/docker/lima-version\" as very latest release"
c8c19bbaf7fb13352c2521819495e1ed61f1e3d799d3de42b8d9bd70ee0ea265
+ timeout 3m bash -euxc 'until curl -f --retry 30 --retry-connrefused http://127.0.0.1:8080; do sleep 3; done'
+ curl -f --retry 30 --retry-connrefused http://127.0.0.1:8080/

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:02 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:03 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:04 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:05 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:06 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:07 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:08 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:09 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:10 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:11 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:12 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:13 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:14 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:15 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:16 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:17 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:18 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:19 --:--:--     0
  0     0    0     0    0     0      0      0 --:--:--  0:00:20 --:--:--     0
[...]
  0     0    0     0    0     0      0      0 --:--:--  0:02:59 --:--:--     0
```

`LIMA_SSH_PORT_FORWARDER=true` may work as a workaround.

jandubois:
Fixes #2859

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -443,9 +443,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        template:
-        - default.yaml
-        - fedora.yaml
+        include:
+        - {template: default.yaml, ssh-port-forwarder: true}  # SSH port forwarder is currently still the default
+        - {template: fedora.yaml, ssh-port-forwarder: false}  # gRPC port forwarder will become default in the future
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
       with:
@@ -467,6 +467,8 @@ jobs:
       run: brew uninstall --ignore-dependencies --force qemu
     - name: Test
       run: ./hack/test-templates.sh templates/${{ matrix.template }}
+      env:
+        LIMA_SSH_PORT_FORWARDER: ${{ matrix.ssh-port-forwarder }}
     - if: failure()
       uses: ./.github/actions/upload_failure_logs_if_exists
       with:
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -307,9 +307,17 @@ if [[ -n ${CHECKS["port-forwards"]} ]]; then
 			fi
 			limactl shell "$NAME" $sudo $CONTAINER_ENGINE info
 			limactl shell "$NAME" $sudo $CONTAINER_ENGINE pull --quiet ${nginx_image}
-			limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p 8888:80 ${nginx_image}
 
+			limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p 8888:80 ${nginx_image}
 			timeout 3m bash -euxc "until curl -f --retry 30 --retry-connrefused http://${hostip}:8888; do sleep 3; done"
+			limactl shell "$NAME" $sudo $CONTAINER_ENGINE rm -f nginx
+
+			if [ "$(uname)" = "Darwin" ]; then
+				# Only macOS can bind to port 80 without root
+				limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p 127.0.0.1:80:80 ${nginx_image}
+				timeout 3m bash -euxc "until curl -f --retry 30 --retry-connrefused http://localhost:80; do sleep 3; done"
+				limactl shell "$NAME" $sudo $CONTAINER_ENGINE rm -f nginx
+			fi
 		fi
 	fi
 	set +x
--- a/pkg/portfwd/listener_darwin.go
+++ b/pkg/portfwd/listener_darwin.go
@@ -68,10 +68,13 @@ func (p pseudoLoopbackListener) Accept() (net.Conn, error) {
 		conn.Close()
 		return nil, err
 	}
-	if remoteAddrIP != "127.0.0.1" {
-		logrus.WithError(err).Debugf("pseudoloopback forwarder: rejecting non-loopback remoteAddr %q", remoteAddr)
+	if !IsLoopback(remoteAddrIP) {
+		err := fmt.Errorf("pseudoloopback forwarder: rejecting non-loopback remoteAddr %q", remoteAddr)
+		logrus.Debug(err)
+		conn.Close()
 		return nil, err
 	}
+	logrus.Infof("pseudoloopback forwarder: accepting connection from %q", remoteAddr)
 	return conn, nil
 }
 
@@ -89,7 +92,7 @@ func (pk *pseudoLoopbackPacketConn) ReadFrom(bytes []byte) (n int, addr net.Addr
 	if err != nil {
 		return 0, nil, err
 	}
-	if remoteAddrIP != "127.0.0.1" {
+	if !IsLoopback(remoteAddrIP) {
 		return 0, nil, fmt.Errorf("pseudoloopback forwarder: rejecting non-loopback remoteAddr %q", remoteAddr)
 	}
 	return n, remoteAddr, nil
@@ -100,8 +103,12 @@ func (pk *pseudoLoopbackPacketConn) WriteTo(bytes []byte, remoteAddr net.Addr) (
 	if err != nil {
 		return 0, err
 	}
-	if remoteAddrIP != "127.0.0.1" {
+	if !IsLoopback(remoteAddrIP) {
 		return 0, fmt.Errorf("pseudoloopback forwarder: rejecting non-loopback remoteAddr %q", remoteAddr)
 	}
 	return pk.PacketConn.WriteTo(bytes, remoteAddr)
+}
+
+func IsLoopback(addr string) bool {
+	return net.ParseIP(addr).IsLoopback()
 }
--- a/pkg/hostagent/port_darwin.go
+++ b/pkg/hostagent/port_darwin.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/lima-vm/lima/pkg/bicopy"
+	"github.com/lima-vm/lima/pkg/portfwd"
 	"github.com/lima-vm/sshocker/pkg/ssh"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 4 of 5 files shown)


jandubois:
A separate issue is if the pseudoLoopbackForwarder should accept connections from `[::1]` in addition to `127.0.0.1`.

jandubois:
@AkihiroSuda This fixes the crash, but doesn't actually forward to `127.0.0.1`:

```
$ curl http://127.0.0.1:80
curl: (7) Failed to connect to 127.0.0.1 port 80 after 0 ms: Couldn't connect to server
```

jandubois:
> @AkihiroSuda This fixes the crash, but doesn't actually forward to `127.0.0.1`:

Actually, it only stops accepting connections from `127.0.0.1` after it has rejected a connection from `[::1]` once.

This also reminds of something @Nino-K told me: when the host agent sends a duplicate port to add, then the host agent removes all existing port forwardings, or something like that. Will see if I can find this in his WIP PR.

AkihiroSuda:
Thanks, can we have a test?
```diff
diff --git a/hack/test-templates.sh b/hack/test-templates.sh
index 57ff37a4..0d689312 100755
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -307,9 +307,11 @@ if [[ -n ${CHECKS["port-forwards"]} ]]; then
                        fi
                        limactl shell "$NAME" $sudo $CONTAINER_ENGINE info
                        limactl shell "$NAME" $sudo $CONTAINER_ENGINE pull --quiet ${nginx_image}
-                       limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p 8888:80 ${nginx_image}
-
-                       timeout 3m bash -euxc "until curl -f --retry 30 --retry-connrefused http://${hostip}:8888; do sleep 3; done"
+                       for hostport in 8888 80; do
+                               limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p ${hostrport}:80 ${nginx_image}
+                               timeout 3m bash -euxc "until curl -f --retry 30 --retry-connrefused http://${hostip}:${hostport}; do sleep 3; done"
+                               limactl shell "$NAME" $sudo $CONTAINER_ENGINE rm -f nginx
+                       done
                fi
        fi
        set +x
```

jandubois:
This PR now fixes the crash and also avoids closing the forwarder the first time a non-local address tries to connect to it.

```console
$ curl 127.0.0.1
<!DOCTYPE html>
…

$ curl localhost
curl: (56) Recv failure: Connection reset by peer

$ curl 127.0.0.1
<!DOCTYPE html>
…
```

I guess we could use a custom error type instead of matching on the string of the error message to make it a bit more robust.

I also think we should accept connections from `[::1]`, but that should be a separate PR.

jandubois:
> Fixes #2859

I don't actually know if this PR fixes the hanging problem; I only know it fixes the crash @AkihiroSuda has shown in the same issue.

And it makes the forwarding listeners a bit more robust in general against accidental removal.

I don't plan to make any more changes tonight; @rfay maybe you can test this PR (or wait until it has been merged into `master`)? Just so we know if this fixes all known issues, or if there is still something else.

AkihiroSuda:
CI is failing

jandubois:
> CI is failing

Yes, because we cannot bind to `${hostip}:80` I think. I'll see if I can still fix it tonight.

jandubois:
It is still failing in CI, but works for me locally. I'm out of time now and will look into this tomorrow. Unless somebody beats me to it...

hack/test-templates.sh
```diff
@@ -307,9 +307,17 @@ if [[ -n ${CHECKS["port-forwards"]} ]]; then
...
 			limactl shell "$NAME" $sudo $CONTAINER_ENGINE info
 			limactl shell "$NAME" $sudo $CONTAINER_ENGINE pull --quiet ${nginx_image}
-			limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p 8888:80 ${nginx_image}
 
+			limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p 8888:80 ${nginx_image}
 			timeout 3m bash -euxc "until curl -f --retry 30 --retry-connrefused http://${hostip}:8888; do sleep 3; done"
+			limactl shell "$NAME" $sudo $CONTAINER_ENGINE rm -f nginx
+
+			if [ "$(uname)" = "Darwin" ]; then
+			  # Only macOS can bind to port 80 without root
```
AkihiroSuda, line 316: Linux can, too, by setting `net.ipv4.ip_unprivileged_port_start`

hack/test-templates.sh
AkihiroSuda, line 1: The CI should now have a matrix to test the both port forwarding drivers

jandubois:
> Yes, because we cannot bind to `${hostip}:80` I think. I'll see if I can still fix it tonight.

I've restricted the test binding to port 80 to only macOS, which is also the only platform where the lower port makes a difference due to the pseudoloopback forwarders.

However, since we have reverted to the SSH forwarder, this test will not currently test the gRPC implementation.

Should we run at least one of the VZ tests (default or fedora) with `LIMA_SSH_PORT_FORWARDER=false`?

hack/test-templates.sh
jandubois: Yes, but you need to be root to set it.

hack/test-templates.sh
jandubois: Isn't that going to be a bit much? Should we just switch to the GRPC forwarder for a couple of tests?

jandubois: I've switched the VZ-Fedora test to use the gRPC forwarder.

We can switch some of the Linux tests too, but I don't think we should run **all** of these tests twice:

```
        - alpine.yaml
        - debian.yaml  # reverse-sshfs
        - fedora.yaml
        - archlinux.yaml
        - opensuse.yaml
        - docker.yaml
        - ../hack/test-templates/alpine-iso-9p-writable.yaml  # Covers alpine-iso.yaml
        - ../hack/test-templates/net-user-v2.yaml
        - ../hack/test-templates/test-misc.yaml  # TODO: merge net-user-v2 into test-misc
```

I feel like we can improve this in a follow-up PR though, I think this one fixes all bugs known (and reproducible) to me, including support for IPv6.

pkg/portfwd/listener_darwin.go
```diff
@@ -100,8 +103,12 @@ func (pk *pseudoLoopbackPacketConn) WriteTo(bytes []byte, remoteAddr net.Addr) (
...
 		return 0, err
 	}
-	if remoteAddrIP != "127.0.0.1" {
+	if !IsLoopback(remoteAddrIP) {
 		return 0, fmt.Errorf("pseudoloopback forwarder: rejecting non-loopback remoteAddr %q", remoteAddr)
 	}
 	return pk.PacketConn.WriteTo(bytes, remoteAddr)
 }
+
+func IsLoopback(addr string) bool {
```
AkihiroSuda, line 112: https://pkg.go.dev/net#IP.IsLoopback is more robust

.github/workflows/test.yml
```diff
@@ -443,9 +443,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        template:
-        - default.yaml
-        - fedora.yaml
+        include:
+        - {template: default.yaml, ssh-port-forwarder: true}
+        - {template: fedora.yaml, ssh-port-forwarder: false}
```
AkihiroSuda, line 448: Can we have a comment to clarify that ssh-port-forwarder: false is not required by fedora, but added for the coverage of the tests?

pkg/portfwd/listener_darwin.go
jandubois: Updated

.github/workflows/test.yml
jandubois: sure

AkihiroSuda:
> Fixes #2859

Does this PR fix the hang too? Or just fixes the panic?

AkihiroSuda (approved):
Thanks

jandubois:
> Does this PR fix the hang too?

I have not been able to reproduce the hangs. Do you have any repro steps?

> Or just fixes the panic?

It fixes the panic, adds support for IPv6, and makes sure the forwarder isn't removed when a connection is attempted from a rejected (non-loopback) address.

AkihiroSuda:
> I have not been able to reproduce the hangs. Do you have any repro steps?

Just running docker.yaml on the CI seems enough.
And I don’t know other way to reproduce the hang.

jandubois:
> Just running docker.yaml on the CI seems enough.

It seems almost all the PRs merged for v1.0.1 have been running `docker.yaml` without hanging, otherwise they wouldn't be green: https://github.com/lima-vm/lima/pulls?q=is%3Apr+is%3Aclosed+milestone%3Av1.0.1

Anyways, I can't remember seeing the docker test hang in this PR, but also not in any of the others I did after v1.0.0.

AkihiroSuda:
I think I have restarted several failing jobs, so they are marked  green

jandubois:
> I think I have restarted several failing jobs, so they are marked green

Ok, let's see if this is going to stop now.

PR #2863, opened 2024-11-07
Update `inject-cmdline-to-template.sh`

norio-nomura:
- use `limactl edit` instead of `yq -i`
- simplify `yq` filter
- add `armv7l` and `riscv64` to arch
- change `default_cmdline` for minimal flavor  
  use `root=/dev/vda1` instead of `root=LABEL=cloudimg-rootfs` to avoid failing boot
  ```console
  [    0.127771] /dev/root: Can't open blockdev
  [    0.127797] VFS: Cannot open root device "LABEL=cloudimg-rootfs" or unknown-block(0,0): error -6
  [    0.127908] Please append a correct "root=" boot option; here are the available partitions:
  ```
- ~~inject provision[0].script `systemctl enable systemd-time-wait-sync.service`  
  Booting with a kernel image lacks an RTC, so time synchronization is needed to prevent network errors.~~

### DIFF
--- a/hack/inject-cmdline-to-template.sh
+++ b/hack/inject-cmdline-to-template.sh
@@ -57,6 +57,8 @@ done
 if [[ -z ${location} ]]; then
 	echo "Failed to get the image location for ${template}" >&2
 	exit 1
+elif [[ ${location} == https://cloud-images.ubuntu.com/minimal/* ]]; then
+	readonly default_cmdline="root=/dev/vda1 ro console=tty1 console=ttyAMA0"
 elif [[ ${location} == https://cloud-images.ubuntu.com/* ]]; then
 	readonly default_cmdline="root=LABEL=cloudimg-rootfs ro console=tty1 console=ttyAMA0"
 else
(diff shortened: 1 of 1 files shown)


norio-nomura:
> - inject provision[0].script `systemctl enable systemd-time-wait-sync.service`  

Removed the above and instead opened #2894

AkihiroSuda:
What is the problem with `root=LABEL=cloudimg-rootfs`?

norio-nomura:
> What is the problem with `root=LABEL=cloudimg-rootfs`?

The kernel image with minimal flavor image fails to boot:
```console
[    0.127771] /dev/root: Can't open blockdev
[    0.127797] VFS: Cannot open root device "LABEL=cloudimg-rootfs" or unknown-block(0,0): error -6
[    0.127908] Please append a correct "root=" boot option; here are the available partitions:
```

AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

PR #2864, opened 2024-11-07
Revert the default port forwarder to SSH

AkihiroSuda:
The GRPC port forwarder is still too unstable (issue #2859, etc.)


jandubois:
Note for the release notes that this means 1.0.1 will not support UDP forwarding, as that relies on the gRPC forwarder.

jandubois (approved):
Thanks, LGTM

PR #2873, opened 2024-11-08
test-templates.sh: fix wrong comment "binds to 0.0.0.0 by default"

AkihiroSuda:
This is a non-default behavior configured via `test-port-forwarding.pl`.


jandubois (approved):
Thanks, LGTM

PR #2874, opened 2024-11-08
Use printf instead of $'' to pass a multi-line string to bash

### LINKED ISSUE #2856: Changing the VM user's shell to fish causes lima to wait forever for the SSH requirement to be satisfied
### Description

When setting the lima VM user's shell to fish, lima will get stuck forever waiting for the the SSH requirement to be satisfied.

[lima-start.log](https://github.com/user-attachments/files/17651935/lima-start.log)
The loop is in the following lines:
```log
time="2024-11-06T13:14:17-06:00" level=info msg="[hostagent] Waiting for the essential requirement 1 of 2: \"ssh\""
time="2024-11-06T13:14:17-06:00" level=debug msg="[hostagent] executing script \"ssh\""
time="2024-11-06T13:14:17-06:00" level=debug msg="[hostagent] executing ssh for script \"ssh\": /usr/bin/ssh [ssh -F /dev/null -o IdentityFile=\"/Users/burner/.config/lima/_config/user\" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o NoHostAuthenticationForLocalhost=yes -o GSSAPIAuthentication=no -o PreferredAuthentications=publickey -o Compression=no -o BatchMode=yes -o IdentitiesOnly=yes -o Ciphers=\"^aes128-gcm@openssh.com,aes256-gcm@openssh.com\" -o User=burner -o ControlMaster=auto -o ControlPath=\"/Users/burner/.config/lima/alpine/ssh.sock\" -o ControlPersist=yes -p 49540 127.0.0.1 -- /bin/bash -c $'while read -r line; do [ -n \"$line\" ] && export \"$line\"; done<<EOF\\n$(sudo cat /mnt/lima-cidata/param.env)\\nEOF\\n/bin/bash']"
time="2024-11-06T13:14:17-06:00" level=debug msg="[hostagent] stdout=\"\", stderr=\"fish: Expected a variable name after this $.\\n/bin/bash -c $'while read -r line; do [ -n \\\"$line\\\" ] && export \\\"$line\\\"; done<<EOF\\\\n$(sudo cat /mnt/lima-cidata/param.env)\\\\nEOF\\\\n/bin/bash'\\n             ^\\n\", err=failed to execute script \"ssh\": stdout=\"\", stderr=\"fish: Expected a variable name after this $.\\n/bin/bash -c $'while read -r line; do [ -n \\\"$line\\\" ] && export \\\"$line\\\"; done<<EOF\\\\n$(sudo cat /mnt/lima-cidata/param.env)\\\\nEOF\\\\n/bin/bash'\\n             ^\\n\": exit status 127"
```
Lima is running a script that was presumably meant for bash or another standard shell, but it runs it in whatever the user's configured shell is. If the user's shell is set to fish, this causes syntax a syntax error.

Reproduced on alpine & fedora VMs. I can provide any extra logs or make some dedicated VMs specifically for testing this if needed.

OS: Mac OS 14.7.1
CPU: Apple M1 Pro
Lima version: 1.0.0 (via homebrew)

jandubois:
The script will be executed by the default shell of the ssh user, and `fish` does not support the $'' syntax.

Fixes #2856

Seems to work:

```console
$ limactl shell alpine
jan@lima-alpine /U/jan>
```

### DIFF
--- a/pkg/hostagent/requirements.go
+++ b/pkg/hostagent/requirements.go
@@ -3,6 +3,7 @@ package hostagent
 import (
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/lima-vm/lima/pkg/limayaml"
@@ -61,14 +62,17 @@ func (a *HostAgent) waitForRequirements(label string, requirements []requirement
 //	EOF
 //	/usr/bin/env ruby
 //
-// ssh.ExecuteScript will strip the `#!` prefix from the first line and invoke the rest
-// of the line as the command. The full script is then passed via STDIN. We use the $' '
-// form of shell quoting to be able to use \n as newline escapes to fit everything on a
+// ssh.ExecuteScript will strip the `#!` prefix from the first line and invoke the
+// rest of the line as the command. The full script is then passed via STDIN. We use
+// "$(printf '…')" to be able to use \n as newline escapes, to fit everything on a
 // single line:
 //
-//	#!/bin/bash -c $'while … done<<EOF\n$(sudo …)\nEOF\n/usr/bin/env ruby'
+//	#!/bin/bash -c "$(printf 'while … done<<EOF\n$(sudo …)\nEOF\n/usr/bin/env ruby')"
 //	#!/usr/bin/env ruby
 //	…
+//
+// An earlier implementation used $'…' for quoting, but that isn't supported if the
+// user switched the default shell to fish.
 func prefixExportParam(script string) (string, error) {
 	interpreter, err := ssh.ParseScriptInterpreter(script)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


jandubois:
Seems like `TEST| [INFO] Testing that PARAM env variables are exported to all types of provisioning scripts and probes` is failing...

AkihiroSuda:
Can we move this to v1.0.2?

jandubois:
> Can we move this to v1.0.2?

Sure. It is a regression to pre-1.0.0 releases, but only affects users that have switched their default shell inside the instance to `fish`.

jandubois:
> Seems like `TEST| [INFO] Testing that PARAM env variables are exported to all types of provisioning scripts and probes` is failing...

Just a small quoting issue; should be all good now.

pkg/hostagent/requirements.go
```diff
@@ -77,7 +81,14 @@ func prefixExportParam(script string) (string, error) {
...
 	// TODO we should have a symbolic constant for `/mnt/lima-cidata`
 	exportParam := `while read -r line; do [ -n "$line" ] && export "$line"; done<<EOF\n$(sudo cat /mnt/lima-cidata/param.env)\nEOF\n`
-	return fmt.Sprintf("#!/bin/bash -c $'%s%s'\n%s", exportParam, interpreter, script), nil
+
+	// double up all '%' characters so we can pass them through unchanged in the format string of printf
+	interpreter = strings.ReplaceAll(interpreter, "%", "%%")
+	exportParam = strings.ReplaceAll(exportParam, "%", "%%")
+	// strings will be interpolated into single-quoted strings, so protect any existing single quotes
+	interpreter = strings.ReplaceAll(interpreter, "'", `'"'"'`)
+	exportParam = strings.ReplaceAll(exportParam, "'", `'"'"'`)
```
jandubois, line 90: All these replacements are not really needed because the input strings don't include `%` or `'` characters. but it feels like the right thing to do to make them correct for that case as well.

norio-nomura (approved):
LGTM 👍🏻 
However, I feel that `ssh.ExecuteScript` should use `shellwords.Parse` to split the first line into a command and arguments rather than using the entire line as the command itself.

jandubois:
> LGTM 👍🏻

🙏 

> However, I feel that `ssh.ExecuteScript` should use `shellwords.Parse` to split the first line into a command and arguments rather than using the entire line as the command itself.

Given that `ssh.ExecuteScript` is in https://github.com/lima-vm/sshocker/blob/master/pkg/ssh/ssh.go (in a different repo), and that this is a separate issue from what this PR is trying to fix, I assume you will create a different issue for that.

PR #2877, opened 2024-11-08
Make sure that root is in the sudoers file on Alpine

jandubois:
Because the Alpine cloud image doesn't include `/etc/sudoers`, and cloud-init just writes a single line to it:

```
#includedir /etc/sudoers.d
```

Without an entry for `root` we cannot run user provisioning scripts because we use `sudo -u $USER …` to do it.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/00-alpine-user-group.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/00-alpine-user-group.sh
@@ -1,12 +1,21 @@
 #!/bin/sh
+test -f /etc/alpine-release || exit 0
+
+# Make sure that root is in the sudoers file.
+# This is needed to run the user provisioning scripts.
+SUDOERS=/etc/sudoers.d/00-root-user
+if [ ! -f $SUDOERS ]; then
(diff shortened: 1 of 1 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot/00-alpine-sudoers-root.sh
AkihiroSuda, line 1: Could this be merged to 00-alpine-user-group.sh ?

We have too many 00- files 😅

jandubois: > Could this be merged to 00-alpine-user-group.sh ?

I merged it, and now the Alpine test is failing (it passed before). I don't see how this PR could cause it to fail 😞

jandubois: > now the Alpine test is failing

I found and fixed it: an incorrectly negated check that prevented the instance from rebooting and therefore the proxy settings from `/etc/environment` where not set, which made the test script to fail on

```
got=$(limactl shell "$NAME" env | grep FTP_PROXY)
```

AkihiroSuda (approved):
Thanks

PR #2882, opened 2024-11-08
copy k3s server-ca to guest

brandonros:


jandubois:
Can you please provide the motivation for making this change? Why is it needed, how is it going to be used?

PR #2894, opened 2024-11-12
Add `boot/00-check-rtc-and-wait-ntp.sh` to cidata

norio-nomura:
In vz, the VM lacks an RTC when booting with a kernel image (see: https://developer.apple.com/forums/thread/760344). This causes incorrect system time until NTP synchronizes it, leading to TLS errors. To avoid TLS errors, this script waits for NTP synchronization if RTC is unavailable.

This script does the following:
- Exits with 0 if `/dev/rtc0` exists.
- Exits with 0 if `systemctl` is not available.
- Enables `systemd-time-wait-sync.service` to wait for NTP synchronization at an earlier stage on subsequent boots.
- Waits for NTP synchronization within the script for the first boot.

Log output during execution:
```console
LIMA 2024-08-08T23:51:15+09:00| Executing /mnt/lima-cidata/boot/00-check-rtc-and-wait-ntp.sh
Created symlink /etc/systemd/system/sysinit.target.wants/systemd-time-wait-sync.service → /usr/lib/systemd/system/systemd-time-wait-sync.service.
TimeUSec=Thu 2024-08-08 23:51:15 JST, Waiting for NTP synchronization...
TimeUSec=Thu 2024-08-08 23:51:16 JST, Waiting for NTP synchronization...
...
TimeUSec=Thu 2024-08-08 23:51:41 JST, Waiting for NTP synchronization...
TimeUSec=Thu 2024-08-08 23:51:42 JST, Waiting for NTP synchronization...
TimeUSec=Tue 2024-11-12 11:43:37 JST, NTP synchronization complete.
NTPMessage={ Leap=0, Version=4, Mode=4, Stratum=2, Precision=-25, RootDelay=991us, RootDispersion=259us, Reference=11FD1CFB, OriginateTimestamp=Thu 2024-08-08 23:51:43 JST, ReceiveTimestamp=Tue 2024-11-12 11:43:36 JST, TransmitTimestamp=Tue 2024-11-12 11:43:36 JST, DestinationTimestamp=Thu 2024-08-08 23:51:43 JST, Ignored=no, PacketCount=1, Jitter=0 }
```

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
@@ -9,6 +9,11 @@ test ! -c /dev/rtc0 || exit 0
 # This script is intended for services running with systemd.
 command -v systemctl >/dev/null 2>&1 || exit 0
 
+echo_with_time_usec() {
+	time_usec=$(timedatectl show --property=TimeUSec)
+	echo "${time_usec}, ${1}"
+}
+
 # Enable `systemd-time-wait-sync.service` to wait for NTP synchronization at an earlier stage.
 systemctl enable systemd-time-wait-sync.service
 
@@ -16,19 +21,43 @@ systemctl enable systemd-time-wait-sync.service
 max_retry=60 retry=0
 until ntp_synchronized=$(timedatectl show --property=NTPSynchronized --value) && [ "${ntp_synchronized}" = "yes" ] ||
 	[ "${retry}" -gt "${max_retry}" ]; do
-	time_usec=$(timedatectl show --property=TimeUSec)
-	echo "${time_usec}, Waiting for NTP synchronization..."
+	if [ "${retry}" -eq 0 ]; then
+		# If /dev/rtc is not available, the system time set during the Linux kernel build is used.
+		# The larger the difference between this system time and the NTP server time, the longer the NTP synchronization will take.
+		# By setting the system time to the modification time of this script, which is likely to be closer to the actual time,
+		# the NTP synchronization time can be shortened.
+		echo_with_time_usec "Setting the system time to the modification time of ${0}."
+
+		# To set the time to a specified time, it is necessary to stop systemd-timesyncd.
+		systemctl stop systemd-timesyncd
+
+		# Since `timedatectl set-time` fails if systemd-timesyncd is not stopped,
+		# ensure that it is completely stopped before proceeding.
+		until pid_of_timesyncd=$(systemctl show systemd-timesyncd --property=MainPID --value) && [ "${pid_of_timesyncd}" -eq 0 ]; do
+			echo_with_time_usec "Waiting for systemd-timesyncd to stop..."
+			sleep 1
+		done
+
+		# Set the system time to the modification time of this script.
+		modification_time=$(stat -c %y "${0}")
+		echo_with_time_usec "Setting the system time to ${modification_time}."
+		timedatectl set-time "${modification_time}"
+
+		# Restart systemd-timesyncd
+		systemctl start systemd-timesyncd
+	else
+		echo_with_time_usec "Waiting for NTP synchronization..."
+	fi
 	retry=$((retry + 1))
 	sleep 1
 done
 # Print the result of NTP synchronization
 ntp_message=$(timedatectl show-timesync --property=NTPMessage)
-time_usec=$(timedatectl show --property=TimeUSec)
 if [ "${ntp_synchronized}" = "yes" ]; then
-	echo "${time_usec}, NTP synchronization complete."
+	echo_with_time_usec "NTP synchronization complete."
 	echo "${ntp_message}"
 else
-	echo "${time_usec}, NTP synchronization timed out."
+	echo_with_time_usec "NTP synchronization timed out."
 	echo "${ntp_message}"
 	exit 1
 fi
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
@@ -1,0 +1,25 @@
+#!/bin/sh
+set -eu
+
+# In vz, the VM lacks an RTC when booting with a kernel image (see: https://developer.apple.com/forums/thread/760344).
+# This causes incorrect system time until NTP synchronizes it, leading to TLS errors.
+# To avoid TLS errors, this script waits for NTP synchronization if RTC is unavailable.
+test ! -c /dev/rtc0 || exit 0
+
+# This script is intended for services running with systemd.
+command -v systemctl >/dev/null 2>&1 || exit 0
+
+# Enable `systemd-time-wait-sync.service` to wait for NTP synchronization at an earlier stage.
+systemctl enable systemd-time-wait-sync.service
+
+# For the first boot, where the above setting is not yet active, wait for NTP synchronization here.
+until ntp_synchronized=$(timedatectl show --property=NTPSynchronized --value) && [ "${ntp_synchronized}" = "yes" ]; do
+	time_usec=$(timedatectl show --property=TimeUSec)
+	echo "${time_usec}, Waiting for NTP synchronization..."
+	sleep 1
+done
+# Print the result of NTP synchronization
+ntp_message=$(timedatectl show-timesync --property=NTPMessage)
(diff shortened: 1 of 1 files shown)


norio-nomura:
> Log output during execution:

It seems there is no pattern to the time spent waiting for NTP synchronization. Sometimes it waits for about 30 seconds, while other times it doesn't wait at all because it is already synchronized.

pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
```diff
@@ -0,0 +1,25 @@
...
+test ! -c /dev/rtc0 || exit 0
+
+# This script is intended for services running with systemd.
+command -v systemctl >/dev/null 2>&1 || exit 0
+
+# Enable `systemd-time-wait-sync.service` to wait for NTP synchronization at an earlier stage.
+systemctl enable systemd-time-wait-sync.service
+
+# For the first boot, where the above setting is not yet active, wait for NTP synchronization here.
+until ntp_synchronized=$(timedatectl show --property=NTPSynchronized --value) && [ "${ntp_synchronized}" = "yes" ]; do
```
AkihiroSuda, line 16: The loop should give up after a few trials

AkihiroSuda: I also wonder if this should happen in background

norio-nomura:
opened issue #2905 regarding the problem when using a kernel image with `vz`.

nirs:
> In vz, the VM lacks an RTC when booting with a kernel image (see: https://developer.apple.com/forums/thread/760344). This causes incorrect system time until NTP synchronizes it, leading to TLS errors.

Why do we get TLS errors? can we avoid them without waiting for NTP update?

pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
nirs: Can we trigger synchronous NTP update instead of waiting for unpredictable time?

norio-nomura:
> Why do we get TLS errors? can we avoid them without waiting for NTP update?

The `curl` error message posted in #2905 is as follows:

```console
[    6.489211] cloud-init[770]: + curl -fsSL https://get.docker.com
[    6.558633] cloud-init[770]: curl: (60) SSL certificate problem: certificate is not yet valid
[    6.558703] cloud-init[770]: More details here: https://curl.se/docs/sslcerts.html
[    6.558761] cloud-init[770]: curl failed to verify the legitimacy of the server and therefore could not
[    6.558834] cloud-init[770]: establish a secure connection to it. To learn more about this situation and
[    6.558926] cloud-init[770]: how to fix it, please visit the web page mentioned above.
[    6.559955] cloud-init[770]: LIMA 2024-08-08T23:51:16+09:00| WARNING: Failed to execute /mnt/lima-cidata/provision.system/00000002
```

The system time at that point was approximately `2024-08-08T23:51:16+09:00`. Checking the certificate for `https://get.docker.com` shows:

```console
$ curl -w '%{certs}' https://get.docker.com -o /dev/null | grep date
100 22115  100 22115    0     0  74788      0 --:--:-- --:--:-- --:--:-- 74966
Start date:Sep  2 00:00:00 2024 GMT
Expire date:Oct  1 23:59:59 2025 GMT
Start date:Aug 23 22:25:30 2022 GMT
Expire date:Aug 23 22:25:30 2030 GMT
Start date:May 25 12:00:00 2015 GMT
Expire date:Dec 31 01:00:00 2037 GMT
```

The certificate appears to be valid from `Sep 2 00:00:00 2024 GMT`, so the system judged it as "not yet valid," causing the connection to fail.

#2905 also includes logs from `apt-get`, though the details there are unclear.

pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
norio-nomura: By the time this script is reached, the NTP sync for `systemd-timesyncd.service` has already been triggered. Additionally, `timedatectl` has no option to wait for NTP synchronization, which results in this type of loop.

nirs: So why it takes sometimes 30 seconds?

Do we have a way to start the instance using the current host time instead of depending on NTP? Since we rebuild the cidata.iso on each reboot, we can inject there the current host time, and have the guest clock starting at the cidata.iso build time. This should be in the worst case few seconds after the real start time.

nirs:
> ```
> curl -fsSL https://get.docker.com
> ```

Why not use -k to skip verification of the certificate? We don't really need to verify the server which we fully control. 

> ```
> Start date:Sep  2 00:00:00 2024 GMT
> Expire date:Oct  1 23:59:59 2025 GMT
> Start date:Aug 23 22:25:30 2022 GMT
> Expire date:Aug 23 22:25:30 2030 GMT
> Start date:May 25 12:00:00 2015 GMT
> Expire date:Dec 31 01:00:00 2037 GMT
> ```

Who create the certificate? can we create it with start date in the distant past, and expire date in the distant future, to ensure that it is always considered valid, even when the guest clock is broken?

nirs:
Ok, I see we fail to validate the remote certificate when accessing https://get.docker.com which we have no control of so we cannot avoid the verification.

nirs:
> > Log output during execution:
> 
> It seems there is no pattern to the time spent waiting for NTP synchronization. Sometimes it waits for about 30 seconds, while other times it doesn't wait at all because it is already synchronized.

Adding unpredictable 30 seconds wait when starting is not great, but if this happens only when starting with kernel image I think we are fine.

pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
nirs: Or maybe better, can we inject the start time using the kernel command line? This is relevant only when using kernel image which we use to allow using kernel command line.

AkihiroSuda:
> Sometimes it waits for about 30 seconds

Not specific to NTP, but we have to have a method to show the progress information in `limactl start` CLI

norio-nomura:
> but if this happens only when starting with kernel image I think we are fine.

As far as I know, `/dev/rtc0` is only unavailable when using a kernel image with vz.

norio-nomura:
By adding a process to set the script's modification time as the system time on the first iteration of the loop waiting for NTP synchronization, the loop now completes in about 1 second each time. It no longer takes up to 30 seconds.

pkg/cidata/cidata.TEMPLATE.d/boot/00-check-rtc-and-wait-ntp.sh
```diff
@@ -0,0 +1,60 @@
...
+}
+
+# For the first boot, where the above setting is not yet active, wait for NTP synchronization here.
+max_retry=60 retry=0
+until ntp_synchronized=$(timedatectl show --property=NTPSynchronized --value) && [ "${ntp_synchronized}" = "yes" ] ||
+	[ "${retry}" -gt "${max_retry}" ]; do
+	if [ "${retry}" -eq 0 ]; then
+		# If /dev/rtc is not available, the system time set during the Linux kernel build is used.
+		# The larger the difference between this system time and the NTP server time, the longer the NTP synchronization will take.
+		# By setting the system time to the modification time of this script, which is likely to be closer to the actual time,
```
AkihiroSuda, line 24: cidata/user-data might be closer

norio-nomura: When running this script, sub-second precision of the file's modification time is irrelevant since NTP synchronization is assumed. Using the script file itself ensures the file reliably exists.

AkihiroSuda: Potentially, a future version of Lima may install the cidata script without go:embed.

In that case, the timestamp of the cidata script will corresponds to the timestamp in the git repo, which may be years behind the current clock

norio-nomura: changed to use `user-data`

AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

PR #2900, opened 2024-11-13
vz: implement suspend/resume

norio-nomura:
- [x] Only `aarch64` is supported due to lack of support in `Code-Hex/vz/v3`.
- [x] Attempt to resume if the `vz-machine-state` file is found in the instance directory. If this fails, proceed with a cold boot.
- [x] On stop, always attempt to suspend to the `vz-machine-state` file; if this fails, perform a shutdown.
- [x] Guest OS and Guest Agent behavior are not considered here. Handling may be needed for suspending while port forwarding is active.

I will mark this as Ready for Review once the following are addressed:
- [ ] Add `.suspend = true` as an opt-in option in `limayaml`.
- [ ] Implement a `--cold-boot` option for `limactl start` if needed (since resume cannot occur if `vz-machine-state` doesn’t exist, a forced resume option should be unnecessary).
- [ ] Implement a `--shut-down` or `--suspend` option for `limactl stop` if required.

This implements part of #598.


norio-nomura:
Using suspend/resume shortens the startup time of the Docker VM I use from 37 seconds to 13 seconds.

afbjorklund:
Do you have an issue describing the feature?

Did you look at the PR

norio-nomura:
> Did you look at the PR #642

Yes, initially I started implementing this as an extension of #642, but since it didn’t align at all with the behavior of the API provided by `Virtualization.framework`, I implemented it from scratch.

norio-nomura:
> I implemented it from scratch.

Ah, however, this PR does not implement `limactl suspend` or `limactl resume`. Instead, an option is added to limayaml that allows start/stop to perform resume/suspend actions.

afbjorklund:
You might want to add that as a separate feature, i.e. describe the auto-suspend

templates/default.yaml
```diff
@@ -525,6 +525,11 @@ hostResolver:
 # 🟢 Builtin default: /usr/local
 guestInstallPrefix: null
 
+# When "suspend" is enabled, Lima suspends the VM instead of shutting it down.
+# The VM can be resumed with `limactl start` if the saved state is available.
+# 🟢 Builtin default: false
+suspend: null
```
AkihiroSuda, line 531: Do we need this yaml property?
Can we just have `limactl suspend` command? (Or `limactl stop --suspend`)

norio-nomura: It’s hard to imagine needing to run `limactl suspend` before restarting the host OS just to take advantage of suspend’s benefits.

nirs: Adding flags to stop makes it more complicated. I would go with adding suspend/resume commands and keep the ux simple.

You either do:

    limactl stop
    limactl start

Or:

    limactl suspend
    limactl resume

This is also virsh interface (using destroy instead of stop).

norio-nomura: I delegate the startup and shutdown of my Lima Docker instance to `launchd`, so the Docker instance starts and stops with macOS login and logout. In this PR, I aim to change that behavior to use resume/suspend instead.

For this, I want `hostagent` to suspend when it receives a SIGINT (sent by `limactl stop` or `launchd`) if the `.suspend` option is set to `true`, and to resume on launch if it was previously suspended.  
Adding `limactl (suspend|resume)` would require a separate mechanism to send/receive something other than SIGINT to handle suspension, increasing the complexity of the changes and not aligning with `launchd` shutdown handling. These aren’t changes I’m interested in.

norio-nomura: > Do we need this yaml property?

Passing it as a command-line option to `limactl start` that would then be forwarded to `hostagent` could work as well. However, I’m also considering setting `.suspend = true` in `override.conf`, so I’d prefer to keep it as an option in `limayaml`.

norio-nomura: > Adding `limactl (suspend|resume)` would require a separate mechanism to send/receive something other than SIGINT to handle suspension,

Ah, for my request to add `--shut-down` or `--suspend` options to `limactl stop`, communication beyond SIGINT would indeed be required. In that case, I feel that adding `limactl suspend` could be a good solution. 

By the way, what would be the best method to send a request to the hostagent? Should I use a signal other than SIGINT, or a socket?

jandubois: Shouldn't this just use the gRPC communication between guest and hostagent?

jandubois: I don't like the additional `resume` command. I don't care if the VM is suspended or stopped, I should always be able to use `start` to get it running again.

I think SIGINT should suspend a running VM (if possible), and `stop` should stop it (`stop` would no longer send SIGINT, but uses grpc to do it).

Now the only missing functionality would be to suspend a VM by command and not by sending SIGINT. This could be done by a `suspend` subcommand, but I think `stop --suspend` works better, given that we won't have a `resume` command.

norio-nomura: The communication I'm referring to is the instruction from `limactl suspend` to `limactl hostagent` to trigger suspend.  
`limactl stop` uses SIGINT.  
Currently, I'm considering sending a request with `.suspend = true` to `ha.sock`, followed by sending SIGINT.  
Also, I am thinking that it might be better to name it `.suspendOnStop` instead of `.suspend`.

jandubois: Sorry, I was confused and thought this was between host and guestagent.

norio-nomura: I plan to try the following implementation:

- Add `.suspend`, `.suspendOnStop`, or `.suspendOnSigInt` to `limayaml`.
- `limactl resume`: Not needed.
- `limactl start`: If the suspend state is still present, resume.
- `limactl suspend`: Request `.suspendOnSigInt = true` to `ha.sock` and send SIGINT to `hostagent`.
- `limactl stop`: Request `.suspendOnSigInt = false` to `ha.sock` and send SIGINT to `hostagent`.
- `hostagent`: Upon receiving SIGINT, if `.suspend` is true, suspend; if false, shutdown.

jandubois: I don't understand why you need the extra field in `limayaml`.

I think I would just use SIGUSR1 to mean `stop` and leave SIGINT to mean `suspend`. That way everything works automatically, including your `launchd` scenario, and you can still explicitly shut down the VM when you want to.

jandubois: On further thoughts I think `limactl stop` shut do a suspend, and `limactl stop --shutdown` (and `limactl stop -f`) should shut down and not suspend.

Eventually we could add a `limactl reboot` (or `restart`) command because I think the only time you really want to shutdown the instance is when you want a fresh boot.

jandubois: That is a good point, I agree.

norio-nomura:
In addition to `vz-machine-state`, it might be necessary to save the `cidata.iso` and `lima.yaml` at the point of suspend for use during resume.

norio-nomura:
> In addition to `vz-machine-state`, it might be necessary to save the `cidata.iso` and `lima.yaml` at the point of suspend for use during resume.

And to apply the latest `lima.yaml` to the guest, a cold boot might be required.

jandubois:
> to apply the latest `lima.yaml` to the guest, a cold boot might be required.

`limactl edit` already requires the machine to be stopped.

norio-nomura:
> > to apply the latest `lima.yaml` to the guest, a cold boot might be required.
> 
> `limactl edit` already requires the machine to be stopped.

I always edit `lima.yaml` in VSCode. 😝

norio-nomura:
> Eventually we could add a `limactl reboot` (or `restart`) command because I think the only time you really want to shutdown the instance is when you want a fresh boot.

A restart command might be needed sooner than I expected. Currently, I restart using:  
```console
launchctl kickstart -kp gui/501/io.lima-vm.autostart.docker; tail -F ~/.lima/docker/{ha.stderr,launchd.stderr,serialv}.log
```  
However, with `launchctl`, it ends up performing suspend/resume instead, making it impossible to actually restart.

jandubois:
> I always edit `lima.yaml` in VSCode. 😝

So how would you deal with this?

When you `suspend`, but the `lima.yaml` file has been edited more recently than the `cidata.iso`, then you will touch the `lima.yaml` file to make sure it is newer than the saved state.

And when you `start`, and the `lima.yaml` has been edited more recently than the `vz-machine-state`, then you will ask the user if they want to drop the suspended state?

I guess that works, but is maybe too magical?

norio-nomura:
Perhaps it would be better to shut down instead of suspending if the `lima.yaml` has been modified since startup?

norio-nomura:
> And when you `start`, and the `lima.yaml` has been edited more recently than the `vz-machine-state`, then you will ask the user if they want to drop the suspended state?

Exactly. For that, I’m thinking of saving the `lima.yaml` and `cidata.iso` from the time of startup. Resume would always use those saved versions, while only a cold boot would use the latest `lima.yaml`.

jandubois:
> Perhaps it would be better to shut down instead of suspending if the `lima.yaml` has been modified since startup?

Maybe, but you would lose all your running state/workloads, so shouldn't there be a prompt?

That's why `limactl edit` requires you to stop first.

Now that I think about it more, I think if you edit `lima.yaml` outside of `limactl edit`, then it is your responsibility to do a `limactl reboot` to trigger the changes to take effect. Lima itself should continue to suspend/resume without worrying about your changes.

norio-nomura:
Would it be better if `limactl suspend` detects changes to `lima.yaml` and prompts for either `limactl stop` or `limactl reboot`, instead of having `hostagent` detect it when receiving SIGINT?

jandubois:
> Would it be better if `limactl suspend` detects changes to `lima.yaml` and prompts for either `limactl stop` or `limactl reboot`

This does not work if the suspend happens in the background, e.g. because you are logging out. Same thing when you automatically resume. In both case you have no chance to prompt the user.

If the user edits `lima.yaml` without stopping the instance, then the user is responsible for eventually applying the changes by requesting a shutdown. We can ask when the instance is stopped or resumed, and stdin/stdout are both connected to a TTY. But otherwise the changes should be ignored.

norio-nomura:
Updated PR description.

templates/default.yaml
```diff
@@ -525,6 +525,11 @@ hostResolver:
 # 🟢 Builtin default: /usr/local
 guestInstallPrefix: null
 
+# When "suspendOnSigInt" is enabled, Lima suspends the VM instead of shutting it down.
+# The VM can be resumed with `limactl start` if the saved state is available.
+# 🟢 Builtin default: false
+suspendOnSigInt: null
```
nirs, line 531: This is not a good name for user interface. We are leaking questionable implementation detail (stopping using SIGINT instead of SIGTERM). The user configuration should not leak such details.

Using suspend for this feature will make it hard to support suspend and resume as used by libvirt:
- https://www.libvirt.org/manpages/virsh.html#suspend
- https://www.libvirt.org/manpages/virsh.html#resume

When a vm is suspended, the qemu process is stopped (I guess with SIGSTOP). When resuming it continues to run (I guess with SIGCONT). Both suspend and resume are very quick. This is a very useful feature for testing, when you want to simulate issues like inaccessible vm.

What vz call suspend is called "save" in libvit:
https://www.libvirt.org/manpages/virsh.html#save

This is much slower, writing and compressing guest memory to disk. This should not be the default behavior since it slow, requires lot of resources, and fragile. For example it will not work if the disks are modified after saving the guest state.

To avoid all the issues, we can call this option `autoSave` or `saveOnStop`. If enabled, instance state is saved when stopping the instance and restored when starting the instance.

pkg/vz/suspend_resume.go
nirs, line 1: save_restore.go? persist.go?

pkg/vz/suspend_resume.go
```diff
@@ -0,0 +1,18 @@
...
+package vz
+
+import (
+	"fmt"
+	"runtime"
+
+	"github.com/Code-Hex/vz/v3"
+)
+
+func suspendVM(vm *vz.VirtualMachine, machineStatePath string) error {
```
nirs, line 12: saveVM?

pkg/vz/suspend_resume.go
```diff
@@ -0,0 +1,18 @@
...
+	return fmt.Errorf("suspend is not supported on the vz driver for the architecture %s", runtime.GOARCH)
+}
+
+func resumeVM(vm *vz.VirtualMachine, machineStatePath string) error {
```
nirs, line 16: restoreVM?

nirs (commented):
Automatically saving guest state is nice improvement but it should not be the default due to resource usage and the fact that it is not needed or wanted in many cases.

We should not use suspend/resume for this feature to allow adding support for quick suspend/resume that do not stop the instance. This also match better the VZ API:
- https://developer.apple.com/documentation/virtualization/vzvirtualmachine/4168516-savemachinestatetourl?language=objc
- https://developer.apple.com/documentation/virtualization/vzvirtualmachine/4168515-restoremachinestatefromurl?language=objc

templates/default.yaml
nirs: Having an option in the yaml and in the command line sounds like best way:
- If you always want the feature, you use change the default or override yaml
- If you want the feature only for some vms, you enable it in the instance yaml
- If you want the feature only in some cases, you enable it when stopping
- If works like other features, easier to use, easier to maintain

templates/default.yaml
norio-nomura: > To avoid all the issues, we can call this option `autoSave` or `saveOnStop`. If enabled, instance state is saved when stopping the instance and restored when starting the instance.

I prefer `saveOnStop`.

norio-nomura: I agree that aligning the terminology with libvirt makes sense.

norio-nomura:
updated commits and PR description to aligning the terminology with libvirt:
- suspend -> save
- resume -> restore
- `.suspendOnSigInt` -> `.saveOnStop`

afbjorklund:
Why not use the same names as in the qemu implementation?

norio-nomura:
> Why not use the same names as in the qemu implementation?

Because its behavior differs from the QEMU implementation.  
In the QEMU implementation, suspend does not save the VM state to disk, and resume does not restore the saved VM state.

afbjorklund:
Right, then it is a good thing to use different names. Actually the commands were called STOP and CONT in qemu.

For suspending to disk, the state is **also** called SUSPEND (standby/S3) or SUSPEND_DISK (hibernate/S4). My bad.

norio-nomura:
The changes to ensure that the same `lima.yaml` and `cidata.iso` are used for save/restore might potentially evolve into a snapshot implementation. Therefore, I would like to address those in a separate PR.

pkg/vz/save_restore_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+	if !vm.CanPause() {
+		return fmt.Errorf("can't pause the VZ machine")
+	}
+
+	// Remove the old machine state file if it exists,
+	// because saving the machine state will fail if the file already exists.
+	if err := os.Remove(machineStatePath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		logrus.WithError(err).Errorf("Failed to remove the old VZ machine state file %q", machineStatePath)
+		return err
+	}
```
nirs, line 24: We can write machine state to `vz-machine-state.tmp` and then replace the current state by renaming `vz-machine-state.tmp` to `vz-machine-state`. This ensures that we always have consistent state, either the previous one or the new one.

pkg/vz/save_restore_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+	logrus.Info("Pausing VZ")
+	if err := vm.Pause(); err != nil {
+		return err
+	}
+
+	// If we can't stop the machine after pausing, saving the machine state will be useless.
+	// So we should check this before saving the machine state.
+	if !vm.CanStop() {
+		return fmt.Errorf("can't stop the VZ machine after pausing")
+	}
```
nirs, line 35: If we fail here the machine is left in paused state. Can we resume it to return to normal operation?

pkg/vz/save_restore_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Remove the old machine state file if it exists,
+	// because saving the machine state will fail if the file already exists.
+	if err := os.Remove(machineStatePath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		logrus.WithError(err).Errorf("Failed to remove the old VZ machine state file %q", machineStatePath)
+		return err
+	}
+
+	logrus.Info("Pausing VZ")
```
nirs, line 26: Can we have more descriptive log message?

pkg/vz/save_restore_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+		return err
+	}
+
+	// If we can't stop the machine after pausing, saving the machine state will be useless.
+	// So we should check this before saving the machine state.
+	if !vm.CanStop() {
+		return fmt.Errorf("can't stop the VZ machine after pausing")
+	}
+
+	logrus.Info("Saving VZ machine state for resuming later")
```
nirs, line 37: for restoring later?

pkg/vz/save_restore_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+	if err := vm.SaveMachineStateToPath(machineStatePath); err != nil {
+		// If we fail to save the machine state, we should resume the machine to call RequestStop() later
+		logrus.WithError(err).Errorf("Failed to save the machine state to %q", machineStatePath)
+		if resumeError := vm.Resume(); resumeError != nil {
+			return resumeError
+		}
```
nirs, line 43: We have mulltiple error handlers calling resume. I think we can have one error handling resuming the machine. One way to implement this to more all the functions that need to resume on errors to separate function and resume when the function fail, maybe with certain errors.

pkg/vz/save_restore_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+		return err
+	}
+	return nil
+}
+
+func restoreVM(vm *vz.VirtualMachine, machineStatePath string) error {
+	if _, err := os.Stat(machineStatePath); err != nil {
+		return err
+	}
+	logrus.Info("Saved VZ machine state found, resuming VZ")
```
nirs, line 63: Maybe:

    logrus.Infof("Resuming VZ machine from %q", machineStatePath)

pkg/vz/save_restore_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+		return err
+	}
+	if err := vm.Resume(); err != nil {
+		return err
+	}
+	// Remove the machine state file after resuming the machine
+	if err := os.Remove(machineStatePath); err != nil {
+		// We should log the error but continue the process, because the machine state is already restored
+		logrus.WithError(err).Errorf("Failed to remove the VZ machine state file %q", machineStatePath)
+	}
```
nirs, line 74: Should we always remove the saved state? what if you want to always start with the same saved state?

pkg/vz/save_restore_arm64.go
norio-nomura: Currently, I don't see the value in keeping the `vz-machine-state` after a resume. This is because the current implementation does not save disk snapshots, and there is no mechanism to revert the guest OS's changes to the diffdisk after a resume to restore the state to match when the `vz-machine-state` was saved. 

By the time we reach this point, the guest OS should already have gone through `StatusRunning`, so I don't think it's worth complicating things to maintain consistency with the previous `vz-machine-state`.

pkg/vz/save_restore_arm64.go
norio-nomura: > what if you want to always start with the same saved state?

As I mentioned in a reply to another comment earlier, achieving that would require a snapshot feature that includes saving and restoring the diffdisk.

pkg/vz/vz_driver_darwin.go
```diff
@@ -192,6 +193,13 @@ func (l *LimaVzDriver) RunGUI() error {
 }
 
 func (l *LimaVzDriver) Stop(_ context.Context) error {
+	machineStatePath := filepath.Join(l.Instance.Dir, filenames.VzMachineState)
+	if err := saveVM(l.machine.VirtualMachine, machineStatePath); err != nil {
+		logrus.WithError(err).Warn("Failed to save VZ. Falling back to shutdown")
```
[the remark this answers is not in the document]
norio-nomura: The commit this comment refers to has been force updated and is now outdated. Could you please check the latest commit again?

pkg/vz/vz_driver_darwin.go
```diff
@@ -193,11 +205,13 @@ func (l *LimaVzDriver) RunGUI() error {
...
-		logrus.WithError(err).Warn("Failed to save VZ. Falling back to shutdown")
-	} else {
-		return nil
+	if l.config.SaveOnStop {
+		machineStatePath := filepath.Join(l.Instance.Dir, filenames.VzMachineState)
+		if err := saveVM(l.machine.VirtualMachine, machineStatePath); err != nil {
+			logrus.WithError(err).Warn("Failed to save VZ. Falling back to shutdown")
+		} else {
+			return nil
+		}
```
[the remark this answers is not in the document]
norio-nomura: > The code to add saveOnStop must be added before you change Stop().

That should already be the case. Could you double-check to ensure you're not looking at an outdated commit?

cmd/limactl/copy.go
```diff
@@ -77,6 +77,8 @@ func copyAction(cmd *cobra.Command, args []string) error {
 			}
 			if inst.Status == store.StatusStopped {
 				return fmt.Errorf("instance %q is stopped, run `limactl start %s` to start the instance", instName, instName)
+			} else if inst.Status == store.StatusSaved {
+				return fmt.Errorf("instance %q is saved, run `limactl start %s` to restore the instance", instName, instName)
```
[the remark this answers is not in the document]
norio-nomura: Got it! That approach sounds good. I'll give it a try.

norio-nomura:
I've applied the review suggestions.  
I made changes to past commits and force-pushed, so please fetch the latest and review again.

pkg/vz/save_restore_arm64.go
norio-nomura: I have moved this part to where resume is needed after an error.

pkg/vz/save_restore_arm64.go
norio-nomura: Thank you. By consolidating the processing that requires resume after an error into `savePausedVM`, the error handling became much cleaner.

cmd/limactl/copy.go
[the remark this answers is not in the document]
norio-nomura: I removed the commit that added `StatusSaved` and replaced it with a commit that adds `Instance.Saved`.

norio-nomura:
I force-pushed to remove a stray code block that had no effect.

pkg/vz/save_restore_arm64.go
nirs: Makes sense. Maybe add a comment here to explain why we don't keep the state?

pkg/vz/save_restore_arm64.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+func saveVM(vm *vz.VirtualMachine, machineStatePath string) error {
+	if !vm.CanPause() {
+		return fmt.Errorf("can't pause the VZ machine")
+	}
+
+	// Remove the old machine state file if it exists,
+	// because saving the machine state will fail if the file already exists.
```
nirs, line 20: Removing to avoid failure on save is not a good reson, save can replace existing file.

pkg/vz/save_restore_arm64.go
nirs: Right, we can add this later.

pkg/vz/vz_driver_darwin.go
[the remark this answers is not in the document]
nirs: This is still the case - the first commit ignore the option added in the second commit. If we need to revert the second commit we change the behavior to always save.

pkg/store/instance.go
```diff
@@ -43,6 +43,7 @@ type Instance struct {
 	// Hostname, not HostName (corresponds to SSH's naming convention)
 	Hostname        string             `json:"hostname"`
 	Status          Status             `json:"status"`
+	Saved           bool               `json:"saved"`
```
nirs, line 46: Can document here what saved mean?

pkg/hostagent/api/client/client.go
```diff
@@ -62,3 +65,35 @@ func (c *client) Info(ctx context.Context) (*api.Info, error) {
...
+	method := "GET"
+	var body io.Reader
+	if config != nil {
+		method = "PATCH"
+		b, err := json.Marshal(config)
+		if err != nil {
+			return nil, err
+		}
+		body = bytes.NewBuffer(b)
+	}
```
nirs, line 80: Doing both GET and PATCH in the same function is confusing, both for understanding this code, and both for code using this function. If we want to avoid duplicate code for sending a request and getting a response, move the common code to a helper.

pkg/hostagent/api/client/client.go
```diff
@@ -16,6 +18,7 @@ import (
 type HostAgentClient interface {
 	HTTPClient() *http.Client
 	Info(context.Context) (*api.Info, error)
+	DriverConfig(context.Context, interface{}) (interface{}, error)
```
nirs, line 21: Would more clear as:

    // Get current driver config
    DriveConfig(context.Context) (interface{}, error)

    // Patch runtime config and return updated config
    PatchDriveConfig(context.Context, interface{}) (interface{}, error)

pkg/hostagent/api/server/server.go
```diff
@@ -50,6 +50,36 @@ func (b *Backend) GetInfo(w http.ResponseWriter, r *http.Request) {
...
+		b.onError(w, err, http.StatusInternalServerError)
+		return
+	}
+	m, err := json.Marshal(config)
+	if err != nil {
+		b.onError(w, err, http.StatusInternalServerError)
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
```
nirs, line 77: Isn't OK the default?

pkg/hostagent/api/server/server.go
```diff
@@ -50,6 +50,36 @@ func (b *Backend) GetInfo(w http.ResponseWriter, r *http.Request) {
...
+	_, _ = w.Write(m)
+}
+
 func AddRoutes(r *http.ServeMux, b *Backend) {
 	r.Handle("/v1/info", http.HandlerFunc(b.GetInfo))
+	r.Handle("GET /v1/driver/config", http.HandlerFunc(b.DriverConfig))
+	r.Handle("PATCH /v1/driver/config", http.HandlerFunc(b.DriverConfig))
```
nirs, line 84: The name of the new handler is not consistent with existing GetInfo. GetConfig and PatchConfig would be consistent.

pkg/instance/stop.go
```diff
@@ -9,18 +9,42 @@ import (
...
 	hostagentevents "github.com/lima-vm/lima/pkg/hostagent/events"
+	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/sirupsen/logrus"
 )
 
-func StopGracefully(inst *store.Instance) error {
+func StopGracefully(inst *store.Instance, saveOnStop bool) error {
```
nirs, line 21: Maybe `save bool` or `saveMachineState bool`?

pkg/instance/stop.go
```diff
@@ -9,18 +9,42 @@ import (
...
 	if inst.Status != store.StatusRunning {
 		return fmt.Errorf("expected status %q, got %q (maybe use `limactl stop -f`?)", store.StatusRunning, inst.Status)
 	}
 
+	if saveOnStop && inst.Saved {
+		logrus.Warn("saved VZ machine state is found. It will be overwritten by the new one.")
```
nirs, line 27: Why warning? this is expected condition, nothing to warn about. Use Debug?

pkg/instance/stop.go
```diff
@@ -9,18 +9,42 @@ import (
...
+
+	if inst.VMType == limayaml.VZ {
+		haSock := filepath.Join(inst.Dir, filenames.HostAgentSock)
+		haClient, err := hostagentclient.NewHostAgentClient(haSock)
+		if err != nil {
+			logrus.WithError(err).Error("Failed to create a host agent client")
+		}
+		ctx, cancel := context.WithTimeout(context.TODO(), 3*time.Second)
+		defer cancel()
+		disableSaveOnStopConfig := struct {
```
nirs, line 38: Why is this "disableSaveOnStopConfig"? why not patch?

pkg/instance/stop.go
```diff
@@ -9,18 +9,42 @@ import (
...
+			SaveOnStop bool `json:"saveOnStop"`
+		}{SaveOnStop: saveOnStop}
+		_, err = haClient.DriverConfig(ctx, disableSaveOnStopConfig)
+		if err != nil {
+			return fmt.Errorf("failed to disable saveOnStop: %w", err)
```
nirs, line 43: Why are we treating this as disable when the value can be true?

pkg/instance/stop.go
```diff
@@ -9,18 +9,42 @@ import (
...
+		}
+	} else if saveOnStop {
+		return fmt.Errorf("save is not supported for %q", inst.VMType)
```
nirs, line 46: Doing input validation first is better and more clear, something like:

    if saveOnStop && inst.VMType != limayaml.VZ {
        return fmt.Errorf(...)
    }

This keeps the happy path very clear, and it is easy to very that we did not forget to handle all abnormal cases.

pkg/vz/vz_driver_darwin.go
```diff
@@ -204,6 +207,34 @@ func (l *LimaVzDriver) RunGUI() error {
...
+	if newConfig.SaveOnStop {
+		if runtime.GOARCH != "arm64" {
+			return nil, fmt.Errorf("saveOnStop is not supported on %s", runtime.GOARCH)
+		} else if runtime.GOOS != "darwin" {
+			return nil, fmt.Errorf("saveOnStop is not supported on %s", runtime.GOOS)
+		} else if macOSProductVersion, err := osutil.ProductVersion(); err != nil {
+			return nil, fmt.Errorf("failed to get macOS product version: %w", err)
+		} else if macOSProductVersion.Major < 14 {
+			return nil, fmt.Errorf("saveOnStop is not supported on macOS %d", macOSProductVersion.Major)
+		}
```
nirs, line 228: This looks like a copy of the block validating the config in a previous commit. Maybe we can have one place doing this check:

    save_restore.Validate(instance) error

Then we can call it in both places,

cmd/limactl/save.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/lima-vm/lima/pkg/instance"
+	networks "github.com/lima-vm/lima/pkg/networks/reconcile"
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/spf13/cobra"
+)
+
+func newSaveCommand() *cobra.Command {
+	saveCmd := &cobra.Command{
+		Use:               "save INSTANCE",
+		Short:             "Save an instance",
```
nirs: This also stop the instance but the name or the help do not explain this. I think this is confusing user interface even if we improve the help text, and it is better to have:

    limactl stop --save

Using --save and --force should be an error.

This also more consistent with saveOnStop instance config. You either enable save on stop for the instance, or for specific stop of the instance.

pkg/limayaml/validate.go
```diff
@@ -523,4 +523,7 @@ func warnExperimental(y *LimaYAML) {
 	if y.MountInotify != nil && *y.MountInotify {
 		logrus.Warn("`mountInotify` is experimental")
 	}
+	if y.SaveOnStop != nil && *y.SaveOnStop {
+		logrus.Warn("`saveOnStop` is experimental")
+	}
```
nirs, line 528: This makes sense but we should add it to the commit adding the option. Do yo plan to squash all the commits to one commit later?

norio-nomura:
I have other priorities to focus on, so I converted this to a draft for now.

AkihiroSuda:
Needs rebase

msimkunas:
@AkihiroSuda How do Lima VMs created with the VZ driver behave when a M-series Mac suspends/sleeps and wakes up? Does Apple's VZ handle this automatically?

Are these save/restore commands discussed in this issue only needed to allow the user to save VM state manually? I'm assuming that suspending the laptop should not result in undefined behavior/corrupted VMs?

PR #2906, opened 2024-11-14
Default `guestIP` to `0.0.0.0` when `ignore` is true

### LINKED ISSUE #2901: portForwards ignore not respected without guestIP: "0.0.0.0"
### Description

In lima 1.0.0 we could disable port forwarding (except ssh port) using:

```yaml
portForwards:
  - ignore: true
    proto: any 
```

In lima 1.0.1 this seems to be broken. The port rule is detected, but then we see logs about port forwarding.

```
% limactl start --tty=0 test.yaml
...
INFO[0012] [hostagent] TCP (except for SSH) and UDP port forwarding is disabled 
...
INFO[0089] [hostagent] Forwarding TCP from [::]:10250 to 127.0.0.1:10250 
INFO[0092] [hostagent] Forwarding TCP from [::]:6443 to 127.0.0.1:6443 
```

I think this is the change in 1.0.1 to disable the grpc port forwarding. The ignore rule probably never worked with the ssh forwarder.

Can be reproduced by this k8s based yaml:
[test.tar.gz](https://github.com/user-attachments/files/17738230/test.tar.gz)

jandubois:
This allows the user to disable all port forwarding with

```yaml
portForwards:
- proto: any
  ignore: true
```

Fixes #2901

### DIFF
--- a/templates/default.yaml
+++ b/templates/default.yaml
@@ -421,7 +421,7 @@ networks:
 # # default: hostPortRange: [1, 65535]
 #
 # - guestPort: 8888
-#   ignore: true (don't forward this port)
+#   ignore: true (don't forward this port, guestIP will default to 0.0.0.0 in this case since Lima 1.0.2)
 #
 # - guestPort: 7443
 #   guestIP: "0.0.0.0"       # Will match *any* interface
--- a/hack/test-port-forwarding.pl
+++ b/hack/test-port-forwarding.pl
@@ -222,8 +222,7 @@ portForwards:
   hostPortRange: [2000, 2009]
   ignore: true
 
-- guestIP: 0.0.0.0
-  guestPortRange: [3010, 3019]
+- guestPortRange: [3010, 3019]
   hostPortRange: [2010, 2019]
   ignore: true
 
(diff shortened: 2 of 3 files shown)


AkihiroSuda:
Can we have a test?

templates/default.yaml
```diff
@@ -421,7 +421,7 @@ networks:
 # # default: hostPortRange: [1, 65535]
 #
 # - guestPort: 8888
-#   ignore: true (don't forward this port)
+#   ignore: true (don't forward this port, guestIP will default to 0.0.0.0 in this case)
```
AkihiroSuda, line 424:
```suggestion
#   ignore: true (don't forward this port, guestIP will default to 0.0.0.0 in this case since Lima v1.0.2)
```

nirs:
Having simpler way to ignore port forwarding is nice, but it is less consistent. I think consistency is more important than making the rule for ignoring everything little bit simpler.

I would go with documenting that guestIP must be "0.0.0.0" or provide an alias like "any", so the rule to ignore all ports would be:

```yaml
portForwards:
  - ignore: true
    proto: any
    guestIP: any
```

jandubois:
> I would go with documenting that guestIP must be "0.0.0.0" or provide an alias like "any", so the rule to ignore all ports would be:

Sounds good to me too, but I think adding an `any` alias is not worth the effort, as the field is currently https://github.com/lima-vm/lima/blob/77cbbda09a1d565f76aceef3f89da00c7949bff3/pkg/limayaml/limayaml.go#L250

Would have to be turned into a `string`, then needs validation code, etc., for basically just a redundant alias.

jandubois:
Closing because it is not backwards compatible, and requiring the user to specify the `0.0.0.0` guest IP isn't that onerous.

Issue #2791, opened 2024-10-24
Various LIMA_CIDATA_* variables are not documented


jandubois:
The following variables are missing from `website/content/en/docs/dev/internals/_index.md`:

```
LIMA_CIDATA_DISKS
LIMA_CIDATA_DISK_%d_NAME
LIMA_CIDATA_DISK_%d_DEVICE
LIMA_CIDATA_DISK_%d_FORMAT
LIMA_CIDATA_DISK_%d_FSTYPE
LIMA_CIDATA_DISK_%d_FSARGS
LIMA_CIDATA_GUEST_INSTALL_PREFIX
LIMA_CIDATA_ROSETTA_ENABLED
LIMA_CIDATA_ROSETTA_BINFMT
LIMA_CIDATA_SKIP_DEFAULT_DEPENDENCY_RESOLUTION
LIMA_CIDATA_VMTYPE
LIMA_CIDATA_VSOCK_PORT
LIMA_CIDATA_VIRTIO_PORT
LIMA_CIDATA_PLAIN
```

PR #2907, opened 2024-11-14
Show that guestIP should be specified to ignore ports for forwarding

jandubois:
Alternative for #2906

### DIFF
--- a/templates/default.yaml
+++ b/templates/default.yaml
@@ -420,8 +420,9 @@ networks:
 # # default: guestPortRange: [1, 65535]
 # # default: hostPortRange: [1, 65535]
 #
-# - guestPort: 8888
-#   ignore: true (don't forward this port)
+# - guestIP: 0.0.0.0 # otherwise defaults to 127.0.0.1
+#   proto: any       # tcp and udp
+#   ignore: true     # don't forward these ports (guestPortRange, in this case 1-65535)
 #
 # - guestPort: 7443
 #   guestIP: "0.0.0.0"       # Will match *any* interface


templates/default.yaml
```diff
@@ -420,8 +420,10 @@ networks:
 # # default: guestPortRange: [1, 65535]
 # # default: hostPortRange: [1, 65535]
 #
-# - guestPort: 8888
-#   ignore: true (don't forward this port)
+# - guestIP: 0.0.0.0 # otherwise defaults to 127.0.0.1
+# # default: guestPortRange: [1, 65535]
+#   proto: any       # tcp and udp
+#   ignore: true     # don't forward this port
```
nirs, line 426: This format is very confusing, mixing incline comments and comment lines. It will be much more clear like this:

```yaml
# Disable port forwarding (any protocol, any guest IP):
# - ignore: true
#   proto: any
#   guestIP: "0.0.0.0"
```

jandubois: > This format is very confusing, mixing incline comments and comment lines.

It is supposed to be valid YAML if you just uncomment the lines by deleting the first 2 characters. So the first line would still need a double comment marker:

```yaml
# # Disable port forwarding (any protocol, any guest IP):
# - ignore: true
#   proto: any
#   guestIP: "0.0.0.0"
```

We just don't have a comment on top for any of the other examples in that section, so I think it is still confusing, but I'm fine changing it, if you think it is better. Should probably add explanations to all the example rules then.

Ultimately I think we should have a page in the docs that documents all the individual settings, and then shows some examples in a cookbook style. The only problem is that it then becomes even more likely that documentation won't get kept in sync with the implementation.

nirs:
We also need to think about this message:

```console
INFO[0012] [hostagent] TCP (except for SSH) and UDP port forwarding is disabled 
```

And the special behavior not to log about "Not forwarding port". Maybe both should be used only if guestIP is "0.0.0.0"?

templates/default.yaml
nirs: > > This format is very confusing, mixing incline comments and comment lines.
> 
> It is supposed to be valid YAML if you just uncomment the lines by deleting the first 2 characters. So the first line would still need a double comment marker:
> 
> ```yaml
> # # Disable port forwarding (any protocol, any guest IP):
> # - ignore: true
> #   proto: any
> #   guestIP: "0.0.0.0"
> ```

It is valid if you use only single comment for the comment line, and it looks better. 

> We just don't have a comment on top for any of the other examples in that section, so I think it is still confusing, but I'm fine changing it, if you think it is better. Should probably add explanations to all the example rules then.

True, this change should be done for the entire file. We can keep the current format for now to keep the file more consistent.

> Ultimately I think we should have a page in the docs that documents all the individual settings, and then shows some examples in a cookbook style. 

Yes!

> The only problem is that it then becomes even more likely that documentation won't get kept in sync with the implementation.

If the documentation is part of the source I don't think this is an issue. You have to remember to update templates/default.yaml or docs/cookbook.md, what is the difference?

jandubois: > If the documentation is part of the source I don't think this is an issue. You have to remember to update templates/default.yaml or docs/cookbook.md, what is the difference?

Shouldn't be, but I just noticed #2791 recently, so obviously we do have a problem updating the docs. We just need to all train ourselves to look for it in code reviews.

jandubois: > It is valid if you use only single comment for the comment line, and it looks better.

The first line is not valid YAML if you uncomment the whole block (remove the first two columns):

CleanShot 2024-11-14 at 13 49 57@2x

This is something you can do with many IDEs by marking the block and pressing e.g `⌘-/` (JetBrains) or `⌘-;` (Emacs).

Anyways, it isn't super important, and I already agreed that the current format is not easy to understand unless you already know what it means.

jandubois: I've removed the `guestPortRange` double-comment and folded it into the comment on `ignore`.

I'm not willing to put the top comment in that breaks the current schema unless we rework all of the examples.

nirs (approved):

nirs:
You can link this to #2901

PR #2908, opened 2024-11-14
Extract options.apply helper

nirs:
Extracting a helper to apply Opt function to options to simplify and clarify downloader.Download(). This remove duplicate loops repeated 4 times in this package.


jandubois (approved):
Thanks, LGTM

PR #2922, opened 2024-11-19
Consistently connect to Unix sockets

tamird:
See commit message.

### DIFF
--- a/pkg/networks/usernet/client.go
+++ b/pkg/networks/usernet/client.go
@@ -139,8 +139,9 @@ func NewClient(endpointSock string, subnet net.IP) *Client {
 func create(sock string, subnet net.IP, base string) *Client {
 	client := &http.Client{
 		Transport: &http.Transport{
-			DialContext: func(context.Context, string, string) (net.Conn, error) {
-				return net.Dial("unix", sock)
+			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
+				var d net.Dialer
+				return d.DialContext(ctx, "unix", sock)
 			},
 		},
 	}
--- a/pkg/qemu/qemu_driver.go
+++ b/pkg/qemu/qemu_driver.go
@@ -417,9 +417,9 @@ func (a *qArgTemplateApplier) applyTemplate(qArg string) (string, error) {
 				if !ok {
 					return "", fmt.Errorf("non-string argument %+v", v)
 				}
-				addr := &net.UnixAddr{
-					Net:  "unix",
-					Name: s,
+				addr, err := net.ResolveUnixAddr("unix", s)
+				if err != nil {
+					return "", err
 				}
 				conn, err := net.DialUnix("unix", nil, addr)
 				if err != nil {
--- a/pkg/vz/network_darwin.go
+++ b/pkg/vz/network_darwin.go
@@ -18,7 +18,11 @@ import (
 )
 
 func PassFDToUnix(unixSock string) (*os.File, error) {
-	unixConn, err := net.Dial("unix", unixSock)
+	unixAddr, err := net.ResolveUnixAddr("unix", unixSock)
+	if err != nil {
+		return nil, err
+	}
+	unixConn, err := net.DialUnix("unix", nil, unixAddr)
 	if err != nil {
 		return nil, err
 	}
@@ -27,7 +31,7 @@ func PassFDToUnix(unixSock string) (*os.File, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = fd.Put(unixConn.(*net.UnixConn), server)
+	err = fd.Put(unixConn, server)
 	if err != nil {
 		return nil, err
 	}


pkg/networks/usernet/client.go
```diff
@@ -140,7 +140,8 @@ func create(sock string, subnet net.IP, base string) *Client {
 	client := &http.Client{
 		Transport: &http.Transport{
 			DialContext: func(context.Context, string, string) (net.Conn, error) {
-				return net.Dial("unix", sock)
+				network := "unix"
+				return net.DialUnix(network, nil, &net.UnixAddr{Name: sock, Net: network})
```
nirs, line 144: This is not an improvement and we don't do any type assertions here.

pkg/vz/network_darwin.go
```diff
@@ -27,7 +28,7 @@ func PassFDToUnix(unixSock string) (*os.File, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = fd.Put(unixConn.(*net.UnixConn), server)
+	err = fd.Put(unixConn, server)
```
nirs, line 31: Here we get rid of the type assertion, so making the code little worse has somme benefit.

pkg/vz/network_darwin.go
```diff
@@ -37,7 +38,8 @@ func PassFDToUnix(unixSock string) (*os.File, error) {
 // DialQemu support connecting to QEMU supported network stack via unix socket.
 // Returns os.File, connected dgram connection to be used for vz.
 func DialQemu(unixSock string) (*os.File, error) {
-	unixConn, err := net.Dial("unix", unixSock)
+	network := "unix"
+	unixConn, err := net.DialUnix(network, nil, &net.UnixAddr{Name: unixSock, Net: network})
```
nirs, line 42: Here again we don't get anything for making the code worse.

What if we add our own dialUnix() returning a *UnixCon, and hiding the ugly interface that standard library provides? We can use this helper in all 3 cases.

```go
func dialUnix(path string) (*UnixCon, error) {
    return net.DialUnix("unix", nil, &net.UnixAddr{Name: path, Net: "unix"})
}
```

pkg/vz/network_darwin.go
```diff
@@ -27,7 +28,7 @@ func PassFDToUnix(unixSock string) (*os.File, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = fd.Put(unixConn.(*net.UnixConn), server)
```
AkihiroSuda, line 30: I don't think we have to introduce a new `netutil` package just for avoiding this single type assertion.

AkihiroSuda (changes requested):
I don't think we have to introduce a new `netutil` package just for avoiding this single type assertion.

If you want to avoid a potential panic, you can just check `conn, ok := unixConn.(*net.UnixConn); if !ok { return errors.New(...)}`

tamird:
> I don't think we have to introduce a new `netutil` package just for avoiding this single type assertion.

Because this pattern occurs in a number of places (connecting to a UNIX socket, not the type assertion), this is reducing all that duplication.

> If you want to avoid a potential panic, you can just check `conn, ok := unixConn.(*net.UnixConn); if !ok { return errors.New(...)}`

I don't think a panic is a real concern. This is just cleanup.

nirs:
> @nirs this is what you wanted, no?

I suggested to add a helper which would solve the issue for the vz package. The current change is little bit bigger and handle all unix sockets in lima (good), but it is hard to sell it as a way to avoid type assertions since we have only one.

I did have time to review all the changes, but I think this will be easier to swallow if we change the commit/pr title/message to describe why creating unix socket in this way is better.

tamird:
Updated both the PR title and the commit message.

tamird:
Would appreciate a review here!

AkihiroSuda:
The amount of the deduplicated code is trivial.
I'm not sure this is worth adding a new "util" pkg.

pkg/httpclientutil/httpclientutil_others.go
```diff
@@ -17,9 +19,8 @@ func NewHTTPClientWithSocketPath(socketPath string) (*http.Client, error) {
 	}
 	hc := &http.Client{
 		Transport: &http.Transport{
-			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
-				var d net.Dialer
-				return d.DialContext(ctx, "unix", socketPath)
+			DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+				return netutil.DialUnix("unix", socketPath)
```
nirs, line 23: Here we ignore the context, which may ignore a deadline no? I would keep the code as is for now and focus on the places we use net.DialUnix(). Keeping the change small helps to review it faster is easy to merge faster.

pkg/httpclientutil/httpclientutil_windows.go
```diff
@@ -17,8 +19,7 @@ func NewHTTPClientWithSocketPath(socketPath string) (*http.Client, error) {
 	hc := &http.Client{
 		Transport: &http.Transport{
 			DialContext: func(ctx context.Context, _, _ string) (net.Conn, error) {
-				var d net.Dialer
-				return d.DialContext(ctx, "unix", socketPath)
+				return netutil.DialUnix("unix", socketPath)
```
nirs, line 22: Same here, we don't want drop the context. We can work on this later if you think this context is not really used.

pkg/networks/usernet/client.go
```diff
@@ -140,7 +141,7 @@ func create(sock string, subnet net.IP, base string) *Client {
 	client := &http.Client{
 		Transport: &http.Transport{
 			DialContext: func(context.Context, string, string) (net.Conn, error) {
-				return net.Dial("unix", sock)
+				return netutil.DialUnix("unix", sock)
```
nirs, line 144: In this case we already dropped the context. Maybe we should also drop it in other places, but it is not required to add a helper and can be done later.

pkg/qemu/qemu_driver.go
```diff
@@ -396,10 +397,8 @@ func (l *LimaQemuDriver) ListSnapshots(_ context.Context) (string, error) {
 	return List(qCfg, l.Instance.Status == store.StatusRunning)
 }
 
-func (l *LimaQemuDriver) GuestAgentConn(ctx context.Context) (net.Conn, error) {
-	var d net.Dialer
-	dialContext, err := d.DialContext(ctx, "unix", filepath.Join(l.Instance.Dir, filenames.GuestAgentSock))
-	return dialContext, err
+func (l *LimaQemuDriver) GuestAgentConn(context.Context) (net.Conn, error) {
+	return netutil.DialUnix("unix", filepath.Join(l.Instance.Dir, filenames.GuestAgentSock))
```
nirs, line 401: Here again we dropped the context. This change is mostly dropping the context from many places instead of adding a helper. Lets keep the code the same.

pkg/qemu/qemu_driver.go
```diff
@@ -417,11 +416,7 @@ func (a *qArgTemplateApplier) applyTemplate(qArg string) (string, error) {
 				if !ok {
 					return "", fmt.Errorf("non-string argument %+v", v)
 				}
-				addr := &net.UnixAddr{
-					Net:  "unix",
-					Name: s,
-				}
-				conn, err := net.DialUnix("unix", nil, addr)
+				conn, err := netutil.DialUnix("unix", s)
```
nirs, line 419: Nice cleanup - assuming that our helper creates the net.UnixAddr internally (but it does not).

pkg/netutil/netutil.go
```diff
@@ -0,0 +1,11 @@
+package netutil
+
+import "net"
+
+func DialUnix(network, address string) (*net.UnixConn, error) {
+	addr, err := net.ResolveUnixAddr(network, address)
```
nirs, line 6: Why do we need to resolved the address? Why not create a net.UnixAddr, doing basically what the code we want to replace did?

nirs (commented):
If we remove the unneeded and possibly unwanted change to ignore the context in many places, the actual change is very small, maybe only in vz/network_darwin.go?

But since the change is so small, I'm not sure it worth the effort. The benefit is very small. Better to invest time on stuff that add more value.

pkg/netutil/netutil.go
tamird: Because that's what is done elsewhere. This is called out in the commit message now.

pkg/qemu/qemu_driver.go
tamird: Sure. It is not exactly clear to me what cancellation actually does in the Unix socket case, but done.

AkihiroSuda (approved):
Thanks

PR #2923, opened 2024-11-19
Simplify QEMU write path

tamird:
Signed-off-by: Tamir Duberstein <tamird@gmail.com>

### DIFF
--- a/pkg/vz/network_darwin.go
+++ b/pkg/vz/network_darwin.go
@@ -109,13 +109,12 @@ func (c *qemuPacketConn) Write(b []byte) (int, error) {
 		return 0, err
 	}
 
-	start := 0
-	for start < size {
-		nw, err := c.Conn.Write(b[start:])
+	for len(b) != 0 {
+		n, err := c.Conn.Write(b)
 		if err != nil {
 			return 0, err
 		}
-		start += nw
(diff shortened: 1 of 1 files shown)


pkg/vz/network_darwin.go
```diff
@@ -116,13 +116,12 @@ func (c *qemuPacketConn) Write(b []byte) (int, error) {
...
-	start := 0
-	for start < size {
-		nw, err := c.Conn.Write(b[start:])
+	for len(b) != 0 {
+		n, err := c.Conn.Write(b)
 		if err != nil {
 			return 0, err
 		}
-		start += nw
+		b = b[n:]
```
nirs, line 124: The new code saves one line, but it makes the intent of the code less clear, and in case of errors it is harder to report how many bytes we copied. I would keep the code as is but lets see what others think.

tamird: This pattern is pretty common in the standard library: https://cs.opensource.google/search?q=for%5C%20len%5C(.*%5C)%5C%20(!%3D%7C%3E)%5C%200&sq=&ss=go%2Fgo

nirs (approved):
Lets go with the more idiomatic way.

PR #2933, opened 2024-11-21
Update go-qcow2reader to 0.6.0                                                                                                                                        

nirs:
This must be manual updated since this version made incompatible
changes.

With this version we cannot track conversion progress using the image,
since convert reads only the allocated extents of the image. We need to
pass the progress bar using convert.Options.

pb.ProgressBar need to be adapted to convert.Updater interface, so
progressbar returns now our own type. This can be used later for other
improvement like hiding the progress bar.


AkihiroSuda (approved):
Thanks

PR #2944, opened 2024-11-25
cleanup: refactor and use bicopy everywhere

tamird:
See commit message.

### DIFF
--- a/pkg/bicopy/bicopy.go
+++ b/pkg/bicopy/bicopy.go
@@ -1,22 +1,3 @@
-// From https://raw.githubusercontent.com/norouter/norouter/v0.6.5/pkg/agent/bicopy/bicopy.go
-/*
-   Copyright (C) NoRouter authors.
-
-   Copyright (C) libnetwork authors.
-
-   Licensed under the Apache License, Version 2.0 (the "License");
-   you may not use this file except in compliance with the License.
-   You may obtain a copy of the License at
-
-       http://www.apache.org/licenses/LICENSE-2.0
-
-   Unless required by applicable law or agreed to in writing, software
-   distributed under the License is distributed on an "AS IS" BASIS,
-   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-   See the License for the specific language governing permissions and
-   limitations under the License.
-*/
-
 package bicopy
 
 import (
@@ -26,56 +7,25 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-// Bicopy is from https://github.com/rootless-containers/rootlesskit/blob/v0.10.1/pkg/port/builtin/parent/tcp/tcp.go#L73-L104
-// (originally from libnetwork, Apache License 2.0).
-func Bicopy(x, y io.ReadWriter, quit <-chan struct{}) {
-	type closeReader interface {
-		CloseRead() error
-	}
-	type closeWriter interface {
-		CloseWrite() error
-	}
+type NamedReadWriter struct {
+	ReadWriter io.ReadWriter
+	Name       string
+}
+
+func Bicopy(context string, x, y NamedReadWriter) {
 	var wg sync.WaitGroup
-	broker := func(to, from io.ReadWriter) {
-		if _, err := io.Copy(to, from); err != nil {
-			logrus.WithError(err).Debug("failed to call io.Copy")
-		}
-		if fromCR, ok := from.(closeReader); ok {
-			if err := fromCR.CloseRead(); err != nil {
-				logrus.WithError(err).Debug("failed to call CloseRead")
-			}
-		}
-		if toCW, ok := to.(closeWriter); ok {
-			if err := toCW.CloseWrite(); err != nil {
-				logrus.WithError(err).Debug("failed to call CloseWrite")
-			}
-		}
-		wg.Done()
-	}
-
 	wg.Add(2)
-	go broker(x, y)
-	go broker(y, x)
-	finish := make(chan struct{})
 	go func() {
-		wg.Wait()
-		close(finish)
+		defer wg.Done()
+		if _, err := io.Copy(x.ReadWriter, y.ReadWriter); err != nil {
+			logrus.WithError(err).Errorf("%s: io.Copy(%s, %s)", context, x.Name, y.Name)
+		}
 	}()
-
-	select {
-	case <-quit:
-	case <-finish:
-	}
-	if xCloser, ok := x.(io.Closer); ok {
-		if err := xCloser.Close(); err != nil {
-			logrus.WithError(err).Debug("failed to call xCloser.Close")
+	go func() {
+		defer wg.Done()
+		if _, err := io.Copy(y.ReadWriter, x.ReadWriter); err != nil {
+			logrus.WithError(err).Errorf("%s: io.Copy(%s, %s)", context, y.Name, x.Name)
 		}
-	}
-	if yCloser, ok := y.(io.Closer); ok {
-		if err := yCloser.Close(); err != nil {
-			logrus.WithError(err).Debug("failed to call yCloser.Close")
-		}
-	}
-	<-finish
-	// TODO: return copied bytes
+	}()
+	wg.Wait()
 }
--- a/pkg/hostagent/port_darwin.go
+++ b/pkg/hostagent/port_darwin.go
@@ -134,23 +134,26 @@ func (plf *pseudoLoopbackForwarder) Serve() error {
 			ac.Close()
 			continue
 		}
-		go func(ac *net.TCPConn) {
-			if fErr := plf.forward(ac); fErr != nil {
-				logrus.Error(fErr)
-			}
-		}(ac)
+		go plf.forward(ac)
 	}
 }
 
-func (plf *pseudoLoopbackForwarder) forward(ac *net.TCPConn) error {
+func (plf *pseudoLoopbackForwarder) forward(ac *net.TCPConn) {
 	defer ac.Close()
 	unixConn, err := net.DialUnix("unix", nil, plf.unixAddr)
 	if err != nil {
-		return err
+		logrus.WithError(err).Errorf("pseudoloopback forwarder: failed to dial %q", plf.unixAddr)
+		return
 	}
 	defer unixConn.Close()
-	bicopy.Bicopy(ac, unixConn, nil)
-	return nil
+
+	bicopy.Bicopy("pseudoloopback forwarder", bicopy.NamedReadWriter{
+		ReadWriter: ac,
+		Name:       "tcp",
+	}, bicopy.NamedReadWriter{
+		ReadWriter: unixConn,
+		Name:       "unix",
+	})
 }
 
 func (plf *pseudoLoopbackForwarder) Close() error {
--- a/pkg/portfwd/client.go
+++ b/pkg/portfwd/client.go
@@ -29,7 +29,14 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
 	}
 
 	rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr, protocol: "tcp"}
-	bicopy.Bicopy(rw, conn, nil)
+
+	bicopy.Bicopy("tcp tunnel", bicopy.NamedReadWriter{
+		ReadWriter: rw,
+		Name:       guestAddr,
+	}, bicopy.NamedReadWriter{
+		ReadWriter: conn,
+		Name:       id,
+	})
 }
 
 func HandleUDPConnection(ctx context.Context, client *guestagentclient.GuestAgentClient, conn net.PacketConn, guestAddr string) {
--- a/pkg/portfwdserver/server.go
+++ b/pkg/portfwdserver/server.go
@@ -32,7 +32,14 @@ func (s *TunnelServer) Start(stream api.GuestService_TunnelServer) error {
 		return err
 	}
 	rw := &GRPCServerRW{stream: stream, id: in.Id}
-	bicopy.Bicopy(rw, conn, nil)
+
+	bicopy.Bicopy("tcp tunnel", bicopy.NamedReadWriter{
+		ReadWriter: rw,
+		Name:       in.Id,
+	}, bicopy.NamedReadWriter{
+		ReadWriter: conn,
+		Name:       in.GuestAddr,
+	})
 	return nil
 }
 
--- a/pkg/vz/network_darwin.go
+++ b/pkg/vz/network_darwin.go
@@ -8,11 +8,11 @@ import (
 	"io"
 	"net"
 	"os"
-	"sync"
 	"syscall"
 	"time"
 
 	"github.com/balajiv113/fd"
+	"github.com/lima-vm/lima/pkg/bicopy"
 
 	"github.com/sirupsen/logrus"
 )
@@ -66,24 +66,13 @@ func forwardPackets(qemuConn *qemuPacketConn, vzConn *packetConn) {
 	defer qemuConn.Close()
 	defer vzConn.Close()
 
-	var wg sync.WaitGroup
-	wg.Add(2)
-
-	go func() {
-		defer wg.Done()
-		if _, err := io.Copy(qemuConn, vzConn); err != nil {
-			logrus.Errorf("Failed to forward packets from VZ to VMNET: %s", err)
-		}
-	}()
-
-	go func() {
-		defer wg.Done()
-		if _, err := io.Copy(vzConn, qemuConn); err != nil {
-			logrus.Errorf("Failed to forward packets from VMNET to VZ: %s", err)
-		}
-	}()
-
-	wg.Wait()
+	bicopy.Bicopy("forwarding packets", bicopy.NamedReadWriter{
+		ReadWriter: qemuConn,
+		Name:       "VMNET",
+	}, bicopy.NamedReadWriter{
+		ReadWriter: vzConn,
+		Name:       "VZ",
+	})
 }
 
 // qemuPacketConn converts raw network packet to a QEMU supported network packet.


pkg/hostagent/port_darwin.go
```diff
@@ -149,8 +151,25 @@ func (plf *pseudoLoopbackForwarder) forward(ac *net.TCPConn) error {
...
 	defer unixConn.Close()
-	bicopy.Bicopy(ac, unixConn, nil)
-	return nil
+
+	g, _ := errgroup.WithContext(context.Background())
+
+	g.Go(func() error {
+		_, err := io.Copy(unixConn, ac)
+		if errors.Is(err, io.EOF) {
+			err = nil
```
nirs, line 160: The previous code does not handle io.EOF, so this implementation may behave differently.

pkg/hostagent/port_darwin.go
```diff
@@ -149,8 +151,25 @@ func (plf *pseudoLoopbackForwarder) forward(ac *net.TCPConn) error {
...
+		}
+		return errors.Join(err, unixConn.CloseRead(), ac.CloseWrite())
```
nirs, line 162: Previous code logged errors when calling CloseRead and CloseWrite, but did not fail. This will return an error in this case.

pkg/hostagent/port_darwin.go
```diff
@@ -149,8 +151,25 @@ func (plf *pseudoLoopbackForwarder) forward(ac *net.TCPConn) error {
...
+	})
+	g.Go(func() error {
+		_, err := io.Copy(ac, unixConn)
+		if errors.Is(err, io.EOF) {
+			err = nil
+		}
+		return errors.Join(err, ac.CloseRead(), unixConn.CloseWrite())
+	})
+
+	return g.Wait()
```
nirs, line 172: This will return an error if CloseWrite or CloseRead failed, changing the behavior.

pkg/hostagent/port_darwin.go
```diff
@@ -149,8 +151,25 @@ func (plf *pseudoLoopbackForwarder) forward(ac *net.TCPConn) error {
 		return err
 	}
 	defer unixConn.Close()
-	bicopy.Bicopy(ac, unixConn, nil)
-	return nil
```
nirs, line 153: Previous code always succeed, but new code may return an error.

nirs (commented):
This changes the behavior but you did not provide any reason for the change. If the purpose is to remove eliminate a package, just inline the package code in the single user. There is no need to change the code.

If you want to reimplement it and/or change the behavior, please explain why.

Also we have good tests for port_darwin? Before rewriting we should have good tests ensuring that we don't change the behavior by mistake. It should be simple test coping data over unix sockets.

pkg/hostagent/port_darwin.go
tamird: Indeed, this is matching the behavior here: https://github.com/lima-vm/lima/blob/ca778e338ab95524555ba4d23bd6398be5a6ee0f/pkg/portfwd/client.go#L30-L43

However that behavior appears to be nonsense given the documentation of [`io.Copy`](https://pkg.go.dev/io?utm_source=godoc#Copy):
> A successful Copy returns err == nil, not err == EOF. Because Copy is defined to read from src until EOF, it does not treat an EOF from Read as an error to be reported.

So I've removed that and mentioned in the commit message.

pkg/hostagent/port_darwin.go
tamird: This is the same comment in triplicate, so resolving this one.

pkg/hostagent/port_darwin.go
tamird: This is the same comment in triplicate, so resolving this one.

pkg/hostagent/port_darwin.go
nirs: Check how we test DialQemu.

pkg/hostagent/port_darwin.go
nirs: This is not the same comment. You suggest to return error instead of logging them. Please explain why this is better and how the code calling the modified code handles the errors.

It can be useful to show logs with errors before and after the change.

tamird: The caller of this function just logs the error. I have indicated this in the commit message.

pkg/bicopy/bicopy.go
```diff
@@ -1,81 +1,44 @@
...
 import (
+	"errors"
 	"io"
 	"sync"
-
-	"github.com/sirupsen/logrus"
 )
 
-// Bicopy is from https://github.com/rootless-containers/rootlesskit/blob/v0.10.1/pkg/port/builtin/parent/tcp/tcp.go#L73-L104
-// (originally from libnetwork, Apache License 2.0).
```
AkihiroSuda, line 30: This reference seems now lost

pkg/vz/network_darwin.go
```diff
@@ -62,24 +62,9 @@ func forwardPackets(qemuConn *qemuPacketConn, vzConn *packetConn) {
 	defer qemuConn.Close()
 	defer vzConn.Close()
 
-	var wg sync.WaitGroup
-	wg.Add(2)
-
-	go func() {
-		defer wg.Done()
-		if _, err := io.Copy(qemuConn, vzConn); err != nil {
-			logrus.Errorf("Failed to forward packets from VZ to VMNET: %s", err)
```
nirs, line 71: We need to keep this error message, so we get a specific error immediately when it happens.

pkg/vz/network_darwin.go
```diff
@@ -62,24 +62,9 @@ func forwardPackets(qemuConn *qemuPacketConn, vzConn *packetConn) {
...
-		}
-	}()
-
-	go func() {
-		defer wg.Done()
-		if _, err := io.Copy(vzConn, qemuConn); err != nil {
-			logrus.Errorf("Failed to forward packets from VMNET to VZ: %s", err)
```
nirs, line 78: Same here, so I don't think it is a good idea to share a trivial helper that run 2 io.Copy goroutines. It is better to use the same code we have here in other place with have more specific error messages.

tamird: Restored detailed error messages.

pkg/bicopy/bicopy.go
tamird: Yeah, intentionally so. There's nothing left from the old implementation apart from the name (and the name `broker`).

pkg/vz/network_darwin.go
```diff
@@ -66,24 +66,9 @@ func forwardPackets(qemuConn *qemuPacketConn, vzConn *packetConn) {
...
-		}
-	}()
-
-	wg.Wait()
+	if err := bicopy.Bicopy(qemuConn, vzConn, "VMNET", "VZ"); err != nil {
+		logrus.WithError(err).Error("failed to forward packets")
+	}
```
nirs, line 71: Previously we logged the error immediately when it happens. Now we my log no errors since after one copy fail, the second may be blocked forever. This will make it harder or impossible to debug issues.

I'm not sure that the current code is ensuring that if one copy fail the other one will also fail, and we don't have test for this error case. Before replacing the code lets understand if this is an issue and how we can fix it.

When we have 3 correct implementations we may consider unifying them.

pkg/portfwd/client.go
```diff
@@ -24,26 +25,10 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
 	}
 
-	g, _ := errgroup.WithContext(ctx)
-
 	rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr}
-	g.Go(func() error {
-		_, err := io.Copy(rw, conn)
-		if errors.Is(err, io.EOF) {
-			return nil
-		}
```
nirs, line 34: For this code, I would start with removing the handling of io.EOF if we think that it is not possible.

pkg/portfwd/client.go
```diff
@@ -24,26 +25,10 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
-	g.Go(func() error {
-		_, err := io.Copy(conn, rw)
-		if errors.Is(err, io.EOF) {
-			return nil
-		}
-		return err
-	})
 
-	if err := g.Wait(); err != nil {
-		logrus.Debugf("error in tcp tunnel for id: %s error:%v", id, err)
```
nirs, line 46: The previous code tried to log both errors, assuming that if one copy fail, the other will fail soon, but this may not be the case. This may never log any error if the first io.Copy() fails and the other is stuck on blocking socket.

A better way to handle this is to use the same code we have in vz/network_darwin.go, logging the first error immediately.

pkg/bicopy/bicopy.go
nirs, line 1: I did not review the previous bicopy.go code. What are the issues in the previous code that you are trying to fix?

nirs:
I think the common behavior in all 3 cases is that we start an endless io.Copy() in both directions. The copy is not expected to complete in normal conditions. This is the case for vz/network_darwin.go, I' not sure about the other 2, maybe port forwarding may end earlier in normal conditions.

In the vz network case, the copy should stop only when the vm is shutting down, or if socket_vmnet crashed or terminated. In this case we want to log an error or warning message about the peer closing the socket. Any other error should be logged as an error. In base case we want to log the issue immediately without a delay, so we don't lose errors.

When one of the io.Copy() goroutines fail, we want to other to fail immediately. I'm not sure this is ensured by current code. This should fixed and tested before we unify the implementation.

When both io.Copy() finished, we already logged the error, so there is nothing to return to the caller, even if we have a caller waiting for the result.

Usually saving the first error (synchronized with a lock or atomics) and returning it will be more helpful than returning 8 possible errors joined. After something fail, the next errors are usually just noise. If we can ensure that the second copy fail immediately when the first fail, we can return the first error to the caller and the caller can log it. This will be more useful for a library since different callers may want to use different loggers. May not be relevant to lima use case.

If we improve bicopy we want to share them with upstream. Did you open upstream issues for the problems you are trying to fix in bicopy?

pkg/bicopy/bicopy.go
tamird: Relative to the old code, the new code is simpler (no CloseRead and CloseWrite which aren't needed in the absence of cancellation) and is now used everywhere.

pkg/bicopy/bicopy.go
```diff
@@ -1,81 +1,71 @@
...
 
+func broker(dst, src io.ReadWriter, dstName, srcName string) error {
+	_, errCopy := io.Copy(dst, src)
+	if errCopy != nil {
+		errCopy = fmt.Errorf("io.Copy(%s, %s): %w", dstName, srcName, errCopy)
+	}
+	return errors.Join(
+		errCopy,
+		closeWrite(dst, dstName),
+		closeRead(src, srcName),
```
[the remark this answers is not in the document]
tamird: Removed.

pkg/bicopy/bicopy.go
```diff
@@ -1,81 +1,71 @@
...
+	var wg sync.WaitGroup
 	wg.Add(2)
-	go broker(x, y)
-	go broker(y, x)
-	finish := make(chan struct{})
+	var errLeft, errRight error
 	go func() {
-		wg.Wait()
-		close(finish)
+		errLeft = broker(left, right, leftName, rightName)
```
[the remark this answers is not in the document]
tamird: Removed.

pkg/bicopy/bicopy.go
```diff
@@ -1,81 +1,71 @@
...
+	go func() {
+		errRight = broker(right, left, rightName, leftName)
+	}()
+	wg.Wait()
+	return errors.Join(
+		errLeft,
+		errRight,
+		ioClose(left, leftName),
+		ioClose(right, rightName),
+	)
```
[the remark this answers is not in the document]
tamird: Sure. I moved all the logging back into bicopy.

pkg/portfwd/client.go
tamird: Eh. What's the benefit of doing that in a separate change? Given that it takes multiple days to get a review I don't see the upside. It is most definitely not possible, it's right there in the docs.

pkg/portfwd/client.go
tamird: Sure. I moved all the logging back into bicopy.

pkg/vz/network_darwin.go
tamird: Sure. I moved all the logging back into bicopy.

pkg/portfwd/client.go
nirs: If you post a trivial change like this it will be merged pretty quickly. People have limited time to review, and it is much easier to review tiny changes. Changes like this one, changing multiple components are hard to review.

tamird:
> If we improve bicopy we want to share them with upstream. Did you open upstream issues for the problems you are trying to fix in bicopy?

I'm just removing features we aren't using. If they need early cancellation that's just fine.

pkg/bicopy/bicopy.go
nirs: OK, so the reason for this change are:

- Maintaining one implementation instead of 3 (good)
- Removing cancellation support which we don't use. Good unless we think that we may need it in the future. I don't know if this may be needed for port forwarding or hostagent.
- Removing pointless io.EOF check (good)

Anything else?

tamird: Looks complete to me.

nirs (commented):
I like the simplified Bicopy!

nirs:
> > If we improve bicopy we want to share them with upstream. Did you open upstream issues for the problems you are trying to fix in bicopy?
> 
> I'm just removing features we aren't using. If they need early cancellation that's just fine.

Good reason why this is not relevant to upstream.

pkg/vz/network_darwin.go
```diff
@@ -66,24 +66,7 @@ func forwardPackets(qemuConn *qemuPacketConn, vzConn *packetConn) {
...
-
-	go func() {
-		defer wg.Done()
-		if _, err := io.Copy(vzConn, qemuConn); err != nil {
-			logrus.Errorf("Failed to forward packets from VMNET to VZ: %s", err)
-		}
-	}()
-
-	wg.Wait()
+	bicopy.Bicopy(qemuConn, vzConn, "VMNET", "VZ")
```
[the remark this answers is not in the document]
tamird: I golfed on this a bit and went with the `Named*` struct. PTAL.

pkg/bicopy/bicopy.go
```diff
@@ -26,56 +7,22 @@ import (
...
-			if err := toCW.CloseWrite(); err != nil {
-				logrus.WithError(err).Debug("failed to call CloseWrite")
-			}
-		}
-		wg.Done()
+func broker(dst, src io.ReadWriter, dstName, srcName string) {
+	if _, err := io.Copy(dst, src); err != nil {
+		logrus.WithError(err).Errorf("io.Copy(%s, %s)", dstName, srcName)
 	}
+}
```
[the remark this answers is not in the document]
nirs: We don't ensure this today, so it should not block your change. Current code assumes that if we fail to read or write from socket and io.Copy() returns, the other io.Copy() will also returns since it is trying to read of write using the same socket. This may be enough, but I'm not sure.

pkg/hostagent/port_darwin.go
```diff
@@ -134,23 +134,26 @@ func (plf *pseudoLoopbackForwarder) Serve() error {
...
+
+	bicopy.Bicopy(bicopy.NamedReadWriter{
+		ReadWriter: ac,
+		Name:       "tcp",
+	}, bicopy.NamedReadWriter{
+		ReadWriter: unixConn,
+		Name:       "unix",
+	})
```
nirs, line 156: Nice cleanup and we get better logs with this change.

pkg/bicopy/bicopy.go
```diff
@@ -26,56 +7,25 @@ import (
...
-	type closeReader interface {
-		CloseRead() error
-	}
-	type closeWriter interface {
-		CloseWrite() error
-	}
-	var wg sync.WaitGroup
-	broker := func(to, from io.ReadWriter) {
-		if _, err := io.Copy(to, from); err != nil {
-			logrus.WithError(err).Debug("failed to call io.Copy")
```
nirs, line 41: The new logs are much better.

pkg/bicopy/bicopy.go
```diff
@@ -26,56 +7,25 @@ import (
...
-	select {
-	case <-quit:
-	case <-finish:
-	}
-	if xCloser, ok := x.(io.Closer); ok {
-		if err := xCloser.Close(); err != nil {
-			logrus.WithError(err).Debug("failed to call xCloser.Close")
+		defer wg.Done()
+		if _, err := io.Copy(x.ReadWriter, y.ReadWriter); err != nil {
+			logrus.WithError(err).Errorf("io.Copy(%s, %s)", x.Name, y.Name)
```
nirs, line 21: How about:

   logrus.Errorf("Failed forward packets from %s to %s: %s", x.Name, y.Name, err)

Matching what we had in vz.network_darwin.go.

The helper is generic, but forwarding packets is the actual use case.

pkg/bicopy/bicopy.go
```diff
@@ -26,56 +7,25 @@ import (
...
-		}
-		if toCW, ok := to.(closeWriter); ok {
-			if err := toCW.CloseWrite(); err != nil {
-				logrus.WithError(err).Debug("failed to call CloseWrite")
-			}
-		}
-		wg.Done()
-	}
+type NamedReadWriter struct {
+	ReadWriter io.ReadWriter
```
nirs, line 11: Any reason to accept io.ReadWriter and not a net.Con? This is not a library for general use, but a lima helper specific to forwarding packets around.

pkg/bicopy/bicopy.go
tamird: We need a way to keep the helper generic. We could plumb in a context string to include in the error?

pkg/bicopy/bicopy.go
tamird: `GrpcClientRW` doesn't implement `net.Conn`.

pkg/portfwd/client.go
```diff
@@ -24,27 +25,15 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
-	}
+	bicopy.Bicopy(bicopy.NamedReadWriter{
+		ReadWriter: rw,
+		Name:       guestAddr,
+	}, bicopy.NamedReadWriter{
+		ReadWriter: conn,
+		Name:       id,
+	})
```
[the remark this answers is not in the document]
tamird: We could include the caller's file and line number in the error message?

pkg/bicopy/bicopy.go
nirs: So if we can ensure that both io.Copy() goroutines finish when one of them fail, we can return either the first error, or both errors to the caller, and the caller can log a more specific error.

nirs: To ensure the completion we need to close, so we probably want to accept a `io.ReadWriteCloser`.

tamird: `GrpcClientRW` doesn't implement `io.Closer`.

nirs: So how can we ensure coping from GrpcClientRW complete when one of the io.Copy() fail?

We can can use Close() if implemented using type assertion but this does not solve the issue for GrpcClientRW. If we want a common helper it should work for all use cases.

tamird: Ensuring that both `io.Copy`s return is outside the scope of this change. As mentioned in [another thread](https://github.com/lima-vm/lima/pull/2944#discussion_r1868234331) this guarantee doesn't exist today, so I would prefer to leave such enhancements to you in another change.

nirs: Looking at portfwd code and the relevant grpc code, I don't see a clear way to use this safely with Bicopy. The way to make progress with this change is leave portfwd as is for now, and use Bicopy in the other case.

We can open an issue fixing portfwd behavior on errors. When we find how to do this we can enhance Bicopy to handle this case.

nirs: The logs are worse, and passing an error message to Bicopy() does not seems like a good API.

pkg/portfwd/client.go
```diff
@@ -24,27 +25,15 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
 	}
 
-	g, _ := errgroup.WithContext(ctx)
-
 	rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr}
-	g.Go(func() error {
-		_, err := io.Copy(rw, conn)
-		if errors.Is(err, io.EOF) {
-			return nil
-		}
```
nirs, line 34: Looking at io.Copy implementation:

```
func copyBuffer(dst Writer, src Reader, buf []byte) (written int64, err error) {
	// If the reader has a WriteTo method, use it to do the copy.
	// Avoids an allocation and a copy.
	if wt, ok := src.(WriterTo); ok {
		return wt.WriteTo(dst)
	}
	// Similarly, if the writer has a ReadFrom method, use it to do the copy.
	if rf, ok := dst.(ReaderFrom); ok {
		return rf.ReadFrom(src)
	}
	if buf == nil {
		size := 32 * 1024
		if l, ok := src.(*LimitedReader); ok && int64(size) > l.N {
			if l.N < 1 {
				size = 1
			} else {
				size = int(l.N)
			}
		}
		buf = make([]byte, size)
	}
	for {
		nr, er := src.Read(buf)
		if nr > 0 {
			nw, ew := dst.Write(buf[0:nr])
			if nw < 0 || nr < nw {
				nw = 0
				if ew == nil {
					ew = errInvalidWrite
				}
			}
			written += int64(nw)
			if ew != nil {
				err = ew
				break
			}
			if nr != nw {
				err = ErrShortWrite
				break
			}
		}
		if er != nil {
			if er != EOF {
				err = er
			}
			break
		}
	}
	return written, err
}
```

It will not return io.EOF from the copy loop - but it may delegate the copy to ReadFrom() or WriteTo(), which can be implemented by the actual types we copy. I don't think we have any guarantee that we will never get io.EOF.

So it may be better to keep this if we don't want to report it as an error.

tamird: `ReadFrom` is not permitted to return EOF:
https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/io/io.go;l=186

`WriteTo` returns "Any error encountered during the write":
https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/io/io.go;l=197
EOF does not occur on write, only on read.

pkg/bicopy/bicopy.go
tamird: I have added a context string to preserve "forwarding packets" and similar.

pkg/portfwd/client.go
nirs: Good about ReadFrom - so we can remove the io.EOF checks than.

pkg/portfwd/client.go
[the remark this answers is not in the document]
nirs: The logger can be configured to include file:line info, but having a good error message is more important.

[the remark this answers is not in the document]
tamird: Configuring the logger in this way won't work because the file/line it would report would be in `bicopy.go` rather than in the caller.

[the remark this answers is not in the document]
nirs: This is fine, the error was in Bicopy and there is reason to lie and report the line of the caller.

If we want to see the file:line of the caller, log the error in the caller.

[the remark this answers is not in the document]
tamird: I believe we're done here, the context is now logged. The fidelity is as good as it was before this change. Resolving.

tamird:
Looks like I caught a test flake - I don't see bicopy in the stack trace.
```
TestGetPorts/loadBalancer_service
2024-12-03T23:19:21.6922190Z panic: test timed out after 10m0s
```

jandubois:
> Looks like I caught a test flake - I don't see bicopy in the stack trace.

I've triggered a re-run, but I'm waiting for approval from another reviewer before merging; not going to work my way through 72 comments... 😄

pkg/bicopy/bicopy.go
```diff
@@ -26,56 +7,25 @@ import (
...
+	go func() {
+		defer wg.Done()
+		if _, err := io.Copy(y.ReadWriter, x.ReadWriter); err != nil {
+			logrus.WithError(err).Errorf("%s: io.Copy(%s, %s)", context, y.Name, x.Name)
 		}
-	}
-	<-finish
-	// TODO: return copied bytes
+	}()
+	wg.Wait()
```
balajiv113, line 30: Do we really need bicopy to handle the error and not throw it to the caller ??

I would prefer it to be handled by the caller itself. It gives us better control. Handling error ourself looks generic but all we are doing is just logging. But in some cases we may need to act on error in caller side

tamird: There has been a lot of back and forth on this already. I'm inclined to leave it as is. Another option is to pass a handler callback, but nobody needs this today.

balajiv113: `but nobody needs this today` - If errors are returned back, callers will have more control to handle logging parts. Some cases might need more detailed logging rather than generic message `context: io.Copy(y, x)` which is not self explanatory.

If you have a valid reason for not returning error please let me know. 

Also i think `func Bicopy(context string, x, y NamedReadWriter) error {}` should be good as well. We don't need to complicate with handlers / callbacks

tamird: @nirs expressed concern that if `io.Copy` in one direction returns, the `io.Copy` in the other direction might not - and returning an error cannot happen until both functions complete. That's why the latest versions logs directly instead of returning an error.

balajiv113: It will eventually fail because one end of stream is closed already. 

Since we have wg.Done at last. Returning error after both io.Copy are closed looks correct to me.

balajiv113: @nirs What do you think ??

balajiv113: We could even use errgroup. Which will close the group when a non-nil error is returned. 

Code can look something like

```
wg, ctx := errgroup.WithContext(ctx)
defer b.Close()
defer a.Close()

wg.Go(func() error {
         _, err1 = io.Copy(a, b)
	return err1
})
wg.Go(func() error {
	_, err1 = io.Copy(b, a)
	return err1
})
return wg.Wait()
```

Even in case close() is not there, we will not do close but all other stays the same

balajiv113: With errgroup as soon as one is returned non-nil its over. It will close the other as well.

If we are using reader/writer then we will not even have close so it's just stopping the loop

nirs: The goroutine is canceled magically? how does it stop?

nirs: I don't see anything in the docs about cancelling goroutine blocked on a syscall.
https://pkg.go.dev/golang.org/x/sync/errgroup#WithContext

balajiv113: Yes you are right, My mistake.

Only context is closed. So we need to check for context close and close the connection.

For pure reader / writer until that syscall is returned we have no other option. But most of our reader / writer have close method so io.Closable should be fine

nirs: I think this should also be added to tcpproxy upstream - ensure termination and return an error. With both we can use it instead of Bicopy.

pkg/bicopy/bicopy.go
```diff
@@ -26,56 +7,25 @@ import (
...
-	select {
-	case <-quit:
-	case <-finish:
-	}
-	if xCloser, ok := x.(io.Closer); ok {
-		if err := xCloser.Close(); err != nil {
-			logrus.WithError(err).Debug("failed to call xCloser.Close")
+		defer wg.Done()
+		if _, err := io.Copy(x.ReadWriter, y.ReadWriter); err != nil {
+			logrus.WithError(err).Errorf("%s: io.Copy(%s, %s)", context, x.Name, y.Name)
```
nirs, line 21: This logs an error like:

    "forwarding packets: io.Copy(VMNENT, VZ)" ... {"error": "actual error from io.Copy"}

For code already calling bicopy, this is an improvement, since before we logged a DEBUG message without any information on which direction failed, or what is the purpose of the copy:

    "failed to call io.Copy" ... ... {"error": "actual error from io.Copy"}

But for the call from network_darwin this is worse. Before we had this clear error:

    "Failed to forward packets from VZ to VMNET: actual error from io.Copy"

I would like to keep the previous error message or something very close. I don't see the value of logging the error separately using logrous.WithError().

It can be done using:

```go
logrus.Errorf("Failed to %s from %s to %s: %s", context, y.Name, x.Name, err)
```

pkg/hostagent/port_darwin.go
```diff
@@ -134,23 +134,26 @@ func (plf *pseudoLoopbackForwarder) Serve() error {
...
 	}
 }
 
-func (plf *pseudoLoopbackForwarder) forward(ac *net.TCPConn) error {
+func (plf *pseudoLoopbackForwarder) forward(ac *net.TCPConn) {
 	defer ac.Close()
 	unixConn, err := net.DialUnix("unix", nil, plf.unixAddr)
 	if err != nil {
-		return err
+		logrus.WithError(err).Errorf("pseudoloopback forwarder: failed to dial %q", plf.unixAddr)
```
nirs, line 145: This error adds context (good) but logs the error separately (bad). Before we logged only the error from net.DialUnix (line 139).

pkg/hostagent/port_darwin.go
```diff
@@ -134,23 +134,26 @@ func (plf *pseudoLoopbackForwarder) Serve() error {
...
-	bicopy.Bicopy(ac, unixConn, nil)
-	return nil
+
+	bicopy.Bicopy("pseudoloopback forwarder", bicopy.NamedReadWriter{
+		ReadWriter: ac,
+		Name:       "tcp",
+	}, bicopy.NamedReadWriter{
+		ReadWriter: unixConn,
+		Name:       "unix",
+	})
```
nirs, line 156: The context does not match other context ("pseudoloopback fowarder" vs "forward packets") and the names of the readwriters are little bit too generic, I'm not sure what is the context on this code, may be others can suggest better values.

pkg/portfwd/client.go
```diff
@@ -29,7 +29,14 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
 	}
 
 	rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr, protocol: "tcp"}
-	bicopy.Bicopy(rw, conn, nil)
+
+	bicopy.Bicopy("tcp tunnel", bicopy.NamedReadWriter{
```
nirs, line 33: This is the tunnel for tcp port forwarding, so maybe a better context would be "forward tcp packets for port N". We probably can get the port from conn.RemoteAddr(), since we forward port N to port N on the host.

pkg/portfwd/client.go
```diff
@@ -29,7 +29,14 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
+		ReadWriter: rw,
+		Name:       guestAddr,
+	}, bicopy.NamedReadWriter{
+		ReadWriter: conn,
+		Name:       id,
```
nirs, line 38: id is a description of the entire port forwading action: "tcp-%s-%s" (see line 17), so it is not a good for the conn.

pkg/portfwdserver/server.go
```diff
@@ -32,7 +32,14 @@ func (s *TunnelServer) Start(stream api.GuestService_TunnelServer) error {
 		return err
 	}
 	rw := &GRPCServerRW{stream: stream, id: in.Id}
-	bicopy.Bicopy(rw, conn, nil)
+
+	bicopy.Bicopy("tcp tunnel", bicopy.NamedReadWriter{
+		ReadWriter: rw,
+		Name:       in.Id,
```
nirs, line 38: Not sure what Id is. Can you share the output we get in runtime?

pkg/portfwdserver/server.go
```diff
@@ -32,7 +32,14 @@ func (s *TunnelServer) Start(stream api.GuestService_TunnelServer) error {
 		return err
 	}
 	rw := &GRPCServerRW{stream: stream, id: in.Id}
-	bicopy.Bicopy(rw, conn, nil)
+
+	bicopy.Bicopy("tcp tunnel", bicopy.NamedReadWriter{
```
nirs, line 36: We can probably use the same context used in the client with the guest port.

tamird:
Since I don't remember why I care about this and it's not fixing obvious bugs, I'm closing. If you want to reuse this work, you have my permission to do so without attribution.

PR #2954, opened 2024-12-02
website: add a guide for running Lima on GHA

AkihiroSuda:
Lima is also useful for:
- Running commands on non-Ubuntu operating systems such as Fedora
- Emulating multiple hosts

Test run: https://github.com/lima-vm/lima/actions/runs/12113943143/job/33769697797?pr=2955

### DIFF
--- a/website/content/en/docs/examples/gha.md
+++ b/website/content/en/docs/examples/gha.md
@@ -1,0 +1,92 @@
+---
+title: GitHub Actions
+weight: 10
+---
+
+## Running Lima on GitHub Actions
+
+On GitHub Actions, Lima is useful for:
+- Running commands on non-Ubuntu operating systems (e.g., Fedora for testing SELinux)
+- Emulating multiple hosts
+
+While these tasks can be partially accomplished with containers like Docker, those containers still rely on the Ubuntu host's kernel and cannot utilize features missing in Ubuntu, such as SELinux.
+
+In contrast, Lima runs virtual machines that do not depend on the Ubuntu host's kernel.
+
+The following GitHub Actions workflow illustrates how to run multiple instances of Fedora using Lima.
+The instances are connected by the `user-v2` network.
+
+```yaml
+name: Fedora
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+jobs:
+  fedora:
+    runs-on: ubuntu-24.04
+    steps:
+    - name: Check out code
+      uses: actions/checkout@v4
+
+    - name: "Install QEMU"
+      run: |
+        set -eux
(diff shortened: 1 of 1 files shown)


website/content/en/docs/examples/gha.md
```diff
@@ -0,0 +1,90 @@
...
+---
+
+## Running Lima on GitHub Actions
+
+Lima is also useful for:
+- Running commands on non-Ubuntu operating systems (e.g., Fedora for testing SELinux)
+- Emulating multiple hosts.
+
+These tasks can be also partially accomplished with containers like Docker,
+but they are still stick to Ubuntu's kernel configuration and cannot use non-Ubuntu features such as SELinux.
```
jandubois, line 13: I don't know what this means: "they are still stick to Ubuntu's kernel configuration"

jandubois (changes requested):
Looks good; just one place that seems unclear to me.

I think it would be even better to create our own action for this, so users could write:

```yaml
   - name: Run Fedora VM
     uses: lima-vm/lima-acation
     with:
       ...
```

But that can come later.

website/content/en/docs/examples/gha.md
AkihiroSuda: Reworded

```
These tasks can be also partially accomplished with containers like Docker.
However, those containers still have to run on the Ubuntu host's kernel, and cannot use features missing in Ubuntu, such as SELinux.

On the other hand, Lima executes virtual machines that does not use the Ubuntu host's kernel.
```

AkihiroSuda:
> uses: lima-vm/lima-action

SGTM, but some organizations may prefer to just copy a few lines of yaml snippet rather than to introduce a dependency on a third-party action.
Probably we will have to keep maintaining the both action and the yaml snippet.

website/content/en/docs/examples/gha.md
AkihiroSuda: Asked Copilot to improve it further
```
Lima is also useful for:
- Running commands on non-Ubuntu operating systems (e.g., Fedora for testing SELinux)
- Emulating multiple hosts

While these tasks can be partially accomplished with containers like Docker, those containers still rely on the Ubuntu host's kernel and cannot utilize features missing in Ubuntu, such as SELinux.

In contrast, Lima runs virtual machines that do not depend on the Ubuntu host's kernel.
```

jandubois:
> but some organizations may prefer to just copy a few lines of yaml snippet rather than to introduce a dependency on a third-party action.

I m familiar with all third-party actions needing approval at the GitHub org level...

jandubois (approved):

PR #2974, opened 2024-12-04
CI and docs: update socket_vmnet (1.2.0)

AkihiroSuda:
https://github.com/lima-vm/socket_vmnet/compare/v1.1.5...v1.2.0

### DIFF
--- a/website/content/en/docs/config/network.md
+++ b/website/content/en/docs/config/network.md
@@ -148,8 +148,8 @@ writable by `root` and will reject `socket_vmnet` installed by Homebrew into a u
 # This assumes that Xcode Command Line Tools are already installed
 git clone https://github.com/lima-vm/socket_vmnet
 cd socket_vmnet
-# Change "v1.1.5" to the actual latest release in https://github.com/lima-vm/socket_vmnet/releases
-git checkout v1.1.5
+# Change "v1.2.0" to the actual latest release in https://github.com/lima-vm/socket_vmnet/releases
+git checkout v1.2.0
 make
 sudo make PREFIX=/opt/socket_vmnet install.bin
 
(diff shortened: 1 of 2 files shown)


website/content/en/docs/config/network.md
```diff
@@ -148,8 +148,8 @@ writable by `root` and will reject `socket_vmnet` installed by Homebrew into a u
 # This assumes that Xcode Command Line Tools are already installed
 git clone https://github.com/lima-vm/socket_vmnet
 cd socket_vmnet
-# Change "v1.1.5" to the actual latest release in https://github.com/lima-vm/socket_vmnet/releases
-git checkout v1.1.5
+# Change "v1.2.0" to the actual latest release in https://github.com/lima-vm/socket_vmnet/releases
+git checkout v1.2.0
```
nirs, line 152: Change looks fine, but since we don't build latest version from git, why not use the latest release tarball?

nirs (approved):

website/content/en/docs/config/network.md
AkihiroSuda: I don't get. The binary releases are built from the git repo.

nirs: Sure the result is the same, but we don't depend on being able to build socket_vment. We use the same method used by users to install it, which is much simpler. Get the latest release from github, download the tarball, and extract it. And with this change we can remove this configuration and avoid this another detail we have to maintain forever.

nirs: I don't want to block this simple change, just a thought on the next step.

jandubois (approved):
Thanks, LGTM

website/content/en/docs/config/network.md
jandubois: > I don't want to block this simple change, just a thought on the next step.

Ok, I've approved and merged then.

Please file a new issue/PR for using the upstream release instead. I do see the (minor) point that this validates that the releases can be installed and used. Also means that we don't need knowledge about how to build `socket_vmnet` in the `lima` repo. But then that is pretty much abstracted away in the `Makefile` anyways.

PR #2971, opened 2024-12-03
portfwd: Fix handling the last UDP  message

nirs:
PacketCon.ReadFrom documentation says[1]:

> Callers should always process the n > 0 bytes returned before
> considering the error err.

But we handled err first, and dropped the last read bytes. Fixed by sending a message if n > 0, and handling the error after the send.

I'm not sure what how this bug affects the system. It was found while reviewing #2969.

stream.Recv() does is not documented, but the trivial implementation ensure that we get nil message on any error. Add comment to make it more clear.

[1] https://pkg.go.dev/net#PacketConn

Fixes #2970

Thanks: Tamir Duberstein <tamird@gmail.com>

PR #2985, opened 2024-12-06
Revamp GRPC Port forwarding tunnels to use existing proxy

### LINKED ISSUE #2968: GRPC port forwarding hangs when using mysql connection
### Description

**Steps to reproduce**
- Run `limactl start --name=lima-vz --vm-type=vz --mount-type=virtiofs template://docker`
- Run `docker run --detach --name mariadb -p 127.0.0.1:10099:3606 --env MARIADB_DATABASE=db --env MARIADB_ROOT_PASSWORD=root mariadb:latest --port 3606`
- Run `mysql -uroot -proot --database=db --host=127.0.0.1 --port=10099 -e "SHOW TABLES;" ` from host and notice it hangs
- Same running from guest works

Note: Identified this issue by running ddev tests

balajiv113:
Fixes #2968 and several GRPC port forwarding issues. 

FYI - Running ddev testsuite to verify end-end (Done)
https://buildkite.com/test-305/ddev/builds/19#01939bef-af2a-4b7d-9ea8-fa1ca461f037

### DIFF
--- a/pkg/portfwd/client.go
+++ b/pkg/portfwd/client.go
@@ -2,20 +2,18 @@ package portfwd
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"io"
 	"net"
+	"time"
 
+	"github.com/containers/gvisor-tap-vsock/pkg/services/forwarder"
+	"github.com/lima-vm/lima/pkg/bicopy"
 	"github.com/lima-vm/lima/pkg/guestagent/api"
 	guestagentclient "github.com/lima-vm/lima/pkg/guestagent/api/client"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/sync/errgroup"
 )
 
 func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgentClient, conn net.Conn, guestAddr string) {
-	defer conn.Close()
-
 	id := fmt.Sprintf("tcp-%s-%s", conn.LocalAddr().String(), conn.RemoteAddr().String())
 
 	stream, err := client.Tunnel(ctx)
@@ -24,26 +22,17 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
 		return
 	}
 
-	g, _ := errgroup.WithContext(ctx)
-
-	rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr}
-	g.Go(func() error {
-		_, err := io.Copy(rw, conn)
-		return err
-	})
-	g.Go(func() error {
-		_, err := io.Copy(conn, rw)
-		return err
-	})
-
-	if err := g.Wait(); err != nil {
-		logrus.Debugf("error in tcp tunnel for id: %s error:%v", id, err)
+	// Handshake message to start tunnel
+	if err := stream.Send(&api.TunnelMessage{Id: id, Protocol: "tcp", GuestAddr: guestAddr}); err != nil {
+		logrus.Errorf("could not start tcp tunnel for id: %s error:%v", id, err)
+		return
 	}
+
+	rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr, protocol: "tcp"}
+	bicopy.Bicopy(rw, conn, nil)
 }
 
 func HandleUDPConnection(ctx context.Context, client *guestagentclient.GuestAgentClient, conn net.PacketConn, guestAddr string) {
-	defer conn.Close()
-
 	id := fmt.Sprintf("udp-%s", conn.LocalAddr().String())
 
 	stream, err := client.Tunnel(ctx)
@@ -52,79 +41,46 @@ func HandleUDPConnection(ctx context.Context, client *guestagentclient.GuestAgen
 		return
 	}
 
-	g, _ := errgroup.WithContext(ctx)
-
-	g.Go(func() error {
-		buf := make([]byte, 65507)
-		for {
-			n, addr, err := conn.ReadFrom(buf)
-			// We must handle n > 0 bytes before considering the error.
-			// https://pkg.go.dev/net#PacketConn
-			if n > 0 {
-				msg := &api.TunnelMessage{
-					Id:            id + "-" + addr.String(),
-					Protocol:      "udp",
-					GuestAddr:     guestAddr,
-					Data:          buf[:n],
-					UdpTargetAddr: addr.String(),
-				}
-				if err := stream.Send(msg); err != nil {
-					return err
-				}
-			}
-			if err != nil {
-				// https://pkg.go.dev/net#PacketConn does not mention io.EOF semantics.
-				if errors.Is(err, io.EOF) {
-					return nil
-				}
-				return err
-			}
-		}
-	})
-
-	g.Go(func() error {
-		for {
-			// Not documented: when err != nil, in is always nil.
-			in, err := stream.Recv()
-			if err != nil {
-				if errors.Is(err, io.EOF) {
-					return nil
-				}
-				return err
-			}
-			addr, err := net.ResolveUDPAddr("udp", in.UdpTargetAddr)
-			if err != nil {
-				return err
-			}
-			_, err = conn.WriteTo(in.Data, addr)
-			if err != nil {
-				return err
-			}
-		}
-	})
-
-	if err := g.Wait(); err != nil {
-		logrus.Debugf("error in udp tunnel for id: %s error:%v", id, err)
+	// Handshake message to start tunnel
+	if err := stream.Send(&api.TunnelMessage{Id: id, Protocol: "udp", GuestAddr: guestAddr}); err != nil {
+		logrus.Errorf("could not start udp tunnel for id: %s error:%v", id, err)
+		return
 	}
+
+	proxy, err := forwarder.NewUDPProxy(conn, func() (net.Conn, error) {
+		rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr, protocol: "udp"}
+		return rw, nil
+	})
+	if err != nil {
+		logrus.Errorf("error in udp tunnel proxy for id: %s error:%v", id, err)
+		return
+	}
+
+	defer func() {
+		err := proxy.Close()
+		if err != nil {
+			logrus.Errorf("error in closing udp tunnel proxy for id: %s error:%v", id, err)
+		}
+	}()
+	proxy.Run()
 }
 
 type GrpcClientRW struct {
-	id     string
-	addr   string
-	stream api.GuestService_TunnelClient
+	id   string
+	addr string
+
+	protocol string
+	stream   api.GuestService_TunnelClient
 }
 
-var _ io.ReadWriter = (*GrpcClientRW)(nil)
+var _ net.Conn = (*GrpcClientRW)(nil)
 
-func (g GrpcClientRW) Write(p []byte) (n int, err error) {
-	if len(p) == 0 {
-		return 0, nil
-	}
+func (g *GrpcClientRW) Write(p []byte) (n int, err error) {
 	err = g.stream.Send(&api.TunnelMessage{
 		Id:        g.id,
 		GuestAddr: g.addr,
 		Data:      p,
-		Protocol:  "tcp",
+		Protocol:  g.protocol,
 	})
 	if err != nil {
 		return 0, err
@@ -132,18 +88,35 @@ func (g GrpcClientRW) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
-func (g GrpcClientRW) Read(p []byte) (n int, err error) {
-	// Not documented: when err != nil, in is always nil.
+func (g *GrpcClientRW) Read(p []byte) (n int, err error) {
 	in, err := g.stream.Recv()
 	if err != nil {
-		if errors.Is(err, io.EOF) {
-			return 0, nil
-		}
 		return 0, err
-	}
-	if len(in.Data) == 0 {
-		return 0, nil
 	}
 	copy(p, in.Data)
 	return len(in.Data), nil
+}
+
+func (g *GrpcClientRW) Close() error {
+	return g.stream.CloseSend()
+}
+
+func (g *GrpcClientRW) LocalAddr() net.Addr {
+	return &net.UnixAddr{Name: "grpc", Net: "unixpacket"}
+}
+
+func (g *GrpcClientRW) RemoteAddr() net.Addr {
+	return &net.UnixAddr{Name: "grpc", Net: "unixpacket"}
+}
+
+func (g *GrpcClientRW) SetDeadline(_ time.Time) error {
+	return nil
+}
+
+func (g *GrpcClientRW) SetReadDeadline(_ time.Time) error {
+	return nil
+}
+
+func (g *GrpcClientRW) SetWriteDeadline(_ time.Time) error {
+	return nil
 }
--- a/pkg/portfwdserver/server.go
+++ b/pkg/portfwdserver/server.go
@@ -4,66 +4,79 @@ import (
 	"errors"
 	"io"
 	"net"
+	"time"
 
+	"github.com/lima-vm/lima/pkg/bicopy"
 	"github.com/lima-vm/lima/pkg/guestagent/api"
 )
 
-type TunnelServer struct {
-	Conns map[string]net.Conn
-}
+type TunnelServer struct{}
 
 func NewTunnelServer() *TunnelServer {
-	return &TunnelServer{
-		Conns: make(map[string]net.Conn),
-	}
+	return &TunnelServer{}
 }
 
 func (s *TunnelServer) Start(stream api.GuestService_TunnelServer) error {
-	for {
-		in, err := stream.Recv()
+	// Receive the handshake message to start tunnel
+	in, err := stream.Recv()
+	if err != nil {
 		if errors.Is(err, io.EOF) {
 			return nil
 		}
-		if err != nil {
-			return err
-		}
-		if len(in.Data) == 0 {
-			continue
-		}
-
-		conn, ok := s.Conns[in.Id]
-		if !ok {
-			conn, err = net.Dial(in.Protocol, in.GuestAddr)
-			if err != nil {
-				return err
-			}
-			s.Conns[in.Id] = conn
-
-			writer := &GRPCServerWriter{id: in.Id, udpAddr: in.UdpTargetAddr, stream: stream}
-			go func() {
-				_, _ = io.Copy(writer, conn)
-				delete(s.Conns, writer.id)
-			}()
-		}
-		_, err = conn.Write(in.Data)
-		if err != nil {
-			return err
-		}
+		return err
 	}
+
+	// We simply forward data form GRPC stream to net.Conn for both tcp and udp. So simple proxy is sufficient
+	conn, err := net.Dial(in.Protocol, in.GuestAddr)
+	if err != nil {
+		return err
+	}
+	rw := &GRPCServerRW{stream: stream, id: in.Id}
+	bicopy.Bicopy(rw, conn, nil)
+	return nil
 }
 
-type GRPCServerWriter struct {
-	id      string
-	udpAddr string
-	stream  api.GuestService_TunnelServer
+type GRPCServerRW struct {
+	id     string
+	stream api.GuestService_TunnelServer
 }
 
-var _ io.Writer = (*GRPCServerWriter)(nil)
+var _ net.Conn = (*GRPCServerRW)(nil)
 
-func (g GRPCServerWriter) Write(p []byte) (n int, err error) {
-	if len(p) == 0 {
-		return 0, nil
-	}
-	err = g.stream.Send(&api.TunnelMessage{Id: g.id, Data: p, UdpTargetAddr: g.udpAddr})
+func (g *GRPCServerRW) Write(p []byte) (n int, err error) {
+	err = g.stream.Send(&api.TunnelMessage{Id: g.id, Data: p})
 	return len(p), err
+}
+
+func (g *GRPCServerRW) Read(p []byte) (n int, err error) {
+	in, err := g.stream.Recv()
+	if err != nil {
+		return 0, err
+	}
+	copy(p, in.Data)
+	return len(in.Data), nil
+}
+
+func (g *GRPCServerRW) Close() error {
+	return nil
+}
+
+func (g *GRPCServerRW) LocalAddr() net.Addr {
+	return &net.UnixAddr{Name: "grpc", Net: "unixpacket"}
+}
+
+func (g *GRPCServerRW) RemoteAddr() net.Addr {
+	return &net.UnixAddr{Name: "grpc", Net: "unixpacket"}
+}
+
+func (g *GRPCServerRW) SetDeadline(_ time.Time) error {
+	return nil
+}
+
+func (g *GRPCServerRW) SetReadDeadline(_ time.Time) error {
+	return nil
+}
+
+func (g *GRPCServerRW) SetWriteDeadline(_ time.Time) error {
+	return nil
 }


pkg/portfwdserver/server.go
```diff
@@ -1,69 +1,85 @@
...
-	if len(p) == 0 {
-		return 0, nil
-	}
-	err = g.stream.Send(&api.TunnelMessage{Id: g.id, Data: p, UdpTargetAddr: g.udpAddr})
+func (g GRPCServerRW) Write(p []byte) (n int, err error) {
+	err = g.stream.Send(&api.TunnelMessage{Id: g.id, Data: p})
 	return len(p), err
 }
+
+func (g GRPCServerRW) Read(p []byte) (n int, err error) {
```
nirs, line 54: Should we add a grpc module providing the server and client so we have the core of this code in one place? This will tiny dead code in the guest and host agent that may not be optimized out, but it will make it easier to understand and maintain when all the pieces are the same place.

pkg/portfwdserver/server.go
```diff
@@ -1,69 +1,85 @@
...
+	copy(p, in.Data)
+	return len(in.Data), nil
+}
+
+func (g GRPCServerRW) Close() error {
+	return nil
+}
+
+func (g GRPCServerRW) LocalAddr() net.Addr {
+	return &net.UnixAddr{Name: "grpc", Net: "unixpacket"}
```
nirs, line 68: This seems like the right way to support good logging, better than the NamedReadWriter added in #2944.

pkg/portfwdserver/server.go
```diff
@@ -1,69 +1,85 @@
...
+	in, err := g.stream.Recv()
+	if err != nil {
+		return 0, err
+	}
+	copy(p, in.Data)
+	return len(in.Data), nil
+}
+
+func (g GRPCServerRW) Close() error {
+	return nil
```
nirs, line 64: Why this does nothing?

pkg/portfwdserver/server.go
```diff
@@ -1,69 +1,85 @@
...
-			return err
-		}
+	//Receive the handshake message to start tunnel
+	in, err := stream.Recv()
+	if errors.Is(err, io.EOF) {
+		return nil
+	}
+	if err != nil {
+		return err
 	}
```
nirs, line 29: In portfwd we changed these checks to this form:

```
if err != nil {
    if errors.Is(err, io.EOF) {
        return nil
    }
    return err
}
```

pkg/portfwd/client.go
```diff
@@ -24,26 +22,23 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
 
-	if err := g.Wait(); err != nil {
-		logrus.Debugf("error in tcp tunnel for id: %s error:%v", id, err)
+	remote := tcpproxy.DialProxy{
+		DialContext: func(_ context.Context, _, _ string) (net.Conn, error) {
+			rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr, protocol: "tcp"}
+			return rw, nil
+		},
 	}
+	remote.HandleConn(conn)
```
nirs, line 38: Delegating the code to another libary is nice, if the library we used is reliable and well maintained.

Is this same tcpproxy libary that does not log anything on errors and was replaced in vz/network_darwin.go?

We must have 2 properties:
- when one side is closed or fail during a copy, the other side is closed
- when the first copy fail, the error is logged or returned to the caller

Example failure flow:
1. first io.Copy() blocked reading from host side socket
2. second io.Copy() blocked reading from grpc socket
3. host side socket closed, first io.Copy() return
4. second io.Copy() continue to block reading from grpc
5. no error is logged since error handler is waiting for both io.Copy to finish

If this library does not ensure both properties we should not use it. It will impossible to debug when we have issues.

We can mitigate the logging issues by logging errors in Read() and Write() since forwarding will stop on the first error.

pkg/portfwd/client.go
```diff
@@ -130,14 +94,32 @@ func (g GrpcClientRW) Write(p []byte) (n int, err error) {
 func (g GrpcClientRW) Read(p []byte) (n int, err error) {
 	in, err := g.stream.Recv()
 	if err != nil {
-		if errors.Is(err, io.EOF) {
-			return 0, nil
-		}
```
nirs, line 135: Looks good, Read must return EOF

pkg/portfwd/client.go
```diff
@@ -130,14 +94,32 @@ func (g GrpcClientRW) Write(p []byte) (n int, err error) {
...
 		return 0, err
 	}
-	if len(in.Data) == 0 {
-		return 0, nil
-	}
```
nirs, line 140: This is not possible or just not needed, since the return value will be the same anyway if this happens?

pkg/portfwd/client.go
```diff
@@ -130,14 +94,32 @@ func (g GrpcClientRW) Write(p []byte) (n int, err error) {
...
+func (g GrpcClientRW) LocalAddr() net.Addr {
+	return &net.UnixAddr{Name: "grpc", Net: "unixpacket"}
+}
+
+func (g GrpcClientRW) RemoteAddr() net.Addr {
+	return &net.UnixAddr{Name: "grpc", Net: "unixpacket"}
+}
+
+func (g GrpcClientRW) SetDeadline(t time.Time) error {
+	return nil
```
nirs, line 116: Why we don't implement it? should we log a warning to make the issue visible?

pkg/portfwd/client.go
```diff
@@ -130,14 +94,32 @@ func (g GrpcClientRW) Write(p []byte) (n int, err error) {
 func (g GrpcClientRW) Read(p []byte) (n int, err error) {
 	in, err := g.stream.Recv()
 	if err != nil {
-		if errors.Is(err, io.EOF) {
-			return 0, nil
-		}
 		return 0, err
```
nirs, line 97: If we don't have proper logging in tcpproxy, we can mitigate it here by logging real read errors. Same can be done for Write().

pkg/portfwd/client.go
nirs: Looking in current tcpproxy used, it is the same library with the same issues:

```go
	errc := make(chan error, 1)
	go proxyCopy(errc, src, dst)
	go proxyCopy(errc, dst, src)
	<-errc
```

The errors from io.Copy are dropped silently.

```go
func proxyCopy(errc chan<- error, dst, src net.Conn) {
	// Before we unwrap src and/or dst, copy any buffered data.
	if wc, ok := src.(*Conn); ok && len(wc.Peeked) > 0 {
		if _, err := dst.Write(wc.Peeked); err != nil {
			errc <- err
			return
		}
		wc.Peeked = nil
	}

	// Unwrap the src and dst from *Conn to *net.TCPConn so Go
	// 1.11's splice optimization kicks in.
	src = UnderlyingConn(src)
	dst = UnderlyingConn(dst)

	_, err := io.Copy(dst, src)
	errc <- err
}
```

When one io.Copy finish (success of failure), the other io.Copy is not affected. If both are blocked on Read() or Recv(), what would cause the call to return?

We have the same issue in the current code, but at least in the current code we will log the error if both io.Copy() fail.

It can be useful to fix both issues in tcpproxy but it will be easier to fix them in lima now. Is this the same code used for user v2 network?

nirs:
Since this change convert the GRPC endpoint to net.Con interface, why not use it with bcopy? #2944

It has the same terminating issue as tcpproxy, but at least it logs errors. It will be easier to fix it to return an error and terminate the copies.

pkg/portfwd/client.go
```diff
@@ -52,74 +47,43 @@ func HandleUDPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
-		buf := make([]byte, 65507)
-		for {
-			n, addr, err := conn.ReadFrom(buf)
-			if err != nil {
-				// https://pkg.go.dev/net#PacketConn does not mention io.EOF semantics.
-				if errors.Is(err, io.EOF) {
-					return nil
-				}
-				return err
-			}
```
nirs, line 67: This was wrong in current code, we ignored the last n byte read if we got io.EOF. Fixed in #2971.

pkg/portfwd/client.go
```diff
@@ -52,74 +47,43 @@ func HandleUDPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
 
-	if err := g.Wait(); err != nil {
-		logrus.Debugf("error in udp tunnel for id: %s error:%v", id, err)
+	if err != nil {
+		logrus.Errorf("error in udp tunnel proxy for id: %s error:%v", id, err)
+		return
 	}
+
+	defer proxy.Close()
+	proxy.Run()
```
nirs, line 68: We don't get any error, or we ignore it? Are error logged in the proxy? Does ensure termination of both copies on errors or normal completion?

pkg/portfwd/client.go
nirs: Seems that it has the same issue:

```go
func (proxy *UDPProxy) Run() {
	readBuf := make([]byte, UDPBufSize)
	for {
		read, from, err := proxy.listener.ReadFrom(readBuf)
		if err != nil {
			// NOTE: Apparently ReadFrom doesn't return
			// ECONNREFUSED like Read do (see comment in
			// UDPProxy.replyLoop)
			if !isClosedError(err) {
				log.Debugf("Stopping udp proxy (%s)", err)
			}
			break
		}
```

It ignores the last n byte read.

It is calling listener.ReadFrom() which seems to have the same issue:

```go
func (pk *pseudoLoopbackPacketConn) ReadFrom(bytes []byte) (n int, addr net.Addr, err error) {
	n, remoteAddr, err := pk.PacketConn.ReadFrom(bytes)
	if err != nil {
		return 0, nil, err
	}

	remoteAddrIP, _, err := net.SplitHostPort(remoteAddr.String())
	if err != nil {
		return 0, nil, err
	}
	if !IsLoopback(remoteAddrIP) {
		return 0, nil, fmt.Errorf("pseudoloopback forwarder: rejecting non-loopback remoteAddr %q", remoteAddr)
	}
	return n, remoteAddr, nil
}
```

pkg/portfwdserver/server.go
balajiv113: Server doesn't have any stream closing methods. So it doesn't do anything

nirs: So this is implemented only to satisfy the net.Conn interface? Should we log a debug message when about ignoring Close?

This code can be integrated in other code expecting that Close() does something. If doing nothing breaks something, the debug log will help to diagnose the issue.

pkg/portfwd/client.go
balajiv113: Too much logging will become a noise since these are connections. It will be called multiple times and its doesn't help much. SetDeadline is not even usecul in our case as ours is a grpc stream

pkg/portfwdserver/server.go
balajiv113: debug logs are useful only in dev side, also grpc server doesn't provide close, its purely handled on client itself. I don't think adding debug is anywhere useful. 

I don't want to spam it in our guestagent logs which doesn't give much value.

pkg/portfwdserver/server.go
balajiv113: Done

pkg/portfwd/client.go
balajiv113: I want to keep the port forwarding stuffs consistent across portfwclient, portfwserver, user-v2, vz forwardings.

In server we need to use tcpproxy because we need to dial and pass connection. So i wanted to use the same on client side as well.

Additionally, gvisor-tap-vsock userspace network is already using these for proxy we are using the same for vz and user-v2 network. This library has been reliable from that point of view.

I intend to reuse as these are testing across different products (not just lima), so that we can avoid failures with custom implementation like before.

pkg/portfwd/client.go
balajiv113: Not needed. We will not get empty data. Even if it happens technically we should simply write empty data to proxy conn as well. So removed it

pkg/portfwd/client.go
balajiv113: We can raise PR with gvisor-tap-vsock to fix the UDPProxy part

balajiv113: Since we don't have a real failure case for it for now, i intend to do this as a follow up

pkg/portfwd/client.go
balajiv113: `defer proxy.Close()` - Takes care of proper completion. Also we have debug and error logs in UDPProxy

pkg/portfwdserver/server.go
balajiv113: I understand the reason behind. But i don't think we can create a common one. As server and client are 2 different one. Create a common will create cyclic dependency as well. 

At max we can do is, have a Base conn with method like setDeadline, setWriteDeadline etc. But for now i will avoid it. If you need for sure i can do that as well

balajiv113:
Ran DDEV Tests in buildkite. All tests passed

https://buildkite.com/test-305/ddev/builds/19#01939bef-af2a-4b7d-9ea8-fa1ca461f037

pkg/portfwdserver/server.go
```diff
@@ -1,69 +1,84 @@
...
 
-var _ io.Writer = (*GRPCServerWriter)(nil)
+var _ net.Conn = (*GRPCServerRW)(nil)
 
-func (g GRPCServerWriter) Write(p []byte) (n int, err error) {
-	if len(p) == 0 {
-		return 0, nil
-	}
-	err = g.stream.Send(&api.TunnelMessage{Id: g.id, Data: p, UdpTargetAddr: g.udpAddr})
+func (g GRPCServerRW) Write(p []byte) (n int, err error) {
```
tamird, line 48: consider making all of these functions have pointer receivers - when you implement an interface on a value rather than a pointer, it is considered to be implemented for *both* pointer and value, whereas implementing it on a pointer does not produce an implementation on a value.

This is usually not important, but can lead to subtle mistakes in the presence of type assertions; since `GRPCServerRW` *or* `*GRPCServerRW` can be passed as a `net.Conn` it becomes necessary to try to type assert both types, whereas only `x.(*GRPCServerRW)` would be necessary if the interface were implemented on the pointer.

pkg/portfwdserver/server.go
tamird: This is a pointless comment. The purpose of `LocalAddr` is to identify the connection (which this doesn't do) while the `NamedReadWriter` in #2944 is identifying the *purpose* of the connection.

This implementation here is a lie, it doesn't identify the connection and the address it returns cannot be used in any of the normal ways that `net.Addr` is used.

pkg/portfwd/client.go
```diff
@@ -24,26 +22,23 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
-	g.Go(func() error {
-		_, err := io.Copy(rw, conn)
-		return err
-	})
-	g.Go(func() error {
-		_, err := io.Copy(conn, rw)
-		return err
-	})
+	// Handshake message to start tunnel
+	err = stream.Send(&api.TunnelMessage{Id: id, Protocol: "tcp", GuestAddr: guestAddr})
```
tamird, line 26: consider using `if err :=` to avoid reusing the earlier `err`. the current code would continue to compile even if the following `if` is removed, which is slightly more fragile.

pkg/portfwd/client.go
```diff
@@ -52,74 +47,42 @@ func HandleUDPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
-				Data:          buf[:n],
-				UdpTargetAddr: addr.String(),
-			}
-			if err := stream.Send(msg); err != nil {
-				return err
-			}
-		}
-	})
+	// Handshake message to start tunnel
+	err = stream.Send(&api.TunnelMessage{Id: id, Protocol: "udp", GuestAddr: guestAddr})
```
tamird, line 51: ditto

pkg/portfwd/client.go
tamird: Shouldn't you log the error from `Close`?

```
defer func() {
  if err := proxy.Close(); err != nil {
    ...
  }
}()
```

pkg/portfwd/client.go
```diff
@@ -130,14 +93,32 @@ func (g GrpcClientRW) Write(p []byte) (n int, err error) {
...
 		return 0, err
 	}
-	if len(in.Data) == 0 {
-		return 0, nil
-	}
 	copy(p, in.Data)
 	return len(in.Data), nil
 }
+
+func (g GrpcClientRW) Close() error {
```
tamird, line 102: same as the comment in the next file: consider implementing these functions using a pointer receiver.

pkg/portfwdserver/server.go
```diff
@@ -1,69 +1,84 @@
...
+type TunnelServer struct{}
 
 func NewTunnelServer() *TunnelServer {
-	return &TunnelServer{
-		Conns: make(map[string]net.Conn),
-	}
+	return &TunnelServer{}
 }
 
 func (s *TunnelServer) Start(stream api.GuestService_TunnelServer) error {
```
tamird, line 20: TunnelServer is an empty type now. Why does it need to exist? This can be a free function.

balajiv113: I agree but just for usecases like before, it is useful to retain that empty. So that later if we want to support reverse forwarding we can use these with different methods / some caching etc.

pkg/portfwd/client.go
balajiv113: If you check the code, proxy.close always returns nil. So handling it i don't think useful

pkg/portfwdserver/server.go
balajiv113: Thanks completely missed it. Updated

pkg/portfwd/client.go
balajiv113: Done

pkg/portfwd/client.go
balajiv113: Done

pkg/portfwd/client.go
balajiv113: Done

tamird (commented):
I don't love this.

Delegating to a 3rd party library is a nice idea but as @nirs points out the error handling is subtle and in the case of this library, dubious. We also have to provide a largely fake interface implementation and instead of panicking (strict) we return nonsense (lax).

pkg/portfwd/client.go
balajiv113: Which file you mean ???

pkg/portfwd/client.go
balajiv113: Why should we panic ?? 

UDP connect is either way closed. There is no point in sending panic for that case.

pkg/portfwd/client.go
```diff
@@ -24,26 +22,22 @@ func HandleTCPConnection(ctx context.Context, client *guestagentclient.GuestAgen
...
-	rw := &GrpcClientRW{stream: stream, id: id, addr: guestAddr}
-	g.Go(func() error {
-		_, err := io.Copy(rw, conn)
-		return err
-	})
-	g.Go(func() error {
-		_, err := io.Copy(conn, rw)
-		return err
-	})
+	// Handshake message to start tunnel
```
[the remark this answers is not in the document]
balajiv113: No, added it newly to support proxy like model.

This is needed to initiate a tunnel on both ends (host & guest).

pkg/portfwd/client.go
tamird: If you assert that the error *is never nil* then you should write a panic to assert your assumption.

pkg/portfwd/client.go
balajiv113: Updated to use bicopy for server and client (Only tcp connection).

Upstream copied tcpproxy because they had a connection leak issue in latest tcpproxy version

pkg/portfwd/client.go
balajiv113: Am bit occupied, since its very edge case i thought of deferring it to a later point. 
Feel free to raise a PR to gvisor-tap-vsock.

pkg/portfwd/client.go
balajiv113: Am not interested in doing this change. Since the error during close does not have much value for us i don't see a need for this assert / log.

Additionally from udpProxy code its clear its always nil. So what ever we write its more of a dead code only.

balajiv113:
@nirs  & @tamird - Updated logic to use bicopy.

Re-verified the ddev run as well
https://buildkite.com/test-305/ddev/builds/21

pkg/portfwd/client.go
tamird: Dead code serves as proof evident that the error is guaranteed to be nil.

pkg/portfwd/client.go
tamird: Or just panic. If these functions don't do what they're supposed to (and are never called) a panic would be proof of that.

pkg/portfwd/client.go
tamird: Thanks, I'm fine. Please do it before landing this change.

pkg/portfwd/client.go
tamird: > Upstream copied tcpproxy because they had a connection leak issue in latest tcpproxy version

Did you report that issue to upstream? Where is the fix?

pkg/portfwd/client.go
balajiv113: We cannot panic on all cases.

These are left as NO-OP wantedly. These methods may / will be called but its of no value to us as its a grpc tunnel wrapped conn

pkg/portfwd/client.go
balajiv113: Check the git log for such reference

https://github.com/containers/gvisor-tap-vsock/commit/c315e611503f22d0b27f5c9f2857f823c3aff325

FYI - https://github.com/containers/gvisor-tap-vsock/issues/387

pkg/portfwd/client.go
balajiv113: Lets say we get error during close what do you think we need to do ??

Even if we have log how will it help us ???

In tunnels, we open and close lot of connections having these errors in logs adds more of a noise rather than being of any use.

Lets hear what others think. Since we have contradicting opinions. 

@AkihiroSuda what do you think on this ??

pkg/portfwd/client.go
balajiv113: I don't think we need this upstream change as a hard dependency to this PR.

If time permits i will fix prior if not we will take as a follow up.

pkg/portfwd/client.go
tamird: That's why my position is that you should panic rather than log.

pkg/portfwd/client.go
tamird: I am not a maintainer, but I see this as completely unnecessary technical debt.

pkg/portfwd/client.go
balajiv113: @tamird Am the one shared those links so i know what i shared. 

Marking this as resolved as its not relevant to us we are using bicopy only already

pkg/portfwd/client.go
balajiv113: @tamird Please understand the full flow before commenting. All left like that for a reason.

UDP Proxy internally will call SetReadDeadline so we cannot throw error. It will be no-op only. That's why all SetDeadline like methods are returning null.

jandubois: > no-op on SetDeadline is not a reasonable thing to do. It should return an error.

I think you are talking past each other. `SetDeadline()` must exist to implement the interface, but there is not actual implementation of that concept for this particular implementation (there is no deadline). Therefore the method does nothing. It doesn't mean it is an error; just a vestigial method that is not being used.

pkg/portfwd/client.go
balajiv113: To address your concern of this getting missed, Done error log to capture this if in case it happens.

Adding panic in application i don't think its a good idea. Marking this as resolved as err is logged now

pkg/portfwd/client.go
balajiv113: Resolving as we are using bicopy

AkihiroSuda (approved):
Thanks

PR #2987, opened 2024-12-06
Check if there is a newer engine available

### LINKED ISSUE #2978: `limactl start template://podman` installs an outdated version
`limactl start template://podman` installs Podman 5.2.5-1.fc41

It is upgraded to 5.3.1-1.fc41 after running `sudo dnf upgrade` though.

afbjorklund:
Previously only "nerdctl-full" was updated automatically, but docker and podman were never updated after installation.

The dependencies are supposed to be updated when necessary, by using the normal package manager's requirements system...

Note: might want to make this feature optional, by adding cidata variable?

Closes #2978

### DIFF
--- a/templates/docker-rootful.yaml
+++ b/templates/docker-rootful.yaml
@@ -43,7 +43,10 @@ provision:
   script: |
     #!/bin/bash
     set -eux -o pipefail
-    command -v docker >/dev/null 2>&1 && exit 0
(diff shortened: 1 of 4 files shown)


afbjorklund:
> But we only upgrade to the version bundled with Lima itself, not to whatever the latest release is on GitHub, right?

We only update to the Lima version, but I suppose you can upgrade Lima and have an older instance remaining?

PR #2990, opened 2024-12-07
Make sure to install podman package from repo

### LINKED ISSUE #2978: `limactl start template://podman` installs an outdated version

afbjorklund:
Newer versions of Fedora come with an older pre-installed Podman,
make sure to install the package from the repository like before.

Closes #2978

----

Fixes the issue on my machine:

```console
INFO[0019] READY. Run `limactl shell podman` to open the shell. 
INFO[0019] Message from the instance "podman":          
------
podman system connection add lima-podman "unix:///home/anders/.lima/podman/sock/podman.sock"
podman system connection default lima-podman
podman --remote run quay.io/podman/hello
------
$ podman.lima version
Client:       Podman Engine
Version:      4.9.3
API Version:  4.9.3
Go Version:   go1.21.7
Git Commit:   8d2b55ddde1bc81f43d018dfc1ac027c06b26a7f
Built:        Fri Feb 16 11:30:23 2024
OS/Arch:      linux/amd64

Server:       Podman Engine
Version:      5.3.1
API Version:  5.3.1
Go Version:   go1.23.3
Built:        Thu Nov 21 01:00:00 2024
OS/Arch:      linux/amd64
```

It needed --best too, or it would exit:

`Package "podman-5:5.2.5-1.fc41.x86_64" is already installed.`

### DIFF
--- a/templates/podman-rootful.yaml
+++ b/templates/podman-rootful.yaml
@@ -31,7 +31,7 @@ provision:
   script: |
     #!/bin/bash
     set -eux -o pipefail
-    command -v podman >/dev/null 2>&1 && exit 0
+    command -v podman >/dev/null 2>&1 && test -e /etc/lima-podman && exit 0
     if [ ! -e /etc/systemd/system/podman.socket.d/override.conf ]; then
       mkdir -p /etc/systemd/system/podman.socket.d
       cat <<-EOF >/etc/systemd/system/podman.socket.d/override.conf
@@ -43,7 +43,7 @@ provision:
       mkdir -p /etc/tmpfiles.d
       echo "d /run/podman 0700 {{.User}} -" > /etc/tmpfiles.d/podman.conf
     fi
-    dnf -y install podman
+    dnf -y install --best podman && touch /etc/lima-podman
 - mode: system
   script: |
     #!/bin/bash
--- a/templates/podman.yaml
+++ b/templates/podman.yaml
@@ -31,8 +31,8 @@ provision:
   script: |
     #!/bin/bash
     set -eux -o pipefail
-    command -v podman >/dev/null 2>&1 && exit 0
-    dnf -y install podman
+    command -v podman >/dev/null 2>&1 && test -e /etc/lima-podman && exit 0
+    dnf -y install --best podman && touch /etc/lima-podman
 - mode: user
   script: |
     #!/bin/bash


jandubois:
I thought we could put something in `boot.sh` to export a `LIMA_IS_FIRST_BOOT` variable, so templates don't have to implement it themselves, but this is probably good enough for now.

> install the package from the repository like before.

What do you mean by "like before"? Did `dnf` upgrade it before, but now it is broken?

jandubois:
I'm not sure if `/var/tmp/lima-podman` is the right location for this flag; it is not unconceivable that the user wants to clean that directory, and now this would implicitly trigger an update of `podman`.

I would suggest `/etc/lima-podman` instead.

jandubois:
> What do you mean by "like before"? Did `dnf` upgrade it before, but now it is broken?

Ok, I just realized that the issue is that Fedora comes with podman already installed, which disables our code that used to install the latest version on first boot.

So this PR does restore the old behaviour; the only reason I'm not merging is because I think the sentinel file should be in `/etc` instead of `/var/tmp`.

afbjorklund:
> > install the package from the repository like before.
> 
> What do you mean by "like before"? Did `dnf` upgrade it before, but now it is broken?

Like with older versions of Fedora, what was I referring to (but that info was only in issue)

templates/podman-rootful.yaml
```diff
@@ -43,7 +43,7 @@ provision:
       mkdir -p /etc/tmpfiles.d
       echo "d /run/podman 0700 {{.User}} -" > /etc/tmpfiles.d/podman.conf
     fi
-    dnf -y install podman
+    dnf -y install --best podman && touch /etc/lima-podman
```
AkihiroSuda, line 46: `--best` is nice, but `touch /etc/lima-podman` seems controversial

jandubois: I don't see how using `/etc/lima-podman` can create any problems. Every service nowadays seems to store their config in there (including apps like `nanorc` or `screenrc`):

```console
$ lima ls /etc | wc -l
     189
```

But if you don't like it, how about `mkdir -p /var/lib/lima` then? I don't really care; I was only pointing out that a temp dir that a user may reasonably wipe clean is a poor location for a file that should **never** be deleted.

jandubois (approved):
Thanks, LGTM.

Not merging myself as there doesn't seem to be consensus about the sentinel location yet.

AkihiroSuda:
> location

My concern was not about the location.
I'm not sure we really need the `/etc/lima-podman` or whatever else.
Probably just running `dnf install --best podman` once is enough for now?

jandubois:
> Probably just running `dnf install --best podman` once is enough for now?

That's what this is all about: how do you make sure that we are not running `dnf install --best podman` **again** at some random point in the future when the user did not intend to change their podman version?

Before we relied on Fedora not including `podman`, so we ran the install only when `podman` was not available via the `PATH`. This mechanism has been broken since Fedora started shipping with an older version of `podman` pre-installed.

So we want to upgrade to the latest `podman` on first-boot, but from then on leave it alone. So we need some kind of marker that says "we already did upgrade podman once". This marker would be the existence of the `/etc/lima-podman` file or `/var/lib/lima` directory, as neither of them exist on a pristine Fedora image.

How would you solve the issue of running `dnf install --best podman` only once?

nirs:
> > What do you mean by "like before"? Did `dnf` upgrade it before, but now it is broken?
> 
> Ok, I just realized that the issue is that Fedora comes with podman already installed, which disables our code that used to install the latest version on first boot.

So old podman is installed by default and `dnf install podman` installs a newer version?! This sounds like a packaging bug. Do we have a Fedora bug for this?

jandubois:
> So old podman is installed by default and `dnf install podman` installs a newer version?! This sounds like a packaging bug.

Unless I misunderstand, this is simply due to the latest Fedora 41 1.4 image being from Oct 24 containing podman 5.2.5 released on Oct 23.

Since then there have been podman releases 5.3.0 (Nov 13) and 5.3.1 (Nov 21) that you can install via `dnf`. It's not like the bundled version is massively out of date.

AkihiroSuda (approved):
Thanks

PR #2995, opened 2024-12-08
Improve guestagent startup error handling and logging

nirs:
- Clarify error handling when creating guest agent
- Improve startup logging to make the status of the guest agent more clear

### DIFF
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -35,6 +35,7 @@ func New(newTicker func() (<-chan time.Time, func()), iptablesIdle time.Duration
 		if !errors.Is(err, syscall.EPROTONOSUPPORT) && !errors.Is(err, syscall.EAFNOSUPPORT) {
 			return nil, err
 		}
+		logrus.Infof("Auditing is not available: %s", err)
 		return startGuestAgentRoutines(a, false), nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/guestagent/guestagent_linux.go
```diff
@@ -103,6 +107,7 @@ type agent struct {
 // setWorthCheckingIPTablesRoutine sets worthCheckingIPTables to be false
 // when no NETFILTER_CFG audit message was received for the iptablesIdle time.
 func (a *agent) setWorthCheckingIPTablesRoutine(auditClient *libaudit.AuditClient, iptablesIdle time.Duration) {
+	logrus.Info("setWorthCheckingIPTablesRoutine(): monitoring netfilter audit events")
```
AkihiroSuda, line 110: Perhaps:
```suggestion
	logrus.Debug"setWorthCheckingIPTablesRoutine(): monitoring netfilter audit events")
```

Same for other logs too

nirs: I'm using INFO messages for the other monitors (system time, kubernetes services). Why hide this monitor?

I think we need cleaner logging, maybe using specific logger or fields for every monitor instead of adding a prefix for every log, but this can be done later.

AkihiroSuda (approved):

Issue #2963, opened 2024-12-02
`limactl copy` should have `--quiet` flag


AkihiroSuda:
`limactl copy -r` spams an enormous amount of logs when the directory contains a lot of files (e.g., `.git` blobs)

nirs:
Or maybe it should be quiet by default and have a -v, --verbose flag?

khayyamsaleem:
close?

PR #2998, opened 2024-12-10
[limactl copy] change make logs quiet by default and adds a verbose flag to enable verbose log output

olamilekan000:
```
Change makes log output quiet by default and adds a "--verbose" flag for 
detailed logging in the Copy action function.
```

Link to issue - https://github.com/lima-vm/lima/issues/2963#issue-2712591901

### DIFF
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -33,12 +33,18 @@ func newCopyCommand() *cobra.Command {
 	}
 
 	copyCommand.Flags().BoolP("recursive", "r", false, "copy directories recursively")
+	copyCommand.Flags().BoolP("verbose", "v", false, "enable verbose output")
 
 	return copyCommand
 }
 
 func copyAction(cmd *cobra.Command, args []string) error {
 	recursive, err := cmd.Flags().GetBool("recursive")
+	if err != nil {
+		return err
+	}
+
+	verbose, err := cmd.Flags().GetBool("verbose")
 	if err != nil {
 		return err
 	}
@@ -54,9 +60,17 @@ func copyAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+
 	if debug {
-		scpFlags = append(scpFlags, "-v")
+		verbose = true
 	}
+
+	if verbose {
+		scpFlags = append(scpFlags, "-v")
+	} else {
+		scpFlags = append(scpFlags, "-q")
+	}
+
 	if recursive {
 		scpFlags = append(scpFlags, "-r")
 	}
@@ -119,7 +133,7 @@ func copyAction(cmd *cobra.Command, args []string) error {
 	sshCmd.Stdin = cmd.InOrStdin()
 	sshCmd.Stdout = cmd.OutOrStdout()
 	sshCmd.Stderr = cmd.ErrOrStderr()
-	logrus.Debugf("executing scp (may take a long time)): %+v", sshCmd.Args)
+	logrus.Debugf("executing scp (may take a long time): %+v", sshCmd.Args)
 
 	// TODO: use syscall.Exec directly (results in losing tty?)
 	return sshCmd.Run()


AkihiroSuda (dismissed):
Thanks

olamilekan000:
> Thanks

You're welcome!

cmd/limactl/copy.go
```diff
@@ -33,6 +33,7 @@ func newCopyCommand() *cobra.Command {
 	}
 
 	copyCommand.Flags().BoolP("recursive", "r", false, "copy directories recursively")
+	copyCommand.Flags().BoolP("quiet", "q", false, "suppress output logs")
```
nirs, line 36: I would consider making this quiet by default and adding -v --verbose for the verbose output, like the underling command we run and most commands.

AkihiroSuda: SGTM

cmd/limactl/copy.go
```diff
@@ -117,9 +128,17 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
-	sshCmd.Stdout = cmd.OutOrStdout()
-	sshCmd.Stderr = cmd.ErrOrStderr()
-	logrus.Debugf("executing scp (may take a long time)): %+v", sshCmd.Args)
+	if quiet {
+		sshCmd.Stdout = nil
+		sshCmd.Stderr = nil
+	} else {
+		sshCmd.Stdout = cmd.OutOrStdout()
+		sshCmd.Stderr = cmd.ErrOrStderr()
+	}
```
alexandear, line 137: Can be simplified:
```suggestion
	if !quiet {
		sshCmd.Stdout = cmd.OutOrStdout()
		sshCmd.Stderr = cmd.ErrOrStderr()
	}
```

The fields are `nil` by default.

cmd/limactl/copy.go
jandubois: I believe that's what he meant by "sounds good to me" (SGTM).

cmd/limactl/copy.go
```diff
@@ -43,23 +44,33 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
 	scpFlags := []string{}
 	scpArgs := []string{}
+
 	debug, err := cmd.Flags().GetBool("debug")
 	if err != nil {
 		return err
 	}
-	if debug {
+
+	if debug && !quiet {
```
jandubois, line 66: I don't think the `--quiet` option should override `--debug`. It should only apply to the output of the ssh command itself.

cmd/limactl/copy.go
olamilekan000: From the documentation, I can see that --debug flag is inherited. How will the introduction of --verbose flag affect the behaviour of the inherited flag or what distinct behaviour should the both have? @nirs

.gitignore
```diff
@@ -3,3 +3,4 @@ _artifacts/
 lima.REJECTED.yaml
 schema-limayaml.json
 .config
+.vscode
```
AkihiroSuda, line 6: This should be in your local gitignore

cmd/limactl/copy.go
AkihiroSuda, line 1: Please squash the commits

cmd/limactl/copy.go
```diff
@@ -33,6 +33,8 @@ func newCopyCommand() *cobra.Command {
 	}
 
 	copyCommand.Flags().BoolP("recursive", "r", false, "copy directories recursively")
+	copyCommand.Flags().BoolP("quiet", "q", true, "suppress output logs (default: true)")
+	copyCommand.Flags().BoolP("verbose", "v", false, "enable verbose output")
```
AkihiroSuda, line 37: I don't think we need the both

cmd/limactl/copy.go
alexandear: Why we need to squash commits? GitHub can do this automatically during merge. https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

AkihiroSuda: We expect contributors to care about the granularity of the commits (and to be familiar with the usage of `git`.

In this case a single commit should suffice, but for other cases multiple commits might be more preferable when a PR covers multiple topics.

So, GitHub's squash button does not work in general.

cmd/limactl/copy.go
olamilekan000: which are we going with?

cmd/limactl/copy.go
```diff
@@ -117,9 +128,12 @@ func copyAction(cmd *cobra.Command, args []string) error {
 
 	sshCmd := exec.Command(arg0, append(sshArgs, scpArgs...)...)
 	sshCmd.Stdin = cmd.InOrStdin()
-	sshCmd.Stdout = cmd.OutOrStdout()
-	sshCmd.Stderr = cmd.ErrOrStderr()
```
AkihiroSuda, line 121: These lines can be kept.

`scp` should be invoked with `-q` unless `--verbose` is set.

olamilekan000: The` -q `flag alone doesn't seem to quiet the logs as expected. I’ve attached a screenshot to illustrate the issue. The first image shows the args or flags and the next shows the output from the action. @AkihiroSuda 


Screenshot 2024-12-24 at 17 25 35
Screenshot 2024-12-24 at 17 25 53

AkihiroSuda: Permission errors should not be silenced.

olamilekan000: Okay. Got that!

olamilekan000: @AkihiroSuda How do I update the documentation to reflect this new change? I have checked through the repository but couldn't find any pointer.

cmd/limactl/copy.go
```diff
@@ -43,23 +44,35 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
-	arg0, err := exec.LookPath("scp")
+	verbose, err := cmd.Flags().GetBool("verbose")
 	if err != nil {
 		return err
 	}
-	instances := make(map[string]*store.Instance)
-	scpFlags := []string{}
-	scpArgs := []string{}
+
 	debug, err := cmd.Flags().GetBool("debug")
```
AkihiroSuda, line 52: `if err != nil { return err }` seems accidentally removed

cmd/limactl/copy.go
```diff
@@ -43,23 +44,35 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+
+	arg0, err := exec.LookPath("scp")
 	if err != nil {
 		return err
 	}
-	if debug {
+
+	instances := make(map[string]*store.Instance)
+	scpFlags := []string{}
+	scpArgs := []string{}
```
AkihiroSuda, line 64: No need to move

cmd/limactl/copy.go
```diff
@@ -43,23 +44,35 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+	if verbose {
 		scpFlags = append(scpFlags, "-v")
+	} else {
+		scpFlags = append(scpFlags, "-q")
 	}
+
 	if recursive {
 		scpFlags = append(scpFlags, "-r")
 	}
+
```
AkihiroSuda, line 75: No need to add new lines

cmd/limactl/copy.go
AkihiroSuda: No need to update the docs this case IMHO

AkihiroSuda: https://lima-vm.io/docs/reference/limactl_copy/ will be automatically updated after merging this PR

AkihiroSuda (approved):
Thanks

PR #3005, opened 2024-12-10
Fix typo in Port Forwarding documentation

cpick:
I had to re-read the bullet a few times before I figured out it probably meant "spawns".


jandubois:
I know it is just a one-character patch, but we require a DCO line on each commit: https://github.com/lima-vm/lima/pull/3005/checks?check_run_id=34219351806

cpick:
Signed off.

jandubois (approved):
Thanks, LGTM

PR #3008, opened 2024-12-11
Use yq string evaluator instead of stream evaluator

jandubois:
This allows us to pass the YAML content in memory instead of having to write it to a file.

This commit also switches to "eval-all" mode, so multiple documents are all loaded together, allowing merge operations between them.

This is in preparation of multi-file template assembly.


AkihiroSuda (approved):
Thanks

PR #3009, opened 2024-12-11
Move template locator and reader code into new limatmpl package

jandubois:
It was part of cmd/limactl/start.go but should be generally reusable code. Moved related cmd/limactl/guessarg into limatmpl as well.

Also created a new `limactl template` command.

It takes a single template locator argument, loads the template, and writes it to STDOUT.

This is currently not very interesting; it is in preparation of implementing composable (multi-file) templates.


pkg/limatmpl/locator.go
```diff
@@ -0,0 +1,166 @@
+package limatmpl
```
AkihiroSuda, line 1: Maybe:
```suggestion
package templatelocator
```

cmd/limactl/template.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"fmt"
+
+	"github.com/lima-vm/lima/pkg/limatmpl"
+	"github.com/spf13/cobra"
+)
+
+func newTemplateCommand() *cobra.Command {
+	validateCommand := &cobra.Command{
```
AkihiroSuda, line 11: Not "validate"

cmd/limactl/template.go
```diff
@@ -0,0 +1,31 @@
...
+		Use:     "template FILE.yaml|URL",
```
AkihiroSuda, line 12: Probably this should be like `limactl template read` (or `print`), so that we can add more subcommands under `limactl template *`

AkihiroSuda: `limactl validate FILE.yaml` should be probably moved to `limactl template validate LOCATOR`

nirs:
> It takes a single template locator argument, loads the template, and writes it to STDOUT.
> 
> This is currently not very interesting; it is in preparation of implementing composable (multi-file) templates.

This *is* interesting, a possible use case is creating new yaml from template, and then creating one or move vms from the yaml. Can be done today by copying a template from source, or from `/opt/homebrew/share/doc/lima/templates/default.yaml` which is much harder and error prone compared with `template://default`.

 Not sure that template is a good name since we repeat "template":

    limactl template template://default

Maybe it should work with edit?

    limactl edit template://default

If this may be confusing since edit is used for instances, maybe:

    limactl generate template://default

jandubois:
> Not sure that template is a good name since we repeat "template":

As @AkihiroSuda already pointed out, we will need additional subcommands under `template`, so it will no longer stutter:

```
limactl template print template://docker
limactl template assemble my.yaml
limactl template with-defaults my.yaml
limactl template validate your.yaml
```

I'm not sure about the `with-defaults` subcommand name, but it would be similar to `limactl validate --fill`, but in addition to the builtin defaults it would also combine it with `default.yaml` and `override.yaml`.

Maybe we need to have 2 different subcommands for that. I find it challenging to come up with meaningful names.

I will make the `template` command hidden, so we can continue to evolve it without worrying about backwards compatibility.

I still think `templates` is the best name for that group of commands.

I'm more concerned about the confusion with "text templates" inside `lima.yaml`, like we have "host templates", and "guest templates", and potentially "param templates" that all have different sets of variables for interpolation. But this is a concern for writing the docs; at the CLI level a `template` should always be a `lima.yaml` template.

Just for clarification, I consider `my.yaml` to be a template name too, it just isn't using the `template://` URI scheme. That's why I started using the term "template locator" instead, similar to how URLs can use various schemes to specify the location.

jandubois:
Just for illustration (and I guess as a teaser), here is an example of `limactl template assemble my.yaml`[^1]:

[^1]: I cheated, currently the command is `limactl template my.yaml`, but I will change it to use the `assemble` subcommand.

```console
$ cat my.yaml
basedOn: template://fedora
mountTypesUnsupported: [virtiofs]
provision:
# Hello world!
- file: my.sh # Hi there

$ cat my.sh
#!/bin/sh
echo "Hello world!"

$ limactl template assemble my.yaml
# 9p is broken in Linux v6.9, v6.10, and v6.11 (used by Fedora 41).
# The issue was fixed in Linux v6.12-rc5 (https://github.com/torvalds/linux/commit/be2ca38).
mountTypesUnsupported: [virtiofs, "9p"]
provision:
# Hello world!
- script: |  # Hi there
    #!/bin/sh
    echo "Hello world!"
# This template requires Lima v0.7.0 or later.
images:
- location: "https://download.fedoraproject.org/pub/fedora/linux/releases/41/Cloud/x86_64/images/Fedora-Cloud-Base-Generic-41-1.4.x86_64.qcow2"
  arch: "x86_64"
  digest: "sha256:6205ae0c524b4d1816dbd3573ce29b5c44ed26c9fbc874fbe48c41c89dd0bac2"
- location: "https://download.fedoraproject.org/pub/fedora/linux/releases/41/Cloud/aarch64/images/Fedora-Cloud-Base-Generic-41-1.4.aarch64.qcow2"
  arch: "aarch64"
  digest: "sha256:085883b42c7e3b980e366a1fe006cd0ff15877f7e6e984426f3c6c67c7cc2faa"
- location: "https://download.fedoraproject.org/pub/fedora/linux/releases/41/Cloud/x86_64/images/Fedora-Cloud-Base-Generic-41-1.4.x86_64.qcow2"
  arch: "x86_64"
  digest: "sha256:6205ae0c524b4d1816dbd3573ce29b5c44ed26c9fbc874fbe48c41c89dd0bac2"
- location: "https://download.fedoraproject.org/pub/fedora/linux/releases/41/Cloud/aarch64/images/Fedora-Cloud-Base-Generic-41-1.4.aarch64.qcow2"
  arch: "aarch64"
  digest: "sha256:085883b42c7e3b980e366a1fe006cd0ff15877f7e6e984426f3c6c67c7cc2faa"
mounts:
- location: "~"
- location: "/tmp/lima"
  writable: true
```

You can see how both `template://fedora` and `my.sh` got merged into `my.yaml`, and how even the comments on the `provision[0].file` node moved to the `provision[0].script` entry.

And you can see how `mountTypesUnsupported` got merged, kept the comments from `template://fedora` because `my.yaml` didn't have a comment on that node, and the string presentation were maintained (`virtiofs` doesn't have quotes, `"9p"` does).

nirs:
Tanks for the example, it is very helpful. `limactl template assemble` could be nicer as `limactl template build`.

jandubois:
> Tanks for the example, it is very helpful. `limactl template assemble` could be nicer as `limactl template build`.

We can bike-shed the names later. I went through several already, and used `compose` before I switched to `assemble`.

To me `build` is creating something from basic ingredients, whereas `assemble` means just combining existing more complex pieces into a whole. For that I like `assemble` better. I do see that `build` is shorter and easier to type... For now it is `tmpl.Assemble()` to perform the magic.

cmd/limactl/template.go
jandubois: I've added `print` and `validate` subcommands. `validate` currently invokes the old `validateAction()`. The `limactl template` command is now hidden, so it can still be modified, even if it is included in an earlier release.

pkg/limatmpl/locator.go
jandubois: It may not make sense right now, but there will be a `limatmpl/assemble.go` in the future that contains all methods on a `limatmpl.Template` receiver, and then the package name `templatelocator` doesn't make sense anymore.

And while I haven't written it yet, the new `FillDefaults()` will also have to move to the `limatmpl` package (to avoid a circular dependency with `usrlocalsharelima`, which indirectly needs `limayaml`).

So for now at least I would like to keep the `limatmpl` package name.

jandubois:
I've realized that `limactl template print` should probably be `limactl template copy` because in the future it will create a working copy of the template, not print it verbatim.

The difference comes in with the `tmpl.Assemble` implementation that supports "relative template locators" (it is already implemented).

Assume you have `templates/my.yaml` stored in the lima.git repo, and also installed in `/usr/local/share/lima/templates/my.yaml`:

```yaml
basedOn: fedora.yaml
provision:
- file: my.sh
```

You can reference it as `./template/my.yaml`, `template://my`, or `https://raw.githubusercontent.com/lima-vm/lima/refs/heads/master/templates/my.yaml`. This will create different copies, to keep the copy working the same way as the original:

* `limactl template copy ./templates/my.yaml`

  ```yaml
  basedOn: /Users/USERNAME/git/lima/templates/fedora.yaml
  provision:
  - file: /Users/USERNAME/git/lima/templates/my.sh
  ```

* `limactl template copy template://my`[^1]

  ```yaml
  basedOn: template://fedora.yaml
  provision:
  - file: template://my.sh
  ```

* `limactl template copy https://raw.githubusercontent.com/lima-vm/lima/refs/heads/master/templates/my.yaml`

  ```yaml
  basedOn: https://raw.githubusercontent.com/lima-vm/lima/refs/heads/master/templates/fedora.yaml
  provision:
  - file: https://raw.githubusercontent.com/lima-vm/lima/refs/heads/master/templates/my.sh
  ```

[^1]: I'm glossing over the fact that the `template://` scheme now supports file extensions, and can be used for script references too.

This means the copy will continue to reference the same base `fedora.yaml` template and the same `my.sh` provisioning file as the source template, even though the copy will be stored in a different location.

I'll rename the `print` subcommand to `copy` in the PR implementing relative locators, unless there are still other changes required for the current PR.

AkihiroSuda:
> limactl template copy

What about `limactl render` ?

AkihiroSuda:
Side topic:

> basedOn: https://raw.githubusercontent.com/lima-vm/lima/refs/heads/master/templates/fedora.yaml

This has to support specifying sha256 digest

jandubois:
> This has to support specifying sha256 digest

Not sure what that means. Digest for what? For the template content? That seems impractical. If you need to fetch a digest, you can just fetch the template instead and load it locally because the templates are small.

How do you specify a digest for

```
limactl start https://raw.githubusercontent.com/lima-vm/lima/refs/heads/master/templates/fedora.yaml
```

If you want to have an immutable reference, maybe store your templates in an OCI registry and reference the data by hash?

jandubois:
> What about `limactl template render` ?

`render` sounds more like what I currently call `assemble`, that is embed all external references into a single `lima.yaml` so it can be stored in the instance directory.

The `copy` command just produces a copy of the template as-is, except all relative references are replaced by absolute references. Otherwise you end up with a broken copy.

I'm not sure how important the `copy` functionality is; I suspect you normally want `assemble`.

The part I dislike about `render` is that I expect it to also evaluate all the text templates inside the template file, like `home: "/home/{{.User}}"`.

It may all make more sense once the actual implementation is merged (which isn't even part of this PR). I've made the `limactl template` command "hidden" so we can continue to bike-shed it.

AkihiroSuda:
> Digest

The digest of the source template.
e.g.,
```yaml
base:
  location: "https://example.com/a.yaml"
  digest: "sha256:deadbeef"
```

As a side topic, `limactl start https://...` should also have a flag like `--template-digest=sha256:deadbeef`

AkihiroSuda:
> copy

I expect any command named "copy" to take two (or more) arguments: SRC and DST

jandubois:
> I expect any command named "copy" to take two (or more) arguments: SRC and DST

That totally makes sense. I was already thinking about adding an `--output` option so you don't have to use shell redirection to write to a file, but requiring a DST (and supporting `-` for STDOUT) makes much more sense.

Maybe most of the other planned subcommands, like `assemble` should just be options on `copy`?

```
copy --assemble  # embed all base templates and script files
copy --fill      # merge override.yaml, default.yaml, and builtin defaults
copy --render    # evaluate all Go text templates

copy             # make relative locators absolute
copy --verbatim  # copy as-is
```

Practically speaking `--fill` should imply `--assemble`, and `--render` should imply `--fill`.

* `limactl template copy --assemble` is what `limactl create` would write to `lima.yaml`.
* `limactl template copy --fill` is what we pass to validation.
* `limactl template copy --render` is what is used to create cidata and start the instance.

Maybe we can allow the user to override, like `copy --render --assenble=false`. Not sure there is a use-case for this except for Lima developers.

jandubois:
> copy --assemble  # embed all base templates and script files

Maybe `--assemble` (and `tmpl.Assemble()` etc.) should be `--embed` (and `tmpl.Embed()` etc.).

jandubois:
> copy --fill      # merge override.yaml, default.yaml, and builtin defaults
> copy --render    # evaluate all Go text templates

Actually, I'm not sure about this; currently filling defaults and evaluating text templates is intermingled, and may need to continue to be so. So we may just have `--fill` and it would include the rendered text templates.

jandubois:
> As a side topic, `limactl start https://...` should also have a flag like `--template-digest=sha256:deadbeef`

A side topic to the side note... 😄 

I've moved this to a discussion topic at #3019.

jandubois:
I've renamed `print` to `copy` and require a target filename. I've also added a `tmpl` alias.

So `limactl template print template://my` is now `limactl tmpl copy template://my -`.

@AkihiroSuda I don't plan to make any other changes; do you have more feedback?

cmd/limactl/validate.go
AkihiroSuda, line 1: This should now invoke `limactl template validate`

cmd/limactl/template.go
```diff
@@ -0,0 +1,69 @@
...
+
+func newTemplateCommand() *cobra.Command {
+	templateCommand := &cobra.Command{
+		Use:           "template",
+		Aliases:       []string{"tmpl"},
+		Short:         "Lima template management",
+		SilenceUsage:  true,
+		SilenceErrors: true,
+		GroupID:       advancedCommand,
+		Hidden:        true,
```
AkihiroSuda, line 19: The reason should be explained in a comment or `--help`

cmd/limactl/template.go
```diff
@@ -0,0 +1,69 @@
...
+		newTemplateCopyCommand(),
+		newTemplateValidateCommand(),
+	)
+	return templateCommand
+}
+
+func newTemplateCopyCommand() *cobra.Command {
+	templateCopyCommand := &cobra.Command{
+		Use:   "copy TEMPLATE DEST",
+		Short: "Copy template",
```
AkihiroSuda, line 31: Maybe add `Long:` to explain the detailed usage and the motivation

AkihiroSuda: ping @jandubois

cmd/limactl/validate.go
jandubois: Ok, `limactl validate` now invokes `templateValidateAction`.

cmd/limactl/template.go
jandubois: Sorry, I was busy this week with other things; PTAL

cmd/limactl/template.go
```diff
@@ -0,0 +1,137 @@
...
+	return validateCommand
+}
+
+var templateCopyExample = `  Template locators are local files, file://, https://, or template:// URLs
+
+  # Copy default template to STDOUT
+  limactl template copy template://default -
+
+  # Copy template from web location to local file
+  limactl template copy https://templates.r.us/lima.yaml mighty-machine.yaml
```
AkihiroSuda, line 47: What is r.us?

AkihiroSuda: This should probably use example.com

jandubois: > What is r.us?

It was supposed to be a joke, referencing "Toys'R'Us" (https://www.toysrus.ca) and similar names like "Babies'R'Us" or "Pets'R'Us".

> This should probably use example.com

Yes, the `--help` output should not try to be funny (or confusing). Sorry about that!

Changed to `example.com.

jandubois: Just realized that `mighty-machine.yaml` was also a joke (reference to https://www.youtube.com/channel/UCgCV4LtUjiBL8-6cqqMeuFQ). But I think this one is still ok, or do you want me to change it too?

AkihiroSuda (approved):
Thanks

Issue #2512, opened 2024-07-25
Call `limadriver.Validate` as part of `limayaml.Validate`


jandubois:
> They can return errors in validation, instead of warnings?

@afbjorklund Yes, they can. The warning comes from `func (l *LimaVzDriver) Validate() error {` and just write to the log, but don't return an error:

https://github.com/lima-vm/lima/blob/374db8bb91ecd8a3e78bb8d48351fd05e2fa2214/pkg/vz/vz_driver_darwin.go#L134-L139

I think all the warning should return an error instead (but in a separate PR).

I also think `limaDriver.Validate()` should be called as part of `limayaml.Validate()` or through some wrapper that will call both. Calling it as part of `instance.Prepare()` is too late.

_Originally posted by @jandubois in https://github.com/lima-vm/lima/pull/1951#discussion_r1690851617_

jandubois:
An alternative would be to move most or all of the driver-based validation into `limayaml.Validate`.

The logic is already leaking into that function e.g. in #1951 in the code that selects the default `vmType` based on all the other settings. So the knowledge about which settings are compatible with which driver already lives there. Moving the rest of the driver validation into the same function just consolidates all of it into a single place.

I understand that conceptually having a separate per-driver validation function is desirable if/when we ever support external (non-builtin) drivers, but we don't have them currently.

jandubois:
I just noticed that some driver-specific settings are not validated at all. E.g.

```shell
$ limactl start --vm-type qemu --set .rosetta.enabled=true
? Creating an instance "default" Proceed with the current configuration
INFO[0001] Starting the instance "default" with VM driver "qemu"
…

$ yq .rosetta ~/.lima/default/lima.yaml
# Enable Rosetta for Linux (EXPERIMENTAL).
# Hint: try `softwareupdate --install-rosetta` if Lima gets stuck at `Installing rosetta...`
# 🟢 Builtin default: false
enabled: true
…
```

will not even show a warning; the Rosetta setting is just silently ignored by the QEMU driver.

I think this is just another argument in favour of merging all validation into `limayaml.Validate()`.

PR #3010, opened 2024-12-11
Silence `WARN[0000] vmType vz: ignoring networks[0]: [Metric]`

### LINKED ISSUE #2986: v1.0.1: `WARN[0000] vmType vz: ignoring networks[0]: [Metric]`
> `limactl start` now prints
> ```
> WARN[0000] vmType vz: ignoring networks[0]: [Metric]
> ```

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/2868#discussion_r1873714239_

jandubois:
Fix #2986

### DIFF
--- a/pkg/vz/vz_driver_darwin.go
+++ b/pkg/vz/vz_driver_darwin.go
@@ -130,6 +130,7 @@ func (l *LimaVzDriver) Validate() error {
 			"Lima",
 			"Socket",
 			"MACAddress",
+			"Metric",
 			"Interface",
 		); len(unknown) > 0 {
 			logrus.Warnf("vmType %s: ignoring networks[%d]: %+v", *l.Instance.Config.VMType, i, unknown)


AkihiroSuda (approved):
Thanks

pkg/vz/vz_driver_darwin.go
```diff
@@ -130,6 +130,7 @@ func (l *LimaVzDriver) Validate() error {
 			"Lima",
 			"Socket",
 			"MACAddress",
+			"Metric",
 			"Interface",
 		); len(unknown) > 0 {
 			logrus.Warnf("vmType %s: ignoring networks[%d]: %+v", *l.Instance.Config.VMType, i, unknown)
```
nirs, line 136: It is easy to grep for the warning text and understand why it happens but the warning should make it clear that this is a validation warning, and the issue is unknown property. Maybe something like:

    validate vmType vz: ignoring unknown property networks[1]: "Metric"

jandubois: SGTM

If you are going to work on this, you should also take a look at the corresponding code in the `pkg/wsl2/wsl_driver_windows.go` file to try to keep things in sync (I noticed vz uses `ignoring` while `wsl2` uses `Ignoring`, which caused me to miss them on my initial grepping).

jandubois: Of course validation is going to be a big rabbit hole; look at #2512 for more thoughts on how it should be refactored.

Personally I don't want to touch it (validation) until I had a chance to replace the current `FillDefaults` with the upcoming code from template assembly because that might change things around anyways.

nirs: I don't have time for this now, added 3011

Issue #3022, opened 2024-12-14
CentOS Stream 10 guest kernel panic (`invalid opcode: 0000` in `chacha_8block_xor_avx512vl`)


AkihiroSuda:
CentOS Stream 10 guest kernel panics:
```
[    1.554186] Oops: invalid opcode: 0000 [#1] PREEMPT SMP PTI                                                                          23:11:52 [26/831]
[    1.554408] CPU: 1 UID: 0 PID: 1 Comm: swapper/0 Not tainted 6.12.0-32.el10.x86_64 #1
[    1.554408] Hardware name: QEMU Standard PC (Q35 + ICH9, 2009), BIOS rel-1.16.3-0-ga6ed6b701f0a-prebuilt.qemu.org 04/01/2014
[    1.554408] RIP: 0010:chacha_8block_xor_avx512vl+0x6e/0x6a0                                                                                           [    1.554408] Code: 7d 58 57 28 c4 62 7d 58 5f 2c c4 62 7d 58 67 30 c4 62 7d 58 6f 34 c4 62 7d 58 77 38 c4 62 7d 58 7f 3c c5 1d fe 25 52 f0 8b 01 <62> e
1 fd 28 6f c0 62 e1 fd 28 6f c9 62 e1 fd 28 6f d2 62 e1 fd 28
[    1.554408] RSP: 0000:ffffb2aec0013a18 EFLAGS: 00010206                                                                                               
[    1.554408] RAX: 0000000000000000 RBX: ffffb2aec0013cb0 RCX: 0000000000000109
[    1.554408] RDX: ffffffffb93b5ac0 RSI: ffff9aaaf43ff000 RDI: ffffb2aec0013cb0
[    1.554408] RBP: 0000000000000109 R08: 0000000000000014 R09: 17ae28094b4ea975
[    1.554408] R10: 0000000000000000 R11: 09dee45c094af0bd R12: 0000000000000109
[    1.554408] R13: 0000000000000014 R14: ffffb2aec0013cb0 R15: 0000000000000014           
[    1.554408] FS:  0000000000000000(0000) GS:ffff9aab3bc80000(0000) knlGS:0000000000000000
[    1.554408] CS:  0010 DS: 0000 ES: 0000 CR0: 0000000080050033                                                                                         
[    1.554408] CR2: 0000000000000000 CR3: 0000000132c22001 CR4: 00000000003706f0
[    1.554408] Call Trace:                                                  
[...]

PR #3023, opened 2024-12-14
Allow running Enterprise Linux release 10 too

afbjorklund:
Previously failing to install sshfs package.

https://github.com/lima-vm/lima/issues/3022#issuecomment-2543160594


jandubois (approved):
Thanks, LGTM

PR #3034, opened 2024-12-19
ci: raise timeout for golangci

PascalBourdier:
to avoid it locally:
```
ERRO Running error: context loading failed: failed to load packages: failed to load packages: failed to load with go/packages: context deadline exceeded
ERRO Timeout exceeded: try increasing it by passing --timeout option
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ _artifacts/
 lima.REJECTED.yaml
 schema-limayaml.json
 .config
-.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ _artifacts/
 lima.REJECTED.yaml
 schema-limayaml.json
 .config
+.idea/
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -18,7 +18,7 @@
 ---
 run:
   concurrency: 6
-  timeout: 2m
+  timeout: 2m # 1m by default
 linters:
   disable-all: true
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@ _artifacts/
 lima.REJECTED.yaml
 schema-limayaml.json
 .config
+.idea/
```
jandubois, line 6: Please don't put your personal config into the project-specific settings files.

This configuration belongs into your `~/.gitignore` file, so it applies to all your repos.

.golangci.yml
```diff
@@ -18,6 +18,7 @@
 ---
 run:
   concurrency: 6
+  timeout: 2m # 1m by default
```
jandubois, line 21: Our YAML linter rules requires at least 2 spaces before a comment:

```suggestion
  timeout: 2m  # 1m by default
```

jandubois (changes requested):
I don't see any changes in this comment except for the comment on the existing timeout value. It doesn't match what the PR description says, which also mentions local execution.

PascalBourdier:
> I don't see any changes in this comment except for the comment on the existing timeout value. It doesn't match what the PR description says, which also mentions local execution.

I fixed it according you review

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Next time please squash the commits

PR #3043, opened 2024-12-24
templates: update

AkihiroSuda:


jandubois (approved):
Approving because CI has passed; I did not test anything manually.

PR #3047, opened 2024-12-24
templates: add centos-stream-10

AkihiroSuda:


jandubois (commented):
I've confirmed that the template works on aarch64, but shouldn't the image for x86_64 be commented out if it is known to be broken?

AkihiroSuda:
> I've confirmed that the template works on aarch64, but shouldn't the image for x86_64 be commented out if it is known to be broken?

It still may work on somebody's machine?

jandubois (dismissed):
Thanks, LGTM

jandubois:
> It still may work on somebody's machine?

May, but sounds unlikely. It is easy enough to uncomment the image setting to try.

In the end we do want to know when it is working, and having to uncomment the image would give them a hint to let us know.

But if you want to merge as-is, then go ahead.

afbjorklund:
It did work in QEMU

afbjorklund:
Maybe it would help to disable the AVX512 support in the cpuType (i.e. not use "host") and go for AVX2 only?

EL10 is known to **not** run on a lot of older Intel, which is why AlmaLinux re-enabled the support for "v2"

templates/centos-stream-10.yaml
```diff
@@ -0,0 +1,27 @@
...
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+firmware:
+  # CentOS Stream 10 still requires legacyBIOS
+  legacyBIOS: true
+cpuType:
+  # Does not seem to boot with "qemu64"
+  x86_64: "Haswell-v4"
```
AkihiroSuda, line 27: This seems to fix #3022 (Intel MacBook Pro 2020) too

afbjorklund: qemu64-v1 is not supported by EL10, it requires v3 (or v2)

afbjorklund: I thought that it would use "host" by default, on Intel?

AkihiroSuda: On ARM it uses "qemu64", and the kernel just reboots infinitely

afbjorklund: Right, that would be expected. So this was when testing on ARM, got it.

AkihiroSuda: Found a comparison table: https://www.qemu.org/docs/master/system/i386/cpu.html

Surprised to know that qemu64 sticks to `x86_64-v1` by default 👀

afbjorklund: Did you have the qemu line that was used, originally. It was `-cpu host`, right?

AkihiroSuda: Updated the PR to explain the issue for both ARM (qemu64) and Intel (host)

AkihiroSuda: > Did you have the qemu line that was used, originally. It was -cpu host, right?

It is chosen by Lima itself; "qemu64" for ARM, "host" ("-pdpe1gb") for Intel

AkihiroSuda: (GitHub posts seem to be temporarily delayed?)

afbjorklund: > (GitHub posts seem to be temporarily delayed?)

We can use Slack if wanting to have a conversation.

afbjorklund:
The same emulation issue also existed for centos-stream-9 (but didn't for centos-stream-8)

`Fatal glibc error: CPU does not support x86-64-v2`

AkihiroSuda:
> centos-stream-9

Let's fix this separately in:
- https://github.com/lima-vm/lima/issues/3063

afbjorklund (approved):

PR #3048, opened 2024-12-24
templates: archlinux: allow 9p again

AkihiroSuda:
9p was once broken, but it was fixed in Linux v6.12 (https://github.com/torvalds/linux/commit/be2ca38).


jandubois (approved):
Thanks, confirmed on Intel. Not tested on M1 because the image is still using the 5.17 kernel there.

PR #3054, opened 2024-12-25
limactl shell: avoid silencing errors

### LINKED ISSUE #2980: `limactl shell INSTANCE` should show an explicit error
`limactl shell INSTANCE` may fail e.g., when `sshd` in the guest is dead.

However, no error is shown currently.

```console
$ limactl shell foo
$ echo $?
255
```

- - -
`ssh_config(5)`

> ```
>      LogLevel
>              Gives the verbosity level that is used when logging messages from ssh(1).  The possible values are:
>              QUIET, FATAL, ERROR, INFO, VERBOSE, DEBUG, DEBUG1, DEBUG2, and DEBUG3.  The default is INFO.  DEBUG and
>              DEBUG1 are equivalent.  DEBUG2 and DEBUG3 each specify higher levels of verbose output.
> ```

AkihiroSuda:
Fix #2980

Before:
```console
$ limactl shell alpine
$ echo $?
255
```

After:
```console
$ limactl shell alpine
suda@127.0.0.1: Permission denied (publickey,keyboard-interactive).
$ echo $?
255
```

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -185,7 +185,7 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		sshArgs = append(sshArgs, "-o", "SendEnv=COLORTERM")
 	}
 	sshArgs = append(sshArgs, []string{
-		"-q",
+		"-o", "LogLevel=ERROR",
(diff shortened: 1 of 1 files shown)


jandubois (approved):
Thanks, LGTM

PR #3059, opened 2024-12-25
website: docs/installation: update

AkihiroSuda:
- Clarify the supported OS
- Remove `brew install jq`, as `jq` is now included in macOS, and `brew` is not always available.
- Explain the `$PREFIX` variable
- Clarify that `make config` is not needed for most users.

### DIFF
--- a/website/content/en/docs/installation/_index.md
+++ b/website/content/en/docs/installation/_index.md
@@ -2,8 +2,13 @@
 title: Installation
 weight: 1
 ---
-Prerequisite:
+Supported host OS:
+- macOS (the latest version is recommended)
+- Linux
+- NetBSD (untested)
+- Windows (untested)
 
+Prerequisite:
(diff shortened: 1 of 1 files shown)


jandubois:
> * Remove `brew install jq`, as `jq` is now included in macOS, and `brew` is not always available.

Does this mean the instructions are only covering the very latest release of macOS?

AkihiroSuda:
> > * Remove `brew install jq`, as `jq` is now included in macOS, and `brew` is not always available.
> 
> Does this mean the instructions are only covering the very latest release of macOS?

No, covers Linux too, assuming that most devs should have installed jq anyway for other purposes.
At least they are expected to have knowledge on how to handle "jq: command not found" errors.

jandubois (approved):
Thanks, LGTM

jandubois:
> At least they are expected to have knowledge on how to handle "jq: command not found" errors.

Yeah, that's a good point!

PR #3067, opened 2024-12-30
Support for eBPF based port forwarding

balajiv113:
Improve port forwarding by using ebpf


AkihiroSuda:
Needs rebase

AkihiroSuda:
What's the current status?

balajiv113:
Am facing issues with kubernetes when I last worked. Not able to derive a common solution there as they are removing all entries and recreating iptables.

If we use kube api for kubernetes similarly we will have to do for others as well. Then this ebpf overall will be useful only for guest agent binds

AkihiroSuda:
Duplicating the comments about the current state of the PR
- - -
> @jandubois & @AkihiroSuda 
> 
> On the eBPF PR this is the current state,
> - guest direct ports are working perfectly without issues
> - docker/ containers with nftables was able to get it working (need to see if it works for all cases)
> - But with respect to kubernetes I got no luck. As I previously mentioned kubernetes are removing all entries in iptable and reconstructing back due to this we couldn't identify missing one's properly.

_Originally posted by @balajiv113 in https://github.com/lima-vm/lima/issues/4021#issuecomment-3304170079_
- - -

> > * docker/ containers with nftables was able to get it working (need to see if it works for all cases)
> 
> Thanks, that sounds promising!
> 
> So what I remember about some distros not having the prerequisite kernel modules, or missing permissions, was not, or is no longer correct?
> 

_Originally posted by @jandubois in https://github.com/lima-vm/lima/issues/4021#issuecomment-3304181398_
            
- - -

> > So what I remember about some distros not having the prerequisite kernel modules, or missing permissions, was not, or is no longer correct?
> 
> I was able to solve it using kprobe. This is available in almost all distro I checked. 
> 
> Even our FT were passing except docker that's when it hit the kubernetes issue
> 

_Originally posted by @balajiv113 in https://github.com/lima-vm/lima/issues/4021#issuecomment-3304189001_

- - -

> The combination of  @balajiv113 's eBPF PR (https://github.com/lima-vm/lima/pull/3067) w/ the existing [kubernetesservice watcher](https://github.com/lima-vm/lima/blob/master/pkg/guestagent/kubernetesservice/kubernetesservice.go) seems the most robust and compatible solution for now?
> 
> I wish we could remove the dependency on Kubernetes client libraries, but it can be discussed separately in future.

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/issues/4021#issuecomment-3305132170_

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/Microsoft/go-winio v0.6.2
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/balajiv113/fd v0.0.0-20230330094840-143eec500f3e
+	github.com/balajiv113/trackport v0.0.0-20250102132841-d36c7896b3ee
```
AkihiroSuda, line 12: The binary blobs such as https://github.com/balajiv113/trackport/blob/main/pkg/internal/bpftracker/bpf_arm64_bpfel.o have to be removed

pkg/guestagent/guestagent_linux.go
```diff
@@ -195,24 +198,26 @@ func isEventEmpty(ev *api.Event) bool {
...
-			ch <- ev
-		}
-		select {
-		case <-ctx.Done():
-			return
-		case _, ok := <-tickerCh:
-			if !ok {
-				return
-			}
-			logrus.Debug("tick!")
```
AkihiroSuda, line 214: To simplify https://github.com/balajiv113/trackport/tree/main/pkg/internal/bpftracker , can we keep the existing `collectEvent()` for now, and use eBPF only for stimulating the ticker on kretprobe of `bind(2)` ?

go.mod
balajiv113: Those are actually needed it will be embedded in go logics

pkg/guestagent/guestagent_linux.go
balajiv113: Yes yes will do that.

I wanted to purely use eBPF because fentry and fexit will wait for the eBPF logic to complete.

So I wanted to verify if the UDP forwarding issue is resolved in that case

go.mod
jandubois: Can we compile them from source in the Lima repo?

balajiv113: Yes yes that's the ultimate goal after it's all good.

balajiv113: I will clean up stuff and verify it again and see if all still holds good.

AkihiroSuda: It might be even fine to just execute the `bpftrace` command to stimulate the ticker, if  the `go build` stuff is too complicated

https://github.com/lima-vm/lima/blob/7459f4587987ed014c372f17b82de1817feffa2e/cmd/lima-guestagent/daemon_linux.go#L57-L63

balajiv113: We can do that as a temp solution. 

If we use ebpf then i wanted to avoid parsing the proc file again. As this will add further delay only in forwarding. But identifying ip and ports just by using bpftrace its hard. we need multiple variations. That's where using celium helped to write down our logics

jandubois: > But identifying ip and ports just by using bpftrace its hard. we need multiple variations.

This sounds like it is not really a generic solution and requires custom rules for each container runtime.

I wonder how much latency parsing the proc file is adding; maybe it is not noticeable?

As long as we manage to close the port forwarding before a new container can reopen the port, we should be fine. I guess we can do some stress-testing to see how it works.

AkihiroSuda:
Alternative PR
- https://github.com/lima-vm/lima/pull/4066

PR #3072, opened 2025-01-01
Assemble templates using the new `base` setting

jandubois:
This PR creates initial support for multi-file templates, mostly based on the suggestions I gave in https://github.com/lima-vm/lima/discussions/2520#discussioncomment-10193570.

## Instance create vs. start stages

The basic idea behind "template embedding" is that when an instance is created, the `lima.yaml` that is stored in the instance directory will embed settings from additional
base template(s) and external scripts into a single composite document.

Once the instance is created, there are no longer external references and the instance can be started while offline, even if it was created from online template fragments.

## Merging default settings from base templates

Merging of templates is controlled by the new `base` property, which takes the name of
a base template, e.g.

```yaml
base: template://docker
```

It is expected that in most instances only a single base template is needed, but it is possible to provide a list:

```yaml
base:
- variables.yaml
- http://example.com/template.yaml
```

The base template can in turn have their own base templates. For example `my.yaml` instance may be based on `template://docker`, which in turn could be based on `template://ubuntu`.

Merging is performed by copying every property from the base template that does not yet exist in the main template. For list properties the lists are appended. More details and exceptions are documented below.

## Embedding of external provisioning scripts

Both provisioning scripts and probe scripts can now be located with the new `file` property specifying a template locator (details below). The file will be loaded and inlined under the `script` property. Therefore `file` and `script` are mutually exclusive.

```yaml
provision:
- mode: user
  file: my-script.sh
probes:
- file: template://docker-probe.sh
```

## Relative template locators

The values of the `base` and `file` properties are "template locators". They are the usual local filenames or `file://`, `https://`, or `template://`URLs.

In addition they can also be relative locators: names that don't have a URL scheme and don't start with a `/`. These names are resolved relative to the template in which they occur (similar to relative `href` properties in HTML).

```yaml
base: base.yaml
provision:
- file: script.sh
```

If this template is loaded as `templates/main.yaml`, then the referenced template files will be fetched from `templates/base.yaml` and `templates/script.sh`.

If the template is loaded as `template://main`, then the files will be `template://base.yaml` and `template://script.sh`.

This means a template using relative locators will always fetch dependencies from the same (relative) location regardless of how it was loaded itself.

To make this work `template://base.yaml` needs to be equivalent to `template://base`. And it must be possible to use other file extensions, so `template://script.sh` loads the script and doesn't look for `script.sh.yaml`.

## limactl template command

This PR adds new options to the `limactl template copy` subcommand.

* `--verbatim` copies a file _exactly_ as is.

* `--embed` will merge in all external base templates and shell scripts.

* `--embed-all` will also merge in templates and files referenced by `template://` URLs.

* `--fill` will apply `override.yaml`, `default.yaml`, and builtin defaults.

Without additional options the `copy` command will copy the file, but turn any relative locators into absolute locators. So the copy of the template will continue to reference the same base templates and scripts. Use the `--verbatim` option if you intend to copy those files locally as well.

## Detailed merging algorithm

Each template will recusively embed all dependencies in its base templates before merging them in.

Then each property, that does not exist in the main template will be copied over from the base template.

Properties that are lists are appended.

For `provision` and `probes` the order is reversed: the base entries are inserted **before** the main template entries, so that latter are executed last (see #2932 for discussion).

### Special rules

The `dns` entries are not merged. The entries from a base template are only copied when the main template list is empty.

`mountTypesSupported` will have any duplicate entries removed.

The `base` property is removed if the base template has been embedded.

For `provision` and `probes` the `file` property is removed, and the actual content of the script is stored in the usual `script` property.

For `minimumLimaVersion` and `vmOpts.qemy.minimumVersion` the values are compared. and the base version is only copied when it is greater than the main version.

### Combining list entries

Three list properties (`additionalDisks`, `mounts`, and `networks`) use combining logic based on a unique key (disk name, mount point, and interface name, respectively).

If a latter entry matches the key of an earlier entry, then any missing fields in the earlier entry are filled in from the latter entry, which is then deleted. This matches the logic already used for merging `override.yaml` and `default.yaml`.

There is a new wildcard feature now, where `*` will match any existing key, so can be used to provide defaults for all entries. This is mostly in preparation to use this mechanism for the builtin defaults.

For the combining list mechanism to work, **all** base templates must be merged (so all lists must contain the complete set of entries). Otherwise the key matching mechanism wouldn't work globally, but only for the parent template at each level.

### DIFF
--- a/cmd/limactl/template.go
+++ b/cmd/limactl/template.go
@@ -63,7 +63,8 @@ func newTemplateCopyCommand() *cobra.Command {
 		Args:    WrapArgsError(cobra.ExactArgs(2)),
 		RunE:    templateCopyAction,
 	}
-	templateCopyCommand.Flags().Bool("embed", false, "embed dependencies into template")
+	templateCopyCommand.Flags().Bool("embed", false, "embed external dependencies into template")
+	templateCopyCommand.Flags().Bool("embed-all", false, "embed all dependencies into template")
 	templateCopyCommand.Flags().Bool("fill", false, "fill defaults")
 	templateCopyCommand.Flags().Bool("verbatim", false, "don't make locators absolute")
 	return templateCopyCommand
@@ -71,6 +72,10 @@ func newTemplateCopyCommand() *cobra.Command {
 
 func templateCopyAction(cmd *cobra.Command, args []string) error {
 	embed, err := cmd.Flags().GetBool("embed")
+	if err != nil {
+		return err
+	}
+	embedAll, err := cmd.Flags().GetBool("embed-all")
 	if err != nil {
 		return err
 	}
@@ -82,13 +87,15 @@ func templateCopyAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	if fill {
+		embedAll = true
+	}
+	if embedAll {
+		embed = true
+	}
 	if embed && verbatim {
-		return errors.New("--embed and --verbatim cannot be used together")
+		return errors.New("--verbatim cannot be used with any of --embed, --embed-all, or --fill")
 	}
-	if fill && verbatim {
-		return errors.New("--fill and --verbatim cannot be used together")
-	}
-
 	tmpl, err := limatmpl.Read(cmd.Context(), "", args[0])
 	if err != nil {
 		return err
@@ -98,7 +105,8 @@ func templateCopyAction(cmd *cobra.Command, args []string) error {
 	}
 	if !verbatim {
 		if embed {
-			if err := tmpl.Embed(cmd.Context()); err != nil {
+			// Embed default base.yaml only when fill is true.
+			if err := tmpl.Embed(cmd.Context(), embedAll, fill); err != nil {
 				return err
 			}
 		} else {
--- a/cmd/limactl/template.go
+++ b/cmd/limactl/template.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -52,23 +53,73 @@ var templateCopyExample = `  Template locators are local files, file://, https:/
 
 func newTemplateCopyCommand() *cobra.Command {
 	templateCopyCommand := &cobra.Command{
-		Use:     "copy TEMPLATE DEST",
+		Use:     "copy [OPTIONS] TEMPLATE DEST",
 		Short:   "Copy template",
 		Long:    "Copy a template via locator to a local file",
 		Example: templateCopyExample,
 		Args:    WrapArgsError(cobra.ExactArgs(2)),
 		RunE:    templateCopyAction,
 	}
+	templateCopyCommand.Flags().Bool("embed", false, "embed dependencies into template")
+	templateCopyCommand.Flags().Bool("fill", false, "fill defaults")
+	templateCopyCommand.Flags().Bool("verbatim", false, "don't make locators absolute")
 	return templateCopyCommand
 }
 
 func templateCopyAction(cmd *cobra.Command, args []string) error {
+	embed, err := cmd.Flags().GetBool("embed")
+	if err != nil {
+		return err
+	}
+	fill, err := cmd.Flags().GetBool("fill")
+	if err != nil {
+		return err
+	}
+	verbatim, err := cmd.Flags().GetBool("verbatim")
+	if err != nil {
+		return err
+	}
+	if embed && verbatim {
+		return errors.New("--embed and --verbatim cannot be used together")
+	}
+	if fill && verbatim {
+		return errors.New("--fill and --verbatim cannot be used together")
+	}
+
 	tmpl, err := limatmpl.Read(cmd.Context(), "", args[0])
 	if err != nil {
 		return err
 	}
 	if len(tmpl.Bytes) == 0 {
 		return fmt.Errorf("don't know how to interpret %q as a template locator", args[0])
+	}
+	if !verbatim {
+		if embed {
+			if err := tmpl.Embed(cmd.Context()); err != nil {
+				return err
+			}
+		} else {
+			if err := tmpl.UseAbsLocators(); err != nil {
+				return err
+			}
+		}
+	}
+	if fill {
+		limaDir, err := dirnames.LimaDir()
+		if err != nil {
+			return err
+		}
+		// Load() will merge the template with override.yaml and default.yaml via FillDefaults().
+		// FillDefaults() needs the potential instance directory to validate host templates using {{.Dir}}.
+		filePath := filepath.Join(limaDir, tmpl.Name+".yaml")
+		tmpl.Config, err = limayaml.Load(tmpl.Bytes, filePath)
+		if err != nil {
+			return err
+		}
+		tmpl.Bytes, err = limayaml.Marshal(tmpl.Config, false)
+		if err != nil {
+			return err
+		}
 	}
 	writer := cmd.OutOrStdout()
 	target := args[1]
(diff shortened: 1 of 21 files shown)


jandubois:
This PR still needs some integration tests, but I wanted to create it before 2024 was over. 😄

pkg/limayaml/limayaml.go
```diff
@@ -7,6 +7,7 @@ import (
 )
 
 type LimaYAML struct {
+	BasedOn               BaseTemplates `yaml:"basedOn,omitempty" json:"basedOn,omitempty" jsonschema:"nullable"`
```
AkihiroSuda, line 10: basedOn → base, for simplicity?

AkihiroSuda: I still prefer this to be extensible to support digests

pkg/limayaml/limayaml.go
```diff
@@ -215,6 +218,7 @@ type Provision struct {
 	Mode                            ProvisionMode `yaml:"mode,omitempty" json:"mode,omitempty" jsonschema:"default=system"`
 	SkipDefaultDependencyResolution *bool         `yaml:"skipDefaultDependencyResolution,omitempty" json:"skipDefaultDependencyResolution,omitempty"`
 	Script                          string        `yaml:"script" json:"script"`
+	File                            *string       `yaml:"file,omitempty" json:"file,omitempty" jsonschema:"nullable"`
```
AkihiroSuda, line 221: Same, needs to support digest

pkg/limayaml/limayaml.go
AkihiroSuda: Also probably this should be marked as an experimental in v1.1.

jandubois: > basedOn → base, for simplicity?

I don't really care, but I thought `basedOn is more intuitive to somebody who hasn't read the docs: "This template is based on this other template". But maybe `base` confers the same meaning.

The other argument I have is that `basedOn` works for both single and multiple arguments:

```yaml
basedOn: "template://ubuntu"
basedOn: [variables.yaml, base.yaml]
```

Whereas `base: [variables.yaml, base.yaml]` looks slightly less right. However, `bases: base.yaml` is even worse.

Not a strong argument, and we already have various singular/plural issues in our property names already, so I will change this to `base`, even though I still think `basedOn` is better. 😉 

> I still prefer this to be extensible to support digests

We can always add this later, when (if) we actually need it. This PR already uses a custom unmarshaller to treat `basedOn: base.yaml` the same as `basedOn: [base.yaml]` (requested by @nirs on Slack, to optimize for the common use case).

It will be trivial to extend it to treat both as `basedOn: [{locator: base.yaml}]` if we need to.

I already tried to explain why a simple digest per file is not really going to work in https://github.com/lima-vm/lima/discussions/3019#discussioncomment-11569719. Let's keep the discussion over there; this should not be required for this PR.

> Also probably this should be marked as an experimental in v1.1.

Absolutely. This should remain as experimental until all parts of it have been implemented, and we have gained some experience actually using it, so that we can still change it in incompatible ways.

pkg/limayaml/limayaml.go
jandubois: Again, this can be added later when we need it.

AkihiroSuda: Yep, but at least we have to decide whether we will just use the existing file locator structure, or we will have to introduce a new microformatted string
- https://github.com/lima-vm/lima/discussions/3019

pkg/limayaml/limayaml.go
AkihiroSuda: ("compose:" might be an option, but on second thought probably we should just reserve it for the VM equivalent of Docker Compose)

AkihiroSuda: What about `inherit:`?

jandubois: I'm fine with `base` if you don't like `basedOn`.

I don't want to invoke the ghost of "multiple inheritance"... 😄

pkg/limayaml/limayaml.go
jandubois: Let's have the discussion over there for now. It is not urgent to get this PR merged; we can take our time to discuss it.

I just want to keep the discussion is a separate "discussion", so it doesn't get buried when this PR gets closed.

jandubois:
@AkihiroSuda I believe I've addressed all your code review feedback so far, as well as all the implications from our discussion about digest support (the schema has been updated, but the digest computation/checking is left for a future PR).

I've rebased this PR since you originally reviewed it, but any further changes are all in separate commits. I've not yet squashed them, in case it makes it easier for you to continue the review; but let me know if you prefer that I squash them now!

templates/default.yaml
```diff
@@ -279,6 +287,17 @@ containerd:
 # 🟢 Builtin default: not set
 minimumLimaVersion: null
 
+# EXPERIMENTAL
+# Default settings can be imported from base templates. These will be merged in when the instance
+# is created, and the combined template is stored in the instance directory.
+# This setting ca be either a single string (URL), or a list of locators.
+# A locator is again either a string (URL), or an object with "url" and "digest" properties, e.g.
+# base: [{url: ./base.yaml, digest: decafbad}, …]
+# The "digest" property is currently unused.
```
AkihiroSuda, line 296: Should print an error when the property is set?

jandubois: I can do that, but then I could have left-out the property all together; I only added it to show you how it will work.

I can implement it once this PR is merged, but I didn't want to include it because this PR is already way too big.

Anyways, let me know if you want to add an error to validation in this PR.

jandubois:
> I will briefly to turn this into Draft mode again because I think there is another condition in `useAbsLocators` that is not fully implemented, and I need to figure out how to add a test for it.

I think I have all the combinations covered now (famous last words alert), so marking it ready for review again.

And again, let me know if you prefer me to squash all the commits already!

AkihiroSuda (dismissed):
Thanks

@lima-vm/committers 

Can we merge this and release v1.1.0-alpha.0 ?

nirs:
Does this introduce any breaking change or it should be completely compatible with existing templates?

jandubois:
> Does this introduce any breaking change or it should be completely compatible with existing templates?

It should be completely backwards compatible.

> Can we merge this and release v1.1.0-alpha.0 ?

I have another commit that should go into the alpha, but I want to create a separate PR for it instead of piling on even more changes on this one.

It is based on a discussion I had with @nirs on Slack about supporting external data files, so you can e.g. write a YAML config file using automation, and don't have to wrap it into a shell script that you put into a provisioning script. The change allows you to do this:

```
provision:
- mode: data
  path: /etc/conf.d/example
  content: |
    VAR=VALUE
  overwrite: false
- mode: data
  path: "{{.Home}}/kubeadm.yaml"
  file: kubeadm.yaml
  owner: "{{.User}}"
  permissions: 0700
```

The `file` property works the same way it works for provision scripts and probes.

It is pretty similar to the `write_files` feature in cloud-init. I can write some more about it when I open the PR.

cmd/limactl/template.go
```diff
@@ -23,6 +24,9 @@ func newTemplateCommand() *cobra.Command {
 		// The template command is still hidden because the subcommands and options are still under development
 		// and subject to change at any time.
 		Hidden: true,
+		PreRun: func(*cobra.Command, []string) {
+			logrus.Warn("`limactl template` is experimental")
```
AkihiroSuda, line 28: Could you also update https://lima-vm.io/docs/releases/experimental/

jandubois: Documented in f74e152.

Is it time to squash all commits now?

jandubois: > Documented in [f74e152](https://github.com/lima-vm/lima/commit/f74e152799f3112c7e1820b0bf1b8f1a82add529).

Oops, I mean f3c78d44dd73a0883d4ab5f0199f60121d8afa8d (I changed `limactl tmpl` to `limactl template`).

jandubois:
I had to rebase the PR to resolve the merge conflict with recent changes to `pkg/limayaml/marshal.go`. There are no other changes.

AkihiroSuda:
Let's merge, after another rebase to add the licensing boilerplate

jandubois:
> Let's merge, after another rebase to add the licensing boilerplate

Rebased and copyright headers added...

AkihiroSuda (approved):
Thanks

PR #3074, opened 2025-01-01
Filter out stderr message from older ssh versions

### LINKED ISSUE #3073: Extra line printed by lima command after change in 1.0.3
### Description

When running lima on a system with an older OpenSSH, an extra line is now printed by `lima` (but **not** by `ssh`):

```console
$ lima uname
Linux
Shared connection to 127.0.0.1 closed.
$ ssh lima-default uname
Linux
```

This is because the change in LogLevel, where old versions (8.8 and earlier) needed QUIET - but now get ERROR:

commit ab48fb7ad11c4d9b3bccc2a12f8936efa4a1348f

That causes a message to be printed on stderr (breaking the spell), when the `-t` flag is passed to the ssh command.

```console
$ ssh -t lima-default uname
Linux
Shared connection to 127.0.0.1 closed.
$ ssh -t -o LogLevel=QUIET lima-default uname
Linux
```

It has been fixed upstream in more recent version (8.9+), now you need LogLevel to be at least INFO to get the log:

https://github.com/openssh/openssh-portable/commit/a882a09722c9f086c9edb65d0c4022fd965ec1ed

afbjorklund:
Closes #3073

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"bufio"
+	"bytes"
 	"errors"
 	"fmt"
 	"os"
@@ -9,6 +11,7 @@ import (
 	"strings"
 
 	"al.essio.dev/pkg/shellescape"
+	"github.com/coreos/go-semver/semver"
 	"github.com/lima-vm/lima/pkg/sshutil"
 	"github.com/lima-vm/lima/pkg/store"
 	"github.com/mattn/go-isatty"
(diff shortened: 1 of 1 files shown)


AkihiroSuda (changes requested):
Can we rather revert the log level for old ssh?

afbjorklund:
> Can we rather revert the log level for old ssh?

Wouldn't you miss out on the error messages then?

afbjorklund:
There is a problem with missing newlines, from the scanner.

PR #3078, opened 2025-01-02
Fix internal links on the website

afbjorklund:
Currently using a static file in the git repository, but they can link directly to generated site instead.

----

The links on the start page of documentation has an extra step in them:

https://lima-vm.io/docs/

https://github.com/lima-vm/lima/blob/master/docs/multi-arch.md

https://lima-vm.io/docs/config/multi-arch/

Seems to be a left-over from when it was moved from the repo to the site...

commit c9a53ae3e8cc712b73dbe2c55ca6fce0aaa43e2b


AkihiroSuda (approved):
Thanks

PR #3083, opened 2025-01-05
Use sync.OnceValue to implement hasSMEDarwin()

jandubois:


jandubois:
@AkihiroSuda I don't understand why the original implementation used both an `init` function and `sync.Once`. Doesn't the `init` function already guarantee that the code is only executed once?

Using `sync.OnceValue()` is still slightly better because it only executes the code if it is needed at all.

AkihiroSuda (approved):
Thanks

PR #3086, opened 2025-01-05
Recompress the containerd archive with zstd

afbjorklund:
Takes a couple of seconds to recompress (when creating the `cidata.iso`)...

But cuts the runtime down from 4-5 seconds down to less than 1 second

Choosing the default (fast) compression, so the actual file size is the same.

If the nerdctl project supported .tar.zst too, that time could be avoided?


AkihiroSuda (changes requested):
I don't think we should introduce another dependency just for saving a few seconds

AkihiroSuda:
Maybe we can just put an uncompressed tar to cidata.iso?

afbjorklund:
Or just have a better version check, if that is the thing we actually run twice?

afbjorklund:
We could keep the refactoring, if wanting to revisit this in the future:

* https://github.com/lima-vm/lima/pull/3087

PR #3087, opened 2025-01-05
Refactor the containerd archive compression

afbjorklund:
Have tar auto-detect the compression, and pass the filename.

Instead of hardcoding gzip and .tgz in the archive handling.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
@@ -14,7 +14,7 @@ command -v systemctl >/dev/null 2>&1 || exit 0
 # Extract bin/nerdctl and compare whether it is newer than the current /usr/local/bin/nerdctl (if already exists).
 # Takes 4-5 seconds. (FIXME: optimize)
 tmp_extract_nerdctl="$(mktemp -d)"
-tar Cxzf "${tmp_extract_nerdctl}" "${LIMA_CIDATA_MNT}"/nerdctl-full.tgz bin/nerdctl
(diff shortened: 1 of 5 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot/40-install-containerd.sh
```diff
@@ -28,7 +28,7 @@ if [ ! -f "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}"/bin/nerdctl ] || [[ "${tmp_extra
 			sudo -iu "${LIMA_CIDATA_USER}" "XDG_RUNTIME_DIR=/run/user/${LIMA_CIDATA_UID}" "PATH=${PATH}" containerd-rootless-setuptool.sh uninstall
 		)
 	fi
-	tar Cxzf "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}" "${LIMA_CIDATA_MNT}"/nerdctl-full.tgz
+	tar Cxaf "${LIMA_CIDATA_GUEST_INSTALL_PREFIX}" "${LIMA_CIDATA_MNT}"/"${LIMA_CIDATA_CONTAINERD_ARCHIVE}"
```
AkihiroSuda, line 31: How is "a" old?
Does this work with EL8, Ubuntu 20.04, etc.?

AkihiroSuda: Found: It is from 2007
https://git.savannah.gnu.org/cgit/tar.git/commit/?h=release_1_20&id=620a136e74f211c2fb98b45fe44261d5d775f6a9

AkihiroSuda (approved):
Thanks

PR #1045, opened 2022-09-08
yaml: deprecate non-strict YAML

AkihiroSuda:
Having duplicated keys is now prohibited, but having unsupported keys is still allowed (with deprecation warning)

Fix #1044 

```console
$ cat correct.yaml 
images:
- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"

networks:
- lima: shared
- lima: bridged

$ limactl validate correct.yaml 
INFO[0000] "correct.yaml": OK 
```

```console
$ cat duplicated-key.yaml 
images:
- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"

networks:
- lima: shared
  lima: bridged

$ limactl validate duplicated-key.yaml 
FATA[0000] failed to load YAML file "duplicated-key.yaml": failed to unmarshal YAML (main file "/tmp/foo/duplicated-key.yaml"): [6:3] duplicate key "lima"
       4 | networks:
       5 | - lima: shared
    >  6 |   lima: bridged
             ^ 
```

```console
$ cat unknown-key.yaml 
images:
- location: "https://cloud-images.ubuntu.com/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img"

networks:
- lima: shared
lima: bridged

$ limactl validate unknown-key.yaml 
WARN[0000] Non-strict YAML is deprecated and will be unsupported in a future version of Lima  comment="main file \"/tmp/foo/unknown-key.yaml\"" error="[6:1] unknown field \"lima\"\n       4 | networks:\n       5 | - lima: shared\n    >  6 | lima: bridged\n           ^"
INFO[0000] "unknown-key.yaml": OK
```

PR #3104, opened 2025-01-09
Un-deprecate non-strict YAML

jandubois:
We still warn the user because it could be due to typos, but we no longer threaten to make strict mode the default.

Allowing non-strict YAML is very useful for maintainers switching between feature branches. Failing on non-strict YAML would make it hard to e.g. delete instances that have been created by a different branch that introduces a new field.

### DIFF
--- a/pkg/limayaml/marshal.go
+++ b/pkg/limayaml/marshal.go
@@ -45,9 +45,7 @@ func Unmarshal(data []byte, v interface{}, comment string) error {
 		return fmt.Errorf("failed to unmarshal YAML (%s): %w", comment, err)
 	}
 	if err := yaml.UnmarshalWithOptions(data, v, yaml.Strict(), yaml.CustomUnmarshaler[Disk](unmarshalDisk)); err != nil {
-		logrus.WithField("comment", comment).WithError(err).Warn("Non-strict YAML is deprecated and will be unsupported in a future version of Lima")
-		// Non-strict YAML is known to be used by Rancher Desktop:
-		// https://github.com/rancher-sandbox/rancher-desktop/blob/c7ea7508a0191634adf16f4675f64c73198e8d37/src/backend/lima.ts#L114-L117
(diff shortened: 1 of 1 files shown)


AkihiroSuda (dismissed):
Thanks, but this might be considered as a breaking change

jandubois:
> Thanks, but this might be considered as a breaking change

It is, but it has been deprecated for 3 years now (since 0.12.0), and issuing a warning for 3 years: #1045.

So yes, it should be mentioned in the release notes. But if we never want to remove it, then we should maybe get rid of the deprecation warning and just print the non-strict YAML errors.

afbjorklund:
For me it is quite normal to get warnings, when switching between alternative futures and the release version.

```
WARN[0000] Non-strict YAML is deprecated and will be unsupported in a future version of Lima  comment="main file \"/home/anders/.lima/raspberrypi/lima.yaml\"" error="[14:3] unknown field \"address\"\n  11 |   writable: true\n  12 | \n  13 | ssh:\n> 14 |   address: raspberrypi.local\n         ^\n  16 | provision:\n  17 | - mode: system\n  18 |   "
WARN[0000] Unknown driver: ext                          
WARN[0000] instance "raspberrypi" has errors             errors="[field `vmType` must be \"qemu\", \"vz\", \"wsl2\"; got \"ext\"]"
```

If these are changed from warnings to errors, I would need to use a separate LIMA_HOME for experimenting...


Which is "OK", I suppose. Can't say that it hasn't warned :-)

_... will be unsupported in a future version ..._

jandubois:
> If these are changed from warnings to errors, I would need to use a separate LIMA_HOME for experimenting...

Yeah, I was thinking the same thing this morning; it is an inconvenience for maintainers.

So let's just remove the deprecation message, but keep it as a warning only.

AkihiroSuda (approved):
Thanks

Issue #2575, opened 2024-09-01
`limactl create --name=xyz` fails if instance xyz already exists but `limactl create --name=xyz template://default` opens a configuration prompt


MayCXC:
### Description

I expected `limactl create` to exit 1 immediately if the instance exists, whether or not a template file is provided.

PR #3120, opened 2025-01-16
fix inconsistent error in limactl create command when instance name and template is specified

olamilekan000:
```
Change fixes inconsistency in the limactl create command when attempting to create an 
instance with an already existing name, when a template is specified and when it isn't.
```
Fixes https://github.com/lima-vm/lima/issues/2575

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -164,28 +164,6 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string, createOnly bool) (*
 		if err := identifiers.Validate(tmpl.Name); err != nil {
 			return nil, fmt.Errorf("argument must be either an instance name, a YAML file path, or a URL, got %q: %w", tmpl.Name, err)
 		}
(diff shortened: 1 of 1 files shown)


olamilekan000:
@AkihiroSuda can you kindly review? Thanks

cmd/limactl/start.go
```diff
@@ -149,6 +149,7 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string, createOnly bool) (*
 	if err != nil {
 		return nil, err
 	}
+
```
AkihiroSuda, line 152: unneeded

AkihiroSuda: Same for other whitespace changes too

olamilekan000: done.

AkihiroSuda (approved):
Thanks

jandubois (approved):
Thanks, LGTM

PR #3123, opened 2025-01-18
cidata: Use mac address as dhcpd identifier

nirs:
On macOS the DHCP server use the DUID+IAID for both hw_address and identifier, which makes it harder to discover the machine mac address by looking up leases in /var/db/dhcpd_leases. It seems that this issue started in macOS 15.

    % cat /var/db/dhcpd_leases
    {
            name=lima-test
            ip_address=192.168.105.4
            hw_address=ff,5c:bb:eb:5e:0:2:0:0:ab:11:72:3e:92:23:14:1d:a7:ba
            identifier=ff,5c:bb:eb:5e:0:2:0:0:ab:11:72:3e:92:23:14:1d:a7:ba
            lease=0x678b8aee
    }

With this change the record is created using the mac address:

    % cat /var/db/dhcpd_leases
    {
            name=lima-test
            ip_address=192.168.105.4
            hw_address=1,52:55:55:ba:82:85
            identifier=1,52:55:55:ba:82:85
            lease=0x678b90c6
    }

The fix is based on:
https://github.com/cirruslabs/linux-image-templates/pull/39.

Minikube have the similar (networkd specific) fix: https://github.com/kubernetes/minikube/pull/3572

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/network-config
+++ b/pkg/cidata/cidata.TEMPLATE.d/network-config
@@ -8,6 +8,7 @@ ethernets:
     set-name: {{$nw.Interface}}
     dhcp4-overrides:
       route-metric: {{$nw.Metric}}
+    dhcp-identifier: mac
     {{- if and (eq $nw.Interface $.SlirpNICName) (gt (len $.DNSAddresses) 0) }}
     nameservers:
       addresses:


nirs:
The option is documented here:
https://netplan.readthedocs.io/en/latest/netplan-yaml/

> dhcp-identifier (scalar)
> (networkd back end only) Sets the source of DHCP (v4) client identifier. If mac is specified, the MAC address of the link is used. If this option is omitted, or if duid is specified, networkd will generate an RFC4361-compliant client identifier for the interface by combining the link’s IAID and DUID.

The solution is mentioned as a fixf for Windows based DHCP server:
https://netplan.readthedocs.io/en/latest/examples/#how-to-integrate-with-windows-dhcp-server

nirs:
Failed test seems unrelated:

time="2025-01-18T11:12:48Z" level=error msg="[failed to satisfy the optional requirement 1 of 2 \"systemd must be available\": systemd is required to run containerd, but does not seem to be available.\nMake sure that you use an image that supports systemd. If you do not want to run\ncontainerd, please make sure that both 'container.system' and 'containerd.user'\nare set to 'false' in the config file.\n; skipping further checks: stdout=\"\", stderr=\"\": failed to execute script \"systemd must be available\": stdout=\"\", stderr=\"\": fork/exec /usr/bin/ssh: bad file descriptor]"

time="2025-01-18T11:12:48Z" level=warning msg="DEGRADED. The VM seems running, but file sharing and port forwarding may not work. (hint: see \"/Users/runner/.lima/fedora/ha.stderr.log\")"
time="2025-01-18T11:12:48Z" level=fatal msg="degraded, status={Running:true Degraded:true Exiting:false Errors:[failed to satisfy the optional requirement 1 of 2 \"systemd must be available\": systemd is required to run containerd, but does not seem to be available.\nMake sure that you use an image that supports systemd. If you do not want to run\ncontainerd, please make sure that both 'container.system' and 'containerd.user'\nare set to 'false' in the config file.\n; skipping further checks: stdout=\"\", stderr=\"\": failed to execute script \"systemd must be available\": stdout=\"\", stderr=\"\": fork/exec /usr/bin/ssh: bad file descriptor] SSHLocalPort:49369}"

+ ERROR 'Failed to start "fedora"'

afbjorklund:
Here is the breakage:

* https://github.com/crc-org/vfkit/issues/242

AkihiroSuda (approved):
Thanks

Issue #3126, opened 2025-01-19
nestedVirtualization, rosetta, and mountType fields in the Lima config file are not Validated on Start Command


olamilekan000:
### Description

There are no validation for `nestedVirtualization`, `rosetta`, and `mountType` in `limayaml.Validate` function when a user tries to create start a VM.

PR #3127, opened 2025-01-19
add validation for nestedVirtualization, rosetta, and mountType

olamilekan000:
```
change validates config values for `nestedVirtualization`, `rosetta`, and `mountType` 
when a user starts or tries to create a VM.
```

Fixes https://github.com/lima-vm/lima/issues/3126


olamilekan000:
@AkihiroSuda kindly review. Also, I am not sure why unit tests are failing for windows, go version 1.22 and 1.23, even though everything looks fine.

AkihiroSuda:
https://github.com/lima-vm/lima/actions/runs/12853756300/job/35837296730?pr=3127

```
=== RUN   TestValidateRosetta
time="2025-01-19T12:59:05Z" level=debug msg="ResolveVMType: resolved VMType \"qemu\" (explicitly specified in []*LimaYAML{o,y,d}[1])"
    validate_test.go:202: assertion failed: expected an error, got nil
--- FAIL: TestValidateRosetta (0.00s)
```

```
=== RUN   TestValidateMountTypeOS
time="2025-01-19T12:59:05Z" level=debug msg="ResolveVMType: resolved VMType \"qemu\" (default for GOOS=\"linux\")"
time="2025-01-19T12:59:05Z" level=debug msg="ResolveVMType: resolved VMType \"qemu\" (default for GOOS=\"linux\")"
time="2025-01-19T12:59:05Z" level=warning msg="`mountType: virtiofs` on Linux is experimental"
    validate_test.go:272: assertion failed: expected an error, got nil
--- FAIL: TestValidateMountTypeOS (0.01s)
```

olamilekan000:
I have tried to replicate it locally but it passes. Not sure what I'm missing

olamilekan000:
@AkihiroSuda Found and fixed the issue. Thanks

jandubois:
@olamilekan000 I don't know if all your pushes where necessary to deal with CI problems, but please be aware that each full CI run costs about $6 due to our use of large GitHub runners for macOS, and this PR had like 18 force-pushes[^1]. That's ok if the problem is only reproducible in CI, but otherwise please try to resolve problems locally first.

[^1]: I have not checked, maybe the macOS jobs where skipped due to earlier compile/lint failures, in which case the cost will have been minimal.

pkg/limayaml/validate.go
```diff
@@ -372,6 +372,45 @@ func Validate(y *LimaYAML, warn bool) error {
 		}
 	}
 
+	if y.Rosetta.Enabled != nil {
+		if *y.Rosetta.Enabled && *y.VMType != VZ {
+			return fmt.Errorf("field `rosetta.enabled` can only be enabled for VMType %q; got %q", VZ, *y.VMType)
+		}
+	}
```
jandubois, line 379: Not sure why you use a nested condition, especially since you group the conditions differently in the next block. Why not just

```suggestion
	if y.Rosetta.Enabled != nil && *y.Rosetta.Enabled && *y.VMType != VZ {
		return fmt.Errorf("field `rosetta.enabled` can only be enabled for VMType %q; got %q", VZ, *y.VMType)
	}
```

pkg/limayaml/validate.go
```diff
@@ -372,6 +372,45 @@ func Validate(y *LimaYAML, warn bool) error {
...
+
+	if y.NestedVirtualization != nil && *y.NestedVirtualization {
+		if *y.VMType != VZ {
+			return fmt.Errorf("field `nestedVirtualization` can only be enabled for VMType %q; got %q", VZ, *y.VMType)
+		}
+	}
```
jandubois, line 385:
```suggestion
	if y.NestedVirtualization != nil && *y.NestedVirtualization && *y.VMType != VZ {
		return fmt.Errorf("field `nestedVirtualization` can only be enabled for VMType %q; got %q", VZ, *y.VMType)
	}
```

pkg/limayaml/validate.go
```diff
@@ -372,6 +372,45 @@ func Validate(y *LimaYAML, warn bool) error {
...
+
+	if y.MountType != nil {
+		if err := validateMountType(y); err != nil {
+			return err
+		}
+	}
```
jandubois, line 391: `y.MountType` cannot be `nil` after `FillDefaults()` has been called. Same way you didn't need to check if `y.VMType` is `nil`.

```suggestion
	if err := validateMountType(y); err != nil {
		return err
	}
```

pkg/limayaml/validate.go
```diff
@@ -372,6 +372,45 @@ func Validate(y *LimaYAML, warn bool) error {
...
+func validateMountType(y *LimaYAML) error {
+	validMountTypes := map[string]bool{
+		REVSSHFS: true,
+		NINEP:    true,
+		VIRTIOFS: true,
+		WSLMount: true,
+	}
+
+	if !validMountTypes[*y.MountType] {
+		return fmt.Errorf("field `mountType` %s, valid options are: %s, %s, %s, %s", *y.MountType, REVSSHFS, NINEP, VIRTIOFS, WSLMount)
```
jandubois, line 405: Please use `%q` consistently to quote embedded actual values.

Also mirror the text of other similar errors, e.g. for `VMType`.

```suggestion
		return fmt.Errorf("field `mountType` must be: %q, %q, %q, %q; got %q", REVSSHFS, NINEP, VIRTIOFS, WSLMount, *y.MountType)
```

pkg/limayaml/validate.go
```diff
@@ -372,6 +372,45 @@ func Validate(y *LimaYAML, warn bool) error {
...
+	}
+
+	if *y.MountType == VIRTIOFS && runtime.GOOS == "darwin" && IsNativeArch(AARCH64) {
+		if y.VMType == nil || *y.VMType != VZ {
+			return fmt.Errorf("field `mountType` requires vmType 'vz' on macOS (darwin); got %s", *y.VMType)
+		}
+	}
```
jandubois, line 412: Why are you checking for `AARCH64`? The requirement for `VZ` also exists for Intel.

And `y.VMType` cannot be `nil` here.

```suggestion
	if *y.MountType == VIRTIOFS && runtime.GOOS == "darwin" && *y.VMType != VZ {
			return fmt.Errorf("field `mountType` %q on macOS requires vmType %q; got %q", *y.MountType, VZ, *y.VMType)
		}
	}
```

pkg/limayaml/validate.go
```diff
@@ -432,6 +471,9 @@ func validateNetwork(y *LimaYAML) error {
 				return fmt.Errorf("field `%s.macAddress` must be a 48 bit (6 bytes) MAC address; actual length of %q is %d bytes", field, nw.MACAddress, len(hw))
 			}
 		}
+		if nw.VZNAT != nil && *nw.VZNAT && *y.VMType != VZ {
+			return fmt.Errorf("field `%s.vzNAT` needs vmType set to vz; got %s", field, *y.VMType)
```
jandubois, line 475:
```suggestion
			return fmt.Errorf("field `%s.vzNAT` needs vmType set to %q; got %q", field, VZ, *y.VMType)
```

jandubois (changes requested):
I've only reviewed the `validate` changes, not the tests.

There may be more conditions that the validation could check for, but I'm out of time for now to think about it more.

pkg/limayaml/validate.go
olamilekan000: was when I was issues with the tests. I was trying to figure out what was wrong

olamilekan000:
> @olamilekan000 I don't know if all your pushes where necessary to deal with CI problems, but please be aware that each full CI run costs about $6 due to our use of large GitHub runners for macOS, and this PR had like 18 force-pushes[1](#user-content-fn-1-6110d88fd5097f5b14faea5fae050c36). That's ok if the problem is only reproducible in CI, but otherwise please try to resolve problems locally first.
> 
> ## Footnotes
> 1. I have not checked, maybe the macOS jobs where skipped due to earlier compile/lint failures, in which case the cost will have been minimal. [↩](#user-content-fnref-1-6110d88fd5097f5b14faea5fae050c36)

Thanks for bringing this to my attention. The reason for that many pushes was due to me trying to figure why the test passes locally but fails with the CI. Apologies for the inconvenience

jandubois (changes requested):
There are still invalid setting combinations that are not covered by tests.

It is a bit unfortunate that complete test coverage can only be achieved by running the test on each supported os and arch.

I think it would be better if the validation code had global variables for os and arch that can then be overridden in the tests to check for all scenarios with just a single local test run. But that should be a separate PR.

PR #3128, opened 2025-01-19
Cache the OpenSSH version if it is needed again

afbjorklund:
Just to avoid calling it twice:

```
DEBU[0000] OpenSSH version 8.2.1 detected               
DEBU[0000] OpenSSH version 8.2.1 detected   
```

Found some hard-coded instances of "ssh", and some duplicated code for $SSH (which should be LIMA_SSH, I guess)

* https://github.com/lima-vm/lima/pull/3129

Added some paranoid caching of "stat" info, instead of using just the path like /usr/bin/ssh to cache the version...

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -127,7 +127,7 @@ var sshInfo struct {
 //
 // The result always contains the IdentityFile option.
 // The result never contains the Port option.
-func CommonOpts(useDotSSH bool) ([]string, error) {
+func CommonOpts(sshPath string, useDotSSH bool) ([]string, error) {
 	configDir, err := dirnames.LimaConfigDir()
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


pkg/sshutil/sshutil.go
```diff
@@ -277,18 +278,40 @@ func ParseOpenSSHVersion(version []byte) *semver.Version {
...
-func DetectOpenSSHVersion() semver.Version {
+// sshExecutable beyond path also records size and mtime, in the case of ssh upgrades.
+type sshExecutable struct {
+	Path    string
+	Size    int64
+	ModTime time.Time
+}
+
+// sshVersion caches the parsed version of each ssh executable, if it is needed again.
+var sshVersions = map[sshExecutable]*semver.Version{}
```
AkihiroSuda, line 289: Probably needs a mutex

pkg/sshutil/sshutil.go
```diff
@@ -277,18 +278,40 @@ func ParseOpenSSHVersion(version []byte) *semver.Version {
 	return &semver.Version{}
 }
 
-func DetectOpenSSHVersion() semver.Version {
+// sshExecutable beyond path also records size and mtime, in the case of ssh upgrades.
+type sshExecutable struct {
+	Path    string
+	Size    int64
+	ModTime time.Time
+}
```
AkihiroSuda, line 286: Maybe over engineering? idk

AkihiroSuda (approved):
Thanks

Issue #3107, opened 2025-01-11
Support for toolbox containers


afbjorklund:
### Description

I was looking into support for "toolbox containers", both the original root variants and the more recent rootless one:

https://github.com/coreos/toolbox/tree/0.0.9 (systemd)

https://github.com/coreos/toolbox (switched to podman)

https://github.com/flatcar/toolbox (went back to docker)

https://containertoolbx.org/ (rewritten, from Bash to Go)

Unfortunately the more recent `toolbox` requires podman, which in turn _almost_ requires fedora (or get an old podman).

[There is support for Ubuntu distribution](https://debarshiray.wordpress.com/2024/01/20/toolbx-now-offers-built-in-support-for-arch-linux-and-ubuntu/) (and a `podman-toolbox` package), and there is support for Podman in Ubuntu.

But it is not possible to use another container engine such as systemd or nerdctl, since this is [**not** a goal](https://containertoolbx.org/goals/) of the project...

<details>
<pre>
anders@lima-default:~$ toolbox
Command 'toolbox' not found, but can be installed with:
apt install podman-toolbox
Please ask your administrator.
anders@lima-default:~$ sudo apt install podman-toolbox
Installing:                     
  podman-toolbox

Installing dependencies:
  aardvark-dns                     libabsl20230802          libjbig0                        libxcb-render0
  adwaita-icon-theme               libasound2-data          libjpeg-turbo8                  libxcb-shm0
  alsa-topology-conf               libasound2t64            libjpeg8                        libxcomposite1
  alsa-ucm-conf                    libatk-bridge2.0-0t64    liblcms2-2                      libxcursor1
  at-spi2-common                   libatk1.0-0t64           liblerc4                        libxdamage1
  at-spi2-core                     libatspi2.0-0t64         libmalcontent-0-0               libxfixes3
  bubblewrap                       libavahi-client3         libostree-1-1                   libxi6
  buildah                          libavahi-common-data     libpango-1.0-0                  libxinerama1
  catatonit                        libavahi-common3         libpangocairo-1.0-0             libxrandr2
  conmon                           libavahi-glib1           libpangoft2-1.0-0               libxrender1
  containernetworking-plugins      libcairo-gobject2        libpipewire-0.3-0t64            libxtst6
  crun                             libcairo2                libpipewire-0.3-common          libyajl2
  dconf-gsettings-backend          libcolord2               libpixman-1-0                   netavark
[...]

PR #3130, opened 2025-01-19
Add basic root toolbox script for rootless nerdctl

afbjorklund:
Version of coreos toolbox, using nerdctl instead of systemd.

Uses fedora by default for the toolbox, it can be configured.

Added for a comparison with the `systemd-nspawn` version:

https://github.com/afbjorklund/systemd-toolbox (from coreos)

```console
$ lima ./toolbox 
░ Spawning container anders-fedora-latest on /var/lib/toolbox/anders-fedora-latest.
░ Press Ctrl-] three times within 1s to kill container.
[root@lima-default ~]# 
logout
Container anders-fedora-latest exited successfully.
```

```console
$ ./cmd/toolbox.lima 
░ Entering container anders-fedora-latest on fedora:latest.
[root@lima-default /]# 
exit
```

Issue #3107

Note: these are just small shell scripts, and not real programs.

----

The default image is read-write, so it doesn't really _need_ a toolbox.

When running a read-only OS, you need one to install your tools in...

But when running `lima` you could just as well install them on the VM.
This allows you to run `toolbox.lima` and leave the system untouched.

If it would be needed for system maintenance, it needs `sudo nerdctl`.
Currently it is root on the inside, but still user on the outside (instance)

Running as the user with a home directory mount is better off as a program.

Something like toolbx or distrobox, but with support for `nerdctl` as well?

### DIFF
--- a/cmd/toolbox.lima
+++ b/cmd/toolbox.lima
@@ -1,0 +1,100 @@
+#!/bin/bash
+
+set -e
+set -o pipefail
+
+hostname=$(hostname)
+
+case ${hostname} in
+	lima-*)
+		# already in lima
+		;;
+	*)
+		exec lima "$0" "$@"
+		;;
+esac
+
+machine=$(uname -m)
+
+case ${machine} in
+	aarch64)
+		PLATFORM=linux/arm64
+		;;
+	armv7l)
+		PLATFORM=linux/arm/v7
+		;;
+	riscv64)
+		PLATFORM=linux/riscv64
+		;;
+	x86_64)
+		PLATFORM=linux/amd64
+		;;
+	*)
+		echo "Warning: Unknown machine type ${machine}" >&2
+		exit 1
+		;;
+esac
+
+TOOLBOX_DOCKER_IMAGE=fedora
+TOOLBOX_DOCKER_TAG=latest
+TOOLBOX_USER=root
+TOOLBOX_BIND="--volume=/:/media/root --volume=/usr:/media/root/usr --volume=/run:/media/root/run"
+TOOLBOX_FLAGS=""
+# Ex: "--env=KEY=VALUE"
+TOOLBOX_ENV=""
+
+toolboxrc="${HOME}"/.toolboxrc
+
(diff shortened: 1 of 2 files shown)


cmd/toolbox.lima
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
AkihiroSuda, line 1: This is quite different from other `*.lima` wrappers, as this doesn't even execute `limactl`

AkihiroSuda: I also don't like the name of the "toolbox", as it is likely to conflict with something else, e.g., https://android.googlesource.com/platform/system/core/+/froyo/toolbox/Android.mk

AkihiroSuda: https://github.com/containerd/nerdctl/tree/main/examples might be a better place?

afbjorklund: The script executes itself with `lima` (if needed), so it should use limactl (eventually)
As mention in the issue, the "toolbox" is somewhat overloaded with 3+ implementations

afbjorklund:
As mentioned above there is much _need_ for a `toolbox` in the default lima instance, since it is already read-write.
It could be useful if there is ever support for CoreOS, or if we make a smaller BeanOS without a package manager?

So it's fine to leave this as an example, similar to LXC/LXD*, and **not** include any scripts in the default installation.
\* the default image includes support for both lxd and snapd, even though it is not used or needed by nerdctl.lima

"LXD is installed in the default Ubuntu template, so there is no `lxd.yaml`"

It would need to use `sudo nerdctl` to match the `sudo systemd-nspawn`, but then it needs a `default-rootful.yaml`
`FATA[0000] cannot access containerd socket "/run/containerd/containerd.sock": no such file or directory`

While you can still run the toolbox just fine as rootless, you can't manipulate the host like with coreos toolbox.
Even though you have `/media/root` mounted and the container is privileged, you don't have the host access.

I will leave the issue open, for providing something  like WSL or distrobox

Issue #2198, opened 2024-02-09
`limactl copy` should use `rsync` instead of `scp` for copying a directory


AkihiroSuda:
### Description

`limactl copy` should use `rsync` (present on macOS by default) instead of `scp` for copying a directory.
`scp` might be still faster for small directories, but probably the overhead is subtle.

https://github.com/lima-vm/lima/blob/eebc6847fb8f5d55e489fdc2948c6acafc2d5436/cmd/limactl/copy.go#L47

afbjorklund:
You could make it optional, like we ended up doing for `docker-machine scp`?

```
Usage: docker-machine scp [OPTIONS] [arg...]

Copy files between machines

Description:
   Arguments are [[user@]machine:][path] [[user@]machine:][path].

Options:
   
   --recursive, -r	Copy files recursively (required to copy directories)
   --delta, -d		Reduce amount of data sent over network by sending only the differences (uses rsync)
   --quiet, -q		Disables the progress meter as well as warning and diagnostic messages from ssh
```

(my suggestion was `--rsync`, since it does has some rather tricky semantics)

ayushgml:
@AkihiroSuda Can I work on this?

afbjorklund:
I found that I had to add --progress, to make the rsync output more similar to scp default output:

https://github.com/lima-vm/lima/pull/2028/commits/a63be723430d03a1c644a85986d6cc97d5b83dd6

olamilekan000:
is this still open to be worked on? @AkihiroSuda

kairveeehh:
is this an open or closed issue?

jandubois:
> is this an open or closed issue?

It is an open issue. It looks like it is waiting for #3143, which is still work in progress.

PR #3143, opened 2025-01-23
add rsync flag option to copy files using rsync

olamilekan000:
```
change adds rsync as an option to the copy command
```
issue https://github.com/lima-vm/lima/issues/2198

### DIFF
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -4,6 +4,7 @@
 package main
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
@@ -26,10 +27,44 @@ const copyHelp = `Copy files between host and guest
 
 Prefix guest filenames with the instance name and a colon.
 
-Example: limactl copy default:/etc/os-release .
+Backends:
+  auto   - Automatically selects the best available backend (rsync preferred, falls back to scp)
+  rsync  - Uses rsync for faster transfers with resume capability (requires rsync on both host and guest)
+  scp    - Uses scp for reliable transfers (always available)
+
+Examples:
+  # Copy file from guest to host (auto backend)
+  limactl copy default:/etc/os-release .
+
+  # Copy file from host to guest with verbose output
+  limactl copy -v myfile.txt default:/tmp/
+
+  # Copy directory recursively using rsync backend
+  limactl copy --backend=rsync -r ./mydir default:/tmp/
+
+  # Copy using scp backend specifically
+  limactl copy --backend=scp default:/var/log/app.log ./logs/
+
+  # Copy multiple files
+  limactl copy file1.txt file2.txt default:/tmp/
 
 Not to be confused with 'limactl clone'.
 `
+
+type copyTool string
+
+const (
+	rsync copyTool = "rsync"
+	scp   copyTool = "scp"
+	auto  copyTool = "auto"
+)
+
+type copyPath struct {
+	instanceName string
+	path         string
+	isRemote     bool
+	instance     *limatype.Instance
+}
 
 func newCopyCommand() *cobra.Command {
 	copyCommand := &cobra.Command{
@@ -44,6 +79,7 @@ func newCopyCommand() *cobra.Command {
 
 	copyCommand.Flags().BoolP("recursive", "r", false, "Copy directories recursively")
 	copyCommand.Flags().BoolP("verbose", "v", false, "Enable verbose output")
+	copyCommand.Flags().String("backend", "auto", "Copy backend (scp|rsync|auto)")
 
 	return copyCommand
 }
@@ -60,13 +96,6 @@ func copyAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	arg0, err := exec.LookPath("scp")
-	if err != nil {
-		return err
-	}
-	instances := make(map[string]*limatype.Instance)
-	scpFlags := []string{}
-	scpArgs := []string{}
 	debug, err := cmd.Flags().GetBool("debug")
 	if err != nil {
 		return err
@@ -75,6 +104,180 @@ func copyAction(cmd *cobra.Command, args []string) error {
 	if debug {
 		verbose = true
 	}
+
+	copyPaths, err := parseCopyArgs(ctx, args)
+	if err != nil {
+		return err
+	}
+
+	backend, err := cmd.Flags().GetString("backend")
+	if err != nil {
+		return err
+	}
+
+	cpTool, toolPath, err := selectCopyTool(ctx, copyPaths, backend)
+	if err != nil {
+		return err
+	}
+
+	logrus.Debugf("using copy tool %q", toolPath)
+
+	var copyCmd *exec.Cmd
+	switch cpTool {
+	case scp:
+		copyCmd, err = scpCommand(ctx, toolPath, copyPaths, verbose, recursive)
+	case rsync:
+		copyCmd, err = rsyncCommand(ctx, toolPath, copyPaths, verbose, recursive)
+	default:
+		err = fmt.Errorf("invalid copy tool %q", cpTool)
+	}
+	if err != nil {
+		return err
+	}
+
+	copyCmd.Stdin = cmd.InOrStdin()
+	copyCmd.Stdout = cmd.OutOrStdout()
+	copyCmd.Stderr = cmd.ErrOrStderr()
+	logrus.Debugf("executing %v (may take a long time)", copyCmd)
+
+	// TODO: use syscall.Exec directly (results in losing tty?)
+	return copyCmd.Run()
+}
+
+func parseCopyArgs(ctx context.Context, args []string) ([]*copyPath, error) {
+	var copyPaths []*copyPath
+
+	for _, arg := range args {
+		cp := &copyPath{}
+
+		if runtime.GOOS == "windows" {
+			if filepath.IsAbs(arg) {
+				var err error
+				arg, err = ioutilx.WindowsSubsystemPath(ctx, arg)
+				if err != nil {
+					return nil, err
+				}
+			} else {
+				arg = filepath.ToSlash(arg)
+			}
+		}
+
+		parts := strings.SplitN(arg, ":", 2)
+		switch len(parts) {
+		case 1:
+			cp.path = arg
+			cp.isRemote = false
+		case 2:
+			cp.instanceName = parts[0]
+			cp.path = parts[1]
+			cp.isRemote = true
+
+			inst, err := store.Inspect(ctx, cp.instanceName)
+			if err != nil {
+				if errors.Is(err, os.ErrNotExist) {
+					return nil, fmt.Errorf("instance %q does not exist, run `limactl create %s` to create a new instance", cp.instanceName, cp.instanceName)
+				}
+				return nil, err
+			}
+			if inst.Status == limatype.StatusStopped {
+				return nil, fmt.Errorf("instance %q is stopped, run `limactl start %s` to start the instance", cp.instanceName, cp.instanceName)
+			}
+			cp.instance = inst
+		default:
+			return nil, fmt.Errorf("path %q contains multiple colons", arg)
+		}
+
+		copyPaths = append(copyPaths, cp)
+	}
+
+	return copyPaths, nil
+}
+
+func selectCopyTool(ctx context.Context, copyPaths []*copyPath, backend string) (copyTool, string, error) {
+	switch copyTool(backend) {
+	case scp:
+		scpPath, err := exec.LookPath("scp")
+		if err != nil {
+			return "", "", fmt.Errorf("scp not found on host: %w", err)
+		}
+		return scp, scpPath, nil
+	case rsync:
+		rsyncPath, err := exec.LookPath("rsync")
+		if err != nil {
+			return "", "", fmt.Errorf("rsync not found on host: %w", err)
+		}
+		if !rsyncAvailableOnGuests(ctx, copyPaths) {
+			return "", "", errors.New("rsync not available on guest(s)")
+		}
+		return rsync, rsyncPath, nil
+	case auto:
+		if rsyncPath, err := exec.LookPath("rsync"); err == nil {
+			if rsyncAvailableOnGuests(ctx, copyPaths) {
+				return rsync, rsyncPath, nil
+			}
+			logrus.Debugf("rsync not available on guest(s), falling back to scp")
+		} else {
+			logrus.Debugf("rsync not found on host, falling back to scp: %v", err)
+		}
+
+		scpPath, err := exec.LookPath("scp")
+		if err != nil {
+			return "", "", fmt.Errorf("neither rsync nor scp found on host: %w", err)
+		}
+		return scp, scpPath, nil
+	default:
+		return "", "", fmt.Errorf("invalid backend %q, must be one of: scp, rsync, auto", backend)
+	}
+}
+
+func rsyncAvailableOnGuests(ctx context.Context, copyPaths []*copyPath) bool {
+	instances := make(map[string]*limatype.Instance)
+
+	for _, cp := range copyPaths {
+		if cp.isRemote {
+			instances[cp.instanceName] = cp.instance
+		}
+	}
+
+	for instName, inst := range instances {
+		if !checkRsyncOnGuest(ctx, inst) {
+			logrus.Debugf("rsync not available on instance %q", instName)
+			return false
+		}
+	}
+
+	return true
+}
+
+func checkRsyncOnGuest(ctx context.Context, inst *limatype.Instance) bool {
+	sshExe, err := sshutil.NewSSHExe()
+	if err != nil {
+		logrus.Debugf("failed to create SSH executable: %v", err)
+		return false
+	}
+	sshOpts, err := sshutil.SSHOpts(ctx, sshExe, inst.Dir, *inst.Config.User.Name, false, false, false, false)
+	if err != nil {
+		logrus.Debugf("failed to get SSH options for rsync check: %v", err)
+		return false
+	}
+
+	sshArgs := sshutil.SSHArgsFromOpts(sshOpts)
+	checkCmd := exec.CommandContext(ctx, "ssh")
+	checkCmd.Args = append(checkCmd.Args, sshArgs...)
+	checkCmd.Args = append(checkCmd.Args,
+		"-p", fmt.Sprintf("%d", inst.SSHLocalPort),
+		fmt.Sprintf("%s@127.0.0.1", *inst.Config.User.Name),
+		"command -v rsync >/dev/null 2>&1",
+	)
+
+	err = checkCmd.Run()
+	return err == nil
+}
+
+func scpCommand(ctx context.Context, command string, copyPaths []*copyPath, verbose, recursive bool) (*exec.Cmd, error) {
+	instances := make(map[string]*limatype.Instance)
+	scpFlags := []string{}
+	scpArgs := []string{}
 
 	if verbose {
 		scpFlags = append(scpFlags, "-v")
@@ -85,53 +288,32 @@ func copyAction(cmd *cobra.Command, args []string) error {
 	if recursive {
 		scpFlags = append(scpFlags, "-r")
 	}
+
 	// this assumes that ssh and scp come from the same place, but scp has no -V
-	sshExe, err := sshutil.NewSSHExe()
+	sshExeForVersion, err := sshutil.NewSSHExe()
 	if err != nil {
-		return err
+		return nil, err
 	}
-	legacySSH := sshutil.DetectOpenSSHVersion(ctx, sshExe).LessThan(*semver.New("8.0.0"))
-	for _, arg := range args {
-		if runtime.GOOS == "windows" {
-			if filepath.IsAbs(arg) {
-				arg, err = ioutilx.WindowsSubsystemPath(ctx, arg)
-				if err != nil {
-					return err
-				}
-			} else {
-				arg = filepath.ToSlash(arg)
-			}
-		}
-		path := strings.Split(arg, ":")
-		switch len(path) {
-		case 1:
-			scpArgs = append(scpArgs, arg)
-		case 2:
-			instName := path[0]
-			inst, err := store.Inspect(ctx, instName)
-			if err != nil {
-				if errors.Is(err, os.ErrNotExist) {
-					return fmt.Errorf("instance %q does not exist, run `limactl create %s` to create a new instance", instName, instName)
-				}
-				return err
-			}
-			if inst.Status == limatype.StatusStopped {
-				return fmt.Errorf("instance %q is stopped, run `limactl start %s` to start the instance", instName, instName)
-			}
+	legacySSH := sshutil.DetectOpenSSHVersion(ctx, sshExeForVersion).LessThan(*semver.New("8.0.0"))
+
+	for _, cp := range copyPaths {
+		if cp.isRemote {
 			if legacySSH {
-				scpFlags = append(scpFlags, "-P", fmt.Sprintf("%d", inst.SSHLocalPort))
-				scpArgs = append(scpArgs, fmt.Sprintf("%s@127.0.0.1:%s", *inst.Config.User.Name, path[1]))
+				scpFlags = append(scpFlags, "-P", fmt.Sprintf("%d", cp.instance.SSHLocalPort))
+				scpArgs = append(scpArgs, fmt.Sprintf("%s@127.0.0.1:%s", *cp.instance.Config.User.Name, cp.path))
 			} else {
-				scpArgs = append(scpArgs, fmt.Sprintf("scp://%s@127.0.0.1:%d/%s", *inst.Config.User.Name, inst.SSHLocalPort, path[1]))
+				scpArgs = append(scpArgs, fmt.Sprintf("scp://%s@127.0.0.1:%d/%s", *cp.instance.Config.User.Name, cp.instance.SSHLocalPort, cp.path))
 			}
-			instances[instName] = inst
-		default:
-			return fmt.Errorf("path %q contains multiple colons", arg)
+			instances[cp.instanceName] = cp.instance
+		} else {
+			scpArgs = append(scpArgs, cp.path)
 		}
 	}
+
 	if legacySSH && len(instances) > 1 {
-		return errors.New("more than one (instance) host is involved in this command, this is only supported for openSSH v8.0 or higher")
+		return nil, errors.New("more than one (instance) host is involved in this command, this is only supported for openSSH v8.0 or higher")
 	}
+
 	scpFlags = append(scpFlags, "-3", "--")
 	scpArgs = append(scpFlags, scpArgs...)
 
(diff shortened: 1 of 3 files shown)


cmd/limactl/copy.go
```diff
@@ -49,10 +50,24 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+		arg0, err = exec.LookPath("rsync")
+		if err != nil {
+			return err
+		}
+	} else {
+		arg0, err = exec.LookPath("scp")
+		if err != nil {
+			return err
+		}
+	}
```
nirs, line 69: This duplicate the lookup code. We could do:

    tool := "scp"
    if useRsync {
        tool = "rsync"
    }

    arg0, err = exec.LookPath(tool)
    if err != nil {
        return err
	}

cmd/limactl/copy.go
```diff
@@ -128,13 +152,23 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+	var cmdArgs []string
 	sshArgs := sshutil.SSHArgsFromOpts(sshOpts)
+	if useRsync {
+		// for rsync, add the -e flag with SSH options
+		scpFlags = append(scpFlags, "-e", fmt.Sprintf("ssh %s", strings.Join(sshArgs, " ")))
+		cmdArgs = append(cmdArgs, append(scpFlags, scpArgs...)...)
+	} else {
+		// for scp, append SSH options and scpArgs
+		cmdArgs = append(cmdArgs, append(sshArgs, scpArgs...)...)
+	}
```
nirs, line 165: This change complicates the code by adding useRsync condition on every step. This is not maintainable.

A better way to do this is to add a function to create a scp command arguments, and a function to create rsync command arguments. Then use shared code to run the command.

cmd/limactl/copy.go
```diff
@@ -67,6 +82,9 @@ func copyAction(cmd *cobra.Command, args []string) error {
 
 	if verbose {
 		scpFlags = append(scpFlags, "-v")
+		if useRsync {
+			scpFlags = append(scpFlags, "--progress")
```
nirs, line 86: We cannot use variable `scpFlags` to keep rsync flags. This will confuse the next developer. Renaming the variable to toolFlags will avoid this.

cmd/limactl/copy.go
```diff
@@ -34,6 +34,7 @@ func newCopyCommand() *cobra.Command {
 
 	copyCommand.Flags().BoolP("recursive", "r", false, "copy directories recursively")
 	copyCommand.Flags().BoolP("verbose", "v", false, "enable verbose output")
+	copyCommand.Flags().BoolP("rsync", "", false, "use rsync for copying instead of scp")
```
nirs, line 37: This is not flexible enough. What if we want to another tool later? If we add an option, adding a "tool" or "backend" option will be better, with possible values "scp", "rsync".

nirs (commented):
Do we have rsync in the guest in all cases? Minimal vm images do not have rsync.

Also adding options is bad both for users and for testing. If it is better to copy with rsync, we should detect if rsync is available in the guest and host, and use it without adding new option. Otherwise fall back to scp.

If we always have rsync, better to use it instead of scp and minimize the testing matrix.

nirs:
Also, do we have tests for code modified in this PR? if not we should not change it before adding tests. The new rsync flow should also have tests.

olamilekan000:
> @olamilekan000 I've asked before to please test your changes locally instead of force-pushing an update every 10 minutes for 2 hours, triggering potentially costly CI runs all the time.

I'd appreciate it if you can help with steps to run locally. I did try but It didn't work as expected.

cmd/limactl/copy.go
olamilekan000: Yeah, that makes sense. I was only following the discussion in the issue. I thought there was a consensus to go with having rsyn as a flag.

cmd/limactl/copy.go
olamilekan000: the tool option will default to `rsync` right?

AkihiroSuda:
> If we always have rsync, better to use it instead of scp and minimize the testing matrix.

Yes, rsync is always available on macOS and most Linux distros.
At least it is easily installable.

cmd/limactl/copy.go
nirs: If we add an option rsync sounds like a better default. But this is trivial to change if needed.

AkihiroSuda: Let's just remove the flag and only support rsync, unless there is a use case where `scp` is preferable over `rsync`

nirs:
> > @olamilekan000 I've asked before to please test your changes locally instead of force-pushing an update every 10 minutes for 2 hours, triggering potentially costly CI runs all the time.
> 
> I'd appreciate it if you can help with steps to run locally. I did try but It didn't work as expected.

@olamilekan000, Can you describe how did you try to test and why it did not work well?

You can also join the #lima slack channel to get quicker help from other developers.

olamilekan000:
> > > @olamilekan000 I've asked before to please test your changes locally instead of force-pushing an update every 10 minutes for 2 hours, triggering potentially costly CI runs all the time.
> > 
> > 
> > I'd appreciate it if you can help with steps to run locally. I did try but It didn't work as expected.
> 
> @olamilekan000, Can you describe how did you try to test and why it did not work well?
> 
> You can also join the #lima slack channel to get quicker help from other developers.

@nirs I have joined and share my error logs on the channel.

cmd/limactl/copy.go
olamilekan000: `rsync` isn't available by default on Alpine, Arch Linux, and Debian (see [here](https://github.com/lima-vm/lima/actions/runs/12959909604) for pipeline logs). A potential fix for this is to ssh into the guest vm and check if `rsync` is installed. @jandubois suggested running a command like this within the guest VM:
 `rsync --rsh="ssh" remote_user@remote_host rsync --version `.

However, implementing this approach introduces a limitation: rsync won’t work in scenarios where the instance name is unavailable, as the instance name is needed to fetch the hostname using `store.Inspect(instName)`.

For example, this command would fail since it doesn’t provide the instance name:
```
limactl copy -r /etc/config /tmp
```
On the other hand, this would work because it includes the instance name:
```
limactl copy -r default:/etc/config /tmp
```
@nirs @AkihiroSuda @jandubois @afbjorklund   I’d like to hear your thoughts on this approach and any potential alternatives. Would love to know if there's a better way to handle this while ensuring rsync works reliably.

nirs: > `rsync` isn't available by default on Alpine, Arch Linux, and Debian (see [here](https://github.com/lima-vm/lima/actions/runs/12959909604) for pipeline logs). 

So to support these distros with required rsync, the templates will have to install rsync. Can you check what is the cost in time and space?

> A potential fix for this 

This is not a fix, but the way to deal with missing rsync.

> is to ssh into the guest vm and check if `rsync` is installed. @jandubois suggested running a command like this within the guest VM: `rsync --rsh="ssh" remote_user@remote_host rsync --version `.

We can also use `ssh user@vm-ip command -v rsync`. We already use ssh to preform the initial connection so it should be easier to use it to detect guest capabilities.

> However, implementing this approach introduces a limitation: rsync won’t work in scenarios where the instance name is unavailable, as the instance name is needed to fetch the hostname using `store.Inspect(instName)`.
> 
> For example, this command would fail since it doesn’t provide the instance name:
> 
> ```
> limactl copy -r /etc/config /tmp
> ```

The online help instructs to prefix the file name with the instance name. It cannot work without this regardless of the tool use to copy.

So the issue about missing instance name does not exist.

> On the other hand, this would work because it includes the instance name:
> 
> ```
> limactl copy -r default:/etc/config /tmp
> ```

cmd/limactl/copy.go
```diff
@@ -21,6 +21,13 @@ Prefix guest filenames with the instance name and a colon.
 Example: limactl copy default:/etc/os-release .
 `
 
+type copyTool string
+
+const (
+	Rsync copyTool = "rsync"
+	Scp   copyTool = "scp"
```
nirs, line 28: Why these are exported?

cmd/limactl/copy.go
```diff
@@ -49,13 +56,20 @@ func copyAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	arg0, err := exec.LookPath("scp")
-	if err != nil {
-		return err
+	defaultTool := Rsync
+	arg0, err := exec.LookPath(string(defaultTool))
+	if err != nil || !strings.HasSuffix(arg0, "rsync") {
```
nirs, line 61: Why check if hte suffix has "rsync"? exec.LookupPath() can succeed and return something else?

The docs do not mention any such issue:
https://pkg.go.dev/os/exec#LookPath

cmd/limactl/copy.go
```diff
@@ -49,13 +56,20 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+		defaultTool = Scp
```
nirs, line 62: Calling this defaultTool is wrong. This is the actual tool we use. Better to call it copyTool or just tool since we are in the context of the copy command.

cmd/limactl/copy.go
```diff
@@ -49,13 +56,20 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+			return err
+		}
 	}
+	logrus.Infof("using copy tool %q", arg0)
+
 	instances := make(map[string]*store.Instance)
-	scpFlags := []string{}
-	scpArgs := []string{}
+	copyToolFlags := []string{}
+	copyToolArgs := []string{}
```
nirs, line 72: This cannot work since rsync and scp have different flags and arguments. What we need is to add functions to create the copy command, one for rsync and one for scp. Something like:

var command *exec.Cmd
switch tool {
case Rsync:
   command = rsyncCommand(...)
case Scp:
   command = scpCommand(...)
default:
    panic(...)
}

// Run the command...

cmd/limactl/copy.go
```diff
@@ -128,14 +154,28 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
-	logrus.Debugf("executing scp (may take a long time): %+v", sshCmd.Args)
+	logrus.Debugf("executing %s (may take a long time): %+v", arg0, sshCmd.Args)
 
 	// TODO: use syscall.Exec directly (results in losing tty?)
 	return sshCmd.Run()
 }
+
+func isCopyToolRsync(copyTool copyTool) bool {
+	return copyTool == Rsync
+}
```
nirs, line 172: This helper is not helpful for 2 reasons:
- We should have exactly one check for the tool
- Even if we need checks it does not help

cmd/limactl/copy.go
```diff
@@ -128,14 +154,28 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+
+func createArgs(sshArgs, copyToolArgs []string, copyTool copyTool) []string {
+	if isCopyToolRsync(copyTool) {
+		rsyncFlags := []string{"-e", fmt.Sprintf("ssh %s", strings.Join(sshArgs, " "))}
+		return append(rsyncFlags, copyToolArgs...)
+	}
+
+	return append(sshArgs, copyToolArgs...)
+}
```
nirs, line 181: Not needed if we have functions to create each command.

cmd/limactl/copy.go
```diff
@@ -128,14 +154,28 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
 		}
 	}
+
 	sshArgs := sshutil.SSHArgsFromOpts(sshOpts)
 
-	sshCmd := exec.Command(arg0, append(sshArgs, scpArgs...)...)
+	sshCmd := exec.Command(arg0, createArgs(sshArgs, copyToolArgs, defaultTool)...)
 	sshCmd.Stdin = cmd.InOrStdin()
 	sshCmd.Stdout = cmd.OutOrStdout()
 	sshCmd.Stderr = cmd.ErrOrStderr()
```
nirs, line 163: We run rsync via ssh? it does not make sense.

Can you share the complete the scp and rsync commands that you want to run?

The commands can be tested from the shell without lima, using the instance ip. limactl just make this easier for the user by building the right command and running it transparently.

cmd/limactl/copy.go
```diff
@@ -128,14 +154,28 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
-	logrus.Debugf("executing scp (may take a long time): %+v", sshCmd.Args)
+	logrus.Debugf("executing %s (may take a long time): %+v", arg0, sshCmd.Args)
```
nirs, line 164: Here we can log the cmd instance instead of arg0 and rest fo the args.

It will be easier to make this change if we work first on preparing the code to add a rsync option. Improve the logs, variables names, and orgnization so adding a second option is easy in the next step.

cmd/limactl/copy.go
olamilekan000: > We can also use ssh user@vm-ip command -v rsync. We already use ssh to preform the initial connection so it should be easier to use it to detect guest capabilities.

@nirs correct me if i'm wrong, is this [where](https://github.com/lima-vm/lima/blob/master/pkg/hostagent/hostagent.go#L442) you're referring to?

cmd/limactl/copy.go
olamilekan000: Initially, I didn't realize that` exec.LookPath` was returning the path to rsync on the host machine rather than the guest machine. I added it to debug the pipeline failures that were occurring on Alpine and some other distributions. The function successfully returned the path to rsync on the pipeline machine, but the copy operation failed because rsync wasn't available on the guest machine.

cmd/limactl/copy.go
nirs: I don't think this it the code checking for initial ssh connection. If you run `limactl --debug start ...` you wll see the ssh commands used to detect if the guest is accessible.

cmd/limactl/copy.go
olamilekan000: for scp 

```
scp -F ~/.lima/deb/ssh.config "myfile.txt" olalekanodukoya@lima-deb:/dump
myfile.txt                                            100%    7     2.1KB/s   00:00
```

cmd/limactl/copy.go
```diff
@@ -118,24 +161,67 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+
+	if recursive {
+		rsyncFlags = append(rsyncFlags, "-r")
+	}
+
+	for _, arg := range args {
+		path := strings.Split(arg, ":")
+		switch len(path) {
+		case 1:
+			inst, ok := instances[instName]
```
nirs, line 201: How this will work? instances is an empty map created in this function.

I think we have only one case - `instance:path`, and it is common to both scp and rsync, so this check should be done first before we consider the copy tool. Input validation must always be done first.

cmd/limactl/copy.go
```diff
@@ -118,24 +161,67 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+			inst, err := store.Inspect(instName)
+			if err != nil {
+				if errors.Is(err, os.ErrNotExist) {
+					return nil, fmt.Errorf("instance %q does not exist, run `limactl create %s` to create a new instance", instName, instName)
+				}
+				return nil, err
+			}
+			sshStr = fmt.Sprintf("ssh -p %s -i %s", fmt.Sprintf("%d", inst.SSHLocalPort), "~/.lima/_config/user")
+			rsyncArgs = append(rsyncArgs, "-avz", "-e", sshStr, path[1])
+			instances[instName] = inst
```
nirs, line 223: I see, you add the instance to the map so the next iteration will find it. This is very complicated and hard to follow way to handle the arguments.

We should instead process the argument first, and convert them to internal structure that will be used to construct the command. This should be common to scp and rsync, so it should be done before we consider the tool.

cmd/limactl/copy.go
```diff
@@ -65,6 +65,48 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+			return err
+		}
+	default:
+		return fmt.Errorf("invalid copy tool %q", cpTool)
+	}
+
+	sshCmd := exec.Command(arg0, append(sshArgs, toolArgs...)...)
+	sshCmd.Stdin = cmd.InOrStdin()
+	sshCmd.Stdout = cmd.OutOrStdout()
+	sshCmd.Stderr = cmd.ErrOrStderr()
```
nirs, line 98: This is not sshCmd - this is very confusing.  Rename to copyCmd?

cmd/limactl/copy.go
```diff
@@ -65,6 +65,48 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+	case scp:
+		sshArgs, toolArgs, err = useScp(args, verbose, recursive)
+		if err != nil {
+			return err
+		}
+	case rsync:
+		toolArgs, err = useRsync(args, verbose, recursive)
+		if err != nil {
+			return err
+		}
```
nirs, line 90: This error handling is common, we can do this after the switch.

cmd/limactl/copy.go
```diff
@@ -65,6 +65,48 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+	arg0, err := exec.LookPath(string(cpTool))
+	if err != nil {
+		arg0, err = exec.LookPath(string(cpTool))
+		if err != nil {
+			return err
+		}
+	}
+	logrus.Infof("using copy tool %q", arg0)
+
+	var sshArgs, toolArgs []string
```
nirs, line 78: Why not single toolArgs list?

cmd/limactl/copy.go
```diff
@@ -65,6 +65,48 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+	case rsync:
+		toolArgs, err = useRsync(args, verbose, recursive)
+		if err != nil {
+			return err
+		}
+	default:
+		return fmt.Errorf("invalid copy tool %q", cpTool)
+	}
+
+	sshCmd := exec.Command(arg0, append(sshArgs, toolArgs...)...)
```
nirs, line 95: appending empty sshArgs to rsync will work but is very confusing.

Also I'm not sure why we ned to handle arg0 separately. The functions can return a complete csp or rsync command ([]string) and here we can create the command like this:

    copyCmd = exec.Command(command[0], command[1:]...)

Or, return `*exec.Command` from the functions instead of lists.

cmd/limactl/copy.go
```diff
@@ -65,6 +65,48 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+	sshCmd.Stdin = cmd.InOrStdin()
+	sshCmd.Stdout = cmd.OutOrStdout()
+	sshCmd.Stderr = cmd.ErrOrStderr()
+	logrus.Debugf("executing scp (may take a long time): %+v", sshCmd.Args)
```
nirs, line 99: Need to remove "scp" and just log the command instance:

    logrus.Debugf("executing %v (may take a long time)", sshCmd)

cmd/limactl/copy.go
```diff
@@ -65,6 +65,48 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+
+	// TODO: use syscall.Exec directly (results in losing tty?)
+	return sshCmd.Run()
+}
+
+func useScp(args []string, verbose, recursive bool) (sshArgs, scpArgs []string, err error) {
```
nirs, line 105: useScp does not describe well the purpose of the function. Calling this `scpCommand` and returning a `*exec.Command` created with "scp" and needed arguments will be better.

cmd/limactl/copy.go
```diff
@@ -118,24 +161,83 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
 	}
-	sshArgs := sshutil.SSHArgsFromOpts(sshOpts)
+	sshArgs = sshutil.SSHArgsFromOpts(sshOpts)
 
-	sshCmd := exec.Command(arg0, append(sshArgs, scpArgs...)...)
-	sshCmd.Stdin = cmd.InOrStdin()
-	sshCmd.Stdout = cmd.OutOrStdout()
-	sshCmd.Stderr = cmd.ErrOrStderr()
-	logrus.Debugf("executing scp (may take a long time): %+v", sshCmd.Args)
+	return sshArgs, scpArgs, nil
```
nirs, line 176: The fact that internally we want to manage sshArgs and scpArgs does not mean we need to return 2 lists. We can make the caller life easier and the code mode consistent if we return a single list, or a `*exec.Cmd` instance created from the lists.

pkg/hostagent/hostagent.go
```diff
@@ -424,6 +424,41 @@ func (a *HostAgent) Info(_ context.Context) (*hostagentapi.Info, error) {
 	return info, nil
 }
 
+func (a *HostAgent) installPackage() error {
```
nirs, line 427: I would not install the package here. If we depend on rsync in the guest, we can install the package using cloud-init.

In our user-data (part of cidata.iso), we can add:

```
packages:
- rsync
```

This will install rsync using cloud-init when the instance is provisioned. For distros that do not support this, the provision part of the yaml can install rsync manually.

pkg/hostagent/hostagent.go
```diff
@@ -424,6 +424,41 @@ func (a *HostAgent) Info(_ context.Context) (*hostagentapi.Info, error) {
...
+			sudo apt-get update && sudo apt-get install -y rsync
+		elif [ -f /etc/alpine-release ]; then
+			sudo apk add rsync
+		elif [ -f /etc/redhat-release ]; then
+			sudo yum install -y rsync
+		elif [ -f /etc/arch-release ]; then
+			sudo pacman -S --noconfirm rsync
+		else
+			echo "Unsupported Linux distribution. Please install rsync manually."
+		fi
```
nirs, line 445: Looks like you are reinventing cloudinit :-)

nirs:
I think we can do this:

- Use cloud-init to install rsync
  - If cloud-init packages does not work for a distro, it must be done in the lima.yaml provision scripts
- Replace scp with rsync

@AkihiroSuda what do you think?

cmd/limactl/copy.go
olamilekan000: >We should instead process the argument first, and convert them to internal structure that will be used to construct the command. This should be common to scp and rsync, so it should be done before we consider the tool.

The argument processing for scp and rsync is quite different, so combining them into a single function could make it complex and harder to maintain. II think it’s better to keep them separate and focus on consistent output structures instead.

hack/test-templates.sh
```diff
@@ -192,7 +192,7 @@ if [ "$got" != "$expected" ]; then
 fi
 
 INFO "Testing limactl copy command"
-tmpfile="$HOME/lima-hostname"
+tmpfile="/var/tmp/lima-hostname"
```
nirs, line 195: Not using $HOME/lima-hostname is a very good change - running tests locally should not create files in your home directory.

But this name may clash with another test, or a file created by someone else.

Best to use a temporary directory that can never clash with anything:

```
% mktemp -d /var/tmp/lima-test-templates.XXXXXX
/var/tmp/lima-test-templates.vuRmiE
```

AkihiroSuda:
> Use cloud-init to install rsync

SGTM

hack/test-templates.sh
olamilekan000: https://github.com/lima-vm/lima/pull/3163
I raised a pr for the change @nirs

nirs: Thanks! but it seems that this is the only test for limactl copy.

I opened #3170 for improving limactl copy tests. I hope we can improve test coverage before we change the code to avoid regressions.

olamilekan000: Okay can be done after https://github.com/lima-vm/lima/pull/3163 gets merged.

AkihiroSuda:
CI is failing

AkihiroSuda:

olamilekan000:
> What's the current status of this?

Update has been made. Currently awaiting review.

pkg/cidata/cidata.TEMPLATE.d/user-data
```diff
@@ -11,6 +11,9 @@ package_upgrade: true
 package_reboot_if_required: true
 {{- end }}
 
+packages:
+  - rsync
```
AkihiroSuda: Packages have to be installed via https://github.com/lima-vm/lima/blob/v1.2.1/pkg/cidata/cidata.TEMPLATE.d/boot/30-install-packages.sh

pkg/hostagent/hostagent.go
```diff
@@ -433,6 +433,7 @@ func (a *HostAgent) startHostAgentRoutines(ctx context.Context) error {
 	if err := a.waitForRequirements("essential", a.essentialRequirements()); err != nil {
 		errs = append(errs, err)
 	}
+
```
AkihiroSuda: ?

cmd/limactl/copy.go
```diff
@@ -74,87 +81,254 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
-	if recursive {
-		scpFlags = append(scpFlags, "-r")
+	cpTool, toolPath, err := selectCopyTool(copyPaths)
+	if err != nil {
+		return err
 	}
-	// this assumes that ssh and scp come from the same place, but scp has no -V
-	legacySSH := sshutil.DetectOpenSSHVersion("ssh").LessThan(*semver.New("8.0.0"))
+
+	logrus.Infof("using copy tool %q", toolPath)
```
AkihiroSuda:
```suggestion
	logrus.Debugf("using copy tool %q", toolPath)
```

cmd/limactl/copy.go
```diff
@@ -30,6 +30,20 @@ Example: limactl copy default:/etc/os-release .
 Not to be confused with 'limactl clone'.
 `
 
+type copyTool string
+
+const (
+	rsync copyTool = "rsync"
```
AkihiroSuda, line 57: Needs a test
https://github.com/lima-vm/lima/blob/c316067062312edd964189fc09257b27f4081dc7/hack/test-templates.sh#L231

cmd/limactl/copy.go
```diff
@@ -74,87 +81,254 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+	copyCmd.Stdin = cmd.InOrStdin()
+	copyCmd.Stdout = cmd.OutOrStdout()
+	copyCmd.Stderr = cmd.ErrOrStderr()
+	logrus.Debugf("executing %v (may take a long time)", copyCmd)
+
+	// TODO: use syscall.Exec directly (results in losing tty?)
+	return copyCmd.Run()
+}
+
+func parseArgs(args []string) ([]*copyPath, error) {
```
AkihiroSuda:
```suggestion
func parseCopyArgs(args []string) ([]*copyPath, error) {
```

cmd/limactl/copy.go
```diff
@@ -74,87 +81,254 @@ func copyAction(cmd *cobra.Command, args []string) error {
...
+			return nil, fmt.Errorf("path %q contains multiple colons", arg)
+		}
+
+		copyPaths = append(copyPaths, cp)
+	}
+
+	return copyPaths, nil
+}
+
+func selectCopyTool(copyPaths []*copyPath) (copyTool, string, error) {
```
AkihiroSuda: Probably there should be an explicit flag `--backend=(scp|rsync|auto)`

olamilekan000: Will `auto ` default to rsync?

AkihiroSuda: Yes when rsync is available

hack/test-templates.sh
```diff
@@ -245,6 +245,7 @@ fi
 INFO "Testing limactl copy command"
 tmpdir="$(mktemp -d "${TMPDIR:-/tmp}"/lima-test-templates.XXXXXX)"
 defer "rm -rf \"$tmpdir\""
+
```
AkihiroSuda: ?

hack/test-templates.sh
```diff
@@ -260,6 +261,73 @@ if [ "$got" != "$expected" ]; then
...
+	rsync_content="$(limactl shell "$NAME" cat /tmp/test-rsync-dir/testfile.txt)"
+	if [ "$rsync_content" != "test content" ]; then
+		ERROR "rsync file content mismatch"
+		exit 1
+	fi
+else
+	INFO "Skipping rsync backend test (rsync not available on host or guest)"
+fi
+
+INFO "Testing limactl copy command with auto backend"
```
AkihiroSuda: How is this different from `"Testing limactl copy command"` (L245) ?

olamilekan000: There isn't much difference but this specifies a backend.` limactl cp --backend=auto "$NAME":/etc/hostname "$tmpfile_auto_host"`

olamilekan000: do you want me to remove the test case? @AkihiroSuda

AkihiroSuda: Yes, seems better to remove

olamilekan000: done.

PR #3148, opened 2025-01-24
Disable GA serial port in QEMU as it is not used by GA

arixmkii:
GA was switched back to use SSH socket forwarding due to stability issues, but QEMU still exposes the serial as a Unix socket, which is

1) unused
2) removed from FS and replaced with forwarded via SSH

This is unnecessary and confusing. Future code changes might introduce a common feature toggle to disable code paths simultaneously in QEMU, HA and GA setup, but for now it at least will make both code paths consistent.

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -920,11 +920,13 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	args = append(args, "-chardev", fmt.Sprintf("socket,id=%s,path=%s,server=on,wait=off", qmpChardev, qmpSock))
 	args = append(args, "-qmp", "chardev:"+qmpChardev)
 
+	// virtserialport doesn't seem to work reliably: https://github.com/lima-vm/lima/issues/2064
+	// should go together with hostagent::New changes
 	// Guest agent via serialport
-	guestSock := filepath.Join(cfg.InstanceDir, filenames.GuestAgentSock)
-	args = append(args, "-chardev", fmt.Sprintf("socket,path=%s,server=on,wait=off,id=qga0", guestSock))
-	args = append(args, "-device", "virtio-serial")
-	args = append(args, "-device", "virtserialport,chardev=qga0,name="+filenames.VirtioPort)
+	// guestSock := filepath.Join(cfg.InstanceDir, filenames.GuestAgentSock)
+	// args = append(args, "-chardev", fmt.Sprintf("socket,path=%s,server=on,wait=off,id=qga0", guestSock))
+	// args = append(args, "-device", "virtio-serial")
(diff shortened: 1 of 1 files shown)


arixmkii:
This is minor, but was confusing to me, when I pulled the latest version and tried to run it to understand all the wiring between different processes.

pkg/qemu/qemu.go
```diff
@@ -920,11 +920,13 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
...
+	// should go together with hostagent::New changes
 	// Guest agent via serialport
-	guestSock := filepath.Join(cfg.InstanceDir, filenames.GuestAgentSock)
-	args = append(args, "-chardev", fmt.Sprintf("socket,path=%s,server=on,wait=off,id=qga0", guestSock))
-	args = append(args, "-device", "virtio-serial")
-	args = append(args, "-device", "virtserialport,chardev=qga0,name="+filenames.VirtioPort)
+	// guestSock := filepath.Join(cfg.InstanceDir, filenames.GuestAgentSock)
+	// args = append(args, "-chardev", fmt.Sprintf("socket,path=%s,server=on,wait=off,id=qga0", guestSock))
+	// args = append(args, "-device", "virtio-serial")
+	// args = append(args, "-device", "virtserialport,chardev=qga0,name="+filenames.VirtioPort)
```
nirs, line 929: If the code is not needed lets delete it.

pkg/qemu/qemu.go
```diff
@@ -920,11 +920,13 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	args = append(args, "-chardev", fmt.Sprintf("socket,id=%s,path=%s,server=on,wait=off", qmpChardev, qmpSock))
 	args = append(args, "-qmp", "chardev:"+qmpChardev)
 
+	// virtserialport doesn't seem to work reliably: https://github.com/lima-vm/lima/issues/2064
+	// should go together with hostagent::New changes
```
nirs, line 924: The comment is not clear. If it is needed to explain why the code is commented it can be removed with the commented code.

pkg/qemu/qemu.go
arixmkii: I'm still using it in an on/off manner (with changes in hostagent.go), while working on QEMU support on Windows hosts. So, it is not 100% useless for sure. May be we can just dismiss this PR for now and I will later create some other more prepared with either removal or feature toggle.

nirs: Maybe we need an option to enable/disable this?

You can also make this a draft, maybe others will find this change helpful.

arixmkii:
Replaced with a more well worked PR.

Issue #3162, opened 2025-01-27
Use Temporary Directory Instead of $HOME/lima-hostname for Tests


olamilekan000:
### Description

The current use of $HOME/lima-hostname for local tests is problematic as it can clash with other tests or user-created files. Switching to a temporary directory would prevent conflicts and improve reliability.

PR #3163, opened 2025-01-27
use temp directory as destination directory for copy command

olamilekan000:
```
Change switches local tests from using $HOME/lima-hostname to a 
temporary directory to prevent conflicts with user-created files 
and permission errors when writing to it. 
```

Fixes https://github.com/lima-vm/lima/issues/3162

### DIFF
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -192,10 +192,11 @@ if [ "$got" != "$expected" ]; then
 fi
 
 INFO "Testing limactl copy command"
-tmpfile="$HOME/lima-hostname"
+tmpdir="$(mktemp -d "${TMPDIR:-/tmp}"/lima-test-templates.XXXXXX)"
+defer "rm -rf \"$tmpdir\""
+tmpfile="$tmpdir/lima-hostname"
 rm -f "$tmpfile"
 limactl cp "$NAME":/etc/hostname "$tmpfile"
-defer "rm -f \"$tmpfile\""
 expected="$(limactl shell "$NAME" cat /etc/hostname)"
 got="$(cat "$tmpfile")"
 INFO "/etc/hostname: expected=${expected}, got=${got}"


hack/test-templates.sh
```diff
@@ -192,7 +192,8 @@ if [ "$got" != "$expected" ]; then
 fi
 
 INFO "Testing limactl copy command"
-tmpfile="$HOME/lima-hostname"
+tmpdir=$(mktemp -d /var/tmp/lima-test-templates.XXXXXX)
```
AkihiroSuda, line 195: The directory has to be removed with `defer`

hack/test-templates.sh
```diff
@@ -192,7 +192,9 @@ if [ "$got" != "$expected" ]; then
...
+defer "rm -rf \"$tmpdir\""
+tmpfile="$tmpdir/lima-hostname"
 rm -f "$tmpfile"
 limactl cp "$NAME":/etc/hostname "$tmpfile"
 defer "rm -f \"$tmpfile\""
```
nirs, line 200: Do we need to to remove the tempfile now? the temporary directory will be removed anyway.

nirs (approved):

hack/test-templates.sh
```diff
@@ -192,10 +192,11 @@ if [ "$got" != "$expected" ]; then
 fi
 
 INFO "Testing limactl copy command"
-tmpfile="$HOME/lima-hostname"
+tmpdir=$(mktemp -d /var/tmp/lima-test-templates.XXXXXX)
```
AkihiroSuda, line 195:
```suggestion
tmpdir="$(mktemp -d /var/tmp/lima-test-templates.XXXXXX)"
```

AkihiroSuda: Also, why `/var/tmp`, not `/tmp`?

AkihiroSuda: On macOS, `/var/tmp` seems almost unused except the following files;
```console
$ ls -la /var/tmp/
total 264
drwxrwxrwt   5 root           wheel     160  1 29 01:06 ./
drwxr-xr-x  34 root           wheel    1088  1 28 18:39 ../
-rw-r--r--   1 _windowserver  wheel  102400  1 28 18:39 cbrgbc_1.sqlite
-rw-r--r--   1 _windowserver  wheel   32768  1 28 18:39 cbrgbc_1.sqlite-shm
-rw-r--r--   1 _windowserver  wheel       0  1 28 18:39 cbrgbc_1.sqlite-wal
```

(Off-topic: what is "cbrgbc"? 🤔)

olamilekan000: > Also, why `/var/tmp`, not `/tmp`?

I think any works since we just need a space where any user can create a file.

olamilekan000: > On macOS, `/var/tmp` seems almost unused except the following files;
> 
> ```
> $ ls -la /var/tmp/
> total 264
> drwxrwxrwt   5 root           wheel     160  1 29 01:06 ./
> drwxr-xr-x  34 root           wheel    1088  1 28 18:39 ../
> -rw-r--r--   1 _windowserver  wheel  102400  1 28 18:39 cbrgbc_1.sqlite
> -rw-r--r--   1 _windowserver  wheel   32768  1 28 18:39 cbrgbc_1.sqlite-shm
> -rw-r--r--   1 _windowserver  wheel       0  1 28 18:39 cbrgbc_1.sqlite-wal
> ```
> 
> (Off-topic: what is "cbrgbc"? 🤔)

not sure what it is but it's definitely created by sqlite. It could be for macOS specific configs

nirs: The advantage of /var/tmp is that this is typically a standard file system on any platform. /tmp is is tmpfs on some Linux distros, which has special issues (direct I/O not supported, SEEK_HOLE/DATA very slow, etc). For our use case it should not matter.

Maybe we should just use $TMPDIR?

On macOS this is the safe user tmp directory used by most programs:
```
% echo $TMPDIR 
/var/folders/qs/_z10qgxn3gsfw8vmkr9b3s4w0000gn/T/
```

It can break if someone modify the env in a stupid way, but if they do they get what they deserve.

jandubois: > On macOS, `/var/tmp` seems almost unused except the following files;

On macOS `/tmp` is also cleaned automatically: files not accessed for 3 days are deleted. The same is done for `$TMPDIR`. I don't think `/var/tmp` is cleaned automatically, but am not sure.
 
> (Off-topic: what is "cbrgbc"? 🤔)

Comic Book Reader …? 🤣

AkihiroSuda: (Probably "cb" = "CoreBrightness", "rgbc" = "Red, Green, Blue, and Clear". The file is used by `/System/Library/HIDPlugins/ColourSensorFilterPlugin.plugin/Contents/MacOS/ColourSensorFilterPlugin`)

hack/test-templates.sh
```diff
@@ -192,10 +192,11 @@ if [ "$got" != "$expected" ]; then
 fi
 
 INFO "Testing limactl copy command"
-tmpfile="$HOME/lima-hostname"
+tmpdir="$(mktemp -d /var/tmp/lima-test-templates.XXXXXX)"
```
nirs, line 195: Based on the discussion, we should use here:

    tmpdir="$(mktemp -d ${TMPDIR:-/tmp}/lima-test-templates.XXXXXX)"

nirs (approved):

hack/test-templates.sh
```diff
@@ -192,10 +192,11 @@ if [ "$got" != "$expected" ]; then
 fi
 
 INFO "Testing limactl copy command"
-tmpfile="$HOME/lima-hostname"
+tmpdir="$(mktemp -d ${TMPDIR:-/tmp}/lima-test-templates.XXXXXX)"
```
nirs, line 195: The linter suggests to quote TMPDIR:
https://github.com/lima-vm/lima/actions/runs/13058754608/job/36436285027?pr=3163

nirs: This linter issue can be detected quickly locally by running `make lint`. This will save the cost of running all the vms in the CI and save time for reviewers looking at the CI failures.

olamilekan000: The linter didn't catch that locally.

AkihiroSuda (approved):
Thanks

PR #3165, opened 2025-01-27
website gha.md: job env variable sets LIMA_VERSION

msgilligan:
1. Fixes the `cache` action which was resolving `${{ env.LIMA_VERSION }}` to an empty string.

2. Makes the build more functional/reproducible by not getting LIMA_VERSION from an external source.

### DIFF
--- a/website/content/en/docs/examples/gha.md
+++ b/website/content/en/docs/examples/gha.md
@@ -50,6 +50,8 @@ jobs:
         gh attestation verify --owner=lima-vm "${FILE}"
         sudo tar Cxzvf /usr/local "${FILE}"
         rm -f "${FILE}"
+        # Export LIMA_VERSION For the GHA cache key
+        echo "LIMA_VERSION=${LIMA_VERSION}" >>$GITHUB_ENV
 
     - name: "Cache ~/.cache/lima"
       uses: actions/cache@v4


msgilligan:
I have made and tested the above change here:
https://github.com/nixos-lima/nixos-lima-config-sample/pull/1/commits/2e5bd7c26e4bce38ca7e6f0822eb37690793a71b

You can see the output from the job before the change here:

```
2025-01-27T18:37:15.7815441Z ##[group]Run actions/cache@v4
2025-01-27T18:37:15.7815682Z with:
2025-01-27T18:37:15.7815851Z   path: ~/.cache/lima
2025-01-27T18:37:15.7816041Z   key: lima-
2025-01-27T18:37:15.7816221Z   enableCrossOsArchive: false
2025-01-27T18:37:15.7816660Z   fail-on-cache-miss: false
2025-01-27T18:37:15.7816884Z   lookup-only: false
2025-01-27T18:37:15.7817065Z   save-always: false
2025-01-27T18:37:15.7817252Z ##[endgroup]
2025-01-27T18:37:16.1812540Z Cache not found for input keys: lima-
```

And after the change here:

```
Run actions/cache@v4
  with:
    path: ~/.cache/lima
    key: lima-v1.0.4
    enableCrossOsArchive: false
    fail-on-cache-miss: false
    lookup-only: false
    save-always: false
  env:
    LIMA_VERSION: v1.0.4
Cache not found for input keys: lima-v1.0.4
```

website/content/en/docs/examples/gha.md
```diff
@@ -44,7 +46,6 @@ jobs:
         GITHUB_TOKEN: ${{ github.token }}  # required by `gh attestation verify`
       run: |
         set -eux
-        LIMA_VERSION=$(curl -fsSL https://api.github.com/repos/lima-vm/lima/releases/latest | jq -r .tag_name)
```
AkihiroSuda, line 47: Can we just write this to `$GITHUB_ENV`?

msgilligan: I think so. But I think it's simpler and better  to have the version configured explicitly. Otherwise you could have two builds from the same commit that use different versions of Lima.

AkihiroSuda: But it is hard to keep updating the version then (unless dependabot can handle it)

msgilligan: Personally, I'd rather not have my CI jobs switch Lima version automatically.

I can see that some might want it to automatically update. However, this is sample code and I think it should lean towards simplicity.

msgilligan: Let me know what you decide and I'll revise the PR if necessary.

AkihiroSuda: I still prefer the automatic update here.
It is obvious how to pin a specific version and it does not need to be documented here.

msgilligan: Ok, I updated and force-pushed.

msgilligan:
@AkihiroSuda And here's an updated, passing, test job: https://github.com/nixos-lima/nixos-lima-config-sample/actions/runs/13019266424/job/36315933324

AkihiroSuda (approved):
Thanks

PR #3173, opened 2025-01-29
Make it possible to set the accelerator for qemu

### LINKED ISSUE #3156: Ability to run aarch64 guests on Linux without KVM
### Description

It should be possible to run aarch64 guest VMs on aarch64 Linux without support for KVM. There should be a way to set the QEMU options to support this. This would allow running Lima on `ubuntu-24.04-arm` GitHub Actions runners and also inside aarch64 Linux Lima instances on pre-M3 macOS systems.

Here is an example GitHub Actions workflow that _should_ work: https://github.com/nixos-lima/nixos-lima-config-sample/blob/b02a4c9cffaeb646feda5ad8940c9ade7599cb32/.github/workflows/ubuntu-arm.yml
This job fails with the following message in the log:

```
{"level":"debug","msg":"qemu[stderr]: Could not access KVM kernel module: No such file or directory","time":"2025-01-26T22:10:43Z"}
```

afbjorklund:
Similar to how you can override the cpu already

Closes #3156

### DIFF
--- a/pkg/instance/start.go
+++ b/pkg/instance/start.go
@@ -144,7 +144,7 @@ func Start(ctx context.Context, inst *store.Instance, limactl string, launchHost
 			if err != nil {
 				return fmt.Errorf("failed to find the QEMU binary for the architecture %q: %w", inst.Arch, err)
 			}
-			if accel := qemu.Accel(inst.Arch); accel == "hvf" {
+			if inst.AccelType == "hvf" {
 				entitlementutil.AskToSignIfNotSignedProperly(qExe)
 			}
 		}
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -54,6 +54,30 @@ var (
 	currentUser = Must(user.Current())
 )
 
+func Accel(arch Arch) string {
+        if IsNativeArch(arch) && IsAccelOS() {
+                switch runtime.GOOS {
+                case "darwin":
+                        return "hvf"
+                case "linux":
+                        return "kvm"
+                case "netbsd":
+                        return "nvmm"
+                case "windows":
+                        return "whpx"
+                }
+        }
+        return "tcg"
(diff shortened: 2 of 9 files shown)


afbjorklund:
Should the accelerator fallback to "tcg", if the default accelerator is not available?

For instance if `/dev/kvm` is missing on Linux, possibly because running in a VM

jandubois:
Is this really necessary? I know we say it is for "debugging", but you can already override QEMU settings with the `QEMU_SYSTEM_X86_64` and `QEMU_SYSTEM_AARCH64` environment variables. It is a bit clunky to use, but is this really a feature for more than a single user?

Having a whole matrix of accel types feels very overengineered for `lima.yaml`.

Same as with CPU types, this feels more like something that should go into a local config file, and not into the individual templates. Like we have `$LIMA_HOME/_config/networks.yaml`.

I understand that with this PR the setup can go into a `$LIMA_HOME/_config/default.yaml`, but I wonder if it is right to make the regular template schema more and more complex with settings that are not really instance-specific.

afbjorklund:
I think we always add the -accel flag even if kvm is missing, but I guess you can filter it out in a custom script wrapper

jandubois:
You can do this:

```console
> export QEMU_SYSTEM_AARCH64="/opt/homebrew/bin/qemu-system-aarch64 -machine virt"
❯ limactl start --tty=false --vm-type qemu
…
WARN[0001] [hostagent] Specifying args ([-machine virt]) via $QEMU_SYSTEM_AARCH64 is supported only for debugging!
WARN[0002] [hostagent] Not adding QEMU argument "-machine" "virt,accel=hvf", as it conflicts with "-machine" "virt"
…
```

It then fails for reasons I don't have time to investigate right now, but the basic mechanism works: if you specify `-machine` in the environment, then it won't be added by Lima.

PR #3180, opened 2025-01-30
limactl disk: Do not use qemu-img for raw disks

nirs:
Fix `limactl disk create` and `limactl disk resize` to use nativeimgutil for creating and resizing disks using raw format.

There are special cases (using direct I/O on certain file systems) when creating a raw image should be done with qemu-img, but these are not relevant to lima on macOS.

With this change we have no dependency on qemu when using VZ instance.

This change does not fix the issue of default "qcow2" disk format. Users must specify the disk format when creating additional disks for VZ instance:

    limactl disk create --format raw

Unfinished:
- Needs tests for nativeimgutil and disk command

Fixes: #2853

### DIFF
--- a/pkg/nativeimgutil/nativeimgutil.go
+++ b/pkg/nativeimgutil/nativeimgutil.go
@@ -2,20 +2,43 @@
 package nativeimgutil
 
 import (
+	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 
-	"github.com/containerd/continuity/fs"
+	containerdfs "github.com/containerd/continuity/fs"
 	"github.com/docker/go-units"
 	"github.com/lima-vm/go-qcow2reader"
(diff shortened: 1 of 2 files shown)


pkg/nativeimgutil/nativeimgutil.go
```diff
@@ -2,21 +2,44 @@
...
+	}
+	f, err := os.Create(dataDisk)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	return f.Truncate(int64(size))
+}
+
+// CreateRawDataDisk resizes a raw data disk.
```
nirs, line 37: Pushing a fix when the CI completes.

AkihiroSuda:
Is this still draft?

nirs:
> Is this still draft?

Posted as draft since it missing tests, but I did not have time to work on this. What I posted was tested and should be ready for merge.

AkihiroSuda (approved):
Thanks

PR #3182, opened 2025-01-31
Fix gcc line in Makefile for Linux

fdmarcin:
I needed this change for `make` to succeed on Fedora 41. Not sure if this is the right way to do this, so I'm open to suggestions. I haven't tested this on other Linux distros.

As it is currently on `master`, I've been getting this error when trying to run `make`:

```go
cgo: C compiler "x86_64-linux-gnu-gcc" not found: exec: "x86_64-linux-gnu-gcc": executable file not found in $PATH
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -485,7 +485,7 @@ ifeq ($(GOOS),darwin)
 to_uname_m = $(foreach arch,$(1),$(shell echo $(arch) | sed 's/amd64/x86_64/'))
 else ifeq ($(GOOS),linux)
 # CC is required for cross-compiling on Linux.
-CC = $(call to_uname_m,$(GOARCH))-linux-gnu-gcc
+CC = gcc
 else ifeq ($(GOOS),windows)
 # artifact in zip format also provided for Windows.
 ARTIFACT_FILE_EXTENSIONS += .zip


fdmarcin:
@norio-nomura Hi! You were the last one to rewrite that section of the Makefile. I've no idea if my changes are correct. Would you mind taking a look? :bow:

fdmarcin:
Alright, `release` fails, so this is probably not the right way to fix this. I'll leave this open, but if the maintainers don't have another idea, this can be closed.

jandubois:
> ```go
> compiler "x86_64-linux-gnu-gcc" not found: exec: "x86_64-linux-gnu-gcc": executable file not found in $PATH
> ```

Did you trying installing the `gcc-x86_64-linux` package (it is a subpackage of `cross-gcc`)?

Your change is breaking cross-compilation, so I think this issue should be closed.

afbjorklund:
I don't think that we _need_ to use the cross-compiler for the local build, only for the actual cross...

In the same way we don't need to set GOOS and GOARCH either, but just call out to `go build`.

It was just something that came with the caching and performance optimizations of the Makefile.

PR #3183, opened 2025-01-31
Allow to sftp-server binary found in path

arixmkii:
Lima allows to override majority of its dependencies (binaries) via $PATH configuration. `sftp-server` is one prominent exception. `sshocker` library will not try to resolve path unsell OpensshSftpServerBinary param is set to non empty string and will always fallback only to well known candidates.

This adds code in Lima to check if the binary could be found in path and in this case will instruct `sshocker` to use it. 

Implementing this will allow to use Lima with an OS flavor agnostic bundle of dependencies if configured via $PATH variable.

### DIFF
--- a/pkg/hostagent/mount.go
+++ b/pkg/hostagent/mount.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/lima-vm/lima/pkg/localpathutil"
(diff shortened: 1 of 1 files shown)


pkg/hostagent/mount.go
```diff
@@ -64,6 +65,10 @@ func (a *HostAgent) setupMount(m limayaml.Mount) (*mount, error) {
 		Readonly:            !(*m.Writable),
 		SSHFSAdditionalArgs: []string{"-o", sshfsOptions},
 	}
+	serverPath, err := exec.LookPath("sftp-server")
```
AkihiroSuda, line 68: This logic should be rather implemented in sshocker

AkihiroSuda: Also curious to know what distro needs this

arixmkii: > Also curious to know what distro needs this

It is from Windows experiment, where I put all "userland" dependencies into "_output/bin/bundle" https://github.com/lima-vm/lima/issues/909#issuecomment-2616343798 I'm exploring how feasible is it to just provide all the dependencies with Lima bundled.

> This logic should be rather implemented in sshocker

I will create PR in sshocker then. Will keep this for now as a draft for references.

coreymheld (approved):

pkg/hostagent/mount.go
arixmkii: PR in sshocker repo https://github.com/lima-vm/sshocker/pull/94

PR #3186, opened 2025-02-02
do not skip udp ports in guest agent

brandonros:
related to https://github.com/lima-vm/lima/discussions/3178

### DIFF
--- a/pkg/guestagent/kubernetesservice/kubernetesservice.go
+++ b/pkg/guestagent/kubernetesservice/kubernetesservice.go
@@ -131,11 +131,6 @@ func (s *ServiceWatcher) GetPorts() []Entry {
 		}
 
 		for _, portEntry := range service.Spec.Ports {
-			if portEntry.Protocol != corev1.ProtocolTCP {
-				// currently only TCP port can be forwarded
-				continue
-			}
-
 			var port int32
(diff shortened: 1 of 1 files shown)


nirs (commented):
It looks good to me, but lets see what @balajiv113 thinks.

Did you test it? can you the results?

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -131,11 +131,6 @@ func (s *ServiceWatcher) GetPorts() []Entry {
 		}
 
 		for _, portEntry := range service.Spec.Ports {
-			if portEntry.Protocol != corev1.ProtocolTCP {
```
brandonros, line 134: add a check if grpc vs ssh is being used?

nirs: If the guest agent knows it would be nice to skip only when using the builtin port forwarder. But it is more important to get this working first, optimizing can be done later.

balajiv113:
Looks good to me. It would be great if we can try out once with a test logic.

AkihiroSuda:
> test

ping @brandonros 

Also, please sign off the commit
https://github.com/lima-vm/lima/pull/3186/checks?check_run_id=36534914016

Issue #3150, opened 2025-01-24
Fails to boot any VM due to panic in hostagent


s1gnate-sync:
### Description

**Bug**
In 90% cases `lima start NAME` fails to boot VM due to panic in hostagent in attempt of creating iso file (via go-diskfs package). It doesn't matter if I clean cache or remove whole lima dir.

### Env
 
- MacOS 15.1.1 on Apple Silicon
- lima installed via brew (limactl version 1.0.4, go-diskfs version 1.4.1)
- Using unmodified
- No special changes made to configuration/templates

### Steps

- Create VM
- Start VM

Obviously I expect it to start, not to panic...

### Log sample
```
~>limactl start rekcod --debug
DEBU[0000] interpreting argument "rekcod" as an instance name
INFO[0000] Creating an instance "rekcod" from template://default (Not from template://rekcod)
WARN[0000] This form is deprecated. Use `limactl create --name=rekcod template://default` instead
INFO[0000] debug mode detected, adding more guest agent candidates: /opt/_output/share/lima/lima-guestagent.Linux-aarch64
DEBU[0000] ResolveVMType: resolved VMType "vz" (existing instance, with "/Users/user/.lima/rekcod/vz-identifier")
INFO[0000] Using the existing instance "rekcod"
DEBU[0000] ResolveVMType: resolved VMType "vz" (existing instance, with "/Users/user/.lima/docker/vz-identifier")
DEBU[0000] ResolveVMType: resolved VMType "vz" (existing instance, with "/Users/user/.lima/podman/vz-identifier")
DEBU[0000] ResolveVMType: resolved VMType "vz" (existing instance, with "/Users/user/.lima/random/vz-identifier")
DEBU[0000] ResolveVMType: resolved VMType "vz" (existing instance, with "/Users/user/.lima/rekcod/vz-identifier")
DEBU[0000] Make sure "user-v2" network is stopped
DEBU[0000] Make sure usernet network is stopped
DEBU[0000] Make sure "shared" network is stopped
DEBU[0000] Make sure "bridged" network is running
DEBU[0000] Make sure "host" network is stopped
INFO[0000] Starting the instance "rekcod" with VM driver "vz"
DEBU[0000] [hostagent] ResolveVMType: resolved VMType "vz" (existing instance, with "/Users/user/.lima/rekcod/vz-identifier")
[...]

PR #3187, opened 2025-02-02
Update go-diskfs package to 1.5.0

s1gnate-sync:
Currently lima uses 1.4.1 version of the package, but the most recent is 1.5.0 and it has some changes in function signature. I've updated lima to reflect changes in the recent version. It's just a preparation for a fix #3150. Here is PR for the package https://github.com/diskfs/go-diskfs/pull/279

PR #3190, opened 2025-02-03
Update go-diskfs package to 1.5.0

s1gnate-sync:
Currently lima uses 1.4.1 version of the package, but the most recent is 1.5.0 and it has some changes in function signature. I've updated lima to reflect changes in the recent version. It's just a preparation for a fix https://github.com/lima-vm/lima/issues/3150. Here is PR for the package https://github.com/diskfs/go-diskfs/pull/279

It's just the same as #3187, only difference is my signature.


AkihiroSuda (approved):
Thanks

PR #3197, opened 2025-02-06
Restrict pessimization of M4 arch to macOS 15.2

jandubois:
It has reportedly been fixed in 15.3:
* https://gitlab.com/qemu-project/qemu/-/issues/2721#note_2318005270
* another report from a colima user on Slack

This commit also caches the results of `osutil.ProductVersion` (on macOS) because additional calls should always return the same results.

The `cortex-a7` CPU fails to run some workloads that earlier M* CPUs already support:

```
$ docker run -it --rm  public.ecr.aws/amazonlinux/amazonlinux:2023
Fatal glibc error: This version of Amazon Linux requires a newer ARM64 processor
                   compliant with at least ARM architecture 8.2-a with Cryptographic
                   extensions. On EC2 this is Graviton 2 or later.
```

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -1194,14 +1194,16 @@ func HasHostCPU() bool {
 	switch runtime.GOOS {
 	case "darwin":
 		if hasSMEDarwin() {
-			// SME is available since Apple M4 running macOS 15.2.
-			//
-			// However, QEMU is not ready to handle SME yet.
-			//
-			// https://github.com/lima-vm/lima/issues/3032
-			// https://gitlab.com/qemu-project/qemu/-/issues/2665
-			// https://gitlab.com/qemu-project/qemu/-/issues/2721
-			return false
+			macOSProductVersion, err := osutil.ProductVersion()
+			if err != nil || macOSProductVersion.Equal(*semver.New("15.2")) {
+				// SME is available since Apple M4 running macOS 15.2, but it was broken on macOS 15.2.
+				// It has been fixed in 15.3.
+				//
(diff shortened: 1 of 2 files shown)


pkg/limayaml/defaults.go
```diff
@@ -1194,14 +1194,18 @@ func HasHostCPU() bool {
...
-			//
-			// However, QEMU is not ready to handle SME yet.
-			//
-			// https://github.com/lima-vm/lima/issues/3032
-			// https://gitlab.com/qemu-project/qemu/-/issues/2665
-			// https://gitlab.com/qemu-project/qemu/-/issues/2721
-			return false
+			macOSProductVersion, err := osutil.ProductVersion()
+			if err != nil || macOSProductVersion.Equal(*semver.New("15.2")) {
+				// SME is available (but broken) since Apple M4 running macOS 15.2.
```
AkihiroSuda, line 1199:
```suggestion
				// SME is available since Apple M4 running macOS 15.2, but it was broken on macOS 15.2.
```

pkg/limayaml/defaults.go
```diff
@@ -1194,14 +1194,18 @@ func HasHostCPU() bool {
...
+				// It has been fixed in 15.3.
+				//
+				// However, QEMU is not ready to handle SME yet.
```
AkihiroSuda, line 1202: Maybe remove this line, if no action is needed on QEMU side?

jandubois: A workaround is needed on 15.2 only, but I've removed the comment. Most users should have upgraded to 15.3.

AkihiroSuda (approved):
Thanks

PR #3201, opened 2025-02-06
Checking for protected dirs must be against mountPoint and not location

### LINKED ISSUE #3200: Mount validation should check the mountPoint and not the location setting
### Discussed in https://github.com/lima-vm/lima/discussions/3199

<div type='discussions-op-text'>

<sup>Originally posted by **msgilligan** February  5, 2025</sup>
If I try to create a Lima instance inside of a a Lima VM (which is using the default home directory name -- `/home/user.linux`), I get the error message: `field 'mounts[0].location' is the reserved internal home directory`

If I try do do something like:
```
- location: "/home"
  mountPoint: "/casa"
```

I get: `field 'mounts[0].location' must not be a system path such as /etc or /usr`

I realize I could rebuild the host Lima instance to not use `/home/user.linux`, but I'd rather not do that.

How can I mount `~` inside the nested instance?

</div>

jandubois:
The validation code was written at a time when all mounts were mounted at `location` inside the guest because there was no separate `mountPoint` setting.

Now that `mountPoint` exists, the only validation for `location` is that the path exists on the host and designates a directory. All validation to protect guest directories must be done against `mountPoint`.

Fixes #3200

### DIFF
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -152,14 +152,6 @@ func Validate(y *LimaYAML, warn bool) error {
 		if err != nil {
 			return fmt.Errorf("field `mounts[%d].location` refers to an unexpandable path: %q: %w", i, f.Location, err)
 		}
-		switch loc {
-		case "/", "/bin", "/dev", "/etc", "/home", "/opt", "/sbin", "/tmp", "/usr", "/var":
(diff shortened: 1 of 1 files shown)


jandubois:
Sorry, just realized that I broke the `location` validation into 2 chunks. I've now moved the `mountPoint` validation after all `location` validation, to keep related things together.

AkihiroSuda (approved):
Thanks

PR #3207, opened 2025-02-08
Make it possible to configure user shell

afbjorklund:
The default shell in FreeBSD is `/bin/tcsh`.

You need to install `/usr/local/bin/bash`.

Maybe others want a different login shell?

For instance `fish` or `zsh`, or whatever...


AkihiroSuda (approved):
Thanks

PR #3209, opened 2025-02-09
Add example with containers on Podman Engine

afbjorklund:
There was no example with using Podman, on https://lima-vm.io

Use the docker client with DOCKER_HOST and compatibility socket,
instead of messing around with podman-remote and CONTAINER_HOST.

For instance Podman Desktop also uses a docker client. Compose too.


AkihiroSuda (approved):
Thanks

PR #3220, opened 2025-02-11
test: Update ubuntu runner as 20.04 is going away

rfay:
Ubuntu 20.04 is going to be turned off shortly as a host runner.

Use 22.04 instead. The note though may imply problems:

`# An old release of Ubuntu is chosen for glibc compatibility`

IMO it might be better to just use ubuntu-latest and remove the comment?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -49,7 +49,7 @@ jobs:
         path: _artifacts/
   release:
     # An old release of Ubuntu is chosen for glibc compatibility
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     needs: artifacts-darwin
     timeout-minutes: 20
     # The maximum access is "read" for PRs from public forked repos


rfay:
I just happened to note this since I had a checkout on my machine and grepped for all the ubuntu-20.04 usages :)

jandubois:
> What symbols are we requiring?

I don't know, but I think this is because we compile with `CGO_ENABLED=1` for access to the glibc host resolver.

afbjorklund:
> I think this is because we compile with CGO_ENABLED=1 for access to the glibc host resolver.

Yeah, the go change was a problem for alpine (since it doesn't use glibc, and didn't have libresolv)

Just wondered if it was a big problem, but then again probably only affects RHEL 8 and such?

This project had some nice tables and workarounds: https://github.com/pypa/manylinux

jandubois:
> This project had some nice tables and workarounds: https://github.com/pypa/manylinux

My head hurts just looking at that...

I think just building our releases on the oldest Ubuntu version that has supported GitHub runners is good enough. If platforms still need special consideration beyond that, then ideally they should distribute Lima through their package system. Or users can compile it themselves: `make && make install`.

AkihiroSuda (approved):
Thanks

jandubois:
> So we might want to delay merging this PR until later in March, in case we do another patch release before then (I don't expect we will; just in case).

Looks like I just jinxed it, and we had to release 1.0.6 a few days later...

I'm re-running failed CI jobs to get everything green, but still think we should hold back merging until late March, just in case. It only affects the `release` workflow, not the `test` one.

AkihiroSuda:
> March 4 14:00 UTC – 22:00 UTC

Shall we merge?

PR #3233, opened 2025-02-14
WSL: Create symlink at /mnt/lima-cidata

mook-as:
Previously we only copied a single file to `/mnt/lima-cidata`, just enough to pass the requirements check.  However, that means various other places that hard-codes `/mnt/lima-cidata` does not resolve correctly.  Replace it with a symlink instead so all files (in particular, `param.env`) can be accessed.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/02-wsl2-setup.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/02-wsl2-setup.sh
@@ -12,6 +12,5 @@ sudo chown "${LIMA_CIDATA_USER}" "${LIMA_CIDATA_HOME}"/.ssh/authorized_keys
 # add $LIMA_CIDATA_USER to sudoers
 echo "${LIMA_CIDATA_USER} ALL=(ALL) NOPASSWD:ALL" | sudo tee -a /etc/sudoers.d/99_lima_sudoers
 
-# copy some CIDATA to the hardcoded path for requirement checks (TODO: make this not hardcoded)
-sudo mkdir -p /mnt/lima-cidata
-sudo cp "${LIMA_CIDATA_MNT}"/meta-data /mnt/lima-cidata/meta-data
+# symlink CIDATA to the hardcoded path for requirement checks (TODO: make this not hardcoded)
+sudo ln -sfFn "${LIMA_CIDATA_MNT}" /mnt/lima-cidata


AkihiroSuda:
cc @pendo324

jandubois (approved):
Thanks, LGTM

It looks like on Windows `LIMA_CIDATA_MNT` is set in https://github.com/lima-vm/lima/blob/5856572195f99d86e8d5df8328d3326d975a9b91/pkg/wsl2/lima-init.TEMPLATE#L5 and this code makes the actual data available in the location expected by the boot scripts.

The WSL2 driver is using its own minimal cloud-init code and only used `meta-data` so far, which means that the settings from `param.env` are not available on Windows. I haven't looked why Windows doesn't seem to need `lima.env`.

afbjorklund:
> The problem is that `/mnt/lima-cidata` is hard-coded in various places that don't have access to the `LIMA_CIDATA_MNT` environment variable.

It is a problem on other platforms as well, that either don't use `/mnt` or only mount the cidata.iso during cloud-init and unmount it afterwards.

Currently the workaround is to `mkdir /mnt`  and copy all the files there, by [patching](https://github.com/afbjorklund/tinycloud/blob/main/files/tinycloudinit.patch) the bootup scripts. Maybe it can be made more configurable?

* TinyCore* (`/tmp/$RANDOM$RANDOM.cinit.rom`, temporary)
* CoreOS (`/media/configdrive`)
* FreeBSD (`/media/nuageinit`, temporary)

\* Also need to make a RAM copy of the guestagent binary, since sadly it is not executable directly from the cidata mount.

Eventually it should probably be moved out to the data disk, but currently it is a question of the mount/execution order...

There is a similar story for `/run`, in lima-boot-done/lima-ssh-ready.

Systems without systemd only have `/var/run`, so another `mkdir`.

jandubois:
> cc @pendo324

@AkihiroSuda Are you waiting for some feedback from @pendo324, or was that just a FYI, and we can go ahead and merge?

jandubois:
> Currently the workaround is to `mkdir /mnt` and copy all the files there, by [patching](https://github.com/afbjorklund/tinycloud/blob/main/files/tinycloudinit.patch) the bootup scripts.

@afbjorklund Your patch is missing `param.env`. 😄 

> Maybe it can be made more configurable?

I think the only remaining place where it can't be configured right now is the requirements code that is running the probes.

I don't see how we can make it configurable unless we set it in `lima.env`.

AkihiroSuda (approved):
Thanks, let's merge this

pendo324:
Was AFK this weekend, change lgtm, thanks for merging it

PR #3236, opened 2025-02-14
Makefile: allow setting KEEP_SYMBOLS=1

AkihiroSuda:
The symbols are needed for applying gomodjail (library sandbox) to Lima. https://github.com/AkihiroSuda/gomodjail

This grows the `limactl` binary from 27M to 30M.

Ideally, Makefile should rather separate the `install-split` target from the `install` target as in GNU's C programs.
However, this is not possible with the current version of Go, as there is no `go tool strip` command.


jandubois (approved):
Thanks, LGTM

I wonder if there should be an error if the setting is neither `0` or `1`, but maybe that would just be clutter.

PR #3238, opened 2025-02-14
go.mod: experimental integration of gomodjail (library sandbox)

AkihiroSuda:
https://github.com/AkihiroSuda/gomodjail

gomodjail imposes syscall restrictions on a specific set of Go modules (excepts ones that use unsafe pointers, reflections, etc.), so as to mitigate their potential vulnerabilities and supply chain attack vectors.

Usage:
```
make KEEP_SYMBOLS=1 binaries install
gomodjail run --go-mod=./go.mod -- limactl rm -f default
```

TODO: pack gomodjail, go.mod, and limactl into a single binary

Hint: use `git diff --word-diff` for reviewing the changes in this commit

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -476,3 +476,35 @@ jobs:
       uses: ./.github/actions/upload_failure_logs_if_exists
       with:
         suffix: ${{ matrix.template }}
+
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #3254, opened 2025-02-17
Use SSHAddress for SSH forwarding in QEMU

### LINKED ISSUE #3241: QEMU commandline should use Instance.SSHAddress insead of hardcoded loopback
### Description

Currently QEMU always forwards SSH to 127.0.0.1 https://github.com/lima-vm/lima/blob/2c84e0c77ab659ae3dcbafa3d199ad6fdaad0772/pkg/qemu/qemu.go#L719 This doesn't play well with some WSL2 integration scenarios. It should match the instance address instead, then this address defining logic would reside in from the single source.

I plan to submit a fix for this.

arixmkii:
Fixes #3241

Extending config (effectively data object to pass into driver) with a new field. The field is filled in with the value from Instance. The field is filled only in Start method as it is only used during QEMU machine startup and make no sense in other methods.

Right now it will always be 127.0.0.1 as it is hardcoded during Instance inspect, but in my Windows experiment this value is updated based on WSL2 networking mode (this code still distant from upstream). There is no immediate use for this change, but there is no good reason to have this hardcoded inside driver, when there is already configurable source of truth.

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -41,6 +41,7 @@ type Config struct {
 	InstanceDir  string
 	LimaYAML     *limayaml.LimaYAML
 	SSHLocalPort int
+	SSHAddress   string
 }
 
 // MinimumQemuVersion is the minimum supported QEMU version.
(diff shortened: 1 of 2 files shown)


pkg/qemu/qemu.go
AkihiroSuda, line 1: This line needs to be updated too?

https://github.com/lima-vm/lima/blob/d2fce41edc738fce6006c070f5efb4573a02d7b4/pkg/store/instance.go#L97

arixmkii: It will. Eventually. I have this changed in my prototype https://github.com/arixmkii/lima/blob/ffb66c42509c3575d74ca78e562e9085f1ecbd5e/pkg/store/instance.go#L99 But it uses new external dependencies and it is too early to enable this code here. This change is just a refactoring of existing code to make it configurable.

AkihiroSuda (approved):
Thanks

PR #3255, opened 2025-02-17
Implement feature toggle for GA forwarding mode in QEMU

### LINKED ISSUE #3177: QEMU: Support feature toggle for GA via serial port
### Description

arixmkii:
Fixes #3177

Current code supports 2 modes:
- forwarding via SSH forwarding
- forwarding by QEMU via serialport

In GA they are exclusive modes - it either talks to device or exposes Unix socket inside VM.

At some point the GA part was switched to Unix socket only, but QEMU start command line was not updated. SSH forwarder takes care of it by removing Unix socket before starting its own, but there is no point in having it, when it can't be used.

This code introduces feature toggle allowing to use either mode. Keeping the "unstable" option could be beneficial for environments, where Unix socket forwarding is tricky (Windows hosts).

### DIFF
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -98,6 +98,7 @@ type VMOpts struct {
 
 type QEMUOpts struct {
 	MinimumVersion *string `yaml:"minimumVersion,omitempty" json:"minimumVersion,omitempty" jsonschema:"nullable"`
+	VirtioGA       *bool   `yaml:"virtioGA,omitempty" json:"virtioGA,omitempty" jsonschema:"nullable"`
 }
 
 type Rosetta struct {
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -131,8 +131,9 @@ func New(instName string, stdout io.Writer, signalCh chan os.Signal, opts ...Opt
 		}
 		vSockPort = port
 	} else if *inst.Config.VMType == limayaml.QEMU {
-		// virtserialport doesn't seem to work reliably: https://github.com/lima-vm/lima/issues/2064
-		virtioPort = "" // filenames.VirtioPort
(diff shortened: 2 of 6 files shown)


arixmkii:
Replaces https://github.com/lima-vm/lima/pulls/arixmkii

arixmkii:
Converted to Draft. The code looks legit, but I haven't done manual testing, nor I had time to integrate it with my CI builds for Windows.

Publishing it for some early reviews to collect feedback that I'm not doing something absolutely stupid.

pkg/limayaml/limayaml.go
```diff
@@ -98,6 +98,7 @@ type VMOpts struct {
 
 type QEMUOpts struct {
 	MinimumVersion *string `yaml:"minimumVersion,omitempty" json:"minimumVersion,omitempty" jsonschema:"nullable"`
+	VirtioGA       *bool   `yaml:"virtioGA,omitempty" json:"virtioGA,omitempty" jsonschema:"nullable"`
```
AkihiroSuda, line 101: Maybe this should be `guestAgentTransportType *string` ?
Also, this may potentially apply to non-QEMU VM drivers?

arixmkii: Good call! I will check if this could be applicable for other platforms. Having it as typed option instead of boolean flag feels like a better design.

arixmkii:
I rebased it, because I believe it still has value before a more complete feature is implemented as requested in review.

I also enabled QEMU integration tests to prove that this actually work (after all previous patches merged already). I had to disable mount tests for QEMU on Windows, because it would need at least these changes:
* https://github.com/lima-vm/lima/pull/3318
* https://github.com/lima-vm/sshocker/pull/97

arixmkii:
`default-windows.yaml` is just a copy of `default.yaml` with mounts section removed.

PR #3239, opened 2025-02-14
Splitted https://lima-vm.io/docs/config/network/ to multiple pages

kairveeehh:
Resolves (#1815 ) , did the fixes .

PR #3259, opened 2025-02-18
docs/config/network: follow-up

AkihiroSuda:
- Fixed broken links
- Removed extra headings
- Clarified that vmnet is only available on macOS

Follow-up to:
- #3239


jandubois (approved):
Thanks, LGTM

PR #3261, opened 2025-02-18
Add optional loading of UEFI firmware via -bios parameter in QEMU

### LINKED ISSUE #3176: QEMU: support UEFI via `-bios`
### Description

Add configuration option to load UEFI firmware through `-bios` command line option. It is not really needed for the currently supported platforms, but on Windows WHPX acceleration experiencing issues with pflash mounted firmware https://gitlab.com/qemu-project/qemu/-/issues/513 Patch exists to workaround this, but using out of tree patches is a complication on its own. There is a workaround provided to load concatenated vars and code via `-bios`.

Current proposal:
* add flag under Firmware section similar to LegacyBIOS
* default value is false or false for non-windows OSes if this is possible/suitable
* add code path to lookup vars counterpart (this is a bit trickier than code, because x86_64 is using vars of i386, but could be handled with trivial conditional reasoning) 
* generate concatenated version in the machine working directory
* generate QEMU command line based on the flag

End goal - support upstream QEMU builds with Lima on Windows hosts.

I plan to work on this feature.

arixmkii:
Fixes #3176

Limiting this to X8664 arch as there is no information if this works for other architectures. This allows to load UEFI firmware on Windows hosts, where pflash can't be used in combination with WHPX acceleration. Default to false on non-Windows hosts and true on Windows (to have working default).

The FW resolution if set to true and running amd64 VM works like this
1. Check if we have it in already loaded image, if so, then consider it valid (already concatenated)
2. ~Check all downloads, if any matches then consider it concatenated image and use like regular download~ (downloading through standard mechanisms is removed to prevent potential incompatibilities)
3. If not yet available in downloaded, then try to find firmware and firmware vars, concatenate and store into downloaded file (so, future runs could find it w/o additional actions)

PR update to use env var instead of Lima yaml

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


arixmkii:
I'm, like many people with JVM background, genuinely bad at naming user facing settings. Any better ideas than ugly `compatUEFIInBIOS` are welcome.

pkg/limayaml/limayaml.go
```diff
@@ -182,6 +182,11 @@ type Firmware struct {
 	// LegacyBIOS is ignored for aarch64.
 	LegacyBIOS *bool `yaml:"legacyBIOS,omitempty" json:"legacyBIOS,omitempty" jsonschema:"nullable"`
 
+	// CompatUEFIViaBIOS forces QEMU to load concatenated firmware with -bios option
+	// CompatUEFIViaBIOS is ignored for non x86_64
+	// This should be deprecated, if the issue in QEMU is fixed https://gitlab.com/qemu-project/qemu/-/issues/513
```
AkihiroSuda, line 187: Since this is already deprecated since its birth, may this should rather be an env var?
Similar to `LIMA_SSH_PORT_FORWARDER` (originally planned to be deprecated in v1.0, now postponed to v1.1)
https://lima-vm.io/docs/config/environment-variables/

pkg/qemu/qemu.go
```diff
@@ -1164,3 +1202,40 @@ func getFirmware(qemuExe string, arch limayaml.Arch) (string, error) {
 	qemuArch := strings.TrimPrefix(filepath.Base(qemuExe), "qemu-system-")
 	return "", fmt.Errorf("could not find firmware for %q (hint: try copying the \"edk-%s-code.fd\" firmware to $HOME/.local/share/qemu/)", arch, qemuArch)
 }
+
+func getFirmwareVars(qemuExe string, arch limayaml.Arch) (string, error) {
+	var targetArch string
+	switch arch {
+	case limayaml.X8664:
+		targetArch = "i386"
```
AkihiroSuda, line 1210: ?

pkg/qemu/qemu.go
```diff
@@ -624,9 +625,44 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
...
+				if err != nil {
+					return "", nil, err
+				}
+				defer varsFile.Close()
+				codeFile, err := os.Open(firmware)
+				if err != nil {
+					return "", nil, err
+				}
+				defer codeFile.Close()
+				downloadedFile, err := os.OpenFile(downloadedFirmware, os.O_CREATE|os.O_WRONLY, 0o644)
```
AkihiroSuda, line 644: Probably the file name should be changed, as the concatenated file would be incompatible with `QemuEfiCodeFD`

pkg/qemu/qemu.go
arixmkii: x86_64 and i386 use the same layout for vars, so, there is no x86_64 version in the distribution. One needs to concatenate i386 to x86_64 code. This is the legacy of x86_64 being derivative of i386. Probably this deserves explanation with a comment.

pkg/qemu/qemu.go
arixmkii: Will change this. No objection here. I just re-used current code for quick prototype, but it is obviously not the most straightforward or easy to understand hack.

pkg/limayaml/limayaml.go
arixmkii: I agree. It might never be removed, because fixes in Windows QEMU version are not of the highest priority. If the ergonomics of the environment var would be insufficient it could be changed later into full-fledged option in Lima yaml. I'm moving this into draft to rework.

arixmkii:
Reworked based on received feedback. Now env var is used as the mode switch. And now original and altered code paths has clearer separation in code. Also, downloading mechanism is disabled if compatibility mode is used.

AkihiroSuda (dismissed):
Thanks

pkg/qemu/qemu.go
```diff
@@ -598,36 +599,89 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	}
 	if !legacyBIOS {
 		var firmware string
+		firmwareInBios := runtime.GOOS == "windows"
+		if envVar := os.Getenv("LIMA_QEMU_UEFI_IN_BIOS"); envVar != "" {
```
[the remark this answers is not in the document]
arixmkii: Updated

AkihiroSuda (approved):
Thanks

Issue #3265, opened 2025-02-20
errors inspecting instance: [field `mounts[0].mountPoint` is the reserved internal home directory


attiss:
### Description

Since I updated to the latest lima version, I get the following error for an existing instance:

```
$ lima --version
limactl version 1.0.6
```

```
$ limactl start xyz
INFO[0000] Using the existing instance "xyz"
FATA[0000] errors inspecting instance: [field `mounts[0].mountPoint` is the reserved internal home directory "/home/xyz.linux"]
```

The instance was created with the below configuration:

```yaml
mountType: virtiofs
mountInotify: true
mounts:
  - location: ~/xyz
    mountPoint: /home/xyz.linux
    writable: true
```

My use case is to mount a folder from the host machine and use that as the whole home directory in the guest (which makes it possible for me to reuse the home if I destroy the instance and create a new one).

I see some related issues and PRs related to the above error:
- https://github.com/lima-vm/lima/issues/3200
- https://github.com/lima-vm/lima/pull/3201

If I'm not mistaken the above PR aimed to fix a bug which validated the `location` instead of the `mountPoint` path - therefore we limited what folders can be mounted from the host instead of limiting what places we can mount to.

I was wondering if it would be possible to keep supporting my use case so people can easily migrate their home from one instance to another?

Thank you!

jandubois:
> If I'm not mistaken the above PR aimed to fix a bug which validated the `location` instead of the `mountPoint` path - therefore we limited what folders can be mounted from the host instead of limiting what places we can mount to.

Yes, that is correct. The 2 things used to be the same before we added the ability to specify a different `mountPoint`, and forgot to update the validation logic.

The intend of the validation logic has always been to prevent you from mounting filesystems on top of certain system directories or the internal home directory.

The whole point of using `/home/$USER.linux` as the home directory inside the instance was to keep it separate from the home directory of the host.

> I was wondering if it would be possible to keep supporting my use case so people can easily migrate their home from one instance to another?

I think in general the assumption is that Lima can modify files in the internal home directory as needed, and therefore might break some host configuration by overwriting their dot files. E.g. Lima will modify the `$HOME/.profile` file to modify the `PATH`.

I think your use case could be implemented by just changing `HOME` in the `/home/$USER.linux/.profile`:

```console
❯ yq .provision ~/.lima/home/lima.yaml
- mode: user
  script: |
    if ! grep 'HOME=' .profile; then
      echo "HOME=/home/{{.User}}" >> .profile
[...]

PR #3266, opened 2025-02-20
warn if mount point is the default home directory

attiss:
This PR makes it possible to mount to the default home folder in the instances so people can easily migrate home folders across instances.

Resolves https://github.com/lima-vm/lima/issues/3265


jandubois:
This PR is not needed/desired. https://github.com/lima-vm/lima/issues/3265#issuecomment-2672381307 describes a way to achieve the same results without loosening the validation.

PR #3286, opened 2025-02-26
Don't panic when deleting instances that don't validate

jandubois:
During Lima development you can end up with Lima instances that don't validate with the current version. In that case deleting them may be impossible:

```console
❯ l delete -f alpine
WARN[0000] Non-strict YAML detected; please check for typos  comment="main file \"/Users/jan/.lima/alpine/lima.yaml\"" error="[20:3] unknown field \"content\"\n  18 | provision:\n  19 | - mode: data\n> 20 |   content: this is my data\n         ^\n  21 |   path: /etc/datafile"
INFO[0000] The  driver process seems already stopped
INFO[0000] Removing *.pid *.sock *.tmp under "/Users/jan/.lima/alpine"
[signal SIGSEGV: segmentation violation code=0x2 addr=0x8 pc=0x10122618c]

goroutine 1 [running]:
github.com/lima-vm/lima/pkg/driverutil.CreateTargetDriverInstance(...)
	/Users/jan/suse/lima/pkg/driverutil/instance.go:12
github.com/lima-vm/lima/pkg/instance.unregister({0x101940da0, 0x1028a6b00}, 0x140005b4640)
	/Users/jan/suse/lima/pkg/instance/delete.go:36 +0x5c
```

This PR skips unregistering the instance and just deletes the files when validation returned any errors.

A second commit does some minor refactoring I did while investigating this issue.

### DIFF
--- a/pkg/instance/delete.go
+++ b/pkg/instance/delete.go
@@ -21,10 +21,11 @@ func Delete(ctx context.Context, inst *store.Instance, force bool) error {
 
 	StopForcibly(inst)
 
-	if err := unregister(ctx, inst); err != nil {
-		return fmt.Errorf("failed to unregister %q: %w", inst.Dir, err)
+	if len(inst.Errors) == 0 {
+		if err := unregister(ctx, inst); err != nil {
+			return fmt.Errorf("failed to unregister %q: %w", inst.Dir, err)
+		}
 	}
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #3298, opened 2025-03-03
chore: update wsl2 rootfs

arixmkii:
Closes #2781 

The rootfs in template hasn't been updated for some time. Now with integration tests in CI it could be done with increased confidence.


AkihiroSuda (approved):
Thanks, eventually this should be supported in https://github.com/lima-vm/lima/blob/master/hack/update-template.sh

PR #3299, opened 2025-03-03
Use loopback for SSH connection in WSL2

### LINKED ISSUE #3242: [WSL2] getWslSSHAddress would fail on busybox distros
### Description

The command line used is not universally supported: https://github.com/lima-vm/lima/blob/2c84e0c77ab659ae3dcbafa3d199ad6fdaad0772/pkg/store/instance_windows.go#L128

Example output from Alpine:

```
~$ hostname -I
hostname: unrecognized option: I
BusyBox v1.37.0 (2024-12-13 21:18:49 UTC) multi-call binary.

Usage: hostname [-sidf] [HOSTNAME | -F FILE]

Show or set hostname or DNS domain name

        -s      Short
        -i      Addresses for the hostname
        -d      DNS domain name
        -f      Fully qualified domain name
        -F FILE Use FILE's content as hostname
~$ echo $?
1
```

arixmkii:
Fixes #3242

According to WSL2 documentation - both mainstream networking modes NAT and mirrored support access to WSL2 running apps via loopback https://learn.microsoft.com/en-us/windows/wsl/networking#accessing-linux-networking-apps-from-windows-localhost

Other Hyper-V networking modes as bridged and may be some others are somewhat supported by WSL2, but are discouraged for common scenarios. 

In this case there is no point to resolve external IP in a distro specific way to successfully run Lima provisioning of the said VM.

I tested this with NAT in GH actions and with mirrored networking locally.

There are setting for preventing access via loopback but there are also firewalls, which would prevent to access over external IP, so, the loopback sounds like a reasonable default and different settings should only be used if specific scenario is detected or if specifically requested. As these scenarios are not documented as of now the code in question could be removed and later reintroduced adjusted to the new usecases.

### DIFF
--- a/pkg/store/instance_windows.go
+++ b/pkg/store/instance_windows.go
@@ -5,7 +5,6 @@ package store
 
 import (
 	"fmt"
-	"os/exec"
 	"regexp"
 	"strings"
 
@@ -26,7 +25,7 @@ func inspectStatus(instDir string, inst *Instance, y *limayaml.LimaYAML) {
 		inst.SSHLocalPort = 22
 
 		if inst.Status == StatusRunning {
-			sshAddr, err := getWslSSHAddress(inst.Name)
+			sshAddr, err := GetSSHAddress(inst.Name)
 			if err == nil {
 				inst.SSHAddress = sshAddr
 			} else {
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #3304, opened 2025-03-04
Improve Subsystem paths handling on Windows hosts

### LINKED ISSUE #3303: Support configurable WSL mount paths
### Description

Latest `limactl cp` fixes use hardcoded `/mnt` prefix. This should be replaced by the actual mount from the WSL2 machine in use.

https://github.com/lima-vm/lima/pull/3300#discussion_r1978517267

I will work on this.

arixmkii:
Fixes #3303 

Implemented changes 
* hardcoded prefixes remove
* shell now tries to switch directory for WSL2 machines (mounts are controlled outside of Lima, so, we will still try)
* using wslpath inside instance for WSL2 machine to determine mounted path, fallback to cygpath for other cases
* adjusted logic for deciding on home directory on Windows hosts - additional code to erase subsystem mount prefix
* ioutilx APIs now return empty string on error, because returning the original input was sometimes confusing (rarely used approach)

I did local testing with WSL2, plan to test this with QEMU, when I rebase my current experimental branch,

### DIFF
--- a/pkg/osutil/user.go
+++ b/pkg/osutil/user.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os/exec"
 	"os/user"
+	"path"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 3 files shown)


arixmkii:
@jandubois please take a look

jandubois:
> @jandubois please take a look

I'm sorry, I hoped I would be able to review (test, actually) this change before having to leave on a trip, but I've run out of time. Unless some other maintainer picks this up, it will have to wait until I'm back online in April.

AkihiroSuda:
Needs rebase

arixmkii:
Rebased

pkg/osutil/user.go
```diff
@@ -149,7 +150,19 @@ func LimaUser(limaVersion string, warn bool) *user.User {
...
+				drivePath, err := ioutilx.WindowsSubsystemPath(filepath.VolumeName(limaUser.HomeDir) + "/")
+				if err != nil {
+					logrus.Debug(err)
+				} else {
+					prefix := path.Dir(strings.TrimSuffix(drivePath, "/"))
+					if prefix != "/" {
+						home = strings.TrimPrefix(home, prefix)
+					}
+					home += ".linux"
+				}
```
AkihiroSuda, line 165: This complicated logic should have unit tests

AkihiroSuda (approved):
Thanks

pkg/osutil/user.go
arixmkii: I would like to write a test for this one, but this have some heavy dependencies. It needs cygpath, which is mostly fine for Windows specific cases, but for wslpath - one needs WSL distribution as it is available only inside container. 

At some point I had idea to remove this part entirely and rely only on hand coded cygpath alternative in the code below this (the second conditional branch), as for the purpose of creating home directory it was good enough.

I will create a ticket, where I summarize what options are available to improve situation.

PR #3317, opened 2025-03-07
add some doc for linux

ningmingxiao:
I'm new user of lima, It's a very cool project.I used to think it could only be used on macOS .
I update  the read me let user know it can run on linux and can create vm by using qemu

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,36 @@ Set up (on macOS):
 brew install lima
 limactl start
 ```
+on linux:
+See <https://lima-vm.io/docs/installation/> for the further information.
+
+To run vm on linux use qemu:
+```
+limactl start \ 
+  --name=default \ 
+  --cpus=1 \ 
+  --memory=1 \ 
+  --vm-type=qemu \ 
+  --rosetta \
+  --mount-type=virtiofs \
+  --mount-writable \
+  --network=lima:user-v2 \ 
+  template://default
+```
+
+To run vm on macos use vz:
+```
+limactl start \
+  --name=default \
+  --cpus=4 \
+  --memory=8 \
+  --vm-type=vz \
+  --rosetta \
+  --mount-type=virtiofs \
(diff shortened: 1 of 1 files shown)


afbjorklund:
This has moved to the website: https://lima-vm.io/docs/installation/
Though one might add a one-liner or so, saying to install from binaries?

afbjorklund:
All of the other parameters are probably better off with the defaults?

afbjorklund (changes requested):
I don't think we want to duplicate all of that (jq/curl/tar) in the README.

But a line or two on how to install on Linux (link to releases), could be nice.

ningmingxiao:
done @afbjorklund

README.md
```diff
@@ -25,6 +25,36 @@ Set up (on macOS):
...
+See <https://lima-vm.io/docs/installation/> for the further information.
+
+To run vm on linux use qemu:
+```
+limactl start \ 
+  --name=default \ 
+  --cpus=1 \ 
+  --memory=1 \ 
+  --vm-type=qemu \ 
+  --rosetta \
```
AkihiroSuda, line 38: You can't do this

README.md
```diff
@@ -25,6 +25,36 @@ Set up (on macOS):
...
+  --mount-type=virtiofs \
```
AkihiroSuda, line 39: You can't do this on macOS, and typically needs a custom virtiofsd on Linux

README.md
```diff
@@ -25,6 +25,36 @@ Set up (on macOS):
...
+  --mount-writable \
+  --network=lima:user-v2 \ 
+  template://default
+```
+
+To run vm on macos use vz:
```
AkihiroSuda, line 45: This is the default

AkihiroSuda:
This seems redundant. The usage is documented in https://lima-vm.io , and it is linked from `README.md`

PR #3318, opened 2025-03-08
Add {{.TempDir}} and {{.GlobalTempDir}} variables to host templates

jandubois:
Allows writing the `/tmp/lima` mount as

```yaml
mounts:
- location: "{{.Temp}}/lima"
  mountPoint: /tmp/lima
```

Note that on macOS this would use `$TMPDIR` and not `/tmp`, so would be a change in behaviour. It would not affect existing instances though.

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -738,7 +738,7 @@ func FillDefault(y, d, o *LimaYAML, filePath string, warn bool) {
 	location := make(map[string]int)
 	for _, mount := range slices.Concat(d.Mounts, y.Mounts, o.Mounts) {
 		if out, err := executeHostTemplate(mount.Location, instDir, y.Param); err == nil {
-			mount.Location = out.String()
+			mount.Location = filepath.Clean(out.String())
 		} else {
 			logrus.WithError(err).Warnf("Couldn't process mount location %q as a template", mount.Location)
 		}
@@ -955,14 +955,22 @@ func executeHostTemplate(format, instDir string, param map[string]string) (bytes
 	tmpl, err := template.New("").Parse(format)
 	if err == nil {
 		limaHome, _ := dirnames.LimaDir()
+		globalTempDir := "/tmp"
+		if runtime.GOOS == "windows" {
+			// On Windows the global temp directory "%SystemRoot%\Temp" (normally "C:\Windows\Temp")
+			// is not writable by regular users.
+			globalTempDir = os.TempDir()
+		}
 		data := map[string]any{
 			"Dir":  instDir,
 			"Name": filepath.Base(instDir),
 			// TODO: add hostname fields for the host and the guest
-			"UID":   currentUser.Uid,
-			"User":  currentUser.Username,
-			"Home":  userHomeDir,
-			"Param": param,
+			"UID":           currentUser.Uid,
+			"User":          currentUser.Username,
+			"Home":          userHomeDir,
+			"Param":         param,
+			"GlobalTempDir": globalTempDir,
+			"TempDir":       os.TempDir(),
 
 			"Instance": filepath.Base(instDir), // DEPRECATED, use `{{.Name}}`
 			"LimaHome": limaHome,               // DEPRECATED, use `{{.Dir}}` instead of `{{.LimaHome}}/{{.Instance}}`
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"path"
 	"path/filepath"
 	"runtime"
 	"slices"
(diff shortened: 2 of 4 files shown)


afbjorklund:
Do you need to model "the real" $TMPDIR, or should it be something like $LIMA_TEMP instead?

arixmkii:
This made default.yaml compatible with Windows. I managed to create and start VM w/o changes to template.

Output of the mount command:
```
C:\msys64\home\Arthu\gitrepos\lima\_output\bin>limactl shell default mount | find "sshfs"
time="2025-03-08T20:21:48+02:00" level=warning msg="treating lima version \"e0487475.m\" from \"C:\\\\Users\\\\Arthu\\\\.lima\\\\default\\\\lima-version\" as very latest release"
bash: line 1: cd: /mnt/c/msys64/home/Arthu/gitrepos/lima/_output/bin: No such file or directory
:/mnt/c/Users/Arthu on /mnt/c/Users/Arthu type fuse.sshfs (ro,nosuid,nodev,relatime,user_id=1000,group_id=1000,allow_other)
:/mnt/c/Users/Arthu/AppData/Local/Temp/lima on /tmp/lima type fuse.sshfs (rw,nosuid,nodev,relatime,user_id=1000,group_id=1000,allow_other)
```

The `/mnt/` prefixes come from WSL2 Alpine tools container (wslpath instead of cygpath), which I use for running QEMU. I'm yet to figure out required fixes to run it with Git bash msys2 stack (mux modes are not supported in msys2 OpenSSH).

arixmkii:
9p mounts

```
C:\msys64\home\Arthu\gitrepos\lima\_output\bin>limactl shell default mount | find "9p"
time="2025-03-08T20:36:53+02:00" level=warning msg="treating lima version \"e0487475.m\" from \"C:\\\\Users\\\\Arthu\\\\.lima\\\\default\\\\lima-version\" as very latest release"
bash: line 1: cd: /mnt/c/msys64/home/Arthu/gitrepos/lima/_output/bin: No such file or directory
mount0 on /mnt/c/Users/Arthu type 9p (ro,relatime,cache=8f,access=client,msize=131072,trans=virtio)
mount1 on /tmp/lima type 9p (rw,relatime,cache=5,access=client,msize=131072,trans=virtio)
```

jandubois:
I'm going to leave this PR open (and in draft mode) because I think adding the `{{.Temp}}` variable and normalizing mountpoints is something we should eventually merge.

But the change to `templates/default.yaml` seems wrong to me and should not be merged! We need further discussion to figure out how to deal with this. I'll be offline from Mar 13 to 31, so can't participate until next month.

jandubois:
I've rebased this PR on latest `master` to solve the merge conflicts. The updated mount location is now localized to just `template://_default/mounts`.

This PR does introduce a user-visible change, in that `/tmp/lima` inside the VM is now mounted to `$TMPDIR/lima` on a Linux or macOS host. It will not change existing instances.

I'm undecided if this is acceptable or not.

AkihiroSuda (changes requested):
For compatibility `{{ .Temp }}` should use `/tmp` on macOS.
Probably we should have `{{ .TempEnv }}` (or `{{ .TempDynamic }}`) for `$TMPDIR`

jandubois:
> For compatibility `{{ .Temp }}` should use `/tmp` on macOS.

I assume you meant Unix, not macOS. `os.TempDir()` will use `TMPDIR` on Linux too.

> Probably we should have `{{ .TempEnv }}` (or `{{ .TempDynamic }}`) for `$TMPDIR`

How about we use `{{ .Tmp }}` for `/tmp` and `{{ .TmpDir }}` for `$TMPDIR`, to make it easier to remember?

On Windows both variables would be the same.

AkihiroSuda:
> How about we use {{ .Tmp }} for /tmp and {{ .TmpDir }} for $TMPDIR, to make it easier to remember?

To me this sounds like the former one is a file and the latter one is a directory

jandubois:
How about `{{ .GlobalTemp }}` for `/tmp` and `{{ .Temp }}` for `$TMPDIR` then?

The `GlobalTemp` variable really only exists for the sake of the `template://_default/mount` template to maintain backwards compatibility. So I would rather keep `Temp` for the version that users should be using.

AkihiroSuda:
> PS: I think I actually prefer {{ .GlobalTempDir }} and {{ .TempDir }}.

👍

pkg/limayaml/defaults.go
```diff
@@ -955,14 +955,21 @@ func executeHostTemplate(format, instDir string, param map[string]string) (bytes
 	tmpl, err := template.New("").Parse(format)
 	if err == nil {
 		limaHome, _ := dirnames.LimaDir()
+		globalTempDir := "/tmp"
+		if runtime.GOOS == "windows" {
+			// There is no global temp dir on Windows, and the system temp dir is not writable by the user.
```
AkihiroSuda, line 960: "global temp dir" != "system temp dir" ?

```suggestion
			// On Windows, the global temp dir (C:\TEMP) is not writable by the user.
```

AkihiroSuda: Maybe `C:\TEMP` → `C:\Windows\Temp`, idk

jandubois: `C:\TEMP` is not a thing on Windows.

I was talking about `%SystemRoot%\SystemTemp`.

There is normally also a `%SystemRoot%\Temp` directory, but that may also not be writable by regular users. It does seem to be writable by local administrators, even when not running elevated.

I'll update the comment, but will use `C:\Windows\Temp` as the directory name.

AkihiroSuda (approved):
Thanks

PR #3319, opened 2025-03-08
qemu: experimental support for S390X

AkihiroSuda:
Supporting S390X is useful for testing a program with big-endian.


pkg/limayaml/defaults.go
```diff
@@ -64,6 +64,7 @@ func defaultCPUType() CPUType {
 		// Since https://github.com/lima-vm/lima/pull/494, we use qemu64 cpu for better emulation of x86_64.
 		X8664:   "qemu64",
 		RISCV64: "rv64", // FIXME: what is the right choice for riscv64?
+		S390X:   "qemu", // FIXME: what is the right choice for riscv64?
```
arixmkii, line 67: Comment copy-paste issue. Should be "for S390X" probably.

AkihiroSuda: Thanks, fixed

jandubois:
How do you run this? I tried `l start --tty=false --arch s390x` (on macOS 15.3.1 on M1) and get this error in `ha.stderr.log`:

```
{"level":"debug","msg":"qemu[stderr]: qemu-system-s390x: -netdev user,id=net0,net=192.168.5.0/24,dhcpstart=192.168.5.15,hostfwd=tcp:127.0.0.1:60022-:22: Could not set up host forwarding rule 'tcp:127.0.0.1:60022-:22'","time":"2025-03-08T22:29:56-08:00"}
{"level":"info","msg":"Driver stopped due to error: \"exit status 1\"","time":"2025-03-08T22:29:56-08:00"}
```

Kconfig
```diff
@@ -31,6 +31,12 @@ config GUESTAGENT_ARCH_RISCV64
 		Build lima-guestagent for "riscv64" Arch
 	default y
 
+config GUESTAGENT_ARCH_S390X
+	bool "guestagent Arch: s390x"
+	help
+		Build lima-guestagent for "s390x" Arch
+	default y
```
afbjorklund, line 38: you could have this off by default (`n`), if it is experimental

afbjorklund:
I know that there were some people fighting hard to keep both `ppc64le` and `s390x` in Kubernetes (and minikube), but most of the time it was either untested or (possibly) run only in emulation. So many wanted it to be removed, therefore.

Think I had some branch that was able to run them (that I since deleted), there was some tweaks that were needed to the startup parameters and what not. But it seems like there will be a rather narrow audience of these, similar to arm32?

AkihiroSuda:
> How do you run this? I tried `l start --tty=false --arch s390x` (on macOS 15.3.1 on M1) and get this error in `ha.stderr.log`:
> 
> ```
> {"level":"debug","msg":"qemu[stderr]: qemu-system-s390x: -netdev user,id=net0,net=192.168.5.0/24,dhcpstart=192.168.5.15,hostfwd=tcp:127.0.0.1:60022-:22: Could not set up host forwarding rule 'tcp:127.0.0.1:60022-:22'","time":"2025-03-08T22:29:56-08:00"}
> {"level":"info","msg":"Driver stopped due to error: \"exit status 1\"","time":"2025-03-08T22:29:56-08:00"}
> ```

It just works for me in that way, on macOS 15.3.1 with QEMU 9.2.2 on M4 Max.

Maybe something was occupying the port 60022 ?

Kconfig
AkihiroSuda: We have been compiling other experimental archs (riscv64, armv7l) by default.

Let's revisit this in a separate issue/PR.

AkihiroSuda: - https://github.com/lima-vm/lima/issues/3321

templates/centos-stream-10.yaml
AkihiroSuda, line 1: EL8, 9, and 10 do not seem to support `virtio-net-pci` on s390x ? 🤔 


Ubuntu 20.04 (kernel 5.4) still works, so this is a matter of the kernel config rather than the version.

AkihiroSuda: Updated to use `virtio-net-ccw`.

The network per se now works, but SSH fails 🤷‍♂️ 

```console
$ ssh -F ~/.lima/centos-stream-10/ssh.config lima-centos-stream-10
suda@127.0.0.1: Permission denied (publickey,gssapi-keyex,gssapi-with-mic).
```

pkg/qemu/qemu.go
```diff
@@ -927,7 +937,11 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	const serialvChardev = "char-serial-virtio"
 	args = append(args, "-chardev", fmt.Sprintf("socket,id=%s,path=%s,server=on,wait=off,logfile=%s", serialvChardev, serialvSock, serialvLog))
 	// max_ports=1 is required for https://github.com/lima-vm/lima/issues/1689 https://github.com/lima-vm/lima/issues/1691
-	args = append(args, "-device", "virtio-serial-pci,id=virtio-serial0,max_ports=1")
+	serialvMaxPorts := 1
+	if *y.Arch == limayaml.S390X {
+		serialvMaxPorts++ // needed to avoid `virtio-serial-bus: Out-of-range port id specified, max. allowed: 0`
```
jandubois, line 942: This is still confusing to me. Why does it report "max. allowed: 0" when you pass `max_ports=1`?

jandubois (approved):
Thanks, LGTM

However, I only tested the `default.yaml` template.

pkg/qemu/qemu.go
AkihiroSuda: Might be a bug of qemu? idk

AkihiroSuda:
Rebased

afbjorklund (approved):
I'm not sure about making armv7l experimental, but s390x changes look OK to me

AkihiroSuda:
> I'm not sure about making armv7l experimental*, but s390x changes look OK to me
> 
> * "unsupported" is fine, just not very new

armv7l has been already experimental: https://lima-vm.io/docs/releases/experimental/

PR #3323, opened 2025-03-09
cmd: add support for restarting a running instance

### LINKED ISSUE #3125: Command to exit the ssh session
### Description

By default, the ssh connection to the lima instance is kept open between commands. (the `ssh.sock`)

But this also means that you **don't** get a new session, if environment changes or any groups are added.

So if you change the prompt or if you add a docker group, it is not reflected in the ongoing shell.

Historically I just deleted the socket, but it could be nice with a `limactl shell` option to exit nicely?

----

```console
$ ssh -F $(limactl ls --format='{{.SSHConfigFile}}' docker) lima-docker -O exit
Exit request sent.
```

Then the next login will be "fresh" again.

EDIT: There is a function in sshocker do to it, called ssh.ExitMaster

For lima, we might want to restart it again as part of the hostagent?

And to re-establish the reverse-sshfs mounts (if any), and so on (ports).

This could also be useful in other places, like after resume from sleep:

* https://github.com/lima-vm/lima/issues/598

unsuman:
This PR adds a new command `limactl restart INSTANCE` which gracefully terminates and restarts Lima instances.

### DIFF
--- a/cmd/limactl/restart.go
+++ b/cmd/limactl/restart.go
@@ -1,0 +1,52 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"github.com/lima-vm/lima/pkg/instance"
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/spf13/cobra"
+)
+
+func newRestartCommand() *cobra.Command {
+	restartCmd := &cobra.Command{
+		Use:               "restart INSTANCE",
+		Short:             "Restart a running instance",
+		Args:              WrapArgsError(cobra.MaximumNArgs(1)),
+		RunE:              restartAction,
+		ValidArgsFunction: restartBashComplete,
+		GroupID:           basicCommand,
+	}
+
+	restartCmd.Flags().BoolP("force", "f", false, "force stop and restart the instance")
+	return restartCmd
+}
+
+func restartAction(cmd *cobra.Command, args []string) error {
+	instName := DefaultInstanceName
+	if len(args) > 0 {
+		instName = args[0]
+	}
+
+	inst, err := store.Inspect(instName)
+	if err != nil {
+		return err
+	}
+
+	force, err := cmd.Flags().GetBool("force")
+	if err != nil {
+		return err
+	}
+
+	ctx := cmd.Context()
+	if force {
+		return instance.RestartForcibly(ctx, inst)
+	}
+
+	return instance.Restart(ctx, inst)
+}
+
+func restartBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
+	return bashCompleteInstanceNames(cmd)
+}
--- a/pkg/instance/restart.go
+++ b/pkg/instance/restart.go
@@ -1,0 +1,45 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package instance
+
+import (
+	"context"
+
+	networks "github.com/lima-vm/lima/pkg/networks/reconcile"
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/sirupsen/logrus"
+)
+
+const launchHostAgentForeground = false
+
+func Restart(ctx context.Context, inst *store.Instance) error {
+	if err := StopGracefully(inst, true); err != nil {
+		return err
+	}
+
+	if err := networks.Reconcile(ctx, inst.Name); err != nil {
+		return err
+	}
+
+	if err := Start(ctx, inst, "", launchHostAgentForeground); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func RestartForcibly(ctx context.Context, inst *store.Instance) error {
+	logrus.Info("Restarting the instance forcibly")
+	StopForcibly(inst)
+
+	if err := networks.Reconcile(ctx, inst.Name); err != nil {
+		return err
+	}
+
+	if err := Start(ctx, inst, "", launchHostAgentForeground); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/instance/stop.go
+++ b/pkg/instance/stop.go
@@ -19,8 +19,12 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func StopGracefully(inst *store.Instance) error {
+func StopGracefully(inst *store.Instance, isRestart bool) error {
 	if inst.Status != store.StatusRunning {
+		if isRestart {
+			logrus.Warn("The instance is not running, continuing with the restart")
+			return nil
+		}
 		return fmt.Errorf("expected status %q, got %q (maybe use `limactl stop -f`?)", store.StatusRunning, inst.Status)
 	}
 
@@ -31,7 +35,13 @@ func StopGracefully(inst *store.Instance) error {
 	}
 
 	logrus.Info("Waiting for the host agent and the driver processes to shut down")
-	return waitForHostAgentTermination(context.TODO(), inst, begin)
+	err := waitForHostAgentTermination(context.TODO(), inst, begin)
+	if err != nil {
+		return err
+	}
+
+	logrus.Info("Waiting for the instance to shut down")
+	return waitForInstanceShutdown(context.TODO(), inst)
 }
 
 func waitForHostAgentTermination(ctx context.Context, inst *store.Instance, begin time.Time) error {
@@ -62,6 +72,31 @@ func waitForHostAgentTermination(ctx context.Context, inst *store.Instance, begi
 	}
 
 	return nil
+}
+
+func waitForInstanceShutdown(ctx context.Context, inst *store.Instance) error {
+	ctx2, cancel := context.WithTimeout(ctx, 3*time.Minute)
+	defer cancel()
+
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			updatedInst, err := store.Inspect(inst.Name)
+			if err != nil {
+				return errors.New("failed to inspect instance status: " + err.Error())
+			}
+
+			if updatedInst.Status == store.StatusStopped {
+				logrus.Infof("The instance %s has shut down", updatedInst.Name)
+				return nil
+			}
+		case <-ctx2.Done():
+			return errors.New("timed out waiting for instance to shut down after 3 minutes")
+		}
+	}
 }
 
 func StopForcibly(inst *store.Instance) {
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -159,6 +159,7 @@ func newApp() *cobra.Command {
 		newUnprotectCommand(),
 		newTunnelCommand(),
 		newTemplateCommand(),
+		newRestartCommand(),
 	)
 	if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
 		rootCmd.AddCommand(startAtLoginCommand())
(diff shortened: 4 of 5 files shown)


cmd/limactl/shell.go
```diff
@@ -49,6 +55,7 @@ func newShellCommand() *cobra.Command {
 
 	shellCmd.Flags().String("shell", "", "shell interpreter, e.g. /bin/bash")
 	shellCmd.Flags().String("workdir", "", "working directory")
+	shellCmd.Flags().BoolVarP(&restart, "restart", "r", false, "restart a running instance")
```
AkihiroSuda, line 58: Why does this need to be a part of `limactl shell`?

Shouldn't we just have `limactl restart INSTANCE...` ?

cmd/limactl/shell.go
```diff
@@ -81,6 +88,31 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+		if err := ssh.ExitMaster(inst.Hostname, inst.SSHLocalPort, sshConfig); err != nil {
+			return err
+		}
+
+		if err := instance.StopGracefully(inst); err != nil {
+			return err
+		}
+
+		launchHostAgentForeground := false
+		time.Sleep(500 * time.Millisecond)
```
jandubois, line 110: Needing a `sleep` is always a red flag; it means there is a race, and there is still a chance that the sleep is not long enough.

I suspect the `sleep` exists because we need to wait for the PID files to be deleted before we can start the instance again. I wonder if it isn't possible to make sure `StopGracefully` will not return until the files are deleted, and return an error when the files still exist after a somewhat longish timeout.

cmd/limactl/shell.go
jandubois: > Why does this need to be a part of `limactl shell`?

I suspect because @afbjorklund asked for it: "it could be nice with a `limactl shell` option to exit nicely?" (in #3125).

> Shouldn't we just have `limactl restart INSTANCE...` ?

I agree. That's what I would have intuitively expected.

jandubois: BTW, the implementation should be in an `instance.Restart()` function in `pkg/instance/restart.go`, so this functionality is available to applications using Lima as a library and not via `limactl`.

unsuman: > Shouldn't we just have limactl restart INSTANCE... ?

Yeah, that's true. I have implemented a new `limactl restart` command for it.

> BTW, the implementation should be in an instance.Restart() function in pkg/instance/restart.go....

I did it now.

cmd/limactl/shell.go
unsuman: > I suspect the sleep exists because we need to wait for the PID files to be deleted before we can start the instance again.

It is indeed the case, the `StopGracefully()` function is not waiting for the host agent and driver processes to shut down, let me see if I can fix this in this PR itself.

cmd/limactl/shell.go
```diff
@@ -49,6 +53,7 @@ func newShellCommand() *cobra.Command {
 
 	shellCmd.Flags().String("shell", "", "shell interpreter, e.g. /bin/bash")
 	shellCmd.Flags().String("workdir", "", "working directory")
+	shellCmd.Flags().BoolVarP(&exit, "exit", "e", false, "exit a ssh session for the instance")
```
AkihiroSuda, line 56: Better to be a separate PR

cmd/limactl/shell.go
```diff
@@ -33,13 +34,16 @@ A custom ssh alias can be used instead by setting the $` + sshutil.EnvShellSSH +
...
+var exit bool
+
 func newShellCommand() *cobra.Command {
 	shellCmd := &cobra.Command{
 		Use:               "shell [flags] INSTANCE [COMMAND...]",
 		Short:             "Execute shell in Lima",
 		Long:              shellHelp,
 		Args:              WrapArgsError(cobra.MinimumNArgs(1)),
 		RunE:              shellAction,
+		Aliases:           []string{"ssh"},
```
AkihiroSuda, line 46: Better to be a separate PR

cmd/limactl/shell.go
AkihiroSuda: `--exit-session` might be more descriptive?

AkihiroSuda, line 56: For consistency with the other codes, `BoolP` should be used rather than `BoolVarP`

pkg/instance/restart.go
```diff
@@ -0,0 +1,64 @@
...
+	ctx := context.Background()
+	if err := StopGracefully(inst); err != nil {
+		return err
+	}
+
+	if err := waitForInstanceShutdown(inst); err != nil {
+		return err
+	}
+
+	launchHostAgentForeground := false
```
AkihiroSuda, line 21:
```suggestion
	const launchHostAgentForeground = false
```

cmd/limactl/shell.go
unsuman: Opening up a separate PR for this and the above requested code changes, shortly!

cmd/limactl/shell.go
unsuman: Got it!

cmd/limactl/restart.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/spf13/cobra"
+)
+
+func newRestartCommand() *cobra.Command {
+	restartCmd := &cobra.Command{
+		Use:               "restart INSTANCE",
+		Short:             "Restart a running instance",
+		Args:              WrapArgsError(cobra.MinimumNArgs(1)),
+		RunE:              restartAction,
+		Aliases:           []string{"reboot"},
```
AkihiroSuda, line 18: Do we need this alias?

pkg/instance/restart.go
```diff
@@ -0,0 +1,66 @@
...
+	"errors"
+	"time"
+
+	"github.com/lima-vm/lima/pkg/store"
+)
+
+const launchHostAgentForeground = false
+
+func Restart(inst *store.Instance) error {
+	ctx := context.Background()
```
AkihiroSuda, line 17: This should be propagated from `cmd.Context()`?

pkg/instance/restart.go
```diff
@@ -0,0 +1,66 @@
...
+	"context"
+	"errors"
+	"time"
+
+	"github.com/lima-vm/lima/pkg/store"
+)
+
+const launchHostAgentForeground = false
+
+func Restart(inst *store.Instance) error {
```
AkihiroSuda, line 16:
```suggestion
func Restart(ctx context.Context, inst *store.Instance) error {
```

Same for other functions too

pkg/instance/restart.go
```diff
@@ -0,0 +1,66 @@
...
+	ctx := context.Background()
+	if err := StopGracefully(inst); err != nil {
+		return err
+	}
+
+	if err := waitForInstanceShutdown(inst); err != nil {
+		return err
+	}
+
```
AkihiroSuda, line 25: Don't we need `networks.Reconcile` ? (see `stop.go` and `start.go`)

cmd/limactl/restart.go
unsuman: Not neccesarily, but we can implement something that is shorter, what do you say? @AkihiroSuda

pkg/instance/restart.go
unsuman: Yup, my bad, we need to call this function

cmd/limactl/shell.go
afbjorklund: Sorry for being unclear in my issue #3125, but `limactl restart` was just the workaround and not the main issue.

Being able to stop and start in one command, and being able to reconnect a ssh session, are 2 different issues.

pkg/instance/restart.go
```diff
@@ -0,0 +1,73 @@
...
+	if err := Start(ctx, inst, "", launchHostAgentForeground); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func waitForInstanceShutdown(ctx context.Context, inst *store.Instance) error {
```
jandubois, line 51: I think this should be part of `waitForHostAgentTermination` in `pkg/instance/stop.go`, or at least called from `StopGracefully`. That makes it available to an external caller and protects `limactl stop && limactl start` from failing too.

unsuman: The reason I added this function is because I was facing this issue while restarting a VM:
```
FATA[0001] instance "default" seems running (hint: remove "/Users/ansumansahoo/.lima/default/ha.pid" if the instance is not running) 
```
But we don't need this function anymore because `networks.Reconcile()` ensures everything is shutdown before a restart.

cmd/limactl/restart.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"github.com/lima-vm/lima/pkg/instance"
+	"github.com/lima-vm/lima/pkg/store"
+	"github.com/spf13/cobra"
+)
+
+func newRestartCommand() *cobra.Command {
+	restartCmd := &cobra.Command{
+		Use:               "restart INSTANCE",
+		Short:             "Restart a running instance",
```
jandubois, line 15: Since `restartAction` only supports 0 or 1 argument, we should throw an error when the user runs `limactl restart foo bar baz` instead of silently ignoring the additional instances:

```go
Args: WrapArgsError(cobra.MaximumNArgs(1)),
```

pkg/instance/restart.go
jandubois: > But we don't need this function anymore because `networks.Reconcile()` ensures everything is shutdown before a restart.

How does it ensure that? It is not immediately obvious to me. Does it just take sufficient time on your machine, so the problem no longer appears (in which case this would be a variant of the `sleep` mechanism, and not a guarantee)?

jandubois (changes requested):
* Needs checking for maximum number of args
* I don't understand why we don't need to wait for the instance to be stopped
* commits should be squashed

jandubois:
I just tested restarting a stopped instance:

```console
❯ limactl restart
FATA[0000] expected status "Running", got "Stopped" (maybe use `limactl stop -f`?)
```

I think this is correct, you need to specify `limactl restart -f`.

But I wonder if the error message is correct. The user obviously expects the instance to be running when the command returns. So the suggested command should probably be `limactl start` (or `limactl restart -f`).

@lima-vm/maintainers WDYT?

AkihiroSuda:
> I just tested restarting a stopped instance:

It probably should just start the instance automatically, with a warning

pkg/instance/restart.go
unsuman: You are correct; I didn't think of it that way. Thanks! To be on the safer side, I will add the function that ensures the instance has been shutdown before starting it again.

pkg/instance/stop.go
```diff
@@ -19,8 +19,12 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-func StopGracefully(inst *store.Instance) error {
+func StopGracefully(ctx context.Context, inst *store.Instance) error {
 	if inst.Status != store.StatusRunning {
+		if isRestart, ok := ctx.Value("restart").(bool); ok && isRestart {
```
unsuman, line 24: Is this the correct way of doing?

AkihiroSuda: Doesn't seem so? Is there any reason to use ctx here?

unsuman: > I just tested restarting a stopped instance:
> 
> It probably should just start the instance automatically, with a warning

I used `ctx` here to find out who is calling `StopGracefully`. If the instance is being restarted and is already shut down, it won’t cause an error and will continue with starting the instance.

AkihiroSuda: Why not just add an argument to the function?

unsuman:
@AkihiroSuda is there any more work needed for this PR? Could you kindly merge this so that it helps my proposal?

AkihiroSuda (approved):
Thanks

PR #3337, opened 2025-03-11
feat: ask to start instance if not running in limactl shell

### LINKED ISSUE #3336: `limactl shell` should ask whether to start the instance when it is not running
`limactl edit` already has `Do you want to start the instance now?  (Y/n)` prompt.

https://github.com/lima-vm/lima/blob/9fd85199673d46188be880fb921442bc8f9134d2/cmd/limactl/edit.go#L148

This should be available in `limactl shell` (`lima`) too.

tatucosmin:
resolves #3336

`limactl shell <instance_name>` now automatically prompts the user to say if they want to start & connect to the instance via shell

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -15,7 +15,9 @@ import (
 
 	"al.essio.dev/pkg/shellescape"
 	"github.com/coreos/go-semver/semver"
+	"github.com/lima-vm/lima/pkg/instance"
 	"github.com/lima-vm/lima/pkg/ioutilx"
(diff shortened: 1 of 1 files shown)


jandubois (changes requested):
LGTM, but please fix lint error.

I assume it complains about the empty line before the `if err != nil` check.

jandubois (changes requested):
Please squash commits!

tatucosmin:
sorry for the mess here, just started contributing and getting everything right first try is not so easy 😢

jandubois (approved):
Thanks, LGTM

PR #3338, opened 2025-03-11
Allow an image URL instead of a template reference

### LINKED ISSUE #3322: `limactl create URL` should accept image URLs too
### Description

Currently `limactl create URL` only accepts YAML URLs.

It should accept image URLs too. (`*.qcow2`, `*.qcow2.xz`, `*.raw`, `*.img`, ...)

Detecting the file type can be hard, though.

jandubois:
The template.Read() method will guess the architecture based on the image filename and construct a template dynamically.

It also strips generic tags from the image name to generate an instance name candidate.

```console
❯ limactl start --containerd none https://dl-cdn.alpinelinux.org/alpine/v3.21/releases/cloud/nocloud_alpine-3.21.2-aarch64-uefi-cloudinit-r0.qcow2
...

❯ limactl shell alpine-3.21.2-r0 uname -a
Linux lima-alpine-3-21-2-r0 6.12.8-0-virt #1-Alpine SMP PREEMPT_DYNAMIC 2025-01-02 12:14:41 aarch64 Linux
```

Closes #3322

This code needs to be updated with `s390x` heuristics once #3319 is merged (or the other way around, depending on which one is merged first).

### DIFF
--- a/pkg/limatmpl/locator.go
+++ b/pkg/limatmpl/locator.go
@@ -12,11 +12,14 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
+	"runtime"
 	"strings"
 	"unicode"
 
 	"github.com/containerd/containerd/identifiers"
 	"github.com/lima-vm/lima/pkg/ioutilx"
+	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/lima-vm/lima/pkg/templatestore"
 	"github.com/sirupsen/logrus"
 )
@@ -28,6 +31,10 @@ func Read(ctx context.Context, name, locator string) (*Template, error) {
 	tmpl := &Template{
 		Name:    name,
 		Locator: locator,
+	}
+
+	if imageTemplate(tmpl, locator) {
+		return tmpl, nil
 	}
 
 	isTemplateURL, templateURL := SeemsTemplateURL(locator)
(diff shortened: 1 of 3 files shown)


pkg/limatmpl/locator.go
```diff
@@ -121,6 +128,88 @@ func Read(ctx context.Context, name, locator string) (*Template, error) {
...
+	"amd64":   limayaml.X8664,
+	"arm64":   limayaml.AARCH64,
+	"armhf":   limayaml.ARMV7L,
+	"armv7l":  limayaml.ARMV7L,
+	"riscv64": limayaml.RISCV64,
+	"x86_64":  limayaml.X8664,
+}
+
+// These generic tags will be stripped from an image name before turning it into an instance name.
+var genericTags = []string{
```
AkihiroSuda, line 147: Needs unit tests

pkg/limatmpl/locator.go
```diff
@@ -121,6 +128,88 @@ func Read(ctx context.Context, name, locator string) (*Template, error) {
...
+	"kvm",          // Oracle
+	"latest",       // Gentoo, CentOS, Rocky, Alma
+	"minimal",      // openSUSE
+	"openstack",    // Gentoo
+	"server",       // Ubuntu
+	"std",          // Alpine-Lima
+	"stream",       // CentOS
+	"uefi",         // Alpine
+	"vm",           // openSUSE
+}
```
AkihiroSuda, line 164: Maybe`fileName.trimPrefix("nocloud_").splitWords[0]` just suffices

jandubois: I kind of like the more descriptive names. You can always use `--name ubuntu https//...` if you want a really short name or predictable name.

So I would prefer to keep the current logic, unless you really dislike it.

Also don't want to add unit tests if you are going to make me remove it again. 😄

AkihiroSuda: @lima-vm/maintainers Thoughts?

jandubois: > `Arch-Linux-x86_64-cloudimg-20250201.304316.qcow2` may still needs an exceptional rule though, if we want the instance name to be `archlinux`, not `arch`

I just noticed that I didn't add `linux` to the list of generic tags. I would want to strip that too, so the default name for this image would be `arch-20250201.304316`.

jandubois:
Changes:

* Add "linux" to list of generic tags
* Created `InstNameFromImageURL` function and added unit tests

pkg/limatmpl/locator.go
jandubois: I've selected all image names from all our templates and filtered out just the `amd64` ones:

```shell
find templates -name "*.yaml" -exec limactl tmpl yq {} '.images[].location' \; | uniq | xargs basename | sort -u | grep -E 'amd|x86'
```

I've then manually transformed each of them into the instance name produced by the current algorithm:

```
almalinux-8
almalinux-8-8.10-20240819
almalinux-9
almalinux-9-9.5-20241120
alpine-3.21.2-r0
alpine-lima-3.20.3
arch
arch-20250201.304316
centos-10
centos-10-20250210.0
centos-9
centos-9-20250210.0
debian-11
debian-11-20241202-1949
debian-12
debian-12-20250210-2019
debian-sid-daily
fedora-41-1.4
gentoo-openstack
ol8u10-b237
ol9u5-b253
opensuse-leap-15.6
opensuse-tumbleweed
rocky-8
rocky-8-8.10-20240528.0
rocky-9
rocky-9-9.5-20241118.0
ubuntu-20.04
ubuntu-22.04
ubuntu-24.04
ubuntu-24.10
```

I will add `daily` and `openstack` to the list of ignored tags.

But otherwise I think the names look reasonable, and I would prefer them to just picking the first word from the string.

AkihiroSuda (approved):
Thanks

PR #3339, opened 2025-03-11
Do tilde-expansion only once in FillDefaults()

jandubois:
We need to do the expansion before we copy the Location over as the default MountPoint because we can't do tilde expansion for guest filenames (the existing call in cidata.go was an error).

<hr>

This will need to be changed even further when we move `FillDefaults` to use the template embedding code: the mountpoint defaults will need to be filled in before mount entries are combined based on a shared mountpoint value. But that is not yet a concern.


AkihiroSuda (approved):
Thanks, ideally we should have an integration test

PR #3342, opened 2025-03-12
feat: added a default `--yes` flag to set the tty

### LINKED ISSUE #3340: limactl: consider adding `-y` (or `-f`?) as an alias for `--tty=false`
`--tty=false` is too long to type, so it might be useful to add `-y` as an alias.

This alias could be also `-f`, but we may want to reserve it for `limactl SOME_POTENTIAL_FUTURE_SUBCOMMAND --file=FILE`?

suryaaprakassh:
fixes #3340 
- created a new boolean flag `--yes' or '-y' in shorthand
- sets the value of --tty to false if used

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -68,6 +68,7 @@ func newApp() *cobra.Command {
 	rootCmd.PersistentFlags().Bool("debug", false, "debug mode")
 	// TODO: "survey" does not support using cygwin terminal on windows yet
 	rootCmd.PersistentFlags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "Enable TUI interactions such as opening an editor. Defaults to true when stdout is a terminal. Set to false for automation.")
+	rootCmd.PersistentFlags().BoolP("yes", "y", false, "Alias of --tty=false")
 	rootCmd.PersistentPreRunE = func(cmd *cobra.Command, _ []string) error {
 		l, _ := cmd.Flags().GetString("log-level")
 		if l != "" {
@@ -128,6 +129,22 @@ func newApp() *cobra.Command {
 		}
 		if nfs {
 			return errors.New("must not run on NFS dir")
+		}
+
+		if cmd.Flags().Changed("yes") && cmd.Flags().Changed("tty") {
+			return errors.New("cannot use both --tty and --yes flags at the same time")
+		}
+
+		if cmd.Flags().Changed("yes") {
+			// Sets the value of the yesValue flag by using the yes flag.
+			yesValue, _ := cmd.Flags().GetBool("yes")
+			if yesValue {
+				// Sets to the default value false
+				err := cmd.Flags().Set("tty", "false")
+				if err != nil {
+					return err
+				}
+			}
 		}
 		return nil
 	}


cmd/limactl/main.go
```diff
@@ -129,6 +130,17 @@ func newApp() *cobra.Command {
 		if nfs {
 			return errors.New("must not run on NFS dir")
 		}
+
+		// Sets the value of the ttyValue flag by using the yes flag.
+		ttyValue, _ := cmd.Flags().GetBool("yes")
```
AkihiroSuda, line 135:
```suggestion
		yesValue, _ := cmd.Flags().GetBool("yes")
```

cmd/limactl/main.go
```diff
@@ -129,6 +130,17 @@ func newApp() *cobra.Command {
...
+		if ttyValue {
+			// Sets to the default value false
+			err := cmd.Flags().Set("tty", fmt.Sprintf("%v", false))
```
AkihiroSuda, line 138:
```suggestion
			err := cmd.Flags().Set("tty", "false")
```

cmd/limactl/main.go
```diff
@@ -68,6 +68,7 @@ func newApp() *cobra.Command {
 	rootCmd.PersistentFlags().Bool("debug", false, "debug mode")
 	// TODO: "survey" does not support using cygwin terminal on windows yet
 	rootCmd.PersistentFlags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "Enable TUI interactions such as opening an editor. Defaults to true when stdout is a terminal. Set to false for automation.")
+	rootCmd.PersistentFlags().BoolP("yes", "y", isatty.IsTerminal(os.Stdout.Fd()), "Sets the TTY value to false, which is the default behavior. It disables interactive terminal features, making it suitable for automation. When enabled, it allows TUI interactions such as opening an editor")
```
AkihiroSuda, line 71:
```suggestion
	rootCmd.PersistentFlags().BoolP("yes", "y", false, "Alias of --tty=false)

File: cmd/limactl/main.go [hunk 4]
```diff
@@ -129,6 +130,17 @@ func newApp() *cobra.Command {
 		if nfs {
 			return errors.New("must not run on NFS dir")
 		}
+
```
AkihiroSuda, line 133: `cmd.Flags().Change("yes")` should be checked here.

cmd/limactl/main.go
suryaaprakassh: this requested change has been made.

cmd/limactl/main.go
suryaaprakassh: made the changes

cmd/limactl/main.go
suryaaprakassh: the condition to check the flag has been added

jandubois (commented):
I wonder what should happen if the user specifies contradictory settings, like `--tty --yes` or `--tty=false --yes=false`.

You could say a later option should override an earlier one, but I don't know if that can easily be done with Cobra.

An alternative (and I suspect the correct one in this case) would be to throw an error instead of randomly picking one setting over the other.

AkihiroSuda:
> An alternative (and I suspect the correct one in this case) would be to throw an error instead of randomly picking one setting over the other.

:+1: on this

Also, please squash the commits

suryaaprakassh:
> An alternative (and I suspect the correct one in this case) would be to throw an error instead of randomly picking one setting over the other.
changes have been made to throw an error when both `--yes` and `--tty` flag are used simultaneously

cmd/limactl/main.go
```diff
@@ -129,6 +130,22 @@ func newApp() *cobra.Command {
...
+		if cmd.Flags().Changed("yes") {
+			// Sets the value of the yesValue flag by using the yes flag.
+			yesValue, _ := cmd.Flags().GetBool("yes")
+			if yesValue {
+				// Sets to the default value false
+				err := cmd.Flags().Set("tty", "false")
+				if err != nil {
+					return err
+				}
+			}
```
AkihiroSuda, line 147: What if `--yes=false` is explicitly specified?

docs/deprecated.md
```diff
@@ -1 +1,2 @@
 Moved to <https://lima-vm.io/docs/releases/deprecated/>
+
```
AkihiroSuda, line 2: Remove

cmd/limactl/main.go
suryaaprakassh: Setting the value of `--yes=false` will not induce any effects and it will be neglected.  we can just give a warning to user to prevent confusion.

cmd/limactl/main.go
```diff
@@ -68,6 +68,7 @@ func newApp() *cobra.Command {
 	rootCmd.PersistentFlags().Bool("debug", false, "debug mode")
 	// TODO: "survey" does not support using cygwin terminal on windows yet
 	rootCmd.PersistentFlags().Bool("tty", isatty.IsTerminal(os.Stdout.Fd()), "Enable TUI interactions such as opening an editor. Defaults to true when stdout is a terminal. Set to false for automation.")
+	rootCmd.PersistentFlags().BoolP("yes", "y", isatty.IsTerminal(os.Stdout.Fd()), "Alias of --tty=false")
```
AkihiroSuda, line 71: The default value seems conflicting with `tty`

suryaaprakassh: The default value of the `yes` flag is changed to `false` which should behave as expected now.

AkihiroSuda (approved):
Thanks

PR #3347, opened 2025-03-12
Support setting extra entries for PATH env variable on Windows hosts

### LINKED ISSUE #3332: [RFC] [Windows] Consider allow extra path set by limactl to improve Windows UX
### Description

`limactl` on Windows requires additional Unix tools to operate (right now only tested against Git bash). The most straightforward they to use `limactl` is from Git bash, but this is not general purpose shell with no extension points (user is only getting the tools provided with the distribution). For daily work one might better prefer full msys2 with a package manager or Powershell as a more Windows flavored alternative, but then git tools under `/usr/bin` have to be added to PATH and take precedence of the tools provided by the shell. It works well for `limactl`, there are number of inherited problems with this approach:

* with msys2 multiple tools from Git will collide with msys2 and running msys2 userspace applications from different roots is bug prone. The big offender here is `bash`, when I tried it, to preserve env I needed to call it through cmd - `cmd /C "bash -c '<your commands>'"` if I needed to preserve env of my working shell. 
* with Powershell bash again is a problem, because when working in PS one might expect bash to be a command to start shell into default WSL instance. 

To support using these shells and `limactl` with more convenience it is possible to allow `limactl` to read dedicated environment variable and then use `os.Setenv` to update path early in the process startup (before any of the tools are needed). `os.Setenv` is not universally supported across platforms, but works in Windows at least, so, the scope might be narrowed for Windows case.

I created a prototype https://github.com/arixmkii/lima/commit/4cefee402e16b24ce0183f11118749451a93a793 but I want to have a brief discussion if this sounds feasible before making into full fledged PR.

arixmkii:
Fixes #3332 

During startup `limactl.exe` will check new experimental `_LIMA_WINDOWS_EXTRA_PATH` environment variable and when found it will be added to PATH environment variable for the running process.

Variable naming implemented according to this proposal https://github.com/lima-vm/lima/issues/3332#issuecomment-2711587045 and it is explicitly mentioned as "experimental" in the docs.

### DIFF
--- a/website/content/en/docs/config/environment-variables.md
+++ b/website/content/en/docs/config/environment-variables.md
@@ -67,13 +67,27 @@ This page documents the environment variables used in Lima.
   export LIMA_USERNET_RESOLVE_IP_ADDRESS_TIMEOUT=5
   ```
 
-### `LIMA_QEMU_UEFI_IN_BIOS`
+### `_LIMA_QEMU_UEFI_IN_BIOS`
 
 - **Description**: Commands QEMU to load x86_64 UEFI images using `-bios` instead of `pflash` drives.
 - **Default**: `false` on Unix like hosts and `true` on Windows hosts
 - **Usage**: 
   ```sh
-  export LIMA_QEMU_UEFI_IN_BIOS=true
+  export _LIMA_QEMU_UEFI_IN_BIOS=true
   ```
 - **Note**: It is expected that this variable will be set to `false` by default in future
   when QEMU supports `pflash` UEFI for accelerated guests on Windows.
+
+### `_LIMA_WINDOWS_EXTRA_PATH`
+
+- **Description**: Additional directories which will be added to PATH by `limactl.exe` process to search for tools.
+  It is useful, when there is a need to prevent collisions between binaries available in active shell and ones
+  used by `limactl.exe` - injecting them only for the running process w/o altering PATH observed by user shell.
+  Is is Windows specific and does nothing for other platforms.
+- **Default**: unset
+- **Usage**:
+  ```bat
+  set _LIMA_WINDOWS_EXTRA_PATH=C:\Program Files\Git\usr\bin
+  ```
+- **Note**: It is an experimental setting and has no guarantees being ever promoted to stable. It may be removed
+  or changed at any stage of project development.
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -28,6 +28,16 @@ const (
 )
 
 func main() {
+	if runtime.GOOS == "windows" {
+		extras, hasExtra := os.LookupEnv("_LIMA_WINDOWS_EXTRA_PATH")
+		if hasExtra && strings.TrimSpace(extras) != "" {
+			p := os.Getenv("PATH")
+			err := os.Setenv("PATH", strings.TrimSpace(extras)+string(filepath.ListSeparator)+p)
+			if err != nil {
+				logrus.Warning("Can't add extras to PATH, relying entirely on system PATH")
+			}
+		}
+	}
 	if err := newApp().Execute(); err != nil {
 		handleExitCoder(err)
 		logrus.Fatal(err)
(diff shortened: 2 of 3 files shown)


jandubois (dismissed):
Thanks, LGTM

I'll leave it for other maintainers to merge, so they have a chance to object to it.

website/content/en/docs/config/environment-variables.md
```diff
@@ -77,3 +77,15 @@ This page documents the environment variables used in Lima.
   ```
 - **Note**: It is expected that this variable will be set to `false` by default in future
   when QEMU supports `pflash` UEFI for accelerated guests on Windows.
+
+### `_LIMA_WINDOWS_EXTRA_PATH`
```
AkihiroSuda, line 81: Why prepend `_`  ?

website/content/en/docs/config/environment-variables.md
```diff
@@ -77,3 +77,15 @@ This page documents the environment variables used in Lima.
...
+
+- **Description**: Additional directories which will be added to PATH by `limactl.exe` process to search for
+  tools. Is is Windows specific and does nothing for other platforms.
+- **Default**: unset
+- **Usage**:
+  ```bat
+  set _LIMA_WINDOWS_EXTRA_PATH=C:\Program Files\Git\usr\bin
```
AkihiroSuda, line 88: Can't we set the standard `PATH`?

website/content/en/docs/config/environment-variables.md
arixmkii: It was requested by @jandubois in https://github.com/lima-vm/lima/issues/3332#issuecomment-2711587045 to highlight the experimental nature of this setting.

> I would probably prefix the variable with _LIMA instead of LIMA to make it obvious that this is an experimental thing.

website/content/en/docs/config/environment-variables.md
arixmkii: It is complicated. This will work for Lima, but would potentially break or introduce a lot of inconveniences for the user.

I will quote some parts of my original research

* with msys2 multiple tools from Git will collide with msys2 and running msys2 userspace applications from different roots is bug prone. The big offender here is bash, when I tried it, to preserve env I needed to call it through cmd - cmd /C "bash -c '<your commands>'" if I needed to preserve env of my working shell.
* with Powershell bash again is a problem, because when working in PS one might expect bash to be a command to start shell into default WSL instance.

Even going with fine tuning the dependencies it unavoidable to have `ssh.exe` of Git installation for Lima to work properly and this is not suitable for various usecases, where user would expect ssh to call either Windows OpenSSH from Powershell or "native" ssh of their msys2 installation. This would be a bit stretched comparison, but this is alike symlinking ssh from chroot installation to /usr/local/bin in a Linux machine. The configuration the user might not expect and really want to avoid.

I'd rather argue that it will be more convenient to use `limactl` in a separate shell by simply running Git bash and have a separate shell of user preference for other tasks than try to handle it with configuration of PATH variable. Separate shell approach is okay-ish for manual use, but effectively defies any automation or scripted usage attempts.

This very special EXTRA_PATH is a try to find middleground balancing complexity and usage possibilities.

website/content/en/docs/config/environment-variables.md
AkihiroSuda: `LIMA_QEMU_UEFI_IN_BIOS` is also kinda experimental but it doesn't have `_`

AkihiroSuda: Either would be fine, but they have to be consistent

website/content/en/docs/config/environment-variables.md
AkihiroSuda: Thanks, could you add that context to the document?

website/content/en/docs/config/environment-variables.md
arixmkii: Renamed the other one to make it consistent.

website/content/en/docs/config/environment-variables.md
arixmkii: Updated the description. I tried to keep it short and concise, but still deliver enough context

AkihiroSuda (approved):
Thanks

PR #3352, opened 2025-03-15
Start QEMU binary as a background process

### LINKED ISSUE #3221: `limactl stop` kills child processes on Windows preventing shutdown sequence
### Description

Some time ago I reported similar issue with CTRL-C command for containers https://github.com/lima-vm/lima/issues/1284 This was fixed by decoupling long living HA into independent console host (new process group).

Then SIGINT to this process was implemented via sending CTRL-C through syscalls https://github.com/lima-vm/lima/blob/7ce6a7a5d1497753c27c30f8aad8f5861c008da0/pkg/osutil/osutil_windows.go#L35

The problem is that it has QEMU VM and potentially SSH master (this requires additional testing) attached withing the same process group and it will immediately kill QEMU.

It causes a lot of error logs, sometimes results in ERROR CODE 130 from `limactl stop` and with QEMU it causes data loss for virtual hdd changes not flushed.

Possible solutions:

1) start QEMU in new process group on Windows
2) use `qemu-system-x86_64w.exe` (`w` here is for Windows GUI like process, which is automatically detached from console host)

This manifests in integration tests "restart" scenario. File is lost.

I can work on implementation, when decision is made on the best option.

arixmkii:
Fixes #3221

~It also includes some tweaks for GH workflow - preserves cleaner PATH variable for Powershell, where tools will be added by the test runner Git bash itself.~

Removed the change above from this PR because it will be reworked in #3357

### DIFF
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -414,8 +414,7 @@ if [[ -n ${CHECKS["restart"]} ]]; then
 	fi
 
 	INFO "Stopping \"$NAME\""
-	# TODO https://github.com/lima-vm/lima/issues/3221
-	limactl stop "$NAME" || [ "${OS_HOST}" = "Msys" ]
+	limactl stop "$NAME"
 	sleep 3
 
 	if [[ -n ${CHECKS["disk"]} ]]; then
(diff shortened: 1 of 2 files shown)


arixmkii:
Vuln check lists only files unrelated to this PR

arixmkii:
Rebased to fix vuln check issues

AkihiroSuda (approved):
Thanks

PR #3353, opened 2025-03-15
Fix mount support in QEMU on Windows hosts

arixmkii:
On Windows there is native path translation needed from location to Unix path mountPoint. And another one is needed for location before passing it to Unix like tooling (sshfs).

I'm not confident if it is better to show original or translated path in Errors. Having native will reduced the amount of lines changed, but would not represent what was actually passed to the command line. It is impossible to translate it earlier, because it is sshfs specific conversion and, for example, QEMU 9P would expect native Windows path on Windows.

### DIFF
--- a/pkg/hostagent/mount.go
+++ b/pkg/hostagent/mount.go
@@ -7,7 +7,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"runtime"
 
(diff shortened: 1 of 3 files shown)


arixmkii:
Rebased to fix vuln check issues

AkihiroSuda (approved):
Thanks

PR #3356, opened 2025-03-17
Add initial support for DragonFly BSD

tuxillo:
DragonFly BSD also supports nvmm acceleration and lima builds out of
 the box.

Right now it has a problem with acceleration and it only uses TCG by default, see discussion here: https://github.com/lima-vm/lima/issues/892#issuecomment-2729728492


AkihiroSuda (approved):
Thanks

PR #3387, opened 2025-03-27
Add Guest Agent transport setting

### LINKED ISSUE #3177: QEMU: Support feature toggle for GA via serial port

arixmkii:
Fixes #3177

Closes #3255 

Adds setting, which could influence, which will then decide, which transport will be in use for connection to GA

Fixes:
* QEMU always creating ga.sock even when SSH forwarder will be used
* Fallback mode to SSH forwarder was unavailable for VZ

If the unsupported option is detected then a warning is printed and fallback will be used instead. The default is VMType and Host OS sensitive (due to limited support of options for some platforms).

### DIFF
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -45,13 +45,14 @@ type LimaYAML struct {
 	DNS          []net.IP          `yaml:"dns,omitempty" json:"dns,omitempty"`
 	HostResolver HostResolver      `yaml:"hostResolver,omitempty" json:"hostResolver,omitempty"`
 	// `useHostResolver` was deprecated in Lima v0.8.1, removed in Lima v0.14.0. Use `hostResolver.enabled` instead.
-	PropagateProxyEnv    *bool          `yaml:"propagateProxyEnv,omitempty" json:"propagateProxyEnv,omitempty" jsonschema:"nullable"`
-	CACertificates       CACertificates `yaml:"caCerts,omitempty" json:"caCerts,omitempty"`
-	Rosetta              Rosetta        `yaml:"rosetta,omitempty" json:"rosetta,omitempty"`
-	Plain                *bool          `yaml:"plain,omitempty" json:"plain,omitempty" jsonschema:"nullable"`
-	TimeZone             *string        `yaml:"timezone,omitempty" json:"timezone,omitempty" jsonschema:"nullable"`
-	NestedVirtualization *bool          `yaml:"nestedVirtualization,omitempty" json:"nestedVirtualization,omitempty" jsonschema:"nullable"`
-	User                 User           `yaml:"user,omitempty" json:"user,omitempty"`
+	PropagateProxyEnv       *bool            `yaml:"propagateProxyEnv,omitempty" json:"propagateProxyEnv,omitempty" jsonschema:"nullable"`
+	CACertificates          CACertificates   `yaml:"caCerts,omitempty" json:"caCerts,omitempty"`
+	Rosetta                 Rosetta          `yaml:"rosetta,omitempty" json:"rosetta,omitempty"`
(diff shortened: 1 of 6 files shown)


arixmkii:
I manually tested it 
* QEMU on Windows
* QEMU on M4 Mac
* VZ on M4 Mac

pkg/limayaml/limayaml.go
```diff
@@ -45,13 +45,14 @@ type LimaYAML struct {
...
-	CACertificates       CACertificates `yaml:"caCerts,omitempty" json:"caCerts,omitempty"`
-	Rosetta              Rosetta        `yaml:"rosetta,omitempty" json:"rosetta,omitempty"`
-	Plain                *bool          `yaml:"plain,omitempty" json:"plain,omitempty" jsonschema:"nullable"`
-	TimeZone             *string        `yaml:"timezone,omitempty" json:"timezone,omitempty" jsonschema:"nullable"`
-	NestedVirtualization *bool          `yaml:"nestedVirtualization,omitempty" json:"nestedVirtualization,omitempty" jsonschema:"nullable"`
-	User                 User           `yaml:"user,omitempty" json:"user,omitempty"`
+	PropagateProxyEnv       *bool            `yaml:"propagateProxyEnv,omitempty" json:"propagateProxyEnv,omitempty" jsonschema:"nullable"`
+	CACertificates          CACertificates   `yaml:"caCerts,omitempty" json:"caCerts,omitempty"`
+	Rosetta                 Rosetta          `yaml:"rosetta,omitempty" json:"rosetta,omitempty"`
+	GuestAgentTransportType *GATransportType `yaml:"guestAgentTransportType,omitempty" json:"guestAgentTransportType,omitempty" jsonschema:"nullable"`
```
AkihiroSuda, line 51: Maybe we should add a new struct `GuestAgent`?

`GuestInstallPrefix` can be moved there too. (The old form should be supported too for a while)

balajiv113:
@arixmkii 
I understand the intent behind this PR.

But i noticed lots of issues related to qemu serial port. At this point of time I don't think it will work out for all of our cases. Major problem was they don't support multiple instances of sock.

With that being said, do you think having this option helps ??

balajiv113:
This option confuses people more because it has drawbacks based on selection in long run.
For instance, we tried with supporting serialport so that we can support pause resume. Ssh forward cannot be paused and resumed properly.

I think we should derive best mode of transport and use it as default always. Giving these config will confuse folks using it also this transport is internal to lima guest agent

arixmkii:
> Major problem was they don't support multiple instances of sock.

This is serial specifics, somewhat expected.

> With that being said, do you think having this option helps ??

It the only option usable on Windows hosts. Unix socket forwarding with SSH alone is not possible, all additional workarounds on top of SSH will have overhead and be more fragile.

It's having something, which works with limitations vs not having any working solution.

> I think we should derive best mode of transport and use it as default always. 

Sounds reasonable, until one steps on platform specific limitations, having unconditional non-configurable different choices on different platforms might be confusing as well.

Using "hacky" looking `if runtime.GOOS == "windows" {} else {}` here is possible, but it didn't feel right for me.

> Giving these config will confuse folks using it also this transport is internal to lima guest agent

Just the discussion, I would disagree here. The setting is purely HostAgent specific, where it later communicates it to both GA and VM driver and makes decision how to establish its own connection.

balajiv113:
> Using "hacky" looking if runtime.GOOS == "windows" {} else {} here is possible, but it didn't feel right for me.

I agree. Can we think of doing something like below

We can have this config but let do these config population as part of default windows itself and this can be a internal config. Not exposed to user 

This way user doesn't require to configure but as you said we will have meaningful way of deciding on transport medium

arixmkii:
> Not exposed to user

So, the main concern is exposing this to limayaml? Am I getting this right?

We can extend Driver API with SupportedTransports, where the priority based list is returned, this would become platform specific, and then HA would find the topmost suitable from this list, which it can use to configure both GA and the driver. This is a rough idea and needs polishing obviously. WDYT?

PR #3401, opened 2025-04-03
fix: ensure usrlocalsharelima.Dir() works when called from tests

### LINKED ISSUE #3208: usrlocalsharelima.Dir() doesn't work when called from tests
During tests the `os.Executable` is located in a temp directory, and the guestagent binary cannot be located relative to that directory.

This is not an issue right now because we don't call this function from our existing tests.

```
failed to find "lima-guestagent.Linux-aarch64" binary for "/var/folders/ds/jk3wz4n96d54wp16s0tg7d2r0000gn/T/go-build4137918305/b001/limatmpl.test", attempted [/var/folders/ds/jk3wz4n96d54wp16s0tg7d2r0000gn/T/go-build4137918305/b001/lima-guestagent.Linux-aarch64 /var/folders/ds/jk3wz4n96d54wp16s0tg7d2r0000gn/T/go-build4137918305/share/lima/lima-guestagent.Linux-aarch64]
```

Konikz:
Fixes #3208

During tests, `usrlocalsharelima.Dir()` fails to find the guestagent binary because the test executable is in a temp directory. This PR adds a test that verifies the binary can be found in both expected locations:

1. Same directory as the test executable
2. `share/lima` directory relative to the test executable

The test creates dummy binaries in both locations and verifies `Dir()` can find them. This ensures the function works correctly in test environments where `os.Executable` returns a temp directory path.

### DIFF
--- a/pkg/usrlocalsharelima/usrlocalsharelima.go
+++ b/pkg/usrlocalsharelima/usrlocalsharelima.go
@@ -13,14 +13,14 @@ import (
 
 	"github.com/lima-vm/lima/pkg/debugutil"
 	"github.com/lima-vm/lima/pkg/limayaml"
+	. "github.com/lima-vm/lima/pkg/must"
 	"github.com/sirupsen/logrus"
 )
 
+// Cache the executable path at package level.
+var self = Must(os.Executable())
+
 func Dir() (string, error) {
-	self, err := os.Executable()
-	if err != nil {
-		return "", err
-	}
 	selfSt, err := os.Stat(self)
 	if err != nil {
 		return "", err
--- a/pkg/usrlocalsharelima/usrlocalsharelima_test.go
+++ b/pkg/usrlocalsharelima/usrlocalsharelima_test.go
@@ -1,0 +1,49 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package usrlocalsharelima
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestDir(t *testing.T) {
+	// Create a temporary directory for testing
+	tmpDir := t.TempDir()
+
+	// Create the expected directory structure
+	shareDir := filepath.Join(tmpDir, "share", "lima")
+	err := os.MkdirAll(shareDir, 0o755)
+	assert.NilError(t, err)
+
+	// Create a dummy guest agent binary with the correct name format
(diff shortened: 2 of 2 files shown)


jandubois:
There is something wrong with this PR; it adds almost 2000 lines, many unrelated to the issue it is supposed to fix. This may be responsible for the various test failures.


Don't merge `master` into your branch, but rebase on latest `master` if there are merge conflicts.

Please squash commits into a single commit and add a DCO signature.

Thanks!

jandubois:
> There is something wrong with this PR; it adds almost 2000 lines, many unrelated to the issue it is supposed to fix. This may be responsible for the various test failures.

@Konikz Are you planning to fix this PR? There is nothing we can do with it until the extraneous commits are removed and the lint failures addressed.

Konikz:
@jandubois I'll rectify the issue with the pr

pkg/usrlocalsharelima/usrlocalsharelima.go
```diff
@@ -16,8 +16,11 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// executable is a variable that can be overridden in tests.
+var executable = os.Executable
```
jandubois, line 20: I would just cache the return value of the function because it is constant over the lifetime of the process. Import `Must` at the top:

```
import (
    …
    . "github.com/lima-vm/lima/pkg/must"
)
```

and then use

```suggestion
var self = Must(os.Executable())
```

later you can simply assign to it:

```go
    self = limactlPath
```

See `osutil/user.go` for an example of using `Must()`.

jandubois (changes requested):
Thanks, LGTM

I would change the logic to use `Must()`, but feel free to ignore if you like your approach better.

However, you do need to squash commits before we can merge.

pkg/usrlocalsharelima/usrlocalsharelima.go
Konikz: Sir @jandubois ,
I have tried implementing the above suggestions, I need your suggestion with the failing tests.

jandubois:
@Konikz I'm sorry, I kind of lost track of this PR. I just took another look and realized that the fix you have been working on was about testing the `Dir` function itself.

I wrote some more background on why I filed the issue in the first place in https://github.com/lima-vm/lima/issues/3208#issuecomment-2877994243, and have closed it because I don't think there is anything else that can be done about.

I'm closing this PR as well, as I don't think we need to test the `Dir` function; it is well covered by the integration tests, that would fail if this function didn't work.

Thank you for working on this, and sorry for not realizing this issue earlier!

PR #3405, opened 2025-04-04
Remove redundant major version number from instance name

jandubois:
almalinux and rocky use image names that include the major version number twice. This commit strips it:

```
almalinux-8-8.10 → almalinux-8.10
almalinux-9-9.5  → almalinux-9.5
rocky-8-8.10     → rocky-8.10
rocky-9-9.5      → rocky-9.5
```


AkihiroSuda (approved):
thanks

PR #3413, opened 2025-04-08
CI: fix the commit hash of golangci-lint-action

AkihiroSuda:
Probably a wrong hash for v7.0.0 was accidentally introduced in https://github.com/lima-vm/lima/pull/3330#discussion_r2023091814 , unless v7.0.0 was retagged.

Anyway, there was no security concern in the previous hashes AFAICS.


alexandear (approved):
LGTM. Thank you.

PR #3419, opened 2025-04-10
Fix slow image pull caused by delayed DNS response

Nino-K:
When IPv6 is disabled, the hostAgent enters a code path that introduces a 1-second sleep whenever it receives an
`AAAA` DNS query. This delay can cause the DNS query to timeout, triggering a retry loop on the client side. Docker, for example, retries up to 10 times at 1-second intervals—resulting in a significant delay before an image is pulled.

A suggested workaround is to remove the sleep, which was originally introduced to address a specific issue that now appears to be outdated or no longer relevant.

[Unexpected A record DNS response](https://github.com/rancher-sandbox/rancher-desktop/issues/3447#issuecomment-1330876897)

### DIFF
--- a/pkg/hostagent/dns/dns.go
+++ b/pkg/hostagent/dns/dns.go
@@ -11,18 +11,14 @@ import (
 	"runtime"
 	"strconv"
 	"strings"
-	"time"
 
 	"github.com/miekg/dns"
 	"github.com/sirupsen/logrus"
 )
 
-const (
-	// Truncate for avoiding "Parse error" from `busybox nslookup`
-	// https://github.com/lima-vm/lima/issues/380
-	truncateSize      = 512
-	ipv6ResponseDelay = time.Second
-)
+// Truncate for avoiding "Parse error" from `busybox nslookup`
+// https://github.com/lima-vm/lima/issues/380
+const truncateSize = 512
 
 var defaultFallbackIPs = []string{"8.8.8.8", "1.1.1.1"}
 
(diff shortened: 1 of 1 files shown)


jandubois:
With current `master` it takes 7-8s before `docker pull` starts fetching an image:

```console
❯ limactl start -y --vm-type qemu template://docker
…
❯ docker pull python | ts -s
00:00:00 Using default tag: latest
00:00:05 latest: Pulling from library/python
00:00:07 71daa2c787b0: Pulling fs layer
…
00:00:30 Status: Downloaded newer image for python:latest
00:00:30 docker.io/library/python:latest
```

With this PR the initial delay is reduced to 1-2s:

```console
❯ docker pull python | ts -s
00:00:00 Using default tag: latest
00:00:02 latest: Pulling from library/python
00:00:02 71daa2c787b0: Pulling fs layer
…
00:00:23 Status: Downloaded newer image for python:latest
00:00:23 docker.io/library/python:latest
```

jandubois (approved):
Thanks, LGTM

I believe it is fine to remove this code, but I would like to see a second approval from another @lima-vm/maintainers

pkg/hostagent/dns/dns.go
```diff
@@ -166,22 +166,6 @@ func (h *Handler) handleQuery(w dns.ResponseWriter, req *dns.Msg) {
...
-				// the correct response even when both Transaction IDs are the same.
-				time.Sleep(ipv6ResponseDelay)
-				// See RFC 2308 section 2.2 which suggests that NODATA is indicated by setting the
-				// RCODE to NOERROR along with zero entries in the response.
-				reply.SetRcode(req, dns.RcodeSuccess)
-				reply.SetReply(req)
-				handled = true
-				break
-			}
-			fallthrough
```
nhinds, line 184: AAAA records previously fell through to the next block when `h.ipv6` is `true`, but with this change they will now skip down to `h.handleDefault(w, req)`, bypassing the `lookupCnameToHost` handling.

`dns.TypeAAAA` should probably still be handled here - or if the intention was to remove `AAAA` handling completely even if IPv6 is enabled, the `else if qtype == dns.TypeAAAA && ipv6` block should be removed from the `case dns.TypeA` section below

Nino-K:
@AkihiroSuda @nhinds, please take a look. I've changed the approach—in the new version, I ended up removing only the sleep since that was the real culprit, not the `NOERROR` response code with zero response.

On another note, my previous commit left the `ipv6ResponseDelay = time.Second` (it was not being used by anything) in place, but the linter didn't flag it. Just wanted to mention it in case this might require an additional linter rule to be enabled.

alexandear (approved):

jandubois:
> On another note, my previous commit left the `ipv6ResponseDelay = time.Second` (it was not being used by anything) in place, but the linter didn't flag it. Just wanted to mention it in case this might require an additional linter rule to be enabled.

@alexandear Any idea why the linter didn't notice this?

jandubois (approved):

Issue #3418, opened 2025-04-10
document how to increase disk size


shenki:
### Description

I've been using a Debian vm under Lima on OSX as a development environment and it's been working fine for a few months. I've filled the disk, and needed more space.

I was confused when `limactl disk` didn't show any output:
```
$ limactl disk list
WARN[0000] No disk found. Run `limactl disk create DISK --size SIZE` to create a disk.
NAME    SIZE    FORMAT    DIR    IN-USE-BY
```

I have disks:
```
$ ls -lhSn
total 195864888
-rw-------@ 1 502  20   100G 10 Apr 11:01 diffdisk
-rw-r--r--@ 1 502  20   328M 23 Jan 16:29 basedisk
```
```
$ qemu-img info diffdisk
image: diffdisk
file format: raw
virtual size: 100 GiB (107374182400 bytes)
disk size: 92.8 GiB
Child node '/file':
    filename: diffdisk
    protocol type: file
    file length: 100 GiB (107374182400 bytes)
    disk size: 92.8 GiB
```

Looking around there's some discussion on the steps required https://github.com/lima-vm/lima/discussions/1446 https://github.com/lima-vm/lima/issues/1448. I went ahead and stopped my VM, rewized it with qemu-img, and started it back up. All was well!

```
qemu-img resize diffdisk 200G
WARNING: Image format was not specified for 'diffdisk' and probing guessed raw.
         Automatically detecting the format is dangerous for raw images, write operations on block 0 will be restricted.
         Specify the 'raw' format explicitly to remove the restrictions.
Image resized.
```
(I assume if I'd use `-t raw` that wouldn't have warned me)

```
 kernel: virtio_blk virtio2: 1/0/0 default/read/poll queues
 kernel: virtio_blk virtio2: [vda] 419430400 512-byte logical blocks (215 GB/200 GiB)
 kernel: GPT:Primary header thinks Alt. header is not at the end of the disk.
 kernel: GPT:209715199 != 419430399
 kernel: GPT:Alternate GPT header not at the end of the disk.
 kernel: GPT:209715199 != 419430399
 kernel: GPT: Use GNU Parted to correct GPT errors.
 kernel:  vda: vda1 vda15
```

I didn't need to resize the partition, it was taken care of by `systemd-growfs`:
```
 systemd[1]: Starting systemd-growfs-root.service - Grow Root File System...
 kernel: EXT4-fs (vda1): resizing filesystem from 26181627 to 52396027 blocks
[...]

PR #3420, opened 2025-04-10
add docs: configure existing VM

songponssw:
Support #3418 

I draft the docs for the "configure existing VM" inside the Example section. 
Let's me know, if there's anything you'd like to add.


songponssw:
My bad! I just realized that what I documented only updates the YAML and doesn't actually resize the disk
I'm going to close this PR.

PR #3423, opened 2025-04-11
Capitalize Command & Command Flag Descriptions (for consistency and clarity)

apachex692:
Hi team, I've fixed some CLI-related sentence casing in this commit. I've tested the codebase before pushing and everything seems to work fine.

### DIFF
--- a/cmd/lima-guestagent/daemon_linux.go
+++ b/cmd/lima-guestagent/daemon_linux.go
@@ -21,12 +21,12 @@ import (
 func newDaemonCommand() *cobra.Command {
 	daemonCommand := &cobra.Command{
 		Use:   "daemon",
-		Short: "run the daemon",
+		Short: "Run the daemon",
 		RunE:  daemonAction,
 	}
-	daemonCommand.Flags().Duration("tick", 3*time.Second, "tick for polling events")
-	daemonCommand.Flags().Int("vsock-port", 0, "use vsock server instead a UNIX socket")
-	daemonCommand.Flags().String("virtio-port", "", "use virtio server instead a UNIX socket")
+	daemonCommand.Flags().Duration("tick", 3*time.Second, "Tick for polling events")
+	daemonCommand.Flags().Int("vsock-port", 0, "Use vsock server instead a UNIX socket")
+	daemonCommand.Flags().String("virtio-port", "", "Use virtio server instead a UNIX socket")
 	return daemonCommand
 }
 
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -24,7 +24,7 @@ func RegisterEdit(cmd *cobra.Command) {
 func registerEdit(cmd *cobra.Command, commentPrefix string) {
 	flags := cmd.Flags()
 
-	flags.Int("cpus", 0, commentPrefix+"number of CPUs") // Similar to colima's --cpu, but the flag name is slightly different (cpu vs cpus)
+	flags.Int("cpus", 0, commentPrefix+"Number of CPUs") // Similar to colima's --cpu, but the flag name is slightly different (cpu vs cpus)
 	_ = cmd.RegisterFlagCompletionFunc("cpus", func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
 		var res []string
 		for _, f := range completeCPUs(runtime.NumCPU()) {
@@ -33,9 +33,9 @@ func registerEdit(cmd *cobra.Command, commentPrefix string) {
 		return res, cobra.ShellCompDirectiveNoFileComp
 	})
 
-	flags.IPSlice("dns", nil, commentPrefix+"specify custom DNS (disable host resolver)") // colima-compatible
+	flags.IPSlice("dns", nil, commentPrefix+"Specify custom DNS (disable host resolver)") // colima-compatible
 
-	flags.Float32("memory", 0, commentPrefix+"memory in GiB") // colima-compatible
+	flags.Float32("memory", 0, commentPrefix+"Memory in GiB") // colima-compatible
 	_ = cmd.RegisterFlagCompletionFunc("memory", func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
 		var res []string
 		for _, f := range completeMemoryGiB(memory.TotalMemory()) {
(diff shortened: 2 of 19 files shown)


jandubois (dismissed):
Thanks, LGTM

I do notice that some tools (e.g. `helm`) do not capitalize the option descriptions, probably because they are not full sentences. But others (e.g. `docker`, `gh`) do, and I think it does look better.

I was also hesitant about the change `ssh` → `SSH` because I thought of it as a command name, and not the technology, but I guess in that case it should have been in backticks.

So I'm fine with merging as-is, but will give other @lima-vm/maintainers a chance to object...

cmd/limactl/main.go
```diff
@@ -75,7 +75,7 @@ func newApp() *cobra.Command {
 	}
 	rootCmd.PersistentFlags().String("log-level", "", "Set the logging level [trace, debug, info, warn, error]")
 	rootCmd.PersistentFlags().String("log-format", "text", "Set the logging format [text, json]")
-	rootCmd.PersistentFlags().Bool("debug", false, "debug mode")
+	rootCmd.PersistentFlags().Bool("debug", false, "Debug mode")
 	// TODO: "survey" does not support using cygwin terminal on windows yet
```
AkihiroSuda, line 79: These ones are not capitalized though

```
  -h, --help                help for limactl
```

```
  -v, --version             version for limactl
```

cmd/limactl/hostagent.go
```diff
@@ -24,15 +24,15 @@ import (
...
 		Args:   WrapArgsError(cobra.ExactArgs(1)),
 		RunE:   hostagentAction,
 		Hidden: true,
 	}
-	hostagentCommand.Flags().StringP("pidfile", "p", "", "write pid to file")
-	hostagentCommand.Flags().String("socket", "", "hostagent socket")
-	hostagentCommand.Flags().Bool("run-gui", false, "run gui synchronously within hostagent")
-	hostagentCommand.Flags().String("nerdctl-archive", "", "local file path (not URL) of nerdctl-full-VERSION-GOOS-GOARCH.tar.gz")
+	hostagentCommand.Flags().StringP("pidfile", "p", "", "Write PID to file")
+	hostagentCommand.Flags().String("socket", "", "Hostagent socket")
```
nirs, line 33: `hostagent` is a name of the process, so we should not call it Hostagent. We can chage to "The hostagent socket" or "Path of hostagent socket.

AkihiroSuda:
I don't see a "typo"

nirs (commented):
Common tools like kubectl do use sentence case, is looks more natural, so +1.

cmd/limactl/main.go
nirs: This comes from Cobra. I think the only way to change is to use our own usage template which is not worth it.

If want consistent output we should use Cobra style. I don't think Cobra will change this since it will break all project using Cobra default style. But maybe Cobra can be extended to support sentence case as Command option.

apachex692: I agree, I couldn't find a way to update it other than the way suggested by @nirs.

apachex692:
> Thanks, LGTM
> 
> I do notice that some tools (e.g. `helm`) do not capitalize the option descriptions, probably because they are not full sentences. But others (e.g. `docker`, `gh`) do, and I think it does look better.
> 
> I was also hesitant about the change `ssh` → `SSH` because I thought of it as a command name, and not the technology, but I guess in that case it should have been in backticks.
> 
> So I'm fine with merging as-is, but will give other @lima-vm/maintainers a chance to object...

You're right about the SSH command, I oversaw it. Will revert it back to what it was.

apachex692:
Hi team, I've incorporated the requested changes. Please do the final review. Thanks! 🚁

cmd/lima-guestagent/daemon_linux.go
```diff
@@ -21,12 +21,12 @@ import (
 func newDaemonCommand() *cobra.Command {
 	daemonCommand := &cobra.Command{
 		Use:   "daemon",
-		Short: "run the daemon",
+		Short: "Run the daemon",
 		RunE:  daemonAction,
 	}
 	daemonCommand.Flags().Duration("tick", 3*time.Second, "tick for polling events")
```
alexandear, line 27:
```suggestion
	daemonCommand.Flags().Duration("tick", 3*time.Second, "Tick for polling events")
```

alexandear:
Could we capitalize descriptions for the `--version` and `--help` flags as well?

```
$ go run ./cmd/... -h
```

image

Also, please update flags descriptions for the `limactl sudoers` command.

apachex692:
https://github.com/lima-vm/lima/pull/3423#discussion_r2039846785

apachex692:
Hi there, squashed the commits and added DCO to commit message. Please review and merge.

jandubois:
> Could we capitalize descriptions for the `--version` and `--help` flags as well?

This was already discussed above and is not really possible because the descriptions are hard-coded inside the Cobra internals. The only way to override it seems to hide the generated `--help` and `--version` commands, and replacing them with our own. Which would be a big effort to provide the same specialized behaviour.

The only other idea I can come up with would be to define a custom help function, call the original help, and then use regex text substitution on the output. Which is rather hacky, and also maybe too much for this.

So I think for now we should ignore these 2 flags. Maybe open an issue/PR against Cobra to make this easier.

apachex692:
I'll take the task of fixing this issue with Cobra.

jandubois (approved):
I still see the uppercase `SSH` in the changes, but I don't really care either way.

Approving, but letting somebody else merge it, in case there is still more feedback.

cmd/limactl/editflags/editflags.go
```diff
@@ -24,7 +24,7 @@ func RegisterEdit(cmd *cobra.Command) {
 func registerEdit(cmd *cobra.Command, commentPrefix string) {
 	flags := cmd.Flags()
 
-	flags.Int("cpus", 0, commentPrefix+"number of CPUs") // Similar to colima's --cpu, but the flag name is slightly different (cpu vs cpus)
+	flags.Int("cpus", 0, commentPrefix+"Number of CPUs") // Similar to colima's --cpu, but the flag name is slightly different (cpu vs cpus)
```
nirs, line 27: Here we concat commentPrefix and a sentence. When commentPrfix is empty this make sense, but what is the other possible values? Can you share the output with all possible values of commentPrefix?

nirs (commented):
I'm not sure about the options using commentPrefix.

I think the rule for SSH is to use "SSH" when we refer to the protocol and "ssh command" when we refer to the actual command. I did not check the actual usage.

Otherwise this looks good.

cmd/limactl/editflags/editflags.go
jandubois: It is used to show the `limactl create` flags in the `limactl start --help` output:

```
Flags:
      --arch string         [limactl create] Machine architecture (x86_64, aarch64, riscv64, armv7l, s390x)
      --containerd string   [limactl create] containerd mode (user, system, user+system, none)
      --cpus int            [limactl create] Number of CPUs
      --disk float32        [limactl create] Disk size in GiB
      --dns ipSlice         [limactl create] Specify custom DNS (disable host resolver) (default [])
      --foreground          Run the hostagent in the foreground
  -h, --help                help for start
```

I think the uppercase variant looks better here too.

I feel like the `[limactl create]` prefix takes up too much space, but that is outside the scope of this PR.

jandubois:
```
 --containerd string   [limactl create] containerd mode (user, system, user+system, none)
```

This looks a bit odd now, so maybe it should be prefixed with "The"?

```
 --containerd string   [limactl create] The containerd mode (user, system, user+system, none)
```

alexandear (approved):

apachex692:
Thank you everyone for the help. Let me know if I should close the PR.

cmd/limactl/editflags/editflags.go
nirs: Yes it looks better. The ‘[limactl create]’ part is very confusing and unhelpful but it not related to this change.

nirs (approved):

PR #3428, opened 2025-04-14
Allow drivers to negotiate transport independent of hostagent

### LINKED ISSUE #3177: QEMU: Support feature toggle for GA via serial port

arixmkii:
Fixes #3177

Closes #3387

Hostagent will agree to whatever driver returns as a transport via BaseDriver interface.

This is re-implemented concept from previous one. Now it follows the maintainer recommendation in https://github.com/lima-vm/lima/pull/3387#issuecomment-2760002905

> I think we should derive best mode of transport and use it as default always. Giving these config will confuse folks using it also this transport is internal to lima guest agent

The change required was to move the configuration code from hostagent into drivers themselves to populate and then make hostagent the settings consumer.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -123,20 +123,13 @@ func New(instName string, stdout io.Writer, signalCh chan os.Signal, opts ...Opt
 		}
 	}
 
-	vSockPort := 0
-	virtioPort := ""
-	if *inst.Config.VMType == limayaml.VZ {
-		vSockPort = 2222
-	} else if *inst.Config.VMType == limayaml.WSL2 {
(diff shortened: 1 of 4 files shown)


arixmkii:
@balajiv113 could you take a look at this? I believe it really close to what you expected as the desired implementation.

arixmkii:
The current implementation is not much different from the original, but the idea is that later, with the drivers as plugins, they would be able to have their own config mechanisms, which will not be limayaml based and the transport type selection could be added at that later point w/o adding user facing complexity in limayaml.

balajiv113 (approved):
These changes LGTM 👍 

Thanks for this change this will help us in long run to support drivers as plugins.

PR #3434, opened 2025-04-16
templates: add fedora-42

AkihiroSuda:
`template://fedora` still refers to Fedora 41, as Fedora 42 does not work on Intel Mac with vz (issue 3334)


jandubois (approved):
Thanks, LGTM

PR #3435, opened 2025-04-16
rootless: allow UID >= 524288

AkihiroSuda:
When the UID is >= 524288, newuidmap was failing.

```
[rootlesskit:parent] error: failed to setup UID/GID map:
newuidmap 3326 [0 222586300 1 1 524288 1073741824] failed:
newuidmap: write to uid_map failed: Invalid argument:
exit status 1
```

Fix moby/moby#49810


jandubois (approved):
Thanks, LGTM

PR #3437, opened 2025-04-17
Fix: resize GuestOS disk

songponssw:
Try to fix: #1448 

**How Lima edit disk**
- The `limactl edit` updates the disk value to the store. It does not resize disk image. Users have to manually resize by `qemu-img`
- Does not support --shrink (fails to start instance)

**What I did**
- create a `limactl disk add` command to add an existing qcow2 disk to Lima.  
- To update GuestOS disk, before start instance, It adds a `diffdisk` of the instance to Lima, resize, and delete.

PR #3439, opened 2025-04-17
Support: limactl disk import command

songponssw:
Split PR from #3437 
Support disk command to add an existing disk to Lima. 

### **Feedback from #3437 PR**
**Cobra command**
  - [x] Fix syntax on `Use`
  - [x] Edit `Short` description
  - [x] Passing as argument instead of using flags
  - [x] Alias add/import

**diskAddAction()**
- [x] Check diskName already taken.
- [x] Use contfs.CopyFile
- [x] Edit log message

### DIFF
--- a/cmd/limactl/disk.go
+++ b/cmd/limactl/disk.go
@@ -9,12 +9,16 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"path/filepath"
 	"text/tabwriter"
 
+	contfs "github.com/containerd/continuity/fs"
 	"github.com/docker/go-units"
+	"github.com/lima-vm/go-qcow2reader"
 	"github.com/lima-vm/lima/pkg/nativeimgutil"
 	"github.com/lima-vm/lima/pkg/qemu"
 	"github.com/lima-vm/lima/pkg/store"
+	"github.com/lima-vm/lima/pkg/store/filenames"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
@@ -44,6 +48,7 @@ func newDiskCommand() *cobra.Command {
 		newDiskDeleteCommand(),
 		newDiskUnlockCommand(),
 		newDiskResizeCommand(),
+		newDiskImportCommand(),
 	)
 	return diskCommand
 }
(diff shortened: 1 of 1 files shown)


cmd/limactl/disk.go
```diff
@@ -418,3 +422,64 @@ func diskResizeAction(cmd *cobra.Command, args []string) error {
 func diskBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
 	return bashCompleteDiskNames(cmd)
 }
+
+func newDiskAddCommand() *cobra.Command {
+	diskAddCommand := &cobra.Command{
+		Use: "add DISK DISKPATH",
```
AkihiroSuda, line 428:
```suggestion
		Use: "add DISK FILE",
```

cmd/limactl/disk.go
```diff
@@ -418,3 +422,64 @@ func diskResizeAction(cmd *cobra.Command, args []string) error {
...
+		Example: `
+Add a disk:
+$ limactl disk add DISK DISKPATH
+`,
+		Short:   "Add an existing disk to Lima",
+		Aliases: []string{"import"},
```
AkihiroSuda, line 434: I'm not sure if we want to add this alias, as potentially we may pick "import" for a different command.

We just have to decide whether the command name should be "add", "import", or something else.

AkihiroSuda:
Is this still draft?

AkihiroSuda:
Please squash the commits and fix the PR title

cmd/limactl/disk.go
songponssw: Changing to "import" and remove the Aliases

songponssw:
Thanks for your suggestion @AkihiroSuda - I've updated the PR name. 
It's ready for review.

AkihiroSuda (dismissed):
Thanks

cmd/limactl/disk.go
```diff
@@ -418,3 +422,63 @@ func diskResizeAction(cmd *cobra.Command, args []string) error {
...
+	case "qcow2", "raw":
+	default:
+		return fmt.Errorf(`disk format %q not supported, use "qcow2" or "raw" instead`, format)
+	}
+
+	if err := os.MkdirAll(diskDir, 0o755); err != nil {
+		return err
+	}
+
+	if err := contfs.CopyFile(filepath.Join(diskDir, "datadisk"), fName); err != nil {
```
AkihiroSuda, line 477: This constant can be used
https://github.com/lima-vm/lima/blob/f3cbd68a97d64f0c71f56ccf7213c944cdf23f6c/pkg/store/filenames/filenames.go#L76

songponssw: Update `datadisk` to `filenames.DataDisk`

AkihiroSuda (approved):
Thanks

Issue #1184, opened 2022-11-20
Multi-node kubernetes clusters


afbjorklund:
### Description

Currently the lima clusters, in k3s.yaml and k8s.yaml, set up a single-node cluster.

If networking between two instances is set up, it is possible to add additional nodes.

Discussion:
* #1181
Initial PR:
* #

afbjorklund:
The template will have a "boolean", if it is a control plane node or a worker node.
(In order to add worker nodes, you will first need to have set up a control plane)

In k3s these are called "server" and "agent", since they run more than just kubernetes.
But the same concept applies. In k8s, they were previously called "master" and "minion".

afbjorklund:
The "k8s.sh" script will follow the same pattern as the "k3s.sh" script is doing...

If there is a K8S_TOKEN, it is a worker node. Otherwise, it is a control plane node.

afbjorklund:
I updated the k3d and kind examples, as an alternative to running multiple virtual machines.

afbjorklund:
The main issue with the virtual machines is the networking and setting the node IP for each.

It also requires more resources, compared with overprovisioning containers on the same host...

PR #3448, opened 2025-04-20
Adding ip address to the SAN certificate for  k8s.yaml

echo-devnull:
Adding this ip will make software trust the api server since the ip address of the api server will be in the SAN certificate list

### DIFF
--- a/templates/k8s.yaml
+++ b/templates/k8s.yaml
@@ -112,6 +112,7 @@ provision:
     apiServer:
       certSANs: # --apiserver-cert-extra-sans
       - "127.0.0.1"
+      - "192.168.5.15"
     networking:
       podSubnet: "10.244.0.0/16" # --pod-network-cidr
     ---


afbjorklund:
The default address of the apiserver is `localhost`, even though the node IP is set to the first interface.

```bash
    # Replace the server address with localhost, so that it works also from the host
    sed -e "/server:/ s|https://.*:\([0-9]*\)$|https://127.0.0.1:\1|" -i $KUBECONFIG
```

When doing multiple k8s nodes we need to fix the networking anyway, since the NAT IP is **not** unique.

* https://github.com/lima-vm/lima/issues/1184

When would the IP be used directly, not kubeconfig?

Why do you need to add it to "extra" SANs, explicitely?

`[certs] apiserver serving cert is signed for DNS names [kubernetes kubernetes.default kubernetes.default.svc kubernetes.default.svc.cluster.local lima-k8s] and IPs [10.96.0.1 192.168.5.15 127.0.0.1]`

templates/k8s.yaml
```diff
@@ -112,6 +112,7 @@ provision:
     apiServer:
       certSANs: # --apiserver-cert-extra-sans
       - "127.0.0.1"
+      - "192.168.5.15"
```
afbjorklund, line 115: This should not be needed (and should not be hardcoded)

afbjorklund (changes requested):

PR #3451, opened 2025-04-21
Deprecate the ansible provision mode and playbook

### LINKED ISSUE #3450: Deprecate the ansible provision playbook
Currently we have an "ansible" type of provision, that runs a playbook from an inventory.

It would be better for the user to run these outside of Lima, after the instance is created.

If really wanting to use Ansible for VM provisioning, one could start it from a shell script?

Generating an inventory is not needed when run locally, with `ansible_connection=local`

afbjorklund:
Now issues a warning when validated, to be removed completely in a future version.

Closes #3450


pkg/limayaml/validate.go
```diff
@@ -277,6 +277,7 @@ func Validate(y *LimaYAML, warn bool) error {
 			if _, err := os.Stat(playbook); err != nil {
 				return fmt.Errorf("field `provision[%d].playbook` refers to an inaccessible path: %q: %w", i, playbook, err)
 			}
+			logrus.Warnf("ansible provision mode is deprecated, use `ansible-playbook %q` instead", playbook)
```
jandubois, line 280: I'm not sure it is clear to the user that `ansible-playbook` should be run directly (on the host? inside the VM?), and that it isn't a new `lima.yaml` setting.

Nit-picking: `provision mode "ansible"`

jandubois (approved):
Thanks, LGTM

pkg/limayaml/limayaml.go
```diff
@@ -221,7 +221,7 @@ const (
 	ProvisionModeUser       ProvisionMode = "user"
 	ProvisionModeBoot       ProvisionMode = "boot"
 	ProvisionModeDependency ProvisionMode = "dependency"
-	ProvisionModeAnsible    ProvisionMode = "ansible"
+	ProvisionModeAnsible    ProvisionMode = "ansible" // DEPRECATED
```
AkihiroSuda, line 224: https://lima-vm.io/docs/releases/deprecated/ has to be updated

templates/default.yaml
```diff
@@ -256,6 +256,8 @@ containerd:
 # # It requires `ansible-playbook` command to be installed.
 # # Environment variables such as ANSIBLE_CONFIG can be used, to control the behavior of the playbook execution.
 # # See ansible docs, and `ansible-config`, for more info https://docs.ansible.com/ansible/latest/playbook_guide/
+# # DEPRECATED The ansible mode is deprecated, and should not be used. Instead call ansible-playbook directly,
+# # either from the host after the instance is started or from the instance by running ansible locally instead.
```
AkihiroSuda, line 260: You can add an `/examples/ansible` page under https://lima-vm.io/docs/examples/ for ansible-playbook

pkg/limayaml/limayaml.go
AkihiroSuda: ping @afbjorklund 🙂

jandubois (approved):
Thanks, LGTM

I think you can add the ansible example later in a separate PR, but up to you...

AkihiroSuda (approved):
Thanks

PR #3453, opened 2025-04-21
Refactor images lists and default mounts into base templates

### LINKED ISSUE #2418: Simplify syntax for selecting images, with locations and digests

jandubois:
Also adds template://default/mounts to setup the default mounts globally.

Closes #2418

If we decide to do this, then we should also create `images/*` templates for the other distributions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -196,7 +196,8 @@ jobs:
     - name: Validate jsonschema
       run: make schema-limayaml.json
     - name: Validate templates
-      run: find -L templates -name '*.yaml' | xargs limactl validate
+      # Can't validate base templates in `_default` because they have no images
+      run: find -L templates -name '*.yaml' ! -path '*/_default/*' | xargs limactl validate
     - name: Install test dependencies
       # QEMU:      required by Lima itself
       # bash:      required by test-templates.sh (OS version of bash is too old)
@@ -208,7 +209,7 @@ jobs:
       run: echo "LIMACTL_CREATE_ARGS=${LIMACTL_CREATE_ARGS} --vm-type=qemu" >>$GITHUB_ENV
     - name: "Inject `no_timer_check` to kernel cmdline"
       # workaround to https://github.com/lima-vm/lima/issues/84
-      run: ./hack/inject-cmdline-to-template.sh templates/default.yaml no_timer_check
+      run: ./hack/inject-cmdline-to-template.sh templates/_images/ubuntu.yaml no_timer_check
     - name: Cache image used by default.yaml
       uses: ./.github/actions/setup_cache_for_template
       with:
@@ -353,7 +354,7 @@ jobs:
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
       with:
-        fetch-depth: 1
+        fetch-depth: 0
     - uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b  # v5.4.0
       with:
         go-version: 1.24.x
@@ -365,7 +366,7 @@ jobs:
       run: echo "LIMACTL_CREATE_ARGS=${LIMACTL_CREATE_ARGS} --vm-type=qemu --network=lima:shared" >>$GITHUB_ENV
     - name: "Inject `no_timer_check` to kernel cmdline"
       # workaround to https://github.com/lima-vm/lima/issues/84
-      run: ./hack/inject-cmdline-to-template.sh templates/default.yaml no_timer_check
+      run: ./hack/inject-cmdline-to-template.sh templates/_images/ubuntu.yaml no_timer_check
     - name: Cache image used by default .yaml
       uses: ./.github/actions/setup_cache_for_template
       with:
@@ -408,6 +409,7 @@ jobs:
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
       with:
+        # To avoid "failed to load YAML file \"templates/experimental/riscv64.yaml\": can't parse builtin Lima version \"3f3a6f6\": 3f3a6f6 is not in dotted-tri format"
         fetch-depth: 0
     - name: Fetch homebrew-core commit messages
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
@@ -452,7 +454,7 @@ jobs:
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
       with:
-        fetch-depth: 1
+        fetch-depth: 0
     - uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b  # v5.4.0
       with:
         go-version: 1.24.x
@@ -487,7 +489,7 @@ jobs:
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
       with:
-        fetch-depth: 1
+        fetch-depth: 0
     - uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b  # v5.4.0
       with:
         go-version: 1.24.x
--- a/pkg/templatestore/templatestore.go
+++ b/pkg/templatestore/templatestore.go
@@ -4,6 +4,8 @@
 package templatestore
 
 import (
+	"errors"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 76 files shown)


templates/ubuntu-25.04.yaml
```diff
@@ -1,37 +1,7 @@
...
-  arch: "armv7l"
-- location: "https://cloud-images.ubuntu.com/releases/plucky/release/ubuntu-25.04-server-cloudimg-s390x.img"
-  arch: "s390x"
-mounts:
-- location: "~"
-- location: "/tmp/lima"
-  writable: true
+minimumLimaVersion: "1.1.0"
+base:
+- template://images/ubuntu-25.04.yaml
```
AkihiroSuda, line 3: Maybe:
```suggestion
- template://_images/ubuntu-25.04.yaml
```

templates/ubuntu-25.04.yaml
```diff
@@ -1,37 +1,7 @@
...
+- template://default/mounts.yaml
```
AkihiroSuda, line 4: Maybe:
```suggestion
- template://default/mounts.yaml
```

templates/images/ubuntu-lts.yaml
AkihiroSuda, line 1: This file could be just a symlink to 24.04

templates/images/ubuntu.yaml
AkihiroSuda, line 1: Should be a symlink to ubuntu-24.10

templates/ubuntu-25.04.yaml
jandubois: Maybe, but I think the `images/*` templates should be user-visible. They can use them as a base in their own templates, and you can even instantiate them directly (if you don't want the default mounts):

```
limactl start template://images/ubuntu-25.04
```

templates/ubuntu-25.04.yaml
jandubois: I don't see any change between the two lines. Did you intend to prefix `default` with an underscore?

I do want to hide the `default/*` templates from `limactl create --list-templates` because they can't be instantiated on their own.

I don't know if adding the `_` is the right way to do it, or if the `default` namespace should just be hidden via a special case in the code?

AkihiroSuda: Yes, I meant `_default`  😓 

I don't prefer to hard-code `default` in the code

templates/images/ubuntu.yaml
jandubois: It should definitely point to the ubuntu-24.10 images list, but using a base template gives us a chance to add the comment why it points to 24.10 and not 25.04, so I think I prefer that.

The performance gain of using a symlink instead of a template merge is not insignificant.

Another reason to use the template mechanism is that symlinks don't necessarily work on Windows (need to be admin, or have some developer mode enabled). I know that the image lists currently don't work on Windows anyways, but the goal should be to eventually have a unified image list for all platforms.

templates/ubuntu-25.04.yaml
jandubois: Well, we'll have to hard-code the test for the leading `_` then, so not much of a change. I'll add the `_` for now; it is easy enough to change it again later.

I don't want to add it to `images` because those are fully functional templates on their own.

jandubois: > I don't want to add it to `images` because those are fully functional templates on their own.

Changed my mind, can also be changed later, if we want to.

jandubois:
I wonder how we should deal with `templates/default.yaml`. Ideally it should also use the base templates, but then we are missing out on the comments.

Maybe include some commented out sample values in the default template, but still use the base templates to get the images and mounts?

jandubois:
The tests also use the templates via filename locator (`templates/docker.yaml`) without installing the templates first.

Which means all the references to base templates should be relative locators: `_images/ubuntu-lts.yaml` instead of `template://_images/ubuntu-lts`. I've changed this PR to accommodate.

Which means the templates in `templates/experimental` cannot use the new base templates because `../_images/ubuntu-lts.yaml` is invalid; all relative references must be in or below the directory of the parent template.

So I think I should revert the change and reference the base templates via `template://` again. Which means that we either need to install the templates before we need use them, or we could add a `LIMA_TEMPLATES` variable so we can override the location of the template store.

I would prefer to add `LIMA_TEMPLATES`.

jandubois:
> Maybe `LIMA_TEMPLATES_PATH`, with the support for PATH separators (`:`)

I'll take a look, but this is now getting into scope-creep for a quick PR that I hoped to do on the side while working on something else. 😄

jandubois:
This turned into a bit bigger PR than originally expected. I think it is getting close now, so please do another review.

We could move other distro images to `_images/*` as well, but since they are not currently duplicated, it would not really simplify anything. It would just make it easier to reuse the image lists from end-user templates.

That may be enough of a reason to do it, but maybe it should be done at a later point in time, when we have some experience/feedback on the current refactoring?

.github/workflows/test.yml
```diff
@@ -353,7 +354,7 @@ jobs:
     steps:
     - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
       with:
-        fetch-depth: 1
+        fetch-depth: 0
```
AkihiroSuda, line 357: ?

pkg/templatestore/templatestore.go
```diff
@@ -20,21 +22,32 @@ type Template struct {
...
-		return nil, err
+	var pathList []string
+	if tmplPath := os.Getenv("LIMA_TEMPLATES_PATH"); tmplPath != "" {
+		pathList = strings.Split(tmplPath, string(filepath.ListSeparator))
+	} else {
+		dir, err := usrlocalsharelima.Dir()
+		if err != nil {
+			return nil, err
+		}
+		pathList = []string{filepath.Join(dir, "templates"), name}
```
AkihiroSuda, line 33:
```suggestion
		pathList = []string{filepath.Join(dir, "templates")}
```

pkg/templatestore/templatestore.go
```diff
@@ -20,21 +22,32 @@ type Template struct {
 }
 
 func Read(name string) ([]byte, error) {
-	dir, err := usrlocalsharelima.Dir()
-	if err != nil {
-		return nil, err
+	var pathList []string
+	if tmplPath := os.Getenv("LIMA_TEMPLATES_PATH"); tmplPath != "" {
```
AkihiroSuda, line 26: https://lima-vm.io/docs/config/environment-variables/ has to be updated

templates/almalinux-8.yaml
```diff
@@ -21,11 +21,9 @@ images:
   arch: "aarch64"
 - location: "https://repo.almalinux.org/almalinux/8/cloud/s390x/images/AlmaLinux-8-GenericCloud-latest.s390x.qcow2"
   arch: "s390x"
+
 mountTypesUnsupported: ["9p"]
-mounts:
-- location: "~"
-- location: "/tmp/lima"
-  writable: true
+base: template://_default/mounts
```
AkihiroSuda, line 26: For consistency the image snippet should be separated out

website/content/en/docs/dev/internals.md
```diff
@@ -135,6 +135,10 @@ The directory contains the following files:
 - `$LIMA_SHELL`: `lima ...` is expanded to `limactl shell --shell ${LIMA_SHELL} ...`.
   - No default : will use the user's shell configured inside the instance
 
+- `$LIMA_TEMPLATES_PATH`: A list of directories to locate templates via
+  the `template://` schema.
+  - Default: the `/usr/local/share/lima/templates` directory.
+
```
AkihiroSuda, line 141: This should be migrated to https://lima-vm.io/docs/config/environment-variables/

jandubois: I'm slightly confused that most (but not all, e.g. not `LIMA_HOME`) are documented in both locations. What are the rules about what goes where?

AkihiroSuda: All the variables should be moved to https://lima-vm.io/docs/config/environment-variables/

AkihiroSuda: I guess I had just forgotten the existence of https://lima-vm.io/docs/dev/internals when I merged the PR for https://lima-vm.io/docs/config/environment-variables/

😞

templates/almalinux-8.yaml
jandubois: I've separated out all image definitions for the regular templates, but not for the experimental ones.

Anything specific to the images themselves (like mount type restrictions, or limitation on using vz or qemu) have been moved to the `_images/*` templates as well, because they should be inherited by any template that uses these images.

I've also normalized most of the templates a little bit, so they generally follow this format:

```yaml
# comments / usage on top

minimumLimaVersion: 1.1.0

base: …

# the rest of the template
```

jandubois:
@AkihiroSuda PTAL! I hope CI will be green (except for the unrelated Windows failure). I will not be making any more changes today, and I hope this is close to "done" now.

Hope I don't jinx it by marking it "Ready for review". 😄

AkihiroSuda (approved):
Thanks

PR #3454, opened 2025-04-21
templates: add linuxbrew

AkihiroSuda:
This template does not use macOS as the guest OS, because it seems hard to automate provisioning macOS instances (accept EULA, skip enabling iCloud, set up SSH...).

To run Homebrew on macOS with a sandbox, Alcoholless can be used instead. The `alclessctl` CLI of Alcoholless is designed to be similar to `limactl`: https://github.com/AkihiroSuda/alcless

```
alclessctl create default
cd ~/SOME_DIRECTORY
alcless brew install xz
alcless xz SOME_FILE
```

Eventually, Alcoholless may be consolidated into Lima.

### DIFF
--- a/templates/linuxbrew.yaml
+++ b/templates/linuxbrew.yaml
@@ -1,0 +1,40 @@
+# Homebrew on Linux.
+#
+# Useful for sandboxing `brew` in an isolated environment.
+#
+# This template does not use macOS as the guest OS, because it seems hard to automate
+# provisioning macOS instances (accept EULA, skip enabling iCloud, set up SSH...).
+
+minimumLimaVersion: 1.1.0
+
+base:
+- template://_images/ubuntu-lts
+# Remove this line to disable the mounts
+- template://_default/mounts
+
+# containerd should be installed via Homebrew when Homebrew supports it
+containerd:
+  system: false
+  user: false
+
+provision:
+- mode: data
+  path: /etc/profile.d/99-linuxbrew.sh
+  # nofile is increased due to https://github.com/Homebrew/brew/issues/9120
+  content: |
+    ulimit -n 65536
+    [ -e /home/linuxbrew/.linuxbrew/bin/brew ] && eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    command -v gcc >/dev/null 2>&1 && exit 0
+    export DEBIAN_FRONTEND=noninteractive
+    apt-get update
+    apt-get install -y build-essential
+- mode: user
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    [ -e /home/linuxbrew ] && exit 0
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
--- a/templates/README.md
+++ b/templates/README.md
@@ -33,6 +33,9 @@ Distro:
 - [`experimental/opensuse-tumbleweed`](./experimental/opensuse-tumbleweed.yaml): [experimental] openSUSE Tumbleweed
 - [`experimental/debian-sid`](./experimental/debian-sid.yaml): [experimental] Debian Sid
 
+Alternative package managers:
+- [`linuxbrew.yaml`](./linuxbrew.yaml): [Homebrew](https://brew.sh) on Linux (Ubuntu)
+
 Container engines:
 - [`apptainer`](./apptainer.yaml): Apptainer
 - [`apptainer-rootful`](./apptainer-rootful.yaml): Apptainer (rootful)


templates/linuxbrew.yaml
```diff
@@ -0,0 +1,53 @@
+# Homebrew on Linux.
+#
+# This template does not use macOS as the guest OS, because it seems hard to automate
+# provisioning macOS instances (accept EULA, skip enabling iCloud, set up SSH...).
+#
+# To run Homebrew on macOS with a sandbox user account, Alcoholless can be used instead.
+# The `alclessctl` CLI of Alcoholless is designed to be similar to `limactl`:
+# https://github.com/AkihiroSuda/alcless
```
AkihiroSuda, line 8: (Happy to move https://github.com/AkihiroSuda/alcless to https://github.com/lima-vm if there is no objection)

templates/linuxbrew.yaml
```diff
@@ -0,0 +1,52 @@
...
+# alcless brew install xz
+# alcless xz SOME_FILE
+# ```
+#
+# Eventually, Alcoholless may be consolidated into Lima.
+
+minimumLimaVersion: 1.1.0
+
+base:
+- template://_images/ubuntu-24.10
```
jandubois, line 22: Should this be "ubuntu-latest", which we just call "ubuntu"?

```suggestion
- template://_images/ubuntu
```

templates/linuxbrew.yaml
```diff
@@ -0,0 +1,52 @@
...
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    command -v gcc >/dev/null 2>&1 && exit 0
+
+    # nofile has to be increased: https://github.com/Homebrew/brew/issues/9120
+    cat <<'EOF' >/etc/profile.d/99-linuxbrew.sh
+    ulimit -n 65536
+    [ -e /home/linuxbrew/.linuxbrew/bin/brew ] && eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
+    EOF
```
jandubois, line 42: This could be a `data` script:

```yaml
- mode: data
  path: /etc/profile.d/99-linuxbrew.sh
  contents: |
    ulimit -n 65536
    [ -e /home/linuxbrew/.linuxbrew/bin/brew ] && eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"
```

templates/linuxbrew.yaml
jandubois: This is probably not the best place to discuss this. I don't have strong objections, but it does not really feel like it is related to "Linux Machines"; it is more like a chroot jail.

I also really don't like the name. I find the whole beer/brew/tap/cask/bottle terminology quite weird, and unless you want to make this part of homebrew, I don't know why you would have picked a name from the same origin. Also `alclessctl` has just too many syllables. 😄

AkihiroSuda: > This is probably not the best place to discuss this. I don't have strong objections, but it does not really feel like it is related to "Linux Machines"

There has been a discussion to extend the project scope to support non-Linux Machines:
- https://github.com/lima-vm/lima/issues/1508

My initial intent was to extend Lima to directly support macOS VMs so as to isolate `brew install <RANDOM_UNREVIEWED_PACKAGE>` from the host, but I ended up creating a "poor man's container with Lima's CLI" using su + sudo + rsync, because macOS doesn't have an equivalent of cloud-init 😞. I strongly guess that Apple internally has it, though.

> it is more like a chroot jail.

It is just su + sudo + rsync

> I also really don't like the name. 

I don't like the name either 😅 . Suggestions are welcome for an alternative name..

> I find the whole beer/brew/tap/cask/bottle terminology quite weird, and unless you want to make this part of homebrew, I don't know why you would have picked a name from the same origin. 

I still _want_ to make it part of the upstream Homebrew, but I don't feel it is negotiable with Homebrew maintainers in the current state, as Alcoholless depends on the ["Untar anywhere (unsupported)" mode](https://docs.brew.sh/Installation#untar-anywhere-unsupported) ([Tier 3](https://docs.brew.sh/Support-Tiers)).

I still wish that this mode could be promoted to Tier 1, or at least Tier 2, if I manage to spread Alcoholless to Homebrew users.

> Also `alclessctl` has just too many syllables. 😄

Yep. Totally unpronounceable 😓

templates/linuxbrew.yaml
AkihiroSuda: Thanks, fixed

templates/linuxbrew.yaml
AkihiroSuda: This was intended to be `ubuntu-lts`

templates/linuxbrew.yaml
jandubois: > There has been a discussion to extend the project scope to support non-Linux Machines:
> 
> * [Support for FreeBSD guests ? #1508](https://github.com/lima-vm/lima/issues/1508)

That is still about virtualization though. But I guess we can still find some over-arching theme that would make it fit, like "running isolated workloads" or something... As I said, I don't have strong objections, just feels a bit like scope creep.

> My initial intent was to extend Lima to directly support macOS VMs so as to isolate `brew install <RANDOM_UNREVIEWED_PACKAGE>` from the host, but I ended up creating a "poor man's container with Lima's CLI" using su + sudo + rsync, because macOS doesn't have an equivalent of cloud-init 😞. I strongly guess that Apple internally has it, though.

Maybe, but I doubt that they would ever make it publicly available.
 
> I don't like the name either 😅 . Suggestions are welcome for an alternative name..

Why not call it `jail`. It is short, pronounceable, and unclaimed, afaict. Also has nice symmetry with `lima` and `limactl`: `jail` and `jailctl`.

AkihiroSuda: "jail" means FreeBSD-style jails by default, so some prefix would have to be prepended to "jail"
e.g., "brewjail", "userjail", etc.

jandubois: I would prefer something short. How about `isola`. Italian for "island" and also hints at "isolated execution"?

I've been trying to find a word that would be more like "peninsula" because the isolation is not perfect. I think `leaf` could work, but there is already a tool with that name.

jandubois (dismissed):
Thanks, LGTM

templates/linuxbrew.yaml
jandubois: We continued to brainstorm on Slack. I kind of like `confine`, as in "you are confined to this directory and cannot escape".

AkihiroSuda: Removed the advert from the yaml

> isola

To me it sounds like an equivalent of mtools ([still actively maintained in 2025 👀](https://ftp.gnu.org/gnu/mtools)) for ISO9660 images 💿

> confine, enclose, restrict

Ungoogleable.

> silo

Taken by https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/s/silo.rb

Thanks for the suggestions anyway.

Let me open a new issue for continuing the discussion

jandubois (approved):
Thanks, LGTM

templates/linuxbrew.yaml
AkihiroSuda: - https://github.com/AkihiroSuda/alcless/issues/10

PR #3461, opened 2025-04-22
templates: Add Debian and Fedora riscv64 images

shenki:
The Debian image is the latest from unstable as there's no release yet for riscv64.

Fedora is a Fedora 42 cloud image. It's not clear how often this will be updated, so the URL may become stale at some point.

### DIFF
--- a/templates/_images/fedora-42.yaml
+++ b/templates/_images/fedora-42.yaml
@@ -7,6 +7,9 @@ images:
   arch: aarch64
   digest: sha256:e10658419a8d50231037dc781c3155aa94180a8c7a74e5cac2a6b09eaa9342b7
 
+- location: https://dl.fedoraproject.org/pub/alt/risc-v/release/42/Cloud/riscv64/images/Fedora-Cloud-Base-Generic-42.20250414-8635a3a5bfcd.riscv64.qcow2
+  arch: riscv64
+  digest: sha256:537c67710f4f1c9112fecaafafc293b649acd1d35b46619b97b5a5a0132241b0
 
 # # NOTE: Intel Mac requires setting vmType to qemu
 # # https://github.com/lima-vm/lima/issues/3334
--- a/templates/_images/fedora-41.yaml
+++ b/templates/_images/fedora-41.yaml
@@ -7,6 +7,10 @@ images:
   arch: aarch64
   digest: sha256:085883b42c7e3b980e366a1fe006cd0ff15877f7e6e984426f3c6c67c7cc2faa
 
+- location: https://dl.fedoraproject.org/pub/alt/risc-v/release/41/Cloud/riscv64/images/Fedora-Cloud-Base-Generic-41.20250224-1026a2d0e311.riscv64.qcow2
+  arch: riscv64
+  digest: sha256:6a8272a858d7f1498f49ce362b34f0b9b959885f63285158947e045abfeece40
+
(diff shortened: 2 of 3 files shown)


templates/experimental/fedora-riscv64.yaml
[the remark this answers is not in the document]
jandubois: > This should be just merged into `fedora.yaml`

I assume you meant `_images/fedora.yaml`.

I wonder why we wouldn't use the "Cloud-Base-Generic" image though, like we do for the other archs: https://dl.fedoraproject.org/pub/alt/risc-v/release/42/Cloud/riscv64/images/Fedora-Cloud-Base-Generic-42.20250414-8635a3a5bfcd.riscv64.qcow2

If it doesn't work for some reason, then we should add a comment about that.

I also haven't checked if the image updater supports this new image URL? We need to be careful that the image updater doesn't start updating only some of the images in a particular template.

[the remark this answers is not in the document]
shenki: > > This should be just merged into `fedora.yaml`
> 
> I assume you meant `_images/fedora.yaml`.
> 
> I wonder why we wouldn't use the "Cloud-Base-Generic" image though, like we do for the other archs: https://dl.fedoraproject.org/pub/alt/risc-v/release/42/Cloud/riscv64/images/Fedora-Cloud-Base-Generic-42.20250414-8635a3a5bfcd.riscv64.qcow2
> 
> If it doesn't work for some reason, then we should add a comment about that.

We can use that one instead.

> I also haven't checked if the image updater supports this new image URL? We need to be careful that the image updater doesn't start updating only some of the images in a particular template.

I tried the update script and it doesn't update anything at this point. The path doesn't match (`/pub/alt`) and the metadata is at https://dl.fedoraproject.org/pub/alt/imagelist-alt instead of https://dl.fedoraproject.org/pub/fedora/imagelist-fedora.

I assume as riscv64 becomes a fully supported architecture in Fedora, the images will eventually end up in the normal location.

templates/_images/fedora-42.yaml
```diff
@@ -7,6 +7,9 @@ images:
   arch: aarch64
   digest: sha256:e10658419a8d50231037dc781c3155aa94180a8c7a74e5cac2a6b09eaa9342b7
 
+- location: https://dl.fedoraproject.org/pub/alt/risc-v/release/42/Cloud/riscv64/images/Fedora-Cloud-Base-Generic-42.20250414-8635a3a5bfcd.riscv64.qcow2
```
AkihiroSuda, line 10: Do you wan to update fedora-41.yaml too?

Also wonder if we can cover alt/risc-v in https://github.com/lima-vm/lima/blob/master/hack/update-template-fedora.sh

shenki: > Do you wan to update fedora-41.yaml too?

I didn't realise we had 41. Yep, I'll test that it boots and add it.

> Also wonder if we can cover alt/risc-v in https://github.com/lima-vm/lima/blob/master/hack/update-template-fedora.sh

I spent a few minutes trying to hack it in, but I didn't get it working. I'm not as much of a shell wizard as rhe author of those scripts!

AkihiroSuda: cc @norio-nomura

shenki:
Fedora 41 booted fine, so I've added that.

AkihiroSuda (approved):
Thanks

PR #3463, opened 2025-04-22
Graduate base/file embedding

jandubois:
Almost all builtin templates now use `base`, so the warnings are annoying.

See also https://github.com/lima-vm/lima/pull/3453#issuecomment-2820496242

### DIFF
--- a/pkg/limatmpl/embed.go
+++ b/pkg/limatmpl/embed.go
@@ -24,14 +24,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var warnBaseIsExperimental = sync.OnceFunc(func() {
-	logrus.Warn("`base` is experimental")
-})
-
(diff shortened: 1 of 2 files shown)


pkg/limatmpl/embed.go
```diff
@@ -24,14 +24,6 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-var warnBaseIsExperimental = sync.OnceFunc(func() {
-	logrus.Warn("`base` is experimental")
-})
-
-var warnFileIsExperimental = sync.OnceFunc(func() {
-	logrus.Warn("`provision[*].file` and `probes[*].file` are experimental")
```
AkihiroSuda, line 32: Seems to be a different topic, but probably fine

AkihiroSuda (approved):
Thanks

Issue #3316, opened 2025-03-07
[Windows] Many templates of Lima are not Windows compatible, when built on Windows


arixmkii:
### Description

Right now templates for QEMU, which have mounts defined are not Windows compatible, because they have Unix absolute path entry "/tmp/lima". They have to be manually changed before they can be used.

Proposed solution:
Provide additional build step which will transform incompatible values into acceptable: something like "/tmp/lima" -> "C:\Temp\lima", which is definitely not imperfect, but somehow better.

So, then one can call command `make binaries windows-templates` and get templates transformed in `_output` directory.

afbjorklund:
We knew this, but it was "good enough" when assuming something like MSYS that would translate paths...

* https://github.com/lima-vm/lima/issues/2339

arixmkii:
@afbjorklund Thank you! will read the discussion there. I have some concerns about translation paths in some edge cases, but I need some time with my Windows test box to experiment with it to confirm.

jandubois:
> Proposed solution:
> Provide additional build step which will transform incompatible values into acceptable

I think we should aim for templates that work on any platform. Otherwise using a template via URL doesn't work; you would have to provide different templates for different platforms, which becomes a maintenance nightmare:

```shell
limactl create --name my-vm https://example.com/template.yaml
```

I've implemented my suggestion from https://github.com/lima-vm/lima/issues/2339#issuecomment-2231525877 in #3318. Please let me know if this would work for you on Windows!

I do have some concerns that this will move the host location on macOS from `/tmp` to `$TMPDIR`, but I think this would be the right semantics anyways.

jandubois:
Also: don't we need a similar solution for mounting `~`? Because we would want to mount e.g. `C:\Users\jan` to `/c/Users/jan` or something similar?

The concept of having `mountPoint` default to `location` doesn't really apply on Windows; the paths always need translation. And the translation may be different based on `vmType`. E.g. for `WSL2` the `mountPoint` would be `/mnt/c/Users/jan` (by default), and Lima doesn't need to bother about mounting the volume. But for `QEMU` the `mountPoint` default may need to be different.

jandubois:
> I do have some concerns that this will move the host location on macOS from `/tmp` to `$TMPDIR`

One thing that bothers me is that this breaks the symmetry of mounting a host temp file into a container, so this would not work:

```console
docker run --volume $TMPDIR/lima:/mnt/lima ...
```

[...]

Issue #302, opened 2021-10-07
Overlapping mounts don't work correctly


jandubois:
I wanted to keep `~` readonly, but make another directory below it (say, `~/tmp`) writable. I get one of 2 results:

1. `~/tmp` is still readonly, all of `~` is visible, or
2. `~/tmp` is writable, but it is the only entry under `~`. `~` is also writable, but files created here are not copied over to the host (unlike files created under `~/tmp`.

It seems like a reasonable requirement to make all of `~` readable, but only a subset writable. Using `/tmp/lima` is a workaround, but fragile as files there are automatically deleted after 3 days.

jandubois:
> Have you consider adding data disks as a feature ? It was considered for machine, but never implemented.

We kind of support data disks (automatically) if the boot image is an ISO. I suspect it only works for Alpine right now.

> But maybe it wouldn't fit into the lima storage model, if it needs to be directly accessible from the host as well.

I don't think it fits. Lima provides the illusion that you can run `nerdctl` directly on it host, and it just works. But it really relies on the host files being mounted at the same location inside the instance, and then `nerdctl` running inside because it needs direct access to both the source files for building, and the containerd layers to directly manipulate the images.

We need to look again into using samba (#188).

My main worry is that once I test `nfsd`, it will turn out to be the fastest option. Unfortunately it is clumsy to configure (you need to manipulate `/etc/exports` on the host, as `root`), and the Apple implementation is (or used to be) buggy.

I've used it a couple of years ago with Vagrant (also for performance reasons), and while being fast, it sometimes didn't realize a host file was updated and served a cached copy. You had to create a sparsebundle layer on the host to avoid the bug. So yeah, I really hope `nfsd` isn't still the most performant option in 2021. Or if it is, that at least the bugs have been fixed!

PR #3475, opened 2025-04-26
Add QEMU on Windows to CI

arixmkii:
For now it will use additional templates, because of incompatible mounts.

This is probably not for 1.1.0.

It is possible to use default.yaml for Windows with changes from #3318 (this would need rebase first, but I checked it using a rebased patch in a forked repo - example run https://github.com/arixmkii/qcw/actions/runs/14681726480/job/41205771124).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -151,8 +151,8 @@ jobs:
       run: sudo make uninstall
 
   windows:
-    name: "Windows tests"
-    runs-on: windows-2022-8-cores
+    name: "Windows tests (WSL2)"
+    runs-on: windows-2025-8-cores
     timeout-minutes: 30
     steps:
     - name: Enable WSL2
@@ -204,6 +204,40 @@ jobs:
         $env:LIMACTL_CREATE_ARGS = '--vm-type=wsl2 --mount-type=wsl2 --containerd=system'
         $env:_LIMA_WINDOWS_EXTRA_PATH = 'C:\Program Files\Git\usr\bin'
         bash.exe -c "./hack/test-templates.sh templates/experimental/wsl2.yaml"
+
+  windows-qemu:
+    name: "Windows tests (QEMU)"
+    runs-on: windows-2025
+    timeout-minutes: 30
+    steps:
+    - name: Set gitconfig
+      run: |
+        git config --global core.autocrlf false
+        git config --global core.eol lf
+    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+      with:
+        # To avoid "can't parse builtin Lima version" errors
+        fetch-depth: 0
+    - uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b  # v5.4.0
+      with:
+        go-version: 1.24.x
+    - name: Unit tests
+      run: go test -v ./...
+    - name: Make
+      run: make
+    - name: Install QEMU
+      run: |
+        winget install --silent --accept-source-agreements --accept-package-agreements --disable-interactivity SoftwareFreedomConservancy.QEMU
+    - name: Integration tests (QEMU, Windows host)
+      run: |
+        $env:PATH = "$pwd\_output\bin;" + 'C:\msys64\usr\bin;' + 'C:\Program Files\QEMU;' + $env:PATH
+        pacman -Sy --noconfirm openbsd-netcat diffutils
+        $env:MSYS2_ENV_CONV_EXCL = 'HOME_HOST;HOME_GUEST;_LIMA_WINDOWS_EXTRA_PATH'
+        $env:HOME_HOST = $(cygpath.exe "$env:USERPROFILE")
+        $env:HOME_GUEST = "$env:HOME_HOST"
+        $env:LIMACTL_CREATE_ARGS = '--vm-type=qemu'
+        $env:_LIMA_WINDOWS_EXTRA_PATH = 'C:\Program Files\Git\usr\bin'
+        bash.exe -c "./hack/test-templates.sh templates/default.yaml"
 
   qemu:
     name: "Integration tests (QEMU, macOS host)"
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -60,6 +60,11 @@ declare -A CHECKS=(
 )
 
 case "$NAME" in
+"default")
+	# CI failure:
+	# "[hostagent] failed to confirm whether /c/Users/runneradmin [remote] is successfully mounted"
+	[ "${OS_HOST}" = "Msys" ] && CHECKS["mount-home"]=
+	;;
 "alpine"*)
 	WARNING "Alpine does not support systemd"
 	CHECKS["systemd"]=


arixmkii:
> time="2025-04-26T17:20:45Z" level=fatal msg="failed to validate YAML file \"C:\\\\a\\\\lima\\\\lima\\\\templates\\\\experimental\\\\default-windows.yaml\": can't parse builtin Lima version \"cfbffd8\": cfbffd8 is not in dotted-tri format"

make/git on Windows incorrectly resolve version. I will check it (no such issues, when checkout and build are done with msys2 tools).

Another topic to check - use chocolatey to install QEMU, because msys2 QEMU installation feels slow.

arixmkii:
Probably would need to move mounts-windows under _default to not fail validation script.

arixmkii:
Chocolatey QEMU package is not well maintained, so, I chose winget instead, which is a great alternative. There is a known limitation that it is not available out of the box in Windows Server 2022, so, there is a hacky action to add it, which is now archived and will not be needed at all after migration to Windows Server 2025, this is highlighted by the comment.

arixmkii:
@jandubois @AkihiroSuda I would like to know your opinions on how reasonable is it to extend CI to support this (to not overload CI and not increase costs significantly). From my side there is no rush and I can see reasons to postpone this until #3316 is addressed (via #3318 refresh or other means). Also it might be reasonable to wait for migration to WinServer 2025 to not use now archived https://github.com/Cyberboss/install-winget action.

I authored it now to have proof of concept confirmed and potentially creating reference starting point for its introduction.

templates/experimental/default-windows.yaml
AkihiroSuda, line 1: Would it be possible to merge the template to `default`?

If it is still difficult, can we use `base: template://default` for deduplication?

.github/workflows/test.yml
```diff
@@ -175,6 +175,44 @@ jobs:
...
+        fetch-depth: 0
+    - uses: actions/setup-go@0aaccfd150d50ccaeb58ebd88d36e91967a5f35b  # v5.4.0
+      with:
+        go-version: 1.24.x
+    - name: Unit tests
+      run: go test -v ./...
+    - name: Make
+      run: make
+    - name: Install winget
+      # TODO remove this step for Windows Server 2025 based runners
```
AkihiroSuda, line 199: Windows 2025 seems now available:
- https://github.com/actions/runner-images/issues/11742

templates/experimental/default-windows.yaml
jandubois: If we decide that #3318 is acceptable, then the regular `template://default` will just work on Windows (I think).

.github/workflows/test.yml
jandubois: I've just created a new runner type for this, so we can try it out:

CleanShot 2025-04-30 at 15 12 54@2x

jandubois: Are we sure we actually need 8 cores? Or maybe we just start with a larger runner, and try later to see how much worse it becomes to run on a standard runner?

The 4-core runner is also limited to 16GB. That is plenty on macOS; not sure what is needed on Windows.

AkihiroSuda: Why do we need to create a custom runner?

jandubois: So it becomes available in our org? Somebody must have created the `windows-2022-8-cores` runner too, as it is also a custom entry for the `lima-vm` org (I don't think it was me). CNCF has only created custom Linux runners:

CleanShot 2025-05-01 at 10 48 15@2x

jandubois: [About Ubuntu and Windows larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#about-ubuntu-and-windows-larger-runners):

> Larger runners with Ubuntu or Windows operating systems are configured in your organization or enterprise.

[About macOS larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#about-macos-larger-runners):

> No additional configuration is required.

jandubois: > Probably it should just work with the standard runner now

Ok, let's try the standard runner first then, to avoid the additional cost.

.github/workflows/test.yml
```diff
@@ -175,6 +175,44 @@ jobs:
         $env:_LIMA_WINDOWS_EXTRA_PATH = 'C:\Program Files\Git\usr\bin'
         bash.exe -c "./hack/test-templates.sh templates/experimental/wsl2.yaml"
 
+  windows-qemu:
```
AkihiroSuda, line 178: Can we now drop these lines?

https://github.com/lima-vm/lima/blob/df13f3003192f8f1a82944edd6296f99a1e125b8/pkg/limayaml/defaults.go#L68-L75

arixmkii: From my experience "max" just didn't work well with WHPX acceleration. I tested it on 3 different machines in the past. I only was able to make it work by disabling specific CPU features, which were different on every machine. It was not user friendly default. I can do some canary testing to compare how it works now with newer QEMU/Windows versions and if the failures are as common as they were before.

.github/workflows/test.yml
arixmkii: I used standard runners for my development. It is possible to start from them and then compare if more cores are really beneficial.

templates/experimental/default-windows.yaml
arixmkii: Will rebase to latest and use default.

arixmkii:
I tried to limit both Windows builds to windows-2025 standard. I see QEMU one failed with errors mounting SSHFS (I observed this instability before with standard runners, they are definitely recurring and could very persistent restarting job). WSL2 faced some error and now test in a locked state (it will be terminated after 30 minutes time out, because I can't cancel it manually). I can say that WSL2 was less stable (comparing to Lima 8-cores runner), when I used standard runners, but I mostly faced errors from sysmtemd, this one is new.

I will restart build setting both to windows-2025-8-cores to compare.

arixmkii:
It didn't help for QEMU
```
time="2025-05-02T19:26:57Z" level=info msg="[hostagent] :/c/Users/runneradmin: No such file or directory"
time="2025-05-02T19:26:57Z" level=warning msg="[hostagent] failed to confirm whether /c/Users/runneradmin [remote] is successfully mounted" error="failed to execute script \"wait-for-remote-ready\": stdout=\"\", stderr=\"mux_client_request_session: read from master failed: Connection reset by peer\\r\\nControlSocket /c/Users/runneradmin/.lima/default/ssh.sock already exists, disabling multiplexing\\r\\nsshfs does not seem to be mounted on /c/Users/runneradmin\\n\": exit status 1"
```

SSHFS is weird on Windows in general and inside runners specifically. Giving some insights on my experience testing this in GH runners for a month or so. It always (or almost always) failed to mount $TEMP, but most of the time managed to mount $HOME, the situation with $TEMP - if temp was tried, but was not mounted the integration tests will still pass. 

Troubleshooting the $TEMP issue locally I first managed to replicate it on my dev machine, but the fix was to clean the content of $TEMP folder. It looked like sftp-server might be sensitive to the folder contents, but I didn't try to test this in details. 

I'm thinking I will test the standard runners and disable mount tests on Windows platform with a comment of them being flaky - which they indeed are. 

Will experiment in my repo on isolated examples and then will update this PR once again.

jandubois:
> It always (or almost always) failed to mount $TEMP, but most of the time managed to mount $HOME

Is this just another instance of #302? Because `$TEMP` will be located at `$HOME\AppData\Temp`?

I always thought the issue was the overlap in the guest, but maybe the overlap on the host is the real problem?

At the time I filed #302 we did not yet have support for specifying a different `mountPoint`, so it was impossible to tell which side was causing the issue.

AkihiroSuda:

arixmkii:
I believe I managed (up to my current understanding of the flaky part) to address this running it locally. I will experiment in GH CI in a forked repo and will post the status here this week.

.github/workflows/test.yml
```diff
@@ -151,8 +151,8 @@ jobs:
       run: sudo make uninstall
 
   windows:
-    name: "Windows tests"
-    runs-on: windows-2022-8-cores
+    name: "Windows tests (WSL2)"
+    runs-on: windows-2025-8-cores
```
AkihiroSuda, line 155: Maybe
```suggestion
    runs-on: windows-2025
```

arixmkii: We can try this. From my observations it was less stable in a forked repo, where I have only default runners. I will apply this change after I will check the comment with qemucpu to max one. 

If this will result in unstable builds it can be reverted later.

arixmkii:
I spent all day doing experiments with CI to make SSHFS behave on Windows. What I tried
* setting HOME and TEMP to different locations to have reduced FS tree for sftp-server
* setting HOME and TEMP to not nested setup (default on windows is TEMP nested deep inside HOME)
* trying different drives C: and D: as locations - may be the disks have different I/O settings in the runner

What I failed to test
* running test as a different user to fully isolate from build using psexec of paexec. I gave up on this idea because it is difficult to pass all needed environment configurations to another user session and there is no good way to redirect logs to stdout for traceability of the test runs

For now it seems that disabling sshfs tests on Windows was a way to go.

I have success to run them in github CI in a forked repo, where I chain 2 VMs - one creating build in a zip form and another one unzipping artifact and running tests on it. As Lima build is I/O heavy it might be that it hits sort of I/O throttling after the build, but I have no way to confirm it. And this chaining strategy giving better results still sometimes result in failures. I will continue to experiment with different setups for sshfs tests in a forked repo.

PR #3478, opened 2025-04-28
Fix version parsing in case of dirty checkout of a tag

### LINKED ISSUE #3477: Lima can't parse version if built from tag with dirty checkout
### Description

Scenario:
1. checkout tag
2. apply patch using `patch` command
3. build
4. run integration tests

Error:
```
time="2025-04-28T17:17:27Z" level=fatal msg="failed to validate YAML file \"D:\\\\a\\\\qcw\\\\qcw\\\\lima-release\\\\templates\\\\experimental\\\\default-windows.yaml\": can't parse builtin Lima version \"v1.1.0.m\": strconv.ParseInt: parsing \"0.m\": invalid syntax"
```

I added 1.1.0 tag artificially here to have no issues in my development build (no tags fetched) with lima version being too low.

The issue will go away if I tag the previous commit to my build, because ".m" would be in the part cut by splitting on '-' symbol.

I believe that building from a tagged commit with local changes should still result in a working binary.

arixmkii:
Fixes #3477

Added basic parsing unit tests.

### DIFF
--- a/pkg/version/versionutil/versionutil.go
+++ b/pkg/version/versionutil/versionutil.go
@@ -11,10 +11,12 @@ import (
 
 // Parse parses a Lima version string by removing the leading "v" character and
 // stripping everything from the first "-" forward (which are `git describe` artifacts and
-// not semver pre-release markers). So "v0.19.1-16-gf3dc6ed.m" will be parsed as "0.19.1".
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #3482, opened 2025-05-01
Finish $LIMA_TEMPLATES_PATH, add $LIMA_HOME/_templates

jandubois:
#3453 added support to locate templates via `LIMA_TEMPLATES_PATH`.

It was however not listing templates from the additional template directories for e.g. `limactl start --list-templates` and didn't include them in shell completions.

This PR adds the missing functionality, and also adds `$LIMA_HOME/_templates` as a default template search location if the directory exists and `LIMA_TEMPLATES_PATH` is not set.

In addition to making user templates addressable via `template://` schema this also allows the user for example to create `$LIMA_HOME/_templates/_default/mounts.yaml` to override the default mounts used by all the builtin templates.

This PR intentionally uses 3 commits that are related, but independent. I don't think they should be squashed.


AkihiroSuda (approved):
Thanks

PR #3484, opened 2025-05-01
Set riscv64 cpu to max

shenki:
https://lore.kernel.org/qemu-devel/20250404152750.332791-1-dbarboza@ventanamicro.com/

> [As more] distros use the 'virt' machine, and they'll start building
> on top of RVA23, and rv64 does not have RVA23 support. In short,
> distros will start to break in the default 'virt' CPU.
>
> Changing the default CPU to 'max' will not cause (intentional) user
> regressions: if the software runs in rv64 it will run in 'max' too given
> that we're adding more extensions as default instead of removing them.

To ensure lima works once Ubuntu 25.10 comes out, which is expected to require RV23, set the cpu type to max.

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -66,7 +66,7 @@ func defaultCPUType() CPUType {
 		ARMV7L:  "cortex-a7",
 		// Since https://github.com/lima-vm/lima/pull/494, we use qemu64 cpu for better emulation of x86_64.
 		X8664:   "qemu64",
-		RISCV64: "rv64", // FIXME: what is the right choice for riscv64?
+		RISCV64: "max",
 		S390X:   "qemu", // FIXME: what is the right choice for s390x?
 	}
 	for arch := range cpuType {


shenki:
cc @danielhb

pkg/limayaml/defaults.go
```diff
@@ -66,7 +66,7 @@ func defaultCPUType() CPUType {
 		ARMV7L:  "cortex-a7",
 		// Since https://github.com/lima-vm/lima/pull/494, we use qemu64 cpu for better emulation of x86_64.
 		X8664:   "qemu64",
-		RISCV64: "rv64", // FIXME: what is the right choice for riscv64?
+		RISCV64: "max",
```
AkihiroSuda, line 69: Do you know if we can use `max` for other architectures too?

AkihiroSuda (approved):
Thanks

pkg/limayaml/defaults.go
AkihiroSuda: I _thought_ that `qemu-system-x86_64 -cpu max` didn't work with TCG mode in the past, but I could be just wrong.

AkihiroSuda: ChatGPT says `-cpu max` is safe on modern QEMU

<details>

<p>

> You’re not imagining it—there really **were** times when `-cpu max` under the **TCG** accelerator simply didn’t work correctly.  Here’s a quick rundown:
> 
> ---
> 
> ## 1. What “-cpu max” really means  
> By design,  
> > **`-cpu max`** enables **all** CPU features that your **accelerator** supports (i.e. KVM, HVF, TCG, etc.)  ([[kvm virtualization - KVM - difference between -cpu host and -cpu max? - Server Fault](https://serverfault.com/questions/977643/kvm-difference-between-cpu-host-and-cpu-max)](https://serverfault.com/questions/977643/kvm-difference-between-cpu-host-and-cpu-max)).  
> So in pure-software (TCG) mode it will try to expose every feature TCG has implemented.
> 
> ---
> 
> ## 2. Historical bugs in TCG + `-cpu max`
> 
> 1. **Linux would hang**  
>    In QEMU 6.1.0 (and the Fedora builds around that time), launching a Linux guest with TCG and `-cpu max` caused SeaBIOS to hang just before the kernel took over.  
>    - Only the **max** model was affected; every other named CPU model worked fine  ([[1999700 – qemu -cpu max + TCG hangs booting Linux because of missing bits in CR4_RESERVED_MASK](https://bugzilla.redhat.com/show_bug.cgi?id=1999700)](https://bugzilla.redhat.com/show_bug.cgi?id=1999700)).  
>    - The root cause was an overly-strict CR4 reserved-bits check in the TCG i386 target.
> 
> 2. **Invalid-opcode on BMI instructions**  
>    Even when it didn’t outright hang, `-cpu max` could turn on modern bit-manipulation extensions (BMI, BMI2, ABM) that TCG didn’t fully decode, leading to “invalid opcode” traps under TCG (while they worked flawlessly under KVM)  ([[Bug #1748296 “TCG throws Invalid Opcode when executing x86 BMI s...” : Bugs : QEMU](https://bugs.launchpad.net/bugs/1748296)](https://bugs.launchpad.net/bugs/1748296)).
> 
> ---
> 
> ## 3. What to do today
> 
> - **On older QEMU (≈6.1.0)**:  
>   Avoid `-cpu max` with TCG. Instead use `-cpu host` (which matches your host CPU) or pick a specific named model (e.g. `Skylake-Server`).
> 
> - **On recent QEMU (≥6.1.0-5)**:  
>   The hanging bug was fixed in the 6.1.0-5 update, and the BMI issue has also been addressed in later patches. You can now safely use `-cpu max` under TCG, though unsupported features will still emit harmless warnings like:  
>   ```
>   warning: TCG doesn't support requested feature: CPUID.01H:ECX.vmx [bit 5]
>   ```
>   and any truly unimplemented instructions will fault, so keep an eye on your guest’s dmesg or console output.
> 
> ---
> 
> **In short**: your memory is correct—`-cpu max` *did* misbehave in TCG mode in the past. If you’ve upgraded to a modern QEMU, those specific issues are fixed, but if you’re on an older 6.1.0-era build, you’ll still hit those TCG-only bugs with `-cpu max`.

</p>

</details>

shenki: I'm not as familiar with the other arches.

danielhb:
> https://lore.kernel.org/qemu-devel/20250404152750.332791-1-dbarboza@ventanamicro.com/
> 
> > [As more] distros use the 'virt' machine, and they'll start building
> > on top of RVA23, and rv64 does not have RVA23 support. In short,
> > distros will start to break in the default 'virt' CPU.
> > Changing the default CPU to 'max' will not cause (intentional) user
> > regressions: if the software runs in rv64 it will run in 'max' too given
> > that we're adding more extensions as default instead of removing them.
> 
> To ensure lima works once Ubuntu 25.10 comes out, which is expected to require RV23, set the cpu type to max.

+1 for this change. I recommend that all tooling that uses QEMU specifies their desired CPU in the command line (-cpu option) instead of relying on QEMU defaults.

PR #3485, opened 2025-05-01
Add dirnames.TemplatesDir() function

jandubois:
All other special directories have a symbolic filename and dirname function.


AkihiroSuda (approved):
Thanks

PR #3487, opened 2025-05-02
qemu: tcg: use `max` by default

### LINKED ISSUE #3063: qemu: Intel on ARM: change the default CPU from `qemu64` to something newer
>  Found a comparison table: https://www.qemu.org/docs/master/system/i386/cpu.html
>
> Surprised to know that qemu64 sticks to `x86_64-v1` by default 👀

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/3047#discussion_r1898053000_

AkihiroSuda:
The previous default value `qemu64` for Intel on ARM was unable to boot recent guests as it only supports x86_64 v1 instructions.

CentOS Stream 10 is known to need x86_64 v3 or later.

Fix #3063

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -548,7 +548,8 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	// Machine
 	switch *y.Arch {
 	case limayaml.X8664:
-		if strings.HasPrefix(cpu, "qemu64") && runtime.GOOS != "windows" {
+		switch accel {
+		case "tcg":
 			// use q35 machine with vmware io port disabled.
 			args = appendArgsIfNoConflict(args, "-machine", "q35,vmport=off")
 			// use tcg accelerator with multi threading with 512MB translation block size
@@ -559,10 +560,10 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 			// This will disable CPU S3/S4 state.
 			args = append(args, "-global", "ICH9-LPC.disable_s3=1")
 			args = append(args, "-global", "ICH9-LPC.disable_s4=1")
-		} else if runtime.GOOS == "windows" && accel == "whpx" {
+		case "whpx":
 			// whpx: injection failed, MSI (0, 0) delivery: 0, dest_mode: 0, trigger mode: 0, vector: 0
 			args = appendArgsIfNoConflict(args, "-machine", "q35,accel="+accel+",kernel-irqchip=off")
-		} else {
+		default:
 			args = appendArgsIfNoConflict(args, "-machine", "q35,accel="+accel)
 		}
 	case limayaml.AARCH64:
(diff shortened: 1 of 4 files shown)


jandubois:
This PR is undoing the effect of #931; is that intentional?

AkihiroSuda:
I don't think this PR affects whpx

jandubois:
> I don't think this PR affects whpx

With this PR you now unconditionally set `cpuType.x86_64: max`.

#931 intentionally prevented the code from setting it to `max` when running on Windows, and kept it at `qemu64`. The comment explicitly says:

```
// WHPX: Unexpected VP exit code 4
```

So maybe this is no longer needed, but maybe it should be tested?

I realize that QEMU on Windows is currently still WIP, so it won't affect actual users, but we may need to rediscover this fact if we just remove it now.

jandubois:
See also https://github.com/lima-vm/lima/issues/909#issuecomment-1169557674

AkihiroSuda:
Fixed to use `qemu64` on Windows

pkg/qemu/qemu.go
```diff
@@ -548,7 +548,7 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	// Machine
 	switch *y.Arch {
 	case limayaml.X8664:
-		if strings.HasPrefix(cpu, "qemu64") && runtime.GOOS != "windows" {
+		if accel == "tcg" && runtime.GOOS != "windows" { // FIXME: Any reason to exclude windows here?
```
jandubois, line 551: I think the check for "windows" was here to prevent the `accel == "whpx"` case to go in this branch, instead of the "else if" branch below.

Since you now test of `accel == "tcg"` instead of `cpu == "qemu64"`, I think you can remove the check for "windows" here.

AkihiroSuda: done

templates/default.yaml
```diff
@@ -334,10 +334,10 @@ os: null
...
 # 🟢 Builtin default: hard-coded arch map with type (see the output of `limactl info | jq .defaultTemplate.cpuType`)
 cpuType:
-#   aarch64: "cortex-a76" # (or "host" when running on aarch64 host)
-#   armv7l: "cortex-a7" # (or "host" when running on armv7l host)
-#   riscv64: "rv64" # (or "host" when running on riscv64 host)
-#   x86_64: "qemu64" # (or "host,-pdpe1gb" when running on x86_64 host)
+#   aarch64: "max" # (or "host" when running on aarch64 host)
+#   armv7l:  "max" # (or "host" when running on armv7l host)
+#   riscv64: "max" # (or "host" when running on riscv64 host)
+#   x86_64:  "max" # (or "host,-pdpe1gb" when running on x86_64 host)
```
jandubois, line 340: Isn't it `host` on Linux, and `host,-pdpe1gb,-avx512vl` on macOS when running on x86_64 host? Or is that "too much information"?

AkihiroSuda: Reworded as `"host" when running on x86_64 host; additional options are appended on Intel Mac`

jandubois (approved):
Thanks, LGTM

chadlwilson:
Thanks for looking at this @AkihiroSuda . Looking forward to being able to build/test Almalinux 10  containers on MacOS via colima 🙏

PR #3488, opened 2025-05-02
qemu: experimental support for PPC64LE

AkihiroSuda:
Usage:
```
limactl start --arch ppc64le
```


pkg/qemu/qemu.go
```diff
@@ -540,7 +540,8 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 			}
 		}
 	}
-	if !strings.Contains(string(features.CPUHelp), strings.Split(cpu, ",")[0]) {
+	// `qemu-system-ppc64 -help` does not show "max", but it is actually accepted
+	if cpu != "max" && !strings.Contains(string(features.CPUHelp), strings.Split(cpu, ",")[0]) {
```
jandubois, line 544: Technically it should probably be:

```suggestion
	if (*y.arch != limayaml.PPC64LE || cpu != "max") && !strings.Contains(string(features.CPUHelp), strings.Split(cpu, ",")[0]) {
```

But I guess we just assume that `max` is valid for all CPUs now.

jandubois (approved):
Thanks, LGTM

I have not verified if the image updater scripts will work for the new PPC64LE images or not. Please file an issue if the scripts still need updating!

AkihiroSuda:
> Thanks, LGTM
> 
> I have not verified if the image updater scripts will work for the new PPC64LE images or not. Please file an issue if the scripts still need updating!

Seems to work

Issue #320, opened 2021-10-12
docs: find out the oldest QEMU version that is compatible with Lima


AkihiroSuda:
QEMU v2.11 is known not to work with Lima v0.7 due to lack of `-netdev help`: https://github.com/rancher-sandbox/rancher-desktop/issues/771#issuecomment-940702172

v2.12 may probably work, but I didn't test.

jandubois:
Given that `-netdev help` output is only used to determine if `vde` is supported, shouldn't we simply set `features.NetDevHelp = []byte{}` instead of throwing an error, if that would allow us to run on Ubuntu 18.04?

AkihiroSuda:
Yes, if that is enough to support Ubuntu 18.04.
I guess more flags were missing in that version of QEMU, though.

afbjorklund:
Some more data points:
* the qemu version in RHEL 7 is too old (1.5.3), the syntax for qemu-img and qemu-system has changed since
* the qemu version in RHEL 8 is still ok (6.2.0)

afbjorklund:
Ubuntu 18.04 is soon (May 31) out of standard support, but still supported with extended support... (Ubuntu Pro)

https://ubuntu.com/blog/18-04-end-of-standard-support

```
qemu-system/bionic-updates 1:2.11+dfsg-1ubuntu7.42 amd64
```

```console
$ qemu-system-x86_64 -netdev help
qemu-system-x86_64: -netdev help: Parameter 'id' is missing
```

AkihiroSuda:
This should suffice for now
https://github.com/lima-vm/lima/blob/68be6f2b9bfc4ca171e626917057f74f2c78dbbf/pkg/driver/qemu/qemu.go#L77-L79

PR #3491, opened 2025-05-02
Clean up codes for QEMU < 7.0

AkihiroSuda:
- QEMU 7.0 is documented as the minimum requirement since 9eb13960f1f5b581e0ba814a70741234695c7672 (Dec 2022)
- Also clean up the hint for QEMU 8.2.0, as QEMU 8.2.0 users should have already updated QEMU to 8.2.1+
- QEMU 4.0 is still exceptionally allowed on linux/amd64 for the compatibility reason, but not tested
- QEMU 6.2 is added to the CI

### DIFF
--- a/pkg/qemu/qemu.go
+++ b/pkg/qemu/qemu.go
@@ -47,10 +47,42 @@ type Config struct {
 	VirtioGA     bool
 }
 
-// MinimumQemuVersion is the minimum supported QEMU version.
-const (
-	MinimumQemuVersion = "4.0.0"
-)
+// minimumQemuVersion returns hardMin and softMin.
+//
+// hardMin is the hard minimum version of QEMU.
+// The driver immediately returns the error when QEMU is older than this version.
+//
+// softMin is the oldest recommended version of QEMU.
+// softMin must be >= hardMin.
+//
+// When updating this function, make sure to update
+// `website/content/en/docs/config/vmtype.md` too.
+func minimumQemuVersion() (hardMin, softMin semver.Version) {
+	var h, s string
+	switch runtime.GOOS {
+	case "darwin":
+		switch runtime.GOARCH {
+		case "arm64":
+			// https://gitlab.com/qemu-project/qemu/-/issues/1990
+			h, s = "8.2.1", "8.2.1"
+		default:
+			// The code specific to QEMU < 7.0 on macOS (https://github.com/lima-vm/lima/pull/703)
+			// was removed in https://github.com/lima-vm/lima/pull/3491
+			h, s = "7.0.0", "8.2.1"
+		}
+	default:
+		// hardMin: Untested and maybe does not even work.
+		// softMin: Ubuntu 22.04's QEMU. The oldest version that can be easily tested on GitHub Actions.
+		h, s = "4.0.0", "6.2.0"
+	}
+	hardMin, softMin = *semver.New(h), *semver.New(s)
+	if softMin.LessThan(hardMin) {
+		// NOTREACHED
+		logrus.Fatalf("internal error: QEMU: soft minimum version %v must be >= hard minimum version %v",
+			softMin, hardMin)
+	}
+	return hardMin, softMin
+}
 
 // EnsureDisk also ensures the kernel and the initrd.
 func EnsureDisk(ctx context.Context, cfg Config) error {
@@ -287,9 +319,6 @@ type features struct {
 	// e.g. "Available CPUs:\n...\nx86 base...\nx86 host...\n...\n"
 	// Not machine-readable, but checking strings.Contains() should be fine.
 	CPUHelp []byte
-
-	// VersionGEQ7 is true when the QEMU version seems v7.0.0 or later
-	VersionGEQ7 bool
 }
 
 func inspectFeatures(exe, machine string) (*features, error) {
@@ -333,7 +362,6 @@ func inspectFeatures(exe, machine string) (*features, error) {
 			f.MachineHelp = stderr.Bytes()
 		}
 	}
-	f.VersionGEQ7 = strings.Contains(string(f.MachineHelp), "-7.0")
 
 	// Avoid error: "No machine specified, and there is no default"
 	cmd = exec.Command(exe, "-cpu", "help", "-machine", machine)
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
Restored the codes for macOS 12

afbjorklund:
I think originally the request was to be able to use Lima with the system version of QEMU,
especially since one use case was to run the newer kernel in an Lima instance if needed...

Ubuntu | QEMU
---- | ----
~~20.04~~ | 4.2
22.04 | 6.2
24.04 | 8.2

RHEL | QEMU
---- | ----
~~7~~ | 1.5
8 | 6.2
9 | 9.0

So it would be nice to keep the _minimum_ version at 6.2, unless 7.0 is somehow required?
Otherwise one has to build a newer version locally, which makes Lima harder to install.

* https://github.com/lima-vm/lima/issues/320

AkihiroSuda:
Maybe we should allow QEMU 6.2 only on linux/amd64 host?
We should also have a ubuntu-22.04 CI?

AkihiroSuda:
Added QEMU 6.2 CI (ubuntu-22.04)

pkg/qemu/qemu.go
```diff
@@ -38,6 +38,10 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// RecommendedQemuVersion is the recommemded QEMU version.
+// See commit 8c48cb3eef178914f1b8fffee324aab0c445fec5 .
```
jandubois, line 42: Is this reference still needed? We use `host: max` now instead of `host: cortex-a76`.

So maybe the comment should be updated that we use `-cpu max` by default, which requires QEMU 7.0.0?

pkg/qemu/qemu_linux_amd64.go
```diff
@@ -0,0 +1,9 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package qemu
+
+// MinimumQemuVersion is the minimum supported QEMU version.
+// This is set to a lower version on linux/amd64 hosts due to a compatibility reason.
+// See qemu_others.go for the "real" minimum supported QEMU version.
+const MinimumQemuVersion = "4.0.0"
```
jandubois, line 9: Is this actually going to work with the defaults? Won't we have to change `-cpu max` to `-cpu host`?

pkg/qemu/qemu.go
```diff
@@ -500,28 +451,26 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 		if version.LessThan(*semver.New(MinimumQemuVersion)) {
 			logrus.Fatalf("QEMU %v is too old, %v or later required", version, MinimumQemuVersion)
 		}
+		if version.LessThan(*semver.New(RecommendedQemuVersion)) {
+			logrus.Warnf("QEMU %v is too old, %v or later is recommended", version, RecommendedQemuVersion)
+		}
 		if y.VMOpts.QEMU.MinimumVersion != nil && version.LessThan(*semver.New(*y.VMOpts.QEMU.MinimumVersion)) {
 			logrus.Fatalf("QEMU %v is too old, template requires %q or later", version, *y.VMOpts.QEMU.MinimumVersion)
 		}
-		if runtime.GOOS == "darwin" && runtime.GOARCH == "arm64" && version.Equal(*semver.New("8.2.0")) {
```
jandubois, line 506: Is there a reason to remove this warning? It is unlikely it will ever be triggered, but checks for a bug in a QEMU version higher than the recommended version.

AkihiroSuda: Maybe we should just set `RecommendedQemuVersion = "8.2.1"` in `qemu_darwin.go`?

pkg/qemu/qemu_linux_amd64.go
AkihiroSuda: idk. Which version introduced `-cpu max`?

pkg/qemu/qemu.go
jandubois: Fine with me

pkg/qemu/qemu_linux_amd64.go
jandubois: I believe `-cpu max` was added in 7.0.0, but I don't know if for all architectures, or if some came later.

jandubois: ChatGPT says `-cpu: max` was introduced gradually:

* x86_64: version 4
* arm64: version 6
* ppc64: version 7.1 or 7.2
* riscv64: version 7
* s390x: not supported at all (is this correct?)

AkihiroSuda: > s390x: not supported at all (is this correct?)

Nope 😅 

`s390x/cpumodel: Add "-cpu max" support` (v3.1.0)
https://github.com/qemu/qemu/commit/c61177881cb

AkihiroSuda: > x86_64: version 4

ChatGPT is just hallucinating 🤖 
Actually it seems added in v2.1.0 ("i386" seems to cover x86_64 too)
https://github.com/qemu/qemu/commit/c62f2630f8e8765677020ab0a464f94d4ac433b6

pkg/qemu/qemu.go
jandubois: I'm sorry, I still think this warning should stay unless you set `MinimumVersion` to `8.2.1` on macOS. I think this would be fine to do; it is just Linux that may need ancient QEMU versions.

AkihiroSuda: Updated to set minimum to 8.2.1 for arm64

jandubois (approved):
Thanks, LGTM

Personally not a big fan of having 5 different files to set a combination of 2 variables. This could be single function with runtime logic instead, that would be easier to understand. But fine if you prefer to do it this way.

pkg/qemu/qemu.go
```diff
@@ -47,10 +47,43 @@ type Config struct {
 	VirtioGA     bool
 }
 
-// MinimumQemuVersion is the minimum supported QEMU version.
-const (
-	MinimumQemuVersion = "4.0.0"
-)
+// HardMinimumVersion is the hard minimum version of QEMU.
+// The driver immediately returns the error when QEMU is older than this version.
+func HardMinimumVersion() string {
```
jandubois, line 52: I don't think these functions need to be exported.

jandubois (approved):
Thanks, LGTM

pkg/qemu/qemu.go
AkihiroSuda: fixed

jandubois (approved):
Thanks, LGTM

PR #3500, opened 2025-05-06
refactor: migrate `cpuType` to `vmOpts.qemu`

### LINKED ISSUE #3486: YAML: `cpuType` should be moved to `vmOpts.qemu`
https://github.com/lima-vm/lima/blob/94ed78c8319bc33089886dc8776effaa640392fc/templates/default.yaml#L321-L340

`cpuType` is specific to QEMU and not portable to other VM drivers, so it should be moved to `vmOpts.qemu`.
The implementation should be moved from `pkg/limayaml` to `pkg/qemu` too.

For compatibility the old form should continue to be supported for a while.

unsuman:
## Description
This PR fixes #3486 by relocating the `cpuType` field to `vmOpts.qemu.cpuType` in both the YAML schema and `limayaml.LimaYAML` struct in Go. To maintain compatibility, the old top-level `cpuType` is still accepted and merged into the new path at runtime, with a warning to encourage migration. Validation and default logic have been moved to the qemu package.


pkg/limayaml/validate.go
```diff
@@ -119,12 +119,22 @@ func Validate(y *LimaYAML, warn bool) error {
...
 		if !slices.Contains(ArchTypes, arch) {
 			return fmt.Errorf("field `cpuType` uses unsupported arch %q", arch)
 		}
 	}
 
+	for arch := range y.VMOpts.QEMU.CPUType {
+		if !slices.Contains(ArchTypes, arch) {
+			return fmt.Errorf("field `vmOpts.qemu.cpuType` uses unsupported arch %q", arch)
+		}
+	}
```
AkihiroSuda, line 136: Probably `defaults.go` should just replicate the top-level CPUType to VMOpts (with a warning when they conflict), then no need to validate the CPUTypes here twice?

AkihiroSuda: The validation should happen in `pkg/qemu`.
`pkg/limayaml` shouldn't be aware of the archs implemented by the QEMU driver

pkg/limayaml/limayaml.go
```diff
@@ -111,6 +112,7 @@ type VMOpts struct {
 
 type QEMUOpts struct {
 	MinimumVersion *string `yaml:"minimumVersion,omitempty" json:"minimumVersion,omitempty" jsonschema:"nullable"`
+	CPUType        CPUType `yaml:"cpuType,omitempty" json:"cpuType,omitempty" jsonschema:"nullable"`
 }
```
AkihiroSuda, line 116: Eventually `VMOpts` should be defined as `map[string]any`, and `QEMUOpts` should be moved to `pkg/qemu`.
Can be another PR.

unsuman: Should I make the changes in this PR itself, or raise an issue from your comment, and start working on it in another PR?

AkihiroSuda: Can be another issue and PR

pkg/limayaml/defaults.go
```diff
@@ -312,28 +271,25 @@ func FillDefault(y, d, o *LimaYAML, filePath string, warn bool) {
...
+		y.VMOpts.QEMU.CPUType = CPUType{}
 	}
-	for k, v := range y.CPUType {
-		if v != "" {
-			overrideCPUType = true
-			cpuType[k] = v
+	// TODO: This check should be removed when we completely eliminate `CPUType` from limayaml.
+	if len(y.CPUType) > 0 {
+		if warn {
+			logrus.Warn("The top-level `cpuType` field is deprecated and will be removed in a future release. Please migrate to `vmOpts.qemu.cpuType`.")
```
unsuman, line 280: I don't know why this is not executing even if `warn` is set `true` by default

templates/default.yaml
```diff
@@ -324,11 +324,21 @@ vmOpts:
...
+    #   aarch64: "max" # (or "host" when running on aarch64 host)
+    #   armv7l:  "max" # (or "host" when running on armv7l host)
+    #   riscv64: "max" # (or "host" when running on riscv64 host)
+    #   x86_64:  "max" # (or "host" when running on x86_64 host; additional options are appended on Intel Mac)
 
 # OS: "Linux".
 # 🟢 Builtin default: "Linux"
 os: null
 
+# DEPRECATED: Use vmOpts.qemu.cpuType instead.
```
jandubois, line 341: I think the duplicate documentation should be removed from the deprecated entry. Just point it to the new location instead.

jandubois (commented):
There is also a linter error to fix.

jandubois (changes requested):
I would prefer that we hold off on merging this PR until we have agreement on how the final config settings are supposed to look like. Otherwise we might end up with double-migrations, which complicates the code, the docs, and confuses users.

See also https://github.com/lima-vm/lima/discussions/3520#discussioncomment-13094218

AkihiroSuda:
Can we merge? @jandubois

jandubois (approved):
Thanks, LGTM

jandubois (approved):
Thanks, LGTM

Issue #3325, opened 2025-03-10
Consider setting `GUESTAGENT_COMPRESS=y` by default


AkihiroSuda:
https://github.com/lima-vm/lima/blob/ddc0081a241b05d3d7379480613f2c2c79ab2136/Kconfig#L34-L38

- https://github.com/lima-vm/lima/pull/2438

This should be enabled by default unless there is a significant drawback?

afbjorklund:
It should be a little slower (due to legacy gzip), but not more than that.

kairveeehh:
would like to take up this issue , should I follow up with a PR or wait for assignment?

AkihiroSuda:
You can open a PR at anytime without waiting for an assignment.

Things to consider:
- Confirm that the overhead isn't more than a few seconds on typical machines
- Print a warning when the both compressed agent and the uncompressed one are present. The one with the newer timestamp should be selected.

kairveeehh:
> You can open a PR at anytime without waiting for an assignment.
> 
> Things to consider:
> - Confirm that the overhead isn't more than a few seconds on typical machines
> - Print a warning when the both compressed agent and the uncompressed one are present. The one with the newer timestamp should be selected.

Working on it.

AkihiroSuda:
@kairveeehh 

> Working on it.

Are you still working on this?

kairveeehh:
> @kairveeehh 
> 
> > Working on it.
> 
> Are you still working on this?

Yes , was a bit busy cause of uni exams but I'll continue to work on it.

AkihiroSuda:
This issue will be fixable in a single-line PR, after finishing this refactoring:
- https://github.com/lima-vm/lima/pull/3504

AkihiroSuda:
@kairveeehh Do you want to open this PR?

```diff
diff --git a/Kconfig b/Kconfig
index d52d6fb0..7e40d787 100644
--- a/Kconfig
+++ b/Kconfig
@@ -47,4 +47,4 @@ config GUESTAGENT_COMPRESS
        bool "guestagent compress"
        help
                Compress lima-guestagent
-       default n
+       default y
diff --git a/config.mk b/config.mk
index 9a02aae2..06ab7787 100644
--- a/config.mk
+++ b/config.mk
@@ -5,4 +5,4 @@ CONFIG_GUESTAGENT_ARCH_ARMV7L=y
 CONFIG_GUESTAGENT_ARCH_PPC64LE=y
[...]

PR #3503, opened 2025-05-07
Makefile: split lima-additional-guestagents-*.tar.gz from  lima-*.tar.gz

AkihiroSuda:
`make artifact` now produces two archives that will be uploaded to <https://github.com/lima-vm/lima/releases>:

1. `lima-<VERSION>-Darwin-arm64.tar.gz`: (`make native`)
  - core components (bin/limactl, share/lima/templates, ...)
  - share/lima/lima-guestagent.Linux-aarch64

2. `lima-additional-guestagents-<VERSION>-Darwin-arm64.tar.gz`: (`make additional-guestagents`)
  - share/lima/lima-guestagent.Linux-armv7l
  - share/lima/lima-guestagent.Linux-ppc64le
  - share/lima/lima-guestagent.Linux-riscv64
  - share/lima/lima-guestagent.Linux-s390x
  - share/lima/lima-guestagent.Linux-x86_64

For compatibility reason, `make` still builds the guestagents for all the architectures by default.

Package maintainers are suggested to split `lima` package to `lima` and `lima-additional-guestagents`.

Fix #3321

PR #3504, opened 2025-05-07
guestagent refactoring

AkihiroSuda:
## Commit 1: guestagent:  prioritize BIN.gz over BIN
    
When both `lima-guestagent.Linux-<ARCH>` and `lima-guestagent.Linux-<ARCH>.gz` are present, the compressed one is chosen with a warning.
    
Preparation for:
- #3325

- - -
## Commit 2: limactl info: show the detected guest agent paths

- - -

## Commit 3: hostagent: verify existence of guestagent earlier

### DIFF
--- a/pkg/usrlocalsharelima/usrlocalsharelima.go
+++ b/pkg/usrlocalsharelima/usrlocalsharelima.go
@@ -80,6 +80,7 @@ func Dir() (string, error) {
 		ostype, arch, self, gaCandidates)
 }
 
+// GuestAgentBinary returns the guest agent binary, possibly with ".gz" suffix.
 func GuestAgentBinary(ostype limayaml.OS, arch limayaml.Arch) (string, error) {
 	if ostype == "" {
 		return "", errors.New("os must be set")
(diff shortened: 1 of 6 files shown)


pkg/usrlocalsharelima/usrlocalsharelima.go
```diff
@@ -91,5 +94,43 @@ func GuestAgentBinary(ostype limayaml.OS, arch limayaml.Arch) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return filepath.Join(dir, "lima-guestagent."+ostype+"-"+arch), nil
+	uncomp := filepath.Join(dir, "lima-guestagent."+ostype+"-"+arch)
+	comp := uncomp + ".gz"
+	res, err := newest([]string{comp, uncomp})
+	if err != nil {
+		logrus.Debug(err)
+		return "", fmt.Errorf("guest agent binary could not be found for %s-%s (Hint: try installing `lima-additional-guestagents` package)", ostype, arch)
```
AkihiroSuda, line 102: lima-additional-guestagents is being packaged in:
- #3503

AkihiroSuda:
> I'm not sure which scenario this is supposed to address; why would somebody have both a compressed and an uncompressed version? Does this only happen for Lima developers / maintainers, or can regular users run into this situation.

Regular users may hit this when unpacking v1.1.x.tar.gz (tentative) after unpacking v1.0.x.tar.gz

jandubois:
> Regular users may hit this when unpacking v1.1.x.tar.gz (tentative) after unpacking v1.0.x.tar.gz

In that case I would still prefer that we will always pick the compressed version, if available, and show a warning when both exist, with a suggestion to delete the uncompressed one.

And with 1.1.0 going forward we should only be releasing the compressed version.

AkihiroSuda:
> In that case I would still prefer that we will always pick the compressed version, if available, and show a warning when both exist, with a suggestion to delete the uncompressed one.

Updated

jandubois (dismissed):
Thanks, LGTM

Only one comment looks off

jandubois (approved):
Thanks, LGTM

PR #3513, opened 2025-05-08
chore: move CreateDataDisk and ResizeDataDisk to util

### LINKED ISSUE #3498: pkg/qemu: split qemu-img functions to imgutil
https://github.com/lima-vm/lima/blob/1f787fb94457ddc15fc44ba1d0928ed57469c305/pkg/qemu/qemu.go#L130-L154

These functions above should be moved from `pkg/qemu` to `pkg/qemu/imgutil`, so that `cmd/limactl` does not need to depend on `pkg/qemu`

https://github.com/lima-vm/lima/blob/1f787fb94457ddc15fc44ba1d0928ed57469c305/cmd/limactl/disk.go#L117

liangyuanpeng:
fixes #3498

### DIFF
--- a/cmd/limactl/disk.go
+++ b/cmd/limactl/disk.go
@@ -16,7 +16,7 @@ import (
 	"github.com/docker/go-units"
 	"github.com/lima-vm/go-qcow2reader"
 	"github.com/lima-vm/lima/pkg/nativeimgutil"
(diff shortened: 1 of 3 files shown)


jandubois (approved):
Thanks, LGTM

PR #3515, opened 2025-05-08
Move jsonschema validation into a utility dir

### LINKED ISSUE #3510: genschema.go shouldn't import multiple jsonschema libraries
https://github.com/lima-vm/lima/blob/00e3398febfc0aabed19c54f7e751dcaf5af5b43/cmd/limactl/genschema.go#L13-L15

Shouldn't need both "github.com/invopop/jsonschema" and "github.com/santhosh-tekuri/jsonschema/v6

afbjorklund:
This performs the same task as py "check-jsonschema" or "jv".

But included with Lima, instead of being a separate CLI tool.

Add simple example, for unit test.

See https://tour.json-schema.org/

Closes #3510

It is a little wasteful to compile the schema everytime, but on the other hand it doesn't take long to run.

~~We don't _actually_ validate that the jsonschema library works here, but we could do that if we wanted to...~~

### DIFF
--- a/cmd/limactl/genschema.go
+++ b/cmd/limactl/genschema.go
@@ -9,10 +9,9 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/goccy/go-yaml"
 	"github.com/invopop/jsonschema"
+	"github.com/lima-vm/lima/pkg/jsonschemautil"
 	"github.com/lima-vm/lima/pkg/limayaml"
-	jsonschema2 "github.com/santhosh-tekuri/jsonschema/v6"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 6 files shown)


pkg/jsonschemautil/testdata/instance.json
```diff
@@ -0,0 +1,4 @@
+{
```
jandubois, line 1: Why does this file exist? It seems to be unused by the test.

afbjorklund: It was the original upstream test, I converted it to yaml

AkihiroSuda (approved):
Thanks

Issue #3497, opened 2025-05-06
pkg/store/disk.go: `inspectDiskSize` does not need to fallback to qemu-img


AkihiroSuda:
https://github.com/lima-vm/lima/blob/1f787fb94457ddc15fc44ba1d0928ed57469c305/pkg/store/disk.go#L64-L82

go-qcow2reader should suffice to inspect the disk size.

I don't remember why I added `inspectDiskSizeWithQemuImg` fall-back in https://github.com/lima-vm/lima/commit/20c2224bc1f74d662dc2b32506782d5cd6450b07 . Probably go-qcow2reader was immature at that time

PR #3524, opened 2025-05-09
removing qemu fallback for inspectDiskSize

jonas-peter:
fixes https://github.com/lima-vm/lima/issues/3497

### DIFF
--- a/pkg/store/disk.go
+++ b/pkg/store/disk.go
@@ -11,7 +11,6 @@ import (
 	"path/filepath"
 
 	"github.com/lima-vm/go-qcow2reader"
-	"github.com/lima-vm/lima/pkg/qemu/imgutil"
 	"github.com/lima-vm/lima/pkg/store/filenames"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/store/disk.go
```diff
@@ -61,35 +60,25 @@ func InspectDisk(diskName string) (*Disk, error) {
...
 	defer f.Close()
 	img, err := qcow2reader.Open(f)
 	if err != nil {
-		return inspectDiskWithQemuImg(fName)
+		return -1, "", err
 	}
 	sz := img.Size()
 	if sz < 0 {
-		return inspectDiskWithQemuImg(fName)
+		return -1, "", err
```
jandubois, line 76: This doesn't work because `err` is `nil` here, and the caller only checks for `err` and not for `size < 0` as an error condition. So it should be something like

```suggestion
		return -1, "", fmt.Errorf("cannot determine size of %q", fName)
```

jonas-peter: yes, of course, sorry!

AkihiroSuda (approved):
Thanks

jandubois:
> fixing #3497

@jonas-peter Note that this format does not automatically link the PR to the issue. You need to use words recognized by GitHub, like "Fixes" or "Closes". See [Linking a pull request to an issue using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) for the full list.

jonas-peter:
> > fixing #3497
> 
> 
> 
> @jonas-peter Note that this format does not automatically link the PR to the issue. You need to use words recognized by GitHub, like "Fixes" or "Closes". See [Linking a pull request to an issue using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) for the full list.

Thanks for the heads up!

PR #3529, opened 2025-05-10
Kconfig: set CONFIG_GUESTAGENT_COMPRESS=y by default

### LINKED ISSUE #3325: Consider setting `GUESTAGENT_COMPRESS=y` by default

AkihiroSuda:
Fix #3325

Before:
```console
$ du -hs _output
311M    _output
```

After:
```console
$ du -hs _output
104M    _output
```


jandubois (approved):
Thanks, LGTM

PR #3531, opened 2025-05-12
CI: set GOTOOLCHAIN=local to prevent auto update of Go

AkihiroSuda:
https://go.dev/doc/toolchain


jandubois (approved):
Thanks, LGTM

afbjorklund:
I had this in my local environment too, to restore the old behaviour (i.e. not `auto`).

PR #3533, opened 2025-05-12
Support editing instance disk size via flag

songponssw:
Add support for editing instance disk size using the `--disk` flag.
e.g. `limactl edit default --disk 20`

### DIFF
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -66,6 +66,11 @@ func registerEdit(cmd *cobra.Command, commentPrefix string) {
 
 	// negative performance impact: https://gitlab.com/qemu-project/qemu/-/issues/334
 	flags.Bool("video", false, commentPrefix+"Enable video output (has negative performance impact for QEMU)")
+
+	flags.Float32("disk", 0, commentPrefix+"Disk size in GiB") // colima-compatible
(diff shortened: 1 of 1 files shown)


cmd/limactl/editflags/editflags.go
```diff
@@ -66,6 +66,11 @@ func registerEdit(cmd *cobra.Command, commentPrefix string) {
 
 	// negative performance impact: https://gitlab.com/qemu-project/qemu/-/issues/334
 	flags.Bool("video", false, commentPrefix+"Enable video output (has negative performance impact for QEMU)")
+
+	flags.Float32("disk", 0, commentPrefix+"Disk size in GiB") // colima-compatible
+	_ = cmd.RegisterFlagCompletionFunc("memory", func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
```
jandubois, line 71: I know you just moved the call, but it looks like it is incorrect:

```suggestion
	_ = cmd.RegisterFlagCompletionFunc("disk", func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
```

jandubois (changes requested):

cmd/limactl/editflags/editflags.go
songponssw: Thanks so much! I've already squashed the commit.

jandubois (approved):
Thanks, LGTM

I wish we would detect the attempt to downsize the disk size, and then reject the change as invalid, but that can be a separate PR.

PR #3535, opened 2025-05-12
Revert yq/v4 from v4.45.2 → v4.45.1

jandubois:
4.45.2 introduces regressions that are being fixed by 4.45.3 and 4.45.4, but 4.45.4 also bumps the required Go version to 1.24 while Lima wants to stay compatible with 1.23 for now.

Therefore we should revert the bump that cause the regressions.

Revert is done manually using `go mod tidy` because `git revert ef376e4` has messy merge conflicts.

See https://github.com/lima-vm/lima/pull/3530#pullrequestreview-2831846295


AkihiroSuda (approved):
Thanks

Issue #3334, opened 2025-03-11
vz on Intel: kernel 6.12 doesn't boot (when `CONFIG_X86_INTEL_MEMORY_PROTECTION_KEYS` is set)


AkihiroSuda:
Linux kernel v6.12 does not boot with vz on Intel.

No error message is printed on video nor on serial.

<details>
<summary>✅ Boots</summary>
<p>

- https://cloud-images.ubuntu.com/releases/24.10/release-20250129/ubuntu-24.10-server-cloudimg-amd64.img (kernel 6.11.0-14-generic)
- https://download.fedoraproject.org/pub/fedora/linux/releases/41/Cloud/x86_64/images/Fedora-Cloud-Base-Generic-41-1.4.x86_64.qcow2 (kernel 6.11.4-301.fc41.x86_64)
- https://dl-cdn.alpinelinux.org/alpine/v3.21/releases/cloud/nocloud_alpine-3.21.2-x86_64-uefi-cloudinit-r0.qcow2 (kernel 6.12.8-0-virt)

</p></details>

<details>
<summary>❌ Doesn't boot (still boots with QEMU)</summary>
<p>

- https://cloud-images.ubuntu.com/plucky/20250310/plucky-server-cloudimg-amd64.img (kernel 6.12.0-16-generic)
- https://geo.mirror.pkgbuild.com/images/v20250201.304316/Arch-Linux-x86_64-cloudimg-20250201.304316.qcow2 (kernel 6.12.10-arch1-1)
- https://download.opensuse.org/tumbleweed/appliances/openSUSE-Tumbleweed-Minimal-VM.x86_64-Cloud.qcow2 (version 20250308, kernel 6.13.5-1-default)
- https://dl.fedoraproject.org/pub/fedora/linux/development/rawhide/Cloud/x86_64/images/Fedora-Cloud-Base-Generic-Rawhide-20250310.n.0.x86_64.qcow2 (kernel 6.14.0-0.rc5.20250307git00a7d39898c8.47.fc43.x86_64)

</p></details>

**The issue might be a regression in kernel 6.12, although Alpine still boots with kernel 6.12.**

> [!NOTE]
[...]

PR #3537, opened 2025-05-13
vz: Intel Mac: warn if macOS < 15.5

AkihiroSuda:
On Intel Mac with vz, macOS 15.5 or later is needed to boot Linux 6.12 and later. (Ubuntu 25.04, Fedora 42, etc.)

Update issue:
- #3334


jandubois (approved):
Thanks, LGTM

Issue #3516, opened 2025-05-08
hugo/docsy issue when trying to preview website changes


jandubois:
I'm unable to preview the website locally

```
❯ make serve
npm run serve

> lima-site@0.7.1 serve
> npm run _serve


> lima-site@0.7.1 _serve
> npm run _hugo-dev -- --minify serve


> lima-site@0.7.1 _hugo-dev
> npm run _hugo -- -e dev -DFE --minify serve


> lima-site@0.7.1 _hugo
> hugo --cleanDestinationDir -e dev -DFE --minify serve

Error: template: swagger/list.html:8: bad character U+002D '-'
make: *** [serve] Error 1

❯ hugo version
hugo v0.147.2+extended+withdeploy darwin/arm64 BuildDate=2025-05-06T11:18:55Z VendorInfo=brew
```

songponssw:
I tried to reproduce this issue on Linux, and I got the same error.

```
thor@linux-01:~/lima/website$ npm list
lima-site@0.7.1 /home/thor/lima/website
├── autoprefixer@10.4.20
├── hugo-extended@0.147.2
└── postcss-cli@10.1.0

thor@linux-01:~/lima/website$ make serve
make -C .. docsy
make[1]: Entering directory '/home/thor/lima'
Guestagents are unzipped each time to check the build configuration; they may be gunzipped afterward.
_output/bin/limactl generate-doc --type docsy website/_output/docsy \
	--output _output --prefix /usr/local
INFO[0000] Replacing "/home/thor/lima/_output" with "/usr/local"
INFO[0000] Replacing "/home/thor" with "~"
make[1]: Leaving directory '/home/thor/lima'
npm run serve

> lima-site@0.7.1 serve
> npm run _serve


> lima-site@0.7.1 _serve
> npm run _hugo-dev -- --minify serve


> lima-site@0.7.1 _hugo-dev
> npm run _hugo -- -e dev -DFE --minify serve


> lima-site@0.7.1 _hugo
> hugo --cleanDestinationDir -e dev -DFE --minify serve

Error: template: swagger/list.html:8: bad character U+002D '-'
make: *** [Makefile:7: serve] Error 1
```

It seems there's a compatibility between docsy and hugo-extended.

**For docsy version 0.7.1 (which is used in Lima), it works fine up to hugo-extended@0.119.0.**

[...]

PR #3085, opened 2025-01-05
website: add simple makefile for testing locally

afbjorklund:
Something like: `cd website && make serve`

Then you can open <http://localhost:1313/>

Closes #3079

I used hugo-extended 0.113.0 (from package.json)

---

Would recommend using something like: [nvm](https://github.com/nvm-sh/nvm) for the `npm` (and `node`) installation and versioning

The "hugo" program can be found on https://gohugo.io/ and "docsy" is found on https://docsy.dev/

PR #3538, opened 2025-05-13
Update docsy to version 0.11.0 and hugo to 0.133.1

afbjorklund:
* Issue #3516

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -3,11 +3,14 @@
 # See also: package.json
 NPM ?= npm
 
-build serve: _output/docsy
+build serve: _output/docsy node_modules/.bin/hugo
 	$(NPM) run $@
 
 _output/docsy:
 	$(MAKE) -C .. docsy
+
+node_modules/.bin/hugo:
+	$(NPM) install
 
 clean:
 	$(NPM) run $@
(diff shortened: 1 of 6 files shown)


afbjorklund:
You need to install the supported hugo version from the npm install (package.json), not the latest/greatest

jandubois (dismissed):
Thanks, LGTM

I would prefer to see some comments about installing the right hugo version in `website/Makefile`, maybe even with a link to https://github.com/gohugoio/hugo/releases/tag/v0.133.1

afbjorklund:
The Makefile was my attempt to make it a little easier, we can probably add a pre-requisite line.

* https://github.com/lima-vm/lima/pull/3085

It also needs to document how to use nvm to get the right version of node.js, so it can get hairy.

jandubois (approved):
Thanks, LGTM

afbjorklund:
The new version does have some subtle improvements, and one "moderate security issue" fixed.

OLD
lima-docs-old
NEW
lima-docs-new

jandubois:
> I think the local version takes predence over any system version, I installed hugo 0.145.0 and it still worked

Looks like I forgot to run `npm install` before running `make serve`.

> It does have to download a few things, but for a node program it is not "too bad" (there are much worse)
> 
> `95M node_modules/`

tfugj4n3l6ez

PR #3543, opened 2025-05-14
Always return an error when a requested instance is not found

### LINKED ISSUE #3541: limactl list <SOMETHING> exits cleanly if VM list is empty
### Description

Normally, when I have some VMs on my machine, if I do a limactl list VM_NAME with a vm that doesn't exist, I get an exit status of 1

```
% limactl list aiodfjwa9r2395r239
WARN[0000] No instance matching aiodfjwa9r2395r239 found.
NAME    STATUS    SSH    VMTYPE    ARCH    CPUS    MEMORY    DISK    DIR
% echo $?
1
```

BUT!

If no VM has been created yet in lima, limactl ignores the instance name and just returns a 0 as if the arg wasn't passed at all:

```
% limactl list aiodfjwa9r2395r239
WARN[0000] No instance found. Run `limactl create` to create an instance.
% echo $?
0
```

This breaks some automated workflows like "create this VM if it doesn't exist already".

Expectation would be that if VM name doesn't exist, exit status is 1 regardless of other VMs presence or absence.

jandubois:
Fixes #3541

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -134,11 +134,11 @@ func listAction(cmd *cobra.Command, args []string) error {
 		logrus.Warnf("The directory %q does not look like a valid Lima directory: %v", store.Directory(), err)
 	}
 
-	allinstances, err := store.Instances()
+	allInstances, err := store.Instances()
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #3556, opened 2025-05-20
Rename infoutil.GetInfo() to info.New()

jandubois:
As discussed on Slack, rename `infoutil` to `limainfo`[^1] because we are over-using the `util` suffix in our package names.

[^1]: I suggested to use just `info`, but think `limainfo` is a better name, that is still reasonably short.


AkihiroSuda (approved):
Thanks

PR #3557, opened 2025-05-20
fix(cidata): do not overwrite bash in alpine vm

### LINKED ISSUE #3555: `01-alpine-ash-as-bash.sh` overwrites existing bash executable
## Description

[01-alpine-ash-as-bash.sh](https://github.com/lima-vm/lima/blob/master/pkg/cidata/cidata.TEMPLATE.d/boot/01-alpine-ash-as-bash.sh) overwrites `/bin/bash` unconditionally, causing bash unusable in Alpine VM.

## Steps To Reproduce

Create a new VM with default Alpine template, then run the following:

```sh
$ sudo apk add bash
$ bash --version
GNU bash, version 5.2.37(1)-release (aarch64-alpine-linux-musl)
```

Then reboot the VM, and run `bash --version` again:

```sh
$ bash --version
/bin/ash: bad option '--version'
```

Bash has been overwritten as ash, even if bash was explicitly installed.

## Host Info

Lima 1.0.7

macOS 15.4.1

MacBook Pro M1 Max

---

Is it a good idea to check the existence of `/bin/bash` before overwriting, and do nothing if it already exists?

If you need, I'm happy to submit a PR.

ycdzj:
Fixes #3555

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/01-alpine-ash-as-bash.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/01-alpine-ash-as-bash.sh
@@ -7,6 +7,9 @@
 # cloud-init scripts can use `#!/bin/bash` and `set -o pipefail`.
 test -f /etc/alpine-release || exit 0
 
+# If bash already exists, do nothing.
+test -x /bin/bash && exit 0
+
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #3564, opened 2025-05-21
Add `bin/lima-additional-guestagents-path`

AkihiroSuda:
When the script is installed as `<PREFIX>/bin/lima-additional-guestagents-path`, the script prints `<PREFIX>/share/lima` that contains the additional guestagents binaries.

The script is intended to be included in the `lima-additional-guestagents` package, so that the main `lima` package can detect the guestagents path via `$PATH`.

Fix #3562

PR #3566, opened 2025-05-21
Locate the limactl executable via os.Args[0] instead of os.Executable()

### LINKED ISSUE #3562: [v1.1 RC] `/opt/homebrew/opt/lima/bin/limactl`, `/home/linuxbrew/.linuxbrew/bin/limactl` can't detect non-native guest agents
https://github.com/AkihiroSuda/homebrew-core/commit/535744dfa73bfe5e61d7e9a6b693c6368f4e2d2c

`/opt/homebrew/bin/limactl` can detect all the guest agents:
```console
$ /opt/homebrew/bin/limactl info | jq .guestAgents
{
  "aarch64": {
    "location": "/opt/homebrew/share/lima/lima-guestagent.Linux-aarch64.gz"
  },
  "armv7l": {
    "location": "/opt/homebrew/share/lima/lima-guestagent.Linux-armv7l.gz"
  },
  "ppc64le": {
    "location": "/opt/homebrew/share/lima/lima-guestagent.Linux-ppc64le.gz"
  },
  "riscv64": {
    "location": "/opt/homebrew/share/lima/lima-guestagent.Linux-riscv64.gz"
  },
  "s390x": {
    "location": "/opt/homebrew/share/lima/lima-guestagent.Linux-s390x.gz"
  },
  "x86_64": {
    "location": "/opt/homebrew/share/lima/lima-guestagent.Linux-x86_64.gz"
  }
}
```

```console
$ ls -l /opt/homebrew/share/lima/
total 0
lrwxr-xr-x   1 suda  admin    72  5 21 05:30 lima-guestagent.Linux-aarch64.gz@ -> ../../Cellar/lima/1.1.0-rc.1/share/lima/lima-guestagent.Linux-aarch64.gz
lrwxr-xr-x   1 suda  admin    94  5 21 05:30 lima-guestagent.Linux-armv7l.gz@ -> ../../Cellar/lima-additional-guestagents/1.1.0-rc.1/share/lima/lima-guestagent.Linux-armv7l.gz
lrwxr-xr-x   1 suda  admin    95  5 21 05:30 lima-guestagent.Linux-ppc64le.gz@ -> ../../Cellar/lima-additional-guestagents/1.1.0-rc.1/share/lima/lima-guestagent.Linux-ppc64le.gz
lrwxr-xr-x   1 suda  admin    95  5 21 05:30 lima-guestagent.Linux-riscv64.gz@ -> ../../Cellar/lima-additional-guestagents/1.1.0-rc.1/share/lima/lima-guestagent.Linux-riscv64.gz
lrwxr-xr-x   1 suda  admin    93  5 21 05:30 lima-guestagent.Linux-s390x.gz@ -> ../../Cellar/lima-additional-guestagents/1.1.0-rc.1/share/lima/lima-guestagent.Linux-s390x.gz
lrwxr-xr-x   1 suda  admin    94  5 21 05:30 lima-guestagent.Linux-x86_64.gz@ -> ../../Cellar/lima-additional-guestagents/1.1.0-rc.1/share/lima/lima-guestagent.Linux-x86_64.gz
drwxr-xr-x  50 suda  admin  1600  5 21 05:30 templates/
```

However, `/opt/homebrew/opt/lima/bin/limactl` can't detect non-native guest agents
```console
$ /opt/homebrew/opt/lima/bin/limactl info | jq .guestAgents
WARN[0000] Failed to resolve the guest agent binary for "x86_64"  error="guest agent binary could not be found for Linux-x86_64 (Hint: try installing `lima-additional-guestagents` package)"
WARN[0000] Failed to resolve the guest agent binary for "armv7l"  error="guest agent binary could not be found for Linux-armv7l (Hint: try installing `lima-additional-guestagents` package)"
WARN[0000] Failed to resolve the guest agent binary for "ppc64le"  error="guest agent binary could not be found for Linux-ppc64le (Hint: try installing `lima-additional-guestagents` package)"
WARN[0000] Failed to resolve the guest agent binary for "riscv64"  error="guest agent binary could not be found for Linux-riscv64 (Hint: try installing `lima-additional-guestagents` package)"
WARN[0000] Failed to resolve the guest agent binary for "s390x"  error="guest agent binary could not be found for Linux-s390x (Hint: try installing `lima-additional-guestagents` package)"
{
  "aarch64": {
    "location": "/opt/homebrew/opt/lima/share/lima/lima-guestagent.Linux-aarch64.gz"
  }
}
```

```console
$ ls -l /opt/homebrew/opt/lima/share/lima/
total 24648
-rw-r--r--   1 suda  admin  12616560  5 21 05:30 lima-guestagent.Linux-aarch64.gz
drwxr-xr-x  50 suda  admin      1600  5 21 05:30 templates/
```

jandubois:
We need the symlink directory to find the the ../share/lima directory relative to it. os.Executable() uses /proc/self/exe on Linux and returns the symlink target, which will break things with a Homebrew installation.

This commit also removes the dead code that pretends to resolve any symlinks. It never worked because it called `os.Stat()` instead of `os.Lstat()`. If it had worked, then code would have been broken on macOS too.

Alternative implementation to #3564 

Fixes #3562

### DIFF
--- a/pkg/usrlocalsharelima/usrlocalsharelima.go
+++ b/pkg/usrlocalsharelima/usrlocalsharelima.go
@@ -8,28 +8,46 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
+	"sync"
 
 	"github.com/lima-vm/lima/pkg/debugutil"
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/sirupsen/logrus"
 )
 
-func Dir() (string, error) {
-	self, err := os.Executable()
-	if err != nil {
-		return "", err
-	}
-	selfSt, err := os.Stat(self)
-	if err != nil {
-		return "", err
-	}
-	if selfSt.Mode()&fs.ModeSymlink != 0 {
-		self, err = os.Readlink(self)
-		if err != nil {
-			return "", err
+// executableViaArgs0 returns the absolute path to the executable used to start this process.
+// It will also append the file extension on Windows, if necessary.
+// This function is different from os.Executable(), which will use /proc/self/exe on Linux
+// and therefore will resolve any symlink used to locate the executable. This function will
+// return the symlink instead because we want to locate ../share/lima relative to the location
+// of the symlink, and not the actual executable. This is important when using Homebrew.
+//
+// If os.Args[0] is invalid, this function still falls back on os.Executable().
+var executableViaArgs0 = sync.OnceValues(func() (string, error) {
+	if os.Args[0] == "" {
+		logrus.Warn("os.Args[0] has not been set")
+	} else {
+		executable, err := exec.LookPath(os.Args[0])
+		if err == nil {
+			// LookPath() will add the `.exe` file extension on Windows, but will not return an
+			// absolute path if the argument contained any of `:/\` (or just `/` on Unix).
+			return filepath.Abs(executable)
 		}
+		logrus.Warnf("os.Args[0] is invalid: %v", err)
+	}
+	return os.Executable()
+})
+
+// Dir returns the location of the <PREFIX>/lima/share directory, relative to the location
+// of the current executable. It checks for multiple possible filesystem layouts and returns
+// the first candidate that contains the native guest agent binary.
+func Dir() (string, error) {
+	self, err := executableViaArgs0()
+	if err != nil {
+		return "", err
 	}
 
 	ostype := limayaml.NewOS("linux")
@@ -80,7 +98,7 @@ func Dir() (string, error) {
 		ostype, arch, self, gaCandidates)
 }
 
-// GuestAgentBinary returns the guest agent binary, possibly with ".gz" suffix.
+// GuestAgentBinary returns the absolute path of the guest agent binary, possibly with ".gz" suffix.
 func GuestAgentBinary(ostype limayaml.OS, arch limayaml.Arch) (string, error) {
 	if ostype == "" {
 		return "", errors.New("os must be set")


pkg/usrlocalsharelima/usrlocalsharelima.go
```diff
@@ -8,16 +8,33 @@ import (
...
+		return "", err
+	}
+	// LookPath() will add the `.exe` file extension on Windows, but will not return an
+	// absolute path if the argument contained any of `:/\` (or just `/` on Unix).
+	return filepath.Abs(executable)
+})
+
 func Dir() (string, error) {
-	self, err := os.Executable()
+	self, err := executableViaArgs0()
```
AkihiroSuda, line 37: os.ReadLink below should be removed

(failed to sleep after all 😅 )

pkg/usrlocalsharelima/usrlocalsharelima.go
```diff
@@ -8,16 +8,33 @@ import (
...
 	"github.com/lima-vm/lima/pkg/debugutil"
 	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/sirupsen/logrus"
 )
 
+// executableViaArgs0 returns the absolute path to the executable used to start this process.
+// It will also append the file extension on Windows, if necessary.
+// This function is different from os.Executable(), which will use /proc/self/exe on Linux
+// and therefore will resolve any symlink used to locate the executable. This function will
+// return the symlink instead.
```
AkihiroSuda, line 25: Needs a comment to explain why symlinks shouldn't be solved here

AkihiroSuda:
#3564 is still better in the robustness; the binary can be executed via an arbitrary path

jandubois:
> #3564 is still better in the robustness; the binary can be executed via an arbitrary path

The only way I can think this can happen is if you invoke `limactl` via `execve` and not setup argv[0] (or set it to an incorrect value). Are there others?

That is unfortunate, but I think this is not a reasonable use case; why would anyone do that?. If it doesn't work the way they expect, it is really their fault and not ours. I don't even think it would be worth documenting.

pkg/usrlocalsharelima/usrlocalsharelima.go
jandubois: Ah, this is dead code because we call `os.Stat`, which will follow the symlink, so the returned mode bits never have the symlink bit set. The code would need to call `os.Lstat()` to actually work... Anyways, will remove it.

jandubois:
> The only way I can think this can happen is if you invoke `limactl` via `execve` and not setup argv[0]

Not sure if this is really necessary, but I've added code to fall back on `os.Executable()` when `os.Args[0]` is empty, or can't be resolved with `exec.LookPath()`.

If the user intentionally passes the path to a different executable, then I don't consider this an error; the user probably wants to resolve the directory relative to that location.

So the only remaining failure path has 3 preconditions:

* Must pass empty string (or invalid name) for `os.Args[0]`
* Must be running on Linux
* Must be using Homebrew style symlink setup

My facetious response would be: "Welcome to cafe Karma! We have no menu; you get what you deserve."

There are many other programs (including all multi-call binaries like `busybox`) that rely on `os.Args[0]` to be set correctly, so why can't we?

jandubois:
[Further testing](https://github.com/lima-vm/lima/pull/3564#issuecomment-2899486768) showed that #3564 is not robust either; it only works when `limactl` is first on the `PATH`.

I think the Args[0] approach is simpler, more reliable, and enables more functionality (allows locating `libexec` directory), so should be the one we pick.

pkg/usrlocalsharelima/usrlocalsharelima.go
```diff
@@ -8,29 +8,47 @@ import (
...
+	if os.Args[0] == "" {
+		logrus.Warn("os.Args[0] has not been set")
+	} else {
+		executable, err := exec.LookPath(os.Args[0])
+		if err == nil {
+			// LookPath() will add the `.exe` file extension on Windows, but will not return an
+			// absolute path if the argument contained any of `:/\` (or just `/` on Unix).
+			return filepath.Abs(executable)
+		}
+		logrus.Warnf("os.Args[0] is invalid: %v", err)
```
AkihiroSuda, line 39: nit: `logrus.WithError.Warn` could be used

AkihiroSuda (approved):
Thanks, let's release v1.1.1 with this and see if it works

PR #3571, opened 2025-05-23
info: silence "guest agent binary could not be found"

AkihiroSuda:
Fix #3570

`pkg/limainfo` expects `fs.ErrNoExist` when the guest agent binary could not be found

https://github.com/lima-vm/lima/blob/44f3da4a6a734730a663612661b851be1ece60cd/pkg/limainfo/limainfo.go#L59-L67


jandubois (approved):
Thanks, LGTM

Issue #3518, opened 2025-05-08
Define the common `interface` for `qemuimgutil` and `nativeimgutil`


AkihiroSuda:
https://github.com/lima-vm/lima/blob/master/pkg/qemu/imgutil/imgutil.go
https://github.com/lima-vm/lima/blob/master/pkg/nativeimgutil/nativeimgutil.go

These package should implement the common interface.

- `pkg/imgutil`: define the interface`
- `pkg/imgutil/nativeimgutil`: native implementation
- `pkg/imgutil/qemuimgutil`: qemu-img implementation

Probably there should be also `pkg/imgutil/proxyimgutil` (name is tentative) that attempts qemuimgutil first and then fall back to nativeimgutil?

PR #3573, opened 2025-05-23
[WIP]: Create a Plugin System for Lima

unsuman:
> [!IMPORTANT]
>  This PR is only for preview purposes and not meant to be merged. Later, the commits can be cherry-picked and raised as separate small PRs.

This PR creates a Plugin System for drivers which can be embedded in the main binary or built as separate binaries, which then communicate with the main binary using gRPC. This work is done as part of my GSoC '25 task.


cmd/limactl/start.go
```diff
@@ -54,6 +56,9 @@ $ limactl create --set='.cpus = 2 | .memory = "2GiB"'
 To see the template list:
 $ limactl create --list-templates
 
+To see the drivers list:
+$ limactl create --list-drivers
```
AkihiroSuda, line 60: The example currently just focuses on templates, not on drivers.

cmd/limactl/disk.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/docker/go-units"
 	"github.com/lima-vm/go-qcow2reader"
 	"github.com/lima-vm/lima/pkg/nativeimgutil"
-	"github.com/lima-vm/lima/pkg/qemu/imgutil"
+	"github.com/lima-vm/lima/pkg/qemuimgutil"
```
AkihiroSuda, line 19: Eventually this should follow:
- https://github.com/lima-vm/lima/issues/3518

Not an urgent topic though

pkg/builtins/drivers.go
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package builtins
```
AkihiroSuda, line 4: builitins should be in `cmd/limactl/main*.go`

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -246,3 +265,44 @@ func (l *LimaVzDriver) GuestAgentConn(_ context.Context) (net.Conn, error) {
...
+func (l *LimaVzDriver) ApplySnapshot(_ context.Context, _ string) error {
+	return errors.New("unimplemented")
+}
+
+func (l *LimaVzDriver) DeleteSnapshot(_ context.Context, _ string) error {
+	return errors.New("unimplemented")
+}
+
+func (l *LimaVzDriver) ListSnapshots(_ context.Context) (string, error) {
+	return "", errors.New("unimplemented")
```
AkihiroSuda, line 302: The common `ErrImplemented` should be defined so that it can be compared with `errors.Is`

pkg/driver/vz/vz_driver_others.go
```diff
@@ -0,0 +1,118 @@
+//go:build !darwin || no_vz
```
AkihiroSuda, line 1: The package shouldn't be compiled at all on non-Darwin

pkg/driver/external/proto/driver.proto
```diff
@@ -0,0 +1,144 @@
...
+    rpc Name(google.protobuf.Empty) returns (NameResponse);
+    rpc SetConfig(SetConfigRequest) returns (SetConfigResponse);
+    
+    rpc GetVSockPort(google.protobuf.Empty) returns (GetVSockPortResponse);
+    rpc GetVirtioPort(google.protobuf.Empty) returns (GetVirtioPortResponse);
+}
+
+message Empty {}
+
+message InstanceConfig {
```
AkihiroSuda, line 27: Can we just pass the byte array of YAML (or JSON), so that we do not need to sync the YAML struct with this proto definition?

pkg/driver/external/driver.proto
```diff
@@ -0,0 +1,98 @@
...
+
+    rpc ForwardGuestAgent(google.protobuf.Empty) returns (ForwardGuestAgentResponse);
+    rpc GuestAgentConn(google.protobuf.Empty) returns (stream GuestAgentConnResponse);
+
+    rpc Name(google.protobuf.Empty) returns (NameResponse);
+    rpc SetConfig(SetConfigRequest) returns (SetConfigResponse);
+    
+    rpc GetVSockPort(google.protobuf.Empty) returns (GetVSockPortResponse);
+    rpc GetVirtioPort(google.protobuf.Empty) returns (GetVirtioPortResponse);
+}
```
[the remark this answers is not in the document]
unsuman: @AkihiroSuda Should I also consolidate these functions in the `driver.Driver` interface, in order to maintain consistency?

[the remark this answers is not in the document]
AkihiroSuda: Yes

pkg/driver/external/client/methods.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	d.logger.Debug("Driver validated successfully")
+	return nil
+}
+
+func (d *DriverClient) Initialize(ctx context.Context) error {
+	d.logger.Debug("Initializing driver instance")
+
+	connCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
AkihiroSuda: timeout should be specified in the caller

pkg/driver/external/server/methods.go
```diff
@@ -62,36 +63,34 @@ func (s *DriverServer) SetConfig(ctx context.Context, req *pb.SetConfigRequest)
...
+	}
+
+	defer conn.Close()
+
+	buf := make([]byte, 32*1<<10)
+
+	for {
+		n, err := conn.Read(buf)
+		if err != nil {
+			if err == io.EOF {
```
AkihiroSuda: `errors.Is` is more robust

pkg/driver/qemu/cmd/main.go
AkihiroSuda, line 1: This should be `cmd/<EXECUTABLE NAME>/main.go`

AkihiroSuda: Makefile has to be updated to compile this by default.

AkihiroSuda:
How to test this ?

Tried `go build -o ~/.local/libexec/lima/drivers/lima-driver-qemu ./pkg/driver/qemu/cmd` , but `~/.local/bin/limactl` did not recognize the driver.

unsuman:
> How to test this ?
> 
> Tried `go build -o ~/.local/libexec/lima/drivers/lima-driver-qemu ./pkg/driver/qemu/cmd` , but `~/.local/bin/limactl` did not recognize the driver.

The driver discovery through the directory is not working rn. I will work on it once I fix the GuestAgentConn() function.

For now, you can test the driver by setting the executable driver's path to `LIMA_DRIVERS_PATH`

cmd/limactl/main_darwin.go
```diff
@@ -0,0 +1,9 @@
+//go:build darwin && !no_vz
```
jandubois: I like to avoid double-negation, when possible:

```suggestion
//go:build darwin && driver_vz
```

Then the `Makefile` will just have a `DRIVER_TAGS=driver_qemu,driver_vz,driver_wsl` setting that the user can override to only include the desired driver(s).

pkg/driver/driver.go
```diff
@@ -35,120 +31,69 @@ type Driver interface {
...
+// GUI defines GUI-related operations.
+type GUI interface {
+	// RunGUI is for starting GUI synchronously by hostagent. This method should be wait and return only after vm terminates
+	// It returns error if there are any failures
+	RunGUI() error
 
-func (d *BaseDriver) Stop(_ context.Context) error {
-	return nil
+	ChangeDisplayPassword(ctx context.Context, password string) error
+	GetDisplayConnection(ctx context.Context) (string, error)
```
jandubois: In Go "getters" don't use a `Get` prefix. You know it is a getter because the method name doesn't include an action word (verb).

```suggestion
	DisplayConnection(ctx context.Context) (string, error)
```

pkg/driver/driver.go
```diff
@@ -35,120 +31,69 @@ type Driver interface {
...
+	Lifecycle
+	GUI
+	SnapshotManager
+	Registration
+	GuestAgent
 
-func (d *BaseDriver) ApplySnapshot(_ context.Context, _ string) error {
-	return errors.New("unimplemented")
-}
+	GetInfo() Info
```
jandubois: I'm not going to comment on additional getters; they all should drop the `Get` prefix.

```suggestion
	Info() Info
```

pkg/driver/driver.go
```diff
@@ -35,120 +31,69 @@ type Driver interface {
...
 
-func (d *BaseDriver) DeleteSnapshot(_ context.Context, _ string) error {
-	return errors.New("unimplemented")
+	// SetConfig sets the configuration for the instance.
+	SetConfig(inst *store.Instance, sshLocalPort int)
```
jandubois: I wonder if it makes sense to have a separate type (like the Kubernetes `Completed*` types):

```go
type ConfiguredDriver struct {
    Driver
}
```

Then you could have a function to turn a `Driver` into a `ConfiguredDriver`:

```suggestion
	Configure(inst *store.Instance, sshLocalPort int) ConfiguredDriver
```

And all the functions/methods that would only work with configured driver would only accept that type instead.

pkg/driver/external/server/server.go
```diff
@@ -0,0 +1,88 @@
...
+	pb.UnimplementedDriverServer
+	driver driver.Driver
+	logger *logrus.Logger
+}
+
+func Serve(driver driver.Driver) {
+	logger := logrus.New()
+	logger.SetLevel(logrus.DebugLevel)
+
+	socketPath := filepath.Join(os.TempDir(), fmt.Sprintf("lima-driver-%s-%d.sock", driver.GetInfo().DriverName, os.Getpid()))
```
jandubois: Should the socket be created somewhere under `LIMA_HOME`? Maybe `$LIMA_HOME/_run`?

pkg/driver/qemu/cmd/lima-driver-qemu/main.go
jandubois, line 1: I think it would be better to move this to `cmd/lima-driver-qemu` at the top level. There should be no `main` packages with a `main` function inside the `pkg` tree.

pkg/driver/qemu/qemu_driver.go
```diff
@@ -27,36 +27,51 @@ import (
...
 )
 
 type LimaQemuDriver struct {
-	*driver.BaseDriver
+	Instance     *store.Instance
+	SSHLocalPort int
+	VSockPort    int
+	VirtioPort   string
+
+	DriverType driver.DriverType
```
jandubois: It feels like a failure of the abstraction if the driver needs to know if it is an internal or external driver. I was hoping that this could be entirely handled by the remoting layer.

pkg/driver/vz/cmd/lima-driver-vz/main.go
```diff
@@ -0,0 +1,18 @@
+//go:build (darwin && amd64) || (darwin && arm64)
```
jandubois: Why is this not just checking for `darwin`?

```suggestion
//go:build darwin
```

pkg/driver/vz/disk.go
jandubois, line 1: There is some inconsistency in the `go:build` line between the files in this directory. E.g. `errors_darwin.go` has

```go
/go:build darwin && !no_vz
```

It looks like including/excluding the driver is controlled by importing `pkg/driver/vz/register` into `main`, so that file is really the only one that needs the `go:build` line. All the rest are not referenced and don't need it.

If we want to keep it for some kind of "belts & suspender" reasons, then all the files should have it.

I also think that we should rename all the file to remove the `_darwin` suffix from the filenames. It is no longer needed because the files will not be referenced on any other OS. Only the `_arm64` suffix on the Rosetta code needs to stay.

pkg/driverutil/instance.go
```diff
@@ -4,20 +4,22 @@
...
 
-func CreateTargetDriverInstance(base *driver.BaseDriver) driver.Driver {
-	limaDriver := base.Instance.Config.VMType
-	if *limaDriver == limayaml.VZ {
-		return vz.New(base)
-	}
-	if *limaDriver == limayaml.WSL2 {
-		return wsl2.New(base)
+// CreateTargetDriverInstance creates the appropriate driver for an instance.
+func CreateTargetDriverInstance(inst *store.Instance, sshLocalPort int) (driver.Driver, error) {
```
jandubois: We are overloading the word `instance` here, for drivers and machines.

I think my previous suggestion to use a different type for `DriverInstance` would work well here:


```suggestion
func CreateConfiguredDriver(inst *store.Instance, sshLocalPort int) (driver.ConfiguredDriver, error) {
```

pkg/driverutil/instance.go
```diff
@@ -4,20 +4,22 @@
...
+	limaDriver := inst.Config.VMType
+	driver, exists := registry.DefaultRegistry.Get(string(*limaDriver), inst.Name)
+	if !exists {
+		return nil, fmt.Errorf("unknown or unsupported VM type: %s", *limaDriver)
 	}
-	return qemu.New(base)
+
+	driver.SetConfig(inst, sshLocalPort)
+
+	return driver, nil
```
jandubois:
```suggestion
	return driver.Configure(inst, sshLocalPort)
```

pkg/driver/qemu/qemu_driver.go
AkihiroSuda: GuestAgentConn needs to know this, at least for now

pkg/driver/external/server/server.go
AkihiroSuda: Why?

cmd/limactl/main_darwin.go
AkihiroSuda: Also `go:build darwin` is already implied by the filename

pkg/driver/external/server/server.go
jandubois: Because it may be a shared directory, like `/tmp`, so anyone would be able to delete it.  Maybe it doesn't matter; idk.

pkg/driver/qemu/qemu_driver.go
jandubois: I see this now. I hope we can fix this.

pkg/limainfo/limainfo.go
```diff
@@ -45,7 +45,7 @@ func New() (*LimaInfo, error) {
 	info := &LimaInfo{
 		Version:         version.Version,
 		DefaultTemplate: y,
-		VMTypes:         driverutil.Drivers(),
+		VMTypes:         registry.DefaultRegistry.List(),
```
jandubois: I guess we need to keep this for backwards compatibility, but I would like to see a map of drivers to their driver locations (empty for builtin drivers). Similar to how we have a `templates` entry.

It helps to debug if `limactl` is picking up the right driver, or if it is maybe shadowed by a different version earlier in the search path.

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+	Path         string
+	ctx          context.Context
+	logger       *logrus.Logger
+	cancelFunc   context.CancelFunc
+}
+
+type Registry struct {
+	internalDrivers map[string]driver.Driver
+	externalDrivers map[string]*ExternalDriver
+	mu              sync.RWMutex
```
jandubois: I don't see why we need a mutex here. The registry is only written to during initialization, and all updates are strictly serialized (all `init` functions run in a defined sequence, and then there is `DiscoverDrivers` which is only called once after all `init` functions are done.

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+func NewRegistry() *Registry {
+	return &Registry{
+		internalDrivers: make(map[string]driver.Driver),
+		externalDrivers: make(map[string]*ExternalDriver),
+	}
+}
```
jandubois: I don't see the point why we pretend there can be more than one registry inside a process. This can be just a single global struct that is initialized in an `init` function, and then used implicitly by all the other functions.

So `registry.DefaultRegistry.List()` would just be `registry.List()` etc.

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+
+func (e *ExternalDriver) Start(instName string) error {
```
jandubois: I think all the driver `Start`, `Stop`, etc. functions don't belong in the `registry` package, but somewhere in the `driver` package (or adjacent to it).

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+				r.externalDrivers[name].InstanceName = instName
+			}
+
+			return externalDriver.Client, true
+		}
+	}
+	return driver, exists
+}
+
+func (r *Registry) RegisterDriver(name, path string) {
```
jandubois: This should be `RegisterExternalDriver` to distinguish it from `Register` (which should be called `RegisterInternalDriver`):

```suggestion
func RegisterExternalDriver(name, path string) {
```

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+	if !exists {
+		externalDriver, exists := r.externalDrivers[name]
+		if exists {
+			externalDriver.logger.Debugf("Using external driver %q", name)
+			if externalDriver.Client == nil || externalDriver.Command == nil || externalDriver.Command.Process == nil {
+				logrus.Infof("Starting new instance of external driver %q", name)
+				if err := externalDriver.Start(instName); err != nil {
+					externalDriver.logger.Errorf("Failed to start external driver %q: %v", name, err)
+					return nil, false
+				}
```
jandubois: I find mixing the concept of fetching a `driver.Driver` object with starting a driver process confusing. I feel like these are separate things, but maybe I need to think about this some more.

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+		logger: logrus.New(),
+	}
+}
+
+func (r *Registry) DiscoverDrivers() error {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return err
+	}
+	stdDriverDir := filepath.Join(homeDir, ".local", "libexec", "lima", "drivers")
```
jandubois: I thought this was going to be `/usr/local/libexec/lima` and not a user-owned directory. We don't store anything except `LIMA_HOME` under the user's home directory.

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+	if driverPaths := os.Getenv("LIMA_DRIVERS_PATH"); driverPaths != "" {
+		paths := filepath.SplitList(driverPaths)
+		for _, path := range paths {
+			if path == "" {
+				continue
+			}
+
+			info, err := os.Stat(path)
+			if err != nil {
+				logrus.Warnf("Error accessing driver path %s: %v", path, err)
```
jandubois: For consistency we should always quote filenames. Especially if they are user-provided and may contain whitespace. But let's just always do it:

```suggestion
				logrus.Warnf("Error accessing driver path %q: %v", path, err)
```

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+		r.registerDriverFile(driverPath)
+	}
+
+	return nil
+}
+
+func (r *Registry) registerDriverFile(path string) {
+	base := filepath.Base(path)
+	if !strings.HasPrefix(base, "lima-driver-") {
+		fmt.Printf("Skipping %s: does not start with 'lima-driver-'\n", base)
```
jandubois: I assume this is debugging code; we don't want to print anything for ignored files.

pkg/registry/registry.go
```diff
@@ -0,0 +1,320 @@
...
+		return
+	}
+
+	name := strings.TrimPrefix(base, "lima-driver-")
+	name = strings.TrimSuffix(name, filepath.Ext(name))
```
jandubois: I guess this is fine; we don't need to allow `.` in driver names. Otherwise I would have said we only strip `.exe` suffix on Windows, but otherwise allow any characters.

pkg/store/instance.go
```diff
@@ -438,3 +439,38 @@ func (inst *Instance) Unprotect() error {
...
 }
+
+func (inst *Instance) MarshalJSON() ([]byte, error) {
+	type Alias Instance
+	errorsAsStrings := make([]string, len(inst.Errors))
+	for i, err := range inst.Errors {
+		if err != nil {
+			errorsAsStrings[i] = err.Error()
+		}
+	}
```
jandubois, line 450: I don't know if this is going to work properly. E.g. if the error is later processed again with `errors.Is()`, then the result will always be `false` because we are not preserving the type.

Not sure if that is important, but it would require some reflection magic to make it work.

pkg/driver/external/server/server.go
jandubois: I forgot that `/tmp` has the sticky bit set, so I guess this is fine.

pkg/driver/vz/disk.go
AkihiroSuda: > It is no longer needed because the files will not be referenced on any other OS.

Still needed to allow running `golangci-lint run ./...`,  `govulncheck ./...`, etc. on non-macOS ?

jandubois: > Still needed to allow running `golangci-lint run ./...`, `govulncheck ./...`, etc. on non-macOS ?

Yeah, probably true. But then it can be removed from the `go:build` line.

pkg/driver/qemu/qemu_driver.go
jandubois: One way to deal with this would be to have both a `GuestAgentConn()` and `ExternalGuestAgentConn()` method on the driver, and the `driverserver.GuestAgentConn()` method in `pkg/driver/external/server/methods.go` would call `s.driver.ExternalGuestAgentConn()` instead of the internal version.

Maybe call it `GuestAgentConnExternal()` so both have the same prefix, idk.

unsuman: Or can we change the method signature to `GuestAgentConn() (net.Conn, type string)` where on the `pkg/driver/external/server/methods.go` side we will check if the incoming connection is not a UDS, we will go with proxying out that connection?

jandubois: > change the method signature to `GuestAgentConn() (net.Conn, type string)`

That would be even better, if you can make it work. Because then the proxying is not part of the driver, but of the server code, where it really belongs.

I just think that the driver should be completely ignorant of whether it is running internal or external.

pkg/registry/registry.go
jandubois: To elaborate a bit, I expected this to use `filepath.Dir(filepath.Dir(usrlocalsharelima.Dir())` as the prefix. Maybe there should be a `usrlocalsharelima.Prefix()` function for that.

Further, I don't know if we need a `$PREFIX/libexec/lima/drivers` directory, or if we can't just use `$PREFIX/libexec` since we filter on `lima-driver-*` filenames anyways.

At most we should be using ``$PREFIX/libexec/lima`; the `drivers` subdirectory is not needed.

pkg/store/instance.go
unsuman: Let's keep this for a later stage, then!

Makefile
```diff
@@ -167,7 +167,7 @@ binaries: limactl helpers guestagents \
 ################################################################################
 # _output/bin
 .PHONY: limactl lima helpers
-limactl: _output/bin/limactl$(exe) lima
+limactl: _output/bin/limactl$(exe) external-drivers lima
```
AkihiroSuda: limactl shouldn't depend on external drivers

Makefile
```diff
@@ -241,18 +241,46 @@ endif
...
-	$(ENVS_$@) $(GO_BUILD) -o $@ ./cmd/limactl
+	$(ENVS_$@) $(GO_BUILD) -tags '$(GO_BUILDTAGS_LIMACTL)' -o $@ ./cmd/limactl
 ifeq ($(GOOS),darwin)
 	codesign -f -v --entitlements vz.entitlements -s - $@
 endif
 
+DRIVER_INSTALL_DIR := _output/libexec/lima
+
+.PHONY: external-drivers
+external-drivers:
```
AkihiroSuda: "additional-drivers" might be more consistent with "additional-guestagents"

Makefile
```diff
@@ -241,18 +241,46 @@ endif
...
+LIMACTL_DRIVER_TAGS += external_vz
+endif
+ifneq (,$(findstring qemu,$(EXTERNAL_DRIVERS)))
+LIMACTL_DRIVER_TAGS += external_qemu
+endif
+ifneq (,$(findstring wsl2,$(EXTERNAL_DRIVERS)))
+LIMACTL_DRIVER_TAGS += external_wsl2
+endif
+
+GO_BUILDTAGS ?=
```
AkihiroSuda, line 255: nit:
```suggestion
GO_BUILD_TAGS ?=
```

AkihiroSuda:
Bug: The progress bar seems no longer shown when downloading an uncached image.
It is still shown for downloading the nerdctl archive, though.
```
$ limactl start
INFO[0000] Starting new instance of external driver "vz" 
INFO[0000] Starting external driver at /Users/suda/.local/libexec/lima/lima-driver-vz 
INFO[0000] External driver vz started successfully       driver=vz
INFO[0000] Starting the instance "default" with VM driver "vz" 
INFO[0000] Reusing existing external driver "vz" instance 
INFO[0048] Attempting to download the nerdctl archive    arch=aarch64 digest="sha256:1b52f32b7d5bbf63005bceb6a3cacd237d2fa8f1d05bb590e8ce58731779b9ee" location="https://github.com/containerd/nerdctl/releases/download/v2.1.2/nerdctl-full-2.1.2-linux-arm64.tar.gz"
Downloading the nerdctl archive (nerdctl-full-2.1.2-linux-arm64.tar.gz)
224.67 MiB / 224.67 MiB [----------------------------------] 100.00% 27.42 MiB/s
INFO[0057] Downloaded the nerdctl archive from "https://github.com/containerd/nerdctl/releases/download/v2.1.2/nerdctl-full-2.1.2-linux-arm64.tar.gz" 
INFO[0057] [hostagent] Starting new instance of external driver "vz" 
INFO[0057] [hostagent] time="2025-06-26T11:41:36+09:00" level=info msg="Starting external driver at /Users/suda/.local/libexec/lima/lima-driver-vz" 
INFO[0057] [hostagent] time="2025-06-26T11:41:36+09:00" level=info msg="External driver vz started successfully" driver=vz 
INFO[0058] [hostagent] hostagent socket created at /Users/suda/.lima/default/ha.sock 
INFO[0058] SSH Local Port: 60022                        
INFO[0058] [hostagent] Waiting for the essential requirement 1 of 2: "ssh" 
INFO[0068] [hostagent] Waiting for the essential requirement 1 of 2: "ssh" 
INFO[0068] [hostagent] The essential requirement 1 of 2 is satisfied 
INFO[0068] [hostagent] Waiting for the essential requirement 2 of 2: "user session is ready for ssh" 
INFO[0068] [hostagent] The essential requirement 2 of 2 is satisfied 
INFO[0068] [hostagent] Waiting for the optional requirement 1 of 2: "systemd must be available" 
INFO[0068] [hostagent] time="2025-06-26T11:41:47+09:00" level=info msg="Getting guest agent connection" 
INFO[0068] [hostagent] Guest agent is running           
INFO[0068] [hostagent] The optional requirement 1 of 2 is satisfied 
INFO[0068] [hostagent] Not forwarding TCP 0.0.0.0:22    
INFO[0068] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed" 
INFO[0068] [hostagent] Not forwarding TCP 127.0.0.54:53 
INFO[0068] [hostagent] Not forwarding TCP 127.0.0.53:53 
INFO[0068] [hostagent] Not forwarding TCP [::]:22       
INFO[0068] [hostagent] Not forwarding UDP 127.0.0.54:53 
INFO[0068] [hostagent] Not forwarding UDP 127.0.0.53:53 
INFO[0068] [hostagent] Not forwarding UDP 192.168.5.15:68 
INFO[0083] [hostagent] Forwarding TCP from 127.0.0.1:32833 to 127.0.0.1:32833 
INFO[0083] [hostagent] The optional requirement 2 of 2 is satisfied 
INFO[0083] [hostagent] Waiting for the guest agent to be running 
INFO[0083] [hostagent] Waiting for the final requirement 1 of 1: "boot scripts must have finished" 
INFO[0092] [hostagent] The final requirement 1 of 1 is satisfied 
INFO[0093] READY. Run `lima` to open the shell.         
INFO[0093] Stopping external driver vz                  
INFO[0093] External driver vz stopped successfully      
INFO[0093] External driver vz stopped successfully
```


Also, "Stopping external driver vz" might be confusing to users, as it sounds like shutting down the VM

Makefile
unsuman: If we remove `external-drivers`(now `additional-drivers`) from limactl as a dependency, then how will it get triggered?
How will this work?
`make ADDITIONAL_DRIVERS="qemu" limactl -B`

AkihiroSuda: Same as additional-guestagents

unsuman:
@AkihiroSuda Since all 4 PRs are now merged should we close this one?

AkihiroSuda:
Yes, thanks

PR #3575, opened 2025-05-23
Allow dotted instance names in `limactl start`

### LINKED ISSUE #3574: VM names with dots are not recognised as names when using with `limactl start` on macOS 15.5
### Description

With lima v1.1.1, on macOS Sonoma 15.5, I cannot start a VM using `limactl start example.com` because it thinks 'example.com' is a path. I must use `limactl start --name=example.com` to have it work correctly. This is a change from the previous version's behaviour.

I am not certain if this bug is specific to lima or macOS, but  Lima is likely using heuristics to determine if an argument is a file path or instance name, and "example.com" triggers the file path logic incorrectly.

jandubois:
`limatmpl.Read()` no longer returns an empty template when called with an instance name because it now needs to be able to read any kind of file via the `template://` mechanism.

This commit changes the callers to no longer call `limatmpl.Read()` with an instance name. This also allows simplification of the locator type selection: every locator that is not a URL can now be treated as a local file.

Fixes #3574

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -7,12 +7,10 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"path"
 	"path/filepath"
 	"runtime"
 	"strings"
 
-	"github.com/containerd/containerd/identifiers"
 	"github.com/lima-vm/lima/cmd/limactl/editflags"
 	"github.com/lima-vm/lima/pkg/editutil"
 	"github.com/lima-vm/lima/pkg/instance"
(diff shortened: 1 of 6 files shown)


cmd/limactl/start.go
```diff
@@ -148,9 +148,16 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string, createOnly bool) (*
 		}
 		tty = false
 	}
-	tmpl, err := limatmpl.Read(cmd.Context(), name, arg)
-	if err != nil {
-		return nil, err
+	var tmpl *limatmpl.Template
+	// Only filenames ending in .yml or .yaml are template filenames; otherwise they are instance names.
+	// This check is necessary because limatmpl.Read() can also be used to read non-template files.
+	if limatmpl.SeemsFilePath(arg) && !(strings.HasSuffix(arg, ".yml") || strings.HasSuffix(arg, ".yaml")) {
```
AkihiroSuda, line 154: Should be case-insensitive?

AkihiroSuda: With unit tests, maybe

jandubois:
@AkihiroSuda PTAL

AkihiroSuda (approved):
Thanks

PR #3578, opened 2025-05-25
Move `identifierutil` → `instance/hostname` package

jandubois:
Part of the crusade against gratuituous use of `util` package suffix.

The package only contains a single `HostnameFromInstName()` function.

It can't be moved into the `instance` package because that would create a circular dependency between `instance` and `cidata`.


pkg/hostagent/hostagent.go
```diff
@@ -288,8 +288,8 @@ func (a *HostAgent) Run(ctx context.Context) error {
 	if limayaml.FirstUsernetIndex(a.instConfig) == -1 && *a.instConfig.HostResolver.Enabled {
 		hosts := a.instConfig.HostResolver.Hosts
 		hosts["host.lima.internal"] = networks.SlirpGateway
-		hostname := identifierutil.HostnameFromInstName(a.instName) // TODO: support customization
-		hosts[hostname] = networks.SlirpIPAddress
+		host := hostname.FromInstName(a.instName) // TODO: support customization
```
nirs, line 291: This is nice:

    hostname.FromInstName(a.instName)

but it seems too extreme to have a package for single function inside an empty directory just to just nicer "namespace".

Also, we the return value is kept in `host`, which seems to be a bigger concept. The best name for the value is `hostname` but it clashes with the package name. If the package was `instance` we could avoid this:

    hostname := instance.HostnameFor(a.instaName)

Can we have single instance package providing instance.HostnameFor(instanceName)?
And maybe move other related functions there?

pkg/instance/hostname/hostname_test.go
```diff
@@ -0,0 +1,16 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package hostname
```
nirs, line 4: Why not hostname_test? we don't need to access private names in the package.

pkg/limayaml/defaults.go
```diff
@@ -24,21 +24,20 @@ import (
...
 	"github.com/lima-vm/lima/pkg/networks"
 	"github.com/lima-vm/lima/pkg/osutil"
 	"github.com/lima-vm/lima/pkg/ptr"
 	"github.com/lima-vm/lima/pkg/store/dirnames"
 	"github.com/lima-vm/lima/pkg/store/filenames"
+	"github.com/lima-vm/lima/pkg/version"
 	"github.com/lima-vm/lima/pkg/version/versionutil"
+	"github.com/pbnjay/memory"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sys/cpu"
```
nirs, line 40: The import changes look suspicious.

pkg/hostagent/hostagent.go
jandubois: > [...] it seems too extreme to have a package for single function inside an empty directory just to just nicer "namespace".

I've only renamed the package (and the function). It was a package with a single function from the start.

> Also, we the return value is kept in `host`, which seems to be a bigger concept. The best name for the value is `hostname` but it clashes with the package name.

There is really no issue with the variable overriding the package name, beyond that I don't like it. I could keep the `hostname` name. Or change it to `h` (like you do with `u, err := url.Parse()`). Or use `hosts[hostname.FromInstName()] = ...`.

I will switch to `name` because it makes the next line look better, but it really doesn't matter:

```
name := hostname.FromInstName(a.instName)
hosts[name] = networks.SlirpIPAddress
```

> If the package was `instance` we could avoid this:
> 
> ```
> hostname := instance.HostnameFor(a.instaName)
> ```

I already explained that `instance` was my first choice, but it doesn't work because would create an import loop with `cidata`.

> Can we have single instance package providing instance.HostnameFor(instanceName)? And maybe move other related functions there?

No, we already have an `instance` package in our tree. Having multiple instance packages in one app is confusing.

pkg/instance/hostname/hostname_test.go
jandubois: There are no strong reasons either way, but there are no private names (right now), so it makes no difference, but makes the tests look more cluttered:

```go
	assert.Equal(t, hostname.FromInstName("default"), "lima-default")
	assert.Equal(t, hostname.FromInstName("ubuntu-24.04"), "lima-ubuntu-24-04")
	assert.Equal(t, hostname.FromInstName("foo_bar.baz"), "lima-foo-bar-baz")
```

vs.

```go
	assert.Equal(t, FromInstName("default"), "lima-default")
	assert.Equal(t, FromInstName("ubuntu-24.04"), "lima-ubuntu-24-04")
	assert.Equal(t, FromInstName("foo_bar.baz"), "lima-foo-bar-baz")
```

and I prefer tests to be as simple as possible. But again it doesn't really matter, and I will change it.

pkg/limayaml/defaults.go
jandubois: Why? I've only ordered the imports alphabetically:

```
	"github.com/coreos/go-semver/semver"
	"github.com/docker/go-units"
	"github.com/goccy/go-yaml"
	"github.com/lima-vm/lima/pkg/instance/hostname"
	"github.com/lima-vm/lima/pkg/ioutilx"
	. "github.com/lima-vm/lima/pkg/must"
	"github.com/lima-vm/lima/pkg/networks"
	"github.com/lima-vm/lima/pkg/ptr"
	"github.com/lima-vm/lima/pkg/store/dirnames"
	"github.com/lima-vm/lima/pkg/store/filenames"
	"github.com/lima-vm/lima/pkg/version"
	"github.com/lima-vm/lima/pkg/version/versionutil"
	"github.com/pbnjay/memory"
	"github.com/sirupsen/logrus"
	"golang.org/x/sys/cpu"
```

Before this commit they looked like this, and I couldn't figure out how they were organized:

```
	"github.com/coreos/go-semver/semver"
	"github.com/docker/go-units"
	"github.com/goccy/go-yaml"
	"github.com/lima-vm/lima/pkg/ioutilx"
	"github.com/lima-vm/lima/pkg/version"
	"github.com/pbnjay/memory"
	"github.com/sirupsen/logrus"
	"golang.org/x/sys/cpu"

	"github.com/lima-vm/lima/pkg/identifierutil"
	. "github.com/lima-vm/lima/pkg/must"
	"github.com/lima-vm/lima/pkg/networks"
	"github.com/lima-vm/lima/pkg/ptr"
	"github.com/lima-vm/lima/pkg/store/dirnames"
	"github.com/lima-vm/lima/pkg/store/filenames"
	"github.com/lima-vm/lima/pkg/version/versionutil"
```

Well maybe the 2nd block was supposed to be external packages, except for `lima-vm`, and the 3rd block was just `lima-vm`? Except the 2nd block has 2 `lima-vm` packages too. And in other files we only have 2 blocks, so I tried to match that style.

AkihiroSuda:
CI failing

pkg/limayaml/defaults.go
AkihiroSuda: > Well maybe the 2nd block was supposed to be external packages, except for lima-vm, and the 3rd block was just lima-vm?

👍 
Should be covered in another PR with a linter rule to enforce that

pkg/cidata/cidata.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	"github.com/docker/go-units"
 	"github.com/lima-vm/lima/pkg/debugutil"
-	"github.com/lima-vm/lima/pkg/identifierutil"
+	instance "github.com/lima-vm/lima/pkg/instance/hostname"
```
nirs, line 24: This is very confusing - typically the package name is the same as the directory name. You said that we already have an instance package, but you added a second instance package in this change, no?

pkg/sshutil/format.go
```diff
@@ -42,11 +42,11 @@ const (
...
-// TODO: consider supporting "url" format (ssh://USER@HOSTNAME:PORT)
-//
-// TODO: consider supporting "json" format
-// It is unclear whether we can just map ssh "config" into JSON, as "config" has duplicated keys.
-// (JSON supports duplicated keys too, but not all JSON implementations expect JSON with duplicated keys)
+	// TODO: consider supporting "url" format (ssh://USER@HOSTNAME:PORT).
+	//
+	// TODO: consider supporting "json" format.
+	// It is unclear whether we can just map ssh "config" into JSON, as "config" has duplicated keys.
+	// (JSON supports duplicated keys too, but not all JSON implementations expect JSON with duplicated keys).
```
nirs, line 49: This unrelated change should in another commit. If we don't want prs with multiple commit it should be in another pr.

pkg/store/instance.go
```diff
@@ -20,7 +20,7 @@ import (
 
 	"github.com/docker/go-units"
 	hostagentclient "github.com/lima-vm/lima/pkg/hostagent/api/client"
-	"github.com/lima-vm/lima/pkg/identifierutil"
+	instancehostname "github.com/lima-vm/lima/pkg/instance/hostname"
```
nirs, line 23: I guess here we import the other instance package? this shows that we should not call the new package "instance".

pkg/sshutil/format.go
jandubois: Yeah, this was kind of forced on me by `gofmt`, and then I had to add the trailing `dot` to keep the linter happy.

I need to investigate why this wasn't get flagged by CI before.

I agree that it should have been a separate PR. Requiring commits to be squashed is not a hard rule; we make exceptions when there are good reasons for it.

PR #3579, opened 2025-05-25
Fix guestagent service file

### LINKED ISSUE #3467: `/etc/init.d/lima-guestagent` uses `LIMA_CIDATA_*` environment variables
### Description

That works when the service is started from within the boot scripts, but not if the service needs to be restarted later.

The easiest fix is probably to just write the relevant variables to `/etc/conf.d/lima-guestagent`.

songponssw:
Fix #3467 

I think that the service file is supposed to be created with variables during the boot process.
I’ve fixed the heredoc to print LIMA_CIDATA_* in the boot script.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
@@ -22,6 +22,9 @@ install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent "${LIMA_CIDATA_GUEST_INSTALL
 
 # Launch the guestagent service
 if [ -f /sbin/openrc-run ]; then
+	# Convert .env to conf.d by wrapping values in double quotes.
+	# Split the variable and value at the first "=" to handle cases where the value contains additional "=" characters.
+	sed -E 's/^([^=]+)=(.*)/\1="\2"/' "${LIMA_CIDATA_MNT}/lima.env" >"/etc/conf.d/lima-guestagent"
 	# Install the openrc lima-guestagent service script
 	cat >/etc/init.d/lima-guestagent <<'EOF'
 #!/sbin/openrc-run


jandubois:
I don't think interpolating the values in the script directly, and having to escape all normal '$' signsis maintenance-friendly. You could define the values first with an interpolated block, and then append the rest of the script verbatim:

```shell
cat >/etc/init.d/lima-guestagent <<EOF
#!/sbin/openrc-run
LIMA_CIDATA_GUEST_INSTALL_PREFIX=$LIMA_CIDATA_GUEST_INSTALL_PREFIX
...

EOF

cat >>/etc/init.d/lima-guestagent <<'EOF'
supervisor=supervise-daemon

log_file="${log_file:-/var/log/${RC_SVCNAME}.log}"
...
EOF
```

But then you can just as well write the values to `/etc/conf.d/lima-guestagent`.

songponssw:
Got it! 
Actually, the `lima.env` file contains the `LIMA_CIDATA_*`, so I just print it into `conf.d/lima-guestagent`.

jandubois:
> Got it! Actually, the `lima.env` file contains the `LIMA_CIDATA_*`, so I just print it into `conf.d/lima-guestagent`.

There is one problem with that though: `lima.env` uses the syntax of `/etc/environment`, but `/etc/conf.d` files are sourced as shell scripts.

So in `lima.env` you have `LIMA_CIDATA_...=some random text" and everything until either a newline or the `#` character will be part of the value.

In shell this can be a syntax error (or even run an unintended command), so the string needs to be quoted using the shell quoting rules.

Now, it looks like we don't have any crazy values in there, that can in turn contain quotes etc, so simply surrounding the value with double-quotes should be good enough.

The only value that I found creating problems was the user comment field (full name):

```
LIMA_CIDATA_COMMENT=Jan Dubois
```

And it is just noise, but I still think we should do this properly:

```console
lima-alpine:~# rc-service lima-guestagent stop
/usr/libexec/rc/sh/openrc-run.sh: /etc/init.d/../conf.d/lima-guestagent: line 5: Dubois: not found
 * Stopping lima-guestagent ...                                                                                         [ ok ]
lima-alpine:~# rc-service lima-guestagent start
/usr/libexec/rc/sh/openrc-run.sh: /etc/init.d/../conf.d/lima-guestagent: line 5: Dubois: not found
 * Starting lima-guestagent ...                                                                                         [ ok ]
```

songponssw:
Appreciate the clarification. I will make the changes.

pkg/cidata/cidata.TEMPLATE.d/boot/25-guestagent-base.sh
```diff
@@ -22,6 +22,8 @@ install -m 755 "${LIMA_CIDATA_MNT}"/lima-guestagent "${LIMA_CIDATA_GUEST_INSTALL
 
 # Launch the guestagent service
 if [ -f /sbin/openrc-run ]; then
+	# Export LIMA_CIDATA_* for service
+	awk -F= '{ print $1 "=\"" $2 "\"" }' "${LIMA_CIDATA_MNT}/lima.env" >"/etc/conf.d/lima-guestagent"
```
jandubois, line 26: This doesn't work if a value contains the `=` character. I would suggest:

```suggestion
	sed -E 's/^([^=]+)=(.*)/\1="\2"/' "${LIMA_CIDATA_MNT}/lima.env" >"/etc/conf.d/lima-guestagent"
```

Please test, and add a comment _why_ this operation is necessary.

jandubois (approved):
Thanks, LGTM

PR #3590, opened 2025-05-27
Change package name of `instance/hostname` from `instance` to `hostname`

jandubois:
https://github.com/lima-vm/lima/pull/3578#discussion_r2107430960


jandubois:
BTW, I tried to move `Hostname` into `LimaYAML` (and `HostnameFromInstName` into the regular `instance` package), but there would then be a circular dependency between `instance` and `limayaml`, so it doesn't work.

AkihiroSuda (approved):
Thanks

Issue #3444, opened 2025-04-18
Log a warning if a mounted directory does not exist


jandubois:
> > `--mount writable` seems a typo of `--mount-writable`
> 
> Maybe we should validate that all directories specified exist before attempting to start the VM?
> 
> In the current behavior Lima just mkdirs it in such as case, probably on purpose.
> A warning should be still printed though. 

 _Originally posted by @AkihiroSuda in [#3438](https://github.com/lima-vm/lima/issues/3438#issuecomment-2814263688)_

PR #3595, opened 2025-05-31
Fix logging non-existent path

songponssw:
Fix: #3444

### DIFF
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -163,6 +163,9 @@ func Validate(y *LimaYAML, warn bool) error {
 			if !errors.Is(err, os.ErrNotExist) {
 				errs = errors.Join(errs, fmt.Errorf("field `mounts[%d].location` refers to an inaccessible path: %q: %w", i, f.Location, err))
 			}
+			if warn {
+				logrus.Warnf("field `mounts[%d].location` refers to a non-existent directory: %q:", i, f.Location)
+			}
 		} else if !st.IsDir() {
 			errs = errors.Join(errs, fmt.Errorf("field `mounts[%d].location` refers to a non-directory path: %q: %w", i, f.Location, err))
 		}


pkg/limayaml/validate.go
```diff
@@ -159,6 +161,7 @@ func Validate(y *LimaYAML, warn bool) error {
 			if !errors.Is(err, os.ErrNotExist) {
 				return fmt.Errorf("field `mounts[%d].location` refers to an inaccessible path: %q: %w", i, f.Location, err)
 			}
+			mountWarnings = append(mountWarnings, fmt.Sprintf("field `mounts[%d].location` non-existent directory: %q:", i, f.Location))
```
AkihiroSuda:
```suggestion
			mountWarnings = append(mountWarnings, fmt.Sprintf("field `mounts[%d].location` refers to a non-existent directory: %q:", i, f.Location))
```

pkg/limayaml/validate.go
```diff
@@ -180,6 +183,12 @@ func Validate(y *LimaYAML, warn bool) error {
 		}
 	}
 
+	if warn && len(mountWarnings) > 0 {
+		for _, warning := range mountWarnings {
+			logrus.Warn(warning)
+		}
+	}
```
AkihiroSuda: Why do we need this slice?
Can we just print a warning as soon as facing an `ErrNotExist`?

songponssw: If there are multiple non-existent paths, I think it would be better to validate all of them and display every missing path at once, rather than stopping at the first one.

songponssw: Just to clarify, it was meant to improve the user experience. 
I'm happy to adjust it according to your feedback.

jandubois: > I think it would be better to validate all of them and display every missing path at once, rather than stopping at the first one.

Yes, but instead of appending the warnings to a slice, you could report on them as you encounter them. You still continue checking the rest.

Creating the slice only makes sense if it was needed for something else as well.

pkg/limayaml/validate.go
```diff
@@ -159,6 +161,7 @@ func Validate(y *LimaYAML, warn bool) error {
 			if !errors.Is(err, os.ErrNotExist) {
 				return fmt.Errorf("field `mounts[%d].location` refers to an inaccessible path: %q: %w", i, f.Location, err)
 			}
+			mountWarnings = append(mountWarnings, fmt.Sprintf("field `mounts[%d].location` refers to a non-existent directory: %q:", i, f.Location))
```
jandubois: This line should just be

```suggestion
			logrus.Warnf("field `mounts[%d].location` refers to a non-existent directory: %q:", i, f.Location)
```

However, I wonder if we shouldn't just treat any error to `stat` the mount location as an error? @AkihiroSuda WDYT?

AkihiroSuda: > However, I wonder if we shouldn't just treat any error to stat the mount location as an error?

Why?

AkihiroSuda:
Please squash the commits

pkg/limayaml/validate.go
jandubois: How does it make sense to mount a directory that does not exist?

If `stat` succeeds, but the entry is not a directory, we throw an error. How is a non-existing entry any different: it also is not a directory.

pkg/limayaml/validate.go
songponssw: I just realized that I had been treating errors and warnings as equally critical, which made me think the function would return immediately after the first error. 
In fact, it just prints a warning message, and we still create the new directory if it doesn't exist, right?

pkg/limayaml/validate.go
jandubois: Never mind, we create the directory, if it doesn't exist. So a warning is fine.

jandubois (changes requested):
* The `mountWarnings` variable is no longer needed. It is now always empty, so the loop at the bottom doesn't do anything.

* The `logrus.Warnf()` call must be wrapped with `if warn {}`.

* All commits need to be squashed.

* The commit needs a DCO signature.

AkihiroSuda (approved):
Thanks

PR #3598, opened 2025-06-01
Fix installation of dummy WSL2 distro

### LINKED ISSUE #3597: WSL2 CI failing: `The imported file is not a valid Linux distribution.`
https://github.com/lima-vm/lima/actions/runs/15362822635/job/43232038944

```
The imported file is not a valid Linux distribution.

Error code: Wsl/Service/RegisterDistro/WSL_E_NOT_A_LINUX_DISTRO

Windows Subsystem for Linux has no installed distributions.

You can resolve this by installing a distribution with the instructions below:



Use 'wsl.exe --list --online' to list available distributions

and 'wsl.exe --install <Distro>' to install.
```

jandubois:
`wsl --import` no longer accepts an empty tarball, but checks that the tarball contains at least `/bin/sh` and `/etc`.

https://github.com/microsoft/WSL/blob/9cd3438/src/linux/init/main.cpp#L2621-L2629

Great that most of WSL2 is now open source! 🎊 

Fixes #3597

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -148,8 +148,15 @@ jobs:
         # otherwise `wsl --list --verbose` (called from Lima) fails:
         # https://github.com/lima-vm/lima/pull/1826#issuecomment-1729993334
         # The distro image itself is not consumed by Lima.
+        # Starting with WSL2 version 2.5.7.0 the distro will be rejected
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks 

We should also eventually try to remove this "dummy" distro

PR #3602, opened 2025-06-04
templates: Add almalinux-10

refi64:
This adds a template for the relatively new AlmaLinux 10 release and updates the `almalinux.yaml` symlink.

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -12,7 +12,8 @@ Default: [`default`](./default.yaml) (⭐Ubuntu, with containerd/nerdctl)
 
 Distro:
 - [`almalinux-8`](./almalinux-8.yaml): AlmaLinux 8
-- [`almalinux-9`](./almalinux-9.yaml), `almalinux.yaml`: AlmaLinux 9
+- [`almalinux-9`](./almalinux-9.yaml): AlmaLinux 9
+- [`almalinux-10`](./almalinux-10.yaml), `almalinux.yaml`: AlmaLinux 10
 - [`almalinux-kitten-10`](./almalinux-kitten-10.yaml), `almalinux-kitten.yaml`: AlmaLinux Kitten 10
 - [`alpine`](./alpine.yaml): ☆Alpine Linux
 - [`alpine-iso`](./alpine-iso.yaml): ☆Alpine Linux (ISO9660 image). Compatible with the `alpine` template used in Lima prior to v1.0.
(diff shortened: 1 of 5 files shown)


templates/README.md
```diff
@@ -12,7 +12,8 @@ Default: [`default`](./default.yaml) (⭐Ubuntu, with containerd/nerdctl)
 
 Distro:
 - [`almalinux-8`](./almalinux-8.yaml): AlmaLinux 8
-- [`almalinux-9`](./almalinux-9.yaml), `almalinux.yaml`: AlmaLinux 9
+- [`almalinux-9`](./almalinux-9.yaml): AlmaLinux 9
+- [`almalinux-10`](./almalinux-9.yaml), `almalinux.yaml`: AlmaLinux 10
```
AkihiroSuda: The link is wrong

templates/almalinux.yaml
```diff
@@ -1 +1 @@
-almalinux-9.yaml
\ No newline at end of file
+almalinux-10.yaml
```
AkihiroSuda, line 1: Probably we should bump up almalinux, rockylinux, oraclelinux, and centos-stream to 10 at once

templates/README.md
refi64: oops, fixed

templates/almalinux.yaml
refi64: I don't mind reverting this part if needed for merge, but I don't think that's really a useful policy. All of those have somewhat different release cycles, especially CentOS Stream where 10 has been the "current" version for quite a while now, and Rocky and OL aren't even out with version 10 yet. Additionally, in practice people tend to choose one or two of them across infra, so there's no real utility to be gained from matching up all the versions.

jandubois: I agree with this. I think the unversioned template should point to the latest "stable" version. There is no reason for us to link the release schedules of different distros.

I think it would be good to write down a policy for when the unversioned templates get updated. E.g. what is the reason `centos-stream.yaml` still links to version 9?

Maybe we need to come up with a somewhat arbitrary definition of what "stable" means for each distro that doesn't define it itself? Even if it is something like "3 months after its initial release".

But given that Alma has Alma Kitten as upstream, I don't know why you would need to wait any longer once the project decides the current snapshot is "stable".

AkihiroSuda (approved):
Thanks

PR #3623, opened 2025-06-10
Support for VBox driver - for converting to external

afbjorklund:
Rebased the old VirtualBox driver from 2023, as an example to be converted into the new driver framework...

* #1277 

The qcow2 format was hardcoded in some places, and the imgutil didn't support VDI format-specific disks

Otherwise it seemed to still be working, with the limitations (only tested x86_64, and only sshfs mounts)

/cc @unsuman

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -235,6 +235,15 @@ func templateArgs(bootScripts bool, instDir, name string, instConfig *limayaml.L
 		args.MountType = "virtiofs"
 	}
 
+	switch *instConfig.VMType {
+	case limayaml.QEMU:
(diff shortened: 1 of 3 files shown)


unsuman:
> Rebased the old VirtualBox driver from 2023, as an example to be converted into the new driver framework...
> 
> * [Support for VBox driver #1277](https://github.com/lima-vm/lima/pull/1277)
> 
> The qcow2 format was hardcoded in some places, and the imgutil didn't support VDI format-specific disks
> 
> Otherwise it seemed to still be working, with the limitations (only tested x86_64, and only sshfs mounts)
> 
> /cc @unsuman

Thanks!

PR #3654, opened 2025-06-24
Search for the data directory using real executable on non-Linux platform

AkinoKaede:
The problem NixOS/nixpkgs#419596 is similar to #3621, but `os.Executable()` only resolves the issue for Linux systems by retrieving the real path via /proc/self/exe. However, this solution does not address the issue on other operating systems, particularly macOS, where symbolic links still need to be resolved manually.

### DIFF
--- a/pkg/usrlocalsharelima/usrlocalsharelima.go
+++ b/pkg/usrlocalsharelima/usrlocalsharelima.go
@@ -57,8 +57,16 @@ func Dir() (string, error) {
 	selfViaOS, err := os.Executable()
 	if err != nil {
 		logrus.WithError(err).Warn("failed to find os.Executable()")
-	} else if len(selfPaths) == 0 || selfViaOS != selfPaths[0] {
-		selfPaths = append(selfPaths, selfViaOS)
+	} else {
+		selfFinalPathViaOS, err := filepath.EvalSymlinks(selfViaOS)
+		if err != nil {
+			logrus.WithError(err).Warn("failed to resolve symlinks")
+			selfFinalPathViaOS = selfViaOS // fallback to the original path
+		}
+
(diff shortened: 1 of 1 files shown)


jandubois (changes requested):
Maybe I misunderstand, but why do we need platform-specific code here? Can't we just always call `filepath.EvalSymlinks()` on the result of `os.Executable()`?

On Linux it shouldn't do anything because `/proc/self/exe` should never return a symlink.

If I'm correct, then this PR is adding 2 extra source code files to avoid calling `filepath.EvalSymlinks()` once in the lifetime of the process. So maybe let's just always call it?

jandubois:
Also, general comments:

* You need to sign the commit with a DCO (see link in the failed check)
* Commits should generally be squashed

AkinoKaede:
> Maybe I misunderstand, but why do we need platform-specific code here? Can't we just always call `filepath.EvalSymlinks()` on the result of `os.Executable()`?
> 
> On Linux it shouldn't do anything because `/proc/self/exe` should never return a symlink.
> 
> If I'm correct, then this PR is adding 2 extra source code files to avoid calling `filepath.EvalSymlinks()` once in the lifetime of the process. So maybe let's just always call it?

You're right that on Linux, `os.Executable()` doesn’t return a symlink, and theoretically, calling `filepath.EvalSymlinks()` isn't necessary. The platform-specific approach in the PR is mainly to avoid the extra call on Linux for performance reasons.

However, after considering the small performance overhead, I decided to support your suggestion and simplify the code by always calling `filepath.EvalSymlinks()`. The performance impact is negligible, and the added complexity doesn’t seem necessary. Thanks for the great suggestion!

jandubois (approved):
Thanks, LGTM

PR #3677, opened 2025-07-03
Implement `limactl network (list|create|delete)`

### LINKED ISSUE #3672: `limactl network` CLI for manipulating `~/.lima/_config/networks.yaml`
There should be the CLI for manipulating `~/.lima/_config/networks.yaml`.

e.g.
```console
$ limactl network list
NAME		MODE		SUBNET
user-v2	user-v2	192.168.104.0/24
shared	shared	192.168.105.0/24
bridged	bridged	-
host		host		192.168.106.0/24

$ limactl network create foo
{
  "name": "foo",
  "mode": "user-v2",
  "gateway": "192.168.107.1",
  "netmask": "255.255.255.0",
  "labels": null
}

$ limactl network delete foo
```

AkihiroSuda:
Fix #3672

e.g.,
```console
$ limactl network create foo --gateway 192.168.42.1/24

$ limactl network list
NAME       MODE       GATEWAY             INTERFACE
bridged    bridged    -                   en0
foo        user-v2    192.168.42.1/24     -
host       host       192.168.106.1/24    -
shared     shared     192.168.105.1/24    -
user-v2    user-v2    192.168.104.1/24    -

$ limactl network list --json foo bridged
{"mode":"user-v2","gateway":"192.168.42.1","netmask":"255.255.255.0"}
{"mode":"bridged","interface":"en0"}

$ limactl network delete --force foo
```

### DIFF
--- a/cmd/limactl/network.go
+++ b/cmd/limactl/network.go
@@ -1,0 +1,271 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"maps"
+	"net"
+	"os"
+	"slices"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/lima-vm/lima/pkg/networks"
+	"github.com/lima-vm/lima/pkg/yqutil"
+)
+
+const networkCreateExample = `  Create a network:
+  $ limactl network create foo --gateway 192.168.42.1/24
+
+  Connect VM instances to the newly created network:
+  $ limactl create --network lima:foo --name vm1
+  $ limactl create --network lima:foo --name vm2
+`
+
+func newNetworkCommand() *cobra.Command {
+	networkCommand := &cobra.Command{
+		Use:     "network",
+		Short:   "Lima network management",
+		Example: networkCreateExample,
+		GroupID: advancedCommand,
+	}
+	networkCommand.AddCommand(
+		newNetworkListCommand(),
+		newNetworkCreateCommand(),
+		newNetworkDeleteCommand(),
+	)
+	return networkCommand
+}
+
+func newNetworkListCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:               "list",
+		Short:             "List networks",
+		Aliases:           []string{"ls"},
+		Args:              WrapArgsError(cobra.ArbitraryArgs),
+		RunE:              networkListAction,
+		ValidArgsFunction: networkBashComplete,
+	}
+	flags := cmd.Flags()
+	flags.Bool("json", false, "JSONify output")
+	return cmd
+}
+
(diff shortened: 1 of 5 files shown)


nirs:
> $ limactl network list --json foo bridged
> {"mode":"user-v2","gateway":"192.168.42.1","netmask":"255.255.255.0"}
> {"mode":"bridged","interface":"en0"}

This is not json format, this is jsonlines format:
https://jsonlines.org

Trying to parse this string:
```
{"mode":"user-v2","gateway":"192.168.42.1","netmask":"255.255.255.0"}
{"mode":"bridged","interface":"en0"}
``` 
will not work with most json parsers.

The fact that Go json encoded and decoder supports this format does not make it json.

To make this valid json we need to return a list:

```
[
  {"mode":"user-v2","gateway":"192.168.42.1","netmask":"255.255.255.0"},
  {"mode":"bridged","interface":"en0"}
]
```

jsonlines format makes sense for logs when you want to stream the data.

> $ limactl network delete foo
> ```

How the case of deleting a network in use is handled? For example deleting the shared network when instances are running using it. I think rejecting the request is the simplest way to avoid issues.

If we have command to create/list/delete networks, maybe support also start/stop and simplify socket_vmnet network management?

AkihiroSuda:
> jsonlines

Intentionally adopted for consistency with `limactl ls` and `limactl disk ls`

> in use

Not handled yet

nirs:
> > jsonlines
> 
> Intentionally adopted for consistency with `limactl ls` and `limactl disk ls`

Makes sense.

Using --json is ok if we document in the help message that we return jsonlines format.
This is easy to handle by reading lines and parsing them separately.

jandubois:
> jsonlines format makes sense for logs when you want to stream the data.

It also makes sense for anything that is processed with a pipeline. It makes your `jq` commands simpler because you don't need to start with `.[]`. And you can always wrap the objects in an array with `jq -s`, but I find I virtually never need that.

FWIW, `docker images --format json` also uses jsonlines format...

cmd/limactl/network.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+	}
+
+	if !strings.Contains(gateway, "/") {
+		gateway += "/24"
+	}
+	gwIP, gwMask, err := net.ParseCIDR(gateway)
+	if err != nil {
+		return fmt.Errorf("failed to parse CIDR %q: %w", gateway, err)
+	}
```
jandubois, line 200: Maybe add a check that this is not a loopback address?

AkihiroSuda: done

cmd/limactl/network.go
```diff
@@ -0,0 +1,271 @@
...
+func networkDeleteAction(cmd *cobra.Command, args []string) error {
+	flags := cmd.Flags()
+	force, err := flags.GetBool("force")
+	if err != nil {
+		return err
+	}
+	if !force {
+		return errors.New("`limactl network delete` currently always requires `--force`")
+		// Because the command currently does not check whether the network being removed is in use
+	}
```
AkihiroSuda, line 257: Updated the PR to always require `--force` for now

jandubois (approved):
Thanks, LGTM

PR #3679, opened 2025-07-03
Add timezone validation

mazzz1y:
My machine accidentally has a symlink in /etc/timezone instead of a proper timezone file, and I spent some time investigating why cloud-init doesn't install anything in the guest, with this error:

```
[    5.994555] cloud-init[685]: 2025-07-03 09:19:09,880 - util.py[WARNING]: Failed loading yaml blob. unacceptable character #x0000: special characters are not allowed
[    5.996367] cloud-init[685]:   in "<unicode string>", position 88
[    5.997206] cloud-init[685]: 2025-07-03 09:19:09,880 - cloud_config.py[WARNING]: Failed at merging in cloud config part from part-001: empty cloud config
```

This PR adds timezone validation using the native Go way. In my case, it prints this warning and falls back to an empty value:

```
WARN[0000] invalid timezone "TZif2\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\b\x00\x00..." 
```

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -20,7 +20,6 @@ import (
 	"strings"
 	"sync"
 	"text/template"
-	"time"
 
 	"github.com/coreos/go-semver/semver"
 	"github.com/docker/go-units"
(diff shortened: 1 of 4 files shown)


pkg/limayaml/defaults.go
```diff
@@ -137,21 +138,22 @@ func MACAddress(uniqueID string) string {
...
+	}
+
+	if tzBytes, err := os.ReadFile("/etc/timezone"); err == nil {
+		if tz := strings.TrimSpace(string(tzBytes)); isValidTimezone(tz) {
+			return tz
 		}
 	}
+
+	if zoneinfoFile, err := filepath.EvalSymlinks("/etc/localtime"); err == nil {
+		if tz := extractTimezoneFromPath(zoneinfoFile); isValidTimezone(tz) {
```
AkihiroSuda: Probably this should be moved to `*_unix.go`

mazzz1y:
I've added platform-specific build constraints as @AkihiroSuda suggested, refactored the function, and added tests for the time-zone extraction function.

pkg/limayaml/defaults.go
mazzz1y: Done, please check now

AkihiroSuda (approved):
Thanks

PR #3696, opened 2025-07-05
Add support for collecting instance IP address

### LINKED ISSUE #3616: Report VM IP addresses to host agent
Right now the only way to discover the IP address assigned to a vznat or vmnet interface is by running `ip a` inside the instance and parsing the output.

It would be nice if the guestagent would report this back to the hostagent, and `limactl list` could include this information.

songponssw:
## What does this PR do?
- Fix #3616 
- Adds support for generating an instance `instance_info.json` file.
- This file includes basic instance information: _name, IPv4, and IPv6 addresses._

### With GuestAgent:
- The raw output from `ip -j a` is embedded as `json_bytes` in the gRPC Events message.
- The GuestAgent listens for this event, extracts the IP information, and writes it to the `instance_info.json` file.

### Without GuestAgent (Plain mode):
- This PR updates the `executeSSH()` function signature to return both the command output and error (instead of just error).
- After completing any requirements setup, HostAgent uses `executeSSH()` to run `ip -j a`, processes the output, and writes the instance info locally.

## Testing
- Setup: Lima upstream master, OS(Ubuntu 22.02)
- templates: almalinux alpine debian fedora opensuse oraclelinux ubuntu
- Test command: `limactl start INSTNAME --network lima:user-v2`
-  **Result**
    - With GuestAgent: All tests passed.
    - Plain Mode: Failed on AlmaLinux and Alpine — instances did not start successfully.

### DIFF
--- a/pkg/guestagent/api/guestservice.proto
+++ b/pkg/guestagent/api/guestservice.proto
@@ -11,11 +11,11 @@ service GuestService {
   rpc PostInotify(stream Inotify) returns (google.protobuf.Empty);
 
   rpc Tunnel(stream TunnelMessage) returns (stream TunnelMessage);
+
+  rpc GetIP(GetIPRequest) returns (GetIPRespond);
 }
 
-message Info {
-  repeated IPPort local_ports = 1;
-}
+message Info { repeated IPPort local_ports = 1; }
 
 message Event {
   google.protobuf.Timestamp time = 1;
@@ -42,3 +42,6 @@ message TunnelMessage {
   string guest_addr = 4;
   string udp_target_addr = 5;
 }
+
+message GetIPRequest { string ip_version = 1; }
+message GetIPRespond { string ip = 1; }
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -20,6 +20,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/guestagent/api"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/iptables"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/kubernetesservice"
+	"github.com/lima-vm/lima/v2/pkg/guestagent/metricutil"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/procnettcp"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/ticker"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/timesync"
@@ -45,8 +46,7 @@ func New(ctx context.Context, ticker ticker.Ticker, iptablesIdle time.Duration) 
 
 	auditStatus, err := auditClient.GetStatus()
 	if err != nil {
-		// syscall.EPERM is returned when using audit from a non-initial namespace
-		// https://github.com/torvalds/linux/blob/633b47cb009d09dc8f4ba9cdb3a0ca138809c7c7/kernel/audit.c#L1054-L1057
+		// syscall.EPERM is returned when using audit from a non-initial namespace https://github.com/torvalds/linux/blob/633b47cb009d09dc8f4ba9cdb3a0ca138809c7c7/kernel/audit.c#L1054-L1057
 		if !errors.Is(err, syscall.EPERM) {
 			return nil, err
 		}
@@ -370,4 +370,9 @@ func (a *agent) HandleInotify(event *api.Inotify) {
 			logrus.Errorf("error in inotify handle. Event: %s, Error: %s", event, err)
 		}
 	}
+}
+
+func (a *agent) IP(ctx context.Context, ipVersion string) (*api.GetIPRespond, error) {
+	out, err := metricutil.GetDefaultIP(ipVersion)
+	return &api.GetIPRespond{Ip: out}, err
 }
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -80,6 +80,8 @@ type HostAgent struct {
 
 	statusMu      sync.RWMutex
 	currentStatus events.Status
+
+	guestIPv4Address string
 }
 
 type options struct {
@@ -472,6 +474,7 @@ func (a *HostAgent) startRoutinesAndWait(ctx context.Context, errCh <-chan error
 func (a *HostAgent) Info(_ context.Context) (*hostagentapi.Info, error) {
 	info := &hostagentapi.Info{
 		SSHLocalPort: a.sshLocalPort,
+		GuestIPv4Address:      a.guestIPv4Address,
 	}
 	return info, nil
 }
@@ -598,6 +601,15 @@ sudo chown -R "${USER}" /run/host-services`
 		}
 		return errors.Join(rmErrs...)
 	})
+
+	if !*a.instConfig.Plain {
+		out, err := a.client.GetIP(ctx, &guestagentapi.GetIPRequest{IpVersion: "4"})
+		if err != nil {
+			logrus.Warnf("Received error from guestagent on collecting IPv4 %s", err)
+		}
+		a.guestIPv4Address = out.Ip
+	}
+
 	return errors.Join(errs...)
 }
 
--- a/pkg/guestagent/api/client/client.go
+++ b/pkg/guestagent/api/client/client.go
@@ -76,3 +76,7 @@ func (c *GuestAgentClient) Tunnel(ctx context.Context) (api.GuestService_TunnelC
 	}
 	return stream, nil
 }
+
+func (c *GuestAgentClient) GetIP(ctx context.Context, req *api.GetIPRequest) (*api.GetIPRespond, error) {
+	return c.cli.GetIP(ctx, req)
+}
--- a/pkg/guestagent/api/guestservice.pb.desc
+++ b/pkg/guestagent/api/guestservice.pb.desc
@@ -1,5 +1,5 @@
 
-…
+û
 guestservice.protogoogle/protobuf/empty.protogoogle/protobuf/timestamp.proto"0
 Info(
 local_ports (2.IPPortR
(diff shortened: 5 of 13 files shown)


pkg/guestagent/api/guestservice.proto
```diff
@@ -21,6 +21,7 @@ message Event {
   repeated IPPort local_ports_added = 2;
   repeated IPPort local_ports_removed = 3;
   repeated string errors = 4;
+  bytes json_bytes = 5;
```
AkihiroSuda: Why do we need two level of serialization: protobuf and json

pkg/guestagent/guestagent_linux.go
```diff
@@ -186,6 +188,24 @@ func (a *agent) collectEvent(ctx context.Context, st eventState) (*api.Event, ev
 	}
 	ev.LocalPortsAdded, ev.LocalPortsRemoved = comparePorts(st.ports, newSt.ports)
 	ev.Time = timestamppb.Now()
+
+	// Get network interfaces info
+	cmd := exec.Command("ip", "-j", "a")
```
AkihiroSuda: How is this frequently called, and how much CPU does this consume?

AkihiroSuda: Does this really work with busybox?

pkg/hostagent/hostagent.go
```diff
@@ -803,3 +826,76 @@ func copyToHost(ctx context.Context, sshConfig *ssh.SSHConfig, port int, local,
...
+		return err
+	}
+
+	// Extract IP address
+	var ifaces []events.NetworkInterfaceInfo
+	if err := json.Unmarshal(ipBytes, &ifaces); err != nil {
+		return err
+	}
+
+	var eth0 events.NetworkInterfaceInfo
```
AkihiroSuda: Why is it useful to report the IP of eth0?
The IP isn't reachable from the host, no?

pkg/store/filenames/filenames.go
```diff
@@ -68,7 +68,8 @@ const (
 	// SocketDir is the default location for forwarded sockets with a relative paths in HostSocket.
 	SocketDir = "sock"
 
-	Protected = "protected" // empty file; used by `limactl protect`
+	Protected = "protected"          // empty file; used by `limactl protect`
+	InstInfo  = "instance_info.json" // store instance information
```
AkihiroSuda: "info" isn't really descriptive name, as every file represents some "info"

AkihiroSuda: Either way, I don't think the IP has to be recorded in this persistent file

pkg/hostagent/hostagent.go
AkihiroSuda: The IP doesn't seem shown in the CLI either.
What's the intended use case of this PR?

pkg/guestagent/guestagent_linux.go
jandubois: It works with the Alpine cloud image from `template://alpine`. It does not work with alpine-lima from `template://alpine-iso`, but can easily be made to work by adding the `iproute2` package to the image.

PS: I assume this question was about Alpine, and not `busybox` because the `ip -j a` would be executed in the guest and not in a container.

afbjorklund: I just used a shell script for apple container, since it "forgot" to update the /etc/hosts file.

`echo "127.0.0.1 localhost" >/etc/hosts`
`echo "$(/sbin/ip -o -4 addr list eth0 | awk '{print $4}' | cut -d/ -f1) $(hostname)" >>/etc/hosts`

afbjorklund: I thought that Alpine used `ip` from busybox by default, but maybe that is what you are saying above

jandubois: A minimal Alpine uses just `busybox`, but you can `apk add iproute2` to install the real one. Our `alpine-lima` ISO image is the only one that is using busybox `ip` by default.

pkg/guestagent/api/guestservice.proto
songponssw: The `json_bytes` field holds the JSON-formatted output from an executed command.

However, I did use two-level serialization to include a label with the output. This label helps the host agent determine which command the output is from.
{
'ip': 'output from ip command',
'cpu': 'output from lscpu command',
}

I will change this implementation to something similar to IPPort by creating a new message type.
```
message CommandOutput {
string label = 1	// e.g. ip,lscpu,top
bool is_json = 2	
bytes payload = 3	// raw output
}
```

pkg/guestagent/guestagent_linux.go
songponssw: Understood. I will create a new API, such as GetIP(), which can be called only when the instance is running.

pkg/hostagent/hostagent.go
songponssw: - I thought `eth0` was the default interface, so we were just showing the IP from that. However, I think we could switch to using the command `ip -j route show default`, which would be more accurate since it gives us the interface associated with the default route. Does that sound good?
- My bad — I’ll include the IP in the CLI output in the next draft.

pkg/store/filenames/filenames.go
songponssw: Got it! Maybe adding the IP as an attribute in `store.instance` would be sufficient.

pkg/hostagent/hostagent.go
AkihiroSuda: > the interface associated with the default route

For what use case?

pkg/guestagent/api/guestservice.proto
AkihiroSuda: > This label helps the host agent determine which command the output is from.

Why does the host agent need to determine it?

pkg/hostagent/hostagent.go
songponssw: I meant to say it gives the ~interface~ `IP address` associated with the default route, not the interface. Sorry for the mix-up!

pkg/guestagent/api/guestservice.proto
songponssw: This message was designed to represent the output from multiple commands, with the label field helping the agent interpret the output and report it to the user.

However, I just realized that since we’ll have specific APIs like GetIP(), GetLsCpu(), and so on, the label is no longer necessary.
The host agent can already infer the source of the output based on which API was called.

pkg/hostagent/hostagent.go
AkihiroSuda: The IP associated with the default route isn't guaranteed to be reachable from the host by the IP

AkihiroSuda: But there might not be a robust way to determine the interface reachable from the host by the IP, so relying on the route metric might be acceptable as a heuristic.

Another heuristic would be to match the interface name (`lima%d`)

songponssw:
# Changes – Draft 2 (edited)

### gRPC
- added `CommandOutput` message to hold output from the client.
- Created dedicated endpoints for GetIPv4 and GetIPv6.

### Hostagent
- agent get ipaddress as the last step in host agent routine and it should all only when the instance spin up.
- Stores IP addresses internally
- Exports IP address through the `Info` strut in hostagent/api

### CLI
- Added `IPv4Address` and `IPv6Address` fields
- IP address are now available, e.g.: `limactl ls --format "{{ .Name }} {{ .IPv4Address }}"`

### Note;
- IPv6 is currently unavailable; no default route is found.
- gRPC compiled grpc with protoc-compiler 6.30.2 

P.S. I'm trying my best to follow the design patterns found in the existing code. If there are any points that need refactoring, please let me know.

pkg/guestagent/api/guestservice.proto
```diff
@@ -11,6 +11,8 @@ service GuestService {
   rpc PostInotify(stream Inotify) returns (google.protobuf.Empty);
 
   rpc Tunnel(stream TunnelMessage) returns (stream TunnelMessage);
+  rpc GetIPv4(google.protobuf.Empty) returns (CommandOutput);
+  rpc GetIPv6(google.protobuf.Empty) returns (CommandOutput);
 }
 
 message Info {
```
AkihiroSuda: I still think that the IP can be added here without wrapping it in `CommandOutput`

https://github.com/lima-vm/lima/issues/3616#issuecomment-3195732103

songponssw:
**Edit**: 
I had a bit of confusion. I thought we were trying to use net.Interface() as a way to query the IP address. 
I’ll continue making the changes according [Akihiro's comment](https://github.com/lima-vm/lima/pull/3696#discussion_r2281731072)


~~# PR Update (draft3)~~

~~## Details~~
~~- supports only IPv4.~~
~~- Hostagent requests the guest for its IP through RPC.~~
~~- `IPv4Address` is added to Hostagent Info, so it can be shown when using `curl` on `ha.sock` at `http://$INST/v1/info`.~~

~~## Determining the IP Address~~
~~- The IP is determined by finding an address that can dial `lima.host.internal`.~~
~~- `net.Interfaces()` returns a list of all interfaces on the machine. I’m not sure whether there will be only 2 interfaces in practice (loopback and primary interface). If that is the case, the implementation could be simplified to pick the primary interface from the list.~~

~~## Not sure how to implement it yet.~~
~~**Support IPv6:**~~
~~- With the current implementation, we would need an IPv6 address for `host.lima.internal`.~~
~~- Otherwise, we could assume it uses the same interfaces as IPv4 or dial another destination such as Google DNS or Cloudflare.~~

~~**Support Plain Mode:**~~
~~- If we want the same implementation to work for both normal and plain mode, we might need to transfer a small binary to the VM to execute and report the IP via RPC.~~
~~- I’m not sure if we can reliably get the IP during VM initialization. This might limit the ability for the client to edit the IP after creation. (I’m not quite sure yet exactly when and how IPs are assigned)~~
~~- In plain mode, there is no `lima.host.internal` in `/etc/hosts`. We would need to dial another destination to figure out the default route.~~

pkg/guestagent/api/guestservice.proto
songponssw: I’ll try adding the IP address directly to RPC and remove all CommandOutput wrapping

songponssw:
This will be implemented in #4175

PR #3697, opened 2025-07-06
guestagent: Fix isEmptyEvent check

nirs:
The intention seems to be checking if an event is equal to empty event, ignoring the event time, by copying the event and setting its time to nil. However instead of copying the event, we use the same object and set the event time to nil.

I don't know if this caused a user visible problem, but it likely log wrong event time if we log the event.

### DIFF
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -191,9 +191,8 @@ func (a *agent) collectEvent(ctx context.Context, st eventState) (*api.Event, ev
 
 func isEventEmpty(ev *api.Event) bool {
 	empty := &api.Event{}
-	copied := ev
-	copied.Time = nil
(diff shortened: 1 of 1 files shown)


pkg/guestagent/guestagent_linux.go
```diff
@@ -191,9 +191,8 @@ func (a *agent) collectEvent(ctx context.Context, st eventState) (*api.Event, ev
 
 func isEventEmpty(ev *api.Event) bool {
 	empty := &api.Event{}
-	copied := ev
```
AkihiroSuda, line 194: This regression seems caused in v0.21

https://github.com/lima-vm/lima/commit/72b7cb7fe2f8097eb363a64b557fd1258bba9f44#diff-4e96077f852132abc5fce986b8a80ec455c810b1db00d2cb23dbc8c4d2d9a365L182

PR #3707, opened 2025-07-09
Fix kernel decompression

### LINKED ISSUE #3705: VZ regression: kernel decompression is broken in master
> This PR disables kernel decompression for vz instances.
> before: https://github.com/lima-vm/lima/blob/b94e2fc19dd4bc0165116bbcafac9503f820eb8e/pkg/vz/disk.go#L45
> after: https://github.com/lima-vm/lima/blob/d32847cef5682569a257c7c60b28c08ca4a2c5a1/pkg/instance/start.go#L120
> 
> Since `vz` doesn’t support compressed kernels, booting with one causes the VM to fail with an internal error:
> > {"level":"fatal","msg":"Error Domain=VZErrorDomain Code=1 Description=\"Internal Virtualization error. The virtual machine failed to start.\" UserInfo={\n    NSLocalizedFailure = \"Internal Virtualization error.\";\n    NSLocalizedFailureReason = \"The virtual machine failed to start.\";\n}","time":"2025-07-09T09:48:08+09:00"}
> 
_Originally posted by @norio-nomura in https://github.com/lima-vm/lima/issues/3691#issuecomment-3050692006_

AkihiroSuda:
Fix #3705

### DIFF
--- a/pkg/instance/start.go
+++ b/pkg/instance/start.go
@@ -117,7 +117,8 @@ func Prepare(ctx context.Context, inst *store.Instance) (*Prepared, error) {
 				continue
 			}
 			if f.Kernel != nil {
-				if _, err := fileutils.DownloadFile(ctx, kernel, f.Kernel.File, false, "the kernel", *inst.Config.Arch); err != nil {
+				// ensure decompress kernel because vz expects it to be decompressed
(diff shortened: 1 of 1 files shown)


norio-nomura (approved):
As a VZ user, I’ve confirmed that this change fixes the issue.
If QEMU users don’t mind the kernel being decompressed, I think this approach is fine.
Thanks! 🙏🏻

PR #3721, opened 2025-07-12
templates: add new template for podman on ubuntu lts

aelsaman:
Add a new template for using Podman on Ubuntu instead of Fedora

### DIFF
--- a/templates/podman-ubuntu-lts.yaml
+++ b/templates/podman-ubuntu-lts.yaml
@@ -13,38 +13,38 @@
 minimumLimaVersion: 1.1.0
 
 base:
-  - template://_images/ubuntu-lts
-  - template://_default/mounts
+- template://_images/ubuntu-lts
+- template://_default/mounts
 
 containerd:
   system: false
   user: false
 provision:
-  - mode: system
-    script: |
-      #!/bin/bash
-      set -eux -o pipefail
-      command -v podman >/dev/null 2>&1 && test -e /etc/lima-podman && exit 0
-      export DEBIAN_FRONTEND=noninteractive
-      apt-get update
-      apt-get install -y podman && touch /etc/lima-podman
-  - mode: user
-    script: |
-      #!/bin/bash
-      set -eux -o pipefail
-      systemctl --user enable --now podman.socket
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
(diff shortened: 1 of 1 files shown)


afbjorklund:
What would be the benefit? We skipped the old Podman template, since they don't make latest version available for Ubuntu (for Docker we are using the vendor version rather than the system version, but Podman doesn't have one)

commit 6cc3f3f0c6179635d09e264769c6e76fa725e4ba

aelsaman:
> What would be the benefit? We skipped the old Podman template, since they don't make latest version available for Ubuntu

well I couldn't find a template that offers Podman on Ubuntu. But you're right, the version published for Ubuntu 24.04 is `4.9.3` which seems to be one year behind the latest `5.5.2`. 

However, the most recent Ubuntu version 25.04 seems to have a much recent version of podman `5.4` see the link below

https://packages.ubuntu.com/plucky/podman

afbjorklund:
It would be possible to provide our own podman, like we do with nerdctl 

But it would also be more work, and only Fedora and friends are "supported"?

https://github.com/afbjorklund/podman-full

Currently lima only has containerd archives built-in

----

EDIT: Note that the `main` branch has 4.9.3, while the `next` branch has 5.5.2

It should be possible to bootstrap it, also with the legacy buildah in podman.
(the build goes much faster with the modern buildkit in docker, but whatever)

I bumped the version of the compilers and dependencies, to match Fedora 42

afbjorklund:
Another alternative would be to use AlmaLinux, it uses a different RHEL branch of podman.

https://repology.org/project/podman/versions

https://github.com/containers/podman/branches/all

```
  origin/v4.5
  origin/v4.6
  origin/v4.6.1-rhel
  origin/v4.7
  origin/v4.8
  origin/v4.9
  origin/v4.9-rhel
  origin/v4.9.1-rhel
  origin/v4.9.3-rhel
  origin/v5.0
  origin/v5.1
  origin/v5.2
  origin/v5.2-rhel
  origin/v5.3
  origin/v5.4
  origin/v5.4-rhel
  origin/v5.5
```

But it is easier to just use Fedora for Podman, like using standard Ubuntu LTS for Docker...

aelsaman:
that make sense. Thanks for the explanation @afbjorklund 

I guess it is best to keep using Fedora for Podman for now. 

Thank you

PR #3722, opened 2025-07-13
rename UDPEstablished to UDPUnconnected

stek29:
I've spent some time debugging issues with UDP forwarding and got sidetracked because of this constant, so I decieded to submit this PR suggesting to change it

there's no "established" for the UDP, and 0x7 is usually called UNCONN in tools like net-tools or ss

strictly speaking there's no "unconnected" constant in the kernel either, but it's a well known way to call it.

this might be a matter of taste, so please feel free to close this PR if you disagree

### DIFF
--- a/pkg/guestagent/procnettcp/procnettcp.go
+++ b/pkg/guestagent/procnettcp/procnettcp.go
@@ -31,7 +31,7 @@ type State = int
 const (
 	TCPEstablished State = 0x1
 	TCPListen      State = 0xA
-	UDPEstablished State = 0x7
+	UDPUnconnected State = 0x7
 )
 
 type Entry struct {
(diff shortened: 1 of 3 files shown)


stek29:
the QEMU linux host seems to be flaky, since each time a different port has failed there

on the vz — I'm not sure what exactly the issue is there

pkg/guestagent/procnettcp/procnettcp.go
```diff
@@ -31,7 +31,9 @@ type State = int
 const (
 	TCPEstablished State = 0x1
 	TCPListen      State = 0xA
-	UDPEstablished State = 0x7
+	UDPUnconnected State = 0x7
+	// Deprecated: Use UDPUnconnected directly instead.
+	UDPEstablished State = UDPUnconnected
```
AkihiroSuda: You can just remove this without the grace period for depreciation

stek29: @AkihiroSuda done

PR #3725, opened 2025-07-14
remove unnecessary legacyBIOS setting

bonifaido:
this setting is not necessary anymore on Oracle Linux 9, but produces a warning on vz vm types


AkihiroSuda (approved):
Thanks

PR #3728, opened 2025-07-14
Create k0s template

plandem:


templates/k0s.yaml
AkihiroSuda, line 1: Please squash the commits and update https://github.com/lima-vm/lima/blob/master/templates/README.md

templates/README.md
```diff
@@ -55,6 +55,7 @@ Container image builders:
 
 Container orchestration:
 - [`faasd`](./faasd.yaml): [Faasd](https://docs.openfaas.com/deployment/edge/)
+- [`k0s`](./k0s.yaml): Kubernetes via k0s
```
afbjorklund: I think it could need a link, since "nobody" has heard of k0s (even though you can of course search for it)

Probably enough with https://k0sproject.io/ without having to link to Mirantis explicitly. (like k3s / Rancher)

afbjorklund (dismissed):
Otherwise it looks OK to me, haven't tested it but assume all the paths etc do work...

It is more or less similar to the k3s template, just slightly different units and installation

templates/README.md
plandem: its in sandbox already (https://www.cncf.io/blog/2025/05/20/a-milestone-for-lightweight-kubernetes-k0s-joins-cncf-sandbox/), so was sure no need for link :)

PR #3746, opened 2025-07-17
Fix: Pass `SSHLocalPort` as an option to configure driver

### LINKED ISSUE #3703: SSHLocalPort as a parameter, and not in instance
Why do we still need sshLocalPort as a parameter, and not just instance? Is that for the obsolete 60022

Eventually I wanted to refactor it into host and port (and not hardcode localhost), but that's a separate topic

_Originally posted by @afbjorklund in https://github.com/lima-vm/lima/issues/3691#issuecomment-3046362610_

unsuman:
This pull request refactors the configuration mechanism for drivers in the `pkg/driver` package, replacing the previous method of passing individual parameters with a more flexible options-based approach. The changes introduce a new `Options` struct and `ConfigOption` type, enabling dynamic configuration and improving code maintainability and extensibility.

Closes #3703


pkg/driver/driver.go
```diff
@@ -79,18 +81,78 @@ type Driver interface {
 
 	Info() Info
 
-	// SetConfig sets the configuration for the instance.
-	Configure(inst *store.Instance, sshLocalPort int) *ConfiguredDriver
+	// Configure sets the configuration for the instance.
+	Configure(opts ...ConfigOption) *ConfiguredDriver
```
afbjorklund: I think we want to keep the instance, since that will always be there.

Configure(inst *store.Instance, opts ...ConfigOption) *ConfiguredDriver

afbjorklund (changes requested):
I only meant to wrap the "int" second parameter in a struct

pkg/driver/external/driver.proto
```diff
@@ -41,8 +41,7 @@ message StartResponse {
 }
 
 message SetConfigRequest {
-  bytes instance_config_json = 1;
-  int64 ssh_local_port = 3;
+  bytes config_opts = 1;
```
afbjorklund: might have to decide on a transport here

pkg/driver/external/driver.proto
```diff
@@ -41,8 +41,7 @@ message StartResponse {
 }
 
 message SetConfigRequest {
-  bytes instance_config_json = 1;
```
afbjorklund, line 44: this one should remain, I think?

pkg/driver/external/driver.proto
```diff
@@ -42,7 +42,7 @@ message StartResponse {
 
 message SetConfigRequest {
   bytes instance_config_json = 1;
-  int64 ssh_local_port = 3;
+  bytes config_opts = 2;
```
AkihiroSuda: Why do we need two jsons ?
Can't we just set the SSH port in the `instance_config_json` ?

unsuman: Do you mean to set it in `Instance.SSHLocalPort`?

AkihiroSuda: Yes

unsuman: Then, is there a need for a ConfigOption?

AkihiroSuda: Probably no

afbjorklund: I thought the lima.yaml (config) said "0" for the port, and then the other parameter passed the current random port?

But if we edit the instance  config in the store each time, then we don't need any runtime parameters

AkihiroSuda (approved):
Thanks

afbjorklund:
I don't understand how it separates the fixed port, versus the dynamic port?
Wouldn't the new instance config now choose the same port every time?

Or are we relying on that it is only changed while loading the yaml, not on disk?
I guess it is not fundamentally different from any other FillDefaults etc

afbjorklund (approved):

PR #3748, opened 2025-07-18
Add feature to keep mounts with plain mode

### LINKED ISSUE #3747: Plain mode, with mounts
### Description

Currently enabling --plain mode disables all types of mounts. But we could keep the "default" ones (9p/virtiofs).

Add a new flag/feature, to keep the mounts in the template even with plain mode. But not for "reverse-sshfs".

* https://github.com/lima-vm/lima/discussions/3734

afbjorklund:
Allows using e.g. virtiofs in plain mode, without needing the guestagent etc.

But only for default mount type, since "reverse-sshfs" still needs non-plain.

Closes #3747

Discussion #3734

### DIFF
--- a/templates/default.yaml
+++ b/templates/default.yaml
@@ -555,6 +555,10 @@ guestInstallPrefix: null
 # 🟢 Builtin default: false
 plain: null
 
+# Keep the mounts, in "plain" mode.
+# 🟢 Builtin default: false
+plainMounts: null
+
 # When the "nestedVirtualization" feature is enabled:
 # - Allows running a VM inside the guest VM.
 # - The guest VM must configure QEMU with the `-cpu host` parameters to run a nested VM:
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -92,6 +92,7 @@ func RegisterCreate(cmd *cobra.Command, commentPrefix string) {
 	})
 
 	flags.Bool("plain", false, commentPrefix+"Plain mode. Disables mounts, port forwarding, containerd, etc.")
+	flags.Bool("plain-mounts", false, commentPrefix+"Keep any mounts, of the default type, in plain mode.")
 }
 
 func defaultExprFunc(expr string) func(v *flag.Flag) (string, error) {
@@ -257,6 +258,7 @@ func YQExpressions(flags *flag.FlagSet, newInstance bool) ([]string, error) {
 		{"disk", d(".disk= \"%sGiB\""), false, false},
 		{"vm-type", d(".vmType = %q"), true, false},
 		{"plain", d(".plain = %s"), true, false},
+		{"plain-mounts", d(".plainMounts = %s"), true, false},
 	}
 	var exprs []string
 	for _, def := range defs {
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -420,7 +420,11 @@ func (a *HostAgent) Info(_ context.Context) (*hostagentapi.Info, error) {
 
 func (a *HostAgent) startHostAgentRoutines(ctx context.Context) error {
 	if *a.instConfig.Plain {
-		logrus.Info("Running in plain mode. Mounts, port forwarding, containerd, etc. will be ignored. Guest agent will not be running.")
+		if *a.instConfig.PlainMounts {
+			logrus.Info("Running in plain mode, with mounts. Port forwarding, containerd, etc. will be ignored. Guest agent will not be running.")
+		} else {
+			logrus.Info("Running in plain mode. Mounts, port forwarding, containerd, etc. will be ignored. Guest agent will not be running.")
+		}
 	}
 	a.onClose = append(a.onClose, func() error {
 		logrus.Debugf("shutting down the SSH master")
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -908,6 +908,16 @@ func FillDefault(y, d, o *LimaYAML, filePath string, warn bool) {
 		y.Plain = ptr.Of(false)
 	}
 
+	if y.PlainMounts == nil {
+		y.PlainMounts = d.PlainMounts
+	}
+	if o.PlainMounts != nil {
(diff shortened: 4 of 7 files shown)


templates/default.yaml
```diff
@@ -555,6 +555,10 @@ guestInstallPrefix: null
 # 🟢 Builtin default: false
 plain: null
 
+# Keep the mounts, in "plain" mode.
+# 🟢 Builtin default: false
+plainMounts: null
```
AkihiroSuda, line 560: I fear the explosion of `plainFoo`, `plainBar`, `plainFooBar` properties.

Can we rather introduce `guestAgent.enabled: false` that will just disable the guest agent without affecting mounts

afbjorklund: The ask was only to be able to do virtiofs mounts with vz, without the guestagent (which is doable with this)

afbjorklund: But currently this is not related to the guestagent at all, just that enabling Plain will wipe out the Mounts

afbjorklund:
The flag could also be different. Perhaps related to mount?

```
      --mount-none          [limactl create] Remove all mounts
      --mount-keep          [limactl create] Keep mounts in plain mode
```

Or keep them by default, and require `--plain --mount-none`

Some templates (like k8s) do an explicit `mounts: []` up front...

msimkunas:
How about `--plain --with-mounts`?

One other alternative is to replace `--plain` and allow the user to disable specific features, e.g. `--disable-mounts`, `--disable-port-forwarding`, etc. but that would probably be too disruptive. 

Or, you could support multiple `--disable-*` flags and keep `--plain` as a metaflag that applies all `--disable-*` flags but this is getting pretty complicated.

msimkunas:
In any case, users may want to use `--plain` but enable port forwarding or just mounting. If you anticipate that in the future more exemption flags may be needed (not just `--with-mounts`) perhaps it is better to think about dedicated flags and treat `--plain` as an alias/metaflag that applies multiple flags at once

afbjorklund:
Having both `--mounts` (string) and `--with-mounts` (bool) is a little vague, and would --with-mounts=false be the same as --mount-none? Hmm

afbjorklund:
> treat --plain as an alias/metaflag that applies multiple flags at once

We could convert `plain` from a boolean to a struct, like we did with `disk`, but I hoped that it wouldn't be needed (complicates the yaml parsing and schema)

* https://github.com/lima-vm/lima/pull/1516

msimkunas:
@afbjorklund hence my other suggestion to introduce `--disable-*` flags for disabling specific features. `--plain` could then just be an alias for multiple such commands, e.g. `--disable-mounts --disable-port-forwarding ...`

afbjorklund:
All the other options do require the guestagent, though. Maybe not the timezone, but.

There is also another setting, "copy the user account", that is currently always enabled.

* https://github.com/lima-vm/lima/issues/3749#issuecomment-3088953430

But this got a bit complicated, it was only about the Mounts specifically

----

```go
func fixUpForPlainMode(y *LimaYAML) {
        if !*y.Plain {
                return
        }
        y.Mounts = nil
        y.PortForwards = nil
        y.Containerd.System = ptr.Of(false)
        y.Rosetta.BinFmt = ptr.Of(false)
        y.Rosetta.Enabled = ptr.Of(false)
        y.TimeZone = ptr.Of("")
}
```

AkihiroSuda:
`--disable-<FEAT>`  is inconsistent with the existing `--FEAT=(true|false)` flags.

So my suggestion is to introduce `--guestagent=(true|false)` (default: `false`) if we want a CLI flag.
Maybe it should rather be set via the YQ syntax (`--set '.guestagent.enabled=false'`).

afbjorklund:
I think we changed a bit away from "plain", and over to specific flags for the guestagent?

And we didn't quite explain what the guestagent is in the first place, or when you need it

jandubois:
So assuming we enabled mounts (except for `reverse-sshfs`) for `--plain` mode in Lima 2.0. Wouldn't `--plain --mount-none` already work?

So I'm not sure I see a need for additional options.

AkihiroSuda:
That's a breaking change and also inconsistent across mount types.
Sshfs does not work with the plain mode.
IIUC even virtiofs doesn't work on some distro in the plain mode as the relevant kmod is not installed by default. (Lazy to confirm it from phone 😅 )

msimkunas:
> That's a breaking change and also inconsistent across mount types.
> Sshfs does not work with the plain mode.
> IIUC even virtiofs doesn't work on some distro in the plain mode as the relevant kmod is not installed by default. (Lazy to confirm it from phone 😅 )

@AkihiroSuda Perhaps the bigger issue is that currently there's no way to specify which boot scripts must be run. In my case, I only need virtiofs support in the VM and nothing else, but currently I can either run all boot scripts or none.

I could perhaps solve this by copying the boot script which sets up the virtiofs mount and using it as a provisioning script though...

jandubois:
> Perhaps the bigger issue is that currently there's no way to specify which boot scripts must be run. In my case, I only need virtiofs support in the VM and nothing else, but currently I can either run all boot scripts or none.

The boot scripts are not an external facing API. We need to be able to modify them in incompatible ways on any release, without having to provide backwards compatibility.

Any configurability comes from settings in `lima.yaml`.

> I could perhaps solve this by copying the boot script which sets up the virtiofs mount and using it as a provisioning script though...

If we decide to make mounts available under `--plain`, then I think we should run the required boot scripts to install the needed drivers, but only if the list of mount points is non-empty.

So if somebody runs a VM with `--plain` and doesn't provide any mounts, they won't get any drivers installed.

afbjorklund:
Will close this PR, until we know how the config for plain mode should be handled. But it's basically a oneliner.

```diff
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -871,7 +871,6 @@ func fixUpForPlainMode(y *LimaYAML) {
        if !*y.Plain {
                return
        }
-       y.Mounts = nil
        y.PortForwards = nil
        y.Containerd.System = ptr.Of(false)
        y.Containerd.User = ptr.Of(false)
```

PR #3766, opened 2025-07-22
lima-guestagent: support external trigger via sighup notification

ciderale:
The Lima guest agent currently uses a polling strategy to update the ports exposed to the host system. For some use cases (e.g. testcontainers (https://github.com/rancher-sandbox/rancher-desktop/issues/3141)), this may not be responsive enough. This PR supplements the periodic ticker with an optional, push-based trigger based on OS signals. External processes can thus trigger a port update via SIGHUP, enabling a more timely response -- for example, by reacting to docker events. 

In my testing, this consistently reduced the time until a port became reachable -- without increasing the polling frequency.
The implementation requires minimal changes to lima-guestagent and is not tied to a specific event source. i.e. it is open to any kind of trigger. 

I hope this change is of value to this interesting project. I'm looking forward to your feedback.

## Validation Experiment

An example implementation of the trigger might look like:

```sh
docker events | while IFS= read -r _; do kill -HUP "$AGENT_PID"; done
```
which can be run in a systemd unit and optionally filter for specific events (e.g. start/stop/kill).

Timing experiment has been done with [ciderale/nixos-lima](https://github.com/ciderale/nixos-lima/tree/general-update) using lima 1.2.0 with [test-portmapping.sh](https://github.com/ciderale/nixos-lima/blob/general-update/scripts/test-portmapping.sh), starting 100 alpine/socat containers (each delayed by 0.3s, on different ports) and measuring time till each request has been fulfilled. The results are:

- Without the docker events SIGHUP trigger (i.e. the normal periodic update with default every 3s): the measured time is uniformly distributed between 240ms and 3160ms. As expected, the average is about half the polling frequency.
- With the proposed, additional SIGHUP trigger: the port is consistently available between 150ms to 220ms.

### DIFF
--- a/cmd/lima-guestagent/daemon_linux.go
+++ b/cmd/lima-guestagent/daemon_linux.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"net"
 	"os"
+	"os/signal"
+	"syscall"
 	"time"
 
 	"github.com/mdlayher/vsock"
@@ -58,7 +60,43 @@ func daemonAction(cmd *cobra.Command, _ []string) error {
 		// without depending on `bpftrace` binary.
 		// The agent binary will need CAP_BPF file cap.
 		ticker := time.NewTicker(tick)
-		return ticker.C, ticker.Stop
+
+		logrus.Info("register for sighup notifications")
+		sighupC := make(chan os.Signal, 1)
+		signal.Notify(sighupC, syscall.SIGHUP)
+
+		c := make(chan time.Time, 1)
+		go func() {
+			defer close(c)
+			defer close(sighupC)
+			for {
+				select {
+				case _, ok := <-sighupC:
+					if !ok {
+						logrus.Info("Sighup channel closed")
+						sighupC = nil;
+					} else {
+						logrus.Debug("Sighup received")
+						c <- time.Now()
+					}
+				case now, ok := <-ticker.C:
+					if !ok {
+						break;
+					} else {
+						c <- now
+					}
+				}
+			}
+			logrus.Info("ticker stopped")
+		}()
+
+		stop := func() {
+			logrus.Info("stopping ticker")
+			ticker.Stop()
+			signal.Stop(sighupC)
+		}
+
+		return c, stop
 	}
 
 	agent, err := guestagent.New(newTicker, tick*20)


jandubois:
Thank you for your PR!

There is however a more comprehensive PR in preparation, that will use the events API for containerd, dockerd, and kubernetes, to send immediate port notifications. For various non-technical reasons it has been in progress for a long time, but should be ready for consideration very soon. It is based on the Rancher Desktop guest agent on Windows. You can take a look at the in-progress changes at https://github.com/Nino-K/lima/pull/2

Given that this PR will hopefully land within the next couple of weeks, I think this signal mechanism should not be merged.

ciderale:
Thank you for the quick response. I see that the referenced PR is more tightly integrated (at the cost of more complexity). The openness of the signal based approach might still be interesting, though. But I see that most use cases (including my current one) are probably covered by that other PR. So I'm looking forward to that other PR to land. Thanks again.

cmd/lima-guestagent/daemon_linux.go
```diff
@@ -58,7 +60,43 @@ func daemonAction(cmd *cobra.Command, _ []string) error {
 		// without depending on `bpftrace` binary.
 		// The agent binary will need CAP_BPF file cap.
 		ticker := time.NewTicker(tick)
-		return ticker.C, ticker.Stop
+
+		logrus.Info("register for sighup notifications")
```
AkihiroSuda: Nit: probably sighup should be spelled in the upper chars, as it might look like "sigh up" (?) or a typo of sign-up

cmd/lima-guestagent/daemon_linux.go
```diff
@@ -58,7 +60,43 @@ func daemonAction(cmd *cobra.Command, _ []string) error {
...
+		sighupC := make(chan os.Signal, 1)
+		signal.Notify(sighupC, syscall.SIGHUP)
```
AkihiroSuda, line 68: Please add a comment line to explain who sends SIGHUP

ciderale: Sure, I have added a comment about the purpose of the SIGHUP mechanism and also improved the log message.

AkihiroSuda:
Turned this to a draft, as we are likely going to use eBPF for watching `bind(2)` syscalls to stimulate the event ticker
- #3067

We may still revisit this PR if there is a use case

PR #3774, opened 2025-07-30
makefile:use local gcc when native compiling

ningmingxiao:
fix 
```
cgo: C compiler "x86_64-linux-gnu-gcc" not found: exec: "x86_64-linux-gnu-gcc": executable file not found in $PATH
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -546,8 +546,10 @@ ifeq ($(GOOS),darwin)
 # returns the architecture name converted from GOARCH to macOS's uname -m.
 to_uname_m = $(foreach arch,$(1),$(shell echo $(arch) | sed 's/amd64/x86_64/'))
 else ifeq ($(GOOS),linux)
+ifeq ($cross_compiling),true)
 # CC is required for cross-compiling on Linux.
 CC = $(call to_uname_m,$(GOARCH))-linux-gnu-gcc
+endif
 else ifeq ($(GOOS),windows)
 # artifact in zip format also provided for Windows.
 ARTIFACT_FILE_EXTENSIONS += .zip


afbjorklund:
The idea is OK, but it should not need an extra variable to be set...

Maybe check `go env GOOS` (without GOOS set) versus `$GOOS`?

ningmingxiao:
can you describe clearly ? thanks @afbjorklund

afbjorklund:
We should be able to detect when doing a cross-compile, without requiring that information from the user?

The main issue is that the format of `uname -s` and GOOS is slightly different (as with `uname -m` and GOARCH)

Most Linux distributions feature both, so it has not been a problem.

i.e. both `gcc` and `x86_64-linux-gnu-gcc`

afbjorklund:
Ah, we already have some handy variables to use:

```Makefile
# returns difference between GOOS GOARCH and GOHOSTOS GOHOSTARCH.
cross_compiling = $(call diff,$(GOOS) $(GOARCH),$(GOHOSTOS) $(GOHOSTARCH))
# returns true if cross_compiling is empty.
native_compiling = $(if $(cross_compiling),,true)
```

We don't need to set anything, for a native build.

i.e. there is no need to hardcode `gcc`, but let go

ningmingxiao:
how about 
```
CC = $(call to_uname_m,$(GOARCH))-linux-gnu-gcc
ifeq ($(native_compiling),true)
	CC = gcc
endif
```
my host doesn't have  x86_64-linux-gnu-gcc
and some os have 
x86_64-anolis-linux-gcc instead of  x86_64-linux-gnu-gcc

afbjorklund:
If your host needs something special, you can override `CC`

`make GOOS=linux GOARCH=... CC=my-gcc`

But you can simplify it to:

```make
ifeq ($(cross_compiling),true)
# CC is required for cross-compiling on Linux.
CC = $(call to_uname_m,$(GOARCH))-linux-gnu-gcc
endif
```

Might need to move the declarations in the file, though.

EDIT: typo

ningmingxiao:
done, thank you.

afbjorklund:
Tests are failing, so the new test is not working properly. ~~(or the comment is wrong, and it always need CC?)~~

afbjorklund:
The cross_compiling variable is empty when false (and the diff, when true), unlike native_compiling \<sigh\>

```diff
+ifneq ($(cross_compiling),)
 # CC is required for cross-compiling on Linux.
 CC = $(call to_uname_m,$(GOARCH))-linux-gnu-gcc
+endif
```

AkihiroSuda:
CI is faling
```
# runtime/cgo
gcc_arm64.S: Assembler messages:
gcc_arm64.S:30: Error: no such instruction: `stp x29,x30,[sp,'
gcc_arm64.S:34: Error: too many memory references for `mov'
gcc_arm64.S:36: Error: no such instruction: `stp x19,x20,[sp,'
gcc_arm64.S:39: Error: no such instruction: `stp x21,x22,[sp,'
gcc_arm64.S:42: Error: no such instruction: `stp x23,x24,[sp,'
gcc_arm64.S:45: Error: no such instruction: `stp x25,x26,[sp,'
gcc_arm64.S:48: Error: no such instruction: `stp x27,x28,[sp,'
gcc_arm64.S:52: Error: too many memory references for `mov'
gcc_arm64.S:53: Error: too many memory references for `mov'
gcc_arm64.S:54: Error: too many memory references for `mov'
gcc_arm64.S:56: Error: no such instruction: `blr x20'
gcc_arm64.S:57: Error: no such instruction: `blr x19'
gcc_arm64.S:59: Error: no such instruction: `ldp x27,x28,[sp,'
gcc_arm64.S:62: Error: no such instruction: `ldp x25,x26,[sp,'
gcc_arm64.S:65: Error: no such instruction: `ldp x23,x24,[sp,'
gcc_arm64.S:68: Error: no such instruction: `ldp x21,x22,[sp,'
gcc_arm64.S:71: Error: no such instruction: `ldp x19,x20,[sp,'
gcc_arm64.S:74: Error: no such instruction: `ldp x29,x30,[sp],'
make[1]: *** [Makefile:251: _output/bin/limactl] Error 1
```
https://github.com/lima-vm/lima/actions/runs/16623363235/job/47033924352?pr=3774

afbjorklund:
Something is broken, the cross_compiling variable is **not** working as intended here: (from amd64 to arm64)

`make artifacts-linux`

`CGO_ENABLED=1 GOOS="linux" GOARCH="amd64" CC="cc"`

`CGO_ENABLED=1 GOOS="linux" GOARCH="arm64" CC="cc" `

AkihiroSuda:
Probably fixed in:
- https://github.com/lima-vm/lima/pull/4228

PR #3777, opened 2025-07-30
feature: add LocalPath to support use local resource to skip download

ningmingxiao:
some airgap environment doesn't support to download image. we can use local image when localPath is set.
for example if   localPath is set we can skip to download from https://cloud-images.ubuntu.com/releases/oracular/release-20250701/ubuntu-24.10-server-cloudimg-amd64.img
```
images:

- location: "https://cloud-images.ubuntu.com/releases/oracular/release-20250701/ubuntu-24.10-server-cloudimg-amd64.img"
  arch: "x86_64"
  digest: "sha256:69f31d3208895e5f646e345fbc95190e5e311ecd1359a4d6ee2c0b6483ceca03"
  localPath: "/tmp/ubuntu-24.10-server-cloudimg-amd64.img"
```

### DIFF
--- a/pkg/fileutils/download.go
+++ b/pkg/fileutils/download.go
@@ -5,9 +5,13 @@ package fileutils
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
+	"io"
+	"os"
 	"path"
+	"path/filepath"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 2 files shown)


jandubois:
Is this necessary? I would have expected this to already work:

```yaml
images:

- location: "https://cloud-images.ubuntu.com/releases/oracular/release-20250701/ubuntu-24.10-server-cloudimg-amd64.img"
  arch: "x86_64"
  digest: "sha256:69f31d3208895e5f646e345fbc95190e5e311ecd1359a4d6ee2c0b6483ceca03"

- location: "/tmp/ubuntu-24.10-server-cloudimg-amd64.img"
  arch: "x86_64"
```

Of if that doesn't work, try `file:///tmp/...`

Reorder the entries if you want to always prefer the local file, if it exists.

ningmingxiao:
thanks, it works well. @jandubois 
```
- location: "/tmp/ubuntu-24.10-server-cloudimg-amd64.img"
  arch: "x86_64"
```
I think we can add some doc to explain how to use  location in airgap environment.

jandubois:
> thanks, it works well. @jandubois

Thanks for confirming! I will close this PR then because it is not needed.

> I think we can add some doc to explain how to use location in airgap environment.

The "documentation" (in `default.yaml`) already states that `location` is a URL, so not sure what else to add. But feel free to create a docs PR if you think it needs further elaboration.

PR #3780, opened 2025-07-31
Stop using hard-coded port 60022 for "default" instance

AkihiroSuda:
The "default" instance no longer uses 60022 as the TCP port.

Discussed in:
- #3712

This change affects the existing "default" instance too, however, users do not need to care about the port number if they use `~/.lima/default/ssh.config`:
```
```


msimkunas:
Just wanted to note that users do care about ports in certain cases. For example, Jetbrains IDEs require you to specify the port manually even if you use the `lima-*` host alias to connect to the VM.

In such cases, the user will want to reserve a host port for exclusive VM use via `ssh.localPort` in the template.

pkg/hostagent/hostagent.go
```diff
@@ -237,17 +237,13 @@ func writeSSHConfigFile(sshPath, instName, instDir, instSSHAddress string, sshLo
...
 
-func determineSSHLocalPort(confLocalPort int, instName string) (int, error) {
+func determineSSHLocalPort(confLocalPort int) (int, error) {
 	if confLocalPort > 0 {
 		return confLocalPort, nil
 	}
 	if confLocalPort < 0 {
 		return 0, fmt.Errorf("invalid ssh local port %d", confLocalPort)
 	}
-	if instName == "default" {
```
afbjorklund, line 247: Did we want to keep this functionality, for instances created before 2.0.0? To not break any existing ones?

https://github.com/lima-vm/lima/issues/3712#issuecomment-3133506245

It would still need to be made explicit for any **new** instances.

website/content/en/docs/faq/_index.md
```diff
@@ -50,7 +50,8 @@ weight: 6
 Password is disabled and locked by default.
 You have to use `limactl shell bash` (or `lima bash`) to open a shell.
 
-Alternatively, you may also directly ssh into the guest: `ssh -p 60022 -i ~/.lima/_config/user -o NoHostAuthenticationForLocalhost=yes 127.0.0.1`.
+Alternatively, you may also directly ssh into the guest: `ssh -p <PORT> -i ~/.lima/_config/user -o NoHostAuthenticationForLocalhost=yes 127.0.0.1`.
+The port number can be inspected by running `limactl list`.
```
afbjorklund, line 54: Maybe we should also have a format here, like `limactl list --format '{{ .SSHLocalPort }}' default`

AkihiroSuda: done

pkg/hostagent/hostagent.go
AkihiroSuda: done

AkihiroSuda:
> In such cases, the user will want to reserve a host port for exclusive VM use via ssh.localPort in the template.

Added `--ssh-port` flag for convenience:
- https://github.com/lima-vm/lima/pull/3791

pkg/version/versionutil/versionutil.go
```diff
@@ -51,3 +51,8 @@ func GreaterThan(limaVersion, oldVersion string) bool {
 func GreaterEqual(limaVersion, oldVersion string) bool {
 	return compare(limaVersion, oldVersion) >= 0
 }
+
+// LessThan returns true if limaVersion < oldVersion.
+func LessThan(limaVersion, oldVersion string) bool {
+	return compare(limaVersion, oldVersion) < 0
```
jandubois: I don't think we need this function, but if you want to keep it, then it should be implemented in terms of the existing functions, that are already covered by unit tests.

```suggestion
	return !GreaterEqual(limaVersion, oldVersion)
```

AkihiroSuda: done

PR #3786, opened 2025-08-03
The wsl2 image source moved upstream in finch

afbjorklund:
Note that it is based on `docker.io/library/fedora:40` image,
in case the Dockerfile should disappear again in the future.

The current image was probably (based on date) created from:
https://github.com/runfinch/finch-core/blob/ab4113b684f69281a126e62a1c19d6e9bfbc25f8/rootfs/Dockerfile

```
public.ecr.aws/docker/library/fedora     41                1e77ba7428ee   13 days ago     161MB
fedora                                   41                1e77ba7428ee   13 days ago     161MB
```

### DIFF
--- a/templates/experimental/wsl2.yaml
+++ b/templates/experimental/wsl2.yaml
@@ -3,7 +3,7 @@
 vmType: wsl2
 
 images:
-# Source: https://github.com/runfinch/finch-core/blob/main/Dockerfile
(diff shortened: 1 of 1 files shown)


jandubois:
> Note that it is based on `docker.io/library/fedora:40` image, in case the Dockerfile should disappear again in the future.

I wonder if we should create a fedora-lima-wsl2 repo and provide the image as a GitHub release instead of relying on a downstream project for the image. Kind of like we do alpine-lima, but even simpler.

Issue #3782, opened 2025-08-01
Creating a new bridged network results in an error.


dharsanb:
### Description

When trying to use the command `limactl network create` to create a bridged network, it results in a error
`failed to parse CIDR "/24": invalid CIDR address: /24 `

I'm assuming the error is here because you cannot add gateway / IP ranges as they are managed outside. https://github.com/lima-vm/lima/blob/bd955e4d4c680984fd3e61bd3d32d35558ee1773/cmd/limactl/network.go#L177-L207

nirs:
@dharsanb can you show the full command that fails?

dharsanb:
Sorry,
Should've posted the full error
```bash
# Command
limactl network create inf --mode=bridged --interface en8
# Error
## FATA[0000] failed to parse CIDR "/24": invalid CIDR address: /24
```

PR #3813, opened 2025-08-09
Fix creating bridged network interface in limactl

dharsanb:
Fixes: https://github.com/lima-vm/lima/issues/3782

### DIFF
--- a/cmd/limactl/network.go
+++ b/cmd/limactl/network.go
@@ -182,6 +182,8 @@ func networkCreateAction(cmd *cobra.Command, args []string) error {
 		if intf == "" {
 			return fmt.Errorf("network mode %q requires specifying interface", mode)
 		}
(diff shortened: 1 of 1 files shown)


dharsanb:
Can I get some help here please.
Thanks.

cmd/limactl/network.go
```diff
@@ -174,23 +174,23 @@ func networkCreateAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	switch mode {
-	case networks.ModeBridged:
+	if mode == networks.ModeBridged {
```
nirs: I don't think you need to change the structure of the code to fix this issue. Keep the switch.

nirs (commented):
Not sure why you change code not related to bridged mode. The best way to fix this is to add a failing test and then add a minimal fix for the bridged case.

jandubois (commented):
Change looks good to me, but please squash commits.

dharsanb:
> Change looks good to me, but please squash commits.

Done.

PR #3818, opened 2025-08-11
Added checking for symlinked cmds

plandem:


afbjorklund:
Why resolve the symlinks?

plandem:
> Why resolve the symlinks?

e.g. i dont want to install docker cli locally, but want to symlink docker.lima as docker. e.g. some tools (e.g. Tilt.dev) will call "docker_build" or "kubectl_build" or even "nerdctl_build", then it will work. with this checking I can symlink it and use as regular docker cli, e.g. `LIMA_INSTANCE=lima_project_a docker`, which will call that wrapper with proper limactl shell docker. without that "self guard", wrapper will go in infinite loop, because "command -v" will return link to wrapper.

plandem:
> I think the change is fine, but should have a comment to explain why the check is needed because stripping the `.lima` suffix from the script names is not the regular use case.

Added comments to commit message. Is it enough or better to comment in scripts?

alexandear:
> Added comments to commit message. Is it enough or better to comment in scripts?

I vote for adding a comment explaining why we need this in three places:

- PR description.
- Commit message (make sure the first line of the commit does not end with a period `.`).
- In the scripts themselves.

Make sure the title of the PR heading equals to the first line of the commit.

afbjorklund:
I missed that you were using symlinks, instead of shell aliases.

Maybe this also needs to be mentioned in the documentation?

https://lima-vm.io/docs/examples/#running-containers

> you may make an alias (or a symlink) by yourself: `alias nerdctl=nerdctl.lima`

There are tabs for nerdctl / docker /podman / kubernetes, but no wrappers or symlinks.

cp -a cmd/lima _output/bin/lima
cp -a cmd/nerdctl.lima _output/bin/nerdctl.lima
cp -a cmd/apptainer.lima _output/bin/apptainer.lima
cp -a cmd/docker.lima _output/bin/docker.lima
cp -a cmd/podman.lima _output/bin/podman.lima
cp -a cmd/kubectl.lima _output/bin/kubectl.lima

Maybe I should have added something about `apptainer` (singularity / SIF), as well?

plandem:
> I missed that you were using symlinks, instead of shell aliases.

I needed it for Tilt.dev which spawns shells (non interactive, so can be some struggle with aliases). But I needed nerdctl and kubectl and you cant  install nerdctl on OSX and use socket for containerd in lima, at least I had no success. if there would be an option to opt-out any local cli tools and use wrappers only, I personally would go this way.

> 
> Maybe this also needs to be mentioned in the documentation?
> 
> https://lima-vm.io/docs/examples/#running-containers
> 
> > you may make an alias (or a symlink) by yourself: `alias nerdctl=nerdctl.lima`

like this? or something more.

plandem:
I think we can keep it as is and just update docs maybe about QoL? e.g. with something like `direnv`, we can create `.envrc` like 
```bash
export LIMA_INSTANCE=MyInstance
export DOCKER_HOST=$(limactl list "$LIMA_INSTANCE" --format "unix://{{.Dir}}/sock/docker.sock")
```

and then `lima` will work without dancing around. its not directly related to any command here except `lima` itself, but anyway.

Issue #3495, opened 2025-05-05
Lima fails to run, if deleting all the git tags before building


afbjorklund:
### Description

It chokes on the generated version, during validate:

`FATA[0001] the YAML is invalid, saved the buffer as "lima.REJECTED.yaml": can't parse builtin Lima version "edec9bc1": edec9bc1 is not in dotted-tri format`

It probably shouldn't generate an invalid version...

```Makefile
VERSION := $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
```

afbjorklund:
There is some workaround for this in the GitHub actions:

```yaml
    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
      with:
        # To avoid "failed to load YAML file \"templates/experimental/riscv64.yaml\": can't parse builtin Lima version \"3f3a6f6\": 3f3a6f6 is not in dotted-tri format"
        fetch-depth: 0
```

commit 762e5108658a38f8e8e94104cccc5778b17fef88 - to avoid shallow clones, without tags

afbjorklund:
This scenario also happens, when you don't set `pkg/version.Version` at all (i.e. run `go build`)

`FATA[0000] the YAML is invalid, saved the buffer as "lima.REJECTED.yaml": can't parse builtin Lima version "<unknown>": <unknown> is not in dotted-tri format`

Ideally there should be a state where the lima version is assumed to be "latest / greatest" (HEAD)
And then it would pass all tests, it would be greater than any existing version in the comparisions.

For sanity, you always want to use a real version - even if you have pass it to `make` as `VERSION`...

olamilekan000:
a pr has been opened for this
https://github.com/lima-vm/lima/pull/3820

olamilekan000:
@jandubois can you assign this to me

PR #3820, opened 2025-08-12
fix limactl run bug  by treating unparseable versions as latest instead of crashing during template validation.

olamilekan000:
```
change fixes limactl run bug  by treating unparseable versions as latest instead of crashing
during template validation.
```

fixes [issue](https://github.com/lima-vm/lima/issues/3495)

### DIFF
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -40,12 +40,10 @@ func Validate(y *LimaYAML, warn bool) error {
 		if _, err := versionutil.Parse(*y.MinimumLimaVersion); err != nil {
 			errs = errors.Join(errs, fmt.Errorf("field `minimumLimaVersion` must be a semvar value, got %q: %w", *y.MinimumLimaVersion, err))
 		}
-		limaVersion, err := versionutil.Parse(version.Version)
-		if err != nil {
-			errs = errors.Join(errs, fmt.Errorf("can't parse builtin Lima version %q: %w", version.Version, err))
-		}
-		if limaVersion != nil && versionutil.GreaterThan(*y.MinimumLimaVersion, limaVersion.String()) {
-			errs = errors.Join(errs, fmt.Errorf("template requires Lima version %q; this is only %q", *y.MinimumLimaVersion, limaVersion.String()))
+		// Unparsable version.Version (like commit hashes or "<unknown>") is treated as "latest/greatest"
+		// and will pass all version comparisons, allowing development builds to work.
+		if !versionutil.GreaterEqual(version.Version, *y.MinimumLimaVersion) {
+			errs = errors.Join(errs, fmt.Errorf("template requires Lima version %q; this is only %q", *y.MinimumLimaVersion, version.Version))
 		}
 	}
 	if y.VMOpts.QEMU.MinimumVersion != nil {
--- a/pkg/limayaml/validate_test.go
+++ b/pkg/limayaml/validate_test.go
@@ -38,19 +38,19 @@ func TestValidateMinimumLimaVersion(t *testing.T) {
 			name:               "minimumLimaVersion greater than current version",
 			currentVersion:     "1.1.1-114-g5bf5e513",
 			minimumLimaVersion: "1.1.2",
-			wantErr:            `template requires Lima version "1.1.2"; this is only "1.1.1"`,
+			wantErr:            `template requires Lima version "1.1.2"; this is only "1.1.1-114-g5bf5e513"`,
 		},
 		{
 			name:               "invalid current version",
 			currentVersion:     "<unknown>",
 			minimumLimaVersion: "0.8.0",
-			wantErr:            `can't parse builtin Lima version "<unknown>": <unknown> is not in dotted-tri format`,
+			wantErr:            "", // Unparsable versions are treated as "latest"
 		},
 		{
 			name:               "invalid minimumLimaVersion",
 			currentVersion:     "1.1.1-114-g5bf5e513",
 			minimumLimaVersion: "invalid",
-			wantErr:            "field `minimumLimaVersion` must be a semvar value, got \"invalid\": invalid is not in dotted-tri format\ntemplate requires Lima version \"invalid\"; this is only \"1.1.1\"",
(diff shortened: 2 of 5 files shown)


olamilekan000:
can someone review this? thanks @afbjorklund @jandubois

pkg/limayaml/validate.go
```diff
@@ -40,12 +40,10 @@ func Validate(y *LimaYAML, warn bool) error {
 		if _, err := versionutil.Parse(*y.MinimumLimaVersion); err != nil {
 			errs = errors.Join(errs, fmt.Errorf("field `minimumLimaVersion` must be a semvar value, got %q: %w", *y.MinimumLimaVersion, err))
 		}
-		limaVersion, err := versionutil.Parse(version.Version)
-		if err != nil {
-			errs = errors.Join(errs, fmt.Errorf("can't parse builtin Lima version %q: %w", version.Version, err))
-		}
-		if limaVersion != nil && versionutil.GreaterThan(*y.MinimumLimaVersion, limaVersion.String()) {
-			errs = errors.Join(errs, fmt.Errorf("template requires Lima version %q; this is only %q", *y.MinimumLimaVersion, limaVersion.String()))
+		// Unparsable versions (like commit hashes or "<unknown>") are treated as "latest/greatest"
```
jandubois: It was not immediately clear to me if this is referring to `version.Version` or `*y.MinimumLimaVersion`, so this would have helped me:

```suggestion
		// Unparsable version.Version (like commit hashes or "<unknown>") is treated as "latest/greatest"
```

pkg/limayaml/validate_test.go
```diff
@@ -38,19 +38,19 @@ func TestValidateMinimumLimaVersion(t *testing.T) {
...
 			minimumLimaVersion: "1.1.2",
-			wantErr:            `template requires Lima version "1.1.2"; this is only "1.1.1"`,
+			wantErr:            `template requires Lima version "1.1.2"; this is only "1.1.1-114-g5bf5e513"`,
 		},
 		{
 			name:               "invalid current version",
 			currentVersion:     "<unknown>",
 			minimumLimaVersion: "0.8.0",
-			wantErr:            `can't parse builtin Lima version "<unknown>": <unknown> is not in dotted-tri format`,
+			wantErr:            "", // Now passes - unparsable versions are treated as "latest"
```
jandubois: The "Now passes" comment does not make sense outside the context of this PR.

```suggestion
			wantErr:            "", // Unparsable versions are treated as "latest".
```

pkg/version/versionutil/versionutil.go
```diff
@@ -31,7 +31,12 @@ func compare(limaVersion, oldVersion string) int {
 	if err != nil {
 		return 1
 	}
-	cmp := version.Compare(*semver.New(oldVersion))
+	// Handle Unparsable oldVersion gracefully - treat as 0.0.0 so Unparsable limaVersion is always greater
+	oldVer, err := semver.NewVersion(oldVersion)
```
jandubois, line 35: This change in behaviour should be covered by a test in `versionsutil_test.go`.

jandubois (changes requested):
Thanks, LGTM

However, some comment changes and an additional test should be added.

Also might be a good idea to update all the GitHub actions that do a checkout with `check-depth: 0` (see https://github.com/lima-vm/lima/issues/3495#issuecomment-2852227944). That should not be necessary after this change, otherwise there is something wrong with it.

jandubois:
Please resolve merge conflicts by rebasing on latest `master`

pkg/version/versionutil/versionutil_test.go
```diff
@@ -41,3 +41,16 @@ func TestParse(t *testing.T) {
...
+	assert.Equal(t, GreaterThan("1.0.0", "invalid-version"), true)
+	assert.Equal(t, GreaterThan("0.1.0", "<unknown>"), true)
+	assert.Equal(t, GreaterThan("0.0.1", "commit-hash-abc123"), true)
+
+	assert.Equal(t, GreaterEqual("1.0.0", "invalid-version"), true)
+	assert.Equal(t, GreaterEqual("0.1.0", "<unknown>"), true)
+	assert.Equal(t, GreaterEqual("0.0.1", "commit-hash-abc123"), true)
+
+	assert.Equal(t, GreaterThan("invalid-lima-version", "1.0.0"), true)
+	assert.Equal(t, GreaterEqual("invalid-lima-version", "1.0.0"), true)
```
jandubois, line 55: Maybe add a test where both old version and min version are invalid?

jandubois (changes requested):
Looks like the colima test is now failing: https://github.com/lima-vm/lima/actions/runs/17002343354/job/48209754473?pr=3820

```
 + colima start
  time="2025-08-16T03:19:41Z" level=fatal msg="lima compatibility error: invalid semver version for Lima: d65ae91 is not in dotted-tri format"
```

olamilekan000:
> Looks like the colima test is now failing: https://github.com/lima-vm/lima/actions/runs/17002343354/job/48209754473?pr=3820
> 
> ```
>  + colima start
>   time="2025-08-16T03:19:41Z" level=fatal msg="lima compatibility error: invalid semver version for Lima: d65ae91 is not in dotted-tri format"
> ```

I think it might be because it is because of this.
```      with:
        ref: ${{ github.event.pull_request.head.sha }}
```
Let me remove it and see what happens.

olamilekan000:
> Thanks, LGTM
> 
> I see we have a bunch of `checkout` actions that explicitly specify `fetch-depth: 1`, which is redundant because it is the default. I wonder if those were added to explicitly show that we don't need the tags? I feel like they could all be removed, but should probably be done in a separate PR.

Yes, I had the same thought too but I didn't want to touch them since the linked issue explicitly mentioned just  `fetch-depth: 0`

jandubois:
> I think it might be because it is because of this.

No, it is not. It is a check in `colima` itself that fails. It will only work with a `limactl` with a proper version: https://github.com/abiosoft/colima/blob/da28a8334746baa56361c05d4c0c2c1e1b2966b3/cmd/start.go#L88-L89

So for the `colima` test step we need to build with `fetch-depth: 0`. Add a comment to the step explaining why this is still required.

jandubois (approved):

PR #3824, opened 2025-08-12
Look for both service names also for openrc

afbjorklund:
While alpine calls it "sshd", in debian it is called "ssh".
So look for both, if running Debian with OpenRC instead...


jandubois (approved):
Thanks, LGTM

Ideally the commit (and PR) message should mention `ssh` to be more descriptive

Issue #3749, opened 2025-07-18
Describe the different major modes of Lima operation


afbjorklund:
Some users that are new to Lima, are surprised that it mirrors the host user and mounts the home directory.

Explain the "machine" way-of-working, like in Docker Machine and Podman Machine, that Lima implements.

Also mention that there is also a plain mode, where those features are disabled and it is more of a "dumb" VM.

Users that are coming from a Vagrant background, have some different expectations on what Lima should do.

afbjorklund:
Example comments:

"To me it's somewhat surprising as a user that Lima auto-mounts anything at all, and I always have to go digging through the docs to remind myself how it works. As far as the docs go, it's not mentioned on the [filesystem mounts](https://lima-vm.io/docs/config/mount/) page; the only reference I can find is under [examples](https://lima-vm.io/docs/examples/)."

"Slightly off-topic but why is the user's home directory automounted? I was a little taken aback by this. I'd rather Lima used a secure-out-of-the-box approach and would not expose the contents of $HOME to the VM by default."

https://youtu.be/R2wGui_NHPA "Pay no attention to the VM behind the curtain!"

* https://github.com/lima-vm/lima/issues/3648

* https://github.com/lima-vm/lima/discussions/1749

afbjorklund:
Currently there is no built-in feature to disable copying the user information into the cloud-config, instead of using the default name from the cloud-init image (or something generic like "lima"). Maybe that should also be optional, eventually.

i.e. exposing the value of $USER to the VM

/etc/cloud/cloud.cfg
```yaml
# A set of users which may be applied and/or used by various modules
# when a 'default' entry is found it will reference the 'default_user'
# from the distro configuration specified below
users:
  - default
```

```yaml
# System and/or distro specific settings
# (not accessible to handlers/transforms)
system_info:
  # This will affect which distro class gets used
  distro: ubuntu
  # Default user name + that default users groups (if added/used)
  default_user:
    name: ubuntu
    lock_passwd: True
    gecos: Ubuntu
    groups: [adm, cdrom, dip, lxd, sudo]
    sudo: ["ALL=(ALL) NOPASSWD:ALL"]
    shell: /bin/bash
```

Not knowing the `uid` and `homedir` up front could be a bit of a pain

Then again, the `name` (for ssh) isn't initially known by the host either...

----

There is a top-level `ssh_authorized_keys` for the `default_user` account.

[...]

Issue #1513, opened 2023-04-27
virtiofs: file ownerships dynamically changes depending on the caller's UID


AkihiroSuda:
```console
suda@lima-vz:~$ whoami
suda

suda@lima-vz:~$ ls -ld /Users/suda/
drwxr-xr-x 106 suda suda 3392 Apr 27 05:04 /Users/suda/

suda@lima-vz:~$ sudo ls -ld /Users/suda/
drwxr-xr-x 106 root root 3392 Apr 27 05:04 /Users/suda/

suda@lima-vz:~$ sudo -u foo ls -ld /Users/suda/
drwxr-xr-x 106 foo foo 3392 Apr 27 05:04 /Users/suda/
```

Lima v0.15.1, macOS 13.3.1 (Intel)

Probably Apple intentionally designed this behavior, but I can't find a documentation for this.

PR #3831, opened 2025-08-14
describe major modes of lima operation

olamilekan000:
```
change explains how the Lima operation mode works.
```

fixes [issue](https://github.com/lima-vm/lima/issues/3749)


jandubois:
No time for a review right now, but I dislike the term "machine mode". I know it comes from the issue that this PR addresses, but most people will not make the connection to "docker-machine" and "podman-machine", and even then might not recognize the implications.

I would suggest that we talk about "integrated mode" (default) vs. "plain mode".

"Integrated" is kind of a long word, so I was looking for alternatives, like "blended mode", "fusion mode", "hybrid mode", "seamless mode", but in the end I think "integrated mode" is the most descriptive one.

What do you think @afbjorklund?

afbjorklund:
The machine reference was mostly internal, we can call it something else in the user documentation...

The individual machines created are usually just referred to as "the instances", in Lima. Perhaps even "THE instance", since having more than one is an advanced topic (like the plain mode)

olamilekan000:
> The machine reference was mostly internal, we can call it something else in the user documentation...
> 
> The individual machines created are usually just referred to as "the instances", in Lima. Perhaps even "THE instance", since having more than one is an advanced topic (like the plain mode)

does integrated or instance mode work?

jandubois:
> does integrated or instance mode work?

Doesn't work for me because "instance" is just a synonym for "machine", you can have "integrated" or "plain" instances.

Lima already means "Linux Machine", so "Lima machine" doesn't make much sense (like "ATM machine").

We use the terms "template" for just the spec of a machine, and "instance" for an actual implementation that can be run.

So I still think "integrated" (default) and "plain" are the best choices so far.

And no, we are not going to make "plain" the default. As "The Highlander" says: "There can be only one!", and "integrated" was here first. 😛

afbjorklund:
We could just call it "default mode", if it is hard to come up with a name.

jandubois:
> We could just call it "default mode", if it is hard to come up with a name.

"default mode" is not descriptive (how is it different from "plain mode").

I think we should just go with "integrated mode" even though it is a mouthful.

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+- No automatic container runtime setup
+- You must manually install and configure container engines
+
+#### 4. No Guest Agent
+- The Lima guest agent does not run
+- No automatic host-guest coordination
+- Reduced background processes in the guest
+
+#### 5. Minimal Host Integration
+- User information may still be mirrored for SSH access
```
afbjorklund, line 125: This is true for now, in the future we might want to add a feature to only use a local user (like "ubuntu", or "lima")

It is a separate feature though, and if it is implemented later on we can just modify this sentence accordingly...

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+## Comparison: Integrated Mode vs Plain Mode
+
+| Feature | Integrated Mode | Plain Mode |
+|---------|--------------|------------|
+| **Home Directory Mount** | ✅ Automatic (`~` read-only) | ❌ None |
+| **Shared Directory** | ✅ `/tmp/lima` writable | ❌ None |
+| **Port Forwarding** | ✅ Automatic | ❌ None |
+| **Container Engine** | ✅ containerd by default | ❌ None |
+| **Guest Agent** | ✅ Running | ❌ Disabled |
+| **User Mirroring** | ✅ Full mirroring | ✅ Basic (for SSH) |
```
afbjorklund, line 169: This does look a bit "odd" in the column, we don't actually need the user account in plain mode do we?

* https://github.com/lima-vm/lima/issues/3844

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+- **Backup Strategy**: Plan for data persistence since there are no automatic mounts
+- **Network Planning**: Design your network access and port forwarding explicitly
+
+### For Both Modes
+- **Resource Monitoring**: Monitor CPU, memory, and disk usage
+- **Regular Maintenance**: Keep the guest OS and packages updated
+- **Configuration Management**: Version control your Lima YAML configurations
+- **Testing**: Test your setup in both development and production-like scenarios
+
+## Troubleshooting Common Issues
```
AkihiroSuda, line 270: FAQs should be consolidated into https://lima-vm.io/docs/faq/ ?

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+- [FAQ](../faq/) - Frequently asked questions
+
+## Summary
+
+Lima's operation modes provide flexibility for different use cases:
+
+- **Integrated Mode (Default)**: Best for development workflows with seamless host-guest integration
+- **Plain Mode**: Best for traditional VM usage with explicit control and better isolation
+
+Choose the mode that best fits your security requirements, workflow preferences, and integration needs. You can always switch between modes by modifying your Lima configuration file.
```
AkihiroSuda, line 308: > You can always switch between modes by modifying your Lima configuration file.

I don't think you can switch an integrated mode into a plain mode in this way

AkihiroSuda: Did ChatGPT or something write this section?

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+- **Explicit Configuration**: Document all manual configurations you make
+- **Security Review**: Regularly audit what access the VM has to host resources
+- **Backup Strategy**: Plan for data persistence since there are no automatic mounts
+- **Network Planning**: Design your network access and port forwarding explicitly
+
+### For Both Modes
+- **Resource Monitoring**: Monitor CPU, memory, and disk usage
+- **Regular Maintenance**: Keep the guest OS and packages updated
+- **Configuration Management**: Version control your Lima YAML configurations
+- **Testing**: Test your setup in both development and production-like scenarios
```
AkihiroSuda, line 268: Too verbose and sounds like out of the topic of the "operation modes".

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+# Port forwarding will be automatic
+```
+
+## Best Practices
+
+### For Integrated Mode
+- **Review Mounted Directories**: Be aware of what host directories are accessible
+- **Use Read-Only Mounts**: Keep the home directory read-only unless you need to write
+- **Monitor Port Forwarding**: Be aware of which ports are being forwarded
+- **Regular Updates**: Keep the guest agent and container engines updated
```
AkihiroSuda, line 256: The guest agent and nerdctl are automatically kept updated

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+
+### For Plain Mode
+- **Explicit Configuration**: Document all manual configurations you make
+- **Security Review**: Regularly audit what access the VM has to host resources
```
AkihiroSuda, line 260: This rather sounds like a practice for integrated mode

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+
+This document explains both modes in detail to help you understand Lima's default behavior and choose the right mode for your needs.
+
+## Integrated Mode (Default)
+
+By default, Lima operates in "integrated mode," which is designed to provide seamless integration between the host and guest systems. This mode prioritizes convenience and developer productivity by automatically setting up various integrations.
+
+### Key Features of Integrated Mode
+
+#### 1. User Mirroring
```
AkihiroSuda, line 23: This is not specific to integrated mode

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+
+#### 3. Port Forwarding
+Lima automatically forwards ports from the guest to the host:
+
+- Published ports in containers are automatically forwarded to `localhost` on the host
+- SSH access is available on a local port (typically 60022 for the "default" instance)
+- Network services running in the guest become accessible from the host
+
+#### 4. Container Engine Integration
+Lima can automatically install and configure container engines:
```
AkihiroSuda, line 52: Link https://lima-vm.io/docs/examples/

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+
+- **containerd**: Installed by default in user mode (rootless)
```
AkihiroSuda, line 54: "user mode" sounds like the antonym of the "kernel mode"

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+- **Docker/Podman**: Available through templates
+- **Kubernetes**: Can be configured through templates like k3s, k8s
+
+#### 5. Guest Agent Services
+A guest agent runs inside the VM to:
+- Manage port forwarding
+- Handle file system mounts
+- Coordinate host-guest integration
```
AkihiroSuda, line 62: What integration?

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+- Provide status information
```
AkihiroSuda, line 63: What status?

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+
+### When to Use Integrated Mode
+
+Integrated mode is ideal for:
+- **Container Development**: When you need Docker/Podman/containerd with host integration
+- **Development Workflows**: When you want to edit files on the host but run them in Linux
+- **Cross-platform Development**: When developing Linux applications on macOS/Windows
+- **CI/CD Testing**: When you need a Linux environment that integrates with host tools
```
AkihiroSuda, line 71: I'd rather use the plain mode for CI to run "normal" Ubuntu that is usually seen on non-Lima environments.

See also https://lima-vm.io/docs/examples/gha/#plain-mode

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+- Handle file system mounts
+- Coordinate host-guest integration
+- Provide status information
+
+### When to Use Integrated Mode
+
+Integrated mode is ideal for:
+- **Container Development**: When you need Docker/Podman/containerd with host integration
+- **Development Workflows**: When you want to edit files on the host but run them in Linux
+- **Cross-platform Development**: When developing Linux applications on macOS/Windows
```
AkihiroSuda, line 70: Not specific to integrated mode

website/content/en/docs/usage/operation-modes.md
AkihiroSuda, line 169: What is "Full mirroring" ?

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+
+| Feature | Integrated Mode | Plain Mode |
+|---------|--------------|------------|
+| **Home Directory Mount** | ✅ Automatic (`~` read-only) | ❌ None |
+| **Shared Directory** | ✅ `/tmp/lima` writable | ❌ None |
+| **Port Forwarding** | ✅ Automatic | ❌ None |
+| **Container Engine** | ✅ containerd by default | ❌ None |
+| **Guest Agent** | ✅ Running | ❌ Disabled |
+| **User Mirroring** | ✅ Full mirroring | ✅ Basic (for SSH) |
+| **Host Integration** | ✅ Seamless | ❌ Minimal |
```
AkihiroSuda, line 170: What is "Host Integration"?

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+| **Convenience** | ✅ High | ⚠️ Manual setup required |
+
+## Understanding the Default Behavior
+
+### Why Does Lima Auto-Mount the Home Directory?
+
+Lima's default behavior of mounting the home directory is designed to provide seamless integration similar to Docker Machine or Podman Machine. This design choice:
+
+1. **Enables Seamless Development**: You can edit files on your host with your preferred tools and run them in the Linux environment
+2. **Maintains File Ownership**: The UID mirroring ensures files have correct permissions
```
AkihiroSuda, line 181: The UID mirroring is kinda broken at least on Apple virtiofs
- https://github.com/lima-vm/lima/issues/1513

website/content/en/docs/usage/operation-modes.md
jandubois: My money is on Claude Sonnet. 😄 (I recognize the style because I just used it last week to write a larger set of reference docs, and both the formatting and document structure closely resembles what it generated for me).

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+3. **Reduces Context Switching**: No need to copy files back and forth between host and guest
```
AkihiroSuda, line 182: "Context Switching" may sound like talking about kernel/user context switches

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+4. **Supports Container Workflows**: Container images can access host files for development
```
AkihiroSuda, line 183: I don't get it

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+
+### "Why can't I access my files?"
+In plain mode, no directories are automatically mounted. You need to:
+- Add explicit mounts in your configuration
+- Or copy files manually using `limactl copy`
+
+### "Ports aren't being forwarded"
+In plain mode, port forwarding is disabled. You need to:
+- Add explicit `portForwards` configuration
+- Or use plain networking and access the VM's IP directly
```
AkihiroSuda, line 285: What is "plain networking"?

website/content/en/docs/usage/operation-modes.md
AkihiroSuda, line 1: Probably this should be rather under https://lima-vm.io/docs/config/ ?

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+### Key Features of Plain Mode
+
+#### 1. No Automatic Mounts
+- No home directory mounting
+- No shared directories
+- You must explicitly configure any mounts you need
+
+#### 2. No Port Forwarding
+- Ports are not automatically forwarded from guest to host
+- You must manually configure any port forwarding rules
```
AkihiroSuda, line 111: This is not merged yet
- https://github.com/lima-vm/lima/pull/3699

website/content/en/docs/usage/operation-modes.md
```diff
@@ -0,0 +1,308 @@
...
+- Network access requires explicit configuration
```
AkihiroSuda, line 112: What network access ?

website/content/en/docs/usage/operation-modes.md
afbjorklund: Possibly related to mounting, but could need some clarification

website/content/en/docs/usage/operation-modes.md
afbjorklund: As far as I know, this goes for Docker and Podman as well

PR #3835, opened 2025-08-14
For IsNFS check all parent directories up to the root until one exist

jandubois:
The current implementation fails if multiple parent directories don't yet exist:

```console
> LIMA_HOME=/path/does/not/exist /limactl ls
FATA[0000] no such file or directory
```


pkg/fsutil/fsutil_linux.go
```diff
@@ -4,10 +4,28 @@
...
+	"path/filepath"
+
 	"golang.org/x/sys/unix"
 )
 
+// IsNFS checks if the path is on NFS. If the path does not exist yet, it will walk
+// up parent directories until one exists, or it hits the root of the volume.
+// Any other stat errors will cause IsNFS to fail.
 func IsNFS(path string) (bool, error) {
+	for len(path) > len(filepath.VolumeName(path))+1 {
```
AkihiroSuda: `VolumeName` doesn't seem meaningful on `_linux.go`

https://pkg.go.dev/path/filepath#VolumeName

```
VolumeName returns leading volume name.
Given "C:\foo\bar" it returns "C:" on Windows.
Given "\\host\share\foo" it returns "\\host\share".
On other platforms it returns "".
```

jandubois: It is not. For some reason I thought I had to make the code compatible with Windows.

pkg/fsutil/fsutil_linux.go
```diff
@@ -4,10 +4,28 @@
...
 import (
+	"errors"
+	"os"
+	"path/filepath"
+
 	"golang.org/x/sys/unix"
 )
 
+// IsNFS checks if the path is on NFS. If the path does not exist yet, it will walk
+// up parent directories until one exists, or it hits the root of the volume.
```
AkihiroSuda: What is "the root of the volume" ? Did you mean a mountpoint? The code doesn't seem to implement that though.

jandubois: Forgot to update the comment when I removed `VolumeName`. It loops until the dir is `/` or `.`. Updated the comment.

PR #3839, opened 2025-08-15
Apple Container external driver for macOS

afbjorklund:
A new macOS driver similar to the WSL2 driver for Windows,
creating virtual machines from container images (as rootfs).

Issue #3788

`make native && ADDITIONAL_DRIVERS=ac make additional-drivers`

(this is a draft, until the driver framework has been finalized)

----

* https://github.com/lima-vm/lima/discussions/3829

PR #3840, opened 2025-08-15
Kata/Docker Container external driver for Linux

afbjorklund:
A new Linux driver similar to the WSL2 driver for Windows,
creating virtual machines from container images (as rootfs).

Depends-On #3839

`make native && ADDITIONAL_DRIVERS=dc make additional-drivers`

(this is a draft, until the driver framework has been finalized)

----

* https://github.com/lima-vm/lima/discussions/3829

Intended for use with [Kata containers](https://katacontainers.io/) or similar runtime

Not related with `template://docker`, inside the instance.

### DIFF
--- a/pkg/driver/dc/dc_driver_linux.go
+++ b/pkg/driver/dc/dc_driver_linux.go
@@ -1,0 +1,336 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package dc
+
+import (
+	"context"
+	"embed"
+	"errors"
+	"fmt"
+	"net"
+	"regexp"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/lima-vm/lima/v2/pkg/driver"
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/lima-vm/lima/v2/pkg/ptr"
+	"github.com/lima-vm/lima/v2/pkg/reflectutil"
+)
+
+var knownYamlProperties = []string{
+	"Arch",
+	"Containerd",
+	"CopyToHost",
+	"CPUType",
+	"Disk",
+	"DNS",
+	"Env",
+	"HostResolver",
+	"Images",
+	"Message",
+	"Mounts",
+	"MountType",
+	"Param",
+	"Plain",
+	"PortForwards",
+	"Probes",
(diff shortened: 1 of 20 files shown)


pkg/driver/dc/dc_driver_linux.go
```diff
@@ -0,0 +1,243 @@
...
+func (l *LimaDcDriver) GuestAgentConn(_ context.Context) (net.Conn, string, error) {
+	return nil, "", nil
+}
+
+func (l *LimaDcDriver) Info() driver.Info {
+	var info driver.Info
+	if l.Instance != nil {
+		info.InstanceDir = l.Instance.Dir
+	}
+	info.DriverName = "dc"
```
AkihiroSuda: Why not call it "docker"?

AkihiroSuda: Anyway there must be a clear documentation that clarifies how this differs from `template://docker`

afbjorklund: An alternative name for it was Kata Containers, but it was hampered a bit about it (kata) being broken...

Currently there is no way to select the runtime, between runc/runv/runq etc, it is a hardcoded string
But the _goal_ for it was to default to a VM runtime - since it is trying to behave like the Apple Containers

pkg/store/instance_linux.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/lima-vm/lima/v2/pkg/limayaml"
+)
+
+func inspectStatus(instDir string, inst *Instance, y *limayaml.LimaYAML) {
+	if inst.VMType == limayaml.DC {
```
afbjorklund: This code will be moved to the driver

* https://github.com/lima-vm/lima/pull/3770

pkg/driver/dc/dc_driver_linux.go
AkihiroSuda: Alternative names:
- in-docker
- docker-as-vmdriver

jandubois: I think just `docker` is fine because of the context where it will be used: `vmType: docker` or `--vm-type docker`.

afbjorklund:
I will close this for now, and reopen it later.

PR #3842, opened 2025-08-15
Add fallback for busybox hostname with wsl2

afbjorklund:
Fix the `hostname` command, instead of switching to the built-in port forwarding.

Problem is that every WSL distribution will otherwise report the IP as `127.0.0.1:22`.

Reverts:
* #3299

Fixes:
* #3242

---

https://learn.microsoft.com/en-us/windows/wsl/networking

Note that `hostname -i` returns 127.0.1.1, on Ubuntu/Alpine

### DIFF
--- a/pkg/store/instance_windows.go
+++ b/pkg/store/instance_windows.go
@@ -121,11 +121,26 @@ func GetWslStatus(instName string) (string, error) {
 // Expected output (whitespace preserved, [] for optional):
 // PS > wsl -d <distroName> bash -c hostname -I | cut -d' ' -f1
 // 168.1.1.1 [10.0.0.1]
+// But busybox hostname does not implement --all-ip-addresses:
+// hostname: unrecognized option: I
 func GetSSHAddress(instName string) (string, error) {
 	distroName := "lima-" + instName
+	// Ubuntu
 	cmd := exec.Command("wsl.exe", "-d", distroName, "bash", "-c", `hostname -I | cut -d ' ' -f1`)
 	out, err := cmd.CombinedOutput()
-	if err != nil {
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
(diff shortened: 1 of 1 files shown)


afbjorklund:
We don't actually have any rootfs images based on Alpine, but it can be tested in a docker container I suppose.

```
docker run -it ubuntu bash -c "hostname -I | cut -d ' ' -f1"
```

```
docker run -it alpine sh -c "ip route get 1 | awk '{gsub(\"^.*src \",\"\"); print $1; exit}'"
```

Unfortunately WSL doesn't provide any _sane_ method, to query the IP of the "distro" from the outside (host)...

https://learn.microsoft.com/en-us/windows/wsl/networking

Compare: `docker inspect --format='{{ .NetworkSettings.IPAddress }}'`

pkg/store/instance_windows.go
```diff
@@ -115,6 +116,33 @@ func GetWslStatus(instName string) (string, error) {
...
+//
+// Expected output (whitespace preserved, [] for optional):
+// PS > wsl -d <distroName> bash -c hostname -I | cut -d' ' -f1
+// 168.1.1.1 [10.0.0.1]
+// But busybox hostname does not implement --all-ip-addresses:
+// hostname: unrecognized option: I
+func GetSSHAddress(instName string) (string, error) {
+	distroName := "lima-" + instName
+	// Ubuntu
+	cmd := exec.Command("wsl.exe", "-d", distroName, "bash", "-c", `hostname -I | cut -d ' ' -f1`)
```
AkihiroSuda:
```suggestion
	cmd := exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "bash", "-c", `hostname -I | cut -d ' ' -f1`)
```
w/ `context.TODO()` for now

PR #3843, opened 2025-08-16
Get rid of `fetch-depth: 1`

jandubois:
It is redundant because it is already the default.

I suspect that we used it because in many places we needed to specify `fetch-depth: 0` to get all the tags. But with one exception for the colima test, this is no longer the case, so I think we should get rid of all of these.

See also: https://github.com/lima-vm/lima/pull/3820#issuecomment-3193348613

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,8 +36,6 @@ jobs:
         # macOS SDK version
         xcrun --show-sdk-version || true
     - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8  # v5.0.0
-      with:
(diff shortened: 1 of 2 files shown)


afbjorklund (approved):
Usually it is just an optimization for huge repos, but I don't think it (shallow clone) is needed here.

jandubois:
> Usually it is just an optimization for huge repos, but I don't think it (shallow clone) is needed here.

Shallow clone is already the default. This is a no-op change, just removing the explicit setting of the default value.

```yaml
    # Number of commits to fetch. 0 indicates all history for all branches and tags.
    # Default: 1
    fetch-depth: ''
```

Issue #2183, opened 2024-02-04
`limactl start` should optionally show the progress of the provision scripts


AkihiroSuda:

afbjorklund:
As long as cloud-init is running them, there is the cloud-init log?

jandubois:
How is this supposed to be implemented? By tailing `/var/log/cloud-init-output.log` to the terminal over ssh?

AkihiroSuda:
> How is this supposed to be implemented? By tailing `/var/log/cloud-init-output.log` to the terminal over ssh?

Yes, probably

olamilekan000:
regarding the title, when you say "optionally show the progress..." does this mean the user can opt in to see the log via a flag on the start command? @AkihiroSuda

jandubois:
> does this mean the user can opt in to see the log via a flag on the start command?

Yes

We need to discuss if this can just piggy-back on the `--debug` option, or if we need an extra one for this.

olamilekan000:
I think the `--debug` flag works since it is used throughout the codebase for verbose logging

olamilekan000:
@jandubois can you assign this to me. Also, are we agreeing to go with the `--debug` flag?

AkihiroSuda:
I'd prefer to have a new --progress flag, as --debug is too verbose

olamilekan000:
okay, noted.

PR #3846, opened 2025-08-17
add progress tracking flag to monitor VM startup state using cloud-init

olamilekan000:
```
change implements progress tracking for monitoring VM state.
```

implements [issue](https://github.com/lima-vm/lima/issues/2183#issuecomment-3185972062)

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -4,6 +4,7 @@
 package hostagent
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
@@ -72,11 +73,14 @@ type HostAgent struct {
 
 	guestAgentAliveCh     chan struct{} // closed on establishing the connection
 	guestAgentAliveChOnce sync.Once
+
+	showProgress bool // whether to show cloud-init progress
 }
 
 type options struct {
 	guestAgentBinary string
 	nerdctlArchive   string // local path, not URL
+	showProgress     bool
 }
 
 type Opt func(*options) error
@@ -91,6 +95,13 @@ func WithGuestAgentBinary(s string) Opt {
 func WithNerdctlArchive(s string) Opt {
 	return func(o *options) error {
 		o.nerdctlArchive = s
+		return nil
+	}
+}
+
+func WithCloudInitProgress(enabled bool) Opt {
+	return func(o *options) error {
+		o.showProgress = enabled
 		return nil
 	}
 }
@@ -214,6 +225,7 @@ func New(instName string, stdout io.Writer, signalCh chan os.Signal, opts ...Opt
 		vSockPort:         vSockPort,
 		virtioPort:        virtioPort,
 		guestAgentAliveCh: make(chan struct{}),
+		showProgress:      o.showProgress,
 	}
 	return a, nil
 }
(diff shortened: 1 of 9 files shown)


pkg/hostagent/hostagent.go
```diff
@@ -777,6 +801,141 @@ func forwardSSH(ctx context.Context, sshConfig *ssh.SSHConfig, port int, local,
...
+
+	if err := cmd.Start(); err != nil {
+		logrus.WithError(err).Warn("Failed to start cloud-init monitoring command")
+		return
+	}
+
+	scanner := bufio.NewScanner(stdout)
+	cloudInitFinished := false
+
+	for scanner.Scan() {
```
norio-nomura, line 866: This loop appears to run indefinitely.
The `ssh -F … sudo tail -n +1 -f /var/log/cloud-init-output.log` process continues after `limactl start --progress` finishes.
Also, some times `limactl start --progress` never stop even if `Cloud-init v. 25.1.2-0ubuntu0~24.04.1 finished` received.

olamilekan000: I have tried to reproduce this but couldn't as everything works fine on my end. Although, I think a timeout context can be added for safety purpose. Can you share how you were able to find the bug?

olamilekan000: I have opened a ticket for the timeout and also implemented the fix. You can check [here](https://github.com/lima-vm/lima/issues/3903) for it.

norio-nomura: > Can you share how you were able to find the bug?

It can be reproduced on my Mac (Apple M2 Pro) with the following:
```console
limactl start template://ubuntu-lts --name=default --progress --set .containerd.user=false
```
built on current HEAD
```console
$ limactl --version
limactl version 1.2.0-193-gd4f4911e
```

pkg/instance/start.go
```diff
@@ -330,6 +353,12 @@ func watchHostAgentEvents(ctx context.Context, inst *store.Instance, haStdoutPat
 				err = xerr
 				return true
 			}
+
+			if showProgress && !cloudInitCompleted {
```
norio-nomura, line 357: Received `ev.Status.Running=true` event may be ignored here.
It should add checking `receivedRunningEvent && cloudInitCompleted` like following:
```golang
} else if ev.Status.Running {
    receivedRunningEvent = true
    ...
} else if receivedRunningEvent && cloudInitCompleted {
    ...
    return true
}
```

norio-nomura:
Found another issue.
`/var/log/cloud-init-output.log` contains the previous boot session's log.
So, it needs to cut the log from the beginning of the latest boot session.

norio-nomura:
border of the session in `/var/log/cloud-init-output.log`:
```log
LIMA 2025-08-25T15:04:03+09:00| Exiting with code 0
Cloud-init v. 25.1.4-0ubuntu0~24.04.1 finished at Mon, 25 Aug 2025 06:04:03 +0000. Datasource DataSourceNoCloud [seed=/dev/vdb].  Up 5.37 seconds
Cloud-init v. 25.1.4-0ubuntu0~24.04.1 running 'init-local' at Mon, 25 Aug 2025 06:07:19 +0000. Up 1.00 seconds.
Cloud-init v. 25.1.4-0ubuntu0~24.04.1 running 'init' at Mon, 25 Aug 2025 06:07:21 +0000. Up 2.59 seconds.
```

norio-nomura:
How can we detect the beginning of the current boot session in the log?

norio-nomura:
filed #3907

PR #3859, opened 2025-08-18
Fix: Server Shutdown Bug

unsuman:
The recently merged PR #3836 introduced a 60-second timeout to shut down the gRPC server if no client connects. However, it has a bug: the timer is not cancelled once a client successfully connects. This causes the server to forcefully shut down after 60 seconds, regardless of whether a client has connected.


AkihiroSuda (approved):
Thanks

Issue #3769, opened 2025-07-25
# Refactor: Lima Config and External Drivers


unsuman:
# Lima Config and External Drivers

This document tries to catalog all the configuration challenges related to supporting external virtualization drivers in Lima.

It is not really well thought through, but more like a first draft to start the discussion.

## Status Quo

Right now all drivers are built-in (`qemu`, `vz`, `wsl2`), and driver-specific configuration is mixed in with generic configuration in various ways.

### Properties that are specific to a driver

These are properties used only by a single driver. They should be moved under `vmOpts`.

#### `cpuTypes`

Specific to `qemu` driver.

#### `rosetta`

Specific to `vz` driver.

#### `video.display`

Seems like a `qemu` specific setting. Not clear if `video.vnc` is supposed to work with other drivers.

#### `hostResolver`

Not sure if it is specific to `qemu`, or works with other drivers too.

### Properties whose valid values depend on the driver

These properties are used by more than one driver, but not necessarily all of them. Some drivers may not support them at all.

#### `arch`

`vz` driver only allows the native arch (`limayaml/validate.do`). `wsl2` ignores value, but should also only allow native arch.

#### `mountType`

`qemu` driver on `macOS` does not support `virtiofs`. `wsl2` driver only supports `wsl2` mount type.

#### `firmware.legacyBIOS`

Default value depends on `arch` and `vmType`.

#### `audio.device`

Mostly a `qemu` setting, but I think `vz` supports `vz` and `none`.

#### `nestedVirtualization`

CPU type must be set to `host` with the `qemu` driver when using nested virtualization.

#### `cpus`, `memory`

For `wsl2` these settings can only be specified globally because all distros are really just containers in a single VM.

#### `images`

The `wsl2` driver needs tarballs of the rootfs, not regular images. There is no property to specify the type of image. We could use the file extension. Currently there is only a single template containing a `wsl2` image (`templates/experimental/wsl2.yaml`), and it only has a single `wsl2` image in the list.

#### `mounts`

Mounts can have mount-type specific sub-properties like `sshfs` and `9p`. Not all drivers support all mount types. A new driver may add a new mount type that needs additional properties under `mounts` that need their own validation logic.

Mount entries are "combined" if they use the same `mountPoint`. Some of the additional properties may have custom merge strategies. E.g. should `fsType` and `fsArgs`[^fsargs] only be merged as a combination?

[^fsargs]: `fsArgs` is currently undocumented in `default.yaml`.

#### `networks`

[...]

PR #3861, opened 2025-08-18
Remove VMType check from the lima-init boot script

afbjorklund:
Do this when setting up the cidata instead, next to the decision on whether to use "cidata/" directory or "cidata.iso".

This means that this particular boot script is not depending on the driver, but can be used from all container drivers.

* https://github.com/lima-vm/lima/pull/3839

We probably want to add an explicit `bool` config for it to the driver, instead of checking for the driver name?

And there are still some features left to add to the script (from cloud-config), like the hostname and timezone.

* #3769

Currently we dont use `user-data`, but `lima.env` (+ ssh key)

We execute this script through the `boot.sh` directly (lima-init).

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -118,6 +118,11 @@ func setupEnv(instConfigEnv map[string]string, propagateProxyEnv bool, slirpGate
 	return env, nil
 }
 
+func useCloudInit(instConfig *limayaml.LimaYAML) bool {
+	// all drivers but WSL2 use cloud-init
+	return *instConfig.VMType != limayaml.WSL2
+}
+
 func templateArgs(ctx context.Context, bootScripts bool, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort, vsockPort int, virtioPort string) (*TemplateArgs, error) {
 	if err := limayaml.Validate(instConfig, false); err != nil {
 		return nil, err
@@ -138,12 +143,13 @@ func templateArgs(ctx context.Context, bootScripts bool, instDir, name string, i
 		Containerd:         Containerd{System: *instConfig.Containerd.System, User: *instConfig.Containerd.User, Archive: archive},
 		SlirpNICName:       networks.SlirpNICName,
 
-		VMType:     *instConfig.VMType,
-		VSockPort:  vsockPort,
-		VirtioPort: virtioPort,
-		Plain:      *instConfig.Plain,
-		TimeZone:   *instConfig.TimeZone,
-		Param:      instConfig.Param,
+		VMType:      *instConfig.VMType,
+		VSockPort:   vsockPort,
+		VirtioPort:  virtioPort,
+		Plain:       *instConfig.Plain,
+		TimeZone:    *instConfig.TimeZone,
+		NoCloudInit: !useCloudInit(instConfig),
+		Param:       instConfig.Param,
 	}
 
 	if instConfig.VMOpts.VZ.Rosetta.Enabled != nil {
@@ -466,7 +472,7 @@ func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name strin
 		})
 	}
 
-	if args.VMType == limatype.WSL2 {
+	if !useCloudInit(instConfig) {
 		layout = append(layout, iso9660util.Entry{
 			Path:   "ssh_authorized_keys",
 			Reader: strings.NewReader(strings.Join(args.SSHPubKeys, "\n")),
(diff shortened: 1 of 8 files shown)


pkg/cidata/template.go
```diff
@@ -105,6 +105,7 @@ type TemplateArgs struct {
 	VirtioPort                      string
 	Plain                           bool
 	TimeZone                        string
+	LimaInit                        bool
```
AkihiroSuda: this might be more clear 
```suggestion
	NoCloudInit                        bool
```

pkg/cidata/cidata.go
```diff
@@ -142,6 +142,7 @@ func templateArgs(bootScripts bool, instDir, name string, instConfig *limayaml.L
 		VirtioPort:     virtioPort,
 		Plain:          *instConfig.Plain,
 		TimeZone:       *instConfig.TimeZone,
+		LimaInit:       *instConfig.VMType == limayaml.WSL2,
```
AkihiroSuda: Availability of cloud-init should be queried to the VM driver

pkg/cidata/cidata.TEMPLATE.d/boot/02-lima-init.sh
AkihiroSuda, line 1: the file name should be changed.
e.g. `02-no-cloudinit.sh`

pkg/cidata/cidata.TEMPLATE.d/boot/02-lima-init.sh
```diff
@@ -4,8 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 # This script replaces the cloud-init functionality of creating a user and setting its SSH keys
-# when using a WSL2 VM.
```
AkihiroSuda:
```
# when cloud-init is not available 
```

pkg/cidata/cidata.TEMPLATE.d/boot/02-lima-init.sh
```diff
@@ -4,8 +4,7 @@
...
-[ "$LIMA_CIDATA_VMTYPE" = "wsl2" ] || exit 0
+[ "$LIMA_CIDATA_LIMA_INIT" = "1" ] || exit 0
```
AkihiroSuda: maybe you can just check the existence of cloud-init here, then the variable can be just dropped

afbjorklund: It is still needed (in the Go code, that is) to know if we should create a "cidata" directory, or a cidata.iso

afbjorklund:
This should be returned by the driver, instead of hardcoded on driver name like now:

```go
func useCloudInit(instConfig *limayaml.LimaYAML) bool {
       // all drivers but WSL2 use cloud-init
       return *instConfig.VMType != limayaml.WSL2
}
```

Not so keen on the variable name, but it does follow the "default is false" convention.

`NoCloudInit`

pkg/cidata/cidata.go
```diff
@@ -116,6 +116,11 @@ func setupEnv(instConfigEnv map[string]string, propagateProxyEnv bool, slirpGate
 	return env, nil
 }
 
+func useCloudInit(instConfig *limayaml.LimaYAML) bool {
+	// all drivers but WSL2 use cloud-init
+	return *instConfig.VMType != limayaml.WSL2
```
AkihiroSuda: (Should be moved to the `DriverFeatures` struct in the driver infra)

AkihiroSuda:
Needs rebase

afbjorklund:
Now the script has been duplicated in each driver instead:

```
pkg/driver/wsl2/boot
└── 02-wsl2-setup.sh
pkg/driver/ac/boot
└── 02-ac-setup.sh
pkg/driver/dc/boot
└── 02-dc-setup.sh
```

https://github.com/lima-vm/lima/blob/master/pkg/driver/wsl2/boot/02-wsl2-setup.sh


```sh
# This script replaces the cloud-init functionality of creating a user and setting its SSH keys
# when using a WSL2 VM.
[ "$LIMA_CIDATA_VMTYPE" = "wsl2" ] || exit 0
```

```sh
# This script replaces the cloud-init functionality of creating a user and setting its SSH keys
# when using a AC VM.
[ "$LIMA_CIDATA_VMTYPE" = "ac" ] || exit 0
```

```sh
# This script replaces the cloud-init functionality of creating a user and setting its SSH keys
# when using a DC VM.
[ "$LIMA_CIDATA_VMTYPE" = "dc" ] || exit 0
```

But it still needs to be fixed in the cidata, still hardcoded:

```go
        if args.VMType == limatype.WSL2 || args.VMType == limatype.AC || args.VMType == limatype.DC {
                layout = append(layout, iso9660util.Entry{
                        Path:   "ssh_authorized_keys",
                        Reader: strings.NewReader(strings.Join(args.SSHPubKeys, "\n")),
                })
                return writeCIDataDir(filepath.Join(instDir, filenames.CIDataISODir), layout)
        }

```

AkihiroSuda (approved):
Thanks

Issue #3851, opened 2025-08-18
`limactl shell --preserve-env` should have tests and docs


AkihiroSuda:
> Needs an integration test and the document to clarify the block list

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/3830#discussion_r2281264252_

Issue #251, opened 2021-09-16
arguments of shell command doesn't support "raw"


RobberPhex:
`limactl shell` command actually take two responsibility:
1. : **lima act like ssh**,
```shell
# expected to get home dir in lima guest vm
$ limactl shell default -- echo '$HOME'
/home/robert.linux
```
Case 1 is right.

2. : **lima act like command tunnel**, in following example, as tunnel to pass arguments to nerdctl.
```
# expected to get home path in container
$ alias docker='limactl shell default -- nerdctl'
docker run -it --rm alpine:latest sh -x -c 'echo $HOME'
+ echo

```
Case 2, **the behavior of limactl is not expected**.
Comparing with docker for desktop:
```shell
$ /usr/local/bin/docker run -it --rm alpine:latest sh -x -c 'echo $HOME'
+ echo /root
/root
```

---

I propose to add `--raw` flag to make limactl support act like command tunnel, which means **what you  pass to limactl, what inner command get**.

jandubois:
> I propose to add `--raw` flag to make limactl support act like command tunnel,

Yes, I've struggled with this use case before too!

However, I think I would prefer to use a different command, e.g. `limactl run` instead of `limactl shell --raw` to provide the different semantics. Let's wait to hear what @AkihiroSuda says!

AkihiroSuda:
> > I propose to add `--raw` flag to make limactl support act like command tunnel,
> 
> Yes, I've struggled with this use case before too!
> 
> However, I think I would prefer to use a different command, e.g. `limactl run` instead of `limactl shell --raw` to provide the different semantics. Let's wait to hear what @AkihiroSuda says!

I don't think I can remember the differences between `run` and `shell` 😅 , and I'd rather prefer to reserve `run` for something similar to `start`.
So `limactl shell --raw` (or maybe `limactl shell --escape`, or `limactl shell --verbatim`) SGTM, but if it has to be a new subcommand (why?) it should be like `limactl shell-raw` / `limactl shell-escape` / `limactl shell-verbatim`.

jandubois:
I think I'm mostly thrown by the name `raw`, which to me means kind of the opposite of what it does here, namely wrapping each argument in an extra set of quotes before joining all the arguments by a space. "Raw" to me means un-processed, not wrapped/formatted.

A descriptive name (instead of `--raw`) would be `--auto-quoted`, but I admit that it is a bit unwieldy, so `--quoted` seems like a reasonable alternative.

[...]

PR #3870, opened 2025-08-20
add preserve-env variable to doc and also add an integration test

olamilekan000:
```
change adds LIMA_SHELLENV_ALLOW and LIMA_SHELLENV_BLOCK env
to the documentation and also include an integration test for the --preserve-env flag.
```
fixes [issue](https://github.com/lima-vm/lima/issues/3851)


pkg/envutil/envutil.go
```diff
@@ -25,7 +25,7 @@ var defaultBlockList = []string{
 	"LD_*",
 	"LOGNAME",
 	"OLDPWD",
-	"PATH",
+	"*PATH",
```
jandubois: I'm not convinced that this is the right thing to do. Why would you not want to import dev paths into the VM by default?

Where do we draw the line? Do we block `RUBY_*`? How about `JAVA_OPTS`? If you want to block most of the variables, you should just set `LIMA_SHELLENV_ALLOW` and have everything else blocked by default.

pkg/envutil/envutil.go
```diff
@@ -78,8 +78,15 @@ func matchesPattern(name, pattern string) bool {
 		return true
 	}
 
-	prefix, found := strings.CutSuffix(pattern, "*")
-	return found && strings.HasPrefix(name, prefix)
+	if prefix, found := strings.CutSuffix(pattern, "*"); found {
```
jandubois: I've been wondering if once we support both prefix and postfix `*` if users would expect that infix `*` should also work?

We could simply replace every `*` with `.*`, and then anchor the pattern with `^pattern$` and perform a regex search. Probably should first make sure there are no other metadata characters in the pattern.

Anyways, not saying you should change this, I just want a discussion if we should support infix `*` or not? E.g. `PERL*LIB` would match `PERL5LIB` and `PERLLIB`.

jandubois (commented):
Not really a review, just 2 observations that need discussion IMO.

pkg/envutil/envutil.go
olamilekan000: I kind of feel it makes sense blocking PATH (and similar variables) by default inorder to avoid unexpected behaviour inside the VM. If the host injects its paths, the VM might end up pointing to binaries or libraries that don’t exist there, which could cause confusing errors.

pkg/envutil/envutil.go
olamilekan000: I actually saw your comment on the parent issue and thought I should make provision for it in this PR but I wanted to see what others would say before going ahead. Your suggestion makes perfect sense.

pkg/envutil/envutil.go
jandubois: I don't know, `--preserve-env` variables may still be overwritten in `~/.profile` inside the guest:

```console
❯ lima sh -c 'tail -1 ~/.profile'
export GOPATH=/home/jan.linux/go

❯ echo $GOPATH
/Users/jan/go

❯ limactl shell --preserve-env default bash -c "printenv GOPATH"
/home/jan.linux/go

❯ limactl shell default bash -c "GOPATH=$GOPATH printenv GOPATH"
/Users/jan/go
```

So this is maybe not how people expect `--preserve-env` to work, but by passing the variables to the shell process, which will then run `~/.profile`, the actual value may still be different from the host. You have to overwrite them inside the shell script to be certain they have the host value.

Assuming we stick with the current functionality, I think there is no reason for aggressive filtering of `*PATH` beyond just `PATH`.

If we want to change how `--preserve-env` is implemented, and the host settings would take precedence over the `~/.profile` ones, then additional filtering would be warranted. But I feel in that case you probably want to stick with setting `LIMA_SHELLENV_ALLOW` instead.

I think the current PR should move ahead with the current implementation and `*PATH` removed, so at least @AkihiroSuda's comments on the closed PR are addressed, and then we can further update the feature based on the result of the discussion thread in a separate PR.

cmd/docker.lima
AkihiroSuda, line 1: Better to be a separate commit (or a PR)

hack/test-preserve-env.sh
AkihiroSuda, line 1: Doesn't seem executed on CI

pkg/envutil/envutil.go
AkihiroSuda: Should be a separate commit, or perhaps a PR

website/content/en/docs/config/environment-variables.md
```diff
@@ -57,6 +57,45 @@ This page documents the environment variables used in Lima.
   lima
   ```
 
+### `LIMA_SHELLENV_ALLOW`
```
AkihiroSuda, line 60: Probably some brief explanation should be printed in `limactl shell --help` too

website/content/en/docs/config/environment-variables.md
```diff
@@ -57,6 +57,45 @@ This page documents the environment variables used in Lima.
...
+- **Default**: A predefined list of system and shell-specific variables that should not be propagated:
+  - Shell variables: `BASH*`, `SHELL`, `SHLVL`, `ZSH*`, `ZDOTDIR`, `FPATH`
+  - System paths: `*PATH`, `PWD`, `OLDPWD`, `TMPDIR`
+  - User/system info: `HOME`, `USER`, `LOGNAME`, `UID`, `GID`, `EUID`, `GROUP`, `HOSTNAME`
+  - Display/terminal: `DISPLAY`, `TERM`, `TERMINFO`, `XAUTHORITY`, `XDG_*`
+  - SSH/security: `SSH_*`
+  - Dynamic linker: `DYLD_*`, `LD_*`
+  - Internal variables: `_*` (variables starting with underscore)
+  
+  See [`GetDefaultBlockList()`](https://github.com/lima-vm/lima/blob/master/pkg/envutil/envutil.go#L133) for the complete list.
```
AkihiroSuda, line 87: The list should be exposed to `limactl info`

website/content/en/docs/config/environment-variables.md
```diff
@@ -57,6 +57,45 @@ This page documents the environment variables used in Lima.
   lima
   ```
 
+### `LIMA_SHELLENV_ALLOW`
+
+- **Description**: Specifies a comma-separated list of environment variable patterns to allow when propagating environment variables to the Lima instance with `--preserve-env`. When set, **only** variables matching these patterns will be passed through, completely overriding the default block list behavior.
```
AkihiroSuda: > default block list

What about the custom block list?

AkihiroSuda: It should be also mentioned that this only applies to Lima v2.0 or later

pkg/envutil/envutil.go
jandubois: Should be a separate PR, together with the whole `*PATH` change. I think in this PR `envutil.go` should not be changed at all.

website/content/en/docs/config/environment-variables.md
jandubois: It already is:

```console
❯ limactl info | jq .shellEnvBlock
[
  "BASH*",
  "DISPLAY",
  "DYLD_*",
  "EUID",
  "FPATH",
  "GID",
  "GROUP",
  "HOME",
  "HOSTNAME",
  "LD_*",
  "LOGNAME",
  "OLDPWD",
  "PATH",
  "PWD",
  "SHELL",
  "SHLVL",
  "SSH_*",
  "TERM",
  "TERMINFO",
  "TMPDIR",
  "UID",
  "USER",
  "XAUTHORITY",
  "XDG_*",
  "ZDOTDIR",
  "ZSH*",
  "_*"
]
```

website/content/en/docs/config/environment-variables.md
jandubois: All block lists are ignored when you use the allow list. The docs should be updated to match.

jandubois: Isn't this is more general issue that we have one set of docs that are supposed to document the current and the upcoming release simultaneously? It would be better to have 2 separate versions, so the docs can be simpler.

pkg/envutil/envutil.go
jandubois: > So this is maybe not how people expect `--preserve-env` to work, but by passing the variables to the shell process, which will then run `~/.profile`, the actual value may still be different from the host. You have to overwrite them inside the shell script to be certain they have the host value.

After subconscious reflection (sleep) I think the current implementation is correct given that `limactl shell` is running commands in a login shell.

It matches the behaviour of `sudo` when running a login shell:

```console
$ RBENV_ROOT=foo sudo --preserve-env bash --login -c 'printenv RBENV_ROOT'
/opt/homebrew/var/rbenv
```

Therefore I think adding the `*PATH` pattern to the default blocklist is not helpful.

---

We have an ancient (from 2021) request about `limactl shell` not actually evaluating arguments as a shell inside the VM (so you can pass through wildcards and variable names without them being evaluated), and I [suggested](https://github.com/lima-vm/lima/issues/251#issuecomment-921494478) to use `limactl exec` for that functionality.

If we ever implement this alternative, then host environment variables should not be shadowed by `~/.profile` settings inside the VM, but we can think about that if we ever implement it. Nobody seems to have complained about it missing in the last few years though...

cmd/limactl/shell.go
```diff
@@ -35,6 +35,11 @@ lima command is provided as an alias for limactl shell $LIMA_INSTANCE. $LIMA_INS
 By default, the first 'ssh' executable found in the host's PATH is used to connect to the Lima instance.
 A custom ssh alias can be used instead by setting the $` + sshutil.EnvShellSSH + ` environment variable.
 
+Environment Variables:
+  --preserve-env: Propagates host environment variables to the guest instance.
+                  Use LIMA_SHELLENV_ALLOW to specify which variables to allow (whitelist).
```
jandubois: We intentionally avoid the use of whitelist/blacklist and use allowlist/blocklist as more descriptive alternatives. I think the description is clear without adding the alternate term:

```suggestion
                  Use LIMA_SHELLENV_ALLOW to specify which variables to allow.
```

pkg/envutil/envutil.go
jandubois: > We have an ancient (from 2021) request about `limactl shell` not actually evaluating arguments as a shell inside the VM (so you can pass through wildcards and variable names without them being evaluated)

I have just tested this, and it seems like `limactl shell` has changed quoting and is passing things through correctly now, so I will close the ancient issue.

```console
$ limactl shell default echo '$HOME'
$HOME
```

> Nobody seems to have complained about it missing in the last few years though...

I guess now I know why...

olamilekan000: When I tried the sample command you shared I got something a little different. The env value was actually preserved.

```
RBENV_ROOT=foo sudo --preserve-env bash --login -c 'printenv RBENV_ROOT'

foo
```
I also did try it with the `limactl shell` command and it was consistent with the result I got locally. Correct me if i'm wrong, what I understand from your earlier point is that since the shell command makes use of the login shell, the guest's value would actually take precedence but that's not the case in this scenario regarding PATH as well.

I have some samples below:
This is the value of the PATH env on my local system.
```echo $PATH
/Users/username/.opam/default/bin:/Users/username/.ops/bin:/Users/username/.gvm/bin:/Users/username/Library/Application:Support/cloud-code/installer/google-cloud-sdk/bin:/Library/Frameworks/Python.framework/Versions/3.11/bin:/usr/lib/google-cloud-sdk/bin:/opt/homebrew/opt/make/libexec/gnubin:/usr/local/sbin:/opt/homebrew/bin:/opt/homebrew/sbin:/Library/Frameworks/Python.framework/Versions/3.10/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Applications/VMware:Fusion:Tech:Preview.app/Contents/Public:/usr/local/go/bin:/Users/username/.ops/bin:/Users/username/.opam/default/bin:/Users/username/.cargo/bin:/Applications/Visual:Studio:Code.app/Contents/Resources/app/bin:/Applications/Visual Studio Code.app/Contents/Resources/app/bin

```
This is what the value looks like in the VM
```
limactl shell default bash 
username@lima-default:/Users/username/works/opensource/lima$ echo $PATH
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/sbin:/sbin:/usr/sbin:/sbin
```

and then when i use the --preserve-env  in the shell command, I get this
```
LIMA_SHELLENV_ALLOW=PATH limactl shell `--preserve-env` default bash -c 'echo "login shell PATH: $PATH"'

login shell PATH: /Users/username/.opam/default/bin:/Users/username/.ops/bin:/Users/username/.gvm/bin:/Users/username/Library/Application:Support/cloud-code/installer/google-cloud-sdk/bin:/Library/Frameworks/Python.framework/Versions/3.11/bin:/usr/lib/google-cloud-sdk/bin:/opt/homebrew/opt/make/libexec/gnubin:/usr/local/sbin:/opt/homebrew/bin:/opt/homebrew/sbin:/Library/Frameworks/Python.framework/Versions/3.10/bin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/Applications/VMware:Fusion:Tech:Preview.app/Contents/Public:/usr/local/go/bin:/Users/username/.ops/bin:/Users/username/.opam/default/bin:/Users/username/.cargo/bin:/Applications/Visual:Studio:Code.app/Contents/Resources/app/bin:/Applications/Visual Studio Code.app/Contents/Resources/app/bin

```
which is same as the PATH in the host machine. Based on the login shell expectation, I don't think the PATH value should be the same as the value on the host machine, hence it shouldn't get overwritten even when it is added to the `LIMA_SHELLENV_ALLOW` list. Is this right?

website/content/en/docs/config/environment-variables.md
```diff
@@ -57,6 +57,45 @@ This page documents the environment variables used in Lima.
...
   ```
 
+### `LIMA_SHELLENV_ALLOW`
+
+- **Description**: Specifies a comma-separated list of environment variable patterns to allow when propagating environment variables to the Lima instance with `--preserve-env`. When set, **only** variables matching these patterns will be passed through, completely overriding the default block list behavior. This feature only applies to Lima v2.0.0 or later.
+- **Default**: unset (when using `--preserve-env`, all variables are propagated except those matching the block list patterns)
+- **Usage**:
+  ```sh
+  export LIMA_SHELLENV_ALLOW="HOME,USER,CUSTOM_*"
+  limactl shell --preserve-env default
```
jandubois: Both `HOME` and `USER` are bad examples because they will be set by the login shell of the VM, so passing them will have no effect.

website/content/en/docs/config/environment-variables.md
```diff
@@ -57,6 +57,45 @@ This page documents the environment variables used in Lima.
...
+  ```
+- **Behavior**:
+  - **Without `--preserve-env`**: No environment variables are propagated (regardless of this setting)
+  - **With `--preserve-env` and `LIMA_SHELLENV_ALLOW` unset**: All variables are propagated except those in the block list
+  - **With `--preserve-env` and `LIMA_SHELLENV_ALLOW` set**: Only variables matching the allow patterns are propagated (block list is ignored)
+- **Note**: Patterns support wildcards using `*` at the end (e.g., `CUSTOM_*` matches `CUSTOM_VAR`, `CUSTOM_PATH`, etc.). This acts as a whitelist that completely overrides normal filtering.
```
jandubois: Don't use `whitelist`. Also this has already been stated before, so the final sentence can be removed:

```suggestion
- **Note**: Patterns support wildcards using `*` at the end (e.g., `CUSTOM_*` matches `CUSTOM_VAR`, `CUSTOM_PATH`, etc.).
```

website/content/en/docs/config/environment-variables.md
```diff
@@ -57,6 +57,45 @@ This page documents the environment variables used in Lima.
...
+
+### `LIMA_SHELLENV_BLOCK`
+
+- **Description**: Specifies a comma-separated list of environment variable patterns to block when propagating environment variables to the Lima instance with `--preserve-env`. Can either replace the default block list or extend it by prefixing with `+`. This feature only applies to Lima v2.0.0 or later.
+- **Default**: A predefined list of system and shell-specific variables that should not be propagated:
+  - Shell variables: `BASH*`, `SHELL`, `SHLVL`, `ZSH*`, `ZDOTDIR`, `FPATH`
+  - System paths: `PATH`, `PWD`, `OLDPWD`, `TMPDIR`
+  - User/system info: `HOME`, `USER`, `LOGNAME`, `UID`, `GID`, `EUID`, `GROUP`, `HOSTNAME`
+  - Display/terminal: `DISPLAY`, `TERM`, `TERMINFO`, `XAUTHORITY`, `XDG_*`
```
jandubois, line 82: Let's not touch this in this PR, but I'm wondering if `TERM` and `TERMINFO` really should be blocked. The VM shell will run in the same terminal after all.

pkg/envutil/envutil.go
jandubois: > When I tried the sample command you shared I got something a little different. The env value was actually preserved.
> 
> ```console
> RBENV_ROOT=foo sudo --preserve-env bash --login -c 'printenv RBENV_ROOT'
> 
> foo
> ```

This happens because you don't set `RBENV_ROOT` in your `~/.profile` that will overwrite your value:

```console
❯ grep RBENV ~/.bash_profile
export RBENV_ROOT=$BREW_PREFIX/var/rbenv
```

> I also did try it with the `limactl shell` command and it was consistent with the result I got locally. Correct me if i'm wrong, what I understand from your earlier point is that since the shell command makes use of the login shell, the guest's value would actually take precedence but that's not the case in this scenario regarding PATH as well.

> [...]
> 
> which is same as the PATH in the host machine. Based on the login shell expectation, I don't think the PATH value should be the same as the value on the host machine, hence it shouldn't get overwritten even when it is added to the `LIMA_SHELLENV_ALLOW` list. Is this right?

What shell are you running, and why do you have backticks around ``` `--preserve-env` ```? That tries to run `--preserve-env` as a command, so I'm curious how this worked for you.

I'm also surprised that you get exactly the same `PATH` as on the host. Are you using the template://default template for your instance?

When I run your command, I get back the `PATH` from the host, with a `:/snap/bin:/usr/sbin:/sbin` suffix.

This happens because `PATH` is not overwritten in the shell profile scripts in the VM, but appended to:

```console
jan@lima-default:~$ cat /etc/profile.d/apps-bin-path.sh
# shellcheck shell=sh

# Expand $PATH to include the directory where snappy applications go.
snap_bin_path="/snap/bin"
if [ -n "${PATH##*${snap_bin_path}}" ] && [ -n "${PATH##*${snap_bin_path}:*}" ]; then
    export PATH="$PATH:${snap_bin_path}"
fi
...

jan@lima-default:~$ tail -5 ~/.profile
# Lima BEGIN
# Make sure iptables and mount.fuse3 are available
PATH="$PATH:/usr/sbin:/sbin"
export PATH
# Lima END
```

Which is why `PATH` needs to be on the default block list.

I have no idea why you don't get the additional directories appended; the only explanation I have is that you use a different template.

hack/test-preserve-env.sh
```diff
@@ -0,0 +1,110 @@
...
+	ERROR "Usage: $0 NAME"
+	exit 1
+fi
+
+NAME="$1"
+
+INFO "Testing --preserve-env flag"
+
+# Clean up any existing test environment variables
+unset LIMA_TEST_PRESERVE_ENV_VAR 2>/dev/null || true
```
jandubois: `unset` always returns success unless an invalid option is given, or the variable is read-only (in which case you want to see the error and fail).

```suggestion
unset LIMA_TEST_PRESERVE_ENV_VAR
```

hack/test-preserve-env.sh
```diff
@@ -0,0 +1,110 @@
...
+	exit 1
+fi
+
+# Final cleanup
+unset LIMA_TEST_PRESERVE_ENV_VAR 2>/dev/null || true
+unset LIMA_TEST_CUSTOM_BLOCK 2>/dev/null || true
+unset LIMA_TEST_ALLOW_VAR 2>/dev/null || true
+unset LIMA_TEST_OTHER_VAR 2>/dev/null || true
+unset LIMA_SHELLENV_ALLOW 2>/dev/null || true
+unset LIMA_SHELLENV_BLOCK 2>/dev/null || true
```
jandubois: The file is called as a script, not included via `source`, so all variables are local to the file and don't need to be reset. Even if exported, the export would not affect the parent process.

pkg/envutil/envutil.go
olamilekan000: >What shell are you running, and why do you have backticks around `--preserve-env`? That tries to run --preserve-env as a command, so I'm curious how this worked for you.

I use `zsh` and the backtick was probably during formating of the command when trying to type in the comment section. This is what it looks like in my shell
```
LIMA_SHELLENV_ALLOW=PATH limactl shell --preserve-env default bash -c 'echo "login shell PATH: $PATH"' 
```

>I'm also surprised that you get exactly the same PATH as on the host. Are you using the template://default template for your instance?

Yes, I am using the default template
```
 limactl ls                                                                                            
default    Running    127.0.0.1:60022    qemu      aarch64    4       4GiB      100GiB    ~/.lima/default
```

olamilekan000: This is what the PATH looks like in the guest when it is not included in the "Allowed List"
```
➜  limactl shell --preserve-env default bash -c 'echo "login shell PATH: $PATH"' 

login shell PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin:/usr/sbin:/sbin
```

jandubois (approved):
Thanks, LGTM

PR #3888, opened 2025-08-21
refactor: reduce context.Background()

### LINKED ISSUE #3875: Reduce `context.Background()`
https://github.com/search?q=repo%3Alima-vm%2Flima%20context.Background&type=code

Similar to:
-  #3874

Some occurrences of `context.Background()` are legit and should not be removed
https://github.com/lima-vm/lima/blob/f3afdb6e4f8a9d3796875cee4b2eab630e27bec9/pkg/hostagent/hostagent.go#L592

ashwat287:
fixes #3875

### DIFF
--- a/cmd/limactl/clone.go
+++ b/cmd/limactl/clone.go
@@ -72,7 +72,7 @@ func cloneAction(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
-		yBytes, err := yqutil.EvaluateExpression(yq, yContent)
+		yBytes, err := yqutil.EvaluateExpression(ctx, yq, yContent)
 		if err != nil {
 			return err
 		}
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -46,7 +46,7 @@ type Lifecycle interface {
 type GUI interface {
 	// RunGUI is for starting GUI synchronously by hostagent. This method should be wait and return only after vm terminates
 	// It returns error if there are any failures
-	RunGUI() error
+	RunGUI(context.Context) error
 
 	ChangeDisplayPassword(ctx context.Context, password string) error
 	DisplayConnection(ctx context.Context) (string, error)
@@ -63,7 +63,7 @@ type SnapshotManager interface {
 // GuestAgent defines operations for the guest agent.
 type GuestAgent interface {
 	// ForwardGuestAgent returns if the guest agent sock needs forwarding by host agent.
-	ForwardGuestAgent() bool
+	ForwardGuestAgent(context.Context) bool
 
 	// GuestAgentConn returns the guest agent connection, or nil (if forwarded by ssh).
 	GuestAgentConn(_ context.Context) (net.Conn, string, error)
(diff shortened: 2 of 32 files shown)


cmd/limactl/clone.go
```diff
@@ -71,7 +71,7 @@ func cloneAction(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return err
 		}
-		yBytes, err := yqutil.EvaluateExpression(yq, yContent)
+		yBytes, err := yqutil.EvaluateExpression(cmd.Context(), yq, yContent)
```
AkihiroSuda:
```suggestion
		yBytes, err := yqutil.EvaluateExpression(ctx, yq, yContent)
```

Same for other occurrences of `cmd.Context()` too

pkg/driver/driver.go
```diff
@@ -63,7 +63,7 @@ type Registration interface {
 // GuestAgent defines operations for the guest agent.
 type GuestAgent interface {
 	// ForwardGuestAgent returns if the guest agent sock needs forwarding by host agent.
-	ForwardGuestAgent() bool
+	ForwardGuestAgent(_ context.Context) bool
```
jandubois: If a function/method has only unnamed parameters, then you don't need to specify the `_` name. It is only needed when you want to provide a name for parameters with a basic type like `bool`. `int`, or `string`, which are not self-explanatory because you cannot mix named and unnamed parameters.

```suggestion
	ForwardGuestAgent(context.Context) bool
```

pkg/hostagent/hostagent.go
```diff
@@ -585,19 +585,20 @@ func (a *HostAgent) watchGuestAgentEvents(ctx context.Context) {
 	remoteUnix := "/run/lima-guestagent.sock"
 
 	a.onClose = append(a.onClose, func() error {
+		ctxBG := context.Background()
```
jandubois: I feel like this is a premature optimization. You introduce an extra variable name that the user needs to keep track of instead of calling a very cheap function multiple times.

It doesn't matter much, but I would always try to minimize the context that the developer needs to carry in their head while reading the code, especially if it is a rather long function.

jandubois (dismissed):
Thanks, LGTM

I have only 2 stylistic nit-picks that you can address, if you want. But we already have existing functions that use `_` for the single unnamed parameter, so it isn't a new issue.

There is of course the issue of failing CI, but I think this might be GitHub flakiness; we'll see...

pkg/hostagent/hostagent.go
ashwat287: Yeah, makes sense—I'll drop `ctxBG` and just call `context.Background()` directly.

pkg/driver/driver.go
ashwat287: Shall we raise a new Issue/PR focusing on removing `_` for single unnamed parameter to make it consistent?

unsuman:
@ashwat287, please fix the merge conflicts

pkg/hostagent/dns/dns.go
```diff
@@ -53,6 +53,15 @@ type Handler struct {
...
 
+type DNSHandler interface {
+	ServeDNS(ctx context.Context, w dns.ResponseWriter, r *dns.Msg)
+}
+
+type DNSServer struct {
+	Server     dns.Server
+	DNSHandler DNSHandler
+}
+
```
unsuman: Why is this needed? This seems out of context for this PR

unsuman (changes requested):
The changes in the `pkg/hostagent/dns/dns.go` might be the reason for CI failures

jandubois:
Superseded by #4946

PR #3892, opened 2025-08-22
feature: "yq" provision mode

norio-nomura:
~~based on #3908~~
```yaml
# Create or edit a file in the guest filesystem by using `yq`.
# The file specified by `path` will be updated with the specified `expression` using `yq --inplace`.
# If the file does not exist, it will be created using `yq --null-input`.
# `format` is optional and defaults to "auto".
# One of the following is expected to work:
#   "auto", "csv", "ini", "json", "props", "tsv", "toml", "xml", "yaml"
# See https://github.com/mikefarah/yq for more info.
# Any missing directories will be created as needed.
# The file permissions will be set to the specified value.
# The file and directory creation will be performed as the specified owner.
# If the existing file is not writable by the specified owner, the operation will fail.
# `path` and `expression` are required.
# `owner` and `permissions` are optional. Defaults to "root:root" and 644.
- mode: yq
  path: "{{.Home}}/.config/docker/daemon.json"
  expression: ".features.containerd-snapshotter = {{.Param.containerdSnapshotter}}"
  format: auto
  owner: "{{.User}}"
  permissions: 644
```

updated:
- ~~based on #3908~~
- dropped `provisionTool`
- applied reviews

### DIFF
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -56,6 +56,7 @@ declare -A CHECKS=(
 	["user-v2"]=""
 	["mount-path-with-spaces"]=""
 	["provision-data"]=""
+	["provision-yq"]=""
 	["param-env-variables"]=""
 	["set-user"]=""
 	["preserve-env"]="1"
@@ -90,6 +91,7 @@ case "$NAME" in
 	CHECKS["clone"]="1"
 	CHECKS["mount-path-with-spaces"]="1"
 	CHECKS["provision-data"]="1"
+	CHECKS["provision-yq"]="1"
 	CHECKS["param-env-variables"]="1"
 	CHECKS["set-user"]="1"
 	;;
@@ -197,6 +199,11 @@ fi
 if [[ -n ${CHECKS["provision-data"]} ]]; then
 	INFO 'Testing that /etc/sysctl.d/99-inotify.conf was created successfully on provision'
 	limactl shell "$NAME" grep -q fs.inotify.max_user_watches /etc/sysctl.d/99-inotify.conf
+fi
+
+if [[ -n ${CHECKS["provision-yq"]} ]]; then
+	INFO 'Testing that /tmp/param-yq.json was created successfully on provision'
+	limactl shell "$NAME" grep -q '"YQ": "yq"' /tmp/param-yq.json
 fi
 
 if [[ -n ${CHECKS["param-env-variables"]} ]]; then
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.sh
@@ -64,15 +64,89 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
 		owner=$(deref "LIMA_CIDATA_DATAFILE_${filename}_OWNER")
 		path=$(deref "LIMA_CIDATA_DATAFILE_${filename}_PATH")
 		permissions=$(deref "LIMA_CIDATA_DATAFILE_${filename}_PERMISSIONS")
+		user="${owner%%:*}"
 		if [ -e "$path" ] && [ "$overwrite" = "false" ]; then
 			INFO "Not overwriting $path"
 		else
 			INFO "Copying $f to $path"
-			# intermediate directories will be owned by root, regardless of OWNER setting
-			mkdir -p "$(dirname "$path")"
+			if ! sudo -iu "${user}" mkdir -p "$(dirname "$path")"; then
+				WARNING "Failed to create directory for ${path} (as user ${user})"
+				WARNING "Falling back to creating directory as root to maintain compatibility"
+				mkdir -p "$(dirname "$path")"
+			fi
 			cp "$f" "$path"
 			chown "$owner" "$path"
 			chmod "$permissions" "$path"
+		fi
+	done
+fi
+
+if [ -d "${LIMA_CIDATA_MNT}"/provision.yq ]; then
+	yq="${LIMA_CIDATA_MNT}/lima-guestagent yq"
+	for f in "${LIMA_CIDATA_MNT}"/provision.yq/*; do
+		filename=$(basename "${f}")
+		format=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_FORMAT")
+		owner=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_OWNER")
+		path=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_PATH")
+		permissions=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_PERMISSIONS")
+		user="${owner%%:*}"
+		# Creating intermediate directories may fail if the user does not have permission.
+		# TODO: Create intermediate directories with the specified group ownership.
+		if ! sudo -iu "${user}" mkdir -p "$(dirname "${path}")"; then
+			WARNING "Failed to create directory for ${path} (as user ${user})"
+			CODE=1
+			continue
+		fi
+		# Since CIDATA is mounted with dmode=700,fmode=700,
+		# `lima-guestagent yq` cannot be executed by non-root users,
+		# and provision.yq/* files cannot be read by non-root users.
+		if [ -f "${path}" ]; then
+			INFO "Updating ${path}"
+			# If the user does not have write permission, it should fail.
+			# This avoids changes being made by the wrong user.
+			if ! sudo -iu "${user}" test -w "${path}"; then
+				WARNING "File ${path} is not writable by user ${user}"
+				CODE=1
+				continue
+			fi
+			# Relies on the fact that yq does not change the owner of the existing file.
+			if ! ${yq} --inplace --from-file "${f}" --input-format "${format}" --output-format "${format}" "${path}"; then
+				WARNING "Failed to update ${path} (as user ${user})"
+				CODE=1
+				continue
+			fi
+		else
+			if [ "${format}" = "auto" ]; then
+				# yq can't determine the output format from non-existing files
+				case "${path}" in
+				*.csv) format=csv ;;
+				*.ini) format=ini ;;
+				*.json) format=json ;;
+				*.properties) format=properties ;;
+				*.toml) format=toml ;;
+				*.tsv) format=tsv ;;
+				*.xml) format=xml ;;
+				*.yaml | *.yml) format=yaml ;;
+				*)
+					format=yaml
+					WARNING "Cannot determine file type for ${path}, using yaml format"
+					;;
+				esac
+			fi
+			INFO "Creating ${path}"
+			if ! ${yq} --null-input --from-file "${f}" --output-format "${format}" | sudo -iu "${user}" tee "${path}"; then
+				WARNING "Failed to create ${path} (as user ${user})"
+				CODE=1
+				continue
+			fi
+		fi
+		if ! sudo -iu "${user}" chown "${owner}" "${path}"; then
+			WARNING "Failed to set owner for ${path} (as user ${user})"
+			CODE=1
+		fi
+		if ! sudo -iu "${user}" chmod "${permissions}" "${path}"; then
+			WARNING "Failed to set permissions for ${path} (as user ${user})"
+			CODE=1
 		fi
 	done
 fi
--- a/pkg/limatype/lima_yaml.go
+++ b/pkg/limatype/lima_yaml.go
@@ -235,6 +235,7 @@ const (
 	ProvisionModeDependency ProvisionMode = "dependency"
 	ProvisionModeAnsible    ProvisionMode = "ansible" // DEPRECATED
 	ProvisionModeData       ProvisionMode = "data"
+	ProvisionModeYQ         ProvisionMode = "yq"
 )
 
 type Provision struct {
@@ -245,6 +246,9 @@ type Provision struct {
 	Playbook                        string             `yaml:"playbook,omitempty" json:"playbook,omitempty"` // DEPRECATED
 	// All ProvisionData fields must be nil unless Mode is ProvisionModeData
 	ProvisionData `yaml:",inline"` // Flatten fields for "strict" YAML mode
+	// ProvisionModeYQ borrows Owner, Path, and Permissions from ProvisionData
+	Expression *string `yaml:"expression,omitempty" json:"expression,omitempty" jsonschema:"nullable"`
+	Format     *string `yaml:"format,omitempty" json:"format,omitempty" jsonschema:"nullable"`
 }
 
 type ProvisionData struct {
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -181,10 +181,10 @@ func Validate(y *limatype.LimaYAML, warn bool) error {
 			}
 		}
 		switch p.Mode {
-		case limatype.ProvisionModeSystem, limatype.ProvisionModeUser, limatype.ProvisionModeBoot, limatype.ProvisionModeData, limatype.ProvisionModeDependency, limatype.ProvisionModeAnsible:
+		case limatype.ProvisionModeSystem, limatype.ProvisionModeUser, limatype.ProvisionModeBoot, limatype.ProvisionModeData, limatype.ProvisionModeDependency, limatype.ProvisionModeAnsible, limatype.ProvisionModeYQ:
 		default:
-			errs = errors.Join(errs, fmt.Errorf("field `provision[%d].mode` must one of %q, %q, %q, %q, %q, or %q",
-				i, limatype.ProvisionModeSystem, limatype.ProvisionModeUser, limatype.ProvisionModeBoot, limatype.ProvisionModeData, limatype.ProvisionModeDependency, limatype.ProvisionModeAnsible))
+			errs = errors.Join(errs, fmt.Errorf("field `provision[%d].mode` must one of %q, %q, %q, %q, %q, %q, or %q",
+				i, limatype.ProvisionModeSystem, limatype.ProvisionModeUser, limatype.ProvisionModeBoot, limatype.ProvisionModeData, limatype.ProvisionModeDependency, limatype.ProvisionModeAnsible, limatype.ProvisionModeYQ))
 		}
 		if p.Mode != limatype.ProvisionModeDependency && p.SkipDefaultDependencyResolution != nil {
 			errs = errors.Join(errs, fmt.Errorf("field `provision[%d].mode` cannot set skipDefaultDependencyResolution, only valid on scripts of type %q",
@@ -192,22 +192,26 @@ func Validate(y *limatype.LimaYAML, warn bool) error {
 		}
 
 		// This can lead to fatal Panic if p.Path is nil, better to return an error here
-		if p.Mode == limatype.ProvisionModeData {
+		switch p.Mode {
+		case limatype.ProvisionModeData, limatype.ProvisionModeYQ:
 			if p.Path == nil {
-				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].path` must not be empty when mode is %q", i, limatype.ProvisionModeData))
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].path` must not be empty when mode is %q", i, p.Mode))
 				return errs
 			}
 			if !path.IsAbs(*p.Path) {
 				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].path` must be an absolute path", i))
 			}
-			if p.Content == nil {
-				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].content` must not be empty when mode is %q", i, limatype.ProvisionModeData))
+			if p.Mode == limatype.ProvisionModeData && p.Content == nil {
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].content` must not be empty when mode is %q", i, p.Mode))
+			}
+			if p.Mode == limatype.ProvisionModeYQ && p.Expression == nil {
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].expression` must not be empty when mode is %q", i, p.Mode))
 			}
 			// FillDefaults makes sure that p.Permissions is not nil
 			if _, err := strconv.ParseInt(*p.Permissions, 8, 64); err != nil {
 				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].permissions` must be an octal number: %w", i, err))
 			}
-		} else {
+		default:
 			if p.Script == "" && p.Mode != limatype.ProvisionModeAnsible {
 				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].script` must not be empty", i))
 			}
@@ -221,10 +225,13 @@ func Validate(y *limatype.LimaYAML, warn bool) error {
 				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].owner` can only be set when mode is %q", i, limatype.ProvisionModeData))
 			}
 			if p.Path != nil {
-				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].path` can only be set when mode is %q", i, limatype.ProvisionModeData))
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].path` can only be set when mode is %q, or %q", i, limatype.ProvisionModeData, limatype.ProvisionModeYQ))
 			}
 			if p.Permissions != nil {
-				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].permissions` can only be set when mode is %q", i, limatype.ProvisionModeData))
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].permissions` can only be set when mode is %q, or %q", i, limatype.ProvisionModeData, limatype.ProvisionModeYQ))
+			}
+			if p.Format != nil {
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].format` can only be set when mode is %q", i, limatype.ProvisionModeYQ))
 			}
 		}
 		if p.Playbook != "" {
@@ -489,13 +496,11 @@ func validateParamIsUsed(y *limatype.LimaYAML) error {
 		}
 		keyIsUsed := false
 		for _, p := range y.Provision {
-			if re.MatchString(p.Script) {
-				keyIsUsed = true
-				break
-			}
-			if p.Content != nil && re.MatchString(*p.Content) {
-				keyIsUsed = true
-				break
+			for _, ptr := range []*string{&p.Script, p.Content, p.Expression, p.Owner, p.Path, p.Permissions} {
+				if ptr != nil && re.MatchString(*ptr) {
+					keyIsUsed = true
+					break
+				}
 			}
 			if p.Playbook != "" {
 				playbook, err := os.ReadFile(p.Playbook)
--- a/templates/default.yaml
+++ b/templates/default.yaml
@@ -246,6 +246,49 @@ containerd:
 #   owner: "root:root"
 #   permissions: 644
 #   overwrite: false
+# # Create or edit a file in the guest filesystem by using `yq`.
+# # The file specified by `path` will be updated by `expression`.
+# # An empty file of the required `format` will be created if it does not yet exist.
+# # `format` defaults to "auto" and will be detected by file extension of `path`.
+# # If the extension is not recognized by `yq` then `format` must be set to a
+# # value from this list:
+# #   "auto", "csv", "ini", "json", "props", "tsv", "toml", "xml", "yaml"
+# # See https://github.com/mikefarah/yq for more info.
+# # Any missing directories will be created as needed.
+# # The file permissions will be set to the specified value.
+# # The file and directory creation will be performed as the specified owner.
+# # If the existing file is not writable by the specified owner, the operation will fail.
+# # `path` and `expression` are required.
+# # `owner` and `permissions` are optional. Defaults to "root:root" and 644.
+# - mode: yq
+#   path: "{{.Home}}/.config/docker/daemon.json"
+#   expression: ".features.containerd-snapshotter = {{.Param.containerdSnapshotter}}"
+#   format: auto
+#   owner: "{{.User}}"
+#   permissions: 644
+#
+# Q. In what order are provision scripts executed?
+# A. All provisions are processed per boot by each module in stages of cloud-init as follows:
+#   1. cloud-init 'init' stage
+#     bootcmd:
+#     - `mode: boot` scripts are executed
+#
+#   2. cloud-init 'config' stage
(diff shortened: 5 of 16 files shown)


AkihiroSuda:
Why not just apt-get yq in a shell script?

jandubois:
I think this is too specialized to be built into Lima given that you can already do this with a regular provisioning script:

```yaml
provision:
- mode: user
  script: |
    if ! command -v yq >/dev/null 2>&1; then
      apt-get update
      apt install -y yq
    fi
    mkdir -p ...
    touch ...
    yq ...
```

norio-nomura:
> Why not just apt-get yq in a shell script?

Because `yq` installed by `apt-get` is https://github.com/kislyuk/yq
This PR installs https://github.com/mikefarah/yq

jandubois:
> Because `yq` installed by `apt-get` is https://github.com/kislyuk/yq

Ok, that is unfortunate. I was wondering why it was dragging in Python.

But `yq` is a standalone binary, so you can also curl/wget it from the GitHub release page.

jandubois:
If the guestagent was already using `yqlib`, then I would be sympathetic to adding a `lima-guestagent yq` subcommand for convenience, but afaict only `limactl` uses `yqlib`.

And we would rather want to bring down the size of the guestagents, so I don't think we should be doing that either.

AkihiroSuda:
> I was wondering why it was dragging in Python.

- Python implementation: 2017 https://github.com/kislyuk/yq/commit/ce72e7289b4ccad52e0cbe78b7cdcd9b5ccd5210
  - Older Python implementation: 2014 https://github.com/abesto/yq/commit/efbddc99c26c49efc05825800bdf0386364b93e4
- Go implementation: 2018 https://github.com/mikefarah/yq/commit/f2e10f21c7be3173ea5dbfe041723509bb798c4a

Dragging in the Python implementation makes a sense, as it seems more "authentic" version

jandubois:
> Dragging in the Python implementation makes a sense, as it seems more "authentic" version

Fair enough. Unfortunately on homebrew `yq` is the standalone Go version, creating a bit of a mess. 😞

https://github.com/Homebrew/homebrew-core/issues/29529

jandubois:
If we want to support adding large local files to the cidata volume, then I would suggest the way to do this would be to extend the `data` provisioning mode. I did try this:

```yaml
base: template://docker
provision:
- mode: data
  path: /usr/local/bin/yq
  file: ~/Downloads/yq_linux_arm
  permissions: 755
```

It didn't work because of the tilde path (I just created a PR to fix that), but also because template embedding would attempt to include the file as a base64 encoded string, which exceeds the (somewhat arbitrary) max size of 4MB we have for `lima.yaml`.

I still think that once created a Lima instance should not reference other local files outside the instance directory. We could get rid of the file size limit (not a good idea, imo), or we could have some mechanism that files would be copied into a cache directory inside the instance directory, and the file reference would be rewritten to point to it.

I don't think the use case is particularly strong though when you can also just fetch the file from GitHub in a provisioning script. We do this with all manner of other prerequisites as well. But I can see the utility when you want to bundle something that is not available for download.

norio-nomura:
This PR adds `--provision-tool yq=<local path to executable>` option to `hostagent` to install the executable as `provision/tool/yq` in cidata.

norio-nomura:
```yaml
provisionTool:
  yq:
  - location: "~/Downloads/yq_linux_amd64"
    arch: "x86_64"
    digest: "sha256:..."
```
The example above is from `.containerd.archives`, but it doesn’t support `~` expansion.

afbjorklund:
@norio-nomura : I think you should add an issue for what you are trying to accomplish here, before diving straight into the implementation and the PR? My gut feeling is that this will meet the same fate as the "ansible" provisioning.

jandubois:
I know this is just an example, but I just realized that you are editing a JSON file, so you can just use `jq` (which is already installed in Ubuntu).

jandubois:
`snap.tmpl` using built-in `jq`:

```yaml
base: template://docker
provision:
- mode: user
  script: |
    set -eux -o pipefail
    DAEMON="{{.Home}}/.config/docker/daemon.json"
    mkdir -p "$(dirname "$DAEMON")"
    [[ ! -e $DAEMON ]] && echo "{}" > "$DAEMON"
    chmod 644 "$DAEMON"
    EXPR='.features["containerd-snapshotter"] = {{.Param.containerdSnapshotter}}'
    jq "$EXPR" "$DAEMON" >"$DAEMON.new"
    mv "$DAEMON.new" "$DAEMON"
param:
  containerdSnapshotter: true
```

Seems to work fine:

```console
❯ limactl start --yes --name snap ./snap.tmpl

❯ limactl shell snap sh -c "cat ~/.config/docker/daemon.json"
{
  "features": {
    "containerd-snapshotter": true
  }
}
```

jandubois:
Unrelated, but I also think we should just always use the `containerdSnapshotter` in the `docker` template. I'm not aware of any reason to use the legacy one.

Even Docker Desktop has been using it by default for over a year now for all new installations. It is not considered experimental anymore.

norio-nomura:
> `snap.tmpl` using built-in `jq`:

`jq` is not built-in on Ubuntu-Minimal that I'm using.

> Seems to work fine:

Yup, I also have some `lima.yaml` files that contain similar provision scripts.

I want to make those tasks much simpler like this PR does.

norio-nomura:
> @norio-nomura : I think you should add an issue for what you are trying to accomplish here, before diving straight into the implementation and the PR? My gut feeling is that this will meet the same fate as the "ansible" provisioning.

If I make an issue before I write a PR, I expect to receive an answer like https://github.com/lima-vm/lima/pull/3892#issuecomment-3215398578 It came.

Being able to make `limactl` with this function at my local Mac is more valuable to me than this PR is merged. I just prioritized the time to realize this function rather than the time spent communicating with the issue.

norio-nomura:
Added `format` field. Now, it can manipulate `systemd`'s unit files like the following:
```yaml
- mode: yq
  path: /etc/systemd/system/docker.service.d/override.conf
  format: ini
  expression: .Socket.SocketUser="{{.User}}"
```

afbjorklund:
> Being able to make `limactl` with this function at my local Mac is more valuable to me than this PR is merged. I just prioritized the time to realize this function rather than the time spent communicating with the issue.

That is OK, but it also runs into the risk of 1) not realizing there is a simpler way to accomplish the same thing and 2) risking that the PR is never merged - or maybe merged and then deprecated and removed again, like Ansible

jandubois:
I like the `provision.mode: yq` feature, but think the `provisionTool` design is too specialized, and would have to either piggy-back on the `provision.mode: data` mechanism or need to be hardcoded into Lima like the `nerdctl` archives.

I think a better approach would be to bundle `yq` into our existing binaries. That way we can re-use all the bundled runtime library code (and even `yqlib` in the case of `limactl`), and don't have to worry about versions/locations etc.

I've created #3908 to show how simple this is to implement. What do you think about using that approach @norio-nomura?

norio-nomura:
> What do you think about using that approach

~~It was a method that was initially considered and rejected as "unacceptable to the community". If it is accepted, there is no problem.~~
That's how I first considered and rejected it in anticipation that "it doesn't seem acceptable to the community". If my prediction is wrong and acceptable, I will not object to using that method.

The `provisionTool` is just a little more versatile to the function that bundles `yq`, and it is not the main purpose, so it doesn't matter if it's gone.

norio-nomura:
I will separate the `provisionTool` related code to a separate commit.

norio-nomura:
updated:
- based on #3908
- dropped `provisionTool`

hack/test-templates.sh
```diff
@@ -57,6 +57,7 @@ declare -A CHECKS=(
 	["mount-path-with-spaces"]=""
 	["provision-data"]=""
 	["param-env-variables"]=""
+	["provision-yq"]=""
```
jandubois, line 59: This is a tiny nitpick, so please ignore unless there are other changes needed:

I would keep the `provision-data` and `provision-yq` tests close together in the source because they are related (both provisioning tests).

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -77,6 +77,64 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
 	done
 fi
 
+if [ -d "${LIMA_CIDATA_MNT}"/provision.yq ]; then
+	yq="${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent yq"
+	if ${yq} --version >/dev/null; then
```
jandubois: I don't think this test is needed since `yq` is built into the guest agent.

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -77,6 +77,64 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
...
+		for expression in "${LIMA_CIDATA_MNT}"/provision.yq/*; do
```
jandubois: I find `expression` misleading, as it really is a filename. All the other provisioning loops use `f`, so I think this should use the same:

```suggestion
		for f in "${LIMA_CIDATA_MNT}"/provision.yq/*; do
```

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -77,6 +77,64 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
...
+			filename=$(basename "${expression}")
+			format=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_FORMAT")
+			path=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_PATH")
+			permissions=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_PERMISSIONS")
+			user=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_USER")
```
jandubois: I think the `yq` provision mode should use `owner` instead of `user` to match the `data` provision mode. The `path` and `permissions` properties use the same name, but `user` is different from `owner`, which is confusing.

```suggestion
			owner=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_OWNER")
```

Of course this applies to all other places in the PR too.

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -77,6 +77,64 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
 	done
 fi
 
+if [ -d "${LIMA_CIDATA_MNT}"/provision.yq ]; then
+	yq="${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent yq"
```
jandubois: Guestagent is only installed when `LIMA_CIDATA_PLAIN` is not set. I assume the provisioning mode should work for plain instance too, so need to execute it from the cidata mount point.

Also need to make sure the guestagent is included in the CIDATA iso for plain instances if there is at least one `yq` provisioning entry.

```suggestion
	yq="${LIMA_CIDATA_MNT}/lima-guestagent yq"
```

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -77,6 +77,64 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
...
+			path=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_PATH")
+			permissions=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_PERMISSIONS")
+			user=$(deref "LIMA_CIDATA_YQ_PROVISION_${filename}_USER")
+			if ! sudo -iu "${user}" mkdir -p "$(dirname "${path}")"; then
+				WARNING "Failed to create directory for ${path} (as user ${user})"
+				CODE=1
+				continue
+			fi
+			# Since CIDATA is mounted with dmode=700,fmode=700,
+			# provision.tool/yq cannot be executed by non-root users,
```
jandubois: This isn't actually true anymore because the guestagent gets installed, but see comment above about calling the guestagent from the cidata mount for plain instances, which makes it true again (except for the command name).

```suggestion
			# `lima-guestagent yq` cannot be executed by non-root users,
```

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -77,6 +77,64 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
...
+				fi
+			else
+				if [ "${format}" = "auto" ]; then
+					# yq can't determine the output format from non-existing files
+					case "${path}" in
+					*.ini) format=ini ;;
+					*.json) format=json ;;
+					*.properties) format=properties ;;
+					*.xml) format=xml ;;
+					*.yaml | *.yml) format=yaml ;;
```
jandubois: I would duplicate the full list from `yq` format autodetection here. At the very least you should add `csv` and `toml`, but I would add everything from https://github.com/mikefarah/yq/blob/76c817009f99310f7719fc006e833b00036e8108/pkg/yqlib/format.go#L20-L83 (except for the single-character versions).

pkg/limatype/lima_yaml.go
```diff
@@ -245,6 +246,10 @@ type Provision struct {
 	Playbook                        string             `yaml:"playbook,omitempty" json:"playbook,omitempty"` // DEPRECATED
 	// All ProvisionData fields must be nil unless Mode is ProvisionModeData
 	ProvisionData `yaml:",inline"` // Flatten fields for "strict" YAML mode
+	// ProvisionModeYQ borrows Path and Permissions from ProvisionData
+	Expression *string `yaml:"expression,omitempty" json:"expression,omitempty" jsonschema:"nullable"`
+	User       *string `yaml:"user,omitempty" json:"user,omitempty" jsonschema:"nullable"`
+	Format     *string `yaml:"format,omitempty" json:"format,omitempty" jsonschema:"nullable"`
```
jandubois: As I mentioned above, I think it should borrow "Owner" as well.

I'm undecided if `expression` should be `expr` for brevity. I already disliked `permissions` as too long, but `mode` was already taken.

Either `expr` or `expression` is fine with me, but would like to hear what others think!

pkg/limayaml/validate.go
```diff
@@ -207,7 +208,22 @@ func Validate(y *limatype.LimaYAML, warn bool) error {
...
+			if !path.IsAbs(*p.Path) {
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].path` must be an absolute path", i))
+			}
+			if p.Expression == nil {
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].expression` must not be empty when mode is %q", i, limatype.ProvisionModeYQ))
+			}
+			// FillDefaults makes sure that p.Permissions is not nil
+			if _, err := strconv.ParseInt(*p.Permissions, 8, 64); err != nil {
+				errs = errors.Join(errs, fmt.Errorf("field `provision[%d].permissions` must be an octal number: %w", i, err))
+			}
```
jandubois: The `Path` and `Permissions` validation is duplicated, which is 75% of the code block. I wonder if they shouldn't be in a shared codepath, and only `Content`/`Expression` are separate?

pkg/limayaml/validate.go
```diff
@@ -493,9 +515,16 @@ func validateParamIsUsed(y *limatype.LimaYAML) error {
 				keyIsUsed = true
 				break
 			}
-			if p.Content != nil && re.MatchString(*p.Content) {
-				keyIsUsed = true
-				break
+			for _, ptr := range []*string{
+				// mode: data
+				p.Content, p.Owner, p.Path,
```
jandubois: Good catch of the additional templates. Maybe add `&p.Script` as well, then you can eliminate the check above as well.

templates/default.yaml
```diff
@@ -246,6 +246,27 @@ containerd:
 #   owner: "root:root"
 #   permissions: 644
 #   overwrite: false
+# # `yq` creates or edits a file in the guest filesystem by using `${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent yq`.
+# # `${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent yq` has the same functionality as https://github.com/mikefarah/yq.
```
jandubois: I think that is too much internals knowledge. Just say it uses `yq`. We already use `yq` in other places, like the `--set` CLI options.

```suggestion
# # Create or edit a file in the guest filesystem by using `yq`.
```

templates/default.yaml
```diff
@@ -246,6 +246,27 @@ containerd:
...
+# # The file specified by `path` will be updated with the specified `expression` using `yq --inplace`.
+# # If the file does not exist, it will be created using `yq --null-input`.
+# # `format` is optional and defaults to "auto", which is passed to `yq` by `--input-format <format>`, `--output-format <format>`.
```
jandubois: Just explain what it does, not how it is implemented.

```suggestion
# # `format` is optional and defaults to "auto".
```

templates/default.yaml
```diff
@@ -246,6 +246,27 @@ containerd:
...
+# # One of the following is expected to work:
+# #   "auto", "a", "ini", "i", "json", "j", "props", "p", "xml", "x", "yaml", "y"
+# # `yq` v4.47.1 can write .ini, .json, .properties, .xml, .yaml, and .yml files.
+# # See https://github.com/mikefarah/yq for more info.
+# # Any missing directories will be created as needed using `mkdir -p`.
+# # The file permissions will be set to the specified value using `chmod`.
+# # The file and directory creation will be performed with the specified user privileges.
```
jandubois:
```suggestion
# # Any missing directories will be created as needed.
# # The file permissions will be set to the specified value.
# # The file and directory creation will be performed as the specified owner.
```

templates/docker-rootful.yaml
```diff
@@ -68,3 +67,5 @@ message: |
   docker context use lima-{{.Name}}
   docker run hello-world
   ------
+param:
+  containerdSnapshotter: false
```
jandubois, line 71: I understand you probably don't want to change the default behaviour, but I would be supportive of changing the default to `true`.

Are there any scenarios where this would break something?

jandubois (changes requested):

pkg/cidata/cidata.TEMPLATE.d/boot.sh
norio-nomura: `user` of `mode: yq` and `owner` of `mode: data` behave a little differently.

For the directory created by `mkdir -p` in `mode: data`, root will be the owner regardless of the `owner`.
If you create `{{.Home}}/.config/docker/daemon.json` with `mode: data`, `sudo chown -R {{. User}} ~/.config` is required for `mode: user`.

If `user` is aligned with `owner`, I want to align the behavior of `mkdir -p` of `mode: data` so that it is executed by the user of `owner`. 🤔

pkg/cidata/cidata.TEMPLATE.d/boot.sh
norio-nomura: As far as I tested, TOML support is not usable:
```console
root@lima-docker:/etc/containerd# cat config.toml 
  version = 2
  [plugins."io.containerd.grpc.v1.cri"]
    enable_cdi = true
  [proxy_plugins]
    [proxy_plugins."stargz"]
      type = "snapshot"
      address = "/run/containerd-stargz-grpc/containerd-stargz-grpc.sock"
root@lima-docker:/etc/containerd# lima-guestagent yq . config.toml 
Error: only scalars (e.g. strings, numbers, booleans) are supported for TOML output at the moment. Please use yaml output format (-oy) until the encoder has been fully implemented
```

CSV support works:
```console
root@lima-docker:/etc/containerd# cat >test.csv <<EOF
a,b,c,d
1,2,3,4
EOF
root@lima-docker:/etc/containerd# lima-guestagent yq '.[0].a = 2' test.csv
a,b,c,d
2,2,3,4
```
I can’t think of how to use CSV in provisioning. I’ll include both.

pkg/cidata/cidata.TEMPLATE.d/boot.sh
jandubois: Maybe `mode: data` should be changed to do the same thing `mode: yq` does now for `mkdir -p`? It makes no sense to me that these 2 modes would behave differently.

pkg/cidata/cidata.TEMPLATE.d/boot.sh
jandubois: > ```
> root@lima-docker:/etc/containerd# lima-guestagent yq . config.toml 
> Error: only scalars (e.g. strings, numbers, booleans) are supported for TOML output at the moment. Please use yaml output format (-oy) until the encoder has been fully implemented
> ```

Oh, that is too bad. I had used `yq` with simple toml files in the past and I assumed they would be fully supported. This makes it considerably less useful. 😞 

> I can’t think of how to use CSV in provisioning. I’ll include both.

Maybe not, but it costs (almost) nothing to add it.

pkg/cidata/cidata.TEMPLATE.d/boot.sh
norio-nomura: I'll change to remove `user` and use `owner` instead.
I'll align the behavior of `mode: data`.
`owner` also includes group, but it is difficult to apply group to the directory to be created with `mkdir -p` alone.
I will probably leave a comment as TODO and use `mkdir -p` as it is. 🤔

templates/docker-rootful.yaml
norio-nomura: As far as I know, there is nothing broken by enabling containerd snapshotter.

I think it would be easier for users to check the changes when a link to PR is written in the release note if there is a separate PR that only changes this default value.

AkihiroSuda: We can change the default for new instances, but it has to be a separate PR

templates/default.yaml
```diff
@@ -246,6 +246,27 @@ containerd:
...
 #   permissions: 644
 #   overwrite: false
+# # `yq` creates or edits a file in the guest filesystem by using `${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent yq`.
+# # `${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent yq` has the same functionality as https://github.com/mikefarah/yq.
+# # The file specified by `path` will be updated with the specified `expression` using `yq --inplace`.
+# # If the file does not exist, it will be created using `yq --null-input`.
+# # `format` is optional and defaults to "auto", which is passed to `yq` by `--input-format <format>`, `--output-format <format>`.
+# # One of the following is expected to work:
+# #   "auto", "a", "ini", "i", "json", "j", "props", "p", "xml", "x", "yaml", "y"
+# # `yq` v4.47.1 can write .ini, .json, .properties, .xml, .yaml, and .yml files.
```
AkihiroSuda: User does not need to, and not even able to, care about the version of the yq library

templates/default.yaml
```diff
@@ -246,6 +246,27 @@ containerd:
 #   owner: "root:root"
 #   permissions: 644
 #   overwrite: false
+# # `yq` creates or edits a file in the guest filesystem by using `${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent yq`.
+# # `${LIMA_CIDATA_GUEST_INSTALL_PREFIX}/bin/lima-guestagent yq` has the same functionality as https://github.com/mikefarah/yq.
+# # The file specified by `path` will be updated with the specified `expression` using `yq --inplace`.
+# # If the file does not exist, it will be created using `yq --null-input`.
+# # `format` is optional and defaults to "auto", which is passed to `yq` by `--input-format <format>`, `--output-format <format>`.
+# # One of the following is expected to work:
+# #   "auto", "a", "ini", "i", "json", "j", "props", "p", "xml", "x", "yaml", "y"
```
AkihiroSuda: Not sure if we want to support the short form ("a", "i", "j", ... ).
Looks cryptic.

templates/docker-rootful.yaml
```diff
@@ -30,16 +30,15 @@ provision:
...
-    fi
     export DEBIAN_FRONTEND=noninteractive
     curl -fsSL https://get.docker.com | sh
+- mode: yq
+  path: /etc/systemd/system/docker.service.d/override.conf
+  format: ini
+  expression: .Socket.SocketUser="{{.User}}"
+- mode: yq
+  path: "/etc/docker/daemon.json"
+  expression: .features.containerd-snapshotter = {{.Param.containerdSnapshotter}}
```
AkihiroSuda, line 41: The file must not be modified for existing instances

jandubois: Template changes by definition only affect new instances. Old instances continue to use the `lima.yaml` in their instance directory.

This is why I keep insisting that once an instance has been created, it should not have any references to files outside the instance directory. So any local changes in your filesystem will not affect the instance when you restart it.

norio-nomura: The template is only read when a new one is created, so there should be no need to worry about changing the existing instance here.

templates/default.yaml
jandubois: > Not sure if we want to support the short form ("a", "i", "j", ... ).

We should not document them here, but they will be automatically supported because `lima-guestagent yq` is the same as upstream `yq`, which supports them.

norio-nomura: The format will be passed as it is, so it can be used as a specification, but it will be excluded from the comments.

norio-nomura:
Pushed a separate commit that applying reviews

pkg/cidata/cidata.TEMPLATE.d/boot.sh
jandubois: > `owner` also includes group, but it is difficult to apply group to the directory to be created with `mkdir -p` alone.
> I will probably leave a comment as TODO and use `mkdir -p` as it is. 🤔

I think we can just document that the group (if specified) only applies to the file and not any directories, that might need to be created. I would not add the TODO comment; this is just how it is going to work.

If somebody really needs to specify the group of parent directories, then they will have to add a provisioning script, that will fix them up. I doubt anybody will ever need this.

We should document the sequence in which provisioning happens:

1. provision.data files will be written
2. provision.yq scripts will be executed, which can modify files created in 1
3. provision.system and .user scripts will execute, and can rely on 1. and 2. having completed.

norio-nomura: changed `mode: data` to try creating a directory by the user before creating a directory by the root

norio-nomura: > We should document the sequence in which provisioning happens:
> 
> 1. provision.data files will be written
> 2. provision.yq scripts will be executed, which can modify files created in 1
> 3. provision.system and .user scripts will execute, and can rely on 1. and 2. having completed.

Is it better if this is written in `default.yaml`? Or another place?

jandubois: Yes, all the documentation is in `default.yaml`. We should rework the comments in there to become part of the regular docs, but for now this is what we have.

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -64,19 +64,94 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
...
 		path=$(deref "LIMA_CIDATA_DATAFILE_${filename}_PATH")
 		permissions=$(deref "LIMA_CIDATA_DATAFILE_${filename}_PERMISSIONS")
+		user="${owner%%:*}"
 		if [ -e "$path" ] && [ "$overwrite" = "false" ]; then
 			INFO "Not overwriting $path"
 		else
 			INFO "Copying $f to $path"
-			# intermediate directories will be owned by root, regardless of OWNER setting
-			mkdir -p "$(dirname "$path")"
+			# TODO: Create intermediate directories with the specified group ownership.
```
jandubois: I don't think we want to implement this; the user can add a provision.system script to ensure this, if they really care.

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -64,19 +64,94 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.data ]; then
...
+		fi
+		# Since CIDATA is mounted with dmode=700,fmode=700,
+		# `lima-guestagent yq` cannot be executed by non-root users,
+		# and provision.yq/* files cannot be read by non-root users.
+		if [ -f "${path}" ]; then
+			INFO "Updating ${path}"
+			# If the user does not have write permission, it should fail.
+			# This avoids changes being made by the wrong user.
+			if ! sudo -iu "${user}" test -w "${path}"; then
+				WARNING "Failed to update ${path} (as user ${user})"
```
jandubois:
```suggestion
				WARNING "File ${path} is not writable by user ${user}"
```

pkg/limayaml/validate.go
```diff
@@ -489,13 +496,17 @@ func validateParamIsUsed(y *limatype.LimaYAML) error {
...
-			}
-			if p.Content != nil && re.MatchString(*p.Content) {
-				keyIsUsed = true
-				break
+			for _, ptr := range []*string{
+				&p.Script,
+				// mode: data
+				p.Content, p.Owner, p.Path,
+				// mode: yq
+				p.Expression, p.Permissions,
```
jandubois: `p.Permissions` is also part of "mode: data". Maybe it is easiest to remove the comments and have just a list?

templates/default.yaml
```diff
@@ -246,6 +246,25 @@ containerd:
 #   owner: "root:root"
 #   permissions: 644
 #   overwrite: false
+# # Create or edit a file in the guest filesystem by using `yq`.
+# # The file specified by `path` will be updated with the specified `expression` using `yq --inplace`.
+# # If the file does not exist, it will be created using `yq --null-input`.
```
jandubois: Let's simplify some more:

```suggestion
# # The file specified by `path` will be updated by `expression`.
# # An empty file of the required `format` will be created if it does not yet exist.
```

templates/default.yaml
```diff
@@ -246,6 +246,25 @@ containerd:
...
+# # `format` is optional and defaults to "auto".
+# # One of the following is expected to work:
```
jandubois:
```suggestion
# # `format` defaults to "auto" and will be detected by file extension of `path`.
# # If the extension is not recognized by `yq` then `format` must be set to a
# # value from this list:
```

jandubois (commented):
Thanks, LGTM

Still have some nit-picks, and I think we are missing the documentation for the order the provisioning types are executed.

norio-nomura:
Pushed a separate commit that applying additional reviews

norio-nomura:
force pushed to fix lint error

AkihiroSuda:
Please squash the commits

norio-nomura:
Thanks! 🙏🏻

PR #3893, opened 2025-08-22
Also check "data" provisioning scripts for param usage

jandubois:
Fixes:

```console
❯ cat test.tmpl
images: [{location: foo.raw}]
provision:
- mode: data
  path: /etc/foo
  content: |
    FOO={{.Param.FOO}}
param:
  FOO: bar

❯ limactl tmpl copy --fill test.tmpl -
FATA[0000] field `param` key "FOO" is not used in any provision, probe, copyToHost, or portForward
```

With this PR:

```
❯ limactl tmpl copy --fill test.tmpl - | yq '.provision[0].content'
FOO=bar
```

### DIFF
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -521,6 +521,10 @@ func validateParamIsUsed(y *LimaYAML) error {
 				keyIsUsed = true
 				break
 			}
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

Eventually we should have a test suite for this

PR #3894, opened 2025-08-22
Add tilde expansion to file locators

jandubois:
Required for tilde expansion in provisioning files, e.g.

```yaml
❯ cat data.tmpl
base: template://docker
provision:
- mode: data
  path: /usr/local/bin/yq
  file: ~/Downloads/yq_linux_arm
  permissions: 755
```

(Note that this particular example will still fail because the file is too large to be included in `lima.yaml` as a base64-encoded string).

PR also includes a minor `localpathutil.Expand()` refactoring.

### DIFF
--- a/pkg/limatmpl/abs.go
+++ b/pkg/limatmpl/abs.go
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+
+	"github.com/lima-vm/lima/v2/pkg/localpathutil"
 )
 
 // UseAbsLocators will replace all relative template locators with absolute ones, so this template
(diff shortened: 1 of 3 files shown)


pkg/limatmpl/abs.go
```diff
@@ -105,6 +107,13 @@ func absPath(locator, basePath string) (string, error) {
 	if err == nil && len(u.Scheme) > 1 {
 		return locator, nil
 	}
+	// Don't expand relative path to absolute unless we know it needs tilde expansion.
+	if locator == "~" || strings.HasPrefix(locator, "~/") {
```
AkihiroSuda: `if locator == "~" || strings.HasPrefix(locator, "~/")` should be probably moved to a function like `Expandable()` in `pkg/localpathutil`

AkihiroSuda (approved):
Thanks

PR #3895, opened 2025-08-22
Add two fields for images exported from containers

afbjorklund:
The original container image has some optional metadata in the ENTRYPOINT and STOPSIGNAL fields, that could be needed later...

But these are not exported in the generic rootfs, by default.

Make it possible to pass these at runtime, for instance to select between OpenRC and systemd - or SIGTERM and systemd.

Some drivers include their config in the rootfs, like wsl.conf.

Required by: (currently hardcoded)

---

Example for systemd

```Dockerfile
ENTRYPOINT [ "/usr/local/bin/entrypoint", "/sbin/init" ]
```

Where /sbin/init was symlinked over to /lib/systemd/systemd

It could also be just the init directly, without a shell wrapper.

```Dockerfile
STOPSIGNAL SIGRTMIN+3
```

The defaults are to run either with the VM's `init`, or with `--init`

And to use acpi or similar for VMs, or SIGTERM for containers

### DIFF
--- a/pkg/instance/start.go
+++ b/pkg/instance/start.go
@@ -74,6 +74,8 @@ func Prepare(ctx context.Context, inst *limatype.Instance) (*Prepared, error) {
 	kernel := filepath.Join(inst.Dir, filenames.Kernel)
 	kernelCmdline := filepath.Join(inst.Dir, filenames.KernelCmdline)
 	initrd := filepath.Join(inst.Dir, filenames.Initrd)
+	entrypoint := filepath.Join(inst.Dir, filenames.Entrypoint)
+	stopsignal := filepath.Join(inst.Dir, filenames.StopSignal)
 	if _, err := os.Stat(baseDisk); errors.Is(err, os.ErrNotExist) {
 		var ensuredBaseDisk bool
 		errs := make([]error, len(inst.Config.Images))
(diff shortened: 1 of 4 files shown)


AkihiroSuda (changes requested):
Ptobably these should be incubated as VMOpts entries.

```
vmOpts:
  container:
    entrypoint: foo
```

afbjorklund:
> Ptobably these should be incubated as VMOpts entries.

They are specific to the image, depending on their "init"
Some images use OpenRC, while some use systemd

Currently it is only implemented by wsl.conf, for WSL2 :

https://learn.microsoft.com/en-us/windows/wsl/wsl-config
> `/etc/wsl.conf`, while inside a WSL distribution

```ini
[boot]
systemd=true
```

But I would rather **not** have to modify each rootfs image.
(or peek inside the tarball, to see what it actually contains)

jandubois:
I don't understand why the filenames need to be configurable; they can be hardwired, like all the other files in the instance directory.

And I also don't understand why this needs to be configurable. Why don't we always write the entrypoint and stopsignal to the corresponding files? If the driver doesn't need them, then it could ignore them.

afbjorklund:
> I don't understand why the filenames need to be configurable

The filenames are hardcoded, it is the content that is configurable.

Similar to the already existing (underdocumented) `kernel.cmdline` file

The driver would just pass to `--entrypoint` (and maybe `--stop-signal`)

jandubois:
> The filenames are hardcoded, it is the content that is configurable.

Thanks, I misread the diff.

@AkihiroSuda I agree with @afbjorklund that these settings should be defined with the image, next to the `kernel` setting.

I.e. they would be included in the `template://_image/DISTRO.yaml` templates and should not be separated out into the derived templates.

Issue #3506, opened 2025-05-07
Eventually `VMOpts` should be defined as `map[string]any`, and `QEMUOpts` should be moved to `pkg/qemu`.


unsuman:
Eventually `VMOpts` should be defined as `map[string]any`, and `QEMUOpts` should be moved to `pkg/qemu`.
Can be another PR.

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/3500#discussion_r2076792327_

unsuman:
I will work on this.

afbjorklund:
@unsuman: see also https://github.com/lima-vm/lima/pull/4176

PR #3896, opened 2025-08-22
Add runtime option for the docker container driver

afbjorklund:
Make it possible to select between "runc" and "runv", etc.

Validation is done later (by the driver), if actually used.

Required by: (currently hardcoded)

### DIFF
--- a/pkg/limayaml/limayaml.go
+++ b/pkg/limayaml/limayaml.go
@@ -107,12 +107,17 @@ type User struct {
 }
 
 type VMOpts struct {
-	QEMU QEMUOpts `yaml:"qemu,omitempty" json:"qemu,omitempty"`
+	QEMU   QEMUOpts   `yaml:"qemu,omitempty" json:"qemu,omitempty"`
+	Docker DockerOpts `yaml:"docker,omitempty" json:"docker,omitempty"`
 }
 
 type QEMUOpts struct {
 	MinimumVersion *string `yaml:"minimumVersion,omitempty" json:"minimumVersion,omitempty" jsonschema:"nullable"`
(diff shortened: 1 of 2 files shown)


pkg/limayaml/limayaml.go
```diff
@@ -107,14 +107,19 @@ type User struct {
 }
 
 type VMOpts struct {
-	QEMU QEMUOpts `yaml:"qemu,omitempty" json:"qemu,omitempty"`
+	QEMU   QEMUOpts   `yaml:"qemu,omitempty" json:"qemu,omitempty"`
+	Docker DockerOpts `yaml:"docker,omitempty" json:"docker,omitempty"`
 }
```
AkihiroSuda, line 112: This contradicts to the current VM driver design.
Driver details must not be put to the yaml definition

pkg/limayaml/limayaml.go
```diff
@@ -107,14 +107,19 @@ type User struct {
...
 
 type VMOpts struct {
```
AkihiroSuda, line 109:
```suggestion
type VMOpts map[string]any
```

pkg/limayaml/limayaml.go
afbjorklund: Do you have a pointer to the new "opts"?
Is that somewhere in the mega-PR perhaps

AkihiroSuda: - https://github.com/lima-vm/lima/issues/3506

afbjorklund: Okay, so there is no actual code for it yet

PR #3900, opened 2025-08-24
refactor: extract common types and functions from `limayaml` & `store`

### LINKED ISSUE #3769: # Refactor: Lima Config and External Drivers

unsuman:
> [!NOTE]
> This is the first PR out of three PRs. Meging all closes #3769 

This PR refactors to use types, constants and functions from the new `pkg/limatype` package instead of the previous `pkg/store` and `pkg/limayaml` packages. This improves code organisation and updates all relevant imports and usages throughout the codebase. The changes are extensive and touch many files, but are primarily focused on type migration rather than functional changes. These changes are made to avoid cycle import issues, making the code more maintainable and modular.


AkihiroSuda:
Lint is failing.
Should be easy to fix, but let us know if you need help.

pkg/driver/wsl2/vm_windows.go
```diff
@@ -13,12 +13,12 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/sirupsen/logrus"
```
AkihiroSuda, line 16: The import blocks are not properly formatted

pkg/limainfo/limainfo.go
```diff
@@ -69,8 +68,7 @@ func New(ctx context.Context) (*LimaInfo, error) {
 		DefaultTemplate: y,
 		VMTypes:         vmTypes,
 		VMTypesEx:       vmTypesEx,
-		GuestAgents:     make(map[limayaml.Arch]GuestAgent),
-		ShellEnvBlock:   envutil.GetDefaultBlockList(),
```
AkihiroSuda, line 73: Where has it gone?

pkg/limatype/lima_instance.go
```diff
@@ -0,0 +1,107 @@
+package limatype
```
AkihiroSuda, line 4: Missing copyright boilerplate

pkg/limainfo/limainfo.go
unsuman: Ah, my bad. It mistakenly got deleted in commit [5d9na56](https://github.com/lima-vm/lima/pull/3770/commits/5d9a5634bce1e434cc9483dbd5f410c453f82bed) 🤕

afbjorklund:
As I mentioned on the chat, do we need to move _everything_ out from limayaml and store just to fix the cycles?

Or can we move just the "problematic" ones that are needed by the drivers, so that one can still import limayaml

unsuman:
> As I mentioned on the chat, do we need to move _everything_ out from limayaml and store just to fix the cycles?
> 
> Or can we move just the "problematic" ones that are needed by the drivers, so that one can still import limayaml

Checked and I can say that not everything has been moved from the `limayaml` to `limatype`, i.e. only the problematic ones. We need to import the `driverutil` pkg in the `store` pkg, the import cycle issue is happening at this point.

[EDIT]: `store` -> `driverutil` -> `driver` -> `store`
*Here's a catch: I don't remember the exact issue that made me refactor the `store/dirnames` and `store/filenames` into limatype

pkg/limainfo/limainfo.go
AkihiroSuda: Can you fix this? @unsuman

AkihiroSuda (changes requested):
Thanks, LGTM, except the ShellEnvBlock regression

AkihiroSuda (changes requested):
Thanks

AkihiroSuda (approved):

PR #3902, opened 2025-08-24
Display the error when the plugin command failed to run

jandubois:
Which is different from running, and failing. One reason this can happen if the plugin is a shell script without a hash-bang line.

Also removed a `Debugf()` call because there is no way to change the logging level when invoking a plugin.

```console
❯ limactl foo
FATA[0000] external command "/Users/jan/bin/limactl-foo" failed: fork/exec /Users/jan/bin/limactl-foo: exec format error
```


AkihiroSuda (approved):
Thanks

Issue #3903, opened 2025-08-24
Add timeout to cloud-init progress monitoring


olamilekan000:
The `watchCloudInitProgress` function could run indefinitely if cloud-init gets stuck or fails to complete, potentially consuming resources and causing issues. A context timeout should be added to prevent indefinite monitoring while allowing sufficient time for normal cloud-init completion.

norio-nomura:
copy from https://github.com/lima-vm/lima/pull/3846#discussion_r2296852850

> Can you share how you were able to find the bug?

It can be reproduced on my Mac (Apple M2 Pro) with the following:
```console
limactl start template://ubuntu-lts --name=default --progress --set .containerd.user=false
```
built on current HEAD
```console
$ limactl --version
limactl version 1.2.0-193-gd4f4911e
```

norio-nomura:
#3904 does not fix this.

olamilekan000:
> [#3904](https://github.com/lima-vm/lima/pull/3904) does not fix this.

Can you check again.

norio-nomura:
It stops before the completion of cloud-init.
```console
➜ ~10:50:18 10:50:21~ (pull/3904) limactl start template://ubuntu-lts --name=default --progress --set .containerd.user=false
INFO[0000] Terminal is not available, proceeding without opening an editor 
INFO[0000] Starting the instance "default" with internal VM driver "vz" 
WARN[0000] vmType vz: ignoring [VMOpts]                 
INFO[0000] Attempting to download the image              arch=aarch64 digest="sha256:bbecbb88100ee65497927ed0da247ba15af576a8855004182cf3c87265e25d35" location="https://cloud-images.ubuntu.com/releases/noble/release-20250704/ubuntu-24.04-server-cloudimg-arm64.img"
INFO[0000] Using cache "/Users/norio/Library/Caches/lima/download/by-url-sha256/87505913829585726d6c8aac9c6981c46b8d67142ea5e1af8d9f2d7e872e3ecb/data" 
INFO[0000] Converting "/Users/norio/.lima/default/basedisk" (qcow2) to a raw disk "/Users/norio/.lima/default/diffdisk" 
3.50 GiB / 3.50 GiB [---------------------------------------] 100.00% 1.34 GiB/s
INFO[0003] Expanding to 100GiB                          
INFO[0003] [hostagent] hostagent socket created at /Users/norio/.lima/default/ha.sock 
[...]

PR #3904, opened 2025-08-24
add timeout to cloud init progress monitoring

olamilekan000:
```
change adds timeout to cloud init monitoring process.
```
fixes [issue](https://github.com/lima-vm/lima/issues/3903)


pkg/hostagent/hostagent.go
```diff
@@ -292,6 +300,17 @@ func (a *HostAgent) emitEvent(_ context.Context, ev events.Event) {
 	}
 }
 
+func (a *HostAgent) emitCloudInitProgressEvent(ctx context.Context, progress *events.CloudInitProgress) {
+	a.statusMu.RLock()
+	currentStatus := a.currentStatus
```
norio-nomura, line 305: 👍🏻

pkg/hostagent/hostagent.go
```diff
@@ -816,15 +838,26 @@ func forwardSSH(ctx context.Context, sshConfig *ssh.SSHConfig, port int, local,
 }
 
 func (a *HostAgent) watchCloudInitProgress(ctx context.Context) {
+	exitReason := "Cloud-init monitoring completed successfully"
+	var cmd *exec.Cmd
+
+	defer func() {
+		if cmd.Process != nil {
+			logrus.Debug("Cleaning up cloud-init monitoring process")
+			cmd.Process.Kill()
```
norio-nomura: This code will never be executed because it won't reach here until the `tail` process stops in the VM.

olamilekan000: Fixed it
```
olalekanodukoya@lima-defaultz:/Users/olalekanodukoya/works/opensource/lima$ ps aux | grep tail
olaleka+    1704 33.3  0.0   6672  1920 pts/0    S+   06:49   0:00 grep --color=auto tail
```

norio-nomura: confirmed that `ssh` process for `tail` killed. :+1:

norio-nomura (commented):
LGTM! 👍🏻

However, I am concerned that the strings passed to `exitReason` and `Warn()` are the same.
Is there a mechanism for Golang to automatically optimize the use of the same string?

olamilekan000:
> LGTM! 👍🏻
> 
> However, I am concerned that the strings passed to `exitReason` and `Warn()` are the same. Is there a mechanism for Golang to automatically optimize the use of the same string?

AFIK, It does. When the Go compiler sees the same string literal (e.g., "Failed to create stdout pipe for cloud-init monitoring") multiple times in the source code, it doesn’t generate a new copy of that string every time. Instead, it puts one copy of that string in the program’s read-only data segment, and every usage just points to it.

https://goplay.tools/snippet/KDjdapaDigp

norio-nomura (approved):
Thank you for letting me know that duplicate strings are optimized. 🙏🏻 
LGTM!

AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

PR #3908, opened 2025-08-25
Add `yq` subcommands to `limactl` and `lima-guestagent`

jandubois:
Adding it to `lime-guestagent` allows us to use `yq` in provisioning scripts without having to install it from a package repository or downloading it from GitHub.

Adding it to `limactl` makes it available to integration tests without having to install it. It` is essentially free because we already use `yqlib` inside `limactl` anyways, so the executable size doesn't really change.

The size of the (compressed) guestagent for aarch64 increases from 12MB to 14MB, which seems acceptable.

These are "hidden" commands that don't show up in --help output.

```console
❯ limactl yq --version
yq (https://github.com/mikefarah/yq/) version v4.47.1

❯ limactl yq -n '.foo="bar"'
foo: bar

❯ lima lima-guestagent yq -n '.foo="bar"'
foo: bar
```

This PR is meant to support the "yq provisioning mode" proposed in #3892, but I think is also valuable on its own.

---

I was initially thinking about making this a busybox-style "multi-call" binary, so you could put a symlink called `yq` into `/usr/local/bin` and have it work as a regular `yq` binary. But I'm worried about shadowing a `yq` version installed by the user with a shadow manager, so didn't add this functionality.

### DIFF
--- a/cmd/lima-guestagent/main_linux.go
+++ b/cmd/lima-guestagent/main_linux.go
@@ -4,8 +4,10 @@
 package main
 
 import (
+	"os"
 	"strings"
 
+	yq "github.com/mikefarah/yq/v4/cmd"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
@@ -14,6 +16,25 @@ import (
 )
 
 func main() {
+	// `lima-guestagent yq` executes the embedded `yq` command instead.
+	if len(os.Args) > 1 && os.Args[1] == "yq" {
+		os.Args = os.Args[1:]
+
+		cmd := yq.New()
+		args := os.Args[1:]
+		_, _, err := cmd.Find(args)
+		if err != nil && args[0] != "__complete" {
+			// default command when nothing matches...
+			newArgs := []string{"eval"}
+			cmd.SetArgs(append(newArgs, os.Args[1:]...))
+		}
+
+		if err := cmd.Execute(); err != nil {
+			os.Exit(1)
+		}
+		return
+	}
+
 	if err := newApp().Execute(); err != nil {
 		logrus.Fatal(err)
 	}
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 
 	"github.com/mattn/go-isatty"
+	yq "github.com/mikefarah/yq/v4/cmd"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 
@@ -32,6 +33,24 @@ const (
 )
 
 func main() {
+	// `limactl yq` executes the embedded `yq` command instead.
+	if len(os.Args) > 1 && os.Args[1] == "yq" {
+		os.Args = os.Args[1:]
+
+		cmd := yq.New()
+		args := os.Args[1:]
+		_, _, err := cmd.Find(args)
+		if err != nil && args[0] != "__complete" {
+			// default command when nothing matches...
+			newArgs := []string{"eval"}
+			cmd.SetArgs(append(newArgs, os.Args[1:]...))
+		}
+
+		if err := cmd.Execute(); err != nil {
+			os.Exit(1)
+		}
+		return
+	}
 	if runtime.GOOS == "windows" {
 		extras, hasExtra := os.LookupEnv("_LIMA_WINDOWS_EXTRA_PATH")
 		if hasExtra && strings.TrimSpace(extras) != "" {


jandubois:
We probably should include a comment that the `yq` command implementation is copied from the upstream repo, and maybe also include the license file in our repo.

cmd/lima-guestagent/main_linux.go
```diff
@@ -14,6 +16,25 @@ import (
 )
 
 func main() {
+	// `lima-guestagent yq` executes the embedded `yq` command instead.
+	if len(os.Args) > 1 && os.Args[1] == "yq" {
```
AkihiroSuda: Why not make this a cobra subcommand

jandubois: Because I had bad experiences with it before. I can't remember the details, but global options defined by one app could not be removed when calling a different app.

Maybe this was due to the complexity of the apps involved (I was also embedding `kubectl`), but I wasted a lot of time on it.

I also don't really see the point (what problem would that be solving?), given that this is essentially a multi-call binary setup where the embedded binary is not related to the rest of the options. Right now the code is directly lifted from the `yq` `main` package (because you cannot import it). Why would we want to complicate it?

If we wanted to show the subcommand in the `--help` output, then we could still add it to Cobra, but I would still want to delegate the `main` functionality before making the first Cobra call.

jandubois: I've moved out the code copied from `yq` into its own function to make the separation clearer.

LICENSE_YQ
```diff
@@ -0,0 +1,30 @@
+Below is a copy of https://github.com/mikefarah/yq/blob/master/LICENSE
+
+This file is included in the Lima repository because `limactl yq` and
```
AkihiroSuda: Why do we need this file specifically for yq?
We do not copy the license files for other Go dependencies

jandubois: I'm not sure we need it.

I felt this is different because we are effectively bundling/distributing the `yq` command now, both for the host and the VM.

It is just compiled into our binaries, which is just a packaging decision (share statically linked libraries). But otherwise the original utility is fully exposed to the end user, which is conceptually different from just using it as a library:

```console
❯ limactl yq --version
yq (https://github.com/mikefarah/yq/) version v4.47.1
```

I will remove it if you think we don't need it.

jandubois:
I've modified the code to turn this into a proper multi-call binary:

```console
❯ limactl --version
limactl version 1.2.0-204-gcfa73e3e.m

❯ limactl yq -n '.foo="bar"'
foo: bar

❯ ln -s $(which limactl) yq

❯ ./yq --version
yq (https://github.com/mikefarah/yq/) version v4.47.1

❯ ./yq -n '.foo="bar"'
foo: bar
```

LICENSE_YQ
AkihiroSuda: The license file is included in https://github.com/lima-vm/lima/releases/download/v1.2.1/lima-1.2.1-go-mod-vendor.tar.gz

(If this is not enough, we have to duplicate the license files for all the deps, not just for yq)

cmd/limactl/main.go
```diff
@@ -31,7 +32,34 @@ const (
 	advancedCommand     = "advanced"
 )
 
+// yqmain is a copy of main() in https://github.com/mikefarah/yq/blob/master/yq.go
```
AkihiroSuda: Better to split a package, with an appropriate SPDX headers

AkihiroSuda: That will also help reducing code clones across `limactl` and `lima-guestagent`

LICENSE_YQ
jandubois: Good enough for me, I will remove it.

cmd/limactl/main.go
jandubois: What is the appropriate license header? I think it needs to maintain the MIT license and include the original copyright in addition to ours.

cmd/yq/yq.go
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-FileCopyrightText: Copyright (c) 2017 Mike Farah
+// SPDX-License-Identifier: MIT
```
AkihiroSuda: IIUC you can just relicense MIT to Apache-2.0

cmd/yq/yq.go
```diff
@@ -0,0 +1,45 @@
...
+
+// This file has been adapted from https://github.com/mikefarah/yq/blob/master/yq.go
```
AkihiroSuda: s/master/[actual version you referred]/

jandubois:
I've added a second commit to add a `yq.lima` symlink both on the host and inside the VM (assuming the guestagent install directory is on the PATH; at least for provisioning scripts I've added it explicitly).

```console
❯ lima yq.lima --version
yq (https://github.com/mikefarah/yq/) version v4.47.1
```

I thought this fits in with our other `docker.lima`, `kubectl.lima` custom versions.

I've kept it in a separate commit, in case we decide we don't want to do this. I can squash if we want to keep.

I also have not tested the Windows command that copies the file instead of making a symlink.

jandubois:
I'm kind of torn about installing `yq.lima` on Windows. I want to do it for consistency, but it feels wasteful to store a copy of `limactl` just because you don't have symlinks when you can also just use `limactl yq` instead of `yq.lima`.

I think I prefer to not install it on Windows; what do you think?

AkihiroSuda:
> yq.lima

This is confusing. The convention of `<SOMETHING>.lima` is to run `<SOMETHING>` inside an instance of Lima.

jandubois:
> I think I prefer to not install it on Windows; what do you think?

An alternative is to just create a symlink on Windows too, and ignore the failure if the user doesn't have "Developer Mode" enabled on their system (to allow non-admins to create symlinks).

jandubois:
> The convention of `<SOMETHING>.lima` is to run `<SOMETHING>` inside an instance of Lima.

Not really. It is a variant of the tool installed by Lima, that will automatically work with the daemons inside a Lima instance. But the tools run locally, as long as they are installed, and only fall back to running inside the VM when they don't exist locally.

The only tools that always run inside the VM are `apptainer.lima` and `nerdctl.lima`. All of `docker.lima`, `kubectl.lima`, and `podman.lima` run on the host.

The goal is to make the difference invisible. The `.lima` suffix is used to avoid shadowing the native tool.

afbjorklund:
> An alternative is to just create a symlink on Windows too

Couldn't you use a shell wrapper on Windows, and call `limactl yq`?

> All of `docker.lima`, `kubectl.lima`, and `podman.lima` run on the host.

The main difference is that they just call the external tool, like `yq` here

Makefile
```diff
@@ -277,10 +277,17 @@ LIMA_CMDS = $(sort lima lima$(bat)) # $(sort ...) deduplicates the list
...
-HELPER_CMDS = nerdctl.lima apptainer.lima docker.lima podman.lima kubectl.lima
+HELPER_CMDS = nerdctl.lima apptainer.lima docker.lima podman.lima kubectl.lima yq.lima$(exe)
 HELPERS_DEPS = $(addprefix _output/bin/,$(HELPER_CMDS))
 helpers: $(HELPERS_DEPS)
 
+_output/bin/yq.lima$(exe): _output/bin/limactl$(exe)
+ifneq ($(GOOS),windows)
+	ln -sf limactl$(exe) $@
+else
+	cp -aL $< $@
```
afbjorklund: You could copy a cmd/yq.lima script here instead of making a copy

Makefile
```diff
@@ -493,6 +500,7 @@ uninstall:
 		"$(DEST)/bin/docker.lima" \
 		"$(DEST)/bin/podman.lima" \
 		"$(DEST)/bin/kubectl.lima" \
+		"$(DEST)/bin/yq.lima$(exe)" \
```
afbjorklund: The exe would have to be removed here, as well (if it is a script)

AkihiroSuda (changes requested):
I still don't think we should expose this dev tool to end users in the same way as other `*.lima` helpers

jandubois:
> I still don't think we should expose this dev tool to end users in the same way as other `*.lima` helpers

I don't know, I thought it was useful, and think it fits the scheme of adding useful utilities with a `.lima` suffix. Especially `lima-hostagent yq` is a bit of a mouthful, but then it will probably be hidden by the `provision.mode: yq` feature anyways.

But whatever, I don't care that much, and have removed the extra commit. I only kept the change that `yq.lima.exe` would invoke `yq`, if the user makes the symlink themselves.

Issue #840, opened 2022-05-01
Show some more information about the guest system


afbjorklund:
### Description

I was looking at adding some more basic details from the guest

This was available in the "host" structure, and in /etc/os-release.

I opened another issue, about the more dynamic information:

* Issue #835 

This issue would be the more static part such as OS and total.

Similar to the output of "screenfetch" or "neofetch", but less clutter.

It is available through shell commands, but could use some JSON ?

* https://github.com/afbjorklund/lima-gui/issues/12

Example info:
* OS
* Kernel

https://pkg.go.dev/github.com/shirou/gopsutil/host

----

```console
$ limactl ls default
NAME       STATUS     SSH                ARCH      CPUS    MEMORY    DISK      DIR
default    Running    127.0.0.1:60022    x86_64    4       4GiB      100GiB    /home/anders/.lima/default
```

Some of these shell scripts are quite long...

https://github.com/KittyKatt/screenFetch

```console
anders@ubuntu:~$ wc -l screenfetch/screenfetch-dev 
6679 screenfetch/screenfetch-dev
anders@ubuntu:~$ lima screenfetch/screenfetch-dev
realpath: /proc/1945/exe: Permission denied
                          ./+o+-       anders@lima-default
                  yyyyy- -yyyyyy+      OS: Ubuntu 22.04 jammy
               ://+//////-yyyyyyo      Kernel: x86_64 Linux 5.15.0-25-generic
           .++ .:/++++++/-.+sss/`      Uptime: 46m
         .:++o:  /++++++++/:--:/-      Packages: 598
        o:+o+:++.`..```.-/oo+++++/     Shell: 
       .:+o:+o/.          `+sssoo+/    Disk: 427G / 556G (81%)
  .++/+:+oo+o:`             /sssooo.   CPU: Intel Core i5-10210U @ 4x 2.112GHz
 /+++//+:`oo+o               /::--:.   GPU: Red Hat, Inc. Virtio GPU (rev 01)
 \+/+o+++`o++o               ++////.   RAM: 463MiB / 3921MiB
  .++.o+++oo+:`             /dddhhh.  
       .+.o+oo:.          `oddhhhh+   
        \+.++o+o``-````.:ohdhhhhh+    
         `:o+++ `ohhhhhhhhyo++os:     
           .o:`.syhhhhhhh/.oo++o`     
               /osyyyyyyo++ooo+++/    
                   ````` +oo+++o\:    
                          `oo++.      
```

[...]

PR #3924, opened 2025-08-27
docs: replace `neofetch` with `uname -a`

AkihiroSuda:
neofetch is no longer maintained: https://github.com/dylanaraps/neofetch

Alternatives like fastfetch are not as famous as neofetch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,7 @@ limactl start
 
 To run Linux commands:
 ```bash
-lima sudo apt-get install -y neofetch
-lima neofetch
+lima uname -a
 ```
 
 To run containers with containerd:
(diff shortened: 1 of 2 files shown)


afbjorklund:
Or maybe we need `fetch.lima` then :-)

https://beucismis.github.io/awesome-fetch/

AkihiroSuda:
> fetch.lima

No, it is reserved for running `fetch(1)` inside an instance of Lima 😄 

https://man.freebsd.org/cgi/man.cgi?query=fetch&apropos=0&sektion=1&manpath=FreeBSD+2.1.5-RELEASE&arch=default&format=html

afbjorklund:
I was mostly referring to the `yq.lima` discussion

But I did include a fetch program in my nerdctl-toolbox

afbjorklund:
We _could_ add some command to show more basic information, though.

* https://github.com/lima-vm/lima/issues/840

jandubois (approved):
Thanks, LGTM

PR #3929, opened 2025-08-28
`docker-rootful.yaml`: Fix path to `/etc/systemd/system/docker.socket.d/override.conf`

norio-nomura:
Sorry, I made a mistake on #3892


norio-nomura:
Thanks! 🙏🏻

Issue #3930, opened 2025-08-28
Shorthand URL syntax for 3rd-party templates on GitHub


msgilligan:
### Description

It would be nice if there where some kind of short-hand syntax for referencing templates in GitHub projects, so users could do something like:

```
limactl start github:nixos-lima/nixos-lima
```

where:
 `github:nixos-lima/nixos-lima` **->**
`https://raw.githubusercontent.com/nixos-lima/nixos-lima/master/templates/default.yaml`

The details of the `github:` URL could be fairly simple or have features (like branch/tag/hash specifiers) similar to schemes in other projects.  For example:

* Nix [flake references](https://nix.dev/manual/nix/2.30/command-ref/new-cli/nix3-flake.html#url-like-syntax)
* [NPM](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?utm_source=chatgpt.com#github-urls)

afbjorklund:
I did this for the GUI

Well partially, at least

```C++

if (url.startsWith("https://github.com")) {
        // Get the "raw" YAML, and not the prettified HTTP
        url = url.replace("github.com", "raw.githubusercontent.com");
        url = url.replace("blob/", "");
}
```

Not the shorthand

jandubois:
I like the shorthand, as long as it allows appending a non-default template name as well, e.g. `github:org/repo/path/template.lima`.

We probably should then add a corresponding scheme for gitlab, for vendor neutrality.

But this makes our current template URLs even more grating because they are not RFC3986 compliant: When a URL path component starts with `//` it MUST be followed by an authority[^1]. So our template URLs should have been either `template:default` or `template:///default` to show the absence of an authority segment.

[^1]: When authority is present, the path must either be empty or begin with a slash ("/") character.  When authority is not present, the path cannot begin with two slash characters ("//").

So if we add `github:org/repo/...` URLs, then we should also support `template:name` URLs (and maybe add a warning for the old format?)

afbjorklund:
I'm fine with renaming from `template://` to `template:`, the "scheme" was always a bit confusing

jandubois:
I've added a URL scheme plugin system in #3937 and the PR summary has a sample implementation of a `github:` scheme.

The PR needs some error handling improvements, and the `github:` scheme should eventually be builtin, but you can experiment with it to figure out what the desired semantics should be for the new scheme.

msgilligan:
The `nixos-lima` repository is now updated to support PR #3937:

```
[...]

PR #3937, opened 2025-08-30
Add a plugin mechanism to provide custom locator URL schemes

jandubois:
The template reader will attempt to call `limactl-url-$SCHEME` to rewrite a custom URL as either a local filename, or a URL with a supported scheme like `https`.

This change allows us to experiment with custom schemes, like the one requested in #3930.

Example implementation for the `github:` scheme is at the end. You can use:

```
github:ORG/REPO
github:ORG/REPO/DIR/
github:ORG/REPO/DIR/FILE
github:ORG/REPO/DIR/FILE@BRANCH
```

`FILE` defaults to `lima`, and files without an extension will have `.yaml` appended.

The default branch is determined by the GitHub API if no `BRANCH` is specified.

```console
❯ limactl tmpl copy github:lima-vm/lima/templates/fedora-42 -
minimumLimaVersion: 1.1.0

base:
- template://_images/fedora-42
- template://_default/mounts
```

This implementation has a few differences to the one requested in #3930:
* uses default branch instead of hard-coded `master`
* uses `lima.yaml` instead of `templates/default.yaml` as the default file

I believe this implementation is more versatile, but we can discuss this below. The plugin format makes it easy to experiment.

Here is a minimalistic `my:` scheme to fetch default templates from the Lima repo:

```console
❯ cat ~/bin/limactl-url-my
#!/bin/bash
echo "https://raw.githubusercontent.com/lima-vm/lima/master/templates/$1.yaml"

❯ limactl tmpl copy my:opensuse-leap -
minimumLimaVersion: 1.1.0

base:
- template://_images/opensuse-leap
- template://_default/mounts
```

And here is `limactl-url-github`. Which should become a builtin scheme (implemented in Go) once we agree on the specific semantics:

```bash
#!/bin/bash

if [ $# -ne 1 ]; then
    echo "Usage: $0 [github:]ORG/REPO[/PATH][@BRANCH]" >&2
    exit 1
fi

input="$1"

# Remove optional github: prefix
if [[ "$input" == github:* ]]; then
    input="${input#github:}"
fi

# Check for explicit branch specification with @ at the end
branch=""
if [[ "$input" == *@* ]]; then
    # Split on last @ to get branch
    branch="${input##*@}"
    input="${input%@*}"
fi

# Split the input into components
IFS='/' read -ra PARTS <<< "$input"

if [ ${#PARTS[@]} -lt 2 ]; then
    echo "Error: Input must be at least ORG/REPO" >&2
    exit 1
fi

org="${PARTS[0]}"
repo="${PARTS[1]}"

# Extract path (everything after ORG/REPO)
if [ ${#PARTS[@]} -gt 2 ]; then
    path=""
    for ((i=2; i<${#PARTS[@]}; i++)); do
        if [ -n "$path" ]; then
            path="$path/"
        fi
        path="$path${PARTS[i]}"
    done
else
    path="lima"
fi

# If path ends with /, it's a directory, so append lima
if [[ "$path" == */ ]]; then
    path="${path}lima"
fi

# If the filename (last component) has no extension, add .yaml
filename="${path##*/}"
if [[ "$filename" != *.* ]]; then
    path="$path.yaml"
fi

# Query default branch if no branch was specified
if [ -z "$branch" ]; then
    branch=$(gh repo view "$org/$repo" --json defaultBranchRef --jq '.defaultBranchRef.name' 2>/dev/null)
    if [ $? -ne 0 ] || [ -z "$branch" ]; then
        echo "Error: Failed to get default branch for $org/$repo. Make sure the repository exists and you have access." >&2
        exit 1
    fi
fi

echo "https://raw.githubusercontent.com/$org/$repo/$branch/$path"
```

### DIFF
--- a/cmd/limactl/template.go
+++ b/cmd/limactl/template.go
@@ -52,13 +52,13 @@ func newValidateCommand() *cobra.Command {
 	return validateCommand
 }
 
-var templateCopyExample = `  Template locators are local files, file://, https://, or template:// URLs
+var templateCopyExample = `  Template locators are local files, file://, https://, or template: URLs
 
   # Copy default template to STDOUT
-  limactl template copy template://default -
+  limactl template copy template:default -
 
   # Copy template from web location to local file and embed all external references
-  # (this does not embed template:// references)
+  # (this does not embed template: references)
   limactl template copy --embed https://example.com/lima.yaml mighty-machine.yaml
 `
 
@@ -177,10 +177,10 @@ External references are embedded and default values are filled in
 before the YQ expression is evaluated.
 
 Example:
-  limactl template yq template://default '.images[].location'
+  limactl template yq template:default '.images[].location'
 
 The example command is equivalent to using an external yq command like this:
-  limactl template copy --fill template://default - | yq '.images[].location'
+  limactl template copy --fill template:default - | yq '.images[].location'
 `
 
 func newTemplateYQCommand() *cobra.Command {
--- a/pkg/limatmpl/locator.go
+++ b/pkg/limatmpl/locator.go
@@ -248,7 +248,10 @@ func SeemsTemplateURL(arg string) (isTemplate bool, templateName string) {
 		return false, ""
 	}
 	if u.Scheme == "template" {
-		return true, path.Join(u.Host, u.Path)
+		if u.Opaque == "" {
+			return true, path.Join(u.Host, u.Path)
+		}
+		return true, u.Opaque
 	}
 	return false, ""
 }
@@ -297,6 +300,16 @@ func TransformCustomURL(ctx context.Context, locator string) (string, error) {
 	u, err := url.Parse(locator)
 	if err != nil || len(u.Scheme) <= 1 {
 		return locator, nil
+	}
+
+	if u.Scheme == "template" {
+		if u.Opaque != "" {
+			return locator, nil
+		}
+		// Fix malformed "template:" URLs.
+		newLocator := "template:" + path.Join(u.Host, u.Path)
+		logrus.Warnf("Template locator %q should be written %q since Lima v2.0", locator, newLocator)
+		return newLocator, nil
 	}
 
 	plugin, err := plugins.Find("url-" + u.Scheme)
--- a/cmd/apptainer.lima
+++ b/cmd/apptainer.lima
@@ -4,7 +4,7 @@ set -eu
 : "${APPTAINER_BINDPATH:=}"
 
 if [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
-  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl create --name=$LIMA_INSTANCE template://apptainer\` to create a new instance" >&2
+  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl create --name=$LIMA_INSTANCE template:apptainer\` to create a new instance" >&2
   exit 1
 elif [ "$(limactl ls -f '{{ .Status }}' "$LIMA_INSTANCE" 2>/dev/null)" != "Running" ]; then
   echo "instance \"$LIMA_INSTANCE\" is not running, run \`limactl start $LIMA_INSTANCE\` to start the existing instance" >&2
--- a/cmd/docker.lima
+++ b/cmd/docker.lima
@@ -8,7 +8,7 @@ set -eu
 : "${DOCKER:=docker}"
 
 if [ "$(limactl ls -q "$LIMA_INSTANCE" 2>/dev/null)" != "$LIMA_INSTANCE" ]; then
-  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl create --name=$LIMA_INSTANCE template://docker\` to create a new instance" >&2
+  echo "instance \"$LIMA_INSTANCE\" does not exist, run \`limactl create --name=$LIMA_INSTANCE template:docker\` to create a new instance" >&2
   exit 1
 elif [ "$(limactl ls -f '{{ .Status }}' "$LIMA_INSTANCE" 2>/dev/null)" != "Running" ]; then
   echo "instance \"$LIMA_INSTANCE\" is not running, run \`limactl start $LIMA_INSTANCE\` to start the existing instance" >&2
--- a/cmd/kubectl.lima
+++ b/cmd/kubectl.lima
@@ -14,8 +14,8 @@ if [ -z "$LIMA_INSTANCE" ]; then
     LIMA_INSTANCE=k8s
   else
     echo "No k3s or k8s existing instances found. Either create one with" >&2
-    echo "limactl create --name=k3s template://k3s" >&2
-    echo "limactl create --name=k8s template://k8s" >&2
+    echo "limactl create --name=k3s template:k3s" >&2
+    echo "limactl create --name=k8s template:k8s" >&2
     echo "or set LIMA_INSTANCE to the name of your Kubernetes instance" >&2
     exit 1
   fi
--- a/cmd/limactl/completion.go
+++ b/cmd/limactl/completion.go
@@ -28,7 +28,7 @@ func bashCompleteTemplateNames(_ *cobra.Command, toComplete string) ([]string, c
 	var comp []string
 	if templates, err := templatestore.Templates(); err == nil {
 		for _, f := range templates {
-			name := "template://" + f.Name
+			name := "template:" + f.Name
 			if !strings.HasPrefix(name, toComplete) {
 				continue
 			}
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -49,7 +49,7 @@ func newCreateCommand() *cobra.Command {
   $ limactl create
 
   To create an instance "default" from a template "docker":
-  $ limactl create --name=default template://docker
+  $ limactl create --name=default template:docker
 
   To create an instance "default" with modified parameters:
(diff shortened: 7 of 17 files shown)


jandubois:
I just realized that the `limactl-url` prefix makes the names overlap with regular plugin names. Not sure if that is an issue or a feature:

```console
❯ limactl url-github lima-vm/lima
https://raw.githubusercontent.com/lima-vm/lima/master/lima.yaml

❯ limactl url-my default
https://raw.githubusercontent.com/lima-vm/lima/master/templates/default.yaml
```

If we don't want this, then we need to use a longer prefix for plugins, like `limactl-plugin-*`. Which does not match what other commands like `git` or `kubectl` do, so I would prefer to keep it as it is. Or I guess we could use `lima-url-*` as a prefix for url schemes. 🤷 I think I'm fine with the overlap.

jandubois:
I've just tested, and you can use tags or commit ids instead of branch names too. These all work:

```console
❯ limactl tmpl copy github:lima-vm/lima/examples/opensuse@v0.23.1 -
❯ limactl tmpl copy github:lima-vm/lima/examples/opensuse@74e2fda81 -
```

msgilligan:
This looks great! 

> a few differences to the one requested in #3930

I did say "something like" in the request to leave room for improvement!

jandubois:
I've updated the PR by refactoring the code to run subcommands with the `limactl` directory first on the `PATH` and also added a `limactl template url CUSTOM_URL` command (mostly for integration tests later):

```
❯ l tmpl url github:lima-vm/lima
https://raw.githubusercontent.com/lima-vm/lima/master/lima.yaml
```

AkihiroSuda:
What is the remaining task to merge this PR? CI?

jandubois:
> What is the remaining task to merge this PR? CI?

I was planning to move to `template:default` style templates (and translate the existing form with a warning to the new format), but also to move the `github:` scheme to be builtin, implemented in Go. I was hoping to get some feedback if the semantics of the bash prototype seem correct to everyone.

Either or both could happen in separate PRs to keep the PR easier to review.

jandubois:
> I was planning to move to `template:default` style templates (and translate the existing form with a warning to the new format), but also to move the `github:` scheme to be builtin, implemented in Go. I was hoping to get some feedback if the semantics of the bash prototype seem correct to everyone.

I think it will be best to create separate PRs for further changes, to keep things easy for review.

I've made one more change in this PR: I've appended `/usr/local/libexec/lima` to the end of the `PATH`, so we look for plugins in the same directory we use to store external drivers. This is based on discussion in https://github.com/lima-vm/lima/pull/3744#issuecomment-3250472833.

I've noticed that we also use `LIMA_DRIVERS_PATH` and `LIMA_TEMPLATES_PATH` to define additional search locations for drivers and templates. And we look in `$LIMA_HOME/_templates` for templates as well. I feel like this is getting out of hand and don't want to add similar mechanisms for plugins, at least not yet. Let me know if you think otherwise.

cmd/limactl/main.go
```diff
@@ -221,8 +222,11 @@ func executeWithPluginSupport(rootCmd *cobra.Command, args []string) error {
 	if len(args) > 0 {
 		cmd, _, err := rootCmd.Find(args)
 		if err != nil || cmd == rootCmd {
-			// Function calls os.Exit() if it found and executed the plugin
-			runExternalPlugin(rootCmd.Context(), args[0], args[1:])
+			_ = executil.WithExecutablePath(func() error {
```
AkihiroSuda: Why ignore err?

cmd/limactl/template.go
```diff
@@ -95,6 +96,25 @@ func fillDefaults(ctx context.Context, tmpl *limatmpl.Template) error {
 	return err
 }
 
+func newTemplateURLCommand() *cobra.Command {
```
AkihiroSuda: Needs docs and tests

AkihiroSuda: Docs can be added under https://deploy-preview-3972--lima-vm.netlify.app/docs/config/plugin after merging:
- #3972

jandubois: This PR is also modifying the same logic that #3973 is currently changing.

And tests should be implemented using BATS from #3969.

I'm turning this back unto "Draft" until these 3 PRs have landed because it will likely need serious refactoring by then.

jandubois:
I just created another URL scheme for testing: `instance:` to reference the template of an already created instance:

```bash
#!/bin/bash
echo "${LIMA_HOME:-$HOME/.lima}/$1/lima.yaml"
```

With this you can create another instance with identical config like this (but unlike the `clone` command you start with a fresh disk):

```console
❯ limactl create -y --name clone instance:default
```

And of course it works all the usual places: `limactl tmpl validate instance:clone` or `limactl tmpl copy instance:default -` etc.

Not sure though if it is useful enough to be builtin.

AkihiroSuda:
> instance:

This is quite confusing; `limactl start instance:NAME` should just mean starting an instance named `NAME` (i.e., `limactl start NAME`)

jandubois:
> > instance:
> 
> This is quite confusing; `limactl start instance:NAME` should just mean starting an instance named `NAME` (i.e., `limactl start NAME`)

Yeah, let's not include it in Lima. I think the only scheme we want to have builtin is `github`.

AkihiroSuda:
Needs rebase

jandubois:
> ```
> github:ORG/REPO
> github:ORG/REPO/DIR/
> github:ORG/REPO/DIR/FILE
> github:ORG/REPO/DIR/FILE@BRANCH
> ```

Extension ideas:

* `github:ORG` is the same as `github:ORG/lima.yaml` or `github:ORG/ORG` if `lima.yaml` repo doesn't exist
* `github:ORG//` is the same as `github:ORG`, but you can now append a DIR, FILE, and BRANCH: `github:mylongprojectname//templates/lima.yaml`

jandubois:
I think this PR is ready for review.

Still missing, but could be done in several separate PRs:
* Implement builtin `github:` scheme in Go
* Switch to `template:` scheme, but keep `template://` working for backwards compatibility
* Add BATS tests for plugins in general and url-* plugins in particular
* Explain url-* plugins in the docs

jandubois:
> Still missing, but could be done in several separate PRs:
> 
> * Switch to `template:` scheme, but keep `template://` working for backwards compatibility

I've made that change as well now (in a separate commit), but I think this PR is getting too big. Can we get this reviewed and merged, and do the remaining tasks in separate PRs?

pkg/limatmpl/locator.go
```diff
@@ -285,3 +295,55 @@ func InstNameFromYAMLPath(yamlPath string) (string, error) {
...
+		return locator, nil
+	}
+
+	if u.Scheme == "template" {
+		if u.Opaque != "" {
+			return locator, nil
+		}
+		// Fix malformed "template:" URLs.
+		newLocator := "template:" + path.Join(u.Host, u.Path)
+		logrus.Warnf("Template locator %q should be written %q", locator, newLocator)
```
AkihiroSuda:
```suggestion
		logrus.Warnf("Template locator %q should be written %q since Lima v2.0", locator, newLocator)
```

AkihiroSuda: This needs to be updated too
https://github.com/lima-vm/lima/blob/94ba0e1c742bad6cd76452b07f2161bc9d720971/cmd/limactl/completion.go#L27-L49

jandubois: I updated `cmd/`, `hack/`, and `README.md`.

I've not yet updated `templates/` and `website/`. I assume we only want to update these close to the release, or should I update everything right away?

README.md
```diff
@@ -39,14 +39,14 @@ lima nerdctl run --rm hello-world
 
 To run containers with Docker:
 ```bash
-limactl start template://docker
+limactl start template:docker
```
AkihiroSuda: The old form should be kept until people actually begin using v2

pkg/limatmpl/locator.go
jandubois: > I've not yet updated `templates/`

And I had to revert `hack/test-upgrades.sh` to keep the command compatible with the old Lima version.

README.md
jandubois: I kind of disagree. The website should document the released version, but the `README.md` should correspond to the state of the branch.

Otherwise we could also not close issues until their PRs have been included in a new release, which would make the whole workflow really awkward.

I'll revert for now, but after 2.0 we should rethink our workflow, and also figure out how to have a version of the docs for the last release, and a different version for the head of `master`.

AkihiroSuda: Regardless to our intention, people will just refer to the README in the default branch and will continue opening "I followed README but it doesn't work" issues  when the content conflicts with the latest release, so we have to care about them

jandubois: > Regardless to our intention, people will just refer to the README in the default branch

I think the way to deal with that is to remove end-user instructions from the README and point them to the docs. There is no reason to repeat usage documentation in it.

---

# README

Lima is Linux Machines running in a VM...

## How to install and use?

Please read the \[Documentation\] and follow the \[Installation\] instructions.

## Community channels

* New releases on GitHub
* GitHub Discussions and Issues
* Slack
* Community meetings (maybe?)
* Social media (if we start posting there)

## Developing Lima

### Required tools and their version

### How to build

### How to run tests

# License

AkihiroSuda: I still prefer to keep the "TLDR" abstract in `README.md`.
First comers do not want to access the website when they do not even know what it is all about.

AkihiroSuda:
On the second thought, a locator plugin should return the YAML contents, not HTTPS URL, so as to support non-HTTP and non-local remotes

PR #3938, opened 2025-08-30
Use errors.As() instead of type assertions

jandubois:
It is more idiomatic because it also works for wrapped errors.


AkihiroSuda (approved):
Thanks

PR #3939, opened 2025-08-31
`hostagent`: Fix: `ha.sock` was not removed on exit

norio-nomura:
Change to close the `http.Server` explicitly, and it will close the `net.UnixListener`. The `net.UnixListener` will then unlink `ha.socket` on close.

### DIFF
--- a/cmd/limactl/hostagent.go
+++ b/cmd/limactl/hostagent.go
@@ -125,12 +125,11 @@ func hostagentAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	go func() {
-		defer os.RemoveAll(socket)
-		defer srv.Close()
 		if serveErr := srv.Serve(l); serveErr != http.ErrServerClosed {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
>  The net.UnixListener will then unlink ha.socket on close.

This behavior doesn't seem documented in https://pkg.go.dev/net#UnixListener

norio-nomura:
> > The net.UnixListener will then unlink ha.socket on close.
> 
> This behavior doesn't seem documented in https://pkg.go.dev/net#UnixListener

https://pkg.go.dev/net#UnixListener.SetUnlinkOnClose
>SetUnlinkOnClose sets whether the underlying socket file should be removed from the file system when the listener is closed.
>The default behavior is to unlink the socket file only when package net created it.

AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

PR #3858, opened 2025-08-18
feature: use Rosetta AOT Caching with CDI

norio-nomura:
## Description
This change introduces device configuration to enable Rosetta AOT Caching in Docker VMs.

- Modify Rosetta Caching Options from Abstract Socket to Unix Domain Socket:  
  Unix Domain Socket can be mounted within a container using the Container Device Interface (CDI) mechanism.  
  This requires merging the following pull request: https://github.com/Code-Hex/vz/pull/195.

- Register Rosettad AOT Caching Daemon as a service:
  - `/etc/systemd/system/rosettad.service` on systemd
  - `/etc/init.d/rosettad` on OpenRC

- Add "lima-vm.io/rosetta=cached" device specification to `{~/.config,/etc}/cdi/rosetta.yaml`  
  see: https://github.com/cncf-tags/container-device-interface/blob/main/SPEC.md

- Add `{~/.config,/etc}/docker/daemon.json` to `docker{,-rootful}.yaml`
  - `.features.cdi = true` to enable CDI

- Add `enable_cdi = true` to `{~/.config,/etc}/containerd/config.toml`

To enable Rosetta AOT Caching in docker, use `--device=lima-vm.io/rosetta=cached`.
see: https://docs.docker.com/build/building/cdi/

## Benchmark
1. Setup docker VM
```console
$ _output/bin/limactl start --name docker-test template://docker --memory 16 --rosetta --tty=false --log-level warn
WARN[0000] vmType vz: ignoring [VMOpts]                 
3.50 GiB / 3.50 GiB [---------------------------------------] 100.00% 1.54 GiB/s
To run `docker` on the host (assumes docker-cli is installed), run the following commands:
------
docker context create lima-docker-test --docker "host=unix:///Users/norio/.lima/docker-test/sock/docker.sock"
docker context use lima-docker-test
docker run hello-world
------
$ docker context create lima-docker-test --docker "host=unix:///Users/norio/.lima/docker-test/sock/docker.sock"
lima-docker-test
Successfully created context "lima-docker-test"
$ docker context use lima-docker-test
lima-docker-test
Current context is now "lima-docker-test"
```
2. Confirm Rosetta AOT Caching is working (`/var/cache/rosettad/*.aotcache` are created)
```console
$ limactl shell docker-test ls -la /var/cache/rosettad
total 12
drwxr-xr-x  3 root root 4096 Aug 18 18:57 .
drwxr-xr-x 15 root root 4096 Aug 18 18:57 ..
[...]

PR #3941, opened 2025-09-02
`docker{,-rootful}.yaml`: Enable the containerd image store by default

norio-nomura:
~~To support building multi-platform images.~~
~~Use `--set .param.containerdSnapshotter=false` flag to disable it.~~

https://github.com/lima-vm/lima/pull/3858#discussion_r2282194496
> https://docs.docker.com/desktop/features/containerd/
> > The containerd image store is enabled by default in Docker Desktop version 4.34 and later, but only for clean installs or if you perform a factory reset.
>
> 4.34 was released on 2024-08-29
> https://docs.docker.com/desktop/release-notes/#4340

edit: remove "To support...", and "Use ..."

### DIFF
--- a/templates/docker-rootful.yaml
+++ b/templates/docker-rootful.yaml
@@ -68,4 +68,4 @@ message: |
   docker run hello-world
   ------
 param:
-  containerdSnapshotter: false
+  containerdSnapshotter: true
--- a/templates/docker.yaml
+++ b/templates/docker.yaml
@@ -75,4 +75,4 @@ message: |
   docker run hello-world
   ------
 param:
-  containerdSnapshotter: false
+  containerdSnapshotter: true


norio-nomura:
> For this purpose, using containerd image store is not necessary, as it is supported by buildx

Removed those lines from commit message.

jandubois:
> For this purpose, using containerd image store is not necessary, as it is supported by buildx

I don't think so, how would it even work? AFAIK the "classic" image store doesn't support multi-platform manifests.

And `buildx` refuse to build for me:

```console
❯ docker buildx build --platform linux/arm/v7,linux/arm64/v8,linux/amd64 --tag multiarch .
[+] Building 0.0s (0/0)                                                                                    docker:lima-docker
ERROR: failed to build: Multi-platform build is not supported for the docker driver.
Switch to a different driver, or turn on the containerd image store, and try again.
Learn more at https://docs.docker.com/go/build-multi-platform/
```

jandubois (approved):
Thanks, LGTM

AkihiroSuda:
> I don't think so, how would it even work? AFAIK the "classic" image store doesn't support multi-platform manifests.

You can't store multi-platform manifests, but you can still build and push them with a custom buildx instance.
(`docker buildx create --use`  and `docker buildx build --push`)

norio-nomura:
Thanks! 🙏🏻

PR #3947, opened 2025-09-03
editflags: add `--mount-only` flag

AkihiroSuda:
Similar to `--mount`, but overrides the existing mounts.

e.g.,
```
limactl start --mount-only "$(pwd)"
````

This will only make the current directory visible to the VM. The host home directory is not visible (unless pwd is the home).

### DIFF
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -44,6 +44,8 @@ func RegisterEdit(cmd *cobra.Command, commentPrefix string) {
 	})
 
 	flags.StringSlice("mount", nil, commentPrefix+"Directories to mount, suffix ':w' for writable (Do not specify directories that overlap with the existing mounts)") // colima-compatible
+	flags.StringSlice("mount-only", nil, commentPrefix+"Similar to --mount, but overrides the existing mounts")
+
 	flags.Bool("mount-none", false, commentPrefix+"Remove all mounts")
 
 	flags.String("mount-type", "", commentPrefix+"Mount type (reverse-sshfs, 9p, virtiofs)") // Similar to colima's --mount-type=(sshfs|9p|virtiofs), but "reverse-sshfs" is Lima is called "sshfs" in colima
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/limactl/editflags/editflags.go [hunk 1]
```diff
@@ -207,30 +223,52 @@ func YQExpressions(flags *flag.FlagSet, newInstance bool) ([]string, error) {
...
-					loc := strings.TrimSuffix(s, ":w")
-					expr += fmt.Sprintf(`{"location": %q, "writable": %v}`, loc, writable)
-					if i < len(ss)-1 {
-						expr += ","
-					}
+				mountListExpr, err := buildMountListExpression(ss)
+				if err != nil {
+					return "", err
+				}
+				expr := `.mounts += ` + mountListExpr + ` | .mounts |= unique_by(.location)`
```
jandubois, line 230: I know this is just existing code that is being moved, but the logic is incorrect: the mounts need to be unique by `.mountPoint`, not by `.location`. The same location can be mounted to several mount points, and this logic would delete all but one of them.

But the mount point may only be filled in later (defaults to location), so we can't do anything here beyond making sure that the new mount points are defined first, so they will overrride existing mounts (mostly relevant only for the `writable` property).

I think this would be better (but have not tested it):

```suggestion
				// mount options take precedence over template settings
				expr := fmt.Sprintf(".mounts = %s + .mounts", mountListExpr)
```

Could also open a new issue to address later.

cmd/limactl/editflags/editflags.go
```diff
@@ -207,30 +223,52 @@ func YQExpressions(flags *flag.FlagSet, newInstance bool) ([]string, error) {
...
+			func(_ *flag.Flag) (string, error) {
+				incompatibleFlagNames := []string{"mount", "mount-only"}
+				for _, name := range incompatibleFlagNames {
+					ss, err := flags.GetStringSlice(name)
+					if err != nil {
+						return "", err
+					}
+					if len(ss) > 0 {
+						return "", errors.New("flag `--mount-none` conflicts with `" + name + "`")
+					}
```
jandubois, line 271: This could potentially be a shared helper function because we have other subcommands that have mutually exclusive options.

AkihiroSuda: I guess we can add a helper when we have move conflicting options

cmd/limactl/editflags/editflags.go
AkihiroSuda: - https://github.com/lima-vm/lima/issues/3959

PR #3949, opened 2025-09-03
`vm_darwin.go`: Fix: `default_ep.sock` was not removed on exit

norio-nomura:
Change to cancel the context that passed to calling `usernet.StartGVisorNetstack` when the VM stopped. `default_fd.sock` won’t be deleted by this change, but the reason is unknown.


norio-nomura:
Thanks! 🙏🏻

PR #3950, opened 2025-09-03
mounts: prohibit relative paths in YAML

### LINKED ISSUE #3948: Mounts: prohibit relative paths in YAML (while allowing them in `limactl create --mount` CLI)
The current YAML surprisingly allows relative paths (maybe a regression):
```
mounts:
- location: .
```

The actual path changes depending on the cwd of the `limactl start` command.

In Lima v2.0, these relative paths should be prohibited to make the command deterministic.

`limactl create --mount DIR` should continue to support relative paths.
A relative path should be resolved into an absolute path at the time of `limactl create`.

AkihiroSuda:
YAMLs with relative paths like `mounts: [{location: .}]` are not longer allowed.

Relative paths are still allowed in CLI: `limactl create --mount=DIR`

Fix #3948

- - -
This PR also contains:
- #3947


jandubois:
I've not looked in detail, but returning an error from `FillDefaults` seems wrong. This is a validation error that should be caught in `Validate`. Otherwise there is no reason to have 2 separate functions.

~I also think the `--mount-only` option should be a separate PR and not mixed in with a bug fix.~

EDIT: I see it is already a separate PR, and just stacked with this one.

AkihiroSuda:
Updated to avoid returning err from `FillDefault`

jandubois (changes requested):
This is still not the right fix. The regression was introduced by #3339. It unconditionally expands all local filenames to absolute names. We should only call `Expand()` when the location actually is a tilde-path:

```diff
--- pkg/limayaml/defaults.go
+++ pkg/limayaml/defaults.go
@@ -713,10 +713,12 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 				mounts[i].NineP.Cache = ptr.Of(Default9pCacheForRO)
 			}
 		}
-		if location, err := localpathutil.Expand(mount.Location); err == nil {
-			mounts[i].Location = location
-		} else {
-			logrus.WithError(err).Warnf("Couldn't expand location %q", mount.Location)
+		if localpathutil.IsTildePath(mount.Location) {
+			if location, err := localpathutil.Expand(mount.Location); err == nil {
+				mounts[i].Location = location
+			} else {
+				logrus.WithError(err).Warnf("Couldn't expand location %q", mount.Location)
+			}
 		}
 		if mount.MountPoint == nil {
 			mountLocation := mounts[i].Location
```

This fixes the existing validation:

```console
❯ cat mount.yaml
images: [{location: /foo}]
mounts: [{location: foo}]

❯ l tmpl validate mount.yaml
FATA[0000] failed to validate YAML file "mount.yaml": field `mounts[0].location` must be an absolute path, got "foo"
```

pkg/limayaml/defaults.go
```diff
@@ -713,6 +713,15 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 				mounts[i].NineP.Cache = ptr.Of(Default9pCacheForRO)
 			}
 		}
+		// The location must be an absolute path or a path that begins with `~`.
+		// This condition should have been validated before reaching here in Load().
```
jandubois: This assumption is not correct. Validation happens **after** FillDefaults.

AkihiroSuda: This assumption was not about `Validate()`.
The assumption is that `FillDefault()` is usually called via `Load()`, which has a minimal pre-flight validation.

The relevant code is now removed.

jandubois: > This assumption was not about `Validate()`. The assumption is that `FillDefault()` is usually called via `Load()`, which has a minimal pre-flight validation.

Your comment says: "This condition should have been validated before reaching here in Load()."

I just pointed out that this is wrong. `Load()` is not supposed to validate at all, not before calling FillDefault(), and not after:

```
// Load loads the yaml and fulfills unspecified fields with the default values.
```

There is unfortunately a call to `validateParamIsUsed()` in `load()`, which is conceptually wrong. It is still not clear to me if this check should be moved, or just removed without replacement, but I don't want to touch it right now.

> The relevant code is now removed.

Thank you, will take another look.

jandubois (approved):
Thanks, LGTM

PR #3951, opened 2025-09-03
templates: default: remove /tmp/lima

### LINKED ISSUE #3648: docs: discourage `/tmp/lima` on shared hosts (and deprecate it?)

AkihiroSuda:
`/tmp/lima` is no longer mounted by default, so as to allow multiple users to run Lima instances independently.

Fix #3648

To restore the previous behavior, add the following to your YAML config:

```yaml
mounts:
- location: "{{.GlobalTempDir}}/lima"
  mountPoint: /tmp/lima
  writable: true
```

or, use the following CLI flag:

```bash
limactl create --mount /tmp/lima:w
```

### DIFF
--- a/website/content/en/docs/config/_index.md
+++ b/website/content/en/docs/config/_index.md
@@ -10,7 +10,7 @@ The current default spec:
 - CPU: 4 cores
 - Memory: 4 GiB
 - Disk: 100 GiB
-- Mounts: `~` (read-only), `/tmp/lima` (writable)
+- Mounts: `~` (read-only), `/tmp/lima` (writable; removed in Lima v2.0)
 - SSH: 127.0.0.1:<Random port>
 
 For environment variables, see [Environment Variables](./environment-variables/).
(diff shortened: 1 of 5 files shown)


website/content/en/docs/config/_index.md
```diff
@@ -10,7 +10,7 @@ The current default spec:
 - CPU: 4 cores
 - Memory: 4 GiB
 - Disk: 100 GiB
-- Mounts: `~` (read-only), `/tmp/lima` (writable)
+- Mounts: `~` (read-only), `/tmp/lima` (writable; removed in Lima v2.0)
```
jandubois, line 13: The comment about `/tmp/lima` should be removed when 2.0 is released.

PR #3956, opened 2025-09-03
make check-generated: check reproducibility of *.pb.go too

### LINKED ISSUE #3953: `make check-generated` should check reproducibility of `*.pb.go`, not just `*.pb.desc`
https://github.com/lima-vm/lima/blob/dc733cf1cad7210ab0ede06846bdb1b99108d031/.github/workflows/test.yml#L38-L39

https://github.com/lima-vm/lima/blob/dc733cf1cad7210ab0ede06846bdb1b99108d031/Makefile#L502-L506

`make check-generated` should check reproducibility of `*.pb.go` too, as such machine-generated codes are well-known hot spots to hide malicious contribution.

Our repo does **not** seem compromised so far, so I'm not treating this issue as a vulnerability.
- https://github.com/lima-vm/lima/commits/master/pkg/guestagent/api/guestservice.pb.go
- https://github.com/lima-vm/lima/commits/master/pkg/guestagent/api/guestservice_grpc.pb.go
- https://github.com/lima-vm/lima/commits/master/pkg/driver/external/driver.pb.go
- https://github.com/lima-vm/lima/commits/master/pkg/driver/external/driver_grpc.pb.go

AkihiroSuda:
Fix #3953


AkihiroSuda:
Needs to pin the version of protoc ?
```
-// - protoc             v6.32.0
+// - protoc             v3.21.12
```

- https://github.com/golang/protobuf/issues/1185

- - -

containerd pins the protoc version like this: https://github.com/containerd/containerd/blob/v2.1.4/script/setup/install-protobuf

AkihiroSuda:
Alternatively we can consider this
```
find . -name '*.pb.go' | xargs perl -pi -E 's/protoc[[:blank:]]+v[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+/protoc version omitted for repro builds/g'
```

Makefile
```diff
@@ -501,9 +501,8 @@ uninstall:
 
 .PHONY: check-generated
 check-generated:
-	@test -z "$$(git status --short | grep ".pb.desc" | tee /dev/stderr)" || \
-		((git diff $$(find . -name '*.pb.desc') | cat) && \
-		(echo "Please run 'make generate' when making changes to proto files and check-in the generated file changes" && false))
+	git diff --exit-code || \
```
nirs, line 504: How does this work? this checks that there are no changes in the entire code, but it should check only the generated files.

hack/gogenerate/protoc.sh
```diff
@@ -0,0 +1,29 @@
...
+# Generate Go code from a .proto file.
+# Expected to be called from `//go:generate` directive.
+
+set -eu
+if [ "$#" -ne 1 ]; then
+	echo >&2 "Usage: $0 FILE"
+	exit 1
+fi
+
+PROTO="$1"                         ## a.proto
```
nirs, line 15: This is the source?

hack/gogenerate/protoc.sh
```diff
@@ -0,0 +1,29 @@
...
+BASE="$(basename "$PROTO" .proto)" ## a
+PB_DESC="${BASE}.pb.desc"          ## a.pb.desc
+PB_GO="${BASE}.pb.go"              ## a.pb.go
+GRPC_PB_GO="${BASE}_grpc.pb.go"    ## a_grpc.pb.go
```
nirs, line 19: These are the generated files?

    a.proto -> a.pb.desc, a.pb.go, a_grpc.pb.go

So when we make a change in a.proto, we expect changes in a.pb.desc, a.pb.go, a_grpc.pb.go, right?

And if there was no change in a.proto, there should be no change in the output files?

Makefile
AkihiroSuda: The entire code has to be checked, as a malicious contributor may inject `//go:generate` that modifies non-pb files

hack/gogenerate/protoc.sh
AkihiroSuda: Yes

hack/gogenerate/protoc.sh
AkihiroSuda: Right

Makefile
nirs: OK, so how `make check-generated` validates that no files was changed by `make generate`?

It seems that the current target assumes that you run it like this:

    git clone ...
    cd lima
    make generate
    make check-generated

This re-generate all generated files, and ensures that all generated files are identical to generated files with are part of the source.

But if you run this like this:

    git clone ...
    cd lima
    make check-generated

It will succeed even if the checked out branch was compromised.

This should work so there is no way to use the check incorrectly. This will also keep the logic of the check in one place and simplify CI.

AkihiroSuda: > It will succeed even if the checked out branch was compromised.

This is a misusage of check-generated.
Any change to GHA YAMLs have to be reviewed carefully.

jandubois (approved):
Thanks, LGTM

PR #3961, opened 2025-09-04
driverutil: allow external driver stderr

### LINKED ISSUE #3960: external driver: qemu fails with `failed to resolve vm`
```console
$ limactl start --name qemu --vm-type qemu 
? Creating an instance "qemu" Proceed with the current configuration
FATA[0000] failed to resolve vm for "/Users/suda/.lima/qemu/lima.yaml": external driver stderr: time="2025-09-04T16:50:33+09:00" level=info msg="QEMU binary \"/opt/homebrew/bin/qemu-system-aarch64\" seems properly signed with the \"com.apple.security.hypervisor\" entitlement" 
```

vz (external), vz (internal), and qemu (internal) still work.

- Lima version: master (e963c321e9635c2753701ed6afd66058a02b93cd)
- QEMU version: 10.1.0 , installed via Homebrew
- Host: macOS 15.6.1 ARM

AkihiroSuda:
Fix #3960

### DIFF
--- a/pkg/driverutil/vm.go
+++ b/pkg/driverutil/vm.go
@@ -90,7 +90,7 @@ func handlePreConfiguredDriverAction(ctx context.Context, y *limatype.LimaYAML, 
 	}
 
 	if stderrBuf.Len() > 0 {
-		return fmt.Errorf("external driver stderr: %s", stderrBuf.String())
+		logrus.Debugf("external driver stderr: %s", stderrBuf.String())
 	}
 
 	*y = res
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
cc @unsuman

unsuman:
Thanks

jandubois (approved):
Thanks, LGTM

Only question is if this should be Info instead of Debug?

But to display it by default, the original message would need to be unwrapped, otherwise it looks too ugly:

```json
{"level":"debug","msg":"external driver stderr: time=\"2025-09-04T19:54:43-07:00\" level=info msg=\"QEMU binary \\\"/opt/homebrew/bin/qemu-system-aarch64\\\" seems properly signed with the \\\"com.apple.security.hypervisor\\\" entitlement\"\n","time":"2025-09-04T19:54:43-07:00"}
```

I guess it is ok to swallow STDERR as long as the exit code was 0.

AkihiroSuda:
> Only question is if this should be Info instead of Debug?

The log level should be rather propagated via `PropagateJSON`
https://github.com/lima-vm/lima/blob/e963c321e9635c2753701ed6afd66058a02b93cd/pkg/logrusutil/logrusutil.go#L16-L19

PR #3972, opened 2025-09-05
docs: split long pages, fix minor errors, etc.

AkihiroSuda:
See individual commits

Preview: https://deploy-preview-3972--lima-vm.netlify.app/docs/


website/content/en/docs/config/plugin/vm.md
```diff
@@ -0,0 +1,7 @@
+---
+title: VM driver plugins
```
jandubois, line 2: I would call them just "drivers" and not "driver plugins" to avoid confusion.

PR #3973, opened 2025-09-05
limactl help and info flag should show available plugins

olamilekan000:
```
change adds functionality to display Plugins in the info and list commands
```

Fixes https://github.com/lima-vm/lima/issues/3608#issuecomment-3257711223


pkg/plugin/plugin.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func UpdatePathForPlugins() error {
+	exe, err := os.Executable()
+	if err != nil {
+		return err
+	}
+
+	binDir := filepath.Dir(exe)
+	prefixDir := filepath.Dir(binDir)
```
AkihiroSuda: use
https://github.com/lima-vm/lima/blob/4b46444446b824d39bde52ace347a64e1b5198e1/pkg/usrlocalsharelima/usrlocalsharelima.go#L170

jandubois:
I'm wondering if the plugins could also be added to tab completion (could be in a separate PR if it is more complex).

olamilekan000:
> I'm wondering if the plugins could also be added to tab completion (could be in a separate PR if it is more complex).

It can be done in another PR

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,203 @@
...
+				plugins = append(plugins, plugin)
+				seen[plugin.Name] = true
+			}
+		}
+	}
+
+	return plugins, nil
+}
+
+func getPluginDirectories() []string {
```
jandubois: This function is not adding the os.Executable() directory in the front of the PATH.

And it is adding the libexec directory at the front; it should be at then back, so users can override it by putting something else on the PATH.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+
+	if filepath.Ext(path) == ".exe" {
+		return true
+	}
+
+	return false
+}
+
+func getPluginDescription(path string) string {
```
jandubois: This doesn't seem like a very useful way to find a description; the best it can deliver is "Alias for foo", which might just as well be wrong because it likely has some modified behaviour too.

If we want to provide descriptions, then I think we need to define some kind of comment pattern we can scan for. And since the plugin can be a compiled binary, we would need to scan the whole file for it because the string would be in some kind of data segment in the file.

So I would suggest to use some kind of delimiters that don't look to horrible in a shell script, and would also be unique in a compiled executable. E.g.

```
# <limactl> Run MCP servers in a sandbox </limactl>
```

Trim whitespace from the text between delimiters and you have your description.

This is just a suggestion; better ideas welcome!

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,203 @@
...
+		if len(matches) > 1 {
+			commandName := matches[1]
+			return "Alias for " + commandName
+		}
+	}
+
+	return ""
+}
+
+func UpdatePathForPlugins() error {
```
jandubois: This function should reuse the `getPluginDirectories()` function and not compute it again independently because then the 2 implementations can get out of sync.

This could be just

```go
pluginDirs := getPluginDirectories()
os.Setenv("PATH", strings.Join(string(filepath.ListSeparator), pluginDirs))
```

pkg/plugin/plugin.go
olamilekan000: I did think about this, but I wasn't sure if it was necessary since it's just an Alias which was why I sort of wen with the "Alias for foo" approach. But since you've mentioned it, I'll just add it

jandubois: I've thought about this some more, and think my suggestion is too much of a hack.

I think this is better:

* Call `limactl-PLUGIN --help`
* If the exit code is not zero → no description
* If the first line of the output contains the string `--help` → no description
* The description is the first line of the output

This automatically makes `limactl PLUGIN --help` too, and even piggy-backs on existing commands, if they are just aliases:

```console
❯ limactl ls --help | head -1
List instances of Lima.
```

So if the plugin is just an alias, it will use the first line of the help of the original command as the description.

Sorry if you already started work on the delimiter-based approach, but I think this one will be better.

jandubois: > * If the first line of the output contains the string `--help` → no description

Actually, I would reject the description if it contains any of these strings:

```
--
invalid
illegal
unrecognized 
unexpected
unknown
missing
too many
too few
not found
not supported
usage:
error
failed
```

This should catch the case where the plugin just passes the `--help` argument to another tool that doesn't implement it.

AkihiroSuda: It might be considered vulnerable to execute all `limactl-*` binaries under `libexec` and `$PATH` by default?
Maybe we can depend on xattr?

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	pathEnv := os.Getenv("PATH")
+	if pathEnv != "" {
+		pathDirs := filepath.SplitList(pathEnv)
+		dirs = append(dirs, pathDirs...)
+	}
+
+	if prefixDir, err := usrlocalsharelima.Prefix(); err == nil {
+		libexecDir := filepath.Join(prefixDir, "libexec", "lima")
```
AkihiroSuda: https://lima-vm.io/docs/config/plugin/cli/#how-it-works needs to be updated to cover libexec

pkg/plugin/plugin.go
jandubois: Outside the scope of this PR, but I wish this worked:

```console
❯ limactl shell default --help | head -1
/bin/bash: --: invalid option
Usage:	/bin/bash [GNU long option] [option] ...
	/bin/bash [GNU long option] [option] script-file ...
GNU long options:
...
```

But it just runs `bash -c '--help'` inside the VM. All the flags for `shell` must come before the INSTANCE name. Maybe this can be changed?

jandubois: > It might be considered vulnerable to execute all `limactl-*` binaries under `libexec` and `$PATH` by default?

Vulnerable to what? Any attack that can place files into a directory on your `PATH` can also replace other files in those directories. It can create `/usr/local/bin/ls` as a wrapper that then invokes `/bin/ls`.

So it doesn't seem to be much of an attack vector.

And the name seems specific enough that nobody would accidentally create a command with a name like that without it being related to `limactl`.

> Maybe we can depend on xattr?

Seems to complicated and obscured for regular users.

AkihiroSuda: > Any attack that can place files into a directory on your PATH can also replace other files in those directories. 

`curl -O /usr/local/bin/limactl-malicious` shouldn't immediately result in executing the malicious code at the time of `limactl --help` or `limactl info`.

jandubois: > `curl -O /usr/local/bin/limactl-malicious` shouldn't immediately result in executing the malicious code at the time of `limactl --help` or `limactl info`.

How is this different from `limactl start` or `limactl shell` immediately executing `/usr/local/bin/ssh` after it has been installed with a malicious version?

Any time we run a subprocess and locate the command via the `PATH` and don't make sure the file and all parent directories are only writable by `root` we run the same risk.

jandubois: For an attacker there are endless openings once they can write to your local disk:

```console
❯ cat /usr/local/bin/ls
#!/bin/sh
echo "Hey, look at me!"
exec /bin/ls "$@"

❯ ls -lh /usr/local/bin/ls
Hey, look at me!
-rwxr-xr-x@ 1 jan  staff    52B Sep  5 21:05 /usr/local/bin/ls
```

jandubois: To follow this argument all the way to the end: how can you trust that `limactl` itself has not been replaced by a malicious version?

So I think there is no good reason not to run `limactl-PLUGIN --help`. The main concern is that you run an executable that is not supposed to be a plugin, but I think the naming scheme makes this extremely unlikely. So we should be good.

AkihiroSuda: At least we should document the behavior.
Maybe we should also consider marking the CLI plugin experimental for now

jandubois: > At least we should document the behavior.

Yes

> Maybe we should also consider marking the CLI plugin experimental for now

Yes. Also the --preserve-env should be experimental so we can still break backwards compatibility.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+	}
+
+	return plugins, nil
+}
+
+func getPluginDirectories() []string {
+	var dirs []string
+
+	if exe, err := os.Executable(); err == nil {
```
jandubois: It just occurred to me that we probably want the logic from `usrlocalsharelima.Dir()` here: call `executableViaArgs0()` and fall back on `os.Executable()` if it fails.


If `limactl` is invoked via a symlink, then we want to locate additional plugins via the directory of the symlink and not via the directory where the executable itself is installed.

This is only relevant to Linux; on macOS `os.Executable` already returns the symlink path.

pkg/plugin/plugin.go
jandubois: Just to summarize: I believe we have settled on using the `--help` approach (including documentation).

olamilekan000: @jandubois 

1. Call limactl-PLUGIN --help
2. If the exit code is not zero → no description
3. If the first line of the output contains the string --help → no description
4. The description is the first line of the output

I have a little issue with step4. In the case where the plugin logs an item before before the actual command is executed, it won't be accurate, and we might end up displaying something inaccurate in the description.
Here's an example.

```sh

#!/bin/sh
# Sample limactl-shell alias that shows running instances
# <limactl> Run shell script </limactl>
echo "Running limactl-shell alias with arguments: $@"
limactl shell "$@"
```
The first item that'd be logged when this plugin is executed will be 

`Running limactl-shell alias with arguments: --help
`

And then we end up with something like

```sh
Available Plugins (Experimental):
  sh       Running limactl-shell alias with arguments: --help
```

jandubois: Well, step 3 says if the line contains the string `--help` then it is not a description, so this would have no description because your first line ends with `--help`.

Note that matching just `--help` has been replaced in https://github.com/lima-vm/lima/pull/3973#discussion_r2326417921 with matching any string from this list:

```
--
invalid
illegal
unrecognized 
unexpected
unknown
missing
too many
too few
not found
not supported
usage:
error
failed
```

So the first pattern in the list (`--`) would have rejected your output as well.

olamilekan000: Oh, I missed the `--`. But it means if it's just a plain echo statement without any arg and it doesn't contain any of the restricted words, It'd treated as a description even though it's incorrect.

jandubois: > It'd treated as a description even though it's incorrect.

It is a heuristic, and by definition that means that sometimes it will be wrong.

I think this is a reasonable compromise. You are more likely to get a description if you make it easy for the plugin author.

As a side effect `limactl PLUGIN --help` will automatically work too.

And typically the end user is creating/installing the plugin (it is an extension mechanism), so they can fix the issue easily.

The alternative of requiring some special markup before/after the description feels clumsy in comparison.

What other alternatives are there?

AkihiroSuda (approved):
Let's merge this, the CLI help strings can be addressed in another PR

AkihiroSuda:
Needs rebase

olamilekan000:
@AkihiroSuda can you help reopen this pr. It was closed by mistake after rebase.

AkihiroSuda:
> @AkihiroSuda can you help reopen this pr. It was closed by mistake after rebase.

Can't reopen, could you open a new one?

olamilekan000:
I have opened another one [here](https://github.com/lima-vm/lima/pull/4009)

PR #3975, opened 2025-09-05
Fix: make templates specify OS compatibility

### LINKED ISSUE #3974: Templates should specify their OS compatibility
For example:

```zsh
➜  lima2 git:(master) ./_output/bin/limactl prune --keep-referred
FATA[0001] failed to resolve vm for "experimental/wsl2": vmType "wsl2" is not a registered driver 
```

It doesn't make sense to validate against a template that is invalid for a particular OS, in this case, `experimental/wsl2` on Darwin. Templates should be able to provide a list of supported OSes, if null then it is supported by all.

unsuman:
Fixes #3974

### DIFF
--- a/templates/default.yaml
+++ b/templates/default.yaml
@@ -11,6 +11,16 @@
 # 🟢 Builtin default: "vz" (on macOS 13.5 and later), "qemu" (on others)
 vmType: null
 
+# Host OS compatibility: List of host operating systems this template supports.
+# Valid values: "darwin" (macOS), "linux", "windows", "freebsd"
+# If not specified or empty, the template can run on any supported host OS.
+# 🟢 Builtin default: [] (runs on all supported host OSes)
+# 🔵 Examples:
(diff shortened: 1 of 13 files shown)


templates/default.yaml
```diff
@@ -11,6 +11,16 @@
 # 🟢 Builtin default: "vz" (on macOS 13.5 and later), "qemu" (on others)
 vmType: null
 
+# Host OS compatibility: List of host operating systems this template supports.
```
AkihiroSuda, line 14: This is an implementation detail of VM driver and must not be YAML fields

unsuman: Ok then, should I move it under `vmOpts`?

AkihiroSuda (changes requested):
Sorry, this design seems to contradict with the goal of the pluggable driver infra. https://github.com/lima-vm/lima/pull/3975/files#r2325320959

Thanks anyway for contributing this, but let's find an alternative solution for the original issue

Issue #3907, opened 2025-08-25
Print current boot session log in cloud-init progress monitoring.


norio-nomura:
### Description

`/var/log/cloud-init-output.log` contains the previous boot session's log.
So, it needs to start printing the log from the beginning of the current boot session.

Border of the session in `/var/log/cloud-init-output.log`::
```log
LIMA 2025-08-25T15:04:03+09:00| Exiting with code 0
Cloud-init v. 25.1.4-0ubuntu0~24.04.1 finished at Mon, 25 Aug 2025 06:04:03 +0000. Datasource DataSourceNoCloud [seed=/dev/vdb].  Up 5.37 seconds
Cloud-init v. 25.1.4-0ubuntu0~24.04.1 running 'init-local' at Mon, 25 Aug 2025 06:07:19 +0000. Up 1.00 seconds.
Cloud-init v. 25.1.4-0ubuntu0~24.04.1 running 'init' at Mon, 25 Aug 2025 06:07:21 +0000. Up 2.59 seconds.
```

olamilekan000:
Can the content of the boot session always be overwritten on every reboot?
If that's not possible, can we have a new log file on every reboot? We can always tail the the content of the latest log file.

norio-nomura:
On some conditions (#2905), VM's clock does not sync to the host at the beginning of the boot session.
On such conditions, time syncing happens while the boot session, comparing the timestamp in the log with host time may not work for detecting the beginning of the current boot session.

norio-nomura:
Following code, depending on `awk`, works as expected:
```golang
	args = append(args,
		"-p", strconv.Itoa(a.sshLocalPort),
		"127.0.0.1",
		"sh", "-c",
		`"sudo tail -n +$(sudo awk '
			BEGIN{b=1; e=1}
			/^Cloud-init.* finished/{e=NR}
			/.*/{if(NR>e){b=e+1}}
			END{print b}
		' /var/log/cloud-init-output.log) -f /var/log/cloud-init-output.log"`,
	)
```

edited code for `awk`

norio-nomura:
> Following code, depending on `awk`, works as expected:
> 
> 	args = append(args,
> 		"-p", strconv.Itoa(a.sshLocalPort),
> 		"127.0.0.1",
> 		"sh", "-c",
> 		`"sudo tail -n +$(sudo awk '
> 			BEGIN{b=1; e=1}
> 			/^Cloud-init.* finished/{e=NR}
[...]

PR #3982, opened 2025-09-06
`--progress`: use `journalctl -u cloud-init-main.service` if `cloud-init-main.service` is detected

norio-nomura:
## Description

In newer versions of cloud-init, `cloud-init-output.log` is not finished with a line containing `Cloud-init` and `finished`. That caused failing to detect both beginning and ending of the latest log. https://github.com/lima-vm/lima/issues/3907#issuecomment-3219319807  
To avoid that, use `journalctl -u cloud-init-main.service`.  
See: https://discourse.ubuntu.com/t/announcement-cloud-init-perfomance-optimization-single-process/47505


norio-nomura:
Thanks! 🙏🏻

PR #3985, opened 2025-09-06
Abort as soon as possible when running as root

### LINKED ISSUE #3978: Checking for "must not run as root" happens too late
Plugin commands and the `yq` command are executed as root. Maybe `yq` is fine[^1], but plugins are not:

[^1]: I think I would prefer to restrict `yq` as well, just on principle.

The reason we check so late is because we make an exception for `limactl generate-doc`. It would be best if we could get rid of this and run doc generation also under a limited user account. Or build a separate `lima-generate-doc` tool; why would an enduser ever run this command?

```console
❯ sudo limactl info
FATA[0000] must not run as the root user

❯ sudo limactl yq -n '.foo=42'
foo: 42

❯ cat /usr/local/libexec/lima/limactl-world
#!/bin/bash
echo "Hello world from $BASH_SOURCE"

❯ sudo limactl world
Hello world from /usr/local/libexec/lima/limactl-world
```

jandubois:
Fixes #3978

This PR makes it impossible to specify global options for `generate-doc` before the subcommand, but you can still specify them after, so this is not a loss in functionality.

I would still prefer that `generate-doc` would be called using a regular user, or would be a separate tool, and that `limactl` refuses to run as `root` completely, but this PR at least restores the old level of protection.

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -33,6 +33,10 @@ const (
 )
 
 func main() {
+	if os.Geteuid() == 0 && (len(os.Args) < 2 || os.Args[1] != "generate-doc") {
+		panic("must not run as the root user")
+	}
+
 	yq.MaybeRunYQ()
 	if runtime.GOOS == "windows" {
 		extras, hasExtra := os.LookupEnv("_LIMA_WINDOWS_EXTRA_PATH")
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

cmd/limactl/main.go
```diff
@@ -33,6 +33,10 @@ const (
 )
 
 func main() {
+	if os.Geteuid() == 0 && (len(os.Args) < 2 || os.Args[1] != "generate-doc") {
```
AkihiroSuda, line 36: Maybe add a comment to explain why `cmd.Name()` cannot be used

jandubois: > Maybe add a comment to explain why `cmd.Name()` cannot be used

I don't know what the comment should say, "Abort as early as possible to avoid running code as root"? That just duplicates what the code already does.

There is no `cmd` at this time, so it seems obvious why it can't be used.

Maybe:

// Check os.Args[] because we don't want to run Cobra code as root.

???

PR #3988, opened 2025-09-07
Restore DynamicSSHAddress functionality for WSL2

afbjorklund:
Restores the functionality:
commit df5e35ab83abb06a7d6501bfad5a77b83b49de62

That was reverted (again):
commit 2076ff1d31f4857aa839c35982f097003bbe831f

```go
func (l *LimaWslDriver) SSHAddress(_ context.Context) (string, error) {
	return "127.0.0.1", nil
}
```

It was first removed in 1.0.7:
commit ebaa10069a9aacc6bdb4a56587c033092d5752cb

----

* https://github.com/lima-vm/lima/pull/3842

* https://github.com/lima-vm/lima/pull/3299

### DIFF
--- a/pkg/driver/wsl2/vm_windows.go
+++ b/pkg/driver/wsl2/vm_windows.go
@@ -282,9 +282,12 @@ func getSSHAddress(ctx context.Context, instName string) (string, error) {
 	// fallback
 	cmd = exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "hostname", "-i")
 	out, err = cmd.CombinedOutput()
-	if err != nil || strings.HasPrefix(string(out), "127.") {
-		return "", fmt.Errorf("failed to get hostname for instance %q, err: %w (out=%q)", instName, err, string(out))
+	if err == nil {
+		ip := net.ParseIP(strings.TrimSpace(string(out)))
+		// some distributions use "127.0.1.1" as the host IP, but we want something that we can route to here
+		if ip != nil && !ip.IsLoopback() {
+			return strings.TrimSpace(string(out)), nil
+		}
 	}
-
-	return strings.TrimSpace(string(out)), nil
+	return "", fmt.Errorf("failed to get hostname for instance %q, err: %w (out=%q)", instName, err, string(out))
 }
--- a/pkg/driver/wsl2/vm_windows.go
+++ b/pkg/driver/wsl2/vm_windows.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -179,6 +180,33 @@ func unregisterVM(ctx context.Context, distroName string) error {
 	return nil
 }
 
+// GetWslStatus runs `wsl --list --verbose` and parses its output.
+// There are several possible outputs, all listed with their whitespace preserved output below.
+//
+// (1) Expected output if at least one distro is installed:
+// PS > wsl --list --verbose
+//
+//	NAME      STATE           VERSION
+//
+// * Ubuntu    Stopped         2
+//
+// (2) Expected output when no distros are installed, but WSL is configured properly:
+// PS > wsl --list --verbose
+// Windows Subsystem for Linux has no installed distributions.
+//
+// Use 'wsl.exe --list --online' to list available distributions
+// and 'wsl.exe --install <Distro>' to install.
+//
+// Distributions can also be installed by visiting the Microsoft Store:
+// https://aka.ms/wslstore
+// Error code: Wsl/WSL_E_DEFAULT_DISTRO_NOT_FOUND
+//
+// (3) Expected output when no distros are installed, and WSL2 has no kernel installed:
+//
+// PS > wsl --list --verbose
+// Windows Subsystem for Linux has no installed distributions.
+// Distributions can be installed by visiting the Microsoft Store:
+// https://aka.ms/wslstore
 func getWslStatus(instName string) (string, error) {
 	distroName := "lima-" + instName
 	out, err := executil.RunUTF16leCommand([]string{
@@ -228,4 +256,35 @@ func getWslStatus(instName string) (string, error) {
 	}
 
 	return instState, nil
+}
+
+// GetSSHAddress runs a hostname command to get the IP from inside of a wsl2 VM.
+//
+// Expected output (whitespace preserved, [] for optional):
+// PS > wsl -d <distroName> bash -c hostname -I | cut -d' ' -f1
+// 168.1.1.1 [10.0.0.1]
+// But busybox hostname does not implement --all-ip-addresses:
+// hostname: unrecognized option: I
+func getSSHAddress(ctx context.Context, instName string) (string, error) {
+	distroName := "lima-" + instName
+	// Ubuntu
+	cmd := exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "bash", "-c", `hostname -I | cut -d ' ' -f1`)
+	out, err := cmd.CombinedOutput()
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
+	}
+	// Alpine
+	cmd = exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "sh", "-c", `ip route get 1 | awk '{gsub("^.*src ",""); print $1; exit}'`)
+	out, err = cmd.CombinedOutput()
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
+	}
+	// fallback
+	cmd = exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "hostname", "-i")
+	out, err = cmd.CombinedOutput()
+	if err != nil || strings.HasPrefix(string(out), "127.") {
+		return "", fmt.Errorf("failed to get hostname for instance %q, err: %w (out=%q)", instName, err, string(out))
+	}
+
+	return strings.TrimSpace(string(out)), nil
 }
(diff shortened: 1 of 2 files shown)


afbjorklund:
It seems like InspectStatus is never called, and it expects the new SSHAddress to do this instead?

This broke the drivers, because they now don't get the dynamic ssh address even after running...

@unsuman

unsuman:
> It seems like InspectStatus is never called, and it expects the new SSHAddress to do this instead?

`driver.InspectStatus()` is called in:
https://github.com/lima-vm/lima/blob/b9b9aa6ff0dd971dc0516ecf559001590a882737/pkg/store/instance.go#L114

> This broke the drivers, because they now don't get the dynamic ssh address even after running...

Drivers? It is only applicable to `wsl2` afaik.

afbjorklund:
> Drivers? It is only applicable to `wsl2` afaik.

I was referring to the three container drivers

pkg/driver/wsl2/vm_windows.go
```diff
@@ -229,3 +229,27 @@ func getWslStatus(instName string) (string, error) {
 
 	return instState, nil
 }
+
+func getSSHAddress(ctx context.Context, instName string) (string, error) {
```
unsuman, line 268: Should this be called from the `wsl_driver_windows.go` too?
```go
func (l *LimaWslDriver) SSHAddress(ctx context.Context) (string, error) {
	return getSSHAddress(ctx, l.Instance.Name)
}
```

afbjorklund: Yes, but it will fail until the container is up and running

So before it was just using "127.0.0.1:0" as a placeholder

afbjorklund:
The driver starts out with the address from `SSHAdress`, later when it is running it is replaced in `InspectStatus`

It works the same in the other container drivers, they each implement two helper methods in order to do this:

```go
func (l *LimaWslDriver) InspectStatus(ctx context.Context, inst *limatype.Instance) string {
        status, err := getWslStatus(inst.Name)
        if err != nil {
                inst.Status = limatype.StatusBroken
        } else {
                inst.Status = status
        }

        inst.SSHLocalPort = 22

                sshAddr, err := getSSHAddress(ctx, inst.Name)
                if err == nil {
                        inst.SSHAddress = sshAddr
                } else {
                }
        }

        return inst.Status
}
```

Ideally we should probably **not** have the VM type in the method name like that, in the VM implementation...

```
func getWslStatus(instName string) (string, error) {
func getSSHAddress(ctx context.Context, instName string) (string, error) {

func getAcStatus(ctx context.Context, instName string) (string, error) {
func getSSHAddress(ctx context.Context, instName string) (string, error) {

func getDcStatus(ctx context.Context, instName string) (string, error) {
func getSSHAddress(ctx context.Context, instName string) (string, error) {
```

We also seem to have lost all documentation of the WSL2 methods in the "refactoring", that is not ideal.

afbjorklund:
@unsuman do you know why the documentation was removed? Should we restore it again?

https://github.com/lima-vm/lima/commit/2076ff1d31f4857aa839c35982f097003bbe831f#diff-518881de03f9b45d606d74cd62e9ba0f6f1515a519cd660c7edc836ab133ca51

afbjorklund:
We are missing a "context" parameter in the first method. It is _supposed_ to use `WithContext` in executil.

```
func getWslStatus(ctx context.Context, instName string) (string, error) {
```

pkg/driver/wsl2/vm_windows.go
```diff
@@ -229,3 +229,27 @@ func getWslStatus(instName string) (string, error) {
...
+	// Alpine
+	cmd = exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "sh", "-c", `ip route get 1 | awk '{gsub("^.*src ",""); print $1; exit}'`)
+	out, err = cmd.CombinedOutput()
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
+	}
+	// fallback
+	cmd = exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "hostname", "-i")
+	out, err = cmd.CombinedOutput()
+	if err != nil || strings.HasPrefix(string(out), "127.") {
```
AkihiroSuda: https://pkg.go.dev/net#IP.IsLoopback should be used rather than checking the "127." string prefix.

Also add a comment to describe why it is expected to be a loopback address, but it cannot be just hard-coded as 127.0.0.1

pkg/driver/wsl2/wsl_driver_windows.go
```diff
@@ -193,7 +193,7 @@ func (l *LimaWslDriver) InspectStatus(ctx context.Context, inst *limatype.Instan
 	inst.SSHLocalPort = 22
 
 	if inst.Status == limatype.StatusRunning {
-		sshAddr, err := l.SSHAddress(ctx)
+		sshAddr, err := getSSHAddress(ctx, inst.Name)
```
AkihiroSuda, line 196: Why not implement the new logic in `l.SSHAddress` ?

unsuman:
> why the documentation was removed? Should we restore it again?

Yes we should restore it, I guess I forgot to add those lines 🥲 . Can you please include them in this PR?
https://github.com/lima-vm/lima/commit/2076ff1d31f4857aa839c35982f097003bbe831f#diff-7f41ba250fd68816deddd4f7b0cf6edd7b669d01d8f6f22661ba69812462ef5dL42-L68
https://github.com/lima-vm/lima/commit/2076ff1d31f4857aa839c35982f097003bbe831f#diff-7f41ba250fd68816deddd4f7b0cf6edd7b669d01d8f6f22661ba69812462ef5dL120-L126

pkg/driver/wsl2/wsl_driver_windows.go
afbjorklund: Because SSHAddress does not check the Status, that is currently in InspectStatus

If we make XXXStatus part of the API instead, we could move the logic to `store` ?

afbjorklund: This happened when we added the "no-op" functions, that other drivers currently use.

```go
func (l *LimaVzDriver) SSHAddress(_ context.Context) (string, error) {
        return "127.0.0.1", nil
}

func (l *LimaVzDriver) InspectStatus(_ context.Context, _ *limatype.Instance) string {
        return ""
}
```

afbjorklund: Normally SSHAddress is **never** called, unless a specific driver feature is enabled.

```go
        // WSL instance SSH address isn't known until after VM start
        if a.driver.Info().Features.DynamicSSHAddress {
                sshAddr, err := a.driver.SSHAddress(ctx)
                if err != nil {
                        return err
                }
                a.instSSHAddress = sshAddr
        }
```

And InspectStatus is not implemented, which means it will check the PID files.

```go
func inspectStatus(ctx context.Context, instDir string, inst *limatype.Instance, y *limatype.LimaYAML) {
        status, err := driverutil.InspectStatus(ctx, inst)
        if err != nil {
                inst.Status = limatype.StatusBroken
                inst.Errors = append(inst.Errors, fmt.Errorf("failed to inspect status: %w", err))
                return
        }

        if status == "" {
                inspectStatusWithPIDFiles(instDir, inst, y)
                return
        }

        inst.Status = status
}
```

afbjorklund: Note that the logic is not new, it was just moved from pkg/store (in 1.x)  to the drivers (2.x).

https://github.com/lima-vm/lima/blob/release/1.2/pkg/store/instance_windows.go

pkg/driver/wsl2/wsl_driver_windows.go
```diff
@@ -193,7 +193,7 @@ func (l *LimaWslDriver) InspectStatus(ctx context.Context, inst *limatype.Instan
 	inst.SSHLocalPort = 22
```
afbjorklund, line 193: @unsuman this old behaviour (from before, in pkg/store) is also a bit weird.

It starts out with "port 0" like any other driver, then go to "22" when running.
But connecting to `127.0.0.1:22` is probably _never_ what you want, anyway...

Will follow up with a fix to that, when I know where the functions end up

pkg/driver/wsl2/vm_windows.go
```diff
@@ -229,3 +257,37 @@ func getWslStatus(instName string) (string, error) {
...
+	out, err = cmd.CombinedOutput()
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
+	}
+	// fallback
+	cmd = exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "hostname", "-i")
+	out, err = cmd.CombinedOutput()
+	if err == nil {
+		ip := net.ParseIP(strings.TrimSpace(string(out)))
+		// some distributions use "127.0.1.1" as the host IP, but we want something that we can route to here
```
jandubois, line 287: I think this command will **always** return `127.0.1.1`. It is documented at https://learn.microsoft.com/en-us/windows/wsl/networking#accessing-a-wsl-2-distribution-from-your-local-area-network-lan

I can confirm that this is still true with the latest WSL2:

```console
C:\Users\suse>wsl -d Ubuntu hostname -i
127.0.1.1

C:\Users\suse>wsl -d rancher-desktop hostname -i
127.0.1.1
```

So I don't know if this fallback will ever help; maybe it should be removed?

afbjorklund: That page seems a bit conflicted. https://learn.microsoft.com/en-us/windows/wsl/networking#identify-ip-address

pkg/driver/wsl2/wsl_driver_windows.go
afbjorklund: * https://github.com/lima-vm/lima/pull/4076

PR #3989, opened 2025-09-07
Drop the log message that qemu is properly signed

jandubois:
The message is pretty much useless. It should at best be a debug message, but I don't think it is needed at all, as long as there is a log message when something is wrong.

### DIFF
--- a/pkg/driver/qemu/entitlementutil/entitlementutil.go
+++ b/pkg/driver/qemu/entitlementutil/entitlementutil.go
@@ -107,7 +107,5 @@ func AskToSignIfNotSignedProperly(ctx context.Context, qExe string) {
 		} else {
 			logrus.Warn("If QEMU does not start up, you may have to sign the QEMU binary with the \"com.apple.security.hypervisor\" entitlement manually. See https://github.com/lima-vm/lima/issues/1742 .")
 		}
-	} else {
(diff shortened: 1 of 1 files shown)


pkg/driver/qemu/entitlementutil/entitlementutil.go
```diff
@@ -107,7 +107,5 @@ func AskToSignIfNotSignedProperly(ctx context.Context, qExe string) {
 		} else {
 			logrus.Warn("If QEMU does not start up, you may have to sign the QEMU binary with the \"com.apple.security.hypervisor\" entitlement manually. See https://github.com/lima-vm/lima/issues/1742 .")
 		}
-	} else {
-		logrus.Infof("QEMU binary %q seems properly signed with the \"com.apple.security.hypervisor\" entitlement", qExe)
```
AkihiroSuda, line 111: Should be harmless to keep this as a debug-level log

jandubois: > Should be harmless to keep this as a debug-level log

Yes, but why? How is this message ever useful?

I don't really mind keeping it (as debug), but I would like to understand the purpose of it.

AkihiroSuda (approved):

PR #3990, opened 2025-09-07
Make `limactl ls --format yaml` use the same schema as the json output

### LINKED ISSUE #3984: `limactl list` produces different data for `json` vs. `yaml` output
I'm working on adding some integration tests and found that the `limactl list` command outputs different structures for `json` vs. `yaml` format:

All of the "host" setting use different names:

```console
❯ l ls default --format json | yq .HostOS
darwin

❯ l ls default --format yaml | yq .hostos
darwin
```

Same for `HostArch`, `LimaHome`, and `IdentityFile`. Arguably these settings belong in the `limactl info` output because they are not specific to the instance.

In the YAML output all the instance specific settings are inside an `instance` object, but in the JSON they are merged together with the `host` settings.

```console
❯ l ls default --format json | yq .name
default

❯ l ls default --format yaml | yq .instance.name
default
```

Returning different schemas for different formats is insanity, but I have no idea how to fix it while preserving backwards compatibility. If we do break compatibility, then we should probably do this in the 2.0 release, so I'll mark this as a high priority issue.

jandubois:
Fixes #3984

It makes sure YAML uses the same key names as JSON for the 4 host properties, and it inlines the instance properties into the root object:

```console
❯ l ls default --format json | yq .HostOS
darwin

❯ l ls default --format yaml | yq .HostOS
darwin

❯ l ls default --format json | yq .config.user
{"name": "jan", "comment": "Jan Dubois", "home": "/home/jan.linux", "shell": "/bin/bash", "uid": 501}

❯ l ls default --format yaml | yq .config.user
name: jan
comment: Jan Dubois
home: /home/jan.linux
shell: /bin/bash
uid: 501
```

### DIFF
--- a/pkg/limainfo/limainfo.go
+++ b/pkg/limainfo/limainfo.go
@@ -7,12 +7,15 @@ import (
 	"context"
 	"errors"
 	"io/fs"
+	"path/filepath"
+	"runtime"
 
 	"github.com/sirupsen/logrus"
 
 	"github.com/lima-vm/lima/v2/pkg/envutil"
 	"github.com/lima-vm/lima/v2/pkg/limatype"
 	"github.com/lima-vm/lima/v2/pkg/limatype/dirnames"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
 	"github.com/lima-vm/lima/v2/pkg/limayaml"
 	"github.com/lima-vm/lima/v2/pkg/registry"
 	"github.com/lima-vm/lima/v2/pkg/templatestore"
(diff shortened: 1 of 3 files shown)


jandubois:
Just noticed that this PR does not include this part of my suggested solution:

* We add the 4 host settings to limactl info output

I would still like to hear some feedback if that is what we want to do. I can add it to this PR, or create a separate one...

jandubois:
> * We add the 4 host settings to limactl info output

I've added `hostArch`, `hostOS`, and `identityFile` to `limactl info` (it already has `limaHome`).

I lower-cased the key names to match the existing naming scheme for `limactl info` output.

jandubois:
I've added one more commit to "deprecate" the host fields, so they don't show up in `limactl ls --list-fields`, but still exist in the JSON object:

```console
❯ l ls default --json | yq .HostOS
darwin
❯ l ls --list-fields | grep HostOS
❯ 
```

pkg/limainfo/limainfo.go
```diff
@@ -29,6 +32,9 @@ type LimaInfo struct {
 	VMTypesEx       map[string]DriverExt         `json:"vmTypesEx"`   // since Lima v2.0.0
 	GuestAgents     map[limatype.Arch]GuestAgent `json:"guestAgents"` // since Lima v1.1.0
 	ShellEnvBlock   []string                     `json:"shellEnvBlock"`
+	HostOS          string                       `json:"hostOS"`       // since Lima 2.0.0
```
AkihiroSuda: nit
```suggestion
	HostOS          string                       `json:"hostOS"`       // since Lima v2.0.0
```

AkihiroSuda (dismissed):
Thanks

AkihiroSuda (approved):

PR #3992, opened 2025-09-07
Don't read VMOpts.VZ struct from cidata directly

afbjorklund:
This is supposed to be provided by the driver, and passed to the
hostagent as features. The cidata and hostagent doesn't know VZ.

Note: this also excludes the VMOpts.VZ from default and override,
similar to the change made to VMOpts.QEMU already. It is unknown.

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -118,7 +118,7 @@ func setupEnv(instConfigEnv map[string]string, propagateProxyEnv bool, slirpGate
 	return env, nil
 }
 
-func templateArgs(ctx context.Context, bootScripts bool, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort, vsockPort int, virtioPort string, noCloudInit bool) (*TemplateArgs, error) {
+func templateArgs(ctx context.Context, bootScripts bool, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort, vsockPort int, virtioPort string, noCloudInit, rosettaEnabled, rosettaBinFmt bool) (*TemplateArgs, error) {
 	if err := limayaml.Validate(instConfig, false); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


afbjorklund:
Still need to fix VMOpts.VZ in the templates:

* #3993

```
templates/docker-rootful.yaml:  {{- if .Instance.Config.VMOpts.VZ.Rosetta.Enabled}}
templates/docker.yaml:  {{- if .Instance.Config.VMOpts.VZ.Rosetta.Enabled}}
```

It is a bit over-engineered at the moment, since the docker templates don't set VMOpts.
i.e. there is no particular reason why it has to mention the AOT and CDI in the message

pkg/cidata/cidata.go
```diff
@@ -118,7 +118,7 @@ func setupEnv(instConfigEnv map[string]string, propagateProxyEnv bool, slirpGate
 	return env, nil
 }
 
-func templateArgs(ctx context.Context, bootScripts bool, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort, vsockPort int, virtioPort string) (*TemplateArgs, error) {
+func templateArgs(ctx context.Context, bootScripts bool, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort, vsockPort int, virtioPort string, rosettaEnabled, rosettaBinFmt bool) (*TemplateArgs, error) {
```
norio-nomura: Will this be changed again when #3899 comes in the future?

afbjorklund: Possibly, but not until the deprecated fields are gone from the top

pkg/hostagent/hostagent.go
```diff
@@ -163,10 +163,18 @@ func New(ctx context.Context, instName string, stdout io.Writer, signalCh chan o
 	vSockPort := limaDriver.Info().VsockPort
 	virtioPort := limaDriver.Info().VirtioPort
 
+	rosettaEnabled := limaDriver.Info().Features.RosettaEnabled
+	rosettaBinFmt := limaDriver.Info().Features.RosettaBinFmt
+	// Disable Rosetta in Plain mode
+	if *inst.Config.Plain {
```
norio-nomura, line 170: If the hostagent could query the plain mode config to the driver, wouldn't the hostagent be able to pass them to the functions that process the template without touching the contents of the config?

afbjorklund: Yes, it would be possible to move the logic from the hostagent to the guest and the boot scripts and have them read $LIMA_CIDATA_PLAIN instead (if one is willing to make them more "smart" like that)

norio-nomura: Also, should the hostagent be able to receive the template string used for `lima.env` from the driver?

afbjorklund: The lima.env is not handled by the driver, it is hostagent that creates the cidata (which includes lima.env)

PR #3997, opened 2025-09-07
Use consistent naming for VMOpts and vmOpts

afbjorklund:
go:   VMOpts.QEMU.CPUType, VMOpts.VZ.Rosetta
yaml: vmOpts.qemu.cpuType, vmOpts.vz.rosetta

Found while moving VMOpts to a separate type.

It is only in the comments actually, but anyway.

### DIFF
--- a/pkg/limatype/lima_yaml.go
+++ b/pkg/limatype/lima_yaml.go
@@ -20,7 +20,7 @@ type LimaYAML struct {
 	OS                 *OS           `yaml:"os,omitempty" json:"os,omitempty" jsonschema:"nullable"`
 	Arch               *Arch         `yaml:"arch,omitempty" json:"arch,omitempty" jsonschema:"nullable"`
 	Images             []Image       `yaml:"images,omitempty" json:"images,omitempty" jsonschema:"nullable"`
-	// Deprecated: Use VMOpts.Qemu.CPUType instead.
+	// Deprecated: Use VMOpts.QEMU.CPUType instead.
 	CPUType               CPUType       `yaml:"cpuType,omitempty" json:"cpuType,omitempty" jsonschema:"nullable"`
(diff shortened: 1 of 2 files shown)


pkg/limatype/lima_yaml.go
```diff
@@ -20,7 +20,7 @@ type LimaYAML struct {
 	OS                 *OS           `yaml:"os,omitempty" json:"os,omitempty" jsonschema:"nullable"`
 	Arch               *Arch         `yaml:"arch,omitempty" json:"arch,omitempty" jsonschema:"nullable"`
 	Images             []Image       `yaml:"images,omitempty" json:"images,omitempty" jsonschema:"nullable"`
-	// Deprecated: Use VMOpts.Qemu.CPUType instead.
+	// Deprecated: Use VMOpts.QEMU.CPUType instead.
```
AkihiroSuda, line 23: Soon:
```suggestion
	// Deprecated: Use VMOpts["qemu"].CPUType instead.
```

afbjorklund: I used the yaml names since they are less ugly, but that is the "after" state and this was the "before" state...

* https://github.com/lima-vm/lima/pull/3899

afbjorklund:
```
pkg/limatype/lima_yaml.go:      VMOpts             VMOpts        `yaml:"vmOpts,omitempty" json:"vmOpts,omitempty"`
pkg/limatype/lima_yaml.go:      // Deprecated: Use vmOpts.qemu.cpuType instead.
pkg/limatype/lima_yaml.go:      // Deprecated: Use vmOpts.vz.rosetta instead.
pkg/limatype/lima_yaml.go:type VMOpts map[VMType]any
```

PR #4008, opened 2025-09-09
`portfwd`: support HostSocket

norio-nomura:
Add missing `hostSocket` support to GRPC port forwarder.
- Added support for listening on Unix domain sockets
- Do not set `HostPortRange` if `HostSocket` is used.

edit: dropped closing `ClosableListeners` on exit

### DIFF
--- a/hack/test-port-forwarding.pl
+++ b/hack/test-port-forwarding.pl
@@ -16,6 +16,7 @@ use strict;
 use warnings;
 
 use Config qw(%Config);
+use File::Spec::Functions qw(catfile);
 use IO::Handle qw();
 use Socket qw(inet_ntoa);
 use Sys::Hostname qw(hostname);
@@ -32,6 +33,11 @@ if ($ipv4 eq "127.0.0.1" && $Config{osname} eq "darwin") {
     $ipv4 = qx(system_profiler SPNetworkDataType -json | jq -r 'first(.SPNetworkDataType[] | select(.ip_address) | .ip_address) | first');
     chomp $ipv4;
 }
+
+my $instDir = qx(limactl list "$instance" --yq .dir);
+chomp $instDir;
+# platform independent way to add trailing path separator
+my $sockDir = catfile($instDir, "sock", "");
 
 # If $instance is a filename, add our portForwards to it to enable testing
 if (-f $instance) {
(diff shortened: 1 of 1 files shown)


norio-nomura:
Dropped closing `ClosableListeners` on exit.
I'll open it as a separate PR.

AkihiroSuda (approved):
Thanks, can we cover this in `test-port-forwarding` ?

norio-nomura:
added a test to `hack/test-port-forwarding.pl`

norio-nomura:
I give up on making hostSocket tests pass on Windows, since I don't have any Windows environments for debugging.

jandubois:
Afaik there is no AF_UNIX interop between WSL2 and Windows; it only worked for WSL, but has never been implemented for WSL2.

norio-nomura:
> I give up on making hostSocket tests pass on Windows, since I don't have any Windows environments for debugging.

I didn’t give up, even though I said I had. But this time, I truly gave up.
Changed the hostSocket test to skip on Windows host.

AkihiroSuda:
With the introduction of  ssh-over-vsock, which port forwarder works best for host sockets?

norio-nomura:
Thanks! 🙏🏻 

> With the introduction of ssh-over-vsock, which port forwarder works best for host sockets?

If it's just the transfer speed, the port forwarder by SSH over VSOCK was fast, but the stability is still unknown.

If the cause of the gRPC forwarder's issue is caused by VSOCK, which is the transporting path, SSH over VSOCK may be affected in the same way.

pkg/portfwd/listener_darwin.go
```diff
@@ -7,12 +7,26 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"path/filepath"
 	"strconv"
 
 	"github.com/sirupsen/logrus"
 )
 
 func Listen(ctx context.Context, listenConfig net.ListenConfig, hostAddress string) (net.Listener, error) {
```
AkihiroSuda, line 16: The function seems called only from `forwardTCP()`?
How does this function work with unix sockets?

norio-nomura: GRPC port forwarder does not support guest-side Unix sockets.
`forwardTCP()`'s TCP indicates "guest side is TCP".

AkihiroSuda: Thanks, got it

PR #4009, opened 2025-09-09
limactl help and info flag should show available plugins

olamilekan000:
```
change adds functionality to display Plugins in the info and list commands
```

Fixes https://github.com/lima-vm/lima/issues/3608#issuecomment-3257711223

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/fsutil"
 	"github.com/lima-vm/lima/v2/pkg/limatype/dirnames"
 	"github.com/lima-vm/lima/v2/pkg/osutil"
+	"github.com/lima-vm/lima/v2/pkg/plugin"
 	"github.com/lima-vm/lima/v2/pkg/version"
 )
 
@@ -165,6 +166,10 @@ func newApp() *cobra.Command {
 	}
 	rootCmd.AddGroup(&cobra.Group{ID: "basic", Title: "Basic Commands:"})
 	rootCmd.AddGroup(&cobra.Group{ID: "advanced", Title: "Advanced Commands:"})
+	rootCmd.AddGroup(&cobra.Group{ID: "plugin", Title: "Available Plugins (Experimental):"})
+
+	// Discover and add plugins as commands
+	addPluginCommands(rootCmd)
 	rootCmd.AddCommand(
 		newCreateCommand(),
 		newStartCommand(),
@@ -215,14 +220,6 @@ func handleExitError(err error) {
 
 // executeWithPluginSupport handles command execution with plugin support.
 func executeWithPluginSupport(rootCmd *cobra.Command, args []string) error {
-	if len(args) > 0 {
-		cmd, _, err := rootCmd.Find(args)
-		if err != nil || cmd == rootCmd {
-			// Function calls os.Exit() if it found and executed the plugin
-			runExternalPlugin(rootCmd.Context(), args[0], args[1:])
-		}
-	}
-
 	rootCmd.SetArgs(args)
 	return rootCmd.Execute()
 }
@@ -232,7 +229,7 @@ func runExternalPlugin(ctx context.Context, name string, args []string) {
 		ctx = context.Background()
 	}
 
-	if err := updatePathEnv(); err != nil {
+	if err := plugin.UpdatePathForPlugins(); err != nil {
 		logrus.Warnf("failed to update PATH environment: %v", err)
 		// PATH update failure shouldn't prevent plugin execution
 	}
@@ -257,23 +254,29 @@ func runExternalPlugin(ctx context.Context, name string, args []string) {
 	logrus.Fatalf("external command %q failed: %v", execPath, err)
 }
 
-func updatePathEnv() error {
-	exe, err := os.Executable()
+func addPluginCommands(rootCmd *cobra.Command) {
+	plugins, err := plugin.DiscoverPlugins()
 	if err != nil {
-		return fmt.Errorf("failed to get executable path: %w", err)
+		logrus.Debugf("Failed to discover plugins: %v", err)
+		return
 	}
 
-	binDir := filepath.Dir(exe)
-	currentPath := os.Getenv("PATH")
-	newPath := binDir + string(filepath.ListSeparator) + currentPath
+	for _, p := range plugins {
+		pluginCmd := &cobra.Command{
+			Use:     p.Name,
+			Short:   p.Description,
+			GroupID: "plugin",
+			Run: func(cmd *cobra.Command, args []string) {
+				pluginName := cmd.Use
+				runExternalPlugin(cmd.Context(), pluginName, args)
+			},
+		}
 
-	if err := os.Setenv("PATH", newPath); err != nil {
-		return fmt.Errorf("failed to set PATH environment: %w", err)
+		pluginCmd.SilenceUsage = true
+		pluginCmd.SilenceErrors = true
+
+		rootCmd.AddCommand(pluginCmd)
 	}
-
-	logrus.Debugf("updated PATH to prioritize %s", binDir)
-
-	return nil
 }
 
 // WrapArgsError annotates cobra args error with some context, so the error message is more user-friendly.
--- a/pkg/limainfo/limainfo.go
+++ b/pkg/limainfo/limainfo.go
@@ -7,13 +7,17 @@ import (
 	"context"
 	"errors"
 	"io/fs"
+	"path/filepath"
+	"runtime"
 
 	"github.com/sirupsen/logrus"
 
 	"github.com/lima-vm/lima/v2/pkg/envutil"
 	"github.com/lima-vm/lima/v2/pkg/limatype"
 	"github.com/lima-vm/lima/v2/pkg/limatype/dirnames"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
 	"github.com/lima-vm/lima/v2/pkg/limayaml"
+	"github.com/lima-vm/lima/v2/pkg/plugin"
 	"github.com/lima-vm/lima/v2/pkg/registry"
 	"github.com/lima-vm/lima/v2/pkg/templatestore"
 	"github.com/lima-vm/lima/v2/pkg/usrlocalsharelima"
@@ -29,6 +33,10 @@ type LimaInfo struct {
 	VMTypesEx       map[string]DriverExt         `json:"vmTypesEx"`   // since Lima v2.0.0
 	GuestAgents     map[limatype.Arch]GuestAgent `json:"guestAgents"` // since Lima v1.1.0
 	ShellEnvBlock   []string                     `json:"shellEnvBlock"`
+	HostOS          string                       `json:"hostOS"`       // since Lima v2.0.0
+	HostArch        string                       `json:"hostArch"`     // since Lima v2.0.0
+	IdentityFile    string                       `json:"identityFile"` // since Lima v2.0.0
+	Plugins         []plugin.Plugin              `json:"plugins"`      // since Lima v2.0.0
 }
 
 type DriverExt struct {
@@ -72,6 +80,8 @@ func New(ctx context.Context) (*LimaInfo, error) {
 		VMTypesEx:       vmTypesEx,
 		GuestAgents:     make(map[limatype.Arch]GuestAgent),
 		ShellEnvBlock:   envutil.GetDefaultBlockList(),
+		HostOS:          runtime.GOOS,
+		HostArch:        limatype.NewArch(runtime.GOARCH),
 	}
 	info.Templates, err = templatestore.Templates()
 	if err != nil {
@@ -81,6 +91,11 @@ func New(ctx context.Context) (*LimaInfo, error) {
 	if err != nil {
 		return nil, err
 	}
+	configDir, err := dirnames.LimaConfigDir()
+	if err != nil {
+		return nil, err
+	}
+	info.IdentityFile = filepath.Join(configDir, filenames.UserPrivateKey)
 	for _, arch := range limatype.ArchTypes {
 		bin, err := usrlocalsharelima.GuestAgentBinary(limatype.LINUX, arch)
 		if err != nil {
@@ -95,5 +110,15 @@ func New(ctx context.Context) (*LimaInfo, error) {
 			Location: bin,
 		}
 	}
+
+	plugins, err := plugin.DiscoverPlugins()
+	if err != nil {
+		logrus.WithError(err).Debug("Failed to discover plugins")
+		// Don't fail the entire info command if plugin discovery fails
+		info.Plugins = []plugin.Plugin{}
+	} else {
+		info.Plugins = plugins
+	}
+
 	return info, nil
 }
--- a/pkg/plugin/plugin.go
+++ b/pkg/plugin/plugin.go
@@ -1,0 +1,194 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package plugin
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/lima-vm/lima/v2/pkg/usrlocalsharelima"
+)
+
+type Plugin struct {
+	Name        string `json:"name"`
+	Path        string `json:"path"`
+	Description string `json:"description,omitempty"`
+}
+
+func DiscoverPlugins() ([]Plugin, error) {
+	var plugins []Plugin
+	seen := make(map[string]bool)
+
+	dirs := getPluginDirectories()
+
+	for _, dir := range dirs {
+		pluginsInDir, err := scanDirectory(dir)
+		if err != nil {
+			logrus.Debugf("Failed to scan directory %s: %v", dir, err)
+			continue
+		}
+
+		for _, plugin := range pluginsInDir {
+			if !seen[plugin.Name] {
+				plugins = append(plugins, plugin)
+				seen[plugin.Name] = true
+			}
+		}
+	}
+
+	return plugins, nil
+}
+
+func getPluginDirectories() []string {
+	var dirs []string
+
+	selfPaths := []string{}
+
+	selfViaArgs0, err := usrlocalsharelima.ExecutableViaArgs0()
+	if err != nil {
+		logrus.WithError(err).Debug("failed to find executable")
+	} else {
+		selfPaths = append(selfPaths, selfViaArgs0)
+	}
+
+	selfViaOS, err := os.Executable()
+	if err != nil {
+		logrus.WithError(err).Debug("failed to find os.Executable()")
+	} else {
+		selfFinalPathViaOS, err := filepath.EvalSymlinks(selfViaOS)
+		if err != nil {
+			logrus.WithError(err).Debug("failed to resolve symlinks")
+			selfFinalPathViaOS = selfViaOS
+		}
+
+		if len(selfPaths) == 0 || selfFinalPathViaOS != selfPaths[0] {
+			selfPaths = append(selfPaths, selfFinalPathViaOS)
+		}
+	}
+
+	for _, self := range selfPaths {
+		binDir := filepath.Dir(self)
+		dirs = append(dirs, binDir)
+	}
+
+	pathEnv := os.Getenv("PATH")
+	if pathEnv != "" {
+		pathDirs := filepath.SplitList(pathEnv)
+		dirs = append(dirs, pathDirs...)
+	}
+
+	if prefixDir, err := usrlocalsharelima.Prefix(); err == nil {
+		libexecDir := filepath.Join(prefixDir, "libexec", "lima")
+		if _, err := os.Stat(libexecDir); err == nil {
+			dirs = append(dirs, libexecDir)
+		}
+	}
+
+	return dirs
+}
+
+func scanDirectory(dir string) ([]Plugin, error) {
+	var plugins []Plugin
+
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+
+		name := entry.Name()
+		if !strings.HasPrefix(name, "limactl-") {
+			continue
+		}
+
+		pluginName := strings.TrimPrefix(name, "limactl-")
+
+		if strings.Contains(pluginName, ".") {
+			if filepath.Ext(name) == ".exe" {
+				pluginName = strings.TrimSuffix(pluginName, ".exe")
+			} else {
+				continue
+			}
+		}
+
+		fullPath := filepath.Join(dir, name)
+
+		if !isExecutable(fullPath) {
+			continue
+		}
+
+		plugin := Plugin{
+			Name: pluginName,
+			Path: fullPath,
+		}
+
+		if desc := extractDescFromScript(fullPath); desc != "" {
+			plugin.Description = desc
+		}
+
+		plugins = append(plugins, plugin)
+	}
+
+	return plugins, nil
+}
+
+func isExecutable(path string) bool {
+	info, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+
+	mode := info.Mode()
+	if mode&0o111 != 0 {
+		return true
+	}
+
+	if filepath.Ext(path) == ".exe" {
+		return true
+	}
+
+	return false
+}
+
+func extractDescFromScript(path string) string {
+	content, err := os.ReadFile(path)
+	if err != nil {
+		logrus.Debugf("Failed to read plugin script %s: %v", path, err)
+		return ""
+	}
+
+	lines := strings.Split(string(content), "\n")
+	for _, line := range lines {
+		line = strings.TrimSpace(line)
+
+		// Look for pattern: # <limactl-desc>Description text</limactl-desc>
+		if strings.HasPrefix(line, "#") && strings.Contains(line, "<limactl-desc>") && strings.Contains(line, "</limactl-desc>") {
+			start := strings.Index(line, "<limactl-desc>") + len("<limactl-desc>")
+			end := strings.Index(line, "</limactl-desc>")
+
+			if start < end {
+				desc := strings.TrimSpace(line[start:end])
+				logrus.Debugf("Plugin %s: extracted description from script: %q", filepath.Base(path), desc)
+				return desc
+			}
+		}
+	}
+
+	logrus.Debugf("Plugin %s: no <limactl-desc> found in script", filepath.Base(path))
+
+	return ""
+}
+
+func UpdatePathForPlugins() error {
+	pluginDirs := getPluginDirectories()
+	newPath := strings.Join(pluginDirs, string(filepath.ListSeparator))
+	return os.Setenv("PATH", newPath)
+}
--- a/website/content/en/docs/config/plugin/cli.md
+++ b/website/content/en/docs/config/plugin/cli.md
@@ -1,12 +1,71 @@
 ---
-title: CLI plugins
+title: CLI plugins (Experimental)
 weight: 2
 ---
 
  | ⚡ Requirement | Lima >= 2.0 |
  |----------------|-------------|
 
-Lima supports a plugin-like command aliasing system similar to `git`, `kubectl`, and `docker`. When you run a `limactl` command that doesn't exist, Lima will automatically look for an external program named `limactl-<command>` in your system's PATH.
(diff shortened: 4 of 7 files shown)


website/content/en/docs/config/plugin/cli.md
```diff
@@ -1,12 +1,71 @@
...
+         "path": "/opt/homebrew/bin/limactl-sh"
+      }
+   ]
+}
+
+```
+
+### Plugin Descriptions
+
+Lima extracts plugin descriptions from script comments using the `<limactl-desc>` format. Include a description comment in your plugin script:
```
AkihiroSuda, line 52: This doesn't work when a plugin is a binary

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,194 @@
...
+		if !isExecutable(fullPath) {
+			continue
+		}
+
+		plugin := Plugin{
+			Name: pluginName,
+			Path: fullPath,
+		}
+
+		if desc := extractDescFromScript(fullPath); desc != "" {
```
AkihiroSuda, line 161: fullPath can be a binary (Mach-O, ELF, PE)

website/content/en/docs/config/plugin/cli.md
```diff
@@ -45,16 +104,36 @@ After creating this alias:
...
-- Share custom commands with your team by distributing scripts
\ No newline at end of file
+- Share custom commands with your team by distributing scripts
+- Package plugins with Lima distributions in the `libexec/lima` directory
+
+## Security Considerations
+
+**Security Note**: Lima will execute any executable named `limactl-*` found in the search directories. Ensure that:
+- You trust the source of any plugins in your `$PATH` or `libexec/lima`
+- You understand that plugin discovery automatically runs `--help` on found executables to generate descriptions
```
AkihiroSuda: Is this still true with limactl-desc ?

website/content/en/docs/config/plugin/cli.md
```diff
@@ -45,16 +104,36 @@ After creating this alias:
...
+## Package Installation
+
+Distribution packages and package managers can install plugins in `<PREFIX>/libexec/lima/` where `<PREFIX>` is typically `/usr/local` or `/opt/homebrew`. This allows plugins to be:
+- Managed by the package manager
+- Isolated from user's `$PATH`
+- Automatically discovered by Lima
+
+## Experimental Status
+
+**Experimental Feature**: The CLI plugin system is currently experimental and may change in future versions. Breaking changes to the plugin API or discovery mechanism may occur without notice.
```
AkihiroSuda, line 128: Update https://lima-vm.io/docs/releases/experimental/ too

website/content/en/docs/config/plugin/cli.md
olamilekan000: Should it just be stated in the docs that it won't extract description if it's a binary?

AkihiroSuda: Yes, and `extractDescFromScript` should just abort parsing the file when the file doesn't begin with `#!`

website/content/en/docs/config/plugin/cli.md
olamilekan000: This part isn't true anymore. The implementation was removed. I can also remove it.

jandubois: > This part isn't true anymore. The implementation was removed. I can also remove it.

Why was this removed, was it moved to a separate PR?

Is the idea to support both `<limactl-desc>` and `--help` for those that don't have a `<limactl-desc>`?

olamilekan000: That'll be addressed in a separate PR. See...  https://github.com/lima-vm/lima/pull/3973#pullrequestreview-3198893615.

jandubois: Ok, I expected that comment means to drop the description part completely from this PR and only list the names. Because now we are back discussing how the `<limactl-desc>` part is going to work instead of getting the PR merged. But that's fine, assuming this is close to consensus (I haven't looked at the code yet, just the comments here).

olamilekan000: Oh that was not how I understood the comment.

AkihiroSuda: In my comment I didn't expect to drop `limactl-desc`; `limactl-desc` seems fine to merge as-is (except that it shouldn't try to parse a binary file as a text script)

cmd/limactl/main.go
```diff
@@ -215,14 +220,6 @@ func handleExitError(err error) {
 
 // executeWithPluginSupport handles command execution with plugin support.
 func executeWithPluginSupport(rootCmd *cobra.Command, args []string) error {
```
jandubois, line 198: There is nothing plugin-specific left in this function. Since it is called only from a single place, I think the 2 statements should just be inlined at the call site, and this function can be eliminated.

cmd/limactl/main.go
```diff
@@ -232,7 +229,7 @@ func runExternalPlugin(ctx context.Context, name string, args []string) {
 		ctx = context.Background()
 	}
 
-	if err := updatePathEnv(); err != nil {
```
jandubois, line 235: GitHub doesn't let me comment on the function definition:

I think now that you have added a `plugin` package, this function should be moved there instead of being "hidden" in the `main` package.

I further think that this function should no longer use `os.LookPath`, but lookup the path using the plugin list. Otherwise we can't be sure that we will run the program listed by `limactl info`.

pkg/limainfo/limainfo.go
```diff
@@ -108,5 +110,15 @@ func New(ctx context.Context) (*LimaInfo, error) {
 			Location: bin,
 		}
 	}
+
+	plugins, err := plugin.DiscoverPlugins()
+	if err != nil {
+		logrus.WithError(err).Debug("Failed to discover plugins")
```
jandubois: I think this should be a warning. This is a condition that should not happen.

```suggestion
		logrus.WithError(err).Warn("Failed to discover plugins")
```

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,199 @@
...
+	var plugins []Plugin
+	seen := make(map[string]bool)
+
+	dirs := getPluginDirectories()
+
+	for _, dir := range dirs {
+		pluginsInDir, err := scanDirectory(dir)
+		if err != nil {
+			logrus.Debugf("Failed to scan directory %s: %v", dir, err)
+			continue
```
jandubois: I think this should be reported as a failure. It only fails when `os.ReadDir` fails.

This is also the only failure condition for this function; right now it always returns `nil`.

```suggestion
			return fmt.Errorf("failed to scan directory %s: %v", dir, err)
```

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,199 @@
...
+
+		for _, plugin := range pluginsInDir {
+			if !seen[plugin.Name] {
+				plugins = append(plugins, plugin)
+				seen[plugin.Name] = true
+			}
+		}
+	}
+
+	return plugins, nil
```
jandubois, line 49: I think the plugins should be sorted alphabetically, just like `templatestore.Templates` returns a sorted list.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+func getPluginDirectories() []string {
+	var dirs []string
+
+	selfPaths := []string{}
+
+	selfViaArgs0, err := usrlocalsharelima.ExecutableViaArgs0()
+	if err != nil {
+		logrus.WithError(err).Debug("failed to find executable")
```
jandubois: We should have the same error strings at the same log level as in `usrlocalsharelima.Dir()`:

```suggestion
		logrus.WithError(err).Warn("failed to find executable from os.Args[0]")
```

Same for the other errors below.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,199 @@
...
+		selfFinalPathViaOS, err := filepath.EvalSymlinks(selfViaOS)
+		if err != nil {
+			logrus.WithError(err).Debug("failed to resolve symlinks")
+			selfFinalPathViaOS = selfViaOS
+		}
+
+		if len(selfPaths) == 0 || selfFinalPathViaOS != selfPaths[0] {
+			selfPaths = append(selfPaths, selfFinalPathViaOS)
+		}
+	}
```
jandubois: I wonder if we should export a `usrlocalsharelima.SelfDirs()` function instead of duplicating so much code here. That would also automatically fix the logging msg issue.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,199 @@
...
+
+	pathEnv := os.Getenv("PATH")
+	if pathEnv != "" {
+		pathDirs := filepath.SplitList(pathEnv)
+		dirs = append(dirs, pathDirs...)
+	}
+
+	if prefixDir, err := usrlocalsharelima.Prefix(); err == nil {
+		libexecDir := filepath.Join(prefixDir, "libexec", "lima")
```
jandubois: I think we should have a `usrlocalsharelima.LibexecLima()` function that returns this directory.

It should also be used in `pkg/registry/registry.go`.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,199 @@
...
+
+		pluginName := strings.TrimPrefix(name, "limactl-")
+
+		if strings.Contains(pluginName, ".") {
+			if filepath.Ext(name) == ".exe" {
+				pluginName = strings.TrimSuffix(pluginName, ".exe")
+			} else {
+				continue
+			}
+		}
```
jandubois: I don't have strong opinions on rejecting plugin names containing a dot (but we should document it, if we do).

But the logic for `.exe` is specific to Windows and should only apply there; the .exe name will need to be rejected on Linux and macOS.

However, we should not just check for `.exe`, but all extensions listed in `PATHEXT` (while ignoring case) because there is no reason a plugin cannot be a batch script:

```
PS C:\Users\suse> echo $env:PATHEXT
.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC;.CPL
```

See the implementation of `os.LookPath` that also uses `PATHEXT`.

Sadly the default `PATHEXT` doesn't seem to contain `.PS1` for Powershell extensions, and I think that would be an obvious choice for a scripted plugin, but we can leave that for another day.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,199 @@
...
+		logrus.Debugf("Failed to read plugin script %s: %v", path, err)
+		return ""
+	}
+
+	if !strings.HasPrefix(string(content), "#!") {
+		logrus.Debugf("Plugin %s: not a script file, skipping description extraction", filepath.Base(path))
+		return ""
+	}
+
+	lines := strings.Split(string(content), "\n")
```
jandubois: This looks too complicated to me. It should be a simple regexp match over the whole contents with `/<limactl-desc>(.*?)</limactl-desc>/`, and the description is the capture group.

jandubois (changes requested):
I haven't looked at the docs changes, but here is some feedback on the code.

cmd/limactl/main.go
olamilekan000: The function was entirely removed, it isn't "hidden".

jandubois: I was talking about `runExternalPlugin`, which is still there.

website/content/en/docs/config/plugin/cli.md
```diff
@@ -1,12 +1,74 @@
...
+
+## Plugin Discovery
+
+Lima discovers plugins by scanning for executables named `limactl-<plugin-name>` in the following locations:
+
+1. **Directory containing the `limactl` binary** (including symlink support)
+2. **All directories in your `$PATH` environment variable**
+3. **`<PREFIX>/libexec/lima`** - For plugins installed by package managers or distribution packages
+
+The plugin discovery respects symlinks, which is particularly important for Homebrew installations where `limactl` may be symlinked.
```
jandubois: This should be explicit about looking in **both** locations: the directory where the symlink is in, and the directory containing the symlink target.

website/content/en/docs/config/plugin/cli.md
```diff
@@ -1,12 +1,74 @@
...
+```bash
+#!/bin/sh
+# <limactl-desc>Docker wrapper that connects to Docker daemon running in Lima instance</limactl-desc>
+set -eu
+
+# Rest of your script...
+```
+
+**Format Requirements:**
+- Must be a comment line starting with `#!`
```
jandubois: This is not correct. The plugin must be a script, so the first line must start with "#!" to specify the command that will interpret the file. But the `<limactl-desc>` line has no further restrictions than needing both the opening and closing tag.

website/content/en/docs/config/plugin/cli.md
```diff
@@ -1,12 +1,74 @@
...
+- Must contain exactly `<limactl-desc>Description text</limactl-desc>`
```
jandubois: It is not clear to me what this means. You can have anything before or after the tags. Maybe it will become clear when you rework the previous line.

website/content/en/docs/config/plugin/cli.md
```diff
@@ -1,12 +1,74 @@
...
+- The description text should be concise and descriptive
+- Only the first matching line will be used
+
+**Limitations:**
+- Description extraction only works for script files that begin with a shabang (`#!`)
```
jandubois: This is duplicated from the previous section, so maybe remove it in the previous section and only list it here.

website/content/en/docs/config/plugin/cli.md
```diff
@@ -45,16 +107,35 @@ After creating this alias:
 ```bash
 limactl sh default           # Equivalent to: limactl shell default
 limactl sh myinstance bash   # Equivalent to: limactl shell myinstance bash
-```
-
 ## How It Works
 
 1. When you run `limactl <unknown-command>`, Lima first tries to find a built-in command
-2. If no built-in command is found, Lima searches for `limactl-<unknown-command>` in your PATH
+2. If no built-in command is found, Lima searches for `limactl-<unknown-command>` in:
```
jandubois: This is repeating the "Plugin Discovery" section. It should only be specified in one place, to make sure descriptions don't diverge when they are updated.

website/content/en/docs/config/plugin/cli.md
```diff
@@ -45,16 +107,35 @@ After creating this alias:
...
+   - The directory containing the `limactl` binary
+   - All directories in your `$PATH`
+   - `<PREFIX>/libexec/lima` (if it exists)
 3. If found, Lima executes the external program and passes all remaining arguments to it
```
jandubois, line 110: You may want to add that the exit code of `lima PLUGIN` will be the exit code of the `lima-plugin` command.

website/content/en/docs/config/plugin/cli.md
```diff
@@ -45,16 +107,35 @@ After creating this alias:
...
 4. If not found, Lima shows the standard "unknown command" error
 
 This system allows you to:
 - Create personal shortcuts and aliases
 - Extend Lima's functionality without modifying the core application
-- Share custom commands with your team by distributing scripts
\ No newline at end of file
+- Share custom commands with your team by distributing scripts
+- Package plugins with Lima distributions in the `libexec/lima` directory
```
jandubois, line 117: I'm not convinced this section is really useful because the content is mostly obvious, and the part about `libexec` is only relevant to package maintainers.

website/content/en/docs/config/plugin/cli.md
```diff
@@ -45,16 +107,35 @@ After creating this alias:
...
+
+## Security Considerations
+
+**Security Note**: Lima will execute any executable named `limactl-*` found in the search directories. Ensure that:
+- You trust the source of any plugins in your `$PATH` or `libexec/lima`
+- Malicious executables with the `limactl-` prefix could be executed if placed in the search directories
```
jandubois: These are not really security considerations specific to Lima or Lima plugins, they apply to all software you install, so this feels like filler material.

We can add it when we implement support for automatically calling all plugins during the discover with a `--help` option.

website/content/en/docs/releases/experimental.md
```diff
@@ -19,6 +19,7 @@ The following commands are experimental and subject to change:
 - `limactl snapshot *`
 - `limactl tunnel`
 - `limactl template *`
+- `limactl-*`
```
jandubois: There should be a line in the section above (just below `External drivers`) that says `limactl plugins`. Because it is the plugin mechanism that is experimental, not a specific bundled plugin.

This section should list command actually bundled with Lima, so the `limactl-mcp` command should be listed here (in it's own PR) because it will be experimental.

website/content/en/docs/config/plugin/cli.md
olamilekan000: In a scenario where you have multiple <limactl-desc> tags something like.

```sh
#!/bin/sh
# Sample limactl-shell alias that shows running instances
# <limactl-desc> Run shell script </limactl-desc>
# <limactl-desc> Run shell script continue</limactl-desc>
echo "Running limactl-shell alias with arguments: $@"
limactl shell "$@"
```

olamilekan000: But I think it can be merged should incase there's something like that.

AkihiroSuda:
BATS failing

```
ok 38 setting both allow list and block list generates a warning in 138ms
not ok 39 limactl info includes the default block list in 124ms
# (in test file hack/bats/tests/preserve-env.bats, line 152)
#   `run -0 limactl yq '.shellEnvBlock[]' <<<"$output"' failed, expected exit code 0, got 1
# time="2025-09-17T09:21:44Z" level=warning msg="Plugin discovery: failed to scan directory /snap/bin: open /snap/bin: no such file or directory"
# time="2025-09-17T09:21:44Z" level=warning msg="Plugin discovery: failed to scan directory /home/runner/.local/bin: open /home/runner/.local/bin: no such file or directory"
# time="2025-09-17T09:21:44Z" level=warning msg="Plugin discovery: failed to scan directory /home/runner/.config/composer/vendor/bin: open /home/runner/.config/composer/vendor/bin: no such file or directory"
# time="2025-09-17T09:21:44Z" level=warning msg="Plugin discovery: failed to scan directory /snap/bin: open /snap/bin: no such file or directory"
# Error: bad file '-': yaml: mapping values are not allowed in this context
# ===== protect =====
ok 40 create dummy instance in 66ms
```

https://github.com/lima-vm/lima/actions/runs/17775630264/job/50573056648?pr=4009

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,241 @@
...
+	if err := UpdatePathForPlugins(); err != nil {
+		logrus.Warnf("failed to update PATH environment: %v", err)
+		// PATH update failure shouldn't prevent plugin execution
+	}
+
+	externalCmd := "limactl-" + name
+	execPath, err := exec.LookPath(externalCmd)
+	if err != nil {
+		return
+	}
```
jandubois: I think I mentioned this before that you should use the path to the plugin from the discovery code and not call `exec.LookPath`, which might find a different version of the plugin in case there are bugs in the discovery algorithm.

Use the same list in both places to make sure the behaviour is always in sync!

cmd/limactl/main.go
```diff
@@ -51,7 +51,7 @@ func main() {
 	rootCmd := newApp()
 	if err := executeWithPluginSupport(rootCmd, os.Args[1:]); err != nil {
 		server.StopAllExternalDrivers()
-		handleExitError(err)
+		usrlocalsharelima.HandleExitError(err)
```
jandubois: `usrlocalsharelima` is kind of an unexpected place for this function. I would move it maybe to its own file in `pkg/osutil/exit.go`.

But this is nitpicking, unless there are other reasons to update the PR once more, this can also be fixed up later.

cmd/limactl/main.go
```diff
@@ -51,7 +51,7 @@ func main() {
 	rootCmd := newApp()
```
jandubois, line 50: Nit-picking: This whole block could be written like this:

```go
	rootCmd := newApp()
	err := executeWithPluginSupport(rootCmd, os.Args[1:])
	server.StopAllExternalDrivers()
	usrlocalsharelima.HandleExitError(err)
	if err != nil {
		logrus.Fatal(err)
	}
```

cmd/limactl/main.go
```diff
@@ -201,79 +203,45 @@ func newApp() *cobra.Command {
...
-			// Function calls os.Exit() if it found and executed the plugin
-			runExternalPlugin(rootCmd.Context(), args[0], args[1:])
+	rootCmd.SetArgs(args)
+
+	if err := rootCmd.ParseFlags(args); err == nil {
+		if debug, _ := rootCmd.Flags().GetBool("debug"); debug {
+			logrus.SetLevel(logrus.DebugLevel)
+			debugutil.Debug = true
 		}
 	}
```
jandubois, line 206: It bothers me that this code block exists in 2 places. Can't it be called early enough that it covers both execution branches?

I think this here might be enough, and you can delete the copy in the PersistentPreRun command?

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+		})
+
+	return plugins, nil
+}
+
+func getPluginDirectories() []string {
+	var dirs []string
+
+	selfDirs := usrlocalsharelima.SelfDirs()
+	dirs = append(dirs, selfDirs...)
```
jandubois: This could be just

```suggestion
	dirs := usrlocalsharelima.SelfDirs()
```

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+
+func DiscoverPlugins() ([]Plugin, error) {
+	var plugins []Plugin
+	seen := make(map[string]bool)
+
+	dirs := getPluginDirectories()
+
+	for _, dir := range dirs {
+		pluginsInDir := scanDirectory(dir)
+		for _, plugin := range pluginsInDir {
```
jandubois: This may be a personal preference, but I prefer to avoid temp variables that are only used once, unless they are necessary to keep the expressions short. But I find this easier to read:

```suggestion
	for _, dir := range getPluginDirectories() {
		for _, plugin := range scanDirectory(dir) {
```

Of course this only works because the functions don't return an error.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+
+	extensions := strings.Split(strings.ToUpper(pathExt), ";")
+	extUpper := strings.ToUpper(ext)
+
+	for _, validExt := range extensions {
+		if validExt == extUpper {
+			return true
+		}
+	}
+	return false
```
jandubois, line 90: I think this can be written as:

```suggestion
    return slices.Contains(extensions, strings.ToUpper(ext))
```

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+		pluginName := strings.TrimPrefix(name, "limactl-")
+
+		if strings.Contains(pluginName, ".") {
+			ext := filepath.Ext(name)
+			if !isWindowsExecutableExt(ext) {
+				continue
+			}
+
+			pluginName = strings.TrimSuffix(pluginName, ext)
+		}
```
jandubois: This logic is not correct because it treats names with multiple dots (like `limactl-hello.world.exe`) differently.

I think you should move the GOOS check here, and just strip Windows executable extensions on Windows, and leave the rest unchanged:

```suggestion
		if runtime.GOOS == "windows" {
			ext := filepath.Ext(pluginName)
			if isWindowsExecutableExt(ext) {
				pluginName = strings.TrimSuffix(pluginName, ext)
			}
		}
```

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+			plugin.Description = desc
+		}
+
+		plugins = append(plugins, plugin)
+	}
+
+	return plugins
+}
+
+func isExecutable(path string) bool {
```
jandubois: I would keep this function next to `isWindowsExecutableExtension`. Either move this function up, or move the other function down here.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+	info, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+
+	mode := info.Mode()
+	if mode&0o111 != 0 {
+		return true
+	}
```
jandubois: This code will not work on Windows; the executable bit is always `0` iirc.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+
+	if runtime.GOOS == "windows" {
+		ext := filepath.Ext(path)
+		return isWindowsExecutableExt(ext)
+	}
```
jandubois: I'm not sure if this code is needed.

The file extension is only needed when you run a command without an extension. E.g. when you run `tar` Windows will try each extension from `PATHEXT` in turn to find an executable: `tar.exe`, `tar.bat`, etc.

But Windows will execute any command that exists as-is, regardless of the extension:

```console
C:\Users\suse>tar
tar: Must specify one of -c, -r, -t, -u, -x

C:\Users\suse>copy C:\Windows\System32\tar.exe hello.world
        1 file(s) copied.

C:\Users\suse>hello.world
hello.world: Must specify one of -c, -r, -t, -u, -x
```

So I think I would just remove this block completely. We want to strip the common extensions from the plugin name for the `--help` output, but here we don't really care.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+
+func extractDescFromScript(path string) string {
+	content, err := os.ReadFile(path)
+	if err != nil {
+		logrus.Debugf("Failed to read plugin script %s: %v", path, err)
+		return ""
+	}
+
+	if !strings.HasPrefix(string(content), "#!") {
+		logrus.Debugf("Plugin %s: not a script file, skipping description extraction", filepath.Base(path))
```
jandubois: I think the debug log should always include the full path name. This is for debugging issues, normal users will not see the output because they don't use `--debug`.

```suggestion
		logrus.Debugf("Plugin %s: not a script file, skipping description extraction", path)
```

Same for the next error below.

pkg/plugin/plugin.go
```diff
@@ -0,0 +1,253 @@
...
+		return ""
+	}
+
+	mergedDesc := descs[0]
+	for _, line := range descs[1:] {
+		mergedDesc += "\n" + strings.Repeat(" ", 22) + line
+	}
+
+	logrus.Debugf("Plugin %s: extracted merged description from script:\n%q", filepath.Base(path), mergedDesc)
+	return mergedDesc
```
jandubois: I think you should ignore everything but the first match. And you definitely should not include multi-line formatting in the description here; that would have to be done at the presentation layer. But I think we should simply recommend to use short, one-line descriptions.

So the whole block should be just `return matches[0]` and that's it.

website/content/en/docs/releases/experimental.md
```diff
@@ -19,6 +19,8 @@ The following commands are experimental and subject to change:
 - `limactl snapshot *`
 - `limactl tunnel`
 - `limactl template *`
+- `limactl plugins` plugin mechanism (see [CLI plugins](../config/plugin/cli))
+  - `limactl-*`
```
jandubois: This line should be in the previous block ("The following features are experimental and subject to change:") under the "External drivers:" line, and look like this:

```suggestion
- `limactl plugins:` plugin mechanism (see [CLI plugins](../config/plugin/cli))
```

There are no specific commands to list down here because plugins would be invoked just like a builtin command, and we can't list `limactl *` here. 😄

website/content/en/docs/config/plugin/cli.md
```diff
@@ -1,12 +1,71 @@
...
+```bash
+#!/bin/sh
+# <limactl-desc>Docker wrapper that connects to Docker daemon running in Lima instance</limactl-desc>
+set -eu
+
+# Rest of your script...
+```
+
+**Format Requirements:**
+- Only files beginning with a shebang (`#!`) are treated as scripts, and their <limactl-desc> lines will be extracted as the plugin description i.e Must contain exactly `<limactl-desc>Description text</limactl-desc>`
```
jandubois: Tag needs to be quoted here:

```suggestion
- Only files beginning with a shebang (`#!`) are treated as scripts, and their `<limactl-desc>` lines will be extracted as the plugin description i.e Must contain exactly `<limactl-desc>Description text</limactl-desc>`
```

jandubois (changes requested):
Looks mostly fine, but I still have some stylistic feedback, and some minor issues.

pkg/plugin/plugin.go
jandubois: Why was this marked as resolved? If you think my suggestion is not applicable, then you have to leave a comment and not just hide the feedback!

olamilekan000: I must have missed the suggestion while working on other feedbacks. APologies.

PR #4016, opened 2025-09-10
`limactl prune --keep-referred`: Do not skip template on failing `driverutil.ResolveVMType()`

norio-nomura:
`locationsFromLimaYAML()` should be called on the template, which fails `driverutil.ResolveVMType()` or not.  
It can be assumed that `driverutil.ResolveVMType()` does not affect any locations in the template.  
Prioritize that there is no omission in the current locations collection rather than changing locations by `driverutil.ResolveVMType()` in the future.

### DIFF
--- a/cmd/limactl/prune.go
+++ b/cmd/limactl/prune.go
@@ -101,8 +101,7 @@ func knownLocations(ctx context.Context) (map[string]limatype.File, error) {
 			return nil, err
 		}
 		if err := driverutil.ResolveVMType(ctx, y, t.Name); err != nil {
-			logrus.Warnf("failed to resolve vm for %q: %v", t.Name, err)
-			continue
+			logrus.Warnf("failed to resolve vmType for %q: %v", t.Name, err)
 		}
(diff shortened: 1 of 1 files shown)


cmd/limactl/prune.go
```diff
@@ -102,7 +102,6 @@ func knownLocations(ctx context.Context) (map[string]limatype.File, error) {
 		}
 		if err := driverutil.ResolveVMType(ctx, y, t.Name); err != nil {
```
jandubois, line 103: Why do we need to keep the warning if the check does not make any difference to us? Can't we delete the whole call?

If we want to keep it, should the log message say "vmType" instead of "vm"?

norio-nomura: There is currently no driver to change the location, but I don't know in the future.
Depending on the driver, you may need nerdctl with special patches or a custom BIOS.

> If we want to keep it, should the log message say "vmType" instead of "vm"?

I'll change to "vmType". 👍🏻

norio-nomura:
Thanks! 🙏🏻

Issue #4002, opened 2025-09-08
Static port forwarding does not work as expected


norio-nomura:
`a.addStaticPortForwardsFromList()` is only called when `--plain=false`, so it does not work as intended when `--plain=true`.

_Originally posted by @norio-nomura in https://github.com/lima-vm/lima/pull/3699#discussion_r2329369179_

PR #4019, opened 2025-09-10
bug fix: static port forward works in plain

Horiodino:
This PR fixes issue [#4002](https://github.com/lima-vm/lima/issues/4002)
where static port forwarding did not work when running with --plain=true, due to incorrect 
insertion in `if` condition.


.github/workflows/test.yml
```diff
@@ -299,6 +299,7 @@ jobs:
         - ../hack/test-templates/alpine-iso-9p-writable.yaml  # Covers alpine-iso.yaml
         - ../hack/test-templates/net-user-v2.yaml
         - ../hack/test-templates/test-misc.yaml  # TODO: merge net-user-v2 into test-misc
+        - ../hack/test-templates/static-port-forward.yaml
```
AkihiroSuda: Can't we merge this to `test-misc.yaml` ?

Horiodino: we can but i think having a dedicated static-port-forward.yaml ensures clear test isolation and intent. Port forwarding behavior ,especially differences between static and dynamic, and plain vs normal modes

AkihiroSuda: Not a fan of increasing the size of the GHA job matrix for this (mostly because of the flakiness of the CI)

pkg/hostagent/hostagent.go
```diff
@@ -535,13 +535,13 @@ sudo chown -R "${USER}" /run/host-services`
...
-		staticPortForwards, err := a.separateStaticPortForwards()
-		if err != nil {
-			errs = append(errs, err)
-		}
-		a.addStaticPortForwardsFromList(ctx, staticPortForwards)
+	staticPortForwards, err := a.separateStaticPortForwards()
+	if err != nil {
+		errs = append(errs, err)
+	}
+	a.addStaticPortForwardsFromList(ctx, staticPortForwards)
```
jandubois: `separateStaticPortForwards` always returns `nil` for the error, so the return value and error handling can be removed. Otherwise the `addStaticPortForwardsFromList` would have to be moved to an `else` branch because return values should not be used when a function returns an error.

Issue #2536, opened 2024-08-07
Bound ports are not released immediately when container is stopped/deleted


rfay:
### Description

This is a replacement for
* https://github.com/abiosoft/colima/issues/644

since the problem is in Lima. 

When a container is stopped, its bound ports are not immediately released. It can take 2 seconds or more for that to happen. DDEV has had to add a sleep after deleting a container to wait for the port to be released.

abiosoft:
Considering that Lima scans the VM periodically for opened ports, there is not an easy way out.

An option could be for Lima to support dynamic port forwarding. 
Colima can monitor docker events and leverage that to forward (and stop forwarding) ports as needed.

rfay:
Thanks. But it seems every other Docker provider does this successfully and immediately.

Nino-K:
We have dynamic port forwarding in Rancher Desktop that listens for events from both [Docker](https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/go/guestagent/pkg/docker/events.go) and [containerd](https://github.com/rancher-sandbox/rancher-desktop/blob/main/src/go/guestagent/pkg/containerd/events.go), which could potentially be leveraged in Lima.

balajiv113:
@jandubois
How about we use LD_PRELOAD with a custom bind logic to get callback within lima ??

Just a initial thought not sure if this works for all cases (like docker, kubernetes and all). But if this works, we should be able to get instant callback before the actual bind happens.

jandubois:
> How about we use LD_PRELOAD with a custom bind logic to get callback within lima ??

I feel a bit nervous of this introducing hard-to-find bugs, and wonder how this will work for different versions of glibc and musl.

And I guess your edit says that it won't work anyways as soon as network netspaces come into play.

So I would say we should integrate the code from the Rancher Desktop guest agent into Lima; it does seem to work quite well. IIRC the Rancher Desktop guest agent is an early fork of the Lima version, so should have the same basic structure for scraping port usage. Let's wait to see when @Nino-K has some time to port it over.

balajiv113:
> I feel a bit nervous of this introducing hard-to-find bugs, and wonder how this will work for different versions of glibc and musl.

True. Had similar concerns as well. Since it doesn't work for multiple namespace there is no benefit out of it

> So I would say we should integrate the code from the Rancher Desktop guest agent into Lima; it does seem to work quite well.

[...]

PR #1855, opened 2023-09-27
fix: guestagent audit detection and fallback

pendo324:
WSL2 just put out a new pre-release version (version 2.0.1 at the time of writing). While testing Lima with the new version, I noticed that AUDIT is enabled in the kernel:

```
# uname -a
Linux EC2AMAZ-3QHKPN9 5.15.123.1-microsoft-standard-WSL2 #1 SMP Mon Aug 7 19:01:48 UTC 2023 x86_64 GNU/Linux

# zfgrep AUDIT /proc/config.gz
CONFIG_AUDIT=y
CONFIG_HAVE_ARCH_AUDITSYSCALL=y
CONFIG_AUDITSYSCALL=y
CONFIG_AUDIT_ARCH=y
# CONFIG_KVM_MMU_AUDIT is not set
# CONFIG_NETFILTER_XT_TARGET_AUDIT is not set
```

and `cap_audit_read`/`cap_audit_write` are set:

```
# cat /proc/$$/status | grep Cap
CapInh: 0000000000000000
CapPrm: 000001ffffffffff
CapEff: 000001ffffffffff
CapBnd: 000001ffffffffff
CapAmb: 0000000000000000

# capsh --decode=000001ffffffffff
0x000001ffffffffff=cap_chown,cap_dac_override,cap_dac_read_search,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_linux_immutable,cap_net_bind_service,cap_net_broadcast,cap_net_admin,cap_net_raw,cap_ipc_lock,cap_ipc_owner,cap_sys_module,cap_sys_rawio,cap_sys_chroot,cap_sys_ptrace,cap_sys_pacct,cap_sys_admin,cap_sys_boot,cap_sys_nice,cap_sys_resource,cap_sys_time,cap_sys_tty_config,cap_mknod,cap_lease,cap_audit_write,cap_audit_control,cap_setfcap,cap_mac_override,cap_mac_admin,cap_syslog,cap_wake_alarm,cap_block_suspend,cap_audit_read,cap_perfmon,cap_bpf,cap_checkpoint_restore
```

However, lima-guestagent was now failing to start, probably for the same reason auditd was also failing to start.

After some digging, I found [this excellent blog post (Google Translate helped out a lot)](https://blog.jp.square-enix.com/iteng-blog/posts/00004-auditd-on-wsl2-windows10/), which led me in the right direction.

[...]

PR #4021, opened 2025-09-10
Subscribe to container engine API for published ports

Nino-K:
This PR aims to capture the published ports within the VM for various container engines. It directly subscribes to the corresponding container engine APIs (Docker, containerd and Kubernetes) to detect published ports immediately as a container is created.

I'm also planning to move the iptables and procnet settings under the portMonitor property; however, that will be addressed in a follow-up PR.


AkihiroSuda:
Seems overengineering.
What is the problem with the existing event watcher?

go.mod
AkihiroSuda, line 1: Too many new dependencies

pkg/guestagent/events/containerd_linux.go
```diff
@@ -0,0 +1,364 @@
...
+	containerPorts, ok := container.Labels[portsKey]
+	if ok {
+		ipPorts, err = extractIPPortsFromLabel(containerPorts)
+		if err != nil {
+			return nil, fmt.Errorf("extracting IPPorts from container %s ports label failed: %w", containerID, err)
+		}
+		return ipPorts, nil
+	}
+	// If the label is not present, we check the network config in the following path:
+	// <DATAROOT>/<ADDRHASH>/containers/<NAMESPACE>/<CID>/network-config.json
```
AkihiroSuda, line 290: This file isn't expected to be parsed externally

jandubois:
> Seems overengineering. What is the problem with the existing event watcher?

There are no event watchers for docker or containerd; we rely on polling `/proc/net/tcp` every couple of seconds (3?), so there is always a delay between the container opening a port and it being forwarded to the host.

See e.g. https://github.com/abiosoft/colima/issues/71

This PR is basically a port of the event watchers from Rancher Desktop on Windows, which does not suffer from this issue.

pkg/guestagent/events/containerd_linux.go
jandubois: I think the issue is that since https://github.com/containerd/nerdctl/pull/4290 the information is no longer available via labels. How are you expected to get the port mapping information?

AkihiroSuda:
> we rely on polling /proc/net/tcp every couple of seconds (3?), so there is always a delay between the container opening a port and it being forwarded to the host.


We also monitor audit events:
https://github.com/lima-vm/lima/blob/69538a761e777688e55d2869cdc8380bdbe1fc18/pkg/guestagent/guestagent_linux.go#L33

Maybe the audit monitor has a bug ?

AkihiroSuda:
There is also a proposal to use eBPF to monitor ports
- https://github.com/lima-vm/lima/pull/3067

cc @balajiv113

pkg/guestagent/events/containerd_linux.go
AkihiroSuda: The expectation is not to depend on container engine implementations; audit events or eBPF should work (see my comments below)

go.mod
Nino-K: Unfortunately, there are many indirect dependencies being used for just a few direct ones. However, I went ahead and compared the guest agent sizes from the master build and my PR, and I’m happy to share the results below:

```
total 231384
-rw-r--r--  1 ninok  wheel    55M Sep 11 10:27 guestagent_master
-rw-r--r--  1 ninok  wheel    58M Sep 11 10:26 guestagent_pr
```

Nino-K:
> Seems overengineering. What is the problem with the existing event watcher?

In addition to what @jandubois also pointed out, the current [implementation](https://github.com/lima-vm/lima/blob/c345be1144d35da2d810f616a2071ea6ab8c2639/pkg/guestagent/guestagent_linux.go#L198-L219)
 of the port collection mechanism uses a 3-second polling ticker, although this is [configurable](https://github.com/lima-vm/lima/blob/c345be1144d35da2d810f616a2071ea6ab8c2639/cmd/lima-guestagent/daemon_linux.go#L28). However, this PR subscribes directly to the APIs, allowing port exposure events to be handled more immediately.

jandubois:
This PR is in response to https://github.com/lima-vm/lima/issues/2536

I thought I read a note from @balajiv113 that the audit approach didn't work out because most cloud images did not have the required kernel modules for it installed, but can't find the reference right now.

And https://github.com/lima-vm/lima/pull/3067 also sounds like it doesn't work anymore for Kubernetes.

jandubois:
https://github.com/lima-vm/lima/pull/1855 may explain why audit monitoring doesn't seem to work.

AkihiroSuda:
Probably we can revisit this TODO to improve the ticking lag

https://github.com/lima-vm/lima/blob/7459f4587987ed014c372f17b82de1817feffa2e/cmd/lima-guestagent/daemon_linux.go#L57-L63

(`AUDIT_SYSCALL` may work too?)

go.mod
AkihiroSuda: My concern is mostly about the maintenance burdens rather than the binary footprint; it is hard to handle dependabot PRs and detect potential supply chain attacks.

Also, the inflation of the Go deps is also challenging for getting Lima in Debian, as Debian makes dpkg for each of those deps.

pkg/guestagent/events/docker_linux.go
```diff
@@ -0,0 +1,228 @@
...
+				Port:     int32(port.PublicPort),
+			})
+		}
+		sendHostAgentEvent(false, ipPorts, ch)
+		runningContainers[container.ID] = ipPorts
+	}
+	return nil
+}
+
+func convertToIPPort(portMap nat.PortMap) ([]*api.IPPort, error) {
```
AkihiroSuda, line 194: Why do we need to parse the Docker API object?
The existing `procnettcp` watcher should be still able to detect the port number IIUC?

The `procnettcp` watcher has an issue about the ticking lags, but I believe this issue can be solved by stimulating the ticker on an `AUDIT_SYSCALL` event (or maybe its eBPF equivalent). See my other comment.

go.mod
jandubois: > My concern is mostly about the maintenance burdens rather than the binary footprint; it is hard to handle dependabot PRs and detect potential supply chain attacks.

True, but these are all dependencies of the `containerd` and `dockerd` client libraries, so should be subject to some scrutiny from upstream already.

Also, independent of this issue, we should maybe consider setting a dependabot cooldown period, e.g. 7 days. It will tell dependabot not to make a PR until there have been no new releases of the dependency for those 7 days. This also helps with accidental breakage in dependencies, that is normally fixed quickly.

I've found the cooldown more important for node dependencies, but it can be setup for each eco system.

> Also, the inflation of the Go deps is also challenging for getting Lima in Debian, as Debian makes dpkg for each of those deps.

I just looked at https://go-team.pages.debian.net/packaging.html but did not find any rationale for __why__ they would go to all that work. What is the benefit?

Anyways, I assume that Debian has packaged all the prerequisites for docker and containerd clients already, so this should not block anything.

pkg/guestagent/events/containerd_linux.go
jandubois: > The expectation is not to depend on container engine implementations; audit events or eBPF should work (see my comments below)

I think this is a poor excuse for `nerdctl` breaking backwards compatibility.

pkg/guestagent/events/docker_linux.go
jandubois: I don't know.

#1855 was 2 years ago, disabling libaudit when not allowed by the kernel.

One year ago, in https://github.com/lima-vm/lima/issues/2536#issuecomment-2370014610 @balajiv113 wrote:

> Completely fine in doing what rancher does for now (As long as size doesn't become too huge and we don't add yet another new container)
>
> On other hand, Ebpf + libaudit changes are almost ready. But i would need some more time to check on different aspects.

It was my understanding that the effort has failed, or is only working on a subset of available distros, but can't find the comments anymore. I remember issues with both Fedora and Arch, but may be mistaken. Maybe it was on Slack?

It would be good to get an update from @balajiv113 directly.

Anyways, here we are, 2 years later, and we have working code that would fix the issue. How long are we going to wait for somebody to figure out a generic way to do this? What if it turns out not to be possible after all?

balajiv113:
@jandubois & @AkihiroSuda 

On the eBPF PR this is the current state,
- guest direct ports are working perfectly without issues
- docker/ containers with nftables was able to get it working (need to see if it works for all cases)
- But with respect to kubernetes I got no luck. As I previously mentioned kubernetes are removing all entries in iptable and reconstructing back due to this we couldn't identify missing one's properly.

jandubois:
> * docker/ containers with nftables was able to get it working (need to see if it works for all cases)

Thanks, that sounds promising!

So what I remember about some distros not having the prerequisite kernel modules, or missing permissions, was not, or is no longer correct?

go.mod
```diff
@@ -17,10 +17,13 @@ require (
 	github.com/cpuguy83/go-md2man/v2 v2.0.7
 	github.com/digitalocean/go-qemu v0.0.0-20221209210016-f035778c97f7
 	github.com/diskfs/go-diskfs v1.7.0 // gomodjail:unconfined
+	github.com/docker/docker v28.3.3+incompatible
```
balajiv113, line 20: Will it help with respect to code maintenance if we can create a go mod to handle all these.

I did create trackport to handle multiple various implementation of port identification 

https://github.com/balajiv113/trackport/blob/main/pkg/trackapi/api.go

https://github.com/balajiv113/trackport/tree/main/pkg/internal

balajiv113:
> So what I remember about some distros not having the prerequisite kernel modules, or missing permissions, was not, or is no longer correct?

I was able to solve it using kprobe. This is available in almost all distro I checked. 

Even our FT were passing except docker that's when it hit the kubernetes issue

jandubois:
> I was able to solve it using kprobe. This is available in almost all distro I checked.

So you are saying we should not merge this PR, but instead wait for you to finish yours?
 
> Even our FT were passing except docker that's when it hit the kubernetes issue

We already have a Kubernetes port monitor in the current guest agent; we will just need to keep it.

pkg/guestagent/events/containerd_linux.go
AkihiroSuda: What backwards compatibility is broken?

AkihiroSuda: The layout of the `/var/lib/nerdctl` directory is an implementation detail, not a part of the public interface.

I'm aware of [Hyrum's Law](https://www.hyrumslaw.com) though 😓

AkihiroSuda:
The combination of  @balajiv113 's eBPF PR (https://github.com/lima-vm/lima/pull/3067) w/ the existing [kubernetesservice watcher](https://github.com/lima-vm/lima/blob/master/pkg/guestagent/kubernetesservice/kubernetesservice.go) seems the most robust and compatible solution for now?

I wish we could remove the dependency on Kubernetes client libraries, but it can be discussed separately in future.

pkg/guestagent/events/containerd_linux.go
jandubois: The backwards compatibility issue is that you used to be able to query container labels for the list of exposed ports. That functionality has been removed in nerdctl 2.1.3 because it broke in an edge case when a large number of ports were exposed.

There is no accommodation for exposing ranges as ranges, they were unrolled to create a label for each port, hitting the containerd limitation when using large ranges. So the labels were removed completely, making it impossible to determine exposed ports via inspection.

jandubois:
> The combination of @balajiv113 's eBPF PR (#3067) w/ the existing [kubernetesservice watcher](https://github.com/lima-vm/lima/blob/master/pkg/guestagent/kubernetesservice/kubernetesservice.go) seems the most robust and compatible solution for now?

Yes, if it works and is indeed robust, then a generic solution is obviously preferable. I was under the impression that this experiment had failed. I'll be happy to learn if I was wrong.

pkg/guestagent/events/containerd_linux.go
AkihiroSuda: The labels are private implementation details too.

The correct way to inspect the port is to:
```console
$ nerdctl container inspect foo | jq .[0].HostConfig.PortBindings
{
  "80/tcp": [
    {
      "HostIp": "0.0.0.0",
      "HostPort": "8080"
    }
  ]
}
```
This works for Docker too.

jandubois: How do you do it over the API? You can do this via the API with docker (and you used to be able to do it with nerdctl containers too).

AkihiroSuda: No containerd API, as the port forwarding is not handled by the daemon

jandubois: I don't really want to argue this anymore, and maybe it is Hyrum's law, but in the end the fact remains that information about the container was available via an API and now is only available by running a subprocess, with the inherent brittleness of that approach (error handling).

PR #4025, opened 2025-09-11
Add foundational BATS tests for `limactl yq` and `limactl list`

jandubois:
They are the basis for more complex tests later, so have filenames starting with a digit, so they are processed first.

May still be missing some edge cases, but they can be added later.

### DIFF
--- a/hack/bats/helpers/limactl.bash
+++ b/hack/bats/helpers/limactl.bash
@@ -1,0 +1,16 @@
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Create a dummy Lima instance for testing purposes. It cannot be started because it doesn't have an actual image.
+# This function intentionally doesn't use create/editflags, but modifies the template with yq instead.
+create_dummy_instance() {
+    local name=$1
+    local expr=$2
+
+    # Template does not validate without an image, and the image must point to a file that exists (for clonefile).
+    local template="{images: [location: /etc/profile]}"
+    if [[ -n $expr ]]; then
+        template="$(limactl yq "$expr" <<<"$template")"
+    fi
+    limactl create --name "$name" - <<<"$template"
+}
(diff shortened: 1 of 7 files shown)


hack/bats/tests/1-list.bats
```diff
@@ -0,0 +1,280 @@
...
+    assert_line --index 1 "bar 2097152"
+}
+
+@test '--list-fields list all available fields' {
+    run -0 limactl ls --list-fields
+    assert_line Name
+    assert_line CPUs
+    assert_line Memory
+    # All field names start with an uppercase letter and don't contain any spaces
+    assert_line --regexp '^[A-Z]'
```
jandubois: This only checks that at least one line starts with an uppercase letter; it doesn't verify that all do.

hack/bats/tests/1-list.bats
```diff
@@ -0,0 +1,280 @@
...
+}
+
+@test '--quiet option can be used with --yq' {
+    run -0 limactl ls --quiet
+    assert_line --index 0 "bar"
+    assert_output_lines_count 3
+
+    run -0 limactl ls --quiet --yq 'select(.name == "foo")'
+    assert_line --index 0 "foo"
+    assert_output_lines_count 1
```
jandubois:
```suggestion
    assert_output "foo"
```

hack/bats/tests/0-yq.bats
AkihiroSuda, line 1: `0-yq.bats` -> `00-yq.bats` or maybe `000-yq.bats`, for ease of sorting when we have 10+ tests in near future

jandubois: We can do `00-yq.bats`, but after a couple of foundational tests, there is not really an obvious order anymore.

I just wanted `yq` first because `list` uses it, and `list` next because it is the needed to verify results in `2-createflags` and `3-editflags`. Maybe we should have `5-info` as well, as another mechanism to query info about the system.

Eventually we may want to move tests into directories, or group them with tags, to make it easier to execute certain slices of the whole set. But it is hard to define the groupings until we have a meaningful number of tests implemented.

PR #4028, opened 2025-09-11
`limactl usernet`: handle signal to close GVisorNetstack

norio-nomura:
`*_ep.sock`, `*_fd.sock`, and `*_qemu.sock` will be properly removed on exit with the following changes:
- Pass `signal.NotifyContext` to `usernet.StartGVisorNetstack`.
- Close `http.Server` instead of `net.Listener`.
- Fix mishandling of checking `http.ErrServerClosed`.
- Check `ctx.Done()` outside of the loop calling `listener.Accept()` in `listenFD`, and `listenQEMU`.
- Use `errors.Is(err, net.ErrClosed)` in `UDPFileConn.Read`.  
  Since https://github.com/golang/go/issues/4373 has been fixed.

### DIFF
--- a/cmd/limactl/usernet.go
+++ b/cmd/limactl/usernet.go
@@ -7,7 +7,9 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/signal"
 	"strconv"
+	"syscall"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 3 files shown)


norio-nomura:
Thanks! 🙏🏻

PR #4031, opened 2025-09-12
`portfwd`: close `ClosableListeners` on exit

norio-nomura:
`HostSocket` will be deleted on close.

Separated from #4008

### DIFF
--- a/pkg/portfwd/listener.go
+++ b/pkg/portfwd/listener.go
@@ -35,6 +35,21 @@ func NewClosableListener() *ClosableListeners {
 	}
 }
 
+func (p *ClosableListeners) Close() {
+	p.listenersRW.Lock()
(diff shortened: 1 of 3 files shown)


norio-nomura:
Thanks! 🙏🏻

pkg/portfwd/listener.go
```diff
@@ -90,6 +105,9 @@ func (p *ClosableListeners) forwardTCP(ctx context.Context, client *guestagentcl
 	for {
 		conn, err := tcpLis.Accept()
 		if err != nil {
+			if opErr, ok := err.(*net.OpError); ok && opErr.Err.Error() == "use of closed network connection" {
```
norio-nomura, line 108: This should be changed to use `if errors.Is(err, net.ErrClosed) {`
opened #4048

PR #4032, opened 2025-09-12
Change `limactl list --format json` not to include empty objects

jandubois:
Unlike the YAML encoder, the JSON encoder will encode empty structs even when they have the `json:",omitempty"` tag (it will only emit nil pointers to structs).

This commit repeatedly removes empty objects until the result no longer gets any shorter. It doesn't check for the omitempty tag, but we have that tag set on every struct.

Example: this removes `"vmOpts":{"qemu":{},"vz":{"rosetta":{}}}`.

### DIFF
--- a/hack/bats/tests/01-list.bats
+++ b/hack/bats/tests/01-list.bats
@@ -79,10 +79,10 @@ local_setup() {
 }
 
 @test '--json is shorthand for --format json' {
-    run -0 limactl ls --format json foo bar
+    run -0 limactl ls foo bar --format json
     format_json=$output
 
-    run -0 limactl ls --json foo bar
+    run -0 limactl ls foo bar --json
     assert_output "$format_json"
 }
 
@@ -91,7 +91,7 @@ local_setup() {
     run -0 limactl ls foo bar --format json
     json=$output
 
-    run -0 limactl ls --format yaml foo bar
+    run -0 limactl ls foo bar --format yaml
     yaml=$output
 
     assert_line --regexp '^name: foo'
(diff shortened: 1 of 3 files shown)


jandubois:
> Ah, it only changes the output format for TTY - then it should be OK

The change to show pretty-printed JSON for TTY output only was already in #3998.

This PR is only removing empty objects from the JSON output to make it identical to the YAML output.

The good thing is we have tests now to make sure the output stays in JSON Lines format when written to a pipe or file.

hack/bats/tests/01-list.bats
```diff
@@ -93,20 +79,19 @@ canonical_json() {
 }
 
 @test '--json is shorthand for --format json' {
-    run -0 limactl ls --format json foo bar
+    run -0 limactl ls foo bar --format json
```
AkihiroSuda, line 82: ?

jandubois: I've changed all (except one) instances of the command to put the `--format` option last to make it easier to tell the difference between the commands.

It is purely for readability. You probably can't tell by looking at the diff; you need to look at the updated file as a whole.

AkihiroSuda: Could be a separate PR, or a separate commit at least?

jandubois: Sure, will change it tomorrow

jandubois: I've moved the formatting changes to a separate commit.

PR #4034, opened 2025-09-15
Fix: `make uninstall` should remove external drivers

unsuman:
Fixes https://github.com/lima-vm/lima/pull/3744#issuecomment-3290200398

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -495,7 +495,11 @@ uninstall:
 		"$(DEST)/bin/kubectl.lima" \
 		"$(DEST)/share/man/man1/lima.1" \
 		"$(DEST)/share/man/man1/limactl"*".1" \
-		"$(DEST)/share/lima" "$(DEST)/share/doc/lima"
+		"$(DEST)/share/lima" \
+		"$(DEST)/share/doc/lima" \
+		"$(DEST)/libexec/lima/lima-driver-qemu$(exe)" \
+		"$(DEST)/libexec/lima/lima-driver-vz$(exe)" \
+		"$(DEST)/libexec/lima/lima-driver-wsl2$(exe)"
 	if [ "$$(readlink "$(DEST)/bin/nerdctl")" = "nerdctl.lima" ]; then rm "$(DEST)/bin/nerdctl"; fi
 	if [ "$$(readlink "$(DEST)/bin/apptainer")" = "apptainer.lima" ]; then rm "$(DEST)/bin/apptainer"; fi
 


afbjorklund:
Wouldn't you normally just `uninstall` what you `install`? Those external drivers probably have a matching set

Makefile
```diff
@@ -495,7 +495,8 @@ uninstall:
 		"$(DEST)/bin/kubectl.lima" \
 		"$(DEST)/share/man/man1/lima.1" \
 		"$(DEST)/share/man/man1/limactl"*".1" \
-		"$(DEST)/share/lima" "$(DEST)/share/doc/lima"
+		"$(DEST)/share/lima" "$(DEST)/share/doc/lima" \
+		"$(DEST)/libexec/lima"
```
jandubois: I wonder if deleting all of `/usr/local/libexec/lima` is the right thing.

I assumed external drivers that are not bundled by Lima would also be installed there. In which case deleting them would be wrong.

Same thing for additional plugins. Can they also be installed in `/usr/local/libexec/lima`? If they can, then again, we should not wipe them.

So to be safe we should just delete:

```suggestion
		"$(DEST)/libexec/lima/lima-driver-qemu$(exe)" \
		"$(DEST)/libexec/lima/lima-driver-vz$(exe)" \
		"$(DEST)/libexec/lima/lima-driver-wsl$(exe)" \
		"$(DEST)/libexec/lima/limactl-mcp$(exe)"
```

jandubois:
> Wouldn't you normally just `uninstall` what you `install`? Those external drivers probably have a matching set

What you install depends on additional environment settings, which may have changed when you run `uninstall`. So I think you should install all things you possibly could have installed, but nothing else.

jandubois:
Note that in the case of `/usr/local/share/lima` and `/usr/local/share/doc/lima` we completely wipe the directories, which seems wrong too. At least for the docs it is conceivable that you want to install e.g. external plugin docs there too.

It is not clear to me if we support storing custom templates in `/usr/local/share/lima/templates` or not. There are alternatives though, like `$LIMA_HOME/_templates` or any directory on `LIMA_TEMPLATES_PATH`.

Makefile
```diff
@@ -495,7 +495,10 @@ uninstall:
 		"$(DEST)/bin/kubectl.lima" \
 		"$(DEST)/share/man/man1/lima.1" \
 		"$(DEST)/share/man/man1/limactl"*".1" \
-		"$(DEST)/share/lima" "$(DEST)/share/doc/lima"
+		"$(DEST)/share/lima" "$(DEST)/share/doc/lima" \
```
jandubois: I know you are just appending the final backslash here, but I think having multiple entries on one line is bad, and it should be split up. Sorry for not noticing in my preview review:

```suggestion
		"$(DEST)/share/lima" \
		"$(DEST)/share/doc/lima" \
```

unsuman: No worries!

Makefile
```diff
@@ -495,7 +495,11 @@ uninstall:
 		"$(DEST)/bin/kubectl.lima" \
 		"$(DEST)/share/man/man1/lima.1" \
 		"$(DEST)/share/man/man1/limactl"*".1" \
-		"$(DEST)/share/lima" "$(DEST)/share/doc/lima"
+		"$(DEST)/share/lima" \
+		"$(DEST)/share/doc/lima" \
+		"$(DEST)/libexec/lima/lima-driver-qemu$(exe)" \
+		"$(DEST)/libexec/lima/lima-driver-vz$(exe)" \
+		"$(DEST)/libexec/lima/lima-driver-wsl$(exe)"
```
jandubois: Isn't the driver name `wsl2` and not just `wsl`?

```Makefile
ifneq (,$(findstring wsl2,$(ADDITIONAL_DRIVERS)))
LIMACTL_DRIVER_TAGS += external_wsl2
endif
```

```suggestion
		"$(DEST)/libexec/lima/lima-driver-wsl2$(exe)"
```

jandubois (approved):

PR #4043, opened 2025-09-17
`Makefile`: Fix `templates/{_default,_images}` handling

norio-nomura:
This change prevents the creation of `_output/share/lima/templates/{_default/_default,_images/_images}` directories during the second make session.

- Add those directories to `MKDIR_TARRGETS` for applying directory creation rules.
- Add `TEMPLATE_DEFAULTS` and `TEMPLATE_IMAGES` variables for files in those directories.
- Add those files to the dependencies of the `templates` target.
- Add rules for those files to depend on their directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -380,16 +380,20 @@ MKDIR_TARGETS += _output/share/lima
 ################################################################################
 # _output/share/lima/templates
 TEMPLATES = $(addprefix _output/share/lima/templates/,$(filter-out experimental,$(notdir $(wildcard templates/*))))
+TEMPLATE_DEFAULTS = ${addprefix _output/share/lima/templates/_default/,$(notdir $(wildcard templates/_default/*))}
+TEMPLATE_IMAGES = $(addprefix _output/share/lima/templates/_images/,$(notdir $(wildcard templates/_images/*)))
 TEMPLATE_EXPERIMENTALS = $(addprefix _output/share/lima/templates/experimental/,$(notdir $(wildcard templates/experimental/*)))
 
 .PHONY: default_template templates template_experimentals
 default_template: _output/share/lima/templates/default.yaml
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -380,16 +380,20 @@ MKDIR_TARGETS += _output/share/lima
 ################################################################################
 # _output/share/lima/templates
 TEMPLATES = $(addprefix _output/share/lima/templates/,$(filter-out experimental,$(notdir $(wildcard templates/*))))
+TEMPLATE_DEFAULTS = ${addprefix _output/share/lima/templates/_default/,$(notdir $(wildcard templates/_default/*))}
```
AkihiroSuda, line 383: nit: 
```suggestion
TEMPLATES_DEFAULT = ${addprefix _output/share/lima/templates/_default/,$(notdir $(wildcard templates/_default/*))}
```

Makefile
```diff
@@ -380,16 +380,20 @@ MKDIR_TARGETS += _output/share/lima
...
+TEMPLATE_IMAGES = $(addprefix _output/share/lima/templates/_images/,$(notdir $(wildcard templates/_images/*)))
```
AkihiroSuda, line 384:
```suggestion
TEMPLATES_IMAGES = $(addprefix _output/share/lima/templates/_images/,$(notdir $(wildcard templates/_images/*)))
```

norio-nomura: These are in accordance with `TEMPLATE_EXPERIMENTALS`.
Is it better to change that too?

AkihiroSuda: Maybe yes, but probably it is an out of scope of this PR.
Anyway, the naming does not really matter for "private" variables.

norio-nomura:
Thanks! 🙏🏻

PR #4044, opened 2025-09-17
`Makefile`: Add the `DEBUG` flag to build binaries with debug information for use by `dlv exec`.

norio-nomura:
macOS Tahoe 26.0 does not allow the debug target of `dlv dap` to use the `com.apple.security.virtualization` entitlement. See: https://github.com/lima-vm/lima/pull/2651#issuecomment-3298156169

But macOS Tahoe 26.0 allows the precompiled binary target of `dlv exec` to use the entitlement. This adds the `DEBUG` flag to build binaries with debug information for use by `dlv exec`.

### Configuration files for VSCode working on macOS Tahoe 26.0 `launch.json`:
```jsonc
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "exec hostagent for input instance",
            "type": "go",
            "request": "launch",
            "mode": "exec",
            // Use integratedTerminal to stop using ctrl+C
            "console": "integratedTerminal",
            "program": "${workspaceFolder}/_output/bin/limactl",
            "env": {
                "LIMA_SSH_OVER_VSOCK": "false",
                "LIMA_SSH_PORT_FORWARDER": "true",
            },
            "cwd": "${userHome}/.lima/${input:targetInstance}",
            "args": [
                "--debug",
                "hostagent",
                "--pidfile",
                "ha.pid",
                "--socket",
                "ha.sock",
                "--guestagent",
                "${workspaceFolder}/_output/share/lima/lima-guestagent.Linux-aarch64",
                "--progress",
                "${input:targetInstance}"
            ],
            "preLaunchTask": "prepare launching hostagent for target instance",
            "postDebugTask": "clean up after stopping hostagent for target instance",
        },
    ],
    "inputs": [
        {
            "id": "targetInstance",
            "type": "promptString",
            "description": "Input target instance parameter for `limactl` command",
        }
    ]
}
```

`tasks.json`:
```jsonc
{
    "version": "2.0.0",
    "tasks": [
        {
            "label": "prepare launching hostagent for target instance",
            "dependsOn":[
                "execute: make CONFIG_GUESTAGENT_COMPRESS=n DEBUG=1",
                "execute: rm -f ${userHome}/.lima/${input:targetInstance}/ha.pid",
            ]
        },
        {
            "label": "clean up after stopping hostagent for target instance",
            "dependsOn":[
                "execute: rm -f ${userHome}/.lima/${input:targetInstance}/ha.pid",
            ]
        },
        {
[...]

PR #4045, opened 2025-09-17
`limactl hostagent`: check process is actually running on detecting `ha.pid`

norio-nomura:
Sometimes, after the `limactl hostagent` process stops (e.g., due to a SIGKILL), the `ha.pid` file remains.  
This change checks if the process with the given PID is still running.  
If not, the `ha.pid` file is ignored, simplifying debugging by reducing cleanup steps to remove `ha.pid`.

see: https://github.com/lima-vm/lima/pull/4044#discussion_r2354249708


AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

Issue #4049, opened 2025-09-18
guestagent: `worthCheckingIPTables` logic seems incorrect


AkihiroSuda:
https://github.com/lima-vm/lima/blob/6c5ee213f3c699736f0780278eba0524bed83006/pkg/guestagent/guestagent_linux.go#L56-L74

https://github.com/lima-vm/lima/blob/6c5ee213f3c699736f0780278eba0524bed83006/pkg/guestagent/guestagent_linux.go#L85-L87

The former part sets `a.worthCheckingIPTables = true` when audit is enabled.
OTOH, the latter one sets `a.worthCheckingIPTables = true` when audit is disabled.

ashwat287:
When the agent starts and auditing is already enabled, the code sets `a.worthCheckingIPTables = true` and does not start `setWorthCheckingIPTablesRoutine`. The routine is the only place that later sets the flag to false or back to true on `NETFILTER_CFG ` events. Because it is not started in this path, `worthCheckingIPTables` stays true permanently. The code path that would use the cached `a.latestIPTables` (the else branch when the flag is false) never executes. Only when the agent itself enables auditing (initially disabled case) does it launch the routine and allow the flag to change state. This creates inconsistent behavior between “auditing already enabled” and “auditing just enabled” starts.

PR #4059, opened 2025-09-19
guestagent: start iptables audit watcher when auditing already enabled

ashwat287:
Fixes: #4049
Fixes a logic gap where, if auditing was already enabled at startup, the code only set worthCheckingIPTables=true and did not launch setWorthCheckingIPTablesRoutine. Without the routine the flag never changes (never set false after idle, never toggled by NETFILTER_CFG events), so the cached a.latestIPTables path is never used and behavior differs from the “auditing just enabled” case.

Change:
- After confirming auditing is (or becomes) enabled, always set worthCheckingIPTables=true for the initial scan and always start setWorthCheckingIPTablesRoutine.
- Leave the non‑auditing fallback path unchanged.

This makes flag state transitions consistent regardless of initial auditStatus.Enabled.

### DIFF
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -67,11 +67,11 @@ func New(ctx context.Context, newTicker func() (<-chan time.Time, func()), iptab
 				return nil, err
 			}
 		}
-
-		go a.setWorthCheckingIPTablesRoutine(auditClient, iptablesIdle)
-	} else {
-		a.worthCheckingIPTables = true
 	}
+
+	a.worthCheckingIPTables = true // allow initial iptables scan
+	go a.setWorthCheckingIPTablesRoutine(auditClient, iptablesIdle)
+
 	logrus.Infof("Auditing enabled (%d)", auditStatus.Enabled)
 	return startGuestAgentRoutines(ctx, a, true), nil
 }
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Thanks, how to test this?

ashwat287:
Shall we create an integration test?

AkihiroSuda:
> Shall we create an integration test?

That would be ideal, but in this case you can just explain how you tested this PR by yourself

ashwat287:
To compare and test the behavior before and after the code changes,

1. I built separate executables for `cmd/lima-guestagent`.
2. Executed them separately and ran the daemon.
3. Made iptables changes.
4. kept observing the logs.

**Observations**:
In the logs after running the executable with the changes in the commit, I observed:

- Just at the start up of daemon, `time="2025-09-24T08:34:30Z" level=info msg="setWorthCheckingIPTablesRoutine(): monitoring netfilter audit events"`. This concludes that, function `setWorthCheckingIPTablesRoutine` is being called which initializes `latestTrue := time.Now()` and enables  `a.worthCheckingIPTables` to be set `false` when left idle.
- After some time: `time="2025-09-24T08:34:50Z" level=debug msg="setWorthCheckingIPTablesRoutine(): setting to false"`. This concludes that  `a.worthCheckingIPTables` is set to `false` when left idle.
- after making iptables changes: `time="2025-09-24T08:36:05Z" level=debug msg="setWorthCheckingIPTablesRoutine(): setting to true"`. Concludes that after changing the iptables `a.worthCheckingIPTables` is again set to `true`.
- Then again after some idle time: `time="2025-09-24T08:36:30Z" level=debug msg="setWorthCheckingIPTablesRoutine(): setting to false"`

Above log events were not logged for the executable without the changes in the commit, which is undesirable as explained in the above comment:

> Fixes a logic gap where, if auditing was already enabled at startup, the code only set worthCheckingIPTables=true and did not launch setWorthCheckingIPTablesRoutine. Without the routine the flag never changes (never set false after idle, never toggled by NETFILTER_CFG events), so the cached a.latestIPTables path is never used and behavior differs from the “auditing just enabled” case.

The above testing strategy shows the difference in behaviors before and after the code changes.

PR #4060, opened 2025-09-20
Create BATS tests for limactl-mcp

jandubois:
I had an idea how to test `limactl-mcp` with BATS, so I wrote this PoC. I think it works well, but must obviously be fleshed out some more.

This PR depends on #3744 and assumes it has been merged already.

### DIFF
--- a/hack/bats/tests/mcp.bats
+++ b/hack/bats/tests/mcp.bats
@@ -1,0 +1,331 @@
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+load "../helpers/load"
+
+NAME=bats
+
+# TODO Move helper functions to shared location
+run_yq() {
+    run -0 --separate-stderr limactl yq "$@"
+}
+
+json_edit() {
+    limactl yq --input-format json --output-format json --indent 0 "$@"
+}
+
+# TODO The reusable Lima instance setup is copied from preserve-env.bats
+# TODO and should be factored out into helper functions.
+local_setup_file() {
+    if [[ -n "${LIMA_BATS_REUSE_INSTANCE:-}" ]]; then
+        run limactl list --format '{{.Status}}' "$NAME"
+        [[ $status == 0 ]] && [[ $output == "Running" ]] && return
+    fi
+    limactl unprotect "$NAME" || :
+    limactl delete --force "$NAME" || :
+    # Make sure that the host agent doesn't inherit file handles 3 or 4.
+    # Otherwise bats will not finish until the host agent exits.
+    limactl start --yes --name "$NAME" template://default 3>&- 4>&-
+}
+
+local_teardown_file() {
+    if [[ -z "${LIMA_BATS_REUSE_INSTANCE:-}" ]]; then
+        limactl delete --force "$NAME"
+    fi
+}
+
+local_setup() {
+    cd "$PATH_BATS_ROOT"
+    coproc MCP { limactl mcp serve "$NAME"; }
+
+    ID=0
+    mcp initialize '{"protocolVersion":"2025-06-18"}'
+
+    # Each mcp request should increment the ID
+    [[ $ID -eq 1 ]]
+
+    run_yq .serverInfo.name <<<"$output"
+    assert_output "lima"
+}
+
+local_teardown() {
+    kill "${MCP_PID:?}" 2>&1 >/dev/null || :
+}
+
+mcp() {
+    local method=$1
+    local params=${2:-}
+
+    local request
+    printf -v request '{"jsonrpc":"2.0","id":%d,"method":"%s"}' "$((++ID))" "$method"
+    if [[ -n $params ]]; then
+        request=$(json_edit ".params=${params}" <<<"$request")
+    fi
+
+    # send request to MCP server stdin
+    echo "$request" >&"${MCP[1]}"
+
+    # read response from MCP server stdout with 5s timeout
+    local json
+    read -t 5 -r json <&"${MCP[0]}"
+
+    # verify that the response matches the request; also validates the output is valid JSON
+    run_yq .id <<<"$json"
+    assert_output "$ID"
+
+    # there must be no error object in the response
+    run_yq .error <<<"$json"
+    assert_output "null"
+
+    # set $output to .result
+    run_yq .result <<<"$json"
+}
+
+tools_call() {
+    local name=$1
+    local args=${2:-}
+
+    local params
+    printf -v params '{"name":"%s"}' "$name"
+    if [[ -n $args ]]; then
+        params=$(json_edit ".arguments=${args}" <<<"$params")
+    fi
+    mcp tools/call "$params"
+}
+
+@test 'list tools' {
+    mcp tools/list
+    run_yq '.tools[].name' <<<"$output"
+    assert_line glob
+    assert_line list_directory
+    assert_line read_file
+    assert_line run_shell_command
+    assert_line search_file_content
+    assert_line write_file
+}
+
+@test 'verify that tools descriptions include input and output schema' {
+    mcp tools/list
+    run_yq '.tools[] | select(.name == "run_shell_command")' <<<"$output"
+    json=$output
+
+    run_yq '.inputSchema.required[]' <<<"$json"
+    assert_line command
+    assert_line directory
+    assert_output_lines_count 2
+
+    run_yq '.inputSchema.properties | keys[]' <<<"$json"
+    assert_line command
+    assert_line description
+    assert_line directory
+    assert_output_lines_count 3
+
+    run_yq '.outputSchema.required[]' <<<"$json"
+    assert_line stdout
+    assert_line stderr
+    assert_output_lines_count 2
+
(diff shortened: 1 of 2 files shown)


jandubois:
One thing I found while writing the test was that `run_shell_command` returns an object that has additional JSON embedded as a string:

```console
$ echo '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"run_shell_command","arguments":{"command":["cat","os-release"],"directory":"/etc"}}}' >&"${MCP[1]}"

$ read -t 1 -r line <&"${MCP[0]}"; jq . <<<"$line"
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "{\"stdout\":\"PRETTY_NAME=\\\"Ubuntu 25.04\\\"\\nNAME=\\\"Ubuntu\\\"\\nVERSION_ID=\\\"25.04\\\"\\nVERSION=\\\"25.04 (Plucky Puffin)\\\"\\nVERSION_CODENAME=plucky\\nID=ubuntu\\nID_LIKE=debian\\nHOME_URL=\\\"https://www.ubuntu.com/\\\"\\nSUPPORT_URL=\\\"https://help.ubuntu.com/\\\"\\nBUG_REPORT_URL=\\\"https://bugs.launchpad.net/ubuntu/\\\"\\nPRIVACY_POLICY_URL=\\\"https://www.ubuntu.com/legal/terms-and-policies/privacy-policy\\\"\\nUBUNTU_CODENAME=plucky\\nLOGO=ubuntu-logo\\n\",\"stderr\":\"\",\"exit_code\":0}"
      }
    ]
  }
}
```

So you need to extract `.text`, unmarshall the result into JSON and then extract `.stdout`, which is quite cumbersome. I don't know if this will bother an LLM, but doesn't this also increase the number of tokens used for the content by escaping all the quotes and backslashes?

Is there a reason this can't be a normal nested object?

jandubois:
Beyond the encoded JSON thing, I don't understand the schema of the output: Why is `content` an array? I don't understand how there can be more than one set of results given that the tool only accepts a single command.

And what does `type: text` mean? Especially given that `text` is a json string. Could it also be `object`? Isn't this all too complex for a tool command?

So I would expect the output of `run_shell_command` to look more like this:

```json
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "stdout": "...",
    "stderr": "",
    "exit_code": 0
  }
}
```

jandubois:
The `glob` and `list_directory` tools seem redundant:

If you make the `pattern` for `glob` optional and have it default to `*`, then `list_directory` is just the same as `glob`.

AkihiroSuda:
> So I would expect the output of run_shell_command to look more like this:

MCP doesn't allow that schema: https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk@v0.6.0/mcp#CallToolResult

`StructuredContent` may be useful for a similar schema, but it might be relatively new and maybe not understood by some AI agents.

https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content also says "For backwards compatibility, a tool that returns structured content SHOULD also return the serialized JSON in a TextContent block."

I guess we can revisit this later.


> The glob and list_directory tools seem redundant:

The current design is to follow Gemini's design
https://github.com/google-gemini/gemini-cli/blob/v0.5.5/docs/tools/shell.md

jandubois:
I have more questions:

---

The `glob` description says:

> The absolute path to the directory to search within. If omitted, searches the tool's root directory.

What is the tool's root directory? How is the LLM to know what this means?

If this command accepts relative patterns, why do the other tools insist on an absolute path? Why can't you pass a filename returned by `glob` to `read_file`? How do you turn the relative file into an absolute file?

---

The description for the `run_in_shell` command says:

> The exact shell command to execute. Defined as a string slice, unlike Gemini's run_shell_command that defines it as a single string.

I think the negative explanation "this is not like something else" is not helpful. Also most LLMs will not even know what the Gemini `run_shell_command` does.

Describe what the tool does, and what parameters it expects in a prescriptive manner.

---

The `write_file` tool does not create the directory for the file to be written.

It returns an error `file does not exist`. I think it should create missing directories. But if it doesn't, the error should be `directory does not exist`.

AkihiroSuda:
Opened an issue about StructuredContent for tracking purpose:
- #4070 

Please consider opening separate issues for other topics too.

> What is the tool's root directory?

The prompt was just copied from https://github.com/google-gemini/gemini-cli/blob/v0.5.5/docs/tools/file-system.md#4-glob-findfiles
Needs to be reworded that it refers to the cwd of the agent, which should correspond to the cwd of the MCP tools too.

> If this command accepts relative patterns, why do the other tools insist on an absolute path? 

The prompt was just copied from Gemini, but `glob` seems different from other tools in the sense that it is specifically made for resolving a vague path pattern to an exact pattern.

> Why can't you pass a filename returned by glob to read_file?

Bug. `glob` should return absolute paths that can be passed to `read_file`.
https://github.com/google-gemini/gemini-cli/blob/v0.5.5/docs/tools/file-system.md#4-glob-findfiles

> How do you turn the relative file into an absolute file?

Maybe we should provide an additional MCP tool, if shelling out `realpath` does not work.

> I think the negative explanation "this is not like something else" is not helpful. Also most LLMs will not even know what the Gemini run_shell_command does.

Gemini may know?

> The write_file tool does not create the directory for the file to be written.

Bug.

jandubois:
> > Why can't you pass a filename returned by glob to read_file?
> 
> Bug. `glob` should return absolute paths that can be passed to `read_file`.

I must have been confused about this; I can no longer reproduce the issue and get full path names now. I probably mixed it up with the results of `list_directory`, which has relative paths because the directory name is a required parameter.

AkihiroSuda:
Needs rebase after merging:
- #4119

jandubois:
I think this is good enough for a first pass. There is always more that could be covered, but there is at least some coverage now for each tool.

There are 2 skipped tests that should be enabled when the bugs #4173 and #4174 have been fixed.

hack/bats/tests/mcp.bats
```diff
@@ -0,0 +1,331 @@
...
+local_setup_file() {
+    if [[ -n "${LIMA_BATS_REUSE_INSTANCE:-}" ]]; then
+        run limactl list --format '{{.Status}}' "$NAME"
+        [[ $status == 0 ]] && [[ $output == "Running" ]] && return
+    fi
+    limactl unprotect "$NAME" || :
+    limactl delete --force "$NAME" || :
+    # Make sure that the host agent doesn't inherit file handles 3 or 4.
+    # Otherwise bats will not finish until the host agent exits.
+    limactl start --yes --name "$NAME" template://default 3>&- 4>&-
```
AkihiroSuda, line 28: Not necessarily needs to be covered in this PR, but we should consider reusing an instance across multiple bats files


https://github.com/lima-vm/lima/blob/8b1f07db05989633984093cace6abb5ee2cf90a5/hack/bats/tests/preserve-env.bats#L20

jandubois: Yes, absolutely.

Just for working on the tests you can already do this:

```console
❯ export LIMA_BATS_REUSE_INSTANCE=1

❯ ./hack/bats/lib/bats-core/bin/bats -T ./hack/bats/tests/mcp.bats -f search_file
mcp.bats
 ✓ search_file_content finds text inside files [507]
 ✓ search_file_content can find unicode characters above U+FFFF [510]
 ✓ search_file_content returns an empty string if it cannot find the pattern [502]

3 tests, 0 failures in 2 seconds

❯ ./hack/bats/lib/bats-core/bin/bats -T ./hack/bats/tests/mcp.bats -f search_file
mcp.bats
 ✓ search_file_content finds text inside files [516]
 ✓ search_file_content can find unicode characters above U+FFFF [512]
 ✓ search_file_content returns an empty string if it cannot find the pattern [501]

3 tests, 0 failures in 3 seconds
```

So you can iterate quickly without restarting the instance every time.

And both `mcp.bats` and `preserve-env.bats` are sharing the same `bats` instance, so it works for both.

AkihiroSuda: Nice, the code clone across the bats files should be moved to `helpers`

jandubois:
I don't know if #4124 should be considered a bug or not. Right now this test assumes the current scheme is the way it is supposed to be.

hack/bats/tests/mcp.bats
```diff
@@ -0,0 +1,331 @@
...
+    assert_line --regexp '/tests/mcp.bats$'
+}
+
+@test 'glob returns an empty list when the pattern does not match' {
+    skip "see https://github.com/lima-vm/lima/issues/4173"
+
+    tools_call glob '{"pattern":"nothing.to.see"}'
+
+    run_yq '.structuredContent.matches[]' <<<"$output"
+    assert_output_lines_count 0
```
jandubois, line 306: This may need to be changed to `refute_output` because I think `$lines` is undefined and not an empty array when there is no output. Can be fixed when the test is no longer skipped.

AkihiroSuda (approved):
Thanks

PR #4066, opened 2025-09-22
guestagent: ticker: watch sys_exit_bind with eBPF

AkihiroSuda:
The event watcher is now triggered immediately on `sys_exit_bind`, not waiting for the next 3-second tick.

This commit resolves the long-standing TODO since the initial commit: https://github.com/lima-vm/lima/blob/7459f4587987ed014c372f17b82de1817feffa2e/cmd/lima-guestagent/daemon_linux.go#L57-L63

Close #3067
Close #3766
Close #4021

### DIFF
--- a/cmd/lima-guestagent/daemon_linux.go
+++ b/cmd/lima-guestagent/daemon_linux.go
@@ -16,6 +16,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/guestagent"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/api/server"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/serialport"
+	"github.com/lima-vm/lima/v2/pkg/guestagent/ticker"
 	"github.com/lima-vm/lima/v2/pkg/portfwdserver"
 )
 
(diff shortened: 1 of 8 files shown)


balajiv113:
Does this work for UDP as well ??

jandubois:
I don't have time to test right now, but I think this will not trigger when the port is closed, right?

One common problem with our polling is that test suites create a container with exposed ports, and after stopping the container create another one again in the next test that exposes the same ports. Due to the 3 second polling the old forwards may not have been removed yet, blocking the new container from exposing the same ports again.

Maybe there is another trace point to also kick the ticker, like `sys_exit_close` or `tcp:tcp_close`, if that works reliably?

AkihiroSuda:
> Does this work for UDP as well ??

Yes

> sys_exit_close

Seems too frequent
```
$ sudo bpftrace -e 'tracepoint:syscalls:sys_exit_close { printf("%d %s %s\n", elapsed / 1000 / 1000 / 1000, probe, comm); }'
Attaching 1 probe...                                   
0 tracepoint:syscalls:sys_exit_close tmux: server      
0 tracepoint:syscalls:sys_exit_close tmux: server      
0 tracepoint:syscalls:sys_exit_close tmux: server      
1 tracepoint:syscalls:sys_exit_close tmux: server      
1 tracepoint:syscalls:sys_exit_close tmux: server      
2 tracepoint:syscalls:sys_exit_close tmux: server      
2 tracepoint:syscalls:sys_exit_close tmux: server      
2 tracepoint:syscalls:sys_exit_close containerd        
2 tracepoint:syscalls:sys_exit_close containerd        
2 tracepoint:syscalls:sys_exit_close containerd        
2 tracepoint:syscalls:sys_exit_close containerd 
...
```


> tcp:tcp_close

Doesn't seem to exist in `/sys/kernel/debug/tracing/available_events`

AkihiroSuda:
> One common problem with our polling is that test suites create a container with exposed ports, and after stopping the container create another one again in the next test that exposes the same ports. Due to the 3 second polling the old forwards may not have been removed yet, blocking the new container from exposing the same ports again.

How does this happen? The guest kernel isn't aware of the usage of the host port, so it doesn't block `bind()` in the guest.
After `bind()` in the guest succeeds, the ticker event is triggered, and the hostagent updates the host port status accordingly.

jandubois:
> How does this happen? The guest kernel isn't aware of the usage of the host port, so it doesn't block `bind()` in the guest.
> After `bind()` in the guest succeeds, the ticker event is triggered, and the hostagent updates the host port status accordingly.

I guess the issue used to be that the test would connect to the dangling port on the host. So maybe just looking for bind events is enough; will need to write a test to verify.

norio-nomura:
After this is merged, the CPU utilization rate of `lima-guestagent` is always close to 90%.

screenshot 2025-09-25 9 36 07

I confirmed that it will be resolved with reverted.

PR #4067, opened 2025-09-22
downloader: CacheKey: include decompression flag

AkihiroSuda:
The caches are now created separately for compressed and decompressed contents. When a decompressed content is cached and a compressed content is requested, the downloader now correctly returns the compressed content.

- URLSHA/data:                 unmodified original data
- URLSHA/sha256.digest:        digest of the original data
- URLSHA+decomp/data:          decompressed data
- URLSHA+decomp/sha256.digest: digest of the *original* (i.e., compressed) data

Caching a decompressed content does not automatically cache the original compressed content.

### DIFF
--- a/pkg/downloader/downloader.go
+++ b/pkg/downloader/downloader.go
@@ -230,7 +230,7 @@ func Download(ctx context.Context, local, remote string, opts ...Opt) (*Result, 
 		return res, nil
 	}
 
-	shad := cacheDirectoryPath(o.cacheDir, remote)
+	shad := cacheDirectoryPath(o.cacheDir, remote, o.decompress)
 	if err := os.MkdirAll(shad, 0o700); err != nil {
 		return nil, err
 	}
@@ -255,7 +255,7 @@ func Download(ctx context.Context, local, remote string, opts ...Opt) (*Result, 
 // nil if the file was copied, nil, nil if the file is not in the cache or the
 // cache needs update, or nil, error on fatal error.
 func getCached(ctx context.Context, localPath, remote string, o options) (*Result, error) {
-	shad := cacheDirectoryPath(o.cacheDir, remote)
+	shad := cacheDirectoryPath(o.cacheDir, remote, o.decompress)
 	shadData := filepath.Join(shad, "data")
 	shadTime := filepath.Join(shad, "time")
 	shadType := filepath.Join(shad, "type")
@@ -301,7 +301,7 @@ func getCached(ctx context.Context, localPath, remote string, o options) (*Resul
 
 // fetch downloads remote to the cache and copy the cached file to local path.
 func fetch(ctx context.Context, localPath, remote string, o options) (*Result, error) {
-	shad := cacheDirectoryPath(o.cacheDir, remote)
+	shad := cacheDirectoryPath(o.cacheDir, remote, o.decompress)
 	shadData := filepath.Join(shad, "data")
 	shadTime := filepath.Join(shad, "time")
 	shadType := filepath.Join(shad, "type")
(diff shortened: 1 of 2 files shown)


pkg/downloader/downloader.go
```diff
@@ -776,8 +776,12 @@ func CacheEntries(opts ...Opt) (map[string]string, error) {
...
 // CacheKey returns the key for a cache entry of the remote URL.
-func CacheKey(remote string) string {
-	return fmt.Sprintf("%x", sha256.Sum256([]byte(remote)))
+func CacheKey(remote string, decompress bool) string {
+	k := fmt.Sprintf("%x", sha256.Sum256([]byte(remote)))
+	if decompress {
+		k += "+decomp"
+	}
+	return k
 }
```
AkihiroSuda, line 784: (On second thought this might be rather confusing when the original data is not compressed)

AkihiroSuda:
ping @lima-vm/maintainers This PR has been open for more than 4 months

pkg/downloader/downloader.go
```diff
@@ -775,8 +775,12 @@ func CacheEntries(opts ...Opt) (map[string]string, error) {
...
 
 // CacheKey returns the key for a cache entry of the remote URL.
-func CacheKey(remote string) string {
-	return fmt.Sprintf("%x", sha256.Sum256([]byte(remote)))
+func CacheKey(remote string, decompress bool) string {
+	k := fmt.Sprintf("%x", sha256.Sum256([]byte(remote)))
+	if decompress {
```
unsuman: To address your [concern](https://github.com/lima-vm/lima/pull/4067/changes#r2369320486), can we?

```suggestion
	if decompress && decompressor(remote) != "" {
```

AkihiroSuda: Thanks, applied

pkg/downloader/downloader.go
```diff
@@ -775,8 +775,12 @@ func CacheEntries(opts ...Opt) (map[string]string, error) {
 }
 
 // CacheKey returns the key for a cache entry of the remote URL.
-func CacheKey(remote string) string {
-	return fmt.Sprintf("%x", sha256.Sum256([]byte(remote)))
+func CacheKey(remote string, decompress bool) string {
+	k := fmt.Sprintf("%x", sha256.Sum256([]byte(remote)))
+	if decompress && decompressor(remote) != "" {
```
jandubois, line 780: I'm sorry, I never got around to reviewing this before it got merged, but I don't see how this can work: `decompressor()` expects a file extension, but you pass the whole URL, so `decompressor(remote)` will always return `""` and you will never append `+decomp` to the cache key.

jandubois: Ideally there should be BATS tests for the downloader functionality, as it has been getting quite complex.

jandubois: Even a unit test comparing `CacheKey("https://example.com/image.qcow2.gz", true)` with `CacheKey("https://example.com/image.qcow2.gz", false)` would detect this specific failure.

jandubois:
Maybe I'm too tired now, but I don't understand how this PR works.

As far as I can tell `fetch()` will always store the raw compressed data, regardless of the value of the decompression flag.

So it looks like if you have both a compressed and an uncompressed entry, they still store the same data, taking twice the space. And the decompression still happens every time `copyLocal()` copies it into an instance directory.

I'll take another look tomorrow, in case I managed to thoroughly confuse myself right now.

pkg/downloader/downloader.go
AkihiroSuda: Opened a PR to fix it:
- https://github.com/lima-vm/lima/pull/4585

AkihiroSuda:
> As far as I can tell fetch() will always store the raw compressed data, regardless of the value of the decompression flag.

You are right. Not sure how I misunderstood this 5 months ago. 🤦 
Let me revert this PR
- https://github.com/lima-vm/lima/pull/4586

PR #2633, opened 2024-09-18
Add k8s-vmnet example

nirs:
This example show how to run multiple connected kubernetes cluster using
a shared network. You can access all the clusters via the shared
network, clusters can access each other.

This configuration was derived from k8s.yaml template, with the
following changes:

- Ensure that the API server and kublet listen on the shared network.
  This is important for pods that need to listen on the host network. An
  example is rook ceph test cluster.

- Disable port forwarding, since we access the host via the shared
  network, and automatic port forwarding cannot work for multiple
  clusters.

- Since we access the API server via the shared network, don't modify
  the kubeconfig copied to the host.

- Enable parallel image pulls for faster provisioning of complex
  clusters.

- Allow unprivileged pods to access block devices. Required for kubevirt
  virtual machines or replicating block volumes using volsync.

These changes were extracted from RamenDR k8s.yaml:
https://github.com/RamenDR/ramen/blob/main/test/drenv/providers/lima/k8s.yaml

Based on #2632 since this depends on this change. Without it we need to fixup the route metric during provisioning.

PR #4068, opened 2025-09-22
add kcp template

olamilekan000:
```
Change adds KCP templates to enable Lima to support multi-tenant Kubernetes control planes out of the box.
```

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -57,6 +57,7 @@ Container orchestration:
 - [`faasd`](./faasd.yaml): [Faasd](https://docs.openfaas.com/deployment/edge/)
 - [`k0s`](./k0s.yaml): [k0s](https://k0sproject.io/) Zero Friction Kubernetes
 - [`k3s`](./k3s.yaml): Kubernetes via k3s
+- [`kcp`](./kcp.yaml): [kcp](https://github.com/kcp-dev/kcp) Kubernetes-like control plane that delivers isolated workspaces and centralized multi-tenant APIs.
 - [`k8s`](./k8s.yaml): Kubernetes via kubeadm
 - [`experimental/u7s`](./experimental/u7s.yaml): [Usernetes](https://github.com/rootless-containers/usernetes): Rootless Kubernetes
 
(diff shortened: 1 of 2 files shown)


jandubois:
I think https://github.com/lima-vm/lima/pull/2633#issuecomment-2359516917 applies to this PR as well.

We can have a list with links to external templates, but we don't want to maintain them as part of the core repo.

olamilekan000:
> I think [#2633 (comment)](https://github.com/lima-vm/lima/pull/2633#issuecomment-2359516917) applies to this PR as well.
> 
> We can have a list with links to external templates, but we don't want to maintain them as part of the core repo.

Oh okay

AkihiroSuda:
> I think https://github.com/lima-vm/lima/pull/2633#issuecomment-2359516917 applies to this PR as well.

The situation was different in that PR; I rejected #2633 because it was overlapping with `k8s` and it could be rather implemented as additional `--set` flags:
- https://github.com/lima-vm/lima/issues/4100

I think we can still add well-known Kubernetes distros, but KCP seems an alternative control plane rather than a Kubernetes distro such as k0s and k3s?

olamilekan000:
> > I think [#2633 (comment)](https://github.com/lima-vm/lima/pull/2633#issuecomment-2359516917) applies to this PR as well.
> 
> The situation was different in that PR; I rejected #2633 because it was overlapping with `k8s` and it could be rather implemented as additional `--set` flags:
> 
> * [template://k8s: support multi-node mode (w/ `.param`) #4100](https://github.com/lima-vm/lima/issues/4100)
> 
> I think we can still add well-known Kubernetes distros, but KCP seems an alternative control plane rather than a Kubernetes distro such as k0s and k3s?

Kcp extends the K8s API server to provide extra capabilities such as multi-tenancy and horizontal scaling of the control plane. By default, it doesn’t include the standard k8s workload APIs (Pods, Deployments, etc.), but can be added if needed.. So we can say it's an alternative CP in a way and to me I don;t think it overlaps with k8s.

PR #4073, opened 2025-09-23
rm docs/*.md; add docs/README.md for the website link

AkihiroSuda:
Two years have passed since migrating the *.md contents to the website, so the *.md files no longer need to be kept.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,1 @@
+Moved to <https://lima-vm.io/docs/>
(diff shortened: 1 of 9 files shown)


afbjorklund:
It is still confusing since they are hiding under `../website` in the same git repository that you already have (offline).

But it will get better when the versioning is fixed, so that you are looking at the right version of the documentation.

PR #4076, opened 2025-09-23
Use driver feature instead of checking the name

afbjorklund:
Other container drivers with dynamic addresses and static ports also want to set the port after start.

This means that we have to provide their names too in "limatype", instead of making it fully external.

```go
if *inst.Config.VMType == limatype.WSL2 || *inst.Config.VMType == limatype.AC || *inst.Config.VMType == limatype.DC
```

There is some weird hard-coded path handling left, that could be the topic of another PR perhaps?

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -105,6 +105,7 @@ type Info struct {
 type DriverFeatures struct {
 	CanRunGUI            bool `json:"canRunGui,omitempty"`
 	DynamicSSHAddress    bool `json:"dynamicSSHAddress"`
+	StaticSSHPort        bool `json:"staticSSHPort"`
 	SkipSocketForwarding bool `json:"skipSocketForwarding"`
 	NoCloudInit          bool `json:"noCloudInit"`
 	RosettaEnabled       bool `json:"rosettaEnabled"`
--- a/pkg/driver/wsl2/wsl_driver_windows.go
+++ b/pkg/driver/wsl2/wsl_driver_windows.go
@@ -309,6 +309,7 @@ func (l *LimaWslDriver) Info() driver.Info {
 
 	info.Features = driver.DriverFeatures{
 		DynamicSSHAddress:    true,
+		StaticSSHPort:        true,
 		SkipSocketForwarding: true,
 		NoCloudInit:          true,
 		CanRunGUI:            l.canRunGUI(),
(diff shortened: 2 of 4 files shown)


afbjorklund:
There is a chicken-and-egg problem here, where the driver must be provided with a host port on creation...

Even though it is ignored, and then replaced with the real address and port as a side-effect of `InspectStatus`

unsuman:
> There is a chicken-and-egg problem here, where the driver must be provided with a host port on creation...

Yes, I tried to tackle this in #3901, but was unable to do so due to time constraints. What are the possible ways to tackle this? 🤔 
Earlier, I tried to first **configure** a driver with the port from the `determineSSHLocalPort()` func and then if the driver possessed the `dynamicSSHAddress` then we would have to create the driver once again

afbjorklund:
Another problem is that the ssh.config is written with the dummy response, so it remains at `127.0.0.1:22` forever.

So while the instance is updated, and `limactl ls` and `limactl shell` work, you can't use plain `ssh` to connect to it.

afbjorklund:
We will need to fix the issues with "no-op and side-effects" driver API and the address/port allocation separately.

* https://github.com/lima-vm/lima/pull/3988

AkihiroSuda:
WSL2 CI failing

pkg/driver/driver.go
```diff
@@ -105,6 +105,7 @@ type Info struct {
 type DriverFeatures struct {
 	CanRunGUI            bool `json:"canRunGui,omitempty"`
 	DynamicSSHAddress    bool `json:"dynamicSSHAddress"`
+	StaticSSHPort        bool `json:"staticSSHPort"`
```
AkihiroSuda, line 108: For consistency with `DynamicSSHAddress`, can we invert the logic and make this `DynamicSSHPort` ?

afbjorklund: We could, but then it would be true by default - where we have mostly tried to keep our struct false by default

afbjorklund: Since DynamicSSHAddress doesn't really do much at the moment, we could look at revising both of them?

afbjorklund: * https://github.com/lima-vm/lima/pull/3988#issuecomment-3310698807

AkihiroSuda (approved):
Thanks

pkg/driver/driver.go
AkihiroSuda: Let's keep `StaticSSHPort` then.

`DynamicSSHAddress` can be revisited in a separate issue

afbjorklund: I will draft a new PR, how we can revise the API (hopefully before 2.0 is out)

Move `InspectStatus` to the core, and make `SSHAddress` and `GetStatus` work

PR #4079, opened 2025-09-24
Make sure we always run BATS with --errexit

jandubois:
Being able to use https://github.com/lima-vm/bats-core/commit/507da84de75fa78798d53eceb42e68851ef5c48b was one of the reasons to use submodules for the bats libraries.

It looks like I forgot to apply a change to enforce that we always use it, and don't have to specify it on the commandline. Without this setting, command failure in helper functions invoked via `run` will simply be ignored instead of failing the test.


hack/bats/helpers/load.bash
```diff
@@ -3,6 +3,9 @@
 
 set -o errexit -o nounset -o pipefail
 
+# Make sure run() will execute all functions with errexit enabled
```
AkihiroSuda: Please add a comment line to clarify that this depends on non-merged PR
https://github.com/bats-core/bats-core/pull/1118

PR #4082, opened 2025-09-24
pkg/hostagent: Remove `ControlMaster` related args from reversesshfs config on Windows

norio-nomura:
HostAgent's `sshConfig` already has some ControlMaster related args in `AdditionalArgs`, so it is necessary to remove them to avoid overrides above `Persist=false`.

### DIFF
--- a/pkg/hostagent/mount.go
+++ b/pkg/hostagent/mount.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"slices"
+	"strings"
 
 	"github.com/lima-vm/sshocker/pkg/reversesshfs"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


norio-nomura:
expecting to avoids error like:
https://github.com/lima-vm/lima/actions/runs/17967205396/job/51101966512?pr=4081#step:7:108
> time="2025-09-24T05:38:26Z" level=info msg="[hostagent] ControlSocket /c/Users/runneradmin/.lima/default/ssh.sock already exists, disabling multiplexing\r"

norio-nomura:
In addition, I think I will make a PR to operate args related to `ControlMaster`, so I will reopen it after that.

norio-nomura:
Updated to rebased on #4093

norio-nomura:
Thanks! 🙏🏻

Issue #4036, opened 2025-09-15
Improve flexibility of `limactl shell --preserve-env`


jandubois:
I would like to make the following changes to the `--preserve-env` functionality:

1. Throw an error when the patterns in `LIMA_SHELLENV_ALLOW` or `LIMA_SHELLENV_BLOCK` use any characters outside of `[a-zA-Z0-9_]`.

2. Allow `*` wildcards anywhere in a pattern and not just at the end.

3. Setting `LIMA_SHELLENV_BLOCK` should no longer be ignored when `LIMA_SHELLENV_ALLOW` is set. The allow list should just override the block list entries, but additional variables that are not blocked should still be implicitly allowed.

    The current behaviour is still available by setting `LIMA_SHELLENV_BLOCK=*`, blocking everything not explicitly allowed by `LIMA_SHELLENV_ALLOW`.

These changes make more comprehensive block lists possible (like `*TOKEN*` patterns), and reserve additional meta-characters that are not valid in environment variable names anyways. Combined they will also make it possible to implement ideas like #3852 later without breaking backwards compatibility. So I would like to see these changes in the 2.0 release.

Of course (3) is changing functionality right now, but we haven't released `--preserve-env` in a release yet, and I think the added functionality is really desirable, especially when you want to use more comprehensive block lists.

---

Optional (can be added after 2.0), but I think worthwhile and easy to implement:

* Add `--allow-env PATTERN` and `--block-env PATTERN` options to `limactl shell`.

They would (just for this command) append the pattern to the respective lists and automatically enable `--preserve-env`.

These commands:

```shell
limactl shell --allow-env "GITHUB_TOKEN" default gh ...
limactl shell --block-env "PRIVATE*" default ...
```

would be equivalent to

```shell
LIMA_SHELLENV_ALLOW="$LIMA_SHELLENV_ALLOW,GITHUB_TOKEN" limactl shell --preserve-env default gh ...
LIMA_SHELLENV_BLOCK="$LIMA_SHELLENV_BLOCK,PRIVATE*" limactl shell --preserve-env default ...
```

Except in case `LIMA_SHELLENV_BLOCK` was empty, it would be adding the pattern to the default list, not the empty list:

```shell
LIMA_SHELLENV_BLOCK="+PRIVATE*" limactl shell --preserve-env default ...
```

olamilekan000:
kindly assign to me

AkihiroSuda:
SGTM

PR #4101, opened 2025-09-26
improve flexibility of limactl shell cmd

olamilekan000:
```
change improves the flexibility to the limactl command when used with 
the --preserve-env flag.
```

Fixes 
https://github.com/lima-vm/lima/issues/4036


pkg/envutil/envutil.go
```diff
@@ -82,8 +112,39 @@ func matchesPattern(name, pattern string) bool {
 		return true
 	}
 
-	prefix, found := strings.CutSuffix(pattern, "*")
-	return found && strings.HasPrefix(name, prefix)
+	parts := strings.Split(pattern, "*")
```
jandubois: I'm just browsing the PR, not doing a proper review, but this approach is too complicated.

Since we have already verified that there are no metacharacters in the pattern except for `*` we can simply replace all `*` by `.*` and wrap the pattern with `^` and `$` to turn it into a regular expression match:

`*FOO*` → `^.*FOO.*$`

pkg/envutil/envutil_test.go
```diff
@@ -213,3 +208,53 @@ func TestGetDefaultBlockList(t *testing.T) {
...
+		{"with comma", "FOO,BAR", false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := validatePattern(tt.pattern)
+			if tt.valid {
+				assert.NilError(t, err, "Expected pattern %q to be valid", tt.pattern)
+			} else {
+				assert.Error(t, err, "pattern \""+tt.pattern+"\" contains invalid characters. Only alphanumeric characters and underscores are allowed")
```
jandubois: Asterisks `*` are also allowed.

pkg/envutil/envutil.go
```diff
@@ -42,27 +44,55 @@ var defaultBlockList = []string{
 	"_*", // Variables starting with underscore are typically internal
 }
 
+func validatePattern(pattern string) error {
+	validPattern := regexp.MustCompile(`^[a-zA-Z0-9_*]*$`)
```
jandubois: If you use a negated character class inside a capture group, then you could report the exact character that is not allowed in the pattern (well, at least the first one, in case there are multiple).

pkg/envutil/envutil.go
```diff
@@ -42,27 +44,55 @@ var defaultBlockList = []string{
...
+	if !validPattern.MatchString(pattern) {
+		return fmt.Errorf("pattern %q contains invalid characters. Only alphanumeric characters, underscores, and asterisk are allowed", pattern)
+	}
+	return nil
+}
+
 // getBlockList returns the list of environment variable patterns to be blocked.
 // The second return value indicates whether the list was explicitly set via LIMA_SHELLENV_BLOCK.
```
jandubois: There is no seconds return value anymore.

pkg/envutil/envutil.go
```diff
@@ -42,27 +44,55 @@ var defaultBlockList = []string{
...
+			logrus.Fatalf("Invalid LIMA_SHELLENV_BLOCK pattern: %v", err)
+		}
+	}
+
 	if !found {
-		return parseEnvList(blockEnv), true
+		return patterns
 	}
-	return slices.Concat(defaultBlockList, parseEnvList(after)), true
+	return slices.Concat(defaultBlockList, patterns)
```
jandubois: This looks complicated. I would write something like this:

```go
append := strings.HasPrefix(blockEnv, "+")
patterns := parseEnvList(strings.TrimPrefix(blockEnv, "+"))

// ... copy validation loop from above

if append {
    return slices.Concat(...)
}
return patterns
```

pkg/envutil/envutil.go
```diff
@@ -96,17 +132,10 @@ func matchesAnyPattern(name string, patterns []string) bool {
...
 // The filtering is controlled by LIMA_SHELLENV_BLOCK and LIMA_SHELLENV_ALLOW environment variables.
 func FilterEnvironment() []string {
-	allowList, isAllowListSet := getAllowList()
-	blockList, isBlockListSet := getBlockList()
-
-	if isBlockListSet && isAllowListSet {
-		logrus.Warn("Both LIMA_SHELLENV_BLOCK and LIMA_SHELLENV_ALLOW are set. Block list will be ignored.")
-		blockList = nil
-	}
 	return filterEnvironmentWithLists(
```
jandubois, line 135: You don't seem to have updated `filterEnvironmentWithLists()`, so it will still ignore the block list and filter out all variables not in the allow list.

hack/bats/tests/preserve-env.bats
```diff
@@ -93,13 +93,13 @@ local_setup() {
     assert_line BAR=bar
 }
 
-@test 'wildcard does only work at the end of the pattern' {
+@test 'wildcard works at the start of the pattern' {
```
jandubois, line 96: The test should be expanded to test wildcards at the beginning, the middle, and at the end, like `*FOO*BAR*` with various names that match/don't match.

There should also be tests for the new allow list functionality, plus one that combines it with the `*` block list to verify that it implements the old functionality. Adding the test should show that the implementation is not yet complete.

pkg/envutil/envutil.go
olamilekan000: I got a little confused with the the behaviour of the `LIMA_SHELLENV_*.` With the current proposed change if I have something like this

```
  "LO_TOKEN": "lo_development",
  "ME_TOKEN": "me_development",
  "LIMA_SHELLENV_ALLOW": "FOO",
  "LIMA_SHELLENV_BLOCK": "FOO,*TOKEN",
``` 

`FOO` will make it to the allow list however, `TOKEN` will be blocked, but then, what happens to the default blocklist, do we still enforce them or since the `LIMA_SHELLENV_BLOCK` has some explicit values, do they still overwrite the default block list? 

ps: The only time the default block list are included is when we have the blocked list is prefixed with the plus operand "+". ` "LIMA_SHELLENV_BLOCK": "+FOO,*TOKEN"`

Looks like I'm overthinking things...

jandubois: > The only time the default block list are included is when we have the blocked list is prefixed with the plus operand "+".

Yes, or if the user does not set the `LIMA_SHELLENV_BLOCK` at all.

The rules are:

1. Any variables matched by ALLOW will be preserved without checking BLOCK.
2. Any remaining variables matched by BLOCK will not be preserved.
3. Any further unmatched variables will be preserved.

So when BLOCK is `*` then everything not explicitly allowed will be blocked in rule (2) and rule (3) has nothing left to check. This matches the way the code currently works when ALLOW is set: everything not explicitly allowed is blocked.

But now we can also have more flexible setups: the user can define a much more comprehensive BLOCK list, and then allow exceptions on a case-by-case basis.

olamilekan000: okay. Thanks for the clarification.

jandubois:
Please run your tests before pushing to the PR:

```
--- FAIL: TestFilterEnvironment (0.00s)
    --- FAIL: TestFilterEnvironment/allowlist (0.00s)
        envutil_test.go:173: assertion failed: 9 (int) != 2 (int)
FAIL
FAIL	github.com/lima-vm/lima/v2/pkg/envutil	0.188s
```

hack/test-preserve-env.sh
jandubois, line 1: I wonder if we shouldn't completely delete this file because the tests are redundant with what we do in BATS now.

If you want to keep these tests, then you should add `LIMA_SHELLENV_BLOCK="*"` to the 2 tests that check the old ALLOW functionality. The first test is meaningless without it, as it would now pass even if you didn't set `LIMA_SHELLENV_ALLOW` at all.

And for the second test you would prove that matching the ALLOW list passes the variable, even if it is matched by the BLOCK list. But for that you have to block `*`.

pkg/envutil/envutil.go
```diff
@@ -39,30 +41,58 @@ var defaultBlockList = []string{
 	"XDG_*",
 	"ZDOTDIR",
 	"ZSH*",
+	"*PASSWORD*",
+	"*TOKEN*",
+	"*SECRET*",
```
jandubois: We did not agree to add more patterns to the default block list. That was part of the extended proposal, which is still under discussion, and not targeted for Lima 2.0.

There are plenty more patterns if we want to catch all the common names for variables containing secrets (see #3852).

pkg/envutil/envutil.go
```diff
@@ -39,30 +41,58 @@ var defaultBlockList = []string{
...
 	}
-	return slices.Concat(defaultBlockList, parseEnvList(after)), true
+	return patterns
 }
 
 // getAllowList returns the list of environment variable patterns to be allowed.
-// The second return value indicates whether the list was explicitly set via LIMA_SHELLENV_ALLOW.
-func getAllowList() ([]string, bool) {
+func getAllowList() []string {
 	if allowEnv := os.Getenv("LIMA_SHELLENV_ALLOW"); allowEnv != "" {
```
jandubois: If is better to do an early exit instead of enclosing a long block of code in an `if` block:

```go
	allowEnv := os.Getenv("LIMA_SHELLENV_ALLOW")
	if allowEnv == "" {
		return nil
	}
	patterns := parseEnvList(allowEnv)
	…

File: pkg/envutil/envutil_test.go [hunk 10]
```diff
@@ -173,9 +168,9 @@ func TestFilterEnvironment(t *testing.T) {
 
 	t.Run("allowlist", func(t *testing.T) {
 		result := filterEnvironmentWithLists(testEnv, []string{"FOO", "USER"}, nil)
-
 		expected := []string{"FOO=bar", "USER=testuser"}
-		assert.Equal(t, len(result), len(expected))
+
+		assert.Equal(t, len(result), len(testEnv))
```
jandubois, line 175: It is not clear from the local variables here why this assertion should hold, so maybe add a comment about it.

pkg/envutil/envutil_test.go
```diff
@@ -213,3 +208,53 @@ func TestGetDefaultBlockList(t *testing.T) {
...
+		{"with numbers", "VAR123", true},
+		{"with trailing asterisk", "FOO*", true},
+		{"with multiple asterisks", "FOO*BAR*", true},
+		{"asterisk at beginning", "*FOO", true},
+		{"asterisk in middle", "FOO*BAR", true},
+		{"only asterisk", "*", true},
+		{"with dash", "FOO-BAR", false},
+		{"with dot", "FOO.BAR", false},
+		{"with space", "FOO BAR", false},
+		{"with slash", "FOO/BAR", false},
```
jandubois: I don't know if such a large list of invalid characters is really necessary here; I think just testing that all allowed characters work, and maybe test for 2 forbidden ones should be enough.

You are missing a test that lowercase characters are allowed.

And maybe you should test the error message, to show that it correctly identifies the invalid character. A single test is sufficient.

hack/bats/tests/preserve-env.bats
jandubois, line 1: I feel there are too many separate tests now; it is becoming a chore just to review all of them and think through each one separately.

I think it should be possible to combine many of them into a single ALLOW and BLOCK setting, and just use additional variables to test the various conditions that should be covered by each of them.

We already have unit test to verify the basic functionality of pattern matching, so this should just confirms that pattern matching works, and then validate the semantics of how ALLOW and BLOCK list work, individually, and combined.

pkg/envutil/envutil.go
olamilekan000: I can exclude them

hack/test-preserve-env.sh
olamilekan000: I think we should delete it, since we have all of them in the bats test.

hack/bats/tests/preserve-env.bats
```diff
@@ -135,16 +125,93 @@ local_setup() {
     assert_line FOO=foo
     assert_line FOOBAR=foobar
     assert_line BAR=bar
-    refute_line --regexp '^BARBAZ='
+    assert_line BARBAZ=barbaz
```
jandubois, line 129: The test should add a variable from the default block list (e.g. `SSH_FOO`) to ALLOW to verify that it is being preserved. Add `FOO*` and `BAR` proves nothing because they would be preserved even if ALLOW is not set at all.

Furthermore this test should set another variable from the default block list, like `SSH_BAR` and validate that it is being blocked.

hack/bats/tests/preserve-env.bats
```diff
@@ -135,16 +125,93 @@ local_setup() {
...
 }
 
-@test 'setting both allow list and block list generates a warning' {
-    export LIMA_SHELLENV_ALLOW=FOO
-    export LIMA_SHELLENV_BLOCK=BAR
+@test 'wildcard patterns work in all positions and combinations' {
+    # Test wildcard at middle, and other combinations
+    export LIMA_SHELLENV_BLOCK="FOO*BAR,*FOO*BAR*,*TEST*,*SUFFIX"
```
jandubois: There is no point in adding `FOO*BAR` because everything it matches is also matched by `*FOO*BAR*`.

I think to keep it simple, this should be the only pattern used in this test.

hack/bats/tests/preserve-env.bats
```diff
@@ -135,16 +125,93 @@ local_setup() {
...
+    assert_line FOOBAZ=foobaz
+    assert_line BAZBAR=bazbar
+    assert_line BAR=bar
+    assert_line NORMAL_VAR=normal_var
+    assert_line UNRELATED=unrelated
+}
+
+@test 'comprehensive allow/block interaction with wildcards and default blocklist' {
+    # Test allowlist with wildcards, and other test rules
+    export LIMA_SHELLENV_ALLOW="SSH_FOO,CUSTOM*,FOO*,*PREFIX,MIDDLE*PATTERN,SUFFIX*"
```
jandubois: Again, I think this is too much. Keep the number of test cases as low as possible while making sure all distinct situations are covered.

When I wrote previously that there are too many tests, I didn't mean to cram them together into a single `@test`, but rather to carefully curate them to cover each edge case, but remove all redundancy.

hack/bats/tests/preserve-env.bats
```diff
@@ -114,37 +114,72 @@ local_setup() {
...
     export FOO=foo
     export FOOBAR=foobar
+    export FOOXYZBAR=fooxyzbar
+    export FOOBAZ=foobaz
+    export BAZBAR=bazbar
     export BAR=bar
-    export BARBAZ=barbaz
+    export XFOOYBARZDOTCOM=xfooybarzdotcom
+    export NORMAL_VAR=normal_var
+    export UNRELATED=unrelated
```
jandubois, line 146: I don't think having a second unrelated variable proves anything, but it doesn't hurt.

hack/bats/tests/preserve-env.bats
```diff
@@ -114,37 +114,72 @@ local_setup() {
...
-@test 'setting both allow list and block list generates a warning' {
-    export LIMA_SHELLENV_ALLOW=FOO
-    export LIMA_SHELLENV_BLOCK=BAR
-    export FOO=foo
-    run -0 --separate-stderr limactl shell --preserve-env "$NAME" printenv FOO
-    assert_output foo
-    assert_stderr --regexp 'level=warning msg="Both LIMA_SHELLENV_BLOCK and LIMA_SHELLENV_ALLOW are set'
+@test 'allowlist overrides default blocklist with wildcards' {
+    export LIMA_SHELLENV_ALLOW="SSH_*,CUSTOM*"
+    export LIMA_SHELLENV_BLOCK="+*TOKEN"
```
jandubois, line 163: I think this test case is mostly redundant with the `allow list can use a , separated list with whitespace ignored` one, except it also tests the "append" functionality. So these tests could be combined.

jandubois (approved):
I think the tests could still be tightened up, but the actual functionality seems to be working as expected.

So I'm going to merge; tests can always be refactored some more later.

Issue #4097, opened 2025-09-25
Third-party template discovery


jandubois:
We have rejected several submissions of additional templates to be bundled with Lima. The main concern is ongoing maintenance:

* We don't want to get bug reports for the templates
* We don't want to update image references
* We don't want to endorse the templates because we don't want to review them

Unfortunately this makes it hard for template authors to advertise their templates, and for users to find them.

---

I propose to create a "self-managing" registry for third-party templates:

### Registry location

The registry will be in the https://github.com/lima-vm/lima.yaml repo.

It will have a `templates.json` file in the root directory which contains the URLs of all known templates and their metadata.

There may be additional GitHub Pages for human consumption.

### Template discovery

Any repo owner can add the `lima.yaml` tag to their repo, promising that the repo contains a `lima.yaml` file in the root directory.

A GitHub action will once per day query GitHub for a list of all repositories with the `lima.yaml` tag. It will verify that the promised `lima.yaml` at the root directory exists. If it does, it will collect metadata (repo description, other repo tags, last modified date for `lima.yaml`). Archived repos are ignored.

A commit with the new `templates.json` will replace the previous version. This automatically takes care of finding new templates and pruning ones that have been renamed or deleted.

We can implement a block-list filter if people start abusing this mechanism to publish misleading or malicious templates. But otherwise the mechanism is fully autonomous.

### User experience

The `templates.json` file will be cached on the user's system. If GitHub supports using ETag or similar, then we can always check; otherwise we can update the cached version once per day (as needed).

Similar to `limactl create --list-templates` we can provide a `limactl create --list-external-templates`.

The menu displayed by `limactl start` can be extended with another entry:

```
 Choose another template (docker, podman, archlinux, fedora, ...)
 Choose an external template
```

We can even extend tab completion for the `github:` scheme (once that is supported).

jandubois:
This proposal could be extended to support e.g. a `.lima-templates` file that can list multiple `lima.yaml` files, so one GitHub repo could provide multiple templates.

But I think this would be a bad idea because we don't want the registry to contain too many similar templates; it just makes it less useful because things become harder to find. So requiring an additional GitHub repo for each additional template creates a natural barrier to spamming the registry with lots of variations of the same template.

We should rather encourage template authors to write templates in such a way that they can be used as `base` templates, with further customization by the user.

Of course the basic mechanism above can always be extended later to allow additional discovery mechanisms if we think they become necessary.

AkihiroSuda:
[...]

PR #4103, opened 2025-09-28
Add meta field to lima.yaml for user annotations

jandubois:
This field will not be used by Lima itself; users can add whatever objects they want.

Possible uses are:

* adding labels that can be used as selectors (see also #3240)

  ```
  limactl edit mysql --set '.meta.database = true'
  limactl list --yq 'select(.config.meta.database).name'
  ```

* adding anything that would normally be just a comment, like a description or keywords

  This makes it possible to query and manipulate the data with applications instead of being solely for human consumption (see also https://github.com/lima-vm/lima/issues/4097#issuecomment-3340228977).

---

I first called the field `metadata`, but it just feels too long, especially since in filter expressions it becomes `.config.metadata`.

It is unfortunate that most of the interesting data for filtering in the `list` command is inside the `.config` object, making all expressions longer and harder to read. The `.name` and `.status` properties are about the only things outside of `.config` that I regularly use.

### DIFF
--- a/pkg/limatype/lima_yaml.go
+++ b/pkg/limatype/lima_yaml.go
@@ -52,11 +52,12 @@ type LimaYAML struct {
 	PropagateProxyEnv *bool          `yaml:"propagateProxyEnv,omitempty" json:"propagateProxyEnv,omitempty" jsonschema:"nullable"`
 	CACertificates    CACertificates `yaml:"caCerts,omitempty" json:"caCerts,omitempty"`
 	// Deprecated: Use VMOpts.VZ.Rosetta instead.
-	Rosetta              Rosetta `yaml:"rosetta,omitempty" json:"rosetta,omitempty"`
-	Plain                *bool   `yaml:"plain,omitempty" json:"plain,omitempty" jsonschema:"nullable"`
-	TimeZone             *string `yaml:"timezone,omitempty" json:"timezone,omitempty" jsonschema:"nullable"`
-	NestedVirtualization *bool   `yaml:"nestedVirtualization,omitempty" json:"nestedVirtualization,omitempty" jsonschema:"nullable"`
-	User                 User    `yaml:"user,omitempty" json:"user,omitempty"`
+	Rosetta              Rosetta        `yaml:"rosetta,omitempty" json:"rosetta,omitempty"`
+	Plain                *bool          `yaml:"plain,omitempty" json:"plain,omitempty" jsonschema:"nullable"`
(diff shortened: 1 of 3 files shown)


pkg/limatype/lima_yaml.go
```diff
@@ -52,11 +52,12 @@ type LimaYAML struct {
...
-	Plain                *bool   `yaml:"plain,omitempty" json:"plain,omitempty" jsonschema:"nullable"`
-	TimeZone             *string `yaml:"timezone,omitempty" json:"timezone,omitempty" jsonschema:"nullable"`
-	NestedVirtualization *bool   `yaml:"nestedVirtualization,omitempty" json:"nestedVirtualization,omitempty" jsonschema:"nullable"`
-	User                 User    `yaml:"user,omitempty" json:"user,omitempty"`
+	Rosetta              Rosetta        `yaml:"rosetta,omitempty" json:"rosetta,omitempty"`
+	Plain                *bool          `yaml:"plain,omitempty" json:"plain,omitempty" jsonschema:"nullable"`
+	TimeZone             *string        `yaml:"timezone,omitempty" json:"timezone,omitempty" jsonschema:"nullable"`
+	NestedVirtualization *bool          `yaml:"nestedVirtualization,omitempty" json:"nestedVirtualization,omitempty" jsonschema:"nullable"`
+	User                 User           `yaml:"user,omitempty" json:"user,omitempty"`
+	Meta                 map[string]any `yaml:"meta,omitempty" json:"meta,omitempty" jsonschema:"nullable"`
```
AkihiroSuda, line 60: `FillDefault` needs to be updated for overriding

AkihiroSuda: Bikeshedding: why not call this "labels"

jandubois: > Bikeshedding: why not call this "labels"

1. Because I think of the value of labels as just strings (`map[string]string`), not objects, and I don't want to put any constraints on the data a user wants to put there.

2. I find `meta` more generic than `labels`, e.g. `meta.keywords` looks better to me than `labels.keywords`. And with point (1), `meta.keywords` can be a `[]string`, but `labels.keywords` would be a comma-separated list that would need to be parsed and could not be used easily with `jq`/`yq`.

jandubois: > `FillDefault` needs to be updated for overriding

I guess. And then it also should be supported by template embedding, for consistency.

At first I thought since Lima doesn't use the value of `Meta`, just declaring the field to make sure we don't get strict YAML errors would be enough. But I guess I was wrong. 😄

Issue #4085, opened 2025-09-24
guestagent: iptable watcher should not be needed at all


AkihiroSuda:
https://github.com/lima-vm/lima/blob/9d815d5027037227ba2b4cddc9d19a583ec48ccc/pkg/guestagent/iptables/iptables.go#L24-L38

This complex and non-robust iptables watcher exists solely for `sudo nerdctl`.

Rootful Docker, even with `userland-proxy: false`, does not need watching iptables, as the port number appears in `/proc/net/tcp`:

```console
# cat /etc/docker/daemon.json 
{
  "features": {
    "cdi": true,
    "containerd-snapshotter": true
  },
  "userland-proxy": false
}
# docker run -d -p 8080:80 httpd:alpine
1042501e95d4e068710a4be09500ec770de7eddb877005989524e3f2175524fc

#  lsof -P -a -i4 -i6 -itcp +c0
COMMAND          PID            USER   FD   TYPE DEVICE SIZE/OFF NODE NAME
systemd            1            root  213u  IPv6   5970      0t0  TCP *:22 (LISTEN)
systemd-resolve  523 systemd-resolve   15u  IPv4   7192      0t0  TCP _localdnsstub:53 (LISTEN)
systemd-resolve  523 systemd-resolve   17u  IPv4   7194      0t0  TCP _localdnsproxy:53 (LISTEN)
dockerd          829            root   28u  IPv4 453813      0t0  TCP *:8080 (LISTEN)
sshd            1362            root    3u  IPv6   5970      0t0  TCP *:22 (LISTEN)
sshd            1512            root    4u  IPv6   8165      0t0  TCP 192.168.5.15:22->host.lima.internal:29186 (ESTABLISHED)
sshd            1609            suda    4u  IPv6   8165      0t0  TCP 192.168.5.15:22->host.lima.internal:29186 (ESTABLISHED)
```

Same for Rootful Podman. `conmon` seems to take the port to let it appear in `/proc/net/tcp`.

Probably we can completely remove the iptables watcher after making some modifications to nerdctl

We may even remove the Kubernetes service watcher, if we can make the same modification to Kubernetes.

AkihiroSuda:
nerdctl needs an equivalent of:
- https://github.com/moby/moby/pull/48132
- https://github.com/containers/podman/pull/23446

PR #4106, opened 2025-09-29
guestagent: ebpfticker: impose a rate limit

### LINKED ISSUE #4094: [v2.0.0-alpha.1] guestagent: the CPU utilization rate of `lima-guestagent` is always close to 90%.
After #4066 is merged, the CPU utilization rate of `lima-guestagent` is always close to 90%.

screenshot 2025-09-25 9 36 07

I confirmed that it will be resolved with reverted.

_Originally posted by @norio-nomura in https://github.com/lima-vm/lima/issues/4066#issuecomment-3331246222_

AkihiroSuda:
Fix #4094

The `iptables` command invoked by the guestagent seems to cause an extreme burst of `bind` events: https://github.com/lima-vm/lima/issues/4094#issuecomment-3346519158

This commit introduces a rate limit so as to avoid consuming too much CPU when a burst of events happen.

The `iptables` watcher is going to be soon removed (#4085), but this rate limit seems still worth keeping after that.

### DIFF
--- a/cmd/lima-guestagent/daemon_linux.go
+++ b/cmd/lima-guestagent/daemon_linux.go
@@ -12,6 +12,7 @@ import (
 	"github.com/mdlayher/vsock"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"golang.org/x/time/rate"
 
 	"github.com/lima-vm/lima/v2/pkg/guestagent"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/api/server"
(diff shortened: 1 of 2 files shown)


jandubois:
> his rate limit seems still worth keeping after that.

Why? Doesn't it just add a race condition, so a port bind may be missed until the next time polling activates. An occasional burst doesn't seem harmful as long as it doesn't constantly use CPU.

I would argue that if we still need the rate limiter then the current eBPF watcher approach is not the right solution.

AkihiroSuda:
> The problem is that the guestagent seems now executing `/usr/sbin/iptables -t nat -S` infinite times without delay 

 _Originally posted by @AkihiroSuda in [#4094](https://github.com/lima-vm/lima/issues/4094#issuecomment-3349786478)_

PR #4107, opened 2025-09-29
guestagent: remove iptables watcher for rootful nerdctl (v2.1.6-beta.0)

### LINKED ISSUE #4085: guestagent: iptable watcher should not be needed at all

AkihiroSuda:
This commit significantly simplifies the guestagent, by removing the complex and non-robust iptables watcher that had existed solely for `sudo nerdctl`.

This iptables watcher is no longer needed since nerdctl v2.1.6, as it exposes proper `/proc/net` entries even for rootful containers.

See:
- containerd/nerdctl#4526.

Fix #4085


pkg/limayaml/containerd.yaml
```diff
@@ -1,9 +1,10 @@
 archives:
-- location: https://github.com/containerd/nerdctl/releases/download/v2.1.5/nerdctl-full-2.1.5-linux-amd64.tar.gz
+# nerdctl version must be >= 2.1.6 since Lima v2.0.0
```
AkihiroSuda: (This may affect alpine-lima, colima, etc.)
@jandubois @abiosoft

jandubois: When is 2.1.6 going to be released? It looks like 2.1.5 just came out.

AkihiroSuda: Probably this week.

AkihiroSuda: If we want to keep supporting `apk add nerdctl` we have to defer merging this PR for half a year

jandubois: I don't know if that is necessary. At least colima and Rancher Desktop should not care; I wonder how many users are running Alpine images and are installing `nerdctl` via `apk`.

You could release-note (and add a comment to the templates) that `nerdctl` for Alpine needs to be installed from GitHub releases. We could justify it because it is a major version bump for 2.0. 

I wouldn't want to do this if it was the 1.3 release, but for 2.0 it seems acceptable.

jandubois:
```console
❯ cat alpine-nerdctl.yaml
base: template://alpine

provision:
- mode: system
  script: |
    [ -f /usr/local/bin/nerdctl ] && exit

    cd /tmp
    ARCH="$(uname -m)"
    ARCH="${ARCH/aarch64/arm64}"
    ARCH="${ARCH/x86_64/amd64}"
    TARBALL="nerdctl-${PARAM_NERDCTL}-linux-${ARCH}.tar.gz"
    wget "https://github.com/containerd/nerdctl/releases/download/v${PARAM_NERDCTL}/$TARBALL"
    tar xz -C /usr/local/bin -f "$TARBALL" nerdctl
    rm "$TARBALL"

param:
  NERDCTL: 2.1.6-beta.0

❯ l start -y --name alpine ./alpine-nerdctl.yaml
...

❯ l shell alpine nerdctl version
WARN[0000] failed to inspect the rootless containerd socket address  error="environment variable XDG_RUNTIME_DIR is not set, see https://rootlesscontaine.rs/getting-started/common/login/"
WARN[0000] unable to determine buildctl version          error="exec: \"buildctl\": executable file not found in $PATH"
Client:
 Version:	v2.1.6-beta.0
 OS/Arch:	linux/arm64
 Git commit:	ce40996a2d5fbf26d0ffa424b0724429aace5b89
 buildctl:
  Version:
```

jandubois: We could add an `alpine-rootful` template if we add this to the template above:

```yaml
- mode: system
  script: |
    [ -x /usr/bin/containerd ] && exit

    apk add containerd
    rc-update add containerd
    rc-service containerd start
```

Then we also have an example to point to in the release notes:

```console
❯ l shell alpine sudo nerdctl pull busybox
docker.io/library/busybox:latest:                                                 resolved       |++++++++++++++++++++++++++++++++++++++|
index-sha256:d82f458899c9696cb26a7c02d5568f81c8c8223f8661bb2a7988b269c8b9051e:    done           |++++++++++++++++++++++++++++++++++++++|
manifest-sha256:cddc8af5547af9de5e6fb66b36d66ef7418561204e1255ae528d0b2c919d09a3: done           |++++++++++++++++++++++++++++++++++++++|
config-sha256:e8291c1a323abf610ebeb32699b8df3e8046823b5dfbf795d888c9c6a73aeff8:   done           |++++++++++++++++++++++++++++++++++++++|
layer-sha256:499bcf3c8ead8ef2570f7ee8e4ab5a4489647d67dd1c1b425fdbd4a74f4b4386:    done           |++++++++++++++++++++++++++++++++++++++|
elapsed: 2.3 s                                                                    total:  10.4 K (4.5 KiB/s)
```

I'm undecided if I like the idea or not. Let me know if you want me to create a PR!

AkihiroSuda: `alpine.yaml` had no support for nerdctl integration so far, so adding a comment like `'apk add nerdctl' will not work until the release of v2.1.6. Install from https://github.com/containerd/nerdctl/releases/ instead` should be fine?

jandubois: Good enough for me; I just don't want to delay this change for another 6 months. I would like us to fix the port monitoring issue for Lima 2.0.

AkihiroSuda: Added the comments in `alpine.yaml`

pkg/guestagent/guestagent_linux.go
```diff
@@ -220,7 +117,7 @@ func (a *agent) Events(ctx context.Context, ch chan *api.Event) {
 	}
 }
 
-func (a *agent) LocalPorts(ctx context.Context) ([]*api.IPPort, error) {
+func (a *agent) LocalPorts(_ context.Context) ([]*api.IPPort, error) {
```
jandubois, line 120: The `_` is only needed when you mix named and unnamed parameters.

```suggestion
func (a *agent) LocalPorts(context.Context) ([]*api.IPPort, error) {
```

AkihiroSuda: Linter doesn't allow that

jandubois: Are you sure?

```console
❯ GOOS=linux golangci-lint run pkg/guestagent/...
0 issues.
```

AkihiroSuda: IIRC CI failed without that, but I could be wrong

Issue #3995, opened 2025-09-07
The `--yes` option is inconsistent and confusing


jandubois:
This doesn't make any sense:

```console
❯ l clone default clone
? Do you want to start the instance now?  (Y/n)
FATA[0004] interrupt

❯ l delete clone
...

❯ l --yes clone default clone
=> does not start the new instance
```

Whereas `limactl --yes create foo` will start the new instance right away.

Same thing for `limactl edit`:

```console
❯ l edit clone --cpus 3
INFO[0000] Instance "clone" configuration edited
? Do you want to start the instance now?  (Y/n)
FATA[0002] interrupt

❯ l --yes edit clone --cpus 4
INFO[0000] Instance "clone" configuration edited
=> does not start the updated instance
```

Linguistically `yes` is not a synonym for "no tty". It is an affirmative. So why doesn't it start the instance?

There are 2 reasonable definitions for `--yes`:

1. It implies the user presses `ENTER` to all normally interactive prompts, selecting the default choice.
2. It is only accepted by commands that can start an instance, and it means "yes, start it"

The fact that `--tty=false` doesn't mean you pick the default choice for prompts makes this even more confusing.

The current definition makes no sense (except on the `limactl start` command).

This cannot be fixed in a backwards compatible way,  so I would like to figure out how (or if) we want to change it for 2.0, when we have a chance to introduce breaking changes.

jandubois:
AFAICT `start`, `clone`, and `edit` are the only subcommands that will issue a prompt, so I think picking option (2) makes sense. It would turn `--yes` from a global flag to one for those 3 subcommands.

I think it would be even clearer to leave `--yes` just for the `start` command and use `--start=false` for `clone` and `edit`.

These options are really only for scripting, to make it possible for them to not prompt. Explicitly asking not to start the instance is clearer than saying you run without a tty, and relying on the implementation that this means not to start the instance.

```bash
limactl edit foo --cpus 3 --no-start
limactl edit foo --cpus 3 --start=false
```

is much clearer than

```bash
limactl edit foo --cpus 3 --yes
limactl edit foo --cpus 3 --tty=false
```

I would leave `--yes` as a hidden flag for backwards compatibility until Lima 3.0.

I kind of like the `-y` flag for `limactl start`, but even there `--yes` is not really intuitive. Something like `--no-prompt` or `--prompt=false` is much more self-explanatory.

[...]

PR #4108, opened 2025-09-29
Use --start flag to start VM for 'clone' and 'edit' commands

Horiodino:
This PR addresses [Issue #3995] regarding the confusing and inconsistent behavior of the --yes flag in limactl clone and limactl edit. The following changes have been made:

Changes

1. Deprecate `--yes` flag for the clone and edit commands. A warning is shown when used, recommending `--start` instead.
2. Add `--start` flag (default: true) to both commands, allowing users to explicitly control whether the instance is started after cloning or editing.

### DIFF
--- a/cmd/limactl/edit.go
+++ b/cmd/limactl/edit.go
@@ -36,6 +36,7 @@ func newEditCommand() *cobra.Command {
 		ValidArgsFunction: editBashComplete,
 		GroupID:           basicCommand,
 	}
+	editCommand.Flags().Bool("start", false, "Start the instance after editing")
 	editflags.RegisterEdit(editCommand, "")
 	return editCommand
 }
@@ -140,19 +141,23 @@ func editAction(cmd *cobra.Command, args []string) error {
 		logrus.Infof("Instance %q configuration edited", inst.Name)
 	}
 
-	if !tty {
-		// use "start" to start it
-		return nil
-	}
 	if inst == nil {
 		// edited a limayaml file directly
 		return nil
 	}
-	startNow, err := askWhetherToStart()
+
+	start, err := flags.GetBool("start")
 	if err != nil {
 		return err
 	}
-	if !startNow {
+
+	if tty && !flags.Changed("start") {
+		start, err = askWhetherToStart()
+		if err != nil {
+			return err
+		}
+	}
+	if !start {
 		return nil
 	}
 	err = reconcile.Reconcile(ctx, inst.Name)
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -155,6 +155,11 @@ func newApp() *cobra.Command {
 		}
 
 		if cmd.Flags().Changed("yes") {
+			switch cmd.Name() {
+			case "clone", "edit", "rename":
+				logrus.Warn("--yes flag is deprecated (--tty=false is still supported and works in the same way. Also consider using --start)")
+			}
+
 			// Sets the value of the yesValue flag by using the yes flag.
 			yesValue, _ := cmd.Flags().GetBool("yes")
 			if yesValue {
--- a/cmd/limactl/clone.go
+++ b/cmd/limactl/clone.go
@@ -34,6 +34,7 @@ Not to be confused with 'limactl copy' ('limactl cp').
 		ValidArgsFunction: cloneBashComplete,
 		GroupID:           advancedCommand,
 	}
+	cloneCommand.Flags().Bool("start", false, "Start the instance after cloning")
 	editflags.RegisterEdit(cloneCommand, "[limactl edit] ")
 	return cloneCommand
 }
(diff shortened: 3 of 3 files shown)


cmd/limactl/main.go
```diff
@@ -143,6 +143,10 @@ func newApp() *cobra.Command {
 		}
 
 		if cmd.Flags().Changed("yes") {
+			if cmd.Name() == "clone" || cmd.Name() == "edit" {
+				logrus.Warn("--yes flag is deprecated for the '" + cmd.Name() + "' command, use --start instead")
```
AkihiroSuda: `--tty=false` should be specified too ?

cmd/limactl/main.go
```diff
@@ -143,6 +143,10 @@ func newApp() *cobra.Command {
 		}
 
 		if cmd.Flags().Changed("yes") {
+			if cmd.Name() == "clone" || cmd.Name() == "edit" {
```
AkihiroSuda: nit: use `switch{}`

cmd/limactl/edit.go
```diff
@@ -36,6 +36,7 @@ func newEditCommand() *cobra.Command {
 		ValidArgsFunction: editBashComplete,
 		GroupID:           basicCommand,
 	}
+	editCommand.Flags().Bool("start", true, "Start the instance after editing")
```
AkihiroSuda: Shouldn't this remain false ?
Same for clone

cmd/limactl/main.go
Horiodino: i wont make any changes but its good for code redability :+1:

cmd/limactl/main.go
jandubois, line 1: I still think we should remove `--yes` from being a global option that shows up in `limactl help`. `start`, `edit`, and `clone` are the only commands that check the value of the `--tty` setting, so it really only applies to `start` now.

So I would make it a local option in those 3 commands, and make it hidden in `edit` and `clone`, so it really becomes just a `start` option.

@AkihiroSuda You seemed to have reservations about making the deprecated option hidden. What is the reason for that?

cmd/limactl/main.go
```diff
@@ -143,6 +143,11 @@ func newApp() *cobra.Command {
 		}
 
 		if cmd.Flags().Changed("yes") {
+			switch cmd.Name() {
+			case "clone", "edit":
+				logrus.Warn("--yes flag is deprecated for the '" + cmd.Name() + "' command, use --start instead and --tty=false")
```
jandubois: I don't really understand the warning. `--start` has different semantics than `--yes`, so `--start=false` may be the same as `--yes=true`, but `--start=true` (start without asking) is different from `--yes=false` (ask about starting).

I think the `edit` and `clone` commands should just print the simplified message:

```suggestion
				logrus.Warn("The --yes flag is deprecated")
```

Users can then run `limactl edit --help` to learn more.

jandubois (changes requested):
Please wait for feedback from @AkihiroSuda before making any changes.

cmd/limactl/main.go
AkihiroSuda: > You seemed to have reservations about making the deprecated option hidden. What is the reason for that?

Because we have been keeping other deprecated commands/options.
(`limactl show-ssh`)

cmd/limactl/main.go
AkihiroSuda: What about:
```suggestion
				logrus.Warn("--yes flag is deprecated (--tty=false is still supported and works in the same way. Also consider using --start)")
```

cmd/limactl/main.go
jandubois: Yes, but I don't understand why this makes sense: the point of deprecation is to transition away from something because it will go away. Keeping it available for new users to discover and use is not helpful.

Only users that already know about these commands (or existing scripts) should continue to use them, but get a warning, so they can transition away before the commands get removed. If there is no intention of removing a command or feature, then there is no point in deprecating it.

jandubois: > So I would make it a local option in those 3 commands, and make it hidden in `edit` and `clone`, so it really becomes just a `start` option.

We definitely need to make it a local option if we don't want to hide the deprecated option. That way we can mark it in the `--help` output for `edit` and `clone` as deprecated (which it isn't right now):

```console
❯ l edit --help
Edit an instance of Lima or a template
...
Global Flags:
...
  -y, --yes                 Alias of --tty=false
```

But also so that `--yes` doesn't show up for every single subcommand when it really is just an option for `start` anymore.

One big reason for getting rid of `--yes` is to avoid user confusion, and we are not achieving this by keeping it very visible.

cmd/limactl/main.go
```diff
@@ -143,6 +143,11 @@ func newApp() *cobra.Command {
 		}
 
 		if cmd.Flags().Changed("yes") {
+			switch cmd.Name() {
+			case "clone", "edit":
```
AkihiroSuda: Now "rename" has to be covered too
- #4207

jandubois (changes requested):
Please resolve merge conflicts and add support for the `limactl rename` command as well.

AkihiroSuda:
@Horiodino Could you update the PR?

Horiodino:
Hey I'm out of town I will do ASAP as I return.

PR #4109, opened 2025-09-29
CI: test UDP port forwarding

### LINKED ISSUE #3657: UDP port forwarding does not work with iperf3 and lacks tests

AkihiroSuda:
Partially fix #3657

### DIFF
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -375,9 +375,10 @@ if [[ -n ${CHECKS["ssh-over-vsock"]} ]]; then
 	fi
 fi
 
-# Use GHCR to avoid hitting Docker Hub rate limit
+# Use GHCR and ECR to avoid hitting Docker Hub rate limit
 nginx_image="ghcr.io/stargz-containers/nginx:1.19-alpine-org"
 alpine_image="ghcr.io/containerd/alpine:3.14.0"
+coredns_image="public.ecr.aws/eks-distro/coredns/coredns:v1.12.2-eks-1-31-latest"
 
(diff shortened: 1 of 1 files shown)


hack/test-templates.sh
```diff
@@ -375,9 +375,10 @@ if [[ -n ${CHECKS["ssh-over-vsock"]} ]]; then
 	fi
 fi
 
-# Use GHCR to avoid hitting Docker Hub rate limit
+# Use GHCR and ECR to avoid hitting Docker Hub rate limit
```
jandubois, line 378: I thought Docker and GitHub had an agreement that GitHub actions running on the GitHub runners on public repos are not subject to the pull request limits. Is this no longer the case?

I suspect the reason is that we are using a non-free runner, so the exemption does not apply. Which I guess is fair enough: if GitHub is being paid for the runner, then it can require a paid DockerHub account as well.

AkihiroSuda: Let's try migrating to Docker Hub in a separate PR

jandubois: It doesn't really matter which registry we use, so I don't think we need to change, except maybe to expand the comment.

I was just surprised to see that we had changed the location because of rate limits, but then I remembered that we were using non-free runners before, which might explain why this was necessary.

AkihiroSuda: Anyway the test has to be runnable on laptops too, not just on CI

jandubois: Sure, but I assume anyone running into pull request limits will just log in with their free DockerHub account.

Anyways, as I wrote above, we don't need to switch the images back; I was just curious.

PR #4117, opened 2025-10-01
guestagent: reuse netlink connections

AkihiroSuda:
Prior to this commit, `pkg/guestagent/sockets` opened and closed a netlink connection on each invocation of `sockets.List()`.

### DIFF
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -20,8 +20,13 @@ import (
 )
 
 func New(ctx context.Context, ticker ticker.Ticker) (Agent, error) {
+	socketsLister, err := sockets.NewLister()
+	if err != nil {
+		return nil, err
+	}
 	a := &agent{
 		ticker:                   ticker,
+		socketLister:             socketsLister,
 		kubernetesServiceWatcher: kubernetesservice.NewServiceWatcher(),
 	}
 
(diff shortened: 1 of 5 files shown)


AkihiroSuda:
@lima-vm/committers Let's merge this to unblock the alpha2 release 🙏

pkg/guestagent/guestagent_linux.go
```diff
@@ -227,3 +232,12 @@ func (a *agent) HandleInotify(event *api.Inotify) {
 		}
 	}
 }
+
+func (a *agent) Close() error {
+	if a.socketLister != nil {
+		if err := a.socketLister.Close(); err != nil {
+			return err
+		}
```
jandubois, line 240:
```suggestion
		return a.socketLister.Close()
```

jandubois (approved):
Thanks, LGTM

My only concern is if this new approach is less robust. Previously a transient failure only affected a single request. Now it can be permanent if clearing the failure state requires closing and reopening the socket.

However, I have no idea how likely a failure is at all, so maybe this is misplaced. The improved efficiency is obviously desirable.

jandubois:
> Let's merge this to unblock the alpha2 release

I'm not sure why this was blocking the release; it seems purely like an optimization that should not be user visible. But I guess getting some more testing done before the release will be worthwhile.

PR #4118, opened 2025-10-01
CI: test Kubernetes port forwarder

AkihiroSuda:


AkihiroSuda:
```
[preflight] Some fatal errors occurred:
  	[ERROR NumCPU]: the number of available CPUs 1 is less than the required 2
  	[ERROR Mem]: the system RAM (955 MB) is less than the minimum 1700 MB
```
https://github.com/lima-vm/lima/actions/runs/18154934137/job/51672563332?pr=4118

AkihiroSuda:
will update to address https://github.com/lima-vm/lima/pull/4120#discussion_r2404102573

afbjorklund:
>  	[ERROR NumCPU]: the number of available CPUs 1 is less than the required 2
>  	[ERROR Mem]: the system RAM (955 MB) is less than the minimum 1700 MB

I actually hadn't looked at this before, but I know that kind is hacking the preflight.

--skip-phases=preflight

Apparently always having 2 vCPU and 2 GiB total RAM is not so important, after all...

But better to do a more selective ignore.

AkihiroSuda:
The default CPU and memory seems enough to pass the CI

.github/workflows/test.yml
```diff
@@ -370,6 +370,19 @@ jobs:
...
+      with:
+        template: templates/k8s.yaml
+    - name: "Run BATS k8s tests"
+      # Wish we could use BATS_TEST_RETRIES=3 as an environment variable here,
+      # but bats does not seem to support it.
+      uses: nick-fields/retry@ce71cc2ab81d554ebbe88c79ab5975992d79ba08  # v3.0.2
+      with:
+        timeout_minutes: 30
+        retry_on: error
+        max_attempts: 3
```
jandubois, line 384: You have to set the variable inside the test that is flaky; the harness sets it to 0, so you can't set it from the outside:

```console
❯ cat random.bats
@test "flaky test that fails 70% of the time" {
    BATS_TEST_RETRIES=10
    echo "Attempt $BATS_TEST_TRY_NUMBER/$BATS_TEST_RETRIES" >&3
    (( RANDOM % 10 < 3))
}

❯ ./hack/bats/lib/bats-core/bin/bats random.bats
random.bats
   flaky test that fails 70% of the time                                                                                                                             1/1
Attempt 1/10
 ✓ flaky test that fails 70% of the time
Attempt 2/10

1 test, 0 failures

❯ ./hack/bats/lib/bats-core/bin/bats random.bats
random.bats
   flaky test that fails 70% of the time                                                                                                                             1/1
Attempt 1/10
   flaky test that fails 70% of the time                                                                                                                             1/1
Attempt 2/10
   flaky test that fails 70% of the time                                                                                                                             1/1
Attempt 3/10
   flaky test that fails 70% of the time                                                                                                                             1/1
Attempt 4/10
   flaky test that fails 70% of the time                                                                                                                             1/1
Attempt 5/10
 ✓ flaky test that fails 70% of the time
Attempt 6/10

1 test, 0 failures
```

Without the logging you cannot tell if the test needed retries to pass (and `BATS_TEST_RETRY_NUMBER` is not defined in `setup` or `teardown`).

jandubois: You can set it at the file level, so it applies to every test in the same file:

```bash
BATS_TEST_RETRIES=10

@test "flaky test that fails 70% of the time" {
    echo "Attempt $BATS_TEST_TRY_NUMBER/$BATS_TEST_RETRIES" >&3
    (( RANDOM % 10 < 3))
}
```

AkihiroSuda: I feel GHA YAML is the right place to set `BATS_TEST_RETRIES`. Maybe we should submit a proposal to the upstream.

Can we just merge this PR?

hack/bats/extras/k8s.bats
```diff
@@ -0,0 +1,64 @@
...
+# Instance names are "${NAME}-0", "${NAME}-1", ...
+NAME="k8s"
+
+get_num_nodes() {
+    local nodes=0
+    for tag in "${BATS_TEST_TAGS[@]}"; do
+        if [[ $tag =~ ^nodes:([0-9]+)$ ]]; then
+            nodes="${BASH_REMATCH[1]}"
+        fi
+    done
```
jandubois, line 19: I'm not convinced that this is a good way to control setup for individual tests.

I think it would be clearer not to rely on `local_setup`, but to have each test call a setup function with an argument, like

```bash
@test 'Single-node configuration' {
   setup_nodes 1
   # ...
}
@test 'Multi-node configuration' {
   setup_nodes "${LIMA_BATS_K8S_NODES:-3}"
   # ...
}
```

And `local_teardown` could use `limactl ls --quiet | grep "^${NAME}-"` to get the list of instances to delete.

This also allows you to control the number of nodes from the outside without modifying the test source code to adjust the tag.

(Ab)using the tags to pass arguments to the setup/teardown functions is a bit confusing, and should be simpler. But we can always change this later, e.g. when there actually is a multi-node test.

hack/bats/extras/k8s.bats
```diff
@@ -0,0 +1,64 @@
...
+}
+
+local_teardown() {
+    local nodes=$(get_num_nodes)
+    for ((i=0; i<nodes; i++)); do
+        limactl delete --force "${NAME}-$i" || :
+    done
+}
+
+k() {
```
jandubois, line 46: I would call it `kubectl` for readability. Local functions always take precedence over running external programs.

Issue #3237, opened 2025-02-14
guestagent binaries are too huge


AkihiroSuda:
The footprint of the guestagent binaries has grown from 6.2M + 6.6M to 46M + 45M + 46M + 48M through the four years 

```console
$ ls -lh lima-0.1.0-Darwin-x86_64.tar.gz 
-rw-r--r--  1 suda  staff   8.4M  2 14 14:23 lima-0.1.0-Darwin-x86_64.tar.gz

$ tar xf lima-0.1.0-Darwin-x86_64.tar.gz

$ ls -lh share/lima/lima-guestagent.Linux-*
-rw-r--r--  1 suda  staff   6.2M  5 14  2021 share/lima/lima-guestagent.Linux-aarch64
-rw-r--r--  1 suda  staff   6.6M  5 14  2021 share/lima/lima-guestagent.Linux-x86_64
```

```console
$ ls -lh lima-1.0.6-Darwin-x86_64.tar.gz 
-rw-r--r--  1 suda  staff    61M  2 14 14:23 lima-1.0.6-Darwin-x86_64.tar.gz

$ tar xf lima-1.0.6-Darwin-x86_64.tar.gz

$ ls -lh share/lima/lima-guestagent.Linux-*
-rw-r--r--  1 suda  staff    46M  2 13 09:04 share/lima/lima-guestagent.Linux-aarch64
-rw-r--r--  1 suda  staff    45M  2 13 09:04 share/lima/lima-guestagent.Linux-armv7l
-rw-r--r--  1 suda  staff    46M  2 13 09:04 share/lima/lima-guestagent.Linux-riscv64
-rw-r--r--  1 suda  staff    48M  2 13 09:04 share/lima/lima-guestagent.Linux-x86_64
```

jandubois:
Most of this comes from adding the Kubernetes port forwarding in `pkg/guestagent/kubernetesservice` which pulled in the `k8s.io/*` dependencies:

```console
❯ tar tvfz lima-0.14.2-Darwin-arm64.tar.gz| grep guest
-rw-r--r--  0 runner staff 7139328 23 Dec  2022 ./share/lima/lima-guestagent.Linux-x86_64
-rw-r--r--  0 runner staff 6946816 23 Dec  2022 ./share/lima/lima-guestagent.Linux-riscv64
-rw-r--r--  0 runner staff 6881280 23 Dec  2022 ./share/lima/lima-guestagent.Linux-aarch64

[...]

PR #4120, opened 2025-10-01
guestagent: k8s: use kubectl instead of client-go

AkihiroSuda:
Part of:
- #3237

TODO: drop dependency on k8s.io/api

-  - -

Merge after:
- https://github.com/lima-vm/lima/pull/4118


AkihiroSuda:
Before (964fb3059d12294dd3d4ee62ed0088bb6ed023b6)
```console
$ du -hs _output/
128M    _output/

$ ls -lh _output/bin/limactl _output/share/lima/lima-guestagent.Linux-*
-rwxr-xr-x@ 1 suda  staff    28M Oct  1 19:47 _output/bin/limactl*
-rw-r--r--@ 1 suda  staff    14M Oct  1 19:47 _output/share/lima/lima-guestagent.Linux-aarch64.gz
-rw-r--r--@ 1 suda  staff    15M Oct  1 19:47 _output/share/lima/lima-guestagent.Linux-armv7l.gz
-rw-r--r--@ 1 suda  staff    14M Oct  1 19:47 _output/share/lima/lima-guestagent.Linux-ppc64le.gz
-rw-r--r--@ 1 suda  staff    15M Oct  1 19:47 _output/share/lima/lima-guestagent.Linux-riscv64.gz
-rw-r--r--@ 1 suda  staff    16M Oct  1 19:47 _output/share/lima/lima-guestagent.Linux-s390x.gz
-rw-r--r--@ 1 suda  staff    16M Oct  1 19:47 _output/share/lima/lima-guestagent.Linux-x86_64.gz
```

After (f74fbdf941c7aa742c68701444edf925c1c1afd4)
```console
$ du -hs _output/
 88M    _output/

$ ls -lh _output/bin/limactl _output/share/lima/lima-guestagent.Linux-*
-rwxr-xr-x@ 1 suda  staff    28M Oct  1 19:49 _output/bin/limactl*
-rw-r--r--@ 1 suda  staff   8.4M Oct  1 19:49 _output/share/lima/lima-guestagent.Linux-aarch64.gz
-rw-r--r--@ 1 suda  staff   8.8M Oct  1 19:49 _output/share/lima/lima-guestagent.Linux-armv7l.gz
-rw-r--r--@ 1 suda  staff   8.4M Oct  1 19:49 _output/share/lima/lima-guestagent.Linux-ppc64le.gz
-rw-r--r--@ 1 suda  staff   8.8M Oct  1 19:49 _output/share/lima/lima-guestagent.Linux-riscv64.gz
-rw-r--r--@ 1 suda  staff   9.2M Oct  1 19:49 _output/share/lima/lima-guestagent.Linux-s390x.gz
-rw-r--r--@ 1 suda  staff   9.3M Oct  1 19:49 _output/share/lima/lima-guestagent.Linux-x86_64.gz
```

jandubois:
I have not looked at this PR at all yet, but wanted to mention a couple of things I discussed with @Nino-K as requirements for his port monitoring PR:

* Keep retrying to connect to k8s indefinitely, as it will not be running yet by the time the guest agent starts.
* When the connection breaks, keep trying to reconnect with a short delay indefinitely, as the user may have stopped and restarted k8s.

For this PR also: the `kubectl` binary may not yet be available on the PATH when you try to invoke it, keep trying. It may also fail because the port may be open, but the apiserver not yet responding, or the kubeconfig may be missing etc. The retry on broken connections should handle this automatically.

Because of the indefinite retries, the kubernetes watcher should be opt-in (configurable in `lima.yaml`), so it only runs when the VM is known to run Kubernetes.

AkihiroSuda:
> The retry on broken connections should handle this automatically.

Yes, this is retried.

> Because of the indefinite retries, the kubernetes watcher should be opt-in

It wasn't opt-in so far, and I don't think it has to be so, as the overhead of polling `LookPath("kubectl")` seems trivial.

hack/test-templates.sh
```diff
@@ -431,6 +439,18 @@ if [[ -n ${CHECKS["container-engine"]} ]]; then
 	fi
 fi
 
+if [[ -n ${CHECKS["k8s"]} ]]; then
+	INFO "Testing Kubernetes"
+	set -x
+	limactl shell "$NAME" kubectl get nodes -o wide
+	limactl shell "$NAME" kubectl create deployment nginx --image="${nginx_image}"
+	limactl shell "$NAME" kubectl create service nodeport nginx --node-port=31080 --tcp=80:80
```
nirs: Why not use yaml file with the deployment and service?

hack/test-templates.sh
```diff
@@ -431,6 +439,18 @@ if [[ -n ${CHECKS["container-engine"]} ]]; then
...
+	timeout 3m bash -euxc "until curl -f --retry 30 --retry-connrefused http://127.0.0.1:31080; do sleep 3; done"
```
nirs: Checking the connection makes sense only after the deployment is available. I would do this:

```
kubectl apply -f nginx.yaml
kubectl rollout status deployment nginx --timeout 60s
```

At this point the service is typically not available, but it should be in few seconds, so we can start checking every second.

Since we wait separately for the deployment, we don't need to wait 3 minutes for the connection, maybe 30 seconds ie enough to detect broken port forwarding.

Notes for the curl command:
- Using --fail will be more readable
- Adding --silent will avoid unhelpful noise in the test logs

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -38,96 +38,161 @@ type Entry struct {
...
+	}
+	return errors.Join(readErr, waitErr)
+}
+
+// readKubectlStream reads kubectl JSON watch events from r and updates the internal
+// services map. The stream is newline-delimited JSON objects representing "WatchEvent".
+func (s *ServiceWatcher) readKubectlStream(r io.Reader) error {
+	scanner := bufio.NewScanner(r)
+	// increase buffer in case of large JSON objects
+	const maxBuf = 10 * 1024 * 1024
```
nirs: We don't have units constants in lima (e.g. KiB, MiB, GiB)? It will make the code more readable.

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -38,96 +38,161 @@ type Entry struct {
...
+	buf := make([]byte, 0, 64*1024)
+	scanner.Buffer(buf, maxBuf)
+
+	for scanner.Scan() {
+		line := scanner.Bytes()
+		line = bytes.TrimSpace(line)
+		if len(line) == 0 {
+			continue
```
nirs, line 162: We expect empty lines in json stream?

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -38,96 +38,161 @@ type Entry struct {
...
-			return false, nil
-		}
 
-		informerFactory := informers.NewSharedInformerFactory(kubeClient, time.Hour)
-		serviceInformer := informerFactory.Core().V1().Services().Informer()
-		informerFactory.Start(ctx.Done())
-		cache.WaitForCacheSync(ctx.Done(), serviceInformer.HasSynced)
+	go func() {
+		for i := 0; ; i++ {
+			if i > 0 {
```
nirs: Why do you skip the first iteration?

The code will more clear if we separate the loop from the action preformed for each iteration.

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -38,96 +38,161 @@ type Entry struct {
...
+			if err := canGetServices(ctx, kubectl, kubeconfig); err != nil {
+				logrus.WithError(err).Debugf("kubectl auth can-i ... failed; will retry")
+				continue
+			}
+			cmd := exec.CommandContext(ctx, kubectl,
+				"get", "--all-namespaces", "service", "--watch", "--output-watch-events", "--output", "json")
+
+			if kubeconfig != "" {
+				cmd.Env = append(os.Environ(), "KUBECONFIG="+kubeconfig)
+			}
```
nirs: It is simpler and more efficient to use `--kubeconfig=`

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -38,96 +38,161 @@ type Entry struct {
...
-		if err != nil {
-			logrus.Tracef("failed to get kube client: %v, will retry in %v", err, retryInterval)
-			return false, nil
-		}
 
-		informerFactory := informers.NewSharedInformerFactory(kubeClient, time.Hour)
-		serviceInformer := informerFactory.Core().V1().Services().Informer()
-		informerFactory.Start(ctx.Done())
-		cache.WaitForCacheSync(ctx.Done(), serviceInformer.HasSynced)
+	go func() {
```
nirs: Separating the function performing the loop will make the code simpler and easier to follow:

```go
func (s *ServiceWatcher) Start(ctx context.Context) {
    ...
    go s.loop(ctx)
}

func (s *ServiceWatcher) loop(ctx context.Context) {
    for {
        ...
    }
}
```

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -38,96 +38,161 @@ type Entry struct {
...
+		if err := json.Unmarshal(ev.Object, &svc); err != nil {
+			return fmt.Errorf("failed to unmarshal service object: %w (line=%q)", err, line)
 		}
 
-		return kubeClient, nil
+		key := svc.Namespace + "/" + svc.Name
+		s.rwMutex.Lock()
+		switch ev.Type {
+		case watch.Added, watch.Modified:
+			s.services[key] = &svc
```
nirs: Doe we need to keep all services in memory? We can extract the relevant details here and minimize memory usage.

hack/test-templates.sh
AkihiroSuda: Because `kubectl create` is shorter

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: How simpler and more efficient?

hack/test-templates.sh
AkihiroSuda: Addressed in:
- #4118

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: I don't think it is necessary.
Everybody who understands this file knows 1024 = KiB and 1024 * 1024 =MiB

afbjorklund: The code I was just reading had `8 << 20` :-) Know your audience and all that

I still think they could have gone with `128 << 20` instead of `1 << 27`, though.

(sorry, off-topic comment)

I think we are exporting go-units?

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: Fixed

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: Fixed

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: Because the first iteration does not need to wait for retryInterval

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: Theoretically yes?

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: Updated to use go-units

AkihiroSuda:
> Trimming the guest agent is nice, anything using client-go becomes huge quickly. But did you measure memory and cpu usage before and after this change?


Before
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
   1589 root      20   0 1284200  63488  36992 S   0.0   1.6   0:10.21 lima-guestagent
```

After
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                                           
   1192 root      20   0 1254740  41600  19200 S   0.0   1.0   0:01.72 lima-guestagent                                                                                                                   
   2932 root      20   0 1284688  44404  33536 S   0.0   1.1   0:00.02 kubectl
```

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -37,114 +38,185 @@ type Entry struct {
...
+		// The first iteration does not need to wait for retryInterval.
+		if i > 0 {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(retryInterval):
+			}
 		}
+		s.attemptToStartKubectl(ctx)
+	}
```
alexandear, line 68: We can remove `i`:
```suggestion
	s.attemptToStartKubectl(ctx)
	for {
		select {
		case <-ctx.Done():
			return
		case <-time.After(retryInterval):
			s.attemptToStartKubectl(ctx)
		}
	}
```

pkg/guestagent/kubernetesservice/kubernetesservice.go
```diff
@@ -37,114 +38,185 @@ type Entry struct {
...
+func (s *ServiceWatcher) readKubectlStream(r io.Reader) error {
+	scanner := bufio.NewScanner(r)
+	// increase buffer in case of large JSON objects
+	const maxBuf = 10 * units.MiB
+	buf := make([]byte, 0, 64*units.KiB)
+	scanner.Buffer(buf, maxBuf)
+
+	for scanner.Scan() {
+		line := scanner.Bytes()
+		line = bytes.TrimSpace(line)
```
alexandear, line 160: nit:
```suggestion
		line := bytes.TrimSpace(scanner.Bytes())
```

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: Yes, but that will result in duplicating the `s.attemptToStartKubectl(ctx)` line.
So I'm not sure which one is better.

pkg/guestagent/kubernetesservice/kubernetesservice.go
AkihiroSuda: Either seems fine to me

PR #4125, opened 2025-10-02
fix copy command examples format

olamilekan000:
```
change fixes copy command examples formatting in the docs.
```

### DIFF
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -28,25 +28,25 @@ const copyHelp = `Copy files between host and guest
 Prefix guest filenames with the instance name and a colon.
 
 Backends:
-  auto   - Automatically selects the best available backend (rsync preferred, falls back to scp)
-  rsync  - Uses rsync for faster transfers with resume capability (requires rsync on both host and guest)
-  scp    - Uses scp for reliable transfers (always available)
+  - **auto**: Automatically selects the best available backend (rsync preferred, falls back to scp)
+  - **rsync**: Uses rsync for faster transfers with resume capability (requires rsync on both host and guest)
+  - **scp**: Uses scp for reliable transfers (always available)
 
 Examples:
-  # Copy file from guest to host (auto backend)
-  limactl copy default:/etc/os-release .
+  #### Copy file from guest to host (auto backend)
+    limactl copy default:/etc/os-release .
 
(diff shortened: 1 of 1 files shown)


cmd/limactl/copy.go
```diff
@@ -28,25 +28,25 @@ const copyHelp = `Copy files between host and guest
...
 
-  # Copy using scp backend specifically
-  limactl copy --backend=scp default:/var/log/app.log ./logs/
+  #### Copy using scp backend specifically
+    limactl copy --backend=scp default:/var/log/app.log ./logs/
 
-  # Copy multiple files
-  limactl copy file1.txt file2.txt default:/tmp/
+  #### Copy multiple files
+    limactl copy file1.txt file2.txt default:/tmp/
```
AkihiroSuda, line 50: The output of `--help` isn't expected to be markdown.

The rendering issue should be better resolved in https://github.com/lima-vm/lima/blob/master/cmd/limactl/gendoc.go

afbjorklund: We could wrap the whole help text in a code block, but that would not be so easy to read either?

But escaping the newlines and the `#` characters would be nice, so that they are not mis-interpreted...

afbjorklund: Something like this: (line breaks with two spaces before newline `  `, escape the comments with `\#`)

----
Backends:\
  auto   - Automatically selects the best available backend (rsync preferred, falls back to scp)  
  rsync  - Uses rsync for faster transfers with resume capability (requires rsync on both host and guest)  
  scp    - Uses scp for reliable transfers (always available)  

Examples:\
  \# Copy file from guest to host (auto backend)  
  limactl copy default:/etc/os-release .  

----

Instead of this: https://lima-vm.io/docs/reference/limactl_copy/

afbjorklund: * #4126

afbjorklund: After fixing the markdown, I found out that the examples were in the wrong place (or it would have been escaped automatically)... But it is still good to fix the output of other commands, to escape the markdown.

olamilekan000: Just seeing that the intension behind this PR has been adressed. Thanks @afbjorklund 
I'll be closing this.

PR #4126, opened 2025-10-02
Escape text in the command markdown documentation

afbjorklund:
Some characters have a special meaning, like the "#" for headings.
    
By default, lines will be concatenated. Add explicit line breaks.

### DIFF
--- a/cmd/limactl/gendoc.go
+++ b/cmd/limactl/gendoc.go
@@ -101,7 +101,26 @@ and $LIMA_WORKDIR.
 	return doc.GenManTree(cmd.Root(), header, dir)
 }
 
+func escapeMarkdown(text string) string {
+	lines := strings.Split(text, "\n")
+	for i := range lines {
+		// Need to escape backticks first, before adding more
+		for _, c := range strings.Split("\\`*_[]()#+-.|", "") {
+			lines[i] = strings.ReplaceAll(lines[i], c, "\\"+c)
+		}
+		if i < len(lines)-1 {
+			if lines[i] != "" && lines[i+1] != "" {
+				lines[i] += "  " // line break
+			}
+		}
+	}
+	return strings.Join(lines, "\n")
+}
+
 func genDocsy(cmd *cobra.Command, dir string) error {
+	for _, c := range cmd.Root().Commands() {
+		c.Long = escapeMarkdown(c.Long)
+	}
 	return doc.GenMarkdownTreeCustom(cmd.Root(), dir, func(s string) string {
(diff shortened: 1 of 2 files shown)


cmd/limactl/gendoc.go
```diff
@@ -101,7 +101,26 @@ and $LIMA_WORKDIR.
 	return doc.GenManTree(cmd.Root(), header, dir)
 }
 
+func escapeMarkdown(text string) string {
+	lines := []string{}
+	for _, line := range strings.Split(text, "\n") {
+		line = strings.ReplaceAll(line, "#", "\\#")
+		lines = append(lines, line)
+	}
+	for i := range lines {
```
jandubois: I know it totally doesn't matter, but why not have just a single loop:

```suggestion
	lines := strings.Split(text, "\n")
	for i := range lines {
		line[i] = strings.ReplaceAll(line[i], "#", "\\#")
```

cmd/limactl/gendoc.go
```diff
@@ -101,7 +101,26 @@ and $LIMA_WORKDIR.
 	return doc.GenManTree(cmd.Root(), header, dir)
 }
 
+func escapeMarkdown(text string) string {
+	lines := []string{}
+	for _, line := range strings.Split(text, "\n") {
+		line = strings.ReplaceAll(line, "#", "\\#")
```
jandubois: Why only escape `#`? Shouldn't we escape all characters that might be interpreted by Markdown?

```
\ ` * _ [ ] ( ) # + - . |
```

afbjorklund: Escaping periods and dashes seems a bit paranoid, at least for the text that we currently have...

afbjorklund: But kept the escaping of the underscores and the parenthesis, even though they seem not needed?

jandubois: A digit followed by a period starts a numbered list:

1. number followed by period

A dash at the beginning of the line starts a bulleted list:

- dash

jandubois: The output is only used to generate HTML, so does it matter if we are over-escaping?

afbjorklund: Sure, it just felt a bit much with the escape at the end of each sentence\\.

afbjorklund: But I can put them back, if we treat the text as machine-readable only...

afbjorklund:
I did double-check, but maybe not obvious.

jandubois (approved):

AkihiroSuda (approved):
Thanks

PR #4132, opened 2025-10-04
Allow passing guestagent location to instance.StartWithPath

jandubois:
This is similar to passing in the full path to limactl to allow an embedding application stored in a different location to launch a VM.

Simplified the internal calls by removing the limactl argument to instance.Start.

Also changed the order to look for external drivers first in LIMA_DRIVERS_PATH before looking in libexec, allowing the user to override the drivers in the default location. This also makes sure that the LIMA_DRIVERS_PATH drivers are found even when libexec cannot be located.

### DIFF
--- a/pkg/instance/start.go
+++ b/pkg/instance/start.go
@@ -44,9 +44,8 @@ type Prepared struct {
 }
 
 // Prepare ensures the disk, the nerdctl archive, etc.
-func Prepare(ctx context.Context, inst *limatype.Instance) (*Prepared, error) {
-	var guestAgent string
-	if !*inst.Config.Plain {
+func Prepare(ctx context.Context, inst *limatype.Instance, guestAgent string) (*Prepared, error) {
+	if !*inst.Config.Plain && guestAgent == "" {
 		var err error
 		guestAgent, err = usrlocalsharelima.GuestAgentBinary(*inst.Config.OS, *inst.Config.Arch)
 		if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/instance/start.go
```diff
@@ -131,20 +130,24 @@ func Prepare(ctx context.Context, inst *limatype.Instance) (*Prepared, error) {
...
+// The showProgress argument tells the hostagent to show provision script progress by tailing cloud-init logs.
+//
+// The limactl argument allows the caller to specify the full path of the limactl executable.
+// The guestAgent argument allows the caller to specify the full path of the guest agent executable.
+// Inside limactl this function is only called by Start, which passes empty strings for both
+// limactl and guestAgent, in which case the location of the current executable is used for
+// limactl and the guest agent is located from the corresponding <prefix>/share/lima directory.
+//
+// StartWithPaths calls Prepare by itself, so you do not need to call Prepare manually before calling Start.
+func StartWithPaths(ctx context.Context, inst *limatype.Instance, launchHostAgentForeground, showProgress bool, limactl, guestAgent string) error {
```
AkihiroSuda, line 150: The function seems to be called only from `Start()`? Then no need to make this function public?

jandubois: The whole point of having `limactl` and `guestAgent` parameters is for calling it externally (outside of limactl itself); internally it always passes empty strings to use the default mechanism to locate them.

So I could have just added `guestAgent` as an extra parameter to `Start`, but I found it silly to pass the empty strings everywhere, so I instead turned it into a wrapper.

I know this is not a supported API, but I still find it useful that it is possible.

AkihiroSuda: What applications calls `Start` externally?

jandubois: > What applications calls `Start` externally?

I have some; probably nobody else; idk. Currently playing with a Kubernetes controller that will provision Lima instances as an example of orchestration. Not yet open source, but will be, if it works out... I accept full responsibility of keeping it updated with whatever changes we make in Lima, but would like the capabilities to be exposed in some way.

AkihiroSuda (approved):
Merging, but I suggest executing `limactl`

PR #4136, opened 2025-10-04
templates: k8s: support multi-node cluster

### LINKED ISSUE #4100: template://k8s: support multi-node mode (w/ `.param`)
> [...]
> https://github.com/lima-vm/lima/pull/2633
> 
> Slightly off-topic here, but we should be able to extend the upstream `k8s.yaml` to include what was proposed in #2633.
> - Flannel can be unconditionally enabled even for a single-node cluster
> - vmnet (or user-v2, in case of non-macOS) should be specified via `limactl create --network`
> - Port forwards can be disabled via `--set`. We may also consider having `--port-forward-none` as an alias of it
> - The PR does not seem to have `kubeadm join`, but it can be supported with `.param` that distinguish the worker nodes from the control planes 

 _Originally posted by @AkihiroSuda in [#4097](https://github.com/lima-vm/lima/issues/4097#issuecomment-3340013714)_

AkihiroSuda:
```
limactl start --name k8s-0 --network lima:user-v2 template:k8s
limactl shell k8s-0 sudo kubeadm token create --print-join-command
limactl start --name k8s-1 --network lima:user-v2 --set '.param.kubeadmInit="false"' template:k8s
limactl shell k8s-1 sudo <JOIN_COMMAND_FROM_ABOVE>
```

Fix #4100

### DIFF
--- a/templates/k8s.yaml
+++ b/templates/k8s.yaml
@@ -10,6 +10,12 @@
 # NAME       STATUS   ROLES                  AGE   VERSION
 # lima-k8s   Ready    control-plane,master   44s   v1.22.3
 
+# Multi-node clusters can be created by starting multiple instances of this template connected via the `lima:user-v2` network.
+# $ limactl start --name k8s-0 --network lima:user-v2 template:k8s
+# $ limactl shell k8s-0 sudo kubeadm token create --print-join-command
+# $ limactl start --name k8s-1 --network lima:user-v2 --set '.param.kubeadmInit="false"' template:k8s
+# $ limactl shell k8s-1 sudo <JOIN_COMMAND_FROM_ABOVE>
+
 minimumLimaVersion: 2.0.0
 
 base: template://_images/ubuntu-lts
@@ -109,14 +115,17 @@ provision:
     apiVersion: kubelet.config.k8s.io/v1beta1
     cgroupDriver: systemd
     EOF
+    {{if eq .Param.kubeadmInit "true"}}
     kubeadm init --config kubeadm-config.yaml
     # Installing a Pod network add-on
     kubectl apply -f https://github.com/flannel-io/flannel/releases/download/v0.26.7/kube-flannel.yml
     # Control plane node isolation
     kubectl taint nodes --all node-role.kubernetes.io/control-plane-
+    # Copy the kubeconfig file to the default location
+    mkdir -p ${HOME:-/root}/.kube && cp -f $KUBECONFIG ${HOME:-/root}/.kube/config
     # Replace the server address with localhost, so that it works also from the host
     sed -e "/server:/ s|https://.*:\([0-9]*\)$|https://127.0.0.1:\1|" -i $KUBECONFIG
-    mkdir -p ${HOME:-/root}/.kube && cp -f $KUBECONFIG ${HOME:-/root}/.kube/config
+    {{end}}
 - mode: system
   script: |
     #!/bin/bash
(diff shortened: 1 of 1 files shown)


templates/k8s.yaml
```diff
@@ -136,44 +145,75 @@ probes:
     fi
   hint: |
     See "/var/log/cloud-init-output.log" in the guest
-- description: "kubernetes images to be pulled"
+# FIXME: probes[].description doesn't seem to recognize .Param.kubeadmInit
```
AkihiroSuda, line 148: FIXME.
Ideally it should be possible to wrap the entire probe entry in {{if}}

templates/k8s.yaml
```diff
@@ -136,44 +145,75 @@ probes:
...
-- guest: "/etc/kubernetes/admin.conf"
-  host: "{{.Dir}}/copied-from-guest/kubeconfig.yaml"
+- guest: "{{if eq .Param.kubeadmInit \"true\"}}/etc/kubernetes/admin.conf{{else}}/dev/null{{end}}"
+  host: "{{if eq .Param.kubeadmInit \"true\"}}{{.Dir}}/copied-from-guest/kubeconfig.yaml{{else}}{{.Dir}}/copied-from-guest/null{{end}}"
   deleteOnStop: true
+# FIXME: set .portForwards[].ignore when kubeadmInit is false
+param:
+  # Set to false for `kubeadm join`
+  kubeadmInit: true
+# FIXME: message doesn't seem to recognize .Param.kubeadmInit
```
AkihiroSuda, line 204: FIXME

templates/k8s.yaml
```diff
@@ -136,44 +145,75 @@ probes:
...
+    {{else}}
+    echo >&2 "skipping the coredns deployment check since kubeadmInit is set to false"
+    {{end}}
 copyToHost:
-- guest: "/etc/kubernetes/admin.conf"
-  host: "{{.Dir}}/copied-from-guest/kubeconfig.yaml"
+- guest: "{{if eq .Param.kubeadmInit \"true\"}}/etc/kubernetes/admin.conf{{else}}/dev/null{{end}}"
+  host: "{{if eq .Param.kubeadmInit \"true\"}}{{.Dir}}/copied-from-guest/kubeconfig.yaml{{else}}{{.Dir}}/copied-from-guest/null{{end}}"
   deleteOnStop: true
+# FIXME: set .portForwards[].ignore when kubeadmInit is false
```
AkihiroSuda, line 200: FIXME. `ignore` doesn't seem to support `{{if}}.

templates/k8s.yaml
```diff
@@ -136,44 +145,75 @@ probes:
...
+param:
+  # Set to false for `kubeadm join`
+  kubeadmInit: true
```
AkihiroSuda, line 203: Bikeshedding: `kubeadmInit` or `KUBEADM_INIT`

templates/k8s.yaml
```diff
@@ -136,44 +145,75 @@ probes:
...
     set -eux -o pipefail
+    {{if eq .Param.kubeadmInit "true"}}
     kubectl wait -n kube-system --timeout=180s --for=condition=available deploy coredns
+    {{else}}
+    echo >&2 "skipping the coredns deployment check since kubeadmInit is set to false"
+    {{end}}
 copyToHost:
-- guest: "/etc/kubernetes/admin.conf"
-  host: "{{.Dir}}/copied-from-guest/kubeconfig.yaml"
+- guest: "{{if eq .Param.kubeadmInit \"true\"}}/etc/kubernetes/admin.conf{{else}}/dev/null{{end}}"
```
AkihiroSuda, line 197: This hacky workaround is needed because `copyToHost` cannot be wrapped in `{{if}}`

templates/k8s.yaml
jandubois: Sorry, only drive-by commenting because I saw the email notification:

I think we should use all uppercase names similar to environment variables for template-specific parameters. They also stand out much better in scripts, and look better if you use `$PARAM_KUBEADM_INIT`.

I prefer to use the shell variables whenever possible because they remain visible in the `cloud-init-output.log` whereas Go templates are just replaced, and the variable name is lost.

This also leaves lower-case params for use later as "global" conventions for stuff that applies to different templates. I don't expect that mixed-case names will have special meaning though, so the decision should be based on usability and aesthetics.

AkihiroSuda: > all uppercase names similar to environment variables 

These names can be easily confused as env vars?

`docker.yaml` isUsingThisCase:
https://github.com/lima-vm/lima/blob/c71bc486917b61d7f4041fd2173461f5950e8926/templates/docker.yaml#L96-L97

AkihiroSuda:
This should suffice for now:
- #4234

PR #4147, opened 2025-10-06
CI: split the BATS tests into a separate job

AkihiroSuda:
The BATS tests should not be associated with `test-misc.yaml`.


jandubois (approved):
Thanks, LGTM

PR #4148, opened 2025-10-06
bats: drop {00, 01, ..}- prefix

AkihiroSuda:
Only some tests had the numeric prefix


jandubois:
The point of the prefix was to make sure these tests run first. If the `yq` and `list` tests are failing, then the failure of the remaining tests is kind of meaningless because they rely on `yq` and `list`.

But I guess it doesn't really matter as long as we run them all, and the mix of numbered and not-numbered tests looks kind of ugly.

jandubois (approved):
Thanks, LGTM

PR #4159, opened 2025-10-07
Allow `github:ORG//` repos to redirect to other repos in the same org

jandubois:
See https://github.com/lima-vm/lima/pull/4134#issuecomment-3369300717 for background.

---

Regarding https://github.com/lima-vm/lima/pull/4134#discussion_r2404972097, it is now possible to implement this yourself, if you want to:

```console
❯ cat ~/bin/limactl-url-gh
#!/bin/bash
echo "github:$1"
❯ l tmpl url gh:jandubois
WARN[0003] The github: scheme is still EXPERIMENTAL
https://raw.githubusercontent.com/jandubois/jandubois/main/templates/demo.yaml
```

### DIFF
--- a/hack/bats/tests/url-github.bats
+++ b/hack/bats/tests/url-github.bats
@@ -6,19 +6,33 @@ load "../helpers/load"
 # The jandubois/jandubois GitHub repo has been especially constructed to test
 # various features of the github URL scheme:
 #
-# * repo defaults to org when not specified
-# * filename defaults to .lima.yaml when only a path is specified
-# * .yaml default extension
-# * .lima.yaml files may be treated as symlinks
-# * default branch lookup when not specified
+#   * repo defaults to org when not specified
+#   * filename defaults to .lima.yaml when only a path is specified
+#   * .yaml default extension
+#   * .lima.yaml files may be treated as symlinks
+#   * default branch lookup when not specified
+#   * github:ORG// repos can redirect to another github:ORG URL in the same ORG
 #
-# The repo files are:
+# The jandubois/jandubois repo files are:
 #
-# ├── .lima.yaml -> templates/demo.yaml
-# ├── docs
-# │   └── .lima.yaml -> ../templates/demo.yaml
-# └── templates
-#     └── demo.yaml
+#   ├── .lima.yaml         -> templates/demo.yaml
+#   ├── back
+#   │   └── .lima.yaml     -> github:jandubois//loop/
+#   ├── docs
+#   │   └── .lima.yaml     -> ../templates/demo.yaml
+#   ├── invalid
+#   │   ├── org
+#   │   │   └── .lima.yaml -> github:lima-vm
+#   │   └── tag
+#   │       └── .lima.yaml -> github:jandubois//@v0.0.0
+#   ├── loop
+#   │   └── .lima.yaml     -> github:jandubois//back/
+#   ├── redirect
+#   │   └── .lima.yaml     -> github:jandubois/lima/templates/default
+#   ├── templates
+#   │   └── demo.yaml      "base: template:default"
+#   └── yaml
+#       └── .lima.yaml     "{}"
 #
 # Both the `main` branch and the `v0.0.0` tag have this layout.
 
@@ -52,7 +66,7 @@ URLS=(
 )
 
 url() {
-	run_e "$1" limactl template url "$2"
+	run_e "$1" limactl --debug template url "$2"
 }
 
 test_jandubois_url() {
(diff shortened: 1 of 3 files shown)


AkihiroSuda:
On second though the "CORS" restriction may not make much sense for GitHub URLs, as a `COMMIT` hash in `https://raw.githubusercontent.com/ORG/REPO/COMMIT/PATH` can (surprisingly) refer to a fork repo outside the `ORG`

jandubois:
> On second though the "CORS" restriction may not make much sense for GitHub URLs, as a `COMMIT` hash in `https://raw.githubusercontent.com/ORG/REPO/COMMIT/PATH` can (surprisingly) refer to a fork repo outside the `ORG`

It is not so surprising once you realize that all forks are part of the same repo on the server side, and serve simply as namespaces for branch and tag names.

This is also the reason GitHub does not allow public forks of private repos. Because if the SHA1 of a private commit is leaked, it becomes accessible through the public fork.

I already mentioned before that we don't restrict redirects on http:// and https:// URLs either, so it is a little odd to have them on github: URLs.

But I think the use case is really to redirect from the github:ORG// repo to a different place in the same ORG, not to be a generic link redirector.

I can remove the restriction, if you want; I don't think it matters much either way.

But we can also remove the restriction later, once an actual use case for cross-org linking has appeared.

jandubois:
While working on the tests for this PR I tripped myself a few times over the problem that except for the root redirect of `github:jandubois`, all others need to specify a trailing slash to mark them as directories:

```console
❯ l tmpl url github:jandubois//redirect
https://raw.githubusercontent.com/jandubois/jandubois/main/redirect.yaml

❯ l tmpl url github:jandubois//redirect/
https://raw.githubusercontent.com/jandubois/lima/master/templates/default.yaml
```

Without the trailing slash `redirect` is treated as `redirect.yaml` and cannot redirect to another repo, even if `redirect.yaml` doesn't exist.

If I can accidentally forget to add the slash while working on the tests, then it is very likely that casual users will also run into it[^1].

[^1]: For similar reasons many websites redirect http://example.com/ to http://example.com (or vice versa).

I think we can do 2 things to mitigate it:

1. Treat `foo` as `foo/` if `foo.yaml` does not exist (and `foo` has no extension).

    So we need to make at least a HEAD request for `foo.yaml`, and if that fails we decide that this was meant to look up `foo/.lima.yaml` instead.

2. Allow all `*.yaml` references (only in ORG repos?) to be symlinks or redirects

    This has the advantage that the user doesn't have to create extra subdirectories as containers for `.lima.yaml`, as there is likely nothing else in there. Instead `foo.yaml` could contain the redirect URL (or symlink).

I think (2) is more useful than (1), but if we do (2) and already fetch the file, then we can just do (1) at the same time at no extra cost.

It is just a fallback when `foo.yaml` doesn't exist, so the alternatives are either return an error right away, or retry the `foo/.lima.yaml` symlink as well.

I think we should implement both (1) and (2). For consistency we should probably also implement them for non-ORG repos, but of course they would only allow symlinks and not redirects to other repos.

This should again be done in a separate PR because this one is already big enough.

But let me know what you think first!

AkihiroSuda:
> ```
> ❯ l tmpl url github:jandubois//redirect/
> https://raw.githubusercontent.com/jandubois/lima/master/templates/default.yaml
> ```


I'd rather expect `github:jandubois//redirect/` to be expanded as `https://raw.githubusercontent.com/jandubois/jandubois/main/redirect/.lima.yaml` regardless to the exisstence of `/redirect.yaml` in the repo.
This should match the behavior of `foo/index.html` in HTTP servers

jandubois:
> I'd rather expect `github:jandubois//redirect/` to be expanded as `https://raw.githubusercontent.com/jandubois/jandubois/main/redirect/.lima.yaml` regardless to the exisstence of `/redirect.yaml` in the repo.
> This should match the behavior of `foo/index.html` in HTTP servers

That's what it does. You are seeing the effect of the redirect from there:

```console
❯ curl https://raw.githubusercontent.com/jandubois/jandubois/main/redirect/.lima.yaml
github:jandubois/lima/templates/default
```

The discussion above is just about URLs **without** a trailing slash like `github:jandubois//redirect`, which will always get a `.yaml` extension and will not be treated as a directory, regardless if the YAML file exists or not.

Currently only `.lima.yaml` files can be redirects (and directory URLs that expand to `.lima.yaml` as the default filename).

I think directories without trailing slashes should be able to redirect as well because it is easy to forget to add one, just like people forget to add one to `https://` URLs, so the server redirects for you automatically.

Either way, that should be a separate PR from this one.

AkihiroSuda:
> 1. Treat foo as foo/ if foo.yaml does not exist (and foo has no extension).

SGTM.
Alternatively it may just return an error message that suggests to attempt adding the `/` suffix


> 2. Allow all *.yaml references (only in ORG repos?) to be symlinks or redirects

This restriction sounds confusing

AkihiroSuda:
Is this PR ready to merge?

jandubois:
> Is this PR ready to merge?

I think so (that's why I asked for reviews last week), but somebody has to approve it...

The changes from the followup discussion should be done in a separate PR. I should probably have created a new issue for it, but it seemed like this PR provides the needed context.

AkihiroSuda (approved):
Thanks, let's merge this

PR #4168, opened 2025-10-08
print an error that wraps ExitError

### LINKED ISSUE #4167: `--network lima:shared` fails without printing an error when the sudoers file is missing
```console
$ limactl start --network lima:shared 
WARN[0000] Template locator "template://_images/ubuntu" should be written "template:_images/ubuntu" since Lima v2.0 
WARN[0000] Template locator "template://_default/mounts" should be written "template:_default/mounts" since Lima v2.0 

$ echo $?
1
```

Lima just fails without printing an error when `--network lima:shared` is specified but `/etc/sudoers.d/lima` is missing.

The error is still printed in the debug log though:
```
DEBU[0000] Make sure "shared" network is running        
DEBU[0000] "/private/etc/sudoers.d/lima" does not exist; passwordLessSudo error: failed to run [sudo --user root --group wheel --non-interactive true]: exit status 1
```

Lima v2.0.0-alpha.2, socket_vmnet v1.2.1, macOS 26.0.1

AkihiroSuda:
Fix #4167

### DIFF
--- a/pkg/osutil/exit.go
+++ b/pkg/osutil/exit.go
@@ -4,18 +4,22 @@
 package osutil
 
 import (
-	"errors"
 	"os"
 	"os/exec"
 )
 
+// HandleExitError calls os.Exit immediately without printing an error, only if the error is an *exec.ExitError (non-nil).
+//
+// The function does not call os.Exit if the error is of any other type, even if it wraps an *exec.ExitError,
+// so that the caller can print the error message.
 func HandleExitError(err error) {
 	if err == nil {
 		return
 	}
 
-	var exitErr *exec.ExitError
-	if errors.As(err, &exitErr) {
+	// Do not use errors.As, because we want to match only *exec.ExitError, not wrapped ones.
+	// https://github.com/lima-vm/lima/pull/4168
+	if exitErr, ok := err.(*exec.ExitError); ok {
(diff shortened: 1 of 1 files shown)


pkg/osutil/exit.go
```diff
@@ -5,18 +5,17 @@ package osutil
 
 import (
 	"errors"
-	"os"
 	"os/exec"
 )
 
-func HandleExitError(err error) {
+func HandleExitError(err error) int {
```
jandubois: The name is now misleading because it just returns the exit error and doesn't handle it.

```suggestion
func ExitError(err error) int {
```

pkg/osutil/exit.go
```diff
@@ -5,18 +5,17 @@ package osutil
...
 	if err == nil {
-		return
+		return 0
 	}
 
 	var exitErr *exec.ExitError
 	if errors.As(err, &exitErr) {
```
jandubois: I think the right fix would be to change from `errors.As` back to a type assertion, so it doesn't match on wrapped ExitErrors (which is what `networks.sudo()` is returning on failure).

This is a regression that I introduced in #3938. Add a comment to make sure it isn't being changed again.

We just don't want to print a message on "naked" ExitErrors that are directly returned by `cmd.Run` etc. because then propagating the exit code is enough.

AkihiroSuda: Fixed

PR #4170, opened 2025-10-08
hack/test-port-forwarding.pl: support `socat` as listener and writer

norio-nomura:
support `socat` as a listener and writer.

```console
$ ./hack/test-port-forwarding.pl default [nc|socat [nc|socat]] [timeout]
```

The `nc` command always sets `SO_REUSEPORT` on the listening TCP port.
The `socat` command provides a `reuseport` option that enables `SO_REUSEPORT` on the listening TCP port.

Ref: https://github.com/lima-vm/lima/pull/4160#issuecomment-3375485955

~~**In the current code, `test-port-forwarding.pl` is executed twice in each job using `nc` and `socat` for comparison.**~~ Dropped an execution using `nc`.

### DIFF
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -433,7 +433,7 @@ if [[ -n ${CHECKS["container-engine"]} ]]; then
 fi
 
 if [[ -n ${CHECKS["port-forwards"]} ]]; then
-	PORT_FORWARDING_CONNECTION_TIMEOUT=3
+	PORT_FORWARDING_CONNECTION_TIMEOUT=1
 	INFO "Testing port forwarding rules using netcat and socat with connection timeout ${PORT_FORWARDING_CONNECTION_TIMEOUT}s"
 	set -x
 	if [[ ${NAME} == "alpine"* ]]; then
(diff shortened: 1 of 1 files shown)


norio-nomura:
As far as I can see, tests that failed in `nc` were also successful in `socat`.
Is the instability of the port forwarding test caused by `nc`?

jandubois:
> Is the instability of the port forwarding test caused by `nc`?

I'm just browsing, but this PR still has failing port forwarding tests, e.g. https://github.com/lima-vm/lima/actions/runs/18364708138/job/52315306508?pr=4170. So it looks like switching to `socat` does not fix them.

norio-nomura:
> I'm just browsing, but this PR still has failing port forwarding tests, e.g. https://github.com/lima-vm/lima/actions/runs/18364708138/job/52315306508?pr=4170.

This PR does tests using both of `nc` and `socat` on the same job.
It has failed only on `nc`.

norio-nomura:
I guessed that the reason why `nc` fails is because the timeout is too short, but even if the timeout is extended to 3 seconds, `nc` continues to fail.

norio-nomura:
the history of workflow runs by this PR: https://github.com/lima-vm/lima/actions?query=branch%3Atest-port-forwarding.pl-use-socat+workflow%3Atest

jandubois:
> This PR does tests using both of `nc` and `socat` on the same job.

I'm sorry, I missed this.

I would be happy to switch to `socat`, even on a trial basis, to see if this helps. I'm so tired of re-running tests all the time.

norio-nomura:
> the history of workflow runs by this PR: https://github.com/lima-vm/lima/actions?query=branch%3Atest-port-forwarding.pl-use-socat+workflow%3Atest

To check failed logs in the workflow on local terminal, we can use `gh`:
`gh run list -b test-port-forwarding.pl-use-socat -w test --json conclusion,databaseId --jq '.[]|select(.conclusion=="failure")|.databaseId'|xargs -n 1 -I % sh -c "gh run view % --log-failed|awk '/\/hack\/test-port-forwarding.pl/{start=1};start==1{print}/Port forwarding rules/{start=0}'"`

norio-nomura:
> > the history of workflow runs by this PR: https://github.com/lima-vm/lima/actions?query=branch%3Atest-port-forwarding.pl-use-socat+workflow%3Atest
> 
> To check failed logs in the workflow on local terminal, we can use `gh`: `gh run list -b test-port-forwarding.pl-use-socat -w test --json conclusion,databaseId --jq '.[]|select(.conclusion=="failure")|.databaseId'|xargs -n 1 -I % sh -c "gh run view % --log-failed|awk '/\/hack\/test-port-forwarding.pl/{start=1};start==1{print}/Port forwarding rules/{start=0}'"`

I checked the `test-port-forwarding.pl`'s failure log of all runs after 9394 that were able to set up `socat` correctly, but none of them failed in `socat`.

norio-nomura:
https://github.com/lima-vm/lima/actions/runs/18367831847/job/52324053793?pr=4170#step:10:673
> time="2025-10-09T06:55:42Z" level=fatal msg="rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): stat /run/user/501/containerd-rootless: no such file or directory"

I often see this failure.

norio-nomura:
Thanks! 🙏🏻

PR #4176, opened 2025-10-10
Remove deprecated Opts from limayaml

afbjorklund:
Also remove the knowledge of the yaml structures
from the internal VM class, and make it use Config.

Follow-up from:
* #3899


afbjorklund:
I think this is a bit early to do now, and it is not ready... Just thought I would show the long-term plan.

It also removes a bit of "stutter" from the previous classes, that assumed there were 2-3 global drivers.

QEMUOpts -> pkg/driver/qemu.Opts
LimaQemuDriver -> pkg/driver/qemu.Driver

VZOpts -> pkg/driver/vz.Opts
LimaVzDriver -> pkg/driver/vz.Driver

PR #4184, opened 2025-10-11
Bump alpine from 3.22.0 → 3.22.2

jandubois:


jandubois:
> LGTM, but we usually update all the images at once

Oops, sorry, for some reason I thought there was no script for Alpine, but I guess I mixed it up with Alpine-ISO.

Alpine 3.22.2 was released 2 days ago, so there should not be another bump before the Lima 2.0 release.

Issue #4007, opened 2025-09-09
Add `limactl list --filter` option


jandubois:
After implementing the `--yq` option I found that I would like to have a `--filter EXPR` option, that would be equivalent to `--yq 'select(EXPR)'`.

This alone doesn't seem worthwhile, but I would want `--filter` to also work with `--format table` and `--format '{{GO-TMPL}}'`.

So you could do something like this:

```console
❯ limactl ls --filter '.status == "Running"'
NAME       STATUS     SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
docker     Running    127.0.0.1:60763    vz        aarch64    4       4GiB      100GiB    ~/.lima/docker
qemu       Running    127.0.0.1:63754    qemu      aarch64    4       4GiB      100GiB    ~/.lima/qemu
```

The `--filter` option can be specified multiple times, so you could have a `limactl-ps` plugin that runs the command above (with `"$@"`), and then use it like this:

```console
❯ limactl ps -f '.config.vmType == "vz"'
NAME       STATUS     SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
docker     Running    127.0.0.1:60763    vz        aarch64    4       4GiB      100GiB    ~/.lima/docker
```

olamilekan000:
can you assign to me

AkihiroSuda:
> The --filter option can be specified multiple times

Can't we just support `--filter "EXPR1 && EXPR2"` ?

jandubois:
> Can't we just support `--filter "EXPR1 && EXPR2"` ?

The filter expression would need to be using yq syntax, so you have to use `and` instead of `&&`:

```console
❯ l ls --yq 'select(.status == "Running" and .config.vmType == "vz")' -q
docker
```

But that doesn't unlock the use in a plugin I showed above. You can do this with `--yq` (once #4006 is merged):

```console
❯ cat ~/bin/limactl-ps
#!/bin/sh
limactl ls --yq 'select(.status == "Running")' "$@"

❯ l ps -q
docker
qemu

❯ l ps -q --yq 'select(.config.vmType == "vz")'
docker
```

You should be able to do the same kind of composable filtering with `--filter` and the default `--format table`. With `--yq` your output can only be JSON or YAML.

So it should look like this:

```console
❯ cat ~/bin/limactl-ps
#!/bin/sh
limactl ls --filter '.status == "Running"' "$@"

❯ l ps
[...]

PR #4187, opened 2025-10-12
add filter option to list command

olamilekan000:
```
change adds --filter option to list command for a user-friendly way to filter instances using yq expressions,
which is equivalent to yq options like --yq 'select(EXPR)'
```
[Fixes](https://github.com/lima-vm/lima/issues/4007)

Screenshot 2026-02-08 at 12 05 07

### DIFF
--- a/cmd/limactl/list.go
+++ b/cmd/limactl/list.go
@@ -6,6 +6,7 @@ package main
 import (
 	"bufio"
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"reflect"
@@ -56,7 +57,17 @@ The output can be presented in one of several formats, using the --format <forma
   --format yaml  - Output in YAML format
   --format table - Output in table format
   --format '{{ <go template> }}' - If the format begins and ends with '{{ }}', then it is used as a go template.
-` + store.FormatHelp,
+
+Filtering instances:
+  --filter EXPR  - Filter instances using yq expression (this is equivalent to --yq 'select(EXPR)')
+                   Can be specified multiple times (combined with AND logic) and it works with all output formats.
+                   Examples:
+                     --filter '.status == "Running"'
+                     --filter '.vmType == "vz"'
+                     --filter '.status == "Running"' --filter '.vmType == "vz"' (Same as AND)
+` + store.FormatHelp + `
+The following legacy flags continue to function:
+  --json - equal to '--format json'`,
 		Args:              WrapArgsError(cobra.ArbitraryArgs),
 		RunE:              listAction,
 		ValidArgsFunction: listBashComplete,
@@ -69,6 +80,7 @@ The output can be presented in one of several formats, using the --format <forma
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
 	listCommand.Flags().Bool("all-fields", false, "Show all fields")
 	listCommand.Flags().StringArray("yq", nil, "Apply yq expression to each instance")
+	listCommand.Flags().StringArrayP("filter", "l", nil, "Filter instances using yq expression (equivalent to --yq 'select(EXPR)')")
 
 	return listCommand
 }
@@ -118,6 +130,10 @@ func listAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	filter, err := cmd.Flags().GetStringArray("filter")
+	if err != nil {
+		return err
+	}
 
 	if jsonFormat {
 		format = "json"
@@ -136,6 +152,11 @@ func listAction(cmd *cobra.Command, args []string) error {
 		}
 		if listFields {
 			return errors.New("option --list-fields conflicts with option --yq")
+		}
+	}
+	if len(filter) != 0 {
+		if listFields {
+			return errors.New("option --list-fields conflicts with option --filter")
 		}
 	}
 
@@ -179,7 +200,7 @@ func listAction(cmd *cobra.Command, args []string) error {
 		instanceNames = allInstances
 	}
 
-	if quiet && len(yq) == 0 {
+	if quiet && len(yq) == 0 && len(filter) == 0 {
 		for _, instName := range instanceNames {
 			fmt.Fprintln(cmd.OutOrStdout(), instName)
 		}
@@ -197,6 +218,27 @@ func listAction(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("unable to load instance %s: %w", instanceName, err)
 		}
 		instances = append(instances, instance)
+	}
+
+	if len(filter) > 0 {
+		var filterExprs []string
+		for _, f := range filter {
+			filterExprs = append(filterExprs, "select("+f+")")
+		}
+		instances, err = filterInstances(instances, filterExprs)
+		if err != nil {
+			return err
+		}
+	}
+
+	if quiet && len(yq) == 0 {
+		for _, instance := range instances {
+			fmt.Fprintln(cmd.OutOrStdout(), instance.Name)
+		}
+		if unmatchedInstances {
+			return unmatchedInstancesError{}
+		}
+		return nil
 	}
 
 	for _, instance := range instances {
@@ -217,10 +259,12 @@ func listAction(cmd *cobra.Command, args []string) error {
 			options.TerminalWidth = w
 		}
 	}
-	// --yq implies --format json unless --format yaml has been explicitly specified
+
+	// --yq implies --format json unless --format has been explicitly specified
 	if len(yq) != 0 && !cmd.Flags().Changed("format") {
 		format = "json"
 	}
+
 	// Always pipe JSON and YAML through yq to colorize it if isTTY
 	if len(yq) == 0 && (format == "json" || format == "yaml") {
 		yq = append(yq, ".")
@@ -316,4 +360,35 @@ func listAction(cmd *cobra.Command, args []string) error {
 
 func listBashComplete(cmd *cobra.Command, _ []string, _ string) ([]string, cobra.ShellCompDirective) {
 	return bashCompleteInstanceNames(cmd)
+}
+
+// filterInstances applies yq expressions to instances and returns the filtered results.
+func filterInstances(instances []*limatype.Instance, yqExprs []string) ([]*limatype.Instance, error) {
+	if len(yqExprs) == 0 {
+		return instances, nil
+	}
+
+	// the yq expression is evaluated with yqutil.EvaluateExpression, which disables environment variable access
+	// and file operations, mitigating injection attacks like ".name=strenv(SOME_SECRET_ENV)" which could
+	// trick Lima into exposing environment variables.
+	yqExpr := strings.Join(yqExprs, " | ")
+
+	var filteredInstances []*limatype.Instance
+	for _, instance := range instances {
+		jsonBytes, err := json.Marshal(instance)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal instance %q: %w", instance.Name, err)
+		}
+
+		result, err := yqutil.EvaluateExpression(yqExpr, jsonBytes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to apply filter %q: %w", yqExpr, err)
+		}
+
+		if len(bytes.TrimSpace(result)) > 0 {
+			filteredInstances = append(filteredInstances, instance)
+		}
+	}
+
+	return filteredInstances, nil
 }
--- a/hack/bats/tests/list.bats
+++ b/hack/bats/tests/list.bats
@@ -274,3 +274,44 @@ local_setup() {
     run_e -1 limactl ls --yq "load(\"${BASH_SOURCE[0]}\")"
     assert_fatal "file operations have been disabled"
 }
+
+@test '--filter option filters instances' {
+    run -0 limactl ls --filter '.name == "foo"'
+    assert_line --index 0 --regexp '^NAME'
+    assert_line --index 1 --regexp '^foo'
+    assert_output_lines_count 2
+}
+
+@test '--filter option works with all output formats' {
+    run -0 limactl ls --filter '.name == "foo"'
+    assert_line --index 1 --regexp '^foo'
+
+    run -0 limactl ls --filter '.name == "foo"' --format json
+    assert_line --index 0 --regexp '^\{"name":"foo",'
+
+    run -0 limactl ls --filter '.name == "foo"' --format '{{.Name}}'
+    assert_output "foo"
+}
+
+@test '--filter option is compatible with --yq' {
+    run -0 limactl ls --filter '.name == "foo"' --yq '.name'
+    assert_output "foo"
+}
+
+@test '--quiet option can be used with --filter' {
+    run -0 limactl ls --quiet --filter '.name == "foo"'
+    assert_output "foo"
+}
+
+@test '--filter option with no matching instances returns empty output' {
+    run -0 limactl ls -q --filter '.name == "kcp"'
+    assert_output ""
+}
+
+@test 'multiple --filter options are combined with AND logic' {
+    run -0 limactl ls -q --filter '.name == "foo"' --filter '.disk == "1M"'
+    assert_output ""
+
+    run -0 limactl ls --filter '.name == "foo"' --filter '.disk == "1048576"'
+    assert_line --index 1 --regexp '^foo'
+}


cmd/limactl/list.go
```diff
@@ -320,3 +363,43 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+		jsonBytes, err := json.Marshal(instance)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal instance %s: %w", instance.Name, err)
+		}
+		jsonLines = append(jsonLines, string(jsonBytes))
+	}
+
+	var yqExprs []string
+	for _, filter := range filters {
+		yqExprs = append(yqExprs, "select("+filter+")")
```
AkihiroSuda: Needs escaping

cmd/limactl/list.go
```diff
@@ -220,15 +242,36 @@ func listAction(cmd *cobra.Command, args []string) error {
 			options.TerminalWidth = w
 		}
 	}
+	if len(filter) != 0 {
+		for _, f := range filter {
+			yq = append(yq, "select("+f+")")
```
AkihiroSuda: Needs escaping

- - -
**EDIT** (2025-10-20): Added an example of bad `f`:
```
. ) | .name=strenv(SOME_SECRET_ENV) | select(.
```

cmd/limactl/list.go
```diff
@@ -57,6 +58,14 @@ The output can be presented in one of several formats, using the --format <forma
   --format yaml  - Output in YAML format
   --format table - Output in table format
   --format '{{ <go template> }}' - If the format begins and ends with '{{ }}', then it is used as a go template.
+
+Filtering instances:
+  --filter EXPR  - Filter instances using yq expression (this is equivalent to --yq 'select(EXPR)')
```
AkihiroSuda, line 62: Maybe we do not need this flag. We can just add --yq select() examples to --help and call it a day

jandubois: That's why I wrote in #4007:

> This alone doesn't seem worthwhile, but I would want `--filter` to also work with `--format table` and `--format '{{GO-TMPL}}'`.

So you can do this (which isn't possible with `--yq`):

```console
❯ cat ~/bin/limactl-ps
#!/bin/sh
limactl ls --filter '.status == "Running"' "$@"

❯ l ps
docker     Running    127.0.0.1:60763    vz        aarch64    4       4GiB      100GiB    ~/.lima/docker
qemu       Running    127.0.0.1:63754    qemu      aarch64    4       4GiB      100GiB    ~/.lima/qemu

❯ l ps -f '.config.vmType == "vz"'
docker     Running    127.0.0.1:60763    vz        aarch64    4       4GiB      100GiB    ~/.lima/docker
```

I've not had time yet to review this PR to see if this is properly implemented.

cmd/limactl/list.go
olamilekan000: I tested numerous escaping approaches as asked, but it breaks the yq expressions. What string escape approach would you recommend? Thanks @AkihiroSuda

cmd/limactl/list.go
```diff
@@ -72,6 +81,7 @@ The following legacy flags continue to function:
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
 	listCommand.Flags().Bool("all-fields", false, "Show all fields")
 	listCommand.Flags().StringArray("yq", nil, "Apply yq expression to each instance")
+	listCommand.Flags().StringArray("filter", nil, "Filter instances using yq expression (equivalent to --yq 'select(EXPR)')")
```
jandubois: I just realized that `-f` is already in use for `--format`. 😞

I suggest we use `-l` as the short name for `--filter`, because it is the next consonant in `filter`, and also it is what `kubectl` uses for `--selector` which is somewhat similar.

An alternative would be to use `--select` and `-s`, but I think I still like `--filter` and `-l` better. What do other @lima-vm/maintainers think?

cmd/limactl/list.go
```diff
@@ -141,6 +155,14 @@ func listAction(cmd *cobra.Command, args []string) error {
 			return errors.New("option --list-fields conflicts with option --yq")
 		}
 	}
+	if len(filter) != 0 {
+		if listFields {
+			return errors.New("option --list-fields conflicts with option --filter")
+		}
+		if len(yq) != 0 {
+			return errors.New("option --filter conflicts with option --yq")
+		}
```
jandubois: I think you should be able to combine `--filter` and `--yq` as long as the output format is `json` or `yaml`.

So you could just drop the check here because `yq` is already incompatible with the other output formats.

cmd/limactl/list.go
```diff
@@ -220,15 +242,36 @@ func listAction(cmd *cobra.Command, args []string) error {
 			options.TerminalWidth = w
 		}
 	}
+	if len(filter) != 0 {
```
jandubois: This check is not necessary; `range` does the right thing on an empty slice.

cmd/limactl/list.go
jandubois: @AkihiroSuda I don't see why this needs escaping. The filter expression should be inserted literally inside the `select()`.

cmd/limactl/list.go
```diff
@@ -220,15 +242,36 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+		for _, f := range filter {
+			yq = append(yq, "select("+f+")")
+		}
+	}
+
 	// --yq implies --format json unless --format yaml has been explicitly specified
-	if len(yq) != 0 && !cmd.Flags().Changed("format") {
+	// But --filter should preserve the original format
+	if len(yq) != 0 && !cmd.Flags().Changed("format") && len(filter) == 0 {
```
jandubois: I don't understand this change. You cannot preserve the default `--format table` by adding a `--filter` expression. Once you use `--yq` the format must be either `yaml` or `json`.

So this change looks incorrect. (Also right now `--yq` is not allowed when you use `--filter`, so `len(filter)` is always `0` when `len(yq)` isn't, but I want you to change that).

cmd/limactl/list.go
```diff
@@ -220,15 +242,36 @@ func listAction(cmd *cobra.Command, args []string) error {
...
 		format = "json"
 	}
 	// Always pipe JSON and YAML through yq to colorize it if isTTY
 	if len(yq) == 0 && (format == "json" || format == "yaml") {
 		yq = append(yq, ".")
 	}
 
+	// handle --filter with table and go-template formats
+	tmpl := strings.HasPrefix(format, "{{") && strings.HasSuffix(format, "}}")
```
jandubois: The `tmpl` name is used throughout the codebase for `limatmpl.Template` instances, so please use a different name, like `isGoTemplate`.

On second thought, you should not check for this at all. Every string that is not `table`, `json`, or `yaml` is treated as a template.

cmd/limactl/list.go
```diff
@@ -220,15 +242,36 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+		filteredInstances, err := filterInstances(instances, filter)
+		if err != nil {
+			return err
+		}
+		err = store.PrintInstances(cmd.OutOrStdout(), filteredInstances, format, &options)
+		if err == nil && unmatchedInstances {
+			return unmatchedInstancesError{}
+		}
+		return err
+	}
```
jandubois: Don't repeat the logic below, just filter the instances and go through the shared logic:

```suggestion
	if len(filter) != 0 && format != "json" && format != "yaml" {
		// yq contains the concatenated filter expressions
		instances, err := filterInstances(instances, yq)
		if err != nil {
			return err
		}
	}
```

cmd/limactl/list.go
```diff
@@ -320,3 +363,43 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+			return nil, fmt.Errorf("failed to marshal instance %s: %w", instance.Name, err)
+		}
+		jsonLines = append(jsonLines, string(jsonBytes))
+	}
+
+	var yqExprs []string
+	for _, filter := range filters {
+		yqExprs = append(yqExprs, "select("+filter+")")
+	}
+	yqExpr := strings.Join(yqExprs, " | ")
```
jandubois: This block is unnecessary; you can pass in the `yq` string from the caller. This function is only used for table and Go template output, in which case there were no `--yq` options, so `yq` only contains the `--filter` expressions.

cmd/limactl/list.go
```diff
@@ -320,3 +363,43 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+		if len(result) > 0 && string(result) != "null" {
+			var instance limatype.Instance
+			if err := json.Unmarshal(result, &instance); err != nil {
+				return nil, fmt.Errorf("failed to unmarshal filtered instance: %w", err)
+			}
+			filteredInstances = append(filteredInstances, &instance)
+		}
+	}
+
+	return filteredInstances, nil
```
jandubois: This is too complicated. You don't need to create the `jsonLines` slice, and then unmarshal the JSON again. The whole function can be something like this (pseudo-code):

```
for _, instance := range instances {
    json.Marshal(instance)
    yq.Eval(yq)
    if len(result) > 0 {
        filteredInstances = append(filteredInstances, instance)
    }
}
return filteredInstances, nil
```

cmd/limactl/list.go
AkihiroSuda: Without escaping the code looks vulnerable to "Bobby Tables".
Unlikely to result in breaking data in our case though

jandubois: How is this different from any code you specify via `--yq`? The only difference is that the `--filter` argument is wrapped inside a `select()` call.

The user can literally send the same command via `--yq "$(printf "select(%s)" "$FILTER")"`.

Any escaping would simply break the filter expression.

cmd/limactl/list.go
olamilekan000: I think the confusion is from how the `yq` variable is set and how the query is created, I have moved it below the check so that the check isn't tampered with.

```sh
	if len(filter) != 0 {
		for _, f := range filter {
			yq = append(yq, "select("+f+")")
		}
	}
```

cmd/limactl/list.go
olamilekan000: This isn't too clear to me but is there anything wrong with the way it currently is?

jandubois: Yes:

```console
❯ l ls -l '.status == "Running"'
NAME          STATUS     SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
alpine-iso    Running    127.0.0.1:50496    vz        aarch64    4       4GiB      100GiB    ~/.lima/alpine-iso

❯ l ls -l '.status == "Running"' --yq .name
FATA[0000] option --filter conflicts with option --yq

❯ l ls --yq 'select(.status == "Running")' --yq .name
alpine-iso
```

I expect the last 2 commands to work exactly the same. There is no reason for the `len(yq)` check to exist.

cmd/limactl/list.go
```diff
@@ -220,15 +242,31 @@ func listAction(cmd *cobra.Command, args []string) error {
...
 		yq = append(yq, ".")
 	}
 
+	for _, f := range filter {
+		yq = append(yq, "select("+f+")")
+	}
+
+	// handle --filter with table and go-template formats
+	isGoTemplate := strings.HasPrefix(format, "{{") && strings.HasSuffix(format, "}}")
+	if len(filter) != 0 && (format == "table" || isGoTemplate) {
```
jandubois:
```suggestion
	if len(filter) != 0 && format != "json" && format != "yaml") {
```

Despite what the `--help` output says, **every** string except `table`, `json`, and `yaml` is treated as a Go template, so the check for the double braces prefix/suffix is not correct:

```console
❯ l ls -f 'Name is: {{.Name}}'
Name is: alpine-iso
Name is: alpine-lima-3.22.2
Name is: foo
```

cmd/limactl/list.go
```diff
@@ -320,3 +358,31 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+		return instances, nil
+	}
+
+	yqExpr := strings.Join(yqExprs, " | ")
+
+	var filteredInstances []*limatype.Instance
+	for _, instance := range instances {
+		jsonBytes, err := json.Marshal(instance)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal instance %s: %w", instance.Name, err)
```
jandubois:
```suggestion
			return nil, fmt.Errorf("failed to marshal instance %q: %w", instance.Name, err)
```

Always use `%q` when printing use input.

Also applies to the next error message.

cmd/limactl/list.go
AkihiroSuda: The difference is that `--filter` is likely considered safe to be exposed directly to a GUI frontend (possibly web UI).
_Bobby_ shouldn't be able to steal the server's [env vars](https://mikefarah.gitbook.io/yq/operators/env-variable-operators) via a filter textbox in a web UI.

`--yq` should be considered a "privileged" flag and should not be exposed to frontends, especially web-based ones.

olamilekan000: I have added a validator for the filter values. Does that work?

cmd/limactl/list.go
```diff
@@ -57,6 +59,14 @@ The output can be presented in one of several formats, using the --format <forma
...
   --format table - Output in table format
   --format '{{ <go template> }}' - If the format begins and ends with '{{ }}', then it is used as a go template.
+
+Filtering instances:
+  --filter EXPR  - Filter instances using yq expression (this is equivalent to --yq 'select(EXPR)')
+                   Can be specified multiple times and it works with all output formats.
+                   Examples:
+                     --filter '.status == "Running"'
+                     --filter '.vmType == "vz"'
+                     --filter '.status == "Running"' --filter '.vmType == "vz"'
```
AkihiroSuda: Is this AND-matching or OR-matching?

jandubois: I assume you want the description updated to clarify.

It is AND so you can do this:

```console
❯ cat ~/bin/limactl-ps
#!/bin/sh
limactl ls --filter 'select(.status == "Running")' "$@"

❯ l ls
NAME                  STATUS     SSH                VMTYPE    ARCH       CPUS    MEMORY      DISK      DIR
alpine-iso            Running    127.0.0.1:50496    vz        aarch64    4       4GiB        100GiB    ~/.lima/alpine-iso
alpine-lima-3.22.2    Stopped    127.0.0.1:0        vz        aarch64    4       4GiB        100GiB    ~/.lima/alpine-lima-3.22.2
default               Running    127.0.0.1:54512    qemu      aarch64    4       4GiB        100GiB    ~/.lima/default
foo                   Stopped    127.0.0.1:0        vz        aarch64    4       1.177MiB    100GiB    ~/.lima/foo

❯ l ps
NAME          STATUS     SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
alpine-iso    Running    127.0.0.1:50496    vz        aarch64    4       4GiB      100GiB    ~/.lima/alpine-iso
default       Running    127.0.0.1:54512    qemu      aarch64    4       4GiB      100GiB    ~/.lima/default

❯ l ps -l '.vmType == "vz"'
NAME          STATUS     SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
alpine-iso    Running    127.0.0.1:50496    vz        aarch64    4       4GiB      100GiB    ~/.lima/alpine-iso
```

cmd/limactl/list.go
```diff
@@ -220,15 +240,35 @@ func listAction(cmd *cobra.Command, args []string) error {
...
 	if len(yq) == 0 && (format == "json" || format == "yaml") {
 		yq = append(yq, ".")
 	}
 
+	for _, f := range filter {
+		// only allow fields, ==, !=, and literals.
+		valid := regexp.MustCompile(`^[a-zA-Z0-9_.\s"'-=><!]+$`)
+		if !valid.MatchString(f) {
+			return fmt.Errorf("unsafe characters in filter expression: %q", f)
+		}
```
jandubois: I don't think this is useful. The whole point of `--filter` and `--yq` is to provide YQ expressions that are going to be evaluated. There is no vulnerability; the expression is provided by the user themselves.

And we already get an error when the YQ expression is invalid, e.g.

```console
❯ l ls -l '.vmType =='
FATA[0000] failed to apply filter select(.vmType ==): '==' expects 2 args but there is 1
```

So I don't understand the purpose of this "validation", especially since it might make legitimate use cases invalid (e.g. you are restricting which characters are allowed inside literal strings).

jandubois (changes requested):
The combination of `--filter` and `--quiet` is not working (the filter is ignored):

```console
❯ l ls -l '.vmType == "qemu"'
default    Running    127.0.0.1:54512    qemu      aarch64    4       4GiB      100GiB    ~/.lima/default

❯ l ls -l '.vmType == "qemu"' -q
alpine-iso
alpine-lima-3.22.2
default
foo
```

cmd/limactl/list.go
jandubois: > The difference is that `--filter` is likely considered safe to be exposed directly to a GUI frontend (possibly web UI).

If this is just about `--filter`, and more generally just about `limactl list` and not every place we use `yq` expressions, then we can easily avoid exposing environment variables by calling `os.Clearenv()` in `listAction()` any time after we set `allInstances`.

Before:

```console
❯ l ls bar --yq 'strenv(PATH)'
…:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/X11/bin:/Users/jan/.local/bin
```

Change:

```diff
@@ -248,6 +249,7 @@ func listAction(cmd *cobra.Command, args []string) error {
 		return err
 	}

+	os.Clearenv()
 	if format == "json" {
 		// The JSON encoder will create empty objects (YAML maps), even when they have the ",omitempty" tag.
 		deleteEmptyObjects := `del(.. | select(tag == "!!map" and length == 0))`
```

After:

```console
❯ l ls bar --yq 'strenv(PATH)'
```

@AkihiroSuda Does this mitigate your concerns?

jandubois: Side note: the `--filter` command can only expose the environment variables when used with JSON or YAML output.

For table or Go templates the filter expression is just evaluated to check if the instance is to be included or not, but the command will print the original instance data, not the one potentially modified by the `--filter` expression. So it can only extract 1 bit of information per call.

But calling `os.Clearenv()` makes all of this irrelevant anyways.

jandubois: Second side note: I also didn't see any functionality in `yq` that would allow you to get a list of all environment variables. You have to know the variable name to query the value. Please let me know if I overlooked anything!

AkihiroSuda: Clearing the env may cause an unexpected effect to the entire process, including the Go runtime.

Can't we just add a new option in the YQ library to prohibit accessing envs?

At least https://pkg.go.dev/github.com/mikefarah/yq/v4@v4.48.1/pkg/yqlib#ExpressionNode could be verified that there is no reference to env vars.

jandubois: > Clearing the env may cause an unexpected effect to the entire process, including the Go runtime.

All we do after clearing the environment is calling `yqlib` to evaluate the expression, printing the result with `fmt.Fprint`, and exiting the process. I can't see how anything would be affected in an unexpected way.

The fact that `os.Getenv` will always return the empty string inside `yqlib` is not unexpected, but desired.

I would be opposed to modifying the process environment for every `yqlib` call, but in `limactl list` and `limactl info` it would be acceptable when all that is left to do is evaluating an expression and printing the result.

jandubois: > Can't we just add a new option in the YQ library to prohibit accessing envs?

I don't know; there doesn't seem to be an easy way to pass settings to the `env` operator etc.

I agree that this would be a good enhancement (so we could make it apply to all our `yqlib` calls), but for `limactl list` it seems like overkill because I don't see an easy way to implement it.

jandubois: @AkihiroSuda Are you ok with using `os.Clearenv()` in this particular case (`limactl list`, just before calling yqlib, printing the result, and exiting)?

I suspect @olamilekan000 is waiting for a decision before addressing the remaining issues, so this would be blocking it from being included in Lima 2.0.

olamilekan000: Yes. That's what I am waiting for

AkihiroSuda: I still don't like to approve `os.Clearenv()` even if it works for the particular case, for  long-term maintainability.

Is there any reason that we have to rush to merge this in v2.0?

cmd/limactl/list.go
AkihiroSuda: What if we just support `--format table` with `--yq`?

cmd/limactl/list.go
AkihiroSuda: > What if we just support `--format table` with `--yq`?

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/4187#discussion_r2443904224_

cmd/limactl/list.go
```diff
@@ -72,6 +82,7 @@ The following legacy flags continue to function:
 	listCommand.Flags().BoolP("quiet", "q", false, "Only show names")
 	listCommand.Flags().Bool("all-fields", false, "Show all fields")
 	listCommand.Flags().StringArray("yq", nil, "Apply yq expression to each instance")
+	listCommand.Flags().StringArrayP("filter", "l", nil, "Filter instances using yq expression (equivalent to --yq 'select(EXPR)')")
```
AkihiroSuda, line 83: The `l` shorthand should be probably reserved for `--label`
- https://github.com/lima-vm/lima/issues/3240

cmd/limactl/list.go
AkihiroSuda: Let me move this to the v2.1+ milestone.

cmd/limactl/list.go
olamilekan000: what should be used?

jandubois: I don't think there is a separate `--label` option; it is the same as `--filter`, which is another reason that I think `-l` is appropriate. But let's wait until KubeCon is over and everyone (or at least @AkihiroSuda) is back, and continue the discussion then. Right now the 2.0 release, and potential regressions is still top-of-mind.

cmd/limactl/list.go
jandubois: I still think calling `os.Clearenv()` is fine as long as we exit the process after printing the results from the yqlib call.

But I've also opened https://github.com/mikefarah/yq/issues/2515. Maybe we can get an option upstream to either pass in an env map, or to disable env lookup functions.

olamilekan000:
Pls what's the conclusion on the implementation for this? @jandubois @AkihiroSuda

AkihiroSuda:
This issue is still unresolved
https://github.com/lima-vm/lima/pull/4405#discussion_r2559735005

cmd/limactl/list.go
```diff
@@ -57,6 +59,14 @@ The output can be presented in one of several formats, using the --format <forma
   --format yaml  - Output in YAML format
   --format table - Output in table format
   --format '{{ <go template> }}' - If the format begins and ends with '{{ }}', then it is used as a go template.
+
+Filtering instances:
+  --filter EXPR  - Filter instances using yq expression (this is equivalent to --yq 'select(EXPR)')
```
AkihiroSuda, line 62: Maybe we have discussed this before, but an alternative design would be to use https://pkg.go.dev/github.com/containerd/containerd/v2/pkg/filters
instead of yq

AkihiroSuda: Draft PR
- https://github.com/lima-vm/lima/pull/4529

hack/bats/tests/list.bats
AkihiroSuda, line 1: This should check YQ expression injections like `.) | .name=strenv(SOME_SECRET_ENV) | select(.`

jandubois: I don't understand. With #4405 merged there should be no need to check YQ expressions at all.

All env and file operators will throw an error:

```console
❯ limactl ls --yq 'strenv(USER)'
FATA[0000] env operations have been disabled
```

And we have a test for that already: https://github.com/lima-vm/lima/blob/a4e47ed0c27a9652f8df9b39679148699a22b21d/hack/bats/tests/list.bats#L268-L271

AkihiroSuda:
This PR needs rebasing at least

olamilekan000:
Okay. I have the intention of closing his up this week at least. I've just been swamped recently 😞

cmd/limactl/list.go
```diff
@@ -317,3 +361,31 @@ func listAction(cmd *cobra.Command, args []string) error {
...
 	return bashCompleteInstanceNames(cmd)
 }
+
+// filterInstances applies yq expressions to instances and returns the filtered results.
+func filterInstances(instances []*limatype.Instance, yqExprs []string) ([]*limatype.Instance, error) {
+	if len(yqExprs) == 0 {
+		return instances, nil
+	}
+
+	yqExpr := strings.Join(yqExprs, " | ")
```
AkihiroSuda, line 374: Add a code comment to explain why this is not vulnerable to expression injections like `. ) | .name=strenv(SOME_SECRET_ENV) | select(.`

cmd/limactl/list.go
```diff
@@ -56,7 +57,17 @@ The output can be presented in one of several formats, using the --format <forma
...
   --format '{{ <go template> }}' - If the format begins and ends with '{{ }}', then it is used as a go template.
-` + store.FormatHelp,
+
+Filtering instances:
+  --filter EXPR  - Filter instances using yq expression (this is equivalent to --yq 'select(EXPR)')
+                   Can be specified multiple times and it works with all output formats.
+                   Examples:
+                     --filter '.status == "Running"'
+                     --filter '.vmType == "vz"'
+                     --filter '.status == "Running"' --filter '.vmType == "vz"'
```
jandubois: I think there was some questions if `--filter` expressions are combined by AND or OR (it is AND), so I think this should be documented in the help.

cmd/limactl/list.go
```diff
@@ -317,3 +361,34 @@ func listAction(cmd *cobra.Command, args []string) error {
...
 	return bashCompleteInstanceNames(cmd)
 }
+
+// filterInstances applies yq expressions to instances and returns the filtered results.
+func filterInstances(instances []*limatype.Instance, yqExprs []string) ([]*limatype.Instance, error) {
+	if len(yqExprs) == 0 {
+		return instances, nil
+	}
+
+	// the yq expression is evaluated with yqutil.EvaluateExpressionWithEncoder, which disables environment variable access
```
jandubois:
```suggestion
	// the yq expression is evaluated with yqutil.EvaluateExpression, which disables environment variable access
```

cmd/limactl/list.go
```diff
@@ -317,3 +361,34 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal instance %q: %w", instance.Name, err)
+		}
+
+		result, err := yqutil.EvaluateExpression(yqExpr, jsonBytes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to apply filter %q: %w", yqExpr, err)
+		}
+
+		if len(result) > 0 {
```
jandubois: Maybe trim whitespace, in case there is some stray space or newline instead of an empty string?


```suggestion
		if len(bytes.TrimSpace(result)) > 0 {
```

hack/bats/tests/list.bats
```diff
@@ -274,3 +274,31 @@ local_setup() {
...
+    run -0 limactl ls --filter '.name == "foo"'
+    assert_line --index 0 --regexp '^NAME'
+    assert_line --index 1 --regexp '^foo'
+    assert_output_lines_count 2
+}
+
+@test '--filter option works with all output formats' {
+    run -0 limactl ls --filter '.name == "foo"'
+    assert_line --index 1 --regexp '^foo'
+
```
jandubois: Almost missed this; we need a linter test for this: there should be no trailing whitespace on any source code. 3 lines down has the same issue.

cmd/limactl/list.go
```diff
@@ -317,3 +361,34 @@ func listAction(cmd *cobra.Command, args []string) error {
...
+
+// filterInstances applies yq expressions to instances and returns the filtered results.
+func filterInstances(instances []*limatype.Instance, yqExprs []string) ([]*limatype.Instance, error) {
+	if len(yqExprs) == 0 {
+		return instances, nil
+	}
+
+	// the yq expression is evaluated with yqutil.EvaluateExpression, which disables environment variable access
+	// and file operations, mitigating injection attacks like ".name=strenv(SOME_SECRET_ENV)" which could
+	// trick Lima into exposing environment variables.
```
jandubois, line 373: I think I mentioned before that this comment is excessive, but it doesn't really matter.

PR #4197, opened 2025-10-14
editflags: support multiple `--set` flags instead of ignoring all but the last such parameter

AndiDog:
I wanted to use lima-based development instructions (from the Kubernetes [CAPI project docs](https://main.cluster-api.sigs.k8s.io/developer/core/tilt.html?highlight=lima#using-lima)):

```
limactl start template://docker-rootful --name "docker" --tty=false \
  --set '.provision += {"mode":"system","script":"#!/bin/bash\nset -eux -o pipefail\ncat << EOF > \"/etc/sysctl.d/99-capi.conf\"\nfs.inotify.max_user_watches = 1048576\nfs.inotify.max_user_instances = 8192\nEOF\nsysctl -p \"/etc/sysctl.d/99-capi.conf\""}' \
  --set '.mounts[0] = {"location": "~", "writable": true}' \
  --memory 12 --cpus 10 --disk 64 \
  --vm-type vz --rosetta=true
```

and it turned out that the sysctl settings that are supposed to be applied in the first `--set` argument are totally ignored without warning because that's how command line parsers usually work. Users should definitely be able to specify this multiple times, so I added support for that.

### DIFF
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -66,7 +66,7 @@ func RegisterEdit(cmd *cobra.Command, commentPrefix string) {
 
 	flags.Bool("rosetta", false, commentPrefix+"Enable Rosetta (for vz instances)")
 
-	flags.String("set", "", commentPrefix+"Modify the template inplace, using yq syntax")
+	flags.StringArray("set", []string{}, commentPrefix+"Modify the template inplace, using yq syntax. Can be passed multiple times.")
 
 	flags.Uint16("ssh-port", 0, commentPrefix+"SSH port (0 for random)") // colima-compatible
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
I thought we had a similar PR in the past but can't find it

jandubois (approved):
> There are already several other options that can be specified multiple times: --dns, --mount, --mount-only, --network. They all build a single expression in their exprFunc.

I've looked at this again, and these options are handled in a significantly different way.

At first I thought this PR was changing too many lines for something that could be localized to just a single short block, but given that `YQExpressions` returns a `[]string` and not just a `string`, I've changed my mind and think this PR is fine as-is.

Sorry for the confusion!

Issue #4186, opened 2025-10-12
Kubernetes wants to stop distributing kubernetes-cni package


afbjorklund:
### Description

Currently we are using the upstream binaries that are installed in `/opt/cni`.

But Kubernetes and kubeadm wants to stop packaging "kubernetes-cni".

* https://github.com/kubernetes/release/issues/4156

Once they are no longer there, we need to switch to `/usr/local/libexec/cni`.

It is the same binaries (same source at least), just that they are installed separately.

```
Files /opt/cni/bin/LICENSE and /usr/local/libexec/cni/LICENSE are identical
Files /opt/cni/bin/README.md and /usr/local/libexec/cni/README.md are identical
Binary files /opt/cni/bin/bandwidth and /usr/local/libexec/cni/bandwidth differ
Binary files /opt/cni/bin/bridge and /usr/local/libexec/cni/bridge differ
Binary files /opt/cni/bin/dhcp and /usr/local/libexec/cni/dhcp differ
Binary files /opt/cni/bin/dummy and /usr/local/libexec/cni/dummy differ
Binary files /opt/cni/bin/firewall and /usr/local/libexec/cni/firewall differ
Only in /opt/cni/bin/: flannel
Binary files /opt/cni/bin/host-device and /usr/local/libexec/cni/host-device differ
Binary files /opt/cni/bin/host-local and /usr/local/libexec/cni/host-local differ
Binary files /opt/cni/bin/ipvlan and /usr/local/libexec/cni/ipvlan differ
Binary files /opt/cni/bin/loopback and /usr/local/libexec/cni/loopback differ
Binary files /opt/cni/bin/macvlan and /usr/local/libexec/cni/macvlan differ
Binary files /opt/cni/bin/portmap and /usr/local/libexec/cni/portmap differ
Binary files /opt/cni/bin/ptp and /usr/local/libexec/cni/ptp differ
Binary files /opt/cni/bin/sbr and /usr/local/libexec/cni/sbr differ
Binary files /opt/cni/bin/static and /usr/local/libexec/cni/static differ
Binary files /opt/cni/bin/tap and /usr/local/libexec/cni/tap differ
Binary files /opt/cni/bin/tuning and /usr/local/libexec/cni/tuning differ
Binary files /opt/cni/bin/vlan and /usr/local/libexec/cni/vlan differ
Binary files /opt/cni/bin/vrf and /usr/local/libexec/cni/vrf differ
```

[...]

PR #4198, opened 2025-10-14
Avoid relying on extra kubernetes packages

afbjorklund:
The second-party package of "cri-tools" (crictl) and also the third-party package "kubernetes-cni" (cni-plugins) will go away.

So **don't** rely on them being installed, but just use "ctr" and "/usr/local/libexec/cni" instead from the `nerdctl-full` install.

* https://github.com/lima-vm/lima/issues/4186

* https://github.com/kubernetes/release/issues/4156

---

Note: the packages are still installed by default, in Kubernetes version 1.34...

But this way, the k8s installation will continue to work - even if they are **not**

The main difference from before, is that there will be small warning printed:

`WARN[0000] Image connect using default endpoints: [unix:///run/containerd/containerd.sock unix:///run/crio/crio.sock unix:///var/run/cri-dockerd.sock]. As the default settings are now deprecated, you should set the endpoint instead.`

It has been deprecated for 5 years, but the user is supposed to configure it:

https://kubernetes.io/docs/tasks/debug/debug-cluster/crictl/ (in `/etc/crictl.yaml`)

And the other cni-plugins are upgraded from version 1.7.1 to version 1.8.0.

It can now use the standard version from `nerdctl`, instead of from kubelet.

### DIFF
--- a/templates/k8s.yaml
+++ b/templates/k8s.yaml
@@ -47,14 +47,6 @@ provision:
     echo "deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v${VERSION}/deb/ /" | sudo tee /etc/apt/sources.list.d/kubernetes.list
     curl -fsSL https://pkgs.k8s.io/core:/stable:/v${VERSION}/deb/Release.key | sudo gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
     apt-get update
-    # cri-tools
-    apt-get install -y cri-tools
-    cat  <<EOF | sudo tee /etc/crictl.yaml
-    runtime-endpoint: unix:///run/containerd/containerd.sock
-    EOF
-    # cni-plugins
-    apt-get install -y kubernetes-cni
-    rm -f /etc/cni/net.d/*.conf*
     apt-get install -y kubelet kubeadm kubectl && apt-mark hold kubelet kubeadm kubectl
     systemctl enable --now kubelet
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
For testing let's merge this first
- https://github.com/lima-vm/lima/pull/4118

templates/k8s.yaml
```diff
@@ -77,6 +69,8 @@ provision:
               runtime_type = "io.containerd.runc.v2"
               [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc.options]
                 SystemdCgroup = true
+      [plugins.'io.containerd.cri.v1.runtime'.cni]
+        bin_dirs = ['/opt/cni/bin','/usr/local/libexec/cni']
```
AkihiroSuda: Probably we do not need this config if we symlink it

afbjorklund: The symlink conflicts with the current package, and is part of the kubernetes-cni legacy (why `/opt`?)

But we need to add it to the path, since the flannel manifest still uses it for _their_ cni plugin install

Should put the newer binaries first, though.

The kubernetes ones are usually forgotten...

`kubernetes-cni/unknown,now 1.7.1-1.1 amd64 [installed,automatic]`
`kubernetes-cni/unknown 1.6.0-1.1 amd64`

"Install CNI plugins (required for most pod network):"
`CNI_PLUGINS_VERSION="v1.3.0"`

afbjorklund:
This is just something that is being discussed, it has not happened yet (in pre-release `v1.35.0-alpha.1`)

https://www.kubernetes.dev/resources/release/

https://build.opensuse.org/project/show/isv:kubernetes:core:prerelease:v1.35

PR #4202, opened 2025-10-15
templates: update Ubuntu to 25.10

AkihiroSuda:


templates/experimental/ubuntu-next.yaml
jandubois, line 1: Should we point `ubuntu-next` to `ubuntu` if the next version is not yet available?

Would be helpful if somebody uses `ubuntu-next` in CI to test with the upcoming version. When we just delete the file, their CI will break.

jandubois (approved):
Thanks, LGTM

templates/experimental/ubuntu-next.yaml
AkihiroSuda: fixed

Issue #3618, opened 2025-06-07
VM drivers for supporting macOS guests (`vz`, `host-user`)


AkihiroSuda:
Supporting macOS guests will be useful for running untrusted packages or AI-generated scripts

## Driver `vz`

The current [vz](https://developer.apple.com/documentation/virtualization) driver could be extended to support macOS guests, however, setting up an instance cannot be easily automated.

The first invocation of `limactl start --vm-type=vz template://macos` will forcibly open the GUI screen, and ask the user to do the following tasks manually:
- Accept EULA
- Create a user account with the specified name
- Skip the iCloud stuff
- Open the System Settings, and enable SSH
- Run a script in the virtual CD-ROM to set up `~/.ssh/authorized_keys` etc.
- Shutdown the VM

### Shell
Same as the current vz for Linux guests.

### Filesystem
Same as the current vz for Linux guests.

### Network
Same as the current vz for Linux guests.

### Port forwarding
TBD. Periodically run `lsof` to scan the ports?

### GPU
Depends on Apple's will.

## Driver `host-user`

The new "host-user" driver (tentative) will incorporate https://github.com/AkihiroSuda/alcless so as to create a pseudo-sandbox account using the plain old su/sudo.
(su and sudo have to be mixed together; see https://github.com/AkihiroSuda/alcless/blob/master/README.md#faqs)

The main advantage of this driver is the native access to the host GPU.
This driver may lose its raisondetre when Apple implements GPU passthrough in vz.

### Shell
Unlike other drivers, `limactl shell` can be implemented without using SSH.

### Filesystem
- Step 1: No support for `mounts`.
- Step 2: Extend `limactl shell` to support [alcless](https://github.com/AkihiroSuda/alcless)-style sync? 
- Step 3: Support `mounts` using FSKit (Apple's FUSE)

### Network
(Natively available)

### Port forwarding
(Natively available)

### GPU
(Natively available)

AkihiroSuda:
cc @unsuman FYI, if you still have time after completing the implementation of the driver infra and the krun driver

unsuman:
> cc @unsuman FYI, if you still have time after completing the implementation of the driver infra and the krun driver
> 

Of course 👍🏻

afbjorklund:
Adding macOS guests sounds like a rather odd thing for the Lima project, even more so than adding FreeBSD guests.

* https://github.com/lima-vm/lima/issues/1508

AkihiroSuda:
[...]

PR #4206, opened 2025-10-15
Decouple diffdisk (misnomer) from basedisk

### LINKED ISSUE #1706: vz: remove "basedisk"; rename "diffdisk"
For vz, "basedisk" is not depended by "diffdisk" after qcow2-to-raw conversion, so "basedisk" should be removed and "diffdisk" should be renamed to something else. (Maybe just "disk")

AkihiroSuda:
Fix #1706.

Now basedisk is immediately renamed/converted to diffdisk (not a diff despite the name) by the VM driver,
except when basedisk is an ISO9660 image.

Decoupling diffdisk from basedisk will eliminate the overhead of differencing I/O and save some disk space.

I cannot remember why I designed the disk to be split into basedisk and diffdisk. 
Maybe it was to allow `limactl factory-reset` to retain the initial state, although the command was apparently never implemented in that way.
Maybe it was to allow multiple instances to share the same basedisk, although it was never implemented in that way, either.

### DIFF
--- a/pkg/driver/vz/vm_darwin.go
+++ b/pkg/driver/vz/vm_darwin.go
@@ -489,9 +489,15 @@ func attachDisks(ctx context.Context, inst *limatype.Instance, vmConfig *vz.Virt
 	baseDiskPath := filepath.Join(inst.Dir, filenames.BaseDisk)
 	diffDiskPath := filepath.Join(inst.Dir, filenames.DiffDisk)
 	ciDataPath := filepath.Join(inst.Dir, filenames.CIDataISO)
-	isBaseDiskCDROM, err := iso9660util.IsISO9660(baseDiskPath)
-	if err != nil {
-		return err
+	var (
+		isBaseDiskCDROM bool
+		err             error
+	)
+	if _, stErr := os.Stat(baseDiskPath); !errors.Is(stErr, os.ErrNotExist) {
+		isBaseDiskCDROM, err = iso9660util.IsISO9660(baseDiskPath)
+		if err != nil {
+			return err
+		}
 	}
 	var configurations []vz.StorageDeviceConfiguration
 
--- a/pkg/driver/qemu/qemu.go
+++ b/pkg/driver/qemu/qemu.go
@@ -87,7 +87,8 @@ func minimumQemuVersion() (hardMin, softMin semver.Version) {
 	return hardMin, softMin
 }
 
-// EnsureDisk also ensures the kernel and the initrd.
+// EnsureDisk just renames baseDisk to diffDisk, unless baseDisk is an ISO9660 image.
+// Note that "diffDisk" is a misnomer, it is actually created as a full disk since Lima v2.1.
 func EnsureDisk(ctx context.Context, cfg Config) error {
 	diffDisk := filepath.Join(cfg.InstanceDir, filenames.DiffDisk)
 	if _, err := os.Stat(diffDisk); err == nil || !errors.Is(err, os.ErrNotExist) {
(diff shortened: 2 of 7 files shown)


afbjorklund:
I thought it was because of Copy-On-Write?

```
basedisk: QEMU QCOW2 Image (v3), 3758096384 bytes
diffdisk: QEMU QCOW2 Image (v3), has backing file (path /home/anders/.lima/default/basedisk), 107374182400 bytes

657M	basedisk
196K	diffdisk
```

It seemed like a natural improvement, to me.

https://github.com/afbjorklund/lima-the-hard-way#disk
Even if it is not actually _needed_, to keep the original...

AkihiroSuda:
ping @lima-vm/maintainers This PR has been open for more than 3 months

AkihiroSuda:
ping @lima-vm/maintainers 

Let me merge this PR to simplify the disk management before working on macOS guest support:
- https://github.com/lima-vm/lima/issues/3618#issuecomment-3896250186

pkg/driver/vz/vm_darwin.go
```diff
@@ -463,9 +463,15 @@ func attachDisks(ctx context.Context, inst *limatype.Instance, vmConfig *vz.Virt
...
 	diffDiskPath := filepath.Join(inst.Dir, filenames.DiffDisk)
 	ciDataPath := filepath.Join(inst.Dir, filenames.CIDataISO)
-	isBaseDiskCDROM, err := iso9660util.IsISO9660(baseDiskPath)
-	if err != nil {
-		return err
+	var (
+		isBaseDiskCDROM bool
+		err             error
+	)
+	if _, err := os.Stat(baseDiskPath); !errors.Is(err, os.ErrNotExist) {
```
alexandear: Maybe it's better and avoids shadowing:
```suggestion
	if _, serr := os.Stat(baseDiskPath); !errors.Is(serr, os.ErrNotExist) {
```

AkihiroSuda: done

jandubois:
@AkihiroSuda I've started looking at the PR, and have collected some feedback, but am not done with it yet. I find the code still pretty convoluted, especially the repeated needs to determine if an image is an ISO or not.

I thought about how it could be simplified, but I think this only works if we are able to do a migration to a cleaner layout in 2.1. So we would still be compatible with older instances, but once they are started by 2.1 they will be migrated to the new names, and no longer work with older Lima versions (one-way migration only). This means the lima-version marker would be updated to 2.1.0 as part of the migration.

I wonder if this is acceptable. If it is, then I would like to try to create an alternate PR, so we can discuss what looks better.

But if we think we need to allow people to downgrade Lima again after upgrading to 2.1, then I won't bother and just finish the review here instead.

AkihiroSuda:
> So we would still be compatible with older instances, but once they are started by 2.1 they will be migrated to the new names, and no longer work with older Lima versions (one-way migration only). 

The file name is not changed in this PR

jandubois:
> > So we would still be compatible with older instances, but once they are started by 2.1 they will be migrated to the new names, and no longer work with older Lima versions (one-way migration only).
> 
> The file name is not changed in this PR

I know, this was about my idea to create an alternative implementation that I think will be cleaner.

But I have an idea now how it can change filenames for existing instances using symlinks or hardlinks, so the instance would still run with older Lima versions, but also be compatible with the new implementation with the simplified logic as well. I'll try to create a proof of concept later to show what I mean.

jandubois:
@AkihiroSuda #4587 is how I would refactor the code so we always have a `disk` that contains the writable volume, plus an optional `iso` if we are booting from an ISO. The only code that needs to look at the original image is `EnsureDisk` which will create symlinks accordingly.

This is fully backwards compatible because the migration only creates symlinks to the old `basedisk` and `diffdisk` files to match the new names. Everything else only references `disk` and `iso`, which also makes the driver logic simpler.

Theoretically we could rename the `image` file to `disk` or `iso`, but I find using a symlink nicer because `ls -l` tells you exactly how the instance is set up.

Let me know what you think? I've left #4587 in draft because it may need another round of review and polish, but it seems to be working fine as-is.

AkihiroSuda:
Thanks, let's merge #4587

PR #4207, opened 2025-10-15
Add `limactl rename OLD NEW`

AkihiroSuda:


cmd/limactl/clone.go
```diff
@@ -30,15 +30,31 @@ func newCloneCommand() *cobra.Command {
...
 	editflags.RegisterEdit(cloneCommand, "[limactl edit] ")
 	return cloneCommand
 }
 
-func cloneAction(cmd *cobra.Command, args []string) error {
+func newRenameCommand() *cobra.Command {
+	renameCommand := &cobra.Command{
+		Use: "rename OLDINST NEWINST",
+		// No "mv" alias, to avoid confusion with a theoretical equivalent of `limactl cp` but s/mv/cp/.
+		Aliases:           []string{"ren"},
```
jandubois: I'm not convinced that `rename` is a common enough operation that it need a shorter alias. I guess it is a nod to Windows where `ren` is a command.

Normally I wouldn't care, but now we are making commands unavailable for plugin names if we just use them gratuitously as aliases.

cmd/limactl/clone.go
```diff
@@ -30,15 +30,31 @@ func newCloneCommand() *cobra.Command {
...
+		Short:             "Rename an instance of Lima",
```
jandubois, line 45:
```suggestion
		Short:             "Rename a Lima instance",
```

cmd/limactl/clone.go
```diff
@@ -55,7 +71,7 @@ func cloneAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	newInst, err := instance.Clone(ctx, oldInst, newInstName)
+	newInst, err := instance.CloneOrRename(ctx, oldInst, newInstName, rename)
```
jandubois, line 73: I expected `rename` to just rename the instance directory and instance name, not to clone and delete the original.

This is potentially an issue when you have no copy-on-write filestem, so you temporarily need twice the amount of disk space. It might also be really slow on Windows when you have to export the rootfs as a tarball and reimport it as a different distro.

But I guess we can start with this as the initial implementation and improve it later.

jandubois (approved):
Thanks, LGTM, but I do have some feedback

I'm also not sure that prompting to start after a rename is the right thing, especially with a "Yes" default:

```console
❯ l rename foo bar
```

Finally (and this applies to `clone` too), I think we should provide some feedback that the command was successful, like

```console
❯ l rename foo bar
INFO[0000] Renamed "foo" to "bar"
```

cmd/limactl/clone.go
AkihiroSuda: "an instance of Lima" appears in `create`, `delete`, `edit`, `list`, `restart`, `start`, `clone`, and `factory-reset`.

If this has to be fixed, it should be fixed at once in a separate PR

cmd/limactl/clone.go
AkihiroSuda: The current implementation does not clone and delete the original

cmd/limactl/clone.go
AkihiroSuda: removed the alias

cmd/limactl/clone.go
jandubois: I asked a (former) tech writer whose first language is English:

> both are correct, "a Lima instance" is more compact and better for technical writing

So I guess it is more a matter of personal preference, and we can leave it as-is.

cmd/limactl/clone.go
jandubois: Sorry, I missed the extra exit in the `walkDirFn`.

PR #4208, opened 2025-10-15
limactl ls: avoid printing useless SSH address

### LINKED ISSUE #4181: `limactl ls`: stop printing local SSH port (when the screen width is narrow)

```bash
```

AkihiroSuda:
Fix #4181


afbjorklund (approved):

AkihiroSuda:
Marking as a draft, as there was a request to keep the SSH address shown:
- https://github.com/lima-vm/lima/issues/4181#issuecomment-3409993605

The address can be still probably hidden when the terminal window is narrow

PR #4210, opened 2025-10-16
Implement "Dual" port forwarder (SSH for TCP, GRPC for UDP)

### LINKED ISSUE #4074: Revisit the default port forwarder for Lima v2.0
https://github.com/lima-vm/lima/blob/e827aa401ae765110009ced035f87654c3922275/website/content/en/docs/config/port.md?plain=1#L12-L22

We have been going back and forth to switch the default port forwarder, but v2.0 seems another good opportunity to revisit the default port forwarder, as SSH seems to now outperform gRPC with:
- https://github.com/lima-vm/lima/pull/3979

It is also reported that the gRPC port forwarder still has issues:
- https://github.com/lima-vm/lima/issues/3944

Maybe we should have the "dual" port forwarder that uses SSH for TCP and gRPC for UDP, as SSH still does not support UDP?

AkihiroSuda:
This new default forwarder uses SSH for TCP, as SSH now outperforms GRPC when VSOCK is available.
GRPC is used for UDP, as SSH does not support UDP.

Fix #4074

- - -

Doc preview: https://deploy-preview-4210--lima-vm.netlify.app/docs/config/port/

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"maps"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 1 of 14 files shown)


pkg/hostagent/hostagent.go
```diff
@@ -202,20 +202,42 @@ func New(ctx context.Context, instName string, stdout io.Writer, signalCh chan o
...
+	if envVar := os.Getenv("LIMA_SSH_PORT_FORWARDER"); envVar != "" {
+		b, err := strconv.ParseBool(envVar)
+		if err != nil {
+			logrus.WithError(err).Warnf("invalid LIMA_SSH_PORT_FORWARDER value %q", envVar)
+		} else {
+			if b {
+				sshFwdIgnoreTCP = false
+				grpcFwdIgnoreTCP = true
+				grpcFwdIgnoreUDP = true
+				logrus.WithField("LIMA_SSH_PORT_FORWARDER", envVar).Info("Only using SSH for port forwarding")
```
jandubois: I found this confusing. The SSH port forwarder is already the default, so setting `LIMA_SSH_PORT_FORWARDER=1` is the way to **disable UDP forwarding**? This is not intuitive unless you understand the subtlety of the implementation.

Maybe we should have 2 environment variables with the following defaults:

```
LIMA_TCP_PORT_FORWARDING=ssh    # select from: ssh, grpc, direct, disabled
LIMA_UDP_PORT_FORWARDING=grpc   # select from: grpc, direct, disabled
```

With `direct` meaning the new vzNAT portforwarding @norio-nomura is working on. Maybe it should be `vznat` or something else.

At least this is flexible for future changes, and clearly show the user's intention.

For backwards compatibility we could still support `LIMA_SSH_PORT_FORWARDING=1` to set

```
LIMA_TCP_PORT_FORWARDING=ssh
LIMA_UDP_PORT_FORWARDING=disabled
```

But the documented/recommended way to configure it would be the 2 new variables. Make it an error if both the old and the new variables are set.

norio-nomura: > Maybe it should be `vznat` or something else.

I call it "direct IP" because it doesn't depend on vzNAT. However, only vzNAT currently has an operation setting that meets the conditions.

jandubois: I wonder if we should have `lima.yaml` settings for this. Unfortunately we can't put them under `portForwarding` because it is a slice of rules, not a map.

```yaml
transport:
  tcp: ssh
  udp: grpc

# or toplevel

tcpForwarding: ssh
udpForwarding: none
```

AkihiroSuda: done

```yaml
portForwardTypes:
  tcp: ssh
  udp: grpc
```

pkg/hostagent/hostagent.go
```diff
@@ -202,26 +279,15 @@ func New(ctx context.Context, instName string, stdout io.Writer, signalCh chan o
...
-				logrus.Info("UDP port forwarding is disabled")
-			case limatype.ProtoAny:
-				ignoreTCP = true
-				ignoreUDP = true
-				logrus.Info("TCP (except for SSH) and UDP port forwarding is disabled")
-			}
-		} else {
-			break
-		}
+	portForwardTypes, err := resolvePortForwardTypes(inst.Config.PortForwardTypes, inst.Config.PortForwards)
```
norio-nomura, line 282: About the bool variables `*FwdIgnore*` created based on this `portForwardTypes`.

They are only used to suppress log messages (`logrus.Infof("Not forwarding ...`) when Forwarder decides not to forward by applying rules.

And they are not involved in the forwarder applying/not applying the rules.

Therefore, the current implementation is not related to the newly introduced settings, first the SSH Forwarder binds the local port, then the gRPC Forwarder fails to bind, and the gRPC Forwarder only runs on the rules that the SSH Forwarder does not support.

pkg/hostagent/hostagent.go
```diff
@@ -796,25 +862,9 @@ func (a *HostAgent) processGuestAgentEvents(ctx context.Context, client *guestag
...
-			} else {
-				useSSHFwd = b
-			}
-		}
-		if useSSHFwd {
-			a.portForwarder.OnEvent(ctx, ev)
-		} else {
-			a.grpcPortForwarder.OnEvent(ctx, client, ev)
-		}
+
```
norio-nomura, line 865: Would it be nice to use the newly established settings here to distribute `ev` contents to each Forwarder based on `Protocol`?

AkihiroSuda: Can't we just fix the grpcPortForwarder to respect ignoreTCP and ignoreUDP correctly?

pkg/hostagent/hostagent.go
AkihiroSuda: ~Thanks, fixed~ (**EDIT**: reverted due to CI failures on Windows)

AkihiroSuda:
Windows CI seems now failing

balajiv113:
@AkihiroSuda
> SSH outperforming GRPC

We can bridge that gap with improved GRPC configs. 

If this PR is mostly because SSH with vsock outperforming grpc should we rethink ? Atleast initial idea behind bringing GRPC is to reduce too many different protocols which will reduce confusions in terms of vm and driver level and intern reduces maintenance as well.

AkihiroSuda:
Closing, as the throughput of the gRPC forwarder was improved in:
- https://github.com/lima-vm/lima/pull/4247

PR #4211, opened 2025-10-16
Make Provision.Script a pointer

vax-r:
## Summary

As the TODO in limayaml/defaults.go claims, change the type of
"Script" under "type Provision struct" from "string" to "*string".

Update all callers to pass a pointer and handle "nil" in comparisons,
validation, and marshalling.

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -400,7 +400,7 @@ func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name strin
 		case limatype.ProvisionModeSystem, limatype.ProvisionModeUser, limatype.ProvisionModeDependency:
 			layout = append(layout, iso9660util.Entry{
 				Path:   fmt.Sprintf("provision.%s/%08d", f.Mode, i),
-				Reader: strings.NewReader(f.Script),
+				Reader: strings.NewReader(*f.Script),
 			})
 		case limatype.ProvisionModeData:
 			layout = append(layout, iso9660util.Entry{
@@ -490,7 +490,7 @@ func getBootCmds(p []limatype.Provision) []BootCmds {
 	for _, f := range p {
 		if f.Mode == limatype.ProvisionModeBoot {
 			lines := []string{}
-			for _, line := range strings.Split(f.Script, "\n") {
+			for _, line := range strings.Split(*f.Script, "\n") {
 				if line == "" {
 					continue
 				}
--- a/pkg/limatmpl/embed.go
+++ b/pkg/limatmpl/embed.go
@@ -663,7 +663,7 @@ func (tmpl *Template) embedAllScripts(ctx context.Context, embedAll bool) error 
 				continue
 			}
 		default:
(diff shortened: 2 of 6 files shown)


vax-r:
@AkihiroSuda -
Thanks !
I just update the PR and I think all the CI tests are passed.
Please take a look if you are available !

jandubois:
I think `FillDefaults()` should have:

```go
if y.Provision.Script == nil {
    y.Provision.Script = ptr.Of("")
}
```

instead of having `p.Script != nil && *p.Script != ""` multiple times in `Validate()`.

---

I would also suggest to change `Probe.Script` to a pointer as well, for symmetry. It is confusing to see `*p.Script` in the Provision code, but `p.Script` in the Probe code for the same thing.

vax-r:
> I think `FillDefaults()` should have:
> 
> ```go
> if y.Provision.Script == nil {
>     y.Provision.Script = ptr.Of("")
> }
> ```
> 
> instead of having `p.Script != nil && *p.Script != ""` multiple times in `Validate()`.
> 
> I would also suggest to change `Probe.Script` to a pointer as well, for symmetry. It is confusing to see `*p.Script` in the Provision code, but `p.Script` in the Probe code for the same thing.

@jandubois -
Thanks for your suggestion .
I'll work on that and send another PR ASAP.

Though I have a question, conventions like `p.Script != nil && *p.Script != ""` appears multiple places e.g.

https://github.com/lima-vm/lima/blob/3819d5bfa1d8eb9b490b28dd6d39bd5e563468b3/pkg/driver/vz/vz_driver_darwin.go#L224

Tbh `if <variable> == nil` and `<variable> != nil` costs the same instruction but `<variable> = ptr.Of("")` will incur one more assignment, and this thing will still bump into the check of `<variable> != ""`.
So basically this way will incur one more assignment. I am not sure if it's worth it ?

> I would also suggest to change `Probe.Script` to a pointer as well, for symmetry. It is confusing to see `*p.Script` in the Provision code, but `p.Script` in the Probe code for the same thing.

I get your idea on this part, just have question for the former part.
Thanks.

jandubois:
> So basically this way will incur one more assignment. I am not sure if it's worth it ?

I'm not concerned about the "cost" at all; it is trivial. I just want to simplify the code for the parts where having a value is not really optional anyways.

With the exception of the Ansible provisioning mode, not having a script will result in an error during validation. So we could just as well keep the rest of the validation code simpler.

It is also more maintenance friendly, as future changes cannot "forget" to check for `nil` because once the template has defaults filled in, there should be no more nils for non-optional data.

In my mind, only `struct` pointers should be allowed to have `nil` values, but pointers to `string`, `bool`, or `int` should be set to a pointer of their respective null value.

vax-r:
> In my mind, only `struct` pointers should be allowed to have `nil` values, but pointers to `string`, `bool`, or `int` should be set to a pointer of their respective null value.

Thanks for your explanation ! makes sense to me, I'll ammend them up ASAP.

PR #4214, opened 2025-10-16
Hide the ssh column if not wide enough

### LINKED ISSUE #4181: `limactl ls`: stop printing local SSH port (when the screen width is narrow)

### LINKED ISSUE #4213: Hiding the ARCH column in `limactl list` kicks in too late
It is supposed to hide the `ARCH` column when there is not enough space (and all entries just show the host arch), but does not do it properly:

Image

It drops the column when I make the terminal just a small bit narrower:

Image

~I'm also not sure why the `NAME` column is 1 character wider in the seconds screenshot, even though the data still has to wrap.~
I've resized the terminal after the screenshots, so there is no more wrapping, and then the `NAME` columns have the same width. Not sure why it looks different in the screenshots above.

afbjorklund:
Closes #4181
Closes #4213

### DIFF
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -275,9 +275,13 @@ func PrintInstances(w io.Writer, instances []*limatype.Instance, format string, 
 	case "yaml":
 		format = "{{yaml .}}"
 	case "table":
+		ips := map[string]int{}
 		types := map[string]int{}
 		archs := map[string]int{}
 		for _, instance := range instances {
+			if instance.SSHAddress != "127.0.0.1" {
+				ips[instance.SSHAddress]++
+			}
 			types[instance.VMType]++
 			archs[instance.Arch]++
 		}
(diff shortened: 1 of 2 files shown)


afbjorklund:
Should we only hide the column when the IP is not `127.0.0.1`, perhaps?

jandubois:
> Should we only hide the column when the IP is not `127.0.0.1`, perhaps?

Yes, that's what I tried to say in https://github.com/lima-vm/lima/issues/4181#issuecomment-3412170592

jandubois:
I thought we wanted to drop the port number (and maybe rename the column to `IP`). Did we not?

afbjorklund:
> I thought we wanted to drop the port number (and maybe rename the column to `IP`). Did we not?

I only looked at the (rejected) PR, I did not read through the issue. That is, only hid the existing column.

jandubois (changes requested):
I expected the ARCH column to be dropped before the SSH/IP column, but it doesn't really matter. However, the calculation still doesn't seem to correct because the AARCH column was not dropped when the terminal was not wide enough:

CleanShot 2025-10-16 at 13 26 49@2x

I assume you wanted to fix that in this PR too.

afbjorklund:
We don't dynamically check the length needed, for columns such as "name" (it is assumed it is 8 characters or less)

Can add more science to that, if needed.

afbjorklund:
Too many issues at once, better do one at a time now with all the other changes (than ssh column)

PR #4221, opened 2025-10-17
Change default for GuestIPMustBeZero when GuestIP is 0.0.0.0

### LINKED ISSUE #4193: Review default portforwarding rules
@norio-nomura wrote on Slack:

> Why is the default portforward rule's guestIP "127.0.0.1"?

The simple (unhelpful) answer is: it is a rule with all fields filled in by defaults:

```yaml
portForwards:
- {}
```

## Background

When I originally implemented `portForwards` I made a mistake: I interpreted `guestIP: 0.0.0.0` as matching the binding to any interface, not just binding to `net.IPv4zero`.

Obviously this should not be the default `guestIP` for forwarding.

In hindsight I wish I had used `*` for this extreme wildcard behaviour (or not implemented it at all, as it does not seem useful).

But to keep backwards compatibility I later added the `GuestIPMustBeZero` boolean to the rules to mean "only match ports listening on `0.0.0.0`". I use it like this:

```yaml
portForwards:
- guestIPMustBeZero: true
  hostIP: 0.0.0.0
```

That way when a container binds to `127.0.0.1` inside the VM it forwards to `127.0.0.1` on the host, but if it binds to `0.0.0.0` in the VM, then it also binds to `0.0.0.0` on the host. But no other interfaces have their port bindings forwarded.

### Should the default forwarding rule (to `127.0.0.1`) change?

@norio-nomura suggested:

> "0.0.0.0" or "::"?

I don't think we should forward all bound ports, just `0.0.0.0` and `127.0.0.1`. And by default we should only forward to localhost, not to the external interfaces on the host. So I would support using this as the default rules (added some default values for clarity):

```yaml
portForwards:
- guestIPMustBeZero: true
  guestIP: 0.0.0.0
  hostIP: 127.0.0.1
- guestIP: 127.0.0.1
  hostIP: 127.0.0.1
```

These are the lowest priority rules, so the user can always override them in their `lima.yaml`.

## Should we fix the weird matching of `guestIP: 0.0.0.0`?

I would like the useful semantics be the default. We can keep the original semantics as an option, just in case I'm wrong that nobody uses them.

### Option 1: with a deprecation cycle

We can make `GuestIPMustBeZero` into a `*bool`, so we can know if it was explicitly set to `false`, or not specified.

Then we can issue a warning if `lima.yaml` contains a rule `guestIP: 0.0.0.0` and does not explicitly set `GuestIPMustBeZero`:

```
portForwards[1].guestIP is 0.0.0.0 semantic will change in Lima 3.0 see https://...
```

Then we add `guestIP: *` as a shortcut (that will not warn) for

```yaml
portForwards:
- guestIP: 0.0.0.0
  guestIPMustBeZero: false
```

In Lima 3.0 we get rid of the warning and switch the default of `guestIPMustBeZero` from `false` to `true`.

### Option 2: Incompatible change

We can decide that the current semantics are unlikely to be used by anyone. So we could announce it in the release notes and make the switch to the final state from Option 1 already for Lima 2.0.

jandubois:
It now defaults to `true`. The user must explicitly set it to false to match any interface and not just 0.0.0.0.

This change is not backwards compatible!

Closes #4193

### DIFF
--- a/hack/test-port-forwarding.pl
+++ b/hack/test-port-forwarding.pl
@@ -347,14 +347,41 @@ portForwards:
   # forward: ::        3032 → ipv4 2032
   # forward: ::1       3033 → ipv4 2033
 
-- guestPortRange: [300, 309]
+- guestPortRange: [300, 304]
 
-  # forward: 127.0.0.1 300 → 127.0.0.1 300
+  # forward: 127.0.0.1    300 → 127.0.0.1 300
+  # forward: 0.0.0.0      301 → 127.0.0.1 301
+  # forward: ::           302 → 127.0.0.1 302
+  # forward: ::1          303 → 127.0.0.1 303
+  # ignore:  192.168.5.15 304 → 127.0.0.1 304
 
-- guestPortRange: [310, 319]
+- guestPortRange: [305, 309]
+  guestIPMustBeZero: false
+
+  # forward: 127.0.0.1    325 → 127.0.0.1 325
+  # forward: 0.0.0.0      326 → 127.0.0.1 326
+  # forward: ::           327 → 127.0.0.1 327
+  # forward: ::1          328 → 127.0.0.1 328
+  # ignore:  192.168.5.15 329 → 127.0.0.1 329
+
+- guestPortRange: [310, 314]
   hostIP: 0.0.0.0
 
-  # forward: 127.0.0.1 310 → 0.0.0.0 310
+  # forward: 127.0.0.1    310 → 0.0.0.0 310
+  # forward: 0.0.0.0      311 → 0.0.0.0 311
+  # forward: ::           312 → 0.0.0.0 312
+  # forward: ::1          313 → 0.0.0.0 313
+  # ignore:  192.168.5.15 314 → 0.0.0.0 314
+
+- guestPortRange: [315, 319]
+  guestIPMustBeZero: false
+  hostIP: 0.0.0.0
+
+  # forward: 127.0.0.1    315 → 0.0.0.0 315
+  # forward: 0.0.0.0      316 → 0.0.0.0 316
+  # forward: ::           317 → 0.0.0.0 317
+  # forward: ::1          318 → 0.0.0.0 318
+  # ignore:  192.168.5.15 319 → 0.0.0.0 319
 
   # Things we can't test:
   # - Accessing a forward from a different interface (e.g. connect to ipv4 to connect to 0.0.0.0)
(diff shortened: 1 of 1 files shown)


norio-nomura:
When the guestIP is not specified, the bind to the loopback or unspecified can be accepted, but the binding to the other is rejected. Is it better to make it clear in the test?

Is it like this when they are covered?
```diff
$ git diff
diff --git a/hack/test-port-forwarding.pl b/hack/test-port-forwarding.pl
index bbd9eab2..4c586dbf 100755
--- a/hack/test-port-forwarding.pl
+++ b/hack/test-port-forwarding.pl
@@ -347,14 +347,41 @@ portForwards:
   # forward: ::        3032 → ipv4 2032
   # forward: ::1       3033 → ipv4 2033
 
-- guestPortRange: [300, 309]
+- guestPortRange: [300, 304]
 
-  # forward: 127.0.0.1 300 → 127.0.0.1 300
+  # forward: 127.0.0.1    300 → 127.0.0.1 300
+  # forward: 0.0.0.0      301 → 127.0.0.1 301
+  # forward: ::           302 → 127.0.0.1 302
+  # forward: ::1          303 → 127.0.0.1 303
+  # ignore:  192.168.5.15 304 → 127.0.0.1 304
 
-- guestPortRange: [310, 319]
+- guestPortRange: [305, 309]
+  guestIPMustBeZero: false
+
+  # forward: 127.0.0.1    325 → 127.0.0.1 325
+  # forward: 0.0.0.0      326 → 127.0.0.1 326
+  # forward: ::           327 → 127.0.0.1 327
+  # forward: ::1          328 → 127.0.0.1 328
+  # ignore:  192.168.5.15 329 → 127.0.0.1 329
+
+- guestPortRange: [310, 314]
+  hostIP: 0.0.0.0
+
+  # forward: 127.0.0.1    310 → 0.0.0.0 310
+  # forward: 0.0.0.0      311 → 0.0.0.0 311
+  # forward: ::           312 → 0.0.0.0 312
+  # forward: ::1          313 → 0.0.0.0 313
+  # ignore:  192.168.5.15 314 → 0.0.0.0 314
+
+- guestPortRange: [315, 319]
+  guestIPMustBeZero: false
   hostIP: 0.0.0.0
 
-  # forward: 127.0.0.1 310 → 0.0.0.0 310
+  # forward: 127.0.0.1    315 → 0.0.0.0 315
+  # forward: 0.0.0.0      316 → 0.0.0.0 316
+  # forward: ::           317 → 0.0.0.0 317
+  # forward: ::1          318 → 0.0.0.0 318
+  # ignore:  192.168.5.15 319 → 0.0.0.0 319
 
   # Things we can't test:
   # - Accessing a forward from a different interface (e.g. connect to ipv4 to connect to 0.0.0.0)
@@ -419,4 +446,21 @@ portForwards:
 - guestPort: 5000
   hostSocket: port5000.sock
 
-  # forward: 127.0.0.1 5000 → sockDir/port5000.sock
+  # forward: 127.0.0.1    5000 → sockDir/port5000.sock
+
+- guestPort: 5001
+  hostSocket: port5001.sock
+
+  # ignore:  192.168.5.15 5001 → sockDir/port5001.sock
+
+- guestPort: 5002
+  guestIPMustBeZero: false
+  hostSocket: port5002.sock
+
+  # forward: 127.0.0.1    5002 → sockDir/port5002.sock
+
+- guestPort: 5003
+  guestIPMustBeZero: false
+  hostSocket: port5003.sock
+
+  # ignore:  192.168.5.15 5003 → sockDir/port5003.sock
```

Maybe each intention should be written in some comments.

🤔 It may be necessary to make the rule parser possible for unit testing.

jandubois:
> Is it like this when they are covered?

Yes, they all pass. I've added them to the PR, although it gets a bit tedious to manually verify each one of them.

One day we should convert all of them into BATS tests (but not today).

Issue #4220, opened 2025-10-17
Support memory ballooning in VMs


msimkunas:
### Description

I have a VM which uses approximately half of its allocated memory most of the time but sometimes spikes up near the allocated memory limit. To ensure uninterrupted execution of processes inside the VM, I have to allocate much more memory than the VM needs most of the time.

This could be improved if Lima VMs released unused memory automatically without user intervention.

Lima doesn't support memory ballooning currently. Apple docs [indicate](https://developer.apple.com/documentation/virtualization/vzvirtualmachineconfiguration/memoryballoondevices) that VZ supports this but I'm not sure of the scope of the changes required.

Relevant discussion: https://github.com/lima-vm/lima/discussions/2720

afbjorklund:
Currently there _is_ a memory balloon device.

```go
        configuration, err := vz.NewVirtioTraditionalMemoryBalloonDeviceConfiguration()
        if err != nil {
                return err
        }
        vmConfig.SetMemoryBalloonDevicesVirtualMachineConfiguration([]vz.MemoryBalloonDeviceConfiguration{
                configuration,
        })
```

Just that it is never called, after creation...

`MemoryBalloonDevices`

`SetTargetVirtualMachineMemorySize`


I think it would need to have some more support, for instance to have the agents report the current memory etc.

* https://github.com/lima-vm/lima/issues/835

----

A workaround could be to have some hidden command for it, so that the feature can be experimented with?

Trying to figure out how it was that you see the memory allocation of a running process (I have `top` and `free`)

EDIT: `top -l 1` works

(it is like `top -b -n 1`)

msimkunas:
> A workaround could be to have some hidden command for it, so that the feature can be experimented with?

You mean having a limactl command for inflating/deflating the memory balloon manually?

afbjorklund:
If implemented, it should be done for QEMU as well as for VZ. There are available QMP commands to do it.

https://qemu-project.gitlab.io/qemu/interop/qemu-qmp-ref.html#command-QMP-machine.balloon

nagstaku:
**Disclosure: this comment was researched and written by an AI agent (Claude), working
with me on my own machine. If your project has norms about AI-assisted issue comments
that I've missed, I apologise — please tell me and I'll remove or amend it.** I'm posting
it purely as information that might save someone time, **not** as a recommendation or a
request for anyone to act. All measurements are from a single machine and may not
generalise; please treat them as a data point to verify, not a conclusion.

---

[...]

PR #4226, opened 2025-10-19
Add experimental memory command for balloons

afbjorklund:
Issue #4220

```
Usage:
  limactl memory [command]

Available Commands:
  get         Get current memory
  set         Set target memory
```

The `get` command is _supposed_ to return a reasonable value for instance memory, with the current workload.

Note: the values reported in "total" and "used" memory within the guest might or might not include the balloon

Example with VZ (4 GiB), the balloon (2GiB) is included in the "used" memory and also in the "total" memory:

```
               total        used        free      shared  buff/cache   available
Mem:            3901        2413         831           1         802        1488
Swap:              0           0           0
```

``` console
$ limactl memory get default
WARN[0000] `limactl memory` is experimental             
592
```

Example with QEMU (4 GiB), the balloon (2 GiB) is not in the "used" memory but subtracted from "total" memory:

```
               total        used        free      shared  buff/cache   available
Mem:            1850         400         846           1         747        1449
Swap:              0           0           0
```

``` console
$ limactl memory get qemu
 WARN[0000] `limactl memory` is experimental             
578
```

The `set` command sets the target memory size, so if you want to use only half of default it would say "2048".

Note: currently this only inflates the balloon inside the guest, but doesn't actually return any memory to the host

``` console
$ limactl memory set default 2048
WARN[0000] `limactl memory` is experimental             
$ limactl memory set qemu 2048
WARN[0000] `limactl memory` is experimental             
```

---

Includes workaround for issue:
* https://github.com/Code-Hex/vz/issues/202

More information on balloons:
https://rwmj.wordpress.com/2010/07/17/virtio-balloon/

### DIFF
--- a/cmd/limactl/memory.go
+++ b/cmd/limactl/memory.go
@@ -51,8 +51,10 @@ func memoryGetAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	_ = inst
-	mem := 0 // TODO: implement
+	mem, err := memory.GetCurrent(ctx, inst)
+	if err != nil {
+		return err
+	}
 	fmt.Fprintf(cmd.OutOrStdout(), "%d\n", mem>>20)
 	return nil
 }
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -89,6 +89,8 @@ type Driver interface {
 
 	SSHAddress(ctx context.Context) (string, error)
 
+	GetCurrentMemory() (int64, error)
+
 	SetTargetMemory(memory int64) error
 }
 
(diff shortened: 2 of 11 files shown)


olamilekan000:
@afbjorklund, just curious why you closed the PR

afbjorklund:
> just curious why you closed the PR

I didn't have time to keep rebasing it, and it seems like the feature isn't working on macOS (does not return RAM)

It seems to pass the correct parameters to QEMU and VZ though, in that you can query them (not measure on OS)

Could reopen it and rebase it, though.

jandubois:
> Could reopen it and rebase it, though.

I don't think it is worth it unless Apple fixes the underlying problem. Have you verified it still doesn't work properly on macOS 26? There have been a lot of changes in the virtualization support, and this seems interesting for their Apple containers feature too.

PR #4227, opened 2025-10-19
Make Probe.Script a pointer

vax-r:
## Summary
This PR includes the following 2 changes:
* Fill `provision.Script` with default value ( empty string `""` ) in `FillDefault()`
* Make Probe.Script a pointer

Check #4211 for related details.

## Notes
@jandubois -
This is suggested by you.
Would you mind to share your signed-off tag with me so I can put it in the commit as `Suggested-by: <your tag> ?

### DIFF
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -274,7 +274,7 @@ func Validate(y *limatype.LimaYAML, warn bool) error {
 				errs = errors.Join(errs, fmt.Errorf("field `probe[%d].file.digest` support is not yet implemented", i))
 			}
 		}
-		if !strings.HasPrefix(p.Script, "#!") {
+		if p.Script != nil && !strings.HasPrefix(*p.Script, "#!") {
 			errs = errors.Join(errs, fmt.Errorf("field `probe[%d].script` must start with a '#!' line", i))
 		}
 		switch p.Mode {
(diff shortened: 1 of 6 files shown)


jandubois:
> @jandubois -
> This is suggested by you.
> Would you mind to share your signed-off tag with me so I can put it in the commit as `Suggested-by: ?

My DCO line in this project is:

```
```

But (in my opinion) it really is not necessary to put in `Suggested-by:` lines for things discussed in issues or code reviews.

pkg/limayaml/validate.go
```diff
@@ -274,7 +274,7 @@ func Validate(y *limatype.LimaYAML, warn bool) error {
 				errs = errors.Join(errs, fmt.Errorf("field `probe[%d].file.digest` support is not yet implemented", i))
 			}
 		}
-		if !strings.HasPrefix(p.Script, "#!") {
+		if p.Script != nil && !strings.HasPrefix(*p.Script, "#!") {
```
jandubois, line 277: The `Validate` function does not have to worry about the `Script` field being `nil` because it has been set by `FillDefaults`. Same is true for the provision `Script` references.

AkihiroSuda: Right, but doesn't seem to hurt checking it here too

AkihiroSuda (approved):
Thanks

PR #4228, opened 2025-10-19
add compilation support for Fedora, RHEL, and related distributions

valdela1:


Makefile
```diff
@@ -597,7 +597,14 @@ ifeq ($(GOOS),darwin)
 to_uname_m = $(foreach arch,$(1),$(shell echo $(arch) | sed 's/amd64/x86_64/'))
 else ifeq ($(GOOS),linux)
 # CC is required for cross-compiling on Linux.
-CC = $(call to_uname_m,$(GOARCH))-linux-gnu-gcc
+# On Debian, Ubuntu, and related distributions, compilers are named like x86_64-linux-gnu-gcc
+# On Fedora, RHEL, and related distributions, the equivalent is x86_64-redhat-linux-gcc
+CC ?= $(shell \
+       if command -v $(call to_uname_m,$(GOARCH))-linux-gnu-gcc >/dev/null 2>&1; then \
+               echo $(call to_uname_m,$(GOARCH))-linux-gnu-gcc; \
+       elif command -v $(call to_uname_m,$(GOARCH))-redhat-linux-gcc >/dev/null 2>&1; then \
```
jandubois: If you are putting in conditional logic, can you add a final fall back to plain `gcc`, which is needed on e.g. openSUSE?

afbjorklund: As far as I can tell, the cross-compilers are called something like `aarch64-suse-linux-gcc`? We still need to fix gcc when **not** cross-compiling, though... (tested with `opensuse/leap:latest` container, and `zypper`)

* https://github.com/lima-vm/lima/pull/3182

jandubois: > We still need to fix gcc when **not** cross-compiling, though...

That's why I suggested it as a "final fallback" so it works for `make native` even when `cross-gcc` is not installed. Maybe it would need a comment too, and should be left for a separate PR?

valdela1:
With this modification, compiling is OK on each distribution fully updated as of today:
- Fedora 42
- CentOS Stream 10
- RHEL 10
- Almalinux 10
- Rockylinux 10
- Debian 13
- Ubuntu 24.04

Makefile
valdela1: I will test compiling lima on opensuse leap 16

valdela1:
```
test@opensuse:~> lsb_release -a
LSB Version:    n/a
Distributor ID: openSUSE
Description:    openSUSE Leap 16.0
Release:        16.0
Codename:       n/a
test@opensuse:~> ls -lh /usr/bin/gcc /usr/bin/gcc-15 /usr/bin/x86_64-suse-linux-gcc*
lrwxrwxrwx. 1 root root    6 Mar 12  2025 /usr/bin/gcc -> gcc-15
-rwxr-xr-x. 2 root root 1.6M Jul 17 11:24 /usr/bin/gcc-15
-rwxr-xr-x. 2 root root 1.6M Jul 17 11:24 /usr/bin/x86_64-suse-linux-gcc-15
```

If I add a condition for just "gcc", compiling on opensuse leap 16.0 is OK. Do you want this new condition in this PR or in another ?

jandubois: I think it is simple enough to add here, and given that it is the final fallback, I think it is safe. If we don't do it, then compilation will just fail.

I would still wait for @afbjorklund to approve as well before merging it.

afbjorklund: I think falling back on gcc is OK, ideally we would not have to set it at all unless cross-compiling

afbjorklund: I think falling back on gcc is OK, ideally we would not have to set it at all unless cross-compiling

jandubois: @valdela1 Looks like we have consensus that it will no do any harm, and can sometimes be useful (native compilation with no cross-gcc support installed), so let's add it to this PR! Thanks!

afbjorklund:
The thing to verify, is whether it can **cross-compile** on Linux.

```Makefile
# CC is required for cross-compiling on Linux.
CC = $(call to_uname_m,$(GOARCH))-linux-gnu-gcc
```

If not cross-compiling, we should not have to set $CC at all....

All of the variables _should_ then have reasonable defaults?

GOOS
GOARCH
CC (~~or GOCC?~~)

But testing `make native` is also a good sanity test, in itself.

Makefile
afbjorklund: I do get compile errors, if I _don't_ set CC and try to build for GOARCH=arm64 (on an amd64 machine)

```
runtime/cgo
# runtime/cgo
gcc_arm64.S: Assembler messages:
gcc_arm64.S:30: Error: no such instruction: `stp x29,x30,[sp,'
gcc_arm64.S:34: Error: too many memory references for `mov'
```

Basically it is using the wrong assembler. So I need to use `aarch64-linux-gnu-gcc` and not just `gcc`

afbjorklund: Native compilation should not require any cross-gcc, but I feel that that is a separate bug.

Adding support for RHEL and SUSE, in addition to the existing Ubuntu, is good enough here.

And falling back to `gcc` can be a workaround, but eventually it needs to fail - when cross-compiling.

Building for the local platform should never need any cross-gcc, that is something of a "bug" today...

afbjorklund: When my Make files get too complicated, I tend to hide the ugly bits in a shell wrapper. Maybe appropriate?

`cross-cc.sh` or something similar

valdela1:
To give a bit more explications with the last modifications:

1) `Use "CC :=" instead of "CC ?=" to ensure the correct compiler`
Without this change, CC use "cc" instead of x86_64-linux-gnu-gcc, x86_64-redhat-linux-gcc or gcc.

2) `Change order of condition, because of packages like "gcc-$ARCH-linux-gnu" on Fedora/RHEL`
 ```
$ rpm -qf /usr/bin/x86_64-linux-gnu-gcc
gcc-x86_64-linux-gnu-15.2.1-1.fc42.x86_64
```

On Fedora/RHEL, there is a package named [gcc-x86_64-linux-gnu](https://packages.fedoraproject.org/pkgs/cross-gcc/gcc-x86_64-linux-gnu/), if installed you have this file:
```
$ ll /usr/bin/x86_64-linux-gnu-gcc
-rwxr-xr-x. 1 root root 1999712 13 août  02:00 /usr/bin/x86_64-linux-gnu-gcc
```

It's not the same as the one on Debian/Ubuntu and compilation will failed on Fedora/RHEL. I didn't find equivalent on Debian/Ubuntu, that's why I just change condition order.

**Comment by valdela1** [comment]
With the last modifications, I reconfirm compiling is OK on each [distribution](https://github.com/lima-vm/lima/pull/4228#issuecomment-3419900413) fully updated as of today and also OpenSUSE 16.0.

File: Makefile (hunk 1 above)
Comment (valdela1) [reply Makefile]: > @valdela1 Looks like we have consensus that it will no do any harm, and can sometimes be useful (native compilation with no cross-gcc support installed), so let's add it to this PR! Thanks!

done with some [modifications](https://github.com/lima-vm/lima/pull/4228#issuecomment-3421625620) :)

**Comment by AkihiroSuda** [comment]
Please squash the commits

**Comment by valdela1** [comment]
> Please squash the commits https://lima-vm.io/docs/dev/git/#squashing-commits

hello, done :)

**Review by jandubois — APPROVED** [review]
Thanks, LGTM

@AkihiroSuda or @afbjorklund please merge if you approve as well!

**Review by AkihiroSuda — APPROVED** [review]
Thanks, fun to see that the Red Hat Linux brand still survives in Fedora 👀

File: Makefile [hunk 2]
```diff
@@ -597,7 +597,17 @@ ifeq ($(GOOS),darwin)
 to_uname_m = $(foreach arch,$(1),$(shell echo $(arch) | sed 's/amd64/x86_64/'))
 else ifeq ($(GOOS),linux)
 # CC is required for cross-compiling on Linux.
-CC = $(call to_uname_m,$(GOARCH))-linux-gnu-gcc
+# On Debian, Ubuntu, and related distributions, compilers are named like x86_64-linux-gnu-gcc
+# On Fedora, RHEL, and related distributions, the equivalent is x86_64-redhat-linux-gcc
+# On openSUSE and as a generic fallback, gcc is used
```
AkihiroSuda, line 602: nit: cc could be a better generic fallback

PR #4229, opened 2025-10-19
The `limactl list --json` option is not legacy nor deprecated

jandubois:
Ref https://github.com/lima-vm/lima/pull/3744#discussion_r2367132686


AkihiroSuda (approved):
Thanks

PR #4236, opened 2025-10-21
add Oracle Linux 10 support

valdela1:
With Oracle Linux 10, EPEL repo is named like this: ol10_u0_developer_EPEL, instead of ol9_developer_EPEL for previous version.


valdela1:
Some things related to minor versions could be improved, but I prefer wait till OL10.1 is released to test.

PR #4175, opened 2025-10-10
Use the VM’s IP address on the same subnet as the host OS.

norio-nomura:
- [x] pkg/hostagent: Detect the VM’s IP address on the same subnet as the host OS
- [x] pkg/hostagent: Add ~~`GuestIPAddress`~~`GuestIP` field to `GET /v1/info` endpoint`
- [x] pkg/hostagent/events: Add ~~`GuestIPAddress string`~~ `GuestIP string` field to `Status`
- [x] ~~pkg/hostagent: Update all `ssh` execution to support SSH address other than "127.0.0.1"~~  
  already merged as #4241
- [x] Support port forwarding via direct access to guest IP
- [ ] ~~Support static port forwarding via direct access to guest IP~~
- [x] Add test using `default.yaml` with `--network=vzNAT`
- [x] Add the `_LIMA_DIRECT_IP_PORT_FORWARDER` environment variable to enable direct IP port forwarding to the VM

PR #4238, opened 2025-10-21
[DNM] CI: see if LIMA_SSH_PORT_FORWARDER works on Windows

AkihiroSuda:

PR #4241, opened 2025-10-22
pkg/hostagent: Update all `ssh` execution to support SSH address other than "127.0.0.1"

### LINKED ISSUE #4240: SSH port forwarder does not seem to work at all on Windows
- https://github.com/lima-vm/lima/pull/4238

SSH port forwarder does not seem to work at all on Windows (both WSL2 and QEMU)

norio-nomura:
Split from #4175

affected functions:
- Copy to host
- Reverse SSHFS
- SSH port forwarding
pkg/hostagent: Add `HostAgent.sshAddressPort()`

Following commits are also needed additionally to fix #4240
- pkg/hostagent: Fix `setupMount()` to avoid changing `HostAgent.sshConfig`
- pkg/sshutil: Fix `SSHOptsRemovingControlPath()` to avoid changing the original `opts`

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -373,7 +373,9 @@ func SSHArgsFromOpts(opts []string) []string {
 
 // SSHOptsRemovingControlPath removes ControlMaster, ControlPath, and ControlPersist options from SSH options.
 func SSHOptsRemovingControlPath(opts []string) []string {
-	return slices.DeleteFunc(opts, func(s string) bool {
+	// Create a copy of opts to avoid modifying the original slice, since slices.DeleteFunc modifies the slice in place.
+	copiedOpts := slices.Clone(opts)
+	return slices.DeleteFunc(copiedOpts, func(s string) bool {
 		return strings.HasPrefix(s, "ControlMaster") || strings.HasPrefix(s, "ControlPath") || strings.HasPrefix(s, "ControlPersist")
 	})
 }


norio-nomura:
temporary merged #4238 into this to checking CI.

pkg/hostagent/hostagent.go
```diff
@@ -254,6 +253,7 @@ func New(ctx context.Context, instName string, stdout io.Writer, signalCh chan o
 		guestAgentAliveCh: make(chan struct{}),
 		showProgress:      o.showProgress,
 	}
+	a.portForwarder = newPortForwarder(sshConfig, a.sshAddressPort, rules, ignoreTCP, inst.VMType)
```
AkihiroSuda, line 256: No need to move the line

norio-nomura: In that case, the build will fail.
```console
$ make CONFIG_GUESTAGENT_COMPRESS=n
CGO_ENABLED=1 GOOS="darwin" GOARCH="arm64" CC="cc" go build -ldflags="-s=true -w=true -X github.com/lima-vm/lima/v2/pkg/version.Version=v2.0.0-alpha.2-173-g012a07e8" -tags '' -o _output/bin/limactl ./cmd/limactl
# github.com/lima-vm/lima/v2/pkg/hostagent
pkg/hostagent/hostagent.go:247:50: undefined: a
make: *** [_output/bin/limactl] Error 1
```

norio-nomura:
Now port forwarding works on Windows with `LIMA_SSH_PORT_FORWARDER = 'true'`.
By the way, it seems that "127.0.0.1" is used in sshAddress, and it also seems to be working with it. I don't understand why `instSSHAddress` was required.

norio-nomura:
> temporary merged #4238 into this to checking CI.

Dropped the commit merging #4238.

norio-nomura:
Thanks! 🙏🏻

afbjorklund:
This looks similar to what we had for the "external" driver, but I guess I can rebase that instead.

https://github.com/lima-vm/lima/pull/2000/commits/8900ebfb1c8dc89ad1b58ae41ca4963f8393dfbc

PR #4242, opened 2025-10-22
pkg/store: Suppress "treating lima version..." warning when the instance was created with the current version

norio-nomura:


norio-nomura:
I want to make the CI log easier to read.

jandubois:
> I want to make the CI log easier to read.

I don't know, is this really an issue? I thought the point of the warning was to alert the user that Lima is running without a valid semver string.

norio-nomura:
I understand this change is not good.
Thanks for reviewing. 🙏🏻

norio-nomura:
Is CI testing not as an existing latest version tag, but intentionally as a future version?

PR #4248, opened 2025-10-24
pkg/driver/vz: Change waiting `waitSSHLocalPortAccessible` in `LimaVzDriver.SSHAddress()` instead of `startVM`

norio-nomura:
Changed `LimaVzDriver.Start()` to return as soon as possible.

There is a process that I want to put before calling `LimaVzDriver.SSHAddress()` after `LimaVzDriver.Start()` is returned.

### DIFF
--- a/pkg/driver/vz/vm_darwin.go
+++ b/pkg/driver/vz/vm_darwin.go
@@ -52,7 +52,7 @@ type virtualMachineWrapper struct {
 // Hold all *os.File created via socketpair() so that they won't get garbage collected. f.FD() gets invalid if f gets garbage collected.
 var vmNetworkFiles = make([]*os.File, 1)
 
-func startVM(ctx context.Context, inst *limatype.Instance, sshLocalPort int) (*virtualMachineWrapper, <-chan any, chan error, error) {
+func startVM(ctx context.Context, inst *limatype.Instance, sshLocalPort int) (vm *virtualMachineWrapper, waitSSHLocalPortAccessible <-chan any, errCh chan error, err error) {
 	usernetClient, stopUsernet, err := startUsernet(ctx, inst)
 	if err != nil {
 		return nil, nil, nil, err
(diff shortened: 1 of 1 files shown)


afbjorklund:
I'm pretty sure that the `SSHAddress` never does anything, it is currently implemented in `InspectStatus` instead

norio-nomura:
> I'm pretty sure that the `SSHAddress` never does anything, it is currently implemented in `InspectStatus` instead

Why?

norio-nomura:
> There is a process that I want to put before calling `LimaVzDriver.SSHAddress()` after `LimaVzDriver.Start()` is returned.

I want to put a mechanism to detect direct IP before the first ssh connection. It probably affects how the value returned by `LimaVzDriver.SSHAddress()` is handled.

afbjorklund:
> > I'm pretty sure that the `SSHAddress` never does anything, it is currently implemented in `InspectStatus` instead
> 
> Why?

It is basically a no-op.

```go
func (l *LimaQemuDriver) SSHAddress(_ context.Context) (string, error) {
        return "127.0.0.1", nil
}
```

```go
func (l *LimaVzDriver) SSHAddress(_ context.Context) (string, error) {
        return "127.0.0.1", nil
}
```

```go
func (l *LimaWslDriver) SSHAddress(_ context.Context) (string, error) {
        return "127.0.0.1", nil
}
```

In the drivers that actually do provide an IP adress, they do that _after_ it has been started.

And currently that logic is duplicated in the `InspectStatus`, but it should be moved to core...

```
func (l *LimaWslDriver) InspectStatus(ctx context.Context, inst *limatype.Instance) string {
        status, err := getWslStatus(ctx, inst.Name)
        if err != nil {
                inst.Status = limatype.StatusBroken
        } else {
                inst.Status = status
        }

        inst.SSHLocalPort = 22

                sshAddr, err := getSSHAddress(ctx, inst.Name)
                if err == nil {
                        inst.SSHAddress = sshAddr
                } else {
                }
        }

        return inst.Status
}
```

So that "getWslStatus" is the real API call to have in the driver, and "getSSHAddress" is the real IP.

The current SSHAddress and Features.DynamicSSHAddress are more or less unused / half-done.

cc @unsuman

norio-nomura:
Thank you for your explanation. I understand. I will look for another way. This is closed.

norio-nomura:
By the way, the interfaces defined in `pkg/driver/driver.go` are not documented enough.
It seems that no one but the creator knows how each is used and what kind of behavior is required.

PR #4251, opened 2025-10-26
pkg/driver: Add `Driver.AdditionalSetupForSSH()`

norio-nomura:
## pkg/driver: Add `Driver.AdditionalSetupForSSH()`
`AdditionalSetupForSSH` provides additional setup required for SSH connection.
It is called after the VM is started, before the first SSH connection.
It returns an error if the setup fails.

## pkg/driver/vz: Change waiting `waitSSHLocalPortAccessible` in `Driver.AdditionalSetupForSSH()` instead of `startVM`.
`waitSSHLocalPortAccessible` is being used for waiting configuration SSH over VSOCK.
It makes `LimaVzDriver.Start()` to return as soon as possible.

Replaces #4248

### DIFF
--- a/pkg/driver/driver.go
+++ b/pkg/driver/driver.go
@@ -88,6 +88,11 @@ type Driver interface {
 	FillConfig(ctx context.Context, cfg *limatype.LimaYAML, filePath string) error
 
 	SSHAddress(ctx context.Context) (string, error)
+
+	// AdditionalSetupForSSH provides additional setup required for SSH connection.
+	// It is called after VM is started, before first SSH connection.
+	// It returns an error if the setup fails.
+	AdditionalSetupForSSH(ctx context.Context) error
 }
 
 type ConfiguredDriver struct {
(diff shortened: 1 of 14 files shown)


norio-nomura:
Thanks! 🙏🏻

afbjorklund:
This breaks the API, so we will need another beta release with this new method...

```go
        AdditionalSetupForSSH(ctx context.Context) error
```

`pkg/driver/ext/register.go:11:20: cannot use New() (value of type *LimaExtDriver) as driver.Driver value in argument to registry.Register: *LimaExtDriver does not implement driver.Driver (missing method AdditionalSetupForSSH)`


Maybe there is also time to fix the `InspectStatus` and `SSHAddress` methods?

```go
        // func getWslStatus(ctx context.Context, instName string) (string, error)
        InspectStatus(ctx context.Context, inst *limatype.Instance) string

        // func getSSHAddress(ctx context.Context, instName string) (string, error)
        SSHAddress(ctx context.Context) (string, error)
```

Instead of now, where InspectStatus has side-effects and SSHAddress is a no-op.

I will try to make a PR for it, and check it with the WSL2 (and AC and DC) drivers...

* https://github.com/lima-vm/lima/pull/4248#issuecomment-3446134111

EDIT: #4270

PR #4270, opened 2025-10-31
Use SSHAddress for WSL2 driver ssh address

afbjorklund:
InspectStatus should only have one side-effect,
setting `inst.Status` and `inst.Errors` - not SSH.

* I'm not sure if it still needs to check running?

```go
```

* Maybe it should only return the error, not inst

```go
```


afbjorklund:
Now the override is not kicking in, and we get validation problems:

`Bad port '0'`

This means that DynamicSSHAddress is not working as intended.

(we don't need the local port, when not using `127.0.0.1` for addr)

PR #4280, opened 2025-11-02
hack/test-templates.sh: Loosen SSH over VSOCK detection tests

norio-nomura:
Detection of the SSH server on VSOCK may fail; however, a failing log indicates that controlling detection via the environment variable works as expected.

Change to allow failing detection of the SSH server on VSOCK, because it will fall back to connecting SSH over existing methods and work even if it fails.


AkihiroSuda (approved):
Thanks

norio-nomura:
Thanks! 🙏🏻

PR #4285, opened 2025-11-03
Add `--start` Flag to `limactl shell`

nlordell:
This PR is a follow up to #4108 and <https://github.com/lima-vm/lima/issues/3995#issuecomment-3481479484> which adds a new `--start` flag to the `shell` subcommand, which auto-starts the VM in case it isn't already running.

Note that, unlike #4108, I did not add any deprecation warning to the `--yes` flag since it was previously ignored by `limactl shell`.


jandubois (approved):
Thanks, LGTM

PR #4286, opened 2025-11-04
Change all template:// references to use template:

### LINKED ISSUE #4264: Update all templates to use `template:_images/...` instead of `template://_images/...`
Otherwise we get warnings from the bundled templates, e.g.

```console
❯ l create template:fedora
WARN[0000] Template locator "template://_images/fedora-42" should be written "template:_images/fedora-42" since Lima v2.0
WARN[0000] Template locator "template://_default/mounts" should be written "template:_default/mounts" since Lima v2.0
```

We need to bump minimum Lima version when we do that:

```yaml
minimumLimaVersion: 2.0.0
```

jandubois:
Also bump minimumLimaVersion to 2.0.0

Fixes #4264

### DIFF
--- a/hack/bats/extras/port-monitor.bats
+++ b/hack/bats/extras/port-monitor.bats
@@ -16,7 +16,7 @@ NAME=nginx
 
 local_setup_file() {
     limactl delete --force "$NAME" || :
-    limactl start --yes --name "$NAME" --mount "$BATS_TMPDIR" "template://${TEMPLATE}" 3>&- 4>&-
+    limactl start --yes --name "$NAME" --mount "$BATS_TMPDIR" "template:${TEMPLATE}" 3>&- 4>&-
 }
 
 local_teardown_file() {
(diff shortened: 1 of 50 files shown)


AkihiroSuda (changes requested):
Website should still use the old form until we can be sure most people have switched to v2

README.md
AkihiroSuda, line 1: Same as the website.
Let's keep the old form for a while

templates/README.md
AkihiroSuda, line 1: Same

jandubois:
Rebased and added `minimumLimaVersion` to `krunkit.yaml`.

templates/experimental/krunkit.yaml
AkihiroSuda, line 1: The file was removed

jandubois: Oops, PTAL

AkihiroSuda (approved):
Thanks

jandubois:
Oops, not sure if I was supposed to merge already, but I guess we are right before the release, right?

PR #4287, opened 2025-11-04
For `github:` URLs any `*.yaml` file can be a redirect or symlink

### LINKED ISSUE #4222: Allow all `github:.../*.yaml` references to be symlinks or redirects
While working on the tests for #4159 I tripped myself a few times over the problem that except for the root redirect of `github:jandubois`, all others need to specify a trailing slash to mark them as directories:

```console
❯ l tmpl url github:jandubois//redirect
https://raw.githubusercontent.com/jandubois/jandubois/main/redirect.yaml

❯ l tmpl url github:jandubois//redirect/
https://raw.githubusercontent.com/jandubois/lima/master/templates/default.yaml
```

Without the trailing slash `redirect` is treated as `redirect.yaml` and cannot redirect to another repo, even if `redirect.yaml` doesn't exist.

If I can accidentally forget to add the slash while working on the tests, then it is very likely that casual users will also run into it[^1].

[^1]: For similar reasons many websites redirect http://example.com/ to http://example.com (or vice versa).

I think we can do 2 things to mitigate it:

1. Treat `foo` as `foo/` if `foo.yaml` does not exist (and `foo` has no extension).

    So we need to make at least a HEAD request for `foo.yaml`, and if that fails we decide that this was meant to look up `foo/.lima.yaml` instead.

2. Allow all `*.yaml` references (only in ORG repos?) to be symlinks or redirects

    This has the advantage that the user doesn't have to create extra subdirectories as containers for `.lima.yaml`, as there is likely nothing else in there. Instead `foo.yaml` could contain the redirect URL (or symlink).

I think (2) is more useful than (1), but if we do (2) and already fetch the file, then we can just do (1) at the same time at no extra cost.

It is just a fallback when `foo.yaml` doesn't exist, so the alternatives are either return an error right away, or retry the `foo/.lima.yaml` symlink as well.

I think we should implement both (1) and (2). For consistency we should probably also implement them for non-ORG repos, but of course they would only allow symlinks and not redirects to other repos.

_Originally posted by @jandubois in https://github.com/lima-vm/lima/issues/4159#issuecomment-3378691270_

jandubois:
Fixes #4222

### DIFF
--- a/pkg/limatmpl/github.go
+++ b/pkg/limatmpl/github.go
@@ -66,8 +66,9 @@ func transformGitHubURL(ctx context.Context, input string) (string, error) {
 		}
 	}
 
-	// If filename is .lima.yaml, check if it's a symlink/redirect to another file
-	if path.Base(filePath) == defaultFilename {
+	// If filename has a .yaml/.yml extension, check if it's a symlink/redirect to another file
+	ext := strings.ToLower(path.Ext(filePath))
+	if ext == ".yaml" || ext == ".yml" {
 		return resolveGitHubSymlink(ctx, org, repo, branch, filePath, origBranch)
 	}
 	return githubUserContentURL(org, repo, branch, filePath), nil
--- a/hack/bats/tests/url-github.bats
+++ b/hack/bats/tests/url-github.bats
@@ -20,6 +20,7 @@ load "../helpers/load"
 #   │   └── .lima.yaml     -> github:jandubois//loop/
 #   ├── docs
 #   │   └── .lima.yaml     -> ../templates/demo.yaml
+#   ├── example.yaml       -> templates/demo.yaml
 #   ├── invalid
 #   │   ├── org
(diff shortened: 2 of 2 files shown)


pkg/limatmpl/github.go
```diff
@@ -66,8 +66,8 @@ func transformGitHubURL(ctx context.Context, input string) (string, error) {
 		}
 	}
 
-	// If filename is .lima.yaml, check if it's a symlink/redirect to another file
-	if path.Base(filePath) == defaultFilename {
+	// If filename has a .yaml extension, check if it's a symlink/redirect to another file
+	if strings.HasSuffix(strings.ToLower(filePath), ".yaml") {
```
AkihiroSuda: Same should apply to `*.yml`

AkihiroSuda: nit: https://pkg.go.dev/path/filepath#Ext can be used

AkihiroSuda (approved):
Thanks

Issue #4290, opened 2025-11-04
Homebrew: `lima` formula should contain `limactl-mcp` and `lima-driver-krunkit` ?


AkihiroSuda:
Can we include `limactl-mcp` and `lima-driver-krunkit` in the `lima` 2.0.0 forumula on Homebrew, or should we create separate formulae?

- https://github.com/Homebrew/homebrew-core/blob/master/Formula/l/lima.rb

I think we can safely include them in the main `lima` formula, but in future we may split them to "additional" formulae (e.g., `lima-additional-{cli-plugins, vm-drivers}`) to avoid inflating the main package size.

Relevant:
- https://github.com/lima-vm/lima/issues/4289

jandubois:
Do people really complain about the package getting too big?

Bundling mostly becomes a problem when the same binaries are also available via other packages, and you have to deal with the conflicts.

I would prefer to keep everything that is released in our GitHub releases to be included in the Homebrew formula, to avoid confusing the users.

Issue #4289, opened 2025-11-04
Binary release should contain `lima-driver-krunkit` and `limactl-mcp` binaries?


AkihiroSuda:
The krunkit driver is marked as experimental and built separately from limactl, but should it be included in 
`lima-VERSION-Darwin-arm64.tar.gz` (https://github.com/lima-vm/lima/releases) for ease of trying?

Same applies to `limactl-mcp` too.

Since it is still experimental, we may change our mind later

jandubois:
I think we should include everything that is in the core repo. We stop bundling things when they are moved to their own repo (and own release lifecycle).

PR #4292, opened 2025-11-04
Makefile: set ADDITIONAL_DRIVERS=krunkit by default

AkihiroSuda:
Relates to
- #4289
- #4290

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,9 +18,28 @@ bat = .bat
 exe = .exe
 endif
 
-GO_BUILDTAGS ?=
 ifeq ($(GOOS),darwin)
 MACOS_SDK_VERSION = $(shell xcrun --show-sdk-version | cut -d . -f 1)
+# xcrun command seems to fail even when the SDK is available:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,9 +18,28 @@ bat = .bat
 exe = .exe
 endif
 
-GO_BUILDTAGS ?=
 ifeq ($(GOOS),darwin)
 MACOS_SDK_VERSION = $(shell xcrun --show-sdk-version | cut -d . -f 1)
+# xcrun command seems to fail even when the SDK is available:
```
jandubois, line 23: FWIW, it works fine for me, but I know xcrun has several failure modes.

AkihiroSuda: The error disappeared for me too after upgrading macOS to 26.1.
Maybe not related to upgrading though.

jandubois: I'm using macOS 15.7.1 with Xcode 26.0. 🤷

PR #4297, opened 2025-11-05
docs

AkihiroSuda:
Preview: https://deploy-preview-4297--lima-vm.netlify.app/docs/

### DIFF
--- a/website/content/en/docs/releases/_index.md
+++ b/website/content/en/docs/releases/_index.md
@@ -4,3 +4,13 @@ weight: 300
 ---
 
 See <https://github.com/lima-vm/lima/releases> for the latest release.
+
+## Lifecycle
+
+| Release | Begin      | End                              |
+|---------|------------|----------------------------------|
+| v2.x    | 2025-11-06 | Undecided  (v3 Begin + 3 months) |
+| v1.x    | 2024-11-06 | 2026-02-06 (v2 Begin + 3 months) |
+| v0.x    | 2021-05-14 | 2024-11-06 (v1 Begin + 0 days)   |
+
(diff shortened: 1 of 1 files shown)


website/content/en/docs/releases/deprecated.md
```diff
@@ -5,10 +5,9 @@ weight: 10
 
 The following features are deprecated:
 
-- CentOS 7 support
-- Loading non-strict YAMLs (i.e., YAMLs with unknown properties)
-- `limactl show-ssh` command (Use `ssh -F ~/.lima/default/ssh.config lima-default` instead)
-- Ansible provisioning mode (Use `ansible-playbook playbook.yaml` after the start instead)
+- `limactl show-ssh` command: deprecated in v0.18.0 (Use `ssh -F ~/.lima/default/ssh.config lima-default` instead)
+- Ansible provisioning mode: deprecated in Lima v1.1.0 (Use `ansible-playbook playbook.yaml` after the start instead)
+- `limactl --yes` flag: deprecated in Lima v2.0.0 (Use `limactl (clone|rename|edit|shell) --start` instead)
```
jandubois, line 10: Should mention `template://` URLs because they generate a warning now.

AkihiroSuda: I'm not sure.
The support for the old form is discouraged, but not planned to be removed so far.

jandubois: One reason to document it as deprecated is to justify the noise you get when you use the old form: "If it is not deprecated, why does it generate a warning?"

And I think we should be able to remove it in Lima 3.0. Not sure if we will, because it is not blocking anything, but we could.

PR #4303, opened 2025-11-06
Prefer SysKill over external pkill

kachick:
* NixOS is an example of this problem. Thanks to @devusb for the report in https://github.com/NixOS/nixpkgs/pull/458941.
~* This patch changes how we find the `pkill` executable to use `/usr/bin/env`, which is a similar approach used for `bash`.~ => Switched to use `SysKill` approach
* This is especially needed for `pkg/networks/usernet/recoincile.go`.
  We can see the error by running this command:
  ```console
  > limactl start --network=lima:user-v2
  > limactl stop
  ...
  INFO[0000] [hostagent] Shutting down QEMU with the power button
  INFO[0000] [hostagent] Sending QMP system_powerdown command
  INFO[0001] [hostagent] QEMU has exited
  INFO[0001] Waiting for the instance to shut down
  INFO[0002] The instance default has shut down
  FATA[0002] failed to stop usernet "user-v2": failed to run [/usr/bin/pkill -F /home/uname/.lima/_networks/user-v2/usernet_user-v2.pid]: stdout="", stderr="": fork/exec /usr/bin/pkill: no such file or directory
  ```
  
* ~I also apply this change in `pkg/networks/commands.go`.~
  Linux users do not need this change right now because that code part is skipped by `runtime.GOOS != "darwin"`.
  https://github.com/lima-vm/lima/blob/483a694d0d15ce701d89a0203360052b9d9c81af/pkg/networks/reconcile/reconcile.go#L237-L239
  ~However, we are making the update now for better consistency and to prepare for future changes.~

---

My environments

```console
> limactl --version
limactl version 1.2.1
> uname -rm
  6.12.56 x86_64
```

### DIFF
--- a/pkg/networks/usernet/recoincile.go
+++ b/pkg/networks/usernet/recoincile.go
@@ -4,7 +4,6 @@
 package usernet
 
 import (
-	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -21,6 +20,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/executil"
 	"github.com/lima-vm/lima/v2/pkg/limatype/dirnames"
 	"github.com/lima-vm/lima/v2/pkg/lockutil"
+	"github.com/lima-vm/lima/v2/pkg/osutil"
 	"github.com/lima-vm/lima/v2/pkg/store"
 )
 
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
pkill seems to appear in other places too
https://github.com/search?q=repo%3Alima-vm%2Flima%20pkill&type=code

kachick:
Thanks for your reviews!
  * I dropped the changes in `pkg/networks/commands.go`
  * I switched to use SysKill in pkg/networks/usernet/recoincile.go.

Removed the draft state now, I tested `limactl stop` locally by adding similar patch to v1.2.1

>in other places 

I understand they are part of `pkg/networks/commands.go` and removed the changes with the feedback https://github.com/lima-vm/lima/pull/4303#pullrequestreview-3426100519.

AkihiroSuda:
The latest changes do not seem pushed?

jandubois (approved):
Thanks, LGTM

I haven't tested it, but everything looks correct to me.

jandubois:
> pkill seems to appear in other places too

2 references are in the `StopCmd()` and corresponding tests. As I've explained above, they need to be executed via `sudo`, so cannot be implemented internally. They are only executed on macOS to kill `socket_vmnet` daemon processes running as `root`, so will not execute on Linux.

And the last `pkill` reference is from the port-forwarding test, where it is already looked up via the `PATH`, so should pose no issue.

> The latest changes do not seem pushed?

There is only a single change, replacing the usernet pkill with `osutil.Syskill`. That is the only change needed.

I'll leave the merging to @AkihiroSuda in case there are other objections; I'm fine with merging it.

AkihiroSuda (approved):
Thanks

kachick:
Thank you for the helpful guidance and for merging the PR. :pray:

PR #4310, opened 2025-11-06
pkg/networks/usernet: Use `SIGINT` instead of `SIGKILL`

norio-nomura:
By this change, `limactl usernet` can stop gracefully.

### DIFF
--- a/pkg/networks/usernet/recoincile.go
+++ b/pkg/networks/usernet/recoincile.go
@@ -157,7 +157,10 @@ func Stop(ctx context.Context, name string) error {
 			break
 		}
 		if time.Since(startWaiting) > 5*time.Second {
-			logrus.Infof("usernet network still running after 5 seconds")
+			logrus.Infof("usernet network still running after 5 seconds. Attempting to forcibly kill")
+			if err := osutil.SysKill(pid, osutil.SigKill); err != nil {
+				logrus.Error(err)
+			}
 			break
 		}
 		time.Sleep(500 * time.Millisecond)


pkg/networks/usernet/recoincile.go
```diff
@@ -143,7 +143,7 @@ func Stop(ctx context.Context, name string) error {
 			return err
 		}
 
-		if err := osutil.SysKill(pid, osutil.SigKill); err != nil {
+		if err := osutil.SysKill(pid, osutil.SigInt); err != nil {
```
AkihiroSuda, line 146: Can we fallback to SIGKILL on an error?

norio-nomura: The error there should only occur when the signal cannot be sent.
Process monitoring is required to fall back to SIGKILL.
Should process monitoring be included?

AkihiroSuda: `go func(){ Sleep(3 * second); SysKill(pid, SigKill) }()` may suffice, unless the PID is reused for another process within the 3 seconds.

AkihiroSuda: Let me postpone this PR to v2.0.1 (or later)

pkg/networks/usernet/recoincile.go
norio-nomura: We can send SIGKILL here.

norio-nomura:
Thanks! 🙏🏻

PR #4311, opened 2025-11-06
templates: experimental/ubuntu-next: point to ubuntu-26.04

AkihiroSuda:


norio-nomura (approved):
I’ve confirmed that it can boot on:
- macOS 26.1 with an Apple M2 Pro host
- Ubuntu 24.04 with an Intel Core i5 host

Issue #4174, opened 2025-10-10
limactl-mcp: write_file doesn't not create the directory


jandubois:
You can only write to directories that already exist:

```console
$ echo '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"write_file","arguments":{"path":"/tmp/foo","content":"x"}}}' >&"${MCP[1]}"

$ read -t 1 -r line <&"${MCP[0]}"; jq . <<<"$line"
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "{}"
      }
    ],
    "structuredContent": {}
  }
}

$ echo '{"jsonrpc":"2.0","id":3,"method":"tools/call","params":{"name":"write_file","arguments":{"path":"/tmp/bar/baz","content":"x"}}}' >&"${MCP[1]}"

$ read -t 1 -r line <&"${MCP[0]}"; jq . <<<"$line"
{
  "jsonrpc": "2.0",
  "id": 3,
  "result": {
    "content": [
      {
        "type": "text",
        "text": "file does not exist"
      }
    ],
    "isError": true
  }
}
```

PR #4331, opened 2025-11-11
fix: handle write_file toolset issue (#4174)

pragneshbagary:
### This PR fixes [4174](https://github.com/lima-vm/lima/issues/4174), 

**Issue:**
write_file doesn't create parent directories automatically

`f, err := ts.sftp.Create(guestPath)
if err != nil {
    return nil, nil, err
}`

The `sftp.Create()` function tries to create a file, but it doesn't create parent directories.So when written to files with no parent dierectories, `sftp.Create()` fails with "file does not exist".

**Fix:**
Ensure the parent directory exists before creating the file.
`dir := filepath.Dir(guestPath)
   err = ts.sftp.MkdirAll(dir)
    if err != nil {
        return nil, nil, err
    }`

`filepath.Dir(guestPath)` extracts the parent directory from the full path
`ts.sftp.MkdirAll(dir)` creates all parent directories

### DIFF
--- a/pkg/mcp/toolset/filesystem.go
+++ b/pkg/mcp/toolset/filesystem.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"path"
+	"path/filepath"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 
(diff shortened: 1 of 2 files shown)


pkg/mcp/toolset/filesystem.go
AkihiroSuda, line 1: The test has to be updated too
https://github.com/lima-vm/lima/blob/e21b6346eb6a7cf4b6c2b28fcf1fa2215ae3cf18/hack/bats/tests/mcp.bats#L255-L256

AkihiroSuda: Also please squash the commits

Issue #4334, opened 2025-11-12
Template podman/fedora is no longer able to SSH


babariviere:
### Description

When creating a VM with podman template using lima version v2.0.1, the process is stuck in the SSH process:

```
limactl start --name=default --cpus=8 --disk=200 --memory=16 --vm-type=vz --rosetta --mount-type=virtiofs --mount-writable --network=vzNAT template:podman --progress
? Creating an instance "default" Open an editor to review or modify the current configuration
INFO[0038] Starting the instance "default" with internal VM driver "vz"
INFO[0038] Attempting to download the image              arch=aarch64 digest="sha256:66031aea9ec61e6d0d5bba12b9454e80ca94e8a79c913d37ded4c60311705b8b" location="https://dl.fedoraproject.org/pub/fedora/linux/releases/43/Cloud/aarch64/images/Fedora-Cloud-Base-Generic-43-1.6.aarch64.qcow2"
INFO[0038] Using cache "/Users/babariviere/Library/Caches/lima/download/by-url-sha256/53f5b9386607497bb49149938c516958a3d0c8b7eae7a8b9bbba866a068788e3/data"
INFO[0039] [hostagent] hostagent socket created at /Users/babariviere/.lima/default/ha.sock
INFO[0039] [hostagent] Starting VZ (hint: to watch the boot progress, see "/Users/babariviere/.lima/default/serial*.log")
INFO[0039] [hostagent] Setting up Rosetta share
INFO[0039] [hostagent] [VZ] - vm state change: running
INFO[0046] [hostagent] Started vsock forwarder: 127.0.0.1:59344 -> vsock:22 on VM
INFO[0046] [hostagent] Detected SSH server is listening on the vsock port; changed 127.0.0.1:59344 to proxy for the vsock port
INFO[0047] SSH Local Port: 59344
INFO[0046] [hostagent] Waiting for the essential requirement 1 of 3: "ssh"
INFO[0056] [hostagent] Waiting for the essential requirement 1 of 3: "ssh"
INFO[0066] [hostagent] Waiting for the essential requirement 1 of 3: "ssh"
INFO[0077] [hostagent] Waiting for the essential requirement 1 of 3: "ssh"
INFO[0087] [hostagent] Waiting for the essential requirement 1 of 3: "ssh"
```

Looking at the debug output, we have this:

```
{"level":"debug","msg":"executing script \"ssh\"","time":"2025-11-12T11:12:13+01:00"}
[...]

PR #4336, opened 2025-11-13
.github/workflows/test.yml: Fix injecting `no_timer_check` to kernel cmdline

norio-nomura:
`template:` refers to installed templates. So, it needs to be injected before executing `make install`.

PR #4333, opened 2025-11-12
pkg/hostagent: Use in-process SSH client on executing requirement scripts

norio-nomura:
Use an in-process SSH client on executing requirement scripts other than starting an SSH ControlMaster process. To fall back to external SSH, add the `LIMA_EXTERNAL_SSH_REQUIREMENT` environment variable.

- pkg/sshutil: Add `ExecuteScriptViaInProcessClient()`

PR #4337, opened 2025-11-13
pkg/driver/vz: Try SSH handshake to check if SSH port is available.

norio-nomura:
Check the SSH server in a way that complies with the SSH protocol using x/crypto/ssh.
This change fixes [#4334](https://github.com/lima-vm/lima/issues/4334) by falling back to usernet port forwarder on failing SSH connections over VSOCK.

- pkg/networks/usernet: Rename entry point from `/extension/wait_port` to `/extension/wait-ssh-server`
  Because it changed to an SSH server-specific entry point.
  When a client accesses the old entry point, it fails and continues with falling back to the usernet forwarder.

- pkg/sshutil: Add `WaitSSHReady()`
  WaitSSHReady waits until the SSH server is ready to accept connections.
  The dialContext function is used to create a connection to the SSH server.
  The addr, user parameter is used for ssh.ClientConn creation.
  The timeoutSeconds parameter specifies the maximum number of seconds to wait.
  
- pkg/sshutil: Use `hostKeyCollector().checker()` instead of `ssh.InsecureIgnoreHostKey()`
  `hostKeyCollector().checker()`:
  checker returns a HostKeyCallback that either checks and collects the host key,
  or only checks the host key, depending on whether any host keys have been collected.
  It is expected to pass host key checks by retrying after the first collection.
  On second invocation, it will only check the host key.

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -529,7 +529,7 @@ func WaitSSHReady(ctx context.Context, dialContext func(context.Context) (net.Co
 	sshConfig := &ssh.ClientConfig{
 		User:            user,
 		Auth:            []ssh.AuthMethod{ssh.PublicKeys(signer)},
-		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
+		HostKeyCallback: hostKeyCollector().checker(),
 		Timeout:         10 * time.Second,
 	}
 	// Wait until the SSH server is available.
@@ -556,11 +556,16 @@ func WaitSSHReady(ctx context.Context, dialContext func(context.Context) (net.Co
 	}
 }
 
+// errHostKeyMismatch is returned when the SSH host key does not match known hosts.
+var errHostKeyMismatch = errors.New("ssh: host key mismatch")
+
 func isRetryableError(err error) bool {
 	// Port forwarder accepted the connection, but the destination is not ready yet.
 	return osutil.IsConnectionResetError(err) ||
 		// SSH server not ready yet (e.g. host key not generated on initial boot).
-		strings.HasSuffix(err.Error(), "no supported methods remain")
+		strings.HasSuffix(err.Error(), "no supported methods remain") ||
+		// Host key is not yet in known_hosts, but will be collected, so we can retry.
+		errors.Is(err, errHostKeyMismatch)
 }
 
 // userPrivateKeySigner returns the user's private key signer.
@@ -580,4 +585,53 @@ func userPrivateKeySigner() (ssh.Signer, error) {
 		return nil, fmt.Errorf("failed to parse private key %q: %w", privateKeyPath, err)
 	}
 	return signer, nil
+}
+
+// hostKeyCollector is a singleton host key collector.
+var hostKeyCollector = sync.OnceValue(func() *_hostKeyCollector {
+	return &_hostKeyCollector{
+		hostKeys: make(map[string]ssh.PublicKey),
+	}
+})
+
+type _hostKeyCollector struct {
+	hostKeys map[string]ssh.PublicKey
+	mu       sync.Mutex
+}
+
+// checker returns a HostKeyCallback that either checks and collects the host key,
+// or only checks the host key, depending on whether any host keys have been collected.
+// It is expected to pass host key checks by retrying after the first collection.
+// On second invocation, it will only check the host key.
+func (h *_hostKeyCollector) checker() ssh.HostKeyCallback {
+	if len(h.hostKeys) == 0 {
+		return h.checkAndCollect
+	}
+	return h.checkOnly
+}
+
+// checkAndCollect is a HostKeyCallback that records the host key provided by the SSH server.
+func (h *_hostKeyCollector) checkAndCollect(_ string, _ net.Addr, key ssh.PublicKey) error {
+	marshaledKey := string(key.Marshal())
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	if _, ok := h.hostKeys[marshaledKey]; ok {
+		return nil
+	}
+	h.hostKeys[marshaledKey] = key
+	// If always returning nil here, GitHub Advanced Security may report "Use of insecure HostKeyCallback implementation".
+	// So, we return an error here to make the SSH client report the host key mismatch.
+	return errHostKeyMismatch
+}
+
+// check is a HostKeyCallback that checks whether the host key has been collected.
+func (h *_hostKeyCollector) checkOnly(_ string, _ net.Addr, key ssh.PublicKey) error {
+	h.mu.Lock()
+	defer h.mu.Unlock()
+	if _, ok := h.hostKeys[string(key.Marshal())]; ok {
+		return nil
+	}
+	// If always returning nil here, GitHub Advanced Security may report "Use of insecure HostKeyCallback implementation".
+	// So, we return an error here to make the SSH client report the host key mismatch.
+	return errHostKeyMismatch
 }


norio-nomura:
~~Expecting CI failures will be fixed by #4336 and #4341~~
Fixed.

norio-nomura:
Following change added to avoid CodeQL error:
- ~~feat: Generate SSH server keys in host agent and use them in guest OS~~
Dropped

norio-nomura:
> Following change added to avoid CodeQL error:
> 
> * feat: Generate SSH server keys in host agent and use them in guest OS

Dropped this feature, because https://github.com/lima-vm/alpine-lima does not support ssh_keys field in user-data, and #4333 depends on this PR won't work with that.

norio-nomura:
Added:

- ~~CodeQL: Exclude `go/insecure-hostkeycallback`~~
  ~~Split `ssh.InsecureIgnoreHostKey()` usage into pkg/sshutil/sshutil_use_insecure_hostkeycallback.go and
  exclude go/insecure-hostkeycallback on that file.
  We already ignore host key verification when executing SSH commands in Lima,
  by using UserKnownHostsFile=/dev/null and StrictHostKeyChecking=no in the SSH command arguments.~~
Dropped.

pkg/sshutil/sshutil_use_insecure_hostkeycallback.go
```diff
@@ -0,0 +1,27 @@
...
+
+// sshClientConfigWithInsecureHostKeyCallback returns an ssh.ClientConfig
+// that uses ssh.InsecureIgnoreHostKey() as the HostKeyCallback.
+// We already ignore host key verification when executing SSH commands in Lima,
+// by using UserKnownHostsFile=/dev/null and StrictHostKeyChecking=no in the SSH command arguments.
+func sshClientConfigWithInsecureHostKeyCallback(user string, timeoutSeconds int, signers ...ssh.Signer) *ssh.ClientConfig {
+	return &ssh.ClientConfig{
+		User:            user,
+		Auth:            []ssh.AuthMethod{ssh.PublicKeys(signers...)},
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
```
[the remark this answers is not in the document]
norio-nomura: Is GitHub Advanced Security different from the CodeQL set in .github/workflows/codeql.yaml?

pkg/sshutil/sshutil.go
```diff
@@ -509,3 +530,234 @@ func detectAESAcceleration() bool {
...
+		User:            user,
+		Auth:            []ssh.AuthMethod{ssh.PublicKeys(signer)},
+		HostKeyCallback: hostKeyChecker.check,
+		Timeout:         10 * time.Second,
+	}
+	// Wait until the SSH server is available.
+	for {
+		conn, err := dialContext(ctx)
+		if err == nil {
+			sshConn, chans, reqs, err := ssh.NewClientConn(conn, addr, sshConfig)
```
AkihiroSuda, line 539: It is quite hard to maintain the code that mixes up `/usr/bin/ssh` with `golang.org/x/crypto/ssh`

AkihiroSuda: The PR looks overengineering.

Can we consider another way to fix the issue #4334 ?

e.g., by writing vsock availability to a serial port

norio-nomura: What I really want to do is #4333, and this is like a grounding.

norio-nomura: I feel that #4333 and this can also be implemented without using host key generation and `known_hosts`.
If the host key generation and the `known_hosts` area disappear, I think the implementation will be simpler.

norio-nomura: > e.g., by writing vsock availability to a serial port

I tried to reconsider this, but I think it's difficult.
On Fedora 43's case:
- Like ubuntu, where normal SSH over VSOCK can be used, `systemd` listens to `VSOCK:22`.
```console
[norio@lima-fedora ~]$ sudo ss -l --vsock -p
Netid State  Recv-Q Send-Q Local Address:Port Peer Address:Port Process                                     
v_str LISTEN 0      0                  *:22              *:*     users:(("systemd",pid=1,fd=228))           
v_str LISTEN 0      0                  3:2222            *:*     users:(("lima-guestagent",pid=1443,fd=13)) 
```
- Accepted connection is denied by some mechanism(audit?).
port availability detection log:
  > [norio@lima-fedora ~]$ sudo journalctl|grep -E '\[(1030|1037)\]'
Nov 20 08:32:47 lima-fedora audit[1030]: AVC avc:  denied  { getattr } for  pid=1030 comm="sshd-session" scontext=system_u:system_r:sshd_session_t:s0-s0:c0.c1023 tcontext=system_u:system_r:sshd_t:s0-s0:c0.c1023 tclass=vsock_socket permissive=1
Nov 20 08:32:47 lima-fedora sshd-session[1030]: banner exchange: Connection from UNKNOWN port 65535: Broken pipe

  Actual log on SSH connection via forwarded to VSOCK is denied:
  > Nov 20 08:32:48 lima-fedora sshd-session[1037]: ssh_dispatch_run_fatal: Connection from UNKNOWN port 65535: Permission denied [preauth]
Nov 20 08:32:48 lima-fedora audit[1037]: AUDIT1112 pid=1037 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:sshd_session_t:s0-s0:c0.c1023 msg='op=login acct="(unknown)" exe="/usr/libexec/openssh/sshd-session" hostname=? addr=UNKNOWN terminal=ssh res=failed'

To writing availability via serial:
- Detecting listener on VSOCK:22 is not sufficient, because Fedora 43 is listening.
- It requires detecting configuration of audit? or another refusing mechanism on other distros. And it must be detected whether they are set to interrupt the connection.

So, I think, it is more reliable to actually try the connection with the SSH protocol, and the portability is also high.
And using `ssh` installed on the host to test the connection to VSOCK will increase the same uncertainty as the execution of requirements that cause CI failure.

norio-nomura:
> * CodeQL: Exclude `go/insecure-hostkeycallback`

This has been dropped since GitHub Advanced Security does not affect that configuration.

Instead of them, added:
- pkg/sshutil: Use `hostKeyCollector().checker()` instead of `ssh.InsecureIgnoreHostKey()`
`hostKeyCollector().checker()`:
checker returns a HostKeyCallback that either checks and collects the host key,
or only checks the host key, depending on whether any host keys have been collected.
It is expected to pass host key checks by retrying after the first collection.
On second invocation, it will only check the host key.

This implementation becomes much simpler than https://github.com/lima-vm/lima/commit/3ae0d293d2501ddd3126833ab481680729983819 :
- Dropped: generating host key and passing it to the guest
- Dropped: generating, using, and updating `ssh_known_hosts`

norio-nomura:
In this PR, it detects that SSH over VSOCK cannot be connected to the Fedora 43 VM as follows, and it will fall back via usernet.
> INFO[0010] [hostagent] Failed to detect SSH server on vsock port, falling back to usernet forwarder  error="failed to create ssh.Conn to \"vsock:22\": ssh: handshake failed: EOF"

pkg/sshutil/sshutil.go
AkihiroSuda: > And using ssh installed on the host to test the connection to VSOCK will increase the same uncertainty as the execution of requirements that cause CI failure.

Why can `crypto/ssh` avoid that uncertainty?
Can we just fix that uncertainty by executing `/usr/bin/ssh` with some option?

norio-nomura: Because `x/crpyto/ssh` does not use inter-process communication.
I think `stderr=\"\": read |0: bad file descriptor"` that I'm trying to fix in #4333 is an error in pipe used for inter-process communication. I don't think it can be fixed with the option to pass to `/usr/bin/ssh`.
It might be fixed if the executing method of `/usr/bin/ssh` changes, but it seemed better to reduce external dependencies at the run time.
After trying #4333 and #4337 so far, I thought `x/crypto/ssh` would be reliable enough to execute requirements.

AkihiroSuda: > It might be fixed if the executing method of /usr/bin/ssh changes, but it seemed better to reduce external dependencies at the run time.

`limactl shell` definitely depends on `/usr/bin/ssh` though

norio-nomura: > `limactl shell` definitely depends on `/usr/bin/ssh` though

It can also be implemented in `x/crypto/ssh`. There's no reason for us to do that.
All SSH communication doesn't have to depend on either `/usr/bin/ssh` or `x/crypto/ssh`.

AkihiroSuda:
We may consider this PR for v2.1+, but for v2.0.2 let's merge the PR below to avoid overengineering:
- https://github.com/lima-vm/lima/pull/4384

pkg/sshutil/sshutil.go
AkihiroSuda: Lima still has to generate `~/.lima/INST/ssh.config` for interoperability with external programs, and has to verify that the generated config works, regardless to the `limactl shell` implementation uses it or not.

PR #4345, opened 2025-11-14
driver(external): wait for `Start()` to complete on the server side 

### LINKED ISSUE #4343: `VZ` gets stuck waiting for `AdditionalSetupForSSH()`
### Description

When `vz` is built as an external driver it gets stuck waiting for `AdditionalSetupForSSH()` to complete when starting an instance but works fine when built as internal:
```
➜  lima2 git:(fix/ext-driver-residue) ✗ lm start
INFO[0001] Starting the instance "default" with external VM driver "vz" 
INFO[0001] Attempting to download the image              arch=aarch64 digest="sha256:696b9ec765c0e3964b7845feb7217ef7e159e026c6b70eb0c8c2fed34e067067" location="https://cloud-images.ubuntu.com/releases/questing/release-20251023/ubuntu-25.10-server-cloudimg-arm64.img"
INFO[0001] Using cache "/Users/ansumansahoo/Library/Caches/lima/download/by-url-sha256/b33d6180955550f21ceb2308eba648114e93a71d7b81ff860b0e2f252c3656eb/data" 
INFO[0006] Attempting to download the nerdctl archive    arch=aarch64 digest="sha256:5398f037ae095d43cf3cb8c30765a24e511e38cafe02977b928a41b26e842ed1" location="https://github.com/containerd/nerdctl/releases/download/v2.2.0/nerdctl-full-2.2.0-linux-arm64.tar.gz"
INFO[0006] Using cache "/Users/ansumansahoo/Library/Caches/lima/download/by-url-sha256/e27422aaa9393572e7dc134eeb158f08328f22ca5e7556c80bdd221497a0a84a/data" 
INFO[0007] [hostagent] hostagent socket created at /Users/ansumansahoo/.lima/default/ha.sock 
^C
```

- Macbook Air M1(MacOS 26.1)
- limactl version 2.0.0-alpha.0-494-ge21b6346.m

### LINKED ISSUE #4348: `DriverClient.Start()` should wait execution of actual driver's `Start()`
https://github.com/lima-vm/lima/issues/4343#issuecomment-3537460687
> I checked the operation with the debugger. The real problem is that `DriverClient.Start()` does not wait for the execution of `LimaVzDriver.Start()`. As a result, `AdditionalSetupForSSH()` is called and hangs up before the waiting channel is set up in `Start()`. `DriverClient.Start()` and `DriverServer.Start()` should be modified.

unsuman:
Fixes #4343

[Edit]: Also fixes #4348

### DIFF
--- a/pkg/driver/external/client/methods.go
+++ b/pkg/driver/external/client/methods.go
@@ -58,6 +58,9 @@ func (d *DriverClient) CreateDisk(ctx context.Context) error {
 	return nil
 }
 
+// Start initiates the driver instance and receives streaming responses. It blocks until
+// receiving the initial success response, then spawns a goroutine to consume subsequent
+// error messages from the stream. Any errors from the driver are sent to the channel.
 func (d *DriverClient) Start(ctx context.Context) (chan error, error) {
 	d.logger.Debug("Starting driver instance")
 
@@ -67,19 +70,37 @@ func (d *DriverClient) Start(ctx context.Context) (chan error, error) {
 		return nil, err
 	}
 
+	// Blocking to receive an initial response to ensure Start() is initiated
+	// at the server-side.
+	initialResp, err := stream.Recv()
+	if err != nil {
+		d.logger.WithError(err).Error("Error receiving initial response from driver start")
+		return nil, err
+	}
+	if !initialResp.Success {
+		return nil, errors.New(initialResp.Error)
+	}
+
+	go func() {
+		<-ctx.Done()
+		if closeErr := stream.CloseSend(); closeErr != nil {
+			d.logger.WithError(closeErr).Warn("Failed to close stream")
+		}
+	}()
+
 	errCh := make(chan error, 1)
 	go func() {
 		for {
-			errorStream, err := stream.Recv()
+			respStream, err := stream.Recv()
 			if err != nil {
-				d.logger.Errorf("Error receiving response from driver: %v", err)
+				d.logger.Infof("Error receiving response from driver: %v", err)
 				return
 			}
-			d.logger.Debugf("Received response: %v", errorStream)
-			if !errorStream.Success {
-				errCh <- errors.New(errorStream.Error)
+			d.logger.Debugf("Received response: %v", respStream)
+			if !respStream.Success {
+				errCh <- errors.New(respStream.Error)
 			} else {
-				errCh <- nil
+				close(errCh)
 				return
 			}
 		}
--- a/pkg/driver/external/server/methods.go
+++ b/pkg/driver/external/server/methods.go
@@ -26,7 +26,18 @@ func (s *DriverServer) Start(_ *emptypb.Empty, stream pb.Driver_StartServer) err
 	errChan, err := s.driver.Start(stream.Context())
 	if err != nil {
 		s.logger.Errorf("Start failed: %v", err)
+		if sendErr := stream.Send(&pb.StartResponse{Success: false, Error: err.Error()}); sendErr != nil {
+			s.logger.Errorf("Failed to send error response: %v", sendErr)
+			return status.Errorf(codes.Internal, "failed to send error response: %v", sendErr)
+		}
 		return status.Errorf(codes.Internal, "failed to start driver: %v", err)
+	}
+
+	// First send a success response upon receiving the errChan to unblock the client
+	// and start receiving errors (if any).
+	if err := stream.Send(&pb.StartResponse{Success: true}); err != nil {
+		s.logger.Errorf("Failed to send success response: %v", err)
+		return status.Errorf(codes.Internal, "failed to send success response: %v", err)
 	}
 
 	for {
--- a/pkg/driver/external/driver.pb.go
+++ b/pkg/driver/external/driver.pb.go
@@ -154,6 +154,11 @@ func (x *InfoResponse) GetInfoJson() []byte {
 	return nil
 }
 
+// StartResponse is a streamed response for Start() RPC. It tries to mimic
+// errChan from pkg/driver/driver.go. The server sends an initial response
+// with success=true when Start() is initiated. If errors occur, they are
+// sent as success=false with the error field populated. When the error channel
+// closes, a final success=true message is sent.
 type StartResponse struct {
 	state         protoimpl.MessageState `protogen:"open.v1"`
 	Success       bool                   `protobuf:"varint,1,opt,name=success,proto3" json:"success,omitempty"`
--- a/pkg/driver/external/driver.proto
+++ b/pkg/driver/external/driver.proto
@@ -46,6 +46,11 @@ message InfoResponse{
   bytes info_json = 1;
 }
 
+// StartResponse is a streamed response for Start() RPC. It tries to mimic
+// errChan from pkg/driver/driver.go. The server sends an initial response 
+// with success=true when Start() is initiated. If errors occur, they are 
+// sent as success=false with the error field populated. When the error channel 
+// closes, a final success=true message is sent.
 message StartResponse {
   bool success = 1;
   string error = 2;
--- a/pkg/driver/external/server/server.go
+++ b/pkg/driver/external/server/server.go
@@ -207,6 +207,9 @@ func handlePreConfiguredDriverAction(ctx context.Context, driver driver.Driver) 
 	}
 }
 
+// Start begins the driver startup process. It sends an initial response to unblock
+// the client and then streams subsequent errors(if any), as the driver initializes.
+// A final success message is streamed upon successful completion.
 func Start(extDriver *registry.ExternalDriver, instName string) error {
 	extDriver.Logger.Debugf("Starting external driver at %s", extDriver.Path)
 	if instName == "" {


unsuman:
Now it waits for `Start()` to complete before running `AdditionalSetupForSSH()`, can you please check?

norio-nomura:
With following patch:
```diff
@@ -0,0 +1,21 @@
diff --git forkSrcPrefix/pkg/driver/vz/vz_driver_darwin.go forkDstPrefix/pkg/driver/vz/vz_driver_darwin.go
index c87995a62c8b53b7724ba4020981c3f5851cd86f..83b074f95bd39c5b7b371cc7861d5b7c3b298037 100644
--- forkSrcPrefix/pkg/driver/vz/vz_driver_darwin.go
+++ forkDstPrefix/pkg/driver/vz/vz_driver_darwin.go
@@ -299,6 +299,7 @@ func (l *LimaVzDriver) CreateDisk(ctx context.Context) error {
 
 func (l *LimaVzDriver) Start(ctx context.Context) (chan error, error) {
 	logrus.Infof("Starting VZ (hint: to watch the boot progress, see %q)", filepath.Join(l.Instance.Dir, "serial*.log"))
+	defer logrus.Infof("Started VZ")
 	vm, waitSSHLocalPortAccessible, errCh, err := startVM(ctx, l.Instance, l.SSHLocalPort)
 	if err != nil {
 		if errors.Is(err, vz.ErrUnsupportedOSVersion) {
@@ -441,6 +442,8 @@ func (l *LimaVzDriver) ForwardGuestAgent() bool {
 }
 
 func (l *LimaVzDriver) AdditionalSetupForSSH(ctx context.Context) error {
+	logrus.Infof("AdditionalSetupForSSH: waiting for SSH port %d to be accessible", l.SSHLocalPort)
+	defer logrus.Infof("AdditionalSetupForSSH: SSH port %d is accessible", l.SSHLocalPort)
 	ticker := time.NewTicker(100 * time.Millisecond)
 	defer ticker.Stop()
 
```
following logs are recorded in `driver.stderr.log`:
```log
time="2025-11-18T18:42:58+09:00" level=debug msg="Received AdditionalSetupForSSH request"
time="2025-11-18T18:42:58+09:00" level=info msg="AdditionalSetupForSSH: waiting for SSH port 50725 to be accessible"
time="2025-11-18T18:42:58+09:00" level=debug msg="Received Start request"
time="2025-11-18T18:42:58+09:00" level=info msg="Starting VZ (hint: to watch the boot progress, see \"/Users/norio/.lima/default/serial*.log\")"
time="2025-11-18T18:42:58+09:00" level=info msg="Started VZ"
time="2025-11-18T18:42:58+09:00" level=debug msg="Stream context done, stopping Start"
time="2025-11-18T18:42:58+09:00" level=info msg="[VZ] - vm state change: running"
time="2025-11-18T18:43:05+09:00" level=info msg="Started vsock forwarder: 127.0.0.1:50725 -> vsock:22 on VM"
time="2025-11-18T18:43:05+09:00" level=info msg="Detected SSH server is listening on the vsock port; changed 127.0.0.1:50725 to proxy for the vsock port"
time="2025-11-18T18:43:06+09:00" level=info msg="AdditionalSetupForSSH: SSH port 50725 is accessible"
time="2025-11-18T18:43:06+09:00" level=debug msg="AdditionalSetupForSSH succeeded"
```

It indicates that `LimaVzDriver.AdditionalSetupForSSH()` is called before calling `LimaVzDriver.Start()`.

I think:
- `DriverServer.Start()` should send initial response that contains `Error` or `Success` that express the result returned by the driver's `Start()`.
- `DriverClient.Start()` should wait initial response, then returns them.
- [Server side stream](https://pkg.go.dev/google.golang.org/grpc#ServerStreamingServer) will be closed when `DriverServer.Start()` returns, so it should not returns until `errChan` is closed by driver or stream is closed by client.
- [Client side stream](https://pkg.go.dev/google.golang.org/grpc#ServerStreamingClient) will not be closed when `DriverClient.Start()` returns, and can be closed by `stream.CloseSend()`. it should be called when `ctx.Done()`.
- The driver's `Start()` should receive `stream.Context()` to handle Server shutdown.

additionally:
- The closing driver's `errChan` event can't be expressed in current API design. But current drivers behavior does not defined when `errChan` should be closed. So, it won't be problems.

Thoughts about driver's implementation that not related to this PR:
- `server.Serve()` is implemented to handling `SIGINT` and `SIGINT`, but `server.Stop()` sends `SIGKILL`.

norio-nomura:
Added link to  gRPC API document on above comment.

unsuman:
Thanks for the detailed explanation!

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -440,7 +440,24 @@ func (l *LimaVzDriver) ForwardGuestAgent() bool {
...
+		}
+
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-timeout:
+			return errors.New("timeout waiting for Start() to initialize")
+		case <-ticker.C:
+		}
+	}
```
unsuman: Reverting these changes also works but I think it can remain so to avoid listening on an empty or uninitialised channel? @norio-nomura

norio-nomura: If this is "we don't know why it's happening, but we know what's going on", I think it's an effective code to deal with, but now that we know why it's happening, it's an unnecessary code.

If there might be something else to deal with, try not to create a situation where the channel has not been created.
When I decided to continue using `waitSSHLocalPortAccessible` in #4251, I didn't expect `AdditionalSetupForSSH()` to be called before `Start()` was called, so I left it created in `Start()`.
However, if I had created a channel at the same time as creating the `LimaVzDriver` instance, the problem with `AdditionalSetupForSSH()` would not have occurred this time, and it might not have caught my attention, and the behavior of `DriverClient.Start()` might not have been known.
And it was causing another problem in another place (already occurring in wsl2?) It may be, so I think it's good to be as it is now.

If it had been specified as a document that "`DriverClient.Start()` does not wait for the actual `Driver.Start()`", it might have been different.

---

I think the current Driver API lacks documentation (doc comment).

Previously, in #4248, I tried to use `SSHAddress()` as the current `AdditionalSetupForSSH()`.
I stopped for the time being because the explanation was given, but to be honest, I didn't understand it well even after reading the explanation.

I want it to be clearly stated in the document.
- When is it used: Before the instance starts? After activation?
- What is it used for?

There may be people who want to judge from the usage situation.
But for example:
https://github.com/lima-vm/lima/blob/8b7289178cb3417bc17e475b2898e97ba61bd145/pkg/hostagent/hostagent.go#L164-L168
Questions such as come to mind:
- Why does this call the same API 5 times in a row?
- Does this assume that something will change on the driver side by calling it continuously?
- Is this method of use correct?

To be honest, I'm not very interested in external drivers, so it doesn't make sense to answer the question here.
However, new contributors who may implement external drivers have no choice but to see how they are used in the current situation where there is no documentation, and I think they will have similar questions.
I don't know if there is anyone who can answer when someone asks that question in the future.

pkg/driver/external/client/methods.go
```diff
@@ -67,8 +67,27 @@ func (d *DriverClient) Start(ctx context.Context) (chan error, error) {
 		return nil, err
 	}
 
+	// Blocking to receive an initial response to ensure Start() is initiated
+	// at the server-side.
+	for {
```
norio-nomura: There is no need to loop here.
`initialResp.Error` should be returned as the return value of `Start()`.
The subsequent `StartResponse.Error` should be sent to `errCh`.

pkg/driver/external/client/methods.go
```diff
@@ -67,8 +67,27 @@ func (d *DriverClient) Start(ctx context.Context) (chan error, error) {
...
+			return nil, err
+		}
+		if initialResp.Success {
+			break
+		}
+	}
+
 	errCh := make(chan error, 1)
 	go func() {
+		defer func() {
```
norio-nomura: This should not be `defer func()`, but `go func()` outside and wait for `ctx.Done()` to call `stream.CloseSend()`.
In the current situation, the only way to get out of the go routine is to shut down on the server side.
When `stream.CloseSend()` is called, `stream.Recv()` should return an error and exit the go routine.

pkg/driver/external/client/methods.go
```diff
@@ -67,8 +67,27 @@ func (d *DriverClient) Start(ctx context.Context) (chan error, error) {
...
+			if closeErr := stream.CloseSend(); closeErr != nil {
+				d.logger.Warnf("Failed to close stream: %v", closeErr)
+			}
+		}()
+
 		for {
 			errorStream, err := stream.Recv()
```
norio-nomura: The variable name `errStream` is strange. The content is `*pb.StartResponse`.

pkg/driver/external/client/methods.go
norio-nomura: When the stream is closed from the server side, it is better to leave it in the log with `Infof` instead of `Errorf()`.

norio-nomura: When the driver closes `errChan` on the server side, it is decided to send `StartResponse{Success:true}` to the client, and it may be good to close `errCh` on the client side here.
Even if the driver closes or does not close `errChan` at any time, even if it is not defined.

pkg/driver/external/client/methods.go
norio-nomura: If `initialResp.Success` is false, there is no need to return `errCh`.

pkg/driver/external/server/methods.go
norio-nomura, line 27: The `err` here should be sent to the client as `initialResponse`.
If `initialResponse.Success` is `true`, the subsequent error received by `errChan` is sent to the client.
When `errChan` is closed, it would be a good idea to send `pb.StartResponse{Success: true}` to express channel closure.

pkg/driver/vz/vz_driver_darwin.go
unsuman: > and it might not have caught my attention, and the behavior of DriverClient.Start() might not have been known.

Many thanks!🙏🏻

> To be honest, I'm not very interested in external drivers, so it doesn't make sense to answer the question here.
> However, new contributors who may implement external drivers have no choice but to see how they are used in the current situation where there is no documentation, and I think they will have similar questions.
> I don't know if there is anyone who can answer when someone asks that question in the future.

Thanks for pointing this out. I implemented all these things as a sprint behaviour during GSOC and also I was new to the codebase. As you’ve pointed out, many of them are inefficient. Also when implementing registry package and driver-server I assumed that one `limactl` process may invoke multiple external drivers but I was wrong and have to work on this too!

Since the driver API hasn’t been finalised yet, a descriptive doc is yet to be created.

pkg/driver/external/client/methods.go
```diff
@@ -67,19 +67,37 @@ func (d *DriverClient) Start(ctx context.Context) (chan error, error) {
...
 
+	// Blocking to receive an initial response to ensure Start() is initiated
+	// at the server-side.
+	initialResp, err := stream.Recv()
+	if err != nil {
+		d.logger.Errorf("Error receiving initial response from driver start: %v", err)
+		return nil, err
+	}
+	if !initialResp.Success {
+		return nil, errors.New(initialResp.Error)
```
norio-nomura, line 81: Would it be better to `stream.CloseSend()` here?
The server side should also be closed, so there may be no problem, but I'm not familiar enough with gRPC to determine whether it's a problem or not.

pkg/driver/external/client/methods.go
```diff
@@ -67,19 +67,37 @@ func (d *DriverClient) Start(ctx context.Context) (chan error, error) {
 		return nil, err
 	}
 
+	// Blocking to receive an initial response to ensure Start() is initiated
+	// at the server-side.
+	initialResp, err := stream.Recv()
+	if err != nil {
+		d.logger.Errorf("Error receiving initial response from driver start: %v", err)
```
norio-nomura: I have previously been asked to use `logger.WithError(err)` when recording `Error` with `logger`.

norio-nomura (changes requested):
It looks like it has become the desired operation. 👍🏻 

After that, since it's a good opportunity, let's add the action decided in this PR as a doc comment to the definition below.
- `pkg/driver/external/driver.proto`: `StartResponse`
- `pkg/driver/external/client/methods.go`: `Start()`
- `pkg/driver/external/server/methods.go`: `Start()`

Actually, it would be better to test the behavior with a mock driver, but I think it's okay to do another PR.

pkg/driver/external/client/methods.go
unsuman: > The server side should also be closed, so there may be no problem, but I'm not familiar enough with gRPC to determine whether it's a problem or not.

Yes, no need to close here!

pkg/driver/external/client/methods.go
unsuman: If that's the case then all the error logging of `pkg/driver/external/*/methods.go` needs to be changed which can be a new issue. I am only changing for this PR.

unsuman: Created #4380

unsuman:
> Actually, it would be better to test the behavior with a mock driver, but I think it's okay to do another PR.

Do you mean to add a test in the CI which builds the current drivers as external and tests the gRPC transport?

norio-nomura:
> > Actually, it would be better to test the behavior with a mock driver, but I think it's okay to do another PR.
> 
> Do you mean to add a test in the CI which builds the current drivers as external and tests the gRPC transport?

No. Isn't it possible to test the behavior of `Start()` with dummy driver in the Unit test? I thought about it, but I don't know if it's easy to do, and it seems too much.

However, there is no doubt that #4343 could have been found earlier if it had been tested in CI.
Personally, I think it would be better to test external drivers in a job scheduled to be performed once a day, rather than a test that is required to pass in PR.

norio-nomura:
Thanks! 🙏🏻

PR #4346, opened 2025-11-14
Allow using swap as a fsType for disks

### LINKED ISSUE #965: Option to add swap space to vm
### Description

Add option to configure a swapping amount for the vm created by lima.

afbjorklund:
This will call "mkswap" instead of mkfs.
    
And will call "swapon" instead of mount.

Improve message, without a bogus "/mnt/lima-swap" dir:

`INFO[0020] [hostagent] Mounting disk "swap" on "[SWAP]"`

Closes #965

----

`limactl disk create swap --size 2g`

```yaml
additionalDisks:
- name: "swap"
  format: true
  fsType: swap
```

```console
$ free -h
               total        used        free      shared  buff/cache   available
Mem:           3.8Gi       522Mi       2.7Gi       4.1Mi       868Mi       3.3Gi
Swap:          2.0Gi          0B       2.0Gi
$ cat /proc/swaps
Filename				Type		Size		Used		Priority
/dev/vdb1                               partition	2095100		0		-2
```

We can mention swapfile and compcache in a followup PR.

* https://en.wikipedia.org/wiki/Swapfile

* https://en.wikipedia.org/wiki/Compcache

Since the user can add those themselves, after it has started.

### DIFF
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -159,6 +159,10 @@ if [[ -n ${CHECKS["disk"]} ]]; then
 		defer "limactl disk delete data"
 		limactl disk create data --size 10G
 	fi
+	if ! limactl disk ls | grep -q "^swap\s"; then
+		defer "limactl disk delete swap"
+		limactl disk create swap --size 2G
+	fi
 fi
 
 set -x
@@ -513,6 +517,10 @@ if [[ -n ${CHECKS["disk"]} ]]; then
 	INFO "Testing disk is attached"
 	set -x
 	if ! limactl shell "$NAME" lsblk --output NAME,MOUNTPOINT | grep -q "/mnt/lima-data"; then
+		ERROR "Disk is not mounted"
+		exit 1
+	fi
+	if ! limactl shell "$NAME" lsblk --output NAME,MOUNTPOINT | grep -q "\[SWAP\]"; then
 		ERROR "Disk is not mounted"
 		exit 1
 	fi
--- a/hack/test-templates/test-misc.yaml
+++ b/hack/test-templates/test-misc.yaml
@@ -98,10 +98,14 @@ probes:
     #!/bin/sh
     touch /tmp/param-$PARAM_PROBE
 
-# in order to use this example, you must first create the disk "data". run:
+# in order to use this example, you must first create the disks. run:
 # $ limactl disk create data --size 10G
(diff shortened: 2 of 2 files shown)


AkihiroSuda:
Maybe add a test in test-misc ?

pkg/store/disk.go
```diff
@@ -56,7 +60,11 @@ func InspectDisk(diskName string) (*Disk, error) {
 		disk.InstanceDir = instDir
 	}
 
-	disk.MountPoint = fmt.Sprintf("/mnt/lima-%s", diskName)
+	if disk.FSType != nil && *disk.FSType == "swap" {
+		disk.MountPoint = "[SWAP]" // only used for log message
```
AkihiroSuda, line 64: Shouldn't this be just empty?
The log should just print the fstype

afbjorklund: The log prints out "mounted on"
`logrus.Infof("Mounting disk %q on %q", disk.Name, disk.MountPoint)`

By default it prints /mnt/lima-swap

pkg/store/disk.go
```diff
@@ -23,11 +23,15 @@ type Disk struct {
 	Instance    string `json:"instance"`
 	InstanceDir string `json:"instanceDir"`
 	MountPoint  string `json:"mountPoint"`
+
+	// if the Disk is in use and the FSType is specified
+	FSType *string `json:"fsType,omitempty"`
```
AkihiroSuda, line 28: Why `*string`, not `string`?
Do we need to distinguish nil from empty?

afbjorklund: We don't actually need it, just thought it would be less noise since it was exposed in all the drivers...

Now it can just copy from the yaml:

```go
type Disk struct {
        Name   string   `yaml:"name" json:"name"` // REQUIRED
        Format *bool    `yaml:"format,omitempty" json:"format,omitempty"`
        FSType *string  `yaml:"fsType,omitempty" json:"fsType,omitempty"`
        FSArgs []string `yaml:"fsArgs,omitempty" json:"fsArgs,omitempty"`
}
```

Not really happy about that, so it would be nice if there was a away to avoid doing it in `InspectDisk`

afbjorklund: Make save the config to a disk.yaml, similar to the lima.yaml that is saved in the instance directory

PR #4349, opened 2025-11-16
Don't use special message for errUnimplemented

afbjorklund:
One day we might want to combine these in lima core.

Instead of having each driver and os do their own...

It's a bit annoying to have one "errors.go" file for each?

```go
import "errors"

var errUnimplemented = errors.New("unimplemented")
```

### DIFF
--- a/pkg/driver/krunkit/errors_darwin_arm64.go
+++ b/pkg/driver/krunkit/errors_darwin_arm64.go
@@ -5,4 +5,4 @@ package krunkit
 
 import "errors"
 
-var errUnimplemented = errors.New("unimplemented by the krunkit driver")
+var errUnimplemented = errors.New("unimplemented")
--- a/pkg/osutil/osutil_windows.go
+++ b/pkg/osutil/osutil_windows.go
@@ -55,5 +55,5 @@ func SignalName(sig os.Signal) string {
 }
 
 func Sysctl(_ context.Context, _ string) (string, error) {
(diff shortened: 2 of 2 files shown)


pkg/driver/krunkit/errors_darwin_arm64.go
```diff
@@ -5,4 +5,4 @@ package krunkit
 
 import "errors"
 
-var errUnimplemented = errors.New("unimplemented by the krunkit driver")
+var errUnimplemented = errors.New("unimplemented")
```
AkihiroSuda, line 8: The new error seems less descriptive

AkihiroSuda:
Why? This seems to make the error less descriptive

afbjorklund:
> Why? This seems to make the error less descriptive

We can pause this refactoring, until ErrUnimplemented is in place...

It was just getting a bit too duplicated:

./pkg/driver/vz/errors_darwin.go
./pkg/driver/wsl2/errors_windows.go
./pkg/driver/qemu/errors.go
./pkg/driver/ac/errors_darwin.go
./pkg/driver/dc/errors_linux.go
./pkg/driver/proxmox/errors.go

pkg/driver/krunkit/errors_darwin_arm64.go
afbjorklund: That was the point, so it can be reused.

PR #4352, opened 2025-11-18
Don't add reverse-sshfs mounts to cloud-init user-data

jandubois:
The template checks the mount type, but when Rosetta is enabled, then all mounts will be added to user-data regardless of the mount type. We need to repeat the check condition exactly.


AkihiroSuda:
CI seems failing

jandubois:
> CI seems failing

Yeah, but I don't know why. I just rebased on latest `master`, just in case. 😉

Issue #4363, opened 2025-11-18
Force stop/remove leaves `lima-driver-*` processes running


unsuman:
Force stopping or removing instances currently leaves external driver processes `lima-driver-*` running as zombies on the system. We should add cleanup logic in `driver/external/server.go` to track external driver PIDs and implement a signal termination function.

unsuman:
Is it a good idea if we create `driver.pid` inside `.lima/{INSTANCE}/` to store the external driver's PID? Since it is used here which might be confusing:
https://github.com/lima-vm/lima/blob/8b7289178cb3417bc17e475b2898e97ba61bd145/pkg/limatype/lima_instance.go#L43

PR #4364, opened 2025-11-18
Fix mismatch of signals for stopping external VMs

### LINKED ISSUE #4362: Signal handling mismatch for stopping external VM drivers
`server.Serve()` is implemented to handling `SIGINT` and `SIGTERM`, but `server.Stop()` sends `SIGKILL`.

_Originally posted by @norio-nomura in https://github.com/lima-vm/lima/issues/4345#issuecomment-3546720381_

unsuman:
Fixes #4362

### DIFF
--- a/pkg/driver/external/server/server.go
+++ b/pkg/driver/external/server/server.go
@@ -292,7 +292,7 @@ func Stop(extDriver *registry.ExternalDriver) error {
 	if extDriver.CancelFunc != nil {
 		extDriver.CancelFunc()
 	}
-	if err := extDriver.Command.Process.Kill(); err != nil {
+	if err := extDriver.Command.Process.Signal(syscall.SIGTERM); err != nil {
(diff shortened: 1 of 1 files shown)


norio-nomura:
With this change, `DriverServer` should be able to shut down gracefully.
As a result of the signal handler's role, `ga.sock` and others are expected to be cleaned up.
When `limactl start` is stopped by ctrl+C in the middle, it will be transmitted to the `DriverServer` side and `ga.sock` will not remain, which is the ideal that this change should bring.

norio-nomura:
> When `limactl start` is stopped by ctrl+C in the middle, it will be transmitted to the `DriverServer` side and `ga.sock` will not remain, which is the ideal that this change should bring.

Sorry, current implementation does not forward ctrl+C on `limactl start`.

unsuman:
I'll fix this in #4363

PR #4366, opened 2025-11-18
Resolve template image paths relative to template directory

### LINKED ISSUE #4365: Resolve relative paths in template images / kernel / initrd fields relative to template directory
Relative paths in the `images` field (and `kernel`/`initrd` sub-fields) are currently resolved relative to the current working directory instead of the template file's directory. All other locations in the template file are resolved relative to the template file so it would make sense to be consistent here.

For example:
Template at `/path/to/templates/myvm.yaml`
```
images:
  - location: ./images/ubuntu.img
```
Currently resolves to: `$(pwd)/images/ubuntu.img`
... but should resolve to: `/path/to/templates/images/ubuntu.img`

skaegi:
Relative paths in image/kernel/initrd locations now resolve relative to the template file instead of pwd. URLs and absolute paths unchanged.

- Added resolveLocationPath() helper function
- URLs and absolute paths remain unchanged
- Tilde and relative paths are expanded
- Added tests for path resolution behaviour

Fixes #4365

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -121,6 +121,29 @@ func defaultGuestInstallPrefix() string {
 	return "/usr/local"
 }
 
+// resolveLocationPath resolves a file location relative to instDir if it's a relative path.
+// URLs and absolute paths are left unchanged. Tilde paths are expanded.
+func resolveLocationPath(location, instDir string) string {
+	// Don't modify URLs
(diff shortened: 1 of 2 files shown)


skaegi:
Hmm... this is not valid. The template is copied into ~/.lima and the relative paths are again wrong. I'm closing this PR and not sure if there is an easy fix and might just have to get used to `--set` ing an absolute url at creation time.

PR #4381, opened 2025-11-20
Don't trim whitespace from boot commands

### LINKED ISSUE #4319: Losing whitespace/indentation in bootcmd scripts
I'm attempting to write a bootcmd that uses a bash heredoc to write some YAML to provide some config overrides to cloudinit before the network is initialized, but the problem is that lima strips whitespace when writing bootcmds to the bootcmd in the user-data:

https://github.com/lima-vm/lima/blob/e21b6346eb6a7cf4b6c2b28fcf1fa2215ae3cf18/pkg/cidata/cidata.go#L491-L499

Here's the relevant portion of my lima.yaml:
```
provision:
- mode: boot
  script: |
    cat <<EOF > /etc/cloud/cloud.cfg.d/99-lima1-disable-route.cfg
    ethernets:
      lima1:
        dhcp4-overrides:
          use-routes: false
    EOF
```

And here's what gets rendered in user-data:
```
bootcmd:
- |
  # We need to embed the params.env as a here-doc because /mnt/lima-cidata is not yet mounted
  while read -r line; do [ -n "$line" ] && export "$line"; done <<'EOF'
  EOF
  cat <<EOF > /etc/cloud/cloud.cfg.d/99-lima1-disable-route.cfg
  ethernets:
  lima1:
  dhcp4-overrides:
  use-routes: false
  EOF
```

jandubois:
It breaks here-docs.

The original code was likely copied from the getCert() function above, but skipping empty lines and stripping whitespace is not appropriate for a shell script.

Fixes #4319

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -489,14 +489,7 @@ func getBootCmds(p []limatype.Provision) []BootCmds {
 	var bootCmds []BootCmds
 	for _, f := range p {
 		if f.Mode == limatype.ProvisionModeBoot {
-			lines := []string{}
-			for line := range strings.SplitSeq(*f.Script, "\n") {
-				if line == "" {
-					continue
-				}
-				lines = append(lines, strings.TrimSpace(line))
-			}
-			bootCmds = append(bootCmds, BootCmds{Lines: lines})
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

Ideally we should have a test to cover this

PR #4391, opened 2025-11-21
CI: migrate test-colima.sh to colima.bats

AkihiroSuda:
Also updated to:
- use the latest Colima
- test port forwarding


AkihiroSuda:
ping @lima-vm/reviewers This has been open for a month

norio-nomura:
I don't understand the meaning of including colima's test in the PR merge conditions.

Previously, when my PR stopped passing only the colima test, I didn't know why the test failed from the log, and I didn't know why it healed on its own.

If some tests including colima do not pass in PR, the colima test will usually be fixed if other tests are fixed.
The problem is if only the colima test does not pass, and the current colima test does not provide enough information to find out the cause, and I think the investigation should be done by the colima project.

If the colima test is scheduled to run once a day in the master branch, we will notice that it is broken.
I think the colima test should be removed from the PR merge conditions.

AkihiroSuda:
> I don't understand the meaning of including colima's test in the PR merge conditions.

So as to find regressions ASAP.

> The problem is if only the colima test does not pass, and the current colima test does not provide enough information to find out the cause, and I think the investigation should be done by the colima project.

This indicates that the Lima PR has a regression that is not covered by other tests.
This should be basically investigated on Lima's side.
The PR author may add debug logs or try it locally to investigate the cause.

> I think the colima test should be removed from the PR merge conditions.

This can be discussed in a separate issue/PR.
Anyway, the colima test seems to have been stable.

.github/workflows/test.yml
```diff
@@ -395,27 +395,23 @@ jobs:
     - name: Install
       run: sudo make install
     - name: Install colima
+    - name: "Install colima"
```
jandubois: This duplicates the previous line, leaving an identically named step with no `run` key.

.github/workflows/test.yml
jandubois: This still references `matrix.colima-version` even though the version is now dynamically determined. Shouldn't the whole `matrix` thing be removed?

hack/bats/extras/colima.bats
```diff
@@ -0,0 +1,16 @@
...
+@test 'Docker' {
+    colima start
+    docker run -p 8080:80 -d --name nginx "${TEST_CONTAINER_IMAGES[nginx]}"
+    sleep 5
+    run curl -sSI http://localhost:8080
+    [ "$status" -eq 0 ]
+    [[ "$output" == *"200 OK"* ]]
+    docker rm -f nginx
+    colima stop
+    colima delete -f
```
jandubois: Should the `colima delete -f` be moved to a `local_teardown` function, so it is executed even if the test fails earlier?

AkihiroSuda: fixed

.github/workflows/test.yml
AkihiroSuda: fixed

.github/workflows/test.yml
AkihiroSuda: fixed

hack/bats/extras/colima.bats
```diff
@@ -0,0 +1,22 @@
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+load "../helpers/load"
+
+local_setup() {
```
jandubois: Since there is only one test in this file it doesn't matter. If there could be more tests, and they all should run against the same colima instance, then the functions should be


```suggestion
local_setup_file() {
```

and same for teardown.

hack/bats/extras/colima.bats
```diff
@@ -0,0 +1,22 @@
...
+}
+
+local_teardown() {
+    colima stop
+    colima delete -f
+}
+
+@test 'Docker' {
+    docker run -p 8080:80 -d --name nginx "${TEST_CONTAINER_IMAGES[nginx]}"
+    sleep 5
```
jandubois, line 17: Any kind of `sleep` is a bit of a red flag, as it can fail on a heavily loaded machine. It would be better to have some other way, but I can't think of anything beyond just retrying on failure, which is not actually better, but more complicated.

hack/bats/extras/colima.bats
```diff
@@ -0,0 +1,22 @@
...
+    run curl -sSI http://localhost:8080
+    [ "$status" -eq 0 ]
+    [[ "$output" == *"200 OK"* ]]
```
jandubois: The canonical BATS way would be:

```suggestion
    run -0 curl -sSI http://localhost:8080
    assert_output --partial "200 OK"
```

AkihiroSuda: fixed

hack/bats/extras/colima.bats
AkihiroSuda: fixed

hack/bats/extras/colima.bats
AkihiroSuda: fixed

PR #4405, opened 2025-11-25
Disable env access and file load operations in yqlib

jandubois:
Before:

```console
❯ l ls --yq 'env(HOME)'
/Users/jan

❯ l ls --yq 'load("/usr/local/share/lima/templates/ubuntu.yaml")'
{
    "minimumLimaVersion": "2.0.0",
    "base": [
        "template:_images/ubuntu-25.10",
        "template:_default/mounts"
    ]
}
```

After:

```console
❯ l ls --yq 'env(HOME)'
FATA[0000] env operations have been disabled

❯ l ls --yq 'load("/usr/local/share/lima/templates/ubuntu.yaml")'
FATA[0000] file operations have been disabled
```

Addresses objections to using YQ expressions for `limactl list --filter` in #4187.

### DIFF
--- a/hack/bats/tests/list.bats
+++ b/hack/bats/tests/list.bats
@@ -264,3 +264,13 @@ local_setup() {
     run -0 limactl ls --quiet --yq 'select(.name == "foo")'
     assert_output "foo"
 }
+
+@test '--yq cannot access environment variables' {
+    run_e -1 limactl ls --yq 'env(HOME)'
+    assert_fatal "env operations have been disabled"
+}
+
+@test '--yq cannot load files' {
+    run_e -1 limactl ls --yq "load(\"${BASH_SOURCE[0]}\")"
+    assert_fatal "file operations have been disabled"
+}
--- a/hack/bats/tests/yq.bats
+++ b/hack/bats/tests/yq.bats
@@ -35,3 +35,15 @@ load "../helpers/load"
     run -0 "$YQ" -n -o json -I 0 .foo=42
     assert_output '{"foo":42}'
 }
+
+@test 'yq multi-call command has support for env access' {
+    export FOO=bar
+    run -0 limactl yq -n 'env(FOO)'
+    assert_output "bar"
+}
+
+@test 'yq multi-call command has support for --security-disable-env-ops' {
+    export FOO=bar
+    run_e -1 limactl yq -n --security-disable-env-ops 'env(FOO)'
+    assert_stderr "Error: env operations have been disabled"
+}
--- a/pkg/yqutil/yqutil.go
+++ b/pkg/yqutil/yqutil.go
@@ -50,6 +50,10 @@ func EvaluateExpressionWithEncoder(expression, content string, encoder yqlib.Enc
 	logging.SetBackend(backend)
 	yqlib.InitExpressionParser()
 
+	// Disable access to environment variables and file loading functions
+	yqlib.ConfiguredSecurityPreferences.DisableEnvOps = true
+	yqlib.ConfiguredSecurityPreferences.DisableFileOps = true
+
 	decoder := yqlib.NewYamlDecoder(yqlib.ConfiguredYamlPreferences)
 	out, err := yqlib.NewStringEvaluator().EvaluateAll(expression, content, encoder, decoder)
 	if err != nil {


hack/bats/tests/list.bats
```diff
@@ -264,3 +264,13 @@ local_setup() {
     run -0 limactl ls --quiet --yq 'select(.name == "foo")'
     assert_output "foo"
 }
+
+@test '--yq cannot access environment variables' {
```
AkihiroSuda, line 268: I think we can allow --yq to access env, as long as the upstream yq continues to allow accessing env by default.

The proposed --filter should be restrictive though

jandubois: I think it is best to restrict it everywhere, except for `limactl yq` itself. 

What is the use case for `limactl list --yq` or `limactl info --yq` to use env variables or load from files?

jandubois: I continue to think that having just a single flavour of YQ expressions throughout Lima is preferable, especially since there are no known use-cases for the prohibited operations. Otherwise you have to document the kind of expression allowed separately for each option that takes a YQ expression.

And you can still do this if you really want to:

```console
limactl list --json | limactl yq '...'
```

Because the `yq` command itself is not limited, only any expression used in Lima internally.

AkihiroSuda: > the use case

Env can be useful for parametizing `limactl (create|start|edit) --set`.

The env can be resolved outside `limactl` though

jandubois: > Env can be useful for parametizing `limactl (create|start|edit) --set`.

This can easily be done by variable interpolation in the shell, or whatever program is invoking `limactl`.

I would still prefer to have a single YQ dialect that applies to all `limactl` options instead of documenting each one individually which operations are available.

I don't understand the desire to add conceptual complexity without a compelling use-case. The only reason to make this available is because the operators already exist. If they didn't, would we open an issue against YQ asking for them to be added?

AkihiroSuda: > I would still prefer to have a single YQ dialect that applies to all limactl options instead of documenting each one individually which operations are available.

When `limactl` is invoked with `yq` as `argv[0]`, it has to emulate the upstream `yq` behavior.

In the current implementation, the yq-emulating mode might not be using `yqutil.EvaluateExpressionWithEncoder`, but the implementation can change in future and it is hard to track that what "dialect" is applied to the global yqlib configuration.

We do not need to introduce any "dialect" if we avoid using yq in implementing `--filter`.
Instead of yq, containerd's filter package could be used: https://github.com/lima-vm/lima/pull/4187#discussion_r2671848242

AkihiroSuda: Let's merge this PR after:
- Add test for the yq multi-call binary
- Add a document to explain our "dialect" of yq language

jandubois:
> Let's merge this PR after:
> 
> * Add test for the yq multi-call binary

I've rebased the PR on latest `master` and added tests for the multi-call functionality.

> * Add a document to explain our "dialect" of yq language

I'm struggling to find a good place in the documentation for this. Like a place users would actually find.

I feel like we also need more documentation for the `create --set` option, including examples. This could go in the `limactl create --help` output, and could then mention the limitation on env and file operators in our YQ dialect. Other options could then point to it.

This might still be a bit unwieldy, so maybe we should start having "concept guides" the way `git` does it:

```console
❯ git help -g
The Git concept guides are:
   core-tutorial    A Git core tutorial for developers
   credentials      Providing usernames and passwords to Git
   cvs-migration    Git for CVS users
   diffcore         Tweaking diff output
   everyday         A useful minimum set of commands for Everyday Git
   faq              Frequently asked questions about using Git
   glossary         A Git Glossary
   namespaces       Git namespaces
   remote-helpers   Helper programs to interact with remote repositories
   submodules       Mounting one repository inside another
   tutorial         A tutorial introduction to Git
   tutorial-2       A tutorial introduction to Git: part two
   workflows        An overview of recommended workflows with Git

'git help -a' and 'git help -g' list available subcommands and some
concept guides. See 'git help <command>' or 'git help <concept>'
to read about a specific subcommand or concept.
See 'git help git' for an overview of the system.
```

There is also `kubectl options` as another precedent; not sure if it has other meta commands.

Anyways, this is quickly becoming out-of-scope for this PR and should be in a separate docs PR.

AkihiroSuda:
Is this PR still a draft?

> concept guides

This can be probably implemented using https://pkg.go.dev/github.com/spf13/cobra#Command.IsAdditionalHelpTopicCommand

jandubois:
> Is this PR still a draft?

I turned it into a draft because we said yesterday it needs documentation before it can be merged, and the documentation is still outstanding (see my comments above).

The code is complete, if you are willing to create an issue for the docs and merge the code as-is.

PR #4203, opened 2025-10-15
templates: update openSUSE Leap to 16

AkihiroSuda:
<details>
<summary>
<s>Fails, as the zypper repos are no longer enabled by default (why?)</s>
</summary>

<p>

```
+ zypper --non-interactive install -y --no-recommends sshfs iptables fuse3
Loading repository data...
Warning: No repositories defined. Operating only with the installed resolvables. Nothing can be installed.
Reading installed packages...
'iptables' not found in package names. Trying capabilities.
No provider of 'iptables' found.
'fuse3' not found in package names. Trying capabilities.
No provider of 'fuse3' found.
'sshfs' not found in package names. Trying capabilities.
No provider of 'sshfs' found.
LIMA 2025-10-15T16:03:21+09:00| WARNING: Failed to execute /mnt/lima-cidata/boot/30-install-packages.sh
```

</p>

</details>

PR #4406, opened 2025-11-25
Ask QEMU for virtio-scsi instead of virtio-scsi-pci

jandubois:
This allows QEMU to pick the specific driver itself. E.g. OpenSUSE Leap 16 for s390x does not have virtio-scsi-pci but does support virtio-scsi-ccw:

```console
❯ limactl start -y --vm-type qemu https://download.opensuse.org/distribution/leap/16.0/appliances/Leap-16.0-Minimal-VM.s390x-s390x-Cloud.qcow2
...
❯ limactl shell leap-16.0.s390x-s390x uname -a
Linux lima-leap-16-0-s390x-s390x 6.12.0-160000.5-default #1 SMP Wed Sep 10 15:26:25 UTC 2025 (3545bbd) s390x s390x s390x GNU/Linux
```

Ref: https://github.com/lima-vm/lima/pull/4203#discussion_r2431392663
Ref: https://bugzilla.suse.com/show_bug.cgi?id=1252096#c10

### DIFF
--- a/pkg/driver/qemu/qemu.go
+++ b/pkg/driver/qemu/qemu.go
@@ -751,7 +751,7 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	// cloud-init
 	args = append(args,
 		"-drive", "id=cdrom0,if=none,format=raw,readonly=on,file="+filepath.Join(cfg.InstanceDir, filenames.CIDataISO),
-		"-device", "virtio-scsi-pci,id=scsi0",
+		"-device", "virtio-scsi,id=scsi0",
 		"-device", "scsi-cd,bus=scsi0.0,drive=cdrom0")
 
 	// Kernel
--- a/templates/_images/opensuse-leap-16.yaml
+++ b/templates/_images/opensuse-leap-16.yaml
@@ -7,6 +7,5 @@ images:
 - location: https://download.opensuse.org/distribution/leap/16.0/appliances/Leap-16.0-Minimal-VM.aarch64-Cloud.qcow2
   arch: aarch64
 
-# s390x image will not boot until https://bugzilla.suse.com/show_bug.cgi?id=1252096 is resolved
 - location: https://download.opensuse.org/distribution/leap/16.0/appliances/Leap-16.0-Minimal-VM.s390x-s390x-Cloud.qcow2
   arch: s390x


pkg/driver/qemu/qemu.go
```diff
@@ -751,7 +751,7 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	// cloud-init
 	args = append(args,
 		"-drive", "id=cdrom0,if=none,format=raw,readonly=on,file="+filepath.Join(cfg.InstanceDir, filenames.CIDataISO),
-		"-device", "virtio-scsi-pci,id=scsi0",
+		"-device", "virtio-scsi,id=scsi0",
 		"-device", "scsi-cd,bus=scsi0.0,drive=cdrom0")
```
nirs, line 755: Why do you need scsi cdrom?

We can use virtio:

    -drive file=cidata.iso,media=cdrom,if=virtio

Or just:

    -cdrom cidata.iso

jandubois: I have no preferences what-so-ever. @AkihiroSuda / @afbjorklund What do you think?

afbjorklund: It depends on the distribution, I think. And it seems unrelated to this change, with a more portable device.

jandubois:
> You could probably do the same for others, not needed to hardcode "pci" like that.

I don't know what to do next now. Should we attempt to drop the "pci" in other places, or wait until it becomes an issue.

And should I try to use `-cdrom cidata.iso` as suggested by Nir, or not?

nirs:
I would do a minimal fix for 2.0.3 - just change virtio-scsi-pci to virtio-scsi - it this works on x86 and s390x we are fine.

Another option, use virtio-scsi-pci in x86 and virtio-scsi-ccw on s390x.

Then open an issue to simplify qemu configuration. I think we should be able to use virtio for everything - this is the most efficient way since virtio is designed to handle virtual devices. Using SCSI or other interfaces is needed only if you have special requirement and unlikely to be need in lima use case.

We can ask about it in qemu-discuss or check with qemu folks in SUSE/Red Hat.

afbjorklund:
I also think you don't need to do anything more here, I was just making an idle comment that most have aliases.

jandubois:
Since all tests are passing and it enables Leap 16 on s390x, I guess I'll merge as-is. Reading the cidata is not really a performance-critical part, so if no other problems are known, maybe this is good enough.

I'm going to remove the comment about the s390x image not working from the template though.

pkg/driver/qemu/qemu.go
afbjorklund: If I recall correctly, it was the matter of `/dev/hda` vs `/dev/sda` vs `/dev/vda` and matching kernel support

Vogtinator: IIRC `-cdrom` is highly arch/machine specific, using IDE on x86(_64), virtio-blk on aarch64 and s390x, spapr-vscsi on ppc64, ...

PR #4414, opened 2025-11-28
docs: stargz: ubuntu-24.04 is no longer needed

AkihiroSuda:
Ubuntu 25.04 and later has been supported since stargz-snapshotter v0.18.0 (Lima v1.2.2).


jandubois (approved):
Thanks, LGTM

PR #4419, opened 2025-12-01
Hostpath mount validation

Horiodino:
**Description:**
added validation to ensure `hostPath` is part of a configured mount. if the path is not mounted, an error is returned.

**Testing:**
added unit tests for `isMounted`

### DIFF
--- a/pkg/mcp/toolset/filesystem.go
+++ b/pkg/mcp/toolset/filesystem.go
@@ -9,7 +9,6 @@ import (
 	"io"
 	"os"
 	"path"
-	"path/filepath"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 
@@ -88,7 +87,7 @@ func (ts *ToolSet) WriteFile(_ context.Context,
 	if err != nil {
 		return nil, nil, err
 	}
-	dir := filepath.Dir(guestPath)
+	dir := path.Dir(guestPath)
 	err = ts.sftp.MkdirAll(dir)
 	if err != nil {
 		return nil, nil, err
--- a/pkg/mcp/toolset/toolset.go
+++ b/pkg/mcp/toolset/toolset.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"slices"
 
@@ -102,13 +103,70 @@ func (ts *ToolSet) Close() error {
 	return err
 }
 
-func (ts *ToolSet) TranslateHostPath(hostPath string) (string, error) {
+func (ts *ToolSet) TranslateHostPath(hostPath string) (guestPath string, err error) {
 	if hostPath == "" {
 		return "", errors.New("path is empty")
 	}
 	if !filepath.IsAbs(hostPath) {
 		return "", fmt.Errorf("expected an absolute path, got a relative path: %q", hostPath)
 	}
-	// TODO: make sure that hostPath is mounted
-	return hostPath, nil
+
+	guestPath, isMounted := ts.translateToGuestPath(hostPath)
+	if !isMounted {
+		return "", fmt.Errorf("path %q is not under any mounted directory", hostPath)
+	}
+	return guestPath, nil
+}
+
+func (ts *ToolSet) translateToGuestPath(hostPath string) (string, bool) {
+	cleanPath := resolvePathForMount(hostPath)
+	bestGuestPath := ""
+	bestLen := -1
+	for _, mount := range ts.inst.Config.Mounts {
+		if mount.MountPoint == nil {
+			continue
+		}
+		location := resolvePathForMount(mount.Location)
+		rel, err := filepath.Rel(location, cleanPath)
+		if err == nil && filepath.IsLocal(rel) {
+			candidate := *mount.MountPoint
+			if rel != "." {
+				candidate = path.Join(candidate, filepath.ToSlash(rel))
+			}
+			if len(location) > bestLen {
+				bestGuestPath, bestLen = candidate, len(location)
+			}
+		}
+	}
+	if bestLen >= 0 {
+		return bestGuestPath, true
+	}
+	return "", false
+}
+
+// resolvePathForMount resolves symlinks before mount path comparison.
+// If pathStr itself does not exist (e.g. write_file target), this function finds
+// the deepest existing ancestor, resolves symlinks for that ancestor, and
+// re-appends the unresolved suffix.
+func resolvePathForMount(pathStr string) string {
+	cleanPath := filepath.Clean(pathStr)
+	if resolved, err := filepath.EvalSymlinks(cleanPath); err == nil {
+		return resolved
+	}
+
+	current := cleanPath
+	tail := []string{}
+	for {
+		resolved, err := filepath.EvalSymlinks(current)
+		if err == nil {
+			parts := append([]string{resolved}, tail...)
+			return filepath.Join(parts...)
+		}
+		parent := filepath.Dir(current)
+		if parent == current {
+			return cleanPath
+		}
+		tail = append([]string{filepath.Base(current)}, tail...)
+		current = parent
+	}
 }
--- a/pkg/mcp/toolset/toolset_test.go
+++ b/pkg/mcp/toolset/toolset_test.go
@@ -1,0 +1,311 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package toolset
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"gotest.tools/v3/assert"
+
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+)
+
+func absPath(p string) string {
+	if runtime.GOOS == "windows" {
+		return "C:" + filepath.FromSlash(p)
+	}
+	return filepath.FromSlash(p)
+}
+
+func TestTranslateHostPath(t *testing.T) {
+	mountPointParent := "/mnt/parent"
+	mountPointChild := "/mnt/child"
+	mountPoint1 := "/mnt/home-user"
+	mountPoint2 := "/mnt/tmp"
+
+	type translateHostPathTest struct {
+		name          string
+		hostPath      string
+		toolSet       ToolSet
+		wantGuestPath string
+		wantErr       bool
+	}
+
+	tests := []translateHostPathTest{
+		{
+			name:     "child mount takes precedence",
+			hostPath: absPath("/parent/child/file.txt"),
+			toolSet: ToolSet{
+				inst: &limatype.Instance{
+					Config: &limatype.LimaYAML{
+						Mounts: []limatype.Mount{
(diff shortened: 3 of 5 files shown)


pkg/mcp/toolset/toolset.go
```diff
@@ -109,6 +110,25 @@ func (ts *ToolSet) TranslateHostPath(hostPath string) (string, error) {
 	if !filepath.IsAbs(hostPath) {
 		return "", fmt.Errorf("expected an absolute path, got a relative path: %q", hostPath)
 	}
-	// TODO: make sure that hostPath is mounted
+	if !ts.isMounted(hostPath) {
+		return "", fmt.Errorf("path %q is not mounted", hostPath)
```
AkihiroSuda: Callers such as `ListDirectory` and `ReadFile` have to handle this case.

Probably it should fall back to the guest path, with an annotation in the tool result to indicate how the path was resolved.

pkg/mcp/toolset/toolset.go
```diff
@@ -109,6 +110,29 @@ func (ts *ToolSet) TranslateHostPath(hostPath string) (string, error) {
 	if !filepath.IsAbs(hostPath) {
 		return "", fmt.Errorf("expected an absolute path, got a relative path: %q", hostPath)
 	}
-	// TODO: make sure that hostPath is mounted
-	return hostPath, nil
+
+	guestPath, isMounted := ts.translateToGuestPath(hostPath)
+	if !isMounted {
+		logrus.Warnf("path %q is not under any mounted directory, using as guest path", hostPath)
```
AkihiroSuda: Logging might be too verbose.

Also, doesn't this need to be visible to the AI agent?

pkg/mcp/msi/filesystem.go
```diff
@@ -35,6 +35,7 @@ type ListDirectoryResultEntry struct {
 
 type ListDirectoryResult struct {
 	Entries []ListDirectoryResultEntry `json:"entries" jsonschema:"The directory content entries."`
+	Logs    []string                   `json:"logs,omitempty" jsonschema:"Logs related to path translation and other operations."`
```
AkihiroSuda: Can we use https://modelcontextprotocol.io/specification/2025-11-25/basic#meta in https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk@v1.1.0/mcp#CallToolResult ?

pkg/mcp/toolset/toolset_test.go
AkihiroSuda, line 1: Tests are failing on Windows

Also please squash the commits

pkg/mcp/toolset/filesystem.go
```diff
@@ -41,9 +41,15 @@ func (ts *ToolSet) ListDirectory(ctx context.Context,
...
 		res.Entries[i].IsDir = ptr.Of(f.IsDir())
 	}
-	return &mcp.CallToolResult{
+	callToolRes := &mcp.CallToolResult{
 		StructuredContent: res,
-	}, res, nil
+	}
+	if logs != "" {
+		callToolRes.Meta = map[string]any{
+			"io.lima-vm/logs": []string{logs},
```
AkihiroSuda: This should be defined as a constant

AkihiroSuda:
```suggestion
			const MetaWarnings = "io.lima-vm.mcp/warnings"
```

AkihiroSuda: Also do not clear existing entries in Meta.
Such entries may not exist in the current code, but may appear in future.

pkg/mcp/toolset/toolset_test.go
AkihiroSuda: ping @Horiodino

Horiodino: got caught up with my end-semester exams, but back now

AkihiroSuda:
Could you try rebasing to see if it passes the CI?

pkg/mcp/toolset/filesystem.go
```diff
@@ -17,13 +17,15 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/ptr"
 )
 
+const MetaWarnings = "io.lima-vm/warnings"
```
jandubois: Why is this constant exported? It is not used outside the package as far as I can tell.

pkg/mcp/toolset/filesystem.go
```diff
@@ -41,9 +43,13 @@ func (ts *ToolSet) ListDirectory(ctx context.Context,
 		res.Entries[i].ModTime = ptr.Of(f.ModTime())
 		res.Entries[i].IsDir = ptr.Of(f.IsDir())
 	}
-	return &mcp.CallToolResult{
+	callToolRes := &mcp.CallToolResult{
 		StructuredContent: res,
-	}, res, nil
+	}
+	if warnings != "" {
+		callToolRes.Meta[MetaWarnings] = warnings
```
jandubois: `Meta` is a `map[string]string`, but never initialized. This code should panic, so I don't think it is covered by any tests.

The same is true for all the other places where `Meta` is being assigned to.

pkg/mcp/toolset/toolset.go
```diff
@@ -102,13 +104,38 @@ func (ts *ToolSet) Close() error {
...
 
-func (ts *ToolSet) TranslateHostPath(hostPath string) (string, error) {
+func (ts *ToolSet) TranslateHostPath(hostPath string) (guestPath, warnings string, err error) {
 	if hostPath == "" {
-		return "", errors.New("path is empty")
+		return "", "", errors.New("path is empty")
 	}
-	if !filepath.IsAbs(hostPath) {
-		return "", fmt.Errorf("expected an absolute path, got a relative path: %q", hostPath)
+	if !filepath.IsAbs(hostPath) && !strings.HasPrefix(hostPath, "/") {
```
jandubois: I don't understand the condition. On Unix the prefix condition is redundant. And on Windows a path starting with a forward slash is missing the drive letter and therefore not really an absolute path.

pkg/mcp/toolset/toolset.go
```diff
@@ -102,13 +104,38 @@ func (ts *ToolSet) Close() error {
...
+	for _, mount := range ts.inst.Config.Mounts {
+		location := filepath.Clean(mount.Location)
+		cleanPath := filepath.Clean(hostPath)
+
+		if cleanPath == location {
+			return *mount.MountPoint, true
+		}
+
+		rel, err := filepath.Rel(location, cleanPath)
+		if err == nil && !strings.HasPrefix(rel, "..") && rel != ".." {
```
jandubois: `rel` being `".."` is already covered by the `HasPrefix` test:

```suggestion
		if err == nil && !strings.HasPrefix(rel, "..") {
```

pkg/mcp/toolset/toolset.go
```diff
@@ -102,13 +104,38 @@ func (ts *ToolSet) Close() error {
...
+}
+
+func (ts *ToolSet) translateToGuestPath(hostPath string) (string, bool) {
+	for _, mount := range ts.inst.Config.Mounts {
+		location := filepath.Clean(mount.Location)
+		cleanPath := filepath.Clean(hostPath)
+
+		if cleanPath == location {
+			return *mount.MountPoint, true
+		}
```
jandubois: There is no test covering this special case.

pkg/mcp/toolset/toolset.go
```diff
@@ -102,13 +104,38 @@ func (ts *ToolSet) Close() error {
...
 	}
-	// TODO: make sure that hostPath is mounted
-	return hostPath, nil
+
+	guestPath, isMounted := ts.translateToGuestPath(hostPath)
+	if !isMounted {
+		warnings = fmt.Sprintf("path %q is not under any mounted directory, using as guest path", hostPath)
+		logrus.Info(warnings)
+	}
+	return guestPath, warnings, nil
```
jandubois: Is just logging a warning the right thing to do? When does it ever make sense to let the MCP access files inside the guest? So shouldn't this be a hard error?

pkg/mcp/toolset/toolset.go
```diff
@@ -102,13 +104,38 @@ func (ts *ToolSet) Close() error {
...
+}
+
+func (ts *ToolSet) translateToGuestPath(hostPath string) (string, bool) {
+	for _, mount := range ts.inst.Config.Mounts {
+		location := filepath.Clean(mount.Location)
+		cleanPath := filepath.Clean(hostPath)
```
jandubois: This could be moved outside the loop because the `hostPath` doesn't change.

jandubois (changes requested):
I've only looked at the code, not really thought through the whole use cases. But I feel like the warnings should be hard errors.

unsuman:
ping @Horiodino

pkg/mcp/toolset/toolset.go
Horiodino: my thought with the warning was to give the AI (and the user reading the logs) visibility into the fact that the path wasn't mounted, so the AI could understand the context but still proceed if it happened to be a valid guest file.

however, you're totally right. if it's just a warning, the AI might silently read or write to the guest filesystem

jandubois:
@Horiodino Thanks for the update. I've generated an AI review at https://jandubois.github.io/lima/20260327-102152-pr-4419.html. Please take a look at at least the important findings.

Note that the review tries to be comprehensive, please use your judgement if any of the suggestions are really worth it, or should be ignored!

AkihiroSuda:
@jandubois PTAL

jandubois:
Sorry, I didn't get around to doing a full manual review yet. I did another AI review before, but didn't post it because I thought some issues in it were incorrect. I just ran it again though, and I think at least the symlink concern is genuine: https://jandubois.github.io/lima/20260420-213558-pr-4419.html

jandubois:
I'm sorry, I know I said only the symlink issue should be fixed because I thought the guest path issue would only affect Windows, but it affects all platforms and I think must be fixed too.

> When a mount uses a distinct mountPoint (for example `/Users/me/project` mounted at `/mnt/project`), this translation is correct for SFTP, but `Glob` and `SearchFileContent` expose the resulting guest paths directly (`sftp.Glob` returns `/mnt/project/...` and `git grep` prints the same). The other MCP file tools still require host paths and now reject those results as not under any mounted directory, so agents cannot open files they just discovered unless the mount point happens to equal the host path.

Ideally we should have a BATS test that confirms that this is done correctly.

My apologies again for not paying closer attention earlier! 😞

jandubois:
@Horiodino Do you still plan to address the last round of feedback?

Horiodino:
Hey, sorry, but I’m currently tied up with other work and won’t be able to dedicate time to this right now.

jandubois:
> Hey, sorry, but I’m currently tied up with other work and won’t be able to dedicate time to this right now.

Do you expect to eventually get back to it, or do you want somebody else to finish this (if somebody else wants to pick it up before you get time again)?

Horiodino:
> > Hey, sorry, but I’m currently tied up with other work and won’t be able to dedicate time to this right now.
> 
> Do you expect to eventually get back to it, or do you want somebody else to finish this (if somebody else wants to pick it up before you get time again)?

No problem, anyone can pick it up.

PR #4421, opened 2025-12-02
fix: Remove premature return after closing sparse diff file so Convert executes.

### LINKED ISSUE #4418: pkg/driverutil.EnsureDisk: diskImageFormat is ignored for ISO mode
https://github.com/lima-vm/lima/blob/cbe724175131421dd9efc530710ca6ae28eb2d88/pkg/driverutil/disk.go#L21-L61

- - -

> diskImageFormat is ignored for ISO mode

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/4327#discussion_r2575858599_

ashwat287:
Fixes #4418 
Ensure diffDisk is converted to the requested format instead of staying sparse.

### DIFF
--- a/pkg/driverutil/disk_test.go
+++ b/pkg/driverutil/disk_test.go
@@ -1,0 +1,150 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package driverutil
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"testing"
+
+	"github.com/lima-vm/go-qcow2reader"
+	"github.com/lima-vm/go-qcow2reader/image"
+	"gotest.tools/v3/assert"
+
+	"github.com/lima-vm/lima/v2/pkg/iso9660util"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
+	"github.com/lima-vm/lima/v2/pkg/osutil"
+)
+
+const (
+	typeRAW  = image.Type("raw")
+	typeASIF = image.Type("asif")
+)
+
+func writeMinimalISO(t *testing.T, path string) {
+	t.Helper()
+	entries := []iso9660util.Entry{
+		{Path: "/hello.txt", Reader: strings.NewReader("hello world")},
+	}
+	assert.NilError(t, iso9660util.Write(path, "TESTISO", entries))
+}
+
+func writeNonISO(t *testing.T, path string) {
+	t.Helper()
+	size := 64 * 1024
+	buf := make([]byte, size)
+	copy(buf[0x8001:], "XXXXX")
+	assert.NilError(t, os.WriteFile(path, buf, 0o644))
+}
+
+func sha256File(t *testing.T, path string) string {
+	t.Helper()
+	b, err := os.ReadFile(path)
+	assert.NilError(t, err)
+	sum := sha256.Sum256(b)
+	return hex.EncodeToString(sum[:])
+}
+
+func detectImageType(t *testing.T, path string) image.Type {
+	t.Helper()
+	f, err := os.Open(path)
+	assert.NilError(t, err)
+	defer f.Close()
+	img, err := qcow2reader.Open(f)
+	assert.NilError(t, err)
+	return img.Type()
+}
+
+func checkDisk(t *testing.T, diff string, expectedType image.Type) {
+	t.Helper()
+	fi, err := os.Stat(diff)
+	assert.NilError(t, err)
+	assert.Assert(t, fi.Size() > 0)
+	assert.Equal(t, detectImageType(t, diff), expectedType)
+}
+
+func isMacOS26OrHigher() bool {
+	if runtime.GOOS != "darwin" {
+		return false
+	}
+	version, err := osutil.ProductVersion()
+	if err != nil {
+		return false
+	}
+	return version.Major >= 26
+}
+
+func TestEnsureDisk_WithISOBaseImage(t *testing.T) {
+	instDir := t.TempDir()
+	base := filepath.Join(instDir, filenames.BaseDisk)
+	diff := filepath.Join(instDir, filenames.DiffDisk)
+
+	writeMinimalISO(t, base)
+	isISO, err := iso9660util.IsISO9660(base)
+	assert.NilError(t, err)
+	assert.Assert(t, isISO)
+	baseHashBefore := sha256File(t, base)
+
+	formats := []image.Type{typeRAW}
+	if isMacOS26OrHigher() {
+		formats = append(formats, typeASIF)
+	}
+
+	for _, format := range formats {
+		assert.NilError(t, EnsureDisk(t.Context(), instDir, "2MiB", format))
+		isISO, err = iso9660util.IsISO9660(base)
+		assert.NilError(t, err)
+		assert.Assert(t, isISO)
+		assert.Equal(t, baseHashBefore, sha256File(t, base))
+		checkDisk(t, diff, format)
+		assert.NilError(t, os.Remove(diff))
+	}
+}
+
+func TestEnsureDisk_WithNonISOBaseImage(t *testing.T) {
+	instDir := t.TempDir()
+	base := filepath.Join(instDir, filenames.BaseDisk)
+	diff := filepath.Join(instDir, filenames.DiffDisk)
+
+	writeNonISO(t, base)
+	isISO, err := iso9660util.IsISO9660(base)
+	assert.NilError(t, err)
+	assert.Assert(t, !isISO)
+
+	formats := []image.Type{typeRAW}
+	if isMacOS26OrHigher() {
+		formats = append(formats, typeASIF)
+	}
+
+	for _, format := range formats {
+		assert.NilError(t, EnsureDisk(t.Context(), instDir, "2MiB", format))
+		checkDisk(t, diff, format)
+		assert.NilError(t, os.Remove(diff))
+	}
+}
+
+func TestEnsureDisk_ExistingDiffDisk(t *testing.T) {
+	instDir := t.TempDir()
+	base := filepath.Join(instDir, filenames.BaseDisk)
+	diff := filepath.Join(instDir, filenames.DiffDisk)
+
+	writeNonISO(t, base)
+
+	formats := []image.Type{typeRAW}
+	if isMacOS26OrHigher() {
+		formats = append(formats, typeASIF)
+	}
+
+	for _, format := range formats {
+		assert.NilError(t, os.WriteFile(diff, []byte("preexisting"), 0o644))
+		origHash := sha256File(t, diff)
+		assert.NilError(t, EnsureDisk(t.Context(), instDir, "2MiB", format))
+		assert.Equal(t, sha256File(t, diff), origHash)
+		assert.NilError(t, os.Remove(diff))
+	}
+}


pkg/driverutil/disk.go
```diff
@@ -50,7 +50,7 @@ func EnsureDisk(ctx context.Context, instDir, diskSize string, diskImageFormat i
 			diffDiskF.Close()
 			return fmt.Errorf("failed to create sparse diff disk %q: %w", diffDisk, err)
 		}
-		return diffDiskF.Close()
+		diffDiskF.Close()
```
AkihiroSuda: How did you test this PR?
The `diffDisk` created in L43  will be overwritten by a new disk converted from `baseDisk` in L57, no?

ashwat287: Yes diffDisk would get converted in L57 when the base image is ISO.
I have include the tests to verify the conditions when base image is ISO and non-ISO
Can you please clarify on what should be the desired behaviors in each test cases?

AkihiroSuda: When the base disk is an ISO9660 image, the diff disk has to be a new empty disk.
("diff" is misnomer here)

ashwat287: The diff disk when created as a sparse file in L49 is empty and has size 0 bytes.
On debugging, I find that after conversion the diffdisk size becomes equal to `diskSizeInBytes`.

To keep the `diffDisk` empty we can either:
1. not convert the `diffDisk` and leave it as a sparse file and eventually ignore `diskImageFormat ` when the base disk is ISO. or,
2. convert the `diffDisk` to the available formats while keeping `diskSizeInBytes=0` in case of ISO base disk. The problem here arises when the disk is being converted to `asif` format, the following error gets thrown `specified size 0 is smaller than the original image size` because of https://github.com/lima-vm/lima/blob/9dbf5727d1277ab0bedba7dc8918941b131308cd/pkg/imgutil/nativeimgutil/nativeimgutil.go#L59-L61

norio-nomura (commented):
This PR does not fix #4418:
```console
$ limactl start ./hack/test-templates/alpine-iso-9p-writable.yaml --vm-type vz --set .vmOpts.vz.diskimageFormat=\"asif\" --set .mountType=null --tty=false --log-level fatal
$ limactl stop alpine-iso-9p-writable
INFO[0000] Sending SIGINT to hostagent process 62151    
INFO[0003] [hostagent] [VZ] - vm state change: stopped  
ERRO[0003] [hostagent] accept tcp 127.0.0.1:58662: use of closed network connection 
INFO[0004] Waiting for the instance to shut down        
INFO[0004] The instance alpine-iso-9p-writable has shut down 
$ diskutil image info ~/.lima/alpine-iso-9p-writable/diffdisk
Image Format: RAW
Format Description: RAW read-write image

	Size Info:
		Empty Bytes: 0
		Min Size Bytes: 512
		Sector Count: 209715200
		Total Bytes: 107374182400

	Encryption Info:
		Is Encrypted: 0

	Partitions
		0:
			content-hint: GUID_partition_scheme
			total-space: 107374182400
			used-space: N/A
			volume-name: N/A
		1:
			content-hint: Microsoft Basic Data
			total-space: 167936
			used-space: N/A
			volume-name: N/A
		2:
			content-hint: EFI
			total-space: 1474560
			used-space: N/A
			volume-name: NO NAME
		3:
			content-hint: Microsoft Basic Data
			total-space: 81270784
			used-space: N/A
			volume-name: N/A
$ limactl --version
limactl version 2.0.2-25-gea17ac7c.m
```

norio-nomura:
> This PR does not fix #4418:

updated console log to add missing ` --set .mountType=null` on `limactl start`

ashwat287:
```
$ limactl start ./hack/test-templates/alpine-iso-9p-writable.yaml --vm-type vz --set .vmOpts.vz.diskImageFormat=\"asif\" --set .mountType=null --tty=false --log-level fatal
$ diskutil image info ~/.lima/alpine-iso-9p-writable/diffdisk                                                                                                                     
Image Format: ASIF
Format Description: Apple sparse image

	Identity Info:
		Stable UUID: F240DA64-A9D1-44B2-841C-1BABD6F3CE17
		UUID: 8C2D97C2-80AE-4545-979D-FC4DE3DCF943

	Size Info:
		Empty Bytes: 107303096320
		Max Size Bytes: 4503599626321920
		Min Size Bytes: 1048576
		Sector Count: 209715200
		Total Bytes: 107374182400

	Encryption Info:
		Is Encrypted: 0

	ASIF Info:
		Bitmapped Entries: 83
		Chunk Size: 1048576
		Dir Pointer Version: 4
		Full Entries: 62
		Is Cache: 0
		Max Sector Count: 8796093022208
		Num Tables: 2
		Uninitialized Entries: 155647
		Unmapped Entries: 102256

	Partitions
		0:
			content-hint: FDisk_partition_scheme
			total-space: 107374182400
			used-space: N/A
			volume-name: N/A
		1:
			content-hint: Linux
			total-space: 107373133824
			used-space: N/A
			volume-name: N/A
```

pkg/driverutil/disk.go
```diff
@@ -38,23 +38,25 @@ func EnsureDisk(ctx context.Context, instDir, diskSize string, diskImageFormat i
...
 	if isBaseDiskISO {
+		destDisk = diffDisk
 		// Create an empty data volume (sparse)
 		diffDiskF, err := os.Create(diffDisk)
 		if err != nil {
 			return err
 		}
 
-		err = diskUtil.MakeSparse(ctx, diffDiskF, 0)
+		err = diskUtil.MakeSparse(ctx, diffDiskF, diskSizeInBytes)
```
norio-nomura: Is this `MakeSparse` still needed?

norio-nomura (approved):
Confirmed that #4418 will be corrected.
I don't know if the call to `MakeSparse` needs to be left, but I don't think it's a problem.

pkg/driverutil/disk.go
ashwat287: Nope, just an empty volume is enough. I have removed it in the amended commit.

pkg/driverutil/disk.go
```diff
@@ -38,23 +38,21 @@ func EnsureDisk(ctx context.Context, instDir, diskSize string, diskImageFormat i
...
 			return err
 		}
 
-		err = diskUtil.MakeSparse(ctx, diffDiskF, 0)
-		if err != nil {
-			diffDiskF.Close()
-			return fmt.Errorf("failed to create sparse diff disk %q: %w", diffDisk, err)
-		}
-		return diffDiskF.Close()
+		diffDiskF.Close()
```
AkihiroSuda: err is unchecked

pkg/driverutil/disk_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+const (
+	typeRAW  = image.Type("raw")
+	typeASIF = image.Type("asif")
+)
+
+func makeTempInstanceDir(t *testing.T) string {
+	t.Helper()
+	return t.TempDir()
+}
```
AkihiroSuda: This function seems useless. You can just use `t.TempDir`.

pkg/driverutil/disk_test.go
```diff
@@ -0,0 +1,175 @@
...
+	assert.Assert(t, fi.Size() > 0)
+	assert.Equal(t, detectImageType(t, diff), expectedType)
+}
+
+func removeAndCheck(t *testing.T, path string) {
+	t.Helper()
+	assert.NilError(t, os.Remove(path))
+	_, err := os.Stat(path)
+	assert.ErrorIs(t, err, os.ErrNotExist)
+}
```
AkihiroSuda: This function seems useless, why not just `os.Remote` it?

pkg/driverutil/disk_test.go
```diff
@@ -0,0 +1,175 @@
...
+
+func isMacOS26OrHigher(t *testing.T) bool {
+	if runtime.GOOS != "darwin" {
+		return false
+	}
+	out, err := exec.CommandContext(t.Context(), "sw_vers", "-productVersion").Output()
```
AkihiroSuda: Use
https://github.com/lima-vm/lima/blob/ae9093895c668b47f66ddf4cc5cb855697a7c2ad/pkg/osutil/osversion_darwin.go#L16

pkg/imgutil/nativeimgutil/nativeimgutil.go
```diff
@@ -62,10 +62,10 @@ func convertTo(destType image.Type, source, dest string, size *int64, allowSourc
 	logrus.Infof("Converting %q (%s) to a %s disk %q", source, srcImg.Type(), destType, dest)
 	switch t := srcImg.Type(); t {
 	case raw.Type:
-		if err = srcF.Close(); err != nil {
-			return err
-		}
```
AkihiroSuda, line 67: ?

pkg/driverutil/disk_test.go
ashwat287: `os.Remove` returning `nil` will only ensure that that the syscall didn't report an error. But it doesn't really asserts that the file is deleted.
By calling `os.Stat` and asserting `os.ErrNotExist` confirms that it is not present in the filesystem.

pkg/imgutil/nativeimgutil/nativeimgutil.go
ashwat287: https://github.com/lima-vm/lima/blob/ae9093895c668b47f66ddf4cc5cb855697a7c2ad/pkg/imgutil/nativeimgutil/nativeimgutil.go#L128
`srcImg.size()` was giving the value `-1`. internally it was not able to run the syscall `os.stat` because it was being closed early when the close statement was outside the if statement. This change ensures that the the file is open when trying to get `scrImg.Size()` and that the value is not `-1`.
https://github.com/lima-vm/lima/blob/ae9093895c668b47f66ddf4cc5cb855697a7c2ad/pkg/imgutil/nativeimgutil/nativeimgutil.go#L123 and https://github.com/lima-vm/lima/blob/ae9093895c668b47f66ddf4cc5cb855697a7c2ad/pkg/imgutil/nativeimgutil/nativeimgutil.go#L54
The above defer statements already take care of closing the file after processing.

pkg/driverutil/disk_test.go
AkihiroSuda: >  But it doesn't really asserts that the file is deleted.

Did you actually observe such a behavior? On what filesystem?

ashwat287: No, I haven't observed such behaviour. And yes, now that i have checked again, I feel that the additional assert by `os.Stat` is redundant here
> But it doesn't really asserts that the file is deleted.

Here i meant that before the process ends the the file might get created again after deletion, not that it doesn't get deleted by `os.remove` at the first place. According to me this can happen if there is any race condition or when the filesystem is buggy, which I feel is a very specific case and doesn't apply here.

I will remove the function `removeAndCheck` and just keep `os.remove`

pkg/driverutil/disk_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+func checkDisk(t *testing.T, diff string, expectedType image.Type) {
+	t.Helper()
+	fi, err := os.Stat(diff)
+	assert.NilError(t, err)
+	assert.Assert(t, fi.Size() > 0)
+	assert.Equal(t, detectImageType(t, diff), expectedType)
+}
+
+var ProductVersion = sync.OnceValues(func() (*semver.Version, error) {
```
AkihiroSuda: Why duplicate
https://github.com/lima-vm/lima/blob/ae9093895c668b47f66ddf4cc5cb855697a7c2ad/pkg/osutil/osversion_darwin.go#L16

ashwat287: Sorry, I missed that it already exists 😵‍💫

pkg/driverutil/disk.go
```diff
@@ -38,23 +38,23 @@ func EnsureDisk(ctx context.Context, instDir, diskSize string, diskImageFormat i
...
-			return fmt.Errorf("failed to create sparse diff disk %q: %w", diffDisk, err)
+		if err = diffDiskF.Close(); err != nil {
+			return err
 		}
-		return diffDiskF.Close()
 	}
 	// Check whether to use ASIF format
 
-	if err = diskUtil.Convert(ctx, diskImageFormat, baseDisk, diffDisk, &diskSizeInBytes, false); err != nil {
+	if err = diskUtil.Convert(ctx, diskImageFormat, destDisk, diffDisk, &diskSizeInBytes, false); err != nil {
```
AkihiroSuda, line 57: This code is hard to read, as `destDisk` is passed as `source`

https://github.com/lima-vm/lima/blob/77d181ab2ec1ac6e43fdf0c146fa926302ea965f/pkg/imgutil/manager.go#L21-L22

pkg/driverutil/disk.go
```diff
@@ -38,23 +38,23 @@ func EnsureDisk(ctx context.Context, instDir, diskSize string, diskImageFormat i
...
 		return fmt.Errorf("failed to convert %q to a disk %q: %w", baseDisk, diffDisk, err)
```
AkihiroSuda, line 58: This error message has to be updated too

pkg/driverutil/disk.go
AkihiroSuda: - https://github.com/lima-vm/lima/pull/4463

PR #4429, opened 2025-12-04
Feature(cli): Add `--sync` to prevent AI agents from breaking the host files

### LINKED ISSUE #3711: Add `limactl shell --sync-host-workdir` (prevents AI agents from breaking the host files)
`limactl shell --sync-host-workdir INSTANCE COMMAND...` will do:
- copy the host workdir to the VM, and set the copied dir as the guest workdir
- run the commands
- sync back the changed content of the workdir (if the user accepts the changes)

This is quite similar to how https://github.com/AkihiroSuda/alcless works:

```console
$ alcless bash -euxc "echo foo > foo"
5:56PM INF ➡️ Syncing the files src=/Users/suda/foo/ dst=default:/Users/alcless_suda_default/Users/suda/foo
+ echo foo
5:56PM INF ⬅️ Syncing the files back (dry run) src=default:/Users/alcless_suda_default/Users/suda/foo/ dst=/Users/suda/foo
>f..t.... foo
5:56PM INF ⬅️ Syncing the files back src=default:/Users/alcless_suda_default/Users/suda/foo/ dst=/Users/suda/foo
⚠️   The following commands will be executed:
rsync -rai --delete -e '/Users/suda/.local/bin/alclessctl shell --workdir=/ --plain' default:/Users/alcless_suda_default/Users/suda/foo/ /Users/suda/foo
❓ Press return to continue, or Ctrl-C to abort

CONTINUE
>f..t.... foo
```

## Implementation

As in https://github.com/AkihiroSuda/alcless, rsync can be used.

The workdir must not be a large directory such as the home directory.
It must not be a mounted directory too.

## Use case

The expected use case is to run AI agents such as `codex`, `gemini`, and `claude` inside the VM so as to avoid breaking the host files.

```console
$ limactl start --mount-none template://claude
$ cd ~/some-project
$ limactl shell --sync-host-workdir claude claude
Changed files:
- foo.go
- Makefile

⚠️Accept the changes?
→ Yes
  No
  View the changed contents
```

unsuman:
This PR adds a `--sync` flag to the `limactl shell` command that allows users to safely run AI commands inside VMs by syncing the host's working directory to the guest and optionally syncing changes back to the host after confirmation. Also added cleanup logic to remove the guest's synced workdir after user decision, ensuring no leftover files in the VM.

### How to test
```zsh
$ limactl start --mount-none
$ cd ~/some-project
$ limactl shell --sync <AI Agent Command>
OR
$ limactl shell --sync # simply shell into the instance and make changes
INFO[0000] Syncing host current directory(/Users/ansumansahoo/temp) to guest instance... 
INFO[0000] Successfully synced host current directory to guest(~/synced-workdir/temp) instance. 
ansumansahoo@lima-default:~/synced-workdir/temp$ vi source_dir/initials.txt 
ansumansahoo@lima-default:~/synced-workdir/temp$ exit
logout
? ⚠️ Accept the changes?  [Use arrows to move, type to filter]
  Yes
  No
> View the changed contents
```
When the program prompts the user to view the changes, a rsync is performed. This copies the guest synced directory to a temporary directory on the host.  This allows diff to be used and the user to see a detailed list of changes.
Fixes #3711

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
@@ -28,6 +30,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/networks/reconcile"
 	"github.com/lima-vm/lima/v2/pkg/sshutil"
 	"github.com/lima-vm/lima/v2/pkg/store"
+	"github.com/lima-vm/lima/v2/pkg/uiutil"
 )
 
 const shellHelp = `Execute shell in Lima
@@ -65,8 +68,13 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("reconnect", false, "Reconnect to the SSH session")
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
+	shellCmd.Flags().String("sync", "", "Copy a host directory to the guest and vice-versa upon exit")
+
 	return shellCmd
 }
+
+// Depth of "/Users/USER" is 3.
+const rsyncMinimumSrcDirDepth = 4
 
 func shellAction(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
@@ -161,29 +169,57 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	syncDirVal, err := flags.GetString("sync")
+	if err != nil {
+		return fmt.Errorf("failed to get sync flag: %w", err)
+	}
+	syncHostWorkdir := syncDirVal != ""
+	if syncHostWorkdir && len(inst.Config.Mounts) > 0 {
+		return errors.New("cannot use `--sync` when the instance has host mounts configured, start the instance with `--mount-none` to disable mounts")
+	}
+
 	// When workDir is explicitly set, the shell MUST have workDir as the cwd, or exit with an error.
 	//
 	// changeDirCmd := "cd workDir || exit 1"                  if workDir != ""
 	//              := "cd hostCurrentDir || cd hostHomeDir"   if workDir == ""
 	var changeDirCmd string
+	var hostCurrentDir string
+	if syncDirVal != "" {
+		hostCurrentDir, err = filepath.Abs(syncDirVal)
+		if err == nil && runtime.GOOS == "windows" {
+			hostCurrentDir, err = mountDirFromWindowsDir(ctx, inst, hostCurrentDir)
+		}
+	} else {
+		hostCurrentDir, err = hostCurrentDirectory(ctx, inst)
+	}
+
+	if err != nil {
+		changeDirCmd = "false"
+		logrus.WithError(err).Warn("failed to get the current directory")
+	}
+	if syncHostWorkdir {
+		if _, err := exec.LookPath("rsync"); err != nil {
+			return fmt.Errorf("rsync is required for `--sync` but not found: %w", err)
+		}
+
+		srcWdDepth := len(strings.Split(hostCurrentDir, string(os.PathSeparator)))
+		if srcWdDepth < rsyncMinimumSrcDirDepth {
+			return fmt.Errorf("expected the depth of the host working directory (%q) to be more than %d, only got %d (Hint: %s)",
+				hostCurrentDir, rsyncMinimumSrcDirDepth, srcWdDepth, "cd to a deeper directory")
+		}
+	}
+
+	var destRsyncDir string
 	workDir, err := cmd.Flags().GetString("workdir")
 	if err != nil {
 		return err
 	}
-	if workDir != "" {
+	switch {
+	case workDir != "":
 		changeDirCmd = fmt.Sprintf("cd %s || exit 1", shellescape.Quote(workDir))
 		// FIXME: check whether y.Mounts contains the home, not just len > 0
-	} else if len(inst.Config.Mounts) > 0 || inst.VMType == limatype.WSL2 {
-		hostCurrentDir, err := os.Getwd()
-		if err == nil && runtime.GOOS == "windows" {
-			hostCurrentDir, err = mountDirFromWindowsDir(ctx, inst, hostCurrentDir)
-		}
-		if err == nil {
-			changeDirCmd = fmt.Sprintf("cd %s", shellescape.Quote(hostCurrentDir))
-		} else {
-			changeDirCmd = "false"
-			logrus.WithError(err).Warn("failed to get the current directory")
-		}
+	case len(inst.Config.Mounts) > 0 || inst.VMType == limatype.WSL2:
+		changeDirCmd = fmt.Sprintf("cd %s", shellescape.Quote(hostCurrentDir))
 		hostHomeDir, err := os.UserHomeDir()
 		if err == nil && runtime.GOOS == "windows" {
 			hostHomeDir, err = mountDirFromWindowsDir(ctx, inst, hostHomeDir)
@@ -193,7 +229,10 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		} else {
 			logrus.WithError(err).Warn("failed to get the home directory")
 		}
-	} else {
+	case syncHostWorkdir:
+		destRsyncDir = *inst.Config.User.Home + hostCurrentDir
+		changeDirCmd = fmt.Sprintf("cd %s", shellescape.Quote(destRsyncDir))
+	default:
 		logrus.Debug("the host home does not seem mounted, so the guest shell will have a different cwd")
 	}
 
@@ -278,6 +317,34 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	sshArgs := append([]string{}, sshExe.Args...)
 	sshArgs = append(sshArgs, sshutil.SSHArgsFromOpts(sshOpts)...)
+
+	var sshExecForRsync *exec.Cmd
+	if syncHostWorkdir {
+		logrus.Infof("Syncing host current directory(%s) to guest instance...", hostCurrentDir)
+		sshExecForRsync = exec.CommandContext(ctx, sshExe.Exe, sshArgs...)
+
+		// Create the destination directory in the guest instance,
+		// we could have done this by using `--rsync-path` but it's more
+		// complex to quote properly.
+		if err := executeSSHForRsync(ctx, *sshExecForRsync, inst.SSHLocalPort, inst.SSHAddress, fmt.Sprintf("mkdir -p %s", shellescape.Quote(destRsyncDir))); err != nil {
+			return fmt.Errorf("failed to create the synced workdir in guest instance: %w", err)
+		}
+
+		// The macOS release of rsync (the latest being 2.6.9) does not support shell escaping of destination path but other versions do.
+		rsyncVer, err := rsyncVersion(ctx)
+		if err != nil {
+			return fmt.Errorf("failed to get rsync version: %w", err)
+		}
+		if rsyncVer.LessThan(*semver.New("3.0.0")) {
+			destRsyncDir = shellescape.Quote(destRsyncDir)
+		}
+
+		if err := rsyncDirectory(ctx, cmd, sshExecForRsync, hostCurrentDir+"/", fmt.Sprintf("%s:%s", *inst.Config.User.Name+"@"+inst.SSHAddress, destRsyncDir)); err != nil {
+			return fmt.Errorf("failed to sync host working directory to guest instance: %w", err)
+		}
+		logrus.Infof("Successfully synced host current directory to guest(%s) instance.", destRsyncDir)
+	}
+
 	if isatty.IsTerminal(os.Stdout.Fd()) || isatty.IsCygwinTerminal(os.Stdout.Fd()) {
 		// required for showing the shell prompt: https://stackoverflow.com/a/626574
 		sshArgs = append(sshArgs, "-t")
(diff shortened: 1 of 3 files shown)


cmd/limactl/shell.go
```diff
@@ -64,9 +66,15 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("reconnect", false, "Reconnect to the SSH session")
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
+	shellCmd.Flags().Bool("sync-host-workdir", false, "Copy the host working directory to the guest to run AI commands inside VMs (prevents AI agents from breaking the host files)")
```
AkihiroSuda: - For consistency with `--mount .`, probably this should be like `--sync .`
- The flag is not really specific to AI commands. Use case with AI should be mentioned in https://lima-vm.io/docs/examples/ai/ though.

cmd/limactl/shell.go
```diff
@@ -64,9 +66,15 @@ func newShellCommand() *cobra.Command {
...
 	return shellCmd
 }
 
+const (
+	rsyncMinimumSrcDirDepth = 4 // Depth of "/Users/USER" is 3.
+	guestSyncedWorkdir      = "~/synced-workdir"
```
AkihiroSuda: Is there any reason that we cannot just use the same path as the host dir?
Is that for avoiding conflicts with mounts?

cmd/limactl/shell.go
```diff
@@ -296,7 +333,138 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+		}
+	}()
+
+	if err := rsyncDirectory(ctx, sshCmd, remoteSource, hostTmpDest); err != nil {
+		logrus.WithError(err).Warn("Failed to sync back the changes to host for viewing")
+		return
+	}
+
+	for {
+		ans, err := uiutil.Select(message, options)
```
AkihiroSuda, line 440: rsync can be slow and may eat up the host disk space, so this prompt should be shown before running rsync to the tmp dir

cmd/limactl/shell.go
```diff
@@ -296,7 +333,138 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+			cleanGuestSyncedWorkdir(ctx, sshCmd)
+			logrus.Info("Successfully synced back the changes to host.")
+			return
+		case 1: // No
+			cleanGuestSyncedWorkdir(ctx, sshCmd)
+			logrus.Info("Skipping syncing back the changes to host.")
+			return
+		case 2: // View the changed contents
+			diffCmd := exec.CommandContext(ctx, "diff", "-ru", "--color=always", hostCurrentDir, filepath.Join(hostTmpDest, filepath.Base(hostCurrentDir)))
+			lessCmd := exec.CommandContext(ctx, "less", "-R")
```
AkihiroSuda: Respect `$PAGER`

cmd/limactl/shell.go
AkihiroSuda, line 1: Needs bats tests and docs

cmd/limactl/shell.go
```diff
@@ -150,29 +158,45 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	syncHostWorkdir, err := flags.GetBool("sync-host-workdir")
+	if err != nil {
+		return fmt.Errorf("failed to get sync-host-workdir flag: %w", err)
+	} else if syncHostWorkdir && len(inst.Config.Mounts) > 0 {
+		return errors.New("cannot use `--sync-host-workdir` when the instance has host mounts configured, use `--mount-none` to disable mounts")
```
AkihiroSuda:
```suggestion
		return errors.New("cannot use `--sync-host-workdir` when the instance has host mounts configured, start the instance with `--mount-none` to disable mounts")
```

cmd/limactl/shell.go
```diff
@@ -182,7 +206,9 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		} else {
 			logrus.WithError(err).Warn("failed to get the home directory")
 		}
-	} else {
+	case syncHostWorkdir:
+		changeDirCmd = fmt.Sprintf("cd %s/%s", guestSyncedWorkdir, filepath.Base(hostCurrentDir))
```
AkihiroSuda: Make sure to quote the path, as it may contain spaces and quote symbols

AkihiroSuda: (Such paths should be tested in bats)

cmd/limactl/shell.go
```diff
@@ -296,7 +333,138 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+func rsyncDirectory(ctx context.Context, sshCmd *exec.Cmd, source, destination string) error {
+	rsyncArgs := []string{
+		"-ah",
+		"-e", sshCmd.String(),
+		source,
+		destination,
+	}
+	rsyncCmd := exec.CommandContext(ctx, "rsync", rsyncArgs...)
+	rsyncCmd.Stdin = os.Stdin
+	rsyncCmd.Stdout = os.Stdout
```
AkihiroSuda: Use https://pkg.go.dev/github.com/spf13/cobra#Command.OutOrStdout

cmd/limactl/shell.go
```diff
@@ -296,7 +333,138 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+// of the host's current working directory into it.
+func rsyncDirectory(ctx context.Context, sshCmd *exec.Cmd, source, destination string) error {
+	rsyncArgs := []string{
+		"-ah",
+		"-e", sshCmd.String(),
+		source,
+		destination,
+	}
+	rsyncCmd := exec.CommandContext(ctx, "rsync", rsyncArgs...)
+	rsyncCmd.Stdin = os.Stdin
```
AkihiroSuda: For what ?

cmd/limactl/shell.go
unsuman: > Is there any reason that we cannot just use the same path as the host dir?

Yes, issue with `rsync` because it only tries to create the base of the path and not the full path so using for example `ansumansahoo@127.0.0.1:~/Users/ansumansahoo/Documents/GOLANG/lima` as a destination path will result into an error `rsync: [Receiver] mkdir "/home/ansumansahoo.linux/Users/ansumansahoo/Documents/GOLANG/lima" failed: No such file or directory (2)`

Or do you mean to use only `ansumansahoo@127.0.0.1:~/lima` regarding this context?

cmd/limactl/shell.go
unsuman: NVM, it was a mistake!

unsuman:
Left with adding bats test and docs.

cmd/limactl/shell.go
AkihiroSuda: > it only tries to create the base of the path and not the full path 

Why not run `mkdir -p`

unsuman:
How can I test existing bats test files?
I tried to run `make bats` first and it always results into 0 tests and 0 failures:
```bash
➜  lima2 git:(feat/rsync) ✗ ./hack/bats/lib/bats-core/bin/bats ./hack/bats/tests/list.bats

0 tests, 0 failures

➜  lima2 git:(feat/rsync) ✗ ./hack/bats/lib/bats-core/bin/bats --version
Bats 1.12.0

➜  lima2 git:(feat/rsync) ✗ lm -v
limactl version 2.0.1-122-g997e1287.m
```
[EDIT]: Ran a sample bats test and it works:
```bash
#!/usr/bin/env bats
@test "simple test" {
    run echo "hello"
    [ "$status" -eq 0 ]
    [ "$output" = "hello" ]
}
```

```bash
➜  lima2 git:(feat/rsync) ✗ ./hack/bats/lib/bats-core/bin/bats test-simple.bats                                                    
test-simple.bats
 ✓ simple test

1 test, 0 failures
```

unsuman:
> always results into 0 tests and 0 failures

This was due to the fact that my Bash version which comes out of the box on MacOS was 3.2 which was silently failing because Bash 3.2 does not support associative arrays(introduced in 4.0). Had to upgrade it using brew and now it works!
https://github.com/lima-vm/lima/blob/d8828f434382b1b1ce04e3d36c53e141969dcf5b/hack/bats/helpers/load.bash#L81

cmd/limactl/shell.go
```diff
@@ -306,7 +333,158 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+					return
+				}
+				rsyncToTempDir = true
+			}
+			diffCmd := exec.CommandContext(ctx, "diff", "-ru", "--color=always", hostCurrentDir, filepath.Join(hostTmpDest, filepath.Base(hostCurrentDir)))
+			pager := os.Getenv("PAGER")
+			if pager == "" {
+				pager = "less"
+			}
+			lessCmd := exec.CommandContext(ctx, pager, "-R")
```
AkihiroSuda: Not every PAGER supports this flag

website/content/en/docs/examples/ai.md
```diff
@@ -116,6 +116,56 @@ See also <https://github.com/github/copilot-cli>.
 {{% /tab %}}
 {{< /tabpane >}}
 
+
+# Syncing Working Directory
+
```
AkihiroSuda, line 134: Add the version requirement (2.1)

website/content/en/docs/examples/ai.md
```diff
@@ -116,6 +116,56 @@ See also <https://github.com/github/copilot-cli>.
...
+The `--sync` flag for `limactl shell` enables bidirectional synchronization of your host working directory with the guest VM. This is particularly useful when running AI agents (like Claude, Copilot, or Gemini) inside VMs to prevent them from accidentally modifying or breaking files on your host system.
+
```
AkihiroSuda, line 148: Probably needs a table to compare mount vs sync

cmd/limactl/shell.go
```diff
@@ -278,6 +303,28 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+		sshExecForRsync = exec.CommandContext(ctx, sshExe.Exe, sshArgs...)
+
+		// Create the destination directory in the guest instance,
+		// we could have done this by using `--rsync-path` but it's more
+		// complex to quote properly.
+		if err := executeSSHForRsync(ctx, sshExecForRsync, fmt.Sprintf("mkdir -p %s", shellescape.Quote(destRsyncDir))); err != nil {
+			return fmt.Errorf("failed to create the synced workdir in guest instance: %w", err)
+		}
+
+		if runtime.GOOS == "darwin" {
```
AkihiroSuda: ?

website/content/en/docs/examples/ai.md
```diff
@@ -116,6 +116,67 @@ See also <https://github.com/github/copilot-cli>.
...
 {{< /tabpane >}}
 
+
+# Syncing Working Directory
+
+ | ⚡ Requirement | Lima >= 2.1 |
+ |----------------|-------------|
+
+The `--sync` flag for `limactl shell` enables bidirectional synchronization of your host working directory with the guest VM. This is particularly useful when running AI agents (like Claude, Copilot, or Gemini) inside VMs to prevent them from accidentally modifying or breaking files on your host system.
+
```
AkihiroSuda, line 148: The comparison table should be shown before the example commands

website/content/en/docs/examples/ai.md
```diff
@@ -116,6 +116,67 @@ See also <https://github.com/github/copilot-cli>.
...
+### Usecase - Running AI Code Assistants Safely
+
+1. Create an isolated instance for AI agents which must be started without host mounts for `--sync` to work:
+
+```bash
+limactl start --name=ai-sandbox --mount-none template://default
```
AkihiroSuda: `--name` should not be needed

cmd/limactl/shell.go
```diff
@@ -307,7 +354,173 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+	sshCmdStr := strings.Join(sshCmdParts, " ")
+
+	rsyncArgs := []string{
+		"-ah",
+		"--delete",
+		"-e", sshCmdStr,
+		source,
+		destination,
+	}
+	rsyncCmd := exec.CommandContext(ctx, "rsync", rsyncArgs...)
```
AkihiroSuda, line 555: Code clones should be reduced

https://github.com/lima-vm/lima/blob/71827e19e671fd3467ac3e1455fda4719548d400/cmd/limactl/copy.go#L353

cmd/limactl/shell.go
unsuman: MacOS’s version of `rsync` (the latest being 2.6.9) doesn’t handle shell escaping itself.  However, Linux versions of `rsync` (specifically 3.x) efficiently manage this. A shell-escaped path in a Linux environment caused path not found issues.
See [this](https://appleinsider.com/inside/macos-sequoia/tips/what-you-should-know-about-apples-switch-from-rsync-to-openrsync#:~:text=Apple%20has%20long%20included%20rsync%20in%20macOS%2C%20but%20the%20last%20official%20version%20included%20is%202.6.9.%20The%20reason%20is%20that%20the%20version%20was%20released%20under%20the%20GPL%202.0%20license%2C%20which%20is%20somewhat%20less%20restrictive%20than%20GPL%203.0.)

cmd/limactl/shell.go
unsuman: Yes, I think we should make a separate `package rsync` same as https://github.com/AkihiroSuda/alcless/blob/master/pkg/rsync/rsync.go
but this can be done by me in another PR?

cmd/limactl/shell.go
unsuman: Better to add a comment

AkihiroSuda: Shouldn’t it check the version, not OS?

cmd/limactl/shell.go
```diff
@@ -65,9 +67,13 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("reconnect", false, "Reconnect to the SSH session")
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
+	shellCmd.Flags().Bool("sync", false, "Copy the host working directory to the guest and vice-versa upon exit")
```
AkihiroSuda: As in `--mount`, can we allow passing an arbitrary path? Typically it should be `.` though

cmd/limactl/shell.go
```diff
@@ -307,7 +361,188 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+		}
+		logrus.Info("Successfully synced back the changes to host.")
+	}
+
+	if !tty {
+		rsyncBackAndCleanup()
+		return
+	}
+
+	message := "⚠️ Accept the changes?"
```
AkihiroSuda, line 421: The changes should be briefly explained
e.g.,
```
⚠️ Accept the changes? (added: 42, deleted: 5, modified: 123)
```

Can be another PR

cmd/limactl/shell.go
```diff
@@ -65,9 +67,15 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("reconnect", false, "Reconnect to the SSH session")
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
+	shellCmd.Flags().String("sync", "", "Copy the host working directory to the guest and vice-versa upon exit")
+	shellCmd.Flags().Lookup("sync").NoOptDefVal = "."
```
AkihiroSuda: Not sure if we want `NoOptDefVal`. 
Probably this one should be omitted from this PR, and should be revisited in a separate issue.

cmd/limactl/shell.go
```diff
@@ -65,9 +67,15 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("reconnect", false, "Reconnect to the SSH session")
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
+	shellCmd.Flags().String("sync", "", "Copy the host working directory to the guest and vice-versa upon exit")
```
AkihiroSuda:
```suggestion
	shellCmd.Flags().String("sync", "", "Copy a host directory to the guest and vice-versa upon exit")
```

website/content/en/docs/examples/ai.md
```diff
@@ -114,20 +114,68 @@ lima vi "/home/${USER}.linux/.bash_profile"
...
 ```
 
-Type `/connect` in the first session for authentication.
-Unlike other agents, this step is not necessary for OpenCode.
+3. Run an AI agent that modifies code:
 
-See also <https://github.com/anomalyco/opencode>.
-{{% /tab %}}
+```bash
+limactl shell --sync default claude "Add error handling to all functions"
```
AkihiroSuda:
```suggestion
limactl shell --sync . default claude "Add error handling to all functions"
```

website/content/en/docs/examples/ai.md
```diff
@@ -114,20 +114,68 @@ lima vi "/home/${USER}.linux/.bash_profile"
...
+```
 
-{{< /tabpane >}}
+Or simply shell into the instance and make changes:
+```bash
+limactl shell --sync default
```
AkihiroSuda: same as above

cmd/limactl/shell.go
```diff
@@ -161,29 +168,57 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+		return errors.New("cannot use `--sync` when the instance has host mounts configured, start the instance with `--mount-none` to disable mounts")
+	}
+
 	// When workDir is explicitly set, the shell MUST have workDir as the cwd, or exit with an error.
 	//
 	// changeDirCmd := "cd workDir || exit 1"                  if workDir != ""
 	//              := "cd hostCurrentDir || cd hostHomeDir"   if workDir == ""
 	var changeDirCmd string
+	var hostCurrentDir string
+	if syncDirVal != "" && syncDirVal != "." {
```
AkihiroSuda: Why need to check `syncDirVal != "."` here?

hack/bats/tests/shell-sync.bats
```diff
@@ -0,0 +1,164 @@
...
+setup() {
+    # Create a temporary test directory and files
+    TEST_SYNC_DIR="$BATS_TEST_TMPDIR/sync-test"
+    mkdir -p "$TEST_SYNC_DIR"
+    touch "$TEST_SYNC_DIR/foo.txt"
+    touch "$TEST_SYNC_DIR/bar.txt"
+    
+    # Create a simple script that makes changes to these files
+    cat > "$TEST_SYNC_DIR/modify.sh" << 'EOF'
+#!/bin/sh
```
AkihiroSuda, line 46: Every shell script should have `set -eu`

cmd/limactl/shell.go
unsuman: Oh, I didn't know that `filepath.Abs()` translates `.` to cwd.

cmd/limactl/shell.go
```diff
@@ -193,7 +228,10 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		} else {
 			logrus.WithError(err).Warn("failed to get the home directory")
 		}
-	} else {
+	case syncHostWorkdir:
+		destRsyncDir = fmt.Sprintf("/home/%s.linux%s", *inst.Config.User.Name, hostCurrentDir)
```
jandubois: The home directory is configurable; don't recompute the default value:

```suggestion
		destRsyncDir = *inst.Config.User.Home + hostCurrentDir
```

There is another instance of this further down in the file.

cmd/limactl/shell.go
```diff
@@ -307,7 +373,188 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+		askUserForRsyncBack(ctx, cmd, inst, sshExecForRsync, hostCurrentDir, destRsyncDir, tty)
+	}
+	return nil
+}
+
+func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype.Instance, sshCmd *exec.Cmd, hostCurrentDir, destRsyncDir string, tty bool) {
+	remoteSource := fmt.Sprintf("%s:%s", *inst.Config.User.Name+"@"+inst.SSHAddress, destRsyncDir)
+	clean := filepath.Clean(hostCurrentDir)
+	parts := strings.Split(clean, string(filepath.Separator))
+	dirForCleanup := shellescape.Quote(fmt.Sprintf("/home/%s.linux/", *inst.Config.User.Name) + parts[1])
```
jandubois: Use `User.Home` instead. Also check that `strings.Split()` has returned at least 2 elements.

cmd/limactl/shell.go
```diff
@@ -307,7 +373,188 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+
+	rsyncBackAndCleanup := func() {
+		if err := rsyncDirectory(ctx, cmd, sshCmd, remoteSource, filepath.Dir(hostCurrentDir)); err != nil {
+			logrus.WithError(err).Warn("Failed to sync back the changes to host")
+			return
+		}
```
jandubois, line 406: Should this function return an error? Why are these failures just logging a warning, and then continue as if everything is ok?

cmd/limactl/shell.go
```diff
@@ -307,7 +373,188 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+			logrus.WithError(err).Warnf("Failed to clean up temporary directory %s", hostTmpDest)
+		}
+	}()
+	rsyncToTempDir := false
+
+	for {
+		ans, err := uiutil.Select(message, options)
+		if err != nil {
+			if errors.Is(err, uiutil.InterruptErr) {
+				logrus.Fatal("Interrupted by user")
```
jandubois: `Fatal` exits the process right away and should be reserved for unrecoverable errors. This feels more like a shortcut for regular control flow because we can't return an error.

cmd/limactl/shell.go
jandubois: Please create an issue for it if you don't include it in the PR.

jandubois (changes requested):
Mostly looking good, but the hard-coded home directory name should be fixed.

cmd/limactl/shell.go
unsuman: Should I raise this issue now? I have already raised a draft PR #4521 , I can make the PR description little more verbose?

AkihiroSuda:
@jandubois Can we merge this and release v2.1 alpha ?
FOSDEM is on this weekend

jandubois:
> Can we merge this and release v2.1 alpha ? FOSDEM is on this weekend

Sorry, I didn't realize this. I thought 2.1 was planned for KubeCon.

I can't review until tomorrow, but I'm fine with you merging it if you and @unsuman both think it is fine. It will just be an alpha, after all.

cmd/limactl/shell.go
```diff
@@ -307,7 +373,180 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+					return fmt.Errorf("failed to sync back the changes from guest instance to host temporary directory: %w", err)
+				}
+				rsyncToTempDir = true
+			}
+			diffCmd := exec.CommandContext(ctx, "diff", "-ru", "--color=always", hostCurrentDir, filepath.Join(hostTmpDest, filepath.Base(hostCurrentDir)))
+			pager := os.Getenv("PAGER")
+			if pager == "" {
+				pager = "less"
+			}
+			lessCmd := exec.CommandContext(ctx, pager)
```
AkihiroSuda: This doesn't work when `PAGER` contains whitespaces

AkihiroSuda:
> I thought 2.1 was planned for KubeCon.

Yes, the plan is unchanged.
For FOSDEM I just want an alpha.

AkihiroSuda:
CI failing

jandubois:
> CI failing

Because the PR needs to be rebased on latest `master`. CI wants to run `colima.bats` which doesn't exist in this PR.

cmd/limactl/shell.go
```diff
@@ -278,6 +316,34 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	}
 	sshArgs := append([]string{}, sshExe.Args...)
 	sshArgs = append(sshArgs, sshutil.SSHArgsFromOpts(sshOpts)...)
+
+	var sshExecForRsync *exec.Cmd
+	if syncHostWorkdir {
+		logrus.Infof("Syncing host current directory(%s) to guest instance...", hostCurrentDir)
+		sshExecForRsync = exec.CommandContext(ctx, sshExe.Exe, sshArgs...)
```
jandubois, line 324: I think this is missing the port and host arguments. It looks like it works if it connects via the ControlMaster socket, but if that is broken, it will not create its own connection:

```console
❯ rm ~/.lima/default/ssh.sock
❯ limactl shell --sync . --yes default pwd
INFO[0000] Syncing host current directory(/Users/jan/suse/lima/hack/bats) to guest instance...
FATA[0000] failed to create the synced workdir in guest instance: exit status 255
```

cmd/limactl/shell.go
```diff
@@ -307,7 +373,182 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+
+			if err := lessCmd.Wait(); err != nil {
+				return fmt.Errorf("failed to wait for less command: %w", err)
+			}
+		}
+	}
+}
+
+func executeSSHForRsync(ctx context.Context, sshCmd *exec.Cmd, command string) error {
+	sshRmCmd := exec.CommandContext(ctx, sshCmd.Path, append(sshCmd.Args, command)...)
```
jandubois:
```suggestion
	// Skip Args[0] (program name) to avoid duplication
	sshRmCmd := exec.CommandContext(ctx, sshCmd.Path, append(sshCmd.Args[1:], command)...)
```

cmd/limactl/shell.go
jandubois: I think the BATS test need to include variants that delete the `ssh.sock` to properly test the explicit SSH commands.

PR #4433, opened 2025-12-07
ci: fix CodeQL Go build step

### LINKED ISSUE #4399: CI: CodeQL setup seems broken
https://github.com/lima-vm/lima/actions/runs/19634391886/job/56221491392

> Go files were found but not processed (1 result)
>    * [Specify a custom build command](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages) that includes one or more `go build` commands to build the `.go` files to be analyzed.

Kim-Yukyung:
### What this PR changes

This PR switches CodeQL from autobuild to manual mode, adds a Go setup step using Go 1.24, and explicitly runs go build ./... so that CodeQL can properly analyze the Go codebase.

### Related Issue

Fixes #4399

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -35,10 +35,15 @@ jobs:
       matrix:
         include:
         - language: go
-          build-mode: autobuild
+          build-mode: manual
     steps:
     - name: Checkout repository
       uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8  # v6.0.1
+
+    - name: Set up Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: '1.24'
 
     - name: Initialize CodeQL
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yaml
```diff
@@ -35,26 +35,26 @@ jobs:
       matrix:
         include:
         - language: go
-          build-mode: autobuild
+          build-mode: manual
```
AkihiroSuda, line 38: Why can't we use the autobuild mode?

Kim-Yukyung: After taking a closer look at the logs, it seems that autobuild is not actually failing. I initially misunderstood the `Go files were found but not processed` message as indicating a required build step, but the analysis itself is completing successfully and the results are being uploaded correctly.

So I agree that switching to manual build is not strictly necessary at this point. Sorry for the confusion.

PR #4442, opened 2025-12-08
Avoid boot loop when there are dnf network issues

### LINKED ISSUE #4440: Fedora Lima image gets stuck in reboot loop if mirror is unavailable
### Description

@jandubois suggested I raise an issue here as well with regards to https://github.com/runfinch/finch/issues/1632

It appears that if the Fedora mirrors are unavailable or inaccessible, dnf needs-restarting will return a non-zero exit code which the cloudinit scripts determine to be the same as needing a reboot.

This results in no visible logging (that I can see) and the VM getting stuck in a reboot loop, chewing up significant host resources and being very difficult to debug.

From the original issue, this was my analysis:

---

When running using a Corporate Proxy that has its own SSL certificates, bringing up a Finch VM is problematic. Right now, upon first boot, we observe that `finch vm init` just hangs for about 15 minutes and then crashes.

Upon pulling all of this configuration and code to pieces, I found that the problem lies within the ISO that is downloaded. The script causing us problems is the following:

```
#!/bin/sh

# SPDX-FileCopyrightText: Copyright The Lima Authors
# SPDX-License-Identifier: Apache-2.0

set -eux

# Check if cloud-init forgot to reboot_if_required
# (only implemented for apt at the moment, not dnf)

if command -v dnf >/dev/null 2>&1; then
	# dnf-utils needs to be installed, for needs-restarting
	if dnf -h needs-restarting >/dev/null 2>&1; then
		# needs-restarting returns "false" if needed (!)
		if ! dnf needs-restarting -r >/dev/null 2>&1; then
			systemctl reboot
		fi
	fi
fi
```

Specifically, take not of the `if ! dnf needs-restarting -r >/dev/null 2>&1; then systemctl reboot`. Whilst it is true that `dnf needs-restarting` will return a non-zero exit code if we need to reboot, it **also** returns a non-zero exit code if it failed to complete.

It turns out that `dnf needs-restarting` dials out to the Fedora repository mirrors... under a corporate proxy that operates on L3/L4 (e.g. as part of a ZTNA), this won't work. You'll just get the following output (which is somewhat unhelpfully suppressed and sent to /dev/null here):

```log
└─[127] <> docker run --rm -it fedora                               
[root@97829c00283b /]# dnf needs-restarting
Updating and loading repositories:
 Fedora 42 - aarch64 - Updates                                                                                             ???% [  <=>             ] |   0.0   B/s |   0.0   B |  00m01s
>>> Curl error (60): SSL peer certificate or SSH remote key was not OK for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f42&arch=aarch64 [SSL certificate problem: u
>>> Curl error (60): SSL peer certificate or SSH remote key was not OK for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f42&arch=aarch64 [SSL certificate problem: u
>>> Curl error (60): SSL peer certificate or SSH remote key was not OK for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f42&arch=aarch64 [SSL certificate problem: u
>>> Curl error (60): SSL peer certificate or SSH remote key was not OK for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f42&arch=aarch64 [SSL certificate problem: u
>>> Curl error (60): SSL peer certificate or SSH remote key was not OK for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f42&arch=aarch64 [SSL certificate problem: u
...
```

This then exits with a non-zero exit code.

This means if you have no side-loaded CA certificates, `finch vm init` will get stuck in a loop of repeatedly restarting the VM every 5 seconds or so, while providing no output of what the issue is, since everything is sent to /dev/null.

afbjorklund:
Only try to reboot when upgradePackages is true, and check the network connectivity before checking if a restart is needed.

Also don't hide the stderr output, in case of any errors, and also show the list of packages that are requiring the reboot.

Closes #4440 

first boot:
```
+ command -v dnf
+ dnf -h needs-restarting
+ dnf check-update
Updating and loading repositories:
Repositories loaded.
+ '[' 0 '!=' 1 ']'
+ dnf needs-restarting -r
Updating and loading repositories:
 Fedora 43 openh264 (From Cisco) - x86_ 100% |   1.6 KiB/s |  10.1 KiB |  00m06s
 Fedora 43 - x86_64 - Updates           100% |   1.7 MiB/s |  21.9 MiB |  00m13s
 Fedora 43 - x86_64                     100% |   2.2 MiB/s |  57.8 MiB |  00m26s
Repositories loaded.
Core libraries or services have been updated since boot-up:
  * glib2
  * gnutls
  * kernel-core
  * kernel-modules-core
  * libxcrypt
  * openssl-libs
  * systemd

Reboot is required to fully utilize these updates.
More information: https://access.redhat.com/solutions/27943
+ systemctl reboot

Broadcast message from root@localhost (Mon 2025-12-08 12:46:02 CET):

The system will reboot now!
```

second boot:
```
+ command -v dnf
+ dnf -h needs-restarting
+ dnf check-update
Updating and loading repositories:
Repositories loaded.
+ '[' 0 '!=' 1 ']'
+ dnf needs-restarting -r
Updating and loading repositories:
Repositories loaded.
No core libraries or services have been updated since boot-up.
Reboot should not be necessary.
```

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # SPDX-FileCopyrightText: Copyright The Lima Authors
 # SPDX-License-Identifier: Apache-2.0
@@ -14,12 +14,21 @@ if command -v dnf >/dev/null 2>&1; then
 	# dnf-utils needs to be installed, for needs-restarting
 	if dnf -h needs-restarting >/dev/null 2>&1; then
 		# check-update returns "false" (100) if updates (!)
+		set +e
 		dnf check-update >/dev/null
 		if [ "$?" != "1" ]; then
+			# needs-restarting messages are translated _()
+			export LC_ALL=C.UTF-8
+			logfile=$(mktemp)
 			# needs-restarting returns "false" if needed (!)
-			if ! dnf needs-restarting -r; then
-				systemctl reboot
+			set -o pipefail
+			dnf needs-restarting -r | tee "$logfile"
+			if [ "$?" = "1" ]; then
+				if grep -q "Reboot is required" "$logfile"; then
+					systemctl reboot
+				fi
 			fi
+			rm "$logfile"
 		fi
 	fi
 fi


pkg/cidata/cidata.TEMPLATE.d/lima.env
```diff
@@ -32,6 +32,11 @@ LIMA_CIDATA_YQ_PROVISION_{{$yqProvision.FileName}}_PATH={{$yqProvision.Path}}
 LIMA_CIDATA_YQ_PROVISION_{{$yqProvision.FileName}}_PERMISSIONS={{$yqProvision.Permissions}}
 {{- end}}
 LIMA_CIDATA_GUEST_INSTALL_PREFIX={{ .GuestInstallPrefix }}
+{{- if .UpgradePackages}}
+LIMA_CIDATA_UPGRADE_PACKAGES=1
+{{- else}}
+LIMA_CIDATA_UPGRADE_PACKAGES=0
+{{- end}}
```
jandubois, line 39: It bothers me that we do these silly translations from booleans to environment variables that have to be set to `1` or the empty string. Or in this case `0`.

Why don't we assign the boolean value directly and deal with it that the variable will be set to `true` or `false`?


```suggestion
LIMA_CIDATA_UPGRADE_PACKAGES={{.UpgradePackages}}
```

pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
```diff
@@ -5,15 +5,20 @@
...
 # (only implemented for apt at the moment, not dnf)
 
 if command -v dnf >/dev/null 2>&1; then
 	# dnf-utils needs to be installed, for needs-restarting
 	if dnf -h needs-restarting >/dev/null 2>&1; then
-		# needs-restarting returns "false" if needed (!)
-		if ! dnf needs-restarting -r >/dev/null 2>&1; then
-			systemctl reboot
+		# check-update returns "false" (100) if updates (!)
+		dnf check-update >/dev/null; if [ "$?" != "1" ]; then
```
jandubois: Tucking an `if` as a second statement on the same line is too unconventional for me. 😄 

```suggestion
		dnf check-update >/dev/null
		if [ "$?" != "1" ]; then
```

jandubois (commented):
I don't understand which part of the PR is supposed to stop the reboot loop in case there is a network problem.

pkg/cidata/cidata.TEMPLATE.d/lima.env
afbjorklund: I was just following the rest of them, either way

pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
afbjorklund: It ended up that way because of the weird exit codes, but lint didn't like it either

pkg/cidata/cidata.TEMPLATE.d/lima.env
afbjorklund: But then it would be more on the lines of:

```bash
$LIMA_CIDATA_UPGRADE_PACKAGES || exit 0
```

afbjorklund: And using `0` is a bit unconventional, not "ifdef"

afbjorklund:
I have no idea why it would still download, if there is some kind of race condition or something?

```
+ dnf check-update
Updating and loading repositories:
Repositories loaded.
+ '[' 0 '!=' 1 ']'
+ dnf needs-restarting -r
Updating and loading repositories:
 Fedora 43 openh264 (From Cisco) - x86_ 100% |   1.6 KiB/s |  10.1 KiB |  00m06s
 Fedora 43 - x86_64 - Updates           100% |   1.7 MiB/s |  21.9 MiB |  00m13s
 Fedora 43 - x86_64                     100% |   2.2 MiB/s |  57.8 MiB |  00m26s
Repositories loaded.
```

Didn't we just download those files? Then again I had another fail mode where it pulled 5 GiB?!?

```console
[root@lima-fedora ~]# ls -1 /var/cache/libdnf5/
fedora-cisco-openh264-65b03c203277e378
fedora-fe4b228be7d3c150
updates-431245df5652aba0
```

----

EDIT: ~~Apparently it is edited all the time (the metadata), and the digests change every time~~:

```
diff -r /var/cache/libdnf5.orig/fedora-cisco-openh264-65b03c203277e378/metalink.xml /var/cache/libdnf5/fedora-cisco-openh264-65b03c203277e378/metalink.xml
2c2
< <metalink version="3.0" xmlns="http://www.metalinker.org/" type="dynamic" pubdate="Tue, 09 Dec 2025 19:04:45 GMT" generator="mirrormanager" xmlns:mm0="http://fedorahosted.
org/mirrormanager">
---
> <metalink version="3.0" xmlns="http://www.metalinker.org/" type="dynamic" pubdate="Tue, 09 Dec 2025 19:05:18 GMT" generator="mirrormanager" xmlns:mm0="http://fedorahosted.
org/mirrormanager">
```

But the second time it also had to pull "filelists", and not just the "primary" (the Requires)

```
Only in libdnf5/fedora-cisco-openh264-65b03c203277e378/repodata: fe7931412901203fb73717d3283209dcf05af265b14e78dbeaf82e20244abe69-filelists.xml.zst
Only in libdnf5/fedora-cisco-openh264-65b03c203277e378/solv: fedora-cisco-openh264-filelists.solvx
Files libdnf5.orig/fedora-fe4b228be7d3c150/metalink.xml and libdnf5/fedora-fe4b228be7d3c150/metalink.xml differ
Only in libdnf5/fedora-fe4b228be7d3c150: packages
Only in libdnf5/fedora-fe4b228be7d3c150/repodata: d98386eee3b666157656ac39450d1abad187bb47557d23cd6b199a0a4dbd9c5e-filelists.xml.zck
Only in libdnf5/fedora-fe4b228be7d3c150/solv: fedora-filelists.solvx
Files libdnf5.orig/updates-431245df5652aba0/metalink.xml and libdnf5/updates-431245df5652aba0/metalink.xml differ
Only in libdnf5/updates-431245df5652aba0/repodata: b2d5d62c3c27785998f95dff3bf68887fd131ddbaa5977f60d635626ef9189ed-filelists.xml.zck
Only in libdnf5/updates-431245df5652aba0/solv: updates-filelists.solvx
```

pkg/cidata/cidata.TEMPLATE.d/lima.env
afbjorklund: Apparently that was my bad, it was only `1` being used

pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
```diff
@@ -5,15 +5,21 @@
 
 set -eux
 
+[ "$LIMA_CIDATA_UPGRADE_PACKAGES" = "1" ] || exit 0
+
 # Check if cloud-init forgot to reboot_if_required
 # (only implemented for apt at the moment, not dnf)
```
AkihiroSuda, line 11: Not new in this PR, but apt and dnf seem swapped each other?

pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
```diff
@@ -5,15 +5,21 @@
...
 	# dnf-utils needs to be installed, for needs-restarting
 	if dnf -h needs-restarting >/dev/null 2>&1; then
-		# needs-restarting returns "false" if needed (!)
-		if ! dnf needs-restarting -r >/dev/null 2>&1; then
-			systemctl reboot
+		# check-update returns "false" (100) if updates (!)
+		dnf check-update >/dev/null
+		if [ "$?" != "1" ]; then
+			# needs-restarting returns "false" if needed (!)
+			if ! dnf needs-restarting -r; then
```
AkihiroSuda: If the command fails due to other reasons, this results in reboot loop too?
Should this match the stdout/stderr to avoid that?

pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
afbjorklund: It tries to say that cloud-init has implemented apt, but not dnf

This is why lima needs a workaround script for dnf, until it has?

pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
afbjorklund: The exit codes are hopelessly confused, scraping the plugin output _could_ make it worse

But I think it is possible, as long as you select the `C` locale (the messages are translated)

AkihiroSuda: When the output is unclear it should default to avoid the potential reboot loop

afbjorklund:
> I don't understand which part of the PR is supposed to stop the reboot loop in case there is a network problem.

The `check-update` will return an error in that case (just need to check that it is not `100`, which means "updates")

Also it now checks if the `upgradePackages` flag is used...

If the feature is not activated (the default), the script exits.

pkg/cidata/cidata.TEMPLATE.d/boot/00-reboot-if-required.sh
afbjorklund: Scraping the log for the "Reboot is required to fully utilize these updates." message.

afbjorklund:
First boot: (after a package update is made available)

```
LIMA 2025-12-19T18:30:19+01:00| Executing /mnt/lima-cidata/boot/00-reboot-if-required.sh
+ '[' 1 = 1 ']'
+ command -v dnf
+ dnf -h needs-restarting
+ set +e
+ dnf check-update
Updating and loading repositories:
Repositories loaded.
+ '[' 0 '!=' 1 ']'
+ export LC_ALL=C.UTF-8
+ LC_ALL=C.UTF-8
++ mktemp
+ logfile=/tmp/tmp.0HAd7emgm1
+ set -o pipefail
+ dnf needs-restarting -r
+ tee /tmp/tmp.0HAd7emgm1
Updating and loading repositories:
Repositories loaded.
Core libraries or services have been updated since boot-up:
  * glibc

Reboot is required to fully utilize these updates.
More information: https://access.redhat.com/solutions/27943
+ '[' 1 == 1 ']'
+ grep -q 'Reboot is required' /tmp/tmp.0HAd7emgm1
+ systemctl reboot
+ rm /tmp/tmp.0HAd7emgm1
```

Second boot:

```
LIMA 2025-12-19T18:30:36+01:00| Executing /mnt/lima-cidata/boot/00-reboot-if-required.sh
+ '[' 1 = 1 ']'
+ command -v dnf
+ dnf -h needs-restarting
+ set +e
+ dnf check-update
Updating and loading repositories:
Repositories loaded.
+ '[' 0 '!=' 1 ']'
+ export LC_ALL=C.UTF-8
+ LC_ALL=C.UTF-8
++ mktemp
+ logfile=/tmp/tmp.YK30LLIfng
+ set -o pipefail
+ dnf needs-restarting -r
+ tee /tmp/tmp.YK30LLIfng
Updating and loading repositories:
Repositories loaded.
No core libraries or services have been updated since boot-up.
Reboot should not be necessary.
+ '[' 0 == 1 ']'
+ rm /tmp/tmp.YK30LLIfng
```

Not sure what is the best way to tell cloud-init to stop executing? Seems wrong to fail the script.

```
      reboot
           Shut down and reboot the system.

           This command mostly equivalent to systemctl start reboot.target --job-mode=replace-irreversibly --no-block, but also prints a wall message to
           all users. This command is asynchronous; it will return after the reboot operation is enqueued, without waiting for it to complete.
```

Perhaps we should go back to `shutdown -r now`

Or `reboot`

Issue #3987, opened 2025-09-07
Number of CPUs is not validated


jandubois:
Setting to `0` is valid because it is the "null" value, meaning to use the default setting.

I wonder if the `--cpus` flag should not print a message in that case.

Setting to negative numbers is not caught by validation, but will cause a failure to start the VM. This must be caught; `limactl edit` should never leave `lima.yaml` in a non-startable state.

We do have validation to reject string values, even though allowing `max` or `host` would maybe make sense (`host` being the number of CPUs of the host, and `max` maybe 2 less?).

```console
❯ l edit -y foo --cpus 1
INFO[0000] Instance "foo" configuration edited

❯ l edit -y foo --cpus 0
INFO[0000] Instance "foo" configuration edited

❯ l edit -y foo --cpus -1
INFO[0000] Instance "foo" configuration edited

❯ l edit -y foo --cpus max
FATA[0000] invalid argument "max" for "--cpus" flag: strconv.ParseInt: parsing "max": invalid syntax
```

afbjorklund:
> Setting to 0 is valid because it is the "null" value, meaning to use the default setting.

Is this documented anywhere? I thought null (`~`) was the null value, in the lima.yaml.

----

I think it should be an integer, but we do allow string for `--memory` now - so maybe that is just backwards thinking...

Having to quote seems like a breaking change, so it would probably have to accept both `cpus: 4` and `cpus: host`

```json
                "cpus": {
                    "oneOf": [
                        {
                            "type": "integer"
                        },
                        {
                            "type": "string"
                        },
                        {
                            "type": "null"
                        }
                    ]
                },
```

EDIT: We don't actually have to change the JSON schema, only the limayaml type (from `int` to `string`)

However, it is not possible to validate the integer without invoking the driver - so not sure we should?

k-veda:
Hello, @sneugle and I are students at UT-Austin taking a course on virtualization. As part of it, we are tasked to work on an open source project. We previously spoke to a maintainer for this repository on the Slack, and were wondering if we could be assigned this issue?

afbjorklund:
> Hello, [@sneugle](https://github.com/sneugle) and I are students at UT-Austin taking a course on virtualization. As part of it, we are tasked to work on an open source project. We previously spoke to a maintainer for this repository on the Slack, and were wondering if we could be assigned this issue?

This issue is more about yaml engineering than it is about virtualization, but contributions are welcome

afbjorklund:
[...]

PR #4451, opened 2025-12-12
Validate number of CPUs

k-veda:
Fix for issue #3987. Added code to function `func YQExpressions()` to give an error if number of CPU's is negative in file `lima/cmd/limactl/editflags/editflags.go`.

### DIFF
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -191,7 +191,21 @@ func YQExpressions(flags *flag.FlagSet, newInstance bool) ([]string, error) {
 	}
 	d := defaultExprFunc
 	defs := []def{
-		{"cpus", d(".cpus = %s"), false, false},
+		{
+			"cpus",
+			func(_ *flag.Flag) ([]string, error) {
+				numCpus, err := flags.GetInt("cpus")
+				if err != nil {
+					return nil, err
+				}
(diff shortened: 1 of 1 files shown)


cmd/limactl/editflags/editflags.go
```diff
@@ -191,7 +191,21 @@ func YQExpressions(flags *flag.FlagSet, newInstance bool) ([]string, error) {
...
 	defs := []def{
-		{"cpus", d(".cpus = %s"), false, false},
+		{
+			"cpus",
+			func(_ *flag.Flag) ([]string, error) {
+				numCpus, err := flags.GetInt("cpus")
+				if err != nil {
+					return nil, err
+				}
+				if numCpus < 0 {
```
AkihiroSuda, line 201: > Setting to negative numbers is not caught by validation, but will cause a failure to start the VM. This must be caught; limactl edit should never leave lima.yaml in a non-startable state.
> 
> Can we just say `-1` means the same amount as the host?
> Other negative values such as `-2` should be invalid. 

 _Originally posted by @AkihiroSuda in [#3987](https://github.com/lima-vm/lima/issues/3987#issuecomment-3652680949)_

AkihiroSuda: (Can be revisited in another PR)

PR #4453, opened 2025-12-13
fix: nil pointer dereference panic in shellAction

paralin:
The shellAction function panics with a nil pointer dereference when attempting to shell into an instance that has configuration validation errors:

    [signal SIGSEGV: segmentation violation code=0x2 addr=0x98 pc=0x...]

This occurs because store.Inspect() returns an instance with inst.Config set to nil when validation fails, but the code at lines 171, 252-256 accesses inst.Config fields without checking for nil.

Example scenario: An instance created with Lima 1.2.1 using QEMU has mountType: 9p in its config. On newer Lima versions, if vmType defaults to vz (or is set explicitly), the config fails validation since VZ only supports reverse-sshfs or
virtiofs mount types. The validation error leaves inst.Config as nil, causing the panic.

This fix adds an early nil check for inst.Config (after store.Inspect at line 90) to return a clear error message instead of panicking:

    instance "docker" has configuration errors: field 'mountType' must be
    "reverse-sshfs" or "virtiofs" for VZ driver, got "9p"

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -94,6 +94,12 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		}
 		return err
 	}
+	if inst.Config == nil {
+		if len(inst.Errors) > 0 {
+			return fmt.Errorf("instance %q has configuration errors: %w", instName, errors.Join(inst.Errors...))
+		}
+		return fmt.Errorf("instance %q has no configuration", instName)
+	}
 	if inst.Status == limatype.StatusStopped {
 		startNow, err := flags.GetBool("start")
 		if err != nil {


paralin:
This was to fix the following error I get:

```
cjs@flame ~ % lima -v
limactl version 2.0.2

cjs@flame ~ % lima
[signal SIGSEGV: segmentation violation code=0x2 addr=0x98 pc=0x104f42f28]

goroutine 1 [running]:
main.shellAction(0x14000532f08, {0x140005805c0, 0x1, 0x1})
        /private/tmp/lima-20251124-4851-jwxwv0/lima-2.0.2/cmd/limactl/shell.go:165 +0x6c8
github.com/spf13/cobra.(*Command).execute(0x14000532f08, {0x14000580590, 0x1, 0x1})
        /Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1015 +0x7d4
github.com/spf13/cobra.(*Command).ExecuteC(0x14000532308)
        /Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1148 +0x350
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/brew/Library/Caches/Homebrew/go_mod_cache/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1071
main.main()
        /private/tmp/lima-20251124-4851-jwxwv0/lima-2.0.2/cmd/limactl/main.go:50 +0x7c
```

paralin:
I have tested and confirmed this works now.

paralin:
Ci fail seems like a flake:

```
   TEST| [INFO] Uninstalling lima
  TEST| [INFO] Installing the old Lima v0.15.1
  curl: (22) The requested URL returned error: 504
```

AkihiroSuda:
> In certain execution paths, cmd.Flags() could return nil, causing a panic:

What paths?

cmd/limactl/shell.go
```diff
@@ -155,14 +161,14 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
-	workDir, err := cmd.Flags().GetString("workdir")
+	workDir, err := flags.GetString("workdir")
 	if err != nil {
 		return err
 	}
 	if workDir != "" {
 		changeDirCmd = fmt.Sprintf("cd %s || exit 1", shellescape.Quote(workDir))
 		// FIXME: check whether y.Mounts contains the home, not just len > 0
-	} else if len(inst.Config.Mounts) > 0 || inst.VMType == limatype.WSL2 {
+	} else if inst.Config != nil && (len(inst.Config.Mounts) > 0 || inst.VMType == limatype.WSL2) {
```
AkihiroSuda: How can this be nil ?

paralin:
@AkihiroSuda this is to fix an actual nil reference crash I can reproduce reliably on the current master branch without this PR applied. The issue occurs when you have an invalid config. The config validation fails and config is nil. The err is not checked and the program therefore hits a nil dereference exception when trying to start the vm (with lima or limactl shell).

paralin:
> > In certain execution paths, cmd.Flags() could return nil, causing a panic:
> 
> What paths?

1. `store.Inspect()` is called (line 90) and returns an `inst` with `inst.Config == nil` when there are validation errors
2. The early check at lines 97-102 handles the case when status is checked before starting
3. However, after `instance.Start()` succeeds (line 130), the code re-inspects the instance (line 135-138) without re-checking if `inst.Config` is nil
4. Later access at line 171 (`inst.Config.Mounts`) and lines 252-256 (all the `inst.Config.*` field accesses) then cause nil pointer dereferences

cmd/limactl/shell.go
paralin: `inst.Config` can be nil when the instance has configuration validation errors.

paralin:
@AkihiroSuda Here is the broken config:

```yaml
images:
  - location: "https://cloud.debian.org/images/cloud/sid/daily/latest/debian-sid-genericcloud-arm64-daily.qcow2"
    arch: "aarch64"

mounts: []

containerd:
  system: false
  user: false

# The problem: mountType 9p is incompatible with VZ driver
# On Apple Silicon, vmType defaults to "vz" but 9p only works with QEMU
mountType: 9p
cpus: 11
memory: 10GiB

provision:
  - mode: system
    script: |
      #!/bin/bash
      set -eux -o pipefail
      command -v docker >/dev/null 2>&1 && exit 0
      apt update
      apt-get install -y docker.io

portForwards:
  - guestSocket: "/var/run/docker.sock"
    hostSocket: "{{.Dir}}/sock/docker.sock"
```

This config was originally created using QEMU (which supports 9p mounts), but changed `vmType` to `vz`  later. The validation fails with: `field 'mountType' must be "reverse-sshfs" or "virtiofs" for VZ driver, got "9p"`.

After the PR is applied, `lima` or `limactl shell` now just prints the correct error (field mountType ...) instead of panicing.

cmd/limactl/shell.go
paralin: You're saying the change on this line is not needed since we have a early-return for inst.Config == nil earlier?

paralin: I see. FIxed the commit message, PR description, etc! Good catch!

paralin:
@AkihiroSuda Thank you for your review. According to your comments the PR description + commit message are fixed and the PR is now only 6 lines long checking the error.

AkihiroSuda (approved):
Thanks

paralin:
@AkihiroSuda Looks good thanks!

PR #2537, opened 2024-08-07
Create amazonlinux2023 template

siennathesane:
Adds a default template reference for Amazon Linux 2023.

PR #4454, opened 2025-12-14
feat(templates): add amazonlinux-2023 distro

### LINKED ISSUE #4047: Add amazon linux template
### Description

sometimes I need to check if something works on amazon linux, but I don't want to use an "EC2" instance for that, I think it would be cool to have this template to run a amazon linux locally with lima.

Lima already have a "fedora" template, for some cases it is enough, but thi isn't the real OS running in EC2 instances

electricalgorithm:
This commit adds a simple template for the Amazon Linux 2023 operating system's latest version. The OS is scheduled to release every two weeks, therefore, we need to find a way to automatically update it from the following AWS website:

- https://cdn.amazonlinux.com/al2023/os-images/latest/

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -eux -o pipefail
+
+# Workaround for Amazon Linux 2023
+if [ -f /etc/os-release ] && grep -q "Amazon Linux" /etc/os-release; then
+	# 1. Create missing mount.virtiofs helper
+	if [ ! -e /sbin/mount.virtiofs ]; then
+		cat >/sbin/mount.virtiofs <<'EOF'
+#!/bin/sh
+exec mount -i -t virtiofs "$@"
+EOF
+		chmod +x /sbin/mount.virtiofs
+	fi
+
+	# Amazon Linux 2023 requires a workaround for virtiofs mounts.
+	# The mount.virtiofs helper script is missing, and cloud-init
+	# may fail to mount the filesystems. This creates the helper
+	# and manually mounts them.
+	USER_DATA="/var/lib/cloud/instance/user-data.txt"
+	if [ -f "${USER_DATA}" ]; then
+		# Parse all mount entries from user-data (mount0, mount1, ...)
+		MOUNT_ENTRIES=$(grep -E '^\s*-\s+\[mount[0-9]+,' "${USER_DATA}" || true)
+		if [ -n "${MOUNT_ENTRIES}" ]; then
+			echo "${MOUNT_ENTRIES}" | while IFS= read -r entry; do
+				MOUNT_TAG=$(echo "${entry}" | grep -oP 'mount\K[0-9]+')
+				MOUNT_POINT=$(echo "${entry}" | awk -F, '{print $2}' | tr -d '[:space:]')
+				if [ -n "${MOUNT_TAG}" ] && [ -n "${MOUNT_POINT}" ] && ! mountpoint -q "${MOUNT_POINT}"; then
+					mkdir -p "${MOUNT_POINT}"
+					mount -t virtiofs "mount${MOUNT_TAG}" "${MOUNT_POINT}" || true
+				fi
+			done
+		fi
+	fi
+fi
--- a/templates/_images/amazonlinux-2023.yaml
+++ b/templates/_images/amazonlinux-2023.yaml
@@ -1,0 +1,9 @@
+# More information: https://docs.aws.amazon.com/linux/al2023/ug/outside-ec2-download.html
+
+images:
+- location: "https://cdn.amazonlinux.com/al2023/os-images/2023.11.20260406.2/kvm/al2023-kvm-2023.11.20260406.2-kernel-6.1-x86_64.xfs.gpt.qcow2"
+  arch: "x86_64"
+  digest: "sha256:72e10d974e832d9b56b84e71fe6f2e22589c9c539cc38e6d00cd84177839e2e8"
+- location: "https://cdn.amazonlinux.com/al2023/os-images/2023.11.20260406.2/kvm-arm64/al2023-kvm-2023.11.20260406.2-kernel-6.1-arm64.xfs.gpt.qcow2"
+  arch: "aarch64"
+  digest: "sha256:c012f9ecb88f377c49e4338c2c6f5702ca885a37e96118119ce67266d0a7524d"
--- a/templates/amazonlinux-2023.yaml
+++ b/templates/amazonlinux-2023.yaml
@@ -1,0 +1,5 @@
+minimumLimaVersion: 2.1.2
+
+base:
+- template:_images/amazonlinux-2023
+- template:_default/mounts
--- a/templates/amazonlinux.yaml
+++ b/templates/amazonlinux.yaml
@@ -1,0 +1,1 @@
+amazonlinux-2023.yaml
--- a/hack/update-template-amazonlinux.sh
+++ b/hack/update-template-amazonlinux.sh
@@ -1,0 +1,190 @@
+#!/usr/bin/env bash
+
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -eu -o pipefail
+
+# Functions in this script assume error handling with 'set -e'.
+# To ensure 'set -e' works correctly:
+# - Use 'set +e' before assignments and '$(set -e; <function>)' to capture output without exiting on errors.
+# - Avoid calling functions directly in conditions to prevent disabling 'set -e'.
+# - Use 'shopt -s inherit_errexit' (Bash 4.4+) to avoid repeated 'set -e' in all '$(...)'.
+shopt -s inherit_errexit || error_exit "inherit_errexit not supported. Please use bash 4.4 or later."
+
+function amazonlinux_print_help() {
+	cat <<HELP
+$(basename "${BASH_SOURCE[0]}"): Update the Amazon Linux 2023 image location in the specified templates
+
+Usage:
+  $(basename "${BASH_SOURCE[0]}") <template.yaml>...
+
+Description:
+  This script updates the Amazon Linux 2023 image location in the specified templates.
+  Image location format:
+    https://cdn.amazonlinux.com/al2023/os-images/<version>/<folder>/<filename>
+
+  Latest version information is fetched from:
+    https://cdn.amazonlinux.com/al2023/os-images/latest/
+
+Examples:
+  Update the Amazon Linux 2023 image location in templates/**.yaml:
+  $ $(basename "${BASH_SOURCE[0]}") templates/**.yaml
+
+Flags:
+  -h, --help          Print this help message
+HELP
+}
+
+function amazonlinux_cache_key_for_image_kernel() {
+	local location=$1
+	case "${location}" in
+	https://cdn.amazonlinux.com/al2023/os-images/*) ;;
+	*) return 1 ;;
+	esac
+	# Cache key based on architecture and variant derived from location
+	# We use a simple heuristic: file path components.
+	# The URL structure is .../os-images/<version>/<folder>/...
+	# We want a key that represents "amazonlinux:al2023:<folder>"
+	local folder
+	if [[ ${location} == *"/kvm-arm64/"* ]]; then
+		folder="kvm-arm64"
+	elif [[ ${location} == *"/kvm/"* ]]; then
+		folder="kvm"
+	else
+		return 1
+	fi
+	echo "amazonlinux:al2023:${folder}"
+}
+
+function amazonlinux_image_entry_for_image_kernel() {
+	local location=$1
+	case "${location}" in
+	https://cdn.amazonlinux.com/al2023/os-images/*) ;;
+	*) return 1 ;;
+	esac
+
+	local latest_url
+	# Get the redirect URL to find the latest version
+	latest_url=$(curl -Ls -o /dev/null -w '%{url_effective}' https://cdn.amazonlinux.com/al2023/os-images/latest/)
+	local version
+	version=$(basename "${latest_url}")
+
+	local arch folder
+	if [[ ${location} == *"x86_64"* ]]; then
+		# shellcheck disable=SC2034
+		arch="x86_64"
+		folder="kvm"
+	elif [[ ${location} == *"arm64"* ]] || [[ ${location} == *"aarch64"* ]]; then
+		# shellcheck disable=SC2034
+		arch="aarch64"
+		folder="kvm-arm64"
+	else
+		error_exit "Unknown arch for amazonlinux location: ${location}"
+	fi
+
+	local base_url="https://cdn.amazonlinux.com/al2023/os-images/${version}/${folder}"
+	local checksums
+	checksums=$(download_to_cache "${base_url}/SHA256SUMS")
+
+	local filename digest line
+	# Find the qcow2 file in SHA256SUMS
+	line=$(grep "\.qcow2$" "${checksums}" | head -n 1)
+	[[ -n ${line} ]] || error_exit "No qcow2 image found in ${base_url}/SHA256SUMS"
+
+	# shellcheck disable=SC2034
+	digest=$(echo "${line}" | awk '{print "sha256:"$1}')
+	filename=$(echo "${line}" | awk '{print $2}')
+
+	# shellcheck disable=SC2034
+	local location="${base_url}/${filename}"
+
+	json_vars location arch digest
+}
+
+# check if the script is executed or sourced
+# shellcheck disable=SC1091
+if [[ ${BASH_SOURCE[0]} == "${0}" ]]; then
+	scriptdir=$(dirname "${BASH_SOURCE[0]}")
+	# shellcheck source=./cache-common-inc.sh
+	. "${scriptdir}/cache-common-inc.sh"
+
+	# shellcheck source=/dev/null # avoid shellcheck hangs on source looping
+	. "${scriptdir}/update-template.sh"
+else
+	# this script is sourced
+	if [[ -v SUPPORTED_DISTRIBUTIONS ]]; then
+		SUPPORTED_DISTRIBUTIONS+=("amazonlinux")
+	else
+		declare -a SUPPORTED_DISTRIBUTIONS=("amazonlinux")
+	fi
+	return 0
+fi
+
+declare -a templates=()
+while [[ $# -gt 0 ]]; do
+	case "$1" in
+	-h | --help)
+		amazonlinux_print_help
+		exit 0
+		;;
+	-d | --debug) set -x ;;
+	*.yaml) templates+=("$1") ;;
+	*)
+		error_exit "Unknown argument: $1"
+		;;
+	esac
+	shift
+done
+
+if [[ ${#templates[@]} -eq 0 ]]; then
+	amazonlinux_print_help
+	exit 0
+fi
+
+declare -A image_entry_cache=()
+
+for template in "${templates[@]}"; do
+	echo "Processing ${template}"
+	# 1. extract location by parsing template
+	yq_filter="
+		.images[] | [.location, .kernel.location, .kernel.cmdline] | @tsv
+	"
+	parsed=$(yq eval "${yq_filter}" "${template}")
+
+	# 2. get the image location
+	arr=()
+	while IFS= read -r line; do arr+=("${line}"); done <<<"${parsed}"
+	locations=("${arr[@]}")
+	for ((index = 0; index < ${#locations[@]}; index++)); do
+		[[ ${locations[index]} != "null" ]] || continue
+		set -e
+		IFS=$'\t' read -r location _kernel_location _kernel_cmdline <<<"${locations[index]}"
+		set +e # Disable 'set -e' to avoid exiting on error for the next assignment.
+		cache_key=$(
+			set -e # Enable 'set -e' for the next command.
+			amazonlinux_cache_key_for_image_kernel "${location}"
+		) # Check exit status separately to prevent disabling 'set -e' by using the function call in the condition.
+		# shellcheck disable=2181
+		[[ $? -eq 0 ]] || continue
+		image_entry=$(
+			set -e # Enable 'set -e' for the next command.
+			if [[ -v image_entry_cache[${cache_key}] ]]; then
+				echo "${image_entry_cache[${cache_key}]}"
+			else
+				amazonlinux_image_entry_for_image_kernel "${location}"
+			fi
+		) # Check exit status separately to prevent disabling 'set -e' by using the function call in the condition.
+		# shellcheck disable=2181
+		[[ $? -eq 0 ]] || continue
+		set -e
+		image_entry_cache[${cache_key}]="${image_entry}"
+		if [[ -n ${image_entry} ]]; then
+			echo "${image_entry}" | jq
+			limactl edit --log-level error --set "
+				.images[${index}] = ${image_entry}|
+				(.images[${index}] | ..) style = \"double\"
+			" "${template}"
+		fi
+	done
+done
--- a/hack/update-template.sh
+++ b/hack/update-template.sh
@@ -174,6 +174,8 @@ if [[ ${BASH_SOURCE[0]} == "${0}" ]]; then
 	. "${scriptdir}/update-template-freebsd.sh"
 	# shellcheck source=./update-template-macos.sh
 	. "${scriptdir}/update-template-macos.sh"
+	# shellcheck source=./update-template-amazonlinux.sh
+	. "${scriptdir}/update-template-amazonlinux.sh"
 else
 	# this script is sourced
 	return 0
--- a/templates/README.md
+++ b/templates/README.md
@@ -40,6 +40,7 @@ lima
 - [`almalinux-kitten-10`](./almalinux-kitten-10.yaml), `almalinux-kitten`: AlmaLinux Kitten 10
 - [`alpine`](./alpine.yaml): ☆Alpine Linux
 - [`alpine-iso`](./alpine-iso.yaml): ☆Alpine Linux (ISO9660 image). Compatible with the `alpine` template used in Lima prior to v1.0.
+- [`amazonlinux-2023`](./amazonlinux-2023.yaml): Amazon Linux 2023
 - [`archlinux`](./archlinux.yaml): ☆Arch Linux
 - [`centos-stream-9`](./centos-stream-9.yaml), `centos-stream`: CentOS Stream 9
 - [`centos-stream-10`](./centos-stream-10.yaml): CentOS Stream 10


electricalgorithm:
Testing it with:
```
$ LIMA_TEMPLATES_PATH=lima/templates limactl start --name al2023-test --arch aarch64 --cpus 4 template:amazonlinux-2023 --debug
```

Correct os-release Info:
```
[gokhankocmarli@lima-al2023-test ~]$ cat /etc/os-release
NAME="Amazon Linux"
VERSION="2023"
ID="amzn"
ID_LIKE="fedora"
VERSION_ID="2023"
PLATFORM_ID="platform:al2023"
PRETTY_NAME="Amazon Linux 2023.9.20251208"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2023"
HOME_URL="https://aws.amazon.com/linux/amazon-linux-2023/"
DOCUMENTATION_URL="https://docs.aws.amazon.com/linux/"
SUPPORT_URL="https://aws.amazon.com/premiumsupport/"
BUG_REPORT_URL="https://github.com/amazonlinux/amazon-linux-2023"
VENDOR_NAME="AWS"
VENDOR_URL="https://aws.amazon.com/"
SUPPORT_END="2029-06-30"
```

Correct uname:
```
[gokhankocmarli@lima-al2023-test ~]$ uname -a
Linux lima-al2023-test 6.1.158-180.294.amzn2023.aarch64 #1 SMP Mon Dec  1 05:36:18 UTC 2025 aarch64 aarch64 aarch64 GNU/Linux
```

Correct repos:
```
[gokhankocmarli@lima-al2023-test ~]$ yum repolist
repo id                                                                         repo name
amazonlinux                                                                     Amazon Linux 2023 repository
kernel-livepatch                                                                Amazon Linux 2023 Kernel Livepatch repository
```

electricalgorithm:
It's weird, but the only problem is that the backspace does not work. I don't know if Lima uses SSH underneath, but I'll investigate the debug logs by re-shelling to the VM.

[Update] Yep, it looks like it uses SSH.
```
DEBU[0000] Failed to detect CPU features. Assuming that AES acceleration is available on this Apple silicon.
DEBU[0000] OpenSSH version 10.0.2 detected, is GSSAPI supported: true
DEBU[0000] executing ssh (may take a long)): [/usr/bin/ssh -F /dev/null -o IdentityFile="/Users/gokhankocmarli/.lima/_config/user" -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o NoHostAuthenticationForLocalhost=yes -o PreferredAuthentications=publickey -o Compression=no -o BatchMode=yes -o IdentitiesOnly=yes -o GSSAPIAuthentication=no -o Ciphers="^aes128-gcm@openssh.com,aes256-gcm@openssh.com" -o User=gokhankocmarli -o ControlMaster=auto -o ControlPath="/Users/gokhankocmarli/.lima/al2023-test/ssh.sock" -o ControlPersist=yes -t -o SendEnv=COLORTERM -o LogLevel=ERROR -p 54658 127.0.0.1 -- cd /Users/gokhankocmarli/ || cd /Users/gokhankocmarli ; exec "$SHELL" --login]
```

Then I believe it could be an issue with SendEnv.

[Update 2] The --preserve-env did not work either. Plus, I realised that the mountpoint did not work. I believe there has to be a directory called /Users/gokhankocmarli, which is my $HOME on the host machine.

afbjorklund:
* https://github.com/lima-vm/lima/issues/4047

Previous attempt with this (Amazon Linux) was:

* https://github.com/lima-vm/lima/pull/2537

I think it also failed to automate the updating?

electricalgorithm:
> * [Add amazon linux template #4047](https://github.com/lima-vm/lima/issues/4047)
> 
> 
> Previous attempt with this (Amazon Linux) was:
> 
>     * [Create amazonlinux2023 template #2537](https://github.com/lima-vm/lima/pull/2537)
> 
> 
> I think it also failed to automate the updating?

Thank you for the info! I have reviewed the comments there and see that two concerns are in common:

- No dynamic approach to retrieve the latest image from CDN. (Should we even want this?)
- Mount problems with virtiofs.

I'll spend some time fixing those. However, I believe that we can move this template to experimental ones and start iteratively on fixing stuff. Some folks may not need mounts or be happy with just having AL2023 with Lima without the latest image since it's possible to easily update the system using `dnf upgrade --releasever=<new-release-date>`. Or even better, use `latest` for releasever.

We could easily patch it by adding a run-script when the image is initialised and upgrading the system to the latest. Wdyt?

afbjorklund:
You can "fix" the mount problems by excluding the methods that don't work, like 9p (check other templates for examples)

But it's a problem if there is no download URL. Maybe something that upstream can provide, if you ask them?

electricalgorithm:
> You can "fix" the mount problems by excluding the methods that don't work, like 9p (check other templates for examples)

I used the same mount point restrictions as the previous PR, the issue still exists but I'm investigating it.

> But it's a problem if there is no download URL. Maybe something that upstream can provide, if you ask them?

I added a new update-template script under hack/ directory. This script automatically fetches the latest version and updates the template. I used the trick of following the URL endpoint /latest and parsing its 302 response to resolve the latest version number (including .1/.0 at the end). I don't know who calls them in what frequency, but I see that it's the common and agreed approach.

electricalgorithm:
**Update**: The mount problem is gone, and it's possible to update the image to the latest one within a new script under `hack/`. Additionally, the system tries to update the latest release when the machine is initialised automatically.

I'll check the unit test and the lining stuff later. Does it look OK to you?

templates/experimental/amazonlinux-2023.yaml
```diff
@@ -0,0 +1,58 @@
...
+- location: "~"
+  mountPoint: "/mnt/host"
+
+message: |
+  ####################################################
+  Amazon Linux 2023 (2023.9.20251208.0)
+
+  To update the latest Amazon Linux 2023 release, run:
+  $ sudo dnf upgrade --releasever=latest
+  ####################################################
```
AkihiroSuda: We do not have this kind of message in other templates

templates/amazonlinux-2023.yaml
```diff
@@ -0,0 +1,58 @@
...
+    #!/bin/bash
+    set -eux -o pipefail
+    # Update to the latest release on the first boot
+    if [ ! -f /etc/lima-al2023-updated ]; then
+      dnf upgrade -y --releasever=latest
+    fi
+
+mounts:
+- location: "~"
+  mountPoint: "/mnt/host"
```
AkihiroSuda: The path is inconsistent with other template

templates/amazonlinux-2023.yaml
```diff
@@ -0,0 +1,58 @@
...
+  arch: "aarch64"
+  digest: "sha256:7aef1f189076a0416cd16062bbf7e0928d81061e68411cab8904447e1bd5eafd"
+
+
+# Do not install containerd by default.
+containerd:
+  system: false
+  user: false
+
+provision:
```
AkihiroSuda: Should be in https://github.com/lima-vm/lima/tree/39b8a4bd6e0761f09efb9235c63a76a246bc000b/pkg/cidata/cidata.TEMPLATE.d/boot

templates/experimental/amazonlinux-2023.yaml
```diff
@@ -0,0 +1,58 @@
...
+images:
+- location: "https://cdn.amazonlinux.com/al2023/os-images/2023.9.20251208.0/kvm/al2023-kvm-2023.9.20251208.0-kernel-6.1-x86_64.xfs.gpt.qcow2"
+  arch: "x86_64"
+  digest: "sha256:c91f244483d9a460869738f7fcfb5e37eb402aec8582d10456cf0c108e1a4d4c"
+- location: "https://cdn.amazonlinux.com/al2023/os-images/2023.9.20251208.0/kvm-arm64/al2023-kvm-2023.9.20251208.0-kernel-6.1-arm64.xfs.gpt.qcow2"
+  arch: "aarch64"
+  digest: "sha256:7aef1f189076a0416cd16062bbf7e0928d81061e68411cab8904447e1bd5eafd"
+
+
+# Do not install containerd by default.
```
AkihiroSuda: Reason?

hack/update-template-amazonlinux-2023.sh
AkihiroSuda, line 1: Other `update-template-*.sh` does not have the version in the file name

templates/amazonlinux-2023.yaml
AkihiroSuda, line 1: Should be symlinked as `amazonlinux.yaml`

hack/update-template-amazonlinux-2023.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+SETUP_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+TEMPLATE_FILE="${SETUP_DIR}/../templates/experimental/amazonlinux-2023.yaml"
```
AkihiroSuda: The template file must not be hard-coded.
See other updaters.

hack/update-template-amazonlinux-2023.sh
AkihiroSuda, line 1: Should be called from https://github.com/lima-vm/lima/blob/master/hack/update-template.sh

templates/amazonlinux-2023.yaml
```diff
@@ -0,0 +1,58 @@
+minimumLimaVersion: 2.0.0
```
AkihiroSuda: Probably this is going to be:

```suggestion
minimumLimaVersion: 2.1.0
```

templates/experimental/amazonlinux-2023.yaml
```diff
@@ -0,0 +1,58 @@
+minimumLimaVersion: 2.0.0
+
+
+
```
AkihiroSuda: Extra empty lines

electricalgorithm: Solved, thanks :)

templates/amazonlinux-2023.yaml
electricalgorithm: Solved, thanks :)

hack/update-template-amazonlinux-2023.sh
electricalgorithm: Solved, thanks :)

hack/update-template-amazonlinux-2023.sh
electricalgorithm: Solved, thanks :)

templates/amazonlinux-2023.yaml
electricalgorithm: Done, thanks

hack/update-template-amazonlinux-2023.sh
electricalgorithm: Done, thanks!

templates/experimental/amazonlinux-2023.yaml
electricalgorithm: I didn't want to install it by default on my trials. Removed it now.

templates/amazonlinux-2023.yaml
electricalgorithm: I don't know what the contextual difference is between provision on the template level and common storage, but I will move the AL-specific connection patches there directly. The one with the mount fix should be in the template.

templates/amazonlinux-2023.yaml
electricalgorithm: What other template? It's the first Amazon Linux template. Do you mean mounting it to a specific directory?

templates/experimental/amazonlinux-2023.yaml
electricalgorithm: Removed it, thanks

templates/amazonlinux-2023.yaml
electricalgorithm: I guess you meant by mounting directly into users' home to /Users/$USER. I will try to do it.

electricalgorithm: It's done. Automatically connects to what mount0 shows. Thanks!

templates/amazonlinux-2023.yaml
electricalgorithm: Move all the weird patches to boot/. Thanks

templates/amazonlinux-2023.yaml
```diff
@@ -0,0 +1,15 @@
...
+base:
+- template:_images/amazonlinux-2023
+- template:_default/mounts
+
+provision:
+- mode: system
+  script: |
+    #!/bin/bash
+    set -eux -o pipefail
+    # Update to the latest release on the first boot
```
afbjorklund: This should use `upgradePackages` instead, if desired

Then cloud-init will call dnf, and reboot if required

electricalgorithm: I haven't found an example of the usage or found the implementation on the Go modules. However, AL2023 does use a different style of releases. Running a simple "yum/dnf upgrade" would not do anything if "releasever" is not defined. I guess this version of upgrade is better. Wdyt?

electricalgorithm:
Could someone possibly help me reproduce these CI checks? I do run shellcheck, shfmt, etc. but I guess there's some configurations different since all seems good on my local. There has to be a guide on running/replicating GitHub actions in this doc: https://lima-vm.io/docs/dev/testing/

templates/amazonlinux-2023.yaml
afbjorklund: You lost me there. The upgradePackages is a template feature:

```yaml
# Upgrade the instance on boot
# Reboot after upgrade if required
# 🟢 Builtin default: false
upgradePackages: null
```

afbjorklund: Or you mean that "dnf upgrade -y" does not work in this distribution?

afbjorklund:
There is `make lint`, but keeping it in sync with `.github` is a constant battle... Maybe https://github.com/nektos/act

templates/amazonlinux-2023.yaml
electricalgorithm: It have a different effect then giving releasever. Amazon Linux uses a "fixed" repository method where every two week, they release another repo. This helps customers to make sure that whenever they use the AL2023 with releasever=20251208, all packages would be the same. So, if one wants to update their packages, they have to upgrade the release version of the repository. 

Standard yum/dnf upgrade or update is not going to work since it does not define which version to upgrade the repo.

afbjorklund: Then it seems to me that the template should not be updating anything at all. If the user writes `20251208.0` they want just that, similar to when you run a docker container. If you want a newer version, you get a newer image? Instead of trying to run the package manager inside a container.  Or that it shouldn't be automatic...

For CentOS it was the opposite, they didn't support snapshots (not even monthly, let alone weekly) and when you upgraded you always got the latest. Not the "x.y.z" thing that RHEL support bought you, but only the "x". So I would just skip the whole section, but otherwise it should create the /etc file after upgrading I think?

afbjorklund:
I still don't want new commits to Lima every two  (just to bump a snapshot version) every two weeks

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -eux -o pipefail
+
+# Workaround for Amazon Linux 2023
+if [ -f /etc/os-release ] && grep -q "Amazon Linux" /etc/os-release; then
+	# 1. Create missing mount.virtiofs helper
```
AkihiroSuda, line 10: Would it be possible to just fix this issue in the Amazon Linux upstream?

templates/_images/amazonlinux-2023.yaml
```diff
@@ -0,0 +1,7 @@
+images:
+- location: "https://cdn.amazonlinux.com/al2023/os-images/2023.9.20251208.0/kvm/al2023-kvm-2023.9.20251208.0-kernel-6.1-x86_64.xfs.gpt.qcow2"
```
AkihiroSuda: Is this URL permanent, or old images are removed periodically?

AkihiroSuda:
ping @electricalgorithm

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
AkihiroSuda: Curious why this issue happens specifically with Amazon Linux. I thought it was very similar to Fedora?

electricalgorithm:
I'll update the PR as soon as possible.

templates/_images/amazonlinux-2023.yaml
electricalgorithm: The URL is permenant. They will be all there. For any case, I added the hack script.

templates/amazonlinux-2023.yaml
electricalgorithm: I understand your point. However, IMO, the updates can be automatically applied since:

- AL2023 have an compatibility promise -- where there cannot be any update that breaks the system packages' compatibility. Therefore, there should not be any concern from any user to use the latest.
- AL2023's package updates are mainly regarding to CVE fixes. As far as I know mainstream users, they should be able to always get the latest security fixes. In the end, it's the same AL2023, with better security.

So, that's why I'd like to stay with the current auto-update until there's an issue ticket. Until then, let's assume that our users wants to pull the latest AL2023 (imagine it like latest Ubuntu 24.04 -- and not 24.04.3 or 24.04.4 for example)  

Would you prefer I remove the provision script entirely and let users manually update if desired?

electricalgorithm:
> I still don't want new commits to Lima every two (just to bump a snapshot version) every two weeks.

It should not be. We can update it on every two weeks by the hack script, but in the end provision script will handle it for us automatically.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
electricalgorithm: I can cut a ticket to the team, but I think it should not be the concern of this PR. We can go with the current upstream approach, and when the fix is there, remove the patch here.

electricalgorithm:
Hey people,

Sorry for being late. Quite busy months in Winter, as you can imagine. I updated the PR with a change to the latest image and addressed your comments. Please review the code once more, although it does not have much changes.

There seems to be issue on GitHub's CI checks. The Lints job failed due to a README's link issue which the PR does not have any modification. I believe there's another issue there. On the vz's integation test job, I couldn't understand if that was my changes breaking it, but after discussing with GitHub's Copilot, looks like it is regarding to setup of the job.

Can you verify that the failures are not related to the PR? If all looks good, please check the version string in the template -- which is 1.2.1 right now and check if it's possible to merge it until the release window closes.

Kindest regards,
Gyokhan 

cc: @afbjorklund @AkihiroSuda

templates/amazonlinux.yaml
```diff
@@ -0,0 +1 @@
+amazonlinux-2023.yaml
```
AkihiroSuda, line 1: templates/README.md has to be updated too

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
```diff
@@ -0,0 +1,29 @@
...
+	fi
+
+	# Cloud-init fails to mount 'mount0' because it doesn't recognize it as a device,
+	# so we extract the intended path from user-data and mount it manually.
+	USER_DATA="/var/lib/cloud/instance/user-data.txt"
+	if [ -f "${USER_DATA}" ]; then
+		MOUNT_POINT=$(grep "mount0" "${USER_DATA}" | awk -F, '{print $2}' | tr -d '[:space:]')
+		if [ -n "${MOUNT_POINT}" ] && ! mountpoint -q "${MOUNT_POINT}"; then
+			mkdir -p "${MOUNT_POINT}"
+			mount -t virtiofs mount0 "${MOUNT_POINT}" || true
```
AkihiroSuda: What about mount1, mount2, ... ?
Only mount0 fails?

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
```diff
@@ -0,0 +1,29 @@
...
+	# 1. Create missing mount.virtiofs helper
+	if [ ! -e /sbin/mount.virtiofs ]; then
+		cat >/sbin/mount.virtiofs <<'EOF'
+#!/bin/sh
+exec mount -i -t virtiofs "$@"
+EOF
+		chmod +x /sbin/mount.virtiofs
+	fi
+
+	# Cloud-init fails to mount 'mount0' because it doesn't recognize it as a device,
```
AkihiroSuda: What is the cause of the issue?
Why does this issue only happen with AL?

templates/amazonlinux-2023.yaml
AkihiroSuda: > Would you prefer I remove the provision script entirely and let users manually update if desired?

That seems more consistent with other templates

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
electricalgorithm: Although the kernel includes the support for virtiofs, there's no helper script under `/sbin`, therefore cloudinit fails. The first part of the patch creates the mount script and makes it executable.

```
[gokhankocmarli@lima-al2023-test ~]$ ls -la /sbin/mount.*
-rwxr-xr-x. 1 root root 200744 Jan 31  2023 /sbin/mount.fuse3
-rwsr-xr-x. 1 root root 219744 Feb  1  2023 /sbin/mount.nfs
lrwxrwxrwx. 1 root root      9 Feb  1  2023 /sbin/mount.nfs4 -> mount.nfs
```
```
[gokhankocmarli@lima-al2023-test ~]$ grep -i virtiofs /proc/filesystems
nodev   virtiofs
```

The second half of the script tries to mount all the mount points defined the user script if the cloud init failed due to missing helper script.


But, you're right. I'll update the commit to automatically read the user-data and add all the mounts not only mount0.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
electricalgorithm: you're right. I'll update the commit to automatically read the user-data and add all the mounts not only mount0.

templates/amazonlinux.yaml
electricalgorithm: I'm updating the commit to add it to the README. Thanks!

templates/amazonlinux-2023.yaml
electricalgorithm: OK. I removed the provision auto-update script from the template. Let's see if there will be people asking for it.

electricalgorithm:
Please review again :) I updated the code to apply your suggestions.

electricalgorithm:
Looks like the failure in jobs is due to the flakiness. Not related to the change.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
```diff
@@ -0,0 +1,38 @@
...
+#!/bin/sh
+exec mount -i -t virtiofs "$@"
+EOF
+		chmod +x /sbin/mount.virtiofs
+	fi
+
+	# Amazon Linux 2023 requires a workaround for virtiofs mounts.
+	# The mount.virtiofs helper script is missing, and cloud-init
+	# may fail to mount the filesystems. This creates the helper
+	# and manually mounts them.
```
AkihiroSuda, line 22: Does it still produce the valid `/etc/fstab`? If so, just running `mount -a` might be fine here?

electricalgorithm: I start a new AL2023 KVM to check and it seems that the fstab does not have the virtiofs configs. Therefore, mount -a does not work and we need to parse them from cloud init config's user-data. That was a nice catch by the way. I'm not super into OSes, thanks for the advice!

electricalgorithm:
Can you re-run the GitHub jobs and merge it if looks good to you since all conversations are resolved @AkihiroSuda?

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
```diff
@@ -0,0 +1,38 @@
...
+	USER_DATA="/var/lib/cloud/instance/user-data.txt"
+	if [ -f "${USER_DATA}" ]; then
+		# Parse all mount entries from user-data (mount0, mount1, ...)
+		MOUNT_ENTRIES=$(grep -E '^\s*-\s+\[mount[0-9]+,' "${USER_DATA}" || true)
```
AkihiroSuda, line 26: This grep didn't work for me

```console
[root@lima-amazonlinux ~]# head  /var/lib/cloud/instance/user-data.txt
#cloud-config
# vim:syntax=yaml

growpart:
  mode: auto
  devices: ['/']
mounts:
- [lima-648190f135cc8172, /Users/suda, virtiofs, "ro,nofail", "0", "0"]
timezone: Asia/Tokyo

[root@lima-amazonlinux ~]# grep -E '^\s*-\s+\[mount[0-9]+,' /var/lib/cloud/instance/user-data.txt
```

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -eux -o pipefail
+
+# Workaround for Amazon Linux 2023
+if [ -f /etc/os-release ] && grep -q "Amazon Linux" /etc/os-release; then
+	# 1. Create missing mount.virtiofs helper
```
AkihiroSuda, line 10: Why is this helper needed in the first place when we have to execute the mount command by ourselves?

AkihiroSuda: The helper doesn't even seem to exist on other distros, such as almalinux-9

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
AkihiroSuda: Why doesn't it generate fstab?

`/var/log/cloud-init.log`:
```
2026-04-16 07:52:01,442 - util.py[DEBUG]: Cloud-init v. 22.2.2 running 'init-local' at Thu, 16 Apr 2026 07:52:01 +0000. Up 1.95 seconds.
[...]
2026-04-16 07:52:04,763 - cc_mounts.py[DEBUG]: mounts configuration is [['lima-648190f135cc8172', '/Users/suda', 'virtiofs', 'ro,nofail', '0', '0']]
2026-04-16 07:52:04,763 - util.py[DEBUG]: Reading from /etc/fstab (quiet=False)
2026-04-16 07:52:04,763 - util.py[DEBUG]: Read 217 bytes from /etc/fstab
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: Attempting to determine the real name of lima-648190f135cc8172
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: changed lima-648190f135cc8172 => None
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: Ignoring nonexistent named mount lima-648190f135cc8172
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: Attempting to determine the real name of ephemeral0
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: changed default device ephemeral0 => None
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: Ignoring nonexistent default named mount ephemeral0
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: Attempting to determine the real name of swap
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: changed default device swap => None
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: Ignoring nonexistent default named mount swap
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: Skipping nonexistent device named None
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: no need to setup swap
2026-04-16 07:52:04,764 - cc_mounts.py[DEBUG]: No modifications to fstab needed
2026-04-16 07:52:04,764 - handlers.py[DEBUG]: finish: init-network/config-mounts: SUCCESS: config-mounts ran successfully
```

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
AkihiroSuda: @electricalgorithm

How did you test your PR?
In what condition `user-data` can have mounts named `mount0`, `mount1`, etc.?

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
AkihiroSuda: >  I think it should not be the concern of this PR

Why not.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
AkihiroSuda: Maybe related to "AL2023 contains a customized version of cloud-init."

https://docs.aws.amazon.com/linux/al2023/ug/cloud-init.html

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-amazonlinux-virtiofs.sh
AkihiroSuda: ping @electricalgorithm

PR #4347, opened 2025-11-14
Fix template locator docs

Alisha-Rehman:
This pull request fixes the documentation by updating all instances of template:// to template: according to the new version 2.0 specification. This change ensures consistency and correctness across the documentation files.

Changes made:

Replaced template:// with template: in all markdown documentation files.

Removed outdated or unnecessary .code-search files.

Why this is important:

Updating the documentation to match the new specification helps users avoid confusion and ensures that templates are referenced correctly when using Lima.

PR #4472, opened 2025-12-18
Fix template syntax for Docker commands

jezmck:


AkihiroSuda:
Thanks, but this seems to be a duplicate of:
- https://github.com/lima-vm/lima/pull/4347

PR #4474, opened 2025-12-18
templates: k8s: pin Kubernetes version to 1.34

### LINKED ISSUE #4473: k8s.bats seems timing out after the release of Kubernetes v1.35.0
https://github.com/lima-vm/lima/actions/runs/20321450742/job/58377862988

```
./hack/bats/extras/k8s.bats
1..2
# ===== /home/runner/work/lima/lima/hack/bats/extras/k8s =====
Error: The operation was canceled.
```

AkihiroSuda:
After the release of Kubernetes v1.35, the test began to fail as containerd dies on the removal of the CNI config.

Fix #4473

### DIFF
--- a/templates/k8s.yaml
+++ b/templates/k8s.yaml
@@ -54,7 +54,9 @@ provision:
     export DEBIAN_FRONTEND=noninteractive
     apt-get update
     apt-get install -y apt-transport-https ca-certificates curl
-    VERSION=$(curl -L -s https://dl.k8s.io/release/stable.txt | sed -e 's/v//' | cut -d'.' -f1-2)
+    # 1.35 fails: https://github.com/lima-vm/lima/issues/4473
+    VERSION=1.34
+    # VERSION=$(curl -L -s https://dl.k8s.io/release/stable.txt | sed -e 's/v//' | cut -d'.' -f1-2)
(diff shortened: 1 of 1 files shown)


afbjorklund:
Was this handled differently in 1.34? Wondered what actually changed, with 1.35

I could only find https://github.com/kubernetes/kubernetes/commit/8118636321ae09d99fbd5328b9c8245f255807b3

afbjorklund:
I thought that I had seem something similar before, due to wiping shared /etc/cni

* https://github.com/kubernetes/kubeadm/issues/1822

https://github.com/kubernetes/kubernetes/commit/e591190cf2fc07738dc2002629367b49269bdf15

Still wonder how it worked before, it if is the bats test doing the `rm -rf /etc/cni`?

afbjorklund (approved):

AkihiroSuda:
CI still timing out 😞

AkihiroSuda:
Let's just downgrade Kubernetes to v1.34 and call it a day

afbjorklund (approved):
This is sad, would be nice if Kubernetes release was tested in Lima as well... (by k8s). Pinning is better than failing.

But from what I understand of the issue, the problem is with `kubeadm reset` rather than with `kubeadm init`?

AkihiroSuda:
> But from what I understand of the issue, the problem is with `kubeadm reset` rather than with `kubeadm init`?

Seems so

PR #4476, opened 2025-12-18
vz: Fix scrollbar issue with runtime swizzle

stuffbucket:
The Code-Hex/vz library wraps VZVirtualMachineView in an NSScrollView with scrollbars hardcoded to YES (for their zoom feature). This causes ~16px of the VM display to be cut off by scrollbar gutters.

This patch uses Objective-C method swizzling to intercept NSScrollView's setDocumentView: method and disable scrollbars when the document view is a VZVirtualMachineView.

This is a self-contained fix that doesn't require forking the vz library or add any dependencies.

Tested successfully on macOS 26.2 (see screenshot) with Debian 13 (Plasma) and Alpine (serial console).

image 70

### DIFF
--- a/pkg/driver/vz/scrollbar_fix_darwin.m
+++ b/pkg/driver/vz/scrollbar_fix_darwin.m
@@ -1,0 +1,44 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+// scrollbar_fix_darwin.m - Runtime patch to disable scrollbars on VZ display window
+//
+// The Code-Hex/vz library wraps VZVirtualMachineView in an NSScrollView with
+// scrollbars hardcoded to YES (for their zoom feature). This causes ~16px of
+// the VM display to be cut off by scrollbar gutters.
+//
+// This file uses Objective-C method swizzling to intercept NSScrollView's
+// setDocumentView: method and disable scrollbars when the document view is
+// a VZVirtualMachineView.
+//
+// This is a temporary workaround until Code-Hex/vz adds a configuration option
+// to disable scrollbars. See: https://github.com/Code-Hex/vz/issues/XXX
+
+#import <Cocoa/Cocoa.h>
+#import <objc/runtime.h>
+
+// Store original implementation pointer
+static IMP original_setDocumentView = NULL;
+
+// Patched implementation that disables scrollbars for VZ views
+static void patched_setDocumentView(id self, SEL _cmd, NSView *view) {
+    // Call original implementation first
+    ((void (*)(id, SEL, NSView *))original_setDocumentView)(self, _cmd, view);
+    
+    // If the document view is a VZVirtualMachineView, disable scrollbars
+    // The scrollbars cause the VM display to be cut off by ~16px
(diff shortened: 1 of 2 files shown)


stuffbucket:
Adding images showing the behavior with the scrollbars. In the lower left there is some text that is occluded by the scroll.  The scrollbars are not shown relative to the size of the desktop or display, they're hardcoded to be shown all the time. In the console case this seems to not create a problem, but in the desktop case it definitely does.
image 71

image 72
image 73

pkg/driver/vz/scrollbar_fix_darwin.m
```diff
@@ -0,0 +1,44 @@
...
+// The Code-Hex/vz library wraps VZVirtualMachineView in an NSScrollView with
+// scrollbars hardcoded to YES (for their zoom feature). This causes ~16px of
+// the VM display to be cut off by scrollbar gutters.
+//
+// This file uses Objective-C method swizzling to intercept NSScrollView's
+// setDocumentView: method and disable scrollbars when the document view is
+// a VZVirtualMachineView.
+//
+// This is a temporary workaround until Code-Hex/vz adds a configuration option
+// to disable scrollbars. See: https://github.com/Code-Hex/vz/issues/XXX
```
AkihiroSuda, line 15: The link is broken

pkg/driver/vz/scrollbar_fix_darwin.m
```diff
@@ -0,0 +1,44 @@
...
+    // If the document view is a VZVirtualMachineView, disable scrollbars
+    // The scrollbars cause the VM display to be cut off by ~16px
+    if (view != nil && [view isKindOfClass:NSClassFromString(@"VZVirtualMachineView")]) {
+        [(NSScrollView *)self setHasVerticalScroller:NO];
+        [(NSScrollView *)self setHasHorizontalScroller:NO];
+    }
+}
+
+// Constructor attribute ensures this runs before main()
+// This patches NSScrollView before Code-Hex/vz creates any windows
```
AkihiroSuda, line 37: Not a fan of monkey patching AppKit, could this be fixed in Code-Hex/vz ?

stuffbucket: @AkihiroSuda I agree. I tried to keep things simple here, not knowing how likely it would be to get a PR accepted in Code-Hex/vz, but I took some time to address the problem in Code-Hex/vz instead. 

I put together a working solution that lets a caller to Code-Hex/vz implement their own GUI UX. The current GUI UI is still the default (although I fixed the scrollbar issue). The gui-linux example was expanded to show how a custom UX could be generated. 

Code-Hex/vz issue: [Code-Hex/vz/issues/210](https://github.com/Code-Hex/vz/issues/210)
Working implementation: [stuffbucket/vz@v0.0.1-pluggable-window-api](https://github.com/stuffbucket/vz/tree/v0.0.1-pluggable-window-api)

AkihiroSuda: > how likely

Seems likely: https://github.com/Code-Hex/vz/issues/210#issuecomment-3707140123


Let me mark this PR as a draft until the Code-Hex/vz issue can be resolved

PR #4482, opened 2025-12-19
limactl copy: fix copying directory with rsync

### LINKED ISSUE #4468: [v2 regression] `limactl copy` puts the destination directory in an extra directory
```bash
mkdir -p /tmp/foo/bar
limactl copy -r /tmp/foo default:/tmp/foo
lima ls -lR /tmp/foo
```

Lima v1.2.2
```
/tmp/foo:
total 0
drwxr-xr-x 2 suda suda 40 Dec 18 10:52 bar

/tmp/foo/bar:
total 0
```

Lima v2.0.2
```
/tmp/foo:
total 0
drwxr-xr-x 3 suda suda 60 Dec 18 10:52 foo

/tmp/foo/foo:
total 0
drwxr-xr-x 2 suda suda 40 Dec 18 10:52 bar

/tmp/foo/foo/bar:
total 0
```

AkihiroSuda:
Fix #4468

### DIFF
--- a/cmd/limactl/copy.go
+++ b/cmd/limactl/copy.go
@@ -392,6 +392,28 @@ func rsyncCommand(ctx context.Context, command string, copyPaths []*copyPath, ve
 		rsyncArgs = append(rsyncArgs, "-e", sshCmd)
 	}
 
+	// Handle trailing slash for directory copies to keep consistent behavior with scp,
+	// which was the original implementation of `limactl copy -r`.
+	// https://github.com/lima-vm/lima/issues/4468
+	if recursive {
+		for i, cp := range copyPaths {
+			//nolint:modernize // stringscutprefix: HasSuffix + TrimSuffix can be simplified to CutSuffix
+			if strings.HasSuffix(cp.path, "/") {
+				if cp.isRemote {
+					for j, cp2 := range copyPaths {
+						if i != j {
+							cp2.path = strings.TrimSuffix(cp2.path, "/")
+						}
+					}
+				} else {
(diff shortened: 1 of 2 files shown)


cmd/limactl/copy.go
```diff
@@ -392,6 +392,28 @@ func rsyncCommand(ctx context.Context, command string, copyPaths []*copyPath, ve
 		rsyncArgs = append(rsyncArgs, "-e", sshCmd)
 	}
 
+	// Handle trailing slash for directory copies to keep consistent behavior with scp,
+	// which was the original implementation of `limactl copy -r`.
+	// https://github.com/lima-vm/lima/issues/4468
+	if recursive {
+		for i, cp := range copyPaths {
```
unsuman, line 399: I think we only need to care about the source's trailing slash when using `rsync`, but I am getting confused with how `scp` works with the prev Lima version? How does `scp` handle trailing slash on the source and the destination side?
Like for example, before lima v2, what should this result to?
```
mkdir -p /tmp/foo/bar
limactl copy -r /tmp/foo/ default:/tmp/foo
```

AkihiroSuda:
```console
$ limactl -v
limactl version 1.2.2

$ mkdir -p /tmp/foo/bar

$ limactl copy -r /tmp/foo/ default:/tmp/foo

$ lima find /tmp/foo
/tmp/foo
/tmp/foo/bar
```

Regardless to whether we like this behavior or not, we should keep the behavior consistent across multiple releases of Lima.

Issue #4263, opened 2025-10-30
Add `--allow-env` and `--block-env` options for `limactl shell`


jandubois:
# Copied from #4036

Optional (can be added after 2.0), but I think worthwhile and easy to implement:

* Add `--allow-env PATTERN` and `--block-env PATTERN` options to `limactl shell`.

They would (just for this command) append the pattern to the respective lists and automatically enable `--preserve-env`.

These commands:

```shell
limactl shell --allow-env "GITHUB_TOKEN" default gh ...
limactl shell --block-env "PRIVATE*" default ...
```

would be equivalent to

```shell
LIMA_SHELLENV_ALLOW="$LIMA_SHELLENV_ALLOW,GITHUB_TOKEN" limactl shell --preserve-env default gh ...
LIMA_SHELLENV_BLOCK="$LIMA_SHELLENV_BLOCK,PRIVATE*" limactl shell --preserve-env default ...
```

Except in case `LIMA_SHELLENV_BLOCK` was empty, it would be adding the pattern to the default list, not the empty list:

```shell
LIMA_SHELLENV_BLOCK="+PRIVATE*" limactl shell --preserve-env default ...
```

vax-r:
@jandubois 
I've opened a PR for it, if you have time you might want to take a look , thanks !

PR #4483, opened 2025-12-21
Add allow-env and block-env option for limactl shell

vax-r:
## Summary
Implement allow-env and block-env option for limactl shell command, which is equivalent of setting
environment variable "LIMA_SHELLENV_ALLOW" and
"LIMA_SHELLENV_BLOCK".

## Related issue
#4263

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -64,6 +64,8 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("reconnect", false, "Reconnect to the SSH session")
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
+	shellCmd.Flags().StringSlice("allow-env", []string{}, "Comma-separated list of environment variable patterns to allow when --preserve-env is set (overrides LIMA_SHELLENV_ALLOW)")
+	shellCmd.Flags().StringSlice("block-env", []string{}, "Comma-separated list of environment variable patterns to allow when --preserve-env is set (overrides LIMA_SHELLENV_BLOCK)")
 	return shellCmd
 }
 
@@ -216,8 +218,16 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	allowListRaw, err := cmd.Flags().GetStringSlice("allow-env")
+	if err != nil {
+		return err
+	}
+	blockListRaw, err := cmd.Flags().GetStringSlice("block-env")
+	if err != nil {
+		return err
+	}
 	if preserveEnv {
-		filteredEnv := envutil.FilterEnvironment()
+		filteredEnv := envutil.FilterEnvironment(allowListRaw, blockListRaw)
 		if len(filteredEnv) > 0 {
 			envPrefix = "env "
 			for _, envVar := range filteredEnv {
--- a/pkg/envutil/envutil.go
+++ b/pkg/envutil/envutil.go
@@ -56,14 +56,19 @@ func validatePattern(pattern string) error {
 }
 
 // getBlockList returns the list of environment variable patterns to be blocked.
-func getBlockList() []string {
-	blockEnv := os.Getenv("LIMA_SHELLENV_BLOCK")
-	if blockEnv == "" {
-		return defaultBlockList
+func getBlockList(blockListRaw []string) []string {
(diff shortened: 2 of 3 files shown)


cmd/limactl/shell.go
```diff
@@ -64,6 +64,8 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("reconnect", false, "Reconnect to the SSH session")
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
+	shellCmd.Flags().String("allow-env", "", "Comma-separated list of environment variable patterns to allow when --preserve-env is set (overrides LIMA_SHELLENV_ALLOW)")
```
AkihiroSuda: Should use https://pkg.go.dev/github.com/spf13/pflag#StringSlice

vax-r:
@AkihiroSuda 
Thanks for your review and suggestion !
I've done the changes you mentioned and use `StringSlice()` instead of `string`.
Let me know if anything needs to be added or changes.

Btw, the test starting WSL2 seems to fail, I've read the log I don't think it's triggered by this commit change?

AkihiroSuda (approved):
Thanks

cmd/limactl/shell.go
```diff
@@ -64,6 +64,8 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("reconnect", false, "Reconnect to the SSH session")
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
+	shellCmd.Flags().StringSlice("allow-env", []string{}, "Comma-separated list of environment variable patterns to allow when --preserve-env is set (overrides LIMA_SHELLENV_ALLOW)")
```
jandubois, line 67: #4263 says:
> They would (just for this command) append the pattern to the respective lists and automatically enable `--preserve-env`.

These options are convenience methods to make it easier to append custom patterns for just a single command (otherwise you would change the env variables if you want the changes to persist for longer). So it makes no sense to require the user to also specify `--preserve-env` as well.

```suggestion
	shellCmd.Flags().StringSlice("allow-env", []string{}, "Comma-separated list of additional environment variable patterns to allow (implies --preserve-env)")
```

cmd/limactl/shell.go
```diff
@@ -64,6 +64,8 @@ func newShellCommand() *cobra.Command {
...
+	shellCmd.Flags().StringSlice("block-env", []string{}, "Comma-separated list of environment variable patterns to allow when --preserve-env is set (overrides LIMA_SHELLENV_BLOCK)")
```
jandubois, line 68: Same as the line above, but the description here has a cut&pasto: needs to say `to block` instead of `to allow`.

pkg/envutil/envutil.go
```diff
@@ -56,15 +56,20 @@ func validatePattern(pattern string) error {
...
 // getBlockList returns the list of environment variable patterns to be blocked.
-func getBlockList() []string {
-	blockEnv := os.Getenv("LIMA_SHELLENV_BLOCK")
-	if blockEnv == "" {
-		return defaultBlockList
+func getBlockList(blockListRaw []string) []string {
+	var shouldAppend bool
+	patterns := blockListRaw
+	if len(patterns) == 0 {
+		blockEnv := os.Getenv("LIMA_SHELLENV_BLOCK")
```
jandubois, line 63: You are replacing the environment variable with the content of the `--block-env` option. #4263 specifies that the option has to be appended to the env variable list. In addition, it says it has to be appended to the default block list if the env variable is empty.

Same thing applies to the allow list (with the exception that there is no default allow list).

Please re-read the issue, and implement the functionality as specified. Or start a discussion in the issue if you think the requested semantics are wrong!

AkihiroSuda:
ping @vax-r

Could you check the comments from @jandubois ?

vax-r:
@jandubois , @AkihiroSuda -
Thanks for your suggestion !
Sorry for the late reply, I am busy at work recently, I'll come back to it during chinese new year, is that acceptable ?

jandubois:
> I'll come back to it during chinese new year, is that acceptable ?

Of course; this is open source. You work on it when you have the time.

You might miss the 2.1 release, but there will always be a 2.2 after that... 😄

jandubois:
It is too late now to make it into 2.1

PR #4489, opened 2025-12-22
Add limactl watch command and JSON events for test scripts

### LINKED ISSUE #4388: Test scripts shouldn't grep log strings
The log strings change day by day; the test scripts shouldn't expect the log strings to be greppable.

Instead they should use `limactl ls --json` or `limactl watch --json` (to be designed and implemented) to obtain proper JSON objects

https://github.com/lima-vm/lima/blob/68be6f2b9bfc4ca171e626917057f74f2c78dbbf/hack/test-templates.sh#L335-L367

https://github.com/lima-vm/lima/blob/68be6f2b9bfc4ca171e626917057f74f2c78dbbf/hack/test-port-forwarding.pl#L148-L153

https://github.com/lima-vm/lima/blob/68be6f2b9bfc4ca171e626917057f74f2c78dbbf/hack/test-port-forwarding.pl#L226-L230

(Probably more)

IrvingMg:
Adds `limactl watch` to stream port forwarding and vsock events, and updates test scripts to parse JSON events instead of grepping log strings. 

Fixes #4388.

### DIFF
--- a/cmd/limactl/watch.go
+++ b/cmd/limactl/watch.go
@@ -1,0 +1,302 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/rjeczalik/notify"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/lima-vm/lima/v2/pkg/hostagent/events"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
+	"github.com/lima-vm/lima/v2/pkg/store"
+)
+
+func newWatchCommand() *cobra.Command {
+	watchCommand := &cobra.Command{
+		Use:   "watch [INSTANCE]...",
+		Short: "Watch events from instances",
+		Long: `Watch events from Lima instances.
+
+Events include status changes (starting, running, stopping), port forwarding
+events, and other instance lifecycle events.
+
+If no instance is specified, events from all instances are watched,
+including newly created instances.
+
+The command will continue watching until interrupted (Ctrl+C).`,
+		Example: `  # Watch events from all instances:
+  $ limactl watch
+
+  # Watch events from a specific instance:
+  $ limactl watch default
+
+  # Include historical events:
+  $ limactl watch --history default
+
+  # Show verbose output (host agent logs, etc.):
+  $ limactl watch --verbose
+
+  # Watch events in JSON format (for scripting):
+  $ limactl watch --json default`,
+		Args:              WrapArgsError(cobra.ArbitraryArgs),
+		RunE:              watchAction,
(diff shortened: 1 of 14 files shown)


cmd/limactl/watch.go
```diff
@@ -0,0 +1,240 @@
...
+			return err
+		case ev := <-eventCh:
+			if jsonFormat {
+				we := watchEvent{Instance: ev.instance, Event: ev.event}
+				j, err := json.Marshal(we)
+				if err != nil {
+					fmt.Fprintf(cmd.ErrOrStderr(), "error marshaling event: %v\n", err)
+					continue
+				}
+				fmt.Fprintln(cmd.OutOrStdout(), string(j))
```
AkihiroSuda: `OutOrStdout` and `ErrOrStderr` should be called only once before entering the loop

IrvingMg: Done. I moved them outside the loop.

IrvingMg:
It appears that the test `test / Integration tests (vz) (default.yaml)` is flaky. Another failure on a different branch can be found here: https://github.com/lima-vm/lima/actions/runs/20450115858/job/58761257374

cmd/limactl/watch.go
AkihiroSuda, line 1: Probably the watch command should not print events that occurred before running the watch command, at least by default

AkihiroSuda, line 1: I tried running `limactl watch` and then running `limactl start` in another terminal, but no new event was printed in the `limactl watch` terminal.

The events were printed only after running `limactl watch` again.

cmd/limactl/watch.go
```diff
@@ -0,0 +1,225 @@
...
+		return err
+	}
+
+	// Determine which instances to watch
+	var instNames []string
+	if len(args) > 0 {
+		instNames = args
+	} else {
+		// Watch all instances
+		allInstances, err := store.Instances()
```
jandubois: This fetches a list of existing instances at the time `limactl watch` is started. It will not start watching any instances that are created later.

This would need some fsnotify/inotify mechanism to see when new instances are created/started.

I think this explains @AkihiroSuda's problem that he doesn't see any events for the instance he starts.

jandubois: Even for existing instances there is a problem when the instance is stopped and restarted. Lima will delete and recreate the log files. But `TailFile()` is called without `ReOpen`, so it will keep watching the closed old file instead of waiting for the file to be recreated, and then watching the new file. I think using `MustExist` is also going to be a problem, but I don't have time now to do a proper review.

cmd/limactl/watch.go
IrvingMg: Done. I've added a `--history` flag to print events before starting the watch.

cmd/limactl/watch.go
IrvingMg: Thanks for the explanation. I've added an fsnotify watcher to detect new instances and changed the TailFile config to ReOpen: true and MustExist: false to handle log file recreation on restart.

cmd/limactl/watch.go
IrvingMg: Done. I've addressed this issue.

IrvingMg:
> Thanks, but please squash the commits https://lima-vm.io/docs/dev/git/#squashing-commits

Done.

PR #4394, opened 2025-11-22
feat: add `vmnet_network` support on macOS 26+

norio-nomura:
## [vmnet_network](https://developer.apple.com/documentation/vmnet/vmnet_network_create(_:_:)?language=objc) support on macOS 26+
- On `vmType: vz`, it uses [`VZVmnetNetworkDeviceAttachment`](https://developer.apple.com/documentation/virtualization/vzvmnetnetworkdeviceattachment?language=objc)
- On `vmType: qemu|krunkit`, it uses [`vmnet_interface_start_with_network`](https://developer.apple.com/documentation/vmnet/vmnet_interface_start_with_network(_:_:_:_:)?language=objc)

Depends on https://github.com/Code-Hex/vz/pull/205

### About what can be done and the restrictions with this API:
- Supports using `vmnet` without root privilege
- Supports `VMNET_SHARED_MODE`, and `VMNET_HOST_MODE`
- Like `vzNAT`, the "bridge10x" interface is created on the host
- Not support `VMNET_BRIDGED_MODE`
  - documented in [vmnet_network_configuration_create](https://developer.apple.com/documentation/vmnet/vmnet_network_configuration_create(_:_:)?language=objc)
- To share `vmnet_network_ref` with multiple processes, it must be the same executable file.
  - documented in [- initWithNetwork:](https://developer.apple.com/documentation/virtualization/vzvmnetnetworkdeviceattachment/init(network:)?language=objc)
  - This restriction seems to be imposed by `VZVmnetNetworkDeviceAttachment` instead of `vmnet_network_create`.
  - ~~`CDHash` is probably used to determine "same executable file". [TN3126: Inside Code Signing: Hashes](https://developer.apple.com/documentation/Technotes/tn3126-inside-code-signing-hashes?language=objc)~~
    - ~~Redoing `codesign` with "Developer ID", where the hash of the executable file changes, does not interfere with the sharing of `vmnet_network_ref`.~~
    - ~~`CDHash` does not change before and after redoing `codesign` with "Developer ID".~~
    - ~~Even if it is `codesign` using the same "Developer ID", if `CDHash` is changed due to source code changes, etc., `vmnet_network_ref` cannot be shared.~~
  - **On macOS 26.2, "same executable file" restriction seems to be relaxed; The external vz driver can connect to subnet that created by `limactl` process.**
- The network created based on `vmnet_network_ref` is owned by the process that created `vmnet_network_ref`, not the process that started the network based on the shared `vmnet_network_ref`.
[...]

PR #4490, opened 2025-12-23
pkg/driver/qemu: Cancel Context on exiting QEMU command

norio-nomura:
By applying this change, go routines started in `LimaQemuDriver.Start` can stop on exiting QEMU command by waiting Context.

refs: https://github.com/lima-vm/lima/pull/4394#issuecomment-3681881048


norio-nomura:
Thanks! 🙏🏻

PR #4500, opened 2025-12-30
feat: Add `limactl ssh` as alias to `limactl shell`

### LINKED ISSUE #3613: Add `limactl ssh` as alias to `limactl shell`
### Description

The common way to run commands and get a shell on a vm is ssh. Common tools provide ssh command to make this easier:

- minikube ssh
- virtctl ssh

But in lima this is called `shell`. Maybe it is more correct, since we can theoretically interact with the vm by other means, but practically this is always ssh.

I try to use `limactl ssh` by mistake many times.

This should be a trivial change in the Cobra command.

Downside: we will not be able to use ssh for another command in the future.

Discussed in #3608

majiayu000:
Fixes #3613

## Changes
- Add `ssh` as Cobra alias for the `shell` command

This follows the existing pattern used by other commands (e.g., `rm` for `delete`, `ls` for `list`, `cp` for `copy`).

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -48,6 +48,7 @@ Hint: try --debug to show the detailed logs, if it seems hanging (mostly due to 
 func newShellCommand() *cobra.Command {
 	shellCmd := &cobra.Command{
 		Use:               "shell [flags] INSTANCE [COMMAND...]",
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
For now can we just print a hint like "No such command: `limactl ssh`. Did you mean `limactl shell`?"

PR #4502, opened 2026-01-02
Suggest shell subcommand when user writes ssh

### LINKED ISSUE #3613: Add `limactl ssh` as alias to `limactl shell`

jandubois:
Fixes #3613

Ref https://github.com/lima-vm/lima/pull/4500#issuecomment-3704248746

```console
❯ l ssh
FATA[0000] unknown command "ssh" for "limactl"

Did you mean this?
	shell
```

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -48,6 +48,7 @@ Hint: try --debug to show the detailed logs, if it seems hanging (mostly due to 
 func newShellCommand() *cobra.Command {
 	shellCmd := &cobra.Command{
 		Use:               "shell [flags] INSTANCE [COMMAND...]",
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #4503, opened 2026-01-04
fix: Merge VMOpts from configurations

### LINKED ISSUE #4427: configuring rosetta in override.yaml has no effect
### Description

hello thank you for Lima it's really cool

i have this in `~/.lima/_config/override.yaml` (and/or `default.yaml`, it doesn't seem to matter):

```yaml
vmOpts:
  vz:
    rosetta:
      enabled: true
      binfmt: true
```

but if i

```shell
limactl create --tty=false &&
limactl start &&
lima cat /proc/sys/fs/binfmt_misc/rosetta
```
i get `cat: /proc/sys/fs/binfmt_misc/rosetta: No such file or directory (os error 2)`

in other words i have to remember to create with `--rosetta`

not the biggest problem but i would like my teammates to be able to `limactl create` and have it do the right thing for our environment (running the occasional arm64 container on an aarch64 host), and it seems like it ought to work from the comments in the default template

Lima: 2.0.2 from Homebrew
Host: MacOS 15

TatsuyaHayashino:
This commit fixes an issue where VMOpts configurations were not being merged from default.yaml, lima.yaml, and override.yaml. The entire VMOpts field was being ignored during the configuration merging process.

Fixes #4427

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -777,6 +777,12 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 	maps.Copy(param, o.Param)
 	y.Param = param
 
+	vmOpts := make(limatype.VMOpts)
+	maps.Copy(vmOpts, d.VMOpts)
+	maps.Copy(vmOpts, y.VMOpts)
+	maps.Copy(vmOpts, o.VMOpts)
+	y.VMOpts = vmOpts
+
 	if y.CACertificates.RemoveDefaults == nil {
 		y.CACertificates.RemoveDefaults = d.CACertificates.RemoveDefaults
 	}
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -419,6 +419,11 @@ func TestFillDefault(t *testing.T) {
 			Shell:   ptr.Of("/bin/tcsh"),
 			UID:     ptr.Of(uint32(8080)),
 		},
+		VMOpts: limatype.VMOpts{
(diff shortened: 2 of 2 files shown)


pkg/limayaml/defaults.go
```diff
@@ -777,6 +777,12 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 	maps.Copy(param, o.Param)
 	y.Param = param
 
+	vmOpts := make(limatype.VMOpts)
+	maps.Copy(vmOpts, d.VMOpts)
+	maps.Copy(vmOpts, y.VMOpts)
+	maps.Copy(vmOpts, o.VMOpts)
+	y.VMOpts = vmOpts
```
alexandear, line 784: nit: We can do the same without creating intermediate map. But it's up to your whether to fix or skip.

```suggestion
	if y.VMOpts == nil {
		y.VMOpts = make(limatype.VMOpts)
	}
	maps.Copy(y.VMOpts, d.VMOpts)
	maps.Copy(y.VMOpts, o.VMOpts)
```

jandubois (approved):
Thanks, LGTM

There is one potential issue: this PR replaces the whole `vmOpts` structure with the version that has the highest priority.

In general we use more fine-grained merging at the key level, with sometimes custom merge semantics for lists.

However, this could only be implemented properly if we move the merge logic for `vmOpts` into the driver interface, and it is unclear how that ultimately should work.

That's why I'm going to merge this now as-is, but we should consider this behaviour "experimental", i.e. subject to change as the driver API evolves.

PR #4516, opened 2026-01-10
cidata: strip control characters from user comment

hekki:
## What this PR does

Removes Unicode Cc (Control) and Cf (Format) characters from the user comment before rendering CIDATA templates.

## Why

Some host user “full name” / GECOS fields can contain invisible control characters (e.g. 0x1e).

When propagated into lima.env, those characters can break shell parsing and lead to cascading failures during boot / cloud-init (e.g. user creation failing, SSH readiness never reached).

## How

Adds removeControlChars() and applies it to templateArgs.Comment.

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -19,6 +19,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"github.com/docker/go-units"
 	"github.com/sirupsen/logrus"
@@ -129,7 +130,7 @@ func templateArgs(ctx context.Context, bootScripts bool, instDir, name string, i
 		Name:               name,
 		Hostname:           hostname.FromInstName(name), // TODO: support customization
 		User:               *instConfig.User.Name,
-		Comment:            *instConfig.User.Comment,
+		Comment:            removeControlChars(*instConfig.User.Comment),
 		Home:               *instConfig.User.Home,
 		Shell:              *instConfig.User.Shell,
 		UID:                *instConfig.User.UID,
@@ -471,6 +472,16 @@ func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name strin
 	}
 
 	return iso9660util.Write(filepath.Join(instDir, filenames.CIDataISO), "cidata", layout)
+}
+
+func removeControlChars(s string) string {
+	out := make([]rune, 0, len(s))
+	for _, r := range s {
+		if unicode.IsPrint(r) {
+			out = append(out, r)
+		}
+	}
+	return string(out)
 }
 
 func getCert(content string) Cert {


pkg/cidata/cidata.go
```diff
@@ -473,6 +474,18 @@ func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name strin
 	return iso9660util.Write(filepath.Join(instDir, filenames.CIDataISO), "cidata", layout)
 }
 
+func removeControlChars(s string) string {
+	out := make([]rune, 0, len(s))
+	for _, r := range s {
+		// Cc: Control, Cf: Format
+		if unicode.Is(unicode.Cc, r) || unicode.Is(unicode.Cf, r) {
```
AkihiroSuda: Can we just use https://pkg.go.dev/unicode#IsControl

hekki: Thank you for the comment. I've modified it to use IsControl.

AkihiroSuda: Please squash the commits

pkg/cidata/cidata.go
```diff
@@ -129,7 +130,7 @@ func templateArgs(ctx context.Context, bootScripts bool, instDir, name string, i
 		Name:               name,
 		Hostname:           hostname.FromInstName(name), // TODO: support customization
 		User:               *instConfig.User.Name,
-		Comment:            *instConfig.User.Comment,
+		Comment:            removeControlChars(*instConfig.User.Comment),
```
AkihiroSuda, line 133: Do we need to remove control chars from other string fields as well ?

jandubois: I don't think so, we can wait until we get an actual bug report. I can see unusual characters in the comment field, but if they are in e.g. the user name, then we should just fail and let the user override the field explicitly.

pkg/cidata/cidata.go
```diff
@@ -473,6 +474,17 @@ func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name strin
 	return iso9660util.Write(filepath.Join(instDir, filenames.CIDataISO), "cidata", layout)
 }
 
+func removeControlChars(s string) string {
+	out := make([]rune, 0, len(s))
+	for _, r := range s {
+		if unicode.IsControl(r) || unicode.Is(unicode.Cf, r) {
+			continue
+		}
+		out = append(out, r)
```
jandubois: I would rather filter out all non-printable character. The current implementation would still leave e.g line and paragraph separators, which are not considered control characters by Unicode (Zl and Zp categories, there are likely others).

```suggestion
		if unicode.IsPrint(r) {
			out = append(out, r)
		}
```

pkg/cidata/cidata.go
hekki: In my environment, the OS's comment field contained control characters, which caused cloud-init/useradd to fail. Therefore, I only sanitized this specific field. For values other than "Comment", I therefore agree with @jandubois's approach.

pkg/cidata/cidata.go
hekki: Thank you, using unicode.IsPrint should help remove more problematic characters.

jandubois:
Sorry, just noticed the commits aren't squashed. We require commits to be squashed unless they address separate issues.

hekki:
> Sorry, just noticed the commits aren't squashed. We require commits to be squashed unless they address separate issues.

Sorry, I forgot about it by accident. I've made the correction.

PR #4522, opened 2026-01-12
Stub out HCL dependencies

jandubois:
The latest version of `yqlib` depends on the `hashicorp/hcl` module, which in turn depends on `apparentlymart/go-textseg`. That one has a custom license and is therefore not approved for use by CNCF projects.

`yqlib` already supports the `yq_nohcl` tag to build without HCL support. Unfortunately the go build system is not tag-aware and always includes all indirect dependencies in `go.mod`/`go.sum`.

This commit stubs out the HCL dependencies, so they will not be fetched, and unconditionally adds the `yq_nohcl` build tag.

Any downstream project will have to do the same, as the `replace` directives in `go.mod` only apply to this repo and are not inherited via the Go build system.

This PR addresses the concerns in https://github.com/lima-vm/lima/pull/4460#pullrequestreview-3580996785.


jandubois:
The BATS test fails because of https://github.com/mikefarah/yq/issues/2572.

I have not checked why the other integration tests failed.

jandubois:
> Any downstream project will have to do the same, as the `replace` directives in `go.mod` only apply to this repo and are not inherited via the Go build system.

I have not tested this, but we probably should move the stubs from `./internal` to `./pkg` so downstream users of the Lima packages can just add the `replace` directory and don't have to create their own stubs.

go.mod
```diff
@@ -3,6 +3,14 @@ module github.com/lima-vm/lima/v2
 
 go 1.24.0
 
+// Avoid pulling in the real HCL module, which depends on github.com/apparentlymart/go-textseg/v15
+// which uses a nonstandard license and therefore cannot be used by CNCF projects.
+replace github.com/hashicorp/hcl/v2 => ./internal/hclstub
```
AkihiroSuda, line 8: Thanks, but not a fan of this complication.
Can we just ask CNCF to permit using HCL? That will help other CNCF projects too.

jandubois: > Can we just ask CNCF to permit using HCL? That will help other CNCF projects too.

Maybe, but I'm not optimistic. The problem is not HCL but github.com/apparentlymart/go-textseg, which has 3 different licenses for different pieces, none of which have been approved.

Can you open a ticket with CNCF about that?

> Thanks, but not a fan of this complication.

Me neither, but what is the alternative (assuming we don't get CNCF to approve the license)? Vendoring all of yqlib and manually removing the HCL dependency? That seems even worse that creating the stubs because we will no longer get dependabot notifications for updates, and there will be manual effort every time we bump yqlib.

This is really the best I can think of.

AkihiroSuda: an alternative would be to just execute yq binary

jandubois: You want to make an external installation of `yq` a requirement for each `--yq` option? And then bundle it in CIDATA for use by the guestagent? That is so much more complex than adding a few stub modules to work around a limitation of the Go dependency resolution.

AkihiroSuda: > Can you open a ticket with CNCF about that?

- https://github.com/cncf/foundation/issues/1242

> You want to make an external installation of yq a requirement for each --yq option? And then bundle it in CIDATA for use by the guestagent? 

Not ideal. Let's see how CNCF handles the request.

AkihiroSuda: One of the reasons I hesitate to accept `replace()` is that my gosocialcheck linter (not integrated to CI yet) does not support `replace()` yet:
- https://github.com/AkihiroSuda/gosocialcheck/issues/7

There might be other tools that do not work well with `replace()` too.

jandubois: `go-licenses` doesn't work with it either. The workaround was to run `go mod vendor` and then run `GOFLAGS="-mod=vendor" go-licenses check ...` (see the change in this PR).

Would that work for `gosocialcheck` too?

jandubois:
> The BATS test fails because of [mikefarah/yq#2572](https://github.com/mikefarah/yq/issues/2572).

On closer inspection this does not look like a bug in `yq`. If anything the new behaviour is more consistent than the old, so I've closed that issue.

I've created #4539 to adjust our BATS test to the new behaviour.
 
> I have not checked why the other integration tests failed.

I believe they are just flaky CI. There is no point in re-running because we are still waiting for a license exception for hcl/v2 from CNCF.

jandubois:
Given https://github.com/cncf/foundation/issues/1242#issuecomment-3918648694

> The Governing Board has approved a blanket exception to allow any CNCF project to use go-textseg under the [Unicode-DFS-2016](https://spdx.org/licenses/Unicode-DFS-2016.html) license.

I think we can assume that `hashicorp/hcl/v2` falls under the license exception for `hashicorp/hcl` and we can simply use the `yq` dependency as-is after adding the license exceptions to the scanner setup.

Issue #1307, opened 2023-01-18
Lima VM clock appears to drift


spacedub:
### Description

I have a Debian VM that has been up for 6 days.

Reported date is:
Wed Jan 18 21:30:48 UTC 2023

macos reported date is:
Wed Jan 18 13:48:26 PST 2023

So, 18 minutes apart.

```
limactl version 0.11.3-27-g7a74ca9
```

I am using https://github.com/lima-vm/lima/blob/master/examples/debian.yaml slightly modified for my VM.

I assume installing ntp would solve this, but then (assuming this is right), should this not be by default?

spacedub:
FWIW, `sudo hwclock --hctosys` is not helping.

Let me know if you need more info about my current rig.

afbjorklund:
You can check the  source of the hwclock, from inside the VM (with `lima`):

`cat /sys/devices/system/clocksource/clocksource0/current_clocksource`

```console
anders@lima-default:/home/anders$ cat /sys/devices/system/clocksource/clocksource0/current_clocksource
kvm-clock
```

jandubois:
This is kind of a duplicate of #850 (note that I put a link to a blogpost about how Docker Desktop deals with this into a comment over there).

Also maybe related #939.

You didn't mention your host OS; are you on macOS or Linux?

spacedub:
@jandubois @afbjorklund I am completely out of my depth on this topic.
Is there a short-term workaround (even something really hack-ish) that I can use?

afbjorklund:
I was surprised to hear that NTP wouldn't work (although coarsely), but it probably needs a local server (like DHCP uses)

Would be nice to have a HVF version of the KVM clocks - but it would probably require some modules on the Linux side...

So yeah, it sounds like the same issue as:
* #850

eldondevat:
This workaround worked for me in the short term: https://github.com/lima-vm/lima/issues/850#issuecomment-3729831954

Issue #850, opened 2022-05-09
Clockdrift in the created guest


dirkdev98:
### Description

Running Lima via Colima (https://github.com/abiosoft/colima/), I have occasional clock drift issues. I am unable to find a cause for this or even a way to reproduce it. It seems to be after I lock my laptop and get back the next morning, but only once in a while (every 10 days or so) and not consistently every day.

Versions:
```
❯ uname -a
Darwin Dirks-MacBook-Pro.local 21.4.0 Darwin Kernel Version 21.4.0: Mon Feb 21 20:36:53 PST 2022; root:xnu-8020.101.4~2/RELEASE_ARM64_T8101 arm64

❯ lima --version
limactl version 0.9.2

❯ colima version  
colima version 0.3.4
git commit: 5a4a70481ca8d1e794677f22524e3c1b79a9b4ae

runtime: docker
arch: aarch64
client: v20.10.14
server: v20.10.11
```

`docker run -it --rm --cap-add SYS_TIME --privileged -e ALLOW_CIDR=0.0.0.0/0 -p 123:123/udp geoffh1977/chrony`

```
Writing New Config File

2022-05-07T08:00:43Z chronyd version 4.1 starting (+CMDMON +NTP +REFCLOCK +RTC +PRIVDROP -SCFILTER +SIGND +ASYNCDNS +NTS +SECHASH +IPV6 -DEBUG)
2022-05-07T08:00:43Z Could not read valid frequency and skew from driftfile /var/lib/chrony/chrony.drift
2022-05-07T08:00:43Z Initial frequency -100000.000 ppm
2022-05-07T08:00:47Z System's initial offset : 978.140307 seconds slow of true (step)
2022-05-07T08:00:52Z Backward time jump detected!
2022-05-07T08:02:00Z Selected source 162.159.200.1 (pool.ntp.org)
2022-05-07T08:02:00Z System clock wrong by 978.685231 seconds
2022-05-07T08:18:18Z System clock was stepped by 978.685231 seconds
2022-05-07T08:02:57Z Backward time jump detected!
2022-05-07T08:02:57Z Can't synchronise: no selectable sources
2022-05-07T08:05:07Z Selected source 94.198.159.15 (pool.ntp.org)
2022-05-07T08:05:07Z System clock wrong by 977.327132 seconds
2022-05-07T08:21:24Z System clock was stepped by 977.327132 seconds
2022-05-07T08:05:06Z Backward time jump detected!
[...]

PR #4527, opened 2026-01-14
hostagent: add host-to-guest time synchronization

mfenderov:
The VZ virtualization framework's virtual RTC drifts from host time,
causing timestamp issues with workloads like TestContainers.

This adds a gRPC-based time sync where the host agent periodically
pushes current time to the guest agent, which adjusts the system
clock if drift exceeds 100ms.

Addresses #850, #1307

### DIFF
--- a/pkg/guestagent/api/guestservice.proto
+++ b/pkg/guestagent/api/guestservice.proto
@@ -11,6 +11,7 @@ service GuestService {
   rpc PostInotify(stream Inotify) returns (google.protobuf.Empty);
 
   rpc Tunnel(stream TunnelMessage) returns (stream TunnelMessage);
+  rpc SyncTime(TimeSyncRequest) returns (TimeSyncResponse);
 }
 
 message Info {
@@ -41,4 +42,14 @@ message TunnelMessage {
   bytes data = 3;
   string guest_addr = 4;
   string udp_target_addr = 5;
+}
+
+message TimeSyncRequest {
+  google.protobuf.Timestamp host_time = 1;
+}
+
+message TimeSyncResponse {
+  bool adjusted = 1;
+  int64 drift_ms = 2;
+  string error = 3;
 }
--- a/pkg/guestagent/api/server/server.go
+++ b/pkg/guestagent/api/server/server.go
@@ -6,6 +6,7 @@ package server
 import (
 	"context"
 	"net"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
@@ -14,6 +15,7 @@ import (
 
 	"github.com/lima-vm/lima/v2/pkg/guestagent"
 	"github.com/lima-vm/lima/v2/pkg/guestagent/api"
+	"github.com/lima-vm/lima/v2/pkg/guestagent/timesync"
 	"github.com/lima-vm/lima/v2/pkg/portfwdserver"
 )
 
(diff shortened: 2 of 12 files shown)


pkg/guestagent/api/server/server.go
AkihiroSuda, line 1: Why wasn’t this working?
https://github.com/lima-vm/lima/blob/b224e2d3274aeeeba7fe884038572441d934497c/pkg/guestagent/guestagent_linux.go#L217

This can be removed now?

pkg/guestagent/api/guestservice.proto
```diff
@@ -11,6 +11,7 @@ service GuestService {
   rpc PostInotify(stream Inotify) returns (google.protobuf.Empty);
 
   rpc Tunnel(stream TunnelMessage) returns (stream TunnelMessage);
+  rpc SyncTime(TimeSyncRequest) returns (TimeSyncResponse);
```
AkihiroSuda, line 14: This method is more privileged than existing ones, so we have to make sure that an unprivileged process in the VM cannot call this method via vsock.

pkg/guestagent/api/server/server.go
AkihiroSuda: Also curious if we can just periodically poke systemd-timesyncd

mfenderov: Using systemd-timesyncd has trade-offs:
- Requires NTP connectivity (may not be available in isolated/air-gapped VMs)
- More latency (waits for NTP server response vs direct time set)
- The host-based sync is more reliable since the host time is always available

mfenderov: The existing `fixSystemTimeSkew()` uses `/dev/rtc` as the time source. For QEMU, this works because QEMU maintains the RTC accurately. However, for VZ, the virtual RTC drifts together with the system clock after host sleep/wake - both are maintained by the hypervisor and drift together. So syncing system time from RTC doesn't help when both are drifting.

AkihiroSuda: Thanks, `fixSystemTimeSkew` can be now removed?

AkihiroSuda: > isolated/air-gapped VMs

We should support it, but it is not supported currently

pkg/guestagent/api/server/server.go
```diff
@@ -78,3 +84,38 @@ func (s *GuestServer) PostInotify(server api.GuestService_PostInotifyServer) err
 func (s *GuestServer) Tunnel(stream api.GuestService_TunnelServer) error {
 	return s.TunnelS.Start(stream)
 }
+
+const hostCID = 2
+
+func (s *GuestServer) SyncTime(ctx context.Context, req *api.TimeSyncRequest) (*api.TimeSyncResponse, error) {
+	if p, ok := peer.FromContext(ctx); ok {
+		if addr, ok := p.Addr.(*vsock.Addr); ok {
+			if addr.ContextID != hostCID {
```
AkihiroSuda: This check should globally apply to all the methods.
Probably `bind(2)` can be used to limit the connections by CID.

AkihiroSuda: Also, please squash the commits

mfenderov: > This check should globally apply to all the methods. Probably `bind(2)` can be used to limit the connections by CID.

Linux vsock `bind(2)` only sets the local CID/port - there's no socket option to filter by remote CID. Instead, I added a `cidFilteredListener` wrapper that checks the remote CID in `Accept()` and rejects non-host connections before they reach gRPC. This applies globally to all methods.

pkg/guestagent/api/server/server.go
mfenderov: > We should support it, but it is not supported currently

I can help with it 😉

pkg/guestagent/api/server/server.go
```diff
@@ -78,3 +80,28 @@ func (s *GuestServer) PostInotify(server api.GuestService_PostInotifyServer) err
...
+	resp := &api.TimeSyncResponse{
+		Adjusted: false,
+		DriftMs:  drift.Milliseconds(),
+	}
+
+	const driftThreshold = 100 * time.Millisecond
+	if drift > driftThreshold || drift < -driftThreshold {
+		if err := timesync.SetSystemTime(hostTime); err != nil {
+			logrus.WithError(err).Warn("SyncTime: failed to set system time")
+			return resp, nil
```
jandubois, line 99: Should the `TimeSyncResponse` include an `err` field, so the host would be notified that time sync has failed?

This could maybe one day be reported in `limactl ls`? Because looking at guestagent logs inside the instance is cumbersome.

mfenderov: Good catch! Done 👍

pkg/hostagent/timesync.go
```diff
@@ -0,0 +1,72 @@
...
+		return
+	}
+
+	select {
+	case <-time.After(timeSyncStartupDelay):
+	case <-ctx.Done():
+		return
+	}
+
+	ticker := time.NewTicker(timeSyncInterval)
```
balajiv113, line 32: Minor concern but can we do some perf test and see.

Usually when we start a vm without containerd during idle state cpu usage should come down to almost zero (no works in vm, so no cpu usage as well)

The moment we start sending events from host to guest in a loop, I feel like this will not be the case.

I will also take a look at current perf. But with this ticker we should monitor the cpu as well

balajiv113: I think this is fine. I just noticed CPU spikes to 1% only on these ticker. Previously it's idle around ~0.6%

Issue #3452, opened 2025-04-21
`READY. Run ... to open the shell` is confusing when the shell is already launched.


AkihiroSuda:
```console
$ lima
? Do you want to start the instance now?  Yes
INFO[0000] Starting the instance "default" with VM driver "vz" 
INFO[0000] [hostagent] hostagent socket created at /Users/suda/.lima/default/ha.sock 
INFO[0000] [hostagent] Starting VZ (hint: to watch the boot progress, see "/Users/suda/.lima/default/serial*.log") 
INFO[0001] SSH Local Port: 60022                        
INFO[0001] [hostagent] [VZ] - vm state change: running  
INFO[0001] [hostagent] Waiting for the essential requirement 1 of 2: "ssh" 
INFO[0011] [hostagent] Waiting for the essential requirement 1 of 2: "ssh" 
INFO[0011] [hostagent] The essential requirement 1 of 2 is satisfied 
INFO[0011] [hostagent] Waiting for the essential requirement 2 of 2: "user session is ready for ssh" 
INFO[0011] [hostagent] The essential requirement 2 of 2 is satisfied 
INFO[0011] [hostagent] Waiting for the optional requirement 1 of 2: "systemd must be available" 
INFO[0011] [hostagent] Guest agent is running           
INFO[0011] [hostagent] The optional requirement 1 of 2 is satisfied 
INFO[0011] [hostagent] Forwarding TCP from 127.0.0.1:44273 to 127.0.0.1:44273 
INFO[0011] [hostagent] Waiting for the optional requirement 2 of 2: "containerd binaries to be installed" 
INFO[0011] [hostagent] Not forwarding TCP 127.0.0.53:53 
INFO[0011] [hostagent] Not forwarding TCP 127.0.0.54:53 
INFO[0011] [hostagent] Not forwarding TCP 0.0.0.0:22    
INFO[0011] [hostagent] Not forwarding TCP [::]:22       
INFO[0011] [hostagent] Not forwarding UDP 127.0.0.54:53 
INFO[0011] [hostagent] Not forwarding UDP 127.0.0.53:53 
INFO[0011] [hostagent] The optional requirement 2 of 2 is satisfied 
INFO[0011] [hostagent] Not forwarding UDP 192.168.5.15:68 
INFO[0011] [hostagent] Waiting for the guest agent to be running 
INFO[0011] [hostagent] Waiting for the final requirement 1 of 1: "boot scripts must have finished" 
[...]

PR #4538, opened 2026-01-20
fix: suppress "READY. Run." message

tushar1977:
Issue - #3452

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -132,7 +132,7 @@ func shellAction(cmd *cobra.Command, args []string) error {
 			}
 		}
 
-		err = instance.Start(ctx, inst, false, false)
+		err = instance.Start(instance.WithLaunchingShell(ctx), inst, false, false)
 		if err != nil {
 			return err
 		}
--- a/pkg/instance/start.go
+++ b/pkg/instance/start.go
@@ -342,10 +342,12 @@ func watchHostAgentEvents(ctx context.Context, inst *limatype.Instance, haStdout
 				return false
 			}
 
-			if *inst.Config.Plain {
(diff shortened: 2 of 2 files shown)


cmd/limactl/shell.go
```diff
@@ -132,6 +133,7 @@ func shellAction(cmd *cobra.Command, args []string) error {
 			}
 		}
 
+		ctx = limactlutil.WithShellAttached(ctx)
```
jandubois: I don't think we need an extra package for this; the functionality is specific to starting an instance, so should be part of the `instance` package.

jandubois (changes requested):
I don't think we need (or want) an extra package for this. But if we did, then the new file would need SPDX copyright/license headers. That is the remaining lint error from CI.

We also request that all commits are squashed together. Don't merge `master` into the PR; rebase on latest `master` instead and squash your own changes.

The commit message should explain the changes in the commit. "Fixing the message bug" is not descriptive at all.

Finally the commit needs a DCO signature before we can merge it; that is a CNCF requirement.

cmd/limactl/shell.go
tushar1977: sure

tushar1977:
> I don't think we need (or want) an extra package for this. But if we did, then the new file would need SPDX copyright/license headers. That is the remaining lint error from CI.
> 
> We also request that all commits are squashed together. Don't merge `master` into the PR; rebase on latest `master` instead and squash your own changes.
> 
> The commit message should explain the changes in the commit. "Fixing the message bug" is not descriptive at all.
> 
> Finally the commit needs a DCO signature before we can merge it; that is a CNCF requirement.

Doing it in couple of minutes.

tushar1977:
let me fix linting issues also

tushar1977:
@jandubois i have squashed all my commits in one commit.

jandubois (approved):
Thanks, LGTM

PR #4539, opened 2026-01-20
Fix YQ test with empty input document

jandubois:
The test is using `<<<""` to specify an empty input document. It is however not really empty, but contains a single newline character.

yq versions before 4.50.1 would filter out this newline, but now it is being preserved. This commit changes the test to use a truly empty input document, as was originally intended.

This PR fixes the BATS failure in #4460. It will still be blocked from merging because of the HCL licensing issue.


AkihiroSuda (approved):
Thanks

Issue #3994, opened 2025-09-07
The make lint target is complaining about local changes


afbjorklund:
### Description

```
Please run 'make generate' when making changes to proto files and check-in the generated file changes
make: *** [Makefile:504: check-generated] Error 1
```

It should only do this for *.pb.go, and not for *.go

* https://github.com/lima-vm/lima/pull/3956

I want to be able to run golangci-lint, before I commit

PR #4549, opened 2026-01-27
Makefile: separate local lint from generated checks

suhas-developer07:
### Summary

This PR separates local linting from CI enforcement.

### Changes
- `make lint` now runs linters without requiring a clean git diff
- `make lint-ci` enforces `check-generated` before linting
- CI behavior remains strict and unchanged

### Context
This addresses the developer experience issue raised in #3994 while preserving the reproducibility guarantees introduced in #3956.

### Notes
Local `protolint` failures on generated files are unchanged and out of scope for this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -575,7 +575,7 @@ bats: native limactl-plugins
 	PATH=$$PWD/_output/bin:$$PATH ./hack/bats/lib/bats-core/bin/bats --timing ./hack/bats/tests
 
 .PHONY: lint
-lint:
+lint: check-generated
 	golangci-lint run ./...
 	yamllint .
 	ls-lint
@@ -585,8 +585,17 @@ lint:
 	ltag -t ./hack/ltag --check -v
 	protolint .
 
-.PHONY: lint-ci
-lint-ci: check-generated lint
+.PHONY: lint-local
+lint-local:
+	golangci-lint run ./...
+	yamllint .
+	ls-lint
+	find . -name '*.sh' ! -path "./.git/*" | xargs shellcheck
+	find . -name '*.sh' ! -path "./.git/*" | xargs shfmt -s -d
+	go-licenses check --include_tests ./... --allowed_licenses=$$(cat ./hack/allowed-licenses.txt)
+	ltag -t ./hack/ltag --check -v
+	protolint .
+
 
 .PHONY: clean
 clean:


Makefile
```diff
@@ -585,6 +585,9 @@ lint: check-generated
 	ltag -t ./hack/ltag --check -v
 	protolint .
 
+.PHONY: lint-ci
+lint-ci: check-generated lint
```
AkihiroSuda: Doesn't seem executed on CI?

Anyway, probably this target is not needed and you can just run `make check-generated` and `make lint` on CI instead

afbjorklund: Then again, "check-generated" is the one that is failing if you have made any changes (not only to protobuf)

```
.PHONY: lint
lint: check-generated
```

So that would make the whole thing a no-op, and you could just leave it the way it was - at least for CI.

afbjorklund:
I don't think you need to change anything for CI, I will just add another local target for when you don't want it.

Similar to `make native` when you don't want to build all weird architectures (and there is no way to configure anymore)

Or `make exe` when you just want to build and not want to clean, there could be a target to just run the lint commands.

I was trying with `pre-commit`, but that mostly made it worse - and not all the lint tools have python versions...

Makefile
afbjorklund: As far as I know, CI does not run `make lint`

suhas-developer07:
Thanks for the clarification — that makes sense 

I misunderstood how CI invokes linting and assumed make lint was part of the CI flow. Thanks for pointing out that it isn’t.

I agree that CI behavior shouldn’t change. Adding a separate local-only target for running the lint commands without check-generated sounds like the right approach.

I can update the PR to introduce a local convenience target (e.g. lint-local) that just runs the lint commands, while keeping make lint unchanged. Let me know if that works for you.

Makefile
```diff
@@ -585,6 +585,18 @@ lint: check-generated
...
+lint-local:
+	golangci-lint run ./...
+	yamllint .
+	ls-lint
+	find . -name '*.sh' ! -path "./.git/*" | xargs shellcheck
+	find . -name '*.sh' ! -path "./.git/*" | xargs shfmt -s -d
+	go-licenses check --include_tests ./... --allowed_licenses=$$(cat ./hack/allowed-licenses.txt)
+	ltag -t ./hack/ltag --check -v
+	protolint .
+
```
AkihiroSuda, line 598: The code clone has to be eliminated

AkihiroSuda: Why not just decouple `make lint` from `make check-generated` and call it a day?

PR #4557, opened 2026-02-02
feature(cli): allow multiple template sources on command line

### LINKED ISSUE #3404: Allow additional base templates to be specified during `limactl create`
It should be possible to run this command:

```
limactl create https://example.com/template.yaml ./secrets.yaml
```

And the effect should be the same as if the user has used this local template:

```yaml
base: [https://example.com/template.yaml ./secrets.yaml]
```

This allows the user to supply locally defined settings to an external template.

If `secrets.yaml` had been specified inside `template.yaml`, then it would have been resolved relative to the template location, and not relative to the current directory, i.e. it would have tried to fetch `https://example.com/secrets.yaml`.

For consistency the same should work for `limactl start` too (when it is used to create and start a new instance in a single step).

jkylling:
Enable passing multiple template files/URLs to `limactl create` and `limactl start`.

Example usage:
  limactl create template:docker my-overrides.yaml
  limactl create https://example.com/base.yaml secrets.yaml

This is equivalent to using the `base:` property in YAML:
  base: [https://example.com/template.yaml, $PWD/secrets.yaml]

Fixes #3404

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -26,6 +26,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/limatype/dirnames"
 	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
 	"github.com/lima-vm/lima/v2/pkg/limayaml"
+	"github.com/lima-vm/lima/v2/pkg/localpathutil"
 	"github.com/lima-vm/lima/v2/pkg/networks/reconcile"
 	"github.com/lima-vm/lima/v2/pkg/registry"
 	"github.com/lima-vm/lima/v2/pkg/store"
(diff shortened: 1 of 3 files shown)


jandubois (changes requested):
This PR drops most of the special cases for specifying templates on the command line. Even the simple single-arg case no longer works:

```console
❯ limactl start template:alpine
FATA[0000] cannot derive instance name from "template:alpine": filename "template:alpine" is invalid: identifier "template:alpine" must match ^[A-Za-z0-9]+(?:[._-](?:[A-Za-z0-9]+))*$
```

Using an image URL no longer creates a short instance name:

```console
❯ limactl start https://dl-cdn.alpinelinux.org/alpine/v3.23/releases/cloud/nocloud_alpine-3.23.2-aarch64-uefi-cloudinit-r0.qcow2
? Creating an instance "nocloud_alpine-3.23.2-aarch64-uefi-cloudinit-r0.qcow2" Proceed with the current configuration
...
```

It looks like the special handling for `-` (stdin) only works for the first argument.

I haven't reviewed the code yet, but we should probably start by creating a bunch of BATS tests for the current behaviour, so we can be sure refactoring to implement multiple input templates is not breaking any of them.

PR #4561, opened 2026-02-05
Create reproducible .gz file

bmwiedemann:
We get a bit-identical .gz file output using `gzip -n`

See https://reproducible-builds.org/ for why this is good.


jandubois:
Thanks! Please sign your commit with a DCO!

AkihiroSuda (approved):
Thanks

PR #4566, opened 2026-02-09
docs: add a tip to reduce `-ngl` value in krunkit

unsuman:
Adds a tip suggesting users reduce the `-ngl` parameter if models take too long to load or experience performance issues.


AkihiroSuda (approved):
Thanks

PR #4569, opened 2026-02-10
fix(shell-sync): show file contents in new directories by using `-N` flag

unsuman:
This PR fixes when a new directory was added, the diff command would only report the directory itself and hiding the actual files(and content) inside it.


AkihiroSuda (approved):
Thanks

PR #4570, opened 2026-02-10
fix: panic when run as root

ningmingxiao:
fix 
limactl list
```
panic: must not run as the root user

goroutine 1 [running]:
main.main()
	/home/runner/work/lima/lima/cmd/limactl/main.go:36 +0x110

```


jandubois (approved):
Thanks, LGTM

PR #4573, opened 2026-02-11
CI: set go-version to 'stable' (resolves to 1.26.x currently)

AkihiroSuda:


AkihiroSuda:
Seems to need:
- https://github.com/lima-vm/lima/pull/4571

afbjorklund:
There doesn't seem to be a way to know which go version is "stable", when **not** using GitHub...?

i.e. you have to parse https://github.com/actions/go-versions/blob/main/versions-manifest.json

arixmkii:
> Yeah, but this change is only changing GitHub actions, so does it matter? 

Potentially for LTS branches. But it should be possible to pin the version in release branches, but keep it open via "stable" in master one.

If pinning is a thing for release branches it should be mentioned in comments or developers guide.

afbjorklund:
I'm just wondering how I can test locally to avoid getting failures in CI, but it is harder when all the details are hidden in GitHub Actions under `.github` (instead of e.g. make)

I don't have anything against merging the PR

Issue #4265, opened 2025-10-30
Restore `templates/debian-11.yaml`


jandubois:
As you intended to keep `templates/_images/debian-11.yaml`, you might have removed `templates/debian-11.yaml` by accident?

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/pull/4029#discussion_r2471838440_

jandubois:
We do keep templates for EOL versions of other distros, we just don't link to them from the README.

PR #4574, opened 2026-02-13
restore templates/debian-11.yaml

olamilekan000:
```
change restores debian-11 template
```
fixes: [4265](https://github.com/lima-vm/lima/issues/4265)

### DIFF
--- a/templates/debian-11.yaml
+++ b/templates/debian-11.yaml
@@ -1,0 +1,5 @@
+minimumLimaVersion: 2.0.0
+
+base:
+- template:_images/debian-11
+- template:_default/mounts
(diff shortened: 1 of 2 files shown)


templates/debian-11.yaml
AkihiroSuda, line 1: https://github.com/lima-vm/lima/blob/master/templates/README.md needs to be updated too ?

jandubois: For some reason I believed that we removed templates from the README after their EOL date, but I see that isn't true.

I which case I wonder why we don't list all the Ubuntu-* releases.

olamilekan000: Should I proceed to add to README? Maybe with an indication that it is deprecated

jandubois: Yes, please add it to the README. Don't mark it as deprecated; we don't do that for the other distros either. If we want to do that, then that is an independent change from restoring the missing template.

jandubois (approved):
Thanks, LGTM

PR #4578, opened 2026-02-13
Rename /home/${USER}.linux to /home/${USER}.guest

### LINKED ISSUE #4577: Rename `/home/${USER}.linux` to `/home/${USER}.guest`
`/home/${USER}.linux` doesn't make much sense when the host OS is Linux too, or the guest OS is non-Linux (unimplemented yet).

For new instances I suggest creating the home directory as `/home/${USER}.guest`.
To maintain compatibility,  `/home/${USER}.linux` should be created as a symlink to `{$USER}.guest`.

For existing instances, the symlink should be created in the opposite way: `/home/{$USER}.guest` should be a symlink that points to `${USER}.linux`.

AkihiroSuda:
Fix #4577

Newly created instances now use `/home/${USER}.guest` as the home directory. To maintain compatibility,  `/home/${USER}.linux` is created as a symlink to `${USER}.guest`.

For existing instances, the symlink is created in the opposite direction.

### DIFF
--- a/pkg/osutil/user.go
+++ b/pkg/osutil/user.go
@@ -111,7 +111,6 @@ func LimaUser(ctx context.Context, limaVersion string, warn bool) *user.User {
 			warnings = append(warnings, warning)
 			limaUser.Username = fallbackUser
 		}
-		limaUser.HomeDir = "/home/{{.User}}.linux"
 		if runtime.GOOS == "windows" {
 			idu, err := call(ctx, []string{"id", "-u"})
 			if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/osutil/user.go
```diff
@@ -150,6 +149,14 @@ func LimaUser(ctx context.Context, limaVersion string, warn bool) *user.User {
 	}
 	// Make sure we return a pointer to a COPY of limaUser
 	u := *limaUser
+	limaVersionUnknown := limaVersion == "" || limaVersion == "<unknown>"
+	if limaVersionUnknown || versionutil.GreaterEqual(limaVersion, "2.1.0") {
+		limaUser.HomeDir = "/home/{{.User}}.guest"
+		// boot script symlinks "/home/{{.User}}.linux" to "{{.User}}.guest" for compatibility
+	} else {
+		limaUser.HomeDir = "/home/{{.User}}.linux"
```
jandubois: The code is modifying `limaUser`, which should be treated as constant because it is returned by `once.Do`. This means the function will return the wrong value on the first call. Fix:

```suggestion
	if limaVersionUnknown || versionutil.GreaterEqual(limaVersion, "2.1.0") {
		u.HomeDir = "/home/{{.User}}.guest"
		// boot script symlinks "/home/{{.User}}.linux" to "{{.User}}.guest" for compatibility
	} else {
		u.HomeDir = "/home/{{.User}}.linux"
```

jandubois (changes requested):
Looks mostly good, but has one bug (see below).

Also `pkg/cidata/template_test.go` has a hardcoded `/home/foo.linux`; maybe it should be updated to the new convention?

pkg/osutil/user.go
AkihiroSuda: Thanks, fixed

jandubois (approved):
Thanks, LGTM

PR #4579, opened 2026-02-13
limactl should never ignore errors from store.Inspect

jandubois:


cmd/limactl/shell.go
```diff
@@ -107,10 +107,10 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		}
 		return err
 	}
+	if len(inst.Errors) > 0 {
+		return fmt.Errorf("instance %q has configuration errors: %w", instName, errors.Join(inst.Errors...))
+	}
```
AkihiroSuda, line 112: Even when some errors exist, `limactl shell` should remain as much functional as possible, otherwise user will practically lose data

cmd/limactl/shell.go
```diff
@@ -107,10 +107,10 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
 	if inst.Config == nil {
```
AkihiroSuda, line 113:
```suggestion
	if inst.Config == nil || inst.SSHLocalPort <= 0 {
```

cmd/limactl/shell.go
jandubois: AFAICT the only errors that can happen that would still allow you to connect to the instance are an ill-formed `message` field, and something that make `os.ReadFile(limaVersionFile)` return an error that is not `os.ErrNotExists`.

This is something that is fixable, so there should be no data loss.

All other cases will either have `inst.Config` being `nil` and/or `inst.SSHLocalPort` being 0. Neither of those will allow a connection.

cmd/limactl/shell.go
jandubois: Not really worth it, I think. We already get:

```console
❯ l shell alpine-iso uname -a
Bad port '0'
```

But I will change to always print out the errors, so user can at least tell what is wrong, e.g.

```console
❯ l shell alpine-iso uname -a
ERRO[0000] Instance "alpine-iso" has configuration errors: failed to connect to "/Users/jan/.lima/alpine-iso/ha.sock": stat /Users/jan/.lima/alpine-iso/ha.sock: no such file or directory
Bad port '0'
```

And it will show the error, when they still can connect, like

```console
❯ l shell alpine-iso uname -a
ERRO[0000] Instance "alpine-iso" has configuration errors: message "{{.foo}" is not a valid template: template: format:1: bad character U+007D '}'
Linux lima-alpine-iso 6.12.51-0-virt #1-Alpine SMP PREEMPT_DYNAMIC 2025-10-07 15:12:03 aarch64 Linux
```

cmd/limactl/shell.go
```diff
@@ -107,10 +107,10 @@ func shellAction(cmd *cobra.Command, args []string) error {
 		}
 		return err
 	}
+	if len(inst.Errors) > 0 {
+		logrus.Errorf("Instance %q has configuration errors: %s", instName, errors.Join(inst.Errors...))
```
AkihiroSuda: nit: use `logrus.WithError`

jandubois: Fixed:

```console
❯ l shell alpine-iso uname -a
ERRO[0000] Instance "alpine-iso" has configuration errors  error="failed to connect to \"/Users/jan/.lima/alpine-iso/ha.sock\": stat /Users/jan/.lima/alpine-iso/ha.sock: no such file or directory\nmessage \"{{.foo}\" is not a valid template: template: format:1: bad character U+007D '}'"
Bad port '0'
```

AkihiroSuda (approved):
Thanks

PR #4592, opened 2026-02-16
store.Inspect: fall back to ssh.config for SSH port

### LINKED ISSUE #4583: Extract ssh port from ssh.config
When the host agent dies, we can no longer connect to the instance via `limactl shell` because we query the host agent for the ssh port number.

But even when the host agent is dead, we can still use the `ssh.config` data to connect:

```console
❯ cd ~/.lima/alpine-iso

❯ rm ha.sock

❯ l shell alpine-iso uname -a
Bad port '0'

❯ ssh -F ssh.config lima-alpine-iso uname -a
Linux lima-alpine-iso 6.12.51-0-virt #1-Alpine SMP PREEMPT_DYNAMIC 2025-10-07 15:12:03 aarch64 Linux
```

So if we can still retrieve the instance config (including instance directory), then we have enough information to connect even without the host agent.

jandubois:
When the host agent is unavailable, the SSH local port remains 0 because `Inspect` can only retrieve it from the host agent. Extract the port from the instance's `ssh.config` file as a fallback; this file persists after the host agent dies and contains the port assigned at startup.

Fixes #4583

```console
❯ cd ~/.lima/alpine-iso
❯ l shell alpine-iso uname -a
Linux lima-alpine-iso 6.12.51-0-virt #1-Alpine SMP PREEMPT_DYNAMIC 2025-10-07 15:12:03 aarch64 Linux

❯ rm ha.sock
❯ l shell alpine-iso uname -a
ERRO[0000] Instance "alpine-iso" has configuration errors  error="failed to connect to \"/Users/jan/.lima/alpine-iso/ha.sock\": stat /Users/jan/.lima/alpine-iso/ha.sock: no such file or directory"
Linux lima-alpine-iso 6.12.51-0-virt #1-Alpine SMP PREEMPT_DYNAMIC 2025-10-07 15:12:03 aarch64 Linux

❯ echo "Hello world" >ssh.config
❯ l shell alpine-iso uname -a
ERRO[0000] Instance "alpine-iso" has configuration errors  error="failed to connect to \"/Users/jan/.lima/alpine-iso/ha.sock\": stat /Users/jan/.lima/alpine-iso/ha.sock: no such file or directory\nfailed to extract SSH local port from \"/Users/jan/.lima/alpine-iso/ssh.config\": port not found in \"/Users/jan/.lima/alpine-iso/ssh.config\""
Bad port '0'
```

### DIFF
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -87,6 +87,14 @@ func Inspect(ctx context.Context, instName string) (*limatype.Instance, error) {
 			}
 		}
 	}
+	if inst.SSHLocalPort == 0 {
+		sshConfigPath := filepath.Join(instDir, filenames.SSHConfig)
+		if port, err := sshPortFromConfig(sshConfigPath); err == nil {
+			inst.SSHLocalPort = port
+		} else if !errors.Is(err, os.ErrNotExist) {
+			inst.Errors = append(inst.Errors, fmt.Errorf("failed to extract SSH local port from %q: %w", sshConfigPath, err))
+		}
+	}
 
 	inst.CPUs = *y.CPUs
 	memory, err := units.RAMInBytes(*y.Memory)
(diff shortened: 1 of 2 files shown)


pkg/store/instance.go
```diff
@@ -226,6 +234,21 @@ func ReadPIDFile(path string) (int, error) {
 	return pid, nil
 }
 
+// sshPortFromConfig extracts the SSH port from an ssh config file.
+func sshPortFromConfig(configPath string) (int, error) {
```
AkihiroSuda, line 238: (This function is not robust to cover a config with multiple host entries, but we do not need to care about it for Lima-generated config files, at least with the current version)

AkihiroSuda (approved):
Thanks

pkg/store/instance.go
AkihiroSuda, line 238: nit: This could be easier for unit testing
```suggestion
func sshPortFromConfig(config []byte) (int, error) {
```

PR #4587, opened 2026-02-16
Introduce stable disk/iso filenames with symlink-based migration

jandubois:
Alternative to #4206.

Replace the legacy basedisk/diffdisk naming with self-documenting filenames: "image" (downloaded VM image), "disk" (VM disk), and "iso" (optional CDROM).

New instances download to "image" and create symlinks: non-ISO images get disk→image; ISO images get iso→image plus a real empty disk. QEMU no longer creates a qcow2 overlay backed by basedisk; it symlinks disk→image directly.

Existing instances are migrated by `MigrateDiskLayout`, which creates disk→diffdisk and (for ISO basedisks) iso→basedisk symlinks. The original files remain so older Lima versions can still use them.

Boot paths now check for the "iso" symlink via `osutil.FileExists` instead of calling `iso9660util.IsISO9660` at startup.

---

#### Using ISO image

```console
❯ ls -lh ~/.lima/alpine-iso/{image,*disk,iso}
ls: /Users/jan/.lima/alpine-iso/image: No such file or directory
-rw-r--r--@ 1 jan  staff    91M Oct 12 20:22 /Users/jan/.lima/alpine-iso/basedisk
-rw-r--r--@ 1 jan  staff   100G Feb 15 16:43 /Users/jan/.lima/alpine-iso/diffdisk
lrwxr-xr-x@ 1 jan  staff     8B Feb 15 16:42 /Users/jan/.lima/alpine-iso/disk -> diffdisk
lrwxr-xr-x@ 1 jan  staff     8B Feb 15 16:42 /Users/jan/.lima/alpine-iso/iso -> basedisk

❯ ls -lh ~/.lima/alpine-new/{image,*disk,iso}
-rw-r--r--@ 1 jan  staff   100G Feb 15 16:43 /Users/jan/.lima/alpine-new/disk
-rw-r--r--@ 1 jan  staff   103M Jan  7 15:01 /Users/jan/.lima/alpine-new/image
lrwxr-xr-x@ 1 jan  staff     5B Feb 15 16:43 /Users/jan/.lima/alpine-new/iso -> image
```

#### Regular non-ISO image

```console
❯ ls -lh ~/.lima/default/{image,*disk}
ls: /Users/jan/.lima/default/image: No such file or directory
-rw-r--r--@ 1 jan  staff   838M Nov 24 20:38 /Users/jan/.lima/default/basedisk
-rw-r--r--@ 1 jan  staff   100G Feb 15 16:51 /Users/jan/.lima/default/diffdisk
[...]

Issue #4135, opened 2025-10-04
A locator plugin should return the YAML contents, not HTTPS URL


AkihiroSuda:
> On the second thought, a locator plugin should return the YAML contents, not HTTPS URL, so as to support non-HTTP and non-local remotes

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/issues/3937#issuecomment-3367798067_

jandubois:
This is slightly problematic because then most locator plugins must implement network transport, or expect `curl` or `wget` to be available.

I would suggest that a plugin can return either a URL/filename, or the YAML content:

If the output matches `/^{/` or `/^\S+:\s/`, or includes a newline, it is treated as YAML.

Otherwise it is interpreted as a URL (or local filename).

This mechanism can be implemented later, as it is fully backwards compatible. So we can just ship the URL mechanism in 2.0, if we run out of time.

AkihiroSuda:
> This is slightly problematic because then most locator plugins must implement network transport, or expect curl or wget to be available.

Not necessary. A plugin can just return a content like:
```
base:
- https://...
```

jandubois:
> Not necessary. A plugin can just return a content like:

No, that does not work for provisioning scripts and probes:

```yaml
provision:
- mode: system
  file: github:ORG/REPO/script.sh
```

jandubois:
The other issue we have to resolve is how to determine the instance name when the template is provided as YAML? We would have to require that the user specifies `--name`, the same way we do it when we read the template from STDIN via `-`.

Maybe it is a stupid idea (I need to sleep), but we could print the name to STDERR when STDOUT is YAML and not a URL?

Maybe require STDERR to be in JSON in this case, so it is extensible if later we want to return even more metadata?

jandubois:
> If the output matches `/^{/` or `/^\S+:\s/`, or includes a newline, it is treated as YAML.
> 
> Otherwise it is interpreted as a URL (or local filename).

The same logic is used in #4134 to determine if a file contains YAML or a symlink (`LooksLikeSymlink` in `pkg/limatmpl/github.go`).

AkihiroSuda:
Stderr is for errors.
Why not just print structured data in stdout?

jandubois:
> Stderr is for errors.

That is just a convention. Even `limactl` is using it also for logging and not just for errors.

It is just a second file descriptor, so we can capture a separate output stream.

[...]

PR #4595, opened 2026-02-16
Support macOS guests

### LINKED ISSUE #3618: VM drivers for supporting macOS guests (`vz`, `host-user`)

AkihiroSuda:
Usage:
```
limactl create --video template:macos
limactl start macos
limactl shell macos
```

The password prompt is shown during creating an instance, so as to run `chown root:wheel ~/.lima/_mnt/0/Library/LaunchDaemons/...`, which is required for the `lima-macos-init` launch daemon to run.

The password for GUI login is randomly generated and stored in `/Users/${USER}.guest/password` in the VM.

Fix #3618

TODOs (in follow-up PRs):
- [x] Docs
- [x] Switch back the default shell to `zsh`
- [x] Graceful `limactl stop` (currently it has to be stopped from the guest) https://github.com/lima-vm/lima/issues/4610
- [ ] Propagate additional cloud-init data
  - [x] Hostname
  - [x] Timezone
  - [x] DNS
  - [ ] CA Certificates https://github.com/lima-vm/lima/issues/4611
- [ ] Support non-plain mode
  - [x] mounts  https://github.com/lima-vm/lima/issues/4612
  - [ ] port forwards https://github.com/lima-vm/lima/issues/4613
- [x] ~Let `limactl sudoers` generate the entry for chowning files
  (`chown root:wheel ~/.lima/_mnt/0/Library/LaunchDaemons/...`)~ https://github.com/lima-vm/lima/issues/4615
- [ ] Support headless mode (`--video=false`) https://github.com/lima-vm/lima/issues/4614
- [x] Support macOS in `hack/update-template.sh` https://github.com/lima-vm/lima/issues/4616

- - -
Depends on:
- [x] #4587
- [x] #4601


cmd/lima-guestagent/fake_cloud_init_darwin.go
```diff
@@ -0,0 +1,26 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	_ "embed"
```
jandubois: Why? There is no `//go:embed` directive in the file.

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -43,13 +51,20 @@ CODE=0
...
-			CODE=1
-		fi
-	done
+	boot="${LIMA_CIDATA_MNT}/boot"
+	if [ "${UNAME}" != "Linux" ]; then
+		# TODO: rename boot to boot.Linux
+		boot="${boot}.${UNAME}"
+	fi
+	if [ -e "${boot}" ]; then
+		for f in "${LIMA_CIDATA_MNT}"/boot/*; do
```
jandubois:
```suggestion
		for f in "${LIMA_CIDATA_MNT}/${boot}"/*; do
```

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -330,14 +335,80 @@ func validateConfig(_ context.Context, cfg *limatype.LimaYAML) error {
...
+	if l.Instance.Config.OS != nil && *l.Instance.Config.OS == limatype.DARWIN {
+		disk := filepath.Join(l.Instance.Dir, filenames.Disk)
+		if !osutil.FileExists(disk) {
+			if err := l.createDiskMacOSGuest(ctx); err != nil {
+				return err
+			}
+		}
+
+		// Wait for a few seconds to avoid "Failed to lock auxiliary storage" error
+		time.Sleep(3 * time.Second)
```
jandubois: Shouldn't this use some retry mechanism with a longer timeout? 3 seconds is not a lot if the host is under load.

pkg/imgutil/nativeimgutil/asifutil/asif_darwin.go
```diff
@@ -49,3 +68,75 @@ func ResizeASIF(path string, size int64) error {
...
+			continue
+		}
+		device := fields[0]
+		typ := strings.Join(fields[1:], " ")
+		if typ == "Apple_APFS_Container" {
+			seenContainers++
+			// FIXME: not robust?
+			if seenContainers == 2 {
+				return &AttachedDisk{
+					DataSlice: device + "s5",
```
jandubois: I know this already has a FIXME comment, but it seems pretty fragile: If macOS versions change the partition scheme, this silently picks the wrong slice.

I guess so far `s5` has always been correct, but maybe do something like `diskutil apfs list <container>` to identify volumes by role?

pkg/guestpatch/macos/macos_darwin.go
```diff
@@ -0,0 +1,170 @@
...
+	for _, file := range filesToTouch {
+		if err := osutil.Touch(file); err != nil {
+			return fmt.Errorf("failed to touch %q: %w", file, err)
+		}
+	}
+
+	const initSh = `#!/bin/bash
+set -eux
+#!/bin/sh
+set -eux
```
jandubois: Duplicated lines. Top hashbang requests `/bin/bash`.

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -330,14 +335,80 @@ func validateConfig(_ context.Context, cfg *limatype.LimaYAML) error {
 }
 
 func (l *LimaVzDriver) Create(_ context.Context) error {
-	_, err := getMachineIdentifier(l.Instance)
+	idenfierFile := filepath.Join(l.Instance.Dir, filenames.VzIdentifier)
```
jandubois:
```suggestion
	identifierFile := filepath.Join(l.Instance.Dir, filenames.VzIdentifier)
```

jandubois:
> On second thought, is the `nil` check necessary?

I've wondered before why we use pointers for so many of the fields in `LimaYAML`.

For some, especially booleans, we want to be able to distinguish if the value was specified as the null value (i.e. `false`, `0`, `""`), or not provided at all. But for many values the null value is not a valid choice, so e.g. the OS could just be `OS` instead of `*OS`.

Or am I missing something?

cmd/lima-guestagent/fake_cloud_init_darwin.go
AkihiroSuda: Fixed

pkg/cidata/cidata.TEMPLATE.d/boot.sh
AkihiroSuda: Fixed

pkg/driver/vz/vz_driver_darwin.go
AkihiroSuda: Fixed

pkg/imgutil/nativeimgutil/asifutil/asif_darwin.go
AkihiroSuda: Fixed to parse XML

pkg/guestpatch/macos/macos_darwin.go
AkihiroSuda: Fixed

pkg/driver/vz/vz_driver_darwin.go
AkihiroSuda: Fixed

AkihiroSuda:
> On second thought, is the nil check necessary? Because FillDefault guarantees that it isn't nil. We don't do nil checks for other fields in the driver code (or shouldn't have to).

Removed the nil check

> I've wondered before why we use pointers for so many of the fields in LimaYAML.

Probably we wanted to treat all the types in the consistent manner.
That may help simplifying `FillDefault` with reflection in future?

AkihiroSuda:
Split plist parser to:
- #4601

jandubois:
> That may help simplifying `FillDefault` with reflection in future?

I don't think `FillDefaults()` will be implemented using reflection, but will use template embedding instead:

* we need a way for custom URL schemes to return the data, not just a redirect (#4135)

* define an `internal:` scheme that can return the following:
  - `internal:override`: contents of `$LIMA_HOME/_config/override.yaml` or empty string
  - `internal:default`: contents of `$LIMA_HOME/_config/default.yaml` or empty string
  - `internal:builtin`: builtin defaults as a LimaYAML config
  - `internal:user`: the template provided by the user

* `limayaml.FillDefaults()` then just invokes `limatmpl.Embed()` on
  ```yaml
  base:
  - internal:override
  - internal:user
  - internal:default
  - internal:builtin
  ```

* URL scheme need to be registered dynamically because `internal` needs to be a closure in order to implement `internal:user`.

There are some other challenges that I currently no longer remember, but this would re-use the existing code for combining templates instead of having a second implementation.

Lohann:
For reference, I managed to skip the password and login via SSH at startup using a few tricks from here:
https://github.com/sickcodes/osx-optimizer

The only downside is that I had to modify the the `*.plist` files manually in the Volume before startup, once I couldn't login in the VM to use the `defaults` commands, I mounted the VM volume before startup, copied the contents, and compared against the content after startup and manual setup.

AkihiroSuda:
> For reference, I managed to skip the password and login via SSH at startup using a few tricks from here:

Thanks, but I don't think we want to skip it by default

pkg/osutil/mount_darwin.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Mount mounts the device.
+// Root privileges it not necessary.
```
jandubois:
```suggestion
// Root privileges is not necessary.
```

website/content/en/docs/usage/_index.md
```diff
@@ -60,8 +60,9 @@ To make the host mount writable, run `limactl start` with `--mount-writable`.
 To disable the mount, `limactl start` with `--mount-none` or `--plain`.
 
 The guest home directory exists independently on the following path:
-- `/home/${USER}.guest` (since Lima v2.1)
-- `/home/${USER}.linux` (prior to Lima v2.1)
+- `/Users/${USER}.guest` (on macOS hosts)
+- `/home/${USER}.guest`  (on other hosts, since Lima v2.1)
```
jandubois: Doesn't this depend on the guest OS instead?


```suggestion
- `/Users/${USER}.guest` (on macOS guests)
- `/home/${USER}.guest`  (on other guests, since Lima v2.1)
```

pkg/guestpatch/macos/macos_darwin.go
```diff
@@ -0,0 +1,191 @@
...
+	if err = patchMacOSDiskPrivilegedPhase(ctx, dataDevPath, mnt); err != nil {
+		return fmt.Errorf("failed to patch macOS disk (phase 2): %w", err)
+	}
+
+	return nil
+}
+
+func patchMacOSDiskUnprivilegedPhase(ctx context.Context, dataSliceDevice, mnt string) error {
+	// Enable "noowners" to allow non-root users to write to the mounted volume.
+	// The ownership is fixed up in patchMacOSDiskPhase2.
```
jandubois:
```suggestion
	// The ownership is fixed up in patchMacOSDiskPrivilegedPhase.
```

I find the names hard to distinguish; there are just 2 characters "Un" hidden in a very long name; it would be better to have a difference at the very front. "patchMacOS" is also kind of redundant given that the directory is `guestpatch/macos".

pkg/guestpatch/macos/macos_darwin.go
```diff
@@ -0,0 +1,191 @@
...
+		return fmt.Errorf("failed to acquire mount slot: %w", err)
+	}
+	defer func() {
+		if mntReleaseErr := mntRelease(); mntReleaseErr != nil {
+			logrus.WithError(mntReleaseErr).Warnf("failed to release mount slot %q", mnt)
+		}
+	}()
+
+	if err = patchMacOSDiskUnprivilegedPhase(ctx, dataDevPath, mnt); err != nil {
+		return fmt.Errorf("failed to patch macOS disk (phase 1): %w", err)
```
jandubois: You still use phase 1/2 in the error message, which doesn't match back to the function names anymore.

pkg/guestpatch/macos/macos_darwin.go
```diff
@@ -0,0 +1,191 @@
...
+		return err
+	}
+	defer func() {
+		if err := osutil.Umount(ctx, mnt); err != nil {
+			logrus.WithError(err).Warnf("failed to unmount %q", mnt)
+		}
+	}()
+	chownCmd := chownCommand(mnt)
+	cmd := exec.CommandContext(ctx, "sudo", chownCmd...)
+	logrus.Infof("Executing command (chowning the newly installed files to root:wheel): \n%v", cmd.Args)
```
jandubois: It is kind of surprising that you have to enter the host admin password to prepare the VM files, so maybe some explanation should be printed?

The first time I ran this I was wondering how to get the root password of the VM before it was running.

I have no idea how to avoid this, but it would be much better not to require the host root password for installation.

jandubois (approved):
Thanks, LGTM

Let's have further changes in separate smaller PRs.

jandubois:
I don't get clean shutdown behaviour:

```console
❯ l stop macos
INFO[0000] Sending SIGINT to hostagent process 46875
INFO[0000] [hostagent] 2026/02/23 12:27:39 tcpproxy: for incoming conn 127.0.0.1:63644, error dialing "192.168.5.15:22": connect tcp 192.168.5.15:22: no route to host
ERRO[0004] [hostagent] vz timeout while waiting for stop status
FATA[0190] did not receive an event with the "exiting" status
```

jandubois:
> > For reference, I managed to skip the password and login via SSH at startup using a few tricks from here:
> 
> Thanks, but I don't think we want to skip it by default

Why not? I think having to enter the host root password just to prepare a VM image is a major misfeature:

1. You need the admin password. Enterprise users may not have access.
2. You cannot script creation of macOS guests.

So installation via SSH instead of needing the launch daemon seems preferable to me.

jandubois:
> So installation via SSH instead of needing the launch daemon seems preferable to me.

I told Claude Code with Opus 4.6 to figure out a way to do it (enable SSH without root/admin), but so far it has not been successful.

One learning though was that it seems impossible to gracefully stop a macOS guest from the outside (you can't reboot via ACPI). You need to shutdown from inside the guest, using SSH, or the guestagent. Several other projects seem to have confirmed this.

pkg/osutil/mount_darwin.go
AkihiroSuda: fixed

website/content/en/docs/usage/_index.md
AkihiroSuda: fixed

pkg/guestpatch/macos/macos_darwin.go
AkihiroSuda: fixed

pkg/guestpatch/macos/macos_darwin.go
AkihiroSuda: fixed

pkg/guestpatch/macos/macos_darwin.go
AkihiroSuda: fixed to print an explanation

AkihiroSuda:
> I don't get clean shutdown behaviour:

Documented in `docs/usage/guests/macos.md`

AkihiroSuda:
> > > For reference, I managed to skip the password and login via SSH at startup using a few tricks from here:
> > 
> > 
> > Thanks, but I don't think we want to skip it by default
> 
> Why not? I think having to enter the host root password just to prepare a VM image is a major misfeature:


I assumed that @Lohann was talking about skipping the guest password on GUI login
https://github.com/sickcodes/osx-optimizer?tab=readme-ov-file#skip-the-gui-login-screen-at-your-own-risk

> 1. You need the admin password. Enterprise users may not have access.
> 2. You cannot script creation of macOS guests.
> 
> So installation via SSH instead of needing the launch daemon seems preferable to me.

You cannot SSH into the VM before the lima-macos-init launch daemon starts up

jandubois:
> enable SSH without root/admin

I've given up on this idea now. Even if it was possible, I think Apple would probably consider it a vulnerability and "fix it" in future releases.

I have a new plan though: create a special-purpose APFS patcher that can just modify UID and GID values of existing files. It needs to be able to traverse the btree structure to find the directory entries, but then it should be simple to overwrite the values in-place and recompute the checksum for the directory entry.

Since this is for a fresh installation, there should be no snapshots, and we would not be changing any file allocations, just modifying entries in-place. Only requirement would be to unmount the disk before patching.

AkihiroSuda:
>  I think having to enter the host root password just to prepare a VM image is a major misfeature:

A follow-up PR will update `limactl sudoers` to allow running `chown root:wheel ~/.lima/_mnt/0/...`.

Ideally we should have a pure user space utility to write a file into the disk image.
Similar to mtools for DOS floppy images 💾.

jandubois:
> A follow-up PR will update `limactl sudoers` to allow running `chown root:wheel ~/.lima/_mnt/0/...`.

Which still requires admin access to create the `sudoers` entry. Regular users are still out of luck.

> Ideally we should have a pure user space utility to write a file into the disk image. Similar to mtools for DOS floppy images 💾.

Yes, that would be ideal, but no such tool seem to exist. Most libraries provide only read-only access to APFS. The tool I described in my previous comment seems more achievable in the short term.

Since all the available open source implementations are licensed under the GPL I've used AI to create a spec that just contains the data structures required to implement this (most info is available directly from Apple): https://gist.github.com/jandubois/e80c17e3322b53bf0a89a886a6ea08de

I'll now let a separate AI attempt a "clean-room" implementation from this spec.

jandubois (approved):
Thanks, LGTM

jandubois:
> I'll now let a separate AI attempt a "clean-room" implementation from this spec.

This took significantly longer than expected, but works for me on macOS 26 aarch64: https://github.com/lima-vm/lima/commit/e35e9a369cb95221151ea723a9e282bac2dece59

It is a single commit on top of this PR branch (or the previous version of it), and it eliminates the need to use `sudo`.

This is AI generated code (with massive steering), and still needs proper review, as well as testing on x86_64, and maybe older versions of macOS. So I can't create a PR yet (and am out of time for now), but please give it a try to see if it works for you too!

AkihiroSuda:
> This took significantly longer than expected, but works for me on macOS 26 aarch64: https://github.com/lima-vm/lima/commit/e35e9a369cb95221151ea723a9e282bac2dece59

Thanks, works for me too.
Left some comments there.

PR #4600, opened 2026-02-20
Make DEST optional in `limactl template copy`

jandubois:
Default to stdout when only a source template is specified, matching the common use case of inspecting a template.

### DIFF
--- a/cmd/limactl/template.go
+++ b/cmd/limactl/template.go
@@ -61,11 +61,11 @@ var templateCopyExample = `  Template locators are local files, file://, https:/
 
 func newTemplateCopyCommand() *cobra.Command {
 	templateCopyCommand := &cobra.Command{
-		Use:     "copy [OPTIONS] TEMPLATE DEST",
+		Use:     "copy [OPTIONS] TEMPLATE [DEST]",
 		Short:   "Copy template",
-		Long:    "Copy a template via locator to a local file",
+		Long:    "Copy a template via locator to a local file or stdout (default)",
 		Example: templateCopyExample,
-		Args:    WrapArgsError(cobra.ExactArgs(2)),
+		Args:    WrapArgsError(cobra.RangeArgs(1, 2)),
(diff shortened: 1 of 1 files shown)


cmd/limactl/template.go
```diff
@@ -51,8 +51,8 @@ func newValidateCommand() *cobra.Command {
 
 var templateCopyExample = `  Template locators are local files, file://, https://, or template: URLs
 
-  # Copy default template to STDOUT
-  limactl template copy template:default -
+  # Copy default template to STDOUT (DEST defaults to -)
+  limactl template copy template:default
```
AkihiroSuda: I feel the previous form is easier to understand

jandubois: So you just want this example to stay as-is, but the rest of the change is fine?

Makes the default a bit more subtle, but it is still mentioned in the long help.

AkihiroSuda: I'm inclined to keep everything as-is, but no strong opinion

jandubois: Reverted the change to `templateCopyExample`.

jandubois:
Ping!

This is a trivial (+7/-4) change to provide a default value, that can be reviewed in 2 minutes.

cmd/limactl/template.go
AkihiroSuda: I'm still inclined to feel `copy SRC` without `DST` looks weird. `render SRC` may look more natural.

No strong opinion though.

AkihiroSuda (approved):

cmd/limactl/template.go
jandubois: I have reverted the docs example. This is now literally just the default value for destination. Essentially `copy` without destination becomes `cat`.

I find this reasonably clear: if you don't specify where to copy to, you copy to STDOUT.

If you don't want to use it, you can still specify `-` as the second parameter.

PR #4601, opened 2026-02-20
Add pkg/plist for parsing XML-formatted plist documents

AkihiroSuda:
NOTE: used GPT-5 mini for some portions

- - -

Didn't use third-party libraries (e.g. https://pkg.go.dev/howett.net/plist) as they seem overengineered

### DIFF
--- a/pkg/osutil/machineid.go
+++ b/pkg/osutil/machineid.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/xml"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -16,6 +17,8 @@ import (
 	"sync"
 
 	"github.com/sirupsen/logrus"
+
+	"github.com/lima-vm/lima/v2/pkg/plist"
 )
 
 var MachineID = sync.OnceValue(func() string {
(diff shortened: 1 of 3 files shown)


pkg/plist/plist.go
```diff
@@ -0,0 +1,243 @@
...
+			p.Value = v
+		case xml.EndElement:
+			if t.Name.Local == start.Name.Local {
+				return nil
+			}
+		}
+	}
+}
+
+func (v *Value) UnmarshalXML(decoder *xml.Decoder, start xml.StartElement) error {
```
jandubois: Use consistent naming, either `dec` or `decoder`, but use the same everywhere.

```suggestion
func (v *Value) UnmarshalXML(dec *xml.Decoder, start xml.StartElement) error {
```

pkg/plist/plist.go
```diff
@@ -0,0 +1,243 @@
...
+			if err != nil {
+				if errors.Is(err, io.EOF) {
+					return nil
+				}
+				return err
+			}
+			if ee, ok := tok.(xml.EndElement); ok && ee.Name.Local == start.Name.Local {
+				return nil
+			}
+		}
```
jandubois: Isn't this the same as:

```suggestion
		return decoder.Skip()
```

Same thing for the next case below.

pkg/plist/plist_test.go
```diff
@@ -0,0 +1,125 @@
...
+		for plutilExpectedKey, plutilExpectedValue := range plutilExpected {
+			plutilCmd := exec.CommandContext(ctx, "plutil", "-extract", plutilExpectedKey, "raw", "-n", "-o", "-", testFilePath)
+			var stderr bytes.Buffer
+			plutilCmd.Stderr = &stderr
+			plutilOutput, err := plutilCmd.Output()
+			assert.NilError(t, err, fmt.Sprintf("failed to run %v (stderr=%q)", plutilCmd.Args, stderr.String()))
+			plutilOutputStr := string(plutilOutput)
+			assert.Equal(t, plutilExpectedValue, plutilOutputStr, "plutil output mismatch for key %q", plutilExpectedKey)
+		}
+		assert.NilError(t, err)
```
jandubois: Duplicate? All assignments to `err` already have an assert.

pkg/osutil/machineid.go
```diff
@@ -56,31 +59,28 @@ func machineID(ctx context.Context) (string, error) {
...
-		elem            string
-		elemKeyCharData string
-	)
-	for {
-		tok, err := d.Token()
-		if err != nil {
-			return "", err
+	var p plist.Plist
+	dec := xml.NewDecoder(r)
+	if err := dec.Decode(&p); err != nil {
```
jandubois, line 64: Might be nice to have a convenience function, so caller doesn't need to import `encoding/xml`:


```suggestion
	if p, err := plist.Decode(&r); err != nil {
```

jandubois (approved):
Thanks, LGTM

Only minor stylistic feedback

pkg/plist/plist.go
AkihiroSuda: fixed

pkg/plist/plist.go
AkihiroSuda: fixed

pkg/plist/plist_test.go
AkihiroSuda: fixed

pkg/osutil/machineid.go
AkihiroSuda: Not sure. We don't do that for `encoding/json`.

PR #4603, opened 2026-02-21
Switch BATS submodules to upstream

### LINKED ISSUE #4584: Bump BATS submodule
The errexit PR has been merged, so we can update our forks to latest upstream commits and point the submodules to a commit on `master` instead of a PR.

We can also consider dropping our forks and switching the submodules to the upstream repos because it doesn't look like we are carrying any more local changes. Will require developers to run a `git submodule sync` and is a bit of a hassle when you switch between branches (as long as there are old branches that point to the old submodules). So not sure if this is worth it or not.

jandubois:
Point all submodules to the latest commit as of today; there is no release yet including the `errexit` support and releases are somewhat infrequent (2 releases in 2025).

The bats-support submodule has no new commits.

Users should switch local submodules with

```
git submodule deinit --all
git submodule sync
git submodule update --init
```

Fixes #4584

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,16 +1,16 @@
 [submodule "hack/bats/lib/bats-core"]
 	path = hack/bats/lib/bats-core
-	url = https://github.com/lima-vm/bats-core.git
+	url = https://github.com/bats-core/bats-core.git
 	branch = master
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #4604, opened 2026-02-21
Extract LIMA_BATS_REUSE_INSTANCE logic into helper functions

### LINKED ISSUE #4599: Move LIMA_BATS_REUSE_INSTANCE code into a reusable helper function
The logic has been duplicated into several test files already; there should be a single implementation only.

jandubois:
Move the duplicated instance setup/teardown code from five test files into `ensure_lima_instance` and `cleanup_lima_instance` in limactl.bash. The instance configuration is determined by $NAME, preventing tests from sharing a name but using incompatible settings.

Give shell-sync.bats its own instance name (`bats-nomount`) so it no longer contaminates the shared `bats` instance by removing mounts.

Closes #4599

### DIFF
--- a/hack/bats/helpers/limactl.bash
+++ b/hack/bats/helpers/limactl.bash
@@ -5,7 +5,7 @@
 # This function intentionally doesn't use create/editflags, but modifies the template with yq instead.
 create_dummy_instance() {
     local name=$1
-    local expr=$2
+    local expr=${2:-}
 
     # Template does not validate without an image, and the image must point to a file that exists (for clonefile).
     local template="{images: [location: /etc/profile]}"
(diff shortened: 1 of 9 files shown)


hack/bats/helpers/limactl.bash
```diff
@@ -14,3 +14,33 @@ create_dummy_instance() {
     fi
     limactl create --name "$name" - <<<"$template"
 }
+
+# Ensure a Lima instance named $NAME is running. When LIMA_BATS_REUSE_INSTANCE
+# is set, reuse an existing running instance. Otherwise delete and recreate it.
+# The instance configuration is determined by $NAME; add a case below for new names.
+# Close file handles 3 and 4 so the host agent doesn't block BATS from exiting.
+ensure_lima_instance() {
```
AkihiroSuda: This should take `local name=$1` as an argument?
Same as `create_dummy_instance()` above.

AkihiroSuda:
```suggestion
ensure_instance() {
```

Or `instance_ensure()`, or maybe `instance::ensure()`

jandubois: fixed

jandubois: fixed

jandubois:
I've renamed `$NAME` to `$INSTANCE` and moved the calls to `ensure_instance` and `delete_instance` inside `setup_file` and `teardown_file` (when `$INSTANCE` is set).

So no local scaffolding is necessary anymore beyond assigning the desired instance type to `$INSTANCE`.

I also folded `create_dummy_instance` into the same mechanism.

AkihiroSuda (dismissed):
Thanks, but needs rebase

AkihiroSuda (approved):
Thanks

PR #4609, opened 2026-02-24
qemu: support vmOpts.qemu.extraArgs

girtsf:
Add support for passing additional QEMU CLI arguments via template YAML.

- add ExtraArgs to limatype.QEMUOpts
- append extra args to the generated QEMU command line
- document usage in templates/default.yaml

### DIFF
--- a/pkg/driver/qemu/qemu.go
+++ b/pkg/driver/qemu/qemu.go
@@ -481,6 +481,14 @@ func resolveCPUType(y *limatype.LimaYAML) string {
 	return cpuType[*y.Arch]
 }
 
+func qemuExtraArgs(y *limatype.LimaYAML) ([]string, error) {
+	var qemuOpts limatype.QEMUOpts
+	if err := limayaml.Convert(y.VMOpts[limatype.QEMU], &qemuOpts, "vmOpts.qemu"); err != nil {
+		return nil, fmt.Errorf("failed to convert `vmOpts.qemu`: %w", err)
+	}
+	return qemuOpts.ExtraArgs, nil
+}
+
 func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err error) {
 	y := cfg.LimaYAML
 	exe, args, err = Exe(*y.Arch)
@@ -1032,6 +1040,11 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	// QEMU process
 	args = append(args, "-name", "lima-"+cfg.Name)
 	args = append(args, "-pidfile", filepath.Join(cfg.InstanceDir, filenames.PIDFile(*y.VMType)))
+	extraArgs, err := qemuExtraArgs(y)
+	if err != nil {
+		return "", nil, err
+	}
+	args = append(args, extraArgs...)
 
 	return exe, args, nil
 }
--- a/templates/default.yaml
+++ b/templates/default.yaml
@@ -396,6 +396,12 @@ vmOpts:
     #   armv7l:  "max" # (or "host" when running on armv7l host)
     #   riscv64: "max" # (or "host" when running on riscv64 host)
     #   x86_64:  "max" # (or "host" when running on x86_64 host; additional options are appended on Intel Mac)
+
+    # Extra arguments to pass to QEMU.
+    # Each list entry is one CLI token (no shell parsing).
+    # 🟢 Builtin default: []
+    # extraArgs:
+    # - "-usb"
   vz:
     # Specify the disk image format: "raw" or "asif".
     # Currently only applies to the primary disk image.
--- a/pkg/driver/qemu/qemu_test.go
+++ b/pkg/driver/qemu/qemu_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/lima-vm/lima/v2/pkg/limatype"
 	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
+	"github.com/lima-vm/lima/v2/pkg/limayaml"
 )
 
 func TestArgValue(t *testing.T) {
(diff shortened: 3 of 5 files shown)


templates/default.yaml
```diff
@@ -372,6 +372,12 @@ vmOpts:
     #   armv7l:  "max" # (or "host" when running on armv7l host)
     #   riscv64: "max" # (or "host" when running on riscv64 host)
     #   x86_64:  "max" # (or "host" when running on x86_64 host; additional options are appended on Intel Mac)
+
+    # Extra arguments to pass to QEMU.
+    # Each list entry is one CLI token (no shell parsing).
+    # 🟢 Builtin default: []
+    # extraArgs:
+    # - "-usb"
```
AkihiroSuda, line 404: Any reason not to use existing env vars? (`QEMU_SYSTEM_AARCH64`, `QEMU_SYSTEM_X86_64`, etc.)

girtsf: Mostly because I missed the fact that it gets parsed for shell arguments. :)

Though looking at https://github.com/lima-vm/lima/blob/24317e90e27093ee35df8d9df6f79b7f09188cf8/pkg/driver/qemu/qemu.go#L1101, looks like it would log a warning when args are present. And it would be nicer to have it in one place in template file.

Feel free to close the PR if you don't think it would be helpful to others.

jandubois: > Any reason not to use existing env vars? (`QEMU_SYSTEM_AARCH64`, `QEMU_SYSTEM_X86_64`, etc.)

I think we used to say that this is just a debugging feature, but it seems like the fact that you can add additional arguments is not documented at all.

If this is supposed to be a regular (driver) feature, then adding it to the template makes sense.

virtualdxs: FWIW, I have a use case that this would be incredibly helpful for - I need to add a pflash device for OVMF vars, so that the variables don't get written to the esp of the guest. Right now, the only workaround I'm seeing is to write a wrapper script that I call via QEMU_SYSTEM_X86_64 and QEMU_SYSTEM_AARCH64, but that feels really hacky. Being able to append args via the YAML would eliminate the need for a wrapper script entirely.

pkg/driver/qemu/qemu.go
```diff
@@ -479,6 +479,14 @@ func resolveCPUType(y *limatype.LimaYAML) string {
 	return cpuType[*y.Arch]
 }
 
+func qemuExtraArgs(y *limatype.LimaYAML) ([]string, error) {
+	var qemuOpts limatype.QEMUOpts
+	if err := limayaml.Convert(y.VMOpts[limatype.QEMU], &qemuOpts, "vmOpts.qemu"); err != nil {
+		return nil, fmt.Errorf("failed to convert `vmOpts.qemu.extraArgs`: %w", err)
```
jandubois: Since `Convert()` converts all opts, not just `extraArgs`, the error message might be misleading.

```suggestion
	if err := limayaml.Convert(y.VMOpts[limatype.QEMU], &qemuOpts, "vmOpts.qemu"); err != nil {
		return nil, fmt.Errorf("failed to convert `vmOpts.qemu`: %w", err)
```

jandubois (commented):
I noticed 2 things reviewing this PR:

1. (this is maybe outside the scope of of the PR) This adds yet another call to `limayaml.Convert()`. We should call it only once, and then use the returned structure instead of calling multiple times, potentially duplicating warning log entries. There is also some inconsistency if `Convert()` returning an error show result in a log warning, or error propagation.

2. (this requires some discussion): The `extraArgs` are appended at the end of the command line. They are not using the `appendArgsIfNoConflict()` mechanism that is used for other options. Maybe QEMU will always replace earlier values with later values. But will it generate warnings/errors? And we claim that e.g. QEMU_SYSTEM_AARCH64 takes precedence over other options. Should it take precedence over `extraArgs` as well. Right now the behaviour is a bit muddied, and also undocumented.

Please don't start changing the code for (2) until there has been a discussion and consensus!

AkihiroSuda:
> QEMU_SYSTEM_AARCH64 takes precedence over other options. Should it take precedence over extraArgs as well. 

I expect the args to be appended in the following order (w/ `appendIfNoConflict`):
- `QEMU_SYSTEM_${ARCH}` env vars
- Internally generated flags
- `extraArgs`

virtualdxs:
Re: `appendArgsIfNoConflict()`: How should it handle the disallowed args (e.g. -drive)? I'd suggest special-casing them and skipping the check rather than disallowing them, to allow for things like `-drive if=pflash,format=raw,file=/path/to/edk2-i386-vars.fd` to be injected via this mechanism.

AkihiroSuda:
> Re: `appendArgsIfNoConflict()`: How should it handle the disallowed args (e.g. -drive)? I'd suggest special-casing them and skipping the check rather than disallowing them, to allow for things like `-drive if=pflash,format=raw,file=/path/to/edk2-i386-vars.fd` to be injected via this mechanism.

For UEFI vars, I'd rather suggest supporting them as a proper feature without needing setting custom args.

AkihiroSuda:
Looks like we already have the support for efi vars, but not always enabled yet
https://github.com/lima-vm/lima/blob/20dd365a5635c339cc26e641a3c59e24e943fc89/pkg/driver/qemu/qemu.go#L618

https://github.com/lima-vm/lima/commit/e3fd24160b89661b1a4dff6e37afa91ae466f509

virtualdxs:
> For UEFI vars, I'd rather suggest supporting them as a proper feature without needing setting custom args.

That would be nice, yes, but in the meantime it would be better to have any way to configure it. In addition, it would be nice to support unknown/unexpected use cases by minimizing restrictions.

AkihiroSuda:
> > For UEFI vars, I'd rather suggest supporting them as a proper feature without needing setting custom args.
> 
> That would be nice, yes, but in the meantime it would be better to have any way to configure it. In addition, it would be nice to support unknown/unexpected use cases by minimizing restrictions.

Workaround: set `QEMU_SYSTEM_${ARCH}` to a wrapper script that executes qemu with `-drive`

virtualdxs:
Yes, that's the workaround I'm implementing now; it's just very hacky when I'm doing automatically as part of a tool that wraps Lima. I'm struggling to understand, what's the downside of allowing arbitrary -drive arguments? This would allow for more use cases than just EFI vars.

girtsf:
> 1. (this is maybe outside the scope of of the PR) This adds yet another call to `limayaml.Convert()`. We should call it only once, and then use the returned structure instead of calling multiple times, potentially duplicating warning log entries. There is also some inconsistency if `Convert()` returning an error show result in a log warning, or error propagation.

Maybe we do the parsing in `Start` and store the parsed result in `Config`? https://github.com/lima-vm/lima/blob/20dd365a5635c339cc26e641a3c59e24e943fc89/pkg/driver/qemu/qemu_driver.go#L254

> 2. (this requires some discussion): The `extraArgs` are appended at the end of the command line. They are not using the `appendArgsIfNoConflict()` mechanism that is used for other options. Maybe QEMU will always replace earlier values with later values. But will it generate warnings/errors? And we claim that e.g. QEMU_SYSTEM_AARCH64 takes precedence over other options. Should it take precedence over `extraArgs` as well. Right now the behaviour is a bit muddied, and also undocumented.

In my mind `extraArgs` are an advanced usage case and the user probably understands what is ending up on the qemu command line. In this case, the `IfNoConflict` part might get in the way.

Would it be useful to be more explicit in the naming that args are *appended*? `appendExtraArgs`?

PR #4623, opened 2026-02-24
cidata: rename boot to boot.Linux

AkihiroSuda:
Boot scripts are expected to be under `boot.<OS>` directory.

`Driver.BootScripts()` is now expected to return a map with keys `boot.<OS>/<SCRIPT>` instead of `<SCRIPT>`.

The old form `<SCRIPT>` is still supported for backward compatibility.


pkg/driver/driver.go
```diff
@@ -46,6 +46,8 @@ type Lifecycle interface {
 
 	InspectStatus(_ context.Context, inst *limatype.Instance) string
 
+	// BootScripts returns the content of boot scripts to be injected into the vm.
+	// The key must be "boot.<OS>/<SCRIPT>" or "<SCRIPT"> (deprecated alias for "boot.Linux/<SCRIPT>").
```
unsuman:
```suggestion
	// The key must be "boot.<OS>/<SCRIPT>" or "<SCRIPT>" (deprecated alias for "boot.Linux/<SCRIPT>").
```

pkg/cidata/cidata.go
```diff
@@ -391,8 +391,18 @@ func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name strin
 	}
 
 	for filename, content := range driverScripts {
+		layoutPath := path.Join("boot.Linux", filename)
+		if strings.Contains(filename, "/") {
+			// When the filename contains a slash, it must be in the format of "boot.<OS>/<SCRIPT>"
+			if !strings.HasPrefix(filename, "boot.") || strings.Count(filename, "/") != 1 {
+				return fmt.Errorf("invalid boot script filename %q", filename)
```
unsuman:
```suggestion
				return fmt.Errorf("invalid boot script filename %q: must be in format 'boot.<OS>/<SCRIPT>'", filename)
```

AkihiroSuda: fixed

pkg/driver/driver.go
AkihiroSuda: fixed

PR #4627, opened 2026-02-24
Support FreeBSD guests

### LINKED ISSUE #1508: Support for FreeBSD guests ?

AkihiroSuda:
Fix #1508

```
limactl start --mount-none template:freebsd
```

9p mounts seem to need `template:experimental/freebsd-16`.
FreeBSD 15 raises EINVAL on mounting p9fs.

<details><summary>Outdated information</summary>

~FIXME:~
- ~Only plain mode works. No virtiofs mounts, etc.~
- ~Only the first boot works (runcmd in user-data is executed only in the first boot)~
- ~Does not work on non-macOS hosts yet as pkg/iso9660util.WithJoliet is unimplemented~

</details>

NOTE: used Claude for `hack/update-template-freebsd.sh`


fwilhe:
Love it, this is the easiest way to spin up a working FreeBSD vm I've seen so far. Very convenient. Would be very cool if this worked on linux hosts, too.

AkihiroSuda:
> Would be very cool if this worked on linux hosts, too.

Works. Tested in the CI.

fwilhe:
Nice, last time I checked this was an open todo. Thank you, will try it soon.

hack/bats/extras/freebsd.bats
```diff
@@ -0,0 +1,24 @@
...
+load "../helpers/load"
+
+local_setup_file() {
+    limactl start --tty=false template:freebsd
+}
+
+local_teardown_file() {
+    # <DEBUG>
+    set -x
+    tail -n 100 "${LIMA_HOME}/freebsd/*.log"
```
jandubois: No wildcard expansion in quoted strings:

```suggestion
    tail -n 100 "${LIMA_HOME}"/freebsd/*.log
```

templates/freebsd-15.yaml
```diff
@@ -0,0 +1,8 @@
+minimumLimaVersion: 2.1.0
+
+base:
+- template:_images/freebsd-15
+# - template:_default/mounts
+
+# Mounts are not functional in freebsd-15.
+# Seems to working in experimental/freebsd-16.
```
jandubois:
```suggestion
# Seems to be working in experimental/freebsd-16.
```

pkg/hostagent/hostagent.go
```diff
@@ -605,7 +605,8 @@ sudo chown -R "${USER}" /run/host-services`
 	staticPortForwards := a.separateStaticPortForwards()
 	a.addStaticPortForwardsFromList(ctx, staticPortForwards)
 
-	if !*a.instConfig.Plain {
+	hasGuestAgentDaemon := !*a.instConfig.Plain && *a.instConfig.OS == limatype.LINUX
```
jandubois, line 608: What about macOS?

pkg/cidata/cidata.TEMPLATE.d/user-data
```diff
@@ -38,9 +49,18 @@ users:
 {{- if ne .OS "Darwin" }}
 {{/* On macOS, the password is not locked so as to allow GUI login */}}
     sudo: ALL=(ALL) NOPASSWD:ALL
```
jandubois, line 53: Does this work on FreeBSD, or does it have a separate setting for `doas`?

Also do we need to modify any boot scripts to use `doas` instead of `sudo`?

templates/_images/freebsd-15.yaml
```diff
@@ -0,0 +1,11 @@
+# The FreeBSD-Archive URLs are used for permanence.
+images:
+- location: "http://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/VM-IMAGES/15.0-RELEASE/amd64/Latest/FreeBSD-15.0-RELEASE-amd64-BASIC-CLOUDINIT-zfs.raw.xz"
```
jandubois, line 3: Doesn't really matter because we verify the digest, but does the archive not support https?

pkg/cidata/cidata.TEMPLATE.d/boot.essential.FreeBSD/00-freebsd-user-group.sh
```diff
@@ -0,0 +1,16 @@
...
+# This script ensures that the user is created with the expected UID.
+# This script is needed because nuageinit does not create the user with the expected UID.
+
+set -eu
+
+[ "$(stat -f %u "${LIMA_CIDATA_HOME}")" = "${LIMA_CIDATA_UID}" ] && exit 0
+
+pw usermod -n "${LIMA_CIDATA_USER}" -u "${LIMA_CIDATA_UID}"
+# the GID is assumed to match the UID
+pw groupmod -n "${LIMA_CIDATA_USER}" -g "${LIMA_CIDATA_UID}"
```
jandubois: Renaming the group is not enough, you also need to set the GID of the user explicitly (otherwise the gid is still `1001`):

```suggestion
# the GID is assumed to match the UID
pw groupmod -n "${LIMA_CIDATA_USER}" -g "${LIMA_CIDATA_UID}"
pw usermod -n "${LIMA_CIDATA_USER}" -u "${LIMA_CIDATA_UID}" -g "${LIMA_CIDATA_UID}"
```

templates/_images/freebsd-15.yaml
AkihiroSuda: In 2026 they still don't support HTTPS?
```console
$ curl https://ftp-archive.freebsd.org/pub/FreeBSD-Archive/old-releases/VM-IMAGES/15.0-RELEASE/aarch64/Latest/FreeBSD-15.0-RELEASE-arm64-aarch64-BASIC-CLOUDINIT-zfs.raw.xz
curl: (60) SSL: no alternative certificate subject name matches target host name 'ftp-archive.freebsd.org'
More details here: https://curl.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

pkg/hostagent/hostagent.go
AkihiroSuda: No guest agent daemon for macOS either, currently

hack/bats/extras/freebsd.bats
AkihiroSuda: fixed

templates/freebsd-15.yaml
AkihiroSuda: fixed

pkg/cidata/cidata.TEMPLATE.d/boot.essential.FreeBSD/00-freebsd-user-group.sh
AkihiroSuda: Thanks, but maybe we do not need to care about GID.
So far we didn't care about it for Linux
```console
$ limactl shell ubuntu-25.10 id
uid=501(suda) gid=1000(suda) groups=1000(suda)
```

AkihiroSuda: Updated the code not to care about the GID

pkg/cidata/cidata.TEMPLATE.d/user-data
AkihiroSuda: Both sudo and doas work.
Updated the PR to consistently use `sudo` to avoid diverging from the Linux boot scripts.

jandubois (approved):
Thanks, LGTM

Issue #4626, opened 2026-02-24
`limactl create <HTTPS URL to an image>` ignores `--name`


AkihiroSuda:
```
limactl create --name freebsd https://download.freebsd.org/releases/VM-IMAGES/15.0-RELEASE/aarch64/Latest/FreeBSD-15.0-RELEASE-arm64-aarch64-BASIC-CLOUDINIT-zfs.raw.xz
```

This command creates an instance named `freebsd-15.0-release-arm64-basic-zfs`, ignoring `--name freebsd`.

Tested with Lima v2.1.0-beta.0.

Harry-kp:
I'd like to work on this. The bug is in `imageTemplate()` in `pkg/limatmpl/locator.go` — it unconditionally overwrites `tmpl.Name` from the image URL, ignoring the `--name` flag. Every other branch in `Read()` checks `if tmpl.Name == ""` first. Will send a PR shortly.

PR #4628, opened 2026-02-24
fix: `limactl create <image URL>` honors `--name` flag

Harry-kp:
## Summary

- **Fixes** #4626
- `imageTemplate()` in `pkg/limatmpl/locator.go` unconditionally overwrote `tmpl.Name` with the name derived from the image URL, ignoring the `--name` CLI flag. Every other code path in `Read()` already guards the assignment with `if tmpl.Name == ""` — this PR applies the same guard to `imageTemplate()`.

## Reproduction

```
limactl create --name freebsd https://download.freebsd.org/releases/VM-IMAGES/15.0-RELEASE/aarch64/Latest/FreeBSD-15.0-RELEASE-arm64-aarch64-BASIC-CLOUDINIT-zfs.raw.xz
```

**Before**: instance created as `freebsd-15.0-release-arm64-basic-zfs`, ignoring `--name freebsd`
**After**: instance created as `freebsd`

## Changes

- `pkg/limatmpl/locator.go`: Guard `tmpl.Name` assignment in `imageTemplate()` with `if tmpl.Name == ""`
- `pkg/limatmpl/locator_test.go`: Add `TestReadImageURLRespectsName` covering both the `--name` override and the default name derivation paths

## Test plan

- [x] `go test ./pkg/limatmpl/ -run TestReadImageURLRespectsName` — passes
- [x] All existing `TestInstNameFromImageURL` tests — still pass

### DIFF
--- a/pkg/limatmpl/locator.go
+++ b/pkg/limatmpl/locator.go
@@ -200,7 +200,9 @@ images:
   arch: %q
 `
 	tmpl.Bytes = fmt.Appendf(nil, template, imageArch, locator, imageArch)
-	tmpl.Name = InstNameFromImageURL(locator, imageArch)
+	if tmpl.Name == "" {
+		tmpl.Name = InstNameFromImageURL(locator, imageArch)
+	}
 	return true
 }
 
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #4634, opened 2026-02-26
Eliminate sudo from macOS guest disk patching

### LINKED ISSUE #4615: macOS guest: remove dependency on host sudo
> > I'll now let a separate AI attempt a "clean-room" implementation from this spec.
> 
> This took significantly longer than expected, but works for me on macOS 26 aarch64: https://github.com/lima-vm/lima/commit/e35e9a369cb95221151ea723a9e282bac2dece59
> 
> It is a single commit on top of this PR branch (or the previous version of it), and it eliminates the need to use `sudo`.
> 
> This is AI generated code (with massive steering), and still needs proper review, as well as testing on x86_64, and maybe older versions of macOS. So I can't create a PR yet (and am out of time for now), but please give it a try to see if it works for you too!
> 
> _Originally posted by @jandubois in https://github.com/lima-vm/lima/issues/4595#issuecomment-3949706144_
>

jandubois:
Replace `sudo chown root:wheel` with direct APFS block modification. The new `pkg/apfs` package patches file UID/GID on the raw disk image by walking the APFS B-tree, modifying inode records, and recomputing Fletcher-64 checksums. No mount, no root privileges needed.

The disk patching now runs in two phases: mount the image to write guest files (noowners), then attach without mounting to fix ownership via the APFS container's character device node.

This PR has been created using Opus 4.6.

Fixes #4615

### DIFF
--- a/pkg/apfs/chown_darwin_test.go
+++ b/pkg/apfs/chown_darwin_test.go
@@ -1,0 +1,79 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package apfs
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"syscall"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+// TestChownIntegration creates a real APFS disk image with hdiutil,
+// writes a test file, runs Chown on the raw image, and verifies
+// ownership via os.Stat after remounting with ownership enabled.
+// No root required.
+func TestChownIntegration(t *testing.T) {
+	tmpDir := t.TempDir()
+	imgPath := filepath.Join(tmpDir, "test.dmg")
+
+	// Create a 64MB APFS disk image (GPT + APFS container).
+	ctx := t.Context()
+	cmd := exec.CommandContext(ctx, "hdiutil", "create",
+		"-size", "64m",
+		"-fs", "APFS",
+		"-volname", "TestVol",
+		imgPath,
+	)
+	out, err := cmd.CombinedOutput()
+	assert.NilError(t, err, "hdiutil create failed: %s", out)
+
+	// Attach, write a test file, and detach. Default mount uses
+	// noowners, so the on-disk UID will be 99 (nobody).
+	mntDir := filepath.Join(tmpDir, "mnt")
+	cmd = exec.CommandContext(ctx, "hdiutil", "attach", imgPath, "-mountpoint", mntDir)
+	out, err = cmd.CombinedOutput()
+	assert.NilError(t, err, "hdiutil attach failed: %s", out)
+
+	assert.NilError(t, os.WriteFile(filepath.Join(mntDir, "testfile.txt"), []byte("hello"), 0o644))
+
+	cmd = exec.CommandContext(ctx, "hdiutil", "detach", mntDir)
+	out, err = cmd.CombinedOutput()
+	assert.NilError(t, err, "hdiutil detach failed: %s", out)
+
+	// Change ownership to root:wheel via raw APFS modification.
+	// hdiutil creates a single volume with role=0 (APFS_VOL_ROLE_NONE).
+	assert.NilError(t, Chown(imgPath, 0, 0, 0, "testfile.txt"))
+
+	// Verify filesystem integrity after raw block modification.
+	cmd = exec.CommandContext(ctx, "hdiutil", "attach", imgPath, "-nomount")
+	out, err = cmd.CombinedOutput()
+	assert.NilError(t, err, "hdiutil attach -nomount failed: %s", out)
+	dev := strings.Fields(string(out))[0]
+	cmd = exec.CommandContext(ctx, "fsck_apfs", "-n", dev)
+	out, err = cmd.CombinedOutput()
+	assert.NilError(t, err, "fsck_apfs failed: %s", out)
+	cmd = exec.CommandContext(ctx, "hdiutil", "detach", dev)
+	out, err = cmd.CombinedOutput()
+	assert.NilError(t, err, "hdiutil detach failed: %s", out)
+
+	// Re-attach with ownership enabled so os.Stat reflects on-disk UIDs.
+	cmd = exec.CommandContext(ctx, "hdiutil", "attach", imgPath,
+		"-mountpoint", mntDir, "-owners", "on")
+	out, err = cmd.CombinedOutput()
+	assert.NilError(t, err, "hdiutil re-attach failed: %s", out)
+	defer func() {
+		_ = exec.CommandContext(ctx, "hdiutil", "detach", mntDir, "-force").Run()
+	}()
+
+	fi, err := os.Stat(filepath.Join(mntDir, "testfile.txt"))
+	assert.NilError(t, err)
+	stat := fi.Sys().(*syscall.Stat_t)
+	assert.Equal(t, stat.Uid, uint32(0), "expected uid=0, got uid=%d", stat.Uid)
+	assert.Equal(t, stat.Gid, uint32(0), "expected gid=0, got gid=%d", stat.Gid)
+}
--- a/pkg/guestpatch/macos/macos_darwin.go
+++ b/pkg/guestpatch/macos/macos_darwin.go
@@ -8,18 +8,20 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"time"
 
 	"github.com/sirupsen/logrus"
 
+	"github.com/lima-vm/lima/v2/pkg/apfs"
 	"github.com/lima-vm/lima/v2/pkg/imgutil/nativeimgutil/asifutil"
 	"github.com/lima-vm/lima/v2/pkg/limatype/dirnames"
 	"github.com/lima-vm/lima/v2/pkg/lockutil/mntlockutil"
 	"github.com/lima-vm/lima/v2/pkg/osutil"
 )
 
+// attachImageWithRetry retries `diskutil image attach` because the
+// command occasionally returns "Resource temporarily unavailable".
 func attachImageWithRetry(ctx context.Context, disk string, retry int) (*asifutil.AttachedDisk, error) {
 	var (
 		attached *asifutil.AttachedDisk
@@ -35,9 +37,20 @@ func attachImageWithRetry(ctx context.Context, disk string, retry int) (*asifuti
 	return attached, err
 }
 
+// Patch prepares a macOS guest disk for first boot. It writes the
+// LaunchDaemon plist, init script, and setup markers via a noowners
+// mount, then fixes file ownership by patching APFS inode records
+// directly on the raw disk image. No sudo required.
 func Patch(ctx context.Context, disk string) error {
-	// Retry `diskutil image attach -plist -nomount` a few times because
-	// the command occasionally returns "Resource temporarily unavailable".
+	if err := patchWriteGuestFiles(ctx, disk); err != nil {
+		return err
+	}
+	return patchFixOwnership(ctx, disk)
+}
+
+// patchWriteGuestFiles attaches the disk image, mounts the Data
+// volume with noowners, writes guest files, then detaches.
+func patchWriteGuestFiles(ctx context.Context, disk string) error {
 	attached, err := attachImageWithRetry(ctx, disk, 3)
 	if err != nil {
 		return fmt.Errorf("failed to attach disk: %w", err)
@@ -47,7 +60,7 @@ func Patch(ctx context.Context, disk string) error {
 	}
 	dataDevPath := "/dev/" + attached.Data
 	defer func() {
-		// Just detaching the data slice is enough to let the system detach the whole ASIF.
+		// Detaching the data slice is enough to detach the whole ASIF.
 		if err := asifutil.DetachASIF(dataDevPath); err != nil {
 			logrus.WithError(err).Warnf("failed to detach %q (%q)", dataDevPath, disk)
 		}
@@ -68,22 +81,52 @@ func Patch(ctx context.Context, disk string) error {
 		}
 	}()
 
-	if err = patchDiskRegularPhase(ctx, dataDevPath, mnt); err != nil {
-		return fmt.Errorf("failed to patch macOS disk (phase 1/2): %w", err)
+	return writeGuestFiles(ctx, dataDevPath, mnt)
+}
+
+// patchFixOwnership attaches the disk image and patches APFS inode
+// records on the raw container device to set root:wheel ownership.
+func patchFixOwnership(ctx context.Context, disk string) error {
+	attached, err := attachImageWithRetry(ctx, disk, 3)
+	if err != nil {
+		return fmt.Errorf("failed to attach disk for ownership fix: %w", err)
+	}
+	if attached == nil || attached.Data == "" {
+		return errors.New("failed to find data slice in attached disk")
+	}
+	dataDevPath := "/dev/" + attached.Data
+	defer func() {
+		if err := asifutil.DetachASIF(dataDevPath); err != nil {
+			logrus.WithError(err).Warnf("failed to detach %q (%q)", dataDevPath, disk)
+		}
+	}()
+
+	if attached.Container == "" {
+		return errors.New("diskutil did not report an APFS container device")
 	}
 
-	// Fix up the file ownership inside the disk.
-	// Invokes sudo.
-	if err = patchDiskPrivilegedPhase(ctx, dataDevPath, mnt); err != nil {
-		return fmt.Errorf("failed to patch macOS disk (phase 2/2): %w", err)
+	// Patch APFS inode records via the raw container device.
+	// The noowners mount stores files with uid=99 (nobody);
+	// LaunchDaemon plists must be owned by root:wheel for launchd
+	// to load them, so we patch them to UID 0 / GID 0 directly.
+	containerDev := "/dev/r" + attached.Container
+	if err = apfs.Chown(containerDev, apfs.VolRoleData, 0, 0,
+		"private/var/db/.AppleSetupDone",
+		"Library/User Template/.skipbuddy",
+		"usr/local/sbin",
+		"usr/local/sbin/lima-macos-init.sh",
+		"Library/LaunchDaemons/io.lima-vm.lima-macos-init.plist",
+	); err != nil {
+		return fmt.Errorf("failed to fix file ownership on disk: %w", err)
 	}
 
 	return nil
 }
 
-func patchDiskRegularPhase(ctx context.Context, dataSliceDevice, mnt string) error {
-	// Enable "noowners" to allow non-root users to write to the mounted volume.
-	// The ownership is fixed up in patchDiskPrivilegedPhase.
+// writeGuestFiles mounts the data volume with noowners and writes the
+// LaunchDaemon plist, init script, and setup markers.
+func writeGuestFiles(ctx context.Context, dataSliceDevice, mnt string) error {
+	// Mount with "noowners" so non-root users can write to the volume.
 	if err := osutil.Mount(ctx, "apfs", dataSliceDevice, mnt, []string{"noowners"}); err != nil {
 		return err
 	}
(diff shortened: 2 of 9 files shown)


jandubois:
Origin story: https://github.com/lima-vm/lima/pull/4595#issuecomment-3949042391

I've created a spec using Opus 4.6 searching for all available information about writing to an APFS volume. This includes reading the sources of several GPL licensed packages. I asked it to generate a spec for everything we need, so I could have it do a "clean-room" implementation in a separate session without reference to any existing source code.

---

# APFS In-Place UID/GID Patching: Implementation Plan

## Goal

Write a minimal APFS tool (`pkg/apfs/chown.go`) that modifies file ownership
(UID/GID) directly on a raw APFS disk image, bypassing the kernel VFS. This
eliminates the `sudo chown root:wheel` requirement from macOS guest disk patching.

## Why this works

APFS inode records store `owner` (UID) and `group` (GID) as plain little-endian
uint32 fields at fixed offsets within B-tree leaf nodes. Each metadata block is
protected by a single Fletcher-64 checksum. Modifying UID/GID requires changing
8 bytes and recomputing one checksum per affected block. No parent pointer updates,
no B-tree restructuring, no superblock changes.

Validated by Goebel 2019 (Universität der Bundeswehr): raw APFS block modification
with checksum recalculation produces a valid filesystem.

## License constraints

All open-source APFS implementations are GPL (v2 or v3). Lima is Apache 2.0.
We can read GPL code to understand the on-disk format (facts are not copyrightable)
but must write a clean implementation. Apple's APFS Reference PDF is the primary
specification.

## On-disk structures (from Apple APFS Reference)

### Object header — `obj_phys_t` (32 bytes)

| Offset | Size | Field | Type |
|--------|------|-------|------|
| 0 | 8 | `o_cksum` | uint64 — Fletcher-64 checksum |
| 8 | 8 | `o_oid` | uint64 — object identifier |
| 16 | 8 | `o_xid` | uint64 — transaction identifier |
| 24 | 4 | `o_type` | uint32 — type + storage class |
| 28 | 4 | `o_subtype` | uint32 |

### Container superblock — `nx_superblock_t` (starts at block 0)

Key fields (offsets from block start):

| Offset | Field | Notes |
|--------|-------|-------|
| 32 | `nx_magic` | Must be `0x4253584E` ("NXSB") |
| 36 | `nx_block_size` | Typically 4096 |
| 104 | `nx_xp_desc_blocks` | Checkpoint descriptor area size (mask `0x7FFFFFFF`) |
| 112 | `nx_xp_desc_base` | Checkpoint descriptor area start (physical block) |
| 136 | `nx_xp_desc_index` | Current checkpoint start index |
| 140 | `nx_xp_desc_len` | Current checkpoint length |
| 160 | `nx_omap_oid` | Container object map (physical address) |
| 184 | `nx_fs_oid[100]` | Volume OIDs (8 bytes each, 800 bytes total) |

### Volume superblock — `apfs_superblock_t`

| Offset | Field | Notes |
|--------|-------|-------|
| 32 | `apfs_magic` | Must be `0x42535041` ("APSB") |
| 128 | `apfs_omap_oid` | Volume object map (physical address) |
| 136 | `apfs_root_tree_oid` | Filesystem B-tree root (virtual OID) |
| 704 | `apfs_volname` | Volume name (256 bytes, UTF-8) |
| 964 | `apfs_role` | Volume role (Data = 0x0040, aka `APFS_VOL_ROLE_DATA`) |

### Object map — `omap_phys_t`

| Offset | Field | Notes |
|--------|-------|-------|
| 48 | `om_tree_oid` | Physical address of omap B-tree root |

### Omap B-tree keys/values

- `omap_key_t`: `ok_oid` (8) + `ok_xid` (8) = 16 bytes
- `omap_val_t`: `ov_flags` (4) + `ov_size` (4) + `ov_paddr` (8) = 16 bytes

### B-tree node — `btree_node_phys_t`

| Offset | Size | Field |
|--------|------|-------|
| 0 | 32 | `obj_phys_t` header |
| 32 | 2 | `btn_flags` |
| 34 | 2 | `btn_level` (0 = leaf) |
| 36 | 4 | `btn_nkeys` |
| 40 | 4 | `btn_table_space` (nloc_t: off u16 + len u16) |
| 44 | 4 | `btn_free_space` (nloc_t) |
| 48 | 4 | `btn_key_free_list` (nloc_t) |
| 52 | 4 | `btn_val_free_list` (nloc_t) |
| 56 | var | `btn_data[]` — ToC, keys, values |

If `btn_flags & BTNODE_ROOT`, a `btree_info_t` (40 bytes) sits at the end of the
block. This reduces the value area by 40 bytes.

### Table of Contents entries

- `kvloc_t` (8 bytes): key nloc (off u16 + len u16) + value nloc (off u16 + len u16)
  Used when `BTNODE_FIXED_KV_SIZE` is NOT set.
- `kvoff_t` (4 bytes): key offset u16 + value offset u16.
  Used when `BTNODE_FIXED_KV_SIZE` IS set.

ToC starts at: `btn_data + btn_table_space.off`
Key area starts at: `btn_data + btn_table_space.off + btn_table_space.len`
Key offsets in ToC are relative to key area start.
Value offsets are relative to:
- End of block (for non-root nodes)
- End of block minus `sizeof(btree_info_t)` = 40 (for root nodes)

Values grow backward (offset 0 = rightmost byte of value area).

### Filesystem B-tree keys

`j_key_t` (8 bytes): `obj_id_and_type` — upper 4 bits = type, lower 60 bits = CNID.

| Type | Value | Key struct |
|------|-------|------------|
| INODE | 3 | `j_inode_key_t`: just the 8-byte `j_key_t` |
| DIR_REC | 9 | `j_drec_key_t`: j_key_t + name_len(2) + name(var) |
|         |   | `j_drec_hashed_key_t`: j_key_t + name_len_and_hash(4) + name(var) |

For hashed keys: `name_len_and_hash` lower 10 bits = name length, upper 22 = hash.

### Filesystem B-tree values

`j_inode_val_t` (92+ bytes):

| Offset | Size | Field |
|--------|------|-------|
| 0 | 8 | parent_id |
| 8 | 8 | private_id |
| 16–47 | 32 | timestamps (create, mod, change, access) |
| 48 | 8 | internal_flags |
| 56 | 4 | nchildren/nlink |
| 60 | 4 | default_protection_class |
| 64 | 4 | write_generation_counter |
| 68 | 4 | bsd_flags |
| **72** | **4** | **owner (UID)** |
| **76** | **4** | **group (GID)** |
| 80 | 2 | mode |
| 82 | 2 | pad1 |
| 84 | 8 | uncompressed_size |
| 92 | var | xfields |

`j_drec_val_t`: `file_id` (8) + `date_added` (8) + `flags` (2) + xfields.

### Key constants

```
NX_MAGIC            = 0x4253584E
APFS_MAGIC          = 0x42535041
APFS_VOL_ROLE_DATA  = 0x0040  (or check Apple spec; may be different encoding)

OBJ_PHYSICAL        = 0x40000000
OBJ_EPHEMERAL       = 0x80000000
OBJECT_TYPE_NX_SUPERBLOCK  = 0x01
OBJECT_TYPE_BTREE_NODE     = 0x03
OBJECT_TYPE_OMAP           = 0x0B
OBJECT_TYPE_CHECKPOINT_MAP = 0x0C
OBJECT_TYPE_FS             = 0x0D

BTNODE_ROOT          = 0x0001
BTNODE_LEAF          = 0x0002
BTNODE_FIXED_KV_SIZE = 0x0004

APFS_TYPE_INODE   = 3
APFS_TYPE_DIR_REC = 9
OBJ_ID_MASK       = 0x0FFFFFFFFFFFFFFF
OBJ_TYPE_SHIFT    = 60

ROOT_DIR_INODE_NUM = 2   (root directory "/"'s inode number)
```

## Fletcher-64 checksum algorithm

Input: block bytes [8..blockSize), treated as little-endian uint32 words.

```
MOD = 0xFFFFFFFF
sum1, sum2 = 0, 0
for each uint32 word w:
    sum1 = (sum1 + uint64(w)) % MOD
    sum2 = (sum2 + sum1) % MOD
ck_low  = MOD - ((sum1 + sum2) % MOD)
ck_high = MOD - ((sum1 + ck_low) % MOD)
checksum = (ck_high << 32) | ck_low
```

Store in `o_cksum` (bytes 0–7 of the block).

## Navigation path: disk image to inode

1. **Read block 0** → container superblock. Validate magic, checksum.
2. **Scan checkpoint descriptor area** for superblock with highest valid `o_xid`.
   Area = `nx_xp_desc_blocks` blocks starting at `nx_xp_desc_base`.
   For each block: if type is NX superblock, validate checksum + magic, keep
   highest xid.
3. **Read container omap** at `nx_omap_oid` (physical). Get `om_tree_oid`.
4. **For each volume** in `nx_fs_oid[]` (skip zeros):
   Look up virtual OID in container omap B-tree → physical address.
   Read volume superblock. Check `apfs_volname` or `apfs_role` to find
   the Data volume.
5. **Read volume omap** at `apfs_omap_oid`. Get `om_tree_oid`.
6. **Resolve filesystem root** by looking up `apfs_root_tree_oid` (virtual)
   in volume omap → physical address of fs B-tree root.
7. **Resolve path** component by component:
   - Start with root dir CNID = 2
   - For each path component, search fs B-tree for dir rec key:
     `(parent_cnid | (APFS_TYPE_DIR_REC << 60))` + name
   - Get `file_id` from `j_drec_val_t` → next CNID
8. **Find target inode**: search fs B-tree for key
   `(target_cnid | (APFS_TYPE_INODE << 60))`
9. **Modify** `owner` and `group` at offsets +72 and +76 in the `j_inode_val_t`.
10. **Recompute Fletcher-64** for the modified block, write back.

## B-tree search algorithm

For a given B-tree root (physical block address):

```
func search(rootBlock, targetKey):
    node = readBlock(rootBlock)
    while true:
        entries = readToC(node)  // kvoff_t or kvloc_t depending on BTNODE_FIXED_KV_SIZE
        if node.btn_level == 0:  // leaf
            for each entry in entries:
                key = readKey(node, entry)
                if key matches targetKey:
                    return readValue(node, entry), node
            return not found
        else:  // internal node
            // Values in internal nodes are child pointers (uint64 OIDs)
            // Find the last key <= targetKey
            childOID = binary search entries for greatest key <= targetKey
            // Resolve OID to physical address via omap (for virtual trees)
            // or use directly (for physical trees)
            node = readBlock(resolve(childOID))
```

### Key comparison

**Omap B-tree** (fixed-size keys): compare `ok_oid` first, then `ok_xid`.
For lookups, find the entry with matching `ok_oid` and highest `ok_xid` ≤
our target xid.

**Filesystem B-tree** (variable-size keys): compare `obj_id_and_type` as
uint64 first. For dir rec keys with same obj_id_and_type, compare by name
(or name hash for case-insensitive volumes).

### Reading keys/values from a node

```
tocStart = 56 + btn_table_space.off    // 56 = sizeof(btree_node_phys_t fixed part)
keyAreaStart = tocStart + btn_table_space.len

if BTNODE_FIXED_KV_SIZE set:
    toc[i] = kvoff_t at tocStart + i*4
    key = block[keyAreaStart + toc[i].k ..]
    // value offset relative to value area end
else:
    toc[i] = kvloc_t at tocStart + i*8
    key = block[keyAreaStart + toc[i].k.off .. +toc[i].k.len]
    // value offset and length from toc[i].v

valueAreaEnd = blockSize                    // for non-root
             = blockSize - 40              // for root (minus btree_info_t)

// For fixed-size values:
value = block[valueAreaEnd - toc[i].v - valueSize .. valueAreaEnd - toc[i].v]
// For variable-size values:
value = block[valueAreaEnd - toc[i].v.off - toc[i].v.len .. valueAreaEnd - toc[i].v.off]
```

Wait — need to double-check the value offset convention. From the Apple spec and
reference implementations: the value offset is relative to the END of the value
area, measured as a distance backward. So:

```
valueStart = valueAreaEnd - valueOffset - valueLength
```

Actually more precisely: the value area END is the reference point. `v.off` is
the offset from this reference point to the start of the value data.

```
valueAddr = valueAreaEnd - v.off    // This points to the END of the value
// No wait — v.off is the start offset measured from the end, so:
valueAddr = valueAreaEnd - v.off - v.len   // ??? Need to verify
```

**NOTE**: Must verify the exact value offset convention against the Apple spec
during implementation. The key detail is whether `v.off` points to the first
byte or the last byte of the value relative to the reference point.

## Implementation structure

### Package: `pkg/apfs/`

Files:
- `chown.go` — public API: `Chown(diskPath string, filePath string, uid, gid uint32) error`
- `structures.go` — Go struct definitions for all on-disk types
- `fletcher64.go` — checksum computation
- `btree.go` — B-tree node reading and search
- `container.go` — container superblock + checkpoint scanning
- `volume.go` — volume superblock + omap resolution
- `chown_test.go` — tests (see below)

### Public API

```go
// Chown changes the owner and group of a file on an unmounted APFS disk image.
// The path is relative to the volume root (e.g., "Library/LaunchDaemons/foo.plist").
// The disk image must not be mounted.
func Chown(diskPath string, volumeName string, filePath string, uid, gid uint32) error
```

### Integration with macOS guest patching

In `pkg/guestpatch/macos/macos_darwin.go`, after writing the LaunchDaemon plist
via `noowners` mount, call:

```go
apfs.Chown(disk, "Data", "Library/LaunchDaemons/io.lima-vm.lima-macos-init.plist", 0, 0)
```

This replaces the entire privileged phase (`sudo chown root:wheel`).

## Testing strategy

1. **Unit tests** for Fletcher-64 (use known test vectors from Apple spec or
   from a real APFS block).
2. **Integration test**: Create a small APFS disk image using `diskutil` on
   macOS, write a test file with known UID, run Chown, mount the image, verify
   `stat` shows the new UID/GID.
3. **End-to-end**: `limactl start` creates a macOS VM without sudo prompt.

## Risks

1. **Value offset convention**: The exact meaning of value offsets in B-tree nodes
   needs verification during implementation. If we get this wrong, we'll read
   garbage. Mitigated by: validating parsed inode fields (timestamps should be
   reasonable, mode should be a valid file mode, etc.).
2. **Hashed vs non-hashed directory records**: macOS volumes may use either format
   depending on case sensitivity settings. We need to handle both
   `j_drec_key_t` and `j_drec_hashed_key_t`.
3. **Multiple volumes**: A macOS disk has System and Data volumes. We need to
   target the Data volume specifically (by name or role).
4. **Checkpoint descriptor area scanning**: Block 0 may not be the latest
   superblock. Must scan the checkpoint area.

jandubois:
During the implementation session we ran into several problems that were insufficiently covered by the original spec. They were resolved by re-reading the Apple APFS spec and local debugging, with no reference to external source code.

Once the new tests passed, and a macOS VM could be started without `sudo` I asked Opus 4.6 to compare our final code with the `apfsfuse` code base to see if the logic we added matches what that implementation does. Again, no code from `apfsfuse` was using in the implementation of this PR.

This is the analysis of the deviation of our code from the original spec, and comparison to `apfsfuse`:

---

# Deviations from the APFS Chown Plan

The original plan assumed a clean path: open a raw APFS container, walk the
B-tree, patch inodes. Seven problems forced changes during implementation. Each
arose from a gap between the Apple APFS Reference and real macOS disk images.

## 1. GPT partition table wrapping the APFS container

The plan said: "open file O_RDWR, read block 0, validate NX magic." On real
disk images, block 0 holds a protective MBR, not the NX superblock. Both Lima's
ASIF images and DMGs from `hdiutil create` wrap the APFS container inside a GPT
partition.

**Fix**: `openContainer()` checks block 0 for NX magic. If absent, it parses
the GPT header at LBA 1, scans partition entries for the APFS Container type
GUID (`7C3457EF-0000-11AA-AA11-00306543ECAC`), and records the partition's byte
offset in `container.baseOffset`. All subsequent block I/O adds this offset.

**Cost**: ~85 lines of GPT parsing code.

## 2. Integration test runs without root

The plan required root to create APFS volumes. In practice, `hdiutil create`
and `hdiutil attach` work as a regular user. The test creates a DMG, writes a
file, runs `Chown`, and verifies ownership, all without `sudo`.

One subtlety: `hdiutil create -fs APFS` produces a volume with
`APFS_VOL_ROLE_NONE` (role 0), not `APFS_VOL_ROLE_DATA` (0x0040). The test
uses role 0.

## 3. B-tree value offset points to value start, not end

The plan said "values grow backward from valueAreaEnd" but left the offset
formula ambiguous. The Apple APFS Reference says `v.off` is "the byte offset
from the end of the value area," without clarifying whether it points to the
value's beginning or end.

The initial code subtracted the value size:

```
valStart = valueAreaEnd - vOff - valueSize   // wrong
```

The correct formula omits the size:

```
valStart = valueAreaEnd - vOff               // correct
```

Confirmed by the apfs-fuse source: `result.val = m_vals_end - kvp->v`. This bug
appeared in all seven value-access sites across `omapLookup`, `lookupDirEntry`,
and `chownInode`.

## 4. Ownership verification requires `-owners on`

The plan verified ownership by remounting with `hdiutil attach` and calling
`os.Stat`. This always showed uid 501 because `hdiutil attach` defaults to
`noowners` for external disk images. Under `noowners`, macOS maps every on-disk
UID to the current user.

**Fix**: remount with `hdiutil attach -owners on`. The kernel then returns
actual on-disk UIDs, confirming our raw APFS modifications through an
independent code path.

## 5. Sparse shadow images require device node access

The plan passed the disk image file path directly to `apfs.Chown()`. macOS VM
images use Apple's sparse shadow format (`shdw` magic), which our GPT parser
cannot read.

**Fix**: attach the image with `diskutil image attach --plist --nomount` and
pass the raw character device (`/dev/r<partition>`) to `Chown()`. The whole-disk
device (`/dev/rdisk7`) returned "resource busy" because the synthesized APFS
container claimed it; the partition device (`/dev/rdisk7s2`) works.

This required splitting `Patch()` into two phases with separate attach/detach
cycles: one mounted (to write files) and one unmounted (to patch ownership).

## 6. Filesystem B-tree keys sort by (obj_id, type), not raw uint64

The plan described filesystem key comparison as "compare `obj_id_and_type`
(uint64) first." Raw uint64 comparison interleaves object IDs and types
incorrectly: it places a type-4 key for object 100 before a type-9 key for
object 99.

APFS sorts filesystem keys by object ID first, then by type. The fix extracts
each field and compares them independently:

```go
aID := a & objIDMask      // low 60 bits
aType := a >> objTypeShift // high 4 bits
```

## 7. Directory record name hash for B-tree descent

After fixing key comparison, path resolution failed on directories with many
entries spanning multiple leaf nodes. When several internal-node keys share the
same header (same parent CNID + type DIR_REC), the 22-bit name hash determines
which child to descend into. Without comparing the hash, the code always picked
the last matching child and overshot the target.

The hash algorithm (from the Apple APFS Reference):

1. NFD-normalize the filename
2. Case-fold to lowercase (for case-insensitive volumes)
3. Encode as UTF-32LE (4 bytes per codepoint, **without** null terminator)
4. Compute CRC-32C
5. Complement the result (cancels with CRC-32C's standard finalization XOR)
6. Mask to low 22 bits

The critical detail: the CRC input is UTF-32LE, not UTF-8. The spec says
"null-terminated UTF-32 string," but on-disk hashes match computation without
the null terminator.

**Cost**: ~70 lines for `drecNameHash()` and `compareDrecKey()`. Applied only
to `lookupDirEntry`; `chownInode` needs only header comparison since INODE keys
are unique per object.

## Summary

| # | Issue | Lines added | Root cause |
|---|-------|-------------|------------|
| 1 | GPT partition table | ~85 | Plan ignored disk image container format |
| 2 | No root for tests | -3 | `hdiutil` works unprivileged |
| 3 | Value offset formula | 0 (fix) | Ambiguous spec wording |
| 4 | `-owners on` mount | ~2 | `noowners` default on external images |
| 5 | Device node access | ~30 | Sparse shadow format unreadable as file |
| 6 | FS key comparison | ~15 | Plan said "compare uint64" too loosely |
| 7 | Name hash for descent | ~70 | Directories spanning multiple leaf nodes |

The plan estimated ~400 lines for `chown.go`. The final file is ~700 lines.
Most growth came from GPT parsing (1) and the name hash algorithm (7). The core
B-tree traversal logic matches the plan.

## Comparison against apfs-fuse

Four deviations (1, 3, 6, 7) concern the APFS on-disk format. All four match
the apfs-fuse reference implementation after our fixes. Three (2, 4, 5) concern
test infrastructure or Lima integration and have no counterpart in the reference.

**1. GPT** — apfs-fuse has a full GPT parser in `GptPartitionMap.cpp`. It finds
the APFS partition GUID and passes the byte offset to `ApfsContainer`, exactly
as we do. We missed it because the GPT code lives in the application layer
(`apfsfuse/ApfsFuse.cpp`), separate from the APFS library we studied.

**3. Value offset** — apfs-fuse: `result.val = m_vals_start - kvp->v` (no
length subtraction). linux-apfs-rw: `*off = sb->s_blocksize - entry->v.off`
(same). Both confirm that `v.off` points to the value's first byte. Our spec
flagged this as uncertain (lines 270–289) but wrote the wrong formula anyway.

**6. Key comparison** — apfs-fuse's `CompareStdDirKey` rotates the packed field
before comparing: `(val << 4) | (val >> 60)`. This moves the 4-bit type from
bits 63–60 to bits 3–0, making obj_id the most significant field — identical in
effect to our `compareFSKeyHeader()`. Our spec said "compare as uint64 first,"
which reads as raw comparison. The rotation trick is right there in apfs-fuse;
we described it too loosely.

**7. Name hash** — apfs-fuse's `HashFilename()` in `Util.cpp`:
- Converts UTF-8 to UTF-32 (native endian = LE on all targets)
- NFD-normalizes, case-folds
- CRC-32C over the UTF-32 bytes, no null terminator
- Returns raw CRC (no final complement), masked to 22 bits

Our `drecNameHash()` does the same: `^crc32.Checksum()` cancels Go's standard
final XOR, giving raw CRC; UTF-32LE encoding; no null terminator; 22-bit mask.
The spec mentioned hashed keys (line 118) and flagged them as a risk (line 339)
but never extracted the hash algorithm from the reference code.

### Why the spec missed these

| # | What apfs-fuse had | Why we missed it |
|---|-------------------|------------------|
| 1 | `GptPartitionMap.cpp` | GPT code is in the app, not the APFS library |
| 3 | `m_vals_start - kvp->v` | Spec flagged uncertainty but guessed wrong |
| 6 | `(val<<4)\|(val>>60)` rotation | Described as "compare uint64" too loosely |
| 7 | `HashFilename()` with CRC-32C of UTF-32LE | Noted hashed keys as risk, didn't extract algorithm |

No changes need reverification. All four fixes match both apfs-fuse and
linux-apfs-rw.

jandubois:
### Learnings

After you create an implementation plan you should review it multiple times to make sure it does not contain any ambiguities. Most of the implementation phase was spend debugging and fixing issues where our spec was not explicit enough. This would have been much easier to correct before the "clean-room" phase, when we still had access to reference implementations.

pkg/apfs/chown_darwin_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+// TestChownIntegration creates a real APFS disk image with hdiutil,
+// writes a test file, runs Chown on the raw image, and verifies
+// ownership via os.Stat after remounting with ownership enabled.
+// No root required.
+func TestChownIntegration(t *testing.T) {
```
AkihiroSuda, line 21: Can we run `fsck_apfs` ?

pkg/guestpatch/macos/macos_darwin.go
AkihiroSuda, line 1: Please also update https://lima-vm.io/docs/usage/guests/macos/

jandubois:
I've updated the test to call `fsck_apfs` and removed the password prompt info from the docs.

I've added another commit that performs a few extra checks to make sure everything in the filesystem is expected: btree node types, inode number matches private id, inode is fully within the block.

All these seem uncontroversial to me, but I'm less sure about my last guardrail: I verify that the existing UID/GID are both `99`, which is the value set when you create a file while the volume is mounted using `noowners`.

I believe this is our only use case, but I can remove it if it gets in the way.

I've left things in separate commits for ease of review; I can squash when we are ready to merge.

AkihiroSuda:
Thanks, could you squash, then LGTM

jandubois:
Squashed

AkihiroSuda (approved):
Thanks

PR #4635, opened 2026-02-26
macOS guest: support graceful shutdown via `limactl stop`

### LINKED ISSUE #4610: macOS guest: support graceful shutdown via `limactl stop`
Currently a macOS VM has to be stopped from the guest for graceful shutdown

AkihiroSuda:
`limactl stop` invokes `sudo /sbin/shutdown -h now` via SSH

Fix #4610

### DIFF
--- a/pkg/driver/vz/vz_driver_darwin.go
+++ b/pkg/driver/vz/vz_driver_darwin.go
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"runtime"
 	"time"
@@ -35,6 +36,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/osutil"
 	"github.com/lima-vm/lima/v2/pkg/ptr"
 	"github.com/lima-vm/lima/v2/pkg/reflectutil"
+	"github.com/lima-vm/lima/v2/pkg/sshutil"
 )
 
 var knownYamlProperties = []string{
@@ -438,7 +440,22 @@ func (l *LimaVzDriver) RunGUI() error {
 	return fmt.Errorf("RunGUI is not supported for the given driver '%s' and display '%s'", "vz", *l.Instance.Config.Video.Display)
 }
 
-func (l *LimaVzDriver) Stop(_ context.Context) error {
+func (l *LimaVzDriver) requestStopViaSSH(ctx context.Context) error {
+	sshExe, err := sshutil.NewSSHExe()
+	if err != nil {
+		return err
+	}
+	cmd := exec.CommandContext(ctx, sshExe.Exe,
+		append(sshExe.Args, "-F", l.Instance.SSHConfigFile, l.Instance.Hostname, "--",
+			"sudo", "/sbin/shutdown", "-h", "now")...)
+	logrus.Infof("Running shutdown command in the VM: %v", cmd.Args)
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("failed to run %v: %w (output=%s)", cmd.Args, err, string(out))
+	}
+	return nil
+}
+
+func (l *LimaVzDriver) Stop(ctx context.Context) error {
 	logrus.Info("Shutting down VZ")
 	canStop := l.machine.CanRequestStop()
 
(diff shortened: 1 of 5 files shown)


jandubois:
Doesn't work for me because `sudo` requires a password:

```
INFO[0000] [hostagent] Running shutdown command in the VM: [/usr/bin/ssh -F /Users/jan/.lima/macos/ssh.config lima-macos -- sudo /sbin/shutdown -h now]
WARN[0000] [hostagent] Failed to request shutdown via SSH  error="failed to run [/usr/bin/ssh -F /Users/jan/.lima/macos/ssh.config lima-macos -- sudo /sbin/shutdown -h now]: exit status 1 (output=sudo: a terminal is required to read the password; either use ssh's -t option or configure an askpass helper\nsudo: a password is required\n)"
```

Maybe we need a setuid root binary on the ci-data ISO (or in the patched image itself) to run the shutdown (after adjusting the permissions, so regular users can execute the binary).

AkihiroSuda:
> Doesn't work for me because sudo requires a password:

Please make sure to create a fresh instance, otherwise sudoers is not updated to allow passwordless shutdown

jandubois:
> Please make sure to create a fresh instance, otherwise sudoers is not updated to allow passwordless shutdown

This was a fresh instance. Where do I find the log / errors from the code trying to create the file?

Edit: I may be confused and only created a new instance for the mount test; will retest later.

pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
```diff
@@ -251,9 +248,35 @@ func createUser(ctx context.Context, u *cloudinittypes.User) error {
...
+	sudoersPath := "/etc/sudoers.d/90-cloud-init-users"
+	f, err := os.OpenFile(sudoersPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0o400)
+	if err != nil {
+		return fmt.Errorf("failed to open sudoers file %q: %w", sudoersPath, err)
+	}
+	defer f.Close()
+	if _, err = fmt.Fprintf(f, "%s %s\n", userName, sudo); err != nil {
+		return fmt.Errorf("failed to write to sudoers file %q for user %q: %w", sudoersPath, userName, err)
+	}
+	return f.Close()
```
jandubois, line 277: Double-close (harmless): you already have `defer f.Close()` above.

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -438,7 +440,22 @@ func (l *LimaVzDriver) RunGUI() error {
...
+	sshExe, err := sshutil.NewSSHExe()
+	if err != nil {
+		return err
+	}
+	cmd := exec.CommandContext(ctx, sshExe.Exe,
+		append(sshExe.Args, "-F", l.Instance.SSHConfigFile, l.Instance.Hostname, "--",
+			"sudo", "/sbin/shutdown", "-h", "now")...)
+	logrus.Infof("Running shutdown command in the VM: %v", cmd.Args)
+	if out, err := cmd.CombinedOutput(); err != nil {
+		return fmt.Errorf("failed to run %v: %w (output=%s)", cmd.Args, err, string(out))
```
jandubois, line 453: I was worried that ssh might be stopped before the command returns, but I didn't see this error, so I guess it is fine.

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -448,6 +465,14 @@ func (l *LimaVzDriver) Stop(_ context.Context) error {
...
 
+		if *l.Instance.Config.OS == limatype.DARWIN {
+			// macOS VM does not respond to l.machine.RequestStop(),
+			// so we need to run `shutdown -h now` in the VM via SSH for graceful shutdown.
+			if err := l.requestStopViaSSH(ctx); err != nil {
+				logrus.WithError(err).Warn("Failed to request shutdown via SSH")
+			}
+		}
+
 		timeout := time.After(5 * time.Second)
```
jandubois: We need to extend the timeout, at least for macOS:

```
INFO[0000] Sending SIGINT to hostagent process 88292
INFO[0000] [hostagent] Running shutdown command in the VM: [/usr/bin/ssh -F /Users/jan/.lima/macos/ssh.config lima-macos -- sudo /sbin/shutdown -h now]
ERRO[0005] [hostagent] vz timeout while waiting for stop status
INFO[0005] [hostagent] [VZ] - vm state change: stopped
ERRO[0005] [hostagent] accept tcp 127.0.0.1:54332: use of closed network connection
INFO[0005] Waiting for the instance to shut down
INFO[0006] The instance macos has shut down
```

pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
AkihiroSuda: fixed

pkg/driver/vz/vz_driver_darwin.go
AkihiroSuda: The situation is same as the ACPI shutdown for Linux instances?

pkg/driver/vz/vz_driver_darwin.go
AkihiroSuda: Extended to 30 secs

jandubois (approved):
Thanks, LGTM

PR #4637, opened 2026-02-26
Ensure only RAW images in cache(`imgconv/raw`) while keeping the original image

### LINKED ISSUE #4056: Cache should convert to RAW when needed
Many/most distro images are distributed in QCOW2 format, which is not supported by VZ.

So every time a VZ instance is started the image is converted into RAW format first. Which takes just a few seconds, but also increases disk space usage:

Normally the file would be copied via `clonefile` and not take up extra space. Any further changes would be written just to the copy via the copy-on-write mechanism, but the shared base image would only exist once.

The savings become even larger when the user runs multiple instances using the same image:

Normally each additional image would use another `clonefile` and not use extra space. But since we are converting to RAW separately for each instance, they are not even using clones of each other, but new identical copies of the original image.

Assume QCOW2 image is 500MB and you create 2 RAW instances: total space needed is 500MB for the QCOW2 version in the cache, plus 500MB for each instance.

If we had a 500MB RAW image in the cache then total space needed would be 500MB for the cache instance, plus pocket-change for the meta data of the 2 clones, saving 1GB of disk.

---

I suggest that the request to the cache downloaded should include the desired format.

1. The desired format is already in the cache: return a copy/clone of it
2. The image is in a different format in the cache: convert to the desired format, but keep original
3. Download image from source, return it if already in desired format
4. Convert to desired format, then delete the download

Case 2 can happen if the user used the image with QEMU before and now wants to create a VZ instance. It makes sense to keep the original format in case they want to create another QEMU instance. It is also possible that the QEMU instance still exists, in which case deleting the cache copy would not clear up any space because it is just a clone.

Case 4 is the case when a user only uses the default VZ driver. They will only need to keep the desired (RAW) version in the cache. In case they later need a QCOW2 version, they will have to download it again. But from then on both versions will be in the cache.

The cache needs to retain the checksum of the original image and not use the checksum of the converted copy for lookup.

unsuman:
Now when a fresh image is downloaded and if it's a non-ISO image, and the image type not supported by the driver it is then converted to RAW in the cache(`imgconv/raw`) and a `raw.digest` file is created. It also works for existing images in the cache that are not supported by the driver

Fixes #4056

### DIFF
--- a/pkg/downloader/downloader.go
+++ b/pkg/downloader/downloader.go
@@ -16,16 +16,21 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"slices"
 	"strings"
 	"sync/atomic"
 	"time"
 
 	"github.com/cheggaaa/pb/v3"
 	"github.com/containerd/continuity/fs"
+	"github.com/lima-vm/go-qcow2reader/image/raw"
 	"github.com/opencontainers/go-digest"
 	"github.com/sirupsen/logrus"
 
 	"github.com/lima-vm/lima/v2/pkg/httpclientutil"
+	"github.com/lima-vm/lima/v2/pkg/imgutil/nativeimgutil"
+	"github.com/lima-vm/lima/v2/pkg/imgutil/proxyimgutil"
+	"github.com/lima-vm/lima/v2/pkg/iso9660util"
 	"github.com/lima-vm/lima/v2/pkg/localpathutil"
 	"github.com/lima-vm/lima/v2/pkg/lockutil"
 	"github.com/lima-vm/lima/v2/pkg/progressbar"
@@ -57,10 +62,11 @@ type Result struct {
 }
 
 type options struct {
-	cacheDir       string // default: empty (disables caching)
-	decompress     bool   // default: false (keep compression)
-	description    string // default: url
-	expectedDigest digest.Digest
+	cacheDir              string // default: empty (disables caching)
+	decompress            bool   // default: false (keep compression)
+	description           string // default: url
+	expectedDigest        digest.Digest
+	supportedImageFormats []string
 }
 
 func (o *options) apply(opts []Opt) error {
@@ -107,6 +113,13 @@ func WithDescription(description string) Opt {
 func WithDecompress(decompress bool) Opt {
 	return func(o *options) error {
 		o.decompress = decompress
+		return nil
+	}
+}
+
+func WithImageFormats(supportedImageFormats []string) Opt {
+	return func(o *options) error {
+		o.supportedImageFormats = supportedImageFormats
 		return nil
 	}
 }
@@ -267,6 +280,81 @@ func getCached(ctx context.Context, localPath, remote string, o options) (*Resul
 	if _, err := os.Stat(shadData); err != nil {
 		return nil, nil
 	}
+	if _, err := os.Stat(shadDigest); err == nil {
+		if err := validateCachedDigest(shadDigest, o.expectedDigest); err != nil {
+			return nil, err
+		}
+	} else if match, lmCached, lmRemote, err := matchLastModified(ctx, shadTime, remote); err != nil {
+		logrus.WithError(err).Info("Failed to retrieve last-modified for cached digest-less image; using cached image.")
+	} else if !match {
+		logrus.Infof("Re-downloading digest-less image: last-modified mismatch (cached: %#q, remote: %#q)", lmCached, lmRemote)
+		return nil, nil
+	}
+
+	// Some drivers (e.g. vz) can only boot raw images. When the driver tells us
+	// which formats it supports and the cached image is not one of them, we save
+	// a converted raw copy next to the original at <shad>/imgconv/raw. See
+	// website/content/en/docs/dev/internals.md for the layout.
+	if len(o.supportedImageFormats) > 0 {
+		isISO, err := iso9660util.IsISO9660(shadData)
+		if err != nil {
+			logrus.WithError(err).Debugf("Skipping cache image conversion for %q (unable to check ISO9660)", shadData)
+		} else if !isISO {
+			imageFormat, err := nativeimgutil.DetectFormat(shadData)
+			if err != nil {
+				logrus.WithError(err).Debugf("Skipping cache image conversion for %q (unable to detect format)", shadData)
+			} else if !slices.Contains(o.supportedImageFormats, imageFormat) {
+				rawImgConvPath := filepath.Join(shad, "imgconv", "raw")
+				rawImgConvDigestPath := filepath.Join(shad, "imgconv", "raw.digest")
+
+				needConvert := false
+				if rawStat, err := os.Stat(rawImgConvPath); err != nil {
+					if errors.Is(err, os.ErrNotExist) {
+						needConvert = true
+					} else {
+						return nil, err
+					}
+				} else {
+					origStat, err := os.Stat(shadData)
+					if err != nil {
+						return nil, err
+					}
+					if origStat.ModTime().After(rawStat.ModTime()) {
+						needConvert = true
+					}
+				}
+
+				if needConvert {
+					logrus.Infof("Converted raw image is missing or stale; (re)converting now.")
+					converted, rawDigest, err := ensureRawInCache(ctx, shadData, imageFormat, o.expectedDigest)
+					if err != nil {
+						return nil, err
+					}
+					shadData = converted
+					if o.expectedDigest != "" {
+						o.expectedDigest = rawDigest
+						shadDigest = rawImgConvDigestPath
+					}
+				} else {
+					shadData = rawImgConvPath
+					if o.expectedDigest != "" {
+						if currentDigestData, err := os.ReadFile(rawImgConvDigestPath); err == nil {
+							currentDigest := strings.TrimSpace(string(currentDigestData))
+							if d, err := digest.Parse(currentDigest); err == nil {
+								o.expectedDigest = d
+								shadDigest = rawImgConvDigestPath
+							} else {
+								return nil, fmt.Errorf("invalid digest in raw digest file %q: %w", rawImgConvDigestPath, err)
+							}
+						} else {
+							return nil, fmt.Errorf("failed to read raw digest file %q: %w", rawImgConvDigestPath, err)
+						}
+					}
+				}
+			}
+		}
+	}
+
 	ext := path.Ext(remote)
 	logrus.Debugf("file %#q is cached as %#q", localPath, shadData)
 	if _, err := os.Stat(shadDigest); err == nil {
@@ -279,15 +367,8 @@ func getCached(ctx context.Context, localPath, remote string, o options) (*Resul
 			return nil, err
 		}
 	} else {
-		if match, lmCached, lmRemote, err := matchLastModified(ctx, shadTime, remote); err != nil {
-			logrus.WithError(err).Info("Failed to retrieve last-modified for cached digest-less image; using cached image.")
-		} else if match {
-			if err := copyLocal(ctx, localPath, shadData, ext, o.decompress, o.description, o.expectedDigest); err != nil {
-				return nil, err
-			}
-		} else {
-			logrus.Infof("Re-downloading digest-less image: last-modified mismatch (cached: %#q, remote: %#q)", lmCached, lmRemote)
-			return nil, nil
+		if err := copyLocal(ctx, localPath, shadData, ext, o.decompress, o.description, o.expectedDigest); err != nil {
+			return nil, err
 		}
 	}
 	res := &Result{
@@ -323,6 +404,34 @@ func fetch(ctx context.Context, localPath, remote string, o options) (*Result, e
 			return nil, err
 		}
 	}
+
+	// If the driver cannot use the format we just downloaded, save a converted
+	// raw copy at <shad>/imgconv/raw, keeping the original <shad>/data as-is.
+	//
+	// Gated on WithImageFormats() (the caller's signal that this is a VM disk
+	// image) and on the file not being an ISO 9660 image.
+	if len(o.supportedImageFormats) > 0 {
+		isISO, err := iso9660util.IsISO9660(shadData)
+		if err != nil {
+			logrus.WithError(err).Debugf("Skipping cache image conversion for %q (unable to check ISO9660)", shadData)
+		} else if !isISO {
+			format, err := nativeimgutil.DetectFormat(shadData)
+			if err != nil {
+				logrus.WithError(err).Debugf("Skipping cache image conversion for %q (unable to detect format)", shadData)
+			} else if !slices.Contains(o.supportedImageFormats, format) {
+				converted, rawDigest, err := ensureRawInCache(ctx, shadData, format, o.expectedDigest)
+				if err != nil {
+					return nil, fmt.Errorf("failed to convert image to raw: %w", err)
+				} else if converted != "" {
+					shadData = converted
+					if o.expectedDigest != "" {
+						o.expectedDigest = rawDigest
+					}
+				}
+			}
+		}
+	}
+
 	// no need to pass the digest to copyLocal(), as we already verified the digest
 	if err := copyLocal(ctx, localPath, shadData, ext, o.decompress, "", ""); err != nil {
 		return nil, err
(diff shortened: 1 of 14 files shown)


unsuman:
Are RAW images not supported on Windows?

AkihiroSuda:
> Are RAW images not supported on Windows?


https://learn.microsoft.com/en-us/windows/win32/fileio/sparse-file-operations says:
> It is up to the application to maintain sparseness by writing zeros with [FSCTL_SET_ZERO_DATA](https://learn.microsoft.com/en-us/windows/win32/api/winioctl/ni-winioctl-fsctl_set_zero_data).

Maybe it is not implemented in qemu-img.exe ?

pkg/downloader/downloader.go
```diff
@@ -336,6 +376,81 @@ func fetch(ctx context.Context, localPath, remote string, o options) (*Result, e
...
+
+	logrus.Infof("Converting qcow2 image to raw sparse format in cache: %q", imagePath)
+	rawPathTmp := imagePath + ".raw.tmp"
+	defer os.Remove(rawPathTmp)
+	diskUtil := proxyimgutil.NewDiskUtil(ctx)
+	if err := diskUtil.Convert(ctx, raw.Type, imagePath, rawPathTmp, nil, false); err != nil {
+		return false, "", fmt.Errorf("failed to convert %q to raw: %w", imagePath, err)
+	}
+
+	// Ensure the image is sparse to save cache space.
```
AkihiroSuda, line 466: Doesn't diskUtil handle this?

pkg/downloader/downloader.go
```diff
@@ -336,6 +376,81 @@ func fetch(ctx context.Context, localPath, remote string, o options) (*Result, e
...
+		return false
+	}
+	defer f.Close()
+
+	img, err := qcow2reader.Open(f)
+	if err != nil {
+		return false
+	}
+
+	return img.Type() == raw.Type
```
AkihiroSuda: vmdk, vhdx, etc. cannot be converted but still compatible with some drivers (e.g., qemu)

pkg/downloader/downloader.go
unsuman: No, I don't think so. Also `diskUtil.MakeSparse()` doesn't work on Windows too and it silently fails:

https://github.com/lima-vm/lima/blob/0a97b27939e19e51f65b1bb99ed3b2f572290b80/pkg/imgutil/proxyimgutil/proxyimgutil.go#L74-L83

https://github.com/lima-vm/lima/blob/0a97b27939e19e51f65b1bb99ed3b2f572290b80/pkg/qemuimgutil/qemuimgutil.go#L189-L192

unsuman:
> Maybe it is not implemented in qemu-img.exe ?

Yes seems so, as you pointed out

Is a `.img` file qcow2 or raw? I checked it on MacOS it shows raw, does it behave differently on Windows?

pkg/downloader/downloader.go
unsuman: Should we exclude Windows from this change?

pkg/driverutil/disk.go
```diff
@@ -95,10 +96,16 @@ func EnsureDisk(ctx context.Context, instDir, diskSize string, diskImageFormat i
 			return fmt.Errorf("failed to create disk %q: %w", diskPath, err)
 		}
 	} else {
-		if err := diskUtil.Convert(ctx, diskImageFormat, imagePath, diskPath, &diskSizeInBytes, false); err != nil {
-			return fmt.Errorf("failed to convert %q to %q: %w", imagePath, diskPath, err)
+		if downloader.IsRawImage(imagePath) {
```
AkihiroSuda:
```suggestion
		if nativeimgutil.IsRawImage(imagePath) {
```

AkihiroSuda:
> Is a .img file qcow2 or raw? I checked it on MacOS it shows raw, does it behave differently on Windows?

Can't predict the format from the extension

AkihiroSuda:
https://github.com/lima-vm/lima/actions/runs/22440179260/job/64980256758?pr=4637

> qemu-img: SetEndOfFile error: There is not enough space on the disk.

Does `--preallocation` flag work for this on Windows?
https://qemu-project.gitlab.io/qemu/tools/qemu-img.html

pkg/downloader/downloader.go
AkihiroSuda: Maybe submit a patch to qemu and wait until they merge it?

unsuman:
> https://github.com/lima-vm/lima/actions/runs/22440179260/job/64980256758?pr=4637
> 
> > qemu-img: SetEndOfFile error: There is not enough space on the disk.
> 
> Does `--preallocation` flag work for this on Windows? https://qemu-project.gitlab.io/qemu/tools/qemu-img.html

Gemini tells me:

> The --preallocation=off flag is actually the default behavior when extending a file's size. On Linux and macOS, this works perfectly because the underlying system call (ftruncate) naturally creates a "hole" (a sparse file) without consuming physical blocks.
> 
> However, on Windows, QEMU translates this resize request into the Win32 API call SetEndOfFile.
> Windows NTFS is extremely strict. If a file has not been explicitly branded with the FSCTL_SET_SPARSE attribute before SetEndOfFile is called, Windows assumes you want real, contiguous physical space.

pkg/downloader/downloader.go
unsuman: But this will block the PR and it might take weeks to get merged?

pkg/driverutil/disk.go
```diff
@@ -95,10 +96,16 @@ func EnsureDisk(ctx context.Context, instDir, diskSize string, diskImageFormat i
 			return fmt.Errorf("failed to create disk %q: %w", diskPath, err)
 		}
 	} else {
-		if err := diskUtil.Convert(ctx, diskImageFormat, imagePath, diskPath, &diskSizeInBytes, false); err != nil {
```
AkihiroSuda, line 98: Why wasn't this causing the error on Windows?

pkg/downloader/downloader.go
```diff
@@ -336,6 +376,81 @@ func fetch(ctx context.Context, localPath, remote string, o options) (*Result, e
...
+	f, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+	defer f.Close()
+
+	digester := algo.Digester()
+	if _, err := io.Copy(digester.Hash(), f); err != nil {
+		return "", err
+	}
```
AkihiroSuda: nit: you can use https://pkg.go.dev/github.com/opencontainers/go-digest#Algorithm.FromReader

pkg/downloader/downloader.go
AkihiroSuda: Not a fan of inflating the amount of platform-specific code.

Maybe give up on ensuring the raw cache on the downloading phase?

An alternative idea:
- The VM driver requests converting qcow2 to raw, only if the driver is incapable of booting qcow2.
  -  The cache key should be probably the sha256 of qcow2, not the URL. e.g., `~/Library/Caches/imgconv/by-orig-sha256/deadbeef/data`
- If the raw cache is available, the cache manager returns it. Otherwise, it converts the qcow2 cache to raw, and, prunes the qcow2 cache.

pkg/driverutil/disk.go
unsuman: Possibly because the `ubuntu-25.10-server-cloudimg-amd64.img` is a `qcow2` image under the hood on Windows and when we used `qemu-img resize .... 100G`, qemu did not talk to the Windows NTFS at all. Instead, it edited the `.qcow2` header([bytes 24-31](https://www.qemu.org/docs/master/interop/qcow2.html)), keeping the physical file size around 500MB.

pkg/downloader/downloader.go
unsuman: > An alternative idea:

SGTM! Should I refactor the logic of this PR?

AkihiroSuda: Yes, please.
You may update this PR or just open a new PR, just as you like.

jandubois: I assume there is no point for me reviewing this PR anymore because it will be replaced by an implementation that will be controlled by the driver and not the downloader?

unsuman: > imgconv/by-orig-sha256/deadbeef/data

Maintaining two different cache tree tends be complex

jandubois: > * The cache key should be probably the sha256 of qcow2, not the URL. e.g., `~/Library/Caches/imgconv/by-orig-sha256/deadbeef/data`

How does that work if the template doesn't include the digest? How would we find the cached image?

Obviously we can just enumerate all cache entries and look at their metadata; it would still be much faster than downloading anything, but not sure what the advantage is instead of caching by URL.

AkihiroSuda: > Maintaining two different cache tree tends be complex

You can merge the image conversion cache tree to the download cache, but I wonder if it can be more complex.
`limactl create` will have to call the driver API to see if it accepts `img.Type()`, and conditionally call the downloader with the converter when the driver does not accept `img.Type()`.

> How does that work if the template doesn't include the digest? How would we find the cached image?

When `limactl create` prepares qcow2 `image` (nee `basedisk`), it can compute and write the digest to a file like `image.digest`.
When the driver transforms `image` to `disk`, and it doesn't support `img.Type()`, it can look up the conversion cache by the content of `image.digest` to obtain the cached raw image.

jandubois: > When `limactl create` prepares qcow2 `image` (nee `basedisk`), it can compute and write the digest to a file like `image.digest`.

Yes, but when the downloader needs to decide if it should download the file or has it in the cache, it needs the cache key to look it up. And it can't get the checksum without downloading.

As I wrote above, it can search the whole cache, but it seems silly to use a cache key that cannot be used to look up the contents. I still don't understand how this simplifies anything.

unsuman: > The cache key should be probably the sha256 of qcow2, not the URL

I think this can be discussed in an another issue

> limactl create will have to call the driver API to see if it accepts img.Type(), and conditionally call the downloader with the converter when the driver does not accept img.Type().

But for fresh installs, we cannot know the `img.Type()` from the url if it's `.img`,  we will have to download it first 

Let me know if we can agree upon this?

For now drivers will send an array of supported image types to lima, then this will be passed to the downloader, and if it sees that the downloaded image type is not supported by the driver, it will force it to be converted to only `raw` and we will maintain the current logic in this PR to resolve the digest(i.e. if converted, we'll maintain a original.digest)

AkihiroSuda: SGTM, but we have to care about this condition:

```
- Initial state. No cache here.
- vz driver requires https://example.com/foo.qcow2
- Lima downloads the qcow2 image
- Lima converts the qcow2 to the raw cache
- Lima returns the cached raw image
- qemu driver requires the same URL
- ⚠️ Question: Can Lima return the cached raw image here? Or should it exactly return qcow2?
- User prunes the cache
- qemu driver requires the same URL again
- Lima downloads the qcow2 image
- Lima returns the cached qcow2 image, without any conversion
```

My suggestion for this question is to unconditionally return qcow2 to avoid non-deterministic behavior.

unsuman: >  My suggestion for this question is to unconditionally return qcow2 to avoid non-deterministic behavior.

Then we have to save the original image as it is and then alongside maintain an image conversion cache tree.

unsuman:
**Current flow:**
```
- Initial state. No cache here.
- vz driver requires https://example.com/foo.qcow2
- Lima downloads the qcow2 image
- Lima converts the qcow2 to the raw cache(imgconv/raw)
- Lima returns the cached raw image
- qemu driver requires the same URL
- +++ Lima returns the qcow2 image because the driver supports it

- User prunes the cache
- qemu driver requires the same URL again
- Lima downloads the qcow2 image
- Lima returns the cached qcow2 image, without any conversion
- vz driver requires the same url
- Lima converts the .qcow2 image to raw and caches it for future use
```
Only left with adding tests to `pkg/downloader/downloader_test.go`, I'll complete it before our meeting https://github.com/lima-vm/lima/discussions/4678

Also need to update `internals.md`

pkg/fileutils/download.go
```diff
@@ -19,18 +20,23 @@ import (
...
 	logrus.WithFields(fields).Infof("Attempting to download %s", description)
-	res, err := downloader.Download(ctx, dest, f.Location,
+	opts := []downloader.Opt{
 		downloader.WithCache(),
 		downloader.WithDecompress(decompress),
 		downloader.WithDescription(fmt.Sprintf("%s (%s)", description, path.Base(f.Location))),
 		downloader.WithExpectedDigest(f.Digest),
-	)
+	}
+	if strings.Contains(description, "image") && supportedImageFormats != nil {
```
AkihiroSuda: description is not expected to be parsed

AkihiroSuda: This suffices?
```suggestion
	if len(supportedImageFormats) > 0 {
```

AkihiroSuda: Just make sure to pass non-nil `supportedImageFormats` only for images

pkg/driver/driver.go
```diff
@@ -117,11 +115,12 @@ type Info struct {
...
-	RosettaEnabled       bool `json:"rosettaEnabled"`
-	RosettaBinFmt        bool `json:"rosettaBinFmt"`
+	CanRunGUI             bool     `json:"canRunGui,omitempty"`
+	DynamicSSHAddress     bool     `json:"dynamicSSHAddress"`
+	StaticSSHPort         bool     `json:"staticSSHPort"`
+	SkipSocketForwarding  bool     `json:"skipSocketForwarding"`
+	NoCloudInit           bool     `json:"noCloudInit"`
+	RosettaEnabled        bool     `json:"rosettaEnabled"`
+	RosettaBinFmt         bool     `json:"rosettaBinFmt"`
+	SupportedImageFormats []string `json:"supportedImageFormats,omitempty"`
```
AkihiroSuda, line 139: Does the order of the slice matter?

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -528,11 +423,12 @@ func (l *LimaVzDriver) Info() driver.Info {
...
-		SkipSocketForwarding: false,
-		CanRunGUI:            guiFlag,
-		RosettaEnabled:       l.rosettaEnabled,
-		RosettaBinFmt:        l.rosettaBinFmt,
+		DynamicSSHAddress:     false,
+		SkipSocketForwarding:  false,
+		CanRunGUI:             guiFlag,
+		RosettaEnabled:        l.rosettaEnabled,
+		RosettaBinFmt:         l.rosettaBinFmt,
+		SupportedImageFormats: []string{string(raw.Type), string(asif.Type)},
```
AkihiroSuda, line 567: Maybe put asif before raw, assuming that asif is more preferable

jandubois:
[Deep Review: 20260327-223819-pr-4637](https://jandubois.github.io/lima/20260327-223819-pr-4637.html)

jandubois:
I improved my review agent to let subagents run more multiple passes as long as they keep finding significant additional issues.

More comprehensive report: [Deep Review: 20260327-223819-pr-4637](https://jandubois.github.io/lima/20260327-223819-pr-4637.html)

pkg/driver/driver.go
unsuman: No, we only check if the downloaded/cached image type is present inside the slice provided by the drivers

AkihiroSuda:
Needs rebasing

pkg/downloader/downloader.go
```diff
@@ -299,6 +377,11 @@ func getCached(ctx context.Context, localPath, remote string, o options) (*Resul
 	return res, nil
 }
 
+func isNonISOImage(description, remote string, supportedImageFormats []string) bool {
+	remoteNoQuery, _, _ := strings.Cut(remote, "?")
+	return strings.Contains(description, "image") && len(supportedImageFormats) > 0 && path.Ext(remoteNoQuery) != ".iso"
```
AkihiroSuda: `description` isn't expected to be parsed by the machine

pkg/downloader/downloader.go
```diff
@@ -322,6 +405,26 @@ func fetch(ctx context.Context, localPath, remote string, o options) (*Result, e
 			return nil, err
 		}
 	}
+
+	// If the driver cannot use the format we just downloaded, save a converted
+	// raw copy at <shad>/imgconv/raw, keeping the original <shad>/data as-is.
+	if isNonISOImage(o.description, remote, o.supportedImageFormats) {
```
AkihiroSuda:
```suggestion
	if !isISOImage(...) {
```

pkg/downloader/downloader.go
AkihiroSuda: Just checking path.Ext isn't enough?

AkihiroSuda:
Needs rebase

pkg/downloader/downloader.go
```diff
@@ -300,6 +381,13 @@ func getCached(ctx context.Context, localPath, remote string, o options) (*Resul
 	return res, nil
 }
 
+// isISOImage reports whether the file at imagePath is an ISO 9660 image.
+// The check inspects file content rather than the filename extension (which
+// may be missing or misleading, e.g. presigned URLs with query strings).
+func isISOImage(imagePath string) (bool, error) {
+	return iso9660util.IsISO9660(imagePath)
+}
```
AkihiroSuda: Why not directly use `iso9660util.IsISO9660`

AkihiroSuda (dismissed):
Thanks

AkihiroSuda (approved):
Thanks

PR #4644, opened 2026-02-27
templates: update WSL2 template

arixmkii:
Latest finch-rootfs is Fedora 42 based, which is significant update over currently used Fedora 40 based one.

Fedora 40 is already EOL, where Fedora 42 should be good for about 2 months (but this is the latest provided by finch right now). I believe it would be a reasonable bump to add to 2.1 release scope.

I also update web site content to reflect the latest changes to template.


arixmkii:
vz test errors are either flaky tests or upstream issue and are not related to this changeset.

PR #4649, opened 2026-03-02
Revert "CI: temporarily disable release check for PRs"

AkihiroSuda:
Revert:
- #4646 

I was overreacting to https://www.stepsecurity.io/blog/hackerbot-claw-github-actions-exploitation .
We were not vulnerable because we had no workflows that permitted triggers from `pull_request_target`, `issue_comment`, or anything similar events.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,6 +13,13 @@ on:
     - "docs/**"
     - "website/**"
     - "**.md"
+  pull_request:
+    branches:
+    - 'master'
+    paths-ignore:
+    - "docs/**"
+    - "website/**"
+    - "**.md"
 env:
   GO111MODULE: on
   GOTOOLCHAIN: local


jandubois (approved):
We can revert it, if you want, but what actually is the benefit of running the `release` workflow on pull requests? We will still run it when it gets merged into `master`, isn't that enough?

AkihiroSuda:
> We can revert it, if you want, but what actually is the benefit of running the `release` workflow on pull requests? We will still run it when it gets merged into `master`, isn't that enough?

It is not enough to verify that the build artifacts are buildable

jandubois:
> It is not enough to verify that the build artifacts are buildable

That's fine, but please explain: how can a PR be buildable, but have a "release" failure that prevents us from merging it? What kind of failures are these? I'm not arguing about it, I'm curious about what I'm missing.

AkihiroSuda:
> > It is not enough to verify that the build artifacts are buildable
> 
> That's fine, but please explain: how can a PR be buildable, but have a "release" failure that prevents us from merging it? What kind of failures are these? I'm not arguing about it, I'm curious about what I'm missing.

A failure in Makefile or in release.yml itself.
Notably cross compilation in `make artifacts-linux`.

Issue #4652, opened 2026-03-02
Feature: Beta Guest OS like Ubuntu 26.04 LTS


reneleonhardt:
### Description

Would it be desired and possible with minimal effort to allow beta guests like Ubuntu 26.04?

Two years ago we saw the extremely slow adoption of Ubuntu 24.04 LTS after it's release due to several bugs.
Early availability of an `ubuntu-26.04-beta` template would make upgrading and collecting feedback much easier.

Ubuntu 26.04 is in Feature Freeze, Beta is planned for 2026-03-23 (in 3 weeks), Final for 2026-04-23 (in two months).
https://discourse.ubuntu.com/t/ubuntu-26-04-lts-the-roadmap/72740
https://packages.ubuntu.com/search?suite=resolute&section=all&arch=any&searchon=names&keywords=libc

Last Snapshot 4 released https://lists.ubuntu.com/archives/ubuntu-devel/2026-February/043504.html

AkihiroSuda:
We already have `limactl create template:experimental/ubuntu-26.04`

https://github.com/lima-vm/lima/blob/master/templates/experimental/ubuntu-26.04.yaml

PR #4653, opened 2026-03-02
`hack/update-template-ubuntu.sh`: Support daily URLs

norio-nomura:
- Update `hack/update-template-ubuntu.sh` to support daily URLs.
- ~~Add `templates/_images/ubuntu-26.04.yaml`~~
- Also updated `hack/inject-cmdline-to-template.sh` to support daily URLs. But, VM won't boot with kernel image, because kernel image file format has been changed to PE32+ which is not supported by VZ.

ref: #4652

### DIFF
--- a/hack/inject-cmdline-to-template.sh
+++ b/hack/inject-cmdline-to-template.sh
@@ -60,17 +60,18 @@ for ((i = 0; i < ${#locations[@]}; i++)); do
 done
 
 # 4. check the image location is supported
-if [[ -z ${location} ]]; then
-	echo "Failed to get the image location for ${template}" >&2
-	exit 1
-elif [[ ${location} == https://cloud-images.ubuntu.com/minimal/* ]]; then
+case "${location:-}" in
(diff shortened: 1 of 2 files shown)


norio-nomura:
> * But, VM won't boot with kernel image, because kernel image file format has been changed to PE32+ which is not supported by VZ.

The kernel image as the file format supported by VZ can be extracted by using https://raw.githubusercontent.com/torvalds/linux/master/scripts/extract-vmlinux

templates/_images/ubuntu-26.04.yaml
AkihiroSuda, line 1: We already have https://github.com/lima-vm/lima/blob/master/templates/experimental/ubuntu-26.04.yaml

norio-nomura: Dropped `templates/_images/ubuntu-26.04.yaml`

AkihiroSuda (approved):
Thanks

PR #4658, opened 2026-03-03
krunkit: enable support for nested virtualization

abiosoft:
This pull request adds support for nested virtualization to the `krunkit` driver.

Support for nested virtualization in `krunkit` driver:

* Updated `Cmdline` function in `krunkit_darwin_arm64.go` to append the `--nested` argument when `NestedVirtualization` is enabled in the configuration.

Configuration validation:

* Enhanced `validateConfig` in `krunkit_driver_darwin_arm64.go` to check that nested virtualization is only enabled on macOS 15 or newer, returning an error otherwise.

Documentation update:

* Updated comments in `templates/default.yaml` to clarify that nested virtualization is supported on Apple M3 or later with either `vmType: vz` or `vmType: krunkit`.


AkihiroSuda (approved):
Thanks

Issue #4660, opened 2026-03-03
local username validation incorrectly rejects valid usernames containing dots


jrwren:
## Description

Lima's username validation regex rejects usernames containing `.` (dots), but dots are valid characters in Linux usernames per `useradd` and `adduser` specifications.

## Current Behavior

ARN[0008] local username "jay.wren" is not a valid Linux username (must match "^[a-z_][a-z0-9_-]*$"); using "lima" instead

## Expected Behavior

Usernames like `jay.wren` should be accepted, as they are valid on Linux systems.

## Evidence

From `man useradd` on Ubuntu:
> Usernames may contain only lower and upper case letters, digits, underscores, or dashes. They can end with a dollar sign. **Dots are also allowed** but the username may not consist entirely of dots.

The shadow-utils `is_valid_name()` function allows dots:
https://github.com/shadow-maint/shadow/blob/master/lib/chkname.c

## Suggested Fix

Change the regex from:
^[a-z_][a-z0-9_-]*$

To:
^[a-z_][a-z0-9._-]*$

## Environment

- Lima version: [run `limactl --version`]
- OS: [e.g., macOS 14.x]
- Host username: `jay.wren`

jandubois:
> but dots are valid characters in Linux usernames per `useradd` and `adduser` specifications.

`useradd` yes, but `adduser` not.

Lima uses a conservative pattern to only allow default usernames that are accepted by most Linux tools.

You can always override manually, if you really want to:

```console
❯ limactl start -y --name test --set '.user.name="john.doe"' template:alpine-iso
...
❯ limactl shell test id
uid=501(john.doe) gid=1000(john.doe) groups=1000(john.doe)
```

You can configure this in your `~/.lima/_config/default.yaml` file to set it automatically on this host.

jrwren:
Ah, I see. Thank you.

jrwren:
I get:
```
WARN[0000] local username "jay.wren" is not a valid Linux username (must match "^[a-z_][a-z0-9_-]*$"); using "lima" instead 
```

with `--set '.user.name="jay.wren"' ` or with the following in my `~/.lima/_config/default.yaml`

```
user:
  name: "jay.wren"
```

Maybe the override isn't working as intended?

jrwren:
Oh, never mind. Now I see that the warning is lying to me about `using "lima" instead`

Still a bug IMO, as a log message is false, but functionality is working.

PR #4661, opened 2026-03-03
allow . in user.names

jrwren:
for #4660


jandubois:
I don't think we want to accept this change; see my comments in #4660.

PR #4662, opened 2026-03-04
makefile: raise error when VERSION cannot be determined

### LINKED ISSUE #4654: Makefile: raise an error when `$VERSION` is unset
https://github.com/lima-vm/lima/blob/7cf08751f544d4a1dcf6eec3cbc036b39cc13ba0/Makefile#L62

Lima strictly requires the `$VERSION` to be set, but it is not automatically set when built from a non-git tar ball such as 
https://github.com/lima-vm/lima/archive/refs/tags/v2.0.3.tar.gz

tharunn0:
Makefile: raise error when VERSION cannot be determined

When building from a GitHub release tarball, the source tree does not
contain git metadata, so `git describe` may fail and leave VERSION unset.

Add a validation check to stop the build if VERSION cannot be determined,
preventing binaries from being built with missing version metadata.

NOTE: This change only ensures VERSION is not empty and preserves the existing
behavior using `git describe --always`, so development builds can still
fall back to commit hashes.

Fixes #4654

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,8 +59,12 @@ endif
 
 PACKAGE := github.com/lima-vm/lima/v2
 
-VERSION := $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags)
+VERSION ?= $(shell git describe --match 'v[0-9]*' --dirty='.m' --always --tags 2>/dev/null)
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #4664, opened 2026-03-05
templates: add macos-15

AkihiroSuda:
Tried to add macos-13 and macos-14 as well, but no success
```
INFO[0000] Patching macOS disk "/Users/suda/.lima/macos-14/disk"
DEBU[0000] Executing command: [mount -t apfs -o noowners /dev/disk8s5 /Users/suda/.lima/_mnt/0]
FATA[0000] failed to mount "/dev/disk8s5" on "/Users/suda/.lima/_mnt/0": exit status 65
(output="mount_apfs: volume could not be mounted: Illegal byte sequence\nmount: /Users/suda/.lima/_mnt/0 failed with 65\n"
```
- https://github.com/lima-vm/lima/issues/4665

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -44,6 +44,7 @@ Linux distributions:
 - [`experimental/debian-sid`](./experimental/debian-sid.yaml): [experimental] Debian Sid
 
 Non-Linux:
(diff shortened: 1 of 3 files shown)


jandubois (approved):
Thanks, LGTM (haven't tested it, but the templates look fine)

PR #4669, opened 2026-03-06
Just warn when LIMA_HOME is on NFS, don't block

### LINKED ISSUE #4656: Allow to bypass nfs protect
### Description

Hello team :)

Currently, in a setup where I’m running VS Code inside a container ([coder.com](https://github.com/coder/coder)
), I have no choice but to use NFS for the home folder. Would it be possible to consider allowing this protection to be skipped via an environment variable?

kind regards
Thomas

jandubois:
Fixes #4656

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -148,7 +148,7 @@ func newApp() *cobra.Command {
 			return err
 		}
 		if nfs {
-			return errors.New("must not run on NFS dir")
+			logrus.Warn("LIMA_HOME should not be on an NFS mount")
 		}
 
 		if cmd.Flags().Changed("yes") && cmd.Flags().Changed("tty") {


cmd/limactl/main.go
```diff
@@ -148,7 +148,7 @@ func newApp() *cobra.Command {
 			return err
 		}
 		if nfs {
-			return errors.New("must not run on NFS dir")
+			logrus.Warn("must not run on NFS dir")
```
afbjorklund: maybe change the text as well, from "must" to "should" (shall)

the recommendation is still to use a local disk for $LIMA_HOME

afbjorklund:
I am anticipating a user request for turning off the warning with an environment variable, though.

At least that is the resort when `podman` starts complaining about various things in the setup... :-)

jandubois:
> I am anticipating a user request for turning off the warning with an environment variable, though.

Drop all the info and warning output:

```console
❯ cat $(which limactl-run)
#!/bin/bash
limactl --log-level error start "$@"
❯ limactl run -y template:alpine-iso
❯ limactl shell alpine-iso uname -a
Linux lima-alpine-iso 6.18.2-0-virt #1-Alpine SMP PREEMPT_DYNAMIC 2025-12-29 10:24:58 aarch64 Linux
```

If you want to filter specific warnings, you will have to do your own filtering, which unfortunately will change the logging format because STDOUT will not longer be a tty.

PR #4670, opened 2026-03-06
feature: accept `limactl shell [FLAGS...]` through `lima` CLI

### LINKED ISSUE #4651: `lima` CLI should accept `limactl shell` flags such as `--sync`
### Description

https://github.com/lima-vm/lima/blob/master/cmd/lima
https://github.com/lima-vm/lima/blob/master/cmd/lima.bat

`lima` CLI should accept `limactl shell` flags.

e.g.,
```
lima --sync . claude
```

unsuman:
Fixes #4651

### DIFF
--- a/cmd/lima
+++ b/cmd/lima
@@ -21,7 +21,7 @@ if [ "$#" -eq 1 ]; then
     base="$(basename "$0")"
     echo "Usage: ${base} [COMMAND...]"
     echo
-    echo "${base} is an alias for \"${LIMACTL} shell ${LIMA_INSTANCE}\"."
+    echo "${base} is an alias for \"${LIMACTL} shell --instance=${LIMA_INSTANCE}\"."
     echo "The instance name (\"${LIMA_INSTANCE}\") can be changed by specifying \$LIMA_INSTANCE."
     echo
     echo "The shell and initial workdir inside the instance can be specified via \$LIMA_SHELL"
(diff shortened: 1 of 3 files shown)


unsuman:
Does this work for Windows or need to update `cmd/lima.bat`?

cmd/lima
```diff
@@ -19,28 +19,64 @@ set -eu
...
+      break
+      ;;
+    *)
+      # Not a shell flag; stop parsing
+      break
+      ;;
+  esac
+done
+
 set - "$LIMA_INSTANCE" "$@"
```
AkihiroSuda: Maybe we can avoid enumerating all the flag names here by introducing a helper flag like `exec limactl shell --instance="$LIMA_INSTANCE" "$@"`

When `--instance` is specified, `limactl shell` should treat the first arg as the command, not the instance name. So, `lima` args can be almost directly passed to `limactl shell`.

jandubois: It feels a bit weird to have an optional positional argument that becomes required when a certain flag is not specified.

I can't think of a better way of dealing with this though, and think the simplification here makes it worth it.

AkihiroSuda: PR:
- https://github.com/lima-vm/lima/pull/4691

> It feels a bit weird to have an optional positional argument that becomes required when a certain flag is not specified.

Yes, this weird implementation detail is expected to be opaque to end users.

PR #4672, opened 2026-03-06
Try to patch inotify

iliasu-zinc:
### Description

When using virtiofs and inotify it seems that you can intermittently hit permission denied errors for example when running yarn install in a decently sized repository. These errors don't actually seem to be a problem with virtiofs.

virtiofs = works fine but no hot reload inside containers
virtiofs + mount inotify = hot reload inside container works but yarn install will crash with a permission denied issue

With this change hopefully we see the following;

virtiofs + mount inotify = yarn install works + hot reload works

This PR seemingly achieves this aim, but I haven't worked on Go in a long time so I'm not an expert this fix is only me debugging with strace and trying to patch what I see.

### Testing Done

1. yarn install inside a Docker container with a virtiofs-mounted volume completes without EACCES errors
2. Editing a source file on the host triggers a single hot-reload cycle (no loops) <- this was a side-effect of an earlier revision of this change, I included it to try to show how the testing was conducted.

### DIFF
--- a/pkg/guestagent/guestagent_linux.go
+++ b/pkg/guestagent/guestagent_linux.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"google.golang.org/protobuf/types/known/timestamppb"
@@ -48,13 +49,11 @@ func New(ctx context.Context, ticker ticker.Ticker, runtimeDir string) (Agent, e
 var _ Agent = (*agent)(nil)
 
 type agent struct {
-	// Ticker is like time.Ticker.
-	// We can't use inotify for /proc/net/tcp, so we need this ticker to
-	// reload /proc/net/tcp.
 	ticker                   ticker.Ticker
 	socketLister             *sockets.Lister
 	kubernetesServiceWatcher *kubernetesservice.ServiceWatcher
 	runtimeDir               string
+	recentChtimes            map[string]time.Time
 }
 
 type eventState struct {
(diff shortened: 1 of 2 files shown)


pkg/guestagent/guestagent_linux.go
```diff
@@ -48,13 +49,11 @@ func New(ctx context.Context, ticker ticker.Ticker, runtimeDir string) (Agent, e
 var _ Agent = (*agent)(nil)
 
 type agent struct {
-	// Ticker is like time.Ticker.
-	// We can't use inotify for /proc/net/tcp, so we need this ticker to
-	// reload /proc/net/tcp.
```
AkihiroSuda, line 53: Why remove?

pkg/guestagent/guestagent_linux.go
AkihiroSuda, line 1: - Squash the commits
- Add Signed-off-by line to the commit message to sign the DCO
- Make the PR title descriptive, please

pkg/guestagent/guestagent_linux.go
```diff
@@ -48,13 +49,11 @@ func New(ctx context.Context, ticker ticker.Ticker, runtimeDir string) (Agent, e
...
 	ticker                   ticker.Ticker
 	socketLister             *sockets.Lister
 	kubernetesServiceWatcher *kubernetesservice.ServiceWatcher
 	runtimeDir               string
+	recentChtimes            map[string]time.Time
```
AkihiroSuda, line 56: Needs a comment line to explain the key

pkg/guestagent/guestagent_linux.go
```diff
@@ -211,13 +210,39 @@ func (a *agent) Info(ctx context.Context) (*api.Info, error) {
...
+		logrus.Errorf("error in inotify handle. Event: %s, Error: %s", event, err)
+	}
+
+	if a.recentChtimes == nil {
+		a.recentChtimes = make(map[string]time.Time)
+	}
+	a.recentChtimes[location] = now
+
+	// Prevent unbounded growth during bulk operations
+	if len(a.recentChtimes) > 10000 {
```
AkihiroSuda, line 243: 10000 should be defined as a const

pkg/guestagent/guestagent_linux.go
```diff
@@ -211,13 +210,39 @@ func (a *agent) Info(ctx context.Context) (*api.Info, error) {
...
+	if err := os.Chtimes(location, local, local); err != nil {
+		logrus.Errorf("error in inotify handle. Event: %s, Error: %s", event, err)
+	}
+
+	if a.recentChtimes == nil {
+		a.recentChtimes = make(map[string]time.Time)
+	}
+	a.recentChtimes[location] = now
+
+	// Prevent unbounded growth during bulk operations
```
AkihiroSuda, line 242: Can we use LRU algorithm or something?

unsuman:
ping @iliasu-zinc

iliasu-zinc:
> ping @iliasu-zinc

Hi sorry didn't get much time, hopefully in the next week or so I will have time. I need to get through the PR comments, test against the latest release also. There is still some issues with this change as it's not a 100% fix but seems to greatly reduce the frequency of this issue so will need to try and investigate further.

PR #4676, opened 2026-03-07
Replace awesome_bot with lychee for link checking

### LINKED ISSUE #4675: Lint failures because of 429 (Too many requests) errors
I've seen this twice from actions running last night:

```
> Links 
  1. [L095] 429 https://github.com/cncf/foundation/blob/main/code-of-conduct.md  
```

Not sure if this is GitHub-wide, or if we are making too many requests in our workflow. Just creating the issue to track this; maybe we can close if it goes away on its own.

jandubois:
`awesome_bot` triggers GitHub API rate limits (429 errors) because it lacks authentication support. `lychee` accepts a GitHub token, which avoids rate limiting on `github.com` URLs.


AkihiroSuda (approved):
Thanks

PR #4679, opened 2026-03-08
Replace diff with instance ID check for boot signal files

### LINKED ISSUE #4659: [bug] limactl start hangs on opensuse-tumbleweed due to missing `diff`command
### Description

host: macOS 26.3

When starting an [opensuse-tumbleweed]([experimental/opensuse-tumbleweed](https://github.com/lima-vm/lima/tree/master/templates/experimental/opensuse-tumbleweed.yaml)), `limactl` start hangs indefinitely at the following stage:

`INFO [hostagent] Waiting for the final requirement 1 of 1: "boot scripts must have finished"`

Even though the VM is fully accessible via limactl shell and the OS has finished booting, the host agent never receives the "success" signal.

### Root Cause
The Lima host-agent executes a readiness check via SSH that relies specifically on the `diff` command to compare the boot-done signal with the metadata:

```
timeout 30s bash -c 'until sudo diff -q /run/lima-boot-done /mnt/lima-cidata/meta-data 2>/dev/null; do sleep 3; done'
```

Because the `diff` command is missing by default on `tumbleweed`, the shell returns exit code 127 (command not found). Lima interprets any non-zero exit code as "boot scripts have not finished," leading to an infinite retry loop in the host agent.

```
lima-microos:~ # cat /etc/*release
NAME="openSUSE Tumbleweed"
# VERSION="20260226"
ID="opensuse-tumbleweed"
ID_LIKE="opensuse suse"
VERSION_ID="20260226"
PRETTY_NAME="openSUSE Tumbleweed"
ANSI_COLOR="0;32"
# CPE 2.3 format, boo#1217921
CPE_NAME="cpe:2.3:o:opensuse:tumbleweed:20260226:*:*:*:*:*:*:*"
#CPE 2.2 format
#CPE_NAME="cpe:/o:opensuse:tumbleweed:20260226"
BUG_REPORT_URL="https://bugzilla.opensuse.org"
SUPPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org"
DOCUMENTATION_URL="https://en.opensuse.org/Portal:Tumbleweed"
LOGO="distributor-logo-Tumbleweed"
```
```
lima-microos:~ # diff
-bash: diff: command not found
```

Captured from `limactl` with debug logging enabled:

```
DEBU[0220] [hostagent] stdout="", stderr="+ timeout 30s bash -c 'until sudo diff -q /run/lima-ssh-ready /mnt/lima-cidata/meta-data 2>/dev/null; do sleep 3; done'\n+ echo 'not ready to start persistent ssh session'\nnot ready to start persistent ssh session\n+ exit 1\n", err=failed to execute script "user session is ready for ssh": stdout="", stderr="+ timeout 30s bash -c 'until sudo diff -q /run/lima-ssh-ready /mnt/lima-cidata/meta-data 2>/dev/null; do sleep 3; done'\n+ echo 'not ready to start persistent ssh session'\nnot ready to start persistent ssh session\n+ exit 1\n": exit status 1
```

Installing `diffutils` resolves the issue.

### Steps to Reproduce
1. Use a minimal opensuse-tumbleweed]([experimental/opensuse-tumbleweed](https://github.com/lima-vm/lima/tree/master/templates/experimental/opensuse-tumbleweed.yaml))
2. Run `limactl start --debug  tumbleweed.yaml`.
3. Observe the hang at the final boot script requirement.
4. Run `limactl shell <vm>` which diff to confirm the binary is missing.

### Suggested Fixes

**Workaround used and that worked:** Install `diffutils` during the provision step.

- **Fallback Mechanism:** Consider falling back to `sha256sum` or `cmp` (if available) if `diff` is not found, as coreutils is more likely to be present in minimal images than diffutils.
- **Better Error Reporting:** If the SSH check returns 127, the host agent could explicitly log that `diff` is missing in the guest rather than treating it as a "booting" state.
- **Template Documentation:** Update the openSUSE templates/examples to ensure `diffutils` is included in the provisioning script for minimal images.

jandubois:
The boot-done and ssh-ready signal files previously copied the entire meta-data file and used `diff -q` to verify completion. This failed on distributions like openSUSE Tumbleweed where `diffutils` is not installed.

Write just the instance ID to the signal files and compare it directly using `cat`, which removes the dependency on `diff` and `sudo`.

Fixes #4659

```console
❯ limactl start -y template:experimental/opensuse-tumbleweed
...
INFO[0078] [hostagent] The final requirement 1 of 1 is satisfied
INFO[0079] READY. Run `limactl shell opensuse-tumbleweed` to open the shell.
```

### DIFF
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -381,24 +381,26 @@ func GenerateCloudConfig(ctx context.Context, instDir, name string, instConfig *
 	return os.WriteFile(filepath.Join(instDir, filenames.CloudConfig), config, 0o444)
 }
 
-func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort int, guestAgentBinary, nerdctlArchive string, vsockPort int, virtioPort string, noCloudInit, rosettaEnabled, rosettaBinFmt bool) error {
+// GenerateISO9660 generates the cidata ISO9660 image (or directory, for noCloudInit)
+// in instDir. It returns the instance ID, which changes on every boot.
+func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort int, guestAgentBinary, nerdctlArchive string, vsockPort int, virtioPort string, noCloudInit, rosettaEnabled, rosettaBinFmt bool) (string, error) {
 	args, err := templateArgs(ctx, true, instDir, name, instConfig, udpDNSLocalPort, tcpDNSLocalPort, vsockPort, virtioPort, noCloudInit, rosettaEnabled, rosettaBinFmt)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	if err := ValidateTemplateArgs(args); err != nil {
-		return err
+		return "", err
 	}
 
 	layout, err := ExecuteTemplateCIDataISO(args)
 	if err != nil {
-		return err
+		return "", err
 	}
 
 	driverScripts, err := drv.BootScripts()
 	if err != nil {
-		return fmt.Errorf("failed to get boot scripts: %w", err)
+		return "", fmt.Errorf("failed to get boot scripts: %w", err)
 	}
 
 	for filename, content := range driverScripts {
@@ -406,7 +408,7 @@ func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name strin
 		if strings.Contains(filename, "/") {
 			// When the filename contains a slash, it must be in the format of "boot.<OS>/<SCRIPT>"
 			if !strings.HasPrefix(filename, "boot.") || strings.Count(filename, "/") != 1 {
-				return fmt.Errorf("invalid boot script filename %q: must be in format 'boot.<OS>/<SCRIPT>'", filename)
+				return "", fmt.Errorf("invalid boot script filename %q: must be in format 'boot.<OS>/<SCRIPT>'", filename)
 			}
 			layoutPath = filename
 		} else {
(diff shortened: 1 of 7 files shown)


pkg/hostagent/requirements.go
```diff
@@ -288,28 +288,22 @@ func (a *HostAgent) finalRequirements() []requirement {
 	req = append(req,
 		requirement{
 			description: "boot scripts must have finished",
-			script: `#!/bin/sh
+			script: fmt.Sprintf(`#!/bin/sh
 set -eux
 timeout=timeout
-sudo=sudo
 A=/run/lima-boot-done
```
AkihiroSuda: A could be named something better now

pkg/cidata/cidata.go
```diff
@@ -381,32 +381,32 @@ func GenerateCloudConfig(ctx context.Context, instDir, name string, instConfig *
 	return os.WriteFile(filepath.Join(instDir, filenames.CloudConfig), config, 0o444)
 }
 
-func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort int, guestAgentBinary, nerdctlArchive string, vsockPort int, virtioPort string, noCloudInit, rosettaEnabled, rosettaBinFmt bool) error {
+func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name string, instConfig *limatype.LimaYAML, udpDNSLocalPort, tcpDNSLocalPort int, guestAgentBinary, nerdctlArchive string, vsockPort int, virtioPort string, noCloudInit, rosettaEnabled, rosettaBinFmt bool) (string, error) {
```
AkihiroSuda, line 386: The ret val should be explained in godoc comment

AkihiroSuda (approved):
Thanks

pkg/hostagent/requirements.go
jandubois: fixed

pkg/cidata/cidata.go
jandubois: fixed

jandubois:
I have no idea how this PR got merged. I force-pushed the requested changes, and added the "fixed" comments, and then GitHub told me that I had merged it. Which should not even be possible because the force-push had dismissed the previous approval.

PR #4680, opened 2026-03-08
Replace diff with sha256sum in boot scripts

jandubois:
The `diff` command is missing on some distributions like openSUSE Tumbleweed. Use `sha256sum` (available in `coreutils` and `busybox`) to compare files instead.

See #4659

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/06-enable-mdns-on-systemd.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/06-enable-mdns-on-systemd.sh
@@ -20,7 +20,7 @@ enable_mdns_conf_content="[Resolve]
 MulticastDNS=yes
 "
 # Create /etc/systemd/resolved.conf.d/00-lima-enable-mdns.conf if its content is different
-if ! diff -q <(echo "${enable_mdns_conf_content}") "${enable_mdns_conf_path}" >/dev/null 2>&1; then
+if [ "$(echo "${enable_mdns_conf_content}" | sha256sum)" != "$(sha256sum <"${enable_mdns_conf_path}" 2>/dev/null)" ]; then
 	mkdir -p "$(dirname "${enable_mdns_conf_path}")"
 	echo "${enable_mdns_conf_content}" >"${enable_mdns_conf_path}"
 	systemctl daemon-reload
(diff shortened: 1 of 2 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot.Linux/06-enable-mdns-on-systemd.sh
```diff
@@ -20,7 +20,7 @@ enable_mdns_conf_content="[Resolve]
 MulticastDNS=yes
 "
 # Create /etc/systemd/resolved.conf.d/00-lima-enable-mdns.conf if its content is different
-if ! diff -q <(echo "${enable_mdns_conf_content}") "${enable_mdns_conf_path}" >/dev/null 2>&1; then
+if [ "$(echo "${enable_mdns_conf_content}" | cksum)" != "$(cksum <"${enable_mdns_conf_path}" 2>/dev/null)" ]; then
```
AkihiroSuda: sha256sum is not available?
I feel cksum is just a remnant of 1970s UNIX and not expected to be used in modern programs

jandubois: `cksum` is part of GNU coreutils. If you have `ls` and `cp`, then you have `cksum`. 🤷

jandubois: That said, `sha256sum` seems to be available on both Alpine and Tumbleweed, so I can change it, you you think that is better.

jandubois: I've updated the commit. One difference is that `cksum` is part of POSIX, and `sha256sum` is not, but I don't think that matters as everything is Linux-specific anyways.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
- https://github.com/lima-vm/lima/pull/4679#issuecomment-4020970676

You might have found a bug of GitHub.
If you force push something (e.g., update the commit message) to this PR, will GitHub automatically merge it again?

jandubois:
> If you force push something (e.g., update the commit message) to this PR, will GitHub automatically merge it again?

I don't know, maybe because I also commented?

jandubois:
I can't tell; this time it didn't, so please approve again.

Last time I added "fixed" to 2 of your PR comments, and when I submitted the second one, it merged the PR even though it had already dismissed the review. Maybe it is a subtle timing thing.

AkihiroSuda (approved):

PR #4688, opened 2026-03-09
macOS guest: remove IPSW after successful installation

### LINKED ISSUE #4619: macOS guest: remove IPSW after successful installation
https://github.com/lima-vm/lima/blob/bfe1b7d92a8dd6d3a09bfb59c9e3f7be6da38135/pkg/driver/vz/vz_driver_darwin.go#L377

AkihiroSuda:
Fix #4619

### DIFF
--- a/pkg/driver/vz/vz_driver_darwin.go
+++ b/pkg/driver/vz/vz_driver_darwin.go
@@ -377,7 +377,7 @@ func (l *LimaVzDriver) CreateDisk(ctx context.Context) error {
 // - `image.ipsw`: hardlink to `image` (".ipsw" suffix is required by VZMacOSInstaller)
 // - `disk`: ASIF disk
 //
-// TODO: consider removing IPSW after successful installation.
+// After successful installation, `image.ipsw` and `image` are removed.
 func (l *LimaVzDriver) createDiskMacOSGuest(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -406,6 +406,16 @@ func (l *LimaVzDriver) createDiskMacOSGuest(ctx context.Context) error {
 		return fmt.Errorf("failed to install macOS: %w", err)
 	}
 
+	filesToClean := []string{
+		ipsw,
+		filepath.Join(l.Instance.Dir, filenames.Image),
+	}
+	for _, file := range filesToClean {
+		if err := os.RemoveAll(file); err != nil {
+			return err
```
jandubois: Do we really want to return an error here? At this point the instance has been created successfully; why would we destroy the instance just because we can't delete the temporary file? We could just log a warning instead.

Also maybe we should use `os.Remove()` because these are files, not directories. That would catch a potential typo.

I know all this is theoretical; I expect this to never fail in practice.

jandubois (approved):
Thanks, LGTM

pkg/driver/vz/vz_driver_darwin.go
AkihiroSuda: Updated to just print warnings

PR #4689, opened 2026-03-09
Remove `mntlockutil`, mv `~/.lima/_mnt/0` to `~/.lima/INST/mnt`

### LINKED ISSUE #4650: Remove `mntlockutil`, mv `~/.lima/_mnt/0` to `~/.lima/INST/mnt`
https://github.com/lima-vm/lima/blob/d3b9571ef0cda50b98298823b71ead676124c60b/pkg/guestpatch/macos/macos_darwin.go#L69-#L82

The path of `~/.lima/_mnt/0` was designed so that it can be hardcoded in `sudoers`.
Since we have removed the necessity of sudo in PR #4634 , we can just move it to `~/.lima/INST/mnt`.

The `mntlockutil` pkg is now useless and can be removed.

AkihiroSuda:
Fix #4650


pkg/guestpatch/macos/macos_darwin.go
```diff
@@ -66,21 +65,16 @@ func patchWriteGuestFiles(ctx context.Context, disk string) error {
...
-		return fmt.Errorf("failed to acquire mount slot: %w", err)
+	instDir := filepath.Dir(disk)
+	mnt := filepath.Join(instDir, filenames.MntDir)
+	if err := os.MkdirAll(mnt, 0o755); err != nil {
+		return fmt.Errorf("failed to create mount point %q: %w", mnt, err)
 	}
 	defer func() {
-		if mntReleaseErr := mntRelease(); mntReleaseErr != nil {
-			logrus.WithError(mntReleaseErr).Warnf("failed to release mount slot %q", mnt)
+		if err := os.Remove(mnt); err != nil {
```
jandubois, line 74: This will fail if the directory is not empty. Is that intentional? Otherwise:

```suggestion
		if err := os.RemoveAll(mnt); err != nil {
```

AkihiroSuda: Intentional, to avoid accidentally removing mounted files

PR #4692, opened 2026-03-09
Fix `limactl shell` to return error for stopped instances

### LINKED ISSUE #4690: `limactl shell --tty=false STOPPED_INSTANCE true` exits with status zero
This test looks wrong

https://github.com/lima-vm/lima/blob/5d9dcf670a1a1e2a0e9da76b3de9ffcf8d048cd6/hack/bats/tests/shell.bats#L8-L11

AkihiroSuda:
`limactl shell --tty=false STOPPED_INSTANCE` was silently exiting with status 0 instead of returning an error. Now it returns a descriptive error message when the instance is stopped and won't be started.

Fixes #4690

NOTE: used Claude Code

### DIFF
--- a/hack/bats/tests/shell.bats
+++ b/hack/bats/tests/shell.bats
@@ -7,12 +7,14 @@ INSTANCE=bats-dummy
 
 @test 'lima stopped lima instance' {
     # check that the "tty" flag is used, also for stdin
-    limactl shell --tty=false "$INSTANCE" true </dev/null
+    run_e -1 limactl shell --tty=false "$INSTANCE" true </dev/null
+    assert_stderr --partial "instance \\\"$INSTANCE\\\" is stopped"
 }
 
 @test 'yes | stopped lima instance' {
     # check that stdin is verified and not just crashing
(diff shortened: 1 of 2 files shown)


hack/bats/tests/shell.bats
```diff
@@ -7,10 +7,10 @@ INSTANCE=bats-dummy
 
 @test 'lima stopped lima instance' {
     # check that the "tty" flag is used, also for stdin
-    limactl shell --tty=false "$INSTANCE" true </dev/null
+    run ! limactl shell --tty=false "$INSTANCE" true </dev/null
```
jandubois: You could check for the error:

```suggestion
    run_e -1 limactl shell --tty=false "$INSTANCE" true </dev/null
    assert_error "instance \"$INSTANCE\" is stopped…"
```

AkihiroSuda: done

jandubois:
> Any reason you didn't use the `assert_error` helper function

Sorry, I just realized this is not a log message, but an error return.

PR #4696, opened 2026-03-09
fix(shell-sync): show both rsync & diff stats for `View the changed contents`

### LINKED ISSUE #4693: `limactl shell --sync`: "View the changed contents" does not show new files
```console
$ limactl shell --sync . default touch new_file
INFO[0000] Syncing host current directory(/Users/suda/test) to guest instance... 
INFO[0000] Successfully synced host current directory to guest(/home/suda.guest/Users/suda/test) instance. 
? ⚠️  Accept the changes? (added: 1, deleted: 0, modified: 0)  [Use arrows to move, type to filter]
  Yes
  No
> View the changed contents
```

"View the changed contents" does not show the `new_file`

unsuman:
Fixes #4693 

Using `diff -ruN` fails to report newly added empty files when comparing directories. A workaround here is to show both the rsync dry-run stats and the diff output:
``` 
> View the changed contents

=== RSYNC DRY-RUN STATISTICS ===
added: 0, deleted: 0, modified: 1

Transfer starting: 10 files
>f..t.... test/new_file

sent 22 bytes  received 286 bytes  3080k bytes/sec
total size is 6191  speedup is 20.10



=== DETAILED DIFF ===
```

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -74,8 +74,11 @@ func newShellCommand() *cobra.Command {
 	return shellCmd
 }
 
-// Depth of "/Users/USER" is 3.
-const rsyncMinimumSrcDirDepth = 4
+const (
+	rsyncMinimumSrcDirDepth = 4 // Depth of "/Users/USER" is 3.
+	colorGray               = "\033[0;90m"
+	colorNone               = "\033[0m"
+)
 
 func shellAction(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
@@ -422,7 +425,7 @@ func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype
 	}
 	dirForCleanup := shellescape.Quote(fmt.Sprintf("%s/", *inst.Config.User.Home) + parts[1])
 
-	rsyncBackAndCleanup := func() error {
+	rsyncBack := func() error {
 		paths := []string{
 			remoteSource,
 			hostCurrentDir,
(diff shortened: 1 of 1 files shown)


cmd/limactl/shell.go
```diff
@@ -528,6 +528,21 @@ func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype
...
+
+			// Write rsync dry-run output first
+			if stats != nil {
+				combinedOutput := fmt.Sprintf(
+					"=== RSYNC DRY-RUN STATISTICS ===\n%s\n\n%s\n\n\n=== DETAILED DIFF ===\n\n\n",
+					stats.String(),
+					stats.Output,
+				)
+
+				if _, err := pipeIn.Write([]byte(combinedOutput)); err != nil {
```
jandubois:
```suggestion
				if _, err := fmt.Fprint(pipeIn, combinedOutput); err != nil {
```

cmd/limactl/shell.go
```diff
@@ -622,6 +637,7 @@ type rsyncStats struct {
 	Added    int
 	Deleted  int
 	Modified int
+	Output   string
```
jandubois: I would call it `Raw` instead of `Output`, but it feels odd to mix parsed and raw data in a single structure. I think returning it as a separate value from the function would be cleaner.

cmd/limactl/shell.go
```diff
@@ -656,7 +672,13 @@ func getRsyncStats(ctx context.Context, source, destination string) (*rsyncStats
 	if err != nil {
 		return nil, err
 	}
-	return parseRsyncStats(string(out)), nil
+	if len(out) == 0 {
```
jandubois: Why do you want to suppress the output when there are no changes? I would prefer to see a "no changes" messages to no output.

cmd/limactl/shell.go
```diff
@@ -528,6 +528,21 @@ func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype
 			if err := lessCmd.Start(); err != nil {
 				return fmt.Errorf("failed to start less: %w", err)
 			}
+
+			// Write rsync dry-run output first
+			if stats != nil {
+				combinedOutput := fmt.Sprintf(
+					"=== RSYNC DRY-RUN STATISTICS ===\n%s\n\n%s\n\n\n=== DETAILED DIFF ===\n\n\n",
```
jandubois: This may be personal preference, but I find this too much yelling. I think more subdued headers would look better, e.g.

```suggestion
					"--- rsync dry-run statistics ---\n%s\n\n%s\n\n\n--- detailed diff ---\n\n\n",
```

jandubois (dismissed):
Sorry, but I really only want to see "No changes", without the whole rsync dry-run and empty diff output.

I thought the command should provide feedback to the user, but I think it should still be special cased to not show

```
--- rsync dry-run statistics ---

No changes

--- detailed diff ---
```

Otherwise the PR is fine, but I would prefer this to still look a little nicer.

unsuman:
I think if there are no changes, it's better not show the interactive prompt at all.

unsuman:
Also made the headers grey 
Screenshot 2026-03-12 at 12 01 57 PM

PR #4704, opened 2026-03-11
Warn users that support for non-Linux guests is experimental

### LINKED ISSUE #4703: Warn("Support for non-Linux guests is experimental")
> limactl create should just print the notice about the experimental status when y.OS != Linux 

 _Originally posted by @AkihiroSuda in [#4548](https://github.com/lima-vm/lima/issues/4548#issuecomment-4037857992)_

unsuman:
Fixes #4703

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -325,6 +325,12 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string, createOnly bool) (*
 	if err := tmpl.Embed(cmd.Context(), true, true); err != nil {
 		return nil, err
 	}
+	if err := tmpl.Unmarshal(); err != nil {
+		return nil, err
+	}
+	if tmpl.Config != nil && tmpl.Config.OS != nil && *tmpl.Config.OS != limatype.LINUX {
+		logrus.Warn("Support for non-Linux guests is experimental")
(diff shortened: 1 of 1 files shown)


cmd/limactl/start.go
```diff
@@ -325,6 +325,9 @@ func loadOrCreateInstance(cmd *cobra.Command, args []string, createOnly bool) (*
 	if err := tmpl.Embed(cmd.Context(), true, true); err != nil {
 		return nil, err
 	}
+	if tmpl.Config != nil && tmpl.Config.OS != nil && *tmpl.Config.OS != limatype.LINUX {
```
jandubois, line 331: You need to call `tmpl.Unmarshal()` to make sure `tmpl.Config` is not `nil`:

```suggestion
	if err := tmpl.Unmarshal(); err != nil {
		return nil, err
	}
	if tmpl.Config.OS != nil && *tmpl.Config.OS != limatype.LINUX {
```

AkihiroSuda (approved):
Thanks

jandubois (approved):
Thanks, LGTM

The `tmpl.Config != nil` check should not be necessary after you call `tmpl.Marshal()` without error, but it doesn't hurt.

PR #4723, opened 2026-03-14
fix(copytool): avoid mutating receiver when opts override is passed to Command()

### LINKED ISSUE #4718: Fix permanent receiver mutation in `copyTool.Command()`
`t.Options = opts` replaces the tool's options for all future calls, mutating shared state permanently.

Fix: 
Use a local variable instead of mutating the receiver:

```
activeOpts := t.Options
if opts != nil {
    activeOpts = opts
}
```

xvchris:
Fixes #4718

`Command()` in both `rsyncTool` and `scpTool` was assigning `t.Options = opts` when a non-nil `opts` was passed. This permanently mutated the tool's stored options, so every subsequent call would use the overridden options rather than the ones set at initialisation.

Use a local variable `activeOpts` instead, which falls back to `t.Options` when no override is provided and takes the caller-supplied value otherwise.


xvchris:
The CI failure in the **Lints** job is unrelated to this PR — lychee (link checker) got a transient TLS handshake error when accessing `https://lima-vm.io/docs/`. The URL is reachable now (HTTP 200). Could a maintainer re-run the failed job? Thanks!

unsuman:
> The CI failure in the Lints job is unrelated to this PR

Yes

Please add a DCO

PR #4725, opened 2026-03-14
vmType: default to QEMU for non-native architectures

### LINKED ISSUE #4710: vmType: default to QEMU for non-native architectures
```console
$ limactl start --name foo --arch x86_64 template:alpine
? Creating an instance "foo" Proceed with the current configuration
FATA[0000] failed to resolve vm for "/Users/suda/.lima/foo/lima.yaml": unsupported arch: "x86_64
```

This error is confusing, it should automatically fallback to QEMU, or, at least it should print a hint to suggest specifying `--vm-type=qemu`

AkihiroSuda:
Fix #4710

### DIFF
--- a/pkg/limatype/lima_yaml.go
+++ b/pkg/limatype/lima_yaml.go
@@ -415,3 +415,7 @@ func DefaultDriver() VMType {
 		return QEMU
 	}
 }
+
+func DefaultNonNativeArchDriver() VMType {
+	return QEMU
+}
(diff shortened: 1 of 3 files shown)


pkg/limatype/lima_yaml.go
```diff
@@ -405,3 +415,7 @@ func DefaultDriver() VMType {
 		return QEMU
```
adam-at-mobomo, line 415: Shouldn't this also be changed to use `DefaultNonNativeArchDriver()`?

AkihiroSuda: No

PR #4727, opened 2026-03-14
fix(copytool): avoid mutating receiver when opts override is passed to Command()

### LINKED ISSUE #4718: Fix permanent receiver mutation in `copyTool.Command()`

xvchris:
Fixes #4718

`Command()` in both `rsyncTool` and `scpTool` assigned `t.Options = opts` when a non-nil `opts` was passed, permanently mutating the receiver's stored options. Every subsequent call would then use the overridden options rather than the ones set at initialisation.

Use a local variable `effectiveOpts` instead of overwriting the receiver field. Also update the `Command` interface comment to accurately describe the non-mutating behaviour.

Signed-off-by: xvchris <chrisjh@foxmail.com>

### DIFF
--- a/pkg/copytool/copytool_test.go
+++ b/pkg/copytool/copytool_test.go
@@ -1,0 +1,27 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package copytool
+
+import (
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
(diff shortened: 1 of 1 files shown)


jandubois (dismissed):
Thanks, LGTM

Fix looks fine, but also adding a test would be better. 😄

PR #4730, opened 2026-03-15
templates/README.md: clean up

AkihiroSuda:


AkihiroSuda:
> I thought the non-Linux section would get a note about their experimental status.

done

PR #4740, opened 2026-03-16
docs: document kernel and initrd in templates/default.yaml

### LINKED ISSUE #4728: `kernel` and `initrd` are not documented in `default.yaml`
`kernel` and `initrd` (https://pkg.go.dev/github.com/lima-vm/lima/v2/pkg/limatype#Image) are not documented in https://github.com/lima-vm/lima/blob/master/templates/default.yaml

qeinstein:
Fixes #4728

### DIFF
--- a/templates/default.yaml
+++ b/templates/default.yaml
@@ -15,8 +15,20 @@ vmType: null
 # 🟢 Builtin default: "default" (corresponds to the host architecture)
 arch: null
 
+
 # OpenStack-compatible disk image.
 # Each image has a `location` URL for the disk image, an `arch` setting, and an optional `digest`.
+# To specify a custom kernel and initial RAM disk, nest `kernel` and `initrd` under an image entry:
+#   - location: "https://example.com/image.qcow2"
+#     # kernel specifies a custom kernel to boot.
+#     kernel:
+#       location: "https://example.com/vmlinuz"
+#       digest: "sha256:..."
+#       cmdline: "console=ttyS0 root=/dev/sda1"
+#     # initrd specifies a custom initial RAM disk.
+#     initrd:
+#       location: "https://example.com/initrd.img"
+#       digest: "sha256:..."
 # 🟢 Builtin default: none (must be specified)
 # 🔵 This file: Ubuntu images (inherited via the `base` mechanism later in this file)
 images: []


templates/default.yaml
```diff
@@ -335,8 +348,8 @@ minimumLimaVersion: 2.0.0
 # 🟢 Builtin default: no base template
 # 🔵 This file: Ubuntu images and default mount points
 base:
-- template:_images/ubuntu
-- template:_default/mounts
+- template: _images/ubuntu
+- template: _default/mounts
```
AkihiroSuda: ?

templates/default.yaml
```diff
@@ -468,7 +481,6 @@ networks:
 # configured in socket_vmnet and not in lima.
 # - socket: "/var/run/socket_vmnet"
 
-
```
AkihiroSuda, line 471: ?

templates/default.yaml
qeinstein: Apologies for the noise. My editor’s auto-indentation shifted this block accidentally while I was adding the documentation. I have reverted it to its original state.

templates/default.yaml
qeinstein: This was a stray character from a bad save. I have cleaned up the file and verified it with yamllint to ensure the structure is correct now. Gracias

templates/default.yaml
```diff
@@ -19,6 +19,15 @@ arch: null
 # Each image has a `location` URL for the disk image, an `arch` setting, and an optional `digest`.
 # 🟢 Builtin default: none (must be specified)
 # 🔵 This file: Ubuntu images (inherited via the `base` mechanism later in this file)
+# kernel specifies a custom kernel to boot.
```
AkihiroSuda: This could be misinterpreted as "Ubuntu images (...) kernel specifies a custom kernel to boot."

templates/default.yaml
```diff
@@ -19,6 +19,15 @@ arch: null
 # Each image has a `location` URL for the disk image, an `arch` setting, and an optional `digest`.
```
AkihiroSuda, line 20: kernel and initrd should go here

templates/default.yaml
qeinstein: Yes, that makes sense
I have made the required changes

templates/default.yaml
```diff
@@ -15,6 +15,17 @@ vmType: null
 # 🟢 Builtin default: "default" (corresponds to the host architecture)
 arch: null
 
+# kernel specifies a custom kernel to boot.
+# kernel:
```
AkihiroSuda: kernel is not a top-level field.
Why did you move there?

PR #4742, opened 2026-03-17
website: document URL handler plugins

### LINKED ISSUE #4299: Document `github:` and custom URL schemes
Document for:
- https://github.com/lima-vm/lima/pull/3937
- https://github.com/lima-vm/lima/pull/4134

jandubois:
I've mostly copied together various examples I used in issue comments. I did leave out the `limactl-diff` example because that's mostly an example plugin that only makes use of the `instance:` scheme.

I wonder if the "composing schemes" at the bottom should be removed; they are getting a bit too confusing maybe? On the other hand they show some more complex things you can do.

---

Add documentation for custom URL scheme plugins, covering:
- How the plugin mechanism works (naming, discovery, invocation)
- Returning URLs, file paths, or generated YAML
- Composing schemes (latest release, multi-repo search)

Examples include dev/prod branch selection, S3 pre-signed URLs, reusing an existing instance's template, and dynamic YAML generation.

Closes #4299

### DIFF
--- a/website/content/en/docs/config/plugin/url.md
+++ b/website/content/en/docs/config/plugin/url.md
@@ -1,0 +1,169 @@
+---
+title: URL handler plugins
+weight: 3
+---
+
+> **Warning**
+> Support for URL handler plugins is experimental
+
+| ⚡ Requirement | Lima >= 2.0 |
+|----------------|-------------|
+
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

Issue #4755, opened 2026-03-18
VZ driver: missing auto-unlock for disks locked to the same instance (parity with QEMU driver)


lukasfroehlich1:
### Description

The QEMU driver auto-unlocks a disk when it's locked to the same instance being started (added in #2092 for #1683). The VZ driver is missing this fix, causing startup failures after unclean shutdowns (crash, macOS reboot, `pkill -9`).

**Error** (from `ha.stderr.log`):
```
{"level":"fatal","msg":"failed to run attach disk \"testdisk\", in use by instance \"testvm\""}
```

`limactl start` surfaces this as:
```
FATA[0002] exiting, status={Running:false Degraded:false Exiting:true Errors:[] SSHLocalPort:0} (hint: see "~/.lima/testvm/ha.stderr.log")
FATA[0002] error starting vm: error at 'starting': exit status 1
```

**Code reference**:

The QEMU driver (`pkg/driver/qemu/qemu.go:705-714`) correctly handles this:
```go
if disk.Instance != "" {
    if disk.InstanceDir != cfg.InstanceDir {
        // Different instance -> real conflict, error out
        return ...
    }
    // Same instance -> stale lock, auto-unlock
    err = disk.Unlock()
}
```

The VZ driver (`pkg/driver/vz/vm_darwin.go:591-592`) does not:
```go
if disk.Instance != "" {
    return fmt.Errorf("failed to run attach disk %q, in use by instance %q", ...)
}
```

**Workaround**: Manually remove `~/.lima/_disks/<name>/in_use_by` before starting.

**Expected fix**: Apply the same same-instance unlock logic from the QEMU driver to the VZ driver.

**Environment**:
- Lima: v1.2.1 (installed via mise)
- macOS: 15.5 (Darwin 25.3.0)
- Architecture: Apple Silicon (aarch64)
- vmType: vz

**Repro steps** (verified on the environment above):

1. Create a lima.yaml with VZ + additional disk:
   ```yaml
   # /tmp/testvm.yaml
   vmType: vz
   images:
     - location: "https://cloud-images.ubuntu.com/releases/24.04/release/ubuntu-24.04-server-cloudimg-arm64.img"
       arch: "aarch64"
   cpus: 2
   memory: "2GiB"
   disk: "10GiB"
   additionalDisks:
     - name: testdisk
   ```

2. Create the disk and start the instance:
   ```
   limactl disk create testdisk --size 1G
   limactl start --name testvm /tmp/testvm.yaml
[...]

PR #2092, opened 2023-12-20
Automatically unlock disk if it is already locked to the same instance (after restarting macOS)

porfirion:
This PR closes #1683 

Sometimes disk is not unlocked while shutdown (for example on macOS restart) and existing link prevents restarting vm without additional commands.
Automatically remove existing link when using with the same vm.

Issue #1683, opened 2023-07-26
After restarting macOS, additionalDisks of the VM failed to mount (`could not attach disk "test", in use by instance "default" `)


JesseChain:
### Steps to Reproduce
1. `limactl disk create test --size 10GB`
2. `limactl start --name default --set='.additionalDisks[0] = "test"' template://ubuntu-lts`
3. Restart the macOS operating system.
4. `limactl start default`
```
~ % limactl start default
INFO[0000] Using the existing instance "default"        
INFO[0000] Attempting to download the nerdctl archive    arch=aarch64 digest="sha256:589dabd962d936b29fd377dcddbb49c07d1c4c27dd4b402bc4b6b20287fe9c37" location="https://github.com/containerd/nerdctl/releases/download/v1.4.0/nerdctl-full-1.4.0-linux-arm64.tar.gz"
INFO[0000] Using cache "/Users/jesse/Library/Caches/lima/download/by-url-sha256/8704652b907393aea54c8d98b48efa0fe9bef4d120fbbf8799bb5264631da634/data" 
ERRO[0002] [hostagent] could not attach disk "test", in use by instance "default" 
INFO[0002] [hostagent] Starting QEMU (hint: to watch the boot progress, see "/Users/jesse/.lima/default/serial.log") 
FATA[0002] exiting, status={Running:false Degraded:false Exiting:true Errors:[] SSHLocalPort:0} (hint: see "/Users/jesse/.lima/default/ha.stderr.log")
```

### Environment Information
- Operating System: macOS 13.4.1
- Lima Version: 0.16.0

AkihiroSuda:
`limactl disk unlock DISKNAME` may work

dnt1996:
@AkihiroSuda Thanks, it works, but this issue is weird, the instance can't mount an additional disk that is occupied by itself, maybe we should keep this issue open until it is resolved.

PR #4756, opened 2026-03-18
vz,qemu: auto-unlock disk locked to the same instance on startup

lukasfroehlich1:
After an unclean shutdown (crash, macOS reboot, `pkill -9`), the `in_use_by` symlink for additional disks persists. The QEMU driver already handled this by checking if the disk was locked to the same instance and auto-unlocking it (added in #2092 for #1683), but the VZ driver did not, causing startup failures with:

```
failed to run attach disk "X", in use by instance "X"
```

This PR extracts the check-unlock-lock logic into `Disk.LockForInstance()` and uses it from both drivers.

Fixes https://github.com/lima-vm/lima/issues/4755.

### DIFF
--- a/pkg/driver/qemu/qemu.go
+++ b/pkg/driver/qemu/qemu.go
@@ -702,21 +702,9 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 			return "", nil, err
 		}
 
-		if disk.Instance != "" {
-			if disk.InstanceDir != cfg.InstanceDir {
-				logrus.Errorf("could not attach disk %q, in use by instance %q", diskName, disk.Instance)
-				return "", nil, err
-			}
-			err = disk.Unlock()
-			if err != nil {
-				logrus.Errorf("could not unlock disk %q to reuse in the same instance %q", diskName, cfg.Name)
-				return "", nil, err
-			}
-		}
 		logrus.Infof("Mounting disk %q on %q", diskName, disk.MountPoint)
-		err = disk.Lock(cfg.InstanceDir)
-		if err != nil {
-			logrus.Errorf("could not lock disk %q: %q", diskName, err)
+		if err = disk.LockForInstance(cfg.InstanceDir); err != nil {
+			logrus.Errorf("could not attach disk %q: %s", diskName, err)
 			return "", nil, err
 		}
 		dataDisk := filepath.Join(disk.Dir, filenames.DataDisk)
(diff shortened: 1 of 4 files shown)


pkg/driver/qemu/qemu.go
```diff
@@ -702,21 +702,9 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
...
-				logrus.Errorf("could not unlock disk %q to reuse in the same instance %q", diskName, cfg.Name)
-				return "", nil, err
-			}
-		}
 		logrus.Infof("Mounting disk %q on %q", diskName, disk.MountPoint)
-		err = disk.Lock(cfg.InstanceDir)
-		if err != nil {
-			logrus.Errorf("could not lock disk %q: %q", diskName, err)
+		if err = disk.LockForInstance(cfg.InstanceDir); err != nil {
+			logrus.Errorf("could not attach disk %q: %q", diskName, err)
```
jandubois: We typically don't quote error messages, just user-data


```suggestion
			logrus.Errorf("could not attach disk %q: %s", diskName, err)
```

pkg/store/disk.go
```diff
@@ -97,3 +97,15 @@ func (d *Disk) Unlock() error {
...
 }
+
+func (d *Disk) LockForInstance(instanceDir string) error {
+	if d.Instance != "" {
+		if d.InstanceDir != instanceDir {
+			return fmt.Errorf("disk %q is in use by instance %q", d.Name, d.Instance)
+		}
+		if err := d.Unlock(); err != nil {
+			return fmt.Errorf("failed to unlock disk %q for reuse in the same instance: %w", d.Name, err)
+		}
```
jandubois, line 108: The disk name is already reported by the caller:

```go
return fmt.Errorf("failed to attach disk %q: %w", diskName, err)
```

No need to repeat it here


```suggestion
		if d.InstanceDir != instanceDir {
			return fmt.Errorf("in use by instance %q", d.Instance)
		}
		if err := d.Unlock(); err != nil {
			return fmt.Errorf("failed to unlock for reuse in the same instance: %w", err)
		}
```

jandubois (approved):
Thanks, LGTM

I've only found very minor nit-picks; not sure if you want to change or not.

Also potentially missing a test case for `LockForInstance` when the symlink cannot be removed, but I don't know how this would even happen, so probably not worth it.

pkg/store/disk.go
lukasfroehlich1: Done.

pkg/driver/qemu/qemu.go
lukasfroehlich1: Done.

jandubois (approved):
Thanks, LGTM

PR #4763, opened 2026-03-22
Fix vz audio device=none still attaches virtio sound devices

### LINKED ISSUE #4761: vz: audio.device=none still attaches virtio sound devices
### Description

When using Lima with `vmType: vz`, setting:

```yaml
audio:
  device: none
```

does not appear to remove guest audio devices.

### Expected behavior

No virtual audio device should be attached to the guest when audio.device: none is set.

### Actual behavior

The VZ path accepts audio.device: none, and one helper returns early without attaching audio, but attachOtherDevices() still unconditionally adds virtio sound input/output devices.

Relevant code:

- pkg/driver/vz/vz_driver_darwin.go
- pkg/driver/vz/vm_darwin.go

### Why this matters

For hardening / reduced attack surface, muting audio is not sufficient. If the virtual sound device is still presented to the guest, the guest kernel can still enumerate it, and a
privileged attacker could still try to access it by loading a driver.

### Suggested fix

Conditionally skip VZ sound device attachment entirely when:

- audio.device: none
- possibly also when audio.device is empty, depending on intended default behavior

### Environment

- Lima version: 2.1.0
- macOS version: 15.7.4 (build 24G517)
- Host arch: arm64

qianchongyang:
## Problem

Setting `audio.device: none` in Lima with `vmType: vz` doesn't prevent virtio sound devices from being attached. The `attachOtherDevices()` function unconditionally adds sound devices even when audio is disabled, creating a security attack surface.

## Solution

Modified `attachOtherDevices()` to conditionally skip VZ sound device attachment when `audio.device: none` is configured. Sound input/output devices are now only attached when a valid audio device is specified.

## Validation

```yaml
# lima.yaml
vmType: vz
audio:
  device: none

# Guest no longer enumerates virtio sound devices
# ✅ Security hardening requirement met
```

Fixes #4761

### DIFF
--- a/pkg/driver/vz/vm_darwin.go
+++ b/pkg/driver/vz/vm_darwin.go
@@ -812,34 +812,7 @@ func attachOtherDevices(inst *limatype.Instance, vmConfig *vz.VirtualMachineConf
 		return err
 	}
 
-	// Set audio device
-	inputAudioDeviceConfig, err := vz.NewVirtioSoundDeviceConfiguration()
-	if err != nil {
-		return err
-	}
-	inputStream, err := vz.NewVirtioSoundDeviceHostInputStreamConfiguration()
-	if err != nil {
-		return err
-	}
-	inputAudioDeviceConfig.SetStreams(
-		inputStream,
-	)
-
-	outputAudioDeviceConfig, err := vz.NewVirtioSoundDeviceConfiguration()
-	if err != nil {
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
Thanks, but a duplicate:
- https://github.com/lima-vm/lima/pull/4762

PR #4772, opened 2026-03-24
pkg/networks/usernet: use domain.tld zones

### LINKED ISSUE #3161: Resolving .internal domains from inside the container
### Description

Feature request: I'd like to resolve domains like `registry.internal` from inside my VM

My understanding of lima's internals is quite blurry — it looks like `.internal` domains are a special case in the hostagent resolver and are never forwarded to the VMs configured nameserver.

julego:
The usernet client stores dns records in tld zones, and as such the default host.lima.internal record breaks resolving .internal domains that may exist on the host network (e.g. compute.internal, dev.internal, etc).

This is a proposal to store dns records in domain.tld zones to be able to use such domains.

It may close #3161 also.

### DIFF
--- a/pkg/networks/usernet/dnshosts/dnshosts.go
+++ b/pkg/networks/usernet/dnshosts/dnshosts.go
@@ -95,7 +95,7 @@ func (z zoneHost) name() string {
 	if i < 0 {
 		return string(z)
 	}
-	return string(z)[i+1:] + "."
+	return string(z)[z.zoneIndex()+1:] + "."
 }
 
 func (z zoneHost) recordName() string {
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
CI failing
```
ping: lima-net-user-v2-1.internal: Name or service not known
```

https://github.com/lima-vm/lima/actions/runs/23485110542/job/69213743913?pr=4772

julego:
I just checked the test, it failed because the dns server implemented in gvisor-tap-vsock doesn't seem to handle domain.tld zones so well.

With domain.tld zones, sublevel records such as project.dev.tld are well resolved by gvisor-tap-vsock (what I was testing on my dev env) but root records like dev.tld are not (whether the tld is ".internal" or not). Hence the test failing with "lima-net-user-v2-1.internal".

I'll have a look at this but gvisor-tap-vsock will probably need a patch too, I don't think there's a direct fix in lima.

PR #4789, opened 2026-03-31
CI: release Windows artifacts

AkihiroSuda:


Makefile
```diff
@@ -722,6 +725,7 @@ goarchs_native_and_others = $(GOHOSTARCH) $(filter-out $(GOHOSTARCH),$(artifact_
 artifacts: $$(addprefix artifact-$$(GOOS)-,$$(goarchs_native_and_others))
 artifacts-darwin: $$(call generate_manpages_if_needed,darwin) $$(addprefix artifact-darwin-,$$(goarchs_native_and_others))
 artifacts-linux: $$(call generate_manpages_if_needed,linux) $$(addprefix artifact-linux-,$$(goarchs_native_and_others))
+# FIXME: remove shell scripts from Windows artifacts?
 artifacts-windows: $$(call generate_manpages_if_needed,windows) $$(addprefix artifact-windows-,$$(goarchs_native_and_others))
```
AkihiroSuda, line 728: Looks like we have to release the artifacts as zip to support winget in future
- https://github.com/microsoft/winget-pkgs/discussions/101705

jandubois:
Shouldn't the workflow also get an `artifacts-windows` job so the Windows artifacts are uploaded on regular CI runs?

AkihiroSuda:
> Shouldn't the workflow also get an `artifacts-windows` job so the Windows artifacts are uploaded on regular CI runs?

Not sure. We haven't uploaded Linux artifacts in regular CI so far.

PR #4790, opened 2026-03-31
containerd: default uninstall containerd

ningmingxiao:


unsuman:
Please sign the DCO check!

ningmingxiao:
I use lima in offline  host  have to set   --containerd=none  to disable containerd, I also afraid this commit  will influence some people , or we can use env or global config to set default value @unsuman

unsuman:
> I use lima in offline host have to set --containerd=none to disable containerd, I also afraid this commit will influence some people , or we can use env or global config to set default value @unsuman

Yes. I think a better approach for your usecase would be to set a global config(`$LIMA_HOME/_config/default.yaml`) or `override.yaml`:

```
containerd:
  system: false
  user: false
```

cmd/limactl/editflags/editflags.go
```diff
@@ -104,7 +104,7 @@ func RegisterCreate(cmd *cobra.Command, commentPrefix string) {
 		return []string{"x86_64", "aarch64", "riscv64", "armv7l", "s390x", "ppc64le"}, cobra.ShellCompDirectiveNoFileComp
 	})
 
-	flags.String("containerd", "", commentPrefix+"containerd mode (user, system, user+system, none)")
+	flags.String("containerd", "none", commentPrefix+"containerd mode (user, system, user+system, none)")
```
AkihiroSuda, line 107: The default configuration has to leave it unmodified

PR #4801, opened 2026-04-04
fix(shell): copy sshCmd.Args to a new slice before appending in executeSSHForRsync

### LINKED ISSUE #4717: Prevent silent slice corruption in `executeSSHForRsync`
### Description

The function takes `exec.Cmd` by value and appends to `Args`. This relies on fragile Go slice capacity semantics and risks overwriting caller data.

Fix: Explicitly copy `sshCmd.Args` to a newly allocated slice before appending

opjt:
Closes #4717

Explicitly copy `sshCmd.Args` to a newly allocated slice before appending
instead of relying on Go slice capacity semantics.


unsuman (approved):
LGTM!

PR #4807, opened 2026-04-07
CI: archlinux: set upgradePackages=true

AkihiroSuda:
FIx #4805

PR #4811, opened 2026-04-07
Fix GHA syntax error in test workflow

jandubois:
Introduced in #4807 and not caught in code review. 😢

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -335,7 +335,7 @@ jobs:
     - name: "Show cache"
       run: ./hack/debug-cache.sh
     - name: "Patch archlinux.yaml (workaround for https://github.com/lima-vm/lima/issues/4805)"
-      run: yq -i .upgradePackages=true templates/${{ matrix.template }
+      run: yq -i .upgradePackages=true templates/${{ matrix.template }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -335,7 +335,7 @@ jobs:
     - name: "Show cache"
       run: ./hack/debug-cache.sh
     - name: "Patch archlinux.yaml (workaround for https://github.com/lima-vm/lima/issues/4805)"
-      run: yq -i .upgradePackages=true templates/${{ matrix.template }
+      run: "sed -i '/^minimumLimaVersion:/a upgradePackages: true' templates/${{ matrix.template }}"
```
AkihiroSuda: Maybe just run `sudo pacman -Syu && sudo reboot` explicitly via `limactl shell` instead

jandubois:
I've reverted this PR to the original version, that just fixes the syntax error in the workflow file.

Let's merge this change, just so that all tests will run again. CI will still be broken because Archlinux is still failing, but we can address this in a separate PR. I just feel uncomfortable leaving the repo in a state where it looks like PRs are green when actually the tests are not being run at all.

PR #4817, opened 2026-04-09
Docs: Add KubeCon 2026 talk and pavilion slides

unsuman:


AkihiroSuda (approved):
Thanks

Besides these talks we should also consider collecting blog articles (CNCF, etc.)

PR #4825, opened 2026-04-10
Re-introducing USB pass-though but using YAML argument "usb" instead o…

FlUxIuS:
Compared to v1.2.3, attaching a USB device is impossible without using the `video` parameter in a YAML configuration for an instance to inject `qemu-xhci,id=usb` when running QEMU.

So we introduced a `usb: true|false` parameters that allows to do so.

### DIFF
--- a/pkg/cidata/cloud-config.yaml
+++ b/pkg/cidata/cloud-config.yaml
@@ -1,1 +1,189 @@
-cidata.TEMPLATE.d/user-data
+#cloud-config
+# vim:syntax=yaml
+
+growpart:
+  mode: auto
+  devices: ['/']
+
+{{- if eq .OS "FreeBSD" }}
+packages:
+  # boot.sh depends on sudo.
+  # TODO: consider replacing sudo with doas.
+  # FIXME: The hostagent script depends on sudo too.
+  # https://github.com/lima-vm/lima/issues/4594
+  - sudo
+{{- end }}
+
+{{- if .UpgradePackages }}
+package_update: true
+package_upgrade: true
+package_reboot_if_required: true
+{{- end }}
+
+{{- if or .RosettaEnabled (and .Mounts (or (eq .MountType "9p") (eq .MountType "virtiofs"))) }}
+mounts:
+  {{- if .RosettaEnabled }}{{/* Mount the rosetta volume before systemd-binfmt.service(8) starts */}}
+- [vz-rosetta, /mnt/lima-rosetta, virtiofs, defaults, "0", "0"]
+  {{- end }}
+  {{- if and .Mounts (or (eq .MountType "9p") (eq .MountType "virtiofs")) }}
+    {{- range $m := $.Mounts}}
+- [{{$m.Tag}}, {{$m.MountPoint}}, {{$m.Type}}, "{{$m.Options}}", "0", "0"]
+    {{- end }}
+  {{- end }}
+{{- end }}
+
+{{- if .TimeZone }}
+timezone: {{.TimeZone}}
+{{- end }}
+
+users:
+  - name: "{{.User}}"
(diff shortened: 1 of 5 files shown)


pkg/cidata/cloud-config.yaml
```diff
@@ -1 +0,0 @@
-cidata.TEMPLATE.d/user-data
```
AkihiroSuda, line 1: Should remain the symlink

pkg/driver/qemu/qemu.go
```diff
@@ -852,6 +852,9 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 		args = append(args, "-device", "virtio-keyboard-pci")
 		args = append(args, "-device", "virtio-"+input+"-pci")
 		args = append(args, "-device", "qemu-xhci,id=usb-bus")
+	} else if *y.USB {
+		// Add USB controller without display for USB device passthrough
+		args = append(args, "-device", "qemu-xhci,id=usb-bus")
```
AkihiroSuda, line 857: I guess the `qemu-xhci` device can be unconditionally added?

pkg/limayaml/default.yaml
```diff
@@ -1 +0,0 @@
-../../templates/default.yaml
```
AkihiroSuda, line 1: This should remain symlink

pkg/limatype/lima_yaml.go
```diff
@@ -35,6 +35,7 @@ type LimaYAML struct {
 	Firmware              Firmware      `yaml:"firmware,omitempty" json:"firmware,omitempty"`
 	Audio                 Audio         `yaml:"audio,omitempty" json:"audio,omitempty"`
 	Video                 Video         `yaml:"video,omitempty" json:"video,omitempty"`
+	USB                   *bool         `yaml:"usb,omitempty" json:"usb,omitempty" jsonschema:"nullable"`
```
AkihiroSuda, line 38:
```suggestion
	USB                   USB        `yaml:"usb,omitempty" json:"usb,omitempty" jsonschema:"nullable"`
```
For consistency with `Audio` and `Video`

pkg/driver/qemu/qemu.go
```diff
@@ -852,6 +852,9 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 		args = append(args, "-device", "virtio-keyboard-pci")
 		args = append(args, "-device", "virtio-"+input+"-pci")
 		args = append(args, "-device", "qemu-xhci,id=usb-bus")
+	} else if *y.USB {
+		// Add USB controller without display for USB device passthrough
```
AkihiroSuda, line 856: Where is the passthrough implementation?
How did you test it?

pkg/driver/qemu/qemu.go
FlUxIuS: Yes, maybe I'm also focusing too much on making it work on macOS Apple Silicon. So maybe the parameter `usb` should be `qemu-xhci` instead of `true`

FlUxIuS: or something else

pkg/driver/qemu/qemu.go
FlUxIuS: USB passthrough is already possible in Lima but the video argument had to be enabled. So in this commit we introduce the USB argument to granuarly enable USB passthrough without video if not needed.

AkihiroSuda: > USB passthrough is already possible in Lima

How was that possible?

AkihiroSuda: ping. 
Also let me mark this PR as a draft.

PR #4827, opened 2026-04-10
fix: LIMA_SHELLENV_ALLOW doesn't remove variables

### LINKED ISSUE #4826: LIMA_SHELLENV_ALLOW doesn't remove environment variables not in the list
### Description

[Environment Variables | Lima](https://lima-vm.io/docs/config/environment-variables/) states that `LIMA_SHELLENV_ALLOW` does the following:

> Specifies a comma-separated list of environment variable patterns to allow when propagating environment variables to the Lima instance with --preserve-env. When set, only variables matching these patterns will be passed through, completely overriding the default block list behavior. This feature only applies to Lima v2.0.0 or later.

I read this as, if `LIMA_SHELLENV_ALLOW=FOO`, the _only_ propagated environment variable would be `FOO` and no others. However, what appears to happen is that all environment variables are preserved instead.

If I read the tests, my observed behaviour appears to match the tested behaviour:

https://github.com/lima-vm/lima/blob/508d1d36156d24240fa66eb196e1b68a8be29332/pkg/envutil/envutil_test.go#L169-L177

The tested behaviour seems to be "when filtered with an allowlist, all variables are passed through" (ie, `assert.Equal(t, len(result), len(testEnv))` --- the output slice is the same as the original testenv).

I _think_ this test should be `assert.Equal(t, len(result), len(expected))`. That is, "only those environment variables in the allowlist are preserved and the others are filtered out".

Is my reading of the docs correct and this is a bug?

mikerhodes:
When `LIMA_SHELLENV_ALLOW` was set, `limactl shell --preserve-env` didn't remove environment variables that were not on the list specified by `LIMA_SHELLENV_ALLOW`. This didn't match the documented behavior.

This commit updates the code to filter out all variables not on the allowlist, if the allowlist is configured.

Fix the associated test to match the expected behavior.

Fixes https://github.com/lima-vm/lima/issues/4826.

### DIFF
--- a/pkg/envutil/envutil.go
+++ b/pkg/envutil/envutil.go
@@ -141,6 +141,12 @@ func FilterEnvironment() []string {
 
 func filterEnvironmentWithLists(env, allowList, blockList []string) []string {
 	var filtered []string
+	allowed := func(name string) bool {
+		if len(allowList) > 0 {
+			return matchesAnyPattern(name, allowList)
+		}
+		return !matchesAnyPattern(name, blockList)
+	}
 
 	for _, envVar := range env {
 		parts := strings.SplitN(envVar, "=", 2)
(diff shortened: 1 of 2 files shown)


jandubois (changes requested):
As I've commented in https://github.com/lima-vm/lima/issues/4826#issuecomment-4228384199, it is unfortunately the documentation, that is out-of-date, and not the code.

We should change the documentation to match the intent of the feature.

Sorry for wasting your time on the code change! 😞

mikerhodes:
Ach, no problem! Closing.

PR #4829, opened 2026-04-10
Update k8s.yaml for bootstrapToken configuration in kubeadm

techthumb:
Joining an existing cluster currently fails due to an error with the join command:

```
error: can not mix '--config' with arguments [discovery-token-ca-cert-hash token]
```

### DIFF
--- a/templates/k8s.yaml
+++ b/templates/k8s.yaml
@@ -122,9 +122,14 @@ provision:
     apiVersion: kubeadm.k8s.io/v1beta4
     nodeRegistration:
       criSocket: unix:///run/containerd/containerd.sock
+    discovery:
+      bootstrapToken:
+        apiServerEndpoint: "{{.Param.url}}"
+        token: "{{.Param.token}}"
+        caCertHashes:
+        - "{{.Param.discoveryTokenCaCertHash}}"
     EOF
-    kubeadm join --config kubeadm-config.yaml {{.Param.url}} --token {{.Param.token}} \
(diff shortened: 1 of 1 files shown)


afbjorklund (approved):
LGTM

AkihiroSuda:
Why couldn't we catch this in CI?
https://github.com/lima-vm/lima/blob/d67fd0c8db94d232cc9f463033b69198c52be5da/.github/workflows/test.yml#L387

techthumb:
I believe the 'Multi-node' test is incorrect.

I deployed the YAML from the test and have all 3 pods running on a single node k8s.

```
$ kubectl get pods --output=wide
NAME        READY   STATUS    RESTARTS   AGE   IP           NODE         NOMINATED NODE   READINESS GATES
dnstest-0   1/1     Running   0          40s   10.244.0.4   lima-k8s-0   <none>           <none>
dnstest-1   1/1     Running   0          37s   10.244.0.5   lima-k8s-0   <none>           <none>
dnstest-2   1/1     Running   0          36s   10.244.0.6   lima-k8s-0   <none>           <none>
```

It looks more like a 'Multi-pod' rather than 'Multi-node' test.

I suggest that the test should really be:

```bash
[ $(kubectl get nodes -o jsonpath='{range .items[*]}{.status.conditions[?(@.type=="Ready")].status}{"\n"}{end}' | grep -c True) -ge 2 ]
```

techthumb:
More fixes at https://github.com/lima-vm/lima/pull/4835

PR #4831, opened 2026-04-11
Fix race condition in 04-persistent-data-volume.sh

### LINKED ISSUE #4830: Race condition in 04-persistent-data-volume.sh causes mount failure and potential data loss
The analysis in this issue has been performed by Opus 4.6.

## Summary

`04-persistent-data-volume.sh` relies on the udev symlink `/dev/disk/by-label/data-volume` at several points. A race between `growpart`/`e2fsck` and udev's device probing can destroy this symlink mid-script, causing mount failures. Under worse timing, the symlink never exists at all, and the script's `else` branch reformats a disk it misidentifies as unused -- destroying the data volume.

## Environment

- Lima with Alpine Linux (ramdisk root, eudev)
- macOS host, Apple Silicon, VZ hypervisor
- Observed on Rancher Desktop 1.22.0 (lima-vm bundled version)

### Normal boot (symlink exists, mount succeeds by narrow margin)

The following dmesg timeline is from a machine where the boot **succeeds** -- but only barely:

```
 0.74s  virtio_blk: [vda] detected, partition table read → vda1
 0.81s  EXT4-fs (vda1): recovery required on readonly filesystem
 1.58s  EXT4-fs (vda1): recovery complete, mounted ro, then unmounted
 6.39s  udevd starts (eudev-3.2.14)
        → initial coldplug probe succeeds → /dev/disk/by-label/data-volume created
10.59s  "vda: vda1" — kernel re-reads partition table (triggered by growpart)
11.68s  udevd: incorrect ext4 checksum on /dev/vda1    ← probe fails
11.69s  EXT4-fs (vda1): mounted filesystem r/w          ← mount succeeds (just barely)
```

**What happens:**

1. `growpart /dev/vda 1` runs `sfdisk`, which triggers a kernel partition table re-read. The kernel generates a new uevent for `vda1`.
 
2. `e2fsck -f -p /dev/disk/by-label/data-volume` opens the device and forces a full filesystem check, modifying the superblock.
 
3. udevd processes the growpart uevent **concurrently** with e2fsck. Its libblkid probe reads a partially modified ext4 superblock, fails with "incorrect ext4 checksum," drops `ID_FS_LABEL`, and **deletes the `/dev/disk/by-label/data-volume` symlink**.
 
4. The `mount` syscall resolves the symlink to `/dev/vda1` milliseconds before udevd deletes it. The mount succeeds, but the symlink is gone for the rest of the session.**Confirmed on a live VM:** after a normal boot, `ls /dev/disk/by-label/` shows only`alpine-lima` and `cidata` -- no `data-volume`. Running `udevadm trigger` for `vda1` restores the symlink. `udevadm info /dev/vda1` confirms `ID_FS_LABEL` is missing until re-triggered.

### Failed boot (symlink absent or deleted before mount)

Reported in rancher-sandbox/rancher-desktop#10133. The user observed two distinct failures:

**Failure 1 (growpart present in log):** The initial udev probe succeeds and creates the symlink. The script enters the `if` branch. `growpart` triggers a re-probe; uevd deletes the symlink. If the `mount` syscall runs after the deletion (the opposite of the "just barely" case above), it fails:

```
mount: mounting /dev/disk/by-label/data-volume on /mnt/data failed: No such file or directory
```

**Failure 2 (no growpart in log):** The initial udev probe also fails (ext4 recovery at 0.8s may overlap with the coldplug probe). The symlink never exists. The script enters the `else` branch.

## The else branch: silent data loss

When the `else` branch runs on a VM that already has a data volume, two additional bugs compound the problem:

### Bug 1: Unsafe "disk in use" check

The else branch determines whether a disk is in use by scanning `/proc/mounts`:

```bash
for PART in $(awk '/^\/dev\// {gsub("/dev/", ""); print $1}' /proc/mounts); do
    if [ "${DISK}" == "${PART}" ] || [ -e /sys/block/"${DISK}"/"${PART}" ]; then
        IN_USE=true
        break
    fi
done
```

On a ramdisk Alpine boot, the root is tmpfs. The data disk (`vda`) has no mounted partitions yet. The only `/dev/` entries in `/proc/mounts` are unrelated devices (the boot ISO, cidata). The script concludes `vda` is unused and runs:

```bash
echo 'type=83' | sfdisk --label dos /dev/"${DISK}"    # destroys existing partition table
mkfs.ext4 -L data-volume /dev/"${PART}"                # reformats the partition
```

**This destroys all data on the volume** -- Docker images, container state, persisted `/etc`, `/var/lib`, `/home`, and `/root`.

### Bug 2: Mount after mkfs uses udev symlink

After reformatting, the else branch mounts via the symlink:

```bash
mount -t ext4 /dev/disk/by-label/data-volume /mnt/data
```

udevd has not yet processed the new filesystem label, so the symlink does not exist. The mount fails with "No such file or directory." The script exits, and the data volume is never mounted.

## Impact

- `/var/lib/docker` remains on the tmpfs root filesystem
- tmpfs fills up (the reporter observed 5.8 GB used, 55 MB free, 99% full)
- Docker builds and pulls fail with "no space left on device"
- All previously stored data is lost (if the else branch ran)

## Suggested fixes

1. **Resolve the device path with `blkid`, not udev symlinks.** Replace all uses of `/dev/disk/by-label/data-volume` with the output of `blkid --label data-volume` (or equivalent). `blkid` reads the superblock directly and does not depend on udev state.

2. **Add `udevadm settle` after growpart.** `growpart` triggers a partition table re-read. Settling udev before running `e2fsck` prevents the concurrent-probe race.

3. **Fix the "disk in use" check in the else branch.** Instead of scanning `/proc/mounts`, check whether the disk has existing partitions via `lsblk` or `/sys/block/${DISK}/*/`:

   ```bash
   if [ -n "$(lsblk --list --noheadings --output name,type /dev/"${DISK}" | awk '$2 == "part"')" ]; then
       IN_USE=true
   fi
   ```

4. **After mkfs in the else branch, settle udev before mounting.** Add `udevadm settle` (or use the resolved device path) between `mkfs.ext4` and `mount`.

Fix 1 alone eliminates the symlink dependency entirely. Fixes 2-4 are defense-in-depth.

## References

- rancher-sandbox/rancher-desktop#10133 -- user report with logs
- Script: `pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-persistent-data-volume.sh`

jandubois:
A race between growpart/e2fsck and udev device probing can delete the /dev/disk/by-label/data-volume symlink, causing mount failures and potential data loss on Alpine ramdisk VMs.

growpart triggers a kernel partition table re-read, which generates a udev re-probe for the data partition. When e2fsck runs concurrently, udevd's libblkid probe reads a partially modified ext4 superblock, fails with "incorrect ext4 checksum", and removes the symlink.

Core fixes:

1. Resolve the device via blkid output (which probes devices directly) instead of the udev symlink. This eliminates the dependency on udev state entirely. The sed pattern anchors LABEL= on whitespace so it does not match PARTLABEL= on util-linux blkid, and quits after the first match so a stray duplicate label never produces a newline-separated device list.

2. Add udevadm settle after growpart to prevent the concurrent probe from racing with e2fsck. Both settle calls (after growpart and after mkfs) tolerate failure with `|| true`; the blkid resolution already removes the core dependency on udev state, so a missing or stuck udevadm must not crash the boot under `set -e`.

3. Fix the else-branch "disk in use" check to reject any disk that already has partitions or carries a filesystem signature, not just mounted devices in /proc/mounts. The old check could misidentify a partitioned-but-unmounted or raw-formatted data disk as unused and reformat it.

Derive DATA_DISK via `lsblk --output pkname` instead of stripping a single trailing digit, so the parent disk is resolved correctly for any partition naming scheme. Scope the /proc/mounts awk substitution to `$1`.

Ref: rancher-sandbox/rancher-desktop#10133
Closes: #4830

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-persistent-data-volume.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/04-persistent-data-volume.sh
@@ -51,20 +51,31 @@ done
 chmod +x /mnt.sh
 
 mkdir -p /mnt/data
-if [ -e /dev/disk/by-label/data-volume ]; then
-	# Find which disk is data volume on
-	DATA_DISK=$(blkid | grep "data-volume" | awk '{split($0,s,":"); sub(/\d$/, "", s[1]); print s[1]};')
+
+# Resolve the data volume device via blkid, not the udev symlink.
+# The /dev/disk/by-label/ symlink depends on udev having probed the device.
+# A race between growpart (which triggers a partition table re-read and thus
+# a udev re-probe) and e2fsck (which modifies the superblock) can cause the
+# re-probe to see an inconsistent ext4 checksum and delete the symlink.
+# BusyBox blkid doesn't support --label, so we parse the output instead.
+DATA_VOLUME=$(blkid | sed -n 's/^\([^:]*\):.*LABEL="data-volume".*/\1/p')
+
+if [ -n "${DATA_VOLUME}" ]; then
+	DATA_DISK="${DATA_VOLUME%[0-9]}"
 	# growpart command may be missing in older VMs
 	if command -v growpart >/dev/null 2>&1 && command -v resize2fs >/dev/null 2>&1; then
 		# Automatically expand the data volume filesystem
 		growpart "$DATA_DISK" 1 || true
+		# growpart triggers a partition table re-read; settle udev before
+		# touching the device to avoid racing with the re-probe.
+		udevadm settle
 		# Only resize when filesystem is in a healthy state
-		if e2fsck -f -p /dev/disk/by-label/data-volume; then
-			resize2fs /dev/disk/by-label/data-volume || true
+		if e2fsck -f -p "${DATA_VOLUME}"; then
+			resize2fs "${DATA_VOLUME}" || true
 		fi
 	fi
 	# Mount data volume
-	mount -t ext4 /dev/disk/by-label/data-volume /mnt/data
+	mount -t ext4 "${DATA_VOLUME}" /mnt/data
 	# Update /etc files that might have changed during this boot
 	cp /etc/network/interfaces /mnt/data/etc/network/
 	cp /etc/resolv.conf /mnt/data/etc/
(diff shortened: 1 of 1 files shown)


jandubois:
Round 2 AI review: https://jandubois.github.io/lima/20260420-120300-pr-4831.html

I don't think any of the issues are worth addressing.

AkihiroSuda (approved):
Thanks

PR #4832, opened 2026-04-11
Fix LIMA_SHELLENV_ALLOW docs to match implementation

### LINKED ISSUE #4826: LIMA_SHELLENV_ALLOW doesn't remove environment variables not in the list

jandubois:
The docs incorrectly stated that LIMA_SHELLENV_ALLOW restricts propagation to only the listed variables. In fact, the allow list exempts variables from the block list; variables matching neither list still pass through. Also corrected the wildcard docs: `*` is supported anywhere in a pattern, not only as a suffix.

Fixes #4826

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -44,8 +44,8 @@ A custom ssh alias can be used instead by setting the $` + sshutil.EnvShellSSH +
 
 Environment Variables:
   --preserve-env: Propagates host environment variables to the guest instance.
-                  Use LIMA_SHELLENV_ALLOW to specify which variables to allow.
-                  Use LIMA_SHELLENV_BLOCK to specify which variables to block (extends default blocklist with +).
+                  Use LIMA_SHELLENV_BLOCK to specify variables to block (prefix with + to extend the default blocklist).
+                  Use LIMA_SHELLENV_ALLOW to exempt specific variables from the blocklist.
 
 Hint: try --debug to show the detailed logs, if it seems hanging (mostly due to some SSH issue).
 `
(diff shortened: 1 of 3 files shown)


mikerhodes:
Author of https://github.com/lima-vm/lima/issues/4826 here. New docs look good to me --- describe the behaviour clearly and would have shown me how to solve the problem I hit.

unsuman (approved):
Thanks!

PR #4841, opened 2026-04-14
fix: remove kubernetes cli call when there is no config

### LINKED ISSUE #4791: Kubernetes service watcher is always enabled, no matter the template used
### Description

Reviewing process' running inside a Lima Ubuntu VM, I was tracking down why I was seeing several repeated calls to a development server running in port `8080`. Following my process tree up I found it was the guest agent's Kubernetes watcher continuously running `kubectl auth can-i get service` and hitting my unrelated development server running on port 8080 as that's the default address `kubectl` will hit if no config exists. https://github.com/lima-vm/lima/blob/master/pkg/guestagent/guestagent_linux.go#L31-L35

When running a VM with a Kubernetes master node on it, I can understand the usefulness of this feature, but using any non-container based template image it feels wrong having it loop and fail. Or worse when you start a dev server on 8080 and wonder why something is trying to access the Kubernetes API. :P  

Would it be better if this service in the guest agent only started if a configuration in the template enabled it? I don't see an obvious config option at the moment looking through the docs and examples.

dustinblackman:
## Overview

Fixes https://github.com/lima-vm/lima/issues/4791

When the guest agent starts up the kubernetes service watcher, it is incorrectly calling `kubectl` when no kube config exists, falling back to the default kubernetes endpoint of `127.0.0.1:8080`. This PR adds a condition where if no kubernetes config is found, the function exits early without calling `kubectl`. The loop to check for if an kubernetes config exists remains unchanged.

### DIFF
--- a/pkg/guestagent/kubernetesservice/kubernetesservice.go
+++ b/pkg/guestagent/kubernetesservice/kubernetesservice.go
@@ -73,6 +73,10 @@ func (s *ServiceWatcher) attemptToStartKubectl(ctx context.Context) {
 		return
 	}
 	kubeconfig := chooseKubeconfig()
+	if kubeconfig == "" {
+		logrus.Debug("No kubeconfig found; skipping kubernetes service watcher")
+		return
+	}
 	// TODO: ensure that kubeconfig points to a local cluster
 	if err := canGetServices(ctx, kubectl, kubeconfig); err != nil {
 		logrus.WithError(err).Debugf("kubectl auth can-i ... failed; will retry")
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

AkihiroSuda:
CI failing
```
--- FAIL: TestChooseKubeconfig (0.00s)
    --- PASS: TestChooseKubeconfig/returns_KUBECONFIG_env_var_when_set (0.00s)
    --- FAIL: TestChooseKubeconfig/returns_empty_string_when_KUBECONFIG_unset_and_no_files_exist (0.00s)
```

AkihiroSuda:
ping @dustinblackman

AkihiroSuda:
Marking the PR as a draft as the CI is failing. Please click `Ready for review` when ready.

PR #4842, opened 2026-04-14
cli(shell-sync): add BATS tests for metadata change

### LINKED ISSUE #4722: [limactl shell --sync] Add BATS test for metadata change
### Description

Currently, there is no test coverage to verify that file metadata changes are correctly synchronized when using the `limactl shell --sync` command.

unsuman:
Fixes #4722

### DIFF
--- a/hack/bats/tests/shell-sync.bats
+++ b/hack/bats/tests/shell-sync.bats
@@ -29,6 +29,24 @@ teardown() {
     fi
 }
 
+file_uid() {
+    local path="$1"
+    if [[ "$OSTYPE" == darwin* ]]; then
+        stat -f '%u' "$path"
+    else
+        stat -c '%u' "$path"
+    fi
+}
+
+file_perm_octal() {
+    local path="$1"
+    if [[ "$OSTYPE" == darwin* ]]; then
+        stat -f '%A' "$path"
+    else
+        stat -c '%a' "$path"
+    fi
+}
+
 @test 'shell --sync preserves working directory path from host to guest' {
     cd "$TEST_SYNC_DIR"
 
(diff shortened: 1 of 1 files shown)


hack/bats/tests/shell-sync.bats
```diff
@@ -136,6 +150,55 @@ EOF
...
+    assert_symlink_to "$TEST_SYNC_DIR/target_b.txt" "$TEST_SYNC_DIR/current_link.txt"
+}
+
+@test 'shell --sync preserves host ownership after guest modification' {
+    cd "$TEST_SYNC_DIR"
+
+    touch "$TEST_SYNC_DIR/ownership-test.txt"
+    orig_uid=$(file_uid "$TEST_SYNC_DIR/ownership-test.txt")
+
+    run -0 bash -c "limactl shell --sync . --yes '$INSTANCE' sh -ceu 'echo \"modified by guest\" > ownership-test.txt'"
```
jandubois: This test is kind of pointless. Writing to the file will just truncate the existing file, it will not change ownership. But even if you were to change the file owner inside the guest explicitly (via `sudo`), the file back on the host should not change because `rsync` on the host does run as the local user and not as `root`.

I would either scrap this test, or change it to something like

```suggestion
    run -0 bash -c "limactl shell --sync . --yes '$INSTANCE' sh -ceu 'sudo chown root ownership-test.txt'"
```

hack/bats/tests/shell-sync.bats
```diff
@@ -136,6 +150,55 @@ EOF
...
+    run file_perm_octal "$TEST_SYNC_DIR/metadata-perms.sh"
+    assert_output "744"
+}
+
+@test 'shell --sync reflects file rename metadata operation from guest to host' {
+    cd "$TEST_SYNC_DIR"
+
+    echo "rename me" > "$TEST_SYNC_DIR/old_name.txt"
+
+    run -0 bash -c "limactl shell --sync . --yes '$INSTANCE' mv old_name.txt new_name.txt"
```
jandubois: `mv` is basically just `create` and `unlink`, both of which are already tested. Doesn't hurt to add, but seems redundant.

jandubois (dismissed):
Thanks, LGTM

Two tests however look useless to me, so I would remove before merging.

jandubois (approved):
Thanks, LGTM

PR #4846, opened 2026-04-16
CI: add zizmor (GHA linter)

AkihiroSuda:
See https://docs.zizmor.sh/audits/

### DIFF
--- a/.github/actions/setup_cache_for_template/action.yml
+++ b/.github/actions/setup_cache_for_template/action.yml
@@ -26,8 +26,11 @@ runs:
     run: |
       set -eux
       source hack/cache-common-inc.sh
-      print_cache_informations_from_template "${{ inputs.template }}" "${{ inputs.arch }}" >> "$GITHUB_OUTPUT"
+      print_cache_informations_from_template "${INPUTS_TEMPLATE}" "${INPUTS_ARCH}" >> "$GITHUB_OUTPUT"
     shell: bash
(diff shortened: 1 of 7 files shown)


jandubois:
Are you sure this is properly configured? I don't see any output in the action, but when I run it locally with `zizmor .github` I see:

```
57 findings (12 suppressed, 44 fixable): 1 informational, 0 low, 23 medium, 21 high
```

AkihiroSuda:
>  I don't see any output in the action

I guess something will appear in https://github.com/lima-vm/lima/security for maintainers,  after merging this PR

jandubois (dismissed):
Thanks, LGTM

The results are indeed recorded: https://github.com/lima-vm/lima/security/code-scanning?query=pr:4846+is:open

I would prefer to see them as regular lint results, to make them more obvious. I understand this is a "security scan" but everybody can already run `zizmor .github` locally in the repo, so they are not a big secret.

jandubois (approved):
Thanks, LGTM

jandubois:
@AkihiroSuda Since you added a 7 day cooldown period, I wonder if we also want to use https://github.com/jandubois/cooldown

It is more about avoiding a series of broken upgrades. Malicious releases are hopefully removed during the cooldown period.

AkihiroSuda:
> cooldown

Yes, we can try it. Feel free to open a PR

PR #4853, opened 2026-04-19
Change uiutil from survey to pterm

### LINKED ISSUE #1881: User interface library "survey" has been archived

afbjorklund:
Closes #1881

image


liketosweep (commented):
Really glad to see this moving forward  `survey` being archived was a pain point worth addressing.

A few things I noticed while looking through the diff that might be worth considering:

- The two failing QEMU integration tests feel related to `go-isatty` being removed from the imports. `survey` used it for explicit TTY detection  `pterm` may not handle non-TTY stdin (containers, piped input, SSH without a PTY) as gracefully, which could explain the failures. Might be worth a quick TTY check before invoking any interactive prompt.
- `pterm` brings in quite a few more transitive dependencies compared to `survey`. Not necessarily a problem, just something to be mindful of from a binary size and supply chain perspective.

On the suggestion side  it might be nice to expose a `defaultValue bool` parameter on `ConfirmInput` down the line. `survey` had this and it's handy for callers who want a sensible fallback in automated or scripted workflows without blocking on user input.

None of this is a criticism the public API is preserved cleanly and the interrupt handling via `OnInterruptFunc` is genuinely an improvement over what was there before. Just flagging in case any of it is useful!

afbjorklund:
Sorry, but that does not seem to be matching the code?

1.
The github.com/mattn/go-isatty mod is unchanged...
Only an older version was removed from the go.sum

2.
There is no significant change in the binary size:
(the number of dependencies are comparable)
`33M	_output/bin/limactl` # before
`33M	_output/bin/limactl` # after

3.
There is no method called `ConfirmInput`, only `Confirm`
And it does have a `defaultParam` parameter already.

4.
The interrupt function is a bit of a workaround to return
an error, but it is potentially more capable (not used here).

liketosweep:
Thanks for the corrections, @afbjorklund. I definitely misread the ` go.sum ` changes as an import removal and missed the  ` defaultParam ` already present in Confirm. Appreciate you taking the time to verify the binary size as well it’s great to know the switch is so lean. Thanks for the clarification!

AkihiroSuda:
Needs rebasing

AkihiroSuda (approved):
Thanks

unsuman (commented):
Thanks, but looking at the output, the current `pterm` implementation feels like a slight degrade compared to `survey`:

- In the old version, the selected option was completely highlighted with a cyan color, which made it very easy to read at a glance. Right now, it's just plain text with a `>` pointer.

- The `?` symbol, the title **`Creating an instance default`** and the `[Use arrows to move, type to filter]` text provided great visual for users.

### Before

Image

### After

Image

afbjorklund:
It is possible that it can be configured, similar to how the colors were changed.

* https://github.com/pterm/pterm/tree/master/_examples/interactive_confirm

* https://github.com/pterm/pterm/tree/master/_examples/interactive_select

Otherwise there needs to be some PRs for such configuration options, like table's box...

* `InteractiveConfirmPrinter.PrefixStyle`

* `InteractiveSelectPrinter.SelectStyle`

afbjorklund:
Here is the patched version:

image

```diff
--- a/pkg/uiutil/uiutil.go
+++ b/pkg/uiutil/uiutil.go
@@ -35,6 +35,8 @@ func Confirm(message string, defaultParam bool) (bool, error) {
        prompt := interactiveConfirm.
                WithDefaultText(message).
                WithDefaultValue(defaultParam).
+               WithPrefixText("?").
+               WithPrefixStyle(&pterm.Style{pterm.FgGreen}).
                WithOnInterruptFunc(func() {
                        interrupted = true
                })
@@ -57,6 +59,7 @@ func Select(message string, options []string) (int, error) {
        // override the default theme colors (cyan/magenta)
        interactiveSelect.TextStyle = &primaryStyle
        interactiveSelect.SelectorStyle = &secondaryStyle
+       interactiveSelect.SelectStyle = &secondaryStyle
        interrupted := false
        prompt := interactiveSelect.
                WithDefaultText(message).
```

`replace github.com/pterm/pterm => github.com/afbjorklund/pterm v0.0.0-20260612103913-fc71d4a958cc`

* https://github.com/pterm/pterm/issues/793

* https://github.com/pterm/pterm/issues/794

afbjorklund:
This is the default pterm style:

image

jandubois:
I don't really have any strong opinion. I think the old UI looks slightly better, but if you have never seen it, then the new one is just fine too.

I'm not sure why we have to switch though. Just because the old module is archived doesn't mean it will stop working. And it seems unlikely that it will contain a vulnerability that affects us.

So I would only switch of we needed some new functionality that is not available from the current package.

But again, in the end, I don't care that much, and if you have a reason to switch, then go ahead.

afbjorklund:
I think we can leave this change, until it has all the configuration needed to not make it "worse"?

But like you say, if we think the old UI/UX is better then we can just continue to use survey/v2

The new library seems to be having a bit of a PR queue as well, so maybe can wait and see...

It is nice to know that we have a reasonable option though, both charm.sh and tview had "issues".

PR #4856, opened 2026-04-20
Update nobin config with base64 and confusables checking

jandubois:
* Move config to `.nobin.yaml`
* Check for unexpected base64 data of 32 bytes or more
* check for use of unicode runes that can be confused with Latin runes

Both base64 and confusables rely on heuristics. They will not match every possible candidate to avoid false positives. Known exceptions are skipped in the config file.

See https://github.com/jandubois/nobin for more information.

You can also check dependencies by running:

```
go mod vendor
go tool -modfile=./hack/tools/go.mod nobin
```

It currently finds 62 files with issues. I haven't checked in detail, but they do look reasonable. I don't think we want to add skip rules for them; it would become a maintenance hassle.

### DIFF
--- a/.nobin.yaml
+++ b/.nobin.yaml
@@ -1,0 +1,31 @@
+all: false
+gitignore: true
+skip:
+- docs/reports/**/*.pdf
+- pkg/**/*.pb.desc
+- website/static/images/**/*.{gif,png}
+skip-ext: []
+
+# Allowed characters
+allow: []
+allow-bell: false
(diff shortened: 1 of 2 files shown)


.nobin.yaml
```diff
@@ -0,0 +1,31 @@
+all: false
```
AkihiroSuda, line 1: Please add a header to explain what nobin is, and what this file is for

PR #4870, opened 2026-04-23
hack/test-templates: remove WORKAROUND_UPDATE_REBOOT

AkihiroSuda:
The workaround should no longer be needed with the latest image.

Revert most of:
- #4821


jandubois (approved):
Thanks, LGTM

PR #4873, opened 2026-04-24
limactl: add --param shortcut for template parameters

jandubois:
Add a `--param NAME=VALUE` option to the shared edit flag handling used by `limactl create`, `start`, `edit`, `clone`, and `rename`. The option expands to the equivalent yq update for `.param.NAME`, so users can write `--param release=v1.35` instead of `--set '.param.release="v1.35"'`.

Validate parameter names and reject values for params that are not already defined by the template/config, so typos fail before modifying the YAML.

PR generated with Codex GPT-5.5


AkihiroSuda (approved):
Thanks

Issue #4792, opened 2026-03-31
Lima template:experimental/ubuntu-26.04 broken after 2026-03-28 update


whoschek:
### Description

Lima template:experimental/ubuntu-26.04 picks up the OS image from https://cloud-images.ubuntu.com/resolute/current/ 
The images available there were dated February 2026 and worked fine with `limactl start`. On 2026-03-28, Ubuntu 26.04 switched from an alpha to beta release in that download URL, which makes Lima VM boot get stuck forever in a `Waiting for port to become available on 192.168.5.15:22` loop, as shown below.

I don't know if this is simply a glitch in the Ubuntu beta that will be rectified in the final Ubuntu release, or an intended breaking change, or possibly something that Lima could (or should?) work around somehow. I'm reporting it here just FYI, and in the hope that anyone might have more insight how to best get this fixed eventually.

Here is what still works today:

```
$ limactl --version
limactl version 2.1.0

$ uname -a
Darwin MacBook-Pro.local 24.6.0 Darwin Kernel Version 24.6.0: Mon Jan 19 21:56:28 PST 2026; root:xnu-11417.140.69.708.3~1/RELEASE_ARM64_T6020 arm64

$ limactl --tty=false create --name=mylimavm template:ubuntu-25.10; limactl --tty=false start --progress=true --log-level=debug mylimavm
<This works fine>
$ limactl delete --force mylimavm
```

Here is how template:experimental/ubuntu-26.04 nomore works after the 2026-03-28 update:

```
$ limactl --tty=false create --name=mylimavm template:experimental/ubuntu-26.04; limactl --tty=false start --progress=true --log-level=debug mylimavm
INFO[0000] Terminal is not available, proceeding without opening an editor
INFO[0000] Attempting to download the image              arch=aarch64 digest= location="https://cloud-images.ubuntu.com/resolute/current/resolute-server-cloudimg-arm64.img"
INFO[0000] Using cache "/Users/foo/Library/Caches/lima/download/by-url-sha256/ecfbe382d6cd7377cad5019591dc0c9579cc094058d17620bded9b7dd648571e/data"
INFO[0002] Run `limactl start mylimavm` to start the instance.
DEBU[0000] interpreting argument "mylimavm" as an instance name
INFO[0000] Using the existing instance "mylimavm"
DEBU[0000] Make sure "user-v2" network is stopped
DEBU[0000] Make sure usernet network is stopped
DEBU[0000] Make sure "shared" network is stopped
[...]

PR #4874, opened 2026-04-24
templates: add ubuntu-26.04

AkihiroSuda:
Not default yet, as the first boot takes 2 minutes due to:
- #4792

See also https://bugs.launchpad.net/ubuntu/+source/cloud-init/+bug/2136392


jandubois (approved):
Thanks, LGTM

PR #4877, opened 2026-04-24
Fix: external driver transport irregularities

### LINKED ISSUE #3970: Remove pre-grpc driver config validation transport
I think we can completely remove the pre-grpc stdio pipe transport if we merge `FillConfig()` and `Configure()`. It was originally added because we assumed validation would run across all available drivers in the registry, which could slow down things when using gRPC. However, with the new VM-type resolution, we’re guaranteed to either resolve the default driver or use the one explicitly specified by the user.

_**What do you all think about this?**_

unsuman:
This PR fixes a bunch of external driver transport irregularities.

- Merged `FillConfig()` into `Configure()`
  - The `Driver` interface method `FillConfig(ctx, cfg, filePath)` has been removed. Its logic is now included into `Configure`, which is updated to return `(*ConfiguredDriver, error)`. The previous `--pre-driver-action` flag is removed from `server.Serve()`. Fixes #3970

- gRPC server socket moved from tmp to the instance directory
  - Previously the socket was created in the OS temp directory with a PID-suffixed name (`/tmp/lima-driver-<name>-<pid>.sock`). It is now created inside the instance directory (`~/.lima/<instance>/lima-driver-<name>.sock`).

- Single persistent gRPC server, shared with hostagent
  - `server.Start()` now checks whether a server is already reachable on the socket before spawning a new process. If the socket is connectable, the existing server is reused. The server's 60-second "no client connected" self-shutdown timer and the `listenerTracker` wrapper are also removed, since the server is now expected to remain alive with the hostagent.

- Simplified shutdown: `Stop()` and `Disconnect()`. Shutdown is now split into two clean operations:
  - **`server.Stop(instDir, force bool)`**: scans `lima-driver-*.pid` in the instance directory, signals each live process with `SIGTERM` (or `SIGKILL` when `force=true`), polls for exit, then removes both the PID file and the corresponding `lima-driver-*.sock`.
  - **`server.Disconnect()`**: closes the in-process gRPC client connections, called by the CLI at the end of a command so hostagent's server continues running.

I’ll squash the commits once this review is complete.


jandubois:
PR has merge conflicts. As you keep pushing updates to the branch, it would be useful to know when you are done, and the PR is ready for review.

The easy way to do this is to mark it as draft once you start updating it again, and only undraft once you are sure you are done.

The alternative is to explicitly ask for reviews when you are done, but why not use the draft mechanism?

And yes, I'm totally aware of the way one always thinks it is just one more change, and then it is really ready... 😄

unsuman:
Sure I'll keep this in mind next time onwards! Once rebased this PR is ready to be reviewed.

> and then it is really ready

This feeling is so real 😁

AkihiroSuda:
Needs rebase

pkg/driver/external/server/server.go
```diff
@@ -189,193 +144,183 @@ func handleInspectStatus(ctx context.Context, driver driver.Driver) {
...
-		cancel()
 		return fmt.Errorf("failed to determine instance directory: %w", err)
 	}
+	socketPath := driverSocketPath(instanceDir, extDriver.Name)
+
+	// If the socket already exists and is connectable, reuse the existing server.
+	if isServerRunning(socketPath) {
+		extDriver.Logger.Debugf("Reusing existing external driver server for %#q at %s", extDriver.Name, socketPath)
+		extDriver.Client, err = client.NewDriverClient(socketPath, extDriver.Logger)
+		if err != nil {
```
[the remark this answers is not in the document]
unsuman: This race is very hard to reproduce and I cannot think of a way to actually initiate this race, so I think it's a overkill for what's actually at stake here

jandubois:
Sorry, running out of time again for reviews, but I have an AI review that may have something useful (check at least the issue it claims as "critical"):

https://jandubois.github.io/lima/20260706-203659-pr-4877.html

unsuman:
Thanks for the AI review which helped us catch a regression(C1) that I had overlooked. Also fixed I1 and I2.

AkihiroSuda:
> address comments by copilot in server.go

The comments should be addressed in each of commits that corresponds to a single topic.
(Probably you can let your AI tidy up the commits)

website/content/en/docs/dev/internals.md
```diff
@@ -107,7 +107,8 @@ Host agent:
 - `ha.stderr.log`: hostagent stderr (human-readable messages)
 
 External drivers:
-- `lima-driver-<driver-name>-<owner>.pid`: external driver process PID, where `<owner>` is either `ha` (written by the hostagent) or `cli` (written by the CLI process). Used by `limactl stop --force` and `limactl rm --force` to discover and kill orphaned external driver processes. Each process only removes its own PID file on graceful exit; force stop globs all matching files and kills them.
+- `lima-driver-<driver-name>.pid`: external driver gRPC server process PID. A single server instance runs per Lima instance and is shared by all callers (CLI, hostagent, inspect). The server is started on first use and stopped only when the instance shuts down.
+- `lima-driver-<driver-name>.sock`: Unix socket for gRPC communication with the external driver server.
```
AkihiroSuda: We have a tight limit on the full path of the socket
https://github.com/lima-vm/lima/blob/9686218c13e8e8915bf0ad925dd1d7348151b31f/pkg/osutil/osutil_others.go#L14

AkihiroSuda:
```suggestion
- `drv-<driver-name>.sock`: Unix socket for gRPC communication with the external driver server.
```

or `<driver-name>.drv.sock`

pkg/limayaml/validate.go
```diff
@@ -617,21 +610,21 @@ func warnExperimental(y *limatype.LimaYAML) {
 	if *y.OS == limatype.WINDOWS {
 		logrus.Warnf("`os: %s` is experimental", limatype.WINDOWS)
 	}
-	if *y.MountType == limatype.VIRTIOFS && runtime.GOOS == "linux" {
+
+	if y.MountType != nil && *y.MountType == limatype.VIRTIOFS && runtime.GOOS == "linux" {
```
AkihiroSuda: Seems to be an irrelevant change?

pkg/limayaml/validate.go
```diff
@@ -617,21 +610,21 @@ func warnExperimental(y *limatype.LimaYAML) {
...
 		logrus.Warn("`mountType: virtiofs` on Linux is experimental")
 	}
-	switch *y.Arch {
-	case limatype.RISCV64, limatype.ARMV7L, limatype.S390X, limatype.PPC64LE:
-		logrus.Warnf("`arch: %s ` is experimental", *y.Arch)
+	if y.Arch != nil {
```
AkihiroSuda: Seems to be an irrelevant change?

AkihiroSuda (approved):
Thanks, but next time please split PR into multiple ones

AkihiroSuda:
https://lima-vm.io/docs/releases/breaking/ should be updated too?

AkihiroSuda:
Merging, but please consider updating https://lima-vm.io/docs/releases/breaking/

PR #4880, opened 2026-04-24
CI: pin Windows QEMU to 10.2.0

jandubois:
QEMU 11.0.0 (installed via winget) breaks systemd boot on the Windows runner under TCG emulation. systemd fails to fork its sandboxing environment with "Failed to fork off sandboxing environment for executing generators: Protocol error", so SSH never starts and the hostagent times out.

The last green `Windows tests (QEMU)` run on master used QEMU 10.2.0 on 2026-04-22; every push since has picked up QEMU 11.0.0 and failed the same way.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -245,7 +245,10 @@ jobs:
       run: make
     - name: Install QEMU
       run: |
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

AkihiroSuda (approved):

PR #4872, opened 2026-04-23
Add parameters for Kubernetes installation

afbjorklund:
We added a variable for the hardcoded k8s version:

`https://pkgs.k8s.io/core:/stable:/v1.36/deb/`

Previously the stability was hardcoded to "stable":

`https://pkgs.k8s.io/core:/stable:/v${VERSION}/deb/`

There was also no way to select the version to install.

It would always use https://dl.k8s.io/release/stable.txt

So rename "VERSION" to "RELEASE".

And add a new "STABILITY" variable.

Make "VERSION" be the k8s version...

`https://pkgs.k8s.io/core:/${STABILITY}:/${RELEASE}/deb/`

Issue #4844

```yaml
param:
  stability: "stable" 
  release: "stable"
  version: "stable"
```

```yaml
param:
  stability: "prerelease"
  release: "v1.37"
  version: "v1.37.0-alpha.1"
```

See https://kubernetes.io/blog/2023/08/15/pkgs-k8s-io-introduction/ for the different params.

And https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/

Closes #4844
Closes #4871 

----

Example, for installing an older version:

```
limactl start ./templates/k8s.yaml --set '.param.release="v1.35"' --set '.param.version="stable-1.35"'
```

EDIT: With the new syntactic sugar:

```
limactl start template:k8s --param "release=v1.35" --param "version=stable-1.35"
```

Note: the second parameter is optional, kubeadm will detect the version skew and do a fallback:

`remote version is much newer: v1.36.2; falling back to: stable-1.35`

Note: older versions of Kubernetes _might_ not be compatible with the Lima version of containerd.

You would have to check the version matrix: https://containerd.io/releases/

Issue #4868, opened 2026-04-23
`template:k8s` fails to install Kubernetes since 2026-04-23 (Kubernetes 1.36): `kubelet : Depends: kubernetes-cni (>= 1.2.0) but it is not installable`


AkihiroSuda:
https://github.com/lima-vm/lima/blob/7e52923a4310af2c0ca41688fcff3cf6f70ce863/templates/k8s.yaml#L53-L61

It fails to run `apt-get install -y kubelet kubeadm kubectl`

```
+ apt-get update
Hit:2 http://ports.ubuntu.com/ubuntu-ports noble InRelease
Get:1 https://prod-cdn.packages.k8s.io/repositories/isv:/kubernetes:/core:/stable:/v1.36/deb  InRelease [1224 B]
Get:3 https://prod-cdn.packages.k8s.io/repositories/isv:/kubernetes:/core:/stable:/v1.36/deb  Packages [1736 B]
Hit:4 http://ports.ubuntu.com/ubuntu-ports noble-updates InRelease
Hit:5 http://ports.ubuntu.com/ubuntu-ports noble-backports InRelease
Hit:6 http://ports.ubuntu.com/ubuntu-ports noble-security InRelease
Fetched 2960 B in 1s (3172 B/s)
Reading package lists...
+ apt-get install -y kubelet kubeadm kubectl
Reading package lists...
Building dependency tree...
Reading state information...
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 kubeadm : Depends: cri-tools (>= 1.30.0) but it is not installable
 kubelet : Depends: kubernetes-cni (>= 1.2.0) but it is not installable
E: Unable to correct problems, you have held broken packages.
```

AkihiroSuda:
Reported:
- https://github.com/kubernetes/kubernetes/issues/138534

afbjorklund:
Not again!

* https://github.com/kubernetes/release/issues/4100

PR #4882, opened 2026-04-24
templates: k8s: pin Kubernetes version to 1.36

AkihiroSuda:


afbjorklund:
I was thinking to leave it at 1.35 for a while (minikube and kind haven't updated either).

But can't we just revert 1f17f993ffa3f7d0bb5c037c4b8f5c29a719bd6f and go back to "stable", before adding the params?

unsuman:
ping @AkihiroSuda

afbjorklund (approved):
The end result would be the same, so we can revert in 2.1.3 and "fix" on master*

\* by adding the version as parameters, so that it is easier to pin 1.36 for 1.37 fails

* https://github.com/lima-vm/lima/pull/4872

But the packages are released (and updated) now, so the 1.36 install works again

* https://github.com/lima-vm/lima/issues/4868

PR #4883, opened 2026-04-24
shell: fix path quoting for rsync 3.0.0 through 3.2.3

### LINKED ISSUE #4720: [limactl shell --sync] Fix path spaces breaking with rsync `3.0.0` through `3.2.3`
### Description

The code quotes `destRsyncDir` for `rsync < 3.0.0` (macOS's ancient 2.6.9). But --protect-args did not become the default until `rsync 3.2.4`. Users running `rsync 3.0.0-3.2.3` (common on older Linux, e.g. Ubuntu 20.04) will see failures when `syncDirVal` contains spaces, as the remote shell splits the unquoted path.

Mysterio-17:
### Problem
`destRsyncDir` is only shell-quoted for rsync < 3.0.0, but `--protect-args`
didn't become the default until rsync 3.2.4. For versions 3.0.0–3.2.3
(e.g. rsync 3.1.3 on Ubuntu 20.04), paths with spaces get split by the
remote shell, breaking `limactl shell --sync`.

### Fix
Changed the version threshold from `3.0.0` to `3.2.4` so the destination
path is properly quoted for all rsync versions that don't default to
`--protect-args`.

Fixes #4720

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -366,12 +366,13 @@ func shellAction(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("failed to create the synced workdir in guest instance: %w", err)
 		}
 
-		// The macOS release of rsync (the latest being 2.6.9) does not support shell escaping of destination path but other versions do.
+		// Quote the destination path for rsync versions before 3.2.4, where --protect-args is not the default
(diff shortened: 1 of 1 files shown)


unsuman (approved):
Thanks, LGTM!

PR #4884, opened 2026-04-24
bats: fix second-boundary race in "shell --instance resolves instance name"

jandubois:
The test ran `limactl shell` twice and asserted that the two stderr outputs were byte-identical, but logrus prefixes each line with a `time="..."` timestamp at second granularity. When the two invocations straddled a second boundary the test failed with only the timestamp differing, as seen on the BATS job for PR #4873.

Normalize the timestamp to a fixed placeholder before comparing, so the test continues to verify that `--instance` and the positional form resolve identically without depending on wall-clock alignment.


AkihiroSuda (approved):
Thanks

Issue #4819, opened 2026-04-09
Windows: drop dependency on `cygpath.exe`


AkihiroSuda:
Beside the CI, it seems used only here
https://github.com/lima-vm/lima/blob/4c791f115ba69abdc29dfcef74f0c65519b45248/pkg/ioutilx/ioutilx.go#L53-L61

arixmkii:
This is needed for SSHFS mounts. And probably for shell opening in the current directory mapped into VM.

This dependency brings problems. For my experimental builds with replacing Windows GIT dependency with WSL Alpine container I had to add workaround to use `wlspath` instead of `cygpath`. 

Both can be replaced by some rules/rewrites for the default scenarios, but the issues appear, when someone has non-default fstab in either msys2 (Windows GIT) or WSL.

afbjorklund:
Replacing with a library sounds like a good idea, not sure if there exists one - or look at the source code?

liketosweep:
Hi @AkihiroSuda. I just submitted a PR for #4810, which handles a direct downstream symptom of this cygpath.exe dependency. I’d love to take on this roadmap item to remove the dependency globally.

To ensure a truly standalone architecture without falling back on other external binaries (like wslpath), I propose this native Go approach:

Native Go Default: Use `path/filepath` and string manipulation to handle standard C:\ <-> /c/ translations natively.

Custom `fstab` Support: Implement a native parser to read common `MSYS2 fstab` locations to support users with custom mount points (addressing @arixmkii's note).

Manual Override: Introduce an environment variable (e.g., LIMA_WINDOWS_MOUNT_PREFIX) as an escape hatch for complex, non-standard setups.

Does this architecture align with the project's vision? If so, I'll start drafting the solution and PR.

AkihiroSuda:
> Native Go Default: Use path/filepath and string manipulation to handle standard C:\ <-> /c/ translations natively.

👍 

> Custom fstab Support: Implement a native parser to read common MSYS2 fstab locations to support users with custom mount points (addressing @arixmkii's note).

When MSYS2 is installed it can just execute MSYS2's cygpath.exe binary

liketosweep:
Makes sense, that significantly simplifies the edge-case handling.

To confirm, the updated plan will be:

Native Fast-Path: Use native Go (path/filepath and string manipulation) for standard C:\ <-> /c/ translations.

[...]

PR #4885, opened 2026-04-24
Windows: support native OpenSSH; no Git for Windows / MSYS2 required

jandubois:
## DO NOT MERGE

This PR has been created with Claude Code using Opus 4.7.

It is for discussion and testing purposes only and has not been reviewed yet!

## Summary

This branch makes `limactl` work on Windows hosts that have only the toolchain shipped in a default Windows 10/11 install (native OpenSSH, `wsl.exe`, `tar`). Lima previously required Git for Windows or MSYS2 on `PATH` for `cygpath`, `ssh`, `ssh-keygen`, `scp`, and `gzip`. After this branch, none of those external tools are required for the core flow on plain Windows.

The PR is offered as research / discussion. The goal is to surface what does and does not work end-to-end, get CI signal across the existing Windows runners plus two new "plain Windows" runners (one per driver), and decide together how much of this we want to land.

Related: #4819. The originating thread there narrowed to "drop the `cygpath.exe` dependency"; investigating that turned up the broader story below.

## Background — what was actually required, and why

Two coupled root causes drove the historical Git-for-Windows / MSYS2 requirement:

1. **Native Windows OpenSSH does not implement SSH multiplexing** ([PowerShell/Win32-OpenSSH#1328](https://github.com/PowerShell/Win32-OpenSSH/issues/1328), still open as of Feb 2026). When Lima needed `ControlMaster` for the legacy ssh-based dynamic port forwarder, native ssh would not work, so Cygwin-built ssh was required.
2. **Cygwin-built ssh expects Cygwin-style paths** (`/c/Users/jan/...`). Hence `cygpath` to translate. Hence the cascade of "why are we calling cygpath everywhere?" sites in the codebase.

Three things make the dependency easier to drop than I expected when starting:

- **Lima's default port forwarder has been Go-native since v1.1.0** (`pkg/portfwd/forward.go`, gRPC-tunnelled via vsock). The legacy ssh-based forwarder is opt-in via `LIMA_SSH_PORT_FORWARDER=true`. So `ControlMaster` is not actually needed by the default flow — verified empirically: starting `python3 -m http.server` in the guest produces a TCP listener owned by `limactl.exe`, not ssh.
- **Native Windows OpenSSH ships `sftp-server.exe`** (when the OpenSSH Server optional feature is installed) and is auto-detected by `sshocker` via `LookPath("sftp-server")`. So reverse-sshfs works natively too once the host-path translation is corrected.
- **Native Windows OpenSSH treats `-F /dev/null` as an empty config**, the same way Cygwin ssh does. Lima's hardcoded `-F /dev/null` argument did not need to change.

## Detection mechanism

A new `sshutil.IsSSHCygwin(sshExe)` checks whether `cygpath.exe` lives in the same directory as `ssh.exe` (the layout used by Git for Windows and MSYS2), after resolving symlinks so chocolatey/scoop shims do not throw the directory check off. Results are cached per-resolved-absolute-path and logged once per path at Debug level. `sshutil.PathForSSH(ctx, sshExe, path)` then dispatches:

- Cygwin-based ssh → `cygpath` for path translation (preserves any custom MSYS2 fstab the user has)
- Native Windows OpenSSH → `filepath.ToSlash` (e.g. `C:/Users/jan/...`), which native `ssh`, `ssh-keygen`, `scp`, and `sftp-server` accept

`ioutilx.WindowsSubsystemPath` (used in the few sites that produce a guest mount-point path rather than a host argument) keeps `cygpath` as the preferred backend but falls back to a native drive-letter conversion (`C:\Users\jan` → `/c/Users/jan`) when `cygpath` is unavailable, so the on-disk Lima config remains identical regardless of toolchain.

## Per-commit breakdown

```
ce3a0696 sshutil: skip cygpath on Windows when ssh is native OpenSSH
bc99aca8 sshutil: detect Win32-OpenSSH version, export PathForSSH
6924004a copytool: support native Windows OpenSSH
df59f8ff downloader: decompress gzip in pure Go
8e56f7e7 ioutilx, hostagent: support reverse-sshfs on plain Windows
1f8a5303 CI: stop forcing Git for Windows on PATH; add plain-Windows smoke test
9858c839 docs: reflect that plain Windows is now a supported host
07328ee2 add diagnosability for plain-Windows path
d720cd4e ioutilx: drop unnecessary else after returning if branch
f21ba237 client: silence flaky nolintlint at the grpc.Dial directive (reverted)
3e09daeb golangci: disable nolintlint
7474e1e1 CI: build the Linux guest agent and fail-loudly in windows-plain
6e599d6e remove _LIMA_WINDOWS_EXTRA_PATH
1218a5d1 CI: split windows-plain by driver, add reverse-sshfs round-trip (partially reverted)
bae5d205 CI: install templates for windows-plain-qemu
e147b971 CI: read symlink targets from git object store
f98222d5 CI: re-skip mount-home on Windows, drop reverse-sshfs round-trip
7f760ab6 sshutil, copytool, ioutilx: address PR review
29cb4234 CI: assert template symlink target stays inside templates/
c78cdc34 docs: sftp-server is an optional Windows feature; toolchain-swap caveat
```

`f21ba237` is reverted by `3e09daeb`; `f98222d5` partially reverts `1218a5d1`; in a non-research PR these would be squashed.

## Review feedback addressed (round 1)

Commit `7f760ab6` folds in fixes for the two Important issues and several Suggestions from the first review pass:

- **I1** — `copytool.checkRsyncOnGuest` now strips the mux options on Windows. The probe previously hit the same ControlMaster failure the `rsync.Command`/`scp.Command` paths work around, so a working rsync install on native Windows OpenSSH was rejected before the guest-side `command -v rsync` ran.
- **I2** — `copytool.parseCopyPaths` now uses `filepath.IsAbs` instead of `filepath.VolumeName != ""` to decide local vs remote. The previous check classified the drive-relative form `C:foo` as local, silently shadowing single-letter instance names. A new `TestParseCopyPathsWindowsDriveLetter` table test locks in `C:\foo` / `C:/foo` / `C:foo` / `instance:path`.
- **S1** + **S3** — `sshutil.IsSSHCygwin` drops its `sync.Once` in favour of a map keyed by the resolved absolute path, with `filepath.EvalSymlinks` applied before the `cygpath.exe` sibling check so chocolatey/scoop shims do not mis-detect.
- **S4** — `ParseOpenSSHVersion` logs the unparsed banner at Debug when it falls back to `0.0.0`, so cipher-selection and scp-URL downgrades are traceable.
- **S6** — the `windows-plain-qemu` template-install step now asserts the resolved symlink target stays inside `templates/` after the path-space chain walk. Defensive only; Lima's current templates don't use `..`.
- **S7** — WSL2 docs now call out that hostagent recomputes reverse-sshfs `LocalPath` on every start; users should not swap between native Windows OpenSSH and Cygwin ssh between `limactl create` and `limactl start` on a QEMU instance with reverse-sshfs mounts. The follow-up would persist `LocalPath` at create time.
- **S8** — `ioutilx.WindowsSubsystemPath`'s native cygpath fallback now inserts the separator explicitly, so a hypothetical drive-relative input `C:foo` does not collapse to `/cfoo`.
- **S9** — WSL2 docs distinguish OpenSSH Client (default on Windows 10 1803+) from OpenSSH Server (optional Feature on Demand, provides `sftp-server.exe`; needed for QEMU + reverse-sshfs).
- **S10** — `parseCopyPaths` now documents UNC-path behaviour inline.

Skipped by design:
- **S5** — transitional warning for `_LIMA_WINDOWS_EXTRA_PATH` still being set. Judged noise over value.

Deferred to follow-ups:
- **S7** persistence of `LocalPath` at create time (design change).
- Commit squash of `f21ba237`+`3e09daeb` and the similar pairs (only if/when this branch moves toward merge).

## Touched call sites

- `pkg/sshutil/sshutil.go` — `DefaultPubKeys` (ssh-keygen), `identityFileEntry` (IdentityFile), `SSHOpts` (ControlPath); `ParseOpenSSHVersion` regex extended to match `OpenSSH_for_Windows_X.YpZ`; `IsSSHCygwin` cache reshaped per review
- `pkg/copytool/copytool.go` — `parseCopyPaths` host-path translation with `filepath.IsAbs` gate (corrected per review I2)
- `pkg/copytool/scp.go`, `pkg/copytool/rsync.go` — strip `ControlMaster`/`ControlPath`/`ControlPersist` on Windows so the underlying ssh does not try to use a mux socket that is unavailable (native) or unreliable (Cygwin). Includes `checkRsyncOnGuest` per review I1.
- `pkg/hostagent/mount.go` — reverse-sshfs `LocalPath` translation now matches the toolchain
- `pkg/ioutilx/ioutilx.go` — native cygpath fallback, plus debug logging of the conversion, hardened per review S8
- `pkg/downloader/downloader.go` — gzip decompression via in-process `compress/gzip`. Other formats still shell out
- `cmd/limactl/main.go` — drop the `_LIMA_WINDOWS_EXTRA_PATH` PATH-injection hook (no longer required for the core flow)
- `.golangci.yml` — disable `nolintlint`, which was producing flaky cross-platform "directive is unused" failures because golangci-lint's analysis cache invalidation can cause SA1019 to drop in and out of the post-processing pipeline

## Tested locally on Windows 11 with QEMU 10.2.0

Hardware: Windows 11 Pro 26100, native OpenSSH `OpenSSH_for_Windows_9.5p2`, with `PATH` cleaned of Git for Windows and MSYS2 entries.

### WSL2 driver (`templates/experimental/wsl2.yaml`, finch rootfs)

- ✅ `limactl create` — generates ed25519 keypair via native ssh-keygen with native paths
- ✅ `limactl start` — boots, runs requirements ("ssh", "user session is ready for ssh", "Explicitly start ssh ControlMaster" all pass with `ControlMaster=no` explicitly set on every ssh invocation)
- ✅ `limactl shell uname -a` → `Linux 6.6.87.2-microsoft-standard-WSL2 x86_64`
- ✅ Dynamic port forwarding — host `Invoke-WebRequest http://127.0.0.1:8888` reaches a `python3 -m http.server` running in the guest. The TCP listener on the host is `limactl.exe` itself (PID owns the socket), confirming the gRPC/vsock forwarder is in use, not `ssh -O forward`
- ✅ `limactl copy` host → guest and guest → host
- ✅ `limactl shell --workdir=/tmp`
- ✅ `limactl stop`, `limactl delete --force`

### QEMU driver with reverse-sshfs (`vmType: qemu`, `mountType: reverse-sshfs`)

- ✅ `limactl create` — default `mountPoint` computed correctly via the new native cygpath fallback (`C:\Users\jan\qemu-share` → `/c/Users/jan/qemu-share`)
- ✅ `limactl start` — sshocker auto-detects `C:\Windows\System32\OpenSSH\sftp-server.exe` via `LookPath("sftp-server")`, mount succeeds
- ✅ Bidirectional read/write across the mount: file written on host visible in guest, file written from guest visible on host with matching content. Verified with both PowerShell `Get-Content` and bash `cat`

Note: this local verification used a custom `qemu-share` mount, not the default template's `~` mount. CI exercises the default template and hits a **pre-existing, non-PR-related** guest-side `fusermount3: mount failed: Permission denied` on Ubuntu 25.10. See "Intentionally not addressed" below.

### Existing Cygwin-ssh path (Git for Windows on `PATH`)

- ✅ `limactl create` + `start` + `shell` + `copy` all still work unchanged when Git for Windows is on PATH. `IsSSHCygwin` correctly identifies the toolchain, `cygpath` is invoked, paths come out as `/c/Users/...`. No regression for users who have Git for Windows installed.

### Unit tests

`go test ./pkg/sshutil/... ./pkg/copytool/... ./pkg/downloader/... ./pkg/ioutilx/... ./pkg/limayaml/...` passes on Windows. Cross-compile of `pkg/sshutil` for darwin and a full Linux build of `./cmd/limactl` both succeed.

### Reverse-sshfs mount into the default template's `~` fails on GH Windows runners (guest-side, Ubuntu 25.10)

`hack/test-templates.sh` keeps the pre-existing skip

```
[ "${OS_HOST}" = "Msys" ] && CHECKS["mount-home"]=
```

for the `default` template case. An earlier commit on this branch (`1218a5d1`) removed that skip on the hypothesis that the new path-translation work covered the original failure. It did not: both `windows-qemu` (MSYS2 sftp-server) and `windows-plain-qemu` (native sftp-server) fail with

```
[hostagent] fusermount3: mount failed: Permission denied
```

inside the Ubuntu 25.10 guest, after `/etc/fuse.conf` has been populated with `user_allow_other` by Lima's cloud-init and the hostagent's pre-mount requirement check has confirmed it. Same symptom on both toolchains, so the failure is below the sftp-server / ssh boundary.

The guest runs with `kernel.apparmor_restrict_unprivileged_userns=1` and `kernel.apparmor_restrict_unprivileged_unconfined=1` (Ubuntu 23.10+ default, applied by `/usr/lib/sysctl.d/10-apparmor.conf`). Lima has an AppArmor profile for `/usr/local/bin/rootlesskit` to work around this restriction for containerd, but no equivalent for `fusermount3`. The leading (unconfirmed) hypothesis is that libfuse3 uses user namespaces for the mount when invoked by a non-root user with `-o allow_other`, and AppArmor denies that for the unconfined `lima` user. Audit is disabled in the kernel (`audit: initializing netlink subsys (disabled)`) so DENIED lines don't show in dmesg without re-enabling it.

This reproduces on the CI runners on both `master` and this branch, and is outside the scope of this PR (native Windows OpenSSH support). Tracking for a separate follow-up; the commit message on `f98222d5` has the full trail.

### `limactl shell --sync` is left unchanged — postponed to a follow-up

`limactl shell --sync DIR` currently does an early `exec.LookPath("rsync")` and exits with `rsync is required for --sync but not found` when rsync is unavailable. After this branch:

- **Plain Windows**: `--sync` still fails immediately. Was already failing before; nothing about this PR changes that.
- **Git for Windows users**: `--sync` *also* fails. Despite my (incorrect) earlier claim, Git for Windows does **not** include rsync in its bundle — only `ssh`, `scp`, `ssh-keygen`, etc. Users have to manually drop rsync into `Git\usr\bin\` (there is a well-known how-to for it). I learned this while writing this PR.
- **MSYS2 users**: `--sync` works after `pacman -S rsync`.
- **macOS / Linux users**: unchanged, rsync is universally present.

A natural fix is to fall back to `scp` when rsync is missing — `pkg/copytool` already has the auto-fallback machinery, but `cmd/limactl/shell.go:223` opts out of it by hardcoding `BackendRsync`. Switching to `BackendAuto` is small. The reason it is **not** in this PR is that the scp fallback is a meaningful behaviour change for `--sync`:

| Feature | rsync | scp |
|---|---|---|
| Bulk copy host ↔ guest | yes | yes |
| `--delete` (propagate file removals) | **yes** | **no** |
| Itemized stats for the "Accept the changes? (added: X, deleted: Y, modified: Z)" prompt | **yes** (`--itemize-changes`) | **no equivalent** |
| Incremental delta transfer on subsequent syncs | **yes** | re-copies the whole tree |

The most consequential gap is `--delete`. If the user adds a file on the host, syncs, deletes it on the host, and re-syncs, rsync would propagate the deletion to the guest; scp would leave a phantom file. The diff-view ("View the changed contents") path also degrades: without `--itemize-changes` the prompt collapses to a generic "Accept all changes? (y/n)" without specifics, unless we re-implement file-tree comparison in Go.

A workable fallback design would be: when rsync is unavailable, switch to `BackendAuto`, log a one-time warning that `--delete` is disabled and stats will be coarse, and either skip the itemized-stats prompt or replace it with a native tree walk. None of that is hard, but the UX policy decision (warn-and-degrade vs flag-gated `--sync-tool=scp` opt-in vs reject) deserves its own discussion. Tracking as a follow-up.

### Other deferred items

- The `nolintlint` linter is disabled in `.golangci.yml`. This silences a real but flaky failure mode (analysis-cache nondeterminism causes the SA1019 deprecation to drop in and out of the input to nolintlint, which then flags the `//nolint:staticcheck` directive at `pkg/driver/external/client/client.go:35` as unused). The proper fix is to migrate `grpc.Dial` to `grpc.NewClient` and remove the directive entirely; out of scope here.
- `hack/test-templates.sh` is unchanged in this PR. It still uses cygpath, bash arrays, netcat, etc., and still runs under MSYS2 bash in the existing `windows-wsl2` and `windows-qemu` jobs. Rewriting it for plain Windows is a separate, larger project.
- Decompressors other than gzip (bzip2, xz, zstd) still shell out. Lima's default WSL2 image is `.tar.gz`, so the gzip path is the one that mattered. The others can be migrated similarly when needed.
- Reverse-sshfs path translation in `pkg/hostagent/mount.go` uses `PathForSSH` (toolchain-aware) rather than `WindowsSubsystemPath` (cygwin-style always). Persisting `LocalPath` at create time would remove the toolchain-swap hazard described in the new WSL2 docs note (review S7); kept out of scope here.

### Existing jobs (`windows-wsl2`, `windows-qemu`)

- Removed `_LIMA_WINDOWS_EXTRA_PATH = 'C:\Program Files\Git\usr\bin'`. After this branch, limactl does not need anything from there. Strict subset of the old environment.
- Removed the `_LIMA_WINDOWS_EXTRA_PATH` entry from `MSYS2_ENV_CONV_EXCL` (it was only there because we were setting the var).
- `hack/test-templates.sh` still skips the `mount-home` check on Msys (see the section above for the reason).

### New jobs

Two new jobs, mirroring the two existing Windows drivers but on a plain-Windows host:

**`windows-plain-wsl2`** — builds with `go build` (no `make`, no MSYS2 bash needed for the build step), scrubs `PATH` of MSYS2 and Git for Windows, then runs a PowerShell smoke test (`create` → `start` → `shell` → `copy` → `stop` → `delete`) against the same WSL2 rootfs template the existing `windows-wsl2` job uses.

**`windows-plain-qemu`** — same shape as `windows-plain-wsl2`, but installs QEMU 10.2.0 and uses `templates/default.yaml` (matching `windows-qemu`). Smoke test covers the same create/start/shell/copy/stop/delete sequence. The reverse-sshfs mount itself is **not** exercised in this job because it hits the pre-existing Ubuntu-25.10 fusermount3 issue described above.

Both new jobs:

- Print `PATH` before and after the scrub.
- Assert `ssh` resolves to `C:\Windows\System32\OpenSSH\ssh.exe` specifically.
- Run a tool inventory of ~20 binaries Lima might shell out to, classified as **required** (must resolve to something native), **forbidden** (`cygpath`, `pacman` — must not resolve at all), and **optional** (logged for context only). Fails with an actionable message if a required tool is missing or a forbidden one is found.
- Pass `--debug` to every `limactl` invocation so the workflow log captures the toolchain-detection result, ssh args, path-translation decisions, etc.
- `if: failure()` step dumps `ha.stderr.log`, `ha.stdout.log`, `serial.log`, `lima.yaml`, `ssh.config` from the instance directory.

## What this PR is hoping CI will confirm

1. **`windows-plain-wsl2` passes** end-to-end. Already confirmed in a previous run on this branch.
2. **`windows-plain-qemu` passes** end-to-end for the non-mount path (create / start / shell / copy / stop / delete via native OpenSSH + native sftp-server autodetection).
3. **`windows-wsl2` still passes** with `_LIMA_WINDOWS_EXTRA_PATH` removed. Strict subset of the old environment, but the only way to confirm there is no hidden dependency is to run it.
4. **Logging at `--debug` produces enough trail to diagnose any failure** without local repro. The new debug logs (toolchain detection, in-process gzip, mux-stripping, reverse-sshfs LocalPath, native cygpath fallback, `ParseOpenSSHVersion` fallback) plus the failure-mode hostagent log dump are designed to make any CI failure self-diagnosable.

If items 1–3 fail in interesting ways, those failures themselves are the data we want from this PR. The branch is structured as small focused commits to make per-commit revert easy.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -298,9 +298,21 @@ jobs:
       with:
         go-version: stable
         cache: false
-    - name: Build limactl
-      # Avoid `make` so the build does not require MSYS2 make / bash.
-      run: go build -o _output\bin\limactl.exe .\cmd\limactl
+    - name: Build limactl and Linux/amd64 guest agent
+      # Avoid `make` so the build does not require MSYS2 make / bash. The
+      # guest agent must be built separately because limactl looks it up at
+      # the path _output/share/lima/lima-guestagent.<OS>-<arch> when starting
+      # an instance, and `go build ./cmd/limactl` does not produce that file.
+      run: |
+        $ErrorActionPreference = 'Stop'
+        go build -o _output\bin\limactl.exe .\cmd\limactl
+        if ($LASTEXITCODE -ne 0) { throw "limactl build failed: $LASTEXITCODE" }
+        New-Item -ItemType Directory -Force -Path _output\share\lima | Out-Null
+        $env:CGO_ENABLED = '0'
+        $env:GOOS = 'linux'
+        $env:GOARCH = 'amd64'
+        go build -o _output\share\lima\lima-guestagent.Linux-x86_64 .\cmd\lima-guestagent
+        if ($LASTEXITCODE -ne 0) { throw "lima-guestagent build failed: $LASTEXITCODE" }
     - name: Scrub PATH and verify only native Windows toolchain is reachable
       run: |
         $ErrorActionPreference = 'Stop'
@@ -396,19 +408,25 @@ jobs:
         LIMA_HOME: ${{ runner.temp }}\lima-plain
       run: |
         $ErrorActionPreference = 'Stop'
+        # Wrap each limactl invocation so a non-zero exit aborts the whole
+        # script. Without this, $ErrorActionPreference does not catch native
+        # command failures and the script silently continues past errors.
+        function Invoke-Limactl {
+          & .\_output\bin\limactl.exe --debug @args
+          if ($LASTEXITCODE -ne 0) { throw "limactl $($args -join ' ') exited $LASTEXITCODE" }
+        }
         if (Test-Path $env:LIMA_HOME) { Remove-Item $env:LIMA_HOME -Recurse -Force }
         New-Item -ItemType Directory -Path $env:LIMA_HOME | Out-Null
         # Use the same WSL2 rootfs template that windows-wsl2 uses, so this
         # job exercises a real boot rather than the smallest synthetic config.
-        & .\_output\bin\limactl.exe --debug create --tty=false --name=plain `
-          .\templates\experimental\wsl2.yaml
-        & .\_output\bin\limactl.exe --debug start plain
-        & .\_output\bin\limactl.exe --debug shell --tty=false plain -- uname -srm
+        Invoke-Limactl create --tty=false --name=plain .\templates\experimental\wsl2.yaml
+        Invoke-Limactl start plain
+        Invoke-Limactl shell --tty=false plain -- uname -srm
         'roundtrip' | Out-File -Encoding ascii "$env:LIMA_HOME\rt.txt"
-        & .\_output\bin\limactl.exe --debug copy "$env:LIMA_HOME\rt.txt" plain:/tmp/rt.txt
-        & .\_output\bin\limactl.exe --debug shell --tty=false plain -- cat /tmp/rt.txt
-        & .\_output\bin\limactl.exe --debug stop plain
-        & .\_output\bin\limactl.exe --debug delete --force plain
+        Invoke-Limactl copy "$env:LIMA_HOME\rt.txt" plain:/tmp/rt.txt
+        Invoke-Limactl shell --tty=false plain -- cat /tmp/rt.txt
+        Invoke-Limactl stop plain
+        Invoke-Limactl delete --force plain
     - name: Dump Lima logs on failure
       if: failure()
       env:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -301,45 +301,135 @@ jobs:
     - name: Build limactl
       # Avoid `make` so the build does not require MSYS2 make / bash.
       run: go build -o _output\bin\limactl.exe .\cmd\limactl
-    - name: Smoke test (create / start / shell / copy / stop / delete)
+    - name: Scrub PATH and verify only native Windows toolchain is reachable
       run: |
         $ErrorActionPreference = 'Stop'
+        Write-Host "PATH before scrub:"
+        $env:PATH -split ';' | ForEach-Object { Write-Host "  $_" }
         # Scrub PATH of Cygwin/MSYS2/Git-for-Windows toolchains so that
         # limactl is forced to use native Windows binaries only. The
         # GitHub runner image has these pre-installed and on PATH by
         # default; we want to simulate a vanilla Windows host where
         # users have not installed Git for Windows or MSYS2.
-        $env:PATH = ($env:PATH -split ';' | Where-Object {
+        $scrubbed = ($env:PATH -split ';' | Where-Object {
           $_ -notmatch '\\msys(64|2)\\' -and
           $_ -notmatch '\\Git\\(usr|mingw\d+)\\bin' -and
           $_ -notmatch '\\Git\\cmd' -and
-          $_ -notmatch '\\Git\\bin'
+          $_ -notmatch '\\Git\\bin' -and
+          $_ -notmatch '\\cygwin\d*\\'
         }) -join ';'
+        # Persist the scrubbed PATH into subsequent steps so the smoke test
+        # and any post-step diagnostics see the same environment.
+        Add-Content -Path $env:GITHUB_ENV -Value "PATH=$scrubbed"
+        $env:PATH = $scrubbed
+        Write-Host ""
+        Write-Host "PATH after scrub:"
+        $env:PATH -split ';' | ForEach-Object { Write-Host "  $_" }
+        Write-Host ""
+        # Required: ssh must be the native Windows OpenSSH binary, since
+        # that is the toolchain we are exercising.
         $sshSrc = (Get-Command ssh).Source
         Write-Host "ssh resolves to: $sshSrc"
         if ($sshSrc -notlike 'C:\Windows\System32\OpenSSH\ssh.exe') {
           Write-Error "Expected native Windows OpenSSH; got $sshSrc"
           exit 1
         }
-        if (Get-Command cygpath -ErrorAction SilentlyContinue) {
-          $cyg = (Get-Command cygpath).Source
-          Write-Error "cygpath is still on PATH at $cyg; PATH scrub did not catch it"
+        # Tool inventory: print the resolved location of every external binary
+        # Lima might shell out to on Windows, so the workflow log captures the
+        # exact toolchain layout the test ran against. Tools split into:
+        #   required-native: must come from C:\Windows or be a built-in like wsl.
+        #   forbidden: must NOT resolve at all on a "plain Windows" host.
+        #   optional: may or may not be present; logged for context. Lima has
+        #             graceful fallbacks (id, rsync) or only needs them for
+        #             non-default code paths (qemu-img, decompressors).
+        $required = @('ssh','scp','ssh-keygen','sftp-server','wsl','tar')
+        $forbidden = @{
+          'cygpath' = 'cygpath comes from MSYS2/Git-for-Windows/Cygwin only'
+          'pacman'  = 'pacman comes from MSYS2 only'
+        }
+        # bash.exe is interesting but ambiguous: Windows 10/11 ship a WSL
+        # launcher named bash.exe in System32, which is fine. Git for Windows
+        # also ships a bash.exe in usr/bin. Log it but do not fail on it.
+        $optional = @('bash','sh','gzip','bzip2','xz','zstd','rsync','id','make','awk','sed','qemu-img')
+
+        Write-Host ""
+        Write-Host "=== Required tools (must resolve) ==="
+        $missingRequired = @()
+        foreach ($cmd in $required) {
+          $found = Get-Command $cmd -ErrorAction SilentlyContinue
+          if ($found) { Write-Host ("  {0,-15} -> {1}" -f $cmd, $found.Source) }
+          else { Write-Host ("  {0,-15} -> NOT FOUND" -f $cmd); $missingRequired += $cmd }
+        }
+        Write-Host ""
+        Write-Host "=== Forbidden tools (must not resolve on plain Windows) ==="
+        $foundForbidden = @()
+        foreach ($cmd in $forbidden.Keys) {
+          $found = Get-Command $cmd -ErrorAction SilentlyContinue
+          if ($found) {
+            Write-Host ("  {0,-15} -> {1}    (FORBIDDEN: {2})" -f $cmd, $found.Source, $forbidden[$cmd])
+            $foundForbidden += $cmd
+          } else {
+            Write-Host ("  {0,-15} -> not found (good)" -f $cmd)
+          }
+        }
+        Write-Host ""
+        Write-Host "=== Optional tools (logged for context, not asserted) ==="
+        foreach ($cmd in $optional) {
+          $found = Get-Command $cmd -ErrorAction SilentlyContinue
+          if ($found) { Write-Host ("  {0,-15} -> {1}" -f $cmd, $found.Source) }
+          else { Write-Host ("  {0,-15} -> not found" -f $cmd) }
+        }
+        Write-Host ""
+        if ($missingRequired.Count -gt 0) {
+          Write-Error ("Required tools not on PATH: " + ($missingRequired -join ', '))
           exit 1
         }
-        $env:LIMA_HOME = "$env:TEMP\lima-plain"
+        if ($foundForbidden.Count -gt 0) {
+          Write-Error ("Forbidden tools still on PATH: " + ($foundForbidden -join ', ') + ". Update the scrub regex in this step.")
+          exit 1
+        }
+    - name: Smoke test (create / start / shell / copy / stop / delete)
+      env:
+        # Surface decisions in path-translation, ssh-args, mux-stripping, etc.
+        # The Debug-level logging is only useful when this job fails, but the
+        # success-case noise is small and CI logs are kept for a long time.
+        LIMA_HOME: ${{ runner.temp }}\lima-plain
+      run: |
+        $ErrorActionPreference = 'Stop'
         if (Test-Path $env:LIMA_HOME) { Remove-Item $env:LIMA_HOME -Recurse -Force }
         New-Item -ItemType Directory -Path $env:LIMA_HOME | Out-Null
         # Use the same WSL2 rootfs template that windows-wsl2 uses, so this
         # job exercises a real boot rather than the smallest synthetic config.
-        & .\_output\bin\limactl.exe create --tty=false --name=plain `
+        & .\_output\bin\limactl.exe --debug create --tty=false --name=plain `
           .\templates\experimental\wsl2.yaml
-        & .\_output\bin\limactl.exe start plain
-        & .\_output\bin\limactl.exe shell --tty=false plain -- uname -srm
+        & .\_output\bin\limactl.exe --debug start plain
+        & .\_output\bin\limactl.exe --debug shell --tty=false plain -- uname -srm
         'roundtrip' | Out-File -Encoding ascii "$env:LIMA_HOME\rt.txt"
-        & .\_output\bin\limactl.exe copy "$env:LIMA_HOME\rt.txt" plain:/tmp/rt.txt
-        & .\_output\bin\limactl.exe shell --tty=false plain -- cat /tmp/rt.txt
-        & .\_output\bin\limactl.exe stop plain
-        & .\_output\bin\limactl.exe delete --force plain
+        & .\_output\bin\limactl.exe --debug copy "$env:LIMA_HOME\rt.txt" plain:/tmp/rt.txt
+        & .\_output\bin\limactl.exe --debug shell --tty=false plain -- cat /tmp/rt.txt
+        & .\_output\bin\limactl.exe --debug stop plain
+        & .\_output\bin\limactl.exe --debug delete --force plain
+    - name: Dump Lima logs on failure
+      if: failure()
+      env:
+        LIMA_HOME: ${{ runner.temp }}\lima-plain
+      run: |
+        # Hostagent logs are the most useful artefact when start fails.
+        # Print everything Lima wrote about the instance so failures show up
+        # in the workflow run instead of being lost when the runner is gone.
+        $instDir = "$env:LIMA_HOME\plain"
+        if (-not (Test-Path $instDir)) {
+          Write-Host "No instance directory at $instDir, nothing to dump."
+          exit 0
+        }
+        Get-ChildItem $instDir | Format-Table Name, Length, LastWriteTime
+        foreach ($name in 'ha.stdout.log','ha.stderr.log','serial.log','lima.yaml','ssh.config') {
+          $p = Join-Path $instDir $name
+          if (Test-Path $p) {
+            Write-Host ("===== {0} =====" -f $p)
+            Get-Content $p
+          }
+        }
 
   qemu:
     name: "Integration tests (QEMU, macOS host)"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -218,11 +218,10 @@ jobs:
       run: |
         $env:PATH = "$pwd\_output\bin;" + 'C:\msys64\usr\bin;' + $env:PATH
         pacman -Sy --noconfirm openbsd-netcat diffutils socat w3m
-        $env:MSYS2_ENV_CONV_EXCL = 'HOME_HOST;HOME_GUEST;_LIMA_WINDOWS_EXTRA_PATH'
+        $env:MSYS2_ENV_CONV_EXCL = 'HOME_HOST;HOME_GUEST'
         $env:HOME_HOST = $(cygpath.exe "$env:USERPROFILE")
         $env:HOME_GUEST = "/mnt$env:HOME_HOST"
         $env:LIMACTL_CREATE_ARGS = '--vm-type=wsl2 --mount-type=wsl2 --containerd=system'
-        $env:_LIMA_WINDOWS_EXTRA_PATH = 'C:\Program Files\Git\usr\bin'
         bash.exe -c "./hack/test-templates.sh templates/experimental/wsl2.yaml"
 
   windows-qemu:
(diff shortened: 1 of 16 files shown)


jandubois:
For the record, the round 1 review mentioned in the summary is https://jandubois.github.io/lima/20260424-190657-pr-4885.html

liketosweep:
Thanks for the solid foundation here, @jandubois! I’ll be testing and reviewing this locally on Windows to help provide a status signal for the research.

I’m also volunteering to pick up the 'pre-existing limitations' for #4819 as follow-up PRs. Specifically, I'm exploring a pure-Go `rsync` implementation; this would not only resolve the binary dependency but also naturally solve the requirements in #4887 as a direct subset.

Excited to help move this toward the finish line!

liketosweep:
@jandubois - I just tested it locally and here is  the report . 

Verification Report: Windows Native Path Translation

---

Test Environment:

Host OS: Windows (Native)

Isolation: Isolated `$env:PATH` (`Git` ,` MSYS2` , and `Mingw` references removed).

Dependency Check: `ssh` verified as native `OpenSSH; cygpath` confirmed unavailable via Get-Command.

---

Methodology:
Compiled `limactl` from the `windows-native-ssh-no-cygpath` branch and executed a "clean-room" start of a WSL2 instance using ` .\limactl-test.exe start --debug` .

---

Key Findings:
The native path translation logic is verified. The `sshutil ` package successfully caught the missing `cygpath`  binary and triggered the pure-Go fallback logic.

---

Captured Log Evidence:

`level=debug msg="cygpath unavailable for \"C:\\\\Users\\\\lts\", attempting native conversion" error="exec: \"cygpath\": executable file not found in %PATH%" `
---
`level=debug msg="native cygpath fallback: \"C:\\\\Users\\\\lts\" -> \"/c/Users/lts\"" `
---

Conclusion:
The host-side CLI logic successfully abstracts away the dependency on external Windows path-translation tools. This confirms the architectural viability for a standalone Windows experience.

jandubois:
Thanks for the feedback @liketosweep!

I would like to hear from other maintainers, and from @arixmkii if this PR looks sensible in principle, and if I should spend the effort on cleaning it up and creating one or more actual mergeable PRs from it. Probably won't have time for it for a little bit, but once I know we want to do this, I can work on it when time opens up.

liketosweep:
Thank you, @jandubois. I completely agree with holding off until there is consensus from the other maintainers and @arixmkii on the architectural direction. I will monitor the discussion here and remain available if further local testing is needed once a decision is reached.

unsuman:
I tested the PR on my Windows Machine(Windows 11 Home 25H2). No MSYS2, no Cygwin, no Git for Windows:
```
PS C:\Windows\system32> Get-Command ssh.exe

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     ssh.exe                                            9.5.5.1    C:\Windows\System32\OpenSSH\ssh.exe


PS C:\Windows\system32> Get-Command ssh-keygen.exe

CommandType     Name                                               Version    Source
-----------     ----                                               -------    ------
Application     ssh-keygen.exe                                     9.5.5.1    C:\Windows\System32\OpenSSH\ssh-keygen.exe

```

### Here are some key observations(We'll have to see if they are universal):
- Source build fails to parse YAML even if when explicit template is given. 
```
  PS> C:\Users\anshu\OneDrive\Documents\lima\_output\bin\limactl.exe start --vm-type qemu C:\Users\anshu\OneDrive\Documents\lima\_output\share\lima\templates\default.yaml
  
time="..." level=fatal msg="failed to unmarshal YAML (template:_images/ubuntu): [1:1] string was used where mapping is expected\n>  1 | ubuntu-25.10.yaml\n       ^\n"
```
  The same binary works perfectly when placed into `C:\Program Files\Lima\bin` (overwriting the official `lima-2.1.1-Windows-AMD64` installation) and run from there.
- `boot.sh` execution failure on the guest side which was causing Lima to get stuck after `user session is ready for ssh`: 
   Used AI and it told me to add a `.gitattributes` entry which forces LF line endings for these files and it worked:
   ```
   pkg/cidata/cidata.TEMPLATE.d/** text eol=lf
   ```
   
 Once these are out of the way, I can finally run a QEMU instance:
  
<details><summary>QEMU log</summary>
<p>

```
PS C:\Windows\system32> limactl start --vm-type qemu
time="2026-04-29T16:30:19+05:30" level=warning msg="local username \"UNSUMAN\\\\anshu\" is not a valid Linux username (must match \"^[a-z_][a-z0-9_-]*$\"); using \"lima\" instead"
time="2026-04-29T16:30:19+05:30" level=warning msg="local uid \"S-1-5-21-2734909792-918164565-3955868323-1001\" is not a valid Linux uid (must be integer); using 1000 uid instead"
time="2026-04-29T16:30:19+05:30" level=warning msg="local gid \"S-1-5-21-2734909792-918164565-3955868323-1001\" is not a valid Linux gid (must be integer); using 1000 gid instead"
time="2026-04-29T16:30:19+05:30" level=info msg="Starting the instance \"default\" with internal VM driver \"qemu\""
time="2026-04-29T16:30:19+05:30" level=info msg="Attempting to download the image" arch=x86_64 digest="sha256:6d91bd33bee09ea9907914136a9ac73c789611059c3ccb8cec894ea6e2d4b245" location="https://cloud-images.ubuntu.com/releases/questing/release-20260320/ubuntu-25.10-server-cloudimg-amd64.img"
time="2026-04-29T16:30:20+05:30" level=info msg="Using cache \"C:\\\\Users\\\\anshu\\\\AppData\\\\Local\\\\lima\\\\download\\\\by-url-sha256\\\\dfc8a98829aa2a5e8524e44fa3208eeed3cfad82f87187dc67b735876f515501\\\\data\""
time="2026-04-29T16:30:20+05:30" level=info msg="Resize instance default's disk from 3.5GiB to 100GiB"
time="2026-04-29T16:30:20+05:30" level=info msg="Attempting to download the nerdctl archive" arch=x86_64 digest="sha256:8a477f35533c6cc1120c19558d8142967c74f25a4b952b481f48104e030de914" location="https://github.com/containerd/nerdctl/releases/download/v2.2.2/nerdctl-full-2.2.2-linux-amd64.tar.gz"
time="2026-04-29T16:30:20+05:30" level=info msg="Using cache \"C:\\\\Users\\\\anshu\\\\AppData\\\\Local\\\\lima\\\\download\\\\by-url-sha256\\\\072a3470f807a0089a48b29fff33dc8b5d3800fc7761ae82e8b55a6b3ce2ace1\\\\data\""
time="2026-04-29T16:30:21+05:30" level=info msg="[hostagent] hostagent socket created at C:\\Users\\anshu\\.lima\\default\\ha.sock"
time="2026-04-29T16:30:23+05:30" level=info msg="[hostagent] Using system firmware (\"C:\\\\Program Files\\\\qemu\\\\share\\\\edk2-x86_64-code.fd\")"
time="2026-04-29T16:30:23+05:30" level=info msg="[hostagent] Using system firmware vars (\"C:\\\\Program Files\\\\qemu\\\\share\\\\edk2-i386-vars.fd\")"
time="2026-04-29T16:30:23+05:30" level=info msg="[hostagent] Starting QEMU (hint: to watch the boot progress, see \"C:\\\\Users\\\\anshu\\\\.lima\\\\default\\\\serial*.log\")"
time="2026-04-29T16:30:23+05:30" level=info msg="SSH Local Port: 5706"
time="2026-04-29T16:30:23+05:30" level=info msg="[hostagent] Waiting for the essential requirement 1 of 4: \"ssh\""
time="2026-04-29T16:30:42+05:30" level=info msg="[hostagent] The essential requirement 1 of 4 is satisfied"
time="2026-04-29T16:30:42+05:30" level=info msg="[hostagent] Waiting for the essential requirement 2 of 4: \"user session is ready for ssh\""
time="2026-04-29T16:30:42+05:30" level=info msg="[hostagent] The essential requirement 2 of 4 is satisfied"
time="2026-04-29T16:30:42+05:30" level=info msg="[hostagent] Waiting for the essential requirement 3 of 4: \"sshfs binary to be installed\""
time="2026-04-29T16:30:51+05:30" level=info msg="[hostagent] The essential requirement 3 of 4 is satisfied"
time="2026-04-29T16:30:51+05:30" level=info msg="[hostagent] Waiting for the essential requirement 4 of 4: \"fuse to \\\"allow_other\\\" as user\""
time="2026-04-29T16:30:51+05:30" level=info msg="[hostagent] The essential requirement 4 of 4 is satisfied"
time="2026-04-29T16:30:51+05:30" level=info msg="[hostagent] Mounting \"C:\\\\Users\\\\anshu\" on \"/c/Users/anshu\""
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] fusermount3: mount failed: Permission denied"
time="2026-04-29T16:31:23+05:30" level=warning msg="[hostagent] failed to confirm whether /c/Users/anshu [remote] is successfully mounted" error="failed to execute script \"wait-for-remote-ready\": stdout=\"\", stderr=\"sshfs does not seem to be mounted on /c/Users/anshu\\n\": exit status 1"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Waiting for the optional requirement 1 of 2: \"systemd must be available\""
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Guest agent is running"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Time sync: guest agent is alive, starting time synchronization"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Not forwarding TCP 0.0.0.0:22"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Not forwarding TCP 127.0.0.53:53"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Not forwarding TCP 127.0.0.54:53"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Forwarding TCP from 127.0.0.1:35961 to 127.0.0.1:35961"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Not forwarding TCP [::]:22"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Not forwarding UDP 127.0.0.54:53"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Not forwarding UDP 127.0.0.53:53"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Not forwarding UDP 192.168.5.15:68"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Forwarding UDP from 127.0.0.1:323 to 127.0.0.1:323"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Forwarding UDP from 0.0.0.0:5353 to 127.0.0.1:5353"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Forwarding UDP from [::1]:323 to 127.0.0.1:323"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Forwarding UDP from [::]:5353 to 127.0.0.1:5353"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] The optional requirement 1 of 2 is satisfied"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Waiting for the optional requirement 2 of 2: \"containerd binaries to be installed\""
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] The optional requirement 2 of 2 is satisfied"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Waiting for the guest agent to be running"
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] Waiting for the final requirement 1 of 1: \"boot scripts must have finished\""
time="2026-04-29T16:31:23+05:30" level=info msg="[hostagent] The final requirement 1 of 1 is satisfied"
time="2026-04-29T16:31:23+05:30" level=info msg="READY. Run `lima` to open the shell."
```

</p>
</details> 

For wsl2, using the `experimental/wsl2.yaml`, I get this error:
`{"level":"debug","msg":"[wsl.exe -d lima-wsl2 bash -c /mnt/c/Users/anshu/AppData/Local/Temp/lima-wsl2-boot-171555286.sh]: \"bash: line 1: /mnt/c/Users/anshu/AppData/Local/Temp/lima-wsl2-boot-171555286.sh: cannot execute: required file not found\\n\""}`

.github/workflows/test.yml
```diff
@@ -253,13 +252,427 @@ jobs:
...
+        $ErrorActionPreference = 'Stop'
+        go build -o _output\bin\limactl.exe .\cmd\limactl
+        if ($LASTEXITCODE -ne 0) { throw "limactl build failed: $LASTEXITCODE" }
+        New-Item -ItemType Directory -Force -Path _output\share\lima | Out-Null
+        $env:CGO_ENABLED = '0'
+        $env:GOOS = 'linux'
+        $env:GOARCH = 'amd64'
+        go build -o _output\share\lima\lima-guestagent.Linux-x86_64 .\cmd\lima-guestagent
+        if ($LASTEXITCODE -ne 0) { throw "lima-guestagent build failed: $LASTEXITCODE" }
+    - name: Scrub PATH and verify only native Windows toolchain is reachable
```
AkihiroSuda, line 316: Can't we just uninstall Cygwin/MSYS2/Git-for-Windows from the CI env?

jandubois:
Superseded by #4998

PR #4886, opened 2026-04-25
docs: add limactl help yq concept guide to explain yqlib limitations

### LINKED ISSUE #4531: `limactl help` should explain the limitation of the YQ expression support
- https://github.com/lima-vm/lima/pull/4405#issuecomment-3757864260

> [...]
> > * Add a document to explain our "dialect" of yq language
> 
> I'm struggling to find a good place in the documentation for this. Like a place users would actually find.
> 
> I feel like we also need more documentation for the `create --set` option, including examples. This could go in the `limactl create --help` output, and could then mention the limitation on env and file operators in our YQ dialect. Other options could then point to it.
> 
> This might still be a bit unwieldy, so maybe we should start having "concept guides" the way `git` does it:
> 
> ```console
> ❯ git help -g
> The Git concept guides are:
>    core-tutorial    A Git core tutorial for developers
>    credentials      Providing usernames and passwords to Git
>    cvs-migration    Git for CVS users
>    diffcore         Tweaking diff output
>    everyday         A useful minimum set of commands for Everyday Git
>    faq              Frequently asked questions about using Git
>    glossary         A Git Glossary
>    namespaces       Git namespaces
>    remote-helpers   Helper programs to interact with remote repositories
>    submodules       Mounting one repository inside another
>    tutorial         A tutorial introduction to Git
>    tutorial-2       A tutorial introduction to Git: part two
>    workflows        An overview of recommended workflows with Git
> 
> 'git help -a' and 'git help -g' list available subcommands and some
> concept guides. See 'git help <command>' or 'git help <concept>'
> to read about a specific subcommand or concept.
> See 'git help git' for an overview of the system.
> ```
> 
> There is also `kubectl options` as another precedent; not sure if it has other meta commands.
> 
> Anyways, this is quickly becoming out-of-scope for this PR and should be in a separate docs PR.

_Originally posted by @jandubois in https://github.com/lima-vm/lima/issues/4405#issuecomment-3757864260_
            
- - -
- https://github.com/lima-vm/lima/pull/4405#issuecomment-3758108539

> [...]
> > concept guides
>
> This can be probably implemented using https://pkg.go.dev/github.com/spf13/cobra#Command.IsAdditionalHelpTopicCommand

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/issues/4405#issuecomment-3758108539_

Mysterio-17:
### Description
Addresses the missing documentation for the Lima `yq` dialect by adding a concept guide that clarifies which `yq` operators are disabled for security reasons.

### Changes
- Added a `limactl help yq` command to document unsupported `yq` operators (`env`, `load`, `load_str`).
- Integrated `limactl yq` into the root command so it appears natively under "Additional help topics".
- Updated the `--set` flag description (used in `limactl create`, etc.) to explicitly point users to `limactl help yq`.

### Testing
- `limactl help`: Verified `yq` is listed under "Additional help topics".
- `limactl help yq`: Verified it correctly prints the new concept guide.
- `limactl create --help`: Verified the `--set` flag correctly references `limactl help yq`.
- `go test ./cmd/limactl/...`: Tests passed successfully with no regressions.

### DIFF
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -68,7 +68,7 @@ func RegisterEdit(cmd *cobra.Command, commentPrefix string) {
 
 	flags.Bool("rosetta", false, commentPrefix+"Enable Rosetta (for vz instances)")
 
-	flags.StringArray("set", []string{}, commentPrefix+"Modify the template inplace, using yq syntax. Can be passed multiple times.")
+	flags.StringArray("set", []string{}, commentPrefix+"Modify the template inplace, using yq syntax. Can be passed multiple times. See 'limactl help yq-restrictions' for limitations.")
 	flags.StringArray("param", []string{}, commentPrefix+"Set a template parameter, e.g. name=value. Can be passed multiple times.")
 
 	flags.Uint16("ssh-port", 0, commentPrefix+"SSH port (0 for random)") // colima-compatible
--- a/cmd/limactl/yq_restrictions.go
+++ b/cmd/limactl/yq_restrictions.go
@@ -1,0 +1,35 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func newYQRestrictionsHelpCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "yq-restrictions",
+		Short: "Restrictions on yq expressions in Lima",
+		Long: `Lima uses yq (v4) syntax for the --set flag and provision mode "yq".
+
+Lima embeds yqlib (https://github.com/mikefarah/yq) as a library and
+disables several operators to prevent template expressions from reading
+the host environment or executing arbitrary commands:
+
+  Disabled by Lima:
+    - env               (environment variable access)
+    - load, load_str    (arbitrary file reads)
+
+  Disabled by yqlib default:
+    - system            (arbitrary command execution)
+
+These restrictions exist because --set expressions and provision.yq
+expressions may come from untrusted template files. Allowing them to
+access environment variables, read files, or execute commands on the
+host would be a security risk.
+
+For full yq v4 expression syntax, see:
+  https://mikefarah.gitbook.io/yq/`,
+	}
+}
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -210,6 +210,7 @@ func newApp() *cobra.Command {
 		newCloneCommand(),
 		newRenameCommand(),
 		newWatchCommand(),
+		newYQRestrictionsHelpCommand(),
 	)
 	addPluginCommands(rootCmd)
 
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -57,6 +57,7 @@ func newCreateCommand() *cobra.Command {
 
   To create an instance "default" with yq expressions:
   $ limactl create --set='.cpus = 2 | .memory = "2GiB"'
+  Note: Some yq operators are restricted. See 'limactl help yq-restrictions'.
 
   To create an instance "default" with a template parameter:
   $ limactl create --name=default --param containerdSnapshotter=false template:docker
--- a/hack/bats/tests/yq.bats
+++ b/hack/bats/tests/yq.bats
@@ -47,3 +47,8 @@ load "../helpers/load"
     run_e -1 limactl yq -n --security-disable-env-ops 'env(FOO)'
     assert_stderr "Error: env operations have been disabled"
 }
+
+@test 'yq multi-call command has system operations disabled by default' {
+    run -1 limactl yq -n 'system("echo hello")'
+    assert_output --partial "system operations are disabled"
+}


cmd/limactl/yq.go
```diff
@@ -0,0 +1,22 @@
...
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func newYQHelpCommand() *cobra.Command {
+	return &cobra.Command{
+		Use:   "yq",
```
AkihiroSuda: This seems to conflict with the existing `limactl yq` command

Mysterio-17: I missed that limactl yq is intercepted as a multicall binary. I've just pushed an update renaming the concept guide to yq-dialect, and I've updated the --set flag description to point there.

hack/cache-common-inc.sh
```diff
@@ -150,7 +150,7 @@ function check_location_with_cache() {
 # ```
 function check_location_without_cache() {
 	local -r location="$1"
-	curl -sIL -w "%{http_code} %header{Content-Length}" "${location}" -o /dev/null
+	curl --retry 5 --retry-connrefused -sIL -w "%{http_code} %header{Content-Length}" "${location}" -o /dev/null
```
unsuman: This seems unrelated to the issue

cmd/limactl/editflags/editflags.go
```diff
@@ -68,7 +68,7 @@ func RegisterEdit(cmd *cobra.Command, commentPrefix string) {
 
 	flags.Bool("rosetta", false, commentPrefix+"Enable Rosetta (for vz instances)")
 
-	flags.StringArray("set", []string{}, commentPrefix+"Modify the template inplace, using yq syntax. Can be passed multiple times.")
+	flags.StringArray("set", []string{}, commentPrefix+"Modify the template inplace, using yq syntax. See 'limactl help yq-dialect' for details. Can be passed multiple times.")
```
unsuman:
```suggestion
	flags.StringArray("set", []string{}, commentPrefix+"Modify the template inplace, using yq syntax. Can be passed multiple times. See 'limactl help yq-dialect' for limitations.")
```

unsuman (commented):
Looks good, please squash the commits!

unsuman (dismissed):
Thanks, LGTM!

jandubois:
I think I would prefer `yq-restrictions` to `yq-dialect`. We don't use a different dialect, we use regular `yq` syntax, with just a few operators disabled.

I wonder if we should also mention that the new system operations are disabled as well? Of course they are also disabled by `yq` itself by default, but may be worth mentioning for completeness.

```console
❯ limactl yq -n 'system("ls")'
Error: system operations are disabled, use --security-enable-system-operator to enable
```

We probably should have a BATS tests to ensure we don't enable it by accident.

Some other observations:

* restricted operations are now listed in 2 files, which is a maintenance issue: `cmd/limactl/yq_dialect.go` and `cmd/limactl/start.go`.

* `--set` works on `create`, `edit`, and `start`, but the long help is only added in `create`. Maybe it should be shared (not duplicated) by all relevant subcommands.

* Maybe the explanation should include not just the "what" but also "why" the operators have been restricted?

* I think the `yq-restrictions` output should include a link to the upstream `yq` docs, explaining that we are using `yq` version 4 syntax.

Mysterio-17:
> I think I would prefer `yq-restrictions` to `yq-dialect`. We don't use a different dialect, we use regular `yq` syntax, with just a few operators disabled.
> 
> I wonder if we should also mention that the new system operations are disabled as well? Of course they are also disabled by `yq` itself by default, but may be worth mentioning for completeness.
> 
> ```
> ❯ limactl yq -n 'system("ls")'
> Error: system operations are disabled, use --security-enable-system-operator to enable
> ```
> 
> We probably should have a BATS tests to ensure we don't enable it by accident.
> 
> Some other observations:
> 
> * restricted operations are now listed in 2 files, which is a maintenance issue: `cmd/limactl/yq_dialect.go` and `cmd/limactl/start.go`.
> * `--set` works on `create`, `edit`, and `start`, but the long help is only added in `create`. Maybe it should be shared (not duplicated) by all relevant subcommands.
> * Maybe the explanation should include not just the "what" but also "why" the operators have been restricted?
> * I think the `yq-restrictions` output should include a link to the upstream `yq` docs, explaining that we are using `yq` version 4 syntax.


Hi @jandubois , I've made the changes:-
- Renamed to `yq-restrictions` — makes more sense since it's regular yq with some ops disabled, not a different dialect.
- Mentioned `system` as well (disabled by yqlib default), separated from the ones Lima explicitly disables (`env`, `load`, `load_str`).
- Added a BATS test for `system()` so we catch it if it ever gets enabled by accident.
- Moved all restriction details into `limactl help yq-restrictions` as the single source — the `--set` flag and `create` example just reference it now, no duplication.
- Included the "why" (untrusted templates) and a link to the upstream yq v4 docs.

Mysterio-17:
> > I think I would prefer `yq-restrictions` to `yq-dialect`. We don't use a different dialect, we use regular `yq` syntax, with just a few operators disabled.
> > I wonder if we should also mention that the new system operations are disabled as well? Of course they are also disabled by `yq` itself by default, but may be worth mentioning for completeness.
> > ```
> > ❯ limactl yq -n 'system("ls")'
> > Error: system operations are disabled, use --security-enable-system-operator to enable
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > We probably should have a BATS tests to ensure we don't enable it by accident.
> > Some other observations:
> > 
> > * restricted operations are now listed in 2 files, which is a maintenance issue: `cmd/limactl/yq_dialect.go` and `cmd/limactl/start.go`.
> > * `--set` works on `create`, `edit`, and `start`, but the long help is only added in `create`. Maybe it should be shared (not duplicated) by all relevant subcommands.
> > * Maybe the explanation should include not just the "what" but also "why" the operators have been restricted?
> > * I think the `yq-restrictions` output should include a link to the upstream `yq` docs, explaining that we are using `yq` version 4 syntax.
> 
> Hi @jandubois , I've made the changes:-
> 
> * Renamed to `yq-restrictions` — makes more sense since it's regular yq with some ops disabled, not a different dialect.
> * Mentioned `system` as well (disabled by yqlib default), separated from the ones Lima explicitly disables (`env`, `load`, `load_str`).
> * Added a BATS test for `system()` so we catch it if it ever gets enabled by accident.
> * Moved all restriction details into `limactl help yq-restrictions` as the single source — the `--set` flag and `create` example just reference it now, no duplication.
> * Included the "why" (untrusted templates) and a link to the upstream yq v4 docs.

Kindly, have a look at this @unsuman.

unsuman (dismissed):

unsuman:
@jandubois PTAL

cmd/limactl/yq_restrictions.go
```diff
@@ -0,0 +1,35 @@
...
+		Use:   "yq-restrictions",
+		Short: "Restrictions on yq expressions in Lima",
+		Long: `Lima uses yq (v4) syntax for the --set flag and provision mode "yq".
+
+Lima embeds yqlib (https://github.com/mikefarah/yq) as a library and
+disables several operators to prevent template expressions from reading
+the host environment or executing arbitrary commands:
+
+  Disabled by Lima:
+    - env, envOp        (environment variable access)
```
jandubois: `envOp` is an internal `yqlib` term; it is not user-facing.

```suggestion
    - env               (environment variable access)
```

jandubois (commented):
Thanks, LGTM, but I would still prefer that `envOp` is removed

Mysterio-17:
Sure, @jandubois I've removed `envOp`. Have a look @unsuman

jandubois:
@Mysterio-17 Please squash commits

Mysterio-17:
> @Mysterio-17 Please squash commits

Done

jandubois (approved):
Thanks, LGTM

PR #4889, opened 2026-04-26
hostagent: close stale GuestAgentClient on reconnect to stop ClientConn leak

### LINKED ISSUE #4888: hostagent: GuestAgentClient leaks gRPC ClientConn (goroutines + FDs) on every guest-agent reconnect
### Description

`HostAgent.getOrCreateClient` (`pkg/hostagent/hostagent.go`) replaces `a.client` with a freshly constructed `*guestagentclient.GuestAgentClient` whenever the previous client's socket is no longer reachable, but it never closes the previous client. `GuestAgentClient` itself (`pkg/guestagent/api/client/client.go`) discards the `*grpc.ClientConn` returned by `grpc.NewClient` — only the generated stub is retained — so the caller has no way to release it either.

Each replacement therefore leaks:

- the underlying `*grpc.ClientConn` and all of its background goroutines (resolver watcher, pick-first balancer, channelz, HTTP/2 reader/keepalive/controlBuffer),
- the `net.Conn` returned by `HostAgent.createConnection` — i.e. an open unix-socket connection to the forwarded `ga.sock` (or the driver's vsock),
- any in-flight `Events`/`Inotify` server-streaming RPCs bound to that connection.

### Trigger (deterministic)

The reconnect path runs on the hot loop in `watchGuestAgentEvents`:

1. `Info(ctx)` against the previous client returns `Unavailable` whenever the guest agent restarts or the VM reboots.
2. `isGuestAgentSocketAccessible` → false.
3. `getOrCreateClient` constructs a new `GuestAgentClient`; the previous one becomes unreachable without `Close()` ever being called.

Reproducer on any Linux guest:

```bash
limactl start default
HA_PID=$(pgrep -f "lima-hostagent.*default" | head -1)
ls /proc/$HA_PID/fd | wc -l                         # baseline FDs
grep Threads /proc/$HA_PID/status                   # baseline threads

for i in $(seq 1 50); do
  limactl shell default sudo systemctl restart lima-guestagent
  sleep 12   # one full watchGuestAgentEvents iteration
done

ls /proc/$HA_PID/fd | wc -l                         # grows ~linearly
grep Threads /proc/$HA_PID/status                   # grows ~linearly
```

Growth is exactly linear in the number of reconnects.

### Impact

- Long-running Lima instances (Colima / Rancher Desktop / Finch use a single Lima VM for the entire user session) accumulate goroutines and FDs across guest reboots, suspend/resume cycles, and guest-agent restarts.
- Eventually `RLIMIT_NOFILE` is hit and subsequent SSH forward / mount setup / `os.OpenFile` calls inside the same hostagent fail with opaque "failed to set up forward" errors.
- Stale stream consumers from the orphaned `ClientConn` can still fire buffered `OnEvent` callbacks, racing with the new client against `a.grpcPortForwarder` (duplicate port-forward setup).

### Fix

Two minimal changes (PR follows):

1. `GuestAgentClient` retains the `*grpc.ClientConn` and exposes `Close()`.
2. `getOrCreateClient` calls `Close()` on the previous client before overwriting `a.client`. A `cleanUp` is also registered in `watchGuestAgentEvents` so the live client is closed on shutdown.

### Environment

Found by code inspection against `master` (commit `5a20c6d3`). Affects all Linux guests with the lima-guestagent daemon (i.e. the default configuration).

mn-ram:
## What

- `GuestAgentClient` now retains `*grpc.ClientConn` and exposes `Close()`.
- `HostAgent.getOrCreateClient` closes the previous `a.client` before overwriting it.
- A `cleanUp` is registered in `watchGuestAgentEvents` so the live client is closed on hostagent shutdown.

## Why

Fixes #4888.

Every time the guest agent restarts or the VM reboots, `getOrCreateClient` overwrites `a.client` with a fresh `GuestAgentClient` whose `*grpc.ClientConn` was never reachable to any caller. The previous `ClientConn` and its goroutines (resolver, balancer, HTTP/2 transport) plus the dialed `net.Conn` to the forwarded `ga.sock` leak permanently. On long-running Lima instances (Colima / Rancher Desktop / Finch) this grows unbounded with reconnect count and eventually exhausts the FD limit.

## Test plan

- `go build ./...`
- `go vet ./pkg/guestagent/... ./pkg/hostagent/...`
- `go test ./pkg/guestagent/... ./pkg/hostagent/...`
- Manual: loop `limactl shell default sudo systemctl restart lima-guestagent` 50× and watch `/proc/<hostagent-pid>/status` Threads count and `ls /proc/<hostagent-pid>/fd | wc -l` — both flat after this PR, both monotonically increasing on `master`.

### DIFF
--- a/pkg/guestagent/api/client/client.go
+++ b/pkg/guestagent/api/client/client.go
@@ -18,7 +18,8 @@ import (
 )
 
 type GuestAgentClient struct {
-	cli api.GuestServiceClient
+	cli  api.GuestServiceClient
+	conn *grpc.ClientConn
 }
 
 func NewGuestAgentClient(dialFn func(ctx context.Context) (net.Conn, error)) (*GuestAgentClient, error) {
@@ -44,8 +45,18 @@ func NewGuestAgentClient(dialFn func(ctx context.Context) (net.Conn, error)) (*G
 	}
 	client := api.NewGuestServiceClient(clientConn)
 	return &GuestAgentClient{
-		cli: client,
+		cli:  client,
+		conn: clientConn,
 	}, nil
+}
+
+// Close releases the underlying gRPC connection. It is safe to call on a nil
+// receiver or a client whose connection has already been closed.
+func (c *GuestAgentClient) Close() error {
+	if c == nil || c.conn == nil {
+		return nil
+	}
+	return c.conn.Close()
 }
 
 func (c *GuestAgentClient) Info(ctx context.Context) (*api.Info, error) {
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -728,7 +728,8 @@ func (a *HostAgent) watchGuestAgentEvents(ctx context.Context) {
 
 	go func() {
 		if a.instConfig.MountInotify != nil && *a.instConfig.MountInotify {
-			if a.client == nil || !isGuestAgentSocketAccessible(ctx, a.client) {
+			client := a.getClient()
+			if client == nil || !isGuestAgentSocketAccessible(ctx, client) {
 				if a.driver.ForwardGuestAgent() {
 					sshAddress, sshPort := a.sshAddressPort()
 					_ = forwardSSH(ctx, a.sshConfig, sshAddress, sshPort, localUnix, remoteUnix, verbForward, false)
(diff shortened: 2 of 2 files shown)


jandubois:
I did re-run the failed CI test, it was just flaky. The PR looks good, but has a potential race condition; see AI review at https://jandubois.github.io/lima/20260426-165750-pr-4889.html

mn-ram:
Thanks @jandubois for the review!

Good catch — setting `a.client = nil` introduced a race with unsynchronized reads.

Fixed in `fae74fb` by adding `getClient()` (locked via `clientMu`) and updating call sites.

Build, vet, and tests pass locally.

Happy to squash commits if preferred.

unsuman:
> Happy to squash commits if preferred.

Yes please!

mn-ram:
> > Happy to squash commits if preferred.
> 
> Yes please!

Done

AkihiroSuda:
> the [v2.2.0](https://github.com/lima-vm/lima/milestone/70) milestone

This PR seems safe to merge in v2.1.2?

unsuman (approved):
Thanks

PR #4892, opened 2026-04-27
wsl2: stop spinning in for { <-ctx.Done() } after instance stop

### LINKED ISSUE #4891: wsl2: provisionVM goroutine hot-loops on for { <-ctx.Done() } after instance stop
### Description

The boot-script goroutine spawned by `provisionVM` in `pkg/driver/wsl2/vm_windows.go` (lines 116–143) wraps a one-shot context-cancellation wait in an infinite `for {}` loop:

```go
go func() {
    cmd := exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "bash", "-c", limaBootFileLinuxPath)
    out, err := cmd.CombinedOutput()
    os.RemoveAll(limaBootFileWinPath)
    ...
    for {
        <-ctx.Done()
        logrus.Info("Context closed, stopping vm")
        if status, err := getWslStatus(ctx, instanceName); err == nil &&
            status == limatype.StatusRunning {
            _ = stopVM(ctx, distroName)
        }
    }
}()
```

Once `ctx` is canceled (the normal trigger for any `limactl stop` or hostagent shutdown), `<-ctx.Done()` returns instantly on every iteration. The goroutine then hot-loops:

- pinning a CPU core,
- spawning a fresh `wsl.exe` subprocess on every iteration via `getWslStatus` (and, until the VM shows as stopped, via `stopVM` too),
- flooding `wslservice.exe`, which is global on the Windows host and shared with every other WSL distro (Docker Desktop, Rancher Desktop, plain Ubuntu shells, etc.).

The hostagent process never exits cleanly because this goroutine never returns; `ha.pid` / `ha.sock` linger until the OS kills the hostagent.

```powershell
# ... wait for ready ...
limactl stop demo
```

Within ~1 second of the stop returning, the hostagent process pegs a CPU core and `wsl.exe` PIDs increment many times per second (visible in Process Explorer or `Get-Process wsl` polled in PowerShell). The hostagent does not exit on its own.

This is structural, not timing-dependent — the loop has no exit condition.

### Fix

The intent was clearly "wait once, then stop the VM." Dropping the `for {` /`}` wrapper restores that. PR follows.

### Environment

Affects all WSL2 instances (`vmType: wsl2`) on Windows. Found by code inspection against `master`.

mn-ram:
Fixes #4891.

The boot-script goroutine in `provisionVM` wraps a single context-cancellation wait in `for {}`. Once `ctx` is canceled (every `limactl stop` / shutdown), `<-ctx.Done()` returns immediately on every iteration, so the goroutine hot-loops calling `getWslStatus` and `stopVM` forever — pinning a CPU core and flooding `wslservice.exe` with `wsl.exe` subprocesses.

Drop the `for {`/`}` wrapper so the goroutine waits once and exits.

## Test plan

- `GOOS=windows go build ./pkg/driver/wsl2/...`
- `GOOS=windows go vet ./pkg/driver/wsl2/...`

### DIFF
--- a/pkg/driver/wsl2/vm_windows.go
+++ b/pkg/driver/wsl2/vm_windows.go
@@ -132,13 +132,11 @@ func provisionVM(ctx context.Context, instanceDir, instanceName, distroName stri
 					"check /var/log/lima-init.log for more details (out=%q)", cmd.Args, err, string(out))
 		}
 
-		for {
-			<-ctx.Done()
-			logrus.Info("Context closed, stopping vm")
-			if status, err := getWslStatus(ctx, instanceName); err == nil &&
-				status == limatype.StatusRunning {
-				_ = stopVM(ctx, distroName)
-			}
+		<-ctx.Done()
+		logrus.Info("Context closed, stopping vm")
+		if status, err := getWslStatus(ctx, instanceName); err == nil &&
+			status == limatype.StatusRunning {
+			_ = stopVM(ctx, distroName)
 		}
 	}()
 


mn-ram:
The `Windows tests (WSL2)` failure looks like a runner issue, not caused by this PR.

Logs:
- create/start/stop all passed
- Failure on immediate restart → `HCS_E_CONNECTION_TIMEOUT`
- `wsl.exe --unregister` hung → job timed out

This error is from Windows (Hyper-V), not the Go code touched here.

That said, earlier loop spam (`getWslStatus` → `wsl.exe`) might have affected timing on restart, so can’t fully rule it out.

Likely a flake — could you please re-run to confirm?

mn-ram:
Friendly follow-up on this PR. @unsuman

The fix is minimal and targeted, and checks are passing. Would appreciate a review when convenient. Happy to provide any additional validation if needed.

unsuman (approved):
Thanks!

jandubois:
This looks pretty much like a PR generated by Claude Code (which is fine, but would be nice to mention).

If you generate a fix with AI, it is very useful to ask the AI afterwards to do a code review. You'll be surprised how often it finds ways to improve the initial version. Or find related pre-existing issues that might be worthwhile either addressing as well, or opening a separate issue for, e.g.

> - Cleanup runs with a canceled context. Both `stopVM` (`vm_windows.go:59`) and `getWslStatus` (`vm_windows.go:208`) feed `ctx` into `executil.RunUTF16leCommand` via `WithContext(ctx)`. By the time `<-ctx.Done()` returns, that context is canceled, so the `wsl.exe --list --verbose` and `wsl.exe --terminate` invocations will likely fail immediately with context canceled. In other words: the very cleanup we kept the goroutine alive for may not actually happen. Detaching with `context.WithoutCancel(ctx)` (or `context.Background()` plus a fresh timeout) for the cleanup commands would be more robust.
> - `getWslStatus` errors are silently swallowed. `err == nil && status == StatusRunning` means any error (including the "context canceled" one above) skips the terminate path entirely. At minimum, `logrus.Warnf` on the error would aid debugging.
> - `_ = stopVM(...)` discards the terminate error without logging.

Might be worth turning into a new issue...

mn-ram:
> This looks pretty much like a PR generated by Claude Code (which is fine, but would be nice to mention).
> 
> If you generate a fix with AI, it is very useful to ask the AI afterwards to do a code review. You'll be surprised how often it finds ways to improve the initial version. Or find related pre-existing issues that might be worthwhile either addressing as well, or opening a separate issue for, e.g.
> 
> > * Cleanup runs with a canceled context. Both `stopVM` (`vm_windows.go:59`) and `getWslStatus` (`vm_windows.go:208`) feed `ctx` into `executil.RunUTF16leCommand` via `WithContext(ctx)`. By the time `<-ctx.Done()` returns, that context is canceled, so the `wsl.exe --list --verbose` and `wsl.exe --terminate` invocations will likely fail immediately with context canceled. In other words: the very cleanup we kept the goroutine alive for may not actually happen. Detaching with `context.WithoutCancel(ctx)` (or `context.Background()` plus a fresh timeout) for the cleanup commands would be more robust.
> > * `getWslStatus` errors are silently swallowed. `err == nil && status == StatusRunning` means any error (including the "context canceled" one above) skips the terminate path entirely. At minimum, `logrus.Warnf` on the error would aid debugging.
> > * `_ = stopVM(...)` discards the terminate error without logging.
> 
> Might be worth turning into a new issue...

Thanks for the detailed review and the follow-up suggestions! Those are great catches:

- Canceled context in cleanup — You're right, ctx is already canceled by the time the <-ctx.Done() arm runs, so stopVM and getWslStatus would fail immediately. Using context.WithoutCancel(ctx) or context.Background() with a timeout for the cleanup commands would be more robust.
- Swallowed getWslStatus errors — Logging a logrus.Warnf on error instead of silently skipping would definitely help with debugging.
- Discarded stopVM error — Should at least log at warn level.

I'll open a separate issue to track these improvements. Thanks again for pointing them out!

PR #4895, opened 2026-04-27
hostagent: stop leaking inotify watchers and re-spawn the gRPC stream after guest-agent reconnect

### LINKED ISSUE #4894: hostagent: startInotify leaks watchers + gRPC stream and silently stops working after guest-agent reconnect
### Description

`startInotify` in `pkg/hostagent/inotify.go` leaks two distinct resources, and a third lifecycle bug means inotify silently stops working entirely after the first guest-agent reconnect.

**1. `notify.Stop` is never called.** `setupWatchers` invokes `notify.Watch(...)` once per writable mount. `rjeczalik/notify` allocates a kernel `inotify_init` FD plus an internal reader goroutine for each `Watch`, and *requires* `notify.Stop(ch)` to release them. `git grep notify.Stop` returns zero matches in the repo. When `startInotify` returns on `<-ctx.Done()`, every kernel watcher and reader goroutine survives.

**2. `inotifyClient.CloseSend()` is never called.** The `PostInotify` RPC is a client-streaming gRPC. Without `CloseSend`, the guest-agent's server handler stays parked reading from a stream the client abandoned, and the HTTP/2 stream slot is never released on the underlying `*grpc.ClientConn`.

**3. `startInotify` is spawned exactly once and never re-entered.** It's launched from a single `go func()` inside `watchGuestAgentEvents` (`pkg/hostagent/hostagent.go` ~line 729). When the guest agent restarts (`systemctl restart lima-guestagent` inside the guest, or any guest reboot), `inotifyClient.Send(event)` returns an error — but the loop only logs `"failed to send inotify"` at warn level and keeps going. The new gRPC client built by `getOrCreateClient` on the next reconnect is never used for inotify; host-side file changes silently stop propagating into the guest until `limactl stop && limactl start`.

```bash
cat <<'EOF' > /tmp/inotify-leak.yaml
mounts:
  - location: "~"
    writable: true
mountInotify: true
EOF
limactl start --tty=false --name=demo /tmp/inotify-leak.yaml

# Verify host-side touches reach the guest:
touch ~/inotify-test-1
limactl shell demo -- stat /home/$(whoami).linux/inotify-test-1   # OK

# Restart the guest agent (or reboot the guest):
limactl shell demo -- sudo systemctl restart lima-guestagent

# Touch another file on the host:
touch ~/inotify-test-2

# Inside the guest, file-watching tooling (Vite/Next.js/nodemon/cargo watch/...)
# no longer sees host-side changes. There is no error in the hostagent log
# beyond a "failed to send inotify" warn that is easy to miss.
```

The watcher-leak side is observable as well:

```bash
HA_PID=$(pgrep -f "lima-hostagent.*demo")
ls /proc/$HA_PID/fd | xargs -I{} readlink /proc/$HA_PID/fd/{} 2>/dev/null \
  | grep -c anon_inode:inotify
# never decreases across the lifetime of the hostagent
```

### Impact

- **Goroutine leak** (rjeczalik/notify reader) and **kernel-FD leak** (one `inotify_init` per writable mount) per Lima instance lifetime.
- **Half-open gRPC client-streams** accumulating on the guest-agent side across reconnects.
- **Silent correctness bug**: file-watch tooling inside the guest stops working after the first guest-agent restart with no actionable signal in the log. This affects every Colima / Rancher Desktop / Finch user who runs Vite, Next.js, `nodemon`, `air`, `watchexec`, `cargo watch`, etc. against a Lima-mounted source tree (a documented and recommended workflow for `mountInotify: true`).

### Fix

Three small changes (PR follows):

1. `defer notify.Stop(mountWatchCh)` after `setupWatchers` succeeds.
2. `defer inotifyClient.CloseSend()` after the stream is opened.
3. Return an error on `inotifyClient.Send` failure, and wrap the spawn site in a `for { startInotify(ctx); select { case <-ctx.Done(): return; case <-time.After(10*time.Second): } }` reconnect loop — mirroring what `watchGuestAgentEvents` already does for the `Events` stream.

### Environment

Found by code inspection against `master` (commit `5a20c6d3`). Affects every Lima Linux instance with `mountInotify: true` (the documented setting for live-reload-capable host mounts).

mn-ram:
Fixes #4894.

`startInotify` never called `notify.Stop` (so the rjeczalik/notify reader goroutine and the per-mount `inotify_init` FDs leak) and never called `inotifyClient.CloseSend()` (so the guest-agent's `PostInotify` handler stays parked). It was also spawned exactly once and only logged on `Send` errors, so after the first guest-agent restart the inotify gRPC stream was dead but the goroutine kept "running" — host-side file changes silently stopped reaching the guest until `limactl stop && limactl start`.

This change:

- adds `defer notify.Stop(mountWatchCh)` and `defer inotifyClient.CloseSend()`,
- returns an error on `inotifyClient.Send` failure (instead of warn-and-loop), and
- wraps the spawn site in a 10s reconnect loop so a fresh `startInotify` is launched against the new client whenever the stream dies — matching what `watchGuestAgentEvents` already does for the `Events` stream.

## Test plan

- `go build ./...`
- `go vet ./pkg/hostagent/...`
- `go test ./pkg/hostagent/...`
- Manual: `mountInotify: true` Lima instance, `touch` a file on the host before/after `systemctl restart lima-guestagent` inside the guest — file changes continue to propagate after the fix; on `master` they stop after the first restart.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -727,17 +727,27 @@ func (a *HostAgent) watchGuestAgentEvents(ctx context.Context) {
 	})
 
 	go func() {
-		if a.instConfig.MountInotify != nil && *a.instConfig.MountInotify {
-			client := a.getClient()
-			if client == nil || !isGuestAgentSocketAccessible(ctx, client) {
-				if a.driver.ForwardGuestAgent() {
-					sshAddress, sshPort := a.sshAddressPort()
-					_ = forwardSSH(ctx, a.sshConfig, sshAddress, sshPort, localUnix, remoteUnix, verbForward, false)
-				}
+		if a.instConfig.MountInotify == nil || !*a.instConfig.MountInotify {
+			return
+		}
+		client := a.getClient()
+		if client == nil || !isGuestAgentSocketAccessible(ctx, client) {
+			if a.driver.ForwardGuestAgent() {
+				sshAddress, sshPort := a.sshAddressPort()
+				_ = forwardSSH(ctx, a.sshConfig, sshAddress, sshPort, localUnix, remoteUnix, verbForward, false)
 			}
-			err := a.startInotify(ctx)
-			if err != nil {
-				logrus.WithError(err).Warn("failed to start inotify")
+		}
+		// Re-spawn startInotify when its gRPC stream dies (typically because
+		// the guest agent restarted). Without this, host-side file changes
+		// silently stop propagating into the guest after the first reconnect.
+		for {
+			if err := a.startInotify(ctx); err != nil && !errors.Is(err, context.Canceled) {
+				logrus.WithError(err).Warn("inotify stream ended; will retry")
+			}
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(10 * time.Second):
 			}
 		}
 	}()
--- a/pkg/hostagent/inotify.go
+++ b/pkg/hostagent/inotify.go
@@ -5,6 +5,7 @@ package hostagent
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


mn-ram:
Hi @unsuman 👋

Just a gentle follow-up on this PR.
It fixes the inotify watcher leak (ref #4894) and follows the same approach as #4889. It should be good to go, but I’m happy to make any changes if needed.

Would appreciate a review whenever you get a chance. Thanks!

unsuman (approved):
Thanks!

PR #4903, opened 2026-04-28
limactl: replace SSH column with IP, drop the local SSH port

### LINKED ISSUE #4181: `limactl ls`: stop printing local SSH port (when the screen width is narrow)

bitalizer:
Closes #4181.

Implements the column design that @jandubois sketched in #4181 (with @afbjorklund's nod): rename the column from `SSH` to `IP`, drop the `:port` suffix, and treat the new column the same way the `ARCH` column is treated — hide it when every instance is on the default `127.0.0.1` loopback and the screen is too narrow to fit everything else.

## Why drop the port

`SSH` showed `127.0.0.1:<port>` (or the misleading `127.0.0.1:0` for stopped instances). The port on its own is not actionable: it has to be paired with the generated identity file and config, so users go through `ssh -F ~/.lima/<name>/ssh.config lima-<name>` — which keys off the instance name, not the port.

## What changes

- `SSH` column → `IP` column, showing only `instance.SSHAddress`.
- `SSHAddress` is preserved on the struct, so the IP work in #4175 (and any consumer of `instance.SSHAddress` like `cmd/limactl/show-ssh.go` / `tunnel.go`) is unaffected.
- The IP column is hidden when **(a)** every instance has the default loopback address **and** **(b)** the terminal is too narrow to fit everything else. If any instance has a non-default address (the case that motivated #4175), the column stays visible regardless of width — because that IP is the only place the information is surfaced.
- The `"127.0.0.1"` literal that was duplicated between `Inspect` (which seeds `SSHAddress`) and the table printer is consolidated into a `defaultSSHAddress` const, so the seeding and the visibility check can't drift apart.

## Sample output

Wide enough (`width >= 100`):
```
NAME       STATUS     IP              VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
default    Running    192.168.64.2    vz        aarch64    4       4GiB      100GiB    ~/.lima/default
alpine     Running    127.0.0.1       vz        aarch64    4       4GiB      100GiB    ~/.lima/alpine
```

Narrow + all-loopback (column auto-hidden, mirroring the existing `ARCH` rule):
```
NAME       STATUS     CPUS    MEMORY    DISK
debug      Stopped    4       4GiB      100GiB
docker     Stopped    4       16GiB     100GiB
```

Narrow + non-loopback (column kept, since the IP is the only place this information shows):
```
NAME       STATUS     IP              CPUS    MEMORY    DISK    DIR
default    Running    192.168.64.2    4       4GiB      100GiB  ~/.lima/default
```

## Tests

- All existing `TestPrintInstanceTable*` cases were updated to the new column layout. The hide-on-narrow rule means width=60/80 with all-loopback now produces a tighter table than before — that's the intended change.
- Added `TestPrintInstanceTable80NonLoopback` covering the "keep IP visible when at least one instance has a non-default address" branch.
- `go test ./pkg/store/... -run TestPrintInstance` is green locally.
- `go vet ./...` clean.

Open question for reviewers: I kept the IP column under a 2-`columnWidth` budget (matching how `SSH` was budgeted before, since `192.168.x.x` ≈ 13 chars). Happy to drop it to 1 if you'd rather optimize for short addresses.

### DIFF
--- a/pkg/store/instance.go
+++ b/pkg/store/instance.go
@@ -31,6 +31,11 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/version/versionutil"
 )
 
+// defaultSSHAddress is the loopback address assigned to an instance's SSHAddress
+// when no other host-reachable IP is known. It is also used by the table printer
+// to decide when the IP column can be hidden on narrow screens.
+const defaultSSHAddress = "127.0.0.1"
+
 // Inspect returns err only when the instance does not exist (os.ErrNotExist).
 // Other errors are returned as *Instance.Errors.
 func Inspect(ctx context.Context, instName string) (*limatype.Instance, error) {
@@ -59,7 +64,7 @@ func Inspect(ctx context.Context, instName string) (*limatype.Instance, error) {
 	inst.Config = y
 	inst.Arch = *y.Arch
 	inst.VMType = *y.VMType
-	inst.SSHAddress = "127.0.0.1"
+	inst.SSHAddress = defaultSSHAddress
 	inst.SSHLocalPort = *y.SSH.LocalPort // maybe 0
 	inst.SSHConfigFile = filepath.Join(instDir, filenames.SSHConfig)
 	inst.HostAgentPID, err = ReadPIDFile(filepath.Join(instDir, filenames.HostAgentPID))
@@ -302,9 +307,11 @@ func PrintInstances(w io.Writer, instances []*limatype.Instance, format string, 
 	case "table":
 		types := map[string]int{}
 		archs := map[string]int{}
+		ips := map[string]int{}
 		for _, instance := range instances {
 			types[instance.VMType]++
 			archs[instance.Arch]++
+			ips[instance.SSHAddress]++
 		}
 		all := options != nil && options.AllFields
 		width := 0
@@ -314,13 +321,21 @@ func PrintInstances(w io.Writer, instances []*limatype.Instance, format string, 
 		columnWidth := 8
 		hideType := false
 		hideArch := false
+		hideIP := false
 		hideDir := false
 
 		// width == 0 means the terminal width is unknown (e.g. stdout is not a TTY),
 		// in which case we treat it as unlimited and never hide any column.
 		columns := 1 // NAME
 		columns += 2 // STATUS
-		columns += 2 // SSH
+		// can we still fit the remaining columns (7) when IP is shown
+		if width != 0 && (columns+2+7)*columnWidth > width && !all {
+			// only hide IP if all instances share the default loopback address
+			hideIP = len(ips) == 1 && instances[0].SSHAddress == defaultSSHAddress
+		}
+		if !hideIP {
+			columns += 2 // IP
+		}
 		// can we still fit the remaining columns (7)
 		if width != 0 && (columns+7)*columnWidth > width && !all {
 			hideType = len(types) == 1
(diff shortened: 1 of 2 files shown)


bitalizer:
Heads-up on CI: 27 of 28 jobs are green (unit, lint, DCO, BATS, all 12 QEMU integration variants, Windows, vmnet, vulncheck, etc). The single red is **`Integration tests (vz) (default.yaml)`**, which failed at `TEST| [ERROR] Failed to start "default"` ~10 min into VM boot — before any `limactl list` / table-printing code path could run. Looks like a hosted-macOS/nested-virt flake on the vz job, not a regression. Happy to rebase if you'd like a fresh run.

afbjorklund:
The IP is not all that interesting either, it would be a nice feature to include DNS and **not** have to use the IP...

Like how they do it for the Apple Container virtual machines, with an embedded DNS service:

https://github.com/apple/container/blob/main/docs/tutorial.md#set-up-a-local-dns-domain-optional

Already you can use the `lima-instance` ssh alias, assuming that you have set up the ssh config to include.

bitalizer:
agreed, resolving instances by DNS name instead of IP would be nicer - the ssh.config alias already gets you part of the way there. that's a bigger change though, so i'd keep it out of this PR which is just swapping the useless port column for the IP. happy to follow up on the embedded-DNS idea separately if there's interest.

rebased onto master to clear the conflict, tests still green.

AkihiroSuda:
How does the new output look like?

jandubois (approved):
The changes look fine, but until #4175 lands the IP address is still always `127.0.0.1`, which is not that interesting.

I'm not sure if showing an address for a stopped instance is correct, but I guess for now it doesn't matter.

The column width will be too small if we ever support IPv6 addresses, but given that we are still locked on `127.0.0.1` this is maybe not a consideration.

So technically this fixes #4181, but is kind of useless until #4175 lands.

jandubois:
> So technically this fixes #4181, but is kind of useless until #4175 lands.

@lima-vm/committers We need to decide if this PR should be merged, or switched to draft until #4175 lands.

Personally I'm fine with merging because it doesn't make the current situation worse, and it gets the PR out of the list of open PRs, which is getting uncomfortably large.

jandubois:
> How does the new output look like?

```console
❯ l ls
NAME       STATUS     IP           VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
alpine     Stopped    127.0.0.1    vz        aarch64    4       4GiB      100GiB    ~/.lima/alpine
default    Stopped    127.0.0.1    vz        aarch64    4       4GiB      100GiB    ~/.lima/default
```

pkg/store/instance.go
```diff
@@ -59,7 +64,7 @@ func Inspect(ctx context.Context, instName string) (*limatype.Instance, error) {
 	inst.Config = y
 	inst.Arch = *y.Arch
 	inst.VMType = *y.VMType
-	inst.SSHAddress = "127.0.0.1"
+	inst.SSHAddress = defaultSSHAddress
```
AkihiroSuda, line 67: Is this constantly set to `defaultSSHAddress`, i.e., 127.0.0.1?

Then this PR doesn't seem to make a sense.

jandubois: Yes, that's what I meant by

> but is kind of useless until #4175 lands

Issue #4898, opened 2026-04-27
Should log messages use `%#q` instead of `%q`?


mie313:
I understand that Logrus is not detecting a TTY on Windows, and therefore, log messages on Windows are fall back to plain text format.
I think they are hard to read because Windows file path contains back-slashs and they need to be escaped with another back-slashs in `%q`.

For example:
```
time="2026-04-27T21:44:04+01:00" level=fatal msg="failed to resolve vm for \"C:\\\\Users\\\\mie31\\\\.lima\\\\default\\\\lima.yaml\": unsupported image type for vmType: wsl2, tarball root file system required: \"https://cloud-images.ubuntu.com/releases/questing/release-20260320/ubuntu-25.10-server-cloudimg-amd64.img\""
```

So, I was just wondering it might be better to use `%#q` in log messages instead of `%q`.

If we use `%#q`, the filw path above will be as follows:
```
time="2026-04-27T21:44:04+01:00" level=fatal msg="failed to resolve vm for `C:\Users\mie31\.lima\default\lima.yaml`: unsupported image type for vmType: wsl2, tarball root file system required: `https://cloud-images.ubuntu.com/releases/questing/release-20260320/ubuntu-25.10-server-cloudimg-amd64.img`"
```

mie313:
Since Logrus use plain text format in non TTY environment, Linux/Mac users also encounter this kind of issue when they dump logs into a file.

For example, in Linux environment, a log will appear on a file as follows:
```
time="2026-04-27T21:54:12+01:00" level=warning msg="[hostagent] treating lima version \"5a20c6d3.m\" from \"/home/mie313/.lima/template/lima-version\" as very latest release"
```

If it makes sense, I am happy to work on it.

jandubois:
I'm in favour, but please wait to hear from other maintainers as well first.

unsuman:
Sounds good!

mie313:
Thank you for your comments.
Could you please assign me?

afbjorklund:
For normal paths on normal systems, it should just change from quotes to backticks - which sounds harmless enough.

jandubois:
> For normal paths on normal systems, it should just change from quotes to backticks - which sounds harmless enough.

Yes, but it also means that nested quotes no longer need backslash-escapes because they are verbatim backticks. So it improves readability ("where does this quoted string end?") even on Unix systems.

AkihiroSuda:
Is this completed?

unsuman:
[...]

PR #4909, opened 2026-04-29
Change %q in logs and errors to %#q [Part1]

mie313:
Change `%q` in logs and errors to `%#q` (all files in `cmd/limactl`) .

Related issue: #4898 

Question:
I have modified 126 files in total, so I need to push another 98 files.
I am worried that 126 files are too many to be reviewd in one PR.
Can I add them in this PR?
Would it be better to raise another PRs?

### DIFF
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -392,7 +392,7 @@ func YQExpressions(flags *flag.FlagSet, newInstance bool, params map[string]stri
 				case "none":
 					return []string{`.containerd.user = false | .containerd.system = false`}, nil
 				default:
-					return nil, fmt.Errorf(`expected one of ["user", "system", "user+system", "none"], got %#q`, s)
+					return nil, fmt.Errorf("expected one of [`user`, `system`, `user+system`, `none`], got %#q", s)
 				}
 			},
 			true,
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -127,12 +127,12 @@ func defaultExprFunc(expr string) func(v *flag.Flag) ([]string, error) {
 func ParsePortForward(spec string) (hostPort, guestPort string, isStatic bool, err error) {
 	parts := strings.Split(spec, ",")
 	if len(parts) > 2 {
-		return "", "", false, fmt.Errorf("invalid port forward format %q, expected HOST:GUEST or HOST:GUEST,static=true", spec)
+		return "", "", false, fmt.Errorf("invalid port forward format %#q, expected HOST:GUEST or HOST:GUEST,static=true", spec)
 	}
 
 	portParts := strings.Split(strings.TrimSpace(parts[0]), ":")
 	if len(portParts) != 2 {
-		return "", "", false, fmt.Errorf("invalid port forward format %q, expected HOST:GUEST", parts[0])
+		return "", "", false, fmt.Errorf("invalid port forward format %#q, expected HOST:GUEST", parts[0])
 	}
 
 	hostPort = strings.TrimSpace(portParts[0])
@@ -143,10 +143,10 @@ func ParsePortForward(spec string) (hostPort, guestPort string, isStatic bool, e
 		if staticValue, ok := strings.CutPrefix(staticPart, "static="); ok {
 			isStatic, err = strconv.ParseBool(staticValue)
 			if err != nil {
-				return "", "", false, fmt.Errorf("invalid value for static parameter: %q", staticValue)
+				return "", "", false, fmt.Errorf("invalid value for static parameter: %#q", staticValue)
 			}
 		} else {
-			return "", "", false, fmt.Errorf("invalid parameter %q, expected 'static=' followed by a boolean value", staticPart)
+			return "", "", false, fmt.Errorf("invalid parameter %#q, expected 'static=' followed by a boolean value", staticPart)
 		}
 	}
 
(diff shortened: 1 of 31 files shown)


jandubois:
> I have modified 126 files in total, so I need to push another 98 files.
> I am worried that 126 files are too many to be reviewd in one PR.

Thank you for asking first!

> Can I add them in this PR?
> Would it be better to raise another PRs?

I think as long as mechanical substitution of `%q` → `%#q` are the only changes in this PR, then combining them all should be fine.

However, I think we have some BATS tests that check for specific error messages, that will include the quotes. If these are more complicated than also simple substitutions in `assert_warning` and `assert_error` arguments, then maybe those should go into a separate PR. But if they too are all straight-forward, then I think they can all go in together.

mie313:
@jandubois Thank you for your response!
As you pointed out, bats test codes need updating.
So, my second commit modified them.
Since they were also simply replacement, I am assuming that all modified files can be combined in this PR.

jandubois:
Please squash commits.

We generally ask for all commits in a PR to be squashed to make sure every commit will pass tests (helps with `git bisect`). We sometimes make exception when a single PR includes multiple separate features that should be kept separate. But in this case the BATS changes are required for the change to pass integration tests.

mie313:
I realised that I need to mention that this PR containts not only mechanical substitution of %q → %#q but also test code changes to align with the replacement.
 @jandubois Can I focus on cmd/limactl and related test code like bats on this PR?

mie313:
>We generally ask for all commits in a PR to be squashed to make sure every commit will pass tests (helps with git bisect).

Sure, I will squash all commits once modifying bats test codes are completed.
(At this moment, the second commit is not ready for review, I will work on it later.)

jandubois:
> Can I focus on cmd/limactl and related test code like bats on this PR?

Sure, as long as the PR passes all checks in CI

mie313:
@jandubois I have squashed all commits.
Would you mind reviewing the PR?

This merged commit fails one check.
I am not sure what caused this error, and I think that it might be fraky error.

hack/bats/tests/protect.bats
```diff
@@ -17,18 +17,18 @@ local_setup_file() {
 
 @test 'protecting a non-existing instance fails' {
     run_e -1 limactl protect "${NOTEXIST}"
-    assert_fatal "failed to inspect instance \"${NOTEXIST}\"…"
+    assert_fatal "failed to inspect instance `${NOTEXIST}`…"
```
jandubois:
```suggestion
    assert_fatal "failed to inspect instance \`${NOTEXIST}\`…"
```

hack/bats/tests/protect.bats
```diff
@@ -41,28 +41,28 @@ local_setup_file() {
 
 @test 'deleting the unprotected clone instance succeeds' {
     run_e -0 limactl delete --force "$CLONE"
-    assert_info "Deleted \"${CLONE}\"…"
+    assert_info "Deleted `${CLONE}`…"
```
jandubois:
```suggestion
    assert_info "Deleted \`${CLONE}\`…"
```

hack/bats/tests/protect.bats
```diff
@@ -41,28 +41,28 @@ local_setup_file() {
...
     run_e -0 limactl unprotect "$INSTANCE"
-    assert_warning "Instance \"${INSTANCE}\" isn't protected. Skipping."
+    assert_warning "Instance \`${INSTANCE}\` isn't protected. Skipping."
     assert_file_not_exists "$LIMA_HOME/$INSTANCE/protected"
 }
 
 @test 'deleting unprotected dummy instance succeeds' {
     run_e -0 limactl delete --force "$INSTANCE"
-    assert_info "Deleted \"${INSTANCE}\"…"
+    assert_info "Deleted `${INSTANCE}`…"
```
jandubois:
```suggestion
    assert_info "Deleted \`${INSTANCE}\`…"
```

cmd/limactl/editflags/editflags.go
```diff
@@ -392,7 +392,7 @@ func YQExpressions(flags *flag.FlagSet, newInstance bool, params map[string]stri
 				case "none":
 					return []string{`.containerd.user = false | .containerd.system = false`}, nil
 				default:
-					return nil, fmt.Errorf(`expected one of ["user", "system", "user+system", "none"], got %q`, s)
+					return nil, fmt.Errorf(`expected one of ["user", "system", "user+system", "none"], got %#q`, s)
```
jandubois: This is inconsistent with the change you made above:

```
fmt.Errorf("network name must be `vzNAT` or `lima:*`, got %#q", s)
```

```suggestion
					return nil, fmt.Errorf(`expected one of [`user`, `system`, `user+system`, `none`], got %#q`, s)
```

I think we should use backticks everywhere if we use `%#q` everywhere.

I have not audited the file for other occurances where strings use double-quotes to denote literal values.

jandubois (changes requested):
There are a few unescaped backticks in `protect.bats`. They happen to pass the test anyways because they interpolate as the empty string, and the string ends with the `…` wildcard.

Also some inconsistency for quoting literal values in strings.

mie313:
Thank you for your review!
I updated the codes in https://github.com/lima-vm/lima/pull/4909/commits/c5e27999aff00ee2caf083e5701c77e83ad97606 following your PR comments.

If it looks ok, I will squash all PRs after your review.

Please kindly note that
- [Integration tests(QEMU, Linux host) failed](https://github.com/lima-vm/lima/actions/runs/25257233342/job/74058525654?pr=4909).
- [Lints tests failed](https://github.com/lima-vm/lima/actions/runs/25257233342/job/74058525565?pr=4909) due to the network error.

As I thinkg these were not caused by my changes, I have decided to ask for the second review.

mie313:
Hi @jandubois,
I was just wondering if you saw my upadted code.
If you review them and let me know if they are ok, that would be helpful.
Once I receive your comment, I will squash all PRs.

jandubois (approved):
Thanks, LGTM

There are still 2 single quotes that should be backticks, but since you have more PRs in this series, you can address it in a later follow-up:

  - `cmd/limactl/editflags/editflags.go:357` — expected 'static=' followed by a boolean value
  - `cmd/limactl/main.go:281` — See '%s --help'.

PR #4910, opened 2026-04-29
openSUSE dropped the 2MB OVMF images from QEMU in version 202602

jandubois:
Only the 4MB versions are available now (following the lead from Arch and Debian).


unsuman (approved):
Thanks!

PR #4914, opened 2026-04-30
shell: clean up stale ssh.sock on --reconnect when the master is already dead

### LINKED ISSUE #4913: limactl shell --reconnect fails when the previous ssh master crashed (stale ssh.sock not cleaned up)
### Description

\`limactl shell --reconnect <inst>\` is the user-visible escape hatch for resetting a wedged SSH session — the documented use case in #3125. It currently fails with a non-zero exit when the SSH ControlMaster file in \`<instDir>/ssh.sock\` exists but the master process is already dead.

\`cmd/limactl/shell.go:180-192\`:

\`\`\`go
if restart && sshutil.IsControlMasterExisting(inst.Dir) {
    logrus.Infof("Exiting ssh session for the instance %q", instName)

    sshConfig := &ssh.SSHConfig{
        ConfigFile:     inst.SSHConfigFile,
        Persist:        false,
        AdditionalArgs: []string{},
    }

    if err := ssh.ExitMaster(inst.Hostname, inst.SSHLocalPort, sshConfig); err != nil {
        return err
    }
}
\`\`\`

Two problems chain here:

1. **\`IsControlMasterExisting\` is liveness-blind.** \`pkg/sshutil/sshutil.go:328-333\` is a pure \`os.Stat\` on \`<instDir>/ssh.sock\` — it does not verify the master process is alive. After any unclean hostagent shutdown (\`kill -9\`, OOM, host crash, panic before the registered \`cleanUp\` runs) the socket file outlives the master.

2. **\`ssh.ExitMaster\` failure is treated as fatal.** When the master is dead, \`ssh -O exit\` fails — typically with \`Control socket connect: Connection refused\` or \`mux_client_request_session: send fds failed\`. \`shellAction\` returns that error directly, so \`--reconnect\` exits non-zero with \`exit status 255\` and no recovery. The user has no in-tree way to clear the stale state — they have to \`rm ~/.lima/<inst>/ssh.sock\` by hand and try again.

This is exactly the inverse of what \`--reconnect\` is supposed to do. The whole point of the flag (per #3125) is to *reset* a broken multiplexed session.

### Trigger (deterministic)

\`\`\`sh
limactl start default
# Simulate any unclean hostagent exit:
kill -9 \$(cat ~/.lima/default/ha.pid)
ls ~/.lima/default/ssh.sock          # file still present, master dead

limactl shell --reconnect default
# -> exits non-zero. Stderr:
#    FATA[...] failed to execute \`ssh -O exit -p ... 127.0.0.1\`, out="...", err=exit status 255
# Without manual rm, every subsequent --reconnect repeats the same failure.
\`\`\`

100% reproducible.

### Why it matters in practice

- Hits anyone whose hostagent terminated uncleanly — macOS memory-pressure kills, panics, the very class of crash recovery that \`--reconnect\` is meant to address.
- Disproportionately impactful on Windows hosts. The Cygwin/msys2 OpenSSH \`ssh.exe\` bundles emulate the Unix-domain control socket as a regular file plus a TCP fallback; that emulation file is even more prone to surviving its master process than a real Unix socket. Lima already strips ControlMaster in three places to work around Cygwin multiplex flakiness (\`pkg/hostagent/hostagent.go:290\`, \`pkg/hostagent/requirements.go:124\`, \`cmd/limactl/shell.go:337\`), but the \`--reconnect\` cleanup path is not one of them — and the stale-file case is exactly where Cygwin emulation hurts most.

### Fix

Two-line semantic change in \`cmd/limactl/shell.go\`:

1. Treat \`ssh.ExitMaster\` failure as a warning, not fatal.
2. After \`ExitMaster\` (whether it returned nil or not), \`os.Remove\` the control socket file. \`ssh -O exit\` already unlinks on success, so this is a no-op on the happy path; it is the only effective recovery on the unhappy path.

Both behaviors preserve existing semantics for the working-master case.

PR follows.

mn-ram:
## Summary

Fixes #4913.

`limactl shell --reconnect <inst>` (added in #3125 as the user-visible escape hatch for resetting a wedged SSH session) currently fails with `exit status 255` when the SSH ControlMaster file in `<instDir>/ssh.sock` exists but the master process is already dead — exactly the scenario where the user expects `--reconnect` to *recover* the broken state.

`sshutil.IsControlMasterExisting` is a pure `os.Stat` (`pkg/sshutil/sshutil.go:328-333`); it does not verify the master process is alive. After any unclean hostagent exit (`kill -9`, OOM, panic before the registered `cleanUp` runs) the socket file outlives the master. In that state `ssh.ExitMaster` shells out `ssh -O exit`, which fails with `Control socket connect: Connection refused`. `shellAction` returned that error directly, leaving the user with no in-tree recovery — they had to `rm ~/.lima/<inst>/ssh.sock` by hand.

This PR:

1. Treats the `ssh.ExitMaster` failure as a warning instead of fatal.
2. Forcibly removes the stale control socket after `ExitMaster`. `ssh -O exit` already unlinks the socket on success, so this is a no-op on the happy path and the only effective recovery on the unhappy path.

Disproportionately affects Windows hosts: Cygwin/msys2 OpenSSH emulates the Unix-domain control socket as a regular file + TCP fallback, which is even more prone to outliving its master than a real Unix-domain socket. Lima already strips `ControlMaster` in three Windows-specific call sites for related Cygwin flakiness (`pkg/hostagent/hostagent.go:290`, `pkg/hostagent/requirements.go:124`, `cmd/limactl/shell.go:337`); this PR closes the matching gap on the `--reconnect` cleanup path.

Closes #4913.

## Test plan

- [x] `go vet ./cmd/limactl/...` — clean
- [x] `go build ./cmd/limactl/...` — clean
- [x] `go test ./cmd/limactl/...` — pass
- [ ] Manual repro:
  ```sh
  limactl start default
  kill -9 $(cat ~/.lima/default/ha.pid)
  ls ~/.lima/default/ssh.sock          # still present
  limactl shell --reconnect default    # before: exit 255; after: succeeds
  ls ~/.lima/default/ssh.sock          # should be gone
  ```
- [ ] Re-run `limactl shell default` afterwards to confirm a clean session is re-established.
- [ ] Confirm the happy path is unaffected: `limactl shell --reconnect default` against a live master should still exit cleanly with no warning.

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -28,6 +29,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/instance"
 	"github.com/lima-vm/lima/v2/pkg/ioutilx"
 	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
 	"github.com/lima-vm/lima/v2/pkg/localpathutil"
 	"github.com/lima-vm/lima/v2/pkg/networks/reconcile"
 	"github.com/lima-vm/lima/v2/pkg/sshutil"
@@ -186,9 +188,30 @@ func shellAction(cmd *cobra.Command, args []string) error {
 			AdditionalArgs: []string{},
 		}
 
+		// IsControlMasterExisting only checks that the control socket file is
+		// present; it does not verify that the master process is alive. After
+		// an unclean hostagent shutdown the file can outlive the master, in
+		// which case `ssh -O exit` fails. Treating that failure as fatal
+		// defeats the purpose of --reconnect (the user is asking us to reset
+		// a broken state).
+		//
+		// However, ExitMaster can also fail for reasons other than a dead
+		// master (transient ssh errors, missing binary, etc.); blindly
+		// removing the socket in those cases would rip it out from under a
+		// still-live multiplexed connection. Disambiguate by running
+		// `ssh -O check` after a failed exit, and only remove the stale
+		// socket when the master is confirmed dead.
 		if err := ssh.ExitMaster(inst.Hostname, inst.SSHLocalPort, sshConfig); err != nil {
-			return err
+			if isControlMasterAlive(inst.Hostname, inst.SSHLocalPort, sshConfig) {
+				return fmt.Errorf("ssh master is still alive but `ssh -O exit` failed: %w", err)
+			}
+			logrus.WithError(err).Warn("ssh master is not responding; removing stale control socket")
+			controlSock := filepath.Join(inst.Dir, filenames.SSHSock)
+			if rmErr := os.Remove(controlSock); rmErr != nil && !errors.Is(rmErr, fs.ErrNotExist) {
+				return fmt.Errorf("failed to remove stale ssh control socket %q: %w", controlSock, rmErr)
+			}
 		}
+		// Success path: `ssh -O exit` already unlinked the control socket.
 	}
 
 	syncDirVal, err := flags.GetString("sync")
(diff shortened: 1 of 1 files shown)


cmd/limactl/shell.go
```diff
@@ -186,8 +188,20 @@ func shellAction(cmd *cobra.Command, args []string) error {
...
+		// it does not verify that the master process is alive. After an unclean
+		// hostagent shutdown the file can outlive the master, in which case
+		// `ssh -O exit` fails. Treating that failure as fatal defeats the whole
+		// purpose of --reconnect (the user is asking us to reset a broken state),
+		// so log the failure and fall through to forcibly removing the stale socket
+		// instead. ssh -O exit also unlinks the socket on success, so the cleanup
+		// is a no-op in the happy path.
 		if err := ssh.ExitMaster(inst.Hostname, inst.SSHLocalPort, sshConfig); err != nil {
-			return err
+			logrus.WithError(err).Warn("failed to exit ssh master cleanly; the master may already be dead — removing the stale control socket")
```
unsuman: This fallback is too broad. `ssh.sock` is removed after any `ssh.ExitMaster` error, not just when the master is actually dead.

mn-ram:
Thanks @unsuman for the review. You're right that the recovery logic was too broad.

Applying your suggestion literally (just `return err` on `ExitMaster` failure) brings the code back to the upstream behaviour, so the patch becomes a functional no-op. Closing this PR; happy to revisit `--reconnect` recovery in a different shape if/when there's interest.

PR #4919, opened 2026-05-01
pkg/reflectutil: add unit tests for UnknownNonEmptyFields

varsha-0007:
`pkg/reflectutil` has no test file, unlike other utility packages
such as `pkg/ptr`, `pkg/textutil`, and `pkg/osutil`.

This PR adds `reflectutil_test.go` covering the following cases:
- Known fields are not reported
- Unknown non-empty fields are reported
- Zero/empty values are ignored
- Pointer-to-struct input works correctly
- Zero-length slice and map are treated as empty
- Non-struct input panics as expected

### DIFF
--- a/pkg/reflectutil/reflectutil.go
+++ b/pkg/reflectutil/reflectutil.go
@@ -38,7 +38,7 @@ func UnknownNonEmptyFields(structOrStructPtr any, knownNames ...string) []string
 	default:
 		panic(fmt.Errorf("expected Ptr or Struct, got %+v", kind))
 	}
-	for i := range val.NumField() {
+	for i := 0; i < val.NumField(); i++ {
 		iField := val.Field(i)
 		if isEmpty(iField) {
 			continue
--- a/pkg/reflectutil/reflectutil.go
+++ b/pkg/reflectutil/reflectutil.go
@@ -30,13 +30,13 @@ func UnknownNonEmptyFields(structOrStructPtr any, knownNames ...string) []string
 	}
 	origVal := reflect.ValueOf(structOrStructPtr)
 	var val reflect.Value
-	switch kind := origVal.Kind(); kind {
-	case reflect.Ptr:
+	switch origVal.Kind() {
+	case reflect.Pointer:
 		val = origVal.Elem()
 	case reflect.Struct:
 		val = origVal
 	default:
-		panic(fmt.Errorf("expected Ptr or Struct, got %+v", kind))
+		panic(fmt.Errorf("expected pointer or struct, got %v", origVal.Kind()))
 	}
 	for i := 0; i < val.NumField(); i++ {
 		iField := val.Field(i)
--- a/pkg/reflectutil/reflectutil.go
+++ b/pkg/reflectutil/reflectutil.go
@@ -38,7 +38,7 @@ func UnknownNonEmptyFields(structOrStructPtr any, knownNames ...string) []string
 	default:
 		panic(fmt.Errorf("expected pointer or struct, got %v", origVal.Kind()))
 	}
-	for i := 0; i < val.NumField(); i++ {
+	for i := range val.NumField() {
 		iField := val.Field(i)
 		if isEmpty(iField) {
 			continue
--- a/pkg/reflectutil/reflectutil_test.go
+++ b/pkg/reflectutil/reflectutil_test.go
@@ -5,7 +5,7 @@ package reflectutil
 
 import (
 	"fmt"
-	"sort"
+	"slices"
 	"strings"
 	"testing"
 
@@ -125,9 +125,8 @@ func TestUnknownNonEmptyFields(t *testing.T) {
 
 			result := UnknownNonEmptyFields(tt.input, tt.knownFields...)
 
-			// Sort both slices before comparing so order does not matter.
-			sort.Strings(result)
-			sort.Strings(tt.expected)
+			slices.Sort(result)
+			slices.Sort(tt.expected)
 			assert.DeepEqual(t, tt.expected, result)
 		})
 	}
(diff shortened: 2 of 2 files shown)


pkg/reflectutil/reflectutil_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"testing"
+
+	"gotest.tools/v3/assert"
+	"gotest.tools/v3/assert/cmp"
+)
+
+type testStruct struct {
+	Known   string
+	Unknown string
+	Count   int
```
jandubois: The `Count` field is not used for anything.

pkg/reflectutil/reflectutil_test.go
```diff
@@ -0,0 +1,79 @@
...
+	Tags    []string
+	Meta    map[string]string
+}
+
+func TestUnknownNonEmptyFields_NoUnknown(t *testing.T) {
```
jandubois: If you look at other test files in the project, they usually use a table of test cases instead of defining a function for each one. All these tests could be entries in a slice that a single function iterates over.

jandubois (commented):
Please also address linter errors.

Bonus suggestions from Claude Code for additional test coverage:

> - Multiple unknown fields: every unknown-field assertion uses `cmp.Contains(result, "Unknown")`, which doesn't check that only `Unknown` is reported, nor the result's full contents/order. A case with two unknown non-empty fields and an `assert.DeepEqual` against a sorted/expected slice would catch regressions `cmp.Contains` won't.
> - Mixed known + unknown: no case where some non-empty fields are known and others aren't (the realistic usage pattern).
> - Nil pointer: `UnknownNonEmptyFields((*testStruct)(nil))` calls `origVal.Elem()` on a `nil` pointer, then `val.NumField()` panics. If the documented contract is "panics on nil", add a test; if not, this is a latent bug worth surfacing to the maintainers rather than silently leaving uncovered.

varsha-0007:
@jandubois Sure, will look into it!

Thanks for the detailed review! 🙂

pkg/reflectutil/reflectutil_test.go
```diff
@@ -4,76 +4,139 @@
...
+		{
+			// All fields are in the known list — result must be empty.
+			name:        "no unknown fields",
+			input:       testStruct{Known: "hello", Unknown: "world"},
+			knownFields: []string{"Known", "Unknown"},
+			expected:    nil,
+		},
+		{
+			// Unknown is non-empty and not in known list — must be reported.
+			name:        "one unknown field",
```
jandubois: This is actually a mix of one known and one unknown field, and an exact duplicate of the mixed case later in this file.

pkg/reflectutil/reflectutil_test.go
```diff
@@ -4,76 +4,139 @@
...
+		},
+		{
+			// Non-empty map is not zero — must appear in result.
+			name:        "non-empty map reported",
+			input:       testStruct{Meta: map[string]string{"key": "val"}},
+			knownFields: nil,
+			expected:    []string{"Meta"},
+		},
+		{
+			// Reviewer suggestion: multiple unknown fields at once.
```
jandubois: Please don't include "history" comments. We have `git blame` and GitHub issue and PR comments for that.

Comments in the source code should explain the WHAT and WHY, but not the history behind changes.

pkg/reflectutil/reflectutil_test.go
```diff
@@ -4,76 +4,139 @@
...
-	assert.Assert(t, cmp.Len(result, 0))
-}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.wantPanic {
+				// Verify panic cases without crashing the whole test run.
+				defer func() {
+					if r := recover(); r == nil {
+						t.Errorf("expected panic but did not panic")
+					}
```
jandubois, line 120: Maybe add a check on the panic message, so you are not silently accepting a different panic than the one you are expecting?

pkg/reflectutil/reflectutil_test.go
```diff
@@ -4,76 +4,139 @@
...
 import (
+	"sort"
 	"testing"
 
 	"gotest.tools/v3/assert"
-	"gotest.tools/v3/assert/cmp"
 )
 
+// testStruct is used as the input type for all test cases.
+// Count field removed — it was unused in any test case (reviewer feedback).
```
jandubois: Don't explain the history.

```suggestion
```

jandubois (changes requested):
Please fix linter errors!

jandubois:
You have fixed the linter errors but have not addressed my feedback.

varsha-0007:
@jandubois, will do it now!!

jandubois (changes requested):
There still is a linter error, and please squash your commits!

pkg/reflectutil/reflectutil.go
```diff
@@ -1,18 +1,31 @@
 // SPDX-FileCopyrightText: Copyright The Lima Authors
+
```
jandubois: What's up with all the random blank lines you have inserted?

If you create/update your changes with AI, then you MUST review them before submitting them to the PR. Please don't waste reviewer time; otherwise it will be faster if we use AI ourselves to create these kind of PRs.

varsha-0007: You're right, I should have reviewed the changes more carefully before pushing. I apologize for wasting your time. Fixing now.

varsha-0007:
hey @jandubois,
I verified the requested fixes locally, and the reflectutil tests/lint now pass.
The two remaining CI failures seem unrelated to this PR (Windows QEMU install and macOS template startup).

Could you please check or rerun them from your side?

jandubois (approved):
Thanks, LGTM

varsha-0007:
Thanks for the merge and for the feedback on the review process. I’ll make sure future PRs are more tightly vetted. See you on the next one!

PR #4920, opened 2026-05-01
Settle udev between sfdisk and lsblk

jandubois:
sfdisk's BLKRRPART triggers a uevent for the new partition. lsblk reads through libudev, which serves stale state until udevd processes that uevent. On Linux KVM Lima runs the uevent lags behind sfdisk's return: lsblk emits no `part` rows, PART becomes empty, and `mkfs.ext4 -L data-volume /dev/` dies with `Device size reported to be zero`. On slower hypervisors (VZ, QEMU on macOS) udevd wins the race and the bug stays latent.

Add `udevadm settle || true` between sfdisk and lsblk, matching the existing settle pattern (after growpart, after mkfs). Tighten the two existing udevadm comments to share the new one's structure.


unsuman (approved):
Thanks, was unable to test this locally but the changes looks safe

PR #4926, opened 2026-05-02
wsl2: detach cleanup from canceled ctx and surface terminate errors

### LINKED ISSUE #4923: wsl2: cleanup commands run with canceled context; errors silently swallowed
## Summary

Follow-up from the review discussion on #4892. The state-change goroutine in `pkg/driver/wsl2/vm_windows.go` has three related robustness issues that remain after the hot-loop fix was merged.

### 1. Cleanup runs with a canceled context

Both `stopVM` (`vm_windows.go:59`) and `getWslStatus` (`vm_windows.go:208`) pass `ctx` into `executil.RunUTF16leCommand` via `WithContext(ctx)`. By the time the `<-ctx.Done()` arm runs, that context is already canceled, so the `wsl.exe --list --verbose` and `wsl.exe --terminate` invocations will likely fail immediately with `context canceled`.

The cleanup we kept the goroutine alive for may not actually execute. Using `context.WithoutCancel(ctx)` (or `context.Background()` with a fresh timeout) for the cleanup commands would be more robust.

### 2. `getWslStatus` errors are silently swallowed

The condition `err == nil && status == StatusRunning` means any error (including a `context canceled` error from issue 1) causes the terminate path to be skipped entirely. At minimum, a `logrus.Warnf` on the error would aid debugging.

### 3. `_ = stopVM(...)` discards the terminate error without logging

The return value of `stopVM` is discarded with `_` and no log message is produced if the terminate fails. Adding a `logrus.WithError(err).Warn(...)` would match the logging style used elsewhere in the codebase.

## Suggested fix

- Use `context.WithoutCancel(ctx)` or `context.Background()` with a timeout for cleanup commands in the `<-ctx.Done()` path
- Log `getWslStatus` errors at warn level instead of silently skipping
- Log `stopVM` errors at warn level instead of discarding

cc @jandubois (per review feedback on #4892)

mn-ram:
Closes #4923. Follow-up to the review feedback on #4892.

After we kept the goroutine alive to actually run the terminate, I went back to look at what it does and realised the cleanup itself was broken in two related ways:

1. We pass `ctx` into `getWslStatus` and `stopVM`, but the only reason we're at this code path is because `ctx` is already cancelled. So `wsl.exe --list --verbose` and `wsl.exe --terminate` get killed by `CommandContext` more or less immediately and return `context canceled`. The whole "stop the VM on shutdown" path was effectively a no-op.

2. The `err == nil && status == StatusRunning` guard meant that whenever `getWslStatus` failed (which, per #1, was basically every shutdown), we silently skipped the terminate. And `_ = stopVM(...)` threw away the only signal that the terminate itself failed. So when this stopped working, nothing in the logs would tell you.

The fix is small:

- `context.WithoutCancel(ctx)` for the cleanup commands so they're not pre-cancelled, with a 30s timeout so we don't hang forever on a wedged `wsl.exe`.
- `logrus.WithError(err).Warn(...)` on both error paths instead of swallowing.

Diff is +17/-3, single file. Behaviour on the happy path (VM is `Running`, terminate succeeds) is unchanged.

## Test plan

- [x] `gofmt -l pkg/driver/wsl2/vm_windows.go` — clean
- [x] `GOOS=windows GOARCH=amd64 go vet ./pkg/driver/wsl2/...` — clean
- [ ] CI: Windows tests (WSL2)
  ```powershell
  limactl start --vm-type=wsl2 --name=demo template://default
  limactl stop demo
  Get-Process wsl  # distro should not show up
  ```
  Before: terminate silently fails on cancelled ctx; if you look at `wsl --list --verbose` after stop, the lima distro often still shows `Running`. After: distro is reliably terminated, and if anything goes wrong you see a clear `Failed to terminate WSL distro during shutdown` warning instead of nothing.

### DIFF
--- a/pkg/driver/wsl2/vm_windows.go
+++ b/pkg/driver/wsl2/vm_windows.go
@@ -14,6 +14,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
@@ -134,9 +135,22 @@ func provisionVM(ctx context.Context, instanceDir, instanceName, distroName stri
 
 		<-ctx.Done()
 		logrus.Info("Context closed, stopping vm")
-		if status, err := getWslStatus(ctx, instanceName); err == nil &&
-			status == limatype.StatusRunning {
-			_ = stopVM(ctx, distroName)
+		// ctx is already canceled at this point, so passing it to wsl.exe
+		// would make the cleanup commands fail with "context canceled"
+		// before they even run. Detach from the cancellation but keep a
+		// short timeout so we don't block forever on a wedged wsl.exe.
+		cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), 30*time.Second)
+		defer cancel()
+		status, err := getWslStatus(cleanupCtx, instanceName)
+		if err != nil {
+			logrus.WithError(err).Warn("Failed to query WSL status during shutdown; skipping terminate")
+			return
+		}
+		if status != limatype.StatusRunning {
+			return
+		}
+		if err := stopVM(cleanupCtx, distroName); err != nil {
+			logrus.WithError(err).Warn("Failed to terminate WSL distro during shutdown")
 		}
 	}()
 


mn-ram:
Hi @jandubois 👋

Following up on your review on #4892 — I’ve opened #4923 to track the issues you pointed out (cleanup running with a canceled ctx, swallowed getWslStatus errors, and the discarded stopVM error), and this PR addresses them.

The change is fairly small (+17/−3 in a single file). I’ve used context.WithoutCancel(ctx) with a 30s timeout for the cleanup commands so they don’t get pre-canceled, and added logrus.WithError(err).Warn(...) in both error paths so failures are visible.

Happy to tweak anything if you’d prefer a different approach (e.g., context.Background() or a different timeout). Thanks again for the helpful suggestions!

AkihiroSuda:
> - [ ] Manual on a Windows host:

This seems unchecked. @mn-ram Did you test it?

PR #4927, opened 2026-05-02
Auto-enable nestedVirtualization on Apple M4 hosts

### LINKED ISSUE #4912: GDB in a lima vm on an M4 host cannot run/attach-to processes
### Description

When you try to spin up a vm on an M4 host, and then use gdb you get this error: `Unable to fetch SVE/SSVE vector length: Invalid argument`.

This seems to have to do with the particular combination of instructions that processor has:

> platforms that provide an SVE context record without supporting SVE instructions (e.g. Apple M4 with SME streaming SVE under Virtualization.Framework).

Googling around this seems to have affected docker-desktop, dotnet, and graalVM as well.

Soumya-codr:
Fixes #4912 Automatically enables nestedVirtualization on Apple M4 hosts (macOS 15+) when using the VZ driver. This resolves the issue where GDB fails to fetch the SVE/SSVE vector length on M4 processors.

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/coreos/go-semver/semver"
 	"github.com/docker/go-units"
 	"github.com/goccy/go-yaml"
 	"github.com/pbnjay/memory"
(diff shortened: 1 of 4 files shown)


Soumya-codr:
The CI failed on the lychee link checker due to a network error reaching lima-vm.io which seems completely unrelated to my changes. Could a maintainer please re-run the failed jobs?

jandubois:
I've explained in #4912 why we will not enable nested virtualization by default.

However, even if we wanted to, this PR is using the wrong approach. It should use `unix.SysctlUint32("hw.optional.arm.FEAT_SME") == 1` instead of looking for M4, M5, M6, M7, M8, M9 strings and guessing about their capabilities.

PR #4935, opened 2026-05-06
fix(limactl): allow $EDITOR with flags

nyanotech:
Allows $VISUAL/$EDITOR to have flags in it, eg for `code -w`, or for some emacs setups

### DIFF
--- a/pkg/editutil/editorcmd/editorcmd.go
+++ b/pkg/editutil/editorcmd/editorcmd.go
@@ -23,7 +23,18 @@ package editorcmd
 import (
 	"os"
 	"os/exec"
+	"strings"
 )
+
+// ShellMetaChars is the set of characters that, when present in an editor
+// command string, require it to be executed via a shell rather than as a
+// direct argv. The list matches git's prepare_shell_cmd in run-command.c.
+const ShellMetaChars = "|&;<>()$`\\\"' \t\n*?[#~=%"
+
+// HasShellMeta reports whether s contains any character from ShellMetaChars.
+func HasShellMeta(s string) bool {
+	return strings.ContainsAny(s, ShellMetaChars)
+}
 
 // Detect detects a text editor command.
 // Returns an empty string when no editor is found.
@@ -39,6 +50,9 @@ func Detect() string {
 	for _, f := range candidates {
 		if f == "" {
 			continue
+		}
+		if HasShellMeta(f) {
+			return f
 		}
 		x, err := exec.LookPath(f)
 		if err == nil {
--- a/pkg/editutil/editutil.go
+++ b/pkg/editutil/editutil.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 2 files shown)


unsuman (changes requested):
Looks good but could you please fix this minor inconsistency?

unsuman:
There is some issue with Github, I cannot comment on the code:
Screenshot 2026-05-06 at 9 29 43 PM

> Investigating - We are investigating reports of degraded performance for Pull Requests
May 06, 2026 - 15:25 UTC

https://www.githubstatus.com/

jandubois:
Can't comment on the code either; here is what I wanted to write:

> This is very simplistic and doesn't support quoted names including whitespace.
>
> And while I think it doesn't matter, this is also a functional change: returns the command now instead of the resolved filename.
>
> I think `Detect` should have tests.

I would recommend looking at the `kubectl edit` implementation and copying the basic logic from there.

nyanotech:
> Can't comment on the code either; here is what I wanted to write:
> 
> > This is very simplistic and doesn't support quoted names including whitespace.edi
> > And while I think it doesn't matter, this is also a functional change: returns the command now instead of the resolved filename.
> > I think `Detect` should have tests.
> 
> I would recommend looking at the `kubectl edit` implementation and copying the basic logic from there.

Good point, I should've looked at some prior art before making this.

I'm thinking I'll do an implementation vaguely like [git's](https://github.com/git/git/blob/94f057755b7941b321fd11fec1b2e3ca5313a4e0/run-command.c#L288), where it passes to a shell if the `$EDITOR` contains spicy characters (including quotes, spaces, `~`), and if not, runs it directly. Compared to kubectl's, this drops the "split the args and then run it directly" path in favor of invoking a shell for those situations as well, and also has more characters in the list of ones it won't try to handle (such as `~`).

Adding extra logic to decide if a shell should be invoked or not makes the split between editutil.go/editorcmd.go feel a bit more intertwined than I'd like; any objections to moving the actual editor invocation to the latter (and keeping the file handling in the former), or merging them into one file?

pkg/editutil/editorcmd/editorcmd.go
```diff
@@ -23,8 +23,19 @@ package editorcmd
 import (
 	"os"
 	"os/exec"
+	"strings"
 )
 
+// ShellMetaChars is the set of characters that, when present in an editor
+// command string, require it to be executed via a shell rather than as a
+// direct argv. The list matches git's prepare_shell_cmd in run-command.c.
+const ShellMetaChars = "|&;<>()$`\\\"' \t\n*?[#~=%"
```
AkihiroSuda, line 32: We don't support this in other places such as `QEMU_SYSTEM_X86_64`

pkg/editutil/editorcmd/editorcmd.go
AkihiroSuda, line 1: Let's just use `shellwords.Parse`
https://github.com/lima-vm/lima/blob/90867ca062fdcec6cb42f694b2a88e7cfbacc833/pkg/driver/qemu/qemu.go#L1042-L1060

AkihiroSuda:
ping @nyanotech

PR #4937, opened 2026-05-07
test: add tests for pkg/must

varsha-0007:
This PR adds unit tests for pkg/must to improve test coverage and verify the expected behaviour of Must() in both success and failure scenarios.

The tests cover:

returning the provided object when the error is nil
panicking when a non-nil error is passed

This ensures the helper behaves correctly across both normal execution and error paths.

**Verification**

Tested locally with:

"go test ./pkg/must"

All tests pass successfully.

### DIFF
--- a/pkg/must/must_test.go
+++ b/pkg/must/must_test.go
@@ -1,0 +1,25 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package must
+
+import (
+	"errors"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestMustSuccess(t *testing.T) {
+	str := Must("string", nil)
+	assert.Equal(t, str, "string")
+}
+
+func TestMustPanic(t *testing.T) {
+	defer func() {
+		r := recover()
+		assert.Assert(t, r != nil, "Must should have panicked")
+	}()
+
+	Must("string", errors.New("test error"))
+}


pkg/must/must_test.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func TestMustSuccess(t *testing.T) {
+	expected := "hello"
+	actual := Must(expected, nil)
+	assert.Equal(t, expected, actual)
+
+	expectedInt := 42
+	actualInt := Must(expectedInt, nil)
+	assert.Equal(t, expectedInt, actualInt)
```
jandubois: The second test is redundant; the generic function is type-agnostic.

pkg/must/must_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+func TestMustPanic(t *testing.T) {
+	expectedErr := errors.New("test error")
+	defer func() {
+		r := recover()
+		assert.Assert(t, r != nil, "Must should have panicked")
+		assert.Equal(t, expectedErr, r)
+	}()
+
+	Must("hello", expectedErr)
```
jandubois: You need to add some failure case here, so the test actually fails if `Must` doesn't panic, e.g.

```go
t.Fatal("Must should have panicked")
```

jandubois (changes requested):

AkihiroSuda:
The test doesn't seem necessary for such a trivial function

AkihiroSuda:
The commits have to be squashed

jandubois (approved):
Thanks, LGTM

jandubois:
> The test doesn't seem necessary for such a trivial function

I do agree, but will merge this time. We should have given that feedback right away, not after asking for revisions. My fault.

varsha-0007:
Thanks for the feedback and for merging this! I completely understand regarding the triviality of the function. I'll keep that in mind for future contributions. Appreciate your time!

PR #4939, opened 2026-05-07
fix:track driver PID to prevent zombie processes

### LINKED ISSUE #4363: Force stop/remove leaves `lima-driver-*` processes running

Aneesh-Hegde:
Fix #4363 .
This PR implements PID tracking for external drivers by writing and reading a `driver.pid` file so the core engine can discover them.
It also refactors the `Stop()` function to gracefully terminate these processes (SIGTERM → Wait → SIGKILL fallback) to prevent zombies during `limactl rm -f`.
A follow-up PR will address the remaining edge case of forwarding SIGINT (Ctrl+C) during `limactl start`.

### DIFF
--- a/pkg/driver/external/server/server.go
+++ b/pkg/driver/external/server/server.go
@@ -15,6 +15,7 @@ import (
 	"os/exec"
 	"os/signal"
 	"path/filepath"
+	"strconv"
 	"sync"
 	"syscall"
 	"time"
@@ -207,6 +208,10 @@ func handlePreConfiguredDriverAction(ctx context.Context, driver driver.Driver) 
 	}
 }
 
+func driverPIDFilePath(instanceDir, driverName, pidFileOwner string) string {
+	return filepath.Join(instanceDir, fmt.Sprintf("lima-driver-%s-%s.pid", driverName, pidFileOwner))
+}
+
 // Start begins the driver startup process. It sends an initial response to unblock
 // the client and then streams subsequent errors(if any), as the driver initializes.
 // A final success message is streamed upon successful completion.
@@ -245,6 +250,15 @@ func Start(extDriver *registry.ExternalDriver, instName string) error {
 		return fmt.Errorf("failed to start external driver: %w", err)
 	}
 
+	pid := cmd.Process.Pid
+	pidFilePath := driverPIDFilePath(instanceDir, extDriver.Name, extDriver.PIDFileOwner)
+	// If the driver crashes immediately, the cleanup paths below will remove this file.
+	if err := os.WriteFile(pidFilePath, []byte(strconv.Itoa(pid)), 0o644); err != nil {
+		cancel()
+		_ = cmd.Process.Kill()
+		return fmt.Errorf("failed to write driver PID file: %w", err)
+	}
+
 	driverLogger := extDriver.Logger.WithField("driver", extDriver.Name)
 
 	scanner := bufio.NewScanner(stdout)
@@ -256,6 +270,8 @@ func Start(extDriver *registry.ExternalDriver, instName string) error {
 			if err := cmd.Process.Kill(); err != nil {
 				driverLogger.Errorf("Failed to kill external driver process: %v", err)
 			}
+			// TODO: reap process (e.g. cmd.Wait()) to avoid temporary zombies
+			_ = os.Remove(pidFilePath)
 			return fmt.Errorf("failed to parse socket path JSON: %w", err)
 		}
 		socketPath = output["socketPath"]
@@ -264,6 +280,7 @@ func Start(extDriver *registry.ExternalDriver, instName string) error {
 		if err := cmd.Process.Kill(); err != nil {
 			driverLogger.Errorf("Failed to kill external driver process: %v", err)
 		}
+		_ = os.Remove(pidFilePath)
 		return errors.New("failed to read socket path from driver")
 	}
 	extDriver.SocketPath = socketPath
@@ -274,6 +291,7 @@ func Start(extDriver *registry.ExternalDriver, instName string) error {
 		if err := cmd.Process.Kill(); err != nil {
 			driverLogger.Errorf("Failed to kill external driver process after client creation failure: %v", err)
 		}
+		_ = os.Remove(pidFilePath)
 		return fmt.Errorf("failed to create driver client: %w", err)
 	}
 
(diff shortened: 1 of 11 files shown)


pkg/store/instance.go
```diff
@@ -175,7 +175,18 @@ func inspectStatus(ctx context.Context, instDir string, inst *limatype.Instance,
 
 func inspectStatusWithPIDFiles(instDir string, inst *limatype.Instance, y *limatype.LimaYAML) {
 	var err error
-	inst.DriverPID, err = ReadPIDFile(filepath.Join(instDir, filenames.PIDFile(*y.VMType)))
```
unsuman, line 178: The `inst.DriverPID` should only hold the `<vm-type>.pid` because this is the actual PID of the VM or the hypervisor process

pkg/limatype/filenames/filenames.go
```diff
@@ -71,6 +71,7 @@ const (
 	QemuEfiCodeFD           = "qemu-efi-code.fd" // efi code; not always created
 	QemuEfiFullFD           = "qemu-efi-full.fd" // concatenated efi vars and code; not always created
 	AnsibleInventoryYAML    = "ansible-inventory.yaml"
+	DriverPIDFile           = "driver.pid"
```
unsuman: This can be little confusing, so maybe `lima-driver-<vm-type>.pid` is fine

unsuman (changes requested):
Starting a Lima instance launches two `limactl` processes (CLI and hostagent). So in external driver mode, there is a window that a pair of `lima-driver-<vm-type>` are also launched. The second process overwrites the `lima-driver-<vm-type>.pid` with the hostagent-launched external driver but after sometime the `limactl` cli process exits when the VM starts. So does the invoked external grpc server process through `server.StopAllExternalDrivers()`, which also results in deleting the `lima-driver-<vm-type>.pid` file as well. Now that the file is deleted, we cannot track or stop the external driver.

I'll have to think about this for sometime!

Aneesh-Hegde:
Addressed all feedback.
do you have any thoughts yet on how you'd prefer to track this `lima-driver `PID so the `limactl rm -f` command can find and kill it? Happy to implement whatever architecture you think is best!

unsuman:
I would suggest that the using of the external driver process PID should only be done for forceful instance removal/stop i.e. terminated by: 
https://github.com/lima-vm/lima/blob/50c551af9670cf28187e6d40045db45120d16852/pkg/instance/stop.go#L109

The external driver pids should be stored with a owner signature like `lima-driver-krunkit-ha.pid`. So during graceful exits these files should be cleaned up only by their parent processes.

Aneesh-Hegde:
Implemented the same as you mentioned @unsuman .PTAL
These are the changes done.
Track external driver PIDs via `lima-driver-<name>-<owner>.pid` files to kill orphaned processes on force stop/remove. Owner suffix (`ha`/`cli`) prevents the CLI from deleting the hostagent's PID file during startup. `StopForcibly` globs and kills all tracked driver processes. Graceful `Stop` uses SIGTERM->SIGKILL escalation with timeouts.

pkg/driverutil/instance.go
```diff
@@ -15,14 +15,15 @@ import (
 )
 
 // CreateConfiguredDriver creates a driver.ConfiguredDriver for the given instance.
-func CreateConfiguredDriver(inst *limatype.Instance, sshLocalPort int) (*driver.ConfiguredDriver, error) {
+func CreateConfiguredDriver(inst *limatype.Instance, sshLocalPort int, owner string) (*driver.ConfiguredDriver, error) {
```
unsuman:
```suggestion
func CreateConfiguredDriver(inst *limatype.Instance, sshLocalPort int, PIDNamespace string) (*driver.ConfiguredDriver, error) {
```

pkg/hostagent/hostagent.go
```diff
@@ -156,7 +156,7 @@ func New(ctx context.Context, instName string, stdout io.Writer, signalCh chan o
 		}
 	}
 
-	limaDriver, err := driverutil.CreateConfiguredDriver(inst, sshLocalPort)
+	limaDriver, err := driverutil.CreateConfiguredDriver(inst, sshLocalPort, "ha")
```
unsuman: Please define a constant for `ha` and `cli`, something like `driverutil.OwnerHostAgent`. The default should be `cli` if `""` is passed. Additionally a comment at the function signature would be great

pkg/instance/stop.go
```diff
@@ -116,6 +117,22 @@ func StopForcibly(inst *limatype.Instance) {
...
+	files, _ := filepath.Glob(driverPIDPattern)
+	for _, pidFile := range files {
+		if b, readErr := os.ReadFile(pidFile); readErr == nil {
+			if pid, convErr := strconv.Atoi(strings.TrimSpace(string(b))); convErr == nil && pid > 0 {
+				logrus.Infof("Sending SIGKILL to external driver process %d (from %s)", pid, filepath.Base(pidFile))
+				if proc, procErr := os.FindProcess(pid); procErr == nil {
+					if err := proc.Kill(); err != nil && !errors.Is(err, os.ErrProcessDone) {
+						logrus.Warnf("Failed to kill external driver process %d: %v", pid, err)
+					}
+				}
```
unsuman: You can reuse `osutil.SysKill()` here

pkg/registry/registry.go
```diff
@@ -34,6 +34,7 @@ type ExternalDriver struct {
 	Ctx          context.Context
 	Logger       *logrus.Logger
 	CancelFunc   context.CancelFunc
+	Owner        string
```
unsuman: Same make this to `PIDNamespace` and a comment explaining the purpose

pkg/driver/external/server/server.go
```diff
@@ -295,12 +313,48 @@ func Stop(extDriver *registry.ExternalDriver) error {
 	if extDriver.CancelFunc != nil {
 		extDriver.CancelFunc()
 	}
-	if err := extDriver.Command.Process.Signal(syscall.SIGTERM); err != nil {
-		extDriver.Logger.Errorf("Failed to kill external driver process: %v", err)
+	if extDriver.Command.Process != nil {
+		err := extDriver.Command.Process.Signal(syscall.SIGTERM)
```
unsuman, line 315: I think this fallback is still not right. Actually`extDriver.CancelFunc()` kills the process immediately. So the order should be: send SIGTERM first, wait for exit, and only use cancel/kill as the final fallback.

Aneesh-Hegde:
Hey @unsuman addressed the feedback. PTAL.

pkg/driverutil/instance.go
```diff
@@ -14,15 +14,24 @@ import (
...
 )
 
+const (
+	OwnerHostAgent = "ha"
+	OwnerCLI       = "cli"
+)
+
 // CreateConfiguredDriver creates a driver.ConfiguredDriver for the given instance.
-func CreateConfiguredDriver(inst *limatype.Instance, sshLocalPort int) (*driver.ConfiguredDriver, error) {
+// pidNamespace identifies the caller (e.g. "ha" or "cli") to isolate driver PID files.
```
AkihiroSuda: This terminology is confusing as it conflicts with
https://man7.org/linux/man-pages/man7/pid_namespaces.7.html

Aneesh-Hegde: Changed to `pidFileOwner`

pkg/driver/external/server/server.go
```diff
@@ -295,12 +313,59 @@ func Stop(extDriver *registry.ExternalDriver) error {
 	if extDriver.CancelFunc != nil {
 		extDriver.CancelFunc()
 	}
```
unsuman: I think this is safe to remove

Aneesh-Hegde: Done

unsuman (dismissed):
Thanks, LGTM!

jandubois:
> @unsuman: So the order should be: send SIGTERM first, wait for exit, and only use cancel/kill as the final fallback.

I don't think this has actually been addressed. CancelFunc() is essentially still a Kill().

Also, `os.FindProcess` is a no-op on Unix; it will never fail. You have to send a `0` signal to the process to find out if it exists and is still alive. Doesn't really matter though, as all it does it prevent a log entry from being written when the process is already gone.

pkg/driver/external/server/server.go
```diff
@@ -292,15 +310,60 @@ func Stop(extDriver *registry.ExternalDriver) error {
...
+		case waitErr := <-done:
+			if waitErr != nil {
+				extDriver.Logger.Debugf("External driver %s exited with: %v", extDriver.Name, waitErr)
+			} else {
+				extDriver.Logger.Debugf("External driver %s exited gracefully", extDriver.Name)
+			}
+		case <-time.After(stopTimeout):
+			extDriver.Logger.Warnf("External driver %s did not exit in time, triggering cancellation", extDriver.Name)
+			if extDriver.CancelFunc != nil {
+				extDriver.CancelFunc()
```
unsuman, line 338: ~~I couldn't catch this earlier but this should be removed from here and the `ExternalDriver` struct~~

[EDIT]: We just use CancelFunc() instead of SIGKILL

unsuman (commented):
Please fix this small issue then the PR is ready to merge

Aneesh-Hegde:
Hey @unsuman made required changes. Now it just uses CancelFunc(). Also changed the `os.FindProcess` logic to just sending syskill call based on @jandubois feedback.

pkg/driver/external/server/server.go
```diff
@@ -207,6 +208,10 @@ func handlePreConfiguredDriverAction(ctx context.Context, driver driver.Driver)
 	}
 }
 
+func driverPIDFilePath(instanceDir, driverName, pidFileOwner string) string {
```
AkihiroSuda, line 211: Please update https://lima-vm.io/docs/dev/internals/#instance-directory-lima_homeinstance to document the PID files and how they are used

Aneesh-Hegde: Added to docs PTAL

unsuman (dismissed):

pkg/driver/external/server/server.go
```diff
@@ -292,15 +310,52 @@ func Stop(extDriver *registry.ExternalDriver) error {
...
+		err := extDriver.Command.Process.Signal(syscall.SIGTERM)
+		if err != nil && !errors.Is(err, os.ErrProcessDone) {
+			extDriver.Logger.Errorf("Failed to send SIGTERM to external driver process: %v", err)
+		}
+
+		done := make(chan error, 1)
+		const stopTimeout = 10 * time.Second
+		go func() {
+			done <- extDriver.Command.Wait()
+		}()
```
jandubois, line 325: I think there is a race condition with the main thread setting `extDriver.Command = nil` at line 360. Not sure if it is reachable, but this feels safer:

```suggestion
		cmd := extDriver.Command
		go func() {
			done <- cmd.Wait()
		}()
```

Aneesh-Hegde:
Addressed the feedback. PTAL

jandubois:
Thanks, LGTM, but please fix the lint issue:

```
website/content/en/docs/security/_index.md:
	31: Trailing whitespace
```

and squash the commits

Aneesh-Hegde:
Done

jandubois (approved):
Thanks, LGTM

PR #4945, opened 2026-05-09
feat: make limactl respect $LIMA_INSTANCE env var

### LINKED ISSUE #4764: `limactl start` should start `$LIMA_INSTANCE` rather than `default`?
https://github.com/lima-vm/lima/blob/69751d55f37977edbecc6222b746e1b7b73b08cf/cmd/lima#L14

While the `lima` wrapper allows overriding `default` via an environment variable `$LIMA_INSTANCE`, the `limactl` CLI doesn't recognize this variable.

We may change the behavior of `limactl` to respect the env var, however, this kind of breaking change should probably happen when bumping up the major version (i.e., Lima v3.0)

Mysterio-17:
### Summary

The `lima` wrapper script respects `$LIMA_INSTANCE` to override the default instance name, but `limactl` subcommands always fell back to the hardcoded `"default"`. This PR makes `limactl` subcommands check `$LIMA_INSTANCE` when no instance name is given explicitly.

### Changes

- Replaced the exported `DefaultInstanceName` const with an unexported `defaultInstanceName()` function that reads `$LIMA_INSTANCE` and falls back to `"default"`
- Updated the following subcommands to use `defaultInstanceName()` as the fallback:
  `start`, `stop`, `restart`, `edit`, `factory-reset`, `start-at-login`, `guest-install`
- Updated `$LIMA_INSTANCE` documentation to reflect the expanded scope

### Tests

- Added `TestDefaultInstanceName` covering both the env-set and env-unset cases
- All existing tests pass

### Note

`limactl shell` is intentionally excluded from this PR, it has different arg parsing (instance name is a required positional arg) and needs separate UX consideration. A follow-up PR will address making `limactl shell` (with no args) also fall back to `$LIMA_INSTANCE`.

### DIFF
--- a/cmd/limactl/edit.go
+++ b/cmd/limactl/edit.go
@@ -54,7 +54,7 @@ func editAction(cmd *cobra.Command, args []string) error {
 	var inst *limatype.Instance
 
 	if arg == "" {
-		arg = DefaultInstanceName
+		arg = defaultInstanceName()
 	}
 	if err := dirnames.ValidateInstName(arg); err == nil {
 		inst, err = store.Inspect(ctx, arg)
--- a/cmd/limactl/factory-reset.go
+++ b/cmd/limactl/factory-reset.go
@@ -32,7 +32,7 @@ func newFactoryResetCommand() *cobra.Command {
 
 func factoryResetAction(cmd *cobra.Command, args []string) error {
 	ctx := cmd.Context()
-	instName := DefaultInstanceName
+	instName := defaultInstanceName()
 	if len(args) > 0 {
 		instName = args[0]
 	}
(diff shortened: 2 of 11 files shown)


Mysterio-17:
Hey @jandubois @unsuman @AkihiroSuda , have a look at this whenever you have time.

AkihiroSuda:
The code looks good, but I don't plan to merge this soon, as I said in:
- https://github.com/lima-vm/lima/issues/4764#issuecomment-4319718778

Mysterio-17:
> The code looks good, but I don't plan to merge this soon, as I said in:
> 
> * [`limactl start` should start `$LIMA_INSTANCE` rather than `default`? #4764 (comment)](https://github.com/lima-vm/lima/issues/4764#issuecomment-4319718778)

Yeah totally, I was just interested in working on this. The follow-up PR for `limactl shell` is also in progress, no rush on merging either.

Issue #4878, opened 2026-04-24
Consider dropping K8s packages


afbjorklund:
Considering stopping to use `pkgs.k8s.io`, and starting to use `dl.k8s.io` directly instead?

https://build.opensuse.org/project/subprojects/isv:kubernetes

https://downloadkubernetes.com/ (should both use the CDN)

This is because of the constant release time failures, when it comes to the OBS packages...

Downsides:

1) Binaries are uncompressed

2) Need to add systemd units.

See https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/install-kubeadm/#installing-kubeadm-kubelet-and-kubectl and "Debian-based distributions" vs "Without a package manager"

```
cd /usr/bin
sudo curl -L --remote-name-all https://dl.k8s.io/release/${RELEASE}/bin/linux/${ARCH}/{kubeadm,kubelet,kubectl}
sudo chmod +x {kubeadm,kubelet,kubectl}

RELEASE_VERSION="v0.16.2" # <-- this version is probably outdated for v1.35 and up
curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/krel/templates/latest/kubelet/kubelet.service" | sudo tee /usr/lib/systemd/system/kubelet.service
sudo mkdir -p /usr/lib/systemd/system/kubelet.service.d
curl -sSL "https://raw.githubusercontent.com/kubernetes/release/${RELEASE_VERSION}/cmd/krel/templates/latest/kubeadm/10-kubeadm.conf" | sudo tee /usr/lib/systemd/system/kubelet.service.d/10-kubeadm.conf
```

(I edited the instructions, to again include `kubectl` and to not relocate to `/usr/local`)

Most people don't want to run `curl` or `tee` as root either, but that's another story.

AkihiroSuda:
> This is because of the constant release time failures, when it comes to the OBS packages...

We are now pinning the dpkg version, so no need to use binary tarballs?
https://github.com/lima-vm/lima/blob/c902e72d5a378162ede31ef2841b6a0bf4330c82/templates/k8s.yaml#L57

AkihiroSuda:
> https://downloadkubernetes.com/

(Off-topic)
The domain once looked like a fake site to me, but it seems legit, as it is linked from k-sigs https://github.com/kubernetes-sigs/downloadkubernetes
But why don't they just use the `*.k8s.io` domain?

afbjorklund:
[...]

PR #4947, opened 2026-05-10
k8s: drop k8s packages and use binaries instead (refs #4878)

ashwat287:


templates/k8s.yaml
AkihiroSuda, line 1: I'm still not convinced:
- https://github.com/lima-vm/lima/issues/4878#issuecomment-4415868543

afbjorklund: It was a item for discussion, considering they dropped the ball three times.

But the added complexity is a problem, and the release dependency an old bug...
(ideally the kubernetes/kubernetes should depend on an external repository)

So I am inclined to hit the "snooze" button, until they also fail [Kubernetes 1.37 ](https://www.kubernetes.dev/resources/release/)?

afbjorklund (changes requested):
Thanks for trying it out, the downsides are indeed bigger than the upside (if any).

"Consider" means it was not decided, and I think we will continue with packages...

ashwat287:
Sure! I raised the PR expecting to initiate the discussion and progress on raised PR. 😄
Thank you!

PR #4948, opened 2026-05-10
Deprecate `_LIMA_QEMU_UEFI_IN_BIOS` flag

### LINKED ISSUE #4607: Deprecate `_LIMA_QEMU_UEFI_IN_BIOS` after QEMU 11.0.0 is released
### Description

Current QEMU tree contains fixes of the issue, which prevented using `pflash` with `whpx` accelerator. It will be delivered with 11.0.0 release, so, this workaround could be skipped altogether when Lima is used with the up to date QEMU.

There are different options to consider:
* switch default based on detected QEMU version (and optionally trigger warning output for informing user);
* abort starting with the older version of QEMU and provider error message (with the workaround code removal completely);
* stating in documentation that on Windows this is the minimal QEMU version supported and provide no diagnostics;
* something not explored in the above list

This is for backlog and not actionable before 11.0.0 releases are available.

arixmkii:
Fixes #4607

Changes default behavior on Windows based on QEMU version. Additionally logs deprecation warnings if the setting is used or if a fallback is triggered.

I manually tested it with QEMU 11.0.0 and 10.2.0.

### DIFF
--- a/pkg/driver/qemu/qemu.go
+++ b/pkg/driver/qemu/qemu.go
@@ -604,8 +604,9 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	noFirmware := *y.Arch == limatype.PPC64LE || *y.Arch == limatype.S390X || legacyBIOS
 	if !noFirmware {
 		var firmware string
-		firmwareInBios := runtime.GOOS == "windows"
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #4955, opened 2026-05-11
fix(validate): correct misleading mountType error on Windows

### LINKED ISSUE #4954: Misleading validation error: "virtiofs" fallback suggests "9p" on Windows, but "9p" is explicitly unsupported
### Description
While testing QEMU `mountType` configurations on a Windows host, I noticed a contradiction in the YAML validation logic. 

If a user specifies `mountType: "virtiofs"` on Windows, the parser throws an error suggesting `9p` is a valid alternative:
> `field 'mountType' must be "reverse-sshfs" or "9p" for QEMU driver on non-Linux...`

However, if the user follows this advice and sets `mountType: "9p"`, the validation immediately fails further down with:
> `mount type "9p" is not supported on Windows`

### Expected Behavior
On Windows hosts, the validation error for `virtiofs` should omit the `9p` suggestion and only direct users toward `reverse-sshfs`. (macOS behavior should remain unchanged).

InsidiousLion:
**Closes #4954**

### Description
This PR fixes a minor desync in the `mountType` validation logic specifically for Windows hosts.

Previously, `validateMountType` would suggest `"9p"` as a valid fallback for `"virtiofs"` on all non-Linux systems. However, `"9p"` is explicitly blocked on Windows later in the exact same function. This resulted in users being advised to use a mount type that would immediately crash the boot process.

This PR splits the `runtime.GOOS != "linux"` check:
- **Windows:** Now exclusively suggests `"reverse-sshfs"`.
- **macOS/Darwin:** Retains the original suggestion of `"reverse-sshfs" or "9p"`.

### Testing
- Verified that providing `mountType: "virtiofs"` on a Windows host now correctly outputs: `field 'mountType' must be "reverse-sshfs" for QEMU driver on Windows`.

### DIFF
--- a/pkg/driver/qemu/qemu_driver.go
+++ b/pkg/driver/qemu/qemu_driver.go
@@ -134,6 +134,12 @@ func validateArch(ctx context.Context, cfg *limatype.LimaYAML) error {
 // Helper method for mount type validation.
 func validateMountType(cfg *limatype.LimaYAML) error {
 	if cfg.MountType != nil && *cfg.MountType == limatype.VIRTIOFS && runtime.GOOS != "linux" {
+		// Windows explicitly does not support 9p, so we should not suggest it.
+		if runtime.GOOS == "windows" {
+			return fmt.Errorf("field `mountType` must be %q for QEMU driver on Windows, got %q",
+				limatype.REVSSHFS, *cfg.MountType)
+		}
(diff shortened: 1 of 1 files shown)


InsidiousLion:
Closing this to fix a DCO signature issue, opening a fresh PR momentarily!

PR #4956, opened 2026-05-11
fix(validate): correct misleading mountType validation error on Windows

### LINKED ISSUE #4954: Misleading validation error: "virtiofs" fallback suggests "9p" on Windows, but "9p" is explicitly unsupported

InsidiousLion:
**Closes #4954**

### Description
This PR fixes a minor desync in the `mountType` validation logic specifically for Windows hosts.

Previously, `validateMountType` would suggest `"9p"` as a valid fallback for `"virtiofs"` on all non-Linux systems. However, `"9p"` is explicitly blocked on Windows later in the exact same function. This resulted in users being advised to use a mount type that would immediately crash the boot process.

This PR splits the `runtime.GOOS != "linux"` check:
- **Windows:** Now exclusively suggests `"reverse-sshfs"`.
- **macOS/Darwin:** Retains the original suggestion of `"reverse-sshfs" or "9p"`.

### Testing
- Verified that providing `mountType: "virtiofs"` on a Windows host now correctly outputs: `field 'mountType' must be "reverse-sshfs" for QEMU driver on Windows`.

### DIFF
--- a/pkg/driver/qemu/qemu_driver.go
+++ b/pkg/driver/qemu/qemu_driver.go
@@ -133,15 +133,45 @@ func validateArch(ctx context.Context, cfg *limatype.LimaYAML) error {
 
 // Helper method for mount type validation.
 func validateMountType(cfg *limatype.LimaYAML) error {
-	if cfg.MountType != nil && *cfg.MountType == limatype.VIRTIOFS && runtime.GOOS != "linux" {
-		return fmt.Errorf("field `mountType` must be %q or %q for QEMU driver on non-Linux, got %q",
-			limatype.REVSSHFS, limatype.NINEP, *cfg.MountType)
-	}
+	// 1. First, check if the user explicitly blocked this mount type in their config.
 	if cfg.MountTypesUnsupported != nil && cfg.MountType != nil && slices.Contains(cfg.MountTypesUnsupported, *cfg.MountType) {
 		return fmt.Errorf("mount type %q is explicitly unsupported", *cfg.MountType)
 	}
-	if runtime.GOOS == "windows" && cfg.MountType != nil && *cfg.MountType == limatype.NINEP {
-		return fmt.Errorf("mount type %q is not supported on Windows", limatype.NINEP)
+
+	// 2. Then, check if the mount type is valid for the current Operating System.
+	if cfg.MountType != nil {
+		switch runtime.GOOS {
+		case "linux":
+			switch *cfg.MountType {
+			case limatype.REVSSHFS, limatype.NINEP, limatype.VIRTIOFS:
+				return nil
+			default:
+				return fmt.Errorf("field `mountType` must be one of %v for QEMU driver on Linux, got %q", []string{limatype.REVSSHFS, limatype.NINEP, limatype.VIRTIOFS}, *cfg.MountType)
+			}
+		case "darwin":
+			switch *cfg.MountType {
+			case limatype.REVSSHFS, limatype.NINEP:
+				return nil
+			default:
+				return fmt.Errorf("field `mountType` must be %q or %q for QEMU driver on macOS, got %q", limatype.REVSSHFS, limatype.NINEP, *cfg.MountType)
+			}
+		case "windows":
+			// Windows version of QEMU does not support 9p yet, so we should not suggest it.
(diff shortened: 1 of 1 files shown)


pkg/driver/qemu/qemu_driver.go
```diff
@@ -134,6 +134,12 @@ func validateArch(ctx context.Context, cfg *limatype.LimaYAML) error {
 // Helper method for mount type validation.
 func validateMountType(cfg *limatype.LimaYAML) error {
 	if cfg.MountType != nil && *cfg.MountType == limatype.VIRTIOFS && runtime.GOOS != "linux" {
+		// Windows explicitly does not support 9p, so we should not suggest it.
```
AkihiroSuda:
```suggestion
		// Windows version of QEMU does not support 9p yet, so we should not suggest it.
		// https://gitlab.com/qemu-project/qemu/-/work_items/974
```

pkg/driver/qemu/qemu_driver.go
```diff
@@ -134,6 +134,12 @@ func validateArch(ctx context.Context, cfg *limatype.LimaYAML) error {
...
+		if runtime.GOOS == "windows" {
+			return fmt.Errorf("field `mountType` must be %q for QEMU driver on Windows, got %q",
+				limatype.REVSSHFS, *cfg.MountType)
+		}
+		// For macOS/Darwin, 9p remains a valid suggestion.
 		return fmt.Errorf("field `mountType` must be %q or %q for QEMU driver on non-Linux, got %q",
 			limatype.REVSSHFS, limatype.NINEP, *cfg.MountType)
 	}
```
AkihiroSuda: Can we use switches like:
```go
switch runtime.GOOS {
  case "linux":
  switch *cfg.MountType {
    case limatype.REVSSHFS, limatype.NINEP, limatype.VIRTIOFS:
      return nil
    default:
      return fmt.Errorf("field `mountType` must be one of %v for QEMU driver on Linux, got %q", []string{limatype.REVSSHFS, limatype.NINEP, limatype.VIRTIOFS}, *cfg.MountType
  }
  case "darwin":
  ...
  case "windows":
  ...
  default:
  ...
}
```

InsidiousLion:
Thanks for the review and the guidance, @AkihiroSuda! I've applied the suggestions.

The validation logic is now cleanly separated using a switch runtime.GOOS block, and I've added the tracking link for the QEMU 9p issue. I also made sure to keep the MountTypesUnsupported check at the very top of the function so it correctly catches explicit bans before the switch block can trigger an early return.

Let me know if anything else is needed here!

pkg/driver/qemu/qemu_driver.go
```diff
@@ -133,21 +133,40 @@ func validateArch(ctx context.Context, cfg *limatype.LimaYAML) error {
...
+			// Windows version of QEMU does not support 9p yet, so we should not suggest it.
+			// https://gitlab.com/qemu-project/qemu/-/work_items/974
+			switch *cfg.MountType {
+			case limatype.REVSSHFS:
+				return nil
+			default:
+				return fmt.Errorf("field `mountType` must be %q for QEMU driver on Windows, got %q", limatype.REVSSHFS, *cfg.MountType)
+			}
+		default:
+			return nil
```
AkihiroSuda: virtiofs isn't supported on BSDs

pkg/driver/qemu/qemu_driver.go
AkihiroSuda, line 1: Please squash the commits

AkihiroSuda (approved):
Thanks

PR #4958, opened 2026-05-11
wsl2: unblock keepAlive / provisionVM errCh sends after instance stop

### LINKED ISSUE #4957: wsl2: keepAlive / provisionVM goroutines deadlock on unbuffered errCh after instance stop
### Description

The `keepAlive` goroutine in `pkg/driver/wsl2/vm_windows.go` (and its sibling writer in `provisionVM` in the same file) writes to an unbuffered `errCh` without a context-aware select, so the writer blocks indefinitely once the hostagent stops draining `errCh` during shutdown — the same shape of bug fixed for the VZ driver in #4922 and for the WSL2 hot-loop in #4892, but on a different code path that those PRs did not touch.

```go
// pkg/driver/wsl2/vm_windows.go
func keepAlive(ctx context.Context, distroName string, errCh chan<- error) {
    keepAliveCmd := exec.CommandContext(ctx, "wsl.exe", "-d", distroName, "bash", "-c",
        "nohup sleep 2147483647d >/dev/null 2>&1")
    go func() {
        if err := keepAliveCmd.Run(); err != nil {
            errCh <- fmt.Errorf("error running wsl keepAlive command: %w", err)
        }
    }()
}
```

`errCh` is allocated unbuffered in `(*LimaWslDriver).Start` (`wsl_driver_windows.go:245`):

```go
errCh := make(chan error)
```

It has exactly one consumer, in `(*HostAgent).startRoutinesAndWait`:

```go
select {
case driverErr := <-errCh:
    logrus.Infof("Driver stopped due to error: %q", driverErr)
case sig := <-a.signalCh:
    logrus.Infof("Received %s, shutting down the host agent", osutil.SignalName(sig))
}
// after this point, no more reads from errCh
if closeErr := a.close(); closeErr != nil { ... }
cancelHA()
return a.driver.Stop(ctx)
```

The race on every `limactl stop` of a WSL2 instance:

1. SIGTERM lands; the outer `select` picks the `signalCh` arm. Nobody will read `errCh` again.
2. `cancelHA()` cancels the driver ctx → `exec.CommandContext` sends `SIGKILL` to the `wsl.exe` subprocess.
3. `keepAliveCmd.Run()` returns with `signal: killed`.
4. The goroutine reaches `errCh <- fmt.Errorf(...)`.
5. **Send blocks forever** — unbuffered channel, no consumer, no `case <-ctx.Done()` fallback.

The goroutine remains parked on `chan send` for the rest of the hostagent process's lifetime, retaining the captured `*exec.Cmd` plus its stdio pipes.

The same hazard exists on the `errCh <- fmt.Errorf(...)` in `provisionVM`'s goroutine (also `vm_windows.go`).

```powershell
# wait for ready
limactl stop demo
```

The bug is structural — unbuffered channel + no consumer + no ctx-aware send — so it fires on every stop. In an in-process test that cancels the driver ctx without exiting the process, a `pprof.Lookup("goroutine")` snapshot shows one goroutine parked at `chan send` in `vm_windows.go:keepAlive.func1`.

### Fix

Mirror the VZ fix in #4922 inside the WSL2 driver:

1. Add a `trySendErr(ctx, errCh, err)` helper that selects on `ctx.Done()` so writers cannot block once the consumer is gone.
2. Use it at both writers in `keepAlive` and `provisionVM`.
3. Buffer `errCh` to size 2 in `(*LimaWslDriver).Start` so the first shutdown-time error is captured even before `trySendErr` is reached.

PR follows.

### Environment

Affects every WSL2 instance (`vmType: wsl2`) on Windows. Found by code inspection against `master`.

mn-ram:
Closes #4957.

Same shape of fix as #4922 (VZ) and #4892 (WSL2 hot-loop), but for the keepAlive / provisionVM writers in `pkg/driver/wsl2/vm_windows.go`, which the earlier PRs did not touch.

### What was broken

Two goroutines in this file write to the driver's `errCh`:

1. `keepAlive`'s supervisor goroutine, when `wsl.exe ... sleep 2147483647d` is killed by ctx cancellation.
2. `provisionVM`'s boot.sh supervisor, on the error path after `cmd.CombinedOutput()`.

`errCh` is allocated unbuffered in `(*LimaWslDriver).Start` (`wsl_driver_windows.go:245`) and has exactly one consumer — the outer `select` in `(*HostAgent).startRoutinesAndWait`. The moment SIGTERM picks the `signalCh` arm, nobody reads `errCh` again. Both writers then sit on an unbuffered send with no `<-ctx.Done()` fallback and block for the rest of the hostagent's lifetime, parked at `chan send`.

### What this PR does

- Adds a small `trySendErr(ctx, errCh, err)` helper in `vm_windows.go` that selects on `ctx.Done()`.
- Routes both writers (`keepAlive` and `provisionVM`) through it.
- Buffers `errCh` to size 2 in `(*LimaWslDriver).Start` so the first shutdown-time error from either writer is captured even before `trySendErr` is reached.

+22 / −5, two files. No API or architecture change.

### Test plan

- [x] `gofmt -l pkg/driver/wsl2/...` — clean
- [x] `GOOS=windows GOARCH=amd64 go build ./pkg/driver/wsl2/...` — clean
- [x] `GOOS=windows GOARCH=amd64 go vet ./pkg/driver/wsl2/...` — clean
- [ ] CI: Windows tests (WSL2)
  ```powershell
  limactl stop demo
  ```
  Before: a goroutine snapshot taken just before hostagent exit shows one goroutine parked at `chan send` in `vm_windows.go:keepAlive.func1`. After: that goroutine returns cleanly when the consumer is gone.

### DIFF
--- a/pkg/driver/wsl2/vm_windows.go
+++ b/pkg/driver/wsl2/vm_windows.go
@@ -127,9 +127,9 @@ func provisionVM(ctx context.Context, instanceDir, instanceName, distroName stri
 		os.RemoveAll(limaBootFileWinPath)
 		logrus.Debugf("%v: %q", cmd.Args, string(out))
 		if err != nil {
-			errCh <- fmt.Errorf(
+			trySendErr(ctx, errCh, fmt.Errorf(
 				"error running wslCommand that executes boot.sh (%v): %w, "+
-					"check /var/log/lima-init.log for more details (out=%q)", cmd.Args, err, string(out))
+					"check /var/log/lima-init.log for more details (out=%q)", cmd.Args, err, string(out)))
 		}
 
 		for {
@@ -159,10 +159,23 @@ func keepAlive(ctx context.Context, distroName string, errCh chan<- error) {
 
 	go func() {
 		if err := keepAliveCmd.Run(); err != nil {
-			errCh <- fmt.Errorf(
-				"error running wsl keepAlive command: %w", err)
+			trySendErr(ctx, errCh, fmt.Errorf(
+				"error running wsl keepAlive command: %w", err))
 		}
 	}()
+}
+
+// trySendErr is a non-blocking send that drops the error once ctx has been
+// cancelled. Without this guard, writers spawned from Start() block forever
+// when the hostagent abandons errCh during shutdown (after the signal
+// handler in startRoutinesAndWait exits its outer select), leaving a
+// non-terminating goroutine parked on `chan send` for the rest of the
+// process lifetime. Same shape of fix as #4922 for the VZ driver.
+func trySendErr(ctx context.Context, errCh chan<- error, err error) {
+	select {
+	case errCh <- err:
+	case <-ctx.Done():
+	}
 }
 
 // unregisterVM calls WSL to unregister a VM.
--- a/pkg/driver/wsl2/wsl_driver_windows.go
+++ b/pkg/driver/wsl2/wsl_driver_windows.go
@@ -242,7 +242,11 @@ func (l *LimaWslDriver) Start(ctx context.Context) (chan error, error) {
 		}
 	}
 
-	errCh := make(chan error)
+	// Buffered so that the first shutdown-time error from provisionVM /
+	// keepAlive does not block the writer if the hostagent has already
+	// stopped draining errCh (the trySendErr helper guards subsequent
+	// sends).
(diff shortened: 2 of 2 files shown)


unsuman:
@mn-ram needs rebase

PR #4968, opened 2026-05-12
Add Apparmor's fusermount3 custom rule for reverse SSHFS

### LINKED ISSUE #4908: We can't use SSHFS on default template (since ubuntu-25.04)
### Description

When I tried default template with Reverse SSHFS, limactl failed to mount my directory:
```bash
limactl start default --mount-type=reverse-sshfs
...
INFO[0029] [hostagent] Mounting "/home/mie31" on "/home/mi31"
INFO[0029] [hostagent] fusermount3: mount failed: Permission denied
...
```

This is because in Ubuntu-25.10, Apparmor sets a rule to restrict fusermount3.
The default rule prohibits us to mount to /${HOME} directory.
In the VM, I confirmed that the sshfs command did not work:
```bash
sshfs ":/home/mie31" "/home/mie31" -o slave -o ro -o allow_other
fusermount3: mount failed: Permission denied
```
(This is what limactl does by default)

I checked older versions of Ubuntu, and ~found that start from 25.10, the rule is set by default.~ EDIT: 25.04 also has the same problem. 24.10 doesn't have the rule.

This disturbs all hosts to use our default template with Reverse SSHFS.
(I guess this issue didn't appear because Linux host use 9p by default, by the way)

I came up with two solutions. 1) Drop Apparmor's fusermount3 rule, or 2) add custom rule then load it to Apparmor. In both cases, I think I can implement a solution by modifying some script in cloud-init.

I'm happy to create a PR once I confirm the solution with maintainers!

mie313:
Add a fusermount3 custom rule so that limactil can mount user's home directory using reverse SSHFS.

Related Issue: #4908

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
@@ -16,6 +16,18 @@ update_fuse_conf() {
 			echo "user_allow_other" >>"${fuse_conf}"
 		fi
 	fi
+
+	# Some distribution (since Ubuntu-25.04) has an apparmor rule for fusermount3. It causes SSHFS mount failed.
+	# Related Issue: https://github.com/lima-vm/lima/issues/4908
+	# Therefore, define a custom rule to loosen the apparmor rule.
+	if [ -e "/etc/apparmor.d/fusermount3" ] && [ ! -e "/etc/apparmor.d/local/fusermount3" ]; then
+		cat >"/etc/apparmor.d/local/fusermount3" <<EOF
+# The following two lines allow VM to be mounted to / unmounted from home directly.
+mount fstype=@{fuse_types} options=(nosuid,nodev) options in (ro,rw,noatime,dirsync,nodiratime,noexec,sync) -> @{HOME},
+umount @{HOME},
+EOF
+		apparmor_parser -r /etc/apparmor.d/fusermount3
+	fi
 }
 
 # update_fuse_conf has to be called after installing all the packages,


pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
```diff
@@ -16,6 +16,15 @@ update_fuse_conf() {
 			echo "user_allow_other" >>"${fuse_conf}"
 		fi
 	fi
+
+	# Some distribution (for example Ubuntu-25.10) has an apparmor rule for fusermount3. It causes SSHFS mount failed.
+	if [ -e "/etc/apparmor.d/fusermount3" ]; then
+		cat >"/etc/apparmor.d/local/fusermount3" <<EOF
```
AkihiroSuda, line 24: This file shouldn't be overwritten if it already exists

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
```diff
@@ -16,6 +16,15 @@ update_fuse_conf() {
...
+mount fstype=@{fuse_types} options=(nosuid,nodev) options in (ro,rw,noatime,dirsync,nodiratime,noexec,sync) -> @{HOME},
```
AkihiroSuda, line 26: Please add a comment to explain what it does

mie313: I added comments to both the shell script file and local/fusermount3 file to be created.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
mie313: Thank you for your review.
I modified the code to append new lines (`>>`) so that local/fusermount3 is not overwritten even if it already exists.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
AkihiroSuda: Now the file grows on every boot
```
# The following two lines allow VM to be mounted to / unmounted from home directly.
mount fstype=@{fuse_types} options=(nosuid,nodev) options in (ro,rw,noatime,dirsync,nodiratime,noexec,sync) -> @{HOME},
# The following two lines allow VM to be mounted to / unmounted from home directly.
mount fstype=@{fuse_types} options=(nosuid,nodev) options in (ro,rw,noatime,dirsync,nodiratime,noexec,sync) -> @{HOME},
# The following two lines allow VM to be mounted to / unmounted from home directly.
mount fstype=@{fuse_types} options=(nosuid,nodev) options in (ro,rw,noatime,dirsync,nodiratime,noexec,sync) -> @{HOME},
[...]
```

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
```diff
@@ -16,6 +16,17 @@ update_fuse_conf() {
 			echo "user_allow_other" >>"${fuse_conf}"
 		fi
 	fi
+
+	# Some distribution (for example Ubuntu-25.10) has an apparmor rule for fusermount3. It causes SSHFS mount failed.
+	# Therefore, define a custom rule to loosen the apparmor rule.
```
AkihiroSuda, line 22: Please add the URL of the issue in the code comment

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
```diff
@@ -16,6 +16,17 @@ update_fuse_conf() {
 			echo "user_allow_other" >>"${fuse_conf}"
 		fi
 	fi
+
+	# Some distribution (for example Ubuntu-25.10) has an apparmor rule for fusermount3. It causes SSHFS mount failed.
```
AkihiroSuda: Maybe (please confirm)
```suggestion
	# Some distribution (since Ubuntu-25.04) has an apparmor rule for fusermount3. It causes SSHFS mount failed.
```

AkihiroSuda (approved):
Thanks

PR #4970, opened 2026-05-13
[Lima 2.2.0] Set minimal QEMU version to 11.0.0 on Windows

### LINKED ISSUE #4949: Set minimal supported QEMU version 11.0.0 on Windows
### Description

Follow up to #4607, #4808, #4933, #4948 

This will allow to remove conditional branching for older QEMU workarounds. And would allow to completely remove piece of logic to support  concatenated firmware in `-bios`.

There is no known vendor providing freely available LTS releases on Windows, so, we could expect users on Windows to upgrade QEMU more frequently than on other platforms. 

Also, as QEMU support in Lima is a pretty new for Windows and Windows releases were added only recently we should not expect huge impact, because of the limited user base.

This is for Lima 2.2 or even later.

arixmkii:
This is for `2.2.0` or later. I need to try it once in the CI to see the reproducer for the root cause of the QEMU pining, because I tried it locally and in a third party GH repo with GHA and can't get the test failure. If CI fails I will try to investigate further w/o hogging CI for the project and postpone patch waiting for later QEMU if needed. There should be a QEMU point release in 2 weeks https://lists.nongnu.org/archive/html/qemu-devel/2026-05/msg02947.html, so, it might bring some fixes as well, but it is a bit iffy when/if the release will be built for Windows and when/if `winget` will update the repo.

Potentially fixes #4949

### DIFF
--- a/pkg/driver/qemu/qemu.go
+++ b/pkg/driver/qemu/qemu.go
@@ -9,7 +9,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io"
 	"io/fs"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


arixmkii:
I messed up lints, but it is ok for now because Windows tests are actually fine. I will factor out unpinning from this PR and re-submit and will rebase and force push the changes, when 2.2.0 merge window opens.

arixmkii:
Vulnchecks are not related to this changeset.

AkihiroSuda (approved):
Thanks

PR #4971, opened 2026-05-13
Unpin QEMU version in Windows tests

arixmkii:
Factored out changes from #4970 related to CI.

I didn't manage to reproduce failure on any setup running QEMU 11.0.0 and the tests passed at least in #4970 where prior version of QEMU are even unsupported. May be it is worth trying to unpin the QEMU version.

Alternatively we can wait for 2-3 weeks for 11.0.1 being released.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -235,10 +235,7 @@ jobs:
       run: make
     - name: Install QEMU
       run: |
-        # Pin to 10.2.0: QEMU 11.0.0 TCG breaks systemd boot on the Windows
-        # runner with "Failed to fork off sandboxing environment: Protocol
-        # error", blocking SSH.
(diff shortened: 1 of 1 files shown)


arixmkii:
@jandubois if you still have a CI failure reproducer I'm open to investigate it deeper.

jandubois:
> @jandubois if you still have a CI failure reproducer I'm open to investigate it deeper.

At the time **all** PRs failed the QEMU on Windows test, which is why I pinned it back to the previous version so we could keep the PR queue moving.

If CI no longer fails, then I'm happy to unpin. Looks like this PR is passing the test, so I'm fine merging it as soon as reruns of the flaky tests are green (on principle; I know they cannot be affected by unpinning QEMU on Windows).

Still curious why it no longer fails, but either the QEMU package got rebuilt, or there are some changes in the GitHub runner image that make it work now. Not worth spending time on, IMO.

jandubois (approved):
Thanks, LGTM

PR #4977, opened 2026-05-14
go.mod: github.com/containers/gvisor-tap-vsock v0.8.9

AkihiroSuda:
https://github.com/containers/gvisor-tap-vsock/releases/tag/v0.8.9
For `Fix issue where udp_proxy enters a tight loop polling the socket if the read returns 0 with err = nil.`

Full changes: https://github.com/containers/gvisor-tap-vsock/compare/v0.8.8...v0.8.9

(Not automatically updated due to dependabot's cooling down configuration)


unsuman (approved):
Thanks

PR #4978, opened 2026-05-14
Missing image and checksum for debian-13 riscv64

afbjorklund:
Tested manually that "latest" is working (in QEMU)

```yaml
- location: https://cloud.debian.org/images/cloud/trixie/daily/latest/debian-13-generic-riscv64-daily.qcow2
  arch: riscv64
```

### DIFF
--- a/templates/_images/debian-13.yaml
+++ b/templates/_images/debian-13.yaml
@@ -7,6 +7,12 @@ images:
 - location: "https://cloud.debian.org/images/cloud/trixie/20260413-2447/debian-13-genericcloud-arm64-20260413-2447.qcow2"
   arch: "aarch64"
   digest: "sha512:9de87e5e9739ec07fe61d40e5a6796e6e97e557ddb7d1eee5cf28619e80c21c0da76fc54713161e52c2719a38b5376e7b3bcfcd468e0b82ab9d33f702a09d601"
+- location: "https://cloud.debian.org/images/cloud/trixie/20260413-2447/debian-13-generic-ppc64el-20260413-2447.qcow2"
+  arch: "ppc64le"
+  digest: "sha512:d8d9345bfa563f95f42788218b2c18e1c7e1bbcb3c2c8a0724c2b98337280abcf9158e9e2483c40a893455669f36a20c7ea149cf1dcc6afbd9d138e96119ca8d"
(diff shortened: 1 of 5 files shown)


templates/_images/debian-13.yaml
```diff
@@ -16,7 +22,10 @@ images:
 - location: https://cloud.debian.org/images/cloud/trixie/daily/latest/debian-13-genericcloud-arm64-daily.qcow2
   arch: aarch64
 
-- location: https://cloud.debian.org/images/cloud/trixie/daily/latest/debian-13-genericcloud-ppc64el-daily.qcow2
+- location: https://cloud.debian.org/images/cloud/trixie/daily/latest/debian-13-generic-ppc64el-daily.qcow2
```
AkihiroSuda, line 25: debian-{11,12,sid} should be updated as well?

afbjorklund: That is possible, I haven't looked at when the arch were introduced

afbjorklund: It seems to depend on the date, they were still there in 20250703-2162

But the "latest" links are broken (404), and the riscv64 is missing (in  yaml).

afbjorklund: Debian 13 trixie is the first release with riscv64 support, though.

https://www.debian.org/News/2025/20250809

afbjorklund:
https://cloud.debian.org/images/cloud/

`generic`: Should run in any environment using cloud-init, for e.g. OpenStack, DigitalOcean and also on bare metal.

`genericcloud`: Identical to generic but with a reduced set of hardware drivers in the kernel. **If it does not work for your use case, you should use the generic images.**

afbjorklund:
The ppc64el images were removed, after 20250703-2162:

https://salsa.debian.org/cloud-team/debian-cloud-images/-/commit/7e528b2257fb07af65880a58a1726489b437c8b5

> Restrict genericcloud images to amd64 and arm64

templates/_images/debian-13.yaml
AkihiroSuda, line 1: Please also update sid.
sid seems to support s390x.
 https://cloud.debian.org/images/cloud/sid/daily/latest/

AkihiroSuda (approved):
Thanks

PR #4987, opened 2026-05-15
fix:prevent docker to run as root in rootless environment

### LINKED ISSUE #4986: `template:docker` starts both the rootless and rootful daemon
`template:docker` seems to start both the rootless and rootful daemon.

The rootful one shouldn't be started by default, as `template:docker-rootful` is expected to be used for rootful mode.

Aneesh-Hegde:
Fixes: #4986 
Disable and mask all rootful Docker and containerd units (`docker.service`, `docker.socket`, `containerd.service`, `containerd.socket`) to prevent the rootful daemon from starting alongside the rootless one.


AkihiroSuda (approved):
Thanks

PR #4990, opened 2026-05-16
Add script to extract the authors from git log

afbjorklund:
As per the AI discussion about authors, added a script to extract "The Lima Authors"

```
Lima
Copyright The Lima Authors.
```

`make AUTHORS`

(inspired by containerd AUTHORS)

The humans are still in the lead (I didn't add you to the mailmap, so you get two entries)

```
  2808	Akihiro Suda
  1286	Jan Dubois
   709	dependabot[bot] 
   337	Anders F Björklund
   185	Norio Nomura
```

The file has email, too.

But didn't list them here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 _output/
 _artifacts/
+AUTHORS
 lima.REJECTED.yaml
 default-template.yaml
 schema-limayaml.json
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,1 @@
+Anders F Björklund <anders.f.bjorklund@gmail.com>
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,10 @@ GO_BUILD := $(strip $(GO) build $(GO_BUILD_GCFLAGS) $(GO_BUILD_LDFLAGS) $(GO_BUI
 all: binaries manpages
 
 ################################################################################
+AUTHORS: .mailmap .git/HEAD
+	git shortlog --summary --email --group=Author --group=trailer:Co-Authored-By | cut -f2 | sort -fu > $@
+
+################################################################################
 # Help
 .PHONY: help
 help:
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,9 @@ all: binaries manpages
 
 ################################################################################
 AUTHORS: .mailmap .git/HEAD
-	git shortlog --summary --email --group=Author --group=trailer:Co-Authored-By | cut -f2 | sort -fu > $@
+	git shortlog --summary --email --group=Author --group=trailer:Co-Authored-By | \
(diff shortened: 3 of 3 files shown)


afbjorklund:
Besides dependabot, there's also:

```
     1	Claude Opus 4.5
     2	Claude Opus 4.6
     1	Claude Sonnet 4.6
     1	Copilot Autofix powered by AI
     1	copilot-swe-agent[bot]
     3	StepSecurity Bot
```

jandubois:
> Besides dependabot, there's also:

I think we should filter them out. For now matching `[bot]`, `<bot@`, or `<noreply@` catches them all: `grep -vE '\[bot\]|<(bot|noreply)@'`.

Is the plan to include the list in the repo itself? It will be a bit of a maintenance problem unless we lazily update it with e.g. a scheduled action.

afbjorklund:
> > Besides dependabot, there's also:
> 
> I think we should filter them out. For now matching `[bot]`, `<bot@`, or `<noreply@` catches them all: `grep -vE '\[bot\]|<(bot|noreply)@'`.

They will not be reported at all, when using `Assisted-by` instead of `Co-Authored-by` (or `Author`)
 
I was just curious about the current situation in the repository, and about the copyright vs author

> Is the plan to include the list in the repo itself? It will be a bit of a maintenance problem unless we lazily update it with e.g. a scheduled action.

No, only the script. The list is already included in the repo, just "hidden" in the `git log` information

If we want to _export_ something, we will need to use the `Signed-Off-by` and also include `vendor`...

afbjorklund:
Also: dependabot doesn't currently add any code directly, only code by others and indirectly.
i.e. the code is in the commits or versions being bumped, not in the line with the reference

So it is not really a part of the discussion about the AI agents, it is more of a tool (like `sed`)
Despite being currently listed as an "author", which is just a quirk of how git commits work?

`Signed-off-by: dependabot[bot] <support@github.com>`

----

EDIT: Seeing the changes introduced by dependabot would be an interesting tool by itself...

Like dumping "vendor" before and after commit, or try to find the external commits if any?

EDIT: It does include the "Changelog" and the "Commits" in the PR itself, though:

https://github.com/lima-vm/lima/pulls?q=is%3Apr+author%3Aapp%2Fdependabot+

So maybe it already has a local version as well, didn't check.

jandubois (approved):
I still don't quite understand why we need this; this seems a generic shell script you can run against any repo.

But it does work, so approving it, if you want it merged, but it would be useful to include a reference how we are going to use it.

afbjorklund:
As mentioned above the question was "who are the Lima authors", but maybe it (the Q) should be included too.

225 AUTHORS (but that number includes duplicate email addresses, for the same author - it needs `.mailmap`)

Issue #4907, opened 2026-04-29
[LFX Mentorship] Improve Windows support (guest and host)


unsuman:
This is a parent issue tracking for work done as part of the [LFX Mentorship 2026 Term 02](https://mentorship.lfx.linuxfoundation.org/) on expanding Lima's Windows support for host and guest:

> Description: Lima launches Linux virtual machines with automatic file sharing and port forwarding (similar to WSL2). Currently, Lima supports Linux, macOS, and FreeBSD as guest operating systems. The primary goal of this project is to expand this capability by adding support for Windows guests. Furthermore, the project aims to improve the stability and user experience of running Lima on Windows hosts. This will be achieved by removing dependencies like `cygpath.exe`, researching and developing a native Hyper-V(or a [HCS](https://learn.microsoft.com/en-us/virtualization/api/hcs/overview)) driver to provide optimized, native virtualization on Windows hosts.
> 
> - Expected Outcome:
>   - Primary:
>     - The ability to successfully launch and run Windows guest virtual machines using `limactl start template:windows` using the QEMU driver on any host.
>     - Deliver a seamless installation experience on Windows hosts by automating configurations and eliminating the need for [manual setup notes](https://github.com/microsoft/winget-pkgs/pull/356038/changes#diff-9c16ba1b4e8cfc88d634b6cc436e0041d2201ac93e1d4fcd78fca09e0667ca3aR29)
>     - Complete removal of the `cygpath.exe` dependency.
>   - Secondary (if time permits):
>     - Investigate and decide between Hyper-V and HCS for the native driver: HCS is the basis of WSL2 and may be available in Windows 11 Home Edition, while Hyper-V is only available in Pro/Enterprise editions. Research availability and integration feasibility.
>     - Potentially integrate the chosen driver (Hyper-V or HCS) as an [external VM driver](https://lima-vm.io/docs/dev/drivers/).
>   - Tertiary Goals (if time permits): Upgrade the existing WSL2 driver to drop image restrictions and allow users to run multiple instances, as well as exploring a simple graphical interface (`limagui.exe`) to launch virtual machines.

mundele2004:
Hi @unsuman ,

I’m very interested in the Lima Windows Support project, particularly because of the systems-level challenges involved in virtualization, platform compatibility, and native Windows integration.

[...]

PR #4993, opened 2026-05-17
ioutilx: replace cygpath with pure Go path conversion

varsha-0007:
Removes the dependency on cygpath.exe by replacing the external 
tool call in WindowsSubsystemPath with pure Go path conversion 
using filepath and strings packages.

This is part of the cygpath.exe removal work tracked in #4907.

- Converts Windows paths (e.g. C:\Users\foo) to Unix-style 
  paths (e.g. /c/Users/foo) natively in Go
- Adds tests for WindowsSubsystemPath


jandubois:
already implemented via #4993, #5270, #5284

PR #4998, opened 2026-05-17
Windows: support native OpenSSH; no Git for Windows / MSYS2 required

jandubois:
This PR has been created with assistance by Claude Opus 4.7.

It is a refactored and cleaned up version of #4885.

This PR consists of 9 commits that I've kept separate to ease review. I believe they should not be squashed for merging, so we can keep the separate commit messages.

## Summary

`limactl` now works on Windows hosts that have only the toolchain shipped in a default Windows 10/11 install (native OpenSSH, `wsl.exe`, `tar`). Lima previously required Git for Windows or MSYS2 on `PATH` for `cygpath`, `ssh`, `ssh-keygen`, `scp`, and `gzip`. After this branch, none of those external tools are required for the core flow on plain
Windows.

Two coupled root causes drove the historical Git-for-Windows / MSYS2 requirement:

1. **Native Windows OpenSSH does not implement SSH multiplexing** ([PowerShell/Win32-OpenSSH#1328](https://github.com/PowerShell/Win32-OpenSSH/issues/1328)). When Lima needed `ControlMaster` for the legacy ssh-based dynamic port forwarder, native ssh would not work, so Cygwin-built ssh was required.

2. **Cygwin-built ssh expects Cygwin-style paths** (`/c/Users/USER/...`), driving the `cygpath` dependency throughout the codebase.

Three things make the dependency droppable:

- **Lima's default port forwarder has been Go-native since v1.1.0** (`pkg/portfwd/forward.go`, gRPC-tunnelled via vsock). The legacy ssh-based forwarder is opt-in via `LIMA_SSH_PORT_FORWARDER=true`, so `ControlMaster` is not actually needed by the default flow.

- **Native Windows OpenSSH ships `sftp-server.exe`** (when the OpenSSH Server optional feature is installed) and is auto-detected by `sshocker`. Reverse-sshfs works natively once the host-path translation is corrected.

- **Native Windows OpenSSH treats `-F /dev/null` as an empty config**, the same way Cygwin ssh does. Lima's hardcoded `-F /dev/null` argument did not need to change.

## Detection mechanism

`sshutil.IsSSHCygwin(sshExe)` checks whether `cygpath.exe` lives in the same directory as `ssh.exe` (the layout used by Git for Windows and MSYS2), after resolving symlinks so chocolatey/scoop shims do not throw the directory check off. Results are cached per resolved absolute path.

`sshutil.PathForSSH(ctx, sshExe, path)` dispatches:

- Cygwin-based ssh → sibling `cygpath` for path translation (preserves any custom MSYS2 fstab the user has configured)

- Native Windows OpenSSH → `filepath.ToSlash` (e.g. `C:/Users/USER/...`), which native `ssh`, `ssh-keygen`, `scp`, and `sftp-server` accept

`sshutil.SftpServerForSSH(ctx, sshExe)` resolves an `sftp-server` binary that matches `ssh.exe`'s toolchain, so reverse-sshfs's locally-spawned sftp-server consumes paths in the form `PathForSSH` produces. Falls back to sshocker's PATH auto-detection (with a Warn log) when no match is found.

`ioutilx.WindowsSubsystemPath` keeps `cygpath` as the preferred backend but falls back to a native drive-letter conversion (`C:\Users\USER` → `/c/Users/USER`) when `cygpath` is unavailable, so the on-disk Lima config remains identical regardless of toolchain.

## CI changes

This PR introduces `.github/actions/windows_*` composite actions (host scrub, plain-host build, smoke test, WSL2 setup, QEMU install, MSYS2 prep, templates installer) and reorganizes the Windows jobs around them.

### Jobs that already existed on master

- **`windows-wsl2`** (formerly keyed `windows:`, renamed for symmetry with the new plain variant) and **`windows-qemu`** now build on the new composite actions instead of inlined steps. `_LIMA_WINDOWS_EXTRA_PATH` is gone from both; the toolchain they exercise is a strict subset of before.

### New jobs

- **`windows-plain-wsl2`** — builds with `go build` (no `make`, no MSYS2 bash), uninstalls MSYS2 and Git for Windows from the Windows runner, verifies across filesystem / PATH / smoking-gun binaries / registry that the host is genuinely vanilla, then runs a PowerShell smoke test (`create` → `start` → `shell` → `copy` → `stop` → `delete`) against `templates/experimental/wsl2.yaml`.

- **`windows-plain-qemu`** — same shape, with QEMU installed via winget and a templates installer that resolves Lima's template symlinks from git's object store (so the action does not need Developer Mode / admin to materialize NTFS symlinks). Smoke test exercises the same lifecycle plus a guest→host copy round-trip against `templates/default.yaml`.

Both new jobs:

- Pass `--debug` to every `limactl` invocation so the workflow log captures the toolchain-detection result, ssh args, path-translation decisions, etc.

- `if: failure()` step dumps `ha.stderr.log`, `ha.stdout.log`, `serial.log`, `lima.yaml`, `ssh.config` from the instance directory.

### DIFF
--- a/.github/actions/windows_msys2_prep/action.yml
+++ b/.github/actions/windows_msys2_prep/action.yml
@@ -1,0 +1,19 @@
+name: 'windows msys2 prep'
+description: >
+  Adds C:\msys64\usr\bin to PATH and installs the MSYS2 packages required by
+  hack/test-templates.sh. The PATH prepend persists into subsequent job steps
+  via $GITHUB_PATH, so the test step can call cygpath / bash / pacman tools
+  directly without re-prepending.
+runs:
+  using: composite
+  steps:
+  - name: Prepend MSYS2 to PATH and install test dependencies
+    shell: pwsh
+    # The value written to $GITHUB_PATH is the static literal C:\msys64\usr\bin,
+    # which is the canonical pattern for adding a directory to subsequent
+    # steps' PATH. No PR-controlled data reaches the env file.
+    run: | # zizmor: ignore[github-env]
+      $ErrorActionPreference = 'Stop'
+      Add-Content -Path $env:GITHUB_PATH -Value 'C:\msys64\usr\bin'
+      & 'C:\msys64\usr\bin\pacman.exe' -Sy --noconfirm openbsd-netcat diffutils socat w3m
+      if ($LASTEXITCODE -ne 0) { throw "pacman failed: $LASTEXITCODE" }
--- a/.github/actions/windows_plain_build/action.yml
+++ b/.github/actions/windows_plain_build/action.yml
@@ -1,0 +1,22 @@
+name: 'windows plain build'
+description: >
+  Builds limactl.exe and the Linux/amd64 guest agent with `go build` on a
+  vanilla Windows host (no MSYS2 make / bash). The guest agent must be built
+  separately because limactl looks it up at
+  _output/share/lima/lima-guestagent.Linux-x86_64 when starting an instance,
+  and `go build ./cmd/limactl` does not produce that file.
+runs:
+  using: composite
+  steps:
+  - name: Build limactl and Linux/amd64 guest agent
+    shell: pwsh
+    run: |
+      $ErrorActionPreference = 'Stop'
+      go build -o _output\bin\limactl.exe .\cmd\limactl
+      if ($LASTEXITCODE -ne 0) { throw "limactl build failed: $LASTEXITCODE" }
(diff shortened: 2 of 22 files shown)


arixmkii:
> support native OpenSSH; no Git for Windows / MSYS2 required

@jandubois Thank you for taking on this! This is actually the most correct way to have the implementation.

I will try to help with review/testing.

AkihiroSuda:
Can this be closed?

jandubois:
Superseded by #5196

PR #5002, opened 2026-05-18
fix(shell-sync): preserve synced guest changes on TUI interruption

### LINKED ISSUE #4997: limactl shell: synced directory is deleted when sync dialog does not exit gracefully
### Description

If a shell session is started with `--sync .` and then `exit` is called the user is presented with a prompt for whether or not they want sync the guest changes back to the host

```
? ⚠️ Accept the changes? (added: 0, deleted: 0, modified: 1, metadata: 1)  [Use arrows to move, type to filter]
> Yes
  No
  View the changed contents
```

If the process is then killed at this point (for example via SIGINT) then the directory on the guest is deleted. This means the work done in the guest is now gone without a straightforward way to recover.

I would expect the data in the guest to persist if limactl does not exit gracefully. This is the case when the diff pager is killed with SIGINT and when the parent process of the `limactl shell` is killed *before* exiting the shell, so I believe this behavior only exists when in the prompt dialog.

On MacOS 26.4.1 with `limactl version 2.1.1`

unsuman:
Fixes #4997

When `SIGINT` is sent during the exit dialogue of `limactl shell --sync`, changes made during the synced session are lost. This PR fixes that.

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -448,6 +448,7 @@ func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype
 	remoteSource := fmt.Sprintf("%s:%s", inst.Name, destRsyncDir)
 	clean := filepath.Clean(hostCurrentDir)
 	dirForCleanup := shellescape.Quote(filepath.Join(*inst.Config.User.Home, clean))
+	cleanupGuestWorkdir := false
 
 	rsyncBack := func() error {
 		paths := []string{
@@ -459,10 +460,14 @@ func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype
 			return fmt.Errorf("failed to sync back the changes from guest instance to host: %w", err)
 		}
 		logrus.Info("Successfully synced back the changes to host.")
+		cleanupGuestWorkdir = true
 		return nil
 	}
 
 	defer func() {
+		if !cleanupGuestWorkdir {
+			return
+		}
 		// Clean up the guest synced workdir
 		if err := executeSSHForRsync(ctx, *sshCmd, inst.SSHLocalPort, inst.SSHAddress, fmt.Sprintf("rm -rf %s", dirForCleanup)); err != nil {
 			logrus.WithError(err).Warn("Failed to clean up guest synced workdir")
@@ -479,6 +484,7 @@ func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype
 	}
 	if stats != nil && stats.String() == "" {
 		logrus.Info("No changes detected")
+		cleanupGuestWorkdir = true
 		return nil
 	}
 	statsMsg := ""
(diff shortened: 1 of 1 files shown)


cmd/limactl/shell.go
```diff
@@ -448,6 +448,7 @@ func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype
 	remoteSource := fmt.Sprintf("%s:%s", inst.Name, destRsyncDir)
 	clean := filepath.Clean(hostCurrentDir)
 	dirForCleanup := shellescape.Quote(filepath.Join(*inst.Config.User.Home, clean))
+	cleanupGuestWorkdir := true
```
jandubois: I think the logic should be reverted: preserve the data, unless it has been confirmed it has been successfully copied or the user decided to discard. That automatically covers other failure scenarios that are not being considered in this PR.
```suggestion
	cleanupGuestWorkdir := false
```

---

Possible other failures detected by Opus 4.7:

The defer still runs rm -rf on every other early-return path inside askUserForRsyncBack, several of which can fire after a successful guest session and would also silently destroy the user's work:
- cmd/limactl/shell.go:504 — `os.MkdirTemp` failure on the host
- cmd/limactl/shell.go:514 — `os.MkdirAll` failure on the host
- cmd/limactl/shell.go:527 — non-Interrupt error from `uiutil.Select` ("failed to open TUI")
- cmd/limactl/shell.go:549 — `rsyncDirectory` to the host temp dir fails (when user picked "View")
- cmd/limactl/shell.go:559/569/589/600/607 — pager/diff pipe failures
- cmd/limactl/shell.go:460 — even the happy "Yes" path: if the final rsyncBack rsync itself fails (disk full, network blip), the defer still wipes the guest dir, losing the only remaining copy.

cmd/limactl/shell.go
```diff
@@ -515,6 +519,11 @@ func askUserForRsyncBack(ctx context.Context, cmd *cobra.Command, inst *limatype
 	for {
 		ans, err := uiutil.Select(message, options)
 		if err != nil {
+			if errors.Is(err, uiutil.InterruptErr) {
+				cleanupGuestWorkdir = false
+				logrus.Infof("Interrupted by user; preserving guest synced workdir at %s", destRsyncDir)
+				return err
```
jandubois: I haven't tested it, but this looks like double-reporting: you show an INFO level log message, and then return with an error message. Maybe return the log message as an error instead?

jandubois (changes requested):
The logic needs to be inverted to cover other failure cases too.

cmd/limactl/shell.go
unsuman: Also, should we prompt the user during --sync if the target directory already exists?

If a user runs `limactl shell --sync` and Lima finds that the directory already exists in the guest, it should check for diff against the host. If there are any changes, Lima will prompt the user with this dialogue before starting the new session.

```
WARN[...] This directory already exists in the guest.
? ⚠️ Sync back changes to host? (added: 0, deleted: 0, modified: 1, metadata: 1)  [Use arrows to move, type to filter]
> Yes
  No (start fresh)
  No (keep guest changes)
  View the changed contents
```

jandubois (approved):
Thanks, LGTM

Issue #4852, opened 2026-04-19
Support for Windows guests


unsuman:
### Description

Lima currently supports three guest OS types (`Linux`, `Darwin`, `FreeBSD`). Guest provisioning is handled via cloud-init, which generates a cidata `ISO9660` image containing `cloud-config.yaml`, boot scripts, and provisioning data. None of this applies to Windows.

### Driver Constraints

- VZ (`Virtualization.framework`) cannot be used for Windows guests because it does not expose a virtual TPM (vTPM) device, which is [required by Windows 11](https://support.microsoft.com/en-us/windows/enable-tpm-2-0-on-your-pc-1fd5a332-360d-4f46-a1e7-ae6b0c90645c).

- WSL2 is also not applicable because it runs Linux distributions inside Windows, not the other way around.

- **QEMU is the preferred driver.** QEMU supports [TPM (via `swtpm`)](https://qemu-project.gitlab.io/qemu/specs/tpm.html#qemu-tpm-device), UEFI firmware with Secure Boot via OVMF, and the full range of virtual hardware Windows expects.

### Provisioning: `autounattend.xml` instead of cloud-init

Windows does not support cloud-init. Instead, unattended installation is driven by [Windows Answer Files](https://learn.microsoft.com/en-us/windows-hardware/manufacture/desktop/update-windows-settings-and-scripts-create-your-own-answer-file-sxs) (`autounattend.xml`). The answer file handles:

- Disk partitioning and OS installation
- User account creation
- Enabling OpenSSH Server for remote access
- Running first-logon scripts (similar to Lima's provisioning scripts)
- Injecting SSH public keys

afbjorklund:
macOS does not support cloud-init either, but there lima provides a "fakecloudinit" adapter

You could make a conversion shim for this, that would translate to the local implementation?

```
// Package fakecloudinit is a fake cloud-init implementation for macOS.
//
// TODO: support other OS that does not have cloud-init.
```

Nesar976:
Hi @unsuman , @afbjorklund , I’d like to work on this.

My plan is to introduce a small conversion layer similar to `fakecloudinit` that translates Lima provisioning into a Windows-compatible `autounattend.xml` setup. I’ll start with basic provisioning (disk setup, user creation, SSH access) and iterate from there.
Please let me know if this approach aligns with your expectations.

unsuman:
[...]

PR #5022, opened 2026-05-21
LFX: Windows guest support proof-of-concept

skataria8443:
Changes in this pull request:
 
## Secure Boot support

This is the first prerequisite for proper Windows guest support.

1. Add QEMU firmware descriptor support for selecting UEFI firmware.
2. Use descriptor-based firmware selection when Secure Boot is enabled through `firmware.secureBoot: true` in the template.
3. Secure Boot defaults to disabled, `preEnrollSecureBootKeys` defaults to the same value as `secureBoot`.
4. Keep the existing firmware selection path as the fallback when Secure Boot is disabled.
5. Keep the existing firmware selection path for legacy BIOS.
6. Support both raw and QCOW2 firmware code/vars files.
7. Support Secure Boot with pre-enrolled keys, typically Microsoft keys provided by the distribution firmware packages.
8. Support Secure Boot with an empty variable store, for guests such as Talos that enroll their own PK, KEK, and db.
9. Support custom QEMU firmware descriptor paths, useful when the distribution does not ship working Secure Boot firmware, e.g. Arch Linux: https://wiki.archlinux.org/title/QEMU#Enabling_Secure_Boot

## TPM support

This is the second prerequisite for proper Windows guest support.

1. Add TPM 2.0 support for QEMU using `swtpm`, which can be enabled through `firmware.tpm2: true` in the template.
2. Use the modern CRB TPM device model instead of legacy TIS.
3. TPM 2.0 defaults to disabled.

## Windows guest support

This adds initial support for automated Windows guest installation using QEMU, with limited configuration support.

1. Add a `windows` template for installing Windows from an ISO.
2. Generate `autounattend.xml` for unattended Windows installation.
3. Add the one-time boot-key workaround needed for Windows installer ISOs that prompt with "Press any key to boot from CD or DVD".
4. Add support for copying selected VirtIO drivers from a configured VirtIO driver ISO into the cidata ISO.
5. Install VirtIO storage and network drivers through `autounattend.xml`.
6. Create the default local Windows user account during unattended setup.
7. Generate and store a per-instance local user password when one is not provided in the template.
8. Install and enable OpenSSH Server during first boot.
9. Configure `administrators_authorized_keys` from Lima's SSH public keys for key-based SSH access.
10. Mark the first Ethernet network profile as Private as a temporary workaround so the OpenSSH firewall rule can apply; more precise networking configuration support will be worked on later.

## Testing and research notes

1. Tested on an Arch Linux x86_64 host with the QEMU driver using OVMF firmware from Fedora.
2. Tested Secure Boot support with Debian and Fedora guests.
3. Tested Secure Boot and TPM 2.0 support with Windows guests.
4. Support for non-QEMU drivers is not included in this pull request.
5. Support for non-Linux hosts and non-x86_64 architectures may be incomplete.
6. Legacy BIOS was not tested.

### DIFF
--- a/.ls-lint.yml
+++ b/.ls-lint.yml
@@ -32,6 +32,7 @@ ignore:
 - hack/common.inc.sh
 - pkg/cidata/cidata.TEMPLATE.d
 - pkg/cidata/cidata.TEMPLATE.d/util/compare_version.sh
+- pkg/cidata/windows.TEMPLATE.d
 - website
 # "ubuntu-24.04" does not follow the kebab-case
 - '**/ubuntu-*\.yaml'
--- a/pkg/cidata/cidata.go
+++ b/pkg/cidata/cidata.go
@@ -4,11 +4,13 @@
 package cidata
 
 import (
+	"bytes"
 	"compress/gzip"
 	"context"
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"maps"
 	"net"
 	"net/url"
@@ -21,6 +23,8 @@ import (
 	"time"
 	"unicode"
 
+	backendfile "github.com/diskfs/go-diskfs/backend/file"
+	"github.com/diskfs/go-diskfs/filesystem/iso9660"
 	"github.com/docker/go-units"
 	"github.com/sirupsen/logrus"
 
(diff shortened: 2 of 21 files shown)


skataria8443:
The linter errors should be fixed now.

jandubois:
AI code review (at the commit before you fixed the linter errors): https://jandubois.github.io/lima/20260521-115429-pr-5022.html

Note that every commit needs a valid DCO sign-off.

skataria8443:
I am working on improving Windows guest support to align with the requirements outlined in #4852. The current changes address most of the Windows guest provisioning flow (e.g. autounattend.xml setup and related Windows guest requirements), but there is still room for refinement and cleanup before the implementation is complete.

I will add DCO sign-offs to all commits in the next revision.

skataria8443:
@jandubois Thanks for sharing the AI code review. I will go through the issues pointed out in the review, verify them manually, and update the implementation where appropriate.

skataria8443:
This was a mentorship application PoC and is archived for reference only. It is not production-ready and should not be merged as-is. I am happy to revisit parts of it if the maintainers think any piece is useful.

PR #5017, opened 2026-05-20
Change %q in logs and errors to %#q [Part2]

mie313:
Change %q in logs and errors to %#q (all files in the following directories) .
- `cmd/lima-guestagent`
-  `pkg/apfs`
-  `pkg/autostart`
-  `pkg/cacheutil`
-  `pkg/cidata`
-  `pkg/copytool`
-  `pkg/downloader`
-  `pkg/driver`
-  `pkg/driverutil`
-  `pkg/editutil`
-  `pkg/envutil`
-  `pkg/fileutils`
-  `pkg/limayaml`

Related issue: #4898 

This PR is a continuation of the work from #4909 .

@jandubois 
I also addressed [your last comments in #4909 ](https://github.com/lima-vm/lima/pull/4909#pullrequestreview-4261131007) in this PR.

PR #5029, opened 2026-05-23
Change %q in logs and errors to %#q [Part3]

mie313:
Change %q in logs and errors to %#q and update the codes so that backticks are used everywhere in logs and errors. (all files in the following directories) .

- `pkg/guestagent`
- `pkg/guestpatch`
- `pkg/hostagent`
- `pkg/httpclientutil`
- `pkg/identifiers`
- `pkg/imgutil`
- `pkg/instance`
- `pkg/iso9660util`
- `pkg/limactlutil`
- `pkg/limainfo`
- `pkg/limatmpl`
- `pkg/limatype`

Related issue: #4898 
This PR is a continuation of the work from #4909 and #5017 .

### DIFF
--- a/pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
+++ b/pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
@@ -47,7 +47,7 @@ func enableSSHD(ctx context.Context) error {
 	cmd := exec.CommandContext(ctx, "/bin/launchctl", "load", "-w", "/System/Library/LaunchDaemons/ssh.plist")
 	logrus.Infof("Executing command: %v", cmd.Args)
 	if output, err := cmd.CombinedOutput(); err != nil {
-		return fmt.Errorf("failed to execute command %v: %w (output=%q)", cmd.Args, err, output)
+		return fmt.Errorf("failed to execute command %v: %w (output=%#q)", cmd.Args, err, output)
 	}
 	return nil
 }
@@ -56,7 +56,7 @@ func processMetaData(ctx context.Context, mnt string) error {
 	metaDataPath := filepath.Join(mnt, "meta-data")
 	metaDataB, err := os.ReadFile(metaDataPath)
 	if err != nil {
-		return fmt.Errorf("failed to read meta data file %q: %w", metaDataPath, err)
+		return fmt.Errorf("failed to read meta data file %#q: %w", metaDataPath, err)
 	}
 	var metaData cloudinittypes.MetaData
 	if err = yaml.Unmarshal(metaDataB, &metaData); err != nil {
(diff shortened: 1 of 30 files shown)


pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
```diff
@@ -222,22 +222,22 @@ func populateHomeDir(ctx context.Context, uid int, homedir string) error {
 func createUser(ctx context.Context, u *cloudinittypes.User) error {
 	homedir := u.Homedir
 	if homedir == "" {
-		return fmt.Errorf("homedir is required for user %q", u.Name)
+		return fmt.Errorf("homedir is required for user %#qq", u.Name)
```
jandubois: You have again included a couple of `%#qq` instances in the PR. How do they happen? Did you create this with AI without checking the output?

mie313: Apologies for the inconvenience, but I have never used AI at all for contributing lima.
Since I modified all codes at once and then divided them into a few PRs, typos the same as before might remain.
I will make sure to double-check not to contain any typos before turning a PR ready for review.

mie313: Fixed all typos and squashed :pray:

pkg/limatmpl/embed.go
```diff
@@ -568,7 +568,7 @@ func encodeScriptReason(script string) string {
 	line := 1
 	for i, r := range script {
 		if !(unicode.IsPrint(r) || r == '\n') {
-			return fmt.Sprintf("unprintable character %q at offset %d", r, i)
+			return fmt.Sprintf("unprintable character %#q at offset %d", r, i)
```
jandubois, line 571: Doesn't really matter, but the `#` makes no difference when printing runes, not strings. It will always use single quotes, not backticks.

pkg/hostagent/requirements.go
```diff
@@ -209,7 +209,7 @@ A possible workaround is to run "apt-get install sshfs" in the guest.
 `,
 		})
 		req = append(req, requirement{
-			description: "fuse to \"allow_other\" as user",
+			description: "fuse to `allow_other` as user",
```
jandubois: Using backticks here will put the whole strings in double-quotes, which will look odd because all other progress messages will show backticks. Either leave this with double-quotes for the inner string, or remove the quotes completely.

```suggestion
			description: "fuse to allow_other as user",
```

jandubois (commented):
Thanks, LGTM

I think I would like to see the "fuse to allow_other as user" string changed; I don't care about the other comment. Both are trivial.

alexandear (approved):

jandubois (approved):
Thanks, LGTM

Issue #5030, opened 2026-05-23
`Integration tests (vz) (default.yaml)` is failing after updating nerdctl to v2.3.1


AkihiroSuda:
> `Integration tests (vz) (default.yaml)` is failing
> https://github.com/lima-vm/lima/actions/runs/26283194704/job/77393125410
> ```
> + limactl shell default nerdctl run -d --name nginx -p 127.0.0.1:8080:80 ghcr.io/stargz-containers/nginx:1.19-alpine-org
> time="2026-05-22T14:04:22Z" level=warning msg="treating lima version \"15f5ce2\" from \"/Users/runner/.lima/default/lima-version\" as very latest release"
> 8f9b5014bb9e5eb02b1b2d8e05950af4d332dba07c61769450bff7079160233f
> + timeout 3m bash -euxc 'until curl -f --retry 30 --retry-connrefused http://127.0.0.1:8080; do sleep 3; done'
> + curl -f --retry 30 --retry-connrefused http://127.0.0.1:8080/
>   % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
>                                  Dload  Upload  Total   Spent   Left   Speed
> 
>   0      0   0      0   0      0      0      0                              0
>   0      0   0      0   0      0      0      0           00:01              0
>   0      0   0      0   0      0      0      0           00:02              0
>   0      0   0      0   0      0      0      0           00:03              0
> ```

_Originally posted by @AkihiroSuda in https://github.com/lima-vm/lima/issues/5024#issuecomment-4525999170_

jandubois:
I don't have time to look into this myself right now, but I got Claude w/ Opus 4.7 try to analyze it. Maybe it has some hints worth following.

---

## Issue 5030 — root-cause analysis

**Bisection.** The failure appeared when nerdctl 2.3.1 merged (commit `cfb9f31d`, CI run 26299823973). Pre-merge runs of `Integration tests (vz) (default.yaml)` passed; every push since then fails the same job.

**Symptom.** curl gets `(52) Empty reply from server` after ~30s — *not* connection refused. Something at macOS `127.0.0.1:8080` accepts the TCP handshake, never returns data, and closes.

### The smoking gun isn't nerdctl, it's the bundled RootlessKit

nerdctl 2.3.0 bumped `nerdctl-full` to **RootlessKit v3.0.0** (major version bump from v2.x). The v3.0.0 changelog headline:

[...]

PR #5024, opened 2026-05-22
nerdctl: update from v2.2.2 to v2.3.1

unsuman:
https://github.com/containerd/nerdctl/releases/tag/v2.3.1

PR #5031, opened 2026-05-24
Pin rootless `containerd` backend to `slirp4netns`

unsuman:
This is a trial PR to validate whether forcing `slirp4netns` solves the recent CI regression observed in #5030 since this is irreproducible locally.

See https://github.com/containerd/nerdctl/pull/4827

`Assisted-by: Claude Opus 4.6`

### DIFF
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -397,142 +397,17 @@ if [[ -n ${CHECKS["container-engine"]} ]]; then
 	limactl shell "$NAME" $sudo $CONTAINER_ENGINE pull --quiet ${nginx_image}
 	limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p 127.0.0.1:8080:80 ${nginx_image}
 
-	if ! timeout 3m bash -euxc "until curl -f --retry 15 --retry-connrefused http://127.0.0.1:8080; do sleep 3; done"; then
-		INFO "=== DEBUG: port forwarding failed (#5030) - FINAL diagnostic run ==="
+	sleep 3
+	if ! limactl shell "$NAME" $sudo $CONTAINER_ENGINE inspect nginx --format '{{.State.Running}}' 2>/dev/null | grep -q "true"; then
+		INFO "Container exited immediately (broken stdout pipe with detach-netns), retrying with redirected output"
+		limactl shell "$NAME" $sudo $CONTAINER_ENGINE rm -f nginx 2>/dev/null || true
+		limactl shell "$NAME" $sudo $CONTAINER_ENGINE run -d --name nginx -p 127.0.0.1:8080:80 \
+			--entrypoint sh ${nginx_image} -c 'exec >/dev/null 2>&1; exec /docker-entrypoint.sh nginx -g "daemon off;"'
+	fi
 
-		# 1. Full container inspect (not truncated)
-		INFO "[1/12] nerdctl inspect nginx"
-		limactl shell "$NAME" $sudo $CONTAINER_ENGINE inspect nginx 2>&1 || true
-
-		# 2. Container logs
-		INFO "[2/12] nerdctl logs nginx"
+	if ! timeout 3m bash -euxc "until curl -f --retry 30 --retry-connrefused http://127.0.0.1:8080; do sleep 3; done"; then
+		limactl shell "$NAME" $sudo $CONTAINER_ENGINE inspect nginx 2>&1 | head -40 || true
 		limactl shell "$NAME" $sudo $CONTAINER_ENGINE logs nginx 2>&1 || true
-
-		# 3. Read json-log file directly from LogPath
-		INFO "[3/12] Direct json-log file read"
-		# shellcheck disable=SC2016
-		limactl shell "$NAME" bash -c '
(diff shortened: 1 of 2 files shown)


unsuman:
## Summary of things done

**The Problem (GitHub issue #5030):**
- Lima's VZ (Apple Virtualization) integration tests started failing after nerdctl was updated to v2.3.1
- The nginx container port-forwarding test gets "Empty reply from server" — TCP connects but no data comes back

**Root Cause:**
- nerdctl 2.3.x bundles RootlessKit v3.0.0 (major version bump)
- RootlessKit v3 introduces `--source-ip-transparent=true` by default
- This installs iptables mangle/fwmark rules that break loopback port forwarding on macOS VZ

**What we tried and failed:**

1. Various retry/timeout approaches for the curl test — didn't help because the issue is fundamental routing, not timing
2. Placing the drop-in inside a `sudo -iu ... sh -ec '...'` heredoc — failed because the directory didn't exist yet
3. Using `chown -R` on `${LIMA_CIDATA_HOME}/.config/systemd` (too broad) — could break other services' ownership
4. Multiple CI trial pushes (trials 1–7+) — each revealed new issues or confirmed previous fixes were incomplete
5. Discovered a **second, pre-existing bug**: `test-port-forwarding.pl` was being passed `"${FILE}"` (a temp YAML path) instead of `"${NAME}"` (the instance name), causing QEMU tests to fail with "unmatched instances"

**What we've implemented (current state):**

1. **Fix 1 (QEMU bug):** In test-templates.sh line 133, changed `"${FILE}"` → `"${NAME}"` so `test-port-forwarding.pl` gets the correct instance name
2. **Fix 2 (VZ issue #5030):** In 40-install-containerd.sh, added a systemd drop-in that sets `CONTAINERD_ROOTLESS_ROOTLESSKIT_FLAGS=--source-ip-transparent=false` before containerd is installed. Uses `<<'EOF'` (no variable expansion) and scoped `chown -R` only on the leaf directory we created

**Current status:**
- Both fixes are committed on branch `fix-nerdctl-rootlesskit`
- Syntax checks pass
- CI was not triggered on the last push (likely a GitHub Actions issue — maybe needs a force push or the workflow isn't configured for this branch)

> Claude Opus 4.6

> Short-term. Consider reverting https://github.com/lima-vm/lima/pull/5024 until the fix lands; the failure burns CI time on every push.

IMO, we should revert back and publish Lima v2.1.2?

unsuman:
## CI Log Summary — Key Findings

### The Core Problem (Not What We Originally Thought)

**The nginx container is CRASHING, not just unreachable.**

```
"Status": "exited"
"ExitCode": 1
"StartedAt": "21:29:37"
"FinishedAt": "21:29:38"   ← died in ~1 second
```

The JSON log file is **0 bytes** — nginx produced no output before dying. This is fundamentally different from "empty reply from server" — **nginx never starts at all**.

---

### What the 12 Diagnostic Tests Show

| Test | Result | Notes |
|------|--------|-------|
| 1. `nerdctl inspect nginx` | Container exited (ExitCode 1) | Port bindings are **empty `{}`** |
| 2. `nerdctl logs nginx` | **Empty** | 0-byte log file |
| 4. `nerdctl run echo` | ✅ Works | Basic containers function fine |
| 5. Filesystem check | ✅ Works | Image content is correct |
| 6. nginx `--network none` | ✅ Works | nginx starts and stays alive |
| 7. nginx `--network host` | ✅ Works | nginx starts and stays alive |
| 8. Manual nginx `-p 8082:80` foreground | ✅ Works | nginx starts, listens on port 80 |
| 12. busybox httpd | ❌ Script bug | `-d` and `--rm` can't be used together in nerdctl |

---

### Important Observations

**1. Rootlesskit command line (from test 9):**
```
rootlesskit --net=gvisor-tap-vsock ... --port-driver=builtin --detach-netns ...
```
- **No `--source-ip-transparent` flag visible** — our fix's env var may not have been active in this run (likely a pre-fix diagnostic run)
- Uses `gvisor-tap-vsock`, not `slirp4netns`

**2. Repeated AppArmor denials (dmesg):**
```
apparmor="DENIED" operation="change_onexec" info="label not found" name="nerdctl-default"
```
- Happens on every container start (tests 4, 5, and all retries)
- But NOT the root cause — echo and filesystem tests still succeed despite this

**3. Port bindings are empty in the failing container's inspect:**
```json
"PortBindings": {}
"Ports": {}
```
Port setup was never recorded, suggesting the container failed **during or before CNI/port setup**.

**4. Multiple shim deaths in the journal** — the test script retried the nginx run 5+ times, each container crashing within 3-5 seconds.

---

### Revised Root Cause Hypothesis

The failure is: **RootlessKit v3's builtin port driver with `--source-ip-transparent=true` causes the container to crash entirely on VZ** (not just misroute packets). The iptables TPROXY rules it tries to install likely fail in the VZ kernel environment, and this failure is fatal to the container startup.

Our fix (`--source-ip-transparent=false`) should still be correct — but this log is from a **pre-fix CI run** (the diagnostic code added earlier). The real question is whether the drop-in was actually loaded, which requires a fresh CI run with our current branch.

---

### Test Script Bug Found in These Logs

```bash
nerdctl run --rm -d ...   # ← invalid, nerdctl doesn't support -d + --rm together
```
This is in the busybox test in the debug block — separate from the main nginx test, and already in a debug-only block so it's lower priority.

[logs-from-ci.txt](https://github.com/user-attachments/files/28261098/logs-from-ci.txt)

PR #5041, opened 2026-05-26
pkg/store: skip containerd.user default for macOS aarch64

### LINKED ISSUE #5037: FillDefault enables containerd.user for macOS guests on aarch64 — nerdctl unnecessarily downloaded

## Root cause

```go
if y.Containerd.User == nil {
    switch *y.Arch {
    default:
    }
}
```

## Why it surfaced now

## Suggested fix

```go
if y.Containerd.User == nil {
    switch *y.Arch {
        } else {
        }
    default:
    }
}
```

## Workaround

```yaml
containerd:
  system: false
  user: false
```

## Environment

MD-Mushfiqur123:
Fixes #5037

### DIFF
--- a/hack/test-templates.sh
+++ b/hack/test-templates.sh
@@ -12,6 +12,14 @@ scriptdir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # shellcheck source=common.inc.sh
 source "${scriptdir}/common.inc.sh"
 
+# Check required prerequisites before doing anything else
+for _cmd in limactl jq curl timeout diff perl; do
+	if ! command -v "$_cmd" &>/dev/null; then
+		ERROR "Required prerequisite not found: $_cmd"
+		exit 1
+	fi
+done
+
 if [ "$#" -ne 1 ]; then
 	ERROR "Usage: $0 FILE.yaml"
 	exit 1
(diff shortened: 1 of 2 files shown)


hack/test-templates.sh
```diff
@@ -12,6 +12,14 @@ scriptdir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
...
 source "${scriptdir}/common.inc.sh"
 
+# Check required prerequisites before doing anything else
+for _cmd in limactl jq curl timeout diff perl; do
+	if ! command -v "$_cmd" &>/dev/null; then
+		ERROR "Required prerequisite not found: $_cmd"
+		exit 1
+	fi
+done
+
```
AkihiroSuda, line 22: Please remove irrelevant commit

MD-Mushfiqur123:
Friendly ping — this PR is ready for review. All CI checks are passing.

AkihiroSuda:
> Friendly ping — this PR is ready for review. All CI checks are passing.

No. DCO check is failing.

hack/test-templates.sh
AkihiroSuda: ping @MD-Mushfiqur123

AkihiroSuda:
Not mergeable as it contains an irrelevant commit
https://github.com/lima-vm/lima/pull/5041#discussion_r3304331940

AkihiroSuda (changes requested):
https://github.com/lima-vm/lima/pull/5041#issuecomment-4591365505

AkihiroSuda:
Merged:
- https://github.com/lima-vm/lima/pull/5090

PR #5043, opened 2026-05-26
portfwd: fix gRPC tunnel connection leak

### LINKED ISSUE #5042: gRPC port forwarder leaks TCP connections in CLOSE_WAIT until MaxConcurrentStreams is exhausted
### Description

Lima’s gRPC port forwarder appears to leak one host-side TCP connection per request when forwarding to an HTTP keep-alive server inside the guest. The leaked connections remain in `CLOSE_WAIT` in the `limactl` hostagent process. Once the count reaches about `2048`, port forwarding stops working, which matches the `grpc.MaxConcurrentStreams(2048)` limit configured in `pkg/guestagent/api/server/server.go`.

This affects normal forwarded ports, not only privileged/pseudoloopback ports. I originally noticed it with Kubernetes API forwarding on `127.0.0.1:6443`, and reproduced it with nginx on `18000`.

### Environment
- Lima revision: `939f4137cda18678ee3c771eaa353f88e1b9a786` (local builded, but same in brew version)
- Host OS: `macOS 15.0.1 (24A348)`
- VM type: `vz`
- Guest image: default template, Ubuntu 25.10
- Port forwarder: gRPC port forwarder (`LIMA_SSH_PORT_FORWARDER` is not enabled)

### Reproduction
Start a default Lima instance:

```bash
limactl start test
```

Inside the VM:

```bash
sudo apt-get update
sudo apt-get install -y nginx
sudo sed -i 's/listen 80 default_server;/listen 18000 default_server;/' /etc/nginx/sites-available/default
sudo sed -i 's/listen \[::\]:80 default_server;/listen [::]:18000 default_server;/' /etc/nginx/sites-available/default
sudo nginx -t
sudo systemctl restart nginx
```

On the host, generate requests:

```bash
while true; do curl http://127.0.0.1:18000; done
```

In another host terminal, watch the leaked connections:

```bash
watch -n 0.1 'lsof -i :18000 | wc -l'
```

### Actual Result
The number of open connections owned by the `limactl` hostagent process steadily increases. The connections remain in `CLOSE_WAIT`.

At `2049` entries, forwarding stops working. This is close to the configured gRPC stream limit:

```go
grpc.MaxConcurrentStreams(2048)
```

After this point, forwarded networking becomes unusable.

### Expected behavior

Short-lived TCP connections through the gRPC port forwarder should be fully released after the client closes the connection.

The number of open FDs should stay roughly stable, and port forwarding should continue working after more than 2048 requests.

### Notes

The number of leaked connections grows until it reaches roughly `2048`, which matches the gRPC server `MaxConcurrentStreams` setting.

I have a local patch that fixes the leak by changing the lifecycle handling of the gRPC tunnel used by the port forwarder. I will send it as a PR.

### Impact

This can break long-running Lima instances that expose frequently used forwarded ports.

For example, a local Kubernetes cluster inside Lima may become unreachable from the host after enough requests to the forwarded Kubernetes API port, while continuing to work normally from inside the VM.

vasileknik76:
### Description

This PR fixes a file descriptor leak in the gRPC port forwarder.

`GrpcClientRW` is used as the remote side of the proxied connection. The proxy implementation relies on half-close semantics and calls [CloseRead]() / `CloseWrite` when one copy direction finishes.

Before this change, `GrpcClientRW` only implemented `Close()`. Because of that, the gRPC tunnel write side was not closed when the host-side TCP client closed the connection. The reverse copy direction could remain blocked in `Recv()`, keeping the accepted host-side TCP socket in `CLOSE_WAIT`.

After enough short-lived forwarded connections, these leaked streams reached the gRPC `MaxConcurrentStreams` limit and port forwarding stopped working.

This change implements half-close handling for `GrpcClientRW`, so the gRPC stream is closed at the point where the proxy expects it to be closed and the TCP connection can be released.

Closes #5042.

### Testing

Reproduced the issue with nginx inside a Lima VM:

```bash
limactl start test
```

Inside the VM:

```bash
sudo apt-get update
sudo apt-get install -y nginx

sudo sed -i 's/listen 80 default_server;/listen 18000 default_server;/' /etc/nginx/sites-available/default
sudo sed -i 's/listen \[::\]:80 default_server;/listen [::]:18000 default_server;/' /etc/nginx/sites-available/default

sudo nginx -t
sudo systemctl restart nginx
```

On the host:

```bash
while true ; do curl http://127.0.0.1:18000; done
```

In another host terminal:

```bash
watch -n 0.1 'lsof -i :18000 | wc -l'
```

Before this change, the number of open FDs kept growing until forwarding stopped around `2048` connections.

After this change, the FD count stays stable and forwarding continues to work.

### DIFF
--- a/pkg/portfwd/client.go
+++ b/pkg/portfwd/client.go
@@ -5,6 +5,7 @@ package portfwd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"sync/atomic"
@@ -51,17 +52,20 @@ func DialContextToGRPCTunnel(client *guestagentclient.GuestAgentClient) func(ctx
 	var connectionCounter atomic.Uint32
 	return func(_ context.Context, network, addr string) (net.Conn, error) {
 		// Passed context.Context is used for timeout on initiate connection, not for the lifetime of the connection.
-		// We use context.Background() here to avoid unexpected cancellation.
-		stream, err := client.Tunnel(context.Background())
+		// Use a dedicated context so CloseRead can unblock a goroutine waiting in stream.Recv().
+		streamCtx, cancel := context.WithCancel(context.Background())
+		stream, err := client.Tunnel(streamCtx)
 		if err != nil {
+			cancel()
 			return nil, fmt.Errorf("could not open tunnel for addr: %s error:%w", addr, err)
 		}
 		// Handshake message to start tunnel
 		id := fmt.Sprintf("%s-%s-%d", network, addr, connectionCounter.Add(1))
 		if err := stream.Send(&api.TunnelMessage{Id: id, Protocol: network, GuestAddr: addr}); err != nil {
+			cancel()
 			return nil, fmt.Errorf("could not start tunnel for id: %s addr: %s error:%w", id, addr, err)
 		}
-		rw := &GrpcClientRW{stream: stream, id: id, addr: addr, protocol: network}
+		rw := &GrpcClientRW{stream: stream, cancel: cancel, id: id, addr: addr, protocol: network}
 		return rw, nil
 	}
 }
@@ -72,6 +76,7 @@ type GrpcClientRW struct {
 
 	protocol string
 	stream   api.GuestService_TunnelClient
+	cancel   context.CancelFunc
 }
 
 var _ net.Conn = (*GrpcClientRW)(nil)
@@ -100,7 +105,24 @@ func (g *GrpcClientRW) Read(p []byte) (n int, err error) {
 
 func (g *GrpcClientRW) Close() error {
 	logrus.Debugf("closing GrpcClientRW for id: %s", g.id)
-	return g.stream.CloseSend()
+	err := g.stream.CloseSend()
+	g.cancel()
+	return err
+}
+
+func (g *GrpcClientRW) CloseRead() error {
+	logrus.Debugf("closing read GrpcClientRW for id: %s", g.id)
+	g.cancel()
+	return nil
+}
+
+func (g *GrpcClientRW) CloseWrite() error {
+	logrus.Debugf("closing write GrpcClientRW for id: %s", g.id)
+	err := g.stream.CloseSend()
+	if errors.Is(err, context.Canceled) {
+		return nil
+	}
+	return err
 }
 
 func (g *GrpcClientRW) LocalAddr() net.Addr {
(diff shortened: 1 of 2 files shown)


pkg/portfwd/client.go
AkihiroSuda, line 1: Thanks, would it be possible to have unit tests?

vasileknik76: Sure

vasileknik76: Added test and remove waste sync.Once for `cancel()`.

If you try run this test on clean master it will fails with "time out" because old version doesn't handle close normally

AkihiroSuda:
Please squash the commits

pkg/portfwd/client.go
```diff
@@ -100,7 +105,22 @@ func (g *GrpcClientRW) Read(p []byte) (n int, err error) {
 
 func (g *GrpcClientRW) Close() error {
 	logrus.Debugf("closing GrpcClientRW for id: %s", g.id)
-	return g.stream.CloseSend()
+	err := g.stream.CloseSend()
+	g.cancel()
+	return err
+}
+
+func (g *GrpcClientRW) CloseRead() error {
```
unsuman, line 113: The existing `Close()` methods for the portfwd server have `logrus.Debugf()`, would be great to have it here as well

pkg/portfwd/client_test.go
```diff
@@ -0,0 +1,97 @@
...
+		if errors.Is(err, io.EOF) {
+			s.tunnelDone <- nil
+			return nil
+		}
+		s.tunnelDone <- err
+		return err
+	}
+}
+
+func (s *testGuestService) GetInfo(context.Context, *emptypb.Empty) (*api.Info, error) {
```
unsuman: Isn’t this redundant for the test case?

jandubois (dismissed):
Thanks, LGTM

pkg/portfwd/client.go
jandubois: Assuming tests are going to be green, I would rather merge this as-is to be in 2.1.2 than have it miss the boat because of a missing debug log. Just in case the PR doesn't get updated before you cut the release.

vasileknik76: Sorry for the delay. I've added the logs to CloseRead and CloseWrite

pkg/portfwd/client_test.go
vasileknik76: Thank you, fixed

jandubois (approved):
Thanks, LGTM

PR #5045, opened 2026-05-26
[DNM] Downgrade RootlessKit

AkihiroSuda:
For analyzing:
- #5030


AkihiroSuda:
The CI still fails with the same error as #5030, even after downgrading RootlessKit to v2.3.6 and installing slirp4netns: https://github.com/lima-vm/lima/actions/runs/26463606367/job/77917852538?pr=5045

This implies that RootlessKit v3.0 is innocent.

PR #5050, opened 2026-05-26
Add user to systemd-journal group so journalctl --user works

### LINKED ISSUE #5047: Add the user to `systemd-journal` group by default so that `journalctl --user` works
Currently `journalctl --user` does not work in the default setup:
```console
$ journalctl  --user
Hint: You are currently not seeing messages from the system.
      Users in groups 'adm', 'systemd-journal' can see all messages.
      Pass -q to turn off this notice.
No journal files were opened due to insufficient permissions.
```

Note: non-systemd distributions do not have `systemd-journal`

AkihiroSuda:
Adds a boot script that adds the default user to the "systemd-journal" group on systemd-based distributions, so that `journalctl --user` works out of the box instead of failing with "No journal files were opened due to insufficient permissions".

The "systemd-journal" group only exists on systemd-based distributions, so the script guards on its existence with getent and is a no-op elsewhere (e.g. Alpine).

Fixes https://github.com/lima-vm/lima/issues/5047

 - --
Note: used Claude

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/00-systemd-user-group.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/00-systemd-user-group.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+set -eu
+
+# If the UID is in 1...999 ("system users"), add the user to the "systemd-journal"
+# group so that `journalctl --user` works out of the box, without failing with
+# "No journal files were opened due to insufficient permissions".
+# https://github.com/lima-vm/lima/issues/5047
+# The "systemd-journal" group only exists on systemd-based distributions,
+# so this is a no-op elsewhere.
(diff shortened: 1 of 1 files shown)


unsuman (dismissed):
Thanks!

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/00-systemd-user-group.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Add the user to the "systemd-journal" group so that `journalctl --user` works
+# out of the box, without failing with "No journal files were opened due to
+# insufficient permissions".
+# https://github.com/lima-vm/lima/issues/5047
+#
+# The "systemd-journal" group only exists on systemd-based distributions,
+# so this is a no-op elsewhere.
+getent group systemd-journal >/dev/null 2>&1 || exit 0
+
```
afbjorklund, line 16: Could add a check if LIMA_CIDATA_UID is 1000 or more, then it should not be needed either

AkihiroSuda: done

jandubois (approved):
Thanks, LGTM

PR #5051, opened 2026-05-26
CI: remove `--memory 1` to avoid potential OOM

AkihiroSuda:


AkihiroSuda:
I thought this may fix #5030, but it doesn't 😞

unsuman:
> I thought this may fix #5030, but it doesn't 😞

I’ve tried this before but I failed

PR #5054, opened 2026-05-27
fix: reject parent directory template locators

immanuwell:
It fixes tiny edge case

`base: ..` slipped past the parent-dir check because it only looked for `../`. That lets `limactl template copy` absolutize it to the parent of the template dir.

Repro on main:
```bash
tmpdir=$(mktemp -d)
printf "base: ..\n" > "$tmpdir/template.yaml"
limactl template copy "$tmpdir/template.yaml" -
```

This prints a parent path instead of rejecting it. now it errors, same vibe as `../base.yaml`.

Tests:
```bash
go test ./...
```

### DIFF
--- a/pkg/limatmpl/abs.go
+++ b/pkg/limatmpl/abs.go
@@ -11,6 +11,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/lima-vm/lima/v2/pkg/localpathutil"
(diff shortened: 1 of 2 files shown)


pkg/limatmpl/abs.go
```diff
@@ -126,7 +127,7 @@ func absPath(locator, basePath string) (string, error) {
 			return "", errors.New("basePath is empty")
 		case basePath == "-":
 			return "", errors.New("can't use relative paths when reading template from STDIN")
-		case strings.Contains(locator, "../"):
+		case containsParentDir(locator):
 			return "", fmt.Errorf("relative locator path %q must not contain '../' segments", locator)
```
jandubois, line 131: Thanks, this is a good improvement. Please also update the error message to match:

```suggestion
			return "", fmt.Errorf("relative locator path %q must not contain '..' path segments", locator)
```

AkihiroSuda:
ping @immanuwell

PR #5055, opened 2026-05-27
pkg/downloader: logging nit

kolyshkin:
When commit 9ed9bbc1 changed fmt.Printf to logrus.Infof, it forgot to remove "\n" from the message. This is not shown with the default test log format, but it is visible in json output, e.g.:

> {"level":"info","msg":"Decompressing data\n","time":"2026-05-27T01:39:52-07:00"}

Remove "\n".

### DIFF
--- a/pkg/downloader/downloader.go
+++ b/pkg/downloader/downloader.go
@@ -558,7 +558,7 @@ func decompressLocal(ctx context.Context, decompressCmd, dst, src, ext, descript
 		if description == "" {
 			description = filepath.Base(src)
 		}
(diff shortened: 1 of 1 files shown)


kolyshkin:
PS I grepped for something like this and could not find anything; looks like this was the only place.

afbjorklund (approved):
Wonder why WSL2 changed that line, seems accidental to me

kolyshkin:
> Wonder why WSL2 changed that line, seems accidental to me

It is indeed accidental (that same commit contains a number of small unrelated fixups, formatting changes etc.).

PR #5056, opened 2026-05-27
vz: reconnect socket_vmnet on bridge failure

### LINKED ISSUE #3020: Lima instance does not recover from socket_vmnet restart
### Description

If we restart socket_vment limactl will fail to forward packets with this log:

```
{"level":"error","msg":"Failed to forward packets from VZ to VMNET: write unix -\u003e/var/run/socket_vmnet: write: broken pipe","time":"2024-12-13T21:29:30+02:00"}
```

From this point the lima0 interface will not receive or send anything since we stopped reading from the the VZ datagram socket and the unix socket connected to socket_vmnet. The only way to recover is to restart the vm.

We can recover from this by changing the forwarding code to reconnect to socket_vment after errors, and start forwarding again. Some packets will be lost, but networking code must be able to recover from lost packets.

This is not easy since we after we get an error on the socket_vmnet socket, we need to stop both io.Copy() goroutines without closing the VZ datagram socketpair. To do this we may need to implement interruptible Copy() instead of using io.Copy().

loncharles:
## Summary

- Add reconnect logic to VZ network forwarding for socket_vmnet connection failures
- On error, the socket_vmnet connection is re-established with exponential backoff (100ms–30s) while keeping the VZ DGRAM socketpair alive
- Replaces `io.Copy` with `copyPackets()` that distinguishes read vs write errors for fatal/recoverable classification

## Motivation

When the socket_vmnet daemon restarts or the UNIX socket connection breaks, `forwardPackets()` exits permanently and VM networking is dead until a full VM restart. The VZ-side socketpair (bound to `VZFileHandleNetworkDeviceAttachment` at VM creation) cannot be replaced, but the socket_vmnet connection is stateless and can be re-dialed.

Fixes #3020

## Performance

Tested on macOS 26 (Apple M1, VZ backend, 1GbE NIC). 3-run averages for each variant:

| | Baseline (2.1.1) | Patched (32K buf) | Patched (64K buf) |
|---|---|---|---|
| Throughput (VM → LAN host) | 944 Mbits/sec | 944 Mbits/sec | 944 Mbits/sec |
| Throughput (VM → macOS host, loopback) | 2.16 Gbits/sec | 2.18 Gbits/sec | 2.15 Gbits/sec |
| limactl CPU during iperf3 (VM → LAN) | ~40% | ~41% | ~42% |
| limactl CPU during iperf3 (loopback) | ~145% | ~145% | ~140% |
| Ping latency (100 pings) | 0.67ms | 0.86ms | 0.98ms |

No measurable throughput or CPU impact. Buffer size (32K matching io.Copy vs 64K) makes no difference on DGRAM sockets where each read returns one packet.

## Reconnect behavior

Tested by killing socket_vmnet (SIGKILL) and immediately restarting:

- Lima detects connection loss within one packet relay cycle
- Exponential backoff: 100ms → 200ms → 400ms → ... → 30s max
- socket_vmnet handles stale socket files on restart (no manual cleanup needed)
- With immediate daemon restart: **zero packet loss**, reconnect in <100ms
- With delayed restart: packets buffer in socketpair kernel buffer (4MB), full recovery when daemon returns
- Docker containers and k3s agent survive the outage

### Goroutine teardown tradeoff

Closing the socket_vmnet connection unblocks the goroutine doing socket_vmnet I/O immediately. The goroutine blocked on `vzConn.Read()` waits until the guest sends its next packet (ARP, NDP, DHCP — typically within seconds). `SetReadDeadline` was considered but rejected: arming/disarming deadlines on the DGRAM socket across reconnect cycles adds complexity and risks false-positive timeouts during normal forwarding.

## Test plan

- [x] Unit tests pass (`TestDialQemu`, `TestForwardPacketsReconnect`)
- [x] `TestForwardPacketsReconnect`: kills fake server, verifies reconnect to new server, packets resume
- [x] No regression in normal operation (ping, iperf3, docker, k3s)
- [x] Throughput benchmarked before/after (iperf3, 3 runs averaged, 3 variants)
- [x] Reconnect tested: SIGKILL socket_vmnet + immediate restart (2 runs, 0% packet loss)
- [x] Reconnect tested: SIGKILL socket_vmnet + delayed restart (backoff to 30s, full recovery)
- [x] Docker containers survive outage
- [x] k3s agent survives outage
- [ ] Soak testing on production host

### DIFF
--- a/pkg/driver/vz/network_darwin.go
+++ b/pkg/driver/vz/network_darwin.go
@@ -42,181 +42,268 @@ func PassFDToUnix(unixSock string) (*os.File, error) {
 	return client, nil
 }
 
-func dialQemuConn(ctx context.Context, unixSock string) (*qemuPacketConn, error) {
-	var dialer net.Dialer
-	conn, err := dialer.DialContext(ctx, "unix", unixSock)
-	if err != nil {
-		return nil, err
-	}
-	return &qemuPacketConn{Conn: conn}, nil
-}
-
 // DialQemu support connecting to QEMU supported network stack via unix socket.
 // Returns os.File, connected dgram connection to be used for vz.
 func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
-	qemuConn, err := dialQemuConn(ctx, unixSock)
+	var dialer net.Dialer
+	conn, err := dialer.DialContext(ctx, "unix", unixSock)
 	if err != nil {
 		return nil, err
 	}
 
 	server, client, err := createSockPair()
 	if err != nil {
+		conn.Close()
 		return nil, err
 	}
 	dgramConn, err := net.FileConn(server)
 	if err != nil {
+		conn.Close()
 		return nil, err
 	}
 	vzConn := &packetConn{Conn: dgramConn}
 
-	go forwardPackets(ctx, unixSock, qemuConn, vzConn)
+	fwdCtx, fwdCancel := context.WithCancel(ctx)
+	qemuConn := newQemuPacketConn(fwdCtx, unixSock, conn)
+
+	go forwardPackets(fwdCancel, qemuConn, vzConn)
 
 	return client, nil
 }
 
-// forwardPackets relays packets between vzConn and qemuConn, reconnecting the
-// qemuConn (socket_vmnet) side on failure. vzConn is bound to
-// Virtualization.framework at VM creation and cannot be replaced; only qemuConn
-// is re-dialed on reconnect.
-func forwardPackets(ctx context.Context, unixSock string, qemuConn *qemuPacketConn, vzConn *packetConn) {
-	defer vzConn.Close()
+// forwardPackets relays packets between vzConn and qemuConn. Reconnection of
+// the qemuConn side is handled internally by qemuPacketConn; this function
+// just runs io.Copy in both directions until a fatal error occurs.
+func forwardPackets(cancel context.CancelFunc, qemuConn *qemuPacketConn, vzConn *packetConn) {
+	done := make(chan struct{}, 2)
 
+	go func() {
+		io.Copy(qemuConn, vzConn)
+		done <- struct{}{}
+	}()
+
+	go func() {
+		io.Copy(vzConn, qemuConn)
+		done <- struct{}{}
+	}()
+
+	<-done
+	if !qemuConn.done() {
+		logrus.Error("VZ network socket error, packet forwarding stopped permanently")
+	}
+	cancel()
+	vzConn.Close()
+	qemuConn.Close()
+	<-done
+}
+
+// qemuPacketConn relays length-prefixed Ethernet frames over a unix stream
+// socket to socket_vmnet, reconnecting automatically on connection failure.
+type qemuPacketConn struct {
+	mu        sync.Mutex
+	cond      *sync.Cond
+	conn      net.Conn
+	gen       uint64
+	reconnect chan uint64
+	ctx       context.Context
+	sock      string
+}
+
+func newQemuPacketConn(ctx context.Context, sock string, conn net.Conn) *qemuPacketConn {
+	c := &qemuPacketConn{
+		conn:      conn,
+		reconnect: make(chan uint64, 1),
+		ctx:       ctx,
+		sock:      sock,
+	}
+	c.cond = sync.NewCond(&c.mu)
+	go c.reconnectLoop()
+	go func() {
+		<-ctx.Done()
+		c.cond.Broadcast()
+	}()
+	return c
+}
+
+// getConn returns the current connection and its generation. Blocks while a
+// reconnect is in progress. Returns ok=false on shutdown.
+func (c *qemuPacketConn) getConn() (conn net.Conn, gen uint64, ok bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for c.conn == nil {
+		if c.ctx.Err() != nil {
+			return nil, c.gen, false
+		}
+		c.cond.Wait()
+	}
+	return c.conn, c.gen, true
+}
+
+// waitReconnect signals the reconnect goroutine and blocks until a new
+// connection is available. Returns false on shutdown.
+func (c *qemuPacketConn) waitReconnect(gen uint64) bool {
+	select {
+	case c.reconnect <- gen:
+	default:
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for c.gen == gen {
+		if c.ctx.Err() != nil {
+			return false
+		}
+		c.cond.Wait()
+	}
+	return c.conn != nil
+}
+
+func (c *qemuPacketConn) reconnectLoop() {
 	const (
 		initialBackoff = 100 * time.Millisecond
 		maxBackoff     = 2 * time.Second
 	)
 
 	for {
-		fatal := forwardUntilError(qemuConn, vzConn)
-		if fatal {
-			logrus.Error("VZ network socket error, packet forwarding stopped permanently")
+		select {
+		case <-c.ctx.Done():
+			c.shutdown()
 			return
-		}
-
-		logrus.Warn("VMNET connection lost, reconnecting")
-
-		backoff := initialBackoff
-		for {
-			if ctx.Err() != nil {
-				return
-			}
-
-			select {
-			case <-ctx.Done():
-				return
-			case <-time.After(backoff):
-			}
-
-			var err error
-			qemuConn, err = dialQemuConn(ctx, unixSock)
-			if err != nil {
-				logrus.WithError(err).Debug("Failed to reconnect to VMNET")
-				backoff *= 2
-				if backoff > maxBackoff {
-					backoff = maxBackoff
-				}
+		case failedGen := <-c.reconnect:
+			c.mu.Lock()
+			if c.gen != failedGen {
+				c.mu.Unlock()
 				continue
 			}
+			old := c.conn
+			c.conn = nil
+			c.mu.Unlock()
 
-			logrus.Info("VMNET connection re-established, resuming packet forwarding")
-			break
-		}
-	}
-}
+			if old != nil {
+				old.Close()
+			}
 
-// forwardUntilError forwards packets between qemuConn and vzConn until an error
-// occurs. Returns true if the error is fatal (vzConn failure), false if
-// recoverable (qemuConn failure). Always closes qemuConn before returning.
-// The goroutine blocked on vzConn.Read may not unblock until the guest sends
-// its next packet (ARP, NDP, DHCP — typically within seconds).
-func forwardUntilError(qemuConn *qemuPacketConn, vzConn *packetConn) bool {
-	errCh := make(chan bool, 2)
+			logrus.Warn("VMNET connection lost, reconnecting")
 
-	var wg sync.WaitGroup
-	wg.Add(2)
+			var dialer net.Dialer
+			backoff := initialBackoff
+			for {
+				select {
+				case <-c.ctx.Done():
+					c.shutdown()
+					return
+				case <-time.After(backoff):
+				}
 
-	// VZ → VMNET: reads vzConn, writes qemuConn
-	go func() {
-		defer wg.Done()
-		readErr, _ := copyPackets(qemuConn, vzConn)
-		errCh <- readErr // readErr=true → vzConn read failed (fatal)
-	}()
+				conn, err := dialer.DialContext(c.ctx, "unix", c.sock)
+				if err != nil {
+					logrus.WithError(err).Debug("Failed to reconnect to VMNET")
+					backoff *= 2
+					if backoff > maxBackoff {
+						backoff = maxBackoff
+					}
+					continue
+				}
 
-	// VMNET → VZ: reads qemuConn, writes vzConn
-	go func() {
-		defer wg.Done()
-		readErr, _ := copyPackets(vzConn, qemuConn)
-		errCh <- !readErr // !readErr → vzConn write failed (fatal)
-	}()
+				if c.ctx.Err() != nil {
+					conn.Close()
+					c.shutdown()
+					return
+				}
 
-	first := <-errCh
-	qemuConn.Close()
-	wg.Wait()
-	second := <-errCh
+				c.mu.Lock()
+				c.conn = conn
+				c.gen++
+				c.cond.Broadcast()
+				c.mu.Unlock()
 
-	return first || second
-}
-
-// maxPacketSize is the maximum Ethernet frame (1514) plus 4-byte QEMU length
-// prefix. Used for relay buffers and as a sanity check on incoming frames.
-const maxPacketSize = 1518
-
-// copyPackets is like io.Copy but reports whether the error was a read error
-// (true) or write error (false), so the caller can distinguish fatal (vzConn)
-// from recoverable (qemuConn) failures.
-func copyPackets(dst io.Writer, src io.Reader) (readErr bool, err error) {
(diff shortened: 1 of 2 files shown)


lonlundgren:
@nirs would you mind reviewing this? It addresses the reconnect approach you described in #3020.

pkg/driver/vz/network_darwin.go
```diff
@@ -61,33 +69,127 @@ func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
...
+// During reconnect, packets from the guest buffer in the socketpair kernel buffer
+// (4MB). If the buffer fills, the guest gets ENOBUFS (packet drops), which is
+// normal for a brief network interruption and recovered by TCP retransmit / ARP
+// retry. Packets from the network side are lost while disconnected.
+func forwardPackets(ctx context.Context, unixSock string, qemuConn *qemuPacketConn, vzConn *packetConn) {
 	defer vzConn.Close()
 
+	const (
+		initialBackoff = 100 * time.Millisecond
+		maxBackoff     = 30 * time.Second
```
nirs: 2 seconds should be good enough maximum. We do a very cheap reconnect attempt and checking every 2 seconds is negligible.

pkg/driver/vz/network_darwin.go
```diff
@@ -61,33 +69,127 @@ func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
...
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(backoff):
+			}
+
+			var err error
+			qemuConn, err = dialQemuConn(ctx, unixSock)
+			if err != nil {
+				logrus.WithError(err).Warn("Failed to reconnect to VMNET")
```
nirs: This is not really a warning but expected condition when we reconnect. A debug message is enough. We already logged a warning when the connection was broken.

pkg/driver/vz/network_darwin.go
```diff
@@ -61,33 +69,127 @@ func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
...
+// (4MB). If the buffer fills, the guest gets ENOBUFS (packet drops), which is
+// normal for a brief network interruption and recovered by TCP retransmit / ARP
+// retry. Packets from the network side are lost while disconnected.
+func forwardPackets(ctx context.Context, unixSock string, qemuConn *qemuPacketConn, vzConn *packetConn) {
 	defer vzConn.Close()
 
+	const (
+		initialBackoff = 100 * time.Millisecond
+		maxBackoff     = 30 * time.Second
+		backoffFactor  = 2.0
```
nirs: We can use integer.

pkg/driver/vz/network_darwin.go
```diff
@@ -61,33 +69,127 @@ func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
...
+			case <-ctx.Done():
+				return
+			case <-time.After(backoff):
+			}
+
+			var err error
+			qemuConn, err = dialQemuConn(ctx, unixSock)
+			if err != nil {
+				logrus.WithError(err).Warn("Failed to reconnect to VMNET")
+				backoff = time.Duration(math.Min(float64(backoff)*backoffFactor, float64(maxBackoff)))
```
nirs: Why do we need floating point to compute the time?

pkg/driver/vz/network_darwin.go
```diff
@@ -61,33 +69,127 @@ func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
...
+			return
+		}
+
+		backoff := initialBackoff
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
```
nirs, line 176: I think this has the same effect:

```go
if ctx.Err() != nil {
    return
}
```

pkg/driver/vz/network_darwin.go
```diff
@@ -61,33 +69,127 @@ func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
...
+
+			logrus.Warnf("VMNET connection lost, reconnecting in %v", backoff)
```
nirs: Why we log here and not after getting an error from forwardUntilError()?

pkg/driver/vz/network_darwin.go
```diff
@@ -61,33 +69,127 @@ func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
...
+
+// copyPackets copies packets from src to dst, returning whether the error was a
+// read error (true) or write error (false). This is used instead of io.Copy
+// because io.Copy does not distinguish read errors from write errors, and the
+// reconnect logic needs that distinction: a qemuConn (socket_vmnet) error is
+// recoverable via reconnect, while a vzConn (VZ socketpair) error is fatal.
+// The copy loop is the same read-buf-write-buf loop io.Copy uses (neither
+// qemuPacketConn nor packetConn implements ReaderFrom/WriterTo).
+func copyPackets(dst io.Writer, src io.Reader) (readErr bool, err error) {
+	buf := make([]byte, 32*1024)
```
nirs: We write single packet (1514 bytes) to/from the datagram socket, so we don't need 32k buffer. We also write single size prefixed packets to socket_vmnet socket, so maximum size is 1518 bytes.

pkg/driver/vz/network_darwin.go
```diff
@@ -61,33 +69,127 @@ func DialQemu(ctx context.Context, unixSock string) (*os.File, error) {
...
+	for {
+		nr, er := src.Read(buf)
+		if nr > 0 {
+			if _, ew := dst.Write(buf[:nr]); ew != nil {
+				return false, ew
+			}
+		}
+		if er != nil {
+			return true, er
+		}
```
nirs, line 269: It will be more clear to handle errors first for read.

But It is not clear to me why we cannot use io.Copy, and why we need to handle read and write errors differently. The same copy loop is used in both directions, so read error may error reading from the datagram socket or error reading from the stream socket, and the same for the other direction.

I think we have 2 possible errors when socket_vment socket is closed:
- read from unix stream socket returns 0 bytes or err != nil
- write to unix stream socket return connection reset/connection refused

So error handling should not be in the copy loop but inside the unix connection.


We can implement the reconnect inside the unix connection: recoverable read or write errors will go into reconnect loop, completing the read/write then the connection is established. The copy loop from unix to datagram can remain the same.

nirs (commented):
Nice!

pkg/driver/vz/network_darwin.go
lonlundgren: Done — maxBackoff is now 2s.

pkg/driver/vz/network_darwin.go
lonlundgren: Done — removed the float, using integer multiplication with an if-cap.

pkg/driver/vz/network_darwin.go
lonlundgren: Done — `backoff *= 2` with `if backoff > maxBackoff`, removed `math` import.

pkg/driver/vz/network_darwin.go
lonlundgren: Done — using `ctx.Err() != nil`.

pkg/driver/vz/network_darwin.go
lonlundgren: Done — moved to right after `forwardUntilError` returns, logged once per outage.

pkg/driver/vz/network_darwin.go
lonlundgren: Done — downgraded to Debug.

pkg/driver/vz/network_darwin.go
lonlundgren: Done — added a `maxPacketSize = 1518` constant and sized the buffer to it. Also added a bounds check in `qemuPacketConn.Read` to return an error instead of panicking on an oversized length prefix.

pkg/driver/vz/network_darwin.go
lonlundgren: Addressed the first seven comments and pushed a revision.

On moving reconnect into `qemuPacketConn` — the question is fair: why differentiate read and write errors at all? If the connection handles its own reconnect transparently, `io.Copy` works as-is and `copyPackets` goes away.

The issue is that reconnect requires coordinating both goroutines. When the stream socket breaks, the Read goroutine and Write goroutine both hit errors at the same time. With a connection-internal approach, both call into the reconnect path concurrently — you need a mutex or generation gate so only one dials while the other waits. But even with that, both goroutines need to restart their forwarding loops after the swap (the Read goroutine may have partial framing state from `qemuPacketConn.Read`'s two-phase header+payload read). So the internal approach still needs to: signal both goroutines, wait for both to drain, swap the connection, restart both. That's the outer loop reimplemented inside the struct with more machinery, not less.

The current design keeps that lifecycle explicit: both goroutines exit on error, the single-threaded outer loop reconnects, fresh goroutines start with a clean connection. The read/write error distinction lets us avoid restarting when the failure is fatal (vzConn side).

We can reconsider the connection-internal route if you think the tradeoff is worth the added complexity.

nirs: Good point about synchronizing the copy goroutines, but we still need to tell the difference between recoverable errors and fatal errors, so the difference cannot be based on the direction. I think the unix connection should detect and return new recoverable error and the copy loop  can try to reconnect only if we get such errors.

Regarding synchronization, depending on getting traffic from guest is not great. It will be better to trigger the reconnect immediately on the first error without having to wait.

I think we will detect the error in 2 cases:

- read error from unix socket - should fail immediately when the socket is closed. At this point.
- write error to unix socket - should fail on the next write after the socket is closed. But when the socket is closed we may be blocked reading from the datagram socket.

If we handle connections inside the unix connection, Write() can block on a condition variable when we are in in a connection retry. Once connection is finished, the Write() can complete. Same for Read() - it can block on the same condition variable. 

So we can have a third goroutine doing the reconnect and signaling the other goroutines when the connection is established.

pkg/driver/vz/network_darwin.go
nirs: Using exact buffer size is little bit fragile is we did not calculate the size correctly, and we don't have the info in this layer. socket_vmnet get the maximum package size from vment but but does not expose this value. I think standard buffer size like 2k or 4k bytes will be better.

pkg/driver/vz/network_darwin.go
lonlundgren: Makes sense, I like this approach better than the initial suggestion. The condvar + third goroutine gives clean coordination without the racing issues. A few things I want to confirm before implementing:

1/ Fatal vs recoverable errors: With io.Copy, we lose visibility into which side produced the error. I'm thinking qemuPacketConn wraps its errors in a recoverable sentinel type, and anything else is treated as fatal (no reconnect). That way a vzConn failure doesn't trigger a spurious reconnect attempt.

2/ Reconnect goroutine lifecycle: Is this a long-lived goroutine that blocks waiting for a "reconnect needed" signal, or spawned on demand when the first error comes in? Either way it needs to respect context cancellation to unblock the condvar waiters on shutdown.

3/ Write atomicity: qemuPacketConn.Write does header + payload as two syscalls on the stream. The lock needs to be held across both so the connection isn't swapped between them. That means a reconnect blocks until any in-flight Write completes. Should be fine in practice (writes are fast, connection is local) but wanted to flag it.

4/ Per-packet locking: Read and Write will both need to acquire a lock on every call. This probably won't affect throughput at 1GbE but we can perf test on a box with a 5GbE NIC to measure the actual impact.

Will submit another update if we're aligned.

nirs: > Makes sense, I like this approach better than the initial suggestion. The condvar + third goroutine gives clean coordination without the racing issues. A few things I want to confirm before implementing:
> 
> 1/ Fatal vs recoverable errors: With io.Copy, we lose visibility into which side produced the error. I'm thinking qemuPacketConn wraps its errors in a recoverable sentinel type, and anything else is treated as fatal (no reconnect). That way a vzConn failure doesn't trigger a spurious reconnect attempt.

If the reconnect mechanism is inside the unix socket, or even better - in the underlying connection, the error remain internal to the reconnecting connection so io.Copy is not in the loop.

We can use this pipeline:

   [ datagram conn ] <-> 2x io.Copy <-> [qemu con] <-> [reconnecting con] <-> unix con

We create the qemuPacketConn with a real Con - we can create it with a reconnectingCon - current code will not be affected. But since we depend on reading the size from the socket and only then reading the payload, I think we will have to keep the reconnect mahanism inside the packetCon, so we can drop a package if we get a size but the closed right after we got the size.

> 2/ Reconnect goroutine lifecycle: Is this a long-lived goroutine that blocks waiting for a "reconnect needed" signal, or spawned on demand when the first error comes in? Either way it needs to respect context cancellation to unblock the condvar waiters on shutdown.

Both options should work, long living grouting will probably be simpler. Loop waiting for reconnect request, reconnecting and signaling waiters.

> 3/ Write atomicity: qemuPacketConn.Write does header + payload as two syscalls on the stream. The lock needs to be held across both so the connection isn't swapped between them. That means a reconnect blocks until any in-flight Write completes. Should be fine in practice (writes are fast, connection is local) but wanted to flag it.

I don't think you need to hold a lock during the read/write, only to check if you can read or write. If the socket is closed the read/write will fail. Also you cannot hold the lock during read - read is blocking and can block for long time.
 
> 4/ Per-packet locking: Read and Write will both need to acquire a lock on every call. This probably won't affect throughput at 1GbE but we can perf test on a box with a 5GbE NIC to measure the actual impact.

socket_vment is pretty slow, maximum throughput is 3-4 Gbits/s - I don't think a lock will make a difference, but it will be easy to test with iperf3.

lonlundgren:
Pushed a revision that moves the reconnect logic into qemuPacketConn per your suggestion. The outer loop and copyPackets are gone. forwardPackets is now just two io.Copy goroutines, and reconnection is transparent to callers.

Design:

- qemuPacketConn.Read/Write snapshot the current connection via getConn() (mutex acquire to read two fields), do I/O on the local copy, and on error call waitReconnect() which signals a dedicated reconnect goroutine and blocks until a new connection is ready.
- Generation counter prevents double-reconnect. Buffered(1) channel with non-blocking send deduplicates concurrent signals from Read and Write goroutines.
- A broadcast-on-context-cancel goroutine ensures condvar waiters always wake on shutdown.
- readPacket/writePacket extracted as standalone functions so the test echo server can use them without constructing a full qemuPacketConn.

Tested on macOS 26 (M1 Mac Mini, VZ, 1GbE NIC). 3-run averages, 18 CPU samples per metric:

| | Baseline (2.1.1) | Refactored |
|---|---|---|
| Throughput VM -> LAN | 944 Mbits/sec | 940 Mbits/sec |
| Throughput VM -> host (loopback) | 2.16 Gbits/sec | 2.17 Gbits/sec |
| limactl CPU (LAN) | ~40% | 41.4% (40.8-42.3) |
| limactl CPU (loopback) | ~145% | 139.9% (136.2-142.2) |

Reconnect tested by killing socket_vmnet (SIGKILL). Backoff retries at 2s cap, full recovery on daemon restart, no VM restart needed.

pkg/driver/vz/network_darwin.go
lonlundgren: @nirs, revision is in if you want to take a look.

lonlundgren:
ping. should we close this if it doesn't fit purpose?

AkihiroSuda:
How to test this PR?

> When the socket_vmnet daemon restarts or the UNIX socket connection breaks, 

How does this happen?

PR #5057, opened 2026-05-28
refactor: enable usestdlibvars linter

alexandear:
https://golangci-lint.run/docs/linters/configuration/#usestdlibvars

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,6 +29,7 @@ linters:
   - unconvert
   - unused
   - usetesting
+  - usestdlibvars
   - whitespace
   settings:
     depguard:
@@ -133,6 +134,10 @@ linters:
       - -ST1000
       - -ST1001  # duplicates revive.dot-imports
       - -ST1022
+    usestdlibvars:
+      http-method: false
+      time-date-month: true
+      time-layout: true
     usetesting:
       os-temp-dir: true
       context-background: true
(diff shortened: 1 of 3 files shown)


.golangci.yml
```diff
@@ -133,6 +134,10 @@ linters:
       - -ST1000
       - -ST1001  # duplicates revive.dot-imports
       - -ST1022
+    usestdlibvars:
+      http-method: false
+      time-date-month: true
+      time-layout: true
```
AkihiroSuda, line 140: Why these exceptions?

alexandear: Because they are turned off by default. 

```yaml
linters:
  settings:
    usestdlibvars:
      # Suggest the use of http.MethodXX.
      # Default: true
      http-method: false
      # Suggest the use of http.StatusXX.
      # Default: true
      http-status-code: false
      # Suggest the use of time.Month in time.Date.
      # Default: false
      time-date-month: true
      # Suggest the use of time.Weekday.String().
      # Default: true
      time-weekday: true
      # Suggest the use of time.Month.String().
      # Default: false
      time-month: true
      # Suggest the use of time.Layout.
      # Default: false
      time-layout: true
      # Suggest the use of crypto.Hash.String().
      # Default: false
      crypto-hash: true
      # Suggest the use of rpc.DefaultXXPath.
      # Default: false
      default-rpc-path: true
      # Suggest the use of sql.LevelXX.String().
      # Default: false
      sql-isolation-level: true
      # Suggest the use of tls.SignatureScheme.String().
      # Default: false
      tls-signature-scheme: true
      # Suggest the use of constant.Kind.String().
      # Default: false
      constant-kind: true
```

jandubois (approved):
Thanks, LGTM

PR #5065, opened 2026-06-01
fix: disable user containerd by default for non-Linux guests

### LINKED ISSUE #5037: FillDefault enables containerd.user for macOS guests on aarch64 — nerdctl unnecessarily downloaded

RoySerbi:
﻿Fixes #5037.

`containerd.user` was defaulted from the guest architecture only, so macOS guests on `x86_64` or `aarch64` enabled user containerd and pulled the nerdctl archive even though containerd/nerdctl are Linux guest functionality.

This keeps the existing default for Linux guests on `x86_64` and `aarch64`, but defaults `containerd.user` to `false` for non-Linux guests.

I also added a small table test for the defaulting behavior across Linux and Darwin guests.

Local check:

- `git diff --check`

I could not run `go test` locally because this Windows environment does not have the Go toolchain installed.

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -528,10 +528,9 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 		y.Containerd.User = o.Containerd.User
 	}
 	if y.Containerd.User == nil {
-		switch *y.Arch {
-		case limatype.X8664, limatype.AARCH64:
+		if *y.OS == limatype.LINUX && (*y.Arch == limatype.X8664 || *y.Arch == limatype.AARCH64) {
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Duplicate:
https://github.com/lima-vm/lima/pull/5041

Also missing DCO sign

RoySerbi:
Thanks @AkihiroSuda — closing as duplicate of #5041.

PR #5071, opened 2026-06-02
fix: append %r to ControlPath to prevent SSH session reuse across users

### LINKED ISSUE #5062: Generated ssh configuration ControlPath is too coarse
### Description

The ssh config files in `~/.lima/MACHINE/ssh.config` include:

```text
Host lima-MACHINE
  ControlMaster auto
  ControlPath "/Users/USERNAME/.lima/MACHINE/ssh.sock"
  ControlPersist yes
  # ...
```

If you want to be able to connect to the machine as more than one user, this configuration breaks it in a confusing way.

```text
user@host$ ssh firstuser@lima-MACHINE
firstuser@lima-MACHINE$ whoami
firstuser
firstuser@lima-MACHINE$ exit
user@host$ ssh seconduser@lima-MACHINE
firstuser@lima-MACHINE$ whoami
firstuser
```

As a workaround you can put a line like this in `~/.ssh/config` before the lima include:

```text
Host lima-MACHINE
  ControlPath ~/.lima/MACHINE/ssh.sock.%r

Include ~/.lima/*/ssh.config
```

But it would be nice if lima would configure this itself.

(You might prefer to use `~/.lima/MACHINE/ssh.sock.%C`, which includes a hash of the username/hostname/port. The hostname doesn't help here, but the port might be useful if the user ever wanted to run a second ssh server on a different port.)

toller892:
## Problem

When connecting to a Lima instance as different users via SSH, the second user silently reuses the first user's ControlMaster session. This is because `ControlPath` does not include the remote username (`%r`), so all users share the same socket file.

**Reproduction:**
```
ssh firstuser@lima-MACHINE   # connects, creates socket at ssh.sock
ssh seconduser@lima-MACHINE  # reuses firstuser's socket -> whoami returns "firstuser"
```

## Fix

Append `.%r` (SSH escape for remote username) to the `ControlPath` template in `SSHOpts()`. This causes SSH to create separate control sockets per user (e.g., `ssh.sock.firstuser`, `ssh.sock.seconduser`).

Also update `IsControlMasterExisting()` to use `filepath.Glob` instead of exact `os.Stat`, so it finds any socket variant (`ssh.sock.*`).

The `filenames.LongestSock` constant already accounts for a 16-char suffix on the socket name, which is sufficient for the `%r` expansion in typical usage.

**Changes:** 1 file, +4/-5 lines.

Fixes #5062

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -327,9 +327,8 @@ func removeOptsFromSSHArgs(sshArgs []string, removeOpts ...string) []string {
 
 // IsControlMasterExisting returns true if the control socket file exists.
 func IsControlMasterExisting(instDir string) bool {
-	controlSock := filepath.Join(instDir, filenames.SSHSock)
-	_, err := os.Stat(controlSock)
-	return err == nil
+	matches, err := filepath.Glob(filepath.Join(instDir, filenames.SSHSock) + "*")
+	return err == nil && len(matches) > 0
 }
 
 // SSHOpts adds the following options to CommonOptions: User, ControlMaster, ControlPath, ControlPersist.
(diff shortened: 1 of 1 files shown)


AkihiroSuda:
> ssh.sock.firstuser

This should be like `firstuser.ssh.sock`? Every socket is expected to have `*.sock` suffix

> The filenames.LongestSock constant already accounts for a 16-char suffix on the socket name, which is sufficient for the %r expansion in typical usage.

What should happen when it is not sufficient?
Probably it should use the hash of the user name?

unsuman:
ping @toller892

PR #5077, opened 2026-06-03
shell: add --sync-exclude flag to exclude paths from --sync

### LINKED ISSUE #4887: shell --sync: Allow excluding files/folders from rsync

itsme-ranger:
Fixes #4887

## Changes

Adds a `--sync-exclude` flag to `limactl shell --sync` that passes `--exclude` to all rsync operations (host→guest initial sync, the dry-run stats pass, and the guest→host sync-back).

```console
# Exclude node_modules and .git from the sync
limactl shell --sync . --sync-exclude=node_modules --sync-exclude=.git default

# Use rsync glob patterns
limactl shell --sync . --sync-exclude='*.o' --sync-exclude=dist myinstance
```

The flag uses rsync's standard `--exclude` syntax and can be specified multiple times. Using `--sync-exclude` without `--sync` returns an error.

## Implementation

- Added `--sync-exclude` as a `StringArray` flag on the shell command
- Built `excludeArgs` as `["--exclude", val, ...]` pairs
- Threaded into the `copytool.Options.AdditionalArgs` of the initial sync CopyTool (which is also reused for rsync-back and the "view changes" diff)
- Added `excludeArgs` parameter to `getRsyncStats` so the dry-run stats pass applies the same filters

## Testing

- `go build ./cmd/limactl/` passes
- `go test ./cmd/limactl/...` passes
- `golangci-lint run ./cmd/limactl/...` reports 0 issues

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -73,6 +73,7 @@ func newShellCommand() *cobra.Command {
 	shellCmd.Flags().Bool("preserve-env", false, "Propagate environment variables to the shell")
 	shellCmd.Flags().Bool("start", false, "Start the instance if it is not already running")
 	shellCmd.Flags().String("sync", "", "Copy a host directory to the guest and vice-versa upon exit")
+	shellCmd.Flags().StringArray("sync-exclude", nil, "Exclude a path from --sync (rsync --exclude syntax); can be passed multiple times")
 
 	return shellCmd
 }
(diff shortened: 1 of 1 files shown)


unsuman:
Seems duplicate of #4917

PR #5081, opened 2026-06-03
MountTag hashing using ":" instead of "\0" separator

ciderale:
Using \0 as a separator makes the hash hard to reproduce in environments that do not allow NUL bytes in strings. This affects, for example, NixOS configurations that generate filesystem mount definitions directly rather than using cloud-init (https://github.com/ciderale/nixos-lima/blob/general-update/modules/lima_mounts.nix#L12). While ":" is a valid file path character, I don't think that this change may cause a problem in practice for the hashed mount tag.

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -95,7 +95,7 @@ func MACAddress(uniqueID string) string {
 // MountTag generates a stable mount tag from location and mountPoint.
 // Both paths are hashed to handle the same location mounted to multiple mount points.
 func MountTag(location, mountPoint string) string {
-	sha := sha256.Sum256([]byte(location + "\x00" + mountPoint))
+	sha := sha256.Sum256([]byte(location + ":" + mountPoint))
 	return fmt.Sprintf("lima-%x", sha[0:8])
 }
 


pkg/limayaml/defaults.go
```diff
@@ -95,7 +95,7 @@ func MACAddress(uniqueID string) string {
 // MountTag generates a stable mount tag from location and mountPoint.
 // Both paths are hashed to handle the same location mounted to multiple mount points.
 func MountTag(location, mountPoint string) string {
-	sha := sha256.Sum256([]byte(location + "\x00" + mountPoint))
+	sha := sha256.Sum256([]byte(location + ":" + mountPoint))
```
AkihiroSuda, line 98: The hash is expected to be stable across versions, so I don't think we can accept this change.

Why does https://github.com/ciderale/nixos-lima need to compute the hash by itself?
If it is really necessary it may invoke another scripting language such as bash or Python

ciderale: Thank you for the quick reply. In NixOS, the aim is to declaratively define the system configuration using the nix language and its module system. My nixos-lima setup can essentially avoid configuration via cloud-init (except for writing the "lima-boot-done" files). Technically, there is a way to call external programs and read them back into the nix configuration, but that should generally be avoided and causes other issues in this MacOS/Linux setup. 

My impression was, that the mountTag is an opaque implementation detail as cloud-init gets the tag as configuration and mounts them accordingly. I was more worried, that my implementation now depends on this hashing -- I have a locally patched lima version. Is there a possibility to change that in some (not too far) future version? 

Another idea was: Would it make sense to allow to provide a mountTag in the configuration explicitly? The current hash computation would be used if no mountTag was specified. That would be a bit more of a change, but that way, there is no coupling to this hashing.

jandubois: I don't quite understand the issue: if you don't use the tag generated by Lima, then it shouldn't matter what kind of tag you calculate in Nix, as long as it is stable. Why do they have to match?

ciderale: Lima-vm is "registering" the tags that eventually are "mountable" in the guest -- the tag is kind of the interface.

The NixOS configuration only does the latter. With my lima-nixos configuration, I also generate the lima configuration yaml. However, since the mountTag is currently not configurable in the lima configuration, I have to match the hashing scheme. 

With the idea of making the mountTag configurable (in lima.yaml), my nix configuration could define both sides. In that case, there would not be a need to match the hashing. 

Making the mountTag (optionally) configurable would be backward compatible, I suppose. The hashing would be used if the configuration has no explicit mountTag. But it is slightly more complex than changing the separator and one still had to use some hashing to configure the mountTag. 

May I ask, in which scenario would a change of the hashing (as in this PR) have a noticeable impact? I currently don't see it, but I guess you have a concrete example where it matters.

jandubois: The whole point of switching from an index to a hash (c30cc31f) was to have a stable mapping across reboots, so `fstab` entries would remain correct early during the boot, even before cloud-init is being processed. Otherwise you would need 2 reboot cycles when the mounts changed.

Changing the hashing function would break existing instances when you upgrade Lima.

I still don't understand why Nix can't extract the tags from `user-data`, which should get written even if you don't use cloud-init. You should never need to recompute hash values in the guest; they should be opaque tags. The fact that they are hashes is an implementation detail.

ciderale: Thank you for explaining the problem scenario requiring the stable hash. That makes sense, and I now see that my setup also benefits from that stability in early boot.

About extracting tags from user-data: yes, that's possible and is essentially the cloud-init style approach. NixOS generally prefers defining mounts, users, and other system settings directly in its configuration language. This declarative model makes the desired system state explicit and helps avoid configuration drift.

Regarding breaking existing instances: would that actually be the case? With the modified hashing, all mountTags would change, so the startup procedure would likely treat existing mounts as obsolete and recreate them accordingly. This differs from the earlier index-based approach, where the meaning of a given index could change over time.

jandubois: > With the modified hashing, all mountTags would change, so the startup procedure would likely treat existing mounts as obsolete and recreate them accordingly.

Yes, that sounds plausible.

I still don't understand why you can't read the `user-data` file from the configuration script instead of recomputing the tag. It is conceptually wrong, even though it would be unlikely to change again.

If you have the location and the mount point (and I assume mount options), why don't you have the tag? Where do you get everything else from, but are missing the tag?

ciderale: In short, my nixos-lima has a unified configuration.nix that generates the lima.yaml configuration and also the NixOS guest configuration. This provides the declarative NixOS experience for both layers simultaneously.

The reason lies in the declarative configuration approach that NixOS uses to achieve reproducibility. A configuration defines all aspects of the system. The configuration is then evaluated and built, before applying to a running system. This can happen on the target machine, but does not have to. At that stage, the on-boot `user-data` is not available. That is why I have access to mountPoint&location, but not the mountTag.

Using the `user-data` information is technically possible, but in my view goes against this philosophy. It would require an imperative (non-trivial) startup script that reads the user-data and reconciles the mounts -- essentially what cloud-init does. This moves the filesystem mounting outside of the declarative configuration. While that may sometimes be unavoidable, it seems worth trying to avoid when possible.

Matching lima's hashing scheme is conceptually not ideal, I agree, but maybe pragmatic. Cleaner would be to allow for defining a `mountTag` in lima.yaml (in addition to location & mountPoint). Then, the opaque tag is the contract between hypervisor and guest os. My configuration.nix could then define mountTags independent of lima's hashing. These mountTags would likely still be derived using some sort of hashing. Since lima's hashing is unlikely to change, it may be pragmatic to match it. Configuring the mountTag explicitly would be cleaner, thought.

AkihiroSuda: > Cleaner would be to allow for defining a mountTag in lima.yaml (in addition to location & mountPoint). 

SGTM

AkihiroSuda: Let me close this PR, feel free to open a new one for defining the tag in the yaml

jandubois: Given that changing the hashing algorithm is not going to break existing instance (they will all remount because all the tags have changed, so there should be no risk of mounting the wrong device), isn't it better not to expose this implementation detail to the end user (well, template author)?

Because now a Nix user would have to manually specify a tag for each mount point.

So unless I misunderstand that changing the hash doesn't break anything, I would prefer the fix from the current PR.

AkihiroSuda: > So unless I misunderstand that changing the hash doesn't break anything

Was it confirmed that it does not leave old entries in fstab?

jandubois: > In short, my nixos-lima has a unified configuration.nix that generates the lima.yaml configuration and also the NixOS guest configuration.

Sorry, I totally misunderstood that this calculation happens again on the host and not inside the guest.

jandubois: > Was it confirmed that it does not leave old entries in fstab?

It was not confirmed yet, and I had no chance to try, but it sounded convincing:

* When the hash function changes, **all** tags change.
* On reboot none of the old tags exist anymore, so **all** the entries will be removed from `fstab`.
* During cloud-init all mounts will be re-created using the new tags.

This was a problem before because the tags were re-used and would point to the wrong location. The mounts were not removed during reboot because the tag, e.g. `mount1` still existed, it just now pointed somewhere else.

But with a changed hash function, none of the old tags should exist anymore, so `fstab` should be clean by the time `cloud-init` starts.

It is like deleting all mounts and adding them all back in a single boot cycle.

ciderale: @jandubois The description matches my understanding. Since there hasn't been any update, I just wanted to check in. I thought you wanted to verify it yourself, but I may have misunderstood.

jandubois (approved):
Thanks, LGTM

The mount tag is never persisted across a Lima version change in a way that could desync. The host and guest tags are always derived together from the same binary at the same start.

PR #4984, opened 2026-05-15
feat(autostart): add LaunchDaemon support for headless macOS servers

resker:
Closes #4983

## Summary

Adds `limactl autostart enable --condition=boot` to register a Lima instance
as a system-level LaunchDaemon. The VM starts at boot without requiring a user
login session, enabling headless macOS server deployments.

Also unifies the existing `start-at-login` functionality under the new
`limactl autostart` command and deprecates `limactl start-at-login`.

## Commands

```sh
# Start at user login (LaunchAgent on macOS, systemd user service on Linux)
limactl autostart enable k3s

# Start at system boot, before any user login (macOS only, prompts for sudo)
limactl autostart enable --condition=boot k3s

# Remove registration (auto-detects which type is installed)
limactl autostart disable k3s
```

## Design

**Privilege model**: `limactl` itself stays non-root (the existing root check
is not bypassed). The `--condition=boot` path execs `sudo` internally for
exactly two operations — writing to `/Library/LaunchDaemons/` and running
`launchctl bootstrap system` — and prompts for a password once at install time.

**`UserName` plist key**: launchd runs the daemon process as the specified user
(defaulting to `$USER`), so the Lima instance runs under the correct user's
home directory and config without requiring that user to be logged in.

**Runtime**: the daemon plist calls `limactl start <instance> --foreground`,
consistent with the existing LaunchAgent approach. Normal `limactl start/stop`
continues to work; no privileged runtime operations are needed.

## Changes

- `pkg/autostart/launchd/io.lima-vm.daemon.INSTANCE.plist` — daemon plist template
- `pkg/autostart/launchd/launchd.go` — `GetDaemonPlistPath`, `DaemonServiceNameFrom`, `DaemonTemplate`
- `pkg/autostart/managers.go` — `extraTemplateVars` field on `TemplateFileBasedManager`
- `pkg/autostart/managers_darwin.go` — `DaemonManager(userName string)` constructor
- `pkg/autostart/managers_linux.go`, `managers_others.go` — `DaemonManager` stubs
- `cmd/limactl/autostart.go` — `limactl autostart` command group (cross-platform)
- `cmd/limactl/autostart_darwin.go` — macOS: login → LaunchAgent, boot → LaunchDaemon
- `cmd/limactl/autostart_others.go` — non-macOS: login only, boot returns unsupported error
- `cmd/limactl/start-at-login.go` — marked deprecated
- `cmd/limactl/main.go` — register `autostart` subcommand
[...]

PR #5086, opened 2026-06-06
feat(autostart): add --keep-alive flag to limactl autostart enable

### LINKED ISSUE #4992: autostart: add KeepAlive to LaunchAgent and LaunchDaemon plists
## Problem

On macOS, both autostart plist templates (`io.lima-vm.autostart.INSTANCE.plist` for LaunchAgent and `io.lima-vm.daemon.INSTANCE.plist` for LaunchDaemon) use `RunAtLoad` but do not include `KeepAlive`. If the Lima host agent process exits unexpectedly (crash, signal, OOM), launchd does not restart it. For example, it's left in a `Broken` state until the next login (LaunchAgent) or reboot (LaunchDaemon), and `limactl start` fails with a stale `ha.sock` error:

```
FATA[0000] errors inspecting instance: [failed to get Info from "~/.lima/k3s/ha.sock":
  Get "http://lima-hostagent/v1/info": dial unix ~/.lima/k3s/ha.sock: connect: connection refused]
```

The LaunchDaemon case (`--condition=boot`) is especially relevant: the point of boot-time startup is generally unattended operation.

## Proposed fix

Add `<key>KeepAlive</key><true/>` to both plist templates:

```xml
<key>KeepAlive</key>
<true/>
```

launchd's built-in throttling prevents spin loops on repeated crashes, so this should be safe to enable unconditionally.

## Questions

- Is `KeepAlive: true` appropriate for the LaunchAgent case as well, or should it be limited to the LaunchDaemon?
- Should this be an option exposed via `limactl autostart enable`, or always on?

Related: #4984 (adds the LaunchDaemon / `--condition=boot` support)

resker:
Depends on #4984.

## Summary

- Adds `--keep-alive` (default: `true`) to `limactl autostart enable` for both LaunchAgent and LaunchDaemon plist templates
- When enabled, launchd restarts the Lima host agent automatically if it exits unexpectedly, preventing instances from remaining in a `Broken` state until the next login or reboot
- The flag is accepted on all platforms:
  - **macOS:** the `KeepAlive` stanza is rendered as a dict with `SuccessfulExit=false`, so launchd restarts the agent only on an unexpected (non-zero) exit — a clean `limactl stop` stays stopped.
  - **Linux:** the flag sets the systemd unit's `Restart=` directive (`on-failure` when enabled, `no` when disabled).

## Testing

- Manually tested on Apple Silicon Mac (macOS 26, arm64)
- Confirmed launchd restarts the agent after a simulated crash with `--keep-alive=true`
- Confirmed no restart with `--keep-alive=false`
- Template-render unit tests cover the keep-alive-enabled LaunchAgent and LaunchDaemon plists

Closes #4992

---

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license and I have signed off all commits per the DCO.

### DIFF
--- a/website/content/en/docs/usage/autostart.md
+++ b/website/content/en/docs/usage/autostart.md
@@ -58,8 +58,8 @@ limactl autostart enable --keep-alive=false default
 ```
 
 This applies to both `--condition=login` (macOS LaunchAgent) and
-`--condition=boot` (macOS LaunchDaemon). On Linux, restart behavior is
-configured separately in the systemd unit file.
+`--condition=boot` (macOS LaunchDaemon). On Linux, the systemd unit always
+uses `Restart=on-failure`, so keep-alive is always enabled regardless of the flag.
 
 ## Lima < 2.2
 
--- a/website/content/en/docs/usage/autostart.md
+++ b/website/content/en/docs/usage/autostart.md
@@ -58,8 +58,8 @@ limactl autostart enable --keep-alive=false default
 ```
 
 This applies to both `--condition=login` (macOS LaunchAgent) and
-`--condition=boot` (macOS LaunchDaemon). On Linux, the systemd unit always
-uses `Restart=on-failure`, so keep-alive is always enabled regardless of the flag.
+`--condition=boot` (macOS LaunchDaemon). On Linux, the flag sets the systemd unit's
+`Restart=` directive: `on-failure` when enabled (the default), or `no` when disabled.
 
 ## Lima < 2.2
 
--- a/website/content/en/docs/usage/autostart.md
+++ b/website/content/en/docs/usage/autostart.md
@@ -48,6 +48,19 @@ The `--user` flag specifies which macOS user the instance runs as (default:
 `$USER`). The plist is installed to
 `/Library/LaunchDaemons/io.lima-vm.daemon.<instance>.plist`.
 
+## Keep-alive behavior
+
+By default (`--keep-alive=true`), launchd will automatically restart the Lima
+host agent if it exits unexpectedly. To disable this:
+
+```bash
+limactl autostart enable --keep-alive=false default
+```
+
+This applies to both `--condition=login` (macOS LaunchAgent) and
+`--condition=boot` (macOS LaunchDaemon). On Linux, restart behavior is
+configured separately in the systemd unit file.
+
 ## Lima < 2.2
 
 Use `limactl start-at-login` (equivalent to `limactl autostart enable --condition=login`):
--- a/cmd/limactl/autostart.go
+++ b/cmd/limactl/autostart.go
@@ -34,6 +34,10 @@ func newAutostartEnableCommand() *cobra.Command {
 		"user", "",
 		"macOS username to run the instance as when --condition=boot (default: $USER)",
 	)
+	flags.Bool(
+		"keep-alive", true,
+		"Restart the instance automatically if the host agent exits unexpectedly",
+	)
 	return cmd
 }
 
--- a/cmd/limactl/autostart_darwin.go
+++ b/cmd/limactl/autostart_darwin.go
@@ -20,7 +20,12 @@ import (
 )
 
 func autostartEnableAction(cmd *cobra.Command, args []string) error {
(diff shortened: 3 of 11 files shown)


resker:
@AkihiroSuda - fyi, also submitted a follow-up PR #5086

AkihiroSuda:
Please rebase

.gitignore
```diff
@@ -4,3 +4,4 @@ lima.REJECTED.yaml
 default-template.yaml
 schema-limayaml.json
 .config
+limactl
```
AkihiroSuda: irrelevant change

resker:
@AkihiroSuda — rebased onto master.

website/content/en/docs/usage/autostart.md
```diff
@@ -48,6 +48,19 @@ The `--user` flag specifies which macOS user the instance runs as (default:
 `$USER`). The plist is installed to
 `/Library/LaunchDaemons/io.lima-vm.daemon.<instance>.plist`.
 
+## Keep-alive behavior
+
+By default (`--keep-alive=true`), launchd will automatically restart the Lima
```
AkihiroSuda, line 53: Have we tested restarting the host agent?
I don't think `limactl` was ever designed to support it.

website/content/en/docs/usage/autostart.md
```diff
@@ -48,6 +48,19 @@ The `--user` flag specifies which macOS user the instance runs as (default:
...
+host agent if it exits unexpectedly. To disable this:
+
+```bash
+limactl autostart enable --keep-alive=false default
+```
+
+This applies to both `--condition=login` (macOS LaunchAgent) and
+`--condition=boot` (macOS LaunchDaemon). On Linux, the systemd unit always
+uses `Restart=on-failure`, so keep-alive is always enabled regardless of the flag.
```
AkihiroSuda: Why not support disabling it on Linux?

resker:
@AkihiroSuda — addressed your remaining comments:

- **Restarting the host agent** (keep-alive): I tested this directly (details on #5088). launchd re-runs `limactl start <instance> --foreground` rather than restarting a separate host-agent process, so it goes through the normal start path — and I confirmed both post-reboot broken states recover cleanly on the next start.
- **Disabling on Linux**: good point — done. systemd already restarts via `Restart=on-failure` by default, so the flag started out as a no-op there (it was really built around the macOS opt-in, where launchd won't restart unless you add the `KeepAlive` key). Since that difference may be confusing, I wired the flag through on Linux so it behaves more uniformly — `--keep-alive=false` now sets `Restart=no`.
- Also dropped the `.gitignore` change.

AkihiroSuda:
Please click `Ready for review` if ready.

resker:
Done — marked ready. All three comments addressed (gitignore dropped, Linux keep-alive parity, host-agent restart validated per #5088).

resker:
Pushed `04125627`:

- **KeepAlive semantics:** changed the launchd stanza from a bare `<true/>` to a `KeepAlive` dict with `SuccessfulExit=false`, so launchd restarts the host agent only on an unexpected (non-zero) exit. The bare `<true/>` would have relaunched even a clean `limactl stop`. This now matches the docs ("exits unexpectedly") and Linux's `Restart=on-failure`.
- **Test coverage:** added template-render tests for the keep-alive-enabled LaunchAgent and LaunchDaemon plists.
- The "no-op on Linux" wording was only in the PR description; the docs already describe the `Restart=` behavior. Fixed the description.

AkihiroSuda:
@norio-nomura Could you take a look?

resker:
@AkihiroSuda Following up on your ping 2026-06-30 — it's been a while w/o response - how'd you suggest to proceed?

AkihiroSuda:
Let me take a look after releasing v2.2 next week

resker:
FYI - validated on macOS 27 beta.

resker:
Bumping this — it's been open a while and #5088 (self-healing LaunchDaemon restart) is stacked on top and waiting on it too. Happy to rebase/adjust either if that helps get eyes on them.

AkihiroSuda (approved):
Thanks

Issue #5075, opened 2026-06-02
limactl start fails with stale ha.sock/ha.pid from unclean shutdown


lonlundgren:
If the system shuts down before `limactl start --foreground` finishes cleanup (signal handler in `hostagent.go`), stale `ha.pid` and `ha.sock` files remain in the instance directory. On the next boot, `limactl start` tries to connect to the stale `ha.sock`, gets connection refused, and exits with a fatal error instead of cleaning up and starting fresh.

The hostagent command already has `defer os.RemoveAll(pidfile)` and signal handling via `signalCh`, but macOS may kill the process before cleanup completes during a fast shutdown. `limactl start` could check whether the PID in `ha.pid` is actually running before attempting to connect to `ha.sock`, and clean up stale files if it's not.

This is one of several issues that make Lima unusable for unattended auto-start via LaunchDaemon. See also #2252.

**Environment:** macOS 15.7.7, Lima 2.1.x (fork), LaunchDaemon with `RunAtLoad`.

resker:
Heads up — #5088 includes a fix for this. On detecting the stale `ha.sock` connection error, `limactl start` now clears leftover `.pid`/`.sock`/`.tmp` files from the instance directory and proceeds with a clean start, without signaling whatever unrelated process may have inherited the reused PID.

I've deliberately not marked it as closing this issue — I only reproduced it on macOS 26 / VZ with a `--condition=boot` LaunchDaemon, and your report may cover ground mine doesn't. @lonlundgren, would you mind confirming against your setup once it lands?

Issue #5074, opened 2026-06-02
limactl start does not verify socket_vmnet succeeded, blocks forever on failure


lonlundgren:
When Lima starts socket_vmnet via `startDaemon()` in `pkg/networks/reconcile/reconcile.go`, it calls `cmd.Start()` and proceeds without checking whether `vmnet_start_interface` succeeded. If socket_vmnet fails (e.g., `VMNET_FAILURE` from vmnet.framework), Lima starts QEMU anyway. With `--foreground`, `limactl start` then blocks indefinitely waiting for SSH to a VM that has no bridged networking.

socket_vmnet exits immediately on `vmnet_start_interface` failure and removes its PID file. Lima could detect this by checking the PID file or process liveness after a short delay.

Additionally, vmnet.framework depends on the `com.apple.NetworkSharing` XPC service. At early boot, this service's Mach port may not be registered yet, causing `vmnet_start_interface` to return `VMNET_FAILURE`. Neither socket_vmnet nor Lima waits for this dependency or retries on failure.

This is one of several issues that make Lima unusable for unattended auto-start via LaunchDaemon. See also #2252.

**Environment:** macOS 15.7.7, Lima 2.1.x (fork), QEMU 11.0.0, Intel x86_64, socket_vmnet bridged mode.

resker:
This should be fixed by #5091, merged 2026-07-21 — `startNetwork()` now waits for socket_vmnet to write its PID file, retrying with backoff on the transient early-exit case and failing fast with the daemon's stderr instead of hanging. The PR referenced this issue but didn't use a closing keyword, so it stayed open. @lonlundgren, does that match what you were hitting?

Issue #2252, opened 2024-03-17
Default instance with a launchagent created by `start-at-login` isn't available after startup


rpkoller:
### Description

Yesterday I've tried the new start at login feature that got in with `0.21.0`.  Before i've shut down my computer yesterday I've used: 

`$> limactl start-at-login default 
INFO[0000] The autostart file "/Users/rkoller/Library/LaunchAgents/io.lima-vm.autostart.default.plist" has been created or updated 

When I've started my computer today and I tried to directly start a project in DDEV a framework for local development environments in docker i got: 

```
$> ddev config
ERRO[0000] app.FindContainerByType(web) failed 
ERRO[0000] app.FindContainerByType(web) failed 
Could not connect to a Docker provider. Please start or install a Docker provider.
For install help go to: https://ddev.readthedocs.io/en/stable/users/install/docker-installation/
 ```

When I checked if the lima instance is running: 

```
$> limactl list
NAME       STATUS     SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
default    Running    127.0.0.1:60022    vz        aarch64    4       8GiB      100GiB    ~/.lima/default
```
My default instance was shown as running. I then did a `limactl stop` directly followed by a `limactl start` and now docker was available to ddev. And if go into the macos system settings the lima login item is listed: 

![Screenshot 2024-03-17 at 16 40 29](https://github.com/lima-vm/lima/assets/1665422/0464362b-8833-4a5c-99ee-0ef4b42eca18)

For the context i am on a M1pro running macOS Sonoma 14.4 and limactl version 0.21.0 (using the docker template with vz and virtiofs). if you need more informations please let me know.

**Comment by roman-kiselenko** [comment]
The `start-at-login` feature generates a `plist` file to autostart lima vm; it doesn't propagate variables or set up any docker-related things. ddev uses colima and colima uses lima under the hood.

I guess is something going wrong with the VM itself at startup.

You can try to debug ddev docker with command: 

```sh
# Output contextual details for the Docker provider
ddev debug dockercheck 
```

roman-kiselenko:
There are logs files under `~/.lima/default/launchd.stderr.log` and ` ~/.lima/default/launchd.stdout.log` could you provide those logs?

rpkoller:
it wasn't a ddev related problem, as i've shown in the second example in the issue summary i've also tried to directly run a docker command and no docker provider was available there neither: 
```
$> docker ps
[...]

Issue #84, opened 2021-06-26
Ubuntu and Debian intermittently fail  to boot: `Kernel panic - not syncing: IO-APIC + timer doesn't work`


AkihiroSuda:
```
==> /Users/runner/.lima/debian/serial.log <==
57
[    0.056000] Call Trace:
58
[    0.056000]  dump_stack+0x6b/0x83
59
+ exit 1
60
[    0.056000]  panic+0x101/0x2d7
61
+ limactl delete -f debian
62
[    0.056000]  setup_IO_APIC+0x80e/0x855
63
[    0.056000]  ? _raw_spin_unlock_irqrestore+0x16/0x20
64
[    0.056000]  ? clear_IO_APIC_pin+0x173/0x240
65
[    0.056000]  x86_late_time_init+0x20/0x30
66
[    0.056000]  start_kernel+0x4eb/0x5a8
67
[    0.056000]  secondary_startup_64_no_verify+0xb0/0xbb
68
[    0.056000] ---[ end Kernel panic - not syncing: IO-APIC + timer doesn't work!  Boot with apic=debug and send a report.  Then try booting with the 'noapic' option. ]---
```

https://github.com/AkihiroSuda/lima/blob/eec56822fe197c15a9ac40a0ca736c3a1f5f833e/examples/debian.yaml

AkihiroSuda:
The default Ubuntu instance hits the same issue intermittently

```
==> /Users/runner/.lima/default/serial.log <==
[    0.048000]  check_timer+0x4a9/0x518
[    0.048000]  ? radix_tree_lookup+0xd/0x10
+ exit 1
[    0.048000]  setup_IO_APIC+0x111/0x120
+ limactl delete -f default
[    0.048000]  apic_intr_mode_init+0x107/0x10a
[    0.048000]  x86_late_time_init+0x24/0x35
[    0.048000]  start_kernel+0x3eb/0x4a7
[    0.048000]  x86_64_start_reservations+0x24/0x26
[    0.048000]  x86_64_start_kernel+0x8b/0x8f
[    0.048000]  secondary_startup_64_no_verify+0xc2/0xcb
[    0.048000] ---[ end Kernel panic - not syncing: IO-APIC + timer doesn't work!  Boot with apic=debug and send a report.  Then try booting with the 'noapic' option. ]---
```

AkihiroSuda:
https://github.com/lima-vm/lima/pull/294 seems working as workaround, so closing this

[...]

PR #5088, opened 2026-06-06
fix(vz): self-healing LaunchDaemon restart after unclean shutdown

### LINKED ISSUE #5087: bug(vz/launchdaemon): VZ driver leaves VM orphaned on shutdown; limactl start fails to recover broken state
## Summary

Two related bugs surface when running a Lima instance as a LaunchDaemon (PR #4984) on macOS. Together they cause the instance to get stuck in a `Broken` state after every system reboot, requiring manual intervention.

## Bug 1: VZ driver does not shut down cleanly on SIGTERM

When launchd sends SIGTERM to `limactl start <instance> --foreground` at system shutdown, Lima attempts to stop the VZ VM via `LimaVzDriver.Stop()`. On affected machines, `l.machine.CanRequestStop()` returns `false`, so `Stop()` returns `errors.New("vz: CanRequestStop is not supported")` and exits fatally without stopping the VM.

The VZ driver process (started by the system-level LaunchDaemon) is left running after the host agent exits. On next boot the instance is in state `"vz driver is running but host agent is not"`.

**Relevant code**: `pkg/driver/vz/vz_driver_darwin.go:527`

```
{"level":"fatal","msg":"vz: CanRequestStop is not supported"}
```

**Observed on**: Apple M4 Mac Mini, macOS 26.5, Lima 2.1.1 + dev build of #4984.

`CanRequestStop` returns `false` when the VM was not configured with `RequestStopHandler` set — this appears to be the case for all VZ instances created by Lima today.

**Fix direction**: set a `RequestStopHandler` on the VZ machine during creation so `CanRequestStop()` returns `true`. Alternatively, when `CanRequestStop` is false, fall back to `requestStopViaSSH` directly (graceful) or force-kill the VZ process (hard stop).

---

## Bug 2: `limactl start` does not recover from broken state

When the instance is in `StatusBroken` with error `"vz driver is running but host agent is not"`, `startAction` in `cmd/limactl/start.go` returns immediately with:

```
errors inspecting instance: [vz driver is running but host agent is not]
```

With `KeepAlive=true` in the LaunchDaemon plist, launchd retries every 10 seconds — but each retry hits the same broken-state check and exits, looping indefinitely. The instance never recovers without manual `limactl stop --force`.

**Relevant code**: `cmd/limactl/start.go:560,580`

**Fix direction**: in `startAction`, when the instance is `StatusBroken` with the specific error `"<vmtype> driver is running but host agent is not"`, automatically attempt `limactl stop --force` to clean up the orphaned driver state before proceeding with the start. This turns a permanent crash-loop into a self-healing recovery.

---

## Reproduction

1. Install Lima instance as a LaunchDaemon via `limactl autostart enable --condition=boot <instance>` (PR #4984)
2. Reboot the system
3. Observe: `limactl list` shows `STATUS=Broken`, `launchd.stderr.log` shows crash-loop

## Workaround

```sh
limactl stop --force <instance>
# launchd retries start automatically within KeepAlive interval
```

## Relationship

Bug 1 causes Bug 2. Fixing Bug 1 (clean VZ shutdown on SIGTERM) prevents the broken state from occurring. Fixing Bug 2 (self-healing start) provides defense-in-depth for any case where the VZ driver is orphaned (crash, OOM kill, etc.).

resker:
## Problem

Filed as #5087. A `KeepAlive`-enabled LaunchDaemon crash-loops after an unclean shutdown, because the host agent can exit before the VM is stopped and `limactl start` then refuses to start from the state that leaves behind.

## Changes

One commit per topic:

**`fix(vz)` — shut down via SSH when `CanRequestStop` is unsupported**
On some macOS versions `VirtualMachine.CanRequestStop()` returns false when no `RequestStopHandler` was registered, and the host agent exited without the guest ever being asked to stop. Falls back to an SSH shutdown, and in both branches waits out the existing stop poll rather than returning early if that request fails — so a VM that is still running yields the poll's timeout error, which is the unsuccessful exit `KeepAlive{SuccessfulExit=false}` restarts on.

**`fix(store)` — name the recoverable broken states, and only when provable**
Sentinel errors (`ErrDriverRunningHostAgentStopped`, `ErrHostAgentUnreachable`) so callers match with `errors.Is` rather than on message strings. The second is reported only on `ECONNREFUSED`, the one outcome proving nothing is listening on `ha.sock`; a timeout, a protocol error, or a socket not yet bound can all come from a live host agent and must not be auto-recovered, or the caller would delete a live agent's files and start a second one. (On Windows `WSAECONNREFUSED` is not mapped to `ECONNREFUSED`, so the sentinel never fires there — reported but not recovered, which is the safe direction.)

**`fix(start)` — recover from the broken states left by an unclean shutdown**
An orphaned driver is force-stopped; an unreachable host agent has its files removed without signaling any process, since the recorded PID may since belong to an unrelated one. Only `ha.pid`/`ha.sock` are removed — a driver running as its own process may still be alive, and hiding it from the next inspection would allow a second driver for the same instance; a driver that runs in the host agent process (vz) records the same PID and is removed with them. The two run in sequence, not as alternatives, because clearing a departed host agent can reveal a driver that still needs stopping — and doing that after `ha.pid` is gone means the force stop cannot signal a reused PID.

**`docs(autostart)`** — documents both states, when they occur, and the SSH fallback.

## Relationship to e529ce08

Since this was opened, e529ce08 made `ReadPIDFile` ignore PID files written during a previous boot, which covers the post-reboot case directly. What remains for the recovery path is PID reuse within a single boot, plus the cases that marker does not cover: platforms without a boot session identifier, and instances started by an older Lima that have no marker.

## Testing

Unit tests for both halves: the socket cases drive the real host agent client against a unix socket rather than synthetic errors, and the cleanup cases cover in-process vs. separate driver.

Validated on Apple Silicon (macOS 26, arm64) with a `--condition=boot` LaunchDaemon running k3s — hard reboot auto-starts cleanly, both broken states recover, k3s and its workloads healthy afterwards. @CFBancroft independently validated the orphaned-VZ recovery on a 2018 Intel Mac mini, and found that a normal Intel shutdown reports `CanRequestStop=true`, so that hardware does not reach the SSH fallback.

Closes #5087

Also addresses the problem reported in #5075 — leaving that open for @lonlundgren to confirm against their setup rather than closing it from here.

Assisted-by: Claude Code (Opus 5)

### DIFF
--- a/cmd/limactl/autostart_darwin.go
+++ b/cmd/limactl/autostart_darwin.go
@@ -20,7 +20,12 @@ import (
 )
 
 func autostartEnableAction(cmd *cobra.Command, args []string) error {
-	condition, err := cmd.Flags().GetString("condition")
+	flags := cmd.Flags()
+	condition, err := flags.GetString("condition")
+	if err != nil {
+		return err
+	}
+	keepAlive, err := flags.GetBool("keep-alive")
 	if err != nil {
 		return err
 	}
@@ -36,12 +41,13 @@ func autostartEnableAction(cmd *cobra.Command, args []string) error {
 
 	switch condition {
 	case "login":
-		if err := autostart.RegisterToStartAtLogin(ctx, inst); err != nil {
+		mgr := autostart.ManagerWith(keepAlive)
+		if err := mgr.RegisterToStartAtLogin(ctx, inst); err != nil {
 			return fmt.Errorf("failed to register instance %#q to start at login: %w", inst.Name, err)
 		}
 		logrus.Infof("Instance %#q registered to start at login", inst.Name)
 	case "boot":
-		userName, err := cmd.Flags().GetString("user")
+		userName, err := flags.GetString("user")
 		if err != nil {
 			return err
 		}
@@ -51,7 +57,7 @@ func autostartEnableAction(cmd *cobra.Command, args []string) error {
 		if userName == "" {
 			return errors.New("could not determine user; pass --user")
 		}
-		return daemonInstall(ctx, inst.Name, inst.Dir, userName)
+		return daemonInstall(ctx, inst.Name, inst.Dir, userName, keepAlive)
 	default:
 		return fmt.Errorf("unknown condition %q: must be \"login\" or \"boot\"", condition)
 	}
@@ -89,18 +95,22 @@ func autostartDisableAction(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func daemonInstall(ctx context.Context, instName, workDir, userName string) error {
+func daemonInstall(ctx context.Context, instName, workDir, userName string, keepAlive bool) error {
 	selfExe, err := os.Executable()
 	if err != nil {
 		return fmt.Errorf("could not determine limactl path: %w", err)
 	}
 
-	content, err := textutil.ExecuteTemplate(launchd.DaemonTemplate, map[string]string{
+	vars := map[string]string{
 		"Binary":   selfExe,
 		"Instance": instName,
 		"WorkDir":  workDir,
 		"UserName": userName,
-	})
+	}
+	if keepAlive {
+		vars["KeepAlive"] = "true"
+	}
+	content, err := textutil.ExecuteTemplate(launchd.DaemonTemplate, vars)
 	if err != nil {
 		return fmt.Errorf("failed to render daemon plist: %w", err)
 	}
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -578,7 +578,22 @@ func startAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 	if len(inst.Errors) > 0 {
-		return fmt.Errorf("errors inspecting instance: %+v", inst.Errors)
+		if isOrphanedDriverError(inst.Errors) {
+			// The VM driver process is running but the host agent is not — this happens when
+			// the host agent exits uncleanly (e.g. launchd SIGTERM at shutdown). Force-stop
+			// to clean up the orphaned driver so we can start fresh.
+			logrus.Warnf("Instance %q has an orphaned driver process; attempting force stop to recover", inst.Name)
+			instance.StopForcibly(inst)
+			inst, err = store.Inspect(cmd.Context(), inst.Name)
+			if err != nil {
+				return err
+			}
+			if len(inst.Errors) > 0 {
+				return fmt.Errorf("errors inspecting instance after force stop: %+v", inst.Errors)
+			}
+		} else {
+			return fmt.Errorf("errors inspecting instance: %+v", inst.Errors)
+		}
 	}
 	switch inst.Status {
 	case limatype.StatusRunning:
@@ -624,6 +639,18 @@ func startAction(cmd *cobra.Command, args []string) error {
 	}
 
 	return instance.Start(ctx, inst, launchHostAgentForeground, progress)
+}
+
+// isOrphanedDriverError returns true when the instance errors indicate that the VM driver
+// process is running but the host agent is not — the recoverable broken state that occurs
+// after an unclean shutdown (e.g. launchd SIGTERM before the host agent could stop the VM).
+func isOrphanedDriverError(errs []error) bool {
+	for _, err := range errs {
+		if strings.Contains(err.Error(), "driver is running but host agent is not") {
+			return true
+		}
+	}
+	return false
 }
 
 func createBashComplete(cmd *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -591,6 +591,19 @@ func startAction(cmd *cobra.Command, args []string) error {
 			if len(inst.Errors) > 0 {
 				return fmt.Errorf("errors inspecting instance after force stop: %+v", inst.Errors)
 			}
+		} else if isStaleHostAgentError(inst.Errors) {
+			// The host agent PID file exists but the socket is unreachable — the PID was
+			// likely reused by another process after a reboot. Remove stale temp files
+			// without sending any signal (the PID may belong to an unrelated process).
+			logrus.Warnf("Instance %q has a stale host agent socket; removing stale files to recover", inst.Name)
+			removeStaleTempFiles(inst.Dir)
+			inst, err = store.Inspect(cmd.Context(), inst.Name)
+			if err != nil {
+				return err
+			}
+			if len(inst.Errors) > 0 {
+				return fmt.Errorf("errors inspecting instance after removing stale files: %+v", inst.Errors)
+			}
 		} else {
 			return fmt.Errorf("errors inspecting instance: %+v", inst.Errors)
 		}
@@ -651,6 +664,39 @@ func isOrphanedDriverError(errs []error) bool {
 		}
 	}
 	return false
+}
+
+// isStaleHostAgentError returns true when the instance errors indicate that the host agent
+// socket is unreachable despite a PID file existing — the recoverable broken state that
+// occurs when the PID is reused by another process after a reboot.
+func isStaleHostAgentError(errs []error) bool {
+	for _, err := range errs {
+		if strings.Contains(err.Error(), filenames.HostAgentSock) {
+			return true
+		}
+	}
+	return false
+}
+
+// removeStaleTempFiles removes all .pid, .sock, and .tmp files from the instance directory.
+// Used to recover from a stale host agent state without sending signals to any process.
+func removeStaleTempFiles(instDir string) {
+	entries, err := os.ReadDir(instDir)
+	if err != nil {
+		logrus.WithError(err).Warnf("Failed to read instance dir %q", instDir)
+		return
+	}
+	for _, entry := range entries {
+		path := filepath.Join(instDir, entry.Name())
+		for _, suffix := range filenames.TmpFileSuffixes {
+			if strings.HasSuffix(path, suffix) {
+				logrus.Infof("Removing stale file %q", path)
+				if err := os.Remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
+					logrus.WithError(err).Warnf("Failed to remove %q", path)
+				}
+			}
+		}
+	}
 }
 
 func createBashComplete(cmd *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -659,7 +659,7 @@ func startAction(cmd *cobra.Command, args []string) error {
 // after an unclean shutdown (e.g. launchd SIGTERM before the host agent could stop the VM).
 func isOrphanedDriverError(errs []error) bool {
 	for _, err := range errs {
-		if strings.Contains(err.Error(), "driver is running but host agent is not") {
+		if errors.Is(err, store.ErrDriverRunningHostAgentStopped) {
 			return true
 		}
 	}
@@ -671,7 +671,7 @@ func isOrphanedDriverError(errs []error) bool {
 // occurs when the PID is reused by another process after a reboot.
 func isStaleHostAgentError(errs []error) bool {
 	for _, err := range errs {
-		if strings.Contains(err.Error(), filenames.HostAgentSock) {
+		if errors.Is(err, store.ErrHostAgentUnreachable) {
 			return true
 		}
 	}
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -593,6 +593,31 @@ func startAction(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
+	// Recover from the broken states that an unclean shutdown can leave behind. The two are
+	// handled in sequence rather than as alternatives: clearing a departed host agent can
+	// reveal a VM driver that is still running, which then has to be stopped on its own.
+	if isStaleHostAgentError(inst.Errors) {
+		// Nothing is listening on the host agent socket, so the host agent is gone. Remove
+		// the files it left behind without signaling any process: its PID may since have
+		// been reused by an unrelated one.
+		logrus.Warnf("Instance %q has a stale host agent socket; removing stale host agent files to recover", inst.Name)
+		removeStaleHostAgentFiles(inst)
+		inst, err = store.Inspect(cmd.Context(), inst.Name)
+		if err != nil {
+			return err
+		}
+	}
+	if isOrphanedDriverError(inst.Errors) {
+		// The VM driver process is running but the host agent is not — this happens when
+		// the host agent exits uncleanly (e.g. launchd SIGTERM at shutdown). Force-stop
+		// to clean up the orphaned driver so we can start fresh.
+		logrus.Warnf("Instance %q has an orphaned driver process; attempting force stop to recover", inst.Name)
+		instance.StopForcibly(inst)
+		inst, err = store.Inspect(cmd.Context(), inst.Name)
+		if err != nil {
+			return err
+		}
+	}
 	if len(inst.Errors) > 0 {
 		return fmt.Errorf("errors inspecting instance: %+v", inst.Errors)
 	}
@@ -648,6 +673,53 @@ func startAction(cmd *cobra.Command, args []string) error {
 	}
 
 	return instance.Start(ctx, inst, launchHostAgentForeground, progress)
+}
+
+// isOrphanedDriverError returns true when the instance errors indicate that the VM driver
+// process is running but the host agent is not — the recoverable broken state that occurs
+// after an unclean shutdown (e.g. launchd SIGTERM before the host agent could stop the VM).
+func isOrphanedDriverError(errs []error) bool {
+	for _, err := range errs {
+		if errors.Is(err, store.ErrDriverRunningHostAgentStopped) {
+			return true
+		}
+	}
+	return false
+}
+
+// isStaleHostAgentError returns true when the instance errors indicate that the host agent
+// socket is unreachable despite a PID file existing — the recoverable broken state that
+// occurs when the PID is reused by another process after a reboot.
+func isStaleHostAgentError(errs []error) bool {
+	for _, err := range errs {
+		if errors.Is(err, store.ErrHostAgentUnreachable) {
+			return true
+		}
+	}
+	return false
+}
+
+// removeStaleHostAgentFiles removes the PID file and socket of a host agent that is known
+// to be gone, without sending a signal to any process. Only files belonging to the host
+// agent are removed: a VM driver that runs as its own process may still be alive, and
+// deleting its PID file would hide it from the next inspection and allow a second driver
+// to be started for the same instance. The exception is a driver that runs the VM in the
+// host agent process (such as vz), which records the very same PID; that PID file refers
+// to the departed host agent too, so it is removed as well.
+func removeStaleHostAgentFiles(inst *limatype.Instance) {
+	paths := []string{
+		filepath.Join(inst.Dir, filenames.HostAgentPID),
+		filepath.Join(inst.Dir, filenames.HostAgentSock),
+	}
+	if inst.DriverPID != 0 && inst.DriverPID == inst.HostAgentPID {
+		paths = append(paths, filepath.Join(inst.Dir, filenames.PIDFile(inst.VMType)))
+	}
+	for _, path := range paths {
+		logrus.Infof("Removing stale file %q", path)
+		if err := os.Remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
+			logrus.WithError(err).Warnf("Failed to remove %q", path)
+		}
+	}
 }
 
 func createBashComplete(cmd *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
--- a/cmd/limactl/autostart.go
+++ b/cmd/limactl/autostart.go
@@ -34,6 +34,10 @@ func newAutostartEnableCommand() *cobra.Command {
 		"user", "",
 		"macOS username to run the instance as when --condition=boot (default: $USER)",
 	)
+	flags.Bool(
+		"keep-alive", true,
+		"Restart the instance automatically if the host agent exits unexpectedly",
+	)
 	return cmd
 }
 
--- a/cmd/limactl/autostart_others.go
+++ b/cmd/limactl/autostart_others.go
@@ -35,7 +35,12 @@ func autostartEnableAction(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
-	if err := autostart.RegisterToStartAtLogin(ctx, inst); err != nil {
+	keepAlive, err := cmd.Flags().GetBool("keep-alive")
+	if err != nil {
+		return err
+	}
+	mgr := autostart.ManagerWith(keepAlive)
+	if err := mgr.RegisterToStartAtLogin(ctx, inst); err != nil {
 		return fmt.Errorf("failed to register instance %#q to start at login: %w", inst.Name, err)
 	}
 	logrus.Infof("Instance %#q registered to start at login", inst.Name)
--- a/cmd/limactl/start_test.go
+++ b/cmd/limactl/start_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/spf13/cobra"
 	"gotest.tools/v3/assert"
 
+	"github.com/lima-vm/lima/v2/pkg/limatype"
 	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
 )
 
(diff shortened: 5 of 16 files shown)


lonlundgren:
Nice job.

This needs additional testing on Intel. HVF has reliability issues that don't show up on VZ that are especially exacerbated by launching a Lima VM closer to boot time.

See: #84 , #2252 , #5074 

We have custom launchdaemon scripts running across stably Intel and Apple Silicon that encompass these fixes and the #5075 fix you included.

resker:
Thanks @loncharles — noted wrt Intel systems... alas I don't have any accessible to validate w/.  I'd had a look at some of your comments and wonder if you'd considered vzNAT as alternative to socket_vmnet - might sidestep the XPC timing issue and hew closer to what macOS is layering in.

lonlundgren:
We absolutely need bridged network in our setup and it is the reason why we chose Lima instead of an alternative and so that we'd have one deployment surface area over Intel and Apple Silicon.

The XPC timing issue is less of a problem in practice, but socket_vmnet has several other repeatable ways it fails when you try to start a VM at boot. And much more reproducible on a machine that has a two-step boot when FileVault is enabled, than one that doesn't. So net net, the better ordering when wanting to boot a VM with a luanchscript is to have the script own start/retry on socket_vmnet creation and then only start Lima after.

The APIC timer kernel panic is independent of networking choice and is highly repeatable on Intel. Many people will hit this. And Lima made a change in their CI to fix it but never shipped any fixes in any of their VM template files.

resker:
fyi, just tested this with socket_vmnet bridged networking on Apple Silicon (most of my use is k3s w/ NAT) — both the LaunchDaemon boot and keep-alive restart work w/o any observed issues.

lonlundgren:
Good stuff. I should have been clear that these are primarily Intel issues and both are intermittent, but real and regular.

This should get some Intel coverage, if not include the other fixes, before merge.

resker:
FYI, #5091 should improve the silent hang described

resker:
loncharles — since you've Intel + FileVault setups running, could you give this a spin? I don't have Intel hw accessible.

EDIT: tagged the wrong person on these originally... removing the "@" to spare them further noise.

resker:
loncharles — thanks again for the detailed notes. I believe this PR should be scoped to the LaunchDaemon restart behavior, with the two issues you raised handled separately, since they're really independent of the self-healing change here:

- **socket_vmnet silent hang** — addressed in #5091 (fail fast + retry with backoff for the XPC timing race).
- **APIC timer kernel panic** — this is a guest-kernel issue on x86 under HVF, independent of networking or how the VM is launched. I looked at shipping the `no_timer_check` fix in the templates, but it can't be done cleanly: the cmdline can only be set by pinning an external kernel + initrd by digest, which goes stale on every Ubuntu kernel update and would pin a kernel for all users to fix an Intel-only panic. That's presumably why it's only in CI today (via `inject-cmdline-to-template.sh`). Probably better tracked under #84 where it can get a proper fix.

EDIT: tagged the wrong person on these originally... removing the "@" to spare them further noise.

.gitignore
```diff
@@ -4,3 +4,4 @@ lima.REJECTED.yaml
 default-template.yaml
 schema-limayaml.json
 .config
+/limactl
```
AkihiroSuda: Irrelevant change

cmd/limactl/start.go
```diff
@@ -626,6 +653,51 @@ func startAction(cmd *cobra.Command, args []string) error {
 	return instance.Start(ctx, inst, launchHostAgentForeground, progress)
 }
 
+// isOrphanedDriverError returns true when the instance errors indicate that the VM driver
+// process is running but the host agent is not — the recoverable broken state that occurs
+// after an unclean shutdown (e.g. launchd SIGTERM before the host agent could stop the VM).
+func isOrphanedDriverError(errs []error) bool {
+	for _, err := range errs {
+		if strings.Contains(err.Error(), "driver is running but host agent is not") {
```
AkihiroSuda: Use `errors.Is` or `errors.As`

cmd/limactl/autostart_darwin.go
```diff
@@ -0,0 +1,168 @@
...
+	return nil
+}
+
+// runSudo executes a command under sudo, inheriting the terminal so password prompts work.
+func runSudo(ctx context.Context, args ...string) error {
+	cmd := exec.CommandContext(ctx, "sudo", args...) //nolint:gosec // args are constructed internally, not from user input
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
+	logrus.Debugf("running: sudo %v", args)
```
AkihiroSuda: In the case of sudo, this should be explicitly printed

AkihiroSuda:
> Note: this PR stacks on https://github.com/lima-vm/lima/pull/5086 (keep-alive flag) which stacks on https://github.com/lima-vm/lima/pull/4984 (LaunchDaemon support). The diff includes those commits as context; the new changes are the four commits on top. Target base will move to master once those merge.

Let me mark this PR as a draft

resker:
@AkihiroSuda — addressed:
- dropped the `.gitignore` change
- recovery detection now matches sentinel errors in `pkg/store` (`ErrDriverRunningHostAgentStopped`, `ErrHostAgentUnreachable`) via `errors.Is` instead of string-matching
- `runSudo` now logs at info level and notes it may prompt for a password

Also re-stacked on the updated #5086.

resker:
@AkihiroSuda — I tested this directly. Two clarifications:

1. launchd isn't restarting a separate host-agent process of its own — the LaunchDaemon just runs `limactl start <instance> --foreground`, and `KeepAlive` re-runs that same command. So "restarting the host agent" really just means "re-running `limactl start`", the normal start path.
2. With vz the VM runs in-process in the host agent, so an unclean exit doesn't leave a separate VM process behind — the broken states this PR handles come from stale PID files after a reboot (a persisted PID that's since been reused by another process).

I reproduced both broken states on a test instance and confirmed `limactl start` recovers each one:
- **orphaned driver** ("driver is running but host agent is not") → force-stops the orphan, then starts cleanly → Running.
- **stale host agent** (ha.pid live but socket unreachable) → removes the stale files and starts cleanly → Running, *without* signaling the possibly-unrelated PID.

resker:
Rebased onto latest master (no functional changes).

resker:
Thanks for looking at #5086, @AkihiroSuda!

This follow on is now unblocked (rebased, no longer marked as draft) & should be ready.

Since your review in June the only change switches the recovery-state matching from error-string comparison to `errors.Is` (in response to your feedback). The launchd question from that thread is answered in my reply above: the LaunchDaemon runs `limactl start <instance> --foreground`, so it isn't restarting a separate host-agent process of its own.

CFBancroft:
Intel test update for #5088.

First, for transparency: I am a Mac mini user, not a Lima developer. I have been working through this test step-by-step with help from ChatGPT. ChatGPT helped me build the PR branch, interpret the Lima/launchd logs, and isolate the tests below. I am posting the actual results from my 2018 Intel Mac mini so the maintainers can decide what is technically useful.

Test system

Mac mini (2018, Macmini8,1)
Intel Core i7 / x86_64
64 GB RAM
macOS Sequoia 15.7.9 (24G830)
Lima 2.2.0
Colima 0.10.3
VZ backend
FileVault off
Dedicated hidden macOS service account for the LaunchDaemon
PR branch: feature/vz-shutdown-recovery
Tested commit: 0ba123d4
PR limactl built locally as x86_64 with Go 1.27.0

Manual recovery test on Intel: PASS

Before recovery, the PR build reported:

vz driver is running but host agent is not

and the instance status was Broken.

I then ran the PR build with:

limactl start colima

It detected the orphaned state and entered the new recovery path:

Instance "colima" has an orphaned driver process; attempting force stop to recover

It then attempted:

Sending SIGKILL to the vz driver process

which returned:

operation not permitted

However, recovery continued successfully. It removed stale *.pid, *.sock, and *.tmp state (including vz.pid), restarted the VZ instance, restored the Docker socket, and eventually reached:

READY

Docker containers restarted normally afterward. I have a DokuWiki container published on host port 5555; after recovery it became reachable both from the Intel Mac mini itself and from another Mac on my LAN.

This also appears relevant to #5075: on this Intel system I directly observed the #5088 recovery code cleaning stale pid/socket/tmp state and successfully recovering the instance.

Unattended pre-login reboot test: BLOCKED before Lima executes

I also tested a real reboot, leaving the Mac mini at the macOS login screen with no GUI user logged in.

DokuWiki did not become reachable.

After logging in to investigate, the system LaunchDaemon showed:

state = not running

last exit code = 78: EX_CONFIG

The macOS unified log showed:

Service could not initialize: posix_spawn(/usr/local/bin/limactl-5088-test), error 0xd - Permission denied

Initially I suspected my locally built PR binary. I therefore tested the stock Homebrew Lima 2.2.0 binary instead.

The same LaunchDaemon failed with:

Service could not initialize: posix_spawn(/usr/local/bin/limactl), error 0xd - Permission denied

I also bypassed the /usr/local/bin/limactl symlink and pointed the LaunchDaemon directly at:

/usr/local/Cellar/lima/2.2.0/bin/limactl

That produced the same result:

posix_spawn(...), error 0xd - Permission denied

However, the stock binary executes successfully when explicitly run as the same service account:

sudo -u containerserver /usr/local/bin/limactl --version

returns:

limactl version 2.2.0

I also compared the stock and PR binaries. Both are x86_64 Mach-O binaries, both use ad-hoc signatures, and neither has a TeamIdentifier. Both are reported as rejected by spctl, so that did not distinguish the PR build from the stock Homebrew build.

Therefore, as far as I can determine with ChatGPT's help, the current posix_spawn EACCES appears to be a separate macOS LaunchDaemon execution-context problem on this machine, rather than a failure of #5088 itself. At boot, limactl never executes, so the #5088 recovery code never gets an opportunity to run.

Current Intel results

Build PR on Intel/x86_64: PASS
Detect broken/orphaned VZ state: PASS
#5088 manual recovery: PASS
Cleanup of stale pid/socket/tmp state: PASS
VZ returns to running state: PASS
Docker socket/container recovery: PASS
DokuWiki reachable again locally/LAN: PASS
Unattended pre-login LaunchDaemon recovery: BLOCKED before Lima executes by macOS posix_spawn(...): Permission denied

I am continuing to investigate the LaunchDaemon posix_spawn problem. If there is a specific Intel test, log, or command that would be useful for #5088, please tell me what to run and I will test it on this Mac mini.

CFBancroft:
Intel Mac test result for Pull Request #5088

I am a regular Mac mini user, not a Lima developer. I performed this testing step-by-step with help from ChatGPT.

Test hardware

2018 Intel Mac mini (Macmini8,1)
Intel Core i7, 6-core 3.2 GHz
64 GB RAM
macOS Sequoia 15.7.9 (24G830)
Lima/Colima using Apple VZ
Lima VM architecture: x86_64
Separate Mac Studio used as the remote test/control machine

How the headless test was performed

This was intended to be a real unattended server test, not a GUI-login test.

I enabled Remote Login on the Intel Mac mini and then rebooted it while the Lima/Colima VM was running.

After reboot, I did not log into any macOS GUI account on the Intel Mac mini. The Mac mini remained at the macOS login screen.

From a separate Mac Studio, I attempted to access my Docker-hosted DokuWiki service on host port 5555. The connection was refused, showing that the container stack had not started.

I then connected from the Mac Studio to the Intel Mac mini using SSH, while the Intel Mac mini was still sitting at its macOS login screen.

All subsequent launchctl, Unified Log, filesystem, and Lima checks were performed through that SSH session. This allowed me to inspect the failed boot without first logging into the Mac mini GUI and changing the headless test conditions.

Build result: PASS

I built the current #5088 branch locally on the Intel Mac. The resulting limactl was a native x86_64 Mach-O executable.

Manual #5088 recovery test: PASS

Before the headless reboot test, I reproduced the broken/orphaned Lima state:

vz driver is running but host agent is not

Running the #5088 limactl start colima manually as my dedicated containerserver account successfully detected the orphaned state, cleaned stale state, restarted VZ, restored the Docker socket, and reached:

READY.

My DokuWiki container on port 5555 then became reachable again.

This confirms that the #5088 recovery logic itself worked on this Intel Mac.

Normal Intel VZ shutdown: PASS

For additional Intel testing, I added a temporary diagnostic to vz_driver_darwin.go to report CanRequestStop().

During a fresh normal start/stop cycle on this Intel Mac, the result was:

CFB Intel test: CanRequestStop=true

The VZ VM shut down normally.

Therefore, the normal Intel shutdown path does not reproduce the CanRequestStop=false condition involved in #5087/#5088.

Headless LaunchDaemon reboot test: BLOCKED before Lima executes

After the real reboot, while the Intel Mac mini remained at the login screen, SSH inspection from the Mac Studio showed that the system LaunchDaemon had attempted to run once and exited with:

last exit code = 78: EX_CONFIG

The macOS Unified Log showed:

Service could not initialize: posix_spawn(.../limactl), error 0xd - Permission denied

Importantly, this occurs before limactl itself starts, so the #5088 recovery code never gets an opportunity to execute during this boot test.

I reproduced the posix_spawn EACCES behavior using multiple executable locations, including:

the Homebrew Lima 2.2.0 binary
my locally built #5088/CFB test binary
a copy of that binary placed under /Users/Shared/Server

The same locally built binary executes successfully when invoked manually as the dedicated containerserver account.

I also unloaded and freshly reloaded the LaunchDaemon after changing the executable location. launchd correctly loaded the new /Users/Shared/Server executable path, but still failed with:

posix_spawn(/Users/Shared/Server/lima-cfb/limactl), error 0xd - Permission denied

Therefore, simply moving the executable does not resolve the LaunchDaemon problem.

Result

On this 2018 Intel Mac mini:

Building #5088 for x86_64: PASS
Running the #5088 build manually: PASS
#5088 recovery from the broken/orphaned VZ state: PASS
Normal Intel VZ shutdown (CanRequestStop=true): PASS
Real reboot with no GUI login: TESTED
Remote verification from a separate Mac Studio over SSH while the Intel Mac remained at the login screen: CONFIRMED
Automatic LaunchDaemon execution after reboot: BLOCKED by macOS posix_spawn EACCES

Therefore, I cannot honestly mark the complete unattended reboot test for #5088 as either PASS or FAIL. The Intel #5088 code works when executed, but my true headless boot test is currently blocked by a separate macOS LaunchDaemon execution problem that occurs before Lima starts.

For Akihiro Suda

I understand that you are in the Tokyo area and mentioned that you do not currently have access to an Intel Mac for testing.

Accessibility note: I am Deaf. I communicate through written English. I also use some Japanese Sign Language (JSL), but I do not use Japanese fingerspelling. If we meet in person, please communicate with me using written English.

I have a train pass between Fujisawa and Ueno on the JR Tokaido Line (Ueno-Tokyo Line), so I can meet you somewhere convenient along that line.

I can bring my 2018 Intel Mac mini, keyboard, mouse, and a very small monitor. We can meet at a café or another place with a power outlet. That way, you can test Lima directly on an actual Intel Mac.

Alternatively, if meeting is unnecessary, please give me the exact commands or test procedure you want me to run. I can run them on the Intel Mac mini and post the complete terminal output here.

CFBancroft:
Additional Technical Details — Exact Intel Test Code and Commands

Follow-up to my previous Intel Mac test report above.

I realized that my previous post described the hardware, headless reboot/SSH test, results, and the LaunchDaemon failure, but I did not include the actual source-code modification and exact commands I used during the Intel testing.

I am adding those details here so the test can be reproduced and so it is clear exactly what I changed and what I observed.

Important: The source modification below is my local CFB Intel experiment on top of Pull Request #5088. It is not code from Pull Request #5088 itself. I made this modification only to observe the Intel VZ shutdown behavior and experiment with the SSH-failure path.

Local Intel diagnostic modification

In:

pkg/driver/vz/vz_driver_darwin.go

immediately after:

logrus.Info("Shutting down VZ")

I added:

logrus.Infof("CFB Intel test: CanRequestStop=%v", l.machine.CanRequestStop())

I also changed:

logrus.WithError(err).Warn("vz: SSH shutdown failed; VM process may require manual cleanup")
return nil

to:

logrus.WithError(err).Warn("vz: SSH shutdown failed; continuing to observe VM state")

The purpose of this second experimental change was to allow execution to continue into the existing 30-second VM-state polling loop rather than immediately returning when the SSH shutdown request failed.

My resulting local diff was:

@@ -501,6 +501,7 @@
 func (l *LimaVzDriver) Stop(ctx context.Context) error {
         logrus.Info("Shutting down VZ")
+        logrus.Infof("CFB Intel test: CanRequestStop=%v", l.machine.CanRequestStop())
         if l.machine.CanRequestStop() {
 ...
-                        logrus.WithError(err).Warn("vz: SSH shutdown failed; VM process may require manual cleanup")
-                        return nil
+                        logrus.WithError(err).Warn("vz: SSH shutdown failed; continuing to observe VM state")
Building on the Intel Mac

I built limactl directly on the 2018 Intel Mac mini:

make limactl

The build completed for:

GOARCH=amd64

and produced a native x86_64 limactl.

The locally built version reported:

limactl version 2.1.2-beta.0-633-g0ba123d4.m

I also built the x86_64 guest agent and templates:

make x86_64-guestagent templates

The guest-agent target produced a gzip-related Error 1, but the resulting compressed x86_64 guest-agent file existed:

_output/share/lima/lima-guestagent.Linux-x86_64.gz

I then ran:

make templates

which completed successfully and populated the required runtime templates.

Starting the existing VM with my modified Intel build

I started my existing Colima VM with the locally compiled binary:

sudo -u containerserver -H env \
LIMA_HOME=/Users/containerserver/.colima/_lima \
/Users/macmini-clainkawada/Developer/lima-5088/_output/bin/limactl start colima

The VM successfully started and ultimately reported:

READY.

Docker became available again and my DokuWiki container on host port 5555 was reachable.

Intel CanRequestStop() result

I then performed a fresh normal start/stop cycle using the modified build.

My diagnostic produced:

CFB Intel test: CanRequestStop=true

The VZ VM shut down normally.

This shows that on my 2018 Intel Mac mini, a normal VZ shutdown uses the CanRequestStop=true path. I therefore could not reproduce CanRequestStop=false simply by performing a normal Intel start/stop cycle.

Pull Request #5088 recovery test

Separately from my local diagnostic modification, I tested the #5088 recovery behavior.

The instance had reported:

vz driver is running but host agent is not

I started it using the #5088 test build:

sudo -u containerserver -H env \
LIMA_HOME=/Users/containerserver/.colima/_lima \
/usr/local/bin/limactl-5088-test start colima

The #5088 code detected the orphaned state and attempted recovery.

During that process, the attempt to send SIGKILL to the old VZ driver reported:

operation not permitted

However, recovery continued. Stale state including vz.pid was removed, VZ restarted, the Docker socket was restored, and Lima ultimately reported:

READY.

My DokuWiki service on port 5555 then became reachable again.

Therefore, manual #5088 broken/orphaned-state recovery worked on this Intel Mac despite the SIGKILL operation-not-permitted message.

Attempt to reproduce the abnormal state manually

I also experimented with killing only the host agent after obtaining its PID from ha.pid.

After sending SIGKILL to that host-agent process, this did not reproduce the original orphaned-VZ condition. The VZ process disappeared as well, and limactl list showed the instance as stopped.

Therefore, I do not count that experiment as evidence either for or against #5088.

Relationship to my previous report

These are the source-code and command details that I accidentally omitted from my previous comment.

My previous comment contains the separate real headless reboot test: the Intel Mac mini was rebooted and left at the macOS login screen, I tested the service from a separate Mac Studio, and then connected to the Intel Mac mini from the Mac Studio using SSH without logging into the Intel Mac GUI.

That test remains blocked by the separate macOS LaunchDaemon failure:

Service could not initialize: posix_spawn(.../limactl), error 0xd - Permission denied

Because that error occurs before limactl executes, I do not consider the headless LaunchDaemon result a failure of #5088 itself.

Summary of this follow-up:

#5088 builds natively on my 2018 Intel Mac: PASS
#5088 manual orphaned-state recovery on Intel: PASS
My diagnostic build runs on Intel: PASS
Normal Intel VZ shutdown: CanRequestStop=true
Deliberately killing the host agent did not reproduce the original orphaned state
Full headless reboot testing remains blocked before Lima executes by the separate macOS launchd posix_spawn EACCES issue

I apologize for not including these exact technical details in my first report. I wanted to add them rather than edit the original comment so that the testing history remains clear.

resker:
@CFBancroft - A few questions:

1. How did you create your LaunchDaemon? Was it w/ `limactl autostart enable --condition=boot <instance>`, or something put together manually? Lima's generated daemon plist sets UserName, so the job runs as that user rather than root — a hand-written one without it will behave quite differently.

2. If it is running as user other than root, does this work?

   `sudo -u containerserver /Users/Shared/Server/lima-cfb/limactl --version`

3. Could you post the output of the following (making sure it doesn't reveal anything sensitive / private beforehand)?

    `sudo plutil -p /Library/LaunchDaemons/<the plist for your Lima job>.plist`

resker:
Rebased onto latest master (clean, no conflicts) and pushed four commits addressing the Copilot review:

- **vz `Stop`** — a failed SSH shutdown request returned `nil`, so the host agent exited 0 while the VM was likely still running, and `KeepAlive{SuccessfulExit=false}` would not restart it. It now falls through to the existing stop poll, as the `CanRequestStop()` branch already does.
- **`ErrHostAgentUnreachable`** — was attached to every failure to reach the host agent, including a timeout and the startup race where `ha.pid` is written before the socket is bound. Now only on `ECONNREFUSED`, the one result proving nothing is listening. Other failures are still recorded and still leave the instance `Broken`; they're just not auto-recovered. (On Windows `WSAECONNREFUSED` isn't mapped to `ECONNREFUSED` by `syscall.Errno.Is`, so the sentinel never fires there — reported but not recovered, which is the safe direction.)
- **Stale-file cleanup** — removed every `.pid`/`.sock`/`.tmp` in the instance dir, which could delete a live driver's PID file and let a second driver start. Now only `ha.pid`/`ha.sock`, plus the driver PID file when the driver runs in-process (vz) and records that same PID.
- The two recovery paths now run in sequence rather than as alternatives, so a driver revealed by clearing the host agent is force-stopped on its own — and since that happens after `ha.pid` is gone, the force stop can't signal a PID that may have been reused.

Added unit tests for both: the socket cases drive the real host agent client against a unix socket rather than synthetic errors, and the cleanup cases cover in-process vs. separate driver.

CFBancroft:
@resker Thank you. Here are the answers to your questions.

I powered on the 2018 Intel Mac mini again tonight (September 6). I left the Mac mini at the macOS login screen and did not log into a GUI account. I am currently accessing it remotely from my separate Mac Studio using SSH. This is the same headless arrangement I was trying to test previously.

1. How was the LaunchDaemon created?

The LaunchDaemon I tested was created manually. I did not create it using:

limactl autostart enable --condition=boot

This may be an important distinction. My previous report should not be interpreted as a test of the LaunchDaemon generated by Lima's own autostart enable --condition=boot command.

I was trying to run the existing Colima/Lima VM under a dedicated macOS service account named containerserver, with no GUI login required.

2. Can containerserver execute the test binary manually?

Yes.

I previously ran:

sudo -u containerserver /Users/Shared/Server/lima-cfb/limactl --version

and received:

limactl version 2.1.2-beta.0-633-g0ba123d4.m

The same locally built binary also successfully starts the existing Lima/Colima VM when I invoke it manually as containerserver with the appropriate LIMA_HOME.

Therefore, the containerserver account itself can execute this binary.

3. Current LaunchDaemon plist

I ran the command you requested:

sudo plutil -p /Library/LaunchDaemons/io.lima-vm.daemon.colima.plist

Output:

{
  "EnvironmentVariables" => {
    "HOME" => "/Users/containerserver"
    "LIMA_HOME" => "/Users/containerserver/.colima/_lima"
  }
  "Label" => "io.lima-vm.daemon.colima"
  "ProcessType" => "Background"
  "ProgramArguments" => [
    0 => "/Users/Shared/Server/lima-cfb/limactl"
    1 => "start"
    2 => "colima"
    3 => "--foreground"
  ]
  "RunAtLoad" => 1
  "StandardErrorPath" => "/var/log/colima-boot-error.log"
  "StandardOutPath" => "/var/log/colima-boot.log"
  "UserName" => "containerserver"
  "WorkingDirectory" => "/Users/containerserver/.colima/_lima/colima"
}

This is the manually created plist that produced the posix_spawn(.../limactl), error 0xd - Permission denied behavior I reported earlier.

I also want to correct/clarify my previous report: because this plist was manually created, my LaunchDaemon failure does not demonstrate that Lima's own limactl autostart enable --condition=boot implementation has the same problem.

If you would like me to test the official Lima-generated boot autostart configuration instead, please give me the exact commands/procedure you want me to use. I will run them on this Intel Mac mini and report the complete output.

Likewise, I see that #5088 has changed since the 0ba123d4 build I tested. If you want me to rebuild and test the current #5088 head on this Intel Mac, please tell me which test sequence you would like me to run, and I will follow that sequence exactly.

resker:
@CFBancroft — thanks, that clarifies things.

It also turned up a real gap, so thank you for that: `limactl autostart enable` generates a plist with no `EnvironmentVariables`, so `LIMA_HOME` isn't propagated and the job looks in `~/.lima` at boot — which would never find a Colima instance under `~/.colima/_lima`. Presumably why you hand-rolled the plist to begin with. Filed as #5488 with a draft fix in #5489 (credited you as reporter).

Independent of this PR, so nothing here changes.

If you're still curious about the EACCES on the manual plist, the one thing I didn't get was:

`ls -lde / /Users /Users/Shared /Users/Shared/Server /Users/Shared/Server/lima-cfb`

Since `sudo -u containerserver ... --version` works the mode bits look fine, so an ACL or a mount option is the likelier culprit.

CFBancroft:
@resker Thanks. I ran the exact command you requested on the 2018 Intel Mac mini.

The Mac mini is still being accessed remotely from my Mac Studio over SSH.

$ ls -lde / /Users /Users/Shared /Users/Shared/Server /Users/Shared/Server/lima-cfb

drwxr-xr-x  22 root  wheel  704 Aug  2 01:31 /
drwxr-xr-x  11 root  admin  352 Aug 31 20:26 /Users
drwxrwxrwt  27 root  wheel  864 Aug 31 23:06 /Users/Shared
drwxr-xr-x   4 root  wheel  128 Sep  1 18:40 /Users/Shared/Server
drwxr-xr-x   3 root  wheel   96 Sep  1 18:41 /Users/Shared/Server/lima-cfb

There were no additional ACL entries printed by ls -lde.

I have not changed any permissions, ACLs, mount settings, the LaunchDaemon, Lima, or Colima after running this command.

If you would like another check on this Intel Mac, please give me the exact command(s) and I will run them and post the output.

resker:
@CFBancroft — that rules out the path then. I ought to point out that this thread isn't ideal for troubleshooting and isn't really topical to the PR at this point — it's not anything Lima is generating.

I'd suggest keeping an eye on #5488 / #5489 and adopting them if they merge. #5489 makes `limactl autostart enable --condition=boot colima` work with Colima's LIMA_HOME, which is what your hand-written plist is trying to do today — Lima generates the plist itself (with UserName and KeepAlive). #5088 here is the other half: recovery after an unclean shutdown.

resker:
@AkihiroSuda — bumping this... since your review: a few Copilot findings were addressed & it's been rebased. CI indicates green — the one red is a cancelled release job. Most of the thread above is unrelated, but @CFBancroft also validated the recovery on an Intel Mac (I've only access to Apple Silicon based systems).

Anything further needed?

AkihiroSuda (commented):
Please tidy up the commits (one commit for one topic)

pkg/store/staleagent_test.go
```diff
@@ -0,0 +1,104 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build !windows
```
AkihiroSuda, line 4: The file should be named `*_test_unix.go`

resker:
@AkihiroSuda - thanks and I've collapsed everything to four commits, one per topic (each builds on its own):

- fix(vz) (SSH shutdown fallback)
- fix(store) (the recoverable-state sentinels)
- fix(start) (the recovery itself)
- docs(autostart). 

But the contributing guide says commits are usually squashed to one before merge — please advise if that's what you meant.

Also renamed the test file to staleagent_unix_test.go, matching pkg/limayaml/defaults_unix_test.go. I read your *_test_unix.go as a typo for that — let me know if you meant something else.

Rebased on master and updated the description to match the new commits.

resker:
@AkihiroSuda — rebased. #5436 landed yesterday and supersedes my `fix(vz)` commit, so I dropped it (thanks @ekalinin — yours also removes `vz.pid` on stop, mine didn't). Three left, still one per topic: `fix(store)`, `fix(start)`, `docs(autostart)`. The docs commit lost its VZ shutdown section for the same reason.

This PR is still applicable even though #5436 and e529ce08 both narrow how these states arise. The description has been rewritten to cover what's left rather than lean on the launchd SIGTERM story those two now cover — mainly an out-of-process driver (qemu) outliving a host agent that died in the same boot, which needs neither a reboot nor PID reuse, and instances with no boot session marker because an older Lima started them. #5075 is that same state, reported independently.

This is on the v2.3.0 milestone and #5436 went in after beta.0, so please advise if anything further is needed to proceed.

- These are one commit per topic, as requested. The contributing guide does say squash to one before merge — so happy to do that if preferred.

PR #5090, opened 2026-06-06
limayaml: do not default containerd.user=true on non-Linux guests

### LINKED ISSUE #5037: FillDefault enables containerd.user for macOS guests on aarch64 — nerdctl unnecessarily downloaded

gaurav0107:
## What

`FillDefault` in `pkg/limayaml/defaults.go` enabled `containerd.user=true` whenever the guest architecture was `x86_64` or `aarch64`, with no check on the guest OS. macOS (`os: Darwin`) and FreeBSD guests therefore defaulted to `containerd.user=true` even though nerdctl is a Linux-only runtime, causing the ~250 MiB nerdctl archive to be downloaded unnecessarily.

## How

Gate the `true` default on `*y.OS == limatype.LINUX`:

```go
if y.Containerd.User == nil {
    if *y.OS == limatype.LINUX {
        switch *y.Arch {
        default:
        }
    } else {
    }
}
```

`y.OS` is already resolved earlier in `FillDefault` (line 157 via `ResolveOS`), so dereferencing is safe.

## Behavior

- Linux + x86_64 / aarch64 → `true` (unchanged)
- Linux + riscv64 / armv7l / ppc64le / s390x → `false` (unchanged)
- Darwin + any arch → `false` (was `true` for x86_64 / aarch64; this is the fix)
- FreeBSD + any arch → `false` (was `true` for x86_64 / aarch64; this is the fix)
- Explicit `containerd.user: true` in user YAML → preserved (override path runs before this block)

## Tests

- `TestContainerdUserDefaultPerOS` — table-driven test covering Linux/Darwin/FreeBSD across `x86_64`, `aarch64`, `riscv64`, `armv7l`.
- `TestContainerdUserExplicitOverride` — verifies an explicit `containerd.user: true` is preserved on a Darwin guest.

Existing `TestFillDefault` already pins Linux + host arch and continues to pass.

```
ok  github.com/lima-vm/lima/v2/pkg/limayaml  0.342s
```

Fixes #5037

### DIFF
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -887,3 +887,54 @@ func TestMountTagDuplicateLocation(t *testing.T) {
 	tag2 := MountTag(location, "/mnt/home-writable")
 	assert.Assert(t, tag1 != tag2)
 }
+
+// TestContainerdUserDefaultPerOS verifies that FillDefault only enables
+// containerd.user=true on Linux guests. nerdctl is a Linux-only runtime,
+// so non-Linux guests (Darwin, FreeBSD) must default to false regardless
+// of the host architecture. Regression test for issue #5037.
+func TestContainerdUserDefaultPerOS(t *testing.T) {
+	cases := []struct {
+		os   limatype.OS
+		arch limatype.Arch
+		want bool
+	}{
+		{limatype.LINUX, limatype.X8664, true},
+		{limatype.LINUX, limatype.AARCH64, true},
+		{limatype.LINUX, limatype.RISCV64, false},
+		{limatype.LINUX, limatype.ARMV7L, false},
+		{limatype.DARWIN, limatype.AARCH64, false},
+		{limatype.DARWIN, limatype.X8664, false},
+		{limatype.FREEBSD, limatype.X8664, false},
+		{limatype.FREEBSD, limatype.AARCH64, false},
+	}
+	for _, tc := range cases {
+		t.Run(fmt.Sprintf("%s/%s", tc.os, tc.arch), func(t *testing.T) {
+			y := limatype.LimaYAML{
+				OS:   ptr.Of(tc.os),
+				Arch: ptr.Of(tc.arch),
+			}
+			var d, o limatype.LimaYAML
+			FillDefault(t.Context(), &y, &d, &o, "", false)
+			assert.Assert(t, y.Containerd.User != nil, "Containerd.User must be set after FillDefault")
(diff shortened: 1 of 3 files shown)


pkg/limayaml/defaults_test.go
```diff
@@ -887,3 +887,54 @@ func TestMountTagDuplicateLocation(t *testing.T) {
 	tag2 := MountTag(location, "/mnt/home-writable")
 	assert.Assert(t, tag1 != tag2)
 }
+
+// TestContainerdUserDefaultPerOS verifies that FillDefault only enables
+// containerd.user=true on Linux guests. nerdctl is a Linux-only runtime,
```
jandubois:
```suggestion
// containerd.user=true on Linux guests for x86_64 and aarch64. nerdctl is a Linux-only runtime,
```

jandubois (changes requested):
Needs an update to `templates/default.yaml` too:

```yaml
  # Enable user-scoped (aka rootless) containerd and its dependencies
  # 🟢 Builtin default: true (for x86_64 and aarch64)
  user: null
```

gaurav0107:
Thanks for the review! Updated `templates/default.yaml` in 6791581 to reflect the new behavior:

```yaml
  # Enable user-scoped (aka rootless) containerd and its dependencies
  # 🟢 Builtin default: true for Linux x86_64 and aarch64 guests, false otherwise
  user: null
```

Followed the wording pattern from 3fc5e85 (`vz: disable ssh.overVsock by default for non-Linux guests`). PTAL.

jandubois:
Thanks, but please squash commits

pkg/limayaml/defaults_test.go
gaurav0107: Applied in 19fab750. Thanks for the review.

gaurav0107:
@jandubois The `templates/default.yaml` comment was updated in 67915812 ("limayaml: clarify containerd.user default is Linux-only in default.yaml") earlier in this branch, and the test-comment suggestion is applied in 19fab750. Both review items should be addressed now — please re-review when convenient.

AkihiroSuda (approved):
Thanks

AkihiroSuda:
Next time please make sure to squash the commits

PR #5027, opened 2026-05-23
qemu: use stream netdev with reconnect for socketVMNet networking

loncharles:
## Summary

Switch QEMU socketVMNet networking on macOS from legacy `-netdev socket` with a pre-dialed fd to `-netdev stream` with built-in reconnect support. Applies to all socketVMNet modes (bridged, shared, host).

- **Feature-detected**: checks `-netdev help` output for `stream` support; falls back to `socket` if unavailable
- **Version-aware reconnect**: `reconnect-ms=500` for QEMU >= 9.2, `reconnect=1` for QEMU 8.0–9.1 (the `reconnect` parameter was renamed to `reconnect-ms` in 9.2 and the old form removed in 10.2), no reconnect for QEMU 7.2–7.9
- **Wire-compatible**: both `stream` and `socket` netdev use the same 4-byte big-endian length-prefixed Ethernet frame protocol that socket_vmnet uses

## Motivation

When the socket_vmnet daemon restarts or the UNIX socket connection breaks, the legacy `-netdev socket` with a pre-dialed fd has no way to recover — the fd is dead and the VM's network is permanently broken until manual VM restart.

The `stream` netdev connects directly to the socket_vmnet UNIX socket and can automatically reconnect, recovering VM networking without restart.

## Performance

In testing on macOS Intel (QEMU 11.0.0, HVF, socket_vmnet bridged to a 5GbE NIC), the stream netdev showed ~2x throughput improvement over legacy socket netdev:

| | socket (before) | stream (after) |
|---|---|---|
| Throughput (VM → LAN host, 3 runs avg) | ~970 Mbits/sec | ~1,840 Mbits/sec |
| QEMU CPU during iperf3 | ~219% | ~216% |
| Ping latency (100 pings) | 0.76ms avg | 0.88ms avg |

Host-to-host baseline on the same NIC is 4.66 Gbits/sec; the remaining gap is in the socket_vmnet relay path.

## Compatibility

- QEMU < 7.2: no change (falls back to existing `socket` netdev)
- QEMU 7.2+: uses `stream` netdev (available since 7.2)
- QEMU 8.0+: adds reconnect support (1-second granularity)
- QEMU 9.2+: uses `reconnect-ms` (millisecond granularity, 500ms)
- macOS only (socketVMNet code path is gated on `runtime.GOOS == "darwin"`)
- No change to usernet or raw socket networking paths

## Test plan

- [x] Verified stream netdev starts and VM boots on QEMU 11.0.0
[...]

PR #5091, opened 2026-06-07
fix(networks): fail fast when socket_vmnet daemon does not start

resker:
## Problem

`startDaemon()` launches socket_vmnet and returns immediately without verifying the process stayed alive. If `vmnet_start_interface` returns `VMNET_FAILURE` — for example because the `com.apple.NetworkSharing` XPC Mach port is not yet registered at early boot — socket_vmnet exits silently. Lima then starts the VM anyway and blocks indefinitely waiting for SSH that will never arrive, with no indication of what went wrong.

This is especially likely when using `limactl autostart enable --condition=boot` (#4984), where Lima starts before all XPC services are guaranteed to be registered.

## Fix

Add `waitForDaemon()`, called in `startNetwork()` immediately after `startDaemon()`. socket_vmnet writes its PID file only after `vmnet_start_interface` succeeds, so the presence of that file is a reliable indicator of successful startup.

- **Transient early exit** (the `VMNET_FAILURE` race): the daemon exits before writing its PID file, so Lima retries with exponential backoff (1/2/4/8s, up to 5 attempts).
- **Stuck process**: if the daemon instead stays alive past a generous timeout (30s) without writing its PID file, that points at a different problem, so Lima fails fast and surfaces the daemon's stderr rather than hanging:

```
socket_vmnet daemon for "bridged" network did not write its PID file within 30s but is still running (stderr: ...)
```

- **Non-transient failure** (an unreadable PID file, a cancelled context, or a failure from `startDaemon` itself such as sudo/exec): returned immediately, without retrying.

## Effect on other use cases

- **Normal startup:** socket_vmnet writes its PID file within ~1s. No user-visible change.
- **Already running (pid != 0):** `startDaemon` is skipped entirely; `waitForDaemon` is never called.
- **Linux / non-darwin:** Early return before this code path. No change.
- **Usernet networking:** Separate code path. No change.
- **QEMU + socket_vmnet:** Same improvement — not VZ-specific.

Ref: #5074 (XPC timing), #4984 (LaunchDaemon boot condition)

Signed-off-by: Robert Esker <resker@gmail.com>

### DIFF
--- a/pkg/networks/reconcile/reconcile.go
+++ b/pkg/networks/reconcile/reconcile.go
@@ -6,6 +6,7 @@ package reconcile
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -112,20 +113,22 @@ func makeVarRun(ctx context.Context, cfg *networks.Config) error {
 	return nil
 }
 
-func startDaemon(ctx context.Context, cfg *networks.Config, name, daemon string) error {
+// startDaemon starts the daemon process and returns its running *exec.Cmd so the
+// caller can wait on the process while polling for its PID file.
+func startDaemon(ctx context.Context, cfg *networks.Config, name, daemon string) (*exec.Cmd, error) {
 	if err := makeVarRun(ctx, cfg); err != nil {
-		return err
+		return nil, err
 	}
 	networksDir, err := dirnames.LimaNetworksDir()
 	if err != nil {
-		return err
+		return nil, err
 	}
 	if err := os.MkdirAll(networksDir, 0o755); err != nil {
-		return err
+		return nil, err
 	}
 	user, err := cfg.User(daemon)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	args := []string{"--user", user.User, "--group", user.Group, "--non-interactive"}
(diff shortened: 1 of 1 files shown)


resker:
@AkihiroSuda — open question on approach: should Lima retry starting socket_vmnet (with backoff) rather than failing fast? The XPC timing issue (#5074) means VMNET_FAILURE at early boot is often transient. Happy to extend this if retry is preferred.

AkihiroSuda:
> @AkihiroSuda — open question on approach: should Lima retry starting socket_vmnet (with backoff) rather than failing fast? The XPC timing issue (#5074) means VMNET_FAILURE at early boot is often transient. Happy to extend this if retry is preferred.

Retrying SGTM
cc @nirs

resker:
@AkihiroSuda — updated with retry logic per your suggestion.

jandubois (dismissed):
I think the `waitForDaemon` design is at odds with the "fail fast" goal of the PR. Instead `startDaemon` should return the `*exec.Cmd`, so you can select on it. So either the pidfile appears within a generous timeout, or the process exits. When the process dies, you retry with the backoff strategy. If the process doesn't die, but the pidfile doesn't appear within the larger timeout, then something else is wrong.

resker:
@jandubois — thanks, makes sense. Reworked it along those lines: `startDaemon` now returns the `*exec.Cmd`, and the wait selects on the process. If it exits before writing the PID file we retry with backoff; if it stays alive past a generous timeout (30s) without a PID file we fail fast and surface stderr. Also rebased onto master, since the branch predated the `pkgsudoers` refactor.

resker:
Pushed a follow-up while this was in the queue: surface non-transient PID-file read errors instead of treating them as a startup timeout, preserve the daemon's exit error in the retry path, and close the stdout/stderr file copies in the parent. The retry/wait rework from the earlier review — `startDaemon` returning the `*exec.Cmd` and selecting on process exit — is included and unchanged by this follow-up.

resker:
Pushed `c6f0888e` addressing the review notes:

- **Clean-exit error formatting:** the early-exit path no longer wraps with `%w` directly — it returns a typed `daemonExitedError` that renders correctly even when the process exits with status 0 (no more `%!w(<nil>)`).
- **Retry scope:** `startDaemonWithRetry` now retries *only* the genuine early-exit race. A non-transient PID-file read error or a cancelled context is surfaced immediately instead of being retried five times.
- On the 10s-vs-30s note: that was a stale line in the PR description, not the code — the 30s timeout is intentional (a daemon still alive past it is a different failure than the startup race). Corrected the description.

resker:
@jandubois - would you mind taking another look?  I've reworked this in response to your review.

Also bumping this to re-trigger CI to try to clear the (erroneous) error.

resker:
Complements #5027 (merged) — that fixes the daemon crashing mid-session; this fixes the daemon failing to start in the first place.

AkihiroSuda:
@jandubois Was your concern resolved?

AkihiroSuda (approved):
Thanks

PR #5098, opened 2026-06-08
feat(screenshot): add limactl screenshot command for VZ GUI VMs

### LINKED ISSUE #5094: limactl screenshot: capture the display of a running VZ VM
In my quest to improve Lima for macOS software development use cases I'm proposing a screenshot function for VZ. Having an out-of-band screenshot function is super helpful for debugging the build of the VM and the software you are testing. Even if the VM is unresponsive or the network is broken, you can see what's happening in the running VM.

I have a working POC of this function, but I wanted to get an issue opened before submitting a PR.

When Lima runs a virtual machine using the VZ driver with a graphical display enabled, the VM's screen appears as a native macOS window. Right now, there's no way to capture that display from the command line — you have to manually take a screenshot with macOS tooling, which doesn't compose well with scripts or automation.

This proposes adding a `limactl screenshot <instance>` command that saves the current VM display to an image file on the host. The default output is a PNG named after the instance; the --output flag lets you specify a different path, and the file format is determined by the extension (`.png` or `.bmp`).

## Scope and limitations

This is a VZ-only, macOS-host feature. It requires the instance to have a graphical display configured (`video.display: vz` or `video.display: default`) and the GUI window must be on screen. QEMU instances and headless VZ instances are not supported — the command will return a clear error in those cases rather than silently failing. Linux and macOS guests are both supported, since the capture happens at the host window level regardless of what's running inside.

## Where changes land

The implementation touches four areas of the codebase, all additive:

- A new optional `Screenshotter` interface in the driver layer, following the same pattern as existing optional driver interfaces. Only the VZ driver implements it.
- A new `/v1/screenshot` endpoint on the hostagent's HTTP API, consistent with the existing `/v1/info` endpoint.
- A new `Screenshot` method on the hostagent client, giving callers a typed way to request a capture.
- A new `limactl screenshot` subcommand that wires these together from the user's perspective.

Note that I am using Commercial LLM tooling to assist with this work.

trodemaster:
This PR implments a screenshot command for the VZ driver. It sounds like adding a QEMU version could follow on. 

Adds `limactl screenshot <instance>` — captures the current display of
a running VZ VM and saves it to an image file on the host.

## Usage

```
limactl screenshot myvm                    # saves myvm-screenshot.png
limactl screenshot myvm --output foo.png   # custom path
limactl screenshot myvm --output foo.bmp   # BMP format
```

## Scope and limitations

VZ driver only, macOS host only. Requires `video.display: vz` or
`video.display: default` and the Lima.app GUI window must be on screen.
QEMU instances and headless VZ instances return a clear error rather
than failing silently.

## Implementation

- New optional `Screenshotter` interface in the driver layer — same
  pattern as existing optional interfaces. Only VZ implements it now;
  afbjorklund noted QEMU's `screendump` HMP command could follow once
  the interface is settled.
- New `GET /v1/screenshot` endpoint on the hostagent HTTP API,
  consistent with `/v1/info`.
- New `Screenshot` method on the hostagent client with typed HTTP
  error handling (404/501/422).
- New `limactl screenshot` subcommand (`//go:build darwin`).

Also fixes a pre-existing bug in `server.go` `onError`: `WriteHeader`
was called before `w.Header().Set(...)`, silently dropping the
Content-Type on error responses.

Closes #5094

Note: I am using Claude Code with Sonnet 4.6 (default) to assist with this work.

### DIFF
--- a/cmd/limactl/main.go
+++ b/cmd/limactl/main.go
@@ -199,6 +199,7 @@ func newApp() *cobra.Command {
 		newGenDocCommand(),
 		newGenSchemaCommand(),
 		newSnapshotCommand(),
+		newScreenshotCommand(),
 		newProtectCommand(),
 		newUnprotectCommand(),
 		newTunnelCommand(),
--- a/cmd/limactl/screenshot.go
+++ b/cmd/limactl/screenshot.go
@@ -1,0 +1,76 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	"github.com/lima-vm/lima/v2/pkg/hostagent/api/client"
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
+	"github.com/lima-vm/lima/v2/pkg/store"
+)
+
+func newScreenshotCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "screenshot INSTANCE",
+		Short:   "Capture a screenshot of the VM display",
+		Args:    WrapArgsError(cobra.ExactArgs(1)),
+		RunE:    screenshotAction,
+		GroupID: advancedCommand,
+	}
+	cmd.Flags().StringP("output", "o", "", "Output path; extension must be .png or .bmp (default: INSTANCE-screenshot.png)")
+	return cmd
+}
+
+func screenshotAction(cmd *cobra.Command, args []string) error {
+	instName := args[0]
+
+	ctx := cmd.Context()
+	inst, err := store.Inspect(ctx, instName)
+	if err != nil {
+		return err
+	}
+	if inst.Status != limatype.StatusRunning {
+		return fmt.Errorf("instance %q is not running (status: %s)", instName, inst.Status)
+	}
+
+	outputPath, _ := cmd.Flags().GetString("output")
+	if outputPath == "" {
+		outputPath = instName + "-screenshot.png"
+	}
+
+	var format string
+	switch strings.ToLower(filepath.Ext(outputPath)) {
+	case ".png":
+		format = "png"
+	case ".bmp":
+		format = "bmp"
+	default:
+		return fmt.Errorf("unsupported output extension %q: must be .png or .bmp", filepath.Ext(outputPath))
+	}
+
+	haSock := filepath.Join(inst.Dir, filenames.HostAgentSock)
+	haClient, err := client.NewHostAgentClient(haSock)
+	if err != nil {
+		return fmt.Errorf("failed to connect to hostagent: %w", err)
+	}
+
+	data, err := haClient.Screenshot(ctx, format)
+	if err != nil {
+		return fmt.Errorf("screenshot failed: %w", err)
+	}
+
+	if err := os.WriteFile(outputPath, data, 0o644); err != nil {
+		return fmt.Errorf("failed to write %q: %w", outputPath, err)
+	}
+	logrus.Infof("Screenshot saved to %q (%d bytes)", outputPath, len(data))
+	return nil
+}
--- a/pkg/driver/vz/screenshot_darwin.go
+++ b/pkg/driver/vz/screenshot_darwin.go
@@ -1,0 +1,112 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build darwin && !no_vz
+
+package vz
+
+/*
+#cgo darwin CFLAGS: -mmacosx-version-min=11 -x objective-c -fno-objc-arc
+#cgo darwin LDFLAGS: -lobjc -framework Foundation -framework AppKit -framework CoreGraphics -framework ImageIO
+#include <stdlib.h>
+#include <string.h>
+#include <dispatch/dispatch.h>
+#include <AppKit/AppKit.h>
+#include <CoreGraphics/CoreGraphics.h>
+#include <ImageIO/ImageIO.h>
+
+static void freeCString(char *s) { free(s); }
+
+// captureWindowImageBytes captures the frontmost non-panel on-screen window of
+// the current process, encodes it using the given ImageIO UTI, and returns a
+// malloc'd buffer. The caller must free() the returned pointer.
+static void *captureWindowImageBytes(int *outLen, const char *uti) {
+    __block void *result = NULL;
+    __block int resultLen = 0;
+
+    NSString *utiStr = [NSString stringWithUTF8String:uti];
+
+    // dispatch_sync to the main queue is required for window server access.
+    // CaptureScreenshot is always called from a Go goroutine, which is never
+    // the main thread, so no deadlock is possible in normal Lima operation.
+    // The isMainThread guard is a safety net for unexpected call paths.
+    void (^captureBlock)(void) = ^{
+        NSWindow *target = nil;
+        for (NSWindow *w in [NSApplication sharedApplication].windows) {
+            if ([w isKindOfClass:[NSPanel class]]) continue;
+            if (!w.isVisible) continue;
+            target = w;
+            break;
+        }
+        if (!target) return;
+
+        CGWindowID wid = (CGWindowID)[target windowNumber];
+        CGImageRef img = CGWindowListCreateImage(
+            CGRectNull,
+            kCGWindowListOptionIncludingWindow,
+            wid,
+            kCGWindowImageBoundsIgnoreFraming | kCGWindowImageShouldBeOpaque
+        );
+        if (!img) return;
+
+        NSMutableData *data = [NSMutableData data];
+        CGImageDestinationRef dst = CGImageDestinationCreateWithData(
+            (__bridge CFMutableDataRef)data,
+            (__bridge CFStringRef)utiStr,
+            1, NULL
+        );
+        if (!dst) {
+            CGImageRelease(img);
+            return;
+        }
+        CGImageDestinationAddImage(dst, img, NULL);
+        bool ok = CGImageDestinationFinalize(dst);
+        CFRelease(dst);
+        CGImageRelease(img);
+        if (!ok) return;
+
+        resultLen = (int)data.length;
+        result = malloc(resultLen);
+        memcpy(result, data.bytes, resultLen);
+    };
+
+    if ([NSThread isMainThread]) {
+        captureBlock();
+    } else {
+        dispatch_sync(dispatch_get_main_queue(), captureBlock);
+    }
+
+    *outLen = resultLen;
+    return result;
+}
+*/
+import "C"
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/lima-vm/lima/v2/pkg/driver"
+)
+
+// CaptureScreenshot captures the VM display window.
+// format is "png" or "bmp"; anything else defaults to PNG.
+// Implements driver.Screenshotter. Requires the GUI app bundle to be running.
+func (l *LimaVzDriver) CaptureScreenshot(format string) ([]byte, error) {
+	if !l.canRunGUI() {
+		return nil, fmt.Errorf("%w (set video.display to \"default\" or \"vz\" to enable screenshots)", driver.ErrNoDisplay)
+	}
+	uti := "public.png"
+	if format == "bmp" {
+		uti = "com.microsoft.bmp"
+	}
+	cuti := C.CString(uti)
+	defer C.freeCString(cuti)
+	var outLen C.int
+	ptr := C.captureWindowImageBytes(&outLen, cuti)
+	if ptr == nil || outLen == 0 {
+		return nil, errors.New("screenshot capture returned no data; GUI window may not be visible")
+	}
+	defer C.free(ptr)
+	return C.GoBytes(ptr, outLen), nil
+}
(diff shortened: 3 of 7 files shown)


pkg/hostagent/api/server/server.go
```diff
@@ -53,6 +55,43 @@ func (b *Backend) GetInfo(w http.ResponseWriter, r *http.Request) {
...
+	format := r.URL.Query().Get("format")
+	if format == "" {
+		format = "png"
+	}
+	data, err := b.Agent.Screenshot(ctx, format)
+	if err != nil {
+		ec := http.StatusInternalServerError
+		msg := err.Error()
+		switch {
+		case strings.Contains(msg, "does not support screenshots"):
```
AkihiroSuda: Use `errors.Is` or `errors.As`

pkg/hostagent/api/client/client.go
```diff
@@ -65,3 +69,27 @@ func (c *client) Info(ctx context.Context) (*api.Info, error) {
...
+	if err != nil {
+		var httpErr *httpclientutil.HTTPStatusError
+		if errors.As(err, &httpErr) {
+			switch httpErr.StatusCode {
+			case http.StatusNotFound:
+				return nil, errors.New("hostagent does not support screenshot — restart the instance to pick up the latest hostagent binary")
+			case http.StatusNotImplemented:
+				return nil, errors.New("this VM's driver does not support screenshots (only VZ VMs with a display are supported)")
+			case http.StatusUnprocessableEntity:
+				return nil, errors.New("VM has no display configured — set video.display to \"default\" or \"vz\" in the instance config")
```
AkihiroSuda:
```suggestion
				return nil, errors.New("VM has no display configured — set video.display (e.g., \"default\") in the instance config")
```

pkg/hostagent/api/client/client.go
```diff
@@ -65,3 +69,27 @@ func (c *client) Info(ctx context.Context) (*api.Info, error) {
...
+	u := fmt.Sprintf("http://%s/%s/screenshot?format=%s", c.dummyHost, c.version, format)
+	resp, err := httpclientutil.Get(ctx, c.HTTPClient(), u)
+	if err != nil {
+		var httpErr *httpclientutil.HTTPStatusError
+		if errors.As(err, &httpErr) {
+			switch httpErr.StatusCode {
+			case http.StatusNotFound:
+				return nil, errors.New("hostagent does not support screenshot — restart the instance to pick up the latest hostagent binary")
+			case http.StatusNotImplemented:
+				return nil, errors.New("this VM's driver does not support screenshots (only VZ VMs with a display are supported)")
```
AkihiroSuda:
```suggestion
				return nil, fmt.Errorf("this VM's driver (%q) does not support screenshots", driverName)
```

AkihiroSuda: "(only VZ VMs with a display are supported)" should be omitted from here as it is likely going to be implemented in other drivers too

cmd/limactl/screenshot.go
```diff
@@ -0,0 +1,73 @@
...
+
+func newScreenshotCommand() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "screenshot INSTANCE",
+		Short:   "Capture a screenshot of the VM display",
+		Args:    WrapArgsError(cobra.ExactArgs(1)),
+		RunE:    screenshotAction,
+		GroupID: advancedCommand,
+	}
+	cmd.Flags().StringP("output", "o", "", "Output path; extension sets format: .png (default) or .bmp")
```
AkihiroSuda: Why not always require an extension?

cmd/limactl/screenshot.go
```diff
@@ -0,0 +1,73 @@
...
+		return fmt.Errorf("instance %q is not running (status: %s)", instName, inst.Status)
+	}
+
+	outputPath, _ := cmd.Flags().GetString("output")
+
+	// Infer format from output extension; default to PNG.
+	format := "png"
+	if strings.ToLower(filepath.Ext(outputPath)) == ".bmp" {
+		format = "bmp"
+	}
```
AkihiroSuda: Error out if the extension is neither `.png` or `.bmp`

trodemaster:
All suggestions have been incorporated. Thanks!

cmd/limactl/main.go
```diff
@@ -199,6 +199,7 @@ func newApp() *cobra.Command {
 		newGenDocCommand(),
 		newGenSchemaCommand(),
 		newSnapshotCommand(),
+		newScreenshotCommand(),
 		newProtectCommand(),
```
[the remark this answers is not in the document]
trodemaster: I think we wanted to have the screenshot command itself not be Darwin specific with a potential QEMU implementation suggested.

pkg/hostagent/api/server/server.go
trodemaster: Done — `GetScreenshot` uses `errors.Is` against the sentinel errors in `driver.go`. Let me know if there's a specific case I missed.

pkg/hostagent/api/client/client.go
trodemaster: Updated to match your suggestion exactly.

pkg/hostagent/api/client/client.go
trodemaster: Removed — the 501 message no longer mentions VZ specifically.

cmd/limactl/screenshot.go
trodemaster: Done — the `switch` on `filepath.Ext` falls through to a `default` that returns an error for anything other than `.png` or `.bmp`.

pkg/hostagent/api/client/client.go
trodemaster: Good point. The hostagent already wraps the driver name into the error (`fmt.Errorf("driver %q: %w", name, ErrDriverNotScreenshotter)`), so on a 501 we now surface the parsed JSON body message directly rather than hardcoding a generic string. The client's `httpErr.Error()` method already extracts `ErrorJSON.Message` from the body, so we just return `httpErr` rather than a new `errors.New`.

cmd/limactl/screenshot.go
trodemaster: The default output path is intentional — `limactl screenshot myvm` works without requiring `--output`, which we think is a better UX for a simple capture command. The default always carries a `.png` extension, and any user-supplied `--output` with an unrecognized extension errors out (addressed in the AS-6 thread), so there is no ambiguity about the inferred format.

pkg/driver/vz/screenshot_darwin.go
```diff
@@ -0,0 +1,94 @@
...
+
+package vz
+
+/*
+#cgo darwin CFLAGS: -mmacosx-version-min=11 -x objective-c -fno-objc-arc
+#cgo darwin LDFLAGS: -lobjc -framework Foundation -framework AppKit -framework CoreGraphics -framework ImageIO
+#include <stdlib.h>
+#include <AppKit/AppKit.h>
+#include <CoreGraphics/CoreGraphics.h>
+#include <ImageIO/ImageIO.h>
```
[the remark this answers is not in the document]
trodemaster: Fixed — added explicit `#include <string.h>` and `#include <dispatch/dispatch.h>` to the CGo preamble. On macOS these come in transitively via AppKit, but the explicit includes make the dependencies clear and guard against future toolchain changes.

pkg/driver/vz/screenshot_darwin.go
```diff
@@ -0,0 +1,94 @@
...
+        NSMutableData *data = [NSMutableData data];
+        CGImageDestinationRef dst = CGImageDestinationCreateWithData(
+            (__bridge CFMutableDataRef)data,
+            (__bridge CFStringRef)utiStr,
+            1, NULL
+        );
+        CGImageDestinationAddImage(dst, img, NULL);
+        CGImageDestinationFinalize(dst);
+        CFRelease(dst);
+        CGImageRelease(img);
```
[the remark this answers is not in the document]
trodemaster: Fixed — `CGImageDestinationCreateWithData` is now NULL-checked; on failure we release the `CGImageRef` and return early (the Go side surfaces this as a "no data" error). We also check the return value of `CGImageDestinationFinalize` and return early without copying on failure.

pkg/hostagent/api/client/client.go
```diff
@@ -65,3 +69,27 @@ func (c *client) Info(ctx context.Context) (*api.Info, error) {
 	}
 	return &info, nil
 }
+
+func (c *client) Screenshot(ctx context.Context, format string) ([]byte, error) {
+	if format == "" {
+		format = "png"
+	}
+	u := fmt.Sprintf("http://%s/%s/screenshot?format=%s", c.dummyHost, c.version, format)
+	resp, err := httpclientutil.Get(ctx, c.HTTPClient(), u)
```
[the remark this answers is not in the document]
trodemaster: Fixed — `url.Values` encoding is now used for the query string, and `Screenshot()` validates `format` before building the URL. In the current flow `format` is always `"png"` or `"bmp"` (set by the CLI from the file extension), but the validation protects any future callers.

pkg/driver/vz/screenshot_darwin.go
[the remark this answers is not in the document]
trodemaster: Fixed — added `<string.h>`.

pkg/driver/vz/screenshot_darwin.go
```diff
@@ -0,0 +1,94 @@
...
+
+        NSMutableData *data = [NSMutableData data];
+        CGImageDestinationRef dst = CGImageDestinationCreateWithData(
+            (__bridge CFMutableDataRef)data,
+            (__bridge CFStringRef)utiStr,
+            1, NULL
+        );
+        CGImageDestinationAddImage(dst, img, NULL);
+        CGImageDestinationFinalize(dst);
+        CFRelease(dst);
```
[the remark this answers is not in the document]
trodemaster: Fixed — added `<string.h>`.

pkg/driver/vz/screenshot_darwin.go
```diff
@@ -0,0 +1,94 @@
...
+// the current process, encodes it using the given ImageIO UTI, and returns a
+// malloc'd buffer. The caller must free() the returned pointer.
+static void *captureWindowImageBytes(int *outLen, const char *uti) {
+    __block void *result = NULL;
+    __block int resultLen = 0;
+
+    NSString *utiStr = [NSString stringWithUTF8String:uti];
+    dispatch_sync(dispatch_get_main_queue(), ^{
+        NSWindow *target = nil;
+        for (NSWindow *w in [NSApplication sharedApplication].windows) {
```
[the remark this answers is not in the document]
trodemaster: Added an `[NSThread isMainThread]` guard: if already on the main thread the block runs inline; otherwise `dispatch_sync` is used. In normal Lima operation this branch is never taken (Go goroutines don't own the macOS main thread), but the guard documents the invariant and prevents deadlock in any unexpected call path.

pkg/hostagent/api/server/server.go
```diff
@@ -53,6 +56,42 @@ func (b *Backend) GetInfo(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	ctx := r.Context()
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	format := r.URL.Query().Get("format")
+	if format == "" {
+		format = "png"
+	}
```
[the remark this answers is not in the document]
trodemaster: Fixed — `format` is now lowercased via `strings.ToLower` and validated against `"png"` and `"bmp"` before being passed to the driver. An unrecognized value returns 400.

pkg/hostagent/api/server/server.go
```diff
@@ -53,6 +56,42 @@ func (b *Backend) GetInfo(w http.ResponseWriter, r *http.Request) {
...
+		case errors.Is(err, driver.ErrNoDisplay):
+			ec = http.StatusUnprocessableEntity
+		}
+		b.onError(w, err, ec)
+		return
+	}
+	ct := "image/png"
+	if format == "bmp" {
+		ct = "image/bmp"
+	}
```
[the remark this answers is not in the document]
trodemaster: Fixed — `format` is now lowercased and validated server-side; the content-type check is consistent with the validated value.

pkg/hostagent/api/client/client.go
```diff
@@ -65,3 +69,32 @@ func (c *client) Info(ctx context.Context) (*api.Info, error) {
...
+			case http.StatusNotImplemented:
+				return nil, errors.New("this VM's driver does not support screenshots")
+			case http.StatusUnprocessableEntity:
+				return nil, errors.New("VM has no display configured — set video.display (e.g., \"default\") in the instance config")
+			}
+		}
+		return nil, err
+	}
+	defer resp.Body.Close()
+	return io.ReadAll(resp.Body)
```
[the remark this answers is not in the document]
trodemaster: Fixed — `io.LimitReader` now caps the read at 32 MB. If the response exceeds that, an error is returned rather than consuming unbounded memory.

pkg/hostagent/api/server/server.go
```diff
@@ -53,6 +56,42 @@ func (b *Backend) GetInfo(w http.ResponseWriter, r *http.Request) {
 	_, _ = w.Write(m)
 }
 
+// GetScreenshot is the handler for GET /v1/screenshot.
+func (b *Backend) GetScreenshot(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodGet {
+		w.WriteHeader(http.StatusMethodNotAllowed)
+		return
+	}
```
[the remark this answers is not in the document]
trodemaster: Fixed — `w.Header().Set("Allow", "GET")` is now set before `WriteHeader(405)`.

AkihiroSuda (approved):
Thanks

PR #4866, opened 2026-04-22
Support attaching `/dev/diskN` block devices on macOS VZ hosts

pirate:
> [!IMPORTANT]
> Superseeded by https://github.com/lima-vm/lima/pull/5117

---
---

Fixes https://github.com/lima-vm/lima/discussions/1314 & https://github.com/lima-vm/lima/issues/2224

📖 Wrote a blog post based on my learnings too: https://docs.sweeting.me/s/sockets-101

## Summary

This adds VZ-only support on macOS for attaching host block devices to Lima VMs.

A VM can now be launched with:

```bash
limactl start --block-device=/dev/disk4 ...
```

It can also be configured in YAML with the top-level `blockDevices` field.

The attached device is exposed to the guest as a virtio block device, and Lima sets a deterministic block device identifier from the host basename. For example, `/dev/disk4` appears in the guest under `/dev/disk/by-id/virtio-disk4`.

## Implementation

- added the `--block-device` CLI flag and top-level `blockDevices` config support
- attached host block devices in the VZ driver with [`VZDiskBlockDeviceStorageDeviceAttachment`](https://developer.apple.com/documentation/virtualization/vzdiskblockdevicestoragedeviceattachment)
- kept block-device-specific VZ runtime logic in `pkg/driver/vz/block_device_darwin.go`
- added a hidden `limactl sudo-open-block-device` helper that opens a host block device and passes the file descriptor back to the unprivileged VM process
- kept the helper path backend-neutral in `pkg/blockdevice`
- moved generic sudo execution and sudoers rendering helpers into `pkg/sudoers`
- kept `pkg/networks` network-specific by limiting it to network sudoers entries and network validation
- updated `limactl sudoers` to compose the shared `/etc/sudoers.d/lima` file from network entries plus the block-device helper entry
- documented top-level `blockDevices` in `templates/default.yaml` and the disk docs
- non-VZ backends reject `blockDevices` explicitly for now

## Layout Notes

The ownership boundaries are now:

- `pkg/networks`: network config, validation, and network-specific sudoers fragments
- `pkg/sudoers`: generic sudo invocation and sudoers file helpers
- `pkg/blockdevice`: generic block-device helper, request validation, fd handoff, and sudoers fragment
- `pkg/driver/vz/block_device_darwin.go`: VZ block-device attachment logic
- `cmd/limactl sudoers`: existing CLI command, combines network sudoers entries with blockdevice entries to create the final `/etc/sudoers.d/lima` file
[...]

PR #5104, opened 2026-06-10
blockdevice: support qemu and krunkit backends on all host platforms

pirate:
Follow-up to #4866, which added `blockDevices` / `--block-device` for VZ on macOS.

## Summary

This extends host block device attachment to the qemu and krunkit backends and to every host OS, with **no new config surface** and **no mandatory sudoers setup**. The same CLI flag and YAML field now work everywhere:

```bash
limactl start --vm-type=qemu --block-device=/dev/sdb template:default   # Linux/BSD
limactl start --vm-type=qemu --block-device=/dev/disk4 template:default # macOS
limactl start --vm-type=krunkit --block-device=/dev/disk4 ...           # macOS arm64
limactl start --vm-type=qemu --block-device=\\.\PhysicalDrive2 ...      # Windows (elevated)
```

| Host | vmType | Mechanism |
| ---- | ------ | --------- |
| macOS | `vz` | fd handoff, unchanged (#4866) |
| macOS | `qemu`, `krunkit` | device node ownership grant via the helper, then the backend opens the path |
| Linux, BSDs | `qemu` | fd handoff via `-add-fd` / `/dev/fdset/N` + `host_device` |
| Windows | `qemu` | direct `\\.\PhysicalDriveN` path passthrough (elevated process) |
| Windows | `wsl2` | still rejected (`wsl --mount` is global to all distros), error points at qemu |

## Zero-setup device access

`pkg/blockdevice` now builds for all Unix hosts and resolves access in order:

1. **Direct open** with the user's own permissions (user-owned ramdisks, Linux `disk` group) — no escalation at all.
2. **Passwordless helper** — `limactl start` probes the privileged helper with a no-op request to check that a NOPASSWD rule or cached sudo timestamp covers it. (`sudo --list` is not usable for this: it answers *permitted*, not *passwordless*.)
3. **One password prompt** — otherwise `limactl start` runs `sudo --validate` in the foreground; the cached timestamp lets the background hostagent run the helper non-interactively moments later. The hostagent cannot prompt itself: it runs in a background process group where reading the terminal raises SIGTTIN.

`limactl start` then opens each device once through the real path, so device-level failures (e.g. SIP-protected disks that even root cannot open) fail fast with a clear foreground error instead of surfacing mid-boot.

`limactl sudoers` now also works on Linux/BSD hosts (emitting only the block-device helper entry, scoped to the current user, since the admin group name varies by distro) for opting out of the prompt in headless/autostart setups.

## Why macOS qemu/krunkit use an ownership grant instead of fd passing

Both fd-passing routes are dead ends on macOS for an external VM process, verified empirically (the second one in CI with a root-owned device, which local user-owned test devices cannot catch):

- `fcntl(F_SETFL)` fails with ENOTTY on macOS disk descriptors (block *and* raw char), and QEMU's fd-set duplication (`qemu_dup_flags`) requires it — so `-add-fd`//`/dev/fdset/N` can never work for disks on macOS.
- Opening `/dev/fd/N` on macOS re-checks the permissions of the underlying vnode (it is not plain `dup(2)` security-wise), so an inherited root-opened descriptor still yields EPERM for the unprivileged backend. krunkit's virtio-blk additionally has no `fd=` option at all (only `path=`/`format=`).

So for these two backends the helper does the one thing that does work and that an admin would do by hand: `chown` the device node to the invoking user (uid taken from sudo's own `SUDO_UID`, never from the request), after which the backend opens the path itself. devfs ownership reverts on detach/reboot. VZ keeps the strictly-cleaner fd handoff since it attaches in-process.

Also fixed en route: `filepath.IsAbs` does not recognize Windows DOS device paths (`\\.\PhysicalDrive2`), so YAML validation rejected them (caught by the Windows CI run).

## Design notes

Native provider features are used wherever they exist: qemu's fd sets, `host_device` driver, `serial=`, `auto-read-only`; sudo's NOPASSWD rules, credential timestamps, and `--validate`; plain `chown` for the macOS access grant. The only custom machinery is the small open-as-root + SCM_RIGHTS helper from #4866 (no provider offers it; libvirt implements the same pattern), now with the additional ownership-grant and no-op probe request types.

Per-OS variants always live in dedicated files selected by build tags, so the dispatch point is the build system itself: `pkg/blockdevice` owns the helper protocol, access probing, and the per-OS attach strategies (`block_device_unix.go` / `block_device_windows.go`, `qemu_darwin.go` / `qemu_fdset.go` / `qemu_windows.go`); `pkg/sudoers` owns generic sudo mechanics; backends contain no per-OS branching and just splice in `blockdevice.QEMUDriveArgs(...)` / call `blockdevice.EnsureDeviceAccessible(...)`. `cmd/limactl` is a thin cross-platform shell with no per-OS stub files.

## Test coverage (all real end-to-end, no mocks, no new test patterns)

A new `CHECKS["block-device"]` in `hack/test-templates.sh` — next to the existing checks, enabled for `default.yaml` and `debian.yaml` — creates a real host block device per OS (root-owned ramdisk via `hdiutil` on macOS, root-owned loop device via `losetup` on Linux, VHD-backed `\\.\PhysicalDriveN` via `diskpart` on Windows), boots with `--block-device`, writes a marker from the guest, and reads it back from the raw device on the host. Because the existing CI jobs already run `test-templates.sh` with these templates, **every existing integration job picks the check up with zero workflow changes**:

- macOS vz (`vz` job, root-owned device → helper fd handoff) — plus the unchanged `vz-block-device.bats` deep mount-roundtrip from #4866
- macOS qemu (`qemu` job, root-owned device → helper ownership grant)
- Linux qemu (`qemu-linux` job `debian.yaml` entry, root-owned loop device → helper fd handoff via fd sets)
- Windows qemu (`windows-qemu` job → diskpart VHD → direct path)

krunkit cannot run on hosted CI (GitHub's arm64 macOS runners lack nested virtualization — the same reason the vz job uses Intel runners, and krunkit is Apple-Silicon-only); the same check runs against it locally via `LIMACTL_CREATE_ARGS=--vm-type=krunkit`, validated on real hardware along with vz/qemu ramdisk roundtrips, the passwordless-helper probe against a real NOPASSWD rule, the helper's chown op, and the fast-fail paths for SIP-protected devices and missing terminals. Unit tests, golangci-lint, and cross-compiles for linux/windows/netbsd/dragonfly/freebsd all pass.


pirate:
@AkihiroSuda I'm more out of my depth with this PR than the last one, as I don't know much about Windows or `krunkit`, I've only used QEMU/vz personally.

Feel free to tear it apart / mark it as draft if you think it needs a lot more work.

AkihiroSuda:
>  I'm more out of my depth with this PR than the last one, as I don't know much about Windows or krunkit, I've only used QEMU/vz personally.

Please do not add untested implementation.
Notably this needs more consideration about the security design:
- https://github.com/lima-vm/lima/issues/5112

hack/test-templates.sh
AkihiroSuda, line 1: New tests should be implemented as BATS

Issue #4593, opened 2026-02-16
`/mnt/lima-cidata` should not need `mode=0700,dmode=0700,overriderockperm,uid=0`


AkihiroSuda:
https://github.com/lima-vm/lima/blob/d7a96ebfd9d8c98ddab4181faa02f3ab28ecba59/pkg/cidata/cidata.TEMPLATE.d/user-data#L50-L53
(Introduced in https://github.com/lima-vm/lima/pull/68/changes/8c75ebd16331d575618b6306f6f7935d8df08411)

Why do we have `mode=0700,dmode=0700,overriderockperm,uid=0` here?
Is there anything we have to hide from non-root?

AkihiroSuda:
Claude helped me remind the reason


>   The restrictions exist for a plausible reason: param.env and etc_environment may contain user credentials (proxy passwords, user-defined params used as secrets). However, the approach is a blunt instrument — it restricts everything including the
>   harmless files just to protect the potentially sensitive ones.
> 
>   A cleaner fix would be to split the ISO contents:
>   - Keep sensitive files (param.env, etc_environment, potentially provision scripts) at 0600 root-only
>   - Relax permissions on the rest (binaries, boot scripts, public metadata) to 0644/0755
> 
>   Or alternatively, document/enforce that param values and environment variables should not contain secrets (and offer a separate secret mechanism), then safely relax all permissions.
> 
>   The overriderockperm + uid=0 combo is also unusual — it aggressively overrides whatever permissions Rock Ridge extensions encode into the ISO. If the ISO were built with correct permissions baked in (via the iso9660util writer), the mount options might
>    not need to override them at all.

Mysterio-17:
Hello @AkihiroSuda , can i work on this issue.

AkihiroSuda:
> Hello [@AkihiroSuda](https://github.com/AkihiroSuda) , can i work on this issue.

Yes, but this one needs careful security review and may not be merged soon

Mysterio-17:
Sure @AkihiroSuda , the blanket `0700 + overriderockperm` makes sense for `param.env`/`etc_environment` which can hold credentials, but it unnecessarily locks down harmless files too... `boot.sh` already has a comment acknowledging this friction. I'll look into setting per-file permissions at ISO creation time and share a proposal before writing any code. Happy to take this slow.

PR #4953, opened 2026-05-11
test(executil): add unit tests for WithContext option

varsha-0007:
Currently, pkg/executil has no test coverage. This PR adds unit tests
for the WithContext option defined in command.go.

Tests added:
- TestWithContext: verifies that WithContext correctly assigns the
  context to the options struct and returns no error
- TestWithContext_NilByDefault: verifies that the default context in
  options is nil before any option is applied

Signed-off-by: varsha-0007, varsha.narvi.07@gmail.com

PR #5109, opened 2026-06-12
feat: allow opting out of passwordless sudo

### LINKED ISSUE #4594: Allow opting out from `sudo: ALL=(ALL) NOPASSWD:ALL` in the VM
https://github.com/lima-vm/lima/blob/d7a96ebfd9d8c98ddab4181faa02f3ab28ecba59/pkg/cidata/cidata.TEMPLATE.d/user-data#L38

`sudo: ALL=(ALL) NOPASSWD:ALL` is quite common for cloud instances, but its bad practice does not necessarily need to be replicated for Lima.

There should be a configuration to disable NOPASSWD sudo and generate a random password for it

~Needs:~
~- https://github.com/lima-vm/lima/issues/4593~

Aneesh-Hegde:
Fixes #4594 

Adds `user.passwordlessSudo` (default `true`, unchanged behavior). When set to `false`, the guest user gets `sudo: ALL=(ALL) ALL` with `lock_passwd: false` instead of `NOPASSWD:ALL`. A random password is generated on first boot via `boot.sh` using `/dev/urandom`, applied via `chpasswd`, and saved to `~/password` in the guest (`0600`).

 This PR is the final piece of a series of prerequisite PRs needed to safely remove `NOPASSWD:ALL`:
 - #5135 — drop `sudo grep` for `fuse*.conf` check (notification file approach)
 - #5147  — drop sudo for non-privileged files in `copyToHost`
 - #5157   — move `/run/host-services` `mkdir`/`chown` to boot script

 **Design notes:**
 - Password is stored guest-side at `~/password`, consistent with macOS guest convention in `macos-26.yaml`
 - Nothing password-related is written to the cidata ISO, avoiding the plaintext visibility concern in #4593
 - `plain: true` is required when `passwordlessSudo: false` as an interim constraint per [GHSA-2j9v-p4xj-cjw2](https://github.com/lima-vm/lima/security/advisories/GHSA-2j9v-p4xj-cjw2), until guestagent's Unix socket tunneling refusal is implemented

 **Known limitations / follow-ups:**
 - [ ] Only applies to cloud-init provisioned guests (QEMU/vz). WSL2 instances ignore `passwordlessSudo`.
 - [ ] `param.env` sudo dependency pending #4593 direction
 - [ ] Proper guestagent Unix socket tunneling refusal when `passwordlessSudo: false` (currently enforced via `plain: true`)
 - [ ] No CLI to retrieve the password, users must `cat ~/password` from inside the guest
 - [ ] Security Constraint: Enforces plain: true when passwordlessSudo: false is set. This mitigates GHSA-2j9v-p4xj-cjw2 by disabling the guestagent (and its Unix socket tunneling) until the guestagent is refactored to refuse socket tunneling safely.

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.sh
@@ -203,6 +203,41 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.user ]; then
 	done
 fi
 
+if [ "${UNAME}" != "Darwin" ] && [ "${LIMA_CIDATA_PASSWORDLESS_SUDO}" != "1" ]; then
+	if [ ! -f "${LIMA_CIDATA_HOME}/.lima-password-set" ]; then
+		PASSWORD=$(head -c 8 /dev/urandom | od -An -tx1 | tr -d ' \n')
+		if command -v chpasswd >/dev/null 2>&1; then
+			if ! echo "${LIMA_CIDATA_USER}:${PASSWORD}" | chpasswd; then
+				WARNING "failed to set password via chpasswd"
+				PASSWORD=""
+				CODE=1
+			fi
+		elif command -v pw >/dev/null 2>&1; then
+			if ! echo "${PASSWORD}" | pw usermod "${LIMA_CIDATA_USER}" -h 0; then
+				WARNING "failed to set password via pw"
+				PASSWORD=""
+				CODE=1
+			fi
+		else
+			WARNING "no supported tool to set password found"
+			PASSWORD=""
+			CODE=1
+		fi
+		if [ -n "${PASSWORD}" ]; then
+			(umask 077 && : >"${LIMA_CIDATA_HOME}/password")
+			chown "${LIMA_CIDATA_USER}:${LIMA_CIDATA_USER}" "${LIMA_CIDATA_HOME}/password" || {
+				WARNING "failed to chown ${LIMA_CIDATA_HOME}/password"
+				CODE=1
+			}
+			echo "${PASSWORD}" >"${LIMA_CIDATA_HOME}/password"
+			touch "${LIMA_CIDATA_HOME}/.lima-password-set"
+			chown "${LIMA_CIDATA_USER}:${LIMA_CIDATA_USER}" "${LIMA_CIDATA_HOME}/.lima-password-set" || {
+				WARNING "failed to chown ${LIMA_CIDATA_HOME}/.lima-password-set"
+				CODE=1
+			}
+		fi
+	fi
+fi
 # Signal that provisioning is done. The instance ID changes on every boot,
 # so any value from a previous boot cycle will be different.
 echo "${LIMA_CIDATA_IID}" >"${RUN}/lima-boot-done"
--- a/pkg/cidata/cidata.TEMPLATE.d/user-data
+++ b/pkg/cidata/cidata.TEMPLATE.d/user-data
@@ -55,6 +55,7 @@ users:
     {{/* (Why doesn't macOS VM support graceful shutdown?) */}}
     sudo: ALL=(ALL) NOPASSWD:/sbin/shutdown -h now
 {{- else }}
+    {{- if .PasswordlessSudo }}
     sudo: ALL=(ALL) NOPASSWD:ALL
     {{- if eq .OS "FreeBSD" }}
     groups:
@@ -62,6 +63,15 @@ users:
     doas: permit nopass :wheel
     {{- end}}
     lock_passwd: true
+    {{- else }}
+    sudo: ALL=(ALL) ALL
+    {{- if eq .OS "FreeBSD" }}
+    groups:
+    - wheel
+    doas: permit :wheel
+    {{- end}}
+    lock_passwd: false
+    {{- end }}
 {{- end }}
 {{- if eq .OS "FreeBSD" }}
     ssh_authorized_keys:
--- a/pkg/cidata/template.go
+++ b/pkg/cidata/template.go
@@ -100,6 +100,7 @@ type TemplateArgs struct {
 	Home                            string // home directory
 	Shell                           string // login shell
 	UID                             uint32
+	PasswordlessSudo                bool
 	SSHPubKeys                      []string
 	Mounts                          []Mount
 	MountType                       string
--- a/pkg/limatype/lima_yaml.go
+++ b/pkg/limatype/lima_yaml.go
@@ -109,11 +109,12 @@ var (
 )
 
 type User struct {
-	Name    *string `yaml:"name,omitempty" json:"name,omitempty" jsonschema:"nullable"`
-	Comment *string `yaml:"comment,omitempty" json:"comment,omitempty" jsonschema:"nullable"`
-	Home    *string `yaml:"home,omitempty" json:"home,omitempty" jsonschema:"nullable"`
-	Shell   *string `yaml:"shell,omitempty" json:"shell,omitempty" jsonschema:"nullable"`
-	UID     *uint32 `yaml:"uid,omitempty" json:"uid,omitempty" jsonschema:"nullable"`
+	Name             *string `yaml:"name,omitempty" json:"name,omitempty" jsonschema:"nullable"`
+	Comment          *string `yaml:"comment,omitempty" json:"comment,omitempty" jsonschema:"nullable"`
+	Home             *string `yaml:"home,omitempty" json:"home,omitempty" jsonschema:"nullable"`
+	Shell            *string `yaml:"shell,omitempty" json:"shell,omitempty" jsonschema:"nullable"`
+	UID              *uint32 `yaml:"uid,omitempty" json:"uid,omitempty" jsonschema:"nullable"`
+	PasswordlessSudo *bool   `yaml:"passwordlessSudo,omitempty" json:"passwordlessSudo,omitempty" jsonschema:"nullable"`
 }
 
 type VMOpts map[VMType]any
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -171,6 +171,9 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 	if y.User.UID == nil {
 		y.User.UID = d.User.UID
 	}
+	if y.User.PasswordlessSudo == nil {
+		y.User.PasswordlessSudo = d.User.PasswordlessSudo
+	}
 	if o.User.Name != nil {
 		y.User.Name = o.User.Name
 	}
@@ -185,6 +188,9 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 	}
 	if o.User.UID != nil {
 		y.User.UID = o.User.UID
+	}
+	if o.User.PasswordlessSudo != nil {
+		y.User.PasswordlessSudo = o.User.PasswordlessSudo
 	}
 	if y.User.Name == nil {
 		y.User.Name = ptr.Of(osutil.LimaUser(ctx, existingLimaVersion, warn, y.OS).Username)
@@ -220,6 +226,15 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 			y.User.UID = ptr.Of(uint32(1000))
 		}
 		// warn = false
+	}
+
+	// fallback to nopasswd
+	if y.User.PasswordlessSudo == nil {
+		if y.OS != nil && *y.OS == limatype.DARWIN {
+			y.User.PasswordlessSudo = ptr.Of(false)
+		} else {
+			y.User.PasswordlessSudo = ptr.Of(true)
+		}
 	}
 	if out, err := executeGuestTemplate(*y.User.Home, instDir, y.User, y.Param); err == nil {
 		y.User.Home = ptr.Of(out.String())
--- a/pkg/limayaml/validate.go
+++ b/pkg/limayaml/validate.go
@@ -424,6 +424,25 @@ func Validate(y *limatype.LimaYAML, warn bool) error {
 			}
 		}
 	}
+
+	// passwordlessSudo: true is not supported on macOS guests yet (lock_passwd unimplemented,
+	// account creation requires a password, and the same account backs the GUI session)
+	if y.OS != nil && *y.OS == limatype.DARWIN &&
+		y.User.PasswordlessSudo != nil && *y.User.PasswordlessSudo {
+		errs = errors.Join(errs, errors.New("`user.passwordlessSudo: true` is not supported on macOS guests"))
+	}
+	if y.VMType != nil && *y.VMType == limatype.WSL2 &&
+		y.User.PasswordlessSudo != nil && !*y.User.PasswordlessSudo {
+		errs = errors.Join(errs, errors.New("`user.passwordlessSudo: false` is not supported on WSL2 guest"))
+	}
+	// GHSA-2j9v-p4xj-cjw2 constraint: require plain mode when passwordless sudo is disabled
+	if y.User.PasswordlessSudo != nil && !*y.User.PasswordlessSudo {
+		if y.OS == nil || *y.OS != limatype.DARWIN {
+			if y.Plain == nil || !*y.Plain {
+				errs = errors.Join(errs, errors.New("`user.passwordlessSudo: false` requires `plain: true`"))
+			}
+		}
+	}
 	if y.Plain != nil && *y.Plain {
 		const portRangeWarnThreshold = 10
 		for i, rule := range y.PortForwards {
@@ -662,6 +681,14 @@ func warnExperimental(y *limatype.LimaYAML) {
 	}
 	if y.MountInotify != nil && *y.MountInotify {
 		logrus.Warn("`mountInotify` is experimental")
+	}
+	if y.User.PasswordlessSudo != nil && !*y.User.PasswordlessSudo {
+		if y.OS == nil || *y.OS != limatype.DARWIN {
+			logrus.Warn("`user.passwordlessSudo: false` is experimental")
+		}
+	}
+	if y.User.PasswordlessSudo != nil && !*y.User.PasswordlessSudo && len(y.Param) > 0 {
+		logrus.Warn("`param` still relies on sudo internally, it may not work as expected when `user.passwordlessSudo: false` is set")
 	}
 }
 
--- a/templates/default.yaml
+++ b/templates/default.yaml
@@ -374,6 +374,12 @@ user:
   # The shell must already exist in the guest image.
   # 🟢 Builtin default: "/bin/bash" (Linux), "/bin/zsh" (macOS), "/bin/sh" (FreeBSD), "cmd.exe" (Windows)
   shell: null
+  # EXPERIMENTAL
+  # If false, a secure 16-character random password will be generated inside
+  # the guest on first boot and saved to `~/password` (with 0600 permissions).
+  # The user should manually rotate this password after the initial login.
+  # 🟢 Builtin default: true (false for macOS)
+  passwordlessSudo: null
 
 vmOpts:
   qemu:
--- a/website/content/en/docs/config/sudo.md
+++ b/website/content/en/docs/config/sudo.md
@@ -1,0 +1,26 @@
+---
+title: Sudo
+weight: 40
+---
+
+| ⚡ Requirement | Lima >= 2.2 |
+|---------------|-------------|
+
+By default, the guest user has passwordless sudo (`NOPASSWD:ALL`).
+
+The password is only generated once, on first boot. If you manually change the
+password and remove `~/password`, it will **not** be regenerated on subsequent
+boots, the file is purely informational for the initial login.
+
+```yaml
+user:
+  passwordlessSudo: false
+```
+
+## Caveats
+
+- Requires `plain: true` on non-macOS guests ([GHSA-2j9v-p4xj-cjw2](https://github.com/lima-vm/lima/security/advisories/GHSA-2j9v-p4xj-cjw2)): the guest agent's Unix socket tunneling is disabled until it can safely refuse privileged socket tunneling with passwordless sudo off.
+- Only applies to cloud-init provisioned guests (QEMU/vz). Setting `user.passwordlessSudo: false` on WSL2 (Windows) guests returns a validation error, as WSL2 does not support this configuration as of now.
+- The guest user is already a member of the `adm` and `systemd-journal` groups, which grants passive read access to `/var/log/*` and the full system journal independent of sudo.
+- `param` still relies on sudo internally (`param.env` is read via `sudo cat` inside the guest), so it may not work as expected when `user.passwordlessSudo: false` is set. A warning is printed in this case.
+- **macOS**: not configurable. Password-required sudo is always on (except `/sbin/shutdown -h now`), matching the built-in [macOS guest](/docs/usage/guests/macos/) behavior. Setting `user.passwordlessSudo: true` on a macOS guest returns a validation error.
--- a/website/content/en/docs/releases/experimental.md
+++ b/website/content/en/docs/releases/experimental.md
@@ -13,6 +13,7 @@ The following features are experimental and subject to change:
 - `audio.device` and `audio.interface`
 - `mountInotify: true`
 - `tpm: true`
+- `user.passwordlessSudo` (See [Sudo](../config/sudo))
 - `External drivers`: building and using drivers as separate executables (see [Virtual Machine Drivers](../dev/drivers))
 - [`vmType: krunkit`](../config/vmtype/krunkit.md)
 - [`github` URL scheme](../templates/github.md): referencing templates on GitHub with `github:` URLs
--- a/hack/bats/tests/passwordless-sudo.bats
+++ b/hack/bats/tests/passwordless-sudo.bats
@@ -1,0 +1,47 @@
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+load "../helpers/load"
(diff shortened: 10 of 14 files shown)


AkihiroSuda (changes requested):
The design has to be discussed before implementation.

- The password is expected to be stored inside guest, not host. It is negotiable to change this design, however, it is not trivial.
https://github.com/lima-vm/lima/blob/c8251205cebc1d20c6684b7f96c6f8bc629a4a99/templates/macos-26.yaml#L11-L13
- The password must not be written in the ISO, as it can be visible to non-root users https://github.com/lima-vm/lima/issues/4593
- The hostagent still depends on `sudo` in several places. How did you test it?
- Needs integration tests (BATS)

Aneesh-Hegde:
Hey @AkihiroSuda sorry for early implementation before discussing the implementation. I have converted PR to a draft before furthur affirmation.
So this is the plan:
> * The password is expected to be stored inside guest, not host. It is negotiable to change this design, however, it is not trivial.
>   https://github.com/lima-vm/lima/blob/c8251205cebc1d20c6684b7f96c6f8bc629a4a99/templates/macos-26.yaml#L11-L13
> * The password must not be written in the ISO, as it can be visible to non-root users #4593

can update `boot.sh` to generate the secret via `/dev/urandom`, apply it via `chpasswd`, and write it to `~/password` (owned by the user, 0600 permissions).
This approach solves a few problems at once: it keeps things consistent with how we handle macOS GUI passwords, avoids passing secrets from the host altogether, and solves the ISO visibility issue.

> * The hostagent still depends on `sudo` in several places. How did you test it?

I initially only validated the template rendering and didn't trace the end-to-end impact on `pkg/hostagent/ automation`. Sure enough, stripping `NOPASSWD` breaks internal (like mounts and port-forwarding).
So what i am thinking to do is instead of removal of `NOPASSWD:ALL`, what if we inject a strictly scoped `/etc/sudoers.d/lima` file? This would grant `NOPASSWD` only for the specific binaries the hostagent requires to function, while keeping general interactive sudo password-protected for the guest user. If this sounds like the right direction, I'll audit the hostagent call sites and propose the exact sudoers allowlist.

AkihiroSuda:
> So what i am thinking to do is instead of removal of NOPASSWD:ALL, what if we inject a strictly scoped /etc/sudoers.d/lima file?

This is the last resort.
The entire hostagent/guestagent architecture should be refactored to remove/reduce the dependency on `limactl shell INST sudo`.

We can also consider having a hidden system user for running `sudo` commands, but I fear it may accidentally look like malware.

Aneesh-Hegde:
Let me do the hostagent sudo audit first and come back with what it actually takes to drop `NOPASSWD:ALL` safely. If it's a handful of well-scoped changes, this PR can land as one complete piece. If it's bigger, I'll split that out as its own issue and keep this PR in draft until everything's ready together. 
do you have a preference on how we track the scope here?

AkihiroSuda:
It is likely that it needs to be split to several PRs

Aneesh-Hegde:
Makes sense. Rough guess at how this could split, I'll refine once I've done the audit:
- Move things like mounts and port-forwarding into guestagent if it's already running as root, so hostagent doesn't need sudo for those.
- Capabilities/setcap for anything that can use that instead of sudo.
- Whatever's left that genuinely needs root, figure out the smallest safe way to handle it.
- Final PR that ties it together, flips the sudo config based on passwordlessSudo, plus the password generation piece from this PR.

Let me do the audit first and see how close that actually is.

Aneesh-Hegde:
most cases have clean paths:

`sudo journalctl/sudo tail` → add guest user to `adm/systemd-journal` group via cloud-init
`sudo grep on fuse*.conf` → relax permissions, file isn't sensitive
`sudo cat param.env` → drops naturally with #4593  (has existing FIXME)
`sudo cat in copyToHost` → user-configured, leaving this alone

The only hard case is `sudo mkdir/chown/ln` on `/run/host-services`. Guestagent runs as the guest user not root, so RPC isn't directly viable here. Two ideas:

- Idea 1: small oneshot root systemd unit via cloud-init just for the `mkdir/chown, ln -sf` moves to a user-level unit after
- Idea 2: move the socket to `/run/user/${UID}/lima/` (already user-writable), no root needed at all but is bigger change though

Revised plan if this looks right:

Phase 1:  small focused PRs for the easy wins (log group membership, fuse.conf)
Phase 2: `/run/host-services` once you confirm direction above
Phase 3: this PR lands with `boot.sh` password generation + passwordlessSudo toggle once `NOPASSWD:ALL` is safely removable

Any suggestions from your end @AkihiroSuda , and which direction would be better for `/run/host-services`?

EDIT: Also will ensure log whenever sudo is accessed as per #5114

Aneesh-Hegde:
@AkihiroSuda gentle ping

AkihiroSuda:
> most cases have clean paths:
> 
> `sudo journalctl/sudo tail` → add guest user to `adm/systemd-journal` group via cloud-init 

This was recently done: https://github.com/lima-vm/lima/pull/5050

> `sudo grep on fuse*.conf` → relax permissions, file isn't sensitive

IIRC on some distro the conf file is automatically created with the strict permissions.
So this should be queried via a notification file akin to `/run/lima-boot-done`, or, via the guestagent API.

>  `sudo cat param.env` → drops naturally with #4593 (has existing FIXME)

Yes, but we haven't fully figured out if we can implement #4953 safely.
Probably we should move param.env codes to the guestagent.

> `sudo cat in copyToHost` → user-configured, leaving this alone

Yes, when the user configured to copy the privileged files.
It still should support copying non-privileged files by dropping sudo, though.

> The only hard case is `sudo mkdir/chown/ln` on `/run/host-services`. Guestagent runs as the guest user not root, so RPC isn't directly viable here. Two ideas:
> 
> * Idea 1: small oneshot root systemd unit via cloud-init just for the `mkdir/chown, ln -sf` moves to a user-level unit after

SGTM, but no need to depend on systemd untis?

> * Idea 2: move the socket to `/run/user/${UID}/lima/` (already user-writable), no root needed at all but is bigger change though

This breaking change is not preferable unless it is really inevitable.

Aneesh-Hegde:
> IIRC on some distro the conf file is automatically created with the strict permissions.
> So this should be queried via a notification file akin to `/run/lima-boot-done`, or, via the guestagent API.

planning to follow the lima-boot-done pattern: `35-setup-packages.sh` already runs as root and modifies fuse.conf, so we can touch `/run/lima-fuse-configured` at the end of `update_fuse_conf()` and replace the `sudo grep` in `requirements.go` with a file check. Does that sound right?

> Yes, but we haven't fully figured out if we can implement #4953 safely.
> Probably we should move param.env codes to the guestagent.

should I hold off until #4593  is clearer, or start scoping the guestagent move now?

> Yes, when the user configured to copy the privileged files.
> It still should support copying non-privileged files by dropping sudo, though.

makes sense, will add a privilege check and drop sudo for non-privileged files.

> SGTM, but no need to depend on systemd untis?
> 
> >

`SSH_AUTH_SOCK` is a host-side runtime variable so `ln -sf` has to stay in hostagent. But `mkdir/chown` can move to a boot script since those don't need runtime context. hostagent can then do the `ln` as the guest user without sudo since the directory is already pre-owned. Also noticed #5121  will add a new script, and just adjust later if implementation is completed.

So what my plan would be:
PR 1 — `fuse*.conf`: notification file approach in 35-setup-packages.sh + drop sudo grep in requirements.go
PR 2 — copyToHost: drop sudo for non-privileged files
PR 3 — `/run/host-services`: move `mkdir/chown` to boot script, drop sudo from ln -sf in hostagent
PR 4 — param.env: move into guestagent (depends on #4593  direction)
PR 5 — this PR: boot.sh password generation + passwordlessSudo toggle, once PRs 1-3 are in

cc @AkihiroSuda . Let me know if this sound right i can start with the implementation.

Aneesh-Hegde:
@AkihiroSuda gentle ping :)

AkihiroSuda:
> Let me know if this sound right i can start with the implementation.

Sounds right, but the following issue has to be considered too:

- - -

>> https://github.com/lima-vm/lima/security/advisories/GHSA-2j9v-p4xj-cjw2
>> ## An arbitrary user in a QEMU VM could gain the root privilege in the VM via the guest agent socket
>>
>> ### Impact
>> On an instance of Lima running with `qemu` driver, an arbitrary user in the VM could access `/run/lima-guestagent.sock` when the guest agent is enabled.
>> 
>> This could result in running an arbitrary command with the root privileges in the VM (**not on the host**), as `lima-guestagent.sock` provides the tunneling service for an arbitrary address, including a Unix socket address for privileged daemons like D-Bus.
>> 
>> This vulnerability is not exploitable on `vz` driver, as the guest agent uses vsocks instead of Unix sockets.
>> 
>> ### Patches
>> Patched in Lima v2.1.3 (8a45892378d22f40505c31a38f786a07701b6d50)
>> 
>> > [!NOTE]
>> > The default user account in the VM can still run an arbitrary command as the root.
>> > This is not a vulnerability, as the user can already run an arbitrary command with `sudo` by design.
>> 
> - - -
> 
> When password-less sudo is disabled, the tunneling service in the guest agent must refuse to tunnel Unix sockets.
> The socket should be still forwarded via SSH (no root access) instead of the privileged guest agent's tunneling service.
> As a baby step, we may just say that "`passwordlessSudo: false` requires `plain: true`" to disable the guest agent, and call it a day.
>  

 _Originally posted by @AkihiroSuda in [#4594](https://github.com/lima-vm/lima/issues/4594#issuecomment-4751670131)_

Aneesh-Hegde:
> but the following issue has to be considered too

Sure, will enforce plain: true when passwordlessSudo: false as the interim constraint, and scope a proper Unix socket tunneling refusal in guestagent as a follow-up. Starting with:


> PR 1 — `fuse*.conf`: notification file approach in 35-setup-packages.sh + drop sudo grep in requirements.go

Aneesh-Hegde:
@AkihiroSuda This PR is good to go for review. PTAL!

pkg/limatype/lima_yaml.go
```diff
@@ -107,11 +107,12 @@ var (
...
-	Comment *string `yaml:"comment,omitempty" json:"comment,omitempty" jsonschema:"nullable"`
-	Home    *string `yaml:"home,omitempty" json:"home,omitempty" jsonschema:"nullable"`
-	Shell   *string `yaml:"shell,omitempty" json:"shell,omitempty" jsonschema:"nullable"`
-	UID     *uint32 `yaml:"uid,omitempty" json:"uid,omitempty" jsonschema:"nullable"`
+	Name             *string `yaml:"name,omitempty" json:"name,omitempty" jsonschema:"nullable"`
+	Comment          *string `yaml:"comment,omitempty" json:"comment,omitempty" jsonschema:"nullable"`
+	Home             *string `yaml:"home,omitempty" json:"home,omitempty" jsonschema:"nullable"`
+	Shell            *string `yaml:"shell,omitempty" json:"shell,omitempty" jsonschema:"nullable"`
+	UID              *uint32 `yaml:"uid,omitempty" json:"uid,omitempty" jsonschema:"nullable"`
+	PasswordlessSudo *bool   `yaml:"passwordlessSudo,omitempty" json:"passwordlessSudo,omitempty" jsonschema:"nullable"`
```
AkihiroSuda, line 117: Hostagent still has sudo in several places:
- https://github.com/lima-vm/lima/blob/069d01a885183a5b9c8b2aa1184d3d76103e72cf/pkg/hostagent/requirements.go#L97
- https://github.com/lima-vm/lima/blob/069d01a885183a5b9c8b2aa1184d3d76103e72cf/pkg/hostagent/hostagent.go#L1102-L1116

pkg/limayaml/defaults.go
```diff
@@ -219,6 +225,10 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 		}
 		// warn = false
 	}
+	// fallback to nopasswd
+	if y.User.PasswordlessSudo == nil {
+		y.User.PasswordlessSudo = ptr.Of(true)
```
AkihiroSuda: false on macOS

TODO: what about Windows? cc @mie313 @unsuman

AkihiroSuda: macOS impl is here: https://github.com/lima-vm/lima/blob/069d01a885183a5b9c8b2aa1184d3d76103e72cf/pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go#L320

If it is hard to implement `passwordlessSudo: true` on macOS, at least it should return an error.

https://lima-vm.io/docs/usage/guests/macos/ has to be updated too.

templates/default.yaml
```diff
@@ -368,6 +375,11 @@ user:
   # Shell. Needs to be an absolute path.
   # 🟢 Builtin default: "/bin/bash"
   shell: null
+  # If false, a secure 16-character random password will be generated inside
+  # the guest on first boot and saved to `~/password` (with 0600 permissions).
+  # The user should manually rotate this password after the initial login.
+  # 🟢 Builtin default: true
+  passwordlessSudo: null
```
AkihiroSuda, line 382: Probably https://lima-vm.io/docs/config/ should have a dedicated `/sudo` page

templates/default.yaml
```diff
@@ -368,6 +375,11 @@ user:
   # Shell. Needs to be an absolute path.
   # 🟢 Builtin default: "/bin/bash"
   shell: null
+  # If false, a secure 16-character random password will be generated inside
+  # the guest on first boot and saved to `~/password` (with 0600 permissions).
+  # The user should manually rotate this password after the initial login.
+  # 🟢 Builtin default: true
```
AkihiroSuda: false on macOS

pkg/limayaml/validate.go
```diff
@@ -406,6 +407,12 @@ func Validate(y *limatype.LimaYAML, warn bool) error {
 			}
 		}
 	}
+	// GHSA-2j9v-p4xj-cjw2 constraint: require plain mode when passwordless sudo is disabled
+	if y.User.PasswordlessSudo != nil && !*y.User.PasswordlessSudo {
+		if y.Plain == nil || !*y.Plain {
+			errs = errors.Join(errs, errors.New("`passwordlessSudo: false` requires `plain: true` (see GHSA-2j9v-p4xj-cjw2)"))
```
AkihiroSuda:
```suggestion
			errs = errors.Join(errs, errors.New("`passwordlessSudo: false` requires `plain: true`"))
```

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -203,6 +203,15 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.user ]; then
 	done
 fi
 
+if [ "${LIMA_CIDATA_PASSWORDLESS_SUDO}" != "1" ]; then
+	if [ ! -f "${LIMA_CIDATA_HOME}/password" ]; then
+		PASSWORD=$(head -c 8 /dev/urandom | od -An -tx1 | tr -d ' \n')
+		echo "${LIMA_CIDATA_USER}:${PASSWORD}" | chpasswd
```
AkihiroSuda: chpasswd is missing on FreeBSD and macOS

pkg/limayaml/defaults.go
Aneesh-Hegde: Yes it is little difficult. `lock_passwd` is just a no-op right now in `fakecloudinit_darwin.go`, so implementing it properly is trickier than on Linux. `sysadminctl` needs a password to even create the account, and that same account backs the GUI session too, so there's no clean passwordless path like we have for headless guests. Doing this right would mean messing with `dscl/pwpolicy`, which risks breaking GUI login for existing macOS users.
for now I've just added a validation error when `passwordlessSudo: true` is set on macOS. can open an upstream issue to track real support if that works

pkg/limatype/lima_yaml.go
Aneesh-Hegde: found a fix for log streaming (adding `adm`+`systemd-journal` groups), but flagging first: those groups aren't scoped to cloud-init, they give read access to all of `/var/log/*` and the full journal. since the guest user already has the sudo password, this doesn't grant new access but it does mean any process running as that user gets passive read access without needing sudo at all. want me to go ahead, or is something more scoped worth looking at?

AkihiroSuda: > it does mean any process running as that user gets passive read access without needing sudo at all.

I think this is fine, but it has to be documented.

website/content/en/docs/config/sudo.md
```diff
@@ -0,0 +1,25 @@
...
+Set `user.passwordlessSudo: false` to require a password for `sudo` inside the guest.
+When disabled, a random password is generated on first boot and saved to `~/password`
+(mode `0600`) inside the guest. Rotate it manually after first login.
+
+```yaml
+user:
+  passwordlessSudo: false
+```
+
+## Constraints
```
AkihiroSuda: This should be called `Caveats` to match other documentations

website/content/en/docs/config/sudo.md
```diff
@@ -0,0 +1,25 @@
+---
+title: Sudo
+weight: X
+---
+
```
AkihiroSuda, line 5: Needs requirement information
https://github.com/search?q=repo%3Alima-vm%2Flima%20%22⚡%20Requirement%22&type=code

templates/default.yaml
```diff
@@ -368,6 +375,11 @@ user:
   # Shell. Needs to be an absolute path.
   # 🟢 Builtin default: "/bin/bash"
   shell: null
+  # If false, a secure 16-character random password will be generated inside
+  # the guest on first boot and saved to `~/password` (with 0600 permissions).
+  # The user should manually rotate this password after the initial login.
+  # 🟢 Builtin default: true (false for macOS)
+  passwordlessSudo: null
```
AkihiroSuda, line 382: Let's mark it experimental for now

https://github.com/lima-vm/lima/blob/66bcdb363cbe371590c363667c4ac2994fab5dfc/pkg/limayaml/validate.go#L610

https://lima-vm.io/docs/releases/experimental/

Aneesh-Hegde:
For the WSL2 case, I added a validation error if `passwordlessSudo: false` is set(also mentioned in docs). As per the [official Microsoft WSL docs](https://learn.microsoft.com/en-us/windows/wsl/setup/environment#set-up-your-linux-username-and-password:~:text=into%20any%20issues.-,Set%20up%20your%20Linux%20username%20and%20password,-Once%20the%20process),

AkihiroSuda:
How does the current PR work with params?
Params are ignored?
https://github.com/lima-vm/lima/pull/5109#discussion_r3503180825

Aneesh-Hegde:
> How does the current PR work with params? Params are ignored? [#5109 (comment)](https://github.com/lima-vm/lima/pull/5109#discussion_r3503180825)

one solution is moving it to guestagent as mentioned in [#4593(comment)](https://github.com/lima-vm/lima/issues/4593#issuecomment-3951488061) or simpler would be relaxing cidata mount which is already mentioned in the same. If the assignee is not active I am happy to work on it
EDIT: Also mentioned in one of the [comments](https://github.com/lima-vm/lima/pull/5109#:~:text=PR%204%20%E2%80%94%20param.env%3A%20move%20into%20guestagent%20(depends%20on%20%234593%20direction)) regarding dependancy.
> PR 4 — param.env: move into guestagent (depends on https://github.com/lima-vm/lima/issues/4593 direction)

Aneesh-Hegde:
@AkihiroSuda should I address the cidata mount permissions fix in this PR or as a follow-up referencing #4593  ?

AkihiroSuda:
> @AkihiroSuda should I address the cidata mount permissions fix in this PR or as a follow-up referencing #4593 ?

A follow-up PR, please.

For now, `passwordlessSudo: false` should print a warning when params are present.
The limitation has to be documented too.

pkg/hostagent/hostagent.go
AkihiroSuda, line 1154: sudo is still needed to read these logs

AkihiroSuda, line 1: This should be a separate PR

AkihiroSuda:
Marking as a draft. Please click the "Ready for review" button when ready.,

pkg/cidata/cidata.TEMPLATE.d/user-data
```diff
@@ -55,13 +55,26 @@ users:
     {{/* (Why doesn't macOS VM support graceful shutdown?) */}}
     sudo: ALL=(ALL) NOPASSWD:/sbin/shutdown -h now
 {{- else }}
+    groups:
+    - adm
+    - systemd-journal
```
AkihiroSuda: Duplicate:
- https://github.com/lima-vm/lima/pull/5050

pkg/cidata/cidata.TEMPLATE.d/user-data
```diff
@@ -55,13 +55,26 @@ users:
     {{/* (Why doesn't macOS VM support graceful shutdown?) */}}
     sudo: ALL=(ALL) NOPASSWD:/sbin/shutdown -h now
 {{- else }}
+    groups:
+    - adm
```
AkihiroSuda: ?

website/content/en/docs/releases/experimental.md
AkihiroSuda, line 15: Experimental YAML properties should go here

pkg/cidata/cidata.TEMPLATE.d/user-data
Aneesh-Hegde: Misconfig my bad

website/content/en/docs/config/sudo.md
```diff
@@ -0,0 +1,25 @@
...
+  passwordlessSudo: false
+```
+
+## Caveats
+
+- Requires `plain: true` on non-macOS guests ([GHSA-2j9v-p4xj-cjw2](https://github.com/lima-vm/lima/security/advisories/GHSA-2j9v-p4xj-cjw2)): the guest agent's Unix socket tunneling is disabled until it can safely refuse privileged socket tunneling with passwordless sudo off.
+- Only applies to cloud-init provisioned guests (QEMU/vz). WSL2 instances ignore this setting.
+- The guest user is already a member of the `adm` and `systemd-journal` groups, which grants passive read access to `/var/log/*` and the full system journal independent of sudo.
+- `param` still relies on sudo internally (`param.env` is read via `sudo cat` inside the guest), so it may not work as expected when `passwordlessSudo: false` is set. A warning is printed in this case.
+- **macOS**: not configurable. Password-required sudo is always on (except `/sbin/shutdown -h now`), matching the built-in [macOS guest](/docs/usage/guests/macos/) behavior. Setting `passwordlessSudo: true` on a macOS guest returns a validation error.
```
AkihiroSuda: It should be marked as unsupported on Windows too?
cc @mie313

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -203,6 +203,21 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.user ]; then
 	done
 fi
 
+if [ "${LIMA_CIDATA_PASSWORDLESS_SUDO}" != "1" ]; then
+	if [ ! -f "${LIMA_CIDATA_HOME}/password" ]; then
```
AkihiroSuda: When a user changes the password manually and remove the `password` file, the password is automatically reset again

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -203,6 +203,21 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.user ]; then
...
+		PASSWORD=$(head -c 8 /dev/urandom | od -An -tx1 | tr -d ' \n')
+		if command -v chpasswd >/dev/null 2>&1; then
+			echo "${LIMA_CIDATA_USER}:${PASSWORD}" | chpasswd
+		elif command -v pw >/dev/null 2>&1; then
+			echo "${PASSWORD}" | pw usermod "${LIMA_CIDATA_USER}" -h 0
+		else
+			echo "lima: warning: no supported tool to set password found, skipping" >&2
```
AkihiroSuda: Consistently use the `WARNING` function

AkihiroSuda: Shouldn't it rather result in a hard error?

pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -203,6 +203,21 @@ if [ -d "${LIMA_CIDATA_MNT}"/provision.user ]; then
...
+		fi
+		echo "${PASSWORD}" >"${LIMA_CIDATA_HOME}/password"
```
AkihiroSuda: Should be written after chown and chmod

AkihiroSuda:
@Aneesh-Hegde Could you click `Resolve conversation` for resolved comments?

website/content/en/docs/releases/experimental.md
```diff
@@ -13,6 +13,7 @@ The following features are experimental and subject to change:
 - `audio.device` and `audio.interface`
 - `mountInotify: true`
 - `tpm: true`
+- `user.passwordlessSudo` (See [Sudo](../config/sudo))
```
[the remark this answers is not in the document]
Aneesh-Hegde: Not needed imho

pkg/cidata/template.go
```diff
@@ -100,6 +100,7 @@ type TemplateArgs struct {
 	Home                            string // home directory
 	Shell                           string // login shell
 	UID                             uint32
+	PasswordlessSudo                bool
 	SSHPubKeys                      []string
 	Mounts                          []Mount
```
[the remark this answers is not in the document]
Aneesh-Hegde: Not needed imo

Aneesh-Hegde:
@AkihiroSuda i think changes are gtg and seems all issues are resolved

AkihiroSuda (approved):
Thanks

PR #5113, opened 2026-06-13
Revert "Support host block-device attachment on macOS VZ"

AkihiroSuda:
This reverts commit 439bf1fad800a9ea93ac23194888bda5b0fdcea3.

Revert #4866
Fix #5112
Close #5104

- - -
> ```
> %everyone ALL=(root:wheel) NOPASSWD:NOSETENV: /usr/local/bin/limactl sudo-open-block-device
> ```
> 
> Even when the `limactl` binary is owned by `root:wheel`, this still allows an arbitrary user to open an arbitrary file as the root, so this is absolutely insecure by design.
> 
> @pirate Sorry for going back and forth, but I have to revert #4866  

 _Originally posted by @AkihiroSuda in [#5112](https://github.com/lima-vm/lima/issues/5112#issuecomment-4699026195)_

Issue #5112, opened 2026-06-13
[Only affects master] `limactl sudo-open-block-device` is insecure


AkihiroSuda:
PR https://github.com/lima-vm/lima/pull/4866 introduced `limactl sudo-open-block-device` helper for opening block devices.
This is insecure unless `limactl` is owned by `root:wheel`, as a user may replace the binary with malicious one.

I'm not treating this issue as a vulnerability, as it only affects the master branch.

Related:
- https://github.com/lima-vm/lima/pull/2734

AkihiroSuda:
```
%everyone ALL=(root:wheel) NOPASSWD:NOSETENV: /usr/local/bin/limactl sudo-open-block-device
```

Even when the `limactl` binary is owned by `root:wheel`, this still allows an arbitrary user to open an arbitrary file as the root, so this is absolutely insecure by design.

@pirate Sorry for going back and forth, but I have to revert #4866

Issue #2224, opened 2024-02-24
USB passthrough rabbit hole


retpolanne:
## QEMU usb passthrough rabbit hole

Okay, so I went down this rabbit hole to see how doable it is to implement USB passthrough. 

Let's go step by step:

1. Lima talks to QEMU through QMP (QEMU Machine Protocol). It uses a lib made by DigitalOcean, go-qemu, which has a function DeviceAdd [1] that seems to be autogenerated. From the qemu qmp ref docs [2] we can see that it accepts device, bus and id. However, according to [3], it should accept other kinds of args as well. In our case, device would be usb-host, and then we would require vendorid and productid.

2. Trying on the qemu monitor, I see this error: 

```
(qemu) device_add usb-host,vendorid=0x1235,productid=0x8211
libusb_detach_kernel_driver: -3 [ACCESS]
libusb_detach_kernel_driver: -3 [ACCESS]
libusb_detach_kernel_driver: -3 [ACCESS]
libusb_detach_kernel_driver: -3 [ACCESS]
libusb_kernel_driver_active: -5 [NOT_FOUND]
libusb_kernel_driver_active: -5 [NOT_FOUND]
libusb_kernel_driver_active: -5 [NOT_FOUND]
libusb_kernel_driver_active: -5 [NOT_FOUND]
libusb_kernel_driver_active: -5 [NOT_FOUND]
libusb_kernel_driver_active: -5 [NOT_FOUND]
libusb_kernel_driver_active: -5 [NOT_FOUND]
```

4. The whole bus could be connected, according to [4]

5. From my tests, if the device_add syntax is wrong or device is inexistent, it freezes the whole instance, seems to be a qemu bug. 

More info at [5] and [6]

## References

[1] https://pkg.go.dev/github.com/digitalocean/go-qemu@v0.0.0-20230711162256-2e3d0186973e/qmp/raw#Monitor.DeviceAdd

[2] https://www.qemu.org/docs/master/interop/qemu-qmp-ref.html#qapidoc-2506

[3] https://qemu-project.gitlab.io/qemu/system/devices/usb.html

[4] https://github.com/lima-vm/lima/pull/1317

[5] https://gitlab.com/qemu-project/qemu/-/issues/1951

[6] https://github.com/libusb/libusb/issues/908

retpolanne:
I wonder if `libusb_get_device_descriptor` is not returning, so qemu gets stuck.

```
[...]

Issue #5067, opened 2026-06-01
v2.2.0 planning


AkihiroSuda:
The master branch should be now thawed for v2.2.0 PRs.
https://github.com/lima-vm/lima/milestone/70

The highlight in v2.2.0 will be:
- https://github.com/lima-vm/lima/issues/4852

arixmkii:
Suggesting this cleanup as part of version 2.2.0 https://github.com/lima-vm/lima/pull/4970 if it doesn't feel too rushed, because deprecation was delivered only in version 2.1.2.

afbjorklund:
I think the additional drivers will have to wait until Lima 3.0, since there is a lot of movement with the Windows changes...

* External driver (ssh)
* Cloud drivers
* Apple Container driver
* Kata Container driver

I had some PRs for the in-tree, but it doesn't seem worth it to keep rebasing them if they are not going to be merged?

The out-of-tree drivers could be developed whenever though, I rebased against the Lima API breakage in 2.0 and 2.1.

* https://github.com/afbjorklund/lima-driver-digitalocean
* https://github.com/afbjorklund/lima-driver-proxmox

afbjorklund:
The uiutil change (pterm) and the kubernetes tarballs were removed from the 2.2.0 milestone, they might re-appear later.

* https://github.com/lima-vm/lima/issues/1881
* https://github.com/lima-vm/lima/issues/4875

AkihiroSuda:
Released v2.2.0-beta.0: https://github.com/lima-vm/lima/releases/tag/v2.2.0-beta.0

GA should be released by July 21 Tue (one week before KubeCon Japan).

AkihiroSuda:
Let me feature-freeze v2.2 by July 14.

Missed PRs can be revisited in August for v2.3
- https://github.com/lima-vm/lima/issues/5223

cc @lima-vm/maintainers

AkihiroSuda:
https://github.com/lima-vm/lima/releases/tag/v2.2.0

PR #5117, opened 2026-06-14
Secure macOS VZ block-device attachment

pirate:
Reintroduces host block-device attachment for builtin macOS VZ after #4866 was reverted in #5113 following the security report in #5112. Users explicitly authorize each device with `limactl sudoers --block-device=/dev/diskN`, then attach it with `--block-device` or YAML `blockDevices`.

## Security and lifecycle

- A dedicated `lima-privileged-block-device` helper opens only the device named in its sudoers-authorized argv. Grants are scoped to the current user; `limactl` itself is never granted root execution.
- The helper must be a regular executable; it and all ancestor directories must be root-owned, non-user-writable, free of symlinks and extended ACLs. Device paths are restricted to Darwin disk nodes and opened with `O_NOFOLLOW`.
- The private Unix socket handoff authenticates both peers, rejects socket symlinks, and validates a random nonce before receiving the descriptor. The receiver checks device type and device identity. Duplicated descriptors have close-on-exec set.
- Descriptors remain valid for the VM lifetime and close on failure or shutdown. Exported network helpers retain their existing process-lifetime behavior for krunkit.
- Whole-disk locks prevent instances sharing a `LIMA_HOME` from concurrently attaching the same disk through raw/block aliases or partitions. Within one VM, disjoint partitions reuse its lock; duplicate aliases and parent/child overlaps are rejected. External VZ driver use is rejected.
- Sudoers verification matches complete logical lines, so appending another command or splicing the grant into a preceding continuation cannot satisfy the expected fragment. Grant rendering is order-independent, and regeneration warns about omitted network and device grants.

The grant remains permission to read and write a device slot until removed from sudoers. It is not tied to a physical disk UUID. The documentation explains slot reuse, boot-disk risks, install layout, and the need to avoid concurrent host/guest or cross-user mounts.

## Review fixes and tests

Addresses the [September 16 review](https://jandubois.github.io/lima/20260916-233839-pr-5117.html) and [September 18 follow-up](https://jandubois.github.io/lima/20260918-123055-pr-5117.html): same-VM lock reuse, continuation-aware sudoers checks, stable grant ordering, missing release helpers, early path validation, unsupported-driver rejection, and cleanup/diagnostic corrections. The BATS suite creates a unique temporary sudoers fragment and revokes it before detaching its RAM disk, preserving existing grants.

The [latest September 18 review](https://jandubois.github.io/lima/20260918-211054-pr-5117.html) additionally led to:

- Root-controlled, readable sudoers files (`root:wheel 0444`), tested directly instead of copying through `sudo cat`; `visudo` runs before installation and privileged BATS retries are disabled so cleanup failures cannot be hidden.
- Startup-only helper preflight, allowing cleanup/deletion without the helper; macOS-version errors precede installation errors, and helper leaf checks reject directories/non-executables.
- Rejection of overlapping attachments while preserving disjoint partitions, accurate network-check warnings, and current-user/multi-user regeneration and revocation guidance.
- Simpler CLI YAML edits and artifact assertions for native privileged helpers. The shared artifact build intentionally also includes the existing Linux networking helper that was previously omitted.

Setup help explains protected binary installation without a source build, grant inspection/validation/installation, and revocation. VZ validates the helper installation before hostagent startup so installation errors reach the CLI. This retains the root-owned installation requirement; it does not add an automatic installer or automatic device authorization.

Regression tests cover real Unix socket connections and symlinks, rejected peers sending no data, descriptor identity, missing stat data, raw/block/partition lock collisions and release, duplicate devices, and runtime external-driver rejection. BATS covers real guest/host disk round trips plus multi-device sudoers checks and removal of temporary grants.

Regression-first checks reproduced overlapping attachments, missing-helper cleanup failure, and the missing network-check warning. New tests cover those paths and helper leaf validation; artifact fixtures fail when the native helper is absent. Earlier tests cover same-VM locking, grant ordering, retention warnings, configured paths, and continued sudoers lines, including real `visudo` behavior. Local affected-package tests, socket/lock race tests, golangci-lint, gomodjail, and protolint pass. A real macOS release archive passed the helper assertion. Manual macOS VM acceptance remains incomplete; no local VMs or privileged disk operations were run for this revision.

Hosted CI for signed commit `cb78c1b3`: 34 passed, 2 skipped, no failures or pending checks. All three VZ block-device BATS cases passed without retries: CLI disk roundtrip, installed-file sudoers checks/revocation, and YAML raw-disk roundtrip. One commit on current `master` (`376d3104`), GitHub signature verified and DCO passing.

Related: #1314, #2224; follow-ups #5104 and abiosoft/colima#1592.

Assisted-by: Codex

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -591,8 +591,24 @@ jobs:
         cache: false
     - name: Make
       run: make
+    - name: Install socket_vmnet
+      env:
+        SOCKET_VMNET_VERSION: v1.2.2
+      run: |
+        (
+          cd ~
+          git clone https://github.com/lima-vm/socket_vmnet
+          cd socket_vmnet
+          git checkout $SOCKET_VMNET_VERSION
+          sudo git config --global --add safe.directory /Users/runner/socket_vmnet
+          sudo make PREFIX=/opt/socket_vmnet install
+        )
     - name: Install
-      run: sudo make install
+      run: |
+        sudo install -d -o root -g wheel -m 0755 /opt/lima
+        sudo make PREFIX=/opt/lima install
+        echo /opt/lima/bin >>$GITHUB_PATH
+        /opt/lima/bin/limactl sudoers | sudo tee /etc/sudoers.d/lima
     - name: Cache image used by templates/${{ matrix.template }}
       uses: ./.github/actions/setup_cache_for_template
       with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -584,6 +584,7 @@ jobs:
     - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
       with:
         persist-credentials: false
+        submodules: true
     - uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16  # v6.5.0
       with:
         go-version: stable
@@ -602,6 +603,8 @@ jobs:
       run: brew uninstall --ignore-dependencies --force qemu
     - name: Test
       run: ./hack/test-templates.sh templates/${{ matrix.template }}
+    - name: Run BATS VZ block-device tests
+      run: ./hack/bats/lib/bats-core/bin/bats --timing ./hack/bats/extras/vz-block-device.bats
     - if: failure()
       uses: ./.github/actions/upload_failure_logs_if_exists
       with:
--- a/cmd/limactl/sudo_open_block_device_command_darwin.go
+++ b/cmd/limactl/sudo_open_block_device_command_darwin.go
@@ -1,0 +1,28 @@
+//go:build darwin
+
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"github.com/spf13/cobra"
+
+	"github.com/lima-vm/lima/v2/pkg/blockdevice"
+)
+
+func isPrivilegedHelperCommand(name string) bool {
+	return name == blockdevice.SudoOpenBlockDeviceCommand
+}
+
+func registerHiddenCommands(rootCmd *cobra.Command) {
+	rootCmd.AddCommand(&cobra.Command{
+		Use:    blockdevice.SudoOpenBlockDeviceCommand,
+		Short:  "Open a host block device via privileged helper",
+		Args:   WrapArgsError(cobra.NoArgs),
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			return blockdevice.ServeSudoOpenBlockDevice(cmd.InOrStdin())
+		},
+	})
+}
--- a/cmd/limactl/sudo_open_block_device_command_darwin.go
+++ b/cmd/limactl/sudo_open_block_device_command_darwin.go
@@ -19,10 +19,10 @@ func registerHiddenCommands(rootCmd *cobra.Command) {
 	rootCmd.AddCommand(&cobra.Command{
 		Use:    blockdevice.SudoOpenBlockDeviceCommand,
 		Short:  "Open a host block device via privileged helper",
-		Args:   WrapArgsError(cobra.NoArgs),
+		Args:   WrapArgsError(cobra.ExactArgs(1)),
 		Hidden: true,
-		RunE: func(cmd *cobra.Command, _ []string) error {
-			return blockdevice.ServeSudoOpenBlockDevice(cmd.InOrStdin())
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return blockdevice.ServeSudoOpenBlockDevice(args[0], cmd.InOrStdin())
 		},
 	})
 }
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -27,7 +27,8 @@ To validate the existing /etc/sudoers.d/lima file:
 $ limactl sudoers --check /etc/sudoers.d/lima
 `,
 		Short: "Generate the content of the /etc/sudoers.d/lima file",
-		Long: fmt.Sprintf(`Generate the content of the /etc/sudoers.d/lima file for enabling vmnet.framework support (socket_vmnet) on macOS.
+		Long: fmt.Sprintf(`Generate the content of the /etc/sudoers.d/lima file for macOS host helpers that require privilege escalation.
+This includes vmnet.framework support (socket_vmnet) and host block-device attachment with --block-device on supported backends.
 The content is written to stdout, NOT to the file.
 This command must not run as the root user.
 See %s for the usage.`, socketVMNetURL),
--- a/cmd/limactl/sudoers.go
+++ b/cmd/limactl/sudoers.go
@@ -28,7 +28,8 @@ $ limactl sudoers --check /etc/sudoers.d/lima
 `,
 		Short: "Generate the content of the /etc/sudoers.d/lima file",
 		Long: fmt.Sprintf(`Generate the content of the /etc/sudoers.d/lima file for macOS host helpers that require privilege escalation.
-This includes vmnet.framework support (socket_vmnet) and host block-device attachment with --block-device on supported backends.
+This includes vmnet.framework support (socket_vmnet). Use --block-device=/dev/rdiskN to also emit opt-in
+host block-device helper entries for the listed devices and current user.
 The content is written to stdout, NOT to the file.
 This command must not run as the root user.
 See %s for the usage.`, socketVMNetURL),
@@ -39,5 +40,7 @@ See %s for the usage.`, socketVMNetURL),
 	cfgFile, _ := networks.ConfigFile()
 	sudoersCommand.Flags().Bool("check", false,
 		fmt.Sprintf("check that the sudoers file is up-to-date with %#q", cfgFile))
+	sudoersCommand.Flags().StringSlice("block-device", nil,
+		"include the macOS VZ host block-device helper for the current user and comma-separated devices")
 	return sudoersCommand
 }
--- a/hack/bats/tests/shell.bats
+++ b/hack/bats/tests/shell.bats
@@ -5,10 +5,6 @@ load "../helpers/load"
 
 INSTANCE=bats-dummy
 
-normalize_stderr_log() {
-    printf '%s\n' "$1" | sed -E 's/^time="[^"]+" //'
-}
-
 @test 'lima stopped lima instance' {
     # check that the "tty" flag is used, also for stdin
     run_e -1 limactl shell --tty=false "$INSTANCE" true </dev/null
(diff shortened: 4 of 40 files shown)


pirate:
I added a lot more security hardening, including a nonce challenge and socket ownership checks, let me know if it needs more @jandubois @AkihiroSuda.

cmd/limactl/sudoers.go
```diff
@@ -38,5 +40,7 @@ See %s for the usage.`, socketVMNetURL),
 	cfgFile, _ := networks.ConfigFile()
 	sudoersCommand.Flags().Bool("check", false,
 		fmt.Sprintf("check that the sudoers file is up-to-date with %#q", cfgFile))
+	sudoersCommand.Flags().Bool("block-device", false,
+		"include the macOS VZ host block-device helper for the current user")
```
AkihiroSuda: Can we limit the list of the block devices here?

pirate: limit in what sense? check against the list that's requested? or limit it to some hardcoded patterns of allowed paths? or limit based on fs permissions?

AkihiroSuda: like `--block-device=/dev/rdisk2,/dev/rdisk3`

66Ton99:
@pirate, what do you need to finish it? Or maybe you need some help?

AkihiroSuda:
Needs rebase

hack/test-templates.sh
```diff
@@ -600,7 +600,9 @@ fi
 if [[ -n ${CHECKS["user-v2"]} ]]; then
 	INFO "Testing user-v2 network"
 	secondvm="$NAME-1"
+	limactl delete -f "$secondvm" >/dev/null 2>&1 || true
```
AkihiroSuda: ?

hack/bats/tests/shell.bats
```diff
@@ -36,12 +40,10 @@ INSTANCE=bats-dummy
     # --instance should resolve the instance name the same way as the positional arg.
     # Normalize the logrus timestamp because the two invocations may land in different seconds.
     run_e -1 limactl shell --tty=false --instance nonexistent
-    inst_output=$(sed 's/^time="[^"]*"/time="TIMESTAMP"/' <<<"$stderr")
+    inst_output=$(normalize_stderr_log "$stderr")
```
AkihiroSuda: ?

AkihiroSuda (dismissed):
Thanks, looks good, but I want another review from @lima-vm/committers and it may miss v2.2:

- https://github.com/lima-vm/lima/issues/5067
- https://github.com/lima-vm/lima/issues/5223

pirate:
Rebased to fix conflicts, note I kept it as 2 commits for easier review but I can squash down to 1 if you prefer.

- 1. original implementation from my first PR
- 2. security hardening added on top

website/content/en/docs/config/disk.md
```diff
@@ -86,3 +86,59 @@ limactl edit default --disk 20
 > **Note:**
 > - Increasing disk size is supported, but shrinking disks is not recommended.
 > - The instance may need to be stopped before editing disk size.
+
+## Attach host block devices
+
+| ⚡ Requirement | Lima >= 2.2, macOS >= 14.0, vmType: vz |
```
AkihiroSuda: This will be in v2.3

jandubois:
I've run a single-round deep review on the current PR state: https://jandubois.github.io/lima/20260716-213905-pr-5117.html

It has one critical, and several important findings. From the results I can tell some agents would find more issues in a multi-round review. Ping me when you have addressed the important issues, and I'll re-run it in "thorough" mode...

jandubois:
I noticed that in the previous review the Gemini subagent delivered an usual poor performance. It didn't use any tools at all and worked purely from the diff included in its prompt.

I've fixed this now, and it added 2 more unique findings. I also added a Gemini-3.5-Flash reviewer, which found one more issue. Updated review at https://jandubois.github.io/lima/20260717-004757-pr-5117.html

AkihiroSuda:
Needs rebasing

website/content/en/docs/config/disk.md
```diff
@@ -86,3 +86,61 @@ limactl edit default --disk 20
...
+blockDevices:
+- /dev/disk4
+- /dev/rdisk5
+```
+{{% /tab %}}
+{{< /tabpane >}}
+
+The `--block-device` flag can be specified multiple times to attach multiple host block devices.
+
+### Sudoers setup
```
unsuman, line 124: Maybe this should come first

.github/workflows/test.yml
```diff
@@ -567,14 +567,31 @@ jobs:
...
       with:
         persist-credentials: false
+        submodules: true
     - uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e  # v7.0.0
       with:
         go-version: stable
         cache: false
     - name: Make
       run: make
+    - name: Install socket_vmnet
```
unsuman, line 577: Why do we need this change?

pkg/blockdevice/block_device_darwin.go
```diff
@@ -0,0 +1,617 @@
...
+	if err := validateSecurePathElement(path, fi, rootUID, rootGID); err != nil {
+		return err
+	}
+	if path != "/" {
+		return validateSecurePath(filepath.Dir(path), rootUID, rootGID)
+	}
+	return nil
+}
+
+func validateSecurePathElement(path string, fi os.FileInfo, rootUID, rootGID uint32) error {
```
unsuman, line 389: A root-owned binary writable via a macOS ACL might still pass? Same goes for: https://github.com/lima-vm/lima/blob/9645a065414dad826939382e277ac3c0f8edb213/pkg/networks/validate.go#L97
But I guess we can do this in a separate PR

unsuman (changes requested):
Overall the hardening looks great. Please address this comment as well: https://github.com/lima-vm/lima/pull/5483#discussion_r3955289642

pirate:
Follow-up on stale/outdated review threads after the latest force-push (`3b2c2adf`):

- `hack/test-templates.sh` `?`: removed; that cleanup was unrelated and is no longer in the PR diff.
- `hack/bats/tests/shell.bats` `?`: removed; that cleanup was unrelated and is no longer in the PR diff.
- package-global VZ fd retention: fixed by scoping normal VZ descriptors to the VM lifecycle while retaining exported krunkit/qemu wrapper fds process-wide.
- fd receive hang: fixed by making fd receive context-aware and avoiding a wait after sudo command failure.
- `FD_CLOEXEC`: fixed by duplicating received helper fds with `F_DUPFD_CLOEXEC`.
- docs version: fixed; the requirement now says Lima `>= 2.3`.
- docs ordering: fixed; sudoers setup now appears before the usage examples.
- VZ workflow `socket_vmnet`: removed from this job; the block-device BATS test installs its per-device sudoers entry after creating the RAM disk, so the extra socket_vmnet setup is not needed here.
- macOS ACL concern: fixed for the block-device helper path by rejecting extended ACL entries on each validated path component, with a real ACL regression test. I left the pre-existing network helper ACL behavior out of this PR as suggested.

cmd/limactl/sudo_open_block_device_command_darwin.go
AkihiroSuda, line 1: This should be now decoupled from limactl and moved to:

https://github.com/lima-vm/lima/blob/a85206649db4ce1f21c206f1e5426042c8bedbc7/Makefile#L321-L323

pirate: Fixed in b12000b5: block-device opening now lives in a dedicated `cmd/lima-privileged-block-device` helper built to `libexec/lima/privileged/lima-privileged-block-device`; `limactl` no longer has a hidden root helper subcommand.

unsuman (dismissed):
Thanks! Once this merges I'll refactor #5485 Linux sudoers handling into `pkg/sudoers` which will also include a sha256 check that can be used for all the binaries inside `/privileged`

AkihiroSuda:
Please squash the commits

unsuman:
Please rebase with master

jandubois:
Updated AI review: https://jandubois.github.io/lima/20260916-233839-pr-5117.html

pirate:
Fixed in 685d7f37:

- I1: runtime executable check; default-build regression replaces ineffective build-tag guard.
- I2: reject socket symlinks; authenticate listener UID before sending nonce/FD.
- I3: match complete sudoers lines; reject appended root commands.
- I4: warn explicitly when generated sudoers omit existing network grants.
- I5: regeneration hints always warn to retain every required device grant.
- I6: unique temporary BATS sudoers fragment; revoke before RAM-disk detach.
- I7: whole-disk locks across instances sharing LIMA_HOME; raw/block/partition aliases covered; limits documented.
- S1: reject duplicate block-device entries.
- S2: checked native stat assertions in the new helper callers; no shared SysStat change.
- S3–S4: document invoked-path lookup and ancestor ownership; actionable install errors.
- S5–S6: accurate template/backend examples; restore identifier-limit explanation.
- S7–S9: explain CI install prefix, gate BATS on default.yaml, clarify comment filtering, remove unrelated churn.
- FD identity: compare received device Rdev with requested node.

Regression tests added; all 33 hosted CI checks passed, including VZ block-device BATS.

unsuman:
CI (Lints) is failing:

> go: updates to go.mod needed; to update it:
> 	go mod tidy
> make: *** [Makefile:733: gomodjail] Error 1
> Error: Process completed with exit code 2.

jandubois:
Thanks for the fixes!

AI review is relentless though, and some of the Important issues deserve consideration: https://jandubois.github.io/lima/20260918-123055-pr-5117.html

jandubois:
AI claims the latest set if fixes have introduced new issues: https://jandubois.github.io/lima/20260918-211054-pr-5117.html

pirate:
Fixed in cb78c1b3 + rebased/squashed onto latest `master`:

- I1, S12: root-owned `0444` sudoers; `visudo` before install; CI checks the actual installed file, not a `sudo cat` copy.
- I2, S1, S4: helper preflight only on startup; deletion works without it; OS-version error first; helper must be a regular executable.
- I3: reject raw/block aliases and parent-child overlaps; retain disjoint partitions; regression coverage for both orders and misleading name prefixes.
- S2: no privileged-test retries masking failed revocation; retain exact-fragment cleanup, no wildcard deletion.
- S3, S5, S6: clarify per-user regeneration, preserving other users, revocation/`--check` limits; restore network-check failure warning.
- S7-S11: simplify yq/test helper, name socket centrally, stop RAM-disk parsing after one match; assert release helper contents, including the intentional Linux networking-helper packaging fix.
- S13/S14 left unchanged: early path validation catches invalid config; stop-channel refactor is pre-existing/unrelated, with device descriptors already closed before the send.

jandubois (approved):
Thanks, LGTM

AI finds one minor issue in the new test: https://jandubois.github.io/lima/20260920-124824-pr-5117.html

Fix, if you feel like it, but I would ignore all the suggestions now.

I'm not going to merge right away to: (1) give you a chance to fix I1 if you want and (2) to allow another maintainer to confirm that we can merge now

pirate:
pushed that one-line test fix and rebased back onto latest master 👍 

(let me know if you want me to rebase again after you fix that quicklinks 404 CI error on master)

unsuman (approved):
Thanks

PR #5120, opened 2026-06-16
Fix off-by-one error in 00-systemd-user-group.sh

jandubois:
The comment explicitly says UID 1..999 (system users), but the code included UID 1000 too.


AkihiroSuda (approved):
Thanks

Issue #5085, opened 2026-06-05
Fedora 44 fails to boot using `Krunkit` (`.ssh.overVsock=true`)


unsuman:
### Description

It gets stuck here:
```
➜ lm start template:fedora --vm-type krunkit --name fedora-44-krunkit
? Creating an instance fedora-44-krunkit Proceed with the current configuration
INFO[0002] Starting the instance "fedora-44-krunkit" with external VM driver "krunkit"
INFO[0002] Attempting to download the image arch=aarch64 digest="sha256:55c60a3b80d3616a08705afd0459e75fe9f03c54aba7a46e4002a41a72fa0d5b" location="[https://download.fedoraproject.org/pub/fedora/linux/releases/44/Cloud/aarch64/images/Fedora-Cloud-Base-Generic-44-1.7.aarch64.qcow2](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)"
INFO[0002] Using cache /Users/ansumansahoo/Library/Caches/lima/download/by-url-sha256/eeaf9b60fec7a86a6a35041b155e622111b5230272fb0df335ebf5ef157e6e18/data
INFO[0003] Attempting to download the nerdctl archive arch=aarch64 digest="sha256:55d68d2613b5f065021146bac21f620cde9e7fdd4bd3eff74cd324f5462e107a" location="[https://github.com/containerd/nerdctl/releases/download/v2.2.2/nerdctl-full-2.2.2-linux-arm64.tar.gz](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html)"
INFO[0003] Using cache /Users/ansumansahoo/Library/Caches/lima/download/by-url-sha256/7a4f1474af2543486c8548fd046bea7eb82f51c3eb376fb4f1896e795a7fdb91/data
INFO[0004] [hostagent] hostagent socket created at /Users/ansumansahoo/.lima/fedora-44-krunkit/ha.sock
INFO[0013] SSH Local Port: 64867
INFO[0012] [hostagent] Waiting for the essential requirement 1 of 3: "ssh"
^C
```

Logs that differ:
<details><summary>serial.log using vz(for fedora 44) and krunkit (for fedora 43)</summary>
<p>

```console
Fedora Linux 44 (Cloud Edition)
[...]

PR #5123, opened 2026-06-16
krunkit: disable ssh.overVsock by default

AkihiroSuda:
ssh.overVsock is known not to work since Fedora 44. 

Workaround for:
- #5085


unsuman (approved):
Thanks!

PR #5125, opened 2026-06-16
fix(makefile): propagate build failure for additional drivers

### LINKED ISSUE #5124: `make` does not fail on a compilation error of an additional VM driver
`make` does not fail even when `pkg/driver/krunkit/krunkit_driver_darwin_arm64.go` has a syntax error.

unsuman:
Fixes #5124


afbjorklund (approved):

AkihiroSuda (approved):
Thanks

PR #5132, opened 2026-06-18
Move the kubeadm config file to /etc

afbjorklund:
Previously it was left in the / directory (since that was the working directory).

```yaml
kind: InitConfiguration
apiVersion: kubeadm.k8s.io/v1beta4
nodeRegistration:
  criSocket: unix:///run/containerd/containerd.sock
---
kind: ClusterConfiguration
apiVersion: kubeadm.k8s.io/v1beta4
apiServer:
  certSANs: # --apiserver-cert-extra-sans
  - "127.0.0.1"
networking:
  podSubnet: "10.244.0.0/16" # --pod-network-cidr
---
kind: KubeletConfiguration
apiVersion: kubelet.config.k8s.io/v1beta1
cgroupDriver: systemd
```

### DIFF
--- a/templates/k8s.yaml
+++ b/templates/k8s.yaml
@@ -98,7 +98,7 @@ provision:
     kubeadm config images pull --cri-socket=unix:///run/containerd/containerd.sock
     systemctl start kubelet
     # Initializing your control-plane node
-    cat <<EOF >kubeadm-config.yaml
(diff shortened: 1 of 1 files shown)


unsuman (approved):
Thanks, LGTM!

PR #5137, opened 2026-06-20
fix(qemu): fallback from hvf to tcg on macOS

### LINKED ISSUE #5133: [QEMU] Support running in a macos VM using VZ (fall back from `hvf` to `tcg` automatically)
### Description

Currently, this defaults to always using hvf on macos without checking whether this works. Can the code be updated to check whether hvf works, and failback to tcg if it doesn't?

<https://github.com/lima-vm/lima/blob/5fdd875f49a5aa36e836b98374f28fe96ee53674/pkg/driver/qemu/qemu.go#L1074>

For context, I'm trying to run colima inside a macos VM that's running using vz and that's failing due to this hardcoding

kavirakesh14:
Fixes #5133

**Description:**
Currently, Lima hardcodes the `hvf` accelerator for macOS VMs. When running Lima inside a nested macOS VM environment(nested virtualization), `hvf` is unavailable, causing QEMU to immediately fail and crash.

To fix this, I separated the accelerator from the -machine string and used the dedicated -accel flag instead. Now, if the code detects hvf as the primary accelerator, it also explicitly appends -accel tcg right after it. This way, QEMU will try hardware acceleration first, but if it's unavailable (like in a nested VM), it safely falls back to software emulation instead of crashing.

**Testing:**
- Ran `go test -v ./pkg/driver/qemu/...` locally.

### DIFF
--- a/pkg/driver/qemu/qemu.go
+++ b/pkg/driver/qemu/qemu.go
@@ -1067,11 +1067,19 @@ func Exe(arch limatype.Arch) (exe string, args []string, err error) {
 	return exe, args, nil
 }
 
+var hvSupport = sync.OnceValues(func() (string, error) {
+	return osutil.Sysctl(context.TODO(), "kern.hv_support")
+})
+
 func Accel(arch limatype.Arch) string {
 	if limatype.IsNativeArch(arch) {
 		switch runtime.GOOS {
 		case "darwin":
-			// TODO: return "tcg" if HVF is not available
+			s, err := hvSupport()
+			if err != nil || s != "1" {
+				logrus.WithError(err).Warn("Hypervisor API (kern.hv_support) is not available. Disabling HVF. Expect very poor performance.")
+				return "tcg"
+			}
 			return "hvf"
 		case "linux":
 			if _, err := os.Stat("/dev/kvm"); err != nil {


afbjorklund:
There is already an `-accel` flag, so the change is in the wrong place...

The function you want is `func Accel(arch limatype.Arch) string`

afbjorklund:
The Linux version issues a warning, when falling back to emulation.

Is it possible to do that on macOS as well, to check if HVF is working?

https://developer.apple.com/documentation/hypervisor

> At runtime, determine whether the Hypervisor APIs are available on a particular machine with the sysctl command, passing `kern.hv_support` as an argument.

kavirakesh14:
Thanks @afbjorklund That is a much better approach, thank you for pointing that out! I have reverted the changes and moved the logic to the Accel() function.

pkg/driver/qemu/qemu.go
```diff
@@ -1071,7 +1071,12 @@ func Accel(arch limatype.Arch) string {
 	if limatype.IsNativeArch(arch) {
 		switch runtime.GOOS {
 		case "darwin":
-			// TODO: return "tcg" if HVF is not available
+			// DONE: return "tcg" if HVF is not available
```
AkihiroSuda: This comment can be now removed

pkg/driver/qemu/qemu.go
```diff
@@ -1071,7 +1071,12 @@ func Accel(arch limatype.Arch) string {
...
+			s, err := osutil.Sysctl(context.Background(), "kern.hv_support")
```
AkihiroSuda: - context.TODO should be used
- sync.OnceValues should be used

kavirakesh14: @AkihiroSuda, Thanks for the review and suggestion. I updated the changes. Kindly review, Thanks

AkihiroSuda (approved):
Thanks

unsuman (approved):

PR #787, opened 2022-04-08
9p: use `fscache` by default for RO mounts

AkihiroSuda:
Fix #786

PR #5140, opened 2026-06-22
mount: set 9p cache=mmap to fix file truncation

ningmingxiao:
on linux after vm start  copy data into share  dir  (for an example /home/ubuntu/containerd.zip)
on host 
```
-rw-r--r--.  1 root   root    607497541 Jun 22 13:11 containerd.zip
[root@anolis bin]# sha256sum /home/ubuntu/containerd.zip
11db316940f1b7bdb5d3d3c7d9d2a072fcc27a7506604338a72ab57d10e9a129  /home/ubuntu/containerd.zip

```
on guest 
```
[ubuntu@lima-test01 ubuntu]$ ls -al /home/ubuntu|grep containerd.zip
-rw-r--r--   1 root   root     34816000 Jun 22 13:10 containerd.zip
[ubuntu@lima-test01 ubuntu]$ sha256sum /home/ubuntu/containerd.zip 
f3682a59925a55cc9c81b047f0ba8f5883df1981ef8426b6ce6576d767b2dbe7  /home/ubuntu/containerd.zip
```
I find file is incomplete.

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -46,7 +46,7 @@ const (
 	Default9pProtocolVersion string = "9p2000.L"
 	Default9pMsize           string = "128KiB"
 	Default9pCacheForRO      string = "fscache"
-	Default9pCacheForRW      string = "mmap"
+	Default9pCache           string = "mmap"
 
 	DefaultVirtiofsQueueSize int = 1024
 )
@@ -777,11 +777,7 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 			mount.Writable = ptr.Of(false)
 		}
 		if mount.NineP.Cache == nil {
-			if *mount.Writable {
-				mounts[i].NineP.Cache = ptr.Of(Default9pCacheForRW)
-			} else {
-				mounts[i].NineP.Cache = ptr.Of(Default9pCacheForRO)
-			}
+			mounts[i].NineP.Cache = ptr.Of(Default9pCache)
 		}
 	}
 
--- a/pkg/limayaml/defaults_test.go
+++ b/pkg/limayaml/defaults_test.go
@@ -214,7 +214,7 @@ func TestFillDefault(t *testing.T) {
 	expect.Mounts[0].NineP.SecurityModel = ptr.Of(Default9pSecurityModel)
 	expect.Mounts[0].NineP.ProtocolVersion = ptr.Of(Default9pProtocolVersion)
 	expect.Mounts[0].NineP.Msize = ptr.Of(Default9pMsize)
-	expect.Mounts[0].NineP.Cache = ptr.Of(Default9pCacheForRO)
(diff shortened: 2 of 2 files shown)


pkg/limayaml/defaults.go
```diff
@@ -780,7 +780,7 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 			if *mount.Writable {
 				mounts[i].NineP.Cache = ptr.Of(Default9pCacheForRW)
 			} else {
-				mounts[i].NineP.Cache = ptr.Of(Default9pCacheForRO)
```
AkihiroSuda, line 783: The variable name mustn't be changed

AkihiroSuda:
For performance reason we chose `fscache`:
- https://github.com/lima-vm/lima/pull/787

ningmingxiao:
how to fix this issue?  @AkihiroSuda

AkihiroSuda:
> how to fix this issue? @AkihiroSuda

Just opt-in to specify cache=none in your own yaml.

Or use `limactl shell --sync`, or `limactl copy`.

ningmingxiao:
how about we don't let user see the new file ?  It will make user confused user don't understand why the file is  incomplete.
I ask gpt  it said that we can set loose vm can't see the new file. @AkihiroSuda

ningmingxiao:
I set it to mmap when readonly, it works well, set loose user still see it.

ningmingxiao:
limactl start --mount-type=virtiofs also works well for me. we can set virtiofs as default mount-type @AkihiroSuda

pkg/limayaml/defaults.go
AkihiroSuda: When I commented "The variable name mustn't be changed", I meant `Default9pSecurityModel` (which isn't even about the caching mode) should not be used for describing the default cache for RO.

Of course I didn't expect `Default9pCacheForRW` to be used for describe the default cache for RO.

Please keep the variable names consistent with the implementation.

AkihiroSuda:
> limactl start --mount-type=virtiofs also works well for me. we can set virtiofs as default mount-type @AkihiroSuda

Only after QEMU supports it on macOS.

afbjorklund:
It also requires root on Linux, so might not available (but as long as you can fall back to virtfs...)

pkg/limayaml/defaults.go
ningmingxiao: should I close this pr or set it to mmap and chage its name doesn't contain rw when use read only?

ningmingxiao:
limactl     start   --name=test03  --cpus=16   --memory=32   --mount-type=virtiofs  --vm-type=qemu --containerd none   template://fedora

ps -ef|grep virtiofsd  
```
ubuntu    245683  245681  0 Jun22 ?        00:02:29 /usr/libexec/virtiofsd --socket-path /home/ubuntu/.lima/fedora/virtiofsd-0.sock --shared-dir /home/ubuntu

```
I find it doesn't need root If I understood correctly  @afbjorklund

PR #5145, opened 2026-06-22
cidata: fix virtiofs mounts with a space in the path

### LINKED ISSUE #5136: Mounts with a space in the path silently fail to mount (unescaped fstab entry)
### Description

A mount whose host path contains a space (e.g. `/Volumes/External HD`) is never mounted in the guest, and no error is reported. The generated `/etc/fstab` entry is malformed and the nofail mount option suppresses the failure, so the mount simply does not appear.

This was surfaced downstream in Colima: https://github.com/abiosoft/colima/issues/1471 (`colima start --mount "/Volumes/External HD:w"`). External drives on macOS commonly contain spaces (External HD, Backup Drive), so this affects a fairly common setup.

**Steps to reproduce:**
1. Create a mount with a space in the path & start:
```
$ mkdir -p "/tmp/lima space test"
$ limactl start --name=spaces --vm-type=vz --mount-type=virtiofs --mount "/tmp/lima space test:w" template://default
```
2. check errors:
```
$ limactl shell spaces -- findmnt --verify
findmnt: /etc/fstab: parse error at line 4 -- ignored
```
3. check line 4 in /etc/fstab:
```
$ limactl shell spaces -- cat /etc/fstab | grep -n "space test"
4:lima-9661e8fd74e6c9ad	/tmp/lima space test	virtiofs	rw,nofail,comment=cloudconfig	0	0
```

**Expected**

The path with the space is mounted at the same location inside the guest.

**Actual**

The mount is silently missing. `df -h / findmnt` in the guest do not show it; other (space-free) mounts work fine.

**Environment**

  - limactl --version: limactl version 2.1.2
  - OS: macOS (Apple Silicon)
  - VM type: `vz` (mountType `virtiofs`)

ekalinin:
## Problem

A `vz` + `virtiofs` mount whose host path contains a space (e.g. `/Volumes/External HD`, `/tmp/lima space test`) is never mounted in the guest, and no error is reported — the mount silently does not appear. External drives on macOS commonly contain spaces, so this affects a common setup. Surfaced downstream in Colima (abiosoft/colima#1471).

## Root cause

cloud-init's `cc_mounts` writes each mount into `/etc/fstab` with a bare `"\t".join(fields)` and does not octal-escape the mount point (canonical/cloud-init#3603). A space/tab in the path produces an unparsable fstab line that `mount(8)` silently skips because of the `nofail` option, so the mount never appears:

    lima-9661e8fd74e6c9ad /tmp/lima space test    virtiofs        rw,nofail,comment=cloudconfig   0       0

(`findmnt --verify` reports `parse error at line N -- ignored`.)

  ## Fix

`cc_mounts` already creates the mount-point directory from the unescaped value, so only the fstab line needs repair. The escape/decode logic lives in two small, self-tested helper scripts under `pkg/cidata/cidata.TEMPLATE.d/util/`, following the `compare_version.sh` convention:

  - `util/escape_fstab.sh` — octal-escapes the mount-point field of cloud-config `virtiofs` entries read on stdin. Parsing with `-F'\t'` isolates the field reliably and makes the pass idempotent (an already-escaped path has no literal space/tab), so it is also forward-compatible with a fixed cloud-init.
  - `util/unescape_fstab.sh` — the inverse decode, used by the SELinux remount in `05-lima-mounts.sh` and (replacing an identical inline `sed`) by `04-persistent-data-volume.sh`.

`boot.Linux/05-lima-mounts.sh` now repairs the fstab via `escape_fstab.sh`, runs `mount -t virtiofs -a`, and decodes via `unescape_fstab.sh` before the SELinux remount. Both helpers are unit-tested via an embedded `SELFTEST` block (`SELFTEST=1 ./util/escape_fstab.sh`), like `compare_version.sh`; `.ls-lint.yml` carries the snake_case exceptions.

No Go/template changes. The fix is scoped to `vz` + `virtiofs`; it complements, but does not depend on, the upstream cloud-init fix canonical/cloud-init#6911 (distros will ship that only years from now).

## Verified

Built `limactl` and booted a `vz` + `virtiofs` instance mounting `/tmp/lima space test`:

  - `findmnt --verify` → `0 parse errors` (was `parse error at line 4`)
  - mount present; test file reads/writes round-trip host ↔ guest
  - idempotent across reboot — a single `\040`-escaped fstab line, no duplicate
  - no-space mounts unaffected; no literal `\040` directory created
  - `SELFTEST=1` passes for both helpers, and escape → unescape round-trips

Fixes: lima-vm/lima#5136
Link: https://github.com/abiosoft/colima/issues/1471

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
@@ -10,6 +10,21 @@ if ! [[ ${LIMA_CIDATA_VMTYPE} == "vz" && ${LIMA_CIDATA_MOUNTTYPE} == "virtiofs" 
 	exit 0
 fi
 
+# cloud-init's cc_mounts writes the mount point into /etc/fstab without octal-escaping it, so a
+# space/tab in the path makes an unparsable line that mount(8) silently skips via nofail.
+# cc_mounts already created the directory from the unescaped value, so just repair the fstab
+# syntax (see util/escape_fstab.sh) and (re)mount.
+# https://github.com/lima-vm/lima/issues/5136
+# https://github.com/abiosoft/colima/issues/1471
+if grep -q virtiofs /etc/fstab; then
+	escape_fstab.sh </etc/fstab >/etc/fstab.lima.tmp &&
+		cat /etc/fstab.lima.tmp >/etc/fstab && rm -f /etc/fstab.lima.tmp
+	# Mount entries cc_mounts skipped due to the previously broken line (already-mounted ones
+	# are a no-op). On Oracle Linux the virtiofs module is installed later in
+	# 30-install-packages.sh (REMOUNT_VIRTIOFS=1 remounts then), so tolerate failure here.
+	mount -t virtiofs -a || true
+fi
+
 # Update fstab entries and unmount/remount the volumes with secontext options
 # when selinux is enabled in kernel
 if [ -d /sys/fs/selinux ]; then
@@ -42,7 +57,9 @@ if [ -d /sys/fs/selinux ]; then
 			fi
 			sed -i -e "$line""s/comment=cloudconfig/comment=cloudconfig,context=\"$label\"/g" /etc/fstab
 			if [[ ${MOUNT_OPTIONS} != *"$label"* ]]; then
-				MOUNT_POINT=$(awk -v line="$line" 'NR==line {print $2}' /etc/fstab)
+				# fstab stores the mount point octal-escaped (e.g. space = "\040"); decode
+				# it before passing the path to mount(8).
+				MOUNT_POINT=$(awk -v line="$line" 'NR==line {print $2}' /etc/fstab | unescape_fstab.sh)
 				OPTIONS=$(awk -v line="$line" 'NR==line {print $4}' /etc/fstab)
 
 				#########################################################
--- a/pkg/cidata/cidata.TEMPLATE.d/util/escape_fstab.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/util/escape_fstab.sh
@@ -1,0 +1,68 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Read an /etc/fstab on stdin and write it to stdout with the mount-point field
+# octal-escaped (per fstab(5)) for cloud-config virtiofs entries whose path
+# contains a space or tab. cloud-init's cc_mounts writes the mount point verbatim,
+# so a space/tab produces an unparsable line that mount(8) silently skips via the
+# nofail option. Fields are tab-separated, so -F'\t' isolates the mount point;
+# already-escaped paths have no literal space/tab, so the transformation is
+# idempotent (and stays correct once cloud-init escapes the field itself).
+#
+# See:
+# https://github.com/lima-vm/lima/issues/5136
+# https://github.com/abiosoft/colima/issues/1471
+# https://github.com/canonical/cloud-init/issues/3603 (cc_mounts does not escape)
+# https://github.com/canonical/cloud-init/issues/6911 (the upstream cloud-init fix)
+
+set -eu
+
+: "${SELFTEST:=}"
+if [ -n "${SELFTEST}" ]; then
+	unset SELFTEST
+	tab=$(printf '\t')
+	check() {
+		local desc=$1 input=$2 want=$3 got
+		got=$(printf '%s\n' "${input}" | "$0")
+		if [ "${got}" = "${want}" ]; then
+			echo "ok: ${desc}"
+		else
+			echo "FAIL: ${desc}" >&2
+			printf '  want: %q\n  got:  %q\n' "${want}" "${got}" >&2
+			return 1
+		fi
+	}
+	echo >&2 "=== Running tests ==="
+	check "space in the mount point is escaped" \
+		"tag${tab}/tmp/dir with spaces${tab}virtiofs${tab}rw,nofail,comment=cloudconfig${tab}0${tab}0" \
+		"tag${tab}/tmp/dir\\040with\\040spaces${tab}virtiofs${tab}rw,nofail,comment=cloudconfig${tab}0${tab}0"
+	check "already-escaped path is unchanged (idempotent)" \
+		"tag${tab}/tmp/dir\\040with\\040spaces${tab}virtiofs${tab}rw,nofail,comment=cloudconfig${tab}0${tab}0" \
+		"tag${tab}/tmp/dir\\040with\\040spaces${tab}virtiofs${tab}rw,nofail,comment=cloudconfig${tab}0${tab}0"
+	check "path without whitespace is unchanged" \
+		"tag${tab}/mnt/nospace${tab}virtiofs${tab}rw,nofail,comment=cloudconfig${tab}0${tab}0" \
+		"tag${tab}/mnt/nospace${tab}virtiofs${tab}rw,nofail,comment=cloudconfig${tab}0${tab}0"
+	check "backslash is escaped before the space" \
+		"tag${tab}/a b\\c${tab}virtiofs${tab}rw,comment=cloudconfig${tab}0${tab}0" \
+		"tag${tab}/a\\040b\\134c${tab}virtiofs${tab}rw,comment=cloudconfig${tab}0${tab}0"
+	check "entry without comment=cloudconfig is unchanged" \
+		"tag${tab}/tmp/dir with spaces${tab}virtiofs${tab}rw${tab}0${tab}0" \
+		"tag${tab}/tmp/dir with spaces${tab}virtiofs${tab}rw${tab}0${tab}0"
+	check "non-virtiofs entry is unchanged" \
+		"/dev/sda1${tab}/data dir${tab}ext4${tab}defaults,comment=cloudconfig${tab}0${tab}0" \
+		"/dev/sda1${tab}/data dir${tab}ext4${tab}defaults,comment=cloudconfig${tab}0${tab}0"
+	echo >&2 "=== All tests passed ==="
+	exit 0
+fi
+
+awk -F'\t' 'BEGIN { OFS = "\t" }
+	$3 == "virtiofs" && $4 ~ /comment=cloudconfig/ && $2 ~ /[ \t]/ {
+		p = $2
+		gsub(/\\/, "\\134", p) # backslash first so introduced escapes are not re-escaped
+		gsub(/ /, "\\040", p)
+		gsub(/\t/, "\\011", p)
+		$2 = p
+	}
+	{ print }'
--- a/pkg/cidata/cidata.TEMPLATE.d/util/unescape_fstab.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/util/unescape_fstab.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright The Lima Authors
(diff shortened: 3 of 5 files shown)


hack/test-templates.sh
AkihiroSuda, line 1: New tests should be added in https://github.com/lima-vm/lima/tree/master/hack/bats

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
AkihiroSuda, line 1: Would it be possible to escape whitespaces when generating the cloud-init YAML?

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
```diff
@@ -10,6 +10,29 @@ if ! [[ ${LIMA_CIDATA_VMTYPE} == "vz" && ${LIMA_CIDATA_MOUNTTYPE} == "virtiofs"
...
 fi
 
+# cloud-init's cc_mounts writes /etc/fstab with a bare "\t".join(fields) and does NOT
+# octal-escape the mount point (canonical/cloud-init#3603); a space/tab in the path makes an
+# unparsable line that mount(8) silently skips via nofail (lima-vm/lima#5136, colima#1471).
+# cc_mounts already created the directory from the unescaped value, so only repair the fstab
+# syntax and (re)mount. -F'\t' isolates the field reliably; escaped paths have no literal
+# space/tab, so this is idempotent and safe once cloud-init is fixed (canonical/cloud-init#6911).
+if grep -q virtiofs /etc/fstab; then
+	awk -F'\t' 'BEGIN { OFS = "\t" }
```
AkihiroSuda: The awk script is hard to read. Should be split to a unit-testable helper script.
https://github.com/lima-vm/lima/tree/master/pkg/cidata/cidata.TEMPLATE.d/util

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
ekalinin: I did try the "escape the path when generating the cloud-init YAML" route first, but it unfortunately doesn't work on its own, because cc_mounts uses each mount value for two purposes:

  1. It writes the `/etc/fstab` line — escaping helps here (`.../dir\040with\040spaces` is a valid fstab field).
  2. It creates the mount-point directory via `util.ensure_dir(<value>)` — using the same string.

So if we escape the path before putting it into `mounts:`, the fstab entry becomes correct, but `ensure_dir` creates a directory literally named `dir\040with\040spaces`. `mount -a` then decodes `\040` and tries to mount onto `.../dir with spaces`, which no longer exists and the mount still fails.

It would also conflict with the upstream cloud-init fix (canonical/cloud-init#6911): that escapes the fstab field itself from the unescaped value, so a pre-escaped input would get double-escaped (`\040` --> `\134040`).

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
ekalinin: Done. I moved the awk into a dedicated, self-tested helper following the existing `util/` convention (same pattern as `util/compare_version.sh`):

hack/test-templates.sh
ekalinin: Thanks, moved.

pkg/cidata/cidata.TEMPLATE.d/util/escape-fstab.sh
AkihiroSuda, line 1: The file name convention and the unit tests are not consistent with the style of https://github.com/lima-vm/lima/blob/master/pkg/cidata/cidata.TEMPLATE.d/util/compare_version.sh .

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
```diff
@@ -42,7 +55,11 @@ if [ -d /sys/fs/selinux ]; then
 			fi
 			sed -i -e "$line""s/comment=cloudconfig/comment=cloudconfig,context=\"$label\"/g" /etc/fstab
 			if [[ ${MOUNT_OPTIONS} != *"$label"* ]]; then
-				MOUNT_POINT=$(awk -v line="$line" 'NR==line {print $2}' /etc/fstab)
+				# fstab stores the mount point octal-escaped (e.g. space = "\040"); decode it
+				# before passing the path to mount(8). Mirrors the unescape sed in
+				# boot.Linux/04-persistent-data-volume.sh.
+				MOUNT_POINT=$(awk -v line="$line" 'NR==line {print $2}' /etc/fstab |
```
AkihiroSuda: This cryptic snippet should be a separate script too

pkg/cidata/cidata.TEMPLATE.d/util/escape-fstab.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# cloud-init's cc_mounts writes the mount point verbatim with a bare
+# "\t".join(fields), so a space/tab in the path produces an unparsable line that
+# mount(8) silently skips because of the nofail option
+# (lima-vm/lima#5136, abiosoft/colima#1471, canonical/cloud-init#3603).
+#
+# Fields are tab-separated, so -F'\t' isolates the mount point even when it
+# contains a space. Already-escaped paths contain no literal space/tab, so the
+# transformation is idempotent (and stays correct once cloud-init escapes the
+# field itself: canonical/cloud-init#6911).
```
AkihiroSuda: ORG/REPO#NNNN → https://github.com/ORG/REPO/issues/NNNN

pkg/cidata/cidata.TEMPLATE.d/util/escape-fstab.sh
ekalinin: Fixed.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
ekalinin: Fixed. Moved into pkg/cidata/cidata.TEMPLATE.d/util/unescape_fstab.sh

pkg/cidata/cidata.TEMPLATE.d/util/escape-fstab.sh
ekalinin: Fixed.

pkg/cidata/cidata.TEMPLATE.d/util/escape_fstab.sh
```diff
@@ -0,0 +1,68 @@
...
+
+awk -F'\t' 'BEGIN { OFS = "\t" }
+	$3 == "virtiofs" && $4 ~ /comment=cloudconfig/ && $2 ~ /[ \t]/ {
+		p = $2
+		gsub(/\\/, "\\134", p) # backslash first so introduced escapes are not re-escaped
+		gsub(/ /, "\\040", p)
+		gsub(/\t/, "\\011", p)
+		$2 = p
+	}
+	{ print }'
```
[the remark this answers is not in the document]
ekalinin: The linter doesn't like this suggestion:

- https://github.com/lima-vm/lima/actions/runs/28196814655/job/83525668155

Besides that:
 - The reported bug (and the colima case, "External HD") is about spaces, which both versions handle identically. A tab in a directory name is pathological - essentially nobody does mkdir $'a\tb', and if they did, other parts of lima would likely choke on it too.
  - It's ~3× the code and much harder to read (field-scanning + a reconstruction loop). That's in direct tension with why we extracted the awk in the first place - make it readable.

ekalinin:
@AkihiroSuda thanks for the reviews! I think everything is addressed now:

  - the awk/sed are split into self-tested `util/escape_fstab.sh` and `util/unescape_fstab.sh`, following the `compare_version.sh` convention (`SELFTEST=1 ./… ` runs the unit tests)
  - issue references are full URLs
  - the unescape helper also de-duplicates the identical snippet in `04-persistent-data-volume.sh`

All review threads are resolved and CI is green. Whenever you have a moment, could you take another look? 🙏

AkihiroSuda:
Please squash the commits

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
```diff
@@ -10,6 +10,21 @@ if ! [[ ${LIMA_CIDATA_VMTYPE} == "vz" && ${LIMA_CIDATA_MOUNTTYPE} == "virtiofs"
...
 
+# cloud-init's cc_mounts writes the mount point into /etc/fstab without octal-escaping it, so a
+# space/tab in the path makes an unparsable line that mount(8) silently skips via nofail.
+# cc_mounts already created the directory from the unescaped value, so just repair the fstab
+# syntax (see util/escape_fstab.sh) and (re)mount.
+# https://github.com/lima-vm/lima/issues/5136
+# https://github.com/abiosoft/colima/issues/1471
+if grep -q virtiofs /etc/fstab; then
+	escape_fstab.sh </etc/fstab >/etc/fstab.lima.tmp &&
+		cat /etc/fstab.lima.tmp >/etc/fstab && rm -f /etc/fstab.lima.tmp
```
jandubois, line 21: Why not use `cp`?

```suggestion
		cp /etc/fstab.lima.tmp /etc/fstab && rm -f /etc/fstab.lima.tmp
```

pkg/cidata/cidata.TEMPLATE.d/util/unescape_fstab.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# SPDX-FileCopyrightText: Copyright The Lima Authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Read a string on stdin and write it to stdout with the octal escapes used in
+# /etc/fstab and /proc/mounts decoded to their literal characters:
+# "\040" -> space, "\011" -> tab, "\012" -> newline, "\134" -> backslash,
+# "\043" -> "#". The inverse of util/escape_fstab.sh.
```
jandubois, line 9: "inverse" is misleading, as "unescape" also converts `\012` (newline) and `\043` ("#"), which "escape" is not generating.

This is correct behaviour but not exactly the "inverse".

jandubois (approved):
Thanks, LGTM

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
ekalinin: Good point! No good reason: `cp` works identically here and is clearer, so it's worth switching.

Since that PR is already merged, I'll open a follow-up PR to change `cat` to `cp` -- any objections?

pkg/cidata/cidata.TEMPLATE.d/util/unescape_fstab.sh
ekalinin: Good catch -- "inverse" is imprecise. 
I'll reword the comment, folding it into the same follow-up PR as the `cat --> cp` change. If you don't mind.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-mounts.sh
ekalinin: fixed https://github.com/lima-vm/lima/pull/5197

pkg/cidata/cidata.TEMPLATE.d/util/unescape_fstab.sh
ekalinin: Fixed https://github.com/lima-vm/lima/pull/5197

PR #5147, opened 2026-06-23
refactor: make sudo optional for copyToHost

Aneesh-Hegde:
part of: #5109 
`copyToHost` previously always used `sudo cat` to read guest files. This drops sudo from the default path attempting `env LANG=C LC_ALL=C cat` first, and only falling back to `sudo cat` if the file returns a permission denied error.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -1230,10 +1230,7 @@ func copyToHost(ctx context.Context, sshConfig *ssh.SSHConfig, sshAddress string
 		"-p", strconv.Itoa(sshPort),
 		sshAddress,
 		"--",
-	)
-	args = append(args,
-		"sudo",
-		"cat",
+		"env", "LANG=C", "LC_ALL=C", "cat",
 		remote,
 	)
 	logrus.Infof("Copying config from %s to %s", remote, local)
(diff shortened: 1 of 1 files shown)


pkg/hostagent/hostagent.go
```diff
@@ -1224,18 +1224,25 @@ func isDeactivatedCloudInitMainService(line string) bool {
...
 		"-p", strconv.Itoa(sshPort),
 		sshAddress,
 		"--",
 	)
-	args = append(args,
-		"sudo",
-		"cat",
-		remote,
-	)
+	if needsSudo {
```
AkihiroSuda: You do not need this flag.
You can just attempt `LANG=C LC_ALL=C cat` without sudo, and fall back to `sudo cat` on  an error with `Permission denied`.

afbjorklund:
It could be useful with a "sudo" flag, though. I had the same need, for the ssh helper.

* https://github.com/lima-vm/sshocker/issues/87

Aneesh-Hegde:
@AkihiroSuda Should I go with the `LANG=C LC_ALL=C cat` fallback approach, or would it make more sense to add a proper sudo flag to sshocker first (as mentioned in sshocker [#87 ](https://github.com/lima-vm/sshocker/issues/87)) and use that here?

AkihiroSuda:
Addition of `needsSudo: true` is a breaking change, so I'm not in favor of it.

https://github.com/lima-vm/sshocker/issues/87 can be discussed separately.

AkihiroSuda (approved):
Thanks

PR #5151, opened 2026-06-24
feat: log all host-side sudo invocations with non-debug level

### LINKED ISSUE #5114: Every invocation of `sudo` on the host has to be logged
https://github.com/lima-vm/lima/blob/f6b7a0cca0abb55a16b9ba0a077622d8b809b9ae/cmd/limactl/autostart_darwin.go#L150-L158
Introduced in:
- https://github.com/lima-vm/lima/pull/4984

The invocation of `sudo` on the host has to be logged with non-debug level.

TODO: find more occurrences of `sudo`

Nachiket-Roy:
#### Summary : 

- Refactored all log levels to use logrus.Infof so that all host-side executions of sudo are logged at a non-debug level.
- Verified that the updated code builds successfully and all unit tests pass.

fixes #5114


AkihiroSuda (approved):
Thanks

PR #5155, opened 2026-06-25
fix port-forward tunnel Read exceeding the supplied buffer

SABITHSAHEB:
1. GrpcClientRW.Read and GRPCServerRW.Read copy a received message with copy(p, in.Data) but return len(in.Data), so when a message is larger than the read buffer the returned count is larger than len(p).
2. the io.Copy inside tcpproxy then slices buf[0:n] past the buffer and panics, and the leftover bytes are dropped; a UDP datagram up to 65507 bytes forwarded to a guest reaches this since tcpproxy reads with a 32 KiB buffer.
Return the copied count and keep the remainder for the next Read so nothing is lost.

### DIFF
--- a/pkg/portfwd/client.go
+++ b/pkg/portfwd/client.go
@@ -77,6 +77,10 @@ type GrpcClientRW struct {
 	protocol string
 	stream   api.GuestService_TunnelClient
 	cancel   context.CancelFunc
+
+	// rxBuf holds bytes received from the stream that did not fit in the
+	// buffer passed to the previous Read call.
+	rxBuf []byte
 }
 
 var _ net.Conn = (*GrpcClientRW)(nil)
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
How to test?

SABITHSAHEB:
Two levels.

Unit tests are in the PR: `go test ./pkg/portfwd/... ./pkg/portfwdserver/...`. `TestGrpcClientRWReadShortBuffer` / `TestGRPCServerRWReadShortBuffer` push a 100-byte `TunnelMessage` through a 10-byte `Read` buffer and assert every `Read` returns `n <= len(buf)` and the payload reassembles with no loss. Against the old code the first `Read` returns 100, which is the defect.

End to end it's a UDP datagram larger than 32 KiB heading into the guest. Run a UDP listener in the guest, forward the port, and from the host send an oversized datagram to it:

```
# guest:  nc -u -l 9999 > /tmp/out
# host:   head -c 40000 /dev/urandom | nc -u -w1 127.0.0.1 9999
```

The host UDP proxy reads with a 65507-byte buffer (gvisor-tap-vsock `UDPBufSize`) and forwards the whole datagram as a single `TunnelMessage`. On the guest, tcpproxy's `io.Copy` reads `GRPCServerRW` with a 32 KiB buffer, so the old `Read` returns `len(in.Data) > len(p)`; `io.Copy` then slices `buf[0:n]` out of range and the guest agent panics, dropping the tail. With the patch the remainder is held in `rxBuf` for the next `Read` and the datagram arrives intact.

AkihiroSuda (approved):
Thanks

PR #5156, opened 2026-06-25
feat: support minimal image variant for LTS templates

Nachiket-Roy:
### Summary
Add support for Ubuntu minimal cloud images without increasing the number of templates. This change introduces a variant-based image selection mechanism that allows users to opt into Ubuntu minimal images while continuing to use the existing Ubuntu templates. The default behavior remains unchanged. The minimal images provide a smaller download and guest footprint compared to the regular Ubuntu images, while still working with the standard Lima provisioning flow.
The change also includes updates to template generation and test coverage to ensure both standard and minimal Ubuntu images continue to work correctly.
### Manual Verification
Users can opt into the minimal image variant by overriding imageOpts.variant during creation:
```bash
limactl create template:ubuntu --set ".imageOpts.variant=\"minimal\""
limactl start ubuntu
```
fixes #5153

PR #5159, opened 2026-06-27
feat(template): support template-less image variant selection

### LINKED ISSUE #5153: Create support for the ubuntu-minimal images
### Description

The regular ubuntu images got really big, but there is a "minimal" version...

https://cloud-images.ubuntu.com/minimal/releases/

Hopefully it can be done without having to increase the number of **templates**.

Note: the minimal images lacks e.g. `git` and `vim`

### LINKED ISSUE #5158: Generic variant tag for image entries
### Description

In #5156 , we added `ubuntu-minimal-VERSION.yaml` as separate templates alongside the existing `ubuntu-VERSION.yaml` ones. That works, but several other distros publish similar variants and would hit the same template-duplication problem if we did the same thing for them:

- Debian publishes `genericcloud`, `generic`, and `nocloud` images for the same version/arch
- openSUSE publishes a standard cloud image and a smaller minimal one

Rather than creating a separate template file for every distro × version × variant combination, I'd like to propose adding a simple variant field directly on each image entry:
```yaml
images:
- location: "https://cloud-images.ubuntu.com/releases/.../ubuntu-24.04-server-cloudimg-amd64.img"
  arch: "x86_64"
  variant: "server"
- location: "https://cloud-images.ubuntu.com/minimal/.../ubuntu-24.04-minimal-cloudimg-amd64.img"
  arch: "x86_64"
  variant: "minimal"
```
The idea is to keep this OS-agnostic and attach the variant directly to each Image entry, similar to how arch is specified today, rather than introducing an OS-specific options structure.
Initially, the field could simply serve as metadata/documentation, but it could also provide a foundation for future generic variant selection if we decide that reducing template duplication across distros is worthwhile.

Nachiket-Roy:
### Description

This PR adds support for image variants (for example, Ubuntu Minimal or Debian NoCloud) without introducing additional template files such as `ubuntu-minimal.yaml` or `debian-nocloud.yaml`.

A new `variant` field is added to image entries, allowing users to select specific image variants dynamically with the existing `--set` flag.

To preserve the current behavior, standard/server images are listed before minimal variants, ensuring that default template usage continues to select standard images.

### How to use

Start Ubuntu 24.04 Minimal:

```bash
llimactl start --image-variant=minimal template:ubuntu-24.04

```

Start Debian 12 NoCloud:

```bash
limactl start --image-variant=nocloud template:debian-12
```

### Tests

Added `TestDefaultSelectedImageIsStandard` to verify that standard images remain the default selection by ensuring they are ordered before minimal variants.

Fixes #5153 
Fixes #5158

### DIFF
--- a/cmd/limactl/editflags/editflags.go
+++ b/cmd/limactl/editflags/editflags.go
@@ -105,6 +105,8 @@ func RegisterCreate(cmd *cobra.Command, commentPrefix string) {
 		return []string{"x86_64", "aarch64", "riscv64", "armv7l", "s390x", "ppc64le"}, cobra.ShellCompDirectiveNoFileComp
 	})
 
+	flags.String("image-variant", "", commentPrefix+"Image variant")
+
 	flags.String("containerd", "", commentPrefix+"containerd mode (user, system, user+system, none)")
 	_ = cmd.RegisterFlagCompletionFunc("containerd", func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
 		return []string{"user", "system", "user+system", "none"}, cobra.ShellCompDirectiveNoFileComp
@@ -374,6 +376,22 @@ func YQExpressions(flags *flag.FlagSet, newInstance bool, params map[string]stri
 			false,
 		},
 		{"ssh-port", d(".ssh.localPort = %s"), false, false},
+		{
+			"image-variant",
+			func(_ *flag.Flag) ([]string, error) {
+				variant, err := flags.GetString("image-variant")
+				if err != nil {
+					return nil, err
+				}
+				if variant == "" {
+					return nil, errors.New("`--image-variant` must not be empty")
+				}
+				expr := fmt.Sprintf(`.images = [.images[] | select(.variant == %q)]`, variant)
+				return []string{expr}, nil
+			},
+			true,
+			false,
+		},
 		{"arch", d(".arch = %q"), true, false},
 		{
 			"containerd",
--- a/pkg/driver/vz/vz_driver_darwin.go
+++ b/pkg/driver/vz/vz_driver_darwin.go
@@ -299,7 +299,7 @@ func validateConfig(_ context.Context, cfg *limatype.LimaYAML) error {
 	}
 
 	for i, image := range cfg.Images {
-		if unknown := reflectutil.UnknownNonEmptyFields(image, "File", "Kernel", "Initrd"); len(unknown) > 0 {
+		if unknown := reflectutil.UnknownNonEmptyFields(image, "File", "Kernel", "Initrd", "Variant"); len(unknown) > 0 {
 			logrus.Warnf("vmType %s: ignoring images[%d]: %+v", *cfg.VMType, i, unknown)
 		}
 	}
--- a/templates/_images/ubuntu-26.04.yaml
+++ b/templates/_images/ubuntu-26.04.yaml
@@ -4,41 +4,72 @@ images:
 - location: "https://cloud-images.ubuntu.com/releases/resolute/release-20260627/ubuntu-26.04-server-cloudimg-amd64.img"
   arch: "x86_64"
   digest: "sha256:3ee4f67f322abb2d1d1f0fffc957f7411404ad6635dd35b026c8ff05ac6e534c"
+  variant: "server"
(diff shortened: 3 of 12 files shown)


AkihiroSuda:
>  --set '.images = [.images[] | select(.variant == "minimal")]' \

There should be `--image-variant=STRING` in https://github.com/lima-vm/lima/blob/4d1c1b2ae24e19c02bfce8e8e2960255cd724b65/cmd/limactl/editflags/editflags.go#L99

templates/_images/ubuntu-26.04.yaml
```diff
@@ -4,41 +4,73 @@ images:
 - location: "https://cloud-images.ubuntu.com/releases/resolute/release-20260520/ubuntu-26.04-server-cloudimg-amd64.img"
   arch: "x86_64"
   digest: "sha256:dced94c031cc1f23dee14419a3723a5b110df9938de0ac31913a2bfd07c755b4"
+  variant: "server"
```
AkihiroSuda, line 7: When it fails to download the standard `server` image due to a transient network failure, it should avoid falling back to `minimal`, to avoid non-deterministic behavior?

cmd/limactl/editflags/editflags.go
```diff
@@ -105,6 +105,11 @@ func RegisterCreate(cmd *cobra.Command, commentPrefix string) {
 		return []string{"x86_64", "aarch64", "riscv64", "armv7l", "s390x", "ppc64le"}, cobra.ShellCompDirectiveNoFileComp
 	})
 
+	flags.String("image-variant", "", commentPrefix+"Image variant")
+	_ = cmd.RegisterFlagCompletionFunc("image-variant", func(*cobra.Command, []string, string) ([]string, cobra.ShellCompDirective) {
+		return []string{"server", "minimal", "genericcloud", "generic", "nocloud"}, cobra.ShellCompDirectiveNoFileComp
```
AkihiroSuda: This seems confusing, as the available variants depends on the template.

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -77,6 +77,7 @@ var knownYamlProperties = []string{
 	"TimeZone",
 	"UpgradePackages",
 	"User",
+	"Variant",
```
AkihiroSuda: Doesn't seem needed here

cmd/limactl/editflags/editflags.go
Nachiket-Roy: Thanks for pointing that! i will try to add a dynamic auto suggestion

pkg/driver/vz/vz_driver_darwin.go
Nachiket-Roy: If we don't keep it then WSL2 or VZ will give warning in terminal regarding `variant` field missing.

AkihiroSuda: The current PR has `.[]images.variant`, but no top-level `.variant`, no?

cmd/limactl/editflags/editflags.go
Nachiket-Roy: Will add more complexity so droping auto suggestion completely

pkg/driver/vz/vz_driver_darwin.go
Nachiket-Roy: Yes there is no top level variant so i can keep it in image level. Thanks again!

AkihiroSuda (dismissed):
Thanks, doesn't this need any change on `hack/update-template*.sh`?

Nachiket-Roy:
https://github.com/lima-vm/lima/pull/5156#discussion_r3481999468
**TL;DR:** No changes needed. The existing `--minimal` flag already supports the new `ubuntu-*-minimal.yaml` templates. Verified successfully with Ubuntu 22.04, 24.04, and 26.04, so the script was left unchanged.

unsuman (changes requested):
Looks good but needs to be documented inside https://github.com/lima-vm/lima/blob/master/templates/default.yaml

AkihiroSuda:
Please squash the commits

afbjorklund (approved):

unsuman (approved):
Thanks!

PR #5138, opened 2026-06-20
Support Windows server 2025 guest

mie313:
This PR is part of #4852.
It supports Windows server 2025 OS as an initial implementation.

It also adds a property to enable TPM emulation following #5079 by @Shardz4. (Thank you so much 🙏 )

PR #5164, opened 2026-06-28
Support Windows 11 x86-64

### LINKED ISSUE #4852: Support for Windows guests

mie313:
This PR implements Windows 11 guest. Currently, it only works on x86-64 architecture.

~Since the branch is created from the branch of #5138 , only the second commit is for Windows 11 implementation. Once #5138 is merged, this PR will be tidied up.~
[Update]
This PR was cleaned up.

### DIFF
--- a/pkg/cidata/template.go
+++ b/pkg/cidata/template.go
@@ -10,12 +10,15 @@ import (
 	"fmt"
 	"io/fs"
 	"path"
+	"path/filepath"
+	"strings"
 
 	"github.com/sethvargo/go-password/password"
 
 	"github.com/lima-vm/lima/v2/pkg/identifiers"
 	"github.com/lima-vm/lima/v2/pkg/iso9660util"
 	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
 	"github.com/lima-vm/lima/v2/pkg/textutil"
 )
 
@@ -28,6 +31,10 @@ const templateFSRoot = "cidata.TEMPLATE.d"
 var windowsTemplateFS embed.FS
 
 const windowsTemplateFSRoot = "wincidata.TEMPLATE.d"
+
+// This is for checking whether Windows OS is 11 or server 2025.
+// For example, Windows 11 x86-64's label is CCCOMA_X64FRE_EN-US_DV9.
+const windowsClientISOLabelPrefix = "CCCOMA_"
 
 type CACerts struct {
 	RemoveDefaults *bool
@@ -126,6 +133,8 @@ type TemplateArgs struct {
 	NoCloudInit                     bool
 	WindowsInitialPassword          string
 	LegacyBIOS                      bool
+	IsWindowsServer                 bool
+	TPM                             bool
 }
 
 func (t *TemplateArgs) generateWindowsInitialPassword() error {
@@ -137,6 +146,19 @@ func (t *TemplateArgs) generateWindowsInitialPassword() error {
 	}
 
 	t.WindowsInitialPassword = pw
+	return nil
+}
+
+// checkWindowsVersion checks if a guest VM is Windows 11 (true) or Windows server 2025 (false).
+func (t *TemplateArgs) checkWindowsVersion(instDir string) error {
+	imagePath := filepath.Join(instDir, filenames.ISO)
+	label, err := iso9660util.Label(imagePath)
+	if err != nil {
+		return fmt.Errorf("failed to get ISO label: %w", err)
+	}
+
+	t.IsWindowsServer = !strings.HasPrefix(label, windowsClientISOLabelPrefix)
+
 	return nil
 }
 
--- a/pkg/cidata/wincidata.TEMPLATE.d/autounattend.xml
+++ b/pkg/cidata/wincidata.TEMPLATE.d/autounattend.xml
@@ -6,22 +6,22 @@
         xmlns:wcm="http://schemas.microsoft.com/WMIConfig/2002/State">
         <DriverPaths>
             <PathAndCredentials wcm:action="add" wcm:keyValue="1">
-                <Path>E:\viostor\2k25\amd64</Path>
+                <Path>E:\viostor\{{ if .IsWindowsServer }}2k25{{ else }}w11{{ end }}\amd64</Path>
             </PathAndCredentials>
             <PathAndCredentials wcm:action="add" wcm:keyValue="2">
-                <Path>E:\vioinput\2k25\amd64</Path>
+                <Path>E:\vioinput\{{ if .IsWindowsServer }}2k25{{ else }}w11{{ end }}\amd64</Path>
             </PathAndCredentials>
             <PathAndCredentials wcm:action="add" wcm:keyValue="3">
-                <Path>E:\NetKVM\2k25\amd64</Path>
+                <Path>E:\NetKVM\{{ if .IsWindowsServer }}2k25{{ else }}w11{{ end }}\amd64</Path>
             </PathAndCredentials>
             <PathAndCredentials wcm:action="add" wcm:keyValue="4">
-                <Path>E:\viofs\2k25\amd64</Path>
+                <Path>E:\viofs\{{ if .IsWindowsServer }}2k25{{ else }}w11{{ end }}\amd64</Path>
             </PathAndCredentials>
             <PathAndCredentials wcm:action="add" wcm:keyValue="5">
-                <Path>E:\Balloon\2k25\amd64</Path>
+                <Path>E:\Balloon\{{ if .IsWindowsServer }}2k25{{ else }}w11{{ end }}\amd64</Path>
             </PathAndCredentials>
             <PathAndCredentials wcm:action="add" wcm:keyValue="6">
-                <Path>E:\viorng\2k25\amd64</Path>
+                <Path>E:\viorng\{{ if .IsWindowsServer }}2k25{{ else }}w11{{ end }}\amd64</Path>
             </PathAndCredentials>
         </DriverPaths>
     </component>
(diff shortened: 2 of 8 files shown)


pkg/cidata/wincidata.TEMPLATE.d/autounattend.xml
```diff
@@ -0,0 +1,195 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mie313, line 1: lima needs to build autounattend.xml based on not only an architecture but also Windows OS version (11 or Server 2025) because of the following reasons:
- Drivers' paths for Windows 11 in Virtio-win are different from paths for Server 2025
- There are some Windows 11 specific configurations. For example, line 41~64 aren't necessary in Server 2025
- Image index is different

lima can check a guest architecture through a yaml file's `arch` field. However, I don't know how to check a guest OS version.

pkg/driver/qemu/qemu_driver.go
```diff
@@ -413,6 +413,20 @@ func (l *LimaQemuDriver) Start(ctx context.Context) (chan error, error) {
 			}
 		}
 	}()
+
+	// QEMU's UEFI boot order doesn't work well for Windows guest OS.
```
mie313, line 428: This is the workaround which we talked in the slack channel.
Currently it just repeats sending a return key and sleeping 1 seconds.

pkg/driver/qemu/qemu_driver.go
```diff
@@ -467,6 +481,37 @@ func checkBinarySignature(ctx context.Context, vmArch string) error {
...
+		return err
+	}
+	qmpClient, err := qmp.NewSocketMonitor("unix", qmpSockPath, 5*time.Second)
+	if err != nil {
+		return err
+	}
+	if err := qmpClient.Connect(); err != nil {
+		return err
+	}
+	defer func() { _ = qmpClient.Disconnect() }()
```
AkihiroSuda, line 567: Connection and disconnection should happen outside the `for range 5` loop in the caller

pkg/driver/qemu/qemu_driver.go
```diff
@@ -413,6 +413,20 @@ func (l *LimaQemuDriver) Start(ctx context.Context) (chan error, error) {
...
 	}()
+
+	// QEMU's UEFI boot order doesn't work well for Windows guest OS.
+	// We send a key periodically to proceed from "Press any key to boot from CD or DVD...".
+	if *qCfg.LimaYAML.OS == limatype.WINDOWS && !*qCfg.LimaYAML.Firmware.LegacyBIOS {
+		logrus.Debug("send a key to QEMU to boot Windows ISO...")
+		// TODO: Ideally, we should check the boot status and break earlier.
+		for range 5 {
+			if err := l.sendEnter(); err != nil {
+				logrus.Infof("%s", err.Error())
```
AkihiroSuda:
```suggestion
				logrus.WithError(err).Error("failed to send a keypress event")
```

pkg/cidata/template.go
```diff
@@ -29,6 +32,10 @@ var windowsTemplateFS embed.FS
 
 const windowsTemplateFSRoot = "wincidata.TEMPLATE.d"
 
+// This is for checking whether Windows OS is 11 or server 2025.
+// For example, Windows 11 x86-64's label is CCCOMA_X64FRE_EN-US_DV9.
+const windows11ISOLabelPrefix = "CCCOMA_"
```
AkihiroSuda: This prefix actually appeared in Windows 10, not 11

pkg/cidata/template.go
```diff
@@ -126,6 +133,8 @@ type TemplateArgs struct {
 	NoCloudInit                     bool
 	WindowsInitialPassword          string
 	LegacyBIOS                      bool
+	IsWindows11                     bool
```
AkihiroSuda: This naming scheme will not work well after the release of Windows 12, 13, 14, ...

AkihiroSuda: Probably better to invert the logic to check whether it is a server:

```suggestion
	IsWindowsServer                     bool
```

templates/windows-11.yaml
```diff
@@ -0,0 +1,32 @@
...
+# You can turn off TPM (in that case, lima bypasses Windows 11's initial TPM check).
+tpm: true
+
+osOpts:
+  Windows:
+    # Virtio-win ISO contains virtio drivers for windows. It is necessary for lots of QEMU options which use virtio (virtio interface for mounting qcow2 disk, virtio-net, virtio-serial, virtio-rng, virtio-keyboard, and virtio-mouse)
+    # In addition, the driver for virtio-fs will be used in the future.
+    # Virtio-win ISO supports both x86_64 and aarch64 (inside the ISO it contains drivers for both archs).
+    virtioWin:
+    - location: https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/archive-virtio/virtio-win-0.1.271-1/virtio-win-0.1.271.iso
```
AkihiroSuda, line 23: Needs sha256 digest

templates/windows-11.yaml
```diff
@@ -0,0 +1,32 @@
+# Windows guest is currently experimental. It only runs on QEMU, and it requires plain: true.
+os: "Windows"
+vmType: "qemu"
+arch: "x86_64"
+
+images:
+# Currently there is no way to download Windows 11 ISO automatically.
+# You have to download the ISO from https://www.microsoft.com/en-us/software-download/windows11 and set the local path here.
+# Please download `English (United States)` version. Otherwise VM installation is stopped in the middle.
+- location: "./images/windows-11-us.iso"
```
AkihiroSuda: Relative paths are not expected

```suggestion
- location: "~/Downloads/Win11_25H2_English_x64_v2.iso"
```

website/content/en/docs/usage/guests/windows.md
```diff
@@ -3,24 +3,28 @@ title: Windows
...
 |-------------------|-----------------------------|
 
 Running Windows guests is experimentally supported since Lima v2.2.
 
 ```bash
-limactl start template:windows-2025
+limactl start template:windows
 ```
 
+The command starts Windows 11 VM. You can also choose Windows server 2025 (template:windows-2025).
```
AkihiroSuda: You can use a tab pane

https://github.com/lima-vm/lima/blob/67b84c58e301fc21341ddac974e9d4bb1345cb45/website/content/en/docs/usage/guests/freebsd.md?plain=1#L11-L22

mie313: Thank you. I modified is as follows:
https://github.com/lima-vm/lima/pull/5164/changes#diff-a5573d7c3e9019dc489f10fe11e3d5eb6fad3b8290a68ed162793ab0ae938b29R11-R22

templates/windows-11.yaml
mie313: Modified it as follows:
https://github.com/lima-vm/lima/pull/5164/changes#diff-331a443cde01025bcba4fece25d592446fe20277e06dd342278a7db62c70c23bR10

templates/windows-11.yaml
mie313: I put sha256 digest calculated by me:
https://github.com/lima-vm/lima/pull/5164/changes#diff-331a443cde01025bcba4fece25d592446fe20277e06dd342278a7db62c70c23bR25-R28

I modified windows-2025.yaml as well:
https://github.com/lima-vm/lima/pull/5164/changes#diff-cb20619b6ec4a81aaf0e1e1fd5782fe319921f357133121c78e07d5c4a280f93R22

I'm sorry, I overlooked your comment in https://github.com/lima-vm/lima/pull/5138#discussion_r3497067272

pkg/cidata/template.go
mie313: Thanks you. I modified it as follows:
https://github.com/lima-vm/lima/pull/5164/changes#diff-c751ec359c364110d9477457172256213a6b002cc1855a0ff9c551b9cf2b6d5eR136

templates/windows-2025.yaml
```diff
@@ -19,6 +19,7 @@ osOpts:
     virtioWin:
     - location: https://fedorapeople.org/groups/virt/virtio-win/direct-downloads/archive-virtio/virtio-win-0.1.271-1/virtio-win-0.1.271.iso
       arch: "x86_64"
+      digest: "sha256:bbe6166ad86a490caefad438fef8aa494926cb0a1b37fa1212925cfd81656429"
```
AkihiroSuda: This could be a separate PR, as it does not relate to Windows 11

pkg/cidata/template.go
mie313: I searched on Google and found web pages referring both Windows 10 and 11, although the number of Windows 10 pages looked greater than the number of Windows 11 pages.
In addition, I confirmed that the Windows 11 x86-64 ISO has the label `CCCOMA_X64FRE_EN-US_DV9`. 
I think Microsoft reuses the name.

Do you think we should not use the prefix for distinguishing Windows OS version? I assume that Microsoft would reuse the name for Windows 12, 13, 14...

templates/windows-2025.yaml
mie313: Thank you. I reverted the change. I will create another PR for windows-2025.yaml.

pkg/cidata/template.go
AkihiroSuda: You can still use it to detect whether it is a server or a client edition.
```suggestion
const windowsClientISOLabelPrefix = "CCCOMA_"
```

or

```suggestion
const windowsServerISOLabelPrefix = "<SERVER PREFIX GOES HERE>"
```

mie313: Thank you. I took the first option.
https://github.com/lima-vm/lima/pull/5164/changes#diff-c751ec359c364110d9477457172256213a6b002cc1855a0ff9c551b9cf2b6d5eR37

AkihiroSuda (approved):
Thanks

PR #5180, opened 2026-07-01
hack: fail early when host prerequisites are missing

### LINKED ISSUE #5038: Test scripts such as `hack/test-template.sh` should fail early when prerequisites (`timeout`, `socat`, etc.) are not installed on the host

ekalinin:
test-templates.sh boots a VM before invoking host tools such as nc, socat, dig, curl, jq and timeout (directly and via test-port-forwarding.pl). When one of them is missing, the script failed only after a full create + boot cycle, wasting minutes on a confusing late error.

Check the host commands required by the enabled checks up front, before `limactl create`, and abort while listing every missing command. The list is gated by the enabled checks, so a tool is only required when the check that uses it runs. Add a reusable check_required_commands() helper to common.inc.sh. Commands that are optional at runtime (rsync, w3m) are intentionally not required.

Fixes #5038

### DIFF
--- a/hack/common.inc.sh
+++ b/hack/common.inc.sh
@@ -18,6 +18,21 @@ function ERROR() {
 	echo >&2 "TEST| [ERROR] $*"
 }
 
+# check_required_commands CMD...
+# Exits 1 (listing every missing command) if any required host command is absent.
+function check_required_commands() {
+	local missing=()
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #5190, opened 2026-07-05
networks: avoid %everyone in default sudoers

### LINKED ISSUE #5118: `limactl sudoers`:  consider avoid allowing `%everyone` by default
Currently, `limactl sudoers` allows `%everyone` to run `socket_vmnet` as the `root:wheel` by default
```
%everyone ALL=(root:wheel) NOPASSWD:NOSETENV: /bin/mkdir -m 775 -p /private/var/run/lima

# Manage "bridged" network daemons

%everyone ALL=(root:wheel) NOPASSWD:NOSETENV: \
    /opt/socket_vmnet/bin/socket_vmnet --pidfile=/private/var/run/lima/bridged_socket_vmnet.pid --socket-group=everyone --vmnet-mode=bridged --vmnet-interface=en0 /private/var/run/lima/socket_vmnet.bridged, \
    /usr/bin/pkill -F /private/var/run/lima/bridged_socket_vmnet.pid

# ...
```
https://github.com/lima-vm/lima/blob/29aeed2a9794ff23b7342cb40a84df312e2d2ebb/pkg/networks/networks.go#L8-L12

Shouldn't it just allow the current user?

ekalinin:
`limactl sudoers` and the generated `networks.yaml` defaulted `group` to `everyone`, which on macOS grants every account (including non-local / directory accounts) permission to run `socket_vmnet` as `root:wheel` and to access the vmnet socket. For the common single-user workstation this is broader than necessary.

Change the default to `admin`. It is tighter than `everyone` (and `staff`), and it matches the realistic audience: installing the sudoers file already requires `sudo`, which on macOS is granted to admins by default, so `%admin` covers exactly the users who can use this feature. A non-admin user must set `group` to a group they belong to (e.g. `staff`).

Existing `networks.yaml` files are not rewritten, so current installations keep their `everyone` value; the new default only applies to fresh setups.

Proper per-user scoping is deferred until `socket_vmnet` gains `--socket-user=USER`.

Closes #5118

### DIFF
--- a/pkg/networks/networks.go
+++ b/pkg/networks/networks.go
@@ -7,7 +7,7 @@ import "net"
 
 type Config struct {
 	Paths    Paths              `yaml:"paths" json:"paths"`
-	Group    string             `yaml:"group,omitempty" json:"group,omitempty"` // default: "everyone"
+	Group    string             `yaml:"group,omitempty" json:"group,omitempty"` // default: "admin"
 	Networks map[string]Network `yaml:"networks" json:"networks"`
 }
 
--- a/website/content/en/docs/config/network/vmnet.md
+++ b/website/content/en/docs/config/network/vmnet.md
@@ -103,7 +103,10 @@ paths:
   varRun: /private/var/run/lima
   sudoers: /private/etc/sudoers.d/lima
 
-group: everyone
+# The group that is allowed to run socket_vmnet via sudo and to access the
+# socket_vmnet socket (macOS only; ignored on other platforms). Non-admin users
+# must set this to a group they belong to (e.g. "staff").
+group: admin
 
 networks:
   shared:
@@ -163,6 +166,20 @@ less etc_sudoers.d_lima  # verify that the file looks correct
 sudo install -o root etc_sudoers.d_lima /etc/sudoers.d/lima
 rm etc_sudoers.d_lima
 ```
+
+The `sudoers` file grants the members of the `group` configured in `networks.yaml`
+permission to run `socket_vmnet` as root. Since Lima v2.2, `group` defaults to
+`admin` instead of `everyone`, so the permission is no longer granted to non-admin
+and non-local accounts. A non-admin user must set `group` to a group they belong
+to (e.g. `staff`).
+
+{{% alert title="Note" color="warning" %}}
+Existing installations keep whatever `group` is already written in
+`$LIMA_HOME/_config/networks.yaml` (the file is never rewritten once it exists).
+To narrow an existing `group: everyone`, edit the field, make sure every user of
+the networks is a member of the new group, and regenerate the `sudoers` file with
+the command above.
+{{% /alert %}}
 
 The IP address is automatically assigned by macOS's bootpd.
 If the IP address is not assigned, try the following commands:
--- a/website/content/en/docs/releases/breaking.md
+++ b/website/content/en/docs/releases/breaking.md
@@ -4,6 +4,8 @@ weight: 20
 ---
 
 ## v2.2.0
+- The default [`socket_vmnet` group](../config/network/vmnet.md) in `networks.yaml` was changed from `everyone` to
+  `admin`. A non-admin user must set `group` to a group they belong to (e.g. `staff`).
 - The default VM driver on Windows hosts was changed from `wsl2` to `qemu`,
   as `wsl2` is not compatible with most templates.
 
--- a/pkg/networks/commands_test.go
+++ b/pkg/networks/commands_test.go
@@ -38,10 +38,6 @@ func TestLogFile(t *testing.T) {
 func TestUser(t *testing.T) {
 	config, err := DefaultConfig()
 	assert.NilError(t, err)
-	if runtime.GOOS != "darwin" && config.Group == "everyone" {
-		// The "everyone" group is a specific macOS feature to include non-local accounts.
-		config.Group = "staff"
-	}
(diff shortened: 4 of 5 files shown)


pkg/networks/config.go
```diff
@@ -25,6 +28,36 @@ var defaultConfigTemplate string
 
 type defaultConfigTemplateArgs struct {
 	SocketVMNet string // "/opt/socket_vmnet/bin/socket_vmnet"
+	Group       string // primary group of the current user, e.g. "staff"
```
AkihiroSuda: Can we just allow the current user, not the group of the current user?

website/content/en/docs/config/network/vmnet.md
```diff
@@ -164,6 +169,19 @@ sudo install -o root etc_sudoers.d_lima /etc/sudoers.d/lima
 rm etc_sudoers.d_lima
 ```
 
+The `sudoers` file grants the members of the `group` configured in `networks.yaml`
+permission to run `socket_vmnet` as root. Since Lima v2, `group` defaults to the
```
AkihiroSuda: Not since v2

ekalinin: Good catch -- v2.0 and v2.1 already ship with `group: everyone`, so "since v2" is wrong. This change targets v2.2; I've updated the doc to say "Since Lima v2.2". Let me know if you'd rather drop the explicit version reference altogether.

pkg/networks/config.go
ekalinin: Sure -- I'll scope the sudoers rule to the current user (`<user> ALL=(root:wheel) …`) instead of a group. `socket_vmnet --socket-group` still needs a group though, so that'll stay the user's primary group (macOS has no per-user group). OK?

AkihiroSuda: Eventually we should have `socket_vmnet --socket-user=USER`

AkihiroSuda: >  I'll scope the sudoers rule to the current user (<user> ALL=(root:wheel) …) instead of a group.

Maybe this work can be deferred until we have `socket_vmnet --socket-user=USER`, and, drop the support for older releases of `socket_vmnet`?

For now, maybe we can just change the default group of `everyone` to `admin` (more preferable than `staff`?)

cc @jandubois @nirs

ekalinin: Sure, let's defer the proper per-user scoping until `socket_vmnet` gains `--socket-user=USER` (then we can require that version and drop the group indirection entirely).

For now I'll switch the default from `staff` to `admin`. It's tighter than `staff`, and it fits the realistic audience: setting up the sudoers file already needs `sudo`, which on macOS defaults to admins only -- so `%admin` covers exactly the users who can use this. Caveat: a standard (non-admin) user isn't in admin, so both the sudoers rule and `--socket-group` would exclude them; I'll document that such setups should set group to a group they belong to (e.g. staff).

This also lets me drop the dynamic primary-group logic and keep it a simple static default. 

Does that sound right? If so I'll update the PR accordingly.

AkihiroSuda: Yes, right, thanks

ekalinin: Done.

AkihiroSuda (dismissed):
Thanks

pkg/networks/networks.go
```diff
@@ -7,7 +7,7 @@ import "net"
 
 type Config struct {
 	Paths    Paths              `yaml:"paths" json:"paths"`
-	Group    string             `yaml:"group,omitempty" json:"group,omitempty"` // default: "everyone"
+	Group    string             `yaml:"group,omitempty" json:"group,omitempty"` // default: "admin"
```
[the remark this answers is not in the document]
AkihiroSuda: >  it may mislead API consumers

No such consumers exist.

website/content/en/docs/config/network/vmnet.md
```diff
@@ -164,6 +167,20 @@ sudo install -o root etc_sudoers.d_lima /etc/sudoers.d/lima
 rm etc_sudoers.d_lima
 ```
 
+The `sudoers` file grants the members of the `group` configured in `networks.yaml`
+permission to run `socket_vmnet` as root. Since Lima v2.2, `group` defaults to
```
AkihiroSuda, line 171: This breaking change also has to be documented in https://github.com/lima-vm/lima/blob/3c7bb540d79f21c55e1f8aac2d3fb35213fee5cc/website/content/en/docs/releases/breaking.md?plain=1#L6-L8

ekalinin: Thanks, done.

website/content/en/docs/releases/breaking.md
```diff
@@ -3,7 +3,9 @@ title: Breaking changes
 weight: 20
 ---
 
-## v2.2.0
+## [v2.2.0](https://github.com/lima-vm/lima/releases/tag/v2.2.0)
```
AkihiroSuda: Inexistent yet

ekalinin: Removed link.

AkihiroSuda (approved):

AkihiroSuda:
@jandubois @abiosoft @ayush-panta

I'm merging this.
This change may potentially affect Rancher Desktop, colima, or Finch.
The old behavior can be restored by setting `group: everyone` in `~/.lima/_config/networks.yaml`.

PR #5195, opened 2026-07-06
downloader: fall back to built-in gzip decompression

jandubois:
Prefer the external `gzip` binary, but decode in-process with `compress/gzip` when it is missing, so a plain Windows host (no `gzip`) can still unpack a .tar.gz` image. `xz`, `bzip2`, and `zstd` stay external-only. A `ctxReader` keeps the in-process decode cancellable, so a canceled download unwinds instead of running to the end.

No new dependency: `compress/gzip` is stdlib and was already linked into `limactl`, so this adds no measurable binary size.

Extracted from #4998

### DIFF
--- a/pkg/downloader/downloader.go
+++ b/pkg/downloader/downloader.go
@@ -5,6 +5,7 @@ package downloader
 
 import (
 	"bytes"
+	"compress/gzip"
 	"context"
 	"crypto/sha256"
 	"errors"
@@ -525,8 +526,6 @@ func decompressorByMagic(file string) string {
 }
 
 func decompressLocal(ctx context.Context, decompressCmd, dst, src, ext, description string) error {
-	logrus.Infof("decompressing %s with %v", ext, decompressCmd)
-
 	st, err := os.Stat(src)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -55,6 +55,7 @@ require (
 	github.com/sirupsen/logrus v1.9.4
 	github.com/spf13/cobra v1.10.2 // gomodjail:unconfined
 	github.com/spf13/pflag v1.0.10
+	github.com/ulikunitz/xz v0.5.15
```
AkihiroSuda: (Why do we already have this in `go.sum`? Apparently not an indirect dependency either)

AkihiroSuda: Not a fan of introducing this dependency.
So far, only FreeBSD images are compressed in xz.

jandubois: It is a dependency of go-diskfs: https://github.com/diskfs/go-diskfs/blob/2bdff12e5d99a077d5a7cbb02552226178163d16/go.mod#L16

I guess `go mod tidy` prunes it out of `go.mod` because we are not currently calling any code that needs it.

jandubois: What in particular is wrong with this dependency (compared to all the others)?

I used it because we already had it in `go.sum`, but I guess the fact that it is not listed as an indirect dependency means we are not actually linking in the code. I can remove it again, but I would like to understand the reason why this one is problematic.

jandubois: Seems like it adds 0.2 MB to the total size (0.7%). Doesn't seem significant, but I'll remove it.

Adding gunzip support is free; the code already seems to be used elsewhere.

AkihiroSuda: The repo contains literally cryptic codes and could be a hot spot to attempt to hide malware.

If I were "Jia Tan" and could somehow hijack the repo, I would have tried to inject a shell code in https://github.com/ulikunitz/xz/blob/master/lzma/fox.lzma (via `//go:embed`) or maybe in https://github.com/ulikunitz/xz/blob/master/internal/randtxt/englm3.go

jandubois: Thanks!

Note that I already removed the xz support from this PR; please review if you have a chance.

AkihiroSuda (approved):
Thanks

PR #5196, opened 2026-07-07
Windows: detect the SSH toolchain and convert paths for native OpenSSH

jandubois:
On Windows, Lima runs `cygpath` to translate key and socket paths, assuming a Cygwin-based ssh (Git for Windows / MSYS2). With only native Windows OpenSSH installed, `cygpath` is unavailable and `limactl create` fails immediately.

Detect which toolchain the selected `ssh` belongs to and convert paths to the form it expects: the sibling `cygpath` (respecting its `fstab`) for Cygwin-based `ssh`, or forward slashes (`C:/Users/...`) for native Windows OpenSSH. Also pick a complete `ssh` install, skipping a partial one like MinGit that lacks `scp` or `ssh-keygen`, and add an `ioutilx` native fallback so path conversion no longer hard-fails when `cygpath` is absent.

This unblocks `limactl create` on plain Windows; `limactl copy` and reverse-sshfs mounts are separate follow-ups.

Extracted from #4998

### DIFF
--- a/pkg/ioutilx/ioutilx.go
+++ b/pkg/ioutilx/ioutilx.go
@@ -50,13 +50,39 @@ func FromUTF16leToString(r io.Reader) (string, error) {
 	return string(out), nil
 }
 
+// WindowsSubsystemPath converts a Windows path to a Cygwin/MSYS form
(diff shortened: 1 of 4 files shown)


jandubois:
Changes already merged via #5205

PR #5197, opened 2026-07-07
cidata: use cp to rewrite /etc/fstab, clarify unescape_fstab comment

ekalinin:
Replace `cat /etc/fstab.lima.tmp >/etc/fstab` with `cp` in `boot.Linux/05-lima-mounts.sh`. 

Also reword the `util/unescape_fstab.sh` header: it decodes the full set of octal escapes the kernel emits in `/proc/mounts` (a superset of what `escape_fstab.sh` produces), so it round-trips escape's output but is not its exact two-sided inverse.

Follow-up to #5145.


jandubois (approved):
Thanks, LGTM

PR #5198, opened 2026-07-07
additionalDisks: support custom mountPoint and optional mounting

### LINKED ISSUE #4065: additionalDisks: support for optional and customizable mount point
### Description

Currently, the additional disks are always mounted at `/mnt/lima-<disk-name>` and expected to be one of the support fileystem formats e.g. ext2-4, xfs

The current behaviour when the filesystem type specified is unsupported is that the disk is simply ignored by Lima during startup.

I would like to propose that we provide explicit option to disable mounting the disk, and an extra option to customize the mount point, which should give us a configuration option similar to the following.

```yaml
additionalDisks:
 - name: string 
   format: bool
   mount: bool # if the disk should be auto-mounted by Lima
   mountPoint: string # custom mountPoint for the disk
   fsType: string
```
**PS:** Colima relies on the current behaviour to prevent Lima from mounting a disk.

ekalinin:
Additional disks were always mounted at the hardcoded path `/mnt/lima-<name>`, with no way to change that location or to attach a disk without Lima mounting it. Issue #4065 asks for both: a custom mount point and an attach-only mode so the guest can manage the disk itself (a flexibility Colima can benefit from).

Add two optional, backwards-compatible fields to each `additionalDisks` entry:
- `mountPoint`: absolute guest path to mount at (default `/mnt/lima-<name>`)
- `mount`: whether Lima auto-mounts the disk (default `true`; false attaches the disk without mounting it)

A separate `mount` boolean is used rather than overloading an empty mountPoint, per the discussion in #4065. `nil` for either field preserves the current behavior. Values are passed to the guest via `LIMA_CIDATA_DISK_<i>_MOUNT / _MOUNTPOINT` and honored by `05-lima-disks.sh` (non-swap disks only). `mountPoint` is validated to be an absolute guest path (`path.IsAbs`, not the host-dependent `filepath.IsAbs`) and not a critical system path.

Fixes #4065

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-disks.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-disks.sh
@@ -18,9 +18,13 @@ for i in $(seq 0 $((LIMA_CIDATA_DISKS - 1))); do
 	FORMAT_DISK="$(get_disk_var "$i" "FORMAT")"
 	FORMAT_FSTYPE="$(get_disk_var "$i" "FSTYPE")"
 	FORMAT_FSARGS="$(get_disk_var "$i" "FSARGS")"
+	MOUNT_DISK="$(get_disk_var "$i" "MOUNT")"
+	MOUNT_POINT="$(get_disk_var "$i" "MOUNTPOINT")"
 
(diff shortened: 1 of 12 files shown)


AkihiroSuda (dismissed):
Thanks

PR #5199, opened 2026-07-07
fix out-of-bounds slice in detectValidPublicKey

SABITHSAHEB:
1. sigLength is a 4-byte length prefix read from the base64-decoded key blob, and the format identifier is taken from decodedKey[4 : 4+sigLength], but the guard only rejects sigLength > len(decodedKey).
2. a ~/.ssh/*.pub entry whose length prefix is within 3 of the decoded length (e.g. equal to it) passes that guard, so the slice runs past the buffer and panics limactl while DefaultPubKeys scans the keys during start.
Compared sigLength against len(decodedKey)-4 so the 4-byte prefix is counted; len is already known to be >= 4, so the subtraction can't underflow. Added a Test_detectValidPublicKey case that panics without the change.

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -514,7 +514,11 @@ func detectValidPublicKey(content string) bool {
 		return false
 	}
 	sigLength := binary.BigEndian.Uint32(decodedKey)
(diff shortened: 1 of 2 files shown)


jandubois (approved):
Thanks, LGTM

PR #5205, opened 2026-07-08
Windows: detect the SSH toolchain and convert paths for native OpenSSH

### LINKED ISSUE #4819: Windows: drop dependency on `cygpath.exe`

mie313:
This PR is derived from #5196 , and most of the work is done by @jandubois .

Changes from the original PR are:
- Modify log/error messages (using `%#q` instead of `%q`)
- Slightly modify comments
- Add a test case `TestDefaultPubKeys_WithoutExistingKey` in sshutil_test.go because `TestDefaultPubKeys` always succeeds if a public key is already created under `.lima/_config/`

### DIFF
--- a/pkg/ioutilx/ioutilx.go
+++ b/pkg/ioutilx/ioutilx.go
@@ -50,13 +50,49 @@ func FromUTF16leToString(r io.Reader) (string, error) {
 	return string(out), nil
 }
 
+// WindowsSubsystemPath converts a Windows path to a Cygwin/MSYS form
+// (C:\Users\jan -> /c/Users/jan) via cygpath on PATH. A caller holding a
+// specific ssh toolchain should use WindowsSubsystemPathWithCygpath, which
+// runs that toolchain's own cygpath and respects its fstab.
 func WindowsSubsystemPath(ctx context.Context, orig string) (string, error) {
-	out, err := exec.CommandContext(ctx, "cygpath", filepath.ToSlash(orig)).CombinedOutput()
-	if err != nil {
-		logrus.WithError(err).Errorf("failed to convert path to mingw, maybe not using Git ssh?")
-		return "", err
+	return WindowsSubsystemPathWithCygpath(ctx, "cygpath", orig)
+}
+
+// WindowsSubsystemPathWithCygpath converts a Windows path with the given
+// cygpathExe ("cygpath" for PATH, or an absolute path to bind the
+// conversion to one Cygwin install). When cygpath is unavailable it falls
+// back to a native conversion of the drive-letter case; Universal Naming Convention
+// (UNC) and other non-drive-letter inputs return an error.
+func WindowsSubsystemPathWithCygpath(ctx context.Context, cygpathExe, orig string) (string, error) {
+	out, err := exec.CommandContext(ctx, cygpathExe, filepath.ToSlash(orig)).CombinedOutput()
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
 	}
-	return strings.TrimSpace(string(out)), nil
+
+	logrus.WithError(err).Debugf("cygpath unavailable for %#q (output: %#q), attempting native conversion", orig, strings.TrimSpace(string(out)))
+
+	return windowsSubsystemPathWithoutCygpath(orig)
+}
+
+func windowsSubsystemPathWithoutCygpath(orig string) (string, error) {
+	// Only an absolute drive-letter path ("C:\foo") has a well-defined
+	// Cygwin form here. A drive-relative path ("C:foo") would become
+	// an unrelated absolute path, so reject it.
+	if !filepath.IsAbs(orig) {
+		return "", fmt.Errorf("cannot convert %#q to a Cygwin-style path: input is not an absolute drive-letter path", orig)
+	}
+
+	// UNC path ("\\server\share\foo") is rejected here.
+	if vol := filepath.VolumeName(orig); len(vol) == 2 && vol[1] == ':' {
+		// orig[2:] starts with a separator for an absolute drive path
+		// (C:\foo, C:/foo); strip it so the result stays canonical.
+		tail := strings.TrimPrefix(filepath.ToSlash(orig[2:]), "/")
+		converted := "/" + strings.ToLower(vol[:1]) + "/" + tail
+		logrus.Debugf("native cygpath fallback: %#q -> %#q", orig, converted)
+		return converted, nil
+	}
+
+	return "", fmt.Errorf("cannot convert %#q to a Cygwin-style path: cygpath is unavailable and input is not an absolute drive-letter path", orig)
 }
 
 func WindowsSubsystemPathForLinux(ctx context.Context, orig, distro string) (string, error) {
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -61,6 +61,13 @@ func NewSSHExe() (SSHExe, error) {
 		}
 	}
 
+	if runtime.GOOS == "windows" {
+		if exe := pickCompleteSSHOnWindows(); exe != "" {
+			sshExe.Exe = exe
+			return sshExe, nil
+		}
+	}
+
 	executable, err := exec.LookPath("ssh")
 	if err != nil {
 		return SSHExe{}, err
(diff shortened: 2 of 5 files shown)


pkg/ioutilx/ioutilx.go
```diff
@@ -50,13 +50,42 @@ func FromUTF16leToString(r io.Reader) (string, error) {
...
+		return strings.TrimSpace(string(out)), nil
 	}
-	return strings.TrimSpace(string(out)), nil
+	logrus.WithError(err).Debugf("cygpath unavailable for %#q (output: %#q), attempting native conversion", orig, strings.TrimSpace(string(out)))
+
+	// Only an absolute drive-letter path ("C:\foo") has a well-defined
+	// Cygwin form here. A drive-relative path ("C:foo") would become
+	// an unrelated absolute path, so reject it.
+	if filepath.IsAbs(orig) {
+		// UNC path ("\\server\share\foo") is rejected here.
```
mie313: I added this comment so that I can understand the if statement skips UNC path.

pkg/ioutilx/ioutilx_test.go
```diff
@@ -0,0 +1,60 @@
...
+			_, err := WindowsSubsystemPathWithCygpath(ctx, bogusCygpath, input)
+			assert.ErrorContains(t, err, "cannot convert", "input=%q", input)
+		}
+	})
+
+	t.Run("lowercases drive letter", func(t *testing.T) {
+		if runtime.GOOS != "windows" {
+			t.Skip("filepath.VolumeName / IsAbs only recognize drive letters on Windows")
+		}
+		got, err := WindowsSubsystemPathWithCygpath(ctx, bogusCygpath, `d:\path`)
```
mie313: In the original PR, the path is `D:\path`. From the test name, I changed the path to a lowercase drive letter.

pkg/sshutil/sshutil.go
```diff
@@ -70,6 +77,166 @@ func NewSSHExe() (SSHExe, error) {
...
+// installs (Windows PE, some enterprise setups) that put it off the C: drive.
+func systemRoot() string {
+	if r := os.Getenv("SystemRoot"); r != "" {
+		return r
+	}
+	return `C:\Windows`
+}
+
+func missingSiblings(dir string, siblings ...string) []string {
+	missing := make([]string, 0, len(siblings))
```
mie313: Since the maximum length of `missing` is `len(siblings)`, I set the capacity.

pkg/sshutil/sshutil.go
```diff
@@ -70,6 +77,166 @@ func NewSSHExe() (SSHExe, error) {
...
+		if _, err := os.Stat(filepath.Join(dir, s)); err != nil {
+			missing = append(missing, s)
+		}
+	}
+	return missing
+}
+
+var (
+	// cygwinDetectCache maps a resolved ssh path to its sibling cygpath.exe,
+	// or "" when that ssh is not Cygwin-based. Each path is probed once. Before using it, we should call cygwinDetectCacheRW's RLock/Lock to protect from concurrent access.
```
mie313: I put an additional comment (`Before using it~`) so that I will never forget to use the lock

pkg/ioutilx/ioutilx.go
```diff
@@ -50,13 +50,42 @@ func FromUTF16leToString(r io.Reader) (string, error) {
...
+// conversion to one Cygwin install). When cygpath is unavailable it fails
+// back to a native conversion of the drive-letter case; Universal naming convention
+// (UNC) and other non-drive-letter inputs return an error.
+func WindowsSubsystemPathWithCygpath(ctx context.Context, cygpathExe, orig string) (string, error) {
+	out, err := exec.CommandContext(ctx, cygpathExe, filepath.ToSlash(orig)).CombinedOutput()
+	if err == nil {
+		return strings.TrimSpace(string(out)), nil
 	}
-	return strings.TrimSpace(string(out)), nil
+	logrus.WithError(err).Debugf("cygpath unavailable for %#q (output: %#q), attempting native conversion", orig, strings.TrimSpace(string(out)))
```
AkihiroSuda, line 72: Split the code below to `windowsSubsystemPathWithoutCygpath` and cover them with unit tests

pkg/ioutilx/ioutilx.go
```diff
@@ -50,13 +50,42 @@ func FromUTF16leToString(r io.Reader) (string, error) {
...
-		return "", err
+	return WindowsSubsystemPathWithCygpath(ctx, "cygpath", orig)
+}
+
+// WindowsSubsystemPathWithCygpath converts a Windows path with the given
+// cygpathExe ("cygpath" for PATH, or an absolute path to bind the
+// conversion to one Cygwin install). When cygpath is unavailable it fails
+// back to a native conversion of the drive-letter case; Universal naming convention
+// (UNC) and other non-drive-letter inputs return an error.
+func WindowsSubsystemPathWithCygpath(ctx context.Context, cygpathExe, orig string) (string, error) {
```
AkihiroSuda, line 66:
```suggestion
func windowsSubsystemPathWithCygpath(ctx context.Context, cygpathExe, orig string) (string, error) {
```

pkg/ioutilx/ioutilx_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+// TestWindowsSubsystemPathWithCygpath_Fallback exercises the native
+// drive-letter fallback path: when the cygpath binary is unreachable,
+// absolute drive-letter inputs convert in-process (e.g., "C:\foo" -> "/c/foo")
+// and inputs without a drive letter return an error.
+func TestWindowsSubsystemPathWithCygpath_Fallback(t *testing.T) {
```
AkihiroSuda:
```suggestion
func TestWindowsSubsystemPath(t *testing.T) {
  for _, withCygpath := range []bool{true, false} {
    t.Run(fmt.Sprintf("withCygpath=%v", withCygpath), func(t *testing.T) {
      if withCygpath {
        if _, err := exec.LookPath(cygpathExe); errors.Is(err, fs.ErrNotFound) {
          t.Skipf("%q: %v", cygpathExe, err)
        }
        testWindowsSubsystemPathWithCygpath(...)
      } else {
        testWindowsSubsystemPathWithoutCygpath(...)
      }
    }
  }
```

pkg/ioutilx/ioutilx_test.go
```diff
@@ -0,0 +1,60 @@
...
+	ctx := t.Context()
+
+	// A bogus binary name guarantees the fallback path runs on every
+	// platform without inventing a Windows-only test.
+	const bogusCygpath = "no-such-cygpath-binary-xyz"
+
+	t.Run("drive-letter backslash", func(t *testing.T) {
+		if runtime.GOOS != "windows" {
+			t.Skip("filepath.VolumeName / IsAbs treat C:\\foo as a relative path on non-Windows")
```
AkihiroSuda: You can remove this line by renaming the file to `*_windows_test.go`

pkg/sshutil/sshutil.go
AkihiroSuda, line 1: Consider moving windows-specific codes to `*_windows.go`

pkg/sshutil/sshutil.go
```diff
@@ -70,6 +77,166 @@ func NewSSHExe() (SSHExe, error) {
...
+	cygwinDetectCache   = map[string]string{}
+	cygwinDetectCacheRW sync.RWMutex
```
AkihiroSuda, line 134: Why not use sync.Once

pkg/ioutilx/ioutilx.go
AkihiroSuda, line 1: cygpath should be now uninstalled from the CI when running the integration tests?

mie313: Does CI mean `windows-qemu` step (https://github.com/lima-vm/lima/blob/master/.github/workflows/test.yml#L221)?
I think it should work without `cygpath.exe`, but I'm not sure where we installed cygpath.exe on a windows-2025 instance.

mie313: I found that `cygpath.exe` is contained in `C:\msys64\usr\bin` . (Probably it was installed by default when I turned on msys2).

I will try to remove `C:\msys64\usr\bin` from https://github.com/lima-vm/lima/blob/master/.github/workflows/test.yml#L244.

By the way, on my local Windows machine, I confirmed that Ubuntu VM was launched with:
- Uninstalling Git for Windows (`ssh.exe` is in `C:\Windows\System32\`)
- Remove `C:\msys64\usr\bin` from PATH
- Remove ` _LIMA_WINDOWS_EXTRA_PATH` (so there is no `cygpath.exe`)

mie313: I tried it on CI (two commits, `CI test` and `test2`) but didn't work well:
- I removed `C:\msys64\usr\bin` from PATH, but it caused that `pacman` was not found (I think packman was also in the path)
- After I removed the path, `cygpath.exe` was still found in `/bin/cygpath.exe`. I didn't find such a path (`/bin` or `\bin`) in PATH
- I printed PATH. However, [the second trial](https://github.com/lima-vm/lima/actions/runs/29052164848/job/86235275062?pr=5205#step:7:20) had a much different PATH from [the first trial](https://github.com/lima-vm/lima/actions/runs/29051125139/job/86231801089?pr=5205#step:7:18). I assume this is because different Windows instances were chosen

Do you have any advice to test Windows host without `cygpath.exe` on CI?

AkihiroSuda: Thanks for trying. You can keep it installed for now.

pkg/ioutilx/ioutilx.go
mie313: `WindowsSubsystemPathWithCygpath` is not just an internal function of `WindowsSubsystemPath`, but also called from `PathForSSH` in `sshutil.go`, because it needs to call a `cygpath` which from the same origin as SSH. (Cygwin-based SSH/cygpath, or OpenSSH based SSH/cygpath).

pkg/ioutilx/ioutilx_test.go
mie313: I think it is difficult to call `cygpath` in the test, because Cygwin-based cygpath creates different results from OpenSSH based one.

pkg/sshutil/sshutil.go
mie313: I think we can not use `sync.Once` here, because the cache will be built by multiple function calls.
Do you think `sync.Map` is better choice?

pkg/ioutilx/ioutilx.go
mie313: Thank you. I removed those 2 CI test commits.

pkg/sshutil/sshutil.go
mie313: I tried it, but since some functions (for example, `DefaultPubKeys`) call windows-specific functions and they are nested, it is hard to separate windows-specific codes into `sshutil_windows.go`.I tried it, but since some functions (for example, `DefaultPubKeys`) call windows-specific functions and they are nested, it is hard to separate windows-specific codes into `sshutil_windows.go`.

I separated test codes into `sshutil_windows_test.go` in https://github.com/lima-vm/lima/pull/5205/commits/8ad2976b1c42e7b2811094e30a4cdaa73954e850.

mie313:
@AkihiroSuda @jandubois 

I pushed additional two commits (https://github.com/lima-vm/lima/pull/5205/commits/11dd569201216a3b97f0e022eb0eed83738b9b0a and https://github.com/lima-vm/lima/pull/5205/commits/8ad2976b1c42e7b2811094e30a4cdaa73954e850) to address Akihiro's reviews without mixing fixes with Jan's original works.
Once review is completed, I will squash those two commits into the original two commits.

pkg/ioutilx/ioutilx_test.go
mie313: I modified the original test to `TestWindowsSubsystemPathWithoutCygpath` as follows:
https://github.com/lima-vm/lima/pull/5205/changes#diff-17e1919f104fdc11274a933af082d23c72e9ad091fcac0eb008310a44cf20e8eR17

pkg/sshutil/sshutil.go
```diff
@@ -70,6 +77,166 @@ func NewSSHExe() (SSHExe, error) {
...
+	if r := os.Getenv("SystemRoot"); r != "" {
+		return r
+	}
+	return `C:\Windows`
+}
+
+func missingSiblings(dir string, siblings ...string) []string {
+	missing := make([]string, 0, len(siblings))
+	for _, s := range siblings {
+		if _, err := os.Stat(filepath.Join(dir, s)); err != nil {
```
AkihiroSuda, line 123:
```suggestion
		if _, err := os.Stat(filepath.Join(dir, filepath.Clean(s))); err != nil {
```

or add a comment to clarify that siblings must not contain non-clean path

AkihiroSuda (dismissed):
Thanks, I still want to see windows-specific codes to be split to `_windows.go`, but it can be worked out later

jandubois:
Unfortunately I didn't get to look at this today. I have an AI review at https://jandubois.github.io/lima/20260712-230128-pr-5205.html, but I haven't looked at it either, so the findings might not be relevant.

I'm going to do a manual review tomorrow.

pkg/sshutil/sshutil.go
jandubois: This change seems to have been dropped; it is needed for the ControlPath to work when cygpath is not installed.

```suggestion
		controlSock, err = PathForSSH(ctx, sshExe, controlSock)
```

pkg/ioutilx/ioutilx.go
```diff
@@ -50,13 +50,49 @@ func FromUTF16leToString(r io.Reader) (string, error) {
...
+	if vol := filepath.VolumeName(orig); len(vol) == 2 && vol[1] == ':' {
+		// orig[2:] starts with a separator for an absolute drive path
+		// (C:\foo, C:/foo); strip it so the result stays canonical.
+		tail := strings.TrimPrefix(filepath.ToSlash(orig[2:]), "/")
+		converted := "/" + strings.ToLower(vol[:1]) + "/" + tail
+		logrus.Debugf("native cygpath fallback: %#q -> %#q", orig, converted)
+		return converted, nil
+	}
+
+	return "", fmt.Errorf("cannot convert %#q to a Cygwin-style path: input is UNC path", orig)
```
jandubois: Input is not necessarily an UNC path, it could also be an extended length path, like `\\?\C:\foo` or a device path. Error should also mention that the root cause is the missing `cygpath` binary:

```suggestion
	return "", fmt.Errorf("cannot convert %q to a Cygwin-style path: cygpath unavailable and input is not an absolute drive-letter path", orig)
```

pkg/sshutil/sshutil.go
```diff
@@ -70,6 +77,166 @@ func NewSSHExe() (SSHExe, error) {
...
+		if _, err := os.Stat(filepath.Join(dir, filepath.Clean(s))); err != nil {
+			missing = append(missing, s)
+		}
+	}
+	return missing
+}
+
+var (
+	// cygwinDetectCache maps a resolved ssh path to its sibling cygpath.exe,
+	// or "" when that ssh is not Cygwin-based. Each path is probed once. Before using it, we should call cygwinDetectCacheRW's RLock/Lock to protect from concurrent access.
```
jandubois: Run-on comment looks looks like a personal note and not documentation.

```suggestion
	// or "" when that ssh is not Cygwin-based. Each path is probed once.
```

pkg/ioutilx/ioutilx.go
```diff
@@ -50,13 +50,49 @@ func FromUTF16leToString(r io.Reader) (string, error) {
...
-	out, err := exec.CommandContext(ctx, "cygpath", filepath.ToSlash(orig)).CombinedOutput()
-	if err != nil {
-		logrus.WithError(err).Errorf("failed to convert path to mingw, maybe not using Git ssh?")
-		return "", err
+	return WindowsSubsystemPathWithCygpath(ctx, "cygpath", orig)
+}
+
+// WindowsSubsystemPathWithCygpath converts a Windows path with the given
+// cygpathExe ("cygpath" for PATH, or an absolute path to bind the
+// conversion to one Cygwin install). When cygpath is unavailable it fails
```
jandubois:
```suggestion
// conversion to one Cygwin install). When cygpath is unavailable it falls
```

pkg/sshutil/sshutil.go
```diff
@@ -70,6 +77,166 @@ func NewSSHExe() (SSHExe, error) {
...
+	if r := os.Getenv("SystemRoot"); r != "" {
+		return r
+	}
+	return `C:\Windows`
+}
+
+func missingSiblings(dir string, siblings ...string) []string {
+	missing := make([]string, 0, len(siblings))
+	for _, s := range siblings {
+		if _, err := os.Stat(filepath.Join(dir, filepath.Clean(s))); err != nil {
```
jandubois: `filepath.Join` returns a cleaned path already, so no need to clean individual fragments:

```go
	if b.Len() == 0 {
		return ""
	}
	return Clean(b.String())
```

```suggestion
		if _, err := os.Stat(filepath.Join(dir, s)); err != nil {
```

pkg/sshutil/sshutil.go
jandubois: It doesn't really matter, but the common case is that `siblings` will be empty. Now you allocate backing storage on every call. It would be more efficient to only allocate on demand:


```suggestion
	var missing []string
```

But this will never be noticeable, so either form is fine.

jandubois (changes requested):
The incorrect handling of the controlpath socket is a real bug.

`TestCygpathForSSH` and `TestSftpServerForSSH` are missing from #5196.

The remaining issues are minor spelling nits.

pkg/sshutil/sshutil.go
mie313: Thank you for pointing that out.
I added `filepath.Clean` so that the input constraint is implied in the code.
However, as it is redundant, I reverted the change.
To address @AkihiroSuda 's [earlier concern](https://github.com/lima-vm/lima/pull/5205#discussion_r3564681441), I added a doc comment to `missingSiblings` clarifying that siblings must not contain non-clean paths.

jandubois: I don't understand why that is needed. `filepath.Join()` will always return a clean path; there is no requirement that the siblings must be clean before being passed in. They will be cleaned before the return.

jandubois:
There are still missing tests, but I'm happy to add them back in a followup PR. The current PR could be merged, but commits must be squashed.

jandubois:
@mie313 Please squash commits so we can merge this!

mie313:
I'm sorry for being late. Now I squashed them.

jandubois (approved):
Thanks, LGTM

pkg/ioutilx/ioutilx.go
AkihiroSuda: Why not just call `WindowsSubsystemPath`?

pkg/ioutilx/ioutilx.go
```diff
@@ -50,13 +50,49 @@ func FromUTF16leToString(r io.Reader) (string, error) {
...
-		logrus.WithError(err).Errorf("failed to convert path to mingw, maybe not using Git ssh?")
-		return "", err
+	return WindowsSubsystemPathWithCygpath(ctx, "cygpath", orig)
+}
+
+// WindowsSubsystemPathWithCygpath converts a Windows path with the given
+// cygpathExe ("cygpath" for PATH, or an absolute path to bind the
+// conversion to one Cygwin install). When cygpath is unavailable it falls
+// back to a native conversion of the drive-letter case; Universal Naming Convention
+// (UNC) and other non-drive-letter inputs return an error.
```
AkihiroSuda, line 65: The fall-back should be attempted in `WindowsSubsystemPath` rather than in `WindowsSubsystemPathWithCygpath` ?

PR #5216, opened 2026-07-09
portfwdserver: fix socket FD leak per forwarded connection

### LINKED ISSUE #5210: guestagent leaks one socket FD per forwarded connection; long-running instances hit RLIMIT_NOFILE and port forwarding silently dies
### Description

`lima-guestagent` leaks one socket file descriptor per **forwarded connection**. On a long-running instance that receives steady traffic through a static `portForwards` rule, the guest agent's FD count grows linearly until it hits `RLIMIT_NOFILE`, after which port forwarding half-dies: the host-side listener still accepts TCP connections, but no data ever flows (clients see an immediate close / TLS handshake failure with zero bytes). The guest agent does not crash, so systemd never restarts it and the instance never recovers on its own — only a guest-agent (or VM) restart clears it.

This is distinct from #4888 (host-side `GuestAgentClient` leak on reconnect, fixed in v2.1.2 by #4889). I observe this leak **on the guest side, on v2.1.4**, and it is proportional to forwarded connections, not reconnects.

### Environment

- Lima 2.1.4 (Homebrew), previously reproduced on 2.1.1
- Host: macOS 15.5 (Darwin 24.5.0), Apple Silicon, `vmType: vz`
- Guest: Ubuntu 24.04.4 LTS, kernel 6.8.0-134-generic, `lima-guestagent version 2.1.4` running as the standard systemd unit
- Static `portForwards` in `lima.yaml` binding host LAN-IP ports to guest ports (the VM runs a k3s node whose ingress receives continuous public HTTP/HTTPS traffic)

### Reproduction

FD count of the guest agent tracks forwarded connections 1:1:

```
# inside the guest
$ pid=$(pgrep -f lima-guestagent); sudo ls /proc/$pid/fd | wc -l
168

# from the host: 20 requests through a forwarded port
$ for i in $(seq 20); do curl -sk -o /dev/null https://<host-lan-ip>:30428/; done

# inside the guest
$ sudo ls /proc/$pid/fd | wc -l
189    # +21 (one extra from ambient traffic)
```

Connections made directly to the guest service (bypassing the forward) do **not** move the count, so it's the tunnel path, not the workload.

Under ~6 forwarded connections/minute of ambient traffic, a freshly restarted agent goes 26 → 58 → 110 → 168 FDs within the first hour and never plateaus.

### What the leaked FDs are

Orphaned sockets: they appear as `socket:[inode]` in `/proc/<pid>/fd`, but the inodes are absent from **every** `/proc/net/*` table (tcp, tcp6, udp, udp6, unix, netlink, packet, raw), and `ss -A vsock` shows only the single established control connection (`3:2222`). i.e. the protocol side of each per-connection stream is fully closed, but the FD is never `close()`d in the guest agent.

```
$ sudo ls -la /proc/$pid/fd | grep -c 'socket:'
107        # of 113 total FDs; only 1 inode found in any /proc/net table
$ sudo ss -A vsock
v_str ESTAB 0 0  3:2222  2:1979827542    # the only live vsock stream
```

### Long-run failure

On the previous boot the agent ran 15 days, consumed the full limit, and forwarding died silently:

```
Max open files  524287  524288  files
```

Its periodic `SyncTime` log lines stop mid-day (when the limit was reached), and the only logged error appears at shutdown:

```
lima-guestagent[714]: level=error msg="failed to save state: open /run/lima-guestagent/event-state.json: too many open files"
```

From that point until the VM was bounced, the hostagent listener accepted TCP on the forwarded ports but transferred zero bytes.

### Workaround

A daily `systemctl restart lima-guestagent` inside the guest resets the FD count; with the host side on ≥ 2.1.2 the hostagent reconnects within seconds and (thanks to #4889) no longer leaks its side on reconnect.

Happy to provide more data or test patches — the affected instance reproduces this within minutes.

AkihiroSuda:
GRPCServerRW.closeCh was a 1-slot buffered channel received from exactly once in TunnelServer.Start, but tcpproxy.DialProxy.HandleConn invokes the Close* methods three times during teardown: CloseRead/CloseWrite from each copy direction's defer, then Close from its own deferred src.Close(). The third send blocked forever, and since HandleConn registers "defer dst.Close()" before "defer src.Close()", the dialed guest connection was never closed: one leaked socket FD and one leaked goroutine per forwarded connection, until the guest agent hit RLIMIT_NOFILE and port forwarding silently died.

Make Close/CloseRead/CloseWrite idempotent and non-blocking with sync.Once + close(closeCh), and close the dialed connection when Start returns so the FD is released even if the guest service keeps its side of the connection open.

Fix #5210

Note: used Claude

### DIFF
--- a/pkg/portfwdserver/server.go
+++ b/pkg/portfwdserver/server.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"io"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/inetaf/tcpproxy"
@@ -39,7 +40,12 @@ func (s *TunnelServer) Start(stream api.GuestService_TunnelServer) error {
 	if err != nil {
 		return err
 	}
-	rw := &GRPCServerRW{stream: stream, id: in.Id, closeCh: make(chan any, 1)}
+	// The tunnel is unusable once this function returns, so close the dialed
+	// connection here as well. This both releases the FD even if the proxy
+	// goroutine is still blocked reading from the guest, and unblocks that
+	// read so the goroutine can finish.
+	defer conn.Close()
+	rw := &GRPCServerRW{stream: stream, id: in.Id, closeCh: make(chan any)}
 	go func() {
 		<-ctx.Done()
 		rw.Close()
@@ -52,7 +58,6 @@ func (s *TunnelServer) Start(stream api.GuestService_TunnelServer) error {
 
 	// The stream will be closed when this function returns.
 	// Wait here until rw.Close(), rw.CloseRead(), or rw.CloseWrite() is called.
-	// We can't close rw.closeCh since the calling order of Close* methods is not guaranteed.
 	<-rw.closeCh
 	logrus.Debugf("closed GRPCServerRW for id: %s", in.Id)
 
@@ -63,6 +68,13 @@ type GRPCServerRW struct {
 	id      string
 	stream  api.GuestService_TunnelServer
 	closeCh chan any
+	// closeOnce guards closeCh. Close, CloseRead, and CloseWrite may be
+	// called in any order and multiple times (e.g., tcpproxy calls
+	// CloseRead/CloseWrite from each copy direction and then Close), so
+	// they must be idempotent and must never block; otherwise the proxy
+	// goroutine gets stuck and never closes the dialed guest connection,
+	// leaking one FD per forwarded connection.
+	closeOnce sync.Once
 
 	// rxBuf holds bytes received from the stream that did not fit in the
 	// buffer passed to the previous Read call.
@@ -91,7 +103,7 @@ func (g *GRPCServerRW) Read(p []byte) (n int, err error) {
 
 func (g *GRPCServerRW) Close() error {
 	logrus.Debugf("closing GRPCServerRW for id: %s", g.id)
-	g.closeCh <- struct{}{}
+	g.closeOnce.Do(func() { close(g.closeCh) })
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


jhgaylor:
Confirmed — this fixes it. Tested on the same setup where I originally hit the bug (#5210).

**Setup:** Ubuntu 24.04 arm64 guest under Lima on an M-series Mac mini, guestagent running via systemd with the vsock gRPC tunnel, static `portForwards` entry (`0.0.0.0:31117` → guest `:31117`, Traefik answering). Test traffic: `curl` with `Connection: close` from a LAN client through the forward, counting `ls /proc/<guestagent-pid>/fd | wc -l` between bursts.

**Stock `lima-guestagent 2.1.4`:**
- baseline: 16 FDs
- after 1 connection: 17
- after 201 connections: **183** (172 of them sockets) — ~0.83 FDs leaked per connection

**Patched guestagent built from this branch (6920d84, `GOOS=linux GOARCH=arm64 CGO_ENABLED=0`):**
- baseline after restart: 16 FDs
- after 201 connections: **16**
- after 100 more: **16**

Forwarding kept working normally throughout (same 404s from the guest service). `go test ./pkg/portfwdserver/` passes on the branch as well.

Thanks for the fast turnaround @AkihiroSuda 🙏

jandubois (approved):
Thanks, LGTM

Fix has also been confirmed by original reporter.

It even got a a rare "Merge as-is" recommendation from AI. Please take a look anyways because it mentions a pre-existing issue relating to half-closes: https://jandubois.github.io/lima/20260716-204108-pr-5216.html

AkihiroSuda:
>  I verified the root cause against inetaf/tcpproxy's source and confirmed both new tests fail on revert. 

Who is "I"?
Is your AI actually capable of running tests, or is it hallucinating?

jandubois:
> Is your AI actually capable of running tests, or is it hallucinating?

Yes, it can run test. Sometimes the agents even write little test programs themselves to verify something.

I call my sandbox "Zilicon Island". It is like "Jurassic Park", except it has roaming AI agents instead of dinosaurs. 😄

jandubois:
Under https://jandubois.github.io/lima/20260716-204108-pr-5216.html#testing-assessment you can actually see the details of the tests the orchestrator performed: changing parts of the PR to confirm that different parts of the test are failing, and how.

PR #5221, opened 2026-07-10
docs(driver): document external driver gRPC API contract

### LINKED ISSUE #4896: Document API contract for external driver transport
The external driver gRPC API in Lima has no documented contract around call ordering or method semantics things like `AdditionalSetupForSSH()` being called before `Start()` finishes, the unclear safety of `SSHAddress()`, overall flow and working of the driver interface etc.

As part of my GSoC tasks last year, I was unable to provide a comprehensive doc that helps someone new to Lima implement external VM drivers. Additionally, I think we should add detailed comments to `driver.proto` file and auto-generate a reference page for the website using `protoc-gen-doc`, so the RPC contract stays in sync with the code and new driver authors have a clear guide, like:

https://github.com/lima-vm/lima/blob/f121314d8fc7584713445f090d8231d8e0e95f43/pkg/driver/external/driver.proto#L49-L57

_Originally posted by @norio-nomura in https://github.com/lima-vm/lima/pull/4345#discussion_r2540041368_

ekalinin:
Document the external driver gRPC API contract by adding detailed doc comments to every RPC and message in `driver.proto`. These comments are the authoritative contract (call ordering, idempotency, per-method semantics) and also improve the godoc for the `driver.Driver` interface.

  - `driver.proto`: service-level call-ordering overview plus per-RPC/field docs (idempotency of Create, conditional SSHAddress, Start streaming, guest agent forwarding)
  - `driver.go`: doc comments on the `Info` and `DriverFeatures` structs carried as JSON in `InfoResponse`
  - `website`: a hand-written "Driver lifecycle and call ordering" section in `drivers.md`, cross-linking to `driver.proto` and the `driver.Driver` godoc

The `protoc-gen-doc` auto-generated reference was descoped from this PR and moved to a separate discussion in #5224.

Fixes #4896

### DIFF
--- a/pkg/driver/external/driver.proto
+++ b/pkg/driver/external/driver.proto
@@ -4,90 +4,215 @@ import "google/protobuf/empty.proto";
 
 option go_package = "github.com/lima-vm/lima/v2/pkg/driver/external";
 
+// Driver is the transport used by an external VM driver. It is the gRPC
+// projection of the Go `driver.Driver` interface (pkg/driver/driver.go): the
+// host agent runs a gRPC client and each external driver binary
+// (lima-driver-<name>) runs the server. The RPCs mirror the interface methods
+// one-to-one, so this file is the authoritative contract for driver authors.
+//
+// Call ordering on the instance start path: instance.Prepare (pkg/instance)
+// runs steps 1-4, then the host agent (pkg/hostagent) boots the VM and runs
+// steps 5-7. Creating an instance without starting it (limactl create) only
+// runs Configure and Create.
+//   1. Configure  - set the instance config; must run before any method that
+//      depends on it.
+//   2. Validate    - reject configs the driver cannot support.
+//   3. Create      - first-time provisioning; a no-op on existing instances.
+//   4. CreateDisk  - create the instance disk(s).
+//   5. Start       - begin booting the VM. It returns once boot has been
+//      initiated (the initial StartResponse), not once the guest has finished
+//      booting; later boot errors arrive on the stream. See StartResponse.
+//   6. AdditionalSetupForSSH - runs right after Start returns (boot initiated,
+//      the guest is not necessarily up yet), before the first SSH connection.
+//   7. SSHAddress  - re-queried after Start only when Info reports
+//      Features.dynamicSSHAddress.
+//
+// Unless stated otherwise, an RPC returning google.protobuf.Empty completes
+// synchronously and reports failures as a gRPC error status. Requests and
+// responses that carry structured data use JSON payloads (Info, Configure) to
+// stay in sync with the Go types without duplicating them here.
+//
+// Concurrency: the start-path RPCs above are invoked sequentially in the
+// order listed. Once the VM is running, the host agent polls read-only RPCs
+// (notably Info and ForwardGuestAgent) from background goroutines, so the
+// driver server must be safe to handle concurrent RPCs.
 service Driver {
+  // Validate reports whether the driver supports the currently configured
+  // instance. It is called early on the start path, before Create. Returning
+  // an error aborts the operation.
   rpc Validate(google.protobuf.Empty) returns (google.protobuf.Empty);
+
+  // Create performs first-time provisioning of the instance (e.g. writing a
+  // "vz-identifier" file). It MUST return successfully (no-op) when called
+  // against an instance that already exists, and it MUST NOT create the disks
+  // (that is CreateDisk's job).
   rpc Create(google.protobuf.Empty) returns (google.protobuf.Empty);
+
+  // CreateDisk creates the instance disk(s). It is called after Create.
   rpc CreateDisk(google.protobuf.Empty) returns (google.protobuf.Empty);
+
+  // Start begins booting the VM. It is a server-streaming RPC that mirrors the
+  // `(chan error, error)` return of the in-process interface: the call returns
+  // once boot has been initiated (the initial StartResponse), not once the
+  // guest has finished booting, and later boot errors arrive on the stream. See
+  // StartResponse for the streaming protocol.
   rpc Start(google.protobuf.Empty) returns (stream StartResponse);
+
+  // Stop terminates the running VM. It blocks until the VM has fully exited
+  // (or the driver's internal shutdown deadline elapses) and returns a gRPC
+  // error status if the shutdown failed.
   rpc Stop(google.protobuf.Empty) returns (google.protobuf.Empty);
+
+  // Delete removes the instance created by Create (and its driver-specific
+  // files). The generic instance directory is removed by the host agent.
   rpc Delete(google.protobuf.Empty) returns (google.protobuf.Empty);
+
+  // BootScripts returns the boot scripts to inject into the VM via cloud-init.
+  // See BootScriptsResponse for the required key format.
   rpc BootScripts(google.protobuf.Empty) returns (BootScriptsResponse);
 
+  // RunGUI starts the GUI synchronously. It blocks and returns only after the
+  // VM terminates. It is only used by drivers that report Features.canRunGui.
   rpc RunGUI(google.protobuf.Empty) returns (google.protobuf.Empty);
+
+  // ChangeDisplayPassword sets the password for the VM display (e.g. VNC).
   rpc ChangeDisplayPassword(ChangeDisplayPasswordRequest) returns (google.protobuf.Empty);
+
+  // GetDisplayConnection returns the connection string for the VM display.
   rpc GetDisplayConnection(google.protobuf.Empty) returns (GetDisplayConnectionResponse);
 
+  // CreateSnapshot creates a VM snapshot identified by the given tag.
   rpc CreateSnapshot(CreateSnapshotRequest) returns (google.protobuf.Empty);
+
+  // ApplySnapshot restores the VM to the snapshot identified by the given tag.
   rpc ApplySnapshot(ApplySnapshotRequest) returns (google.protobuf.Empty);
+
+  // DeleteSnapshot removes the snapshot identified by the given tag.
   rpc DeleteSnapshot(DeleteSnapshotRequest) returns (google.protobuf.Empty);
+
+  // ListSnapshots returns the driver-formatted list of existing snapshots.
   rpc ListSnapshots(google.protobuf.Empty) returns (ListSnapshotsResponse);
 
+  // ForwardGuestAgent reports whether the host agent must forward the guest
+  // agent socket over SSH. When false, the driver provides a direct guest agent
+  // connection via GuestAgentConn (e.g. vsock/virtio).
   rpc ForwardGuestAgent(google.protobuf.Empty) returns (ForwardGuestAgentResponse);
+
+  // GuestAgentConn returns a direct guest agent connection, used when
+  // ForwardGuestAgent is false; it returns nil when the socket is forwarded
+  // over SSH instead. Over this transport the server bridges the connection to
+  // a unix socket in the instance directory when it is not already one, so the
+  // client relies on that socket rather than on a returned connection.
   rpc GuestAgentConn(google.protobuf.Empty) returns (google.protobuf.Empty);
 
+  // Configure sets the instance configuration on the driver. It must be called
+  // before RPCs that depend on the config.
   rpc Configure(SetConfigRequest) returns (SetConfigResponse);
+
+  // Info returns driver metadata and capability flags (see InfoResponse). The
+  // host agent uses it to discover features such as dynamicSSHAddress.
   rpc Info(google.protobuf.Empty) returns (InfoResponse);
+
+  // SSHAddress returns the address used to SSH into the guest. For drivers that
+  // report Features.dynamicSSHAddress the address is not known until after
+  // Start, so the host agent re-queries it at that point.
   rpc SSHAddress(google.protobuf.Empty) returns (SSHAddressResponse);
 
   // AdditionalSetupForSSH provides additional setup required for SSH connection.
-  // It is called after VM is started, before first SSH connection.
+  // It is called right after Start returns (boot initiated, the guest is not
+  // necessarily up yet), before the first SSH connection.
   rpc AdditionalSetupForSSH(google.protobuf.Empty) returns (google.protobuf.Empty);
 }
 
+// BootScriptsResponse carries the boot scripts injected into the VM.
 message BootScriptsResponse {
+  // Maps a script path to its content. Each key must be
+  // "boot.<OS>/<SCRIPT>", or "<SCRIPT>" as a deprecated alias for
+  // "boot.Linux/<SCRIPT>".
   map<string, bytes> scripts = 1;
 }
 
+// SSHAddressResponse carries the guest SSH address.
 message SSHAddressResponse {
+  // Host/IP used to SSH into the guest.
   string address = 1;
 }
 
-message InfoResponse{
+// InfoResponse carries driver metadata and capability flags.
+message InfoResponse {
+  // JSON-encoded driver.Info (name, ports, instance dir and feature flags).
+  // JSON keeps it in sync with the Go type. The full set of fields is defined at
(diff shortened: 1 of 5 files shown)


hack/gogenerate/protoc-doc.sh
unsuman, line 1: Thanks, also needs the CI to check if proto changes are reflected in the docs.

On a second thought, it feels redundant to add a copy of the proto comments in the docs when we could simply point the developer directly to the proto file but no strong opinion here, would love to know what other maintainers think

cc @lima-vm/maintainers

pkg/driver/external/driver.proto
```diff
@@ -4,86 +4,201 @@ import "google/protobuf/empty.proto";
...
+  // GuestAgentConn establishes the guest agent connection. Over this transport
+  // the server bridges the connection to a unix socket in the instance
+  // directory when it is not already a unix socket, so the client relies on
+  // that socket rather than on a returned connection.
   rpc GuestAgentConn(google.protobuf.Empty) returns (google.protobuf.Empty);
 
+  // Configure sets the instance configuration on the driver. It must be called
+  // before RPCs that depend on the config. Note that FillConfig from the Go
+  // interface is not exposed over this transport.
   rpc Configure(SetConfigRequest) returns (google.protobuf.Empty);
```
unsuman: Needs a rewrite since FillConfig will no longer exist. See https://github.com/lima-vm/lima/pull/4877 and will need a rebase after merging

Also:
```suggestion
  rpc Configure(SetConfigRequest) returns (SetConfigResponse);
```

hack/gogenerate/protoc-doc.sh
afbjorklund: I am not against javadocs and friends, I think they serve a purpose - especially on larger projects. This API is probably small enough that it is a little overkill, but the nice thing about rendering the docs is that there is no duplication. If you change the code, then the protoc-gen-doc will update the docs as well.

Now, if it would been a simpler RPC without having to generate **code** and other monstrosities then it would probably have been enough with the code and some samples. But we're talking about gRPC here. The generic types at the bottom are a bit superfluous, but good to have everything in one place.

My main issue was not the docs itself, but the breakage.

i.e. having to duplicate a lot of stubs, and watch them break...

hack/gogenerate/protoc-doc.sh
```diff
@@ -0,0 +1,47 @@
...
+TMP_DIR="$(mktemp -d)"
+trap 'rm -rf "$TMP_DIR"' EXIT
+
+set -x
+protoc --doc_out="$TMP_DIR" --doc_opt=markdown,"$OUT_FILE" "$PROTO"
+set +x
+
+# Prepend docsy front matter and a generated-file marker, then append the
+# generated reference with trailing whitespace stripped, so the whole file stays
+# generated and passes editorconfig-checker. perl is used because `sed -i` is not
```
afbjorklund: You can do some shenanigans with argument ordering, if you want it portable:

`sed -i -e 's/foo/bar/' file && rm -f ./-e`

You could also sed to a copy and mv

Adding perl is more of a tech debt, not even sure if the blank lines were an issue?

website/content/en/docs/dev/external-driver-api.md
AkihiroSuda, line 1: Auto-generated files should be generated during `make docsy` in netlify
https://github.com/lima-vm/lima/blob/1ba38dc238036b21491b06ed88918aa0035e59c0/Makefile#L533-L545

AkihiroSuda, line 1: We already have godoc
https://pkg.go.dev/github.com/lima-vm/lima/v2@v2.1.4/pkg/driver#Driver

AkihiroSuda: I suggest opening a separate issue/PR to discuss whether we want to have protoc-gen-doc

ekalinin: Agreed. I'll drop the generated file and the tooling from this PR and open a separate issue (see #5224). 

This PR will keep only the `driver.proto` doc comments and the lifecycle narrative in `drivers.md`.

hack/gogenerate/protoc-doc.sh
ekalinin: Agreed. Since I am descoping the generation script (see the discussion with AkihiroSuda below), this goes away for now. If we revisit `protoc-gen-doc` in the separate issue, I'll use a portable `sed` (or generate at build time) instead.

pkg/driver/external/driver.proto
ekalinin: Thanks! Will do.

website/content/en/docs/dev/external-driver-api.md
ekalinin: Agreed.

hack/gogenerate/protoc-doc.sh
ekalinin: Thanks everyone for the feedback. The core goal of #4896 is to make the driver contract (call ordering, idempotency, per-method semantics) discoverable, and I agree that belongs in `driver.proto` itself, which also feeds godoc, rather than in a committed second copy.

So I'll descope this PR to:
  - keep the detailed doc comments in `driver.proto` (this is the substance of the issue and improves godoc too);
  - keep the hand-written "Driver lifecycle & call ordering" section in `drivers.md`, but point it at the proto file / godoc instead of a generated page;
  - drop the `protoc-gen-doc` tooling, the generated `external-driver-api.md`, `hack/gogenerate/protoc-doc.sh`, the `hack/tools` additions and the `install-protoc-tools` change.

I'll open a separate issue to discuss whether we want `protoc-gen-doc` at all (#5224), and if so wire it into `make docsy` so it is generated at netlify build time (like `limactl.md`) rather than committed.

unsuman:
Needs rebase

website/content/en/docs/dev/drivers.md
```diff
@@ -65,6 +65,13 @@ type Driver interface {
 }
 ```
 
+   The gRPC transport that carries these methods to an external driver is defined
+   in [`driver.proto`](https://github.com/lima-vm/lima/blob/master/pkg/driver/external/driver.proto),
+   whose per-RPC doc comments are the authoritative contract (call ordering,
+   idempotency and per-RPC semantics) for driver authors. The Go side is
+   documented as godoc on the [`driver.Driver`](https://pkg.go.dev/github.com/lima-vm/lima/v2/pkg/driver#Driver)
```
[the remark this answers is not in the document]
ekalinin: Fixed.

website/content/en/docs/dev/drivers.md
```diff
@@ -92,6 +99,44 @@ func main() {
 limactl create myinstance --vm-type=mydriver template:default
 ```
 
+## Driver lifecycle & call ordering
+
+Drivers are unaware whether they are internal or external, but external driver
+authors need to know when each method is invoked. During instance create/start,
+Lima (`pkg/instance` and `pkg/hostagent`) drives the methods in this order:
```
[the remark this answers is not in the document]
ekalinin: Fixed.

website/content/en/docs/dev/drivers.md
```diff
@@ -92,6 +99,44 @@ func main() {
...
+
+1. **`Configure`** - sets the instance configuration on the driver. It must run
+   before any method that depends on the config.
+2. **`Validate`** - rejects a config the driver cannot support. Returning an
+   error aborts create/start.
```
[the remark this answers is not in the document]
ekalinin: Fixed.

website/content/en/docs/dev/drivers.md
```diff
@@ -92,6 +99,44 @@ func main() {
...
+6. **`AdditionalSetupForSSH`** - runs right after `Start` returns (boot
+   initiated, the guest is not necessarily up yet), before the first SSH
+   connection.
+7. **`SSHAddress`** - re-queried after `Start` only when `Info` reports
+   `Features.dynamicSSHAddress` (e.g. WSL2, whose address is not known until the
+   VM is up).
+
+The guest-agent socket is either forwarded by the host agent or reached over
+SSH, depending on what **`ForwardGuestAgent`** returns; when forwarding is
+needed, **`GuestAgentConn`** establishes the connection.
```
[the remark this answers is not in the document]
ekalinin: Fixed.

pkg/driver/external/driver.proto
```diff
@@ -4,90 +4,204 @@ import "google/protobuf/empty.proto";
...
 option go_package = "github.com/lima-vm/lima/v2/pkg/driver/external";
 
+// Driver is the transport used by an external VM driver. It is the gRPC
+// projection of the Go `driver.Driver` interface (pkg/driver/driver.go): the
+// host agent runs a gRPC client and each external driver binary
+// (lima-driver-<name>) runs the server. The RPCs mirror the interface methods
+// one-to-one, so this file is the authoritative contract for driver authors.
+//
+// Call ordering during instance create/start (as driven by pkg/instance and
+// pkg/hostagent):
```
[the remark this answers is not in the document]
ekalinin: Fixed.

pkg/driver/external/driver.proto
```diff
@@ -4,90 +4,204 @@ import "google/protobuf/empty.proto";
...
+// stay in sync with the Go types without duplicating them here.
+//
+// Concurrency: the create/start RPCs above are invoked sequentially in the
+// order listed. Once the VM is running, the host agent polls read-only RPCs
+// (notably Info and ForwardGuestAgent) from background goroutines, so the
+// driver server must be safe to handle concurrent RPCs.
 service Driver {
+  // Validate reports whether the driver supports the currently configured
+  // instance. It is called early during create/start, before Create. Returning
+  // an error aborts the operation.
```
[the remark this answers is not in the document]
ekalinin: Fixed.

pkg/driver/external/driver.proto
```diff
@@ -4,90 +4,204 @@ import "google/protobuf/empty.proto";
...
+
+  // DeleteSnapshot removes the snapshot identified by the given tag.
   rpc DeleteSnapshot(DeleteSnapshotRequest) returns (google.protobuf.Empty);
+
+  // ListSnapshots returns the driver-formatted list of existing snapshots.
   rpc ListSnapshots(google.protobuf.Empty) returns (ListSnapshotsResponse);
 
+  // ForwardGuestAgent reports whether the guest agent socket must be forwarded
+  // by the host agent. When false, the guest agent is reached over SSH instead.
   rpc ForwardGuestAgent(google.protobuf.Empty) returns (ForwardGuestAgentResponse);
```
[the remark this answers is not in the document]
ekalinin: Fixed

pkg/driver/external/driver.proto
```diff
@@ -4,90 +4,209 @@ import "google/protobuf/empty.proto";
...
   // AdditionalSetupForSSH provides additional setup required for SSH connection.
-  // It is called after VM is started, before first SSH connection.
+  // It is called right after Start returns (boot initiated, the guest is not
+  // necessarily up yet), before the first SSH connection.
   rpc AdditionalSetupForSSH(google.protobuf.Empty) returns (google.protobuf.Empty);
 }
 
+// BootScriptsResponse carries the boot scripts injected into the VM.
 message BootScriptsResponse {
+  // scripts maps a script path to its content. Each key must be
```
alexandear: The generated code is the following:

```go
// BootScriptsResponse carries the boot scripts injected into the VM.
type BootScriptsResponse struct {
	// scripts maps a script path to its content. Each key must be
	// "boot.<OS>/<SCRIPT>", or "<SCRIPT>" as a deprecated alias for
	// "boot.Linux/<SCRIPT>".
	Scripts       map[string][]byte `protobuf:"bytes,1,rep,name=scripts,proto3" json:"scripts,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
	unknownFields protoimpl.UnknownFields
	sizeCache     protoimpl.SizeCache
}
```

We have the field name `Scripts` and the mismatched comment that starts with `scripts`.

I propose removing the field name here and below:

```suggestion
  // Maps a script path to its content. Each key must be
```

pkg/driver/external/driver.proto
```diff
@@ -4,90 +4,209 @@ import "google/protobuf/empty.proto";
...
 }
 
+// SetConfigRequest carries the instance configuration for Configure.
 message SetConfigRequest {
+  // instance_config_json is the JSON-encoded limatype.Instance. JSON keeps it
+  // in sync with the Go type.
   bytes instance_config_json = 1;
 }
 
 message SetConfigResponse {
```
unsuman, line 171: This is undocumented

pkg/driver/external/driver.proto
```diff
@@ -4,90 +4,209 @@ import "google/protobuf/empty.proto";
...
 message SSHAddressResponse {
+  // address is the host/IP used to SSH into the guest.
   string address = 1;
 }
 
-message InfoResponse{
+// InfoResponse carries driver metadata and capability flags.
+message InfoResponse {
+  // info_json is the JSON-encoded driver.Info (name, ports, instance dir and
+  // feature flags). JSON keeps it in sync with the Go type.
```
unsuman: We can link to the available fields: https://pkg.go.dev/github.com/lima-vm/lima/v2/pkg/driver#DriverFeatures

```suggestion
// InfoResponse carries driver metadata and capability flags.
message InfoResponse {
  // info_json is the JSON-encoded driver.Info (name, ports, instance dir and
  // feature flags). JSON keeps it in sync with the Go type. The full set of fields are defined in https://pkg.go.dev/github.com/lima-vm/lima/v2/pkg/driver#DriverFeatures
```

And possibly can we also add comments to [`driver.Info{}`](https://github.com/lima-vm/lima/blob/0c54bfef57f57ecfe9a20542934b7ac1451e83ad/pkg/driver/driver.go#L118) in this PR?

pkg/driver/external/driver.proto
```diff
@@ -4,90 +4,209 @@ import "google/protobuf/empty.proto";
...
   rpc CreateDisk(google.protobuf.Empty) returns (google.protobuf.Empty);
+
+  // Start begins booting the VM. It is a server-streaming RPC that mirrors the
+  // `(chan error, error)` return of the in-process interface: the call returns
+  // once boot has been initiated (the initial StartResponse), not once the
+  // guest has finished booting, and later boot errors arrive on the stream. See
+  // StartResponse for the streaming protocol.
   rpc Start(google.protobuf.Empty) returns (stream StartResponse);
+
+  // Stop terminates the running VM. It returns an error if the shutdown fails.
```
unsuman:
```suggestion
  // Stop terminates the running VM. It blocks until the VM has fully exited
  // (or the driver's internal shutdown deadline elapses) and returns a gRPC
  // error status if the shutdown failed.
```

unsuman (changes requested):
Thanks, the PR is almost ready to be merged for `v2.2` after addressing the comments  but it's non-critical. If it isn’t updated in time(before 21 July) for the GA release, we’ll move it to `v2.3`

pkg/driver/external/driver.proto
ekalinin: Thanks, fixed.

pkg/driver/external/driver.proto
ekalinin: > And possibly can we also add comments to [driver.Info{}](https://github.com/lima-vm/lima/blob/0c54bfef57f57ecfe9a20542934b7ac1451e83ad/pkg/driver/driver.go#L118) in this PR?

Done.

pkg/driver/external/driver.proto
ekalinin: Good point, thanks! Fixed.

ekalinin:
> Thanks, the PR is almost ready to be merged for `v2.2` after addressing the comments but it's non-critical. If it isn’t updated in time(before 21 July) for the GA release, we’ll move it to `v2.3`

Thanks!

I've pushed changes addressing all the review comments, so this should be ready for another look well before 21 July.

unsuman (approved):
Thanks, LGTM!

PR #5222, opened 2026-07-11
Minor improvements to kubernetes documentation

afbjorklund:
Fixes the single quote issue mentioned on Slack, and also the kubernetes.lima issue with instance name.

```
No k3s or k8s existing instances found. Either create one with
limactl create --name=k3s template:k3s
limactl create --name=k8s template:k8s
or set LIMA_INSTANCE to the name of your Kubernetes instance
```

The machine test case can still use special node name and the yq params, this change was for humans...

### DIFF
--- a/templates/k3s.yaml
+++ b/templates/k3s.yaml
@@ -20,7 +20,7 @@
 # (The token for the start command printed here)
 #
 # $ limactl start --name k3s-1 --network lima:user-v2 template:k3s \
-#                 --set '.param.url="<URL_FROM_ABOVE>" | .param.token="<TOKEN_FROM_ABOVE>"'
+#                 --param url="<URL_FROM_ABOVE>" --param token="<TOKEN_FROM_ABOVE>"
 
 minimumLimaVersion: 2.0.0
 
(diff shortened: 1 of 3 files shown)


templates/k3s.yaml
```diff
@@ -13,14 +13,14 @@
 # A multi-node cluster can be created by starting multiple instances of this template
 # connected via the `lima:user-v2` network.
 #
-# $ limactl start --name k3s-0 --network lima:user-v2 template:k3s
-# $ printf "https://lima-%s.internal:6443\n" k3s-0
+# $ limactl start --name k3s --network lima:user-v2 template:k3s
+# $ printf "https://lima-%s.internal:6443\n" k3s
```
AkihiroSuda, line 17: What was the problem with having `-0` suffix?

afbjorklund: The kubectl.lima wrapper doesn't find the instance, so you need to remember to run the commands from the control plane (the KUBECONFIG) in order to access the worker that was created

You can set LIMA_INSTANCE, but it is an extra step

AkihiroSuda: The wrapper could check the existence of `k3s` and fallback to `k3s-0`?

afbjorklund: But why does it need a special name, just because it was started with user-v2 network?

AkihiroSuda (approved):

PR #5230, opened 2026-07-12
reject line breaks in ssh config option values

SABITHSAHEB:
1. sshutil.Format writes each ssh option into the instance ssh.config (and `limactl show-ssh` output) verbatim.
2. the User= value is user.name, which validate.go never checks for line breaks.
3. a newline in it injects extra ssh_config directives such as ProxyCommand (which ssh runs on the host) under the instance Host block; user.name is settable from a shared template.

Format now rejects any option value containing a line break; covered by a new test in pkg/sshutil (`go test ./pkg/sshutil/`).

### DIFF
--- a/pkg/sshutil/format.go
+++ b/pkg/sshutil/format.go
@@ -63,6 +63,13 @@ func quoteOption(o string) string {
 // Format formats the ssh options.
 func Format(w io.Writer, sshPath, instName string, format FormatT, opts []string) error {
 	fakeHostname := hostname.FromInstName(instName) // TODO: support customization
+	for _, o := range opts {
+		// A line break in an option value (e.g. a crafted user.name) would inject
+		// extra ssh_config directives such as ProxyCommand into the generated config.
+		if strings.ContainsAny(o, "\r\n") {
+			return fmt.Errorf("ssh option %#q contains a line break", o)
+		}
+	}
 	switch format {
 	case FormatCmd:
 		args := []string{sshPath}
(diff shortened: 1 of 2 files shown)


pkg/sshutil/format.go
```diff
@@ -63,6 +63,13 @@ func quoteOption(o string) string {
 // Format formats the ssh options.
 func Format(w io.Writer, sshPath, instName string, format FormatT, opts []string) error {
 	fakeHostname := hostname.FromInstName(instName) // TODO: support customization
+	for _, o := range opts {
+		// A line break in an option value (e.g. a crafted user.name) would inject
```
AkihiroSuda, line 67: How craftable?

SABITHSAHEB: The value is `user.name` from the instance's lima.yaml, which Lima never validates. The default template even spells it out: "An explicitly specified username is not validated by Lima." YAML lets a double-quoted scalar carry a literal newline, so

```yaml
user:
  name: "lima\n  ProxyCommand touch /tmp/pwned"
```

unmarshals into `*inst.Config.User.Name` with the newline intact (validate.go only checks `user.shell` and `user.home`). That flows through `SSHOpts` as `User=<name>` and Format writes it verbatim into `~/.lima/<inst>/ssh.config` under the Host block, adding a `ProxyCommand` line. ssh runs ProxyCommand on the host, so anything that later reads that file with `ssh -F` (limactl shell, show-ssh, the mcp sftp path) executes it on the host.

The realistic vector is a shared/remote template: you `limactl start` a template someone hands you, and the generated ssh.config picks up a host command. A template is expected to do things inside the guest, not run commands on the host, so this crosses that boundary. Guarding at Format is the cheap chokepoint since every format and both callers go through it.

AkihiroSuda (approved):
Thanks

PR #5233, opened 2026-07-13
reject local file references from remote templates in absPath

Nashit-h:
1. absPath resolves base/provision/probe locators; for a template fetched over http(s)/github it returned file:// URLs and absolute local paths verbatim.
2. embedBase and embedAllScripts then Read those paths and inline the bytes into a provisioning script, so a downloaded template can read and exfiltrate arbitrary host files (ssh keys, etc.).

Relative references under a remote base already resolve remotely via url.JoinPath, so this only rejects the local file:// and absolute-path cases, next to the existing ../ guard. Added regression tests in abs_test.go covering the file://, absolute-path, and tilde cases under a remote base.

### DIFF
--- a/pkg/limatmpl/abs.go
+++ b/pkg/limatmpl/abs.go
@@ -102,13 +102,37 @@ func basePath(locator string) (string, error) {
 	return withVolume(base)
 }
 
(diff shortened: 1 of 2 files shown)


Nashit-h:
gentle ping

Issue #5171, opened 2026-06-29
`limactl (create|start)` should support resuming partial downloads


AkihiroSuda:
`limactl (create|start)` should support resuming partial downloads.
This might not be straightforward, as it has to support concurrent sessions of limactl.

afbjorklund:
The current behaviour is to leave a "data.tmp.XXXXXX" in the cache folder.

* https://github.com/lima-vm/lima/issues/2722

They can accumulate in size, if you are prone to sometimes doing ctrl-C

Maybe they need some kind of cleanup command (the nuclear prune "works")

ekalinin:
Hi!

Took a look at this one -- the resume request and the `data.tmp.XXXXXX` buildup that @afbjorklund mentioned look like the same underlying problem, so here's a single approach that covers both.

What's going on: the download is streamed into a per-PID temp file (`data.tmp.{pid}.{n}`, the scheme from #2722) and only cleaned up via a `defer`. Since `limactl create/start` don't catch SIGINT, a Ctrl-C skips that cleanup -- and because the name is unique per process, every interrupt leaves a *fresh* orphan behind. That's the pile-up. And with no Range support, those partial bytes are discarded anyway.

The useful part: the cache download already runs under an exclusive dir lock (`lockutil.WithDirLock`), so only one process ever writes per URL. That makes a single stable partial file safe there — which is what unlocks resume.

Worth being explicit about parallelism: this doesn't serialize different images. Every URL gets its own cache dir (`by-url-sha256/<sha256(url)>/`) with its own lock, so distinct images already download fully in parallel -- the stable `data.part` lives inside a single URL's dir. The only thing serialized is the same URL from multiple processes, which is intentional (per #2722): the first one downloads, the rest wait on the lock and reuse the cache instead of re-fetching the same image N times.

The proposal:

1. Resume: write into a fixed `data.part` instead of `data.tmp.{pid}.{n}`. If it already exists (size `N`), send `Range: bytes=N-` plus `If-Range: <Last-Modified/ETag>`. Remote unchanged → `206`, and the new bytes are appended (existing bytes get re-hashed into the digester first, so the checksum still validates); remote changed or no Range support → `200`, and the download restarts. Everything else stays the same -- verify digest, atomic rename to `data`. As a bonus this caps the leftovers at one reusable file per URL instead of an endless pile. The uncached path stays untouched.

[...]

PR #5234, opened 2026-07-13
feat(prune): sweep leftover partial downloads from the cache

ekalinin:
Add `downloader.RemoveStaleTempFiles`, which removes leftover "data.part" and legacy "data.tmp.*" files from the cache while keeping the completed "data" files. Each cache entry is locked before its partial files are removed, so a download in progress is never disturbed.

"limactl prune --keep-referred" now calls it after pruning unreferred entries, giving a non-nuclear way to reclaim space taken by interrupted downloads (e.g. after Ctrl-C) without dropping valid cached images.

Refs #5171

### DIFF
--- a/cmd/limactl/prune.go
+++ b/cmd/limactl/prune.go
@@ -64,7 +64,9 @@ func pruneAction(cmd *cobra.Command, _ []string) error {
 			}
 		}
 	}
-	return nil
+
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks

PR #5237, opened 2026-07-13
Treat instances created before Lima v0.20 as older than any release

jandubois:
Instances created before Lima v0.20 have no `lima-version` file. FillDefault() treated the missing file as an empty version until commit 2076ff1d (v2.0.0-alpha.0), which made `ExistingLimaVersion()` report the current version instead. These old instances then get defaults meant for new ones.

QEMU instances switch from `reverse-sshfs` to `9p` mounts, and the guest home directory defaults to `/home/<user>.guest`, renamed from `/home/<user>.linux` in commit 4a10d927 (v2.1.0-beta.0). On `alpine-iso` instances with user-mode provisioning, `/home/<user>.guest` is missing because `/home` comes from the persistent data volume. Commit 032fc350 already restores the compatibility symlink after that mount, but this corrects the defaults at their source.

`versionutil` already treats the empty version as older than any release; `osutil.LimaUser` was the only consumer that read it as an unknown build and assumed the newest defaults.

### DIFF
--- a/pkg/limayaml/defaults.go
+++ b/pkg/limayaml/defaults.go
@@ -874,20 +874,27 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 	fixUpForPlainMode(y)
 }
 
-// ExistingLimaVersion returns empty if the instance was created with Lima prior to v0.20.
+// ExistingLimaVersion returns the version that created the instance, the empty string
+// for instances created before Lima v0.20 (which have no recorded version), or the
+// current version when the instance doesn't exist yet.
 func ExistingLimaVersion(instDir string) string {
 	if !IsExistingInstanceDir(instDir) {
 		return version.Version
 	}
 
 	limaVersionFile := filepath.Join(instDir, filenames.LimaVersion)
-	if b, err := os.ReadFile(limaVersionFile); err == nil {
+	b, err := os.ReadFile(limaVersionFile)
+	switch {
+	case err == nil:
 		return strings.TrimSpace(string(b))
-	} else if !errors.Is(err, os.ErrNotExist) {
+	case errors.Is(err, os.ErrNotExist):
+		// Instances created before Lima v0.20 have no version file. versionutil treats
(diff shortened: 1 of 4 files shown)


pkg/limayaml/defaults.go
```diff
@@ -874,20 +874,30 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
...
 
 	limaVersionFile := filepath.Join(instDir, filenames.LimaVersion)
-	if b, err := os.ReadFile(limaVersionFile); err == nil {
+	b, err := os.ReadFile(limaVersionFile)
+	switch {
+	case err == nil:
 		return strings.TrimSpace(string(b))
-	} else if !errors.Is(err, os.ErrNotExist) {
+	case errors.Is(err, os.ErrNotExist):
+		return lastVersionWithoutVersionFile
```
AkihiroSuda: Can we just return an empty string

jandubois: Well, that is exactly what this PR is trying to fix.

We could leave it at returning the empty string, and just fix `pkg/osutil/user.go:153` to treat the empty string as very old, and not as unknown.

But I feel it is kind of brittle because all code accessing the value needs to know that the empty string means "older than 0.20" and not "unknown, but likely new or even latest".

This seems to work right now in all other places except in `user.go`, but it is not obvious that this treatment is intentional, and not just correct by accident.

Setting the version to `0.19` should be safe everywhere; there is no code that discriminates on earlier versions (because we never recorded them).

So I can change it, if you want, but I really think this variant is more robust.

jandubois:
I've updated the PR description; the boot failure due to the home directory change only affects `alpine-iso` instances with `user` boot scripts.

jandubois:
@AkihiroSuda I've changed the fix to use the empty string to mean "older than v0.20" and fixed `osutil.LimaUser` instead to treat it correctly. PTAL

AkihiroSuda (approved):
Thanks

PR #5238, opened 2026-07-13
Create the guest home symlink after the data volume is mounted

jandubois:
The symlink never reached an existing Alpine instance, where `/home` is mounted from a persistent data volume, and it was never created at all on WSL2, where the driver creates the user later, in `02-no-cloud-init-setup.sh`.

Note: this PR renames `00-guest-home.sh` → `05-guest-home.sh`


AkihiroSuda (approved):
Thanks

PR #5241, opened 2026-07-14
update(docs): add external driver transport changes to breaking-changes

unsuman:
Addresses https://github.com/lima-vm/lima/pull/4877#issuecomment-4965065126


AkihiroSuda (approved):
Thanks

Issue #5188, opened 2026-07-05
create: cached-image fallback fails when the freshness HEAD check fails (digest-less cache)


wfscot:
### Description

When creating an instance whose image is already in the download cache but the cache entry is digest-less, and the freshness check (HEAD for `last-modified`) fails, `limactl create` logs the advertised fallback:

```
level=info msg="Failed to retrieve last-modified for cached digest-less image; using cached image."
level=info msg="Using cache \"~/Library/Caches/lima/download/by-url-sha256/<sha>/data\""
```

…but then exits fatally instead of using the cache:

```
level=fatal msg="open /Users/<user>/.lima/<name>/image: no such file or directory"
```

(v2.1.4; on v2.0.3 the same failure says `basedisk` instead of `image`.)

Comparing against the successful path shows what's missing. When the HEAD succeeds, create logs:

```
level=info msg="Using cache \"…/data\""
level=info msg="Converting \"/Users/<user>/.lima/<name>/image\" (qcow2) to a raw disk \"/Users/<user>/.lima/<name>/disk\""
level=info msg="Expanding to 30GiB"
```

In the degraded path the `Converting` step never happens: the cached image is apparently never copied into the instance directory, and the next step fatals opening the file the copy would have created. The instance is left registered but broken (shows `Stopped` in `limactl list`; a retry then fails with "already exists").

### How we hit it

`cloud.debian.org` is served by a mirror pool; one mirror IP was serving TLS handshakes in 13–17 s (TCP connect was normal, ~150 ms). The downloader uses `http.DefaultClient`, whose default `TLSHandshakeTimeout` is 10 s, so the HEAD timed out deterministically for as long as DNS kept resolving to that mirror — making the fallback path (and this fatal) 100% reproducible for hours. There appears to be no knob to raise the timeout.

### Reproduction

1. A template with a digest-less https image already present in the cache from an earlier successful create, e.g.:

   ```yaml
   images:
     - location: https://cloud.debian.org/images/cloud/bookworm/latest/debian-12-generic-arm64.qcow2
       arch: aarch64
   ```

2. Make the HEAD request fail (slow/unreachable mirror; anything that trips the 10 s TLS handshake timeout works).
3. `limactl create --name test --tty=false template.yaml`

Fatal as above; the same command succeeds when the HEAD succeeds, and also succeeds with a local-file `location:` (network-free), so only the degraded branch is broken.

### Expected behavior

[...]

PR #5242, opened 2026-07-14
fix(downloader): add regression test for cached-image fallback (#5188)

ekalinin:
When the freshness HEAD check for a digest-less cached image fails, `getCached()` must still copy the cached data to the local path instead of just reporting `StatusUsedCache` without copying. This was broken in v2.0.3/v2.1.4 and got fixed as a side effect of commit b13db25f, but had no dedicated test coverage.

Refs #5188


AkihiroSuda (approved):
Thanks

PR #5243, opened 2026-07-14
fix(cli): `limactl list` to skip broken instances during clone/delete

### LINKED ISSUE #5236: `limactl list` fails completely while another instance is being cloned or deleted
### Description

`limactl clone` creates the new instance directory before it writes `lima.yaml` into it. `limactl delete` removes `lima.yaml` before it removes the directory. In both cases there is a window where `~/.lima/<name>/` exists without a readable `lima.yaml`.

Any `limactl list` that runs during that window exits 1 and prints nothing. It does not just skip or flag the incomplete instance: every other instance disappears from the output too.

Clone in particular can take some time, and users may be doing a parallel operation or lima may be underneath some other tool

## Steps to reproduce

No VM is needed, you can reproduce this with an empty directory.

```console
$ mkdir ~/.lima/probe

$ limactl list
time="..." level=warning msg="The directory `/home/user/.lima` does not look like a valid Lima directory: stat /home/user/.lima/probe/lima.yaml: no such file or directory"
time="..." level=fatal msg="unable to load instance probe: open /home/user/.lima/probe/lima.yaml: no such file or directory"

$ echo $?
1

$ rmdir ~/.lima/probe
$ limactl list          # works again
```

Same result with `--format json`.

## Expected

The incomplete instance is skipped, or reported as `Broken`. The other instances are still listed and the command exits 0.

## Actual

Exit 1, no output, no instances listed.

## Which operations trigger it

Measured by polling `limactl list` every 0.2s for the duration of each operation (limactl 2.1.3, Linux, QEMU):

| operation | polls | `limactl list` failures |
| --- | --- | --- |
| create (`limactl start --name`) | 364 | 0 |
| start | 133 | 0 |
| stop | 24 | 0 |
| clone | 7 | 7 |
| delete | 3 | 2 |

Start, stop and create never remove or precede `lima.yaml`, so they are unaffected. Only clone and delete pass through the bad state.

## Cause

https://github.com/lima-vm/lima/blob/487ec81f7f60459a8cfe44189a3cdd636469528f/cmd/limactl/list.go#L214-L222

A few lines above, an invalid store directory is tolerated with `logrus.Warnf`, and the `Broken` status already exists to represent an instance that cannot be loaded, so the machinery to degrade gracefully is there.

## Workaround

Naming the instances explicitly avoids it, because the incomplete instance is then never inspected:

```console
$ mkdir ~/.lima/probe
$ limactl list default other --format json   # succeeds, warning on stderr only
```

unsuman:
Fixes #5236 

> [!NOTE]
> `Assisted-by: Copilot:Claude Opus 4.7`

### DIFF
--- a/cmd/limactl/delete.go
+++ b/cmd/limactl/delete.go
@@ -13,6 +13,8 @@ import (
 
 	"github.com/lima-vm/lima/v2/pkg/autostart"
 	"github.com/lima-vm/lima/v2/pkg/instance"
+	"github.com/lima-vm/lima/v2/pkg/limatype/dirnames"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
 	"github.com/lima-vm/lima/v2/pkg/networks/reconcile"
 	"github.com/lima-vm/lima/v2/pkg/store"
 )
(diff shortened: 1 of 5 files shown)


AkihiroSuda (approved):
Thanks

PR #5248, opened 2026-07-15
validate network names and fields used in sudoers commands

Nashit-h:
networks.Config.Validate checks paths.* but not the group or the per-network name, mode, and interface, even though sudoers.go writes them verbatim into the generated /etc/sudoers.d/lima file and reconcile.go splits StartCmd on spaces to run socket_vmnet under sudo. A network name containing a newline adds an arbitrary directive to that sudoers file, and a value with a space injects an extra argument into the root command. Validate now requires network names to be valid identifiers and rejects whitespace or control characters in the group, mode, and interface, matching the whitespace rule already enforced on paths.

### DIFF
--- a/pkg/networks/validate.go
+++ b/pkg/networks/validate.go
@@ -13,10 +13,40 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/lima-vm/lima/v2/pkg/identifiers"
 	"github.com/lima-vm/lima/v2/pkg/osutil"
 )
 
 func (c *Config) Validate() error {
+	// The group name and the per-network name/mode/interface are interpolated
+	// verbatim into the sudoers file (sudoers.go) and into the socket_vmnet
+	// command that reconcile.go runs via sudo after splitting it on spaces
+	// (commands.go). A value with whitespace injects an extra argument, and a
+	// newline in a network name adds an arbitrary directive to the generated
+	// sudoers file, so require them to be valid identifiers. Interface is empty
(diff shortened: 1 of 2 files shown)


pkg/networks/validate.go
```diff
@@ -51,7 +74,18 @@ func (c *Config) Validate() error {
...
 		return fmt.Errorf("networks.yaml: %#q (`paths.socketVMNet`) has to be installed", pathsMap["socketVMNet"])
 	}
-	// TODO(jandubois): validate network definitions
+	return nil
+}
+
+// validateNetworkField rejects values that would break out of the generated
+// sudoers file or inject arguments into the sudo-run daemon commands, mirroring
+// the whitespace restriction already enforced on `paths.*` by validatePath.
+func validateNetworkField(field, value string) error {
```
AkihiroSuda: Could we just use https://github.com/lima-vm/lima/blob/de48002a4b26cda22277a4a862ef94354005b315/pkg/identifiers/validate.go#L57

Nashit-h: Good call. Dropped the custom whitespace check and use identifiers.Validate for the name, mode, interface, and group. Interface is empty on non-bridged networks (and group defaults to admin when unset), so those are only checked when a value is present.

AkihiroSuda:
Please squash the commits

Nashit-h:
Squashed down to a single commit. No content change from the last push, just the two commits folded together.

AkihiroSuda (approved):
Thanks

PR #5250, opened 2026-07-15
reject qemu packet exceeding the read buffer size

SABITHSAHEB:
1. qemuPacketConn.Read reads a uint32 size prefix from the QEMU packet socket and slices b[:size] without checking size against len(b).
2. a frame whose size prefix is larger than the read buffer (io.Copy in forwardPackets uses a 32 KiB buffer) makes the slice panic with an out-of-range and takes down the vz networking goroutine.
Return an error when size is larger than len(b), which the method already documents as a broken-protocol condition where the socket must be closed.

### DIFF
--- a/pkg/driver/vz/network_darwin.go
+++ b/pkg/driver/vz/network_darwin.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"os"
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
How does this issue actually happen?

AkihiroSuda (approved):
Not sure how this actually happens, but seems safe to merge anyway

SABITHSAHEB:
To answer the question above: it shouldn't happen with a well-behaved socket_vmnet, since real frames stay far below io.Copy's 32 KiB buffer. But the size prefix is read straight off the wire, so if the stream ever desyncs (a peer bug, a truncated write, or something else listening on the configured socket path) the next 4 payload bytes get parsed as a length, which can be anything up to 4 GiB. Before the check that was a slice panic taking down the process instead of an error that just closes the connection, which is what the Read contract already says should happen. TestQemuPacketConnReadOversize in this PR reproduces it over net.Pipe. Thanks for merging.

PR #5251, opened 2026-07-15
Add parameter for Kubernetes apt-proxy (k8s)

afbjorklund:
Now supports a parameter "proxy" for pulling packages from:

proxy: "http://host.lima.internal:3142"

A pull-through cache for pkgs.k8s.io (and the apt CDN) at:

https://github.com/afbjorklund/apt-proxy

Issue #5258

----

`apt-proxy -cachedir /var/tmp/packages -port 3142 &`

```
Hit:1 http://archive.ubuntu.com/ubuntu resolute InRelease
Hit:2 http://archive.ubuntu.com/ubuntu resolute-updates InRelease
Hit:3 http://archive.ubuntu.com/ubuntu resolute-backports InRelease
Hit:5 http://security.ubuntu.com/ubuntu resolute-security InRelease
Get:4 http://host.lima.internal:3142/HTTPS///pkgs.k8s.io/core:/stable:/v1.36/deb  InRelease [1227 B]
Get:6 http://host.lima.internal:3142/HTTPS///pkgs.k8s.io/core:/stable:/v1.36/deb  Packages [5666 B]
Get:1 http://host.lima.internal:3142/HTTPS///pkgs.k8s.io/core:/stable:/v1.36/deb  kubeadm 1.36.2-2.1 [12.6 MB]
Get:2 http://host.lima.internal:3142/HTTPS///pkgs.k8s.io/core:/stable:/v1.36/deb  kubectl 1.36.2-2.1 [11.8 MB]
Get:3 http://host.lima.internal:3142/HTTPS///pkgs.k8s.io/core:/stable:/v1.36/deb  kubernetes-cni 1.9.1-1.1 [39.0 MB]
Get:4 http://host.lima.internal:3142/HTTPS///pkgs.k8s.io/core:/stable:/v1.36/deb  kubelet 1.36.2-2.1 [13.4 MB]
```

### DIFF
--- a/templates/k8s.yaml
+++ b/templates/k8s.yaml
@@ -53,14 +53,22 @@ provision:
     # Installing kubeadm, kubelet and kubectl
     export DEBIAN_FRONTEND=noninteractive
     apt-get update
+    {{- if .Param.proxy }}
+    apt-get install -y --no-upgrade ca-certificates curl
+    {{- else}}
     apt-get install -y apt-transport-https ca-certificates curl
+    {{- end}}
     STABILITY="{{.Param.stability}}"
     {{if eq .Param.release "stable" }}
     RELEASE=$(curl -L -s https://dl.k8s.io/release/stable.txt | cut -d'.' -f1-2)
     {{else}}
     RELEASE="{{.Param.release}}"
     {{end}}
+    {{- if .Param.proxy }}
+    echo "deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] {{.Param.proxy}}/HTTPS///pkgs.k8s.io/core:/${STABILITY}:/${RELEASE}/deb/ /" | sudo tee /etc/apt/sources.list.d/kubernetes.list
+    {{- else}}
     echo "deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/${STABILITY}:/${RELEASE}/deb/ /" | sudo tee /etc/apt/sources.list.d/kubernetes.list
+    {{- end}}
     curl -fsSL https://pkgs.k8s.io/core:/${STABILITY}:/${RELEASE}/deb/Release.key | sudo gpg --dearmor -o /etc/apt/keyrings/kubernetes-apt-keyring.gpg
     apt-get update
     apt-get install -y kubelet kubeadm kubectl && apt-mark hold kubelet kubeadm kubectl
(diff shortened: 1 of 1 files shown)


templates/k8s.yaml
```diff
@@ -53,14 +53,22 @@ provision:
     # Installing kubeadm, kubelet and kubectl
     export DEBIAN_FRONTEND=noninteractive
     apt-get update
+    {{- if .Param.proxy }}
+    apt-get install -y --no-upgrade ca-certificates curl
+    {{- else}}
     apt-get install -y apt-transport-https ca-certificates curl
```
AkihiroSuda, line 59: This already covers ca-certificates and curl

templates/k8s.yaml
```diff
@@ -53,14 +53,22 @@ provision:
...
+    {{- end}}
     STABILITY="{{.Param.stability}}"
     {{if eq .Param.release "stable" }}
     RELEASE=$(curl -L -s https://dl.k8s.io/release/stable.txt | cut -d'.' -f1-2)
     {{else}}
     RELEASE="{{.Param.release}}"
     {{end}}
+    {{- if .Param.proxy }}
```
AkihiroSuda, line 67: Could we check `HTTP_PROXY` env var? (Already automatically propagated from host env?)

afbjorklund: We need a special proxy, with the "HTTPS///" feature, so a generic proxy might not work

templates/k8s.yaml
afbjorklund: I wanted to **avoid** having to dial out to install the (now not needed) `apt-transport-https`

Currently they are both installed (curl and ca-certificates), so it is more of a no-op, but...

afbjorklund: This command is in the upstream documentation, but what it does here is to _upgrade_ `curl`.

And I only wanted to see that it is available (installed), I don't care (here) if it is up-to-date or not.

AkihiroSuda: I feel apt-transport-https is an essential package of the OS today.
I don't really get why they don't want to include it in the base OS by default.

afbjorklund: Here it is mostly a problem, since it blocks any caching of the packages...
Note: it does check the package signatures, just no TLS for the download

templates/k8s.yaml
afbjorklund: It would need something like https://mitmproxy.org/ and set up the CA:

https://docs.mitmproxy.org/stable/concepts/certificates/#quick-setup

I'm not sure if it is available for the Mac?

And would be more "work" for this template

afbjorklund: Also, it would be `HTTPS_PROXY`

AkihiroSuda (approved):
Thanks

PR #5257, opened 2026-07-16
Add parameter for Kubernetes apt-proxy (k3s)

afbjorklund:
Now supports a parameter "proxy" for pulling packages from:

proxy: "http://host.lima.internal:3142"

It is not a deb package, but we can still use the apt-proxy.

Need a new URL instead of https_proxy, if we want caching...

Issue #5258

----

`apt-proxy -cachedir /var/tmp/packages -port 3142 &`

If using apt-cacher-ng, it needs to be configured to allow k3s:

```
PfilePatternEx: ^/k3s-io/k3s/releases/download/(v.*)/(sha256sum.*|k3s.*)$
```

Otherwise you will get an error, when trying to cache the k3s:

`403 Forbidden file type or location`

### DIFF
--- a/templates/k3s.yaml
+++ b/templates/k3s.yaml
@@ -37,6 +37,10 @@ provision:
   script: |
     #!/bin/sh
     if [ ! -d /var/lib/rancher/k3s ]; then
+    {{- if .Param.proxy }}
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #5263, opened 2026-07-16
Support `provision` feature on Windows VM

mie313:
## What This PR Changes
This PR enables `provision` feature on Windows VM.

Current provision support has some limitations:
- boot mode isn't supported since Windows VM doesn't have an equivalent step to Cloud-init's init stage
- yq mode isn't supported since Windows VM doesn't have yq engine by default
- all data mode files are stored under C:\ProgramData\Lima\provision\provision.data (path needs to be converted to Windows path style, or we need to soften current validation check)
- skipDefaultDependencyResolution in dependency mode is currently ignored, because currently there is no default dependencies

## How I Tested This
Run `./_output/bin/limactl start template:windows-2025 --debug`

I checked ScheduledTask after restarting my VM:
```powershell
> Get-ScheduledTaskInfo -TaskName "LimaProvisionStartupSystem" -TaskPath "\Lima\"


LastRunTime        : 7/16/2026 11:58:07 PM
LastTaskResult     : 0
NextRunTime        :
NumberOfMissedRuns : 0
TaskName           : LimaProvisionStartupSystem
TaskPath           : \Lima\
PSComputerName     :
```
`LastTaskReuslt: 0` means the task was successfully executed in the startup time.

In addition, under `C:\ProgramData\Lima\provision`, there are directories/files for provision:
```powershell
> ls C:\ProgramData\Lima\provision\


    Directory: C:\ProgramData\Lima\provision


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         7/17/2026  12:54 AM                provision.data
d-----         7/17/2026  12:54 AM                provision.dependency
d-----         7/17/2026  12:54 AM                provision.system
d-----         7/17/2026  12:54 AM                provision.user
-ar---         7/17/2026  12:48 AM            823 startup_runner.ps1
```


pkg/cidata/cidata.go
```diff
@@ -432,30 +432,9 @@ func GenerateISO9660(ctx context.Context, drv driver.Driver, instDir, name strin
...
-				Reader: strings.NewReader(*f.Expression),
-			})
-		case limatype.ProvisionModeBoot:
-			continue
-		case limatype.ProvisionModeAnsible:
-			continue
-		default:
-			return "", fmt.Errorf("unknown provision mode %#q", f.Mode)
-		}
+	layout, err = appendProvisionEntries(layout, instConfig.Provision, false)
```
AkihiroSuda: Not new in this PR, but `includeBoot = false` needs an explanation as a code comment line.
Where are they added instead?

jandubois:
AI review: https://jandubois.github.io/lima/20260725-221419-pr-5263.html

pkg/cidata/cidata.go
mie313: I found that `ProvisionModeBoot` was introduced in https://github.com/lima-vm/lima/pull/1094, and the mode is treated in a different place (stored in TamplateArgs.BootCmds, then embedded in user-data).
I added comments as follows: https://github.com/lima-vm/lima/pull/5263/changes#diff-6a911aa7574099e18369e9477b214a031077350dedc30e60b933ad0f61e022c2R572-R574

In addition, since Windows VM doesn't support `ProvisionModeBoot`, `appendProvisionEntries` does nothing for the mode, and I removed `includeBoot` argument.

mie313:
I also addressed some of AI review comments.

AkihiroSuda (approved):
Thanks

PR #5269, opened 2026-07-17
reject backslash parent-dir traversal in template locators

SABITHSAHEB:
1. absPath, the helper that makes a template's base/provision/probes locators absolute, rejects relative locators containing `../` but not the backslash form `..\`.
2. filepath.Join treats `\` as a separator on Windows, so a relative locator like `..\..\x` escapes the base directory; on `limactl start` of a local template, embedAllScripts reads the resolved provision/probe file and inlines its contents into the config, so a downloaded template can read files outside its own directory.

Rejected `..\` in the same guard so the check holds on every platform. Added a regression case for the backslash form.


jandubois (approved):
Thanks, LGTM

PR #5270, opened 2026-07-17
sshutil, ioutilx: follow-ups from the native Windows OpenSSH review

jandubois:
Follow-up to #5205, closing the review findings it deferred: pair `ssh-keygen` with the selected toolchain, make the `cygpath` fallback fail loudly instead of converting silently, and restore the tests #5196 had that the port dropped.

`SftpServerForSSH` stays exported but unwired — reverse-sshfs mounts and `limactl copy` are still separate follow-ups; it has test coverage in the meantime.

Created with Claude Opus 4.8

### DIFF
--- a/pkg/ioutilx/ioutilx.go
+++ b/pkg/ioutilx/ioutilx.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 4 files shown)


AkihiroSuda (approved):
Thanks

PR #5271, opened 2026-07-17
hack: preserve `variant` field when updating image templates

unsuman:
## What This PR Changes

While updating the images for the GA release (https://github.com/lima-vm/lima/compare/master...unsuman:lima:update-templates), I found that the `update-template-*` scripts were dropping the variant field. This happened because limactl edit replaces the entire image entry, stripping out any fields the script doesn't explicitly emit.

## AI Usage

> [!NOTE]
>  `Assisted-by: Copilot: Claude Opus 4.7`


AkihiroSuda (approved):
Thanks

PR #5274, opened 2026-07-17
ci: run quick lints on docs-only PRs

### LINKED ISSUE #5272: Run editorconfig lint on docs-only PRs
In #5268 we modified just a doc file, missing a final newline at the end. Because it had no code changes, we did not run the `lint` step.

Any following PR with code changes will have a linter error in a file it doesn't touch.

jandubois:
The test workflow owns editorconfig-checker, yamllint, ls-lint, and the lychee hyperlink check, but its paths-ignore skips docs-only changes, so a docs PR can land a violation that only surfaces later, on the next code PR that runs the workflow against the whole tree.

Move those four into a new lint-quick workflow with no paths-ignore. Each of them covers files a docs-only PR touches. shellcheck, shfmt, and the Go/protobuf checks stay behind: they need extra tooling and have nothing to lint in the docs tree.

Closes #5272

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -43,31 +43,14 @@ jobs:
       run: make generate check-generated
     - name: Run nobin
       run: make nobin
-    - name: Run editorconfig-checker
-      run: make editorconfig-checker
-    - name: Run yamllint
-      run: make yamllint
     - name: Install shellcheck
       run: |
         sudo apt-get update
         sudo apt-get install -y shellcheck
-    - name: Run file and directory name linter
-      uses: ls-lint/action@02e380fe8733d499cbfc9e22276de5085508a5bd  # v2.3.1
     - name: Run shellcheck
       run: make shellcheck
     - name: Run shfmt
       run: make shfmt
-    - name: Check hyperlinks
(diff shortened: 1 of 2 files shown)


unsuman (dismissed):
Thanks, should `lint / Lint` be fine? or perhaps `lint / File content`?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,33 @@
+name: lint
```
AkihiroSuda: It is confusing to have two kinds of "lint".

The workflow name and the file name could be `lint-docs` too?

.github/workflows/test.yml
AkihiroSuda: This could be moved to lint-docs too?

AkihiroSuda (changes requested):
https://github.com/lima-vm/lima/pull/5274/changes#r3609179248

.github/workflows/lint.yml
AkihiroSuda: Or something like `lint-basic` `lint-quick`

AkihiroSuda (approved):
Thanks

unsuman (approved):
Thanks!

PR #5276, opened 2026-07-18
verify cached image digest before raw conversion

Nashit-h:
getCached replaces o.expectedDigest with the digest it computes for imgconv/raw, so the validateCachedDigest call that follows compares imgconv/raw.digest against a value read from that same file and can never fail. If the cache entry carries no <ALGO>.digest file, which is what a digest-less download leaves behind, nothing has checked the caller's digest against any bytes by that point, and the image is copied out with ValidatedDigest set. Hashing the cached original before the conversion branch takes over closes the gap, and the added test fails on the current tree where the cached image is accepted against an unrelated digest.


jandubois (approved):
Thanks, LGTM

But leaving it open for somebody else to confirm.

unsuman (approved):
Thanks!

Issue #5152, opened 2026-06-24
WSL2: improve support for generic templates and standard Lima configuration fields


Nachiket-Roy:
### Rootfs format restrictions

The WSL2 driver imports guests via:

```text
wsl.exe --import
```

which only accepts filesystem archives (`.tar`/`.tar.gz`).

As a result, templates referencing standard `.qcow2` or `.raw` cloud images cannot be used with the WSL2 backend.

Currently it is unclear whether this limitation should be:

* explicitly validated and documented, or
* abstracted by adding conversion/import helpers.

### Configuration fields ignored by the backend

`validateConfig()` currently warns and ignores several standard Lima configuration fields, including:

* `mounts`
* `networks`

This makes WSL2 behave differently from other drivers and limits template portability.

### Phase 1: Improve validation and documentation

* Explicitly validate supported image formats for `vmType: wsl2`.
* Provide clear error messages when unsupported image types (`qcow2`, `raw`, etc.) are used.
* Document how users can build or provide WSL-compatible rootfs archives.

### Phase 2: Support standard Lima mount semantics

Investigate mapping Lima `mounts:` entries onto WSL's existing host filesystem integration.

Possible approaches:

* translate Windows host paths into their `/mnt/<drive>/...` equivalents;
* create symlinks or bind mounts inside the guest to honor `mountPoint`;
* generate appropriate `/etc/wsl.conf` entries during initialization.

### Phase 3: Reduce backend-specific special casing

Where feasible, allow WSL2 instances to consume a broader set of Lima templates without requiring a dedicated experimental template.

## Open questions

1. Should WSL2 reject unsupported image formats at validation time?
2. What level of `mounts:` compatibility is desirable for WSL2?
3. Is full template portability across drivers a goal for the backend?

I'd be interested in working on this if the direction aligns with the maintainers' vision for WSL2.

afbjorklund:
> This makes WSL2 behave differently from other drivers and limits template portability.

It is, however, similar to the other container drivers - so it is portable within that "family"

* https://github.com/lima-vm/lima/discussions/3829

Supporting squashfs in addition to .tar would be a nice feature (I'm not so keen on OCI)

afbjorklund:
> Document how users can build or provide WSL-compatible rootfs archives.

This is something that is needed, currently it is using Dockerfile to build images.

```yaml
# Source: https://github.com/runfinch/finch-core/blob/main/rootfs/Dockerfile (image based on fedora)
```

[...]

Issue #3991, opened 2025-09-07
Change to Ubuntu for the lima container image?


afbjorklund:
### Description

Currently we are running Fedora as the base for the WSL2 image, but Ubuntu for the VM image.

I know that Podman and Podman Machine are using Fedora because of a lack of `podman` packages.

But for the default, it would be nice if it (container) was more similar to `template://default` (image)

Previously AC was struggling a bit with the systemd support, but that has now been fixed (0.4.0).

We also should consider making it part of the project, similar to https://github.com/lima-vm/alpine-lima

----

https://github.com/lima-vm/lima/blob/master/templates/experimental/wsl2.yaml

https://github.com/runfinch/finch-core/blob/main/rootfs/Dockerfile

My AC testing image was based on Debian, but that was mostly for OpenRC support.

So the suggestion is to go with the same OS, that is: `docker.io/library/ubuntu:rolling`

There is a bunch of other unneeded stuff, in the current image:

* `dnf update`
* `cosign` package
* old `qemu-user-static*`
* disabled network setup?

afbjorklund:
The new "project" is supposed to use a Dockerfile, to provide a rootfs tarball image for download.

Hopefully we can use the same rootfs for all host operating systems, and mostly use the upstream?

```Dockerfile
RUN apt-get update && apt-get install -y --no-install-recommends openssh-server && rm -rf /var/lib/apt/lists/* 
RUN apt-get update && apt-get install -y --no-install-recommends sudo && rm -rf /var/lib/apt/lists/* 

# lima
RUN apt-get update && apt-get install -y --no-install-recommends iptables && rm -rf /var/lib/apt/lists/* 
RUN update-alternatives --set iptables /usr/sbin/iptables-legacy
RUN apt-get update && apt-get install -y --no-install-recommends sshfs && rm -rf /var/lib/apt/lists/* 
RUN echo "user_allow_other" >>/etc/fuse.conf
```

* docker build
* docker create
* docker export
* docker rm

Might want some other tool (than docker), since it leaves some container cruft in the exported image.

` gtar --delete .dockerenv --delete dev --delete proc --delete sys < rootfs.tar > debian-rootfs-arm64.tar`

AkihiroSuda:
The `podman` template was originally using Ubuntu:
- https://github.com/lima-vm/lima/pull/210

We switched to Fedora in:
- https://github.com/lima-vm/lima/pull/876

[...]

Issue #3788, opened 2025-08-03
Lima driver for Apple Container virtual machines


afbjorklund:
### Description

Currently Lima is using the Virtualization.framework (VZ) to create a VM from a regular cloud image...

But it is also possible to use the Containerization Swift library, to create a VM from a container image

This makes it more similar to the WSL2 driver for Windows, with kernel and init provided outside image*.

The easiest way to integrate from a Go program is to use the `container` CLI, [github.com/apple/container](https://github.com/apple/container)

----

The main feature is that the MicroVM boots up in a second (with the right image) - instead of in a minute.

The default kernel is currently from the Kata containers project, it could be made available for others too.

Current issues

* Systemd is not yet supported (not compatible with `vminitd`), only OpenRC

* Some advanced features are missing from default kernel, such as IPv6 or VXLAN

* Apple ARM CPU is required, it does not work on Intel CPU

* Networking between instances does not work on macOS 15

\* for WSL2, the kernel is shared between all system containers. in AC, each one get their own kernel.

Lima could use a custom image for AC, similar to the custom image that is currently used with WSL2.

----

* #3622

* https://github.com/apple/container/discussions/106

afbjorklund:
Currently the wsl2 example (finch) uses a `fedora` image:

https://github.com/runfinch/finch-core/blob/main/rootfs/Dockerfile

For ac, the easiest is using a `debian` image (with openrc):

https://gist.github.com/afbjorklund/99ec0683c82f03e58c22b0d2753b9f50

Note: the `@cloud-server-environment` group install is 286 rpm packages!

The most important packages are `openssh-server` and `cloud-init` (if used)

The default cloud-init implementation also brings a python installation.

Later versions of Fedora also throws in a whole podman installation too...

----

We could also use Alpine, to make it more similar to alpine-lima cloud image

https://github.com/lima-vm/alpine-lima

Using standard Ubuntu (or Fedora) would have to wait for systemd support

* https://github.com/apple/container/issues/92

afbjorklund:
Currently Lima only supports tarball images, it could be extended with container images...

> WSL2 requires a tar formatted rootfs archive instead of a VM image

We can create a tarball with `docker export` and we can create a image with `docker import`.

* https://github.com/apple/container/issues/426

----

https://learn.microsoft.com/en-us/windows/wsl/use-custom-distro

Normally we would not use a docker daemon, but some OCI library

Supporting both registry and archive is doable as well, of course.

[...]

PR #5279, opened 2026-07-18
feat: unify container-based drivers for WSL2, AC, and DC

Nachiket-Roy:
## What This PR Changes

This PR implements Phase 3 of #5152 by unifying the WSL2, Apple Container (AC),
and Docker Container (DC) drivers under a shared container-driver framework.

The work builds on the AC/DC drivers originally proposed in closed PR #3840
and consolidates common validation, configuration defaults, and provisioning
logic into reusable helpers, removing duplicated code across the three drivers.

### 1. Shared Container-Driver Validation
- Introduce `IsContainerDriver()` for identifying container-based drivers.
- Add `driverutil.ValidateContainerDriverConfig()` to centralize validation:
  Linux guest OS + native arch requirements, tarball rootfs validation, mount
  type defaults/validation, unsupported feature checks (e.g. TPM), warnings
  for unknown fields.
- Apply `user.passwordlessSudo: false` validation consistently across all
  container drivers (previously WSL2-only).

### 2. Shared Provisioning
- Move `02-no-cloud-init-setup.sh` into the shared Linux boot templates.
- Remove duplicated provisioning scripts from WSL2/AC/DC.
- Refactor all three drivers to use the shared provisioning path.
- Generate SSH host keys and reset systemd's crash-loop counter for `sshd`
  in the shared script, since NoCloudInit guests never go through cloud-init's
  normal host-key generation.

### 3. Driver Interface & VM Updates
- Update AC/DC `Configure()` to the current `driver.Driver` interface.
- Fix `DownloadFile()` calls (missing `supportedImageFormats` argument).
- Replace context-cancellation spin loops during VM teardown with a single
  cleanup path (previous code busy-looped on a closed channel).
- Fix `ga.sock` bind conflict between the external driver server and the
  hostagent by having `GuestAgentConn` report `"unix"` explicitly.

### 4. DC-Specific Fixes
- Resolve guest SSH address via Docker bridge network inspection instead of
  loopback (loopback doesn't route into the container's network namespace).
- Work around `/etc/hosts` rename failure on Docker's bind-mounted file by
  using `cat > /etc/hosts` instead of `sed -i`.
- **Needs reviewer attention:** to unblock rootlesskit under
  `kernel.apparmor_restrict_unprivileged_userns`, the boot script now mounts
  `securityfs` and loads a custom AppArmor profile into the *host* kernel from
  inside the container. This requires the outer container to run with
  elevated privileges by default. I want to flag this explicitly since it's a
  meaningfully different trust/permission model than the other drivers and
  deserves discussion before merging — see "Known issues" below for the
  overlay-on-overlay problem this is solving.
- Nested overlayfs (Docker's own storage driver + rootless containerd/stargz
  inside the guest) isn't supported by the kernel, so containerd/buildkit/
  stargz working directories are moved to tmpfs.

### 5. Build, Templates, and CI
- Enable AC by default on Darwin builds, DC on Linux builds.
- Update uninstall handling for external container drivers.
- Update template validation to recognize AC/DC.
- Pin experimental AC/DC templates to Ubuntu 24.04 rootfs images.
- Add CI coverage for the DC template.
- **Test-suite changes I want called out explicitly, not just diffed:**
  - `test-templates.sh` now accepts "degraded" systemd status as passing for
    container drivers (some services, e.g. TPM/root-mount units, can't run
    inside a container and are expected to fail there).
  - `dc` was removed from the list of VM types that force `sudo` for
    container-engine commands, since DC runs rootless by design.
  - `test-port-forwarding.pl` now resolves the guest's route IP dynamically
    and skips guest-IP-specific/IPv6 checks for external/container drivers,
    since those addresses aren't meaningful for DC's network setup.

## Linked Issue
Relates to #5152 

## How I Tested This
- `go test ./pkg/limayaml/... ./pkg/driverutil/... ./pkg/cidata/...`
- Added unit tests for `ValidateContainerDriverConfig()`.
- `make native ADDITIONAL_DRIVERS=dc`
- `./_output/bin/limactl validate templates/experimental/ac.yaml`
- `./_output/bin/limactl validate templates/experimental/dc.yaml`
- Manual `dc` instance boot/SSH/nerdctl smoke test on Ubuntu 24.04 host

## AI Usage
Assisted-by: Gemini 3.5

### DIFF
--- a/pkg/driver/ac/vm_darwin.go
+++ b/pkg/driver/ac/vm_darwin.go
@@ -1,0 +1,285 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package ac
+
+import (
+	"bytes"
+	"context"
+	_ "embed"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
+	"github.com/lima-vm/lima/v2/pkg/textutil"
+)
+
+// createVM calls AC to create a VM.
+func createVM(ctx context.Context, distroName string, cpus, memory int, initSystem string) error {
+	imageName := distroName
+	entrypoint := "/sbin/init"
+	// /sbin/init is normally just a symlink to systemd
+	// eventually we might want to look inside the image
+	switch initSystem {
+	case "systemd":
+		entrypoint = "/lib/systemd/systemd"
+	case "openrc":
+		entrypoint = "/sbin/openrc-init"
+	default:
+		logrus.Infof("unknown init system, running only vminitd")
+	}
+	out, err := exec.CommandContext(ctx,
+		"container",
+		"create",
+		"--name",
+		distroName,
+		"--cpus",
+		fmt.Sprintf("%d", cpus),
+		"--memory",
+		fmt.Sprintf("%dM", memory),
+		"--entrypoint",
+		entrypoint,
+		imageName).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to run `container create %s`: %w (out=%q)",
+			distroName, err, out)
+	}
+	return nil
+}
+
+// startVM calls AC to start a VM.
+func startVM(ctx context.Context, distroName string) error {
+	out, err := exec.CommandContext(ctx,
+		"container",
+		"start",
+		distroName).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to run `container start %s`: %w (out=%q)",
+			distroName, err, out)
+	}
+	return nil
+}
+
+// initVM calls AC to import a new VM specifically for Lima.
+func initVM(ctx context.Context, instanceDir, distroName string) error {
+	imageName := distroName
+	baseDisk := filepath.Join(instanceDir, filenames.BaseDiskLegacy)
+	logrus.Infof("Importing distro from %q to %q", baseDisk, imageName)
+	out, err := exec.CommandContext(ctx,
+		"container",
+		"image",
+		"import",
+		baseDisk,
+		imageName).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to run `container image import %s %s`: %w (out=%q)",
+			baseDisk, imageName, err, out)
+	}
+	return nil
+}
+
+// stopVM calls AC to stop a running VM.
+func stopVM(ctx context.Context, distroName string) error {
+	out, err := exec.CommandContext(ctx,
+		"container",
+		"stop",
+		distroName).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to run `container stop %s`: %w (out=%q)",
+			distroName, err, out)
+	}
+	return nil
+}
+
+//go:embed lima-init.TEMPLATE
+var limaBoot string
+
+// copyDir copies a directory.
+func copyDir(ctx context.Context, distroName, src, dst string) error {
+	cmd := exec.CommandContext(ctx, "container", "exec", distroName, "mkdir", "-p", dst)
+	if err := cmd.Run(); err != nil {
+		if exiterr, ok := err.(*exec.ExitError); ok {
+			logrus.Debugf("run stderr: %s", exiterr.Stderr)
+		}
+		return fmt.Errorf("failed to run %v: %w", cmd.Args, err)
+	}
+	logrus.Infof("Copying directory from %q to \"%s:%s\"", src, distroName, dst)
+	tar1 := exec.CommandContext(ctx, "tar", "Cc", src, ".")
+	tar2 := exec.CommandContext(ctx, "container", "exec", "-i", distroName, "tar", "Cx", dst)
+
+	p, err := tar1.StdoutPipe()
+	if err != nil {
+		return err
+	}
+	tar2.Stdin = p
+	if err := tar2.Start(); err != nil {
+		return err
+	}
+	if err := tar1.Run(); err != nil {
+		return err
+	}
+	if err := tar2.Wait(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// provisionVM starts Lima's boot process inside an already imported VM.
+func provisionVM(ctx context.Context, instanceDir, instanceName, distroName string, errCh chan<- error) error {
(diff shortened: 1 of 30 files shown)


AkihiroSuda:
Could this be experimented in a third-party repository?

afbjorklund:
I think it might be a good idea to try and see what is similar between the drivers, but **not** so happy about losing all the context and authorship of the old code - there is not even a commit message, and no co-authored-by - nothing.

There seems to be some debugging and test code left in there, like not using Kata containers, will comment on those. I don't think that it will be ready for Lima 2.3.0, and needs to be broken up into patches before it can be reviewed...

pkg/driver/dc/vm_linux.go
```diff
@@ -0,0 +1,304 @@
...
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/lima-vm/lima/v2/pkg/limatype"
+	"github.com/lima-vm/lima/v2/pkg/limatype/filenames"
+	"github.com/lima-vm/lima/v2/pkg/textutil"
+)
+
+// OCI runtime for container.
+const ociRuntime = "runc"
```
afbjorklund: This should be switched back to Kata containers by default, the "runc" was more of an experiment

We are not so interested in running Docker in Docker, it is more useful to have a VM model (like AC)

pkg/driver/ac/vm_darwin.go
```diff
@@ -0,0 +1,300 @@
...
+		return err
+	}
+	err = os.Chdir(instanceDir)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = os.Chdir(wd) }()
+	out, err := exec.CommandContext(ctx,
+		"container",
+		"build",
```
afbjorklund: ~~There is now proper support for tarball import in `container`, so we don't have to use build anymore.~~

templates/experimental/ac.yaml
```diff
@@ -0,0 +1,23 @@
+vmType: ac
+
+arch: aarch64
+cpus: 4
+memory: 1GiB
+disk: 512GiB
+
+images:
+- location: "https://cloud-images.ubuntu.com/releases/noble/release-20260615/ubuntu-24.04-server-cloudimg-arm64-root.tar.xz"
```
afbjorklund: This should use a container image as the base, and not the cloud image (with kernel, and so on)

templates/experimental/dc.yaml
```diff
@@ -0,0 +1,24 @@
+vmType: dc
+
+images:
```
afbjorklund, line 3: Same here, we want to use container images (as rootfs) for the container drivers - not cloud images

templates/experimental/ac.yaml
```diff
@@ -0,0 +1,23 @@
...
+  digest: "sha256:15188696da114a3ffd3d3554f5858a0c3ac257933656e85feb4e0e83ad542b4a"
+
+mountType: virtiofs
+
+mounts:
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+
+# The built-in containerd installer does not support OpenRC currently.
```
afbjorklund: This comment is obsolete, there is now support for systemd in Apple Container images

templates/experimental/dc.yaml
```diff
@@ -0,0 +1,24 @@
...
+- location: "~"
+- location: "/tmp/lima"
+  writable: true
+
+provision:
+- mode: system
+  script: |
+    #!/bin/bash
+    apt-get update
+    apt-get install -y socat
```
afbjorklund: This needs a comment, and preferrably it should not be needed (or integrated in Lima)

afbjorklund:
By mistake I pushed "runc" as the default in the latest DC draft, but it should be [Kata containers](https://katacontainers.io/).

* https://github.com/lima-vm/lima/discussions/3829

i.e. runtime should default to "io.containerd.kata.v2" (which is the same as the old name "runv")

The "runc" and "runq" was alternatives, while developing.

afbjorklund:
> Could this be experimented in a third-party repository?

We also need a proper issue, for the scope of the drivers.

Nachiket-Roy:
Thanks! I've gone through the comments and addressed most of them.

I've added `Co-authored-by: Anders F Björklund` to the AC and DC commits.

Done:

* Removed the `socat` install from `dc.yaml` and updated the port-forwarding test to skip the socat-based check for container/external drivers.
* Switched AC to import the rootfs tarball directly instead of generating a Dockerfile.
* Removed the outdated OpenRC comment from `ac.yaml`.

There are two things I'd like your input on before going further.

**DC's default runtime.** I looked into switching to Kata as suggested, but ran into an issue where running Docker with `--privileged` (needed for systemd) causes host devices to be passed through, which conflicts with Kata's device initialization and prevents the container from starting. Containerd has a CRI-level option (`privileged_without_host_devices`) to avoid this, but Docker doesn't use that path. Rather than defaulting back to `runc`, I think the better direction is to avoid `--privileged` entirely and boot systemd with only the capabilities and mounts it actually needs (using `--cgroupns=host`, mounting `/sys/fs/cgroup`, and granting only the required capabilities). Does that sound like the right approach? Also, does AC already have a capability/cgroup setup that DC should mirror, rather than me re-deriving it?

**Base images.** I agree that using proper OCI images is the right direction, but the official minimal images don't include systemd or `sshd`, so supporting them cleanly needs some boot-script work. Since that's closely tied to the runtime discussion above, I'd rather resolve that first than merge temporary workarounds.

On the other two points: I'm happy to move the AC/DC experimentation into a separate repository and will set one up. I'll also open a dedicated issue to track the driver work.

pkg/driver/ac/vm_darwin.go
afbjorklund: Sorry the PR was not merged yet, there is still no "import" in the released version of container 1.1.0

afbjorklund: https://github.com/apple/container/commit/cad5c27df3cff7c4774f4b045d07e69a4291b9b8

afbjorklund:
> **Base images.** I agree that using proper OCI images is the right direction, but the official minimal images don't include systemd or `sshd`, so supporting them cleanly needs some boot-script work. S

This is a problem shared with Apple Container Machine, it also needs a derived image with a `/sbin/init` in it.

https://github.com/apple/container/blob/main/docs/container-machine.md#bring-your-own-container-machine-image

That is why it would be a good with a common approach, that could also include the WSL image (currently: Finch).

* #3991

> On the other two points: I'm happy to move the AC/DC experimentation into a separate repository and will set one up. I'll also open a dedicated issue to track the driver work.

Your fork already _is_ a separate repository, and I am not sure we decided to make these drivers external yet?

https://github.com/Nachiket-Roy/lima/tree/feature/unify-container-drivers

They might end up having to have different names, like "apple" and "kata", names have been a problem before.

* #3788

PR #5280, opened 2026-07-18
cidata: write password.txt without a UTF-8 BOM

### LINKED ISSUE #5277: Windows guest: `~/password.txt` shouldn't have UTF-8 BOM (`EF BB BF`)
https://github.com/lima-vm/lima/blob/3821dd4742e675a1483acb84a647cd567be468c9/pkg/cidata/wincidata.TEMPLATE.d/first_logon.ps1#L10-L17

This seems to prepend UTF-8 BOM (`EF BB BF`) to `password.txt`

SAY-5:
## What This PR Changes

On Windows guests, `first_logon.ps1` writes the generated login password to `~/password.txt` with `Out-File -Encoding utf8`. Under Windows PowerShell 5.1 (the engine that runs during first logon) that encoding prepends a UTF-8 byte-order mark (`EF BB BF`) to the file, so the stored password is preceded by three stray bytes. Since the generated password only uses ASCII characters (`a-zA-Z0-9`), switching to `-Encoding ascii` writes the value with no BOM. This also matches the neighboring `boot.done` write on line 83, which already uses `-Encoding ascii`.

Closes #5277

## How I Tested This

The change is confined to the Windows-guest cidata template, which is only exercised by booting a Windows guest, so I could not run it in this environment. I confirmed by inspection that Windows PowerShell's `Out-File -Encoding utf8` emits a BOM while `-Encoding ascii` does not, and that the password charset on line 13 is ASCII-only, so `ascii` encodes it losslessly.

### DIFF
--- a/pkg/cidata/wincidata.TEMPLATE.d/first_logon.ps1
+++ b/pkg/cidata/wincidata.TEMPLATE.d/first_logon.ps1
@@ -14,7 +14,9 @@ $chars = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
 $newPassword = -join ((1..16) | ForEach-Object { $chars[(Get-Random -Maximum $chars.Length)] })
 
 # Store the password under the user directory so that user can know/change it.
-$newPassword | Out-File -FilePath "C:\Users\{{.User}}\password.txt" -Encoding utf8 -NoNewline
+# Use ASCII encoding to avoid the UTF-8 BOM that Windows PowerShell's `Out-File -Encoding utf8`
+# prepends; the generated password only contains ASCII characters.
+$newPassword | Out-File -FilePath "C:\Users\{{.User}}\password.txt" -Encoding ascii -NoNewline
 
 # Change the password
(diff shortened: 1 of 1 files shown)


pkg/cidata/wincidata.TEMPLATE.d/first_logon.ps1
```diff
@@ -14,7 +14,9 @@ $chars = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789'
 $newPassword = -join ((1..16) | ForEach-Object { $chars[(Get-Random -Maximum $chars.Length)] })
 
 # Store the password under the user directory so that user can know/change it.
-$newPassword | Out-File -FilePath "C:\Users\{{.User}}\password.txt" -Encoding utf8 -NoNewline
+# Use ASCII encoding to avoid the UTF-8 BOM that Windows PowerShell's `Out-File -Encoding utf8`
```
AkihiroSuda, line 17: This comment line does not seem needed

AkihiroSuda (approved):

Issue #5283, opened 2026-07-18
`pkg/ioutilx` contains functions that do not resemble `io/ioutil`


AkihiroSuda:
https://github.com/lima-vm/lima/blob/master/pkg/ioutilx/ioutilx.go

- `ReadAtMaximum`: should stay in `ioutilx`
- `FromUTF16le` and similar functions: should be moved to something like `pkg/strutil`? `pkg/unicodeutil?` Or should just stay in `ioutilx`?
- `WindowsSubsystemPath` and similar functions: should be moved to `pkg/fsutil`. The file should also have the `_windows.go` suffix to avoid compilation on non-Windows https://github.com/lima-vm/lima/pull/5284

PR #5284, opened 2026-07-18
pkg/ioutilx: move Windows path helpers to pkg/fsutil

arimu1:
Part of #5283.

Moves the Windows path-conversion helpers out of `pkg/ioutilx` into `pkg/fsutil`, per the issue's plan:

- `WindowsSubsystemPath`, `WindowsSubsystemPathWithCygpath`, `WindowsSubsystemPathForLinux` (+ the unexported cygpath-less fallback) → `pkg/fsutil/fsutil_windows.go`, compiled only on Windows via the filename suffix.
- `ReadAtMaximum` stays in `ioutilx` (per the issue), and the UTF16 helpers are deliberately **not** moved — their eventual home (`strutil`? `unicodeutil`?) is still an open question in the issue, so this PR doesn't preempt that call.

Because all six caller files are cross-platform and already gate these calls behind `runtime.GOOS == "windows"`, the symbols still need to resolve on non-Windows: `pkg/fsutil/fsutil_unix.go` (`//go:build !windows`) provides stub definitions returning an error, unreachable in practice — mirroring the existing `errNotSupported` pattern in `pkg/guestagent/timesync/timesync_others.go`. This keeps every caller's logic unchanged (only the import + `ioutilx.` → `fsutil.` prefix change) and no GOOS build breaks.

The Windows-only unit test moves alongside (`pkg/fsutil/fsutil_windows_test.go`, 98% rename). No forwarding shims were left in `ioutilx`, matching how prior moves in this area (7a5a207, 54db504) rewrote call sites directly.

## Verification

- `go build ./...` on darwin, `GOOS=linux`, and `GOOS=windows` — all clean
- `GOOS=windows go vet ./pkg/fsutil/...` + `go test -c ./pkg/fsutil/...` to compile-check the Windows-only files and test
- `go test` on all touched caller packages (copytool, hostagent, sshutil, limayaml, cmd/limactl) — ok
- `go vet ./...` clean; `golangci-lint run` on touched packages — 0 issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -25,8 +25,8 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/autostart"
 	"github.com/lima-vm/lima/v2/pkg/copytool"
 	"github.com/lima-vm/lima/v2/pkg/envutil"
+	"github.com/lima-vm/lima/v2/pkg/fsutil"
 	"github.com/lima-vm/lima/v2/pkg/instance"
-	"github.com/lima-vm/lima/v2/pkg/ioutilx"
 	"github.com/lima-vm/lima/v2/pkg/limatype"
 	"github.com/lima-vm/lima/v2/pkg/limayaml"
 	"github.com/lima-vm/lima/v2/pkg/localpathutil"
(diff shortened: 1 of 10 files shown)


AkihiroSuda (approved):
Thanks

PR #5288, opened 2026-07-19
docs(agents): add AGENTS.md and CLAUDE.md for AI coding agents

### LINKED ISSUE #5267: Create `AGENTS.md` to guide AI agent behaviour
> You might also consider an agents.md at the root of the project even if only to guide how you do / don't want it to behave (and as you suggest economize on token cost).  For context, I've observed this being used to good effect:
> 
> https://github.com/NVIDIA/aicr/blob/main/AGENTS.md
> 
> _Originally posted by @resker in https://github.com/lima-vm/lima/issues/4989#issuecomment-4465291189_
>

ekalinin:
## What This PR Changes

Adds an `AGENTS.md` at the repository root to guide AI coding agents (Claude Code, Codex, Cursor, and others), and to economize on token cost - agents read one curated, committed file instead of re-deriving the project from scratch. This
follows the [NVIDIA aicr example](https://github.com/NVIDIA/aicr/blob/main/AGENTS.md) referenced in the issue.
Adds `CLAUDE.md` - a thin pointer that tells Claude Code to always read `AGENTS.md` first, so tool-specific and shared guidance stay in one place.

## Linked Issue

Closes #5267

## How I Tested This

Documentation-only change. `make editorconfig-checker` passes; `AGENTS.md` and `CLAUDE.md` are Markdown.

## AI Usage

Assisted-by: Claude:Opus-4.8 [Claude Code]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ lima.REJECTED.yaml
 default-template.yaml
 schema-limayaml.json
 .config
+*.local.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,48 @@
+# AGENTS.md
+
+Guidance for AI coding agents working in the Lima repository. It points to the authoritative docs
+instead of duplicating them, so it stays in sync with the code.
+
+## AI contribution rules
+
+Read @website/content/en/docs/community/contributing.md and follow its "AI Contribution Rules"
+section.
+
+## Build, test, lint
+
+Lima is `github.com/lima-vm/lima/v2` (imports use the `/v2` suffix). The build
+uses a GNU Makefile; output goes to `_output/`.
+
+```bash
+make native        # build limactl + native guestagent + templates (fastest full dev build)
+make minimal       # build just limactl + native guestagent + default template
+go test ./...      # unit tests - never boot VMs
+make bats          # integration tests (BATS); boot real VMs (needs git submodules)
+make lint          # editorconfig, golangci-lint, yamllint, ls-lint, shellcheck, ltag, ...
+make generate      # regenerate protobuf after editing a .proto file
+```
+
+Unit tests never execute VMs; anything that boots a VM is a BATS or template test under `hack/`
+(for example `./hack/test-templates.sh ./templates/default.yaml`). Every commit must be signed off
+with `git commit -s`, or CI fails.
+
+## Where things are
+
+Pointers to the authoritative sources - read these rather than a duplicated copy here.
+
+- Architecture, the three processes (`limactl` / hostagent / guestagent), the on-disk `${LIMA_HOME}`
+  layout, and every `LIMA_CIDATA_*` variable:
+  [`website/content/en/docs/dev/internals.md`](website/content/en/docs/dev/internals.md).
+- Config / data model: `pkg/limatype` (core `LimaYAML` / `Instance` types), `pkg/limayaml`
+  (load / default / validate), `pkg/limatmpl` and `pkg/templatestore` (templates).
+- Drivers (virtualization backends): `pkg/driver/`
+- Guest provisioning: `pkg/cidata/` builds `cidata.iso` (guestagent binary, boot scripts, and the
+  `user-data` file). `user-data` uses the cloud-config YAML format defined by cloud-init, which a
+  guest may consume with an implementation other than Python cloud-init.
+- Instance lifecycle: `pkg/instance/`.
+
+## Conventions
+
+- New Go, shell, Dockerfile, and Makefile files need an SPDX header (`ltag` enforces this in CI;
+  other file types, including markdown, are exempt).
+- Keep the `gomodjail` / `gosocialcheck` annotations in `go.mod`.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,6 @@
+# CLAUDE.md
+
+@AGENTS.md
+
+`AGENTS.md` is the shared guidance for AI coding agents in this repository; the `@` line above makes
+the harness load it. Personal, machine-local overrides may live in `CLAUDE.local.md` (gitignored).


AGENTS.md
```diff
@@ -0,0 +1,147 @@
...
+
+- `pkg/limatype/` — the core `LimaYAML` / `Instance` data structures shared everywhere. Start here
+  to understand what a template/instance contains.
+- `pkg/limayaml/` — loading, defaulting, and validating `lima.yaml`.
+- `pkg/limatmpl/`, `pkg/templatestore/` — template (`templates/*.yaml`) resolution, including the
+  `template:` scheme.
+
+### Guest provisioning
+
+VMs are configured via cloud-init. `limactl` builds `cidata.iso` (see `pkg/cidata/`) containing the
```
afbjorklund: It is not 100% required that guests actually use cloud-init, so Lima uses the "cloud-config" yaml that is defined by cloud-init - but it might be picked up by a different implementation than python cloud-init.

Only need to mention `user-data` file here, in addition to the others. And that it uses this format.

afbjorklund:
Keeping AGENTS.md up-to-date is not going to be "fun", since it duplicates a lot of information, but inevitable...

I invited another robot to the party.

AGENTS.md
```diff
@@ -0,0 +1,147 @@
+# AGENTS.md
+
+This file guides AI coding agents (Claude Code, Codex, Cursor, and others) working in the Lima
```
afbjorklund: I am not sure the brands need to go in here.

AGENTS.md
AkihiroSuda, line 1: Do we really need this amount of AGENTS.md?
I fear it can be easily out of sync with the actual code.

Did you measure how this file contributes to reduction of LLM token costs and tool calling overhead?

CLAUDE.md
AkihiroSuda, line 1: Why do we need agent-specific prompt files in addition to the generic AGENTS.md?

AGENTS.md
unsuman, line 1: Shouldn't this be placed inside `docs/agents/`?

AkihiroSuda: Why?

unsuman: Nothing specific, I thought it would make the repo root a little tidier 😅 
But the AI is good at finding this file when at root! So it should be at root

AGENTS.md
ekalinin: Agreed, removed.

AGENTS.md
ekalinin: Good points, and I agree with both.

On the token/tool-call measurement: no, I didn't measure it, so I've dropped that claim - it wasn't something I could back up. The real motivation is narrower: give agents (and their humans) one curated entry point to the contribution rules, so PRs follow the DCO/Assisted-by/talk-first policy instead of each tool guessing.

On size and drift: also fair. I've trimmed the file and removed the "Architecture" section, which was the main drift risk since it duplicated `dev/internals.md`. What's left is:

  - AI contribution rules, summarized with a link to `contributing.md#ai-contribution-rules` as the source of truth.
  - Build/test/lint commands.
  - A "Where things are" section that only points to internals.md and the key packages rather than copying them.

CLAUDE.md
ekalinin: Strictly, we don't need them committed. The catch is that Claude Code (and some other tools) don't read `AGENTS.md` by default - Claude Code loads `CLAUDE.md`, Codex/others load `AGENTS.md`, Cursor `.cursorrules`, Copilot `.github/copilot-instructions.md`. That's the only reason a `CLAUDE.md` was there.

That said, a committed per-tool file isn't strictly necessary - a Claude Code user can keep a local, gitignored `CLAUDE.md` with a single `@AGENTS.md` import, so the content still comes from the one generic file. If you'd rather not have `CLAUDE.md` in the tree, I'm happy to remove it (and the `.gitignore` line) and keep the PR to just `AGENTS.md`.

afbjorklund: Sounds like Dockerfile vs Containerfile :-)  (but I'm OK with having a CLAUDE.md as an exception)

AGENTS.md
ekalinin: Reworded to say `pkg/cidata/ builds` the `cidata.iso` with a `user-data` file in the `cloud-config` YAML format defined by `cloud-init`, and that a guest may consume it with an implementation other than Python `cloud-init`, rather than implying `cloud-init` is required.
Thanks!

CLAUDE.md
jandubois: Why isn't CLAUDE.md just a symlink to AGENTS.md?

afbjorklund: Probably Windows? It is always Windows.

Then again having "@AGENTS.md" in a textfile, or "AGENTS.md" in a symlink file sounds rather similar...

AGENTS.md
```diff
@@ -0,0 +1,65 @@
...
+(for example `./hack/test-templates.sh ./templates/default.yaml`). Every commit must be signed off
+with `git commit -s`, or CI fails.
+
+## Where things are
+
+Pointers to the authoritative sources - read these rather than a duplicated copy here.
+
+- Architecture, the three processes (`limactl` / hostagent / guestagent), the on-disk `${LIMA_HOME}`
+  layout, and every `LIMA_CIDATA_*` variable:
+  [`dev/internals.md`](website/content/en/docs/dev/internals.md).
```
AkihiroSuda:
```suggestion
  [`website/content/en/docs/dev/internals.md`](website/content/en/docs/dev/internals.md).
```

AGENTS.md
```diff
@@ -0,0 +1,65 @@
...
+- Drivers (virtualization backends): `pkg/driver/` (`qemu`, `vz`, `krunkit`, `wsl2`, and `external`
+  for the gRPC bridge).
+- Guest provisioning: `pkg/cidata/` builds `cidata.iso` (guestagent binary, boot scripts, and the
+  `user-data` file). `user-data` uses the cloud-config YAML format defined by cloud-init, which a
+  guest may consume with an implementation other than Python cloud-init.
+- Instance lifecycle: `pkg/instance/`.
+
+## Conventions
+
+- Files are platform-split with build constraints and `_darwin.go` / `_linux.go` / `_windows.go` /
```
AkihiroSuda: AI is expected to be already aware of this convention, no?

AGENTS.md
```diff
@@ -0,0 +1,65 @@
...
+  `_others.go` / `_unix.go` suffixes; check all variants when touching cross-platform behavior.
+- Source files carry SPDX headers (`ltag` enforces this; markdown is exempt) - copy one from an
```
AkihiroSuda: AI should be aware of this?

AGENTS.md
```diff
@@ -0,0 +1,65 @@
...
+matters most to agents and points to the authoritative docs instead of copying them, so it stays in
+sync with the code.
+
+## AI contribution rules
+
+A summary of the "AI Contribution Rules" section of
+[`contributing.md`](website/content/en/docs/community/contributing.md#ai-contribution-rules), which
+is the source of truth.
+
+- Talk first, code later - open an issue and get maintainer approval before a pull request.
```
AkihiroSuda: This instruction is for human, not for AI

AGENTS.md
```diff
@@ -0,0 +1,65 @@
...
+
+Guidance for AI coding agents working in the Lima repository. The full contribution policy lives in
+[`contributing.md`](website/content/en/docs/community/contributing.md); this file highlights what
+matters most to agents and points to the authoritative docs instead of copying them, so it stays in
+sync with the code.
+
+## AI contribution rules
+
+A summary of the "AI Contribution Rules" section of
+[`contributing.md`](website/content/en/docs/community/contributing.md#ai-contribution-rules), which
```
AkihiroSuda:
```suggestion
[`website/content/en/docs/community/contributing.md`](website/content/en/docs/community/contributing.md#ai-contribution-rules), which
```

AGENTS.md
ekalinin: Agreed. Removed it.

AGENTS.md
ekalinin: Unlike the file-suffix convention, this one is repo-specific: an agent knows what an SPDX header is, but not that Lima requires one on every Go/shell/Dockerfile/Makefile file and enforces it via ltag in CI.

Without the note a new file lands without a header and fails CI. I trimmed it to just that fact (and which file types are exempt) and dropped the generic "copy from an existing file" advice. Happy to remove it entirely if you'd prefer.

AGENTS.md
ekalinin: Right, the suffix naming itself is generic Go knowledge, so I dropped that explanation. I kept only the actionable part - Lima has a lot of these splits, and the common miss is editing one variant (say
  _darwin.go) and forgetting the others. If you think even that is unnecessary, I'll remove the bullet entirely.

CLAUDE.md
```diff
@@ -0,0 +1,12 @@
+# CLAUDE.md
+
+Always read **[AGENTS.md](AGENTS.md)** first.
```
jandubois: 1. This does not necessarily work. I did a test, and only CLAUDE.md was loaded.

2. If CLAUDE.md does not exist, the claude 2.1.220 on macOS was loading AGENTS.md.

3. If CLAUDE.md exists and contains an `@AGENTS.md` line, then AGENTS.md was also loaded.

I would suggest picking option (2). Option (3) may add backwards compatibility to older `claude` versions, but that doesn't seem important to me; everybody always updates in a timely manner.

AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+- One fix per pull request.
+- Humans are responsible for all content - a human reviews and edits any AI-generated code and pull
+  request text, and replies to review comments themselves.
+- DCO: agents MUST NOT add a `Signed-off-by` trailer. Only the human submitting the code signs off,
+  with `git commit -s`.
+- Disclose AI usage with an `Assisted-by: AI_TOOL_NAME` trailer in the pull request description.
+
+## Build, test, lint
+
+Lima is `github.com/lima-vm/lima/v2` (imports use the `/v2` suffix) and requires Go 1.25+. The build
```
jandubois: The Go version is trivially available via `go.mod` and agents know this. And this line is already wrong, as we require 1.25.7+.

```suggestion
Lima is `github.com/lima-vm/lima/v2` (imports use the `/v2` suffix). The build
```

AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+
+Pointers to the authoritative sources - read these rather than a duplicated copy here.
+
+- Architecture, the three processes (`limactl` / hostagent / guestagent), the on-disk `${LIMA_HOME}`
+  layout, and every `LIMA_CIDATA_*` variable:
+  [`website/content/en/docs/dev/internals.md`](website/content/en/docs/dev/internals.md).
+- Config / data model: `pkg/limatype` (core `LimaYAML` / `Instance` types), `pkg/limayaml`
+  (load / default / validate), `pkg/limatmpl` and `pkg/templatestore` (templates).
+- Drivers (virtualization backends): `pkg/driver/` (`qemu`, `vz`, `krunkit`, `wsl2`, and `external`
+  for the gRPC bridge).
```
jandubois: Driver list is available by listing the directory contents:

```suggestion
- Drivers (virtualization backends): `pkg/driver/`
```

AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+- Instance lifecycle: `pkg/instance/`.
+
+## Conventions
+
+- Cross-platform code is split across many `_darwin.go` / `_linux.go` / `_windows.go` / `_others.go`
+  / `_unix.go` files; when changing such behavior, update every variant, not just one.
+- New Go, shell, Dockerfile, and Makefile files need an SPDX header (`ltag` enforces this in CI;
+  other file types, including markdown, are exempt).
+- Keep the `gomodjail` / `gosocialcheck` annotations in `go.mod`.
+- Reuse the small `pkg/*util` helpers (`fsutil`, `osutil`, `sshutil`, ...) instead of reimplementing.
```
jandubois: Generic advice, can be deleted; agents already grep for helpers before writing them.

AGENTS.md
```diff
@@ -0,0 +1,62 @@
...
+A summary of the "AI Contribution Rules" section of
+[`website/content/en/docs/community/contributing.md`](website/content/en/docs/community/contributing.md#ai-contribution-rules), which
+is the source of truth.
+
+- One fix per pull request.
+- Humans are responsible for all content - a human reviews and edits any AI-generated code and pull
+  request text, and replies to review comments themselves.
+- DCO: agents MUST NOT add a `Signed-off-by` trailer. Only the human submitting the code signs off,
+  with `git commit -s`.
+- Disclose AI usage with an `Assisted-by: AI_TOOL_NAME` trailer in the pull request description.
```
jandubois: Duplicating the info is not useful. If the summary and the document go out of sync, you have conflicting advice.

Agents are pretty good at progressive disclosure now, so just having a link is enough. I think `@website/content/en/docs/community/contributing.md` is better because it forces the harness to read the file while the Markdown link is just content that gets inlined into the context.

CLAUDE.md
jandubois: I could not reproduce (2) on a fresh start; it looks like (3) is the only mechanism that works for me.

Sorry for the confusion!

I tested (1) with Opus 5, and I wonder if the reason it didn't work is that to Opus the line looks like prompt injection, so it just ignores it. Maybe it does work with older models, but still, it takes a tool call while the `@` form is free because it is performed by the harness.

AGENTS.md
ekalinin: Agreed. Dropped the summary. The section now just points at the policy with `@website/content/en/docs/community/contributing.md`.

AGENTS.md
ekalinin: Done.

CLAUDE.md
ekalinin: Thanks for testing this so thoroughly. Switched to option (3): CLAUDE.md now contains an @AGENTS.md import line.

AGENTS.md
```diff
@@ -0,0 +1,50 @@
...
+  (load / default / validate), `pkg/limatmpl` and `pkg/templatestore` (templates).
+- Drivers (virtualization backends): `pkg/driver/`
+- Guest provisioning: `pkg/cidata/` builds `cidata.iso` (guestagent binary, boot scripts, and the
+  `user-data` file). `user-data` uses the cloud-config YAML format defined by cloud-init, which a
+  guest may consume with an implementation other than Python cloud-init.
+- Instance lifecycle: `pkg/instance/`.
+
+## Conventions
+
+- Cross-platform code is split across many `_darwin.go` / `_linux.go` / `_windows.go` / `_others.go`
```
AkihiroSuda: I still think this is obvious and does not need to be instructed

.gitignore
```diff
@@ -4,3 +4,4 @@ lima.REJECTED.yaml
 default-template.yaml
 schema-limayaml.json
 .config
+CLAUDE.local.md
```
AkihiroSuda: nit:
```suggestion
*.local.md
```

AkihiroSuda (dismissed):
Thanks

AGENTS.md
ekalinin: Ok, deleted it.

AkihiroSuda (approved):
Thanks

AGENTS.md
```diff
@@ -0,0 +1,48 @@
...
+make minimal       # build just limactl + native guestagent + default template
+go test ./...      # unit tests - never boot VMs
+make bats          # integration tests (BATS); boot real VMs (needs git submodules)
+make lint          # editorconfig, golangci-lint, yamllint, ls-lint, shellcheck, ltag, ...
+make generate      # regenerate protobuf after editing a .proto file
+```
+
+Unit tests never execute VMs; anything that boots a VM is a BATS or template test under `hack/`
+(for example `./hack/test-templates.sh ./templates/default.yaml`). Every commit must be signed off
+with `git commit -s`, or CI fails.
```
afbjorklund, line 27: I'm not sure what the agent is supposed to do with this information, since it can't sign anything?

afbjorklund (approved):

PR #5293, opened 2026-07-20
block ssh ports on every guest IP, not just 0.0.0.0

Nashit-h:
The rules that block port 22 and sshLocalPort are built with GuestIP 0.0.0.0 and no explicit GuestIPMustBeZero, so FillPortForwardDefaults derives it as true and the rule ends up matching only a guest listener bound to 0.0.0.0. A guest that binds 127.0.0.1:22 or ::1:<sshLocalPort> falls through both deny rules and is then picked up by the default catch-all rule, which forwards it to the host, letting a compromised VM take over the host-local SSH endpoint that limactl shell talks to. The added test fails on master with `guest 127.0.0.1:22 -> 127.0.0.1:22` and passes here.

Setting GuestIPMustBeZero explicitly to false on the two deny rules keeps the exact-match semantics for user rules that ask for guestIP 0.0.0.0 while making these behave as the comment says. I pulled the rule construction out of New into portForwardRules so it could be tested; both forwarders read the same slice, so the one change covers the grpc and ssh paths.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -246,18 +246,7 @@ func New(ctx context.Context, instName string, stdout io.Writer, signalCh chan o
 			break
 		}
 	}
-	rules := make([]limatype.PortForward, 0, 3+len(inst.Config.PortForwards))
-	// Block ports 22 and sshLocalPort on all IPs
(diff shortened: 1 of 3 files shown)


AkihiroSuda (approved):
Thanks

PR #5298, opened 2026-07-21
sshutil: parse the Win32-OpenSSH version banner

jandubois:
The regex required a digit right after `OpenSSH_`, so `OpenSSH_for_Windows_9.5p2` parsed as 0.0.0 and Lima treated a current install as ancient. That refused multi-instance `limactl copy` (gated at 8.0) and forced `LogLevel=QUIET` on `limactl shell` (gated at 8.9), which hides ssh errors.

Part of a series continuing the native Windows OpenSSH work from #5205 and #5270: #5298, #5299, #5300, #5301. They are independent and can merge in any order. #5298 and #5301 both touch `pkg/sshutil/sshutil.go`, in different functions, and apply cleanly in either order.

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -632,7 +632,9 @@ func SSHOptsRemovingControlPath(opts []string) []string {
 }
 
 func ParseOpenSSHVersion(version []byte) *semver.Version {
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #5299, opened 2026-07-21
copytool: support native Windows OpenSSH

jandubois:
On Windows, `limactl copy` converted host paths to the Cygwin/MSYS form no matter which tool would consume them, and it passed ControlMaster options unconditionally. Native OpenSSH cannot use that path form and has no multiplexing, so scp failed with `getsockname failed`.

The two backends need different path forms, so every path now takes the form of the binary that reads it. Both also skip multiplexing on Windows.

Part of the native Windows OpenSSH series. It can merge in any order, except that #5337 appends to the same test file, so whichever lands second needs a trivial rebase there. The Windows jobs exercise only the Cygwin scp path, because the runners carry no rsync of their own. That coverage and validation on a plain Windows host both arrive with the CI jobs in #5342.

Assisted-by: Claude Opus 5

PR #5300, opened 2026-07-21
hostagent: support reverse-sshfs on plain Windows

jandubois:
The sftp-server that serves a mount and the host path handed to it now come from one install, the one holding the ssh.exe Lima selected at startup. Otherwise sshocker picks an sftp-server off PATH that may come from a different toolchain than the one the path was converted for, and cannot resolve what it is given.

The builtin driver instead gets the native form, because sshocker serves that one in-process, where a Cygwin path resolves to the wrong directory.

#5299 also touches `pkg/sshutil/sshutil.go`, `pkg/sshutil/sshutil_windows_test.go`, and `pkg/hostagent/hostagent.go`, in different places; the two apply cleanly in either order.

The Windows CI runner uses MSYS2 / Git for Windows, so it exercises the Cygwin path. End-to-end validation on a plain Windows host, with no Cygwin toolchain present, arrives with the plain-Windows CI jobs in #5342.

Assisted-by: Claude Opus 5

### DIFF
--- a/pkg/hostagent/mount.go
+++ b/pkg/hostagent/mount.go
@@ -8,18 +8,38 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"runtime"
 
 	"github.com/lima-vm/sshocker/pkg/reversesshfs"
 	"github.com/sirupsen/logrus"
 
-	"github.com/lima-vm/lima/v2/pkg/fsutil"
 	"github.com/lima-vm/lima/v2/pkg/limatype"
 	"github.com/lima-vm/lima/v2/pkg/sshutil"
 )
 
 type mount struct {
 	close func() error
+}
+
+// mountPathAndSftpServer returns location in the native path form, and the
+// sftp-server from sshExe's toolchain. Every driver gets the native form
+// because sshocker rewrites a Cygwin path to backslashes, which a Cygwin
+// ssh.exe then eats, leaving the server a path it cannot resolve. Both server
+// flavors accept the native form. The server is empty for the builtin driver,
+// which sshocker serves in-process, and when the toolchain ships none.
+func mountPathAndSftpServer(ctx context.Context, sshExe sshutil.SSHExe, driver, location string) (resolvedLocation, sftpServerBinary string) {
+	resolvedLocation = filepath.ToSlash(location)
+	if driver == limatype.SFTPDriverBuiltin {
+		return resolvedLocation, ""
+	}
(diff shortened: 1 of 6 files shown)


pkg/hostagent/mount.go
```diff
@@ -53,27 +52,40 @@ func (a *HostAgent) setupMount(ctx context.Context, m limatype.Mount) (*mount, e
...
 		if err != nil {
 			return nil, err
 		}
+		resolvedLocation, err = sshutil.PathForSSH(ctx, sshExe, m.Location)
+		if err != nil {
+			return nil, err
+		}
+		sftpServerBinary = sshutil.SftpServerForSSH(ctx, sshExe)
+		if sftpServerBinary == "" {
+			logrus.Warnf("reverse-sshfs: no sftp-server matching %#q found; sshocker will auto-detect from PATH (Windows OpenSSH.Server may be missing)", sshExe.Exe)
```
AkihiroSuda: nit: not necessary to mention sshocker

pkg/hostagent/mount.go
```diff
@@ -53,27 +52,40 @@ func (a *HostAgent) setupMount(ctx context.Context, m limatype.Mount) (*mount, e
 	logrus.Infof("Mounting %#q on %#q", m.Location, *m.MountPoint)
 
 	resolvedLocation := m.Location
+	var sftpServerBinary string
 	if runtime.GOOS == "windows" {
```
AkihiroSuda, line 77: Is this specific to Windows?

jandubois: Technically only the Warning is specific to Windows (on Unix it doesn't matter if `sftpserver` is the same directory). But the rest of the code is a no-op on Unix, so there is no point in running it.

I updated the comment to hopefully make this clear.

pkg/hostagent/mount.go
jandubois: Fixed

jandubois (commented):
Addressed both comments.

Also cached `sshExe` to avoid looking it up twice.

jandubois:
Turned to draft mode because it may need another change, but I'm out of time right now.

jandubois:
Last AI review: https://jandubois.github.io/lima/20260726-233238-pr-5300.html

I've fixed S1 and S3; I think the rest don't need to be addressed. This is the 3rd AI review round, and the suggestions are just changes to "fixes" made to address round 2 suggestions. I don't want to continue this, as it will blow up the size of the PR too much.

AkihiroSuda (approved):
Thanks

PR #5301, opened 2026-07-21
sshutil: remove the unused IsSSHCygwin

jandubois:
`PathForSSH` was its only caller. It switched to `cygpathForSSH`, which returns the sibling cygpath it needs, and `IsSSHCygwin` has had no caller since it landed in #5205.

The two assertions dropped from `TestCygpathForSSH` re-checked the boolean that the `cygpathForSSH` assertion on the preceding line already covers, so coverage is unchanged.

Part of a series continuing the native Windows OpenSSH work from #5205 and #5270: #5298, #5299, #5300, #5301. They are independent and can merge in any order. #5298 and #5301 both touch `pkg/sshutil/sshutil.go`, in different functions, and apply cleanly in either order.

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -139,14 +139,6 @@ var (
 	cygwinDetectCacheRW sync.RWMutex
 )
 
-// IsSSHCygwin reports whether sshExe is a Cygwin-based OpenSSH (Git for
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Still draft?

AkihiroSuda (approved):
Thanks

PR #5303, opened 2026-07-21
don't follow symlinks when writing copyToHost destination

Nashit-h:
copyToHost writes the guest-controlled bytes from `cat <guestFile>` to the host path in `host`, using os.WriteFile, which follows a symlink already present at that path. When `host` sits inside a writable mount, a guest can plant a symlink there (say pointing at ~/.ssh/authorized_keys) so the copy lands outside the declared destination. The write now goes through os.OpenRoot on the parent directory, so a final-component symlink whose target escapes that directory is refused.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -1316,7 +1316,9 @@ func copyToHost(ctx context.Context, sshConfig *ssh.SSHConfig, sshAddress string
 			return fmt.Errorf("failed to run %v: %#q: %w", cmd.Args, string(out), err)
 		}
 	}
-	if err := os.WriteFile(local, out, 0o600); err != nil {
+	// The destination may sit inside a writable mount, so don't follow a
+	// symlink planted there by the guest.
+	if err := osutil.WriteFileBeneathDir(local, out, 0o600); err != nil {
 		return fmt.Errorf("can't write to local file %#q: %w", local, err)
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


pkg/hostagent/hostagent.go
```diff
@@ -1316,8 +1316,32 @@ func copyToHost(ctx context.Context, sshConfig *ssh.SSHConfig, sshAddress string
...
 }
+
+// writeFileBeneathDir writes data to name without following a symlink at the
+// final path component. copyToHost writes guest-controlled bytes, and its
+// destination may sit inside a writable mount; a guest that plants a symlink at
+// name (e.g. pointing at ~/.ssh/authorized_keys) would otherwise get those
+// bytes written through the link to a file outside the declared destination.
+// os.OpenRoot confines the write to name's parent directory, so a symlink whose
+// target escapes that directory is refused.
+func writeFileBeneathDir(name string, data []byte, perm os.FileMode) error {
```
afbjorklund: This probably belongs in "util"

Nashit-h: Moved it to osutil as `osutil.WriteFileBeneathDir` (pkg/osutil/file.go), with the test alongside it. hostagent now just calls it.

AkihiroSuda (approved):
Thanks

jandubois:
I know this is already merged, but wouldn't it make more sense to call `os.RemoveFile` on the target path instead of still following symlinks underneath the same directory?

This is probably fine, but always removing the target first is obviously correct.

PR #5294, opened 2026-07-20
Extend the default timeout for Windows guest

mie313:
## What This PR Changes
This PR extends the default timeout duration for Windows guest to 30 minutes, because the initial boot of Windows guests usually takes longer than the current default timeout setting (10min).

## Linked Issue (Required in most cases)
#5255 

Closes #5255 

## How I Tested This
I tried booting Windows 11 VM on my Mac and confirmed that the initial booting completed.

## AI Usage
No

PR #5304, opened 2026-07-21
fix: only extend Windows guest timeout on initial boot

### LINKED ISSUE #5296: [Windows guest] `limactl start`: default `--timeout` should be shorter for non-initial boot
When booting the Windows guest the timeout is unconditionally stretched from 10 min to 30 min (#5294):
https://github.com/lima-vm/lima/blob/549afab3f3efa633f84ed20b0ef26b5cf3518eae/cmd/limactl/start.go#L624-L627

For non-initial boot it should remain 10 min.

Non-initial boot can be detected by checking this file:
https://github.com/lima-vm/lima/blob/406e25d7d67df5ed3bdf7f7eaeb51f3bc50f7e38/pkg/limatype/filenames/filenames.go#L86

opjt:
## What This PR Changes

`limactl start` always stretches the default timeout from 10m to 30m for Windows guests(#5294),  even when just restarting an already-installed instance.   
This change only applies the 30m timeout on the inital boot, by checking for the `win-done-installation` marker file

Closes #5296 

## How I Tested This

I verified the branching logic end-to-end without a real Windows boot:
- Created a minimal `os: windows` \ `vmType: qemu` instance (dummy disk image, no real ISO).
- Ran `limactl start`: with no `win-done-installation` marker, the log shows the timeout was extended to 30m
- After creating the marker file and restarting, the extension log does not appear (default 10m timeout is used)

### DIFF
--- a/cmd/limactl/start.go
+++ b/cmd/limactl/start.go
@@ -622,8 +622,10 @@ func startAction(cmd *cobra.Command, args []string) error {
 	}
 
 	if !cmd.Flags().Changed("timeout") && *inst.Config.OS == limatype.WINDOWS {
-		timeout = instance.WinDefaultWatchHostAgentEventsTimeout
-		logrus.Infof("Extended the default timeout to %v for Windows guest boot.", timeout)
+		if _, err := os.Stat(filepath.Join(inst.Dir, filenames.WinDoneInstallation)); err != nil {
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
Thanks

PR #5307, opened 2026-07-21
website: update hugo-extended, docsy

AkihiroSuda:
## What This PR Changes

- hugo-extended 0.133.1 -> 0.164.0 (via `npm run update:pkg:hugo`)
- docsy 0.11.0 -> 0.15.0, as docsy 0.11.0 is incompatible with Hugo >= 0.141 (Resource.Err removal)
- hugo.toml: allow HTML content files and Node.js parent-directory reads (PostCSS/browserslist), both restricted by the new security defaults introduced in Hugo 0.157
- hugo.toml: rename deprecated `includeFiles` -> `files`, `languageName` -> `label`; bump min Hugo version to 0.157.0
- Refresh caniuse-lite via `npx update-browserslist-db`

Closes https://github.com/lima-vm/lima/security/dependabot/20

## How I Tested This

https://deploy-preview-5307--lima-vm.netlify.app

> [!CAUTION]
> ~Not mergeable yet. The colors are awkward.~

## AI Usage

Assisted-by: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/website/hugo.toml
+++ b/website/hugo.toml
@@ -55,7 +55,7 @@ anchor = "smart"
 
 [languages]
 [languages.en]
-languageName ="English"
+label = "English"
 # Weight used for sorting.
 weight = 1
 [languages.en.params]
@@ -206,12 +206,21 @@ enable = false
   name = "Releases"
   url = "https://github.com/lima-vm/lima/releases"
 
+# Allow HTML content files such as content/en/_index.html
+# (denied by default since Hugo 0.157)
+[security]
+  allowContent = ['.*']
+  # Node.js runs PostCSS (autoprefixer); browserslist reads parent
+  # directories, which the default sandbox (allowRead = ['.']) denies.
+  [security.node.permissions]
+    disable = true
+
 [module]
   # uncomment line below for temporary local development of module
   # replacements = "github.com/google/docsy -> ../../docsy"
   [module.hugoVersion]
     extended = true
-    min = "0.110.0"
+    min = "0.157.0"
   [[module.imports]]
     path = "github.com/google/docsy"
     disable = false
@@ -236,4 +245,4 @@ enable = false
     source="templates"
     target="static/templates"
     language="en"
-    includeFiles = "README.md"
+    files = "README.md"
--- a/website/hugo.toml
+++ b/website/hugo.toml
@@ -55,7 +55,7 @@ anchor = "smart"
 
 [languages]
 [languages.en]
-languageName ="English"
+label = "English"
 # Weight used for sorting.
 weight = 1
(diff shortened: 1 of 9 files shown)


afbjorklund (dismissed):

website/hugo.toml
```diff
@@ -206,12 +206,21 @@ enable = false
...
   url = "https://github.com/lima-vm/lima/releases"
 
+# Allow HTML content files such as content/en/_index.html
+# (denied by default since Hugo 0.157)
+[security]
+  allowContent = ['.*']
+  # Node.js runs PostCSS (autoprefixer); browserslist reads parent
+  # directories, which the default sandbox (allowRead = ['.']) denies.
+  [security.node.permissions]
+    disable = true
```
jandubois: Doesn't this work (from https://gohugo.io/configuration/security/#nodepermissionsallowread)?


```suggestion
    disable = false
    allowRead = ['*']
```

website/hugo.toml
```diff
@@ -206,12 +206,21 @@ enable = false
...
+
 [module]
   # uncomment line below for temporary local development of module
   # replacements = "github.com/google/docsy -> ../../docsy"
   [module.hugoVersion]
     extended = true
-    min = "0.110.0"
+    min = "0.157.0"
```
jandubois: `security.allowContent` and the default HTML-content denial were introduced in Hugo 0.162, not 0.157 (https://github.com/gohugoio/hugo/releases/tag/v0.162.0).


```suggestion
    min = "0.162.0"
```

jandubois (commented):
LGTM, just some little tweaks.

It seems like the URLs for printable pages have changed, e.g. `docs/config/_print` becomes `_print/docs/config`, which is arguably more sensible.

I don't think we really care, but in case we do, here is a snippet courtesy of GPT 5.6 Sol that should restore URL compatibility. I have not checked if it has any bad side effects:

```toml
  [outputFormats.print]
  baseName = "_print/index"
  isHTML = false
  mediaType = "text/html"
  path = ""
  permalinkable = false
```

AkihiroSuda:
> It seems like the URLs for printable pages have changed, e.g. docs/config/_print becomes _print/docs/config, which is arguably more sensible.

I don't think there is somebody relying on the old URL

website/hugo.toml
AkihiroSuda: Applied

website/hugo.toml
AkihiroSuda: Applied

jandubois (approved):
Thanks, LGTM

PR #5314, opened 2026-07-22
end rsync option parsing before the copy paths

Nashit-h:
The rsync backend appends the copy paths straight after the flag list, so rsync parses any path beginning with a dash as an option. A file named `--rsh=<cmd>` that reaches the argv, for instance through a shell glob over a directory the guest can write via a writable mount, overrides the `-e` we pass and rsync runs it on the host. The scp backend already ends its flags with `--`; this does the same for rsync, and the test checks the dashed path lands after the separator.

### DIFF
--- a/pkg/copytool/copytool_test.go
+++ b/pkg/copytool/copytool_test.go
@@ -4,6 +4,7 @@
 package copytool
 
 import (
+	"slices"
 	"testing"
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #5323, opened 2026-07-23
reject invalid guest ip or protocol in forwardingAddresses

Nashit-h:
## What This PR Changes

The port forwarders read `api.IPPort` values off the guest-agent event stream. `forwardingAddresses` (in `pkg/portfwd/forward.go` and `pkg/hostagent/port.go`) hands the guest-supplied `Ip` back as `guest.HostString()`, and the forwarders copy `guest.Protocol` straight into `PortForwardEvent`. `net.ParseIP` was only used for rule matching, so an `Ip` that never parses still rides the catch-all rule through to the address, and both strings then land in the host-agent log and the events that `limactl watch`/`start` print to the terminal. A compromised guest can put escape sequences or forged log lines in `Ip` or `Protocol` and inject them into the operator console.

This validates the guest `Ip` and `Protocol` at the top of both `forwardingAddresses` functions (`Ip` must parse, `Protocol` must be `tcp` or `udp` per the proto contract) and drops anything else before it reaches the address, logs, or events.

No separate issue filed; this is a hardening fix for the guest-to-host trust boundary in the port forwarders. Happy to open one if you'd prefer to track it separately.

## How I Tested This

`go test ./pkg/portfwd/ ./pkg/hostagent/` (added regression tests in both packages that feed hostile `Ip`/`Protocol` values and assert no host or guest address is produced), plus `go build ./...`, `gofmt`, and `go vet`.

## AI Usage

None.

### DIFF
--- a/pkg/hostagent/port.go
+++ b/pkg/hostagent/port.go
@@ -83,6 +83,13 @@ func hostAddress(rule limatype.PortForward, guest *api.IPPort) string {
 
 func (pf *portForwarder) forwardingAddresses(guest *api.IPPort) (hostAddr, guestAddr string) {
 	guestIP := net.ParseIP(guest.Ip)
+	// guest.Ip and guest.Protocol arrive from the guest agent over gRPC and are
+	// otherwise untrusted. Drop entries that are not a valid ip / known protocol so
+	// raw guest bytes never reach the forwarded address, the host logs, or the
+	// port-forward events that limactl prints to the user's terminal.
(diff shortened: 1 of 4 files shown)


AkihiroSuda (changes requested):
Please fulfill https://github.com/lima-vm/.github/blob/main/PULL_REQUEST_TEMPLATE.md?plain=1

Nashit-h:
Filled out the template in the description. There's no separate issue for this one, but I can open one if you'd like it tracked.

Issue #5295, opened 2026-07-20
Reduce usage of `//go:embed` (`containerd.yaml`, `networks.TEMPLATE.yaml`, etc.)


AkihiroSuda:
As in regular template YAMLs (`/usr/local/share/lima/templates/*.yaml`), `containerd.yaml`, `networks.TEMPLATE.yaml`, etc. could be read from the local directory rather than from `//go:embed`?

https://github.com/search?q=repo%3Alima-vm%2Flima+%22%2F%2Fgo%3Aembed%22&type=code

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/limayaml/defaults.go#L61-L62

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/networks/config.go#L23-L24

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/cidata/template.go#L25-L31

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/autostart/launchd/launchd.go#L19-L23

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/autostart/systemd/systemd.go#L19-L20

## Driver-specific embeds (may stay embedded?)

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/driver/wsl2/vm_windows.go#L72-L73

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/driver/wsl2/wsl_driver_windows.go#L176-L177

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/driver/vz/vz_driver_darwin.go#L231-L232

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/pkg/driver/krunkit/krunkit_driver_darwin_arm64.go#L288-L289

## Guest agent embeds (should stay embedded probably, as the guest agent binary has been expected to be a single file)

https://github.com/lima-vm/lima/blob/434aabef3f0aabc9bd21e14720906c15e756249d/cmd/lima-guestagent/install_systemd_linux.go#L112-L113

afbjorklund:
[...]

PR #5324, opened 2026-07-23
pkg: externalize host default assets

fpolica91:
## Summary

- move the containerd and network defaults to `share/lima/defaults`
- load host resources from installed share directories with a source-tree fallback for tests and local builds
- package and validate the external resources in release artifacts
- keep driver-specific and guest-agent assets embedded

## Why

This starts #5295 with two host-only assets that have isolated consumers. The remaining CI-data, autostart, driver, and guest-agent embeds are intentionally unchanged so maintainers can review the resource-loading and packaging approach before expanding the scope.

## Testing

- `go test ./...`
- `go test -trimpath ./pkg/usrlocal ./pkg/limayaml ./pkg/networks`
- `make golangci-lint`
- `make yamllint`
- `make ls-lint`
- `make limactl`
- Linux and Windows cross-builds
- CLI runtime smoke tests for network and containerd defaults
- positive and negative artifact validation fixtures

Related to #5295.


pkg/usrlocal/usrlocal.go
```diff
@@ -126,6 +127,78 @@ func ShareLima() ([]string, error) {
...
+		dirs = append(dirs, sourceDir)
+	}
+	b, err := readFileFromDirs(name, dirs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to locate Lima resource %q: %w", name, err)
+	}
+	return b, nil
+}
+
+func sourceShareLima() string {
```
AkihiroSuda: Why do we need this?

Makefile
```diff
@@ -215,7 +215,7 @@ binaries: limactl helpers limactl-plugins guestagents \
 ################################################################################
 # _output/bin
 .PHONY: limactl lima helpers
-limactl: _output/bin/limactl$(exe) lima
+limactl: _output/bin/limactl$(exe) lima defaults
```
AkihiroSuda: "defaults" sounds confusing, as this target will be use for other potential `/usr/local/share/lima/*` contents too

pkg/usrlocal/usrlocal.go
fpolica91: @AkihiroSuda  This keeps go test ./... working directly from the source tree and CI runs the unit tests before make, thanks for reviewing.

Makefile
fpolica91: renamed

pkg/usrlocal/usrlocal.go
AkihiroSuda: I don't like this complexity.
The unit test should be adjusted to support the new path.

AkihiroSuda:
Please squash the commits

pkg/usrlocal/usrlocal.go
fpolica91: Resolved.

fpolica91:
@AkihiroSuda resolved the comments you left.

AkihiroSuda (dismissed):
Thanks, the implementation looks good, but we still do not have consensus on:
- https://github.com/lima-vm/lima/issues/5295

PR #2320, opened 2024-05-05
apptainer: fix AppArmor incompatibility with Ubuntu 24.04

AkihiroSuda:
Fix #2319

PR #1085, opened 2022-10-02
For apptainer, move TMPDIR away from tmpfs

afbjorklund:
It has some issues with it being mounted as "nodev", so revert to using a normal file system for $TMPDIR.

```console
$ limactl shell singularity singularity run -u -B $HOME:$HOME docker://alpine
INFO:    Converting OCI blobs to SIF format
WARNING: 'nodev' mount option set on /tmp, it could be a source of failure during build process
INFO:    Starting build...
Getting image source signatures
Copying blob 213ec9aee27d done  
Copying config 29f453b10b done  
Writing manifest to image destination
Storing signatures
2022/10/02 07:18:41  info unpack layer: sha256:213ec9aee27d8be045c6a92b7eac22c9a64b44558193775a1a7f626352392b49
2022/10/02 07:18:41  warn xattr{etc/shadow} ignoring ENOTSUP on setxattr "user.rootlesscontainers"
2022/10/02 07:18:41  warn xattr{/tmp/build-temp-98011911/rootfs/etc/shadow} destination filesystem does not support xattrs, further warnings will be suppressed
INFO:    Creating SIF file...
INFO:    Converting SIF file to temporary sandbox...
Apptainer> 
```

Using /var/tmp is the recommended revert, from the Fedora instructions regarding the change (in fc18):

https://fedoraproject.org/wiki/Features/tmp-on-tmpfs

```console
[anders@lima-singularity lima]$ df -h /var /tmp
Filesystem      Size  Used Avail Use% Mounted on
/dev/vda5        99G  708M   98G   1% /
tmpfs           2.0G     0  2.0G   0% /tmp
[anders@lima-singularity lima]$ findmnt /
TARGET SOURCE           FSTYPE OPTIONS
/      /dev/vda5[/root] btrfs  rw,relatime,seclabel,compress=zstd:1,space_cache=v2,subvolid=256,subvol=/root
[anders@lima-singularity lima]$ findmnt /tmp
TARGET SOURCE FSTYPE OPTIONS
/tmp   tmpfs  tmpfs  rw,nosuid,nodev,seclabel,nr_inodes=1048576,inode64
```

PR #5331, opened 2026-07-25
Update the Apptainer template for Ubuntu 26.04

afbjorklund:
Needs one workaround from before:

* https://github.com/lima-vm/lima/pull/1085

And we can drop one earlier workaround:

* https://github.com/lima-vm/lima/pull/2320

### DIFF
--- a/templates/apptainer-rootful.yaml
+++ b/templates/apptainer-rootful.yaml
@@ -21,6 +21,12 @@ provision:
     add-apt-repository -y ppa:apptainer/ppa
     apt-get update
     apt-get install -y apptainer-suid
+    # Ubuntu 24.10 moved /tmp over to tmpfs
+    # Apptainer wants a regular filesystem TMPDIR:
+    # - Building images has a problem with "nodev"
+    # - The image might be too big to fit in memory
+    echo "APPTAINER_TMPDIR=/var/tmp" > /etc/profile.d/apptainer.sh
+    echo "export APPTAINER_TMPDIR" > /etc/profile.d/apptainer.sh
 probes:
 - script: |
     #!/bin/bash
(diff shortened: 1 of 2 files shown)


templates/apptainer-rootful.yaml
```diff
@@ -21,6 +21,9 @@ provision:
     add-apt-repository -y ppa:apptainer/ppa
     apt-get update
     apt-get install -y apptainer-suid
+    # Ubuntu 24.10 moved /tmp over to tmpfs
```
AkihiroSuda, line 24: Could you add a comment to explain why /tmp cannot be used

afbjorklund: I added a comment, eventually it might not be needed in 1.6.0

afbjorklund:
I added a feature request, about Apptainer being able to pick a suitable temporary directory automatically:

* https://github.com/apptainer/apptainer/issues/3657

So packaging `/etc/profile.d/apptainer.sh` with the installation might not be needed in the future.

Similar workarounds have already been added to e.g. Podman https://github.com/podman-container-tools/podman/commit/772af18205eacf95ae55a4167f7f48e917bcd5e1

----

Previously it was only a problem on Fedora, and didn't affect more conservative systems like RHEL and Debian

But now they have also started to use tmp-on-tmpfs, e.g. Ubuntu 26.04 LTS, so now we need the workarounds...

https://fedoraproject.org/wiki/Features/tmp-on-tmpfs
https://web.archive.org/web/20250828080905/http://0pointer.de/blog/projects/tmp.html

https://www.debian.org/releases/trixie/release-notes/issues.html#the-temporary-files-directory-tmp-is-now-stored-in-a-tmpfs
https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#systemd-259

AkihiroSuda:
Please squash the commits

AkihiroSuda (approved):
Thanks

PR #5335, opened 2026-07-25
feat(image): add archVariant for x86-64-v3 image selection

### LINKED ISSUE #5160: Support amd64v3 variants of Ubuntu
https://www.phoronix.com/review/ubuntu-2610-amd64v3
amd64v3 variants aren't so fast, but we should cover it anyway mostly for compatibility testing.

Rather than complicating `--arch`, should we adopt the `--image-variant` proposal for specifying amd64v3 images too?
- https://github.com/lima-vm/lima/issues/5158

ekalinin:
## What This PR Changes

Add an `archVariant` field on image entries, orthogonal to the existing distro `variant` (server/minimal), plus a `--arch-variant` flag to select it. Selection at start time filters by archVariant in parallel with variant, so the base image stays the default and `--arch-variant v3` picks the v3 image. The arch field is left untouched (no x86_64v3).

Wire up the variant entries and teach the update scripts to keep the vendor arch name (`amd64v3, x86_64_v2`) for the stream-JSON/CHECKSUM lookup while emitting `arch: x86_64` for lima:

  - Ubuntu 26.04: `amd64v3` image as `archVariant: v3`.
  - AlmaLinux 10: `x86_64_v2` image as `archVariant: v2` (AlmaLinux 10 ships x86-64-v3 by default and x86-64-v2 for older hardware).

Closes #5160

## How I Tested This

  - Unit tests: `go test ./pkg/limayaml/... ./pkg/instance/... ./cmd/limactl/editflags/...` (added `TestUnmarshalImageArchVariant`).
  - Lint/format: `gofmt, shfmt, shellcheck, yamllint` all clean; `limactl generate-jsonschema` output includes the new archVariant field.
  - Functional check with `limactl create`:
    - `template:ubuntu-26.04 --arch-variant v3` reduces the resolved image list to the amd64v3 entries only; without the flag the base amd64 image is first and stays the default.
    - `template:almalinux-10 --arch-variant v2` reduces the list to the x86_64_v2 entries only; without the flag the default x86_64 (v3) image is first and stays the default.
  - Verified the update-script arch mapping in isolation: `amd64v3 / amd64v2 / amd64v4 and x86_64_v2` map to `x86_64`, while `riscv64` and the other arches are left untouched.

## AI Usage

Assisted-by: Claude Opus 4.8 [Claude Code]

### DIFF
--- a/templates/_images/ubuntu-26.04.yaml
+++ b/templates/_images/ubuntu-26.04.yaml
@@ -5,6 +5,13 @@ images:
   arch: "x86_64"
   digest: "sha256:117816726abbdefc5ef3e38902e81a76f1c76c3610e709999d0885f9d5d9b477"
   variant: "server"
+# x86-64-v3 optimized image. Still runs as x86_64, distinguished by archVariant.
+# Select it with `limactl start --arch-variant v3`.
+- location: "https://cloud-images.ubuntu.com/releases/resolute/release-20260720/ubuntu-26.04-server-cloudimg-amd64v3.img"
+  arch: "x86_64"
+  digest: "sha256:2e7bc9471ec729b7d9ee59958258da6d3943c86b50b6c98d7dedb84cc6724bbf"
+  variant: "server"
+  archVariant: "v3"
 - location: "https://cloud-images.ubuntu.com/releases/resolute/release-20260720/ubuntu-26.04-server-cloudimg-arm64.img"
   arch: "aarch64"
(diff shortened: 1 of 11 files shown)


templates/_images/ubuntu-26.04.yaml
AkihiroSuda, line 1: almalinux-10 has to be updated too
https://repo.almalinux.org/almalinux/10.2/cloud/

ekalinin: Thanks! Done.

AkihiroSuda (approved):
Thanks

afbjorklund (approved):
LGTM

PR #5336, opened 2026-07-25
feat(vz): add osOpts.Darwin.suppressFirstLoginSetup

### LINKED ISSUE #5186: vmOpts.vz.suppressFirstLoginSetup: skip macOS's first-login Setup Assistant screens for automated guest provisioning
I'm proposing an opt-in way to skip macOS's first-login Setup Assistant wizard when Lima provisions a macOS VZ guest. Without this feature, every new macOS guest boots straight into that wizard and just sits there until someone manually clicks through it in the VM's GUI window; with it enabled, Lima pre-populates the guest's setup state during provisioning so the VM comes up at a ready-to-use desktop instead, with no first-login screens to interact with.

macOS's Setup Assistant ("mini-buddy") walks through several first-login screens — Accessibility, Apple Pay, iCloud, Screen Time, Siri, Touch ID, Terms of Address, and more — the same wizard a person sees unboxing a new Mac. That flow doesn't fit an automated, disposable VM, and macOS's own Software Update separately asks the user to opt into or out of automatic updates on that same first login. There's currently no way to skip either dialog.

I have a working implementation and wanted to open this issue before submitting a PR.

This proposes a new `vmOpts.vz.suppressFirstLoginSetup` option (opt-in, default unset). When enabled, Lima writes the SetupAssistant and SoftwareUpdate preference files into the guest as root during account provisioning, before any GUI session starts — marking each wizard pane as already seen and disabling the automatic-update prompt. Because the preferences exist before the first login rather than being changed afterward, macOS treats them as the guest's actual initial state instead of resetting them. The option also accepts a custom SetupAssistant plist, for anyone who wants different defaults than the one Lima ships.

Minimal opt-in, using Lima's built-in plist template:

```yaml
vmOpts:
  vz:
    suppressFirstLoginSetup: {}
```

Supplying a custom `com.apple.SetupAssistant.plist` instead of the built-in one:

```yaml
vmOpts:
  vz:
    suppressFirstLoginSetup:
      plist: |
        <?xml version="1.0" encoding="UTF-8"?>
        <plist version="1.0">
        <dict>
          <key>MiniBuddyLaunchReason</key><integer>0</integer>
          <!-- ...remaining keys per your target OS build/version... -->
        </dict>
        </plist>
```

## Scope and limitations

This is a Darwin-guest, VZ-driver feature — QEMU macOS guests aren't affected, since this hooks into the VZ-specific cloud-init-equivalent provisioning path (fakecloudinit) that only runs on Darwin. It's opt-in only; without the setting, guest behavior doesn't change. Because macOS ties "setup already done" to the exact OS build and version, the built-in plist template queries `sw_vers` from inside the guest at provisioning time rather than hardcoding version strings, so it keeps working as the guest OS updates. Anyone supplying a custom plist is responsible for keeping its build/version stamps accurate for their target OS release.

This workaround is only needed for macOS 26 and earlier guests[^1].

## Where changes land

The implementation threads one boolean (plus an optional plist string) from the instance config down to the guest account-creation step:

- The `lima.yaml` schema gains `vmOpts.vz.suppressFirstLoginSetup`, an opt-in struct with an optional custom `plist` field.
- Instance provisioning (the `cidata` package) reads the option, embeds a custom plist file on the guest provisioning volume if one is supplied, and carries the enable flag through the cloud-init-equivalent user-data payload sent to the guest.
- The Darwin fakecloudinit guest agent does the actual work: it writes the per-user SetupAssistant plist and the system-wide SoftwareUpdate preferences as root, before the account's first login. It also switches the "does this user already exist" check from a filesystem test to an OpenDirectory query, since other provisioning steps can pre-create a home directory before the account itself exists.
- Documentation gains a new section under the macOS guest docs explaining the option and showing the built-in plist template.

## Functional flow

```mermaid
flowchart TD
    A["lima.yaml: vmOpts.vz.suppressFirstLoginSetup"] --> B["cidata: templateArgs()"]
    B --> C{"Custom plist supplied?"}
    C -->|yes| D["Embed setup-assistant.plist on cidata volume"]
    C -->|no| E["Flag only"]
    B --> F["user-data template: suppress_first_login_setup: true"]
    D --> G["Guest boots, cidata volume mounted"]
    E --> G
    F --> G
    G --> H["fakecloudinit: processUserData()"]
    H --> I["createUser(): check dscl / OpenDirectory"]
    I -->|account exists| J["Skip creation"]
    I -->|new account| K["Create account + home directory"]
    K --> L{"suppressFirstLoginSetup set?"}
    L -->|no| M["Done — normal first login, wizard shown"]
    L -->|yes| N["suppressFirstLoginScreens() runs as root"]
    N --> O{"Custom plist on cidata volume?"}
    O -->|yes| P["Use supplied plist verbatim"]
    O -->|no| Q["Build plist: query sw_vers, stamp build/version keys"]
    P --> R["Write com.apple.SetupAssistant.plist to user Preferences"]
    Q --> R
    N --> S["defaults write com.apple.SoftwareUpdate prefs"]
    R --> T["First GUI login"]
    S --> T
    T --> U["Setup Assistant + Update dialog skipped"]
```

---

Note that I am using Claude Code (Sonnet 4.7 & Sonnet 5) to assist with this work.

[^1]: Starting with macOS 27 guests, Apple's Virtualization framework provides native provisioning via [`VZMacGuestProvisioningOptions`](https://developer.apple.com/documentation/virtualization/vzmacguestprovisioningoptions) (announced at WWDC26): the host app supplies a full name, username, password, and can enable auto-login and SSH, and Setup Assistant itself honors those options on first boot — no plist workaround needed. That only applies once both host and guest support the new API, so this feature still fills the gap for macOS 26 and earlier guests.

trodemaster:
macOS VZ guests boot straight into the Setup Assistant ("mini-buddy") first-login wizard and sit there until someone clicks through it in the VM's GUI window -- Accessibility, Apple Pay, iCloud, Screen Time, Siri, Touch ID, Terms of Address, and more, plus a separate Software Update opt-in/out prompt on the same first login. Neither can currently be skipped, which blocks unattended macOS guest provisioning.

This adds an opt-in osOpts.Darwin.suppressFirstLoginSetup option. When set, fakecloudinit writes the per-user SetupAssistant plist and the system-wide SoftwareUpdate preferences as root during account creation, before the account's first login -- since the preferences exist before first login rather than being changed afterward, macOS treats them as the guest's actual initial state instead of resetting them. The built-in plist template queries sw_vers from inside the guest at provisioning time rather than hardcoding version strings, so it keeps working as the guest OS updates; a custom plist can be supplied instead via the option's `plist` field for anyone who wants different defaults.

This also switches the guest account-existence check from a filesystem test to an OpenDirectory (dscl) query, since other provisioning steps can pre-create a home directory before the account itself exists, which was making the filesystem check unreliable.

This is a Darwin-guest, VZ-driver-only feature.  It's opt-in only; without the setting, guest behavior doesn't change. It targets macOS 15/26 guests; macOS 27+ guests will have native provisioning support via VZMacGuestProvisioningOptions once that API is available on both host and guest, so this fills the gap until then.

Tested go build/go vet/go test/gofmt clean on this branch. Verified the osOpts.Darwin.suppressFirstLoginSetup config path end-to-end through YAML unmarshal into DarwinOpts via a direct test against the real limayaml. Also ran live end-to-end boot tests against real macOS 26.5 and macOS 15.6.1 VZ guests, each built with only this change applied and no other local patches: on both, with the option set, the guest's com.apple.SetupAssistant.plist and system-wide SoftwareUpdate preferences were confirmed written before first login with the guest's actual build/version substituted in, the SetupAssistant process never launched, and the VM booted straight to the standard login window instead of the first-login wizard.

Fixes #5186

Assisted-by: Claude Code (Sonnet 4.7 & Sonnet 5)

### DIFF
--- a/pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
+++ b/pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
@@ -16,6 +16,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"text/template"
 
 	"github.com/goccy/go-yaml"
 	"github.com/sethvargo/go-password/password"
@@ -117,7 +118,7 @@ func processUserData(ctx context.Context, mnt string) error {
 		}
 	}
 	for _, u := range userData.Users {
-		if err := createUser(ctx, &u); err != nil {
+		if err := createUser(ctx, &u, mnt); err != nil {
 			errs = append(errs, fmt.Errorf("failed to create user %#q: %w", u.Name, err))
 		}
 	}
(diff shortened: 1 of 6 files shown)


AkihiroSuda:
> QEMU macOS guests aren't affected

QEMU macOS guests aren't supported in the first place.
Please verify the AI output before submission.

pkg/cidata/cidata.TEMPLATE.d/user-data
```diff
@@ -183,6 +183,10 @@ ca_certs:
   {{- end }}
 {{- end }}
 
+{{- if .SuppressFirstLoginSetup }}
+suppress_first_login_setup: true
```
AkihiroSuda: user-data has to conform to the cloud-config spec.

AkihiroSuda (changes requested):
user-data has to conform to the cloud-config spec.

pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
```diff
@@ -219,7 +220,119 @@ func populateHomeDir(ctx context.Context, uid int, homedir string) error {
 	return nil
 }
 
-func createUser(ctx context.Context, u *cloudinittypes.User) error {
+// setupAssistantPlistTemplate is the built-in com.apple.SetupAssistant.plist.
+// The Build/Version stamps are what macOS checks to decide whether setup is
+// already complete; without them macOS resets MiniBuddyLaunchReason to 13 on
```
AkihiroSuda, line 225: > 13

What is this magic number?

website/content/en/docs/usage/guests/macos.md
```diff
@@ -33,6 +33,81 @@ limactl shell macos cat /Users/${USER}.guest/password
...
+By default, macOS shows a series of setup wizard screens (Setup Assistant /
+mini-buddy) on the first GUI login. For automated or headless-style macOS VMs
+this is inconvenient. Set `osOpts.Darwin.suppressFirstLoginSetup` to have Lima
+pre-populate the relevant preference plists during provisioning, before any GUI
+session starts, so the setup screens are skipped automatically:
+
+```yaml
+osOpts:
+  Darwin:
+    suppressFirstLoginSetup: {}
```
AkihiroSuda: Why `{}`, not `true`?

website/content/en/docs/usage/guests/macos.md
```diff
@@ -33,6 +33,81 @@ limactl shell macos cat /Users/${USER}.guest/password
 - Password-less sudo is disabled, except for `/sbin/shutdown -h now` (see [Sudo](/docs/config/sudo/) — this is not currently configurable on macOS)
 - Several features are not implemented yet. See [Caveats](#caveats) below.
 
+## Suppressing first-login setup screens
```
AkihiroSuda:
```suggestion
## Advanced topics
### Suppressing first-login setup screens
| ⚡ Requirement | Lima >= 2.3, macOS >= (ADD VERSION HERE)  |
|-------------------|-----------------------------|
```

website/content/en/docs/usage/guests/macos.md
```diff
@@ -33,6 +33,81 @@ limactl shell macos cat /Users/${USER}.guest/password
...
+
+This writes `com.apple.SetupAssistant.plist` into the guest user's home
+directory and pre-configures `com.apple.SoftwareUpdate` system preferences so
+that the "Update Mac Automatically" dialog is also suppressed. The preferences
+are written as root (via the Lima guest agent) before first login, so macOS
+reads them as the authoritative initial state and does not reset them.
+
+**Default:** unset — setup screens are shown as normal.
+
+### Custom plist
```
AkihiroSuda, line 61: Do we really need to implement and document such details?

pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go
trodemaster: MiniBuddy has state tracking using these numbers. They are not documented and annotate things like macOS has booted and successfully run minibuddy.

website/content/en/docs/usage/guests/macos.md
trodemaster: The intention is that the user could provide the customized plist as described below.

website/content/en/docs/usage/guests/macos.md
trodemaster: The intent of allowing a custom plist is to address changes that Apple can make. We could simplify and use a Boolean and not provide a way to use a custom plist. The risk is that Apple changes something, and then the only way a user can work around it is by a new release of Lima.

pkg/cidata/cidata.TEMPLATE.d/user-data
trodemaster: OK, I have shifted to a file marker approach. Is this workable?
https://github.com/lima-vm/lima/blob/c6b2a4a11a06320da89177cddf12e4d555f63d73/pkg/guestagent/fakecloudinit/fakecloudinit_darwin.go#L397-L406

website/content/en/docs/usage/guests/macos.md
AkihiroSuda: I still feel plist could be provided in a separate option

trodemaster: Ok I have made this change.

https://github.com/lima-vm/lima/pull/5336/changes#diff-86ab405b8c8c00d3f02855b2156b221014f993d5960853369e980beded873ffcR73

website/content/en/docs/usage/guests/macos.md
```diff
@@ -33,6 +33,84 @@ limactl shell macos cat /Users/${USER}.guest/password
 - Password-less sudo is disabled, except for `/sbin/shutdown -h now` (see [Sudo](/docs/config/sudo/) — this is not currently configurable on macOS)
 - Several features are not implemented yet. See [Caveats](#caveats) below.
 
+## Advanced topics
+### Suppressing first-login setup screens
+| ⚡ Requirement | Lima >= 2.3, macOS >= (ADD VERSION HERE)  |
```
AkihiroSuda: `ADD VERSION HERE` needs to be fulfilled

trodemaster: Added macOS 13 as the supported version.

AkihiroSuda (approved):
Thanks

PR #5337, opened 2026-07-26
copytool: report a bad path instead of a missing tool

jandubois:
A path naming an instance that does not exist reached the user as "neither rsync nor scp found on host", and as "rsync not available on guest(s)" under `--backend=rsync`. Both backends now report the parse error itself, and the auto backend's fallback message no longer guesses which tool is missing.

This is independent of the native Windows OpenSSH series, so it can merge in any order. But #5299 appends a test to the end of `pkg/copytool/copytool_test.go` too, so whichever lands second needs a rebase there.

## Console output

`-` is before the change, `+` is after. The last two commands run with `scp` removed from `$PATH`; the middle one still has `rsync`, the last has neither.

```diff
 $ limactl copy --backend=rsync nonexistent:/tmp/x /tmp/y
-FATA[0000] rsync not available on guest(s)
+FATA[0000] instance `nonexistent` does not exist, run `limactl create nonexistent` to create a new instance

 $ limactl copy nonexistent:/tmp/x /tmp/y
-FATA[0000] neither rsync nor scp found on host: scp not found on host: exec: "scp": executable file not found in $PATH
+FATA[0000] instance `nonexistent` does not exist, run `limactl create nonexistent` to create a new instance

 $ limactl copy /tmp/a /tmp/b
-FATA[0000] neither rsync nor scp found on host: scp not found on host: exec: "scp": executable file not found in $PATH
+FATA[0000] no usable copy tool on host: scp not found on host: exec: "scp": executable file not found in $PATH
```

With scp on the host, `limactl copy nonexistent:/tmp/x /tmp/y` already printed the instance error, from the copy command rather than from `New()`, so the auto backend looks the same there.

Assisted-by: Claude Opus 5

### DIFF
--- a/pkg/copytool/copytool_test.go
+++ b/pkg/copytool/copytool_test.go
@@ -44,3 +44,13 @@ func TestRsyncCommandEndsOptionParsing(t *testing.T) {
 	assert.Assert(t, sep != -1, "rsync args must contain the %#q separator: %v", "--", cmd.Args)
 	assert.Assert(t, slices.Index(cmd.Args, dashPath) > sep, "path %#q must come after %#q: %v", dashPath, "--", cmd.Args)
 }
+
+// TestNewAutoSurfacesPathError verifies that the default backend reports a bad
+// path as itself, rather than as a missing copy tool.
+func TestNewAutoSurfacesPathError(t *testing.T) {
+	t.Setenv("LIMA_HOME", t.TempDir())
+	paths := []string{"nonexistent-instance-for-test:/tmp/x", "/tmp/y"}
+
+	_, err := New(t.Context(), string(BackendAuto), paths, &Options{})
+	assert.ErrorContains(t, err, "instance `nonexistent-instance-for-test`")
+}
--- a/pkg/copytool/copytool.go
+++ b/pkg/copytool/copytool.go
@@ -62,23 +62,28 @@ func New(ctx context.Context, backend string, paths []string, opts *Options) (Co
 		if err != nil {
 			return nil, err
 		}
-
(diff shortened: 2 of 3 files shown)


jandubois:
Ready for review.

AI review: https://jandubois.github.io/lima/20260726-220537-pr-5337.html

The commit message has been updated, and the branch has been rebased since the AI review, but the changes are identical. I don't think any findings are actionable.

AkihiroSuda (commented):
This kind of PR could be reviewed easier if the console output were attached

pkg/copytool/copytool_test.go
```diff
@@ -44,3 +44,13 @@ func TestRsyncCommandEndsOptionParsing(t *testing.T) {
...
 }
+
+// TestNewAutoSurfacesPathError verifies that the default backend reports a bad
+// path as itself, rather than as a missing copy tool.
+func TestNewAutoSurfacesPathError(t *testing.T) {
+	t.Setenv("LIMA_HOME", t.TempDir())
+	paths := []string{"nonexistent-instance-for-test:/tmp/x", "/tmp/y"}
+
+	_, err := New(t.Context(), string(BackendAuto), paths, &Options{})
+	assert.ErrorContains(t, err, "instance `nonexistent-instance-for-test`")
```
AkihiroSuda, line 55: This should assert the reason of the error, not just the instance name involved in the error?

jandubois: updated

jandubois:
Rebased

AkihiroSuda (approved):
Thanks

Issue #5328, opened 2026-07-24
[feat] generate json schema and publish to schemastore.org for editor DX


loynoir:
### Description

# feat

generate json schema and publish to schemastore.org for editor DX

## current

Within editor

- copy paste https://github.com/lima-vm/lima/blob/master/templates/archlinux.yaml

- No idea what is the full layout of lima.yaml

- No editor tab completion

- No editor validation

## suggestion

1. `lima` team generate lima.shema.json

2. `lima` team publish to https://github.com/SchemaStore/schemastore/blob/master/src/api/json/catalog.json

To have

- crystal clear layout of lima.yaml

- editor tab completion

- editor validation

afbjorklund:
You can use `make schema-limayaml.json`, to generate the jsonschema file.

* https://github.com/lima-vm/lima/issues/2305

It is not checked in anywhere afaik, and it is not part of the GitHub releases.

It should be up-to-date though, since it runs `make check-jsonschema` to test.

afbjorklund:
But the recommended reference is **not** jsonschema, but default.yaml:

https://github.com/lima-vm/lima/blob/master/templates/default.yaml

https://github.com/lima-vm/lima/discussions/1489#discussioncomment-5619640

It could still be useful with both, just having documentation doesn't give tab completion and syntax...

loynoir:
similar FR and PR

`vcpkg` publish to schemastore

https://github.com/microsoft/vcpkg/issues/34706

https://github.com/SchemaStore/schemastore/pull/3341/changes

https://github.com/SchemaStore/schemastore/blob/master/src/api/json/catalog.json#L2137-L2148

```json
    {
      "name": "vcpkg manifest file",
      "description": "vcpkg manifest file",
      "fileMatch": ["vcpkg.json"],
      "url": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg.schema.json"
    },
    {
      "name": "vcpkg configuration file",
      "description": "vcpkg configuration file",
      "fileMatch": ["vcpkg-configuration.json"],
      "url": "https://raw.githubusercontent.com/microsoft/vcpkg-tool/main/docs/vcpkg-configuration.schema.json"
    },
```
---

The schema store also support lint toml and yaml, but the url seems must `XXX.schema.json`

https://github.com/SchemaStore/schemastore/blob/master/src/api/json/catalog.json#L67-L72

```json
    {
      "name": "Releasaurus Config",
[...]

Issue #2957, opened 2024-12-02
lima-vm.io/docs should show non-master releases


AkihiroSuda:
![image](https://github.com/user-attachments/assets/14832369-12c2-4f12-8a96-3c3db4f5fcd6)

https://lima-vm.io/docs should show non-master releases

olamilekan000:
Does something like this work?

![Screenshot 2025-01-09 at 00 09 21](https://github.com/user-attachments/assets/5a7f2e3b-b35f-49d5-a480-a3d66e167ff6)
Users get redirected to https://github.com/lima-vm/lima/releases when the click the button.

AkihiroSuda:
The latter one seems good (until we can actually build multi-versioned document site)

PR #5339, opened 2026-07-26
Add jsonschema to the release artifacts

afbjorklund:
So that the json can be referenced from schema store:

`https://github.com/lima-vm/lima/releases/latest/download/schema-limayaml.json`

Generate the file before cleaning the native `limactl`.

(since the native limactl is required, in order to generate the jsonschema from go)

## What This PR Changes

Prepares an artifact, that can be used for making a PR to https://schemastore.org

Issue #5328

## How I Tested This

`make artifacts-jsonschema`

## AI Usage

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ DEST := $(shell echo "$(DESTDIR)/$(PREFIX)" | sed 's:///*:/:g; s://*$$::')
 GITHUB_ACTIONS ?= false
 
 GO ?= go
+CP ?= cp
 TAR ?= tar
 ZIP ?= zip
 PLANTUML ?= plantuml # may also be "java -jar plantuml.jar" if installed elsewhere
@@ -179,6 +180,7 @@ help-artifact:
 	@echo  '#   make GOOS=linux GOARCH=arm64 artifact'
 	@echo
 	@echo  'Targets for miscellaneous artifacts:'
+	@echo  '- artifacts-jsonschema      : Build artifacts for schema-limayaml.json'
 	@echo  '- artifacts-misc            : Build artifacts for go.mod, go.sum, and vendor'
 
 ################################################################################
@@ -545,7 +547,9 @@ ifeq ($(native_compiling),true)
 endif
 
 ################################################################################
-default-template.yaml: _output/bin/limactl$(exe)
+default-template.yaml: _output/bin/limactl$(exe) templates/default.yaml \
+                       _output/share/lima/templates/_images/ubuntu.yaml \
+                       _output/share/lima/templates/_default/mounts.yaml
 ifeq ($(native_compiling),true)
 	$< tmpl copy --embed-all templates/default.yaml $@
 endif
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -777,6 +779,10 @@ artifact-%-arm64 artifact-%-aarch64 artifact-arm64 artifact-aarch64: GOARCH = ar
 artifact-%: $$(call generate_manpages_if_needed)
 	make clean artifact GOOS=$(GOOS) GOARCH=$(GOARCH)
 
+.PHONY: artifacts-jsonschema
+artifacts-jsonschema: schema-limayaml.json | _artifacts
```
AkihiroSuda, line 785: Could this be rather pushed to the website?

afbjorklund: There are no versions in the website, afaik

AkihiroSuda, line 785: `schema-instance.json` might be more descriptive?
Or just `schema.json` if it can cover `networks.yaml` etc.

AkihiroSuda: - https://github.com/lima-vm/lima/issues/2957

afbjorklund: The name is globally visible (with the other files), like schema-cloud-config-v1.json and so on.

So it should be called something with "lima" in it (I think), and `Instance` is a different structure?
We only wanted to cover what was in the `lima.yaml`, and in the template files included from it

https://github.com/lima-vm/lima/blob/master/pkg/limatype/lima_yaml.go

I don't think we need any jsonschema for the Instance json, since it will not be edited by hand?
It would of course be _possible_ to generate one for it, only need to change genschema.go if so

https://github.com/lima-vm/lima/blob/master/pkg/limatype/lima_instance.go

afbjorklund: I'm not sure how great docsy is with "raw" json files, but maybe we could document it more readable?

Like how we build the manpages, and include them in the artifacts and then _also_ build as markdown.

With https://github.com/sbrunner/jsonschema2md

https://gist.github.com/afbjorklund/109a05c331c25a64554cb7bd6d1556da#file-schema-limayaml-md

Makefile
```diff
@@ -545,6 +547,8 @@ ifeq ($(native_compiling),true)
 endif
 
 ################################################################################
+default-template.yaml: _output/share/lima/templates/_images/ubuntu.yaml
+default-template.yaml: _output/share/lima/templates/_default/mounts.yaml
 default-template.yaml: _output/bin/limactl$(exe)
```
AkihiroSuda: Should be squashed to one line?

AkihiroSuda: Also please squash the commits

afbjorklund: I added them as dependencies, and moved the go comments to a separate PR (it was unrelated)

The missing build dependencies were also separate, but you don't see them if you do "make" first.

AkihiroSuda (approved):
Thanks

unsuman (approved):
Thanks, LGTM!

PR #5340, opened 2026-07-26
shell: measure the --sync directory depth on the host's own path

jandubois:
`--sync` refuses a directory too close to the filesystem root, but on Windows it measured a path already converted to the `/c/...` form while still splitting on backslashes. Every directory counted as depth 1, so `--sync` could not run there at all. Measuring the native path keeps one threshold meaning the same thing everywhere.

Only Windows separates on a backslash. Elsewhere it is an ordinary filename character, and counting it as a separator would score a directory just below the root deep enough to pass, so the caller now says which form it holds rather than leaving it implied.

That guard was also the only thing stopping `--sync` when the conversion failed or returned nothing, because the empty string scored 1. Both now need their own error. An empty directory reaches rsync as the toolchain root, and with stdout not a terminal the sync back runs `--delete` against it unprompted.

A wsl2 guest already reaches the host directory through the `/mnt` automount, so `--sync` cannot isolate it from host files there. It now refuses that combination up front.

This is pre-existing and unrelated to the native Windows OpenSSH series. No test caught it because the BATS `--sync` tests run only on `ubuntu-24.04` and the Windows jobs never invoke `--sync`. Running BATS on Windows needs submodules, an rsync install, and a vm-type hook in `ensure_instance`, so that waits for the plain-Windows CI work.

Assisted-by: Claude Opus 5

### DIFF
--- a/cmd/limactl/shell.go
+++ b/cmd/limactl/shell.go
@@ -213,23 +213,37 @@ func shellAction(cmd *cobra.Command, args []string) error {
 	if syncHostWorkdir && len(inst.Config.Mounts) > 0 {
 		return errors.New("cannot use `--sync` when the instance has host mounts configured, start the instance with `--mount-none` to disable mounts")
 	}
+	// A wsl2 guest already reaches the host directory through the /mnt automount,
+	// so `--sync` cannot isolate it from host files the way it does elsewhere.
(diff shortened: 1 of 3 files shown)


jandubois:
AI review: https://jandubois.github.io/lima/20260727-171515-pr-5340.html

unsuman (approved):
Thanks, LGTM!

PR #5341, opened 2026-07-26
docs: document the Windows host SSH toolchain

jandubois:
Document the host SSH toolchain now that Lima supports native Windows OpenSSH: which binaries ship by default, why to prefer full Git for Windows over MinGit, and how the Cygwin-vs-native path form is chosen. Also replaces the Known Issues bullet that claimed Windows ships no `ssh.exe`, since only `xz`, `bzip2`, and `zstd` rootfs archives still need an external binary.

This describes the state after #5299 and #5300 land, so it should merge after them. Merging it first would publish toolchain behaviour that master does not yet have.

`paths-ignore` in test.yml keeps the test workflow off this PR, but lint-quick.yml has no such filter and does run.

PR #5342, opened 2026-07-26
CI: factor Windows setup into composite actions; add plain-Windows jobs

jandubois:
Move the Windows runner setup out of `test.yml` into reusable composite actions, and add `windows-plain-{wsl2,qemu}` jobs that run Lima on a host with Git for Windows, MSYS2, and Cygwin removed. The plain-host action checks the toolchain is gone across filesystem, PATH, known binaries, and registry, and that the native OpenSSH client is present, so a runner-image change in either direction fails loudly instead of masking a regression.

With test.yml no longer setting `_LIMA_WINDOWS_EXTRA_PATH`, that env var loses its last user, so the implementation and its docs entry go too. It prepended `C:\Program Files\Git\usr\bin` to limactl's PATH, which is how the MSYS2 jobs got a Cygwin ssh.

The MSYS2 jobs install `openssh` in its place, so they keep exercising the Cygwin toolchain while the plain jobs exercise the native one. They also gain `rsync`. Without it `hack/test-templates.sh` skips its rsync copy backend instead of failing, so the path conversion that backend performs on Windows never runs.

The "windows" job id becomes "windows-wsl2". Its `name:` is unchanged, so required status checks are unaffected.

Note that the new plain-Windows jobs run `limactl copy` in both directions on a host with no Cygwin toolchain, which is what #5299 fixes, so expect them to fail until that lands.

A separate do-not-merge branch runs these jobs together with #5299, #5300, #5337, #5340 and #5341.

Assisted-by: Claude Opus 5

### DIFF
--- a/.github/actions/windows_msys2_prep/action.yml
+++ b/.github/actions/windows_msys2_prep/action.yml
@@ -1,0 +1,24 @@
+name: 'windows msys2 prep'
+description: >
+  Adds C:\msys64\usr\bin to PATH and installs the MSYS2 packages the Windows
+  integration jobs need. The PATH prepend persists into subsequent job steps
+  via $GITHUB_PATH, so the test step can call cygpath / bash / pacman tools
+  directly without re-prepending. The Make step inherits the prepend too, so
+  make's $(shell find ...) finds GNU find instead of System32\FIND.EXE.
+runs:
+  using: composite
+  steps:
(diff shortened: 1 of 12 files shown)


jandubois:
AI review: https://jandubois.github.io/lima/20260728-180750-pr-5342.html

jandubois:
The current CI failures are expected; they go away once all the prerequisite PRs have been merged (see #5343).

PR #5350, opened 2026-07-27
ci: Also run CI jobs on almalinux-9

whoschek:
## What This PR Changes

Make sure AlmaLinux-9 (which is among the canned Lima templates) continues to work.

FYI, the analog change actually fails for AlmaLinux-10 but that's a separate issue with the Lima mDNS script and its missing `kernel-modules-extra-$(uname -r)` dependency.

## How I Tested This

Passed the Lima Github CI matrix.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -308,6 +308,7 @@ jobs:
         - fedora.yaml
         - archlinux.yaml
         - opensuse.yaml
+        - almalinux-9.yaml
         - docker.yaml
         - ../hack/test-templates/alpine-iso-9p-writable.yaml  # Covers alpine-iso.yaml
         - ../hack/test-templates/net-user-v2.yaml


.github/workflows/test.yml
```diff
@@ -308,6 +308,7 @@ jobs:
         - fedora.yaml
         - archlinux.yaml
         - opensuse.yaml
+        - almalinux-9.yaml
```
AkihiroSuda, line 311: We only test Tier1 and Tier2 templates on the CI
https://lima-vm.io/docs/templates/

AkihiroSuda: If we are going to add almalinux-9, we have to consider:
- why not add almalinux-10
- why prefer alma over rocky

whoschek: Almalinux is a major mainstream distro, and a lot of folks are using it. Adding almalinux-10 and rocky too would be even better to gain confidence, sure :-)

AkihiroSuda: Due the flakiness of the CI, it is not realistic to cover both alma and rocky, and their multiple versions.

So we have to come up a reason why we should prioritize almalinux-9 over other ones.

whoschek: Personally, I'd prefer almalinux-10. The only reason I didn't include it in the PR is because it actually fails CI systematically (due to a real bug, not a flakiness issue).

AkihiroSuda: > bug

Could you open an issue ?

whoschek: Ok, I've opened an issue here: https://github.com/lima-vm/lima/issues/5368

PR #5358, opened 2026-07-28
cidata: suppress the boot-done signal when a reboot is pending

jandubois:
Two boot scripts reboot the guest partway through boot, the Alpine one after removing the image's built-in user and `00-reboot-if-required.sh` after a package upgrade. Both reboots are asynchronous, so `boot.sh` runs on to write `/run/lima-boot-done` and the host agent calls the instance ready while `sshd` is already stopping. A `limactl shell` issued right then gets its connection reset, which on July 18 killed `hack/test-templates.sh` on master under set -e.

The host agent retries the `boot-done` requirement for ten minutes, so waiting the reboot out costs nothing when the guest comes back.

Assisted-by: Claude Opus 5

<details>
<summary>The one recorded failure</summary>

Collected by a CI failure sweep over `lima-vm/lima`. Each entry is one failed job. All in `Integration tests (QEMU, Linux host) (alpine.yaml)`.

- [2026-07-18](https://github.com/lima-vm/lima/actions/runs/29664012302/job/88131299873) on `master`

</details>

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.sh
@@ -21,6 +21,12 @@ if [ "${UNAME}" != "Linux" ]; then
 fi
 rm -f "${RUN}/lima-boot-done"
 
+# A boot script that reboots the guest creates this file first, so that the rest
+# of this script knows the guest is on its way down.
+LIMA_REBOOT_REQUIRED="${RUN}/lima-reboot-required"
+export LIMA_REBOOT_REQUIRED
+rm -f "${LIMA_REBOOT_REQUIRED}"
+
 # shellcheck disable=SC2163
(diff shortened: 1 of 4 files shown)


pkg/cidata/cidata.TEMPLATE.d/boot.sh
```diff
@@ -21,6 +21,12 @@ if [ "${UNAME}" != "Linux" ]; then
 fi
 rm -f "${RUN}/lima-boot-done"
 
+# A boot script that reboots the guest creates this file first, so that the rest
+# of this script knows the guest is on its way down.
+LIMA_REBOOT_REQUIRED="${RUN}/lima-reboot-required"
```
AkihiroSuda, line 26: Needs documentation
https://lima-vm.io/docs/dev/internals/#guest-runtime-files

AkihiroSuda (approved):
Thanks

PR #5359, opened 2026-07-28
dependabot: group the codeql-action bumps into one PR

jandubois:
Dependabot treats each codeql-action sub-path as its own dependency, so one upstream release opens three PRs. v4.37.2 arrived as #5351, #5352 and #5353.


AkihiroSuda (approved):
Thanks

PR #5365, opened 2026-07-29
quote guest error string in time sync log

Nashit-h:
syncTimeOnce logs the guest agent's TimeSyncResponse.Error with %s, but that field is a proto string the guest sets, so a compromised guest can return CR/LF and ANSI escape bytes that forge log lines or inject terminal escapes into the host agent's log stream. This runs automatically every 10s. Log it with %#q, matching how the rest of hostagent already renders guest-supplied strings (e.g. the "received error from the guest" line).


AkihiroSuda (approved):
Thanks

PR #5366, opened 2026-07-29
match static hosts case-insensitively in lookupCnameToHost

SABITHSAHEB:
1. `NewHandler` keys `hostToIP` and `cnameToHost` by `dns.CanonicalName`, so both maps are lowercased and fully qualified.
2. `handleQuery` matches against them with the raw `q.Name`, which is never folded. Domain names are case-insensitive (RFC 4343) and stub resolvers vary query-name case deliberately (DNS 0x20), so a guest asking for `HOST.LIMA.INTERNAL` misses the map, falls through to `handleDefault`, and gets whatever the system resolver or upstream returns instead of the configured address. Same for anything an operator pins through `hostResolver.hosts`.
3. Folded the name in `lookupCnameToHost`, the one entry point the A/AAAA and CNAME paths share. The answer still echoes the owner name as it was queried.

`go test ./pkg/hostagent/dns/` covers it. The new subtest fails before the change, with the mixed-case query answered by the upstream resolver rather than the static entry.

### DIFF
--- a/pkg/hostagent/dns/dns.go
+++ b/pkg/hostagent/dns/dns.go
@@ -212,7 +212,7 @@ func (h *Handler) handleQuery(ctx context.Context, w dns.ResponseWriter, req *dn
 					continue
 				}
 			}
-			if cname != "" && cname != q.Name {
+			if cname != "" && cname != dns.CanonicalName(q.Name) {
 				hdr.Rrtype = dns.TypeCNAME
 				a := &dns.CNAME{
 					Hdr:    hdr,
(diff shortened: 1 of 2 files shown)


pkg/hostagent/dns/dns.go
jandubois: With this patch `cname` is now the canonical name, but `q.Name` is the original query name.

```suggestion
			if cname != "" && cname != dns.CanonicalName(q.Name) {
```

jandubois (changes requested):
Thanks, LGTM, but an explicit `CNAME` query with a mixed-case hostname will still return the wrong value.

pkg/hostagent/dns/dns.go
SABITHSAHEB: Good catch, the folded return value made the q.Name comparison trip on case alone. Applied your suggestion; an explicit CNAME query for Host.Lima.Internal now falls through to upstream like the lowercase form instead of getting a synthesized CNAME answer, and go test ./pkg/hostagent/dns/ is green.

jandubois (approved):
Thanks, LGTM

Issue #4810, opened 2026-04-07
Reverse SSHFS mount fails with "Permission denied" due to Path Conversion conflicts


unsuman:
### Description

When running Lima on Windows using the QEMU driver, the automatic mounting of the Windows home directory (C:\Users\user) to the guest VM (at /c/Users/user) consistently fails. This appears to be a global issue affecting both locally and CI pipelines.

```
level=info msg="[hostagent] Mounting \"C:\\Users\\anshu\" on \"/c/Users/anshu\""
level=info msg="[hostagent] Accepting \"/c/Users/anshu\" Unix path, assuming Cygwin/msys2 OpenSSH"
level=info msg="[hostagent] chdir to \"/c/Users/anshu\" failed: No such file or directory\r"
level=info msg="[hostagent] fusermount3: mount failed: Permission denied"
level=warning msg="[hostagent] failed to confirm whether /c/Users/anshu [remote] is successfully mounted"
```

https://github.com/lima-vm/lima/actions/runs/23485110542/job/69213743941?pr=4772#step:7:114

liketosweep:
Hello @unsuman,

I've reproduced this behavior on my Windows host using the `QEMU` driver. I'm seeing the same path conversion failure and `"Permission denied" ` error during the mounting process.

Since I use a Windows/WSL environment for my daily development, I’d like to take this on and investigate the `hostagent` logic to fix how these paths are being handled. May I be assigned to this issue?
/assign

InsidiousLion:
Hi @unsuman and @liketosweep,

I have found the exact root cause for the Permission denied error and verified a workaround. The issue is a path translation mismatch caused when Lima falls back to native Windows OpenSSH.

I have been doing a deep dive into the Windows architecture and I spent today reproducing this exact reverse-sshfs bug on a native Windows host. I wanted to share my debug logs and root cause analysis.

The Root Cause: Path Translation & Native Windows Binaries
The Permission denied error is a symptom of a path translation mismatch between Lima's hostagent and the host's SSH tools.

When observing the --debug logs during the reverse-sshfs provisioning phase, I noticed Lima was attempting to use the native Windows OpenSSH executable:
C:\WINDOWS\System32\OpenSSH\ssh.exe

However, the hostagent constructs the mount command and the IdentityFile arguments using Unix-style paths (e.g., IdentityFile='/c/Users/Abinandhan/.lima/_config/user' and sshfs ':/c/Users/Abinandhan').

[...]

PR #5375, opened 2026-07-31
cidata: allow apparmor mounts on every Lima mount point

jandubois:
Ubuntu 25.04 added an apparmor profile for fusermount3 permitting fuse mounts only under home, `/mnt`, `/tmp` and `/media`. A mount point mirrors the host path, which on macOS and Windows falls outside those, so reverse-sshfs mounts fail with `fusermount3: mount failed: Permission denied`. The #4968 override covered `@{HOME}` alone, so it only helped Linux hosts.

Derive the rules from the cidata mount points instead. Lima rewrites the file on every boot, so a rule added to it by hand is lost.

The Windows `mount-home` check has been disabled since that job was added, so it never caught this failure. Re-enable it.

Related: #4810, #4908.

Assisted-by: Claude Opus 5


AkihiroSuda:
Windows CI failed 4 times
https://github.com/lima-vm/lima/actions/runs/30606575122/job/91402493778?pr=5375

jandubois:
> Windows CI failed 4 times

Expected:

> The Windows check needs https://github.com/lima-vm/lima/pull/5300 to pass, so this merges after it.

You need to rebase once #5300 is merged (or close and reopen). #5343 exists to show that all tests pass (including the new ones) once the PRs have been merged in sequence.

There is also a merge conflict between #5299 and #5337 that needs to be resolved manually when #5337 is to be merged.

These are the consequences of splitting the big PR into smaller chunks. I've tried to make things free-standing and non-overlapping as good as I can. The alternative would have been to push all branches to the upstream repo and stack them on top of each other. GitHub doesn't yet support stacked PRs from a fork (https://github.com/github/gh-stack/issues/46#issuecomment-4268914593).

jandubois:
Rebased on latest `master` and all green now. PTAL @AkihiroSuda

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
```diff
@@ -19,14 +19,34 @@ update_fuse_conf() {
 
 	# Some distribution (since Ubuntu-25.04) has an apparmor rule for fusermount3. It causes SSHFS mount failed.
 	# Related Issue: https://github.com/lima-vm/lima/issues/4908
-	# Therefore, define a custom rule to loosen the apparmor rule.
```
AkihiroSuda, line 22: any reason to remove this line?

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
```diff
@@ -19,14 +19,34 @@ update_fuse_conf() {
...
-EOF
-		apparmor_parser -r /etc/apparmor.d/fusermount3
+	# The profile only allows mounting under home, /mnt, /tmp and /media, but a mount point mirrors
+	# the host path, so it can be anywhere: /Users/USER from a macOS host, /c/Users/USER from a
+	# Windows one.
+	local_conf="/etc/apparmor.d/local/fusermount3"
+	generated="# Generated by Lima. Delete this line to keep your own rules."
+	# Lima wrote this header before it generated a rule per mount point, so an instance created by
+	# an older Lima is regenerated rather than left with the home-only rule.
+	legacy="# The following two lines allow VM to be mounted to / unmounted from home directly."
```
AkihiroSuda: Can we just overwrite the file without checking the old header?

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
jandubois: Because it belongs to the code block below that got deleted.

It is replaced by the comment below, explaining what the profile allows and why that is insufficient.

Maybe it should include a line explaining that we are generating a rule by mount point now, but that is also already implied by line 27.

pkg/cidata/cidata.TEMPLATE.d/boot.Linux/35-setup-packages.sh
jandubois: We could, but you previously asked for it not to be overwritten: https://github.com/lima-vm/lima/pull/4968#discussion_r3231913021

Which makes sense, as this is the only place a user could add their own rules.

jandubois: Hmm, the code doesn't actually do what I think it should be doing anyways.

So should we always overwrite it and just not support users editing it (because we will overwrite on every boot)?

jandubois:
I've updated the comment and simplified the code to overwrite the file on every boot.

AkihiroSuda (approved):

PR #5376, opened 2026-07-31
validate paths.varRun against an allowlist before findBaseDirectory trims it

SABITHSAHEB:
1. findBaseDirectory() trims paths.varRun down to the deepest component that already exists, and validatePath() only ever sees the trimmed value, so whatever has not been created yet is never checked.
2. the untrimmed value is still what MkdirCmd/PIDFile/Sock interpolate into the sudoers file, so a newline in that tail becomes a rule of its own.
3. the check works off an allowlist rather than a denylist: every component of varRun has to pass identifiers.Validate, the same rule 5d96d98 applied to the network name/mode/interface/group. Rejecting only whitespace still let `/private/var/run/lima,ALL` through, since a comma ends the command list entry in sudoers and `ALL` then stands as an unrestricted command of its own.
4. only varRun is checked this way. socketVMNet and sudoers keep the whitespace check they have on master, so existing layouts like `/Users/foo/.local/bin/socket_vmnet` keep validating.

With `varRun: "/private/var/run/lima\n%staff ALL=(root:wheel) NOPASSWD:NOSETENV: ALL"` in networks.yaml, Validate() returns nil on master and `limactl sudoers` emits that second line verbatim, so installing the file grants group staff passwordless root. paths.varRun was the one value left reaching those commands without a full check.

To test: `go test ./pkg/networks/...`. TestValidateRejectsInjectableVarRun covers the newline, space and comma cases and fails on master. TestValidateVarRun also pins down what stays accepted (trailing slash, dotted component) so the tighter rule does not break real configs.

### DIFF
--- a/pkg/networks/validate.go
+++ b/pkg/networks/validate.go
@@ -61,6 +61,13 @@ func (c *Config) Validate() error {
 		pathsMap[name] = path
 		// varPath will be created securely, but any existing parent directories must already be secure
 		if name == "varRun" {
+			// findBaseDirectory drops the trailing components that don't exist yet,
+			// so the whole value has to be checked before it is trimmed: it is
+			// interpolated into the sudoers file and into the sudo command line
+			// long before the directory is created.
+			if err := validateVarRunString(path); err != nil {
+				return fmt.Errorf("networks.yaml field `paths.%s` error: %w", name, err)
+			}
 			path = findBaseDirectory(path)
 		}
 		err := validatePath(path, name == "varRun")
@@ -92,6 +99,39 @@ func findBaseDirectory(path string) string {
 		}
 	}
 	return path
+}
+
+// validateVarRunString validates the parts of paths.varRun that don't depend on
+// the filesystem, so it can be applied to the whole value before findBaseDirectory
+// trims the components that don't exist yet. varRun is interpolated verbatim into
+// the sudoers file, which has no quoting: whitespace ends the command token, a
+// newline starts a directive of its own, and a comma ends the command list entry.
+// reconcile.go also splits the command line on " " before handing it to sudo.
+// Rather than enumerating those metacharacters, require every component to be an
+// identifier, the same way the network name, mode, interface and group are.
+//
+// This is deliberately stricter than validatePath: it only applies to varRun,
+// so existing socketVMNet/sudoers layouts with dotted components keep validating.
+//
+// The raw value is what reaches the sudoers file, so it is checked as-is rather
+// than after filepath.Clean, which would collapse a ".." component.
+func validateVarRunString(varRun string) error {
+	if varRun == "" {
+		return nil
+	}
+	if !filepath.IsAbs(varRun) {
+		return fmt.Errorf("path %#q is not an absolute path", varRun)
+	}
+	for component := range strings.SplitSeq(varRun, "/") {
+		if component == "" {
+			// leading, trailing and repeated separators
+			continue
+		}
+		if err := identifiers.Validate(component); err != nil {
+			return fmt.Errorf("path %#q has an invalid component: %w", varRun, err)
+		}
+	}
+	return nil
 }
 
 func validatePath(path string, allowDaemonGroupWritable bool) error {
--- a/pkg/networks/validate_unix_test.go
+++ b/pkg/networks/validate_unix_test.go
@@ -1,0 +1,80 @@
+//go:build !windows
+
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package networks
+
+import (
+	"errors"
+	"os"
+	"testing"
+
+	"gotest.tools/v3/assert"
+)
+
+func TestValidateRejectsInjectableVarRun(t *testing.T) {
+	// findBaseDirectory() trims the components that don't exist yet, so the
+	// injected sudoers directive never reached the path check.
+	err := (&Config{Paths: Paths{
+		VarRun: "/private/var/run/lima\n%staff ALL=(root:wheel) NOPASSWD:NOSETENV: ALL",
+	}}).Validate()
+	assert.ErrorContains(t, err, "invalid component")
+
+	// A space in the same position injects an extra argument into the sudo command.
+	err = (&Config{Paths: Paths{
+		VarRun: "/private/var/run/lima --extra-root-flag",
+	}}).Validate()
+	assert.ErrorContains(t, err, "invalid component")
+
+	// A comma ends the command list entry, so "ALL" becomes a command of its own.
+	err = (&Config{Paths: Paths{
+		VarRun: "/private/var/run/lima,ALL",
+	}}).Validate()
+	assert.ErrorContains(t, err, "invalid component")
+}
+
+func TestValidateVarRunString(t *testing.T) {
+	for _, path := range []string{
+		"",
+		"/",
+		"/private/var/run/lima",
+		"/private/var/run/lima/",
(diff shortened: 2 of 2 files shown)


pkg/networks/validate.go
```diff
@@ -94,16 +102,31 @@ func findBaseDirectory(path string) string {
...
 	if path == "" {
 		return nil
 	}
 	if path[0] != '/' {
 		return fmt.Errorf("path %#q is not an absolute path", path)
 	}
-	if strings.ContainsRune(path, ' ') {
+	// A newline adds an arbitrary directive to the generated sudoers file; any
+	// other whitespace injects an extra argument into the socket_vmnet command
+	// that reconcile.go builds by splitting on spaces.
```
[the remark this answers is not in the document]
SABITHSAHEB: Good catch, the split in reconcile.go is on literal spaces only. Reworded it so the sudoers side carries the reason for rejecting all whitespace (interpolated unquoted, so any whitespace ends the command token and a newline starts a directive of its own), with the extra-argv case called out as space-specific. The check itself is unchanged and still rejects all whitespace.

pkg/networks/validate.go
```diff
@@ -94,16 +102,32 @@ func findBaseDirectory(path string) string {
 	return path
 }
 
-func validatePath(path string, allowDaemonGroupWritable bool) error {
+// validatePathString validates the properties of a path that don't depend on the
+// filesystem, so that it can also be applied to a directory that doesn't exist yet.
+func validatePathString(path string) error {
```
unsuman: I think this can still pass `/private/var/run/lima,ALL` and we should check for **what is allowed** rather than **what is not allowed**

SABITHSAHEB: You're right, a comma ends the command list entry in sudoers, so `/private/var/run/lima,ALL` still got through and `ALL` became a command of its own.

Switched it to an allowlist: every path component now has to pass `identifiers.Validate`, the same rule the network name, mode, interface and group already use, so it's checking what's permitted rather than enumerating metacharacters. Added the `,ALL` case and a table test pinning down what stays accepted (`/opt/homebrew/opt/socket_vmnet/bin/socket_vmnet`, `/private/etc/sudoers.d/lima`, trailing slash) so the tighter rule doesn't break real configs.

unsuman: I think we should only be strict about the `varRun` field and keep others (`socketVMNet` and `sudoers`) as they are because some existing configs can break like for example: `/Users/foo/.local/`. Also the PR title says `paths.varRun` but it covers all `paths.*`

SABITHSAHEB: Narrowed it to varRun. socketVMNet and sudoers are back to the whitespace check exactly as it is on master, so `/Users/foo/.local/bin/socket_vmnet` still validates. The diff against master is now only the pre-trim check on varRun, so the title matches the scope again. Also squashed to one commit.

unsuman: Not narrowed yet, `validatePath()` still calls `validatePathString()`, so `socketVMNet: /Users/foo/.local/bin/socket_vmnet` breaks. Also please update the test coverage

SABITHSAHEB: You're right, validatePath() was still calling into it so socketVMNet/sudoers were affected too. Split it: validatePath is back to master's whitespace check, and the identifier rule now lives in validateVarRunString which only runs on varRun before the trim. So socketVMNet: /Users/foo/.local/bin/socket_vmnet validates again. Added TestValidatePathAllowsDottedComponents to pin that down, alongside the existing varRun cases.

unsuman (dismissed):
Thanks, LGTM!

pkg/networks/validate.go
```diff
@@ -94,6 +101,36 @@ func findBaseDirectory(path string) string {
...
+// identifier, the same way the network name, mode, interface and group are.
+//
+// This is deliberately stricter than validatePath: it only applies to varRun,
+// so existing socketVMNet/sudoers layouts with dotted components keep validating.
+func validateVarRunString(path string) error {
+	if path == "" {
+		return nil
+	}
+	if path[0] != '/' {
+		return fmt.Errorf("path %#q is not an absolute path", path)
```
AkihiroSuda: Use https://pkg.go.dev/path/filepath#Abs

pkg/networks/validate.go
```diff
@@ -94,6 +101,36 @@ func findBaseDirectory(path string) string {
...
+	}
+	for component := range strings.SplitSeq(path, "/") {
```
AkihiroSuda: use https://pkg.go.dev/path/filepath#SplitList

pkg/networks/validate.go
SABITHSAHEB: Switched to a stdlib IsAbs check, but via path.IsAbs rather than filepath: the Windows CI runs these unit tests and filepath treats "/private/var/run/lima" as relative there, since varRun is always a slash-separated unix path. filepath.Abs would also run Clean on the value, and validating the cleaned string while the raw one is what gets interpolated into the sudoers file would let a ".." collapse hide an injected component, so the raw string has to be what's checked. Added a comment explaining that.

pkg/networks/validate.go
SABITHSAHEB: filepath.SplitList splits on the list separator (":" on unix, for $PATH-style strings), so the whole value would come back as a single element and identifiers.Validate would then reject every valid path. There's no stdlib helper for splitting a path into its components, so I kept the split on "/", which matches the slash-separated unix paths networks.yaml uses. Happy to change it if you had a different shape in mind.

pkg/networks/validate.go
```diff
@@ -94,6 +102,41 @@ func findBaseDirectory(path string) string {
...
+// newline starts a directive of its own, and a comma ends the command list entry.
+// reconcile.go also splits the command line on " " before handing it to sudo.
+// Rather than enumerating those metacharacters, require every component to be an
+// identifier, the same way the network name, mode, interface and group are.
+//
+// This is deliberately stricter than validatePath: it only applies to varRun,
+// so existing socketVMNet/sudoers layouts with dotted components keep validating.
+//
+// varRun is a slash-separated Unix path, so the check uses the path package
+// rather than filepath: filepath's absolute-path rules differ on Windows, and
```
AkihiroSuda, line 118: Is this function ever called on Windows?

SABITHSAHEB: Not in production, no. Both real callers are darwin-only: `limactl sudoers` is a stub on !darwin, and reconcile's startNetwork/stopNetwork return before validateConfig() when runtime.GOOS != "darwin". The one place it does run on Windows is the unit tests, since CI runs `go test ./...` on the windows runners and pkg/networks has no build tags. With filepath.IsAbs the accept cases ("/private/var/run" etc.) would fail there, which is why the check sticks to the path package; on darwin the two are equivalent.

AkihiroSuda: >  The one place it does run on Windows is the unit tests

The unit test should not run on Windows then

SABITHSAHEB: Done. The tests for this are now in validate_unix_test.go under `//go:build !windows`, and validate_test.go is back to what's on master. With the Windows run out of the picture the check uses filepath.IsAbs. I kept IsAbs rather than Abs because the raw string is what reaches the sudoers file, and Clean would fold a `..` component before it's checked.

pkg/networks/validate.go
SABITHSAHEB: Switched to filepath.IsAbs now that the tests no longer run on Windows.

AkihiroSuda (approved):
Thanks

PR #5384, opened 2026-08-03
portfwd: don't close the live udp listener on a duplicate forward

SABITHSAHEB:
forwardUDP registers its cleanup with `defer p.Remove(...)` as its first statement, before the check for a listener already registered under the same key, so the deduplication path carries that cleanup out with it. The host reaches that path whenever the guest agent reports a UDP port it has already reported, which happens on every guest agent reconnect since the agent re-sends its whole local port list, and which the guest can also trigger on demand by repeating an entry inside one event. Remove closes the PacketConn the first goroutine is still serving through HandleUDPConnection and drops the map entry, so the forward is gone while the rule is still in effect and neither side notices. That first goroutine then unwinds out of proxy.Run and runs its own deferred Remove, which can take out a listener a later re-add has since installed under the same key. I found it while comparing the two forwarders side by side: forwardTCP registers the identical defer only after the duplicate check and a successful bind, and the UDP one is the odd one out. Moving the defer to the same position gives forwardUDP the same ownership rule, and the listen-failure path that no longer calls Remove never held the key anyway. The regression test fails on master and passes with the change.

### DIFF
--- a/pkg/portfwd/listener.go
+++ b/pkg/portfwd/listener.go
@@ -130,7 +130,6 @@ func (p *ClosableListeners) forwardTCP(ctx context.Context, dialContext func(ctx
 
 func (p *ClosableListeners) forwardUDP(ctx context.Context, dialContext func(ctx context.Context, network string, addr string) (net.Conn, error), hostAddress, guestAddress string) {
 	key := key("udp", hostAddress, guestAddress)
-	defer p.Remove(ctx, "udp", hostAddress, guestAddress)
 
 	p.udpListenersRW.Lock()
 	_, ok := p.udpListeners[key]
@@ -145,6 +144,7 @@ func (p *ClosableListeners) forwardUDP(ctx context.Context, dialContext func(ctx
 		p.udpListenersRW.Unlock()
 		return
 	}
+	defer p.Remove(ctx, "udp", hostAddress, guestAddress)
 	p.udpListeners[key] = udpConn
 	p.udpListenersRW.Unlock()
 
--- a/pkg/portfwd/listener_test.go
+++ b/pkg/portfwd/listener_test.go
@@ -1,0 +1,68 @@
+// SPDX-FileCopyrightText: Copyright The Lima Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package portfwd
+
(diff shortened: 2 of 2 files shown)


AkihiroSuda:
How to test? (other than unit tests)

SABITHSAHEB:
You can see it end to end on a stock instance with the default grpc port forwarder (`LIMA_SSH_PORT_FORWARDER` unset). The trigger is the guest agent coming up without its saved event state, which is what a guest agent crash or a guest reboot produces (`/run` is a tmpfs): the fresh agent announces its whole port list again, and that duplicate announce is what tears down the live UDP listener on the host.

1. `limactl start default`, then run a UDP echo server inside the guest:

```console
limactl shell default -- python3 -c '
import socket
s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.bind(("127.0.0.1", 10007))
while True:
    d, a = s.recvfrom(1024)
    s.sendto(d, a)'
```

2. Wait for `Forwarding UDP from 127.0.0.1:10007 to 127.0.0.1:10007` in `~/.lima/default/ha.stderr.log`, then check the forward from the host:

```console
$ echo ping | nc -u -w1 127.0.0.1 10007
ping
```

3. Restart the guest agent without its saved state, simulating a crash or reboot (a plain `systemctl restart` saves the state on SIGTERM, so the file has to go while the agent is stopped):

```console
limactl shell default -- sudo sh -c 'systemctl stop lima-guestagent; rm -f /run/lima-guestagent/event-state.json; systemctl start lima-guestagent'
```

4. Within ~15s the hostagent reconnects and logs the same `Forwarding UDP from 127.0.0.1:10007 ...` line again. Repeat the check from step 2.

On master the echo no longer answers, and `lsof -nP -iUDP:10007` on the host shows the hostagent has dropped the socket even though the log just reported the forward as up. With this patch the duplicate announce is a no-op, lsof still shows the listener, and the echo keeps working.

SABITHSAHEB:
any update?

balajiv113 (approved):
Thanks changes looks good

balajiv113 (approved):

PR #5385, opened 2026-08-03
Allow installing k8s on minimal image variant

afbjorklund:
## What This PR Changes

The kubernetes template was failing, when running on the "minimal" ubuntu images...

The "gnupg" package was not installed by default, causing the package installation to fail.

`407M	ubuntu-26.04-minimal-cloudimg-amd64.img`

`821M	ubuntu-26.04-server-cloudimg-amd64.img`

Closes #

## How I Tested This

`limactl start --yes template:k8s --image-variant minimal`

## AI Usage

No

### DIFF
--- a/templates/k8s.yaml
+++ b/templates/k8s.yaml
@@ -53,11 +53,7 @@ provision:
     # Installing kubeadm, kubelet and kubectl
     export DEBIAN_FRONTEND=noninteractive
     apt-get update
-    {{- if .Param.proxy }}
-    apt-get install -y --no-upgrade ca-certificates curl
-    {{- else}}
-    apt-get install -y apt-transport-https ca-certificates curl
-    {{- end}}
+    apt-get install -y --no-upgrade ca-certificates curl gnupg
     STABILITY="{{.Param.stability}}"
     {{if eq .Param.release "stable" }}
     RELEASE=$(curl -L -s https://dl.k8s.io/release/stable.txt | cut -d'.' -f1-2)


templates/k8s.yaml
```diff
@@ -54,7 +54,7 @@ provision:
     export DEBIAN_FRONTEND=noninteractive
     apt-get update
     {{- if .Param.proxy }}
-    apt-get install -y --no-upgrade ca-certificates curl
+    apt-get install -y --no-upgrade ca-certificates curl gnupg
     {{- else}}
     apt-get install -y apt-transport-https ca-certificates curl
```
AkihiroSuda: Isn't it needed in this case?

afbjorklund: Think it got added by default, with "apt". Since it is in the "recommends". Will test.

EDIT: Nope, I was wrong... And "apt-transport-https" is a no-op in Ubuntu 26.04

`ca-certificates is already the newest version (20260601~26.04.1).`

`curl is already the newest version (8.18.0-1ubuntu2.3).`

afbjorklund: So we could probably remove all of it, and only make sure gnupg is installed...

`apt-get install -y --no-upgrade gnupg` (it will already be installed in standard)

Can leave the others too, in case it changes in a future revision or something.

But "https support has been moved into the apt package in 1.5", so that can go.

AkihiroSuda (approved):
Thanks

unsuman (approved):
Thanks!

PR #5393, opened 2026-08-05
fix(wsl): fix ssh address for mirrored networking

### LINKED ISSUE #5392: WSL2: Mirrored networking fails to forward correct SSH address
### Description

Corporate VPNs can be restrictive about their network policies and often do not have exceptions for the WSL NAT architecture used by traditional WSL2 VMs, leading to the VM setup stage failing when connected to such VPNs.

Setting `networkingMode=mirrored` allows a bypass for this and allows the Lima instance to be created, however it seems that it remains broken after the fact. Any attempts to SSH will time out. This is because per Microsoft [docs](https://learn.microsoft.com/en-us/windows/wsl/networking), SSH should connect to `localhost` or `127.0.0.1` to communicate with the machine, however in our code we check for the output of `hostname -I` within the VM, which doesn't work for mirrored networking.

It should be a simple fix of just returning localhost when mirrored networking is enabled.

sondavidb:
## What This PR Changes

This PR returns localhost for mirrored networking mode in WSL2. The existing behavior of checking the hostname from within the VM doesn't work for mirrored networking per the Microsoft [docs](https://learn.microsoft.com/en-us/windows/wsl/networking), and the docs recommend this approach instead

Closes #5392 

## How I Tested This

I tested this with the workflow in runfinch/finch, where running a container would previously time out. With this change, the container runs as expected.

### DIFF
--- a/pkg/driver/wsl2/vm_windows.go
+++ b/pkg/driver/wsl2/vm_windows.go
@@ -257,6 +257,8 @@ func getWslStatus(ctx context.Context, instName string) (string, error) {
 }
 
 // GetSSHAddress runs a hostname command to get the IP from inside of a wsl2 VM.
+// In WSL2 mirrrored mode, this should return 127.0.0.1 per Microsoft docs
+// https://learn.microsoft.com/en-us/windows/wsl/networking
 //
 // Expected output (whitespace preserved, [] for optional):
 // PS > wsl -d <distroName> bash -c hostname -I | cut -d' ' -f1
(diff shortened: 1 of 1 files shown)


afbjorklund:
Sounds good to me. I still don't know how it is supposed to work if you have multiple distros, but I guess we need published ports and proper SSH port for that to work. So that can be handled separately, with the other drivers.

sondavidb:
> I still don't know how it is supposed to work if you have multiple distros

I don't really understand either, if all the instances share the same networking stack though logic leads me to believe that they can't both be on the same port. With multiple distros running with this fix it seems that it keeps on landing my tasks on the correct Lima instance. So not sure what to make of that.

Regardless, based on the official Microsoft documentation this seems correct.

sondavidb:
Hey @afbjorklund wondering if you could have any folks take a look at this?

PR #5399, opened 2026-08-07
fix: Disable LLMNR to prevent Debian SSH login delay

### LINKED ISSUE #5389: LLMNR causes login delay on Debian
### Description

I have noticed an undesired behaviour when using Lima with Debian 13.

# Steps to reproduce

Create a new VM like this: `limactl create --name=myvm template:debian-13`

Start the VM: `limactl start myvm`

Run this command and notice how long it takes to complete:

`ssh -F $HOME/.lima/myvm/ssh.config lima-myvm /usr/bin/true`

Then run almost the same command again, except this time we force pseudo-terminal allocation. Notice how long it takes to complete.

`ssh -t -F $HOME/.lima/myvm/ssh.config lima-myvm /usr/bin/true`

# What I expected to happen

That both commands would finish instantly.

# What actually happens

The first command (no pseudo-terminal) finishes instantly.

The second command, where the pseudo-terminal is allocated, has a quite significant delay before completing. On my system it takes about four seconds to complete.

# Workaround

In the VM, the file `/etc/systemd/resolved.conf.d/00-lima-enable-mdns.conf` is created by Lima. Modify it to look like this:

```
[Resolve]
MulticastDNS=yes
LLMNR=no
```

Now restart the resolver:

`sudo systemctl restart systemd-resolved.service`

When Link-Local Multicast Name Resolution is disabled, there is no login delay.

# Alternative workaround

Add the following lines to `/etc/hosts` in the VM:

```
127.0.0.1 UNKNOWN
::1 UNKNOWN
```

# Why the delay happens

It seems like sshd sets the string `UNKNOWN` as the client hostname. Then it tries to resolve that hostname, an operation which obviously fails after a time-out. I don’t know why the issue only occurs for interactive sessions.

# Suggestion for permanent solution

One possible solution could be to add the `LLMNR=no` directive permanently. I’m not sure what side-effects it might have.

mvanhorn:
## What This PR Changes

Extend the existing `[Resolve]` content in `pkg/cidata/cidata.TEMPLATE.d/boot.Linux/06-enable-mdns-on-systemd.sh` with `LLMNR=no`, while preserving `MulticastDNS=yes` and the current per-link mDNS setup. Keep the existing content-hash comparison and resolver restart behavior so an existing guest receives the corrected drop-in once and subsequent boots remain idempotent. Add a focused assertion to the existing cidata template test that inspects the emitted `boot.Linux/06-enable-mdns-on-systemd.sh` artifact and verifies that both resolver directives are present.

Closes #5389

Closes #

## How I Tested This

Not applicable to this change.

## AI Usage

Not applicable to this change.

AI was used for assistance.

### DIFF
--- a/templates/_images/debian-13.yaml
+++ b/templates/_images/debian-13.yaml
@@ -64,3 +64,6 @@ images:
   variant: "nocloud"
 
 mountTypesUnsupported: [9p]
+
+param:
+  internal_disableLLMNR: "true"
--- a/templates/_images/debian-13.yaml
+++ b/templates/_images/debian-13.yaml
@@ -65,5 +65,10 @@ images:
 
 mountTypesUnsupported: [9p]
 
+# Debian 13's sshd passes an UNKNOWN peer address for vsock connections to Linux
+# audit, which tries to resolve it via LLMNR and delays interactive SSH logins.
+# Remove this workaround when the images include the OpenSSH fix:
+# https://github.com/openssh/openssh-portable/commit/e5055ef26abc
+# See https://github.com/lima-vm/lima/issues/5389
 param:
   internal_disableLLMNR: "true"
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/06-enable-mdns-on-systemd.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/06-enable-mdns-on-systemd.sh
@@ -18,6 +18,9 @@ command -v resolvectl >/dev/null 2>&1 || exit 0
 enable_mdns_conf_path=/etc/systemd/resolved.conf.d/00-lima-enable-mdns.conf
 enable_mdns_conf_content="[Resolve]
 MulticastDNS=yes
+{{if .Param.internal_disableLLMNR -}}
+LLMNR=no
+{{end -}}
 "
 # Create /etc/systemd/resolved.conf.d/00-lima-enable-mdns.conf if its content is different
 if [ "$(echo "${enable_mdns_conf_content}" | sha256sum)" != "$(sha256sum <"${enable_mdns_conf_path}" 2>/dev/null)" ]; then
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/06-enable-mdns-on-systemd.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/06-enable-mdns-on-systemd.sh
@@ -16,12 +16,12 @@ command -v resolvectl >/dev/null 2>&1 || exit 0
 
 # Configure systemd-resolved to enable mDNS resolution globally
 enable_mdns_conf_path=/etc/systemd/resolved.conf.d/00-lima-enable-mdns.conf
-enable_mdns_conf_content="[Resolve]
+enable_mdns_conf_content='[Resolve]
 MulticastDNS=yes
-{{if .Param.internal_disableLLMNR -}}
+{{if eq .Param.internal_disableLLMNR "true" -}}
 LLMNR=no
 {{end -}}
-"
+'
 # Create /etc/systemd/resolved.conf.d/00-lima-enable-mdns.conf if its content is different
 if [ "$(echo "${enable_mdns_conf_content}" | sha256sum)" != "$(sha256sum <"${enable_mdns_conf_path}" 2>/dev/null)" ]; then
 	mkdir -p "$(dirname "${enable_mdns_conf_path}")"
--- a/pkg/cidata/template_test.go
+++ b/pkg/cidata/template_test.go
@@ -125,16 +125,33 @@ func TestTemplate(t *testing.T) {
 	}
 	layout, err := ExecuteTemplateCIDataISO(args)
 	assert.NilError(t, err)
+	var resolverBootScript string
 	for _, f := range layout {
 		t.Logf("=== %#q ===", f.Path)
 		b, err := io.ReadAll(f.Reader)
 		assert.NilError(t, err)
 		t.Log(string(b))
+		if f.Path == "boot.Linux/06-enable-mdns-on-systemd.sh" {
+			resolverBootScript = string(b)
+		}
 		if f.Path == "user-data" {
 			// mounted later
 			assert.Assert(t, !strings.Contains(string(b), "mounts:"))
 			// ca_certs:
 			assert.Assert(t, !strings.Contains(string(b), "trusted:"))
+		}
+	}
+	assert.Assert(t, strings.Contains(resolverBootScript, "MulticastDNS=yes"))
+	assert.Assert(t, !strings.Contains(resolverBootScript, "LLMNR=no"))
+
+	args.Param = map[string]string{"internal_disableLLMNR": "true"}
+	layout, err = ExecuteTemplateCIDataISO(args)
+	assert.NilError(t, err)
+	for _, f := range layout {
+		if f.Path == "boot.Linux/06-enable-mdns-on-systemd.sh" {
+			b, err := io.ReadAll(f.Reader)
+			assert.NilError(t, err)
+			assert.Assert(t, strings.Contains(string(b), "LLMNR=no"))
 		}
 	}
 }
--- a/pkg/cidata/template_test.go
+++ b/pkg/cidata/template_test.go
@@ -144,15 +144,22 @@ func TestTemplate(t *testing.T) {
 	assert.Assert(t, strings.Contains(resolverBootScript, "MulticastDNS=yes"))
 	assert.Assert(t, !strings.Contains(resolverBootScript, "LLMNR=no"))
 
-	args.Param = map[string]string{"internal_disableLLMNR": "true"}
-	layout, err = ExecuteTemplateCIDataISO(args)
-	assert.NilError(t, err)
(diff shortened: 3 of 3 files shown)


AkihiroSuda (changes requested):
Was the cause analyzed?
Why does it happen specifically with Debian?

AkihiroSuda:
> How I Tested This
> Not applicable to this change.

Definitely applicable. How does this affect existing mDNS setup?

cpach:
I have added some more info about the bug in #5389.

mvanhorn:
Yes, and the credit is cpach's - the analysis in #5389 (comment of 2026-08-07T16:56) answers both questions.

Cause: over AF_VSOCK sshd has no peer hostname to canonicalize, so it records the client as the literal string `"UNKNOWN"` (canohost.c). On Linux, `linux_audit_record_event` passes that through to `audit_log_acct_message`, which calls `getaddrinfo("UNKNOWN")`. That is a single-label name, so systemd-resolved attempts LLMNR resolution, and the ~4s login delay is that lookup timing out. It only shows with `-t` because the audit path runs on PTY allocation.

Why Debian specifically: Debian 13 ships an openssh predating the upstream fix e5055ef26abc ("Don't log audit messages with UNKNOWN hostname"). cpach confirmed a current Debian sid image does not reproduce it.

One thing in cpach's own distro table argues against this PR's approach, and I would rather raise it than let it slide: fedora-44 runs `LLMNR=resolve`, not off, and is listed unaffected. If the LLMNR setting were the discriminator, `resolve` should still hit the same timeout. That points at the sshd version as the real variable and makes "disable LLMNR in every guest" look like a broad guest-side change for a bug that is already fixed upstream and will arrive via Debian.

So I do not want to push this as-is on my own judgment. cpach also asked for your call on whether disabling LLMNR is desirable at all. Options as I see them:

1. Close this and wait for Debian to backport the sshd fix.
2. Scope the drop-in to Debian guests only, leaving other distros untouched.
3. Keep it global as written.

Happy to rework it into 2, or close it, whichever you prefer. No strong opinion from me beyond wanting to avoid a global default change for a distro-specific regression.

afbjorklund:
You can also add an "internal_" param only for the Debian 13 image template, like for the Ubuntu 26.04 workaround

```
param:
  internal_disableLLMNR: "true"
```

mvanhorn:
Done as suggested: `LLMNR=no` now sits behind an `internal_disableLLMNR` param, and only the Debian 13 image template sets it, same shape as the Ubuntu 26.04 workaround. Template test covers both the default (LLMNR untouched) and the Debian path.

unsuman:
@mvanhorn Please sign the DCO and squash commits

mvanhorn:
Both were already in as of ca3e6e6, sorry for not saying so: the branch is a single commit and it carries the Signed-off-by trailer. The DCO check is green on the current head.

AkihiroSuda:
Let's see if there is a chance to fix the issue on Debian's side
- https://github.com/lima-vm/lima/issues/5389#issuecomment-5474146549

mvanhorn:
Sounds right, and the Debian-side fix is the one that actually removes the cause.

Two things that might help move it along:

- cpach said in August he might file a Debian bug but hadn't yet, and your ping today is still unanswered. I'm happy to file it against Debian's openssh myself, requesting a Trixie backport of openssh e5055ef26abc ("Don't log audit messages with UNKNOWN hostname"), and link it here so it isn't blocked on one person.
- For what it's worth, this PR is no longer the global change you pushed back on. Since ca3e6e6 the `LLMNR=no` drop-in sits behind `internal_disableLLMNR`, and only the Debian 13 template sets it, the same shape as the Ubuntu 26.04 workaround afbjorklund pointed to.

So it can sit as a stopgap that stops applying once Trixie has the fix, or I'm equally happy to close it and let the Debian backport be the whole answer. Your call, just say which and I'll do it.

cpach:
@mvanhorn @AkihiroSuda I’ll try my best to reach out to the Debian project within 24 hours.

cpach:
@mvanhorn @AkihiroSuda I have now sent an [e-mail to the debian-ssh mailing list](https://lists.debian.org/debian-ssh/2026/08/msg00026.html), awaiting their response.

cpach:
Current status: No reply from Debian yet.

cpach:
I have now received a [reply from the Debian package maintainer](https://lists.debian.org/debian-ssh/2026/09/msg00004.html), and it seems like there’s a good chance he’ll backport the upstream patch into trixie.

ETA: Currently unknown.

mvanhorn:
> Was the cause analyzed?

Yes, by @cpach in #5389 -- [the full write-up is here](https://github.com/lima-vm/lima/issues/5389#issuecomment-5219820915), posted a few hours after you asked, so it never made it onto this PR. The chain:

The SSH connection arrives over `AF_VSOCK`. sshd has no hostname for that kind of peer, so `canohost.c` sets the client hostname to the literal string `"UNKNOWN"`. On Linux, sshd records the session via `linux_audit_record_event`, which calls `audit_log_acct_message` in audit-userspace. The audit library cannot tell that `"UNKNOWN"` is not a real hostname, so it calls `getaddrinfo` on it. Because it is a single-label name, systemd-resolved attempts LLMNR -- and that is where the multi-second timeout comes from. It only shows up with `ssh -t` because the audit event is recorded on PTY allocation.

OpenSSH fixed this upstream in [e5055ef26abc](https://github.com/openssh/openssh-portable/commit/e5055ef26abc) ("Don't log audit messages with UNKNOWN hostname", Sept 2025) by passing `NULL` instead of `"UNKNOWN"`.

> Why does it happen specifically with Debian?

Because Debian 13 ships an sshd predating that fix. @cpach checked the other images: `almalinux-10` is unaffected, `opensuse-leap-16` does not use systemd-resolved, `ubuntu-26.04` already has LLMNR off by default, and a fresh Debian sid image -- which has the OpenSSH fix -- does not reproduce it. So it is not really Debian-specific so much as specific to images whose sshd predates the fix *and* which run systemd-resolved with LLMNR enabled; Debian 13 is the one in-tree template that lands in that intersection.

That is why the param is set only in `templates/_images/debian-13.yaml` rather than globally, and I have added a comment there recording the cause and the upstream commit, so the workaround has an explicit removal condition once Trixie picks up the fix.

I also fixed a bug in the template while here: `{{if .Param.internal_disableLLMNR}}` is true for *any* non-empty string, so setting the param to `"false"` still emitted `LLMNR=no`. It now compares against `"true"`, and `template_test.go` covers `""`, `"false"` and `"true"` -- reverting the template alone fails the `false` subtest.

One open question is still yours rather than mine: @cpach explicitly asked whether disabling LLMNR is the remedy you want at all, versus waiting for the Debian backport or handling it another way. Happy to rework this if you would prefer a different approach.

AkihiroSuda:
I think we can merge this PR and revert it later when Debian cherry-picks the relevant commit, but please squash the commits and fix the DCO error at least

templates/_images/debian-13.yaml
```diff
@@ -64,3 +64,11 @@ images:
   variant: "nocloud"
 
 mountTypesUnsupported: [9p]
+
+# Debian 13's sshd passes an UNKNOWN peer address for vsock connections to Linux
+# audit, which tries to resolve it via LLMNR and delays interactive SSH logins.
+# Remove this workaround when the images include the OpenSSH fix:
+# https://github.com/openssh/openssh-portable/commit/e5055ef26abc
+# See https://github.com/lima-vm/lima/issues/5389
```
AkihiroSuda, line 72: Could you also add the link to the Debian bug tracker

mvanhorn:
Squashed to one commit with the signoff, so DCO should be green now. Tree is unchanged, `go test ./pkg/cidata/...` passes on 5b0208c.

On the Debian bug tracker link: I could not find one. I went back through #5389 and the analysis there is cpach's, tracing it to the older sshd in Debian 13 passing `UNKNOWN` to Linux audit. You suggested backporting the fix to Debian, but as far as I can tell nobody has actually filed anything in the Debian BTS, so there is no link to add. The template currently points at the upstream OpenSSH commit and at #5389.

Want me to file the Debian bug and then add the link, or would you rather leave the template pointing at the two links it has?

cpach:
> On the Debian bug tracker link: I could not find one.

It’s here: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1147100

mvanhorn:
Thanks @cpach, that is the one I could not find. Added in 9211335.

@AkihiroSuda the comment block in `templates/_images/debian-13.yaml` now carries the Debian bug report alongside the upstream OpenSSH commit, so both the fix and the downstream tracking are recorded next to the workaround:

```yaml
# Remove this workaround when the images include the OpenSSH fix:
# https://github.com/openssh/openssh-portable/commit/e5055ef26abc
# See https://github.com/lima-vm/lima/issues/5389
# Debian bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1147100
```

Amended into the existing commit rather than adding a new one, so the PR stays at one commit with the signoff. `go test ./pkg/cidata/...` passes on 9211335.

AkihiroSuda:
CI failing
https://github.com/lima-vm/lima/actions/runs/35633268831/job/106624925502?pr=5399

PR #5400, opened 2026-08-08
sshutil: only prioritize ciphers ssh actually supports

### LINKED ISSUE #2913: ssh in lima seems to exclude aes-ctr and force aes-gcm exclusively
### Description

My problem here is that we do not want to carry openssl as a dependency in other parts of our systems, and our ssh client is compiled `--without-openssl`.

So:
```
$ ssh -Q cipher
aes128-ctr
aes192-ctr
aes256-ctr
chacha20-poly1305@openssh.com
```

Which will not work with lima apparently:
```
$ limactl shell foo
command-line line 0: Unsupported option "gssapiauthentication"
command-line line 0: Bad SSH2 cipher spec '^aes128-gcm@openssh.com,aes256-gcm@openssh.com'.
```

From a casual reading of:
https://github.com/lima-vm/lima/blob/217da28fb2bf8e02e538a397b7977de463cc3568/pkg/sshutil/sshutil.go#L201-L222

It seems to me lima is forcing gcm (when there is acceleration for aes), with a fallback on chacha - so, pretty much, forcing ciphers that also provide integrity, excluding ctr+separate mac.

I appreciate strong opinions :-) - and clearly there is nothing wrong with aes-gcm (though maybe chacha is more contentious) - so, if this was a conscious decision to make lima work exclusively for these ciphers, that is fine.

On the other hand, if this was not a concerted decision - there is nothing wrong with aes-ctr + hmac-sha2 either - so, curious about folks' opinion on this?

Thanks in advance.

pujitha24:
Motivation:
CommonOpts unconditionally emits a `Ciphers=^...` ssh option that
prioritizes aes128-gcm/aes256-gcm@openssh.com (or
chacha20-poly1305@openssh.com without AES acceleration). ssh clients
built `--without-openssl` (e.g. some minimal/embedded distributions)
do not implement any of those AEAD ciphers at all, only ctr/cbc
variants, so ssh rejects the forced cipher spec outright:

  command-line line 0: Bad SSH2 cipher spec
  '^aes128-gcm@openssh.com,aes256-gcm@openssh.com'.

This makes `limactl shell` fail unconditionally for anyone using such
an ssh build. A maintainer confirmed on the report that Lima should
query `ssh -Q cipher` to validate available algorithms before forcing
a cipher preference.

Approach:
- Add detectSupportedCiphers/parseSupportedCiphers, which run
  `ssh -Q cipher` once per resolved ssh executable (cached alongside
  the existing version/GSSAPI detection) and parse the supported
  cipher names into a set. A nil result (query failed, or the client
  is too old to support "-Q") means "unknown", not "unsupported".
- Add filterSupportedCiphers, a small pure helper used by CommonOpts
  to keep only the preferred ciphers that are actually supported,
  preserving order. When SupportedCiphers is nil it returns the
  preferred list unchanged, preserving the historical behavior for
  ssh clients this query cannot inspect.
- When none of the preferred ciphers are supported, CommonOpts now
  omits the Ciphers= option entirely instead of emitting one ssh will
  reject, letting ssh fall back to its own default cipher negotiation
  (which includes aes-ctr etc. on openssl-less builds).
- When ssh does support the preferred ciphers, the emitted Ciphers=
  string is unchanged from before (verified by inspection: the new
  windows/non-windows formatting produces byte-identical output to
  the old hardcoded literals).

Validation:
- go build ./...
- go test ./pkg/sshutil/...  (includes new TestParseSupportedCiphers
  and TestFilterSupportedCiphers, which directly cover the
  --without-openssl scenario: a ctr-only supported-cipher set filters
  out every preferred GCM cipher, matching the reported failure mode)
- gofmt -l and go vet ./pkg/sshutil/... clean
I did not build or run against an actual ssh binary compiled
--without-openssl, and did not run the BATS/VM integration suite;
this change is validated at the unit level against the exact cipher
sets described in the report, not via a live end-to-end reproduction.

Report: https://github.com/lima-vm/lima/issues/2913
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #2913

### DIFF
--- a/pkg/sshutil/sshutil.go
+++ b/pkg/sshutil/sshutil.go
@@ -373,6 +373,12 @@ type openSSHInfo struct {
 
 	// Some distributions omit this feature by default, for example, Alpine, NixOS.
 	GSSAPISupported bool
+
+	// SupportedCiphers is the set of cipher names reported by `ssh -Q cipher`.
+	// It is empty when the query could not be run or produced no output
+	// (e.g. very old ssh clients that do not support "-Q"), in which case
+	// callers should not filter ciphers based on it.
+	SupportedCiphers map[string]bool
 }
 
 var sshInfo struct {
@@ -380,9 +386,6 @@ var sshInfo struct {
 	// aesAccelerated is set to true when AES acceleration is available.
 	// Available on almost all modern Intel/AMD processors.
 	aesAccelerated bool
-
-	// OpenSSH executable information for the version and supported options.
-	openSSH openSSHInfo
 }
 
 // CommonOpts returns ssh option key-value pairs like {"IdentityFile=/path/to/id_foo"}.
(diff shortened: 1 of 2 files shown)


unsuman:
@pujitha24 Please squash the commits

unsuman (approved):
Thanks, LGTM!

afbjorklund (approved):

AkihiroSuda:
Closing per https://github.com/lima-vm/lima/pull/5409#issuecomment-5328485853

You, or anybody else, may still resubmit the PR with same content after manual verification.

pujitha24:
Understood, that's a fair call and I should have led with it rather than after the fact. I don't currently have a way to boot a real Lima VM to validate changes like this one, so I'll hold off on submitting PRs I can't actually test going forward. Thanks for taking the time to explain instead of just closing it outright.

Issue #5329, opened 2026-07-24
gRPC port forwarder does not propagate TCP half-close (FIN)


elprans:
The default (gRPC) port forwarder does not properly implement TCP half-close
semantics. A client that sends its request and then calls `shutdown(SHUT_WR)`
-- sending a FIN while keeping the read side open to receive the response --
never gets its response: the FIN reaches the guest, but the forwarder treats
the half-close as a full close and immediately tears down the tunnel and the
guest-side TCP connection, so the server's response is lost and the client sees
a premature EOF instead of its response.

"FIN marks end-of-request, response follows, server closes" is a standard TCP
idiom (used by the MySQL protocol's close sequence, FastCGI, and various RPC
framings). Such protocols deadlock or fail through the gRPC forwarder, while
working correctly through `LIMA_SSH_PORT_FORWARDER=true` (the SSH
`direct-tcpip` channel protocol has an explicit `CHANNEL_EOF` message distinct
from `CHANNEL_CLOSE`, so half-closes round-trip correctly) and when tested
inside the guest directly.

I believe this is the root cause of (at least part of) the long tail of "gRPC
forwarding hangs" reports: #2968, #2859, #3355, #3944 (closed asking for a
minimal reproducer -- this issue provides one), and abiosoft/colima#1376.

## Half-close is standard TCP behavior

Normative references:

- [RFC 9293, section 3.6.1 "Half-Closed Connections"](https://www.rfc-editor.org/rfc/rfc9293.html#section-3.6.1):
  "Since the two directions of a TCP connection are closed
  independently, it is possible for a connection to be "half closed",
  i.e., closed in only one direction, and a host is permitted to
  continue sending data in the open direction on a half-closed
  connection."
- [RFC 9293, section 3.10.4 "CLOSE Call"](https://www.rfc-editor.org/rfc/rfc9293.html#section-3.10.4): 
  CLOSE means "I have no more to send" but does not mean 
  "I will not receive any more." This is precisely the semantics 
  the forwarder violates.
- [RFC 4254, section 5.3 "Closing a Channel"](https://www.rfc-editor.org/rfc/rfc4254.html#section-5.3):
  `SSH_MSG_CHANNEL_EOF` -- "the channel remains open after this
  message, and more data may still be sent in the other direction" --
  which is why the SSH forwarder relays half-closes correctly.

Real-world TCP relay implementations that propagate half-close:

- `inetaf/tcpproxy` -- Lima's own proxy dependency (at the version
[...]

PR #5401, opened 2026-08-09
portfwdserver: relay TCP half-close from the host

### LINKED ISSUE #5264: portfwd: TCP half-closes might be broken
> Thanks, LGTM
> 
> Fix has also been confirmed by original reporter.
> 
> It even got a a rare "Merge as-is" recommendation from AI. Please take a look anyways because it mentions a pre-existing issue relating to half-closes: https://jandubois.github.io/lima/20260716-204108-pr-5216.html

_Originally posted by @jandubois in https://github.com/lima-vm/lima/pull/5216#pullrequestreview-4719289149_
            

- - -

> ### Half-close from the host tears down the guest connection before the response is relayed (future) [Gemini] [Claude]
> 
> `Start` returns as soon as any `Close*` fires, so a client that calls `shutdown(SHUT_WR)` and waits for a reply loses it. The chain: the host's `io.Copy` finishes, `GrpcClientRW.CloseWrite()` calls `stream.CloseSend()` ([`pkg/portfwd/client.go:129`](https://github.com/lima-vm/lima/blob/6920d840efa25ec22fd78834354532e830470251/pkg/portfwd/client.go#L129)), the guest's `Recv()` returns `io.EOF`, and `closeRead(rw)` closes `closeCh` and ends the tunnel.
> 
> This predates the PR. I confirmed it by running the new tests against the merge-base: `TestTunnelServerClosesGuestConn` fails at the write loop ([`server_test.go:151`](https://github.com/lima-vm/lima/blob/6920d840efa25ec22fd78834354532e830470251/pkg/portfwdserver/server_test.go#L151)), meaning it passed the `startDone` assertion — the old code also returned from `Start` on the same `CloseRead`, and lost the response to a dead stream. This PR changes only how the guest connection dies: a prompt close instead of a lingering leaked FD, which is strictly better.
> 
> Gemini proposed dropping `closeOnce.Do` from `CloseRead`. That enables half-close but changes teardown semantics well beyond a leak fix, and it needs its own tests — the guest→host direction then parks in `conn.Read` until the guest service closes or the host cancels the stream. Treat it as follow-up work, not a change to this PR.

ravi-arnan:
Fix #5264

A client that sends its request, calls `shutdown(SHUT_WR)`, and waits for the response
gets a premature EOF through the gRPC forwarder. The FIN does reach the guest service,
but the forwarder tears the tunnel down before the response can be relayed.

Reproducer, analysis and the linked history are @elprans's, from #5329.

## Cause

`tcpproxy` ends each copy direction with `closeRead(src)` / `closeWrite(dst)`. On the
guest side the stream adapter is `src`, so when the host half-closes:

1. the host calls `stream.CloseSend()`, so the guest's `Recv` returns `io.EOF`;
2. the stream to guest copy finishes. `closeWrite` shuts down the write side of the
   dialed connection, which is right, the guest service does see the FIN. `closeRead`
   calls `GRPCServerRW.CloseRead`;
3. `CloseRead` closed `closeCh`, which returns from `TunnelServer.Start`, whose
   deferred `conn.Close()` then tears down the guest connection;
4. the response the guest service writes a moment later has nowhere to go.

## Fix

A half-close and a full close are already distinguishable on the wire:
`GrpcClientRW.CloseWrite` only calls `stream.CloseSend()`, while `GrpcClientRW.Close`
also cancels the stream context. So `io.EOF` from `Recv` with a live context is a
half-close.

`CloseRead` no longer ends the tunnel. The proxy shuts down the write side of the guest
connection by itself, so there is nothing left for it to do, and the tunnel now lives
until the guest service closes (`CloseWrite`) or the host cancels the stream (`Close`).
`Close` and `CloseWrite` keep the `sync.Once` from #5216, so the FD leak fixed there
stays fixed and none of the three can block.

Only the guest side changes. The host side already gets this right: its `CloseRead`
cancels the stream, but it is reached only after `stream.Recv` has failed, so the stream
is over anyway.

## Tradeoff worth a look

Relaying a half-close means the tunnel outlives the host's FIN. A peer that fully closes
is indistinguishable from one that half-closes, so if the guest service ignores EOF and
never closes, that tunnel now stays up where before it was torn down promptly. This is
inherent to half-close support and is why nginx gates the same behaviour behind
`proxy_half_close`. `inetaf/tcpproxy`, `docker-proxy` and `socat` all relay it by
default. Happy to add an idle timeout instead if you would rather bound it.

## Not fixed here

The reverse direction, a half-close from the guest service towards the host, is still
not relayed. `TunnelMessage` cannot express EOF and a gRPC server ends a stream only by
returning from its handler, so that needs a protocol change plus guest agent version
negotiation. I left it out and noted the gap in a comment; glad to follow up separately.

## Tests

`TestTunnelServerRelaysHalfClose` is new and fails on master with
`response was lost; the half-close tore the tunnel down`.

`TestTunnelServerClosesGuestConn` now cancels the stream context to end the tunnel.
Closing the send side alone is a half-close, which must keep the tunnel up, so on its
own it no longer ends `Start`. Its subject, that the dialed connection is fully closed
and not leaked, is unchanged.

`go test ./...` passes; `-race -count=5` on the two portfwd packages passes;
golangci-lint reports 0 issues.

## Verified end to end

With the reproducer from #5264, on Linux/QEMU (Zorin OS, x86_64, Ubuntu 26.04 guest,
`vmType: qemu`), same tree, only the patch differing. containerd is disabled in the
template because this host cannot reach GitHub releases; it plays no part in the test.

master, `2.2.0-110-gb8b365a0`:

```
== running host-side client through the grpc forwarder
FAIL: premature EOF after half-close; connection was fully closed by the forwarder (got b'', expected b'OK 5\n')

real	0m0,207s
```

with this patch, `2.2.0-110-gb8b365a0.m`:

```
== running host-side client through the grpc forwarder
PASS: server saw the half-close and the response came back: b'OK 5\n'

real	0m1,026s
== guest server log:
listening on 127.0.0.1:12345
accepted ('127.0.0.1', 36454)
request complete (5 bytes), responding
```

The 1.0 s is the reproducer's deliberate 0.5 s server-side processing delay plus
overhead, matching the SSH forwarder's timing in #5264. The FIN propagates immediately.

Assisted-by: Claude Code (Opus 5)

### DIFF
--- a/pkg/portfwdserver/server.go
+++ b/pkg/portfwdserver/server.go
@@ -57,7 +57,8 @@ func (s *TunnelServer) Start(stream api.GuestService_TunnelServer) error {
 	go proxy.HandleConn(rw)
 
 	// The stream will be closed when this function returns.
-	// Wait here until rw.Close(), rw.CloseRead(), or rw.CloseWrite() is called.
+	// Wait here until rw.Close() or rw.CloseWrite() is called. A half-close
+	// from the host does not end the tunnel; see CloseRead below.
 	<-rw.closeCh
 	logrus.Debugf("closed GRPCServerRW for id: %s", in.Id)
 
@@ -68,12 +69,11 @@ type GRPCServerRW struct {
 	id      string
 	stream  api.GuestService_TunnelServer
 	closeCh chan any
-	// closeOnce guards closeCh. Close, CloseRead, and CloseWrite may be
-	// called in any order and multiple times (e.g., tcpproxy calls
-	// CloseRead/CloseWrite from each copy direction and then Close), so
-	// they must be idempotent and must never block; otherwise the proxy
-	// goroutine gets stuck and never closes the dialed guest connection,
-	// leaking one FD per forwarded connection.
+	// closeOnce guards closeCh. Close and CloseWrite may be called in any
+	// order and multiple times (e.g., tcpproxy calls CloseWrite from a copy
+	// direction and then Close), so they must be idempotent and must never
+	// block; otherwise the proxy goroutine gets stuck and never closes the
+	// dialed guest connection, leaking one FD per forwarded connection.
 	closeOnce sync.Once
 
 	// rxBuf holds bytes received from the stream that did not fit in the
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
> Reproducer, analysis and the linked history are @elprans's, from https://github.com/lima-vm/lima/issues/5329.

@elprans Would you be interested in reviewing this PR?

balajiv113 (approved):
Looks good tested the repro and its working fine as mentioned

PR #5405, opened 2026-08-10
wsl2: validate guest-reported address in getSSHAddress

SABITHSAHEB:
getSSHAddress reads the instance address from commands run inside the guest (`hostname -I | cut`, `ip route get 1 | awk`); only the fallback probe checked the output was an IP, the first two returned it verbatim. The pipeline in the first probe exits 0 even when `hostname` doesn't support `-I`, so its usage text could become `inst.SSHAddress` and end up on the ssh/scp/rsync command line instead of falling through to the next probe.

Route all three probes through one helper that accepts only global unicast addresses (private and ULA ranges included), so unspecified, loopback, link-local and multicast values fall through the same way `127.0.1.1` already did in the fallback.

This is input validation rather than a security boundary: a WSL2 guest already has interop access to the Windows host, so a compromised guest doesn't gain anything new from this path.


SABITHSAHEB:
any update?

jandubois:
AI review: https://jandubois.github.io/lima/20260911-232216-pr-5405.html

I'm out of time again for checking the review myself right now; not sure if I1 is a valid concern. Maybe the PR is good as-is; will take a look tomorrow.

jandubois:
So I took a look again. The PR seems generally fine, but:

* the security claim in the PR message seems overstated
* calling every parsable address except localhost routable seems incorrect

I don't know if this can actually happen, but it doesn't hurt to use `!ip.IsGlobalUnicast()` instead of `ip.IsLoopback()` to also filter `0.0.0.0`, multicast, and link-local addresses.

SABITHSAHEB:
Agreed on both. A WSL2 guest already has interop access to the host, so I've dropped the guest-to-host claim from the description; the practical case is the first probe handing ssh a non-address (the pipeline exits 0 even when `hostname` rejects `-I`, so its usage text was returned) instead of falling through. Switched the helper to `!ip.IsGlobalUnicast()` and added test rows for unspecified, link-local, multicast and broadcast.

PR #5408, opened 2026-08-11
only remove a stale socket at the hostSocket path

Nashit-h:
The unix-socket port forwarders call os.RemoveAll on a rule's hostSocket before binding, but hostSocket is validated only as an absolute path and may come from a remote template, so a value that points at an existing directory or regular file gets deleted recursively when the instance starts. Add osutil.RemoveStaleSocket, which removes the path only when it is a socket and otherwise leaves it in place and returns an error, and call it from both prepareUnixSocket and forwardSSH.


AkihiroSuda (approved):
Thanks

PR #5409, opened 2026-08-11
cidata: detect existing additionalDisks by filesystem, not label

### LINKED ISSUE #5407: `additionalDisks`: a disk whose name is longer than 11 characters is silently reformatted on every boot
### Description

An `additionalDisks` entry whose name is too long to fit in the filesystem label is reformatted on every boot, destroying whatever was stored on it. There is no warning, and the disk still mounts normally afterward, so the data loss is silent.

### Mechanism

`pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-disks.sh` decides whether a disk needs first-time setup by testing for a by-label device node, then formats it with a label derived from the same name:

```bash
if [[ ! -b "/dev/disk/by-label/lima-${DISK_NAME}" ]]; then
    if $FORMAT_DISK; then
        echo 'type=linux' | sfdisk --label gpt "/dev/${DEVICE_NAME}"
        mkfs.$FORMAT_FSTYPE $FORMAT_FSARGS -L "lima-${DISK_NAME}" "/dev/${DEVICE_NAME}1"
    fi
fi
```

`mkfs.ext4` caps volume labels at 16 bytes and truncates silently. The `lima-` prefix takes 5 of those, so any name longer than 11 characters is written with a label that does not match the path the guard tests. The guard is therefore true on every boot and `mkfs` runs again.

The mount at the end of the same loop uses the device path, not the label, so the disk always mounts and nothing looks wrong.

The limit depends on the filesystem: ext4 allows 16 bytes, xfs only 12 (so xfs breaks at names longer than 7 characters), btrfs 255. `fsType` is user selectable, so a single hardcoded constant would not be a correct fix.

### Reproduction

```yaml
additionalDisks:
  - name: "myproject-images"   # 16 characters, over the limit
    format: true
    fsType: ext4
```

Start the instance, write a file into `/mnt/lima-myproject-images/`, restart, and the file is gone. In the guest:

```console
$ blkid /dev/vdb1
/dev/vdb1: LABEL="lima-myproject-i" ...      # truncated to 16 bytes

$ ls /dev/disk/by-label/
lima-myproject-i                             # not lima-myproject-images

$ sudo tune2fs -l /dev/vdb1 | grep 'Filesystem created'
Filesystem created:       <the current boot's timestamp>
```

### Possible fix

Stop identifying the disk by its filesystem label. Treating "the partition already carries a filesystem" as the first-time-setup signal removes the length limit and, importantly, still recognises disks formatted by older Lima versions, so an upgrade cannot trigger the very data loss it is meant to fix. A GPT partition label (36 characters, `/dev/disk/by-partlabel/`) can carry the full name for user convenience.

Rejecting over-long names would be the other option, but it would break configs that work correctly once identification no longer depends on the label.

I have a patch along these lines and can open a PR if this direction sounds right.

pujitha24:
Motivation:
An additionalDisks entry whose name is longer than 11 characters was
reformatted on every boot, silently destroying whatever was stored on it.
05-lima-disks.sh decided whether a disk needed first-time formatting by
testing for "/dev/disk/by-label/lima-${DISK_NAME}". mkfs truncates
filesystem labels (16 bytes for ext4, 12 for xfs, minus 5 for the "lima-"
prefix), so for any name over that limit the label written to disk never
equals the one being tested for, the "already set up" check is always
false, and the disk is wiped and reformatted every boot. The disk still
mounted normally afterward (mount uses the device path, not the label),
so nothing looked wrong to the user.

Approach:
Detect first-time setup by asking whether the partition already carries a
filesystem ("blkid -s TYPE -o value") instead of checking for the
truncated label. This removes the name-length limit entirely and also
means a disk formatted by an older Lima version is still recognized as
already set up, so upgrading cannot trigger the same data loss. Two
downstream resize/fsck-check lines had the identical truncated-label
problem (they silently skipped growpart/resize2fs/xfs_growfs for
long-named disks instead of losing data); they're switched to the same
direct device path that mount already uses.

Validation:
- go build ./... and go test ./pkg/cidata/... pass.
- shellcheck and shfmt -s -d pass on the changed file.
- I used an AI coding assistant to help draft this change, and checked
  its reasoning myself: commands in an "if ! ...; then" condition are
  exempt from errexit under `set -eux -o pipefail`, so blkid failing on
  a not-yet-partitioned device won't abort the script. I also had it
  write a standalone bash simulation (not part of this commit) of the
  old by-label condition versus the new blkid-based one across two
  simulated boots of a 16-character disk name: the old condition
  reformats on boot 2, reproducing the reported data loss; the new one
  doesn't.
- I have not booted a real Lima VM with this change, so it hasn't been
  exercised through an actual guest boot or this repo's
  BATS/hack/test-templates.sh integration tests. I don't have a working
  VM environment available to me. If there's a lightweight way to get
  that coverage, or someone is willing to test it on a real VM, I'd
  appreciate the help.

Assisted-by: Claude

Report: https://github.com/lima-vm/lima/issues/5407
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

Fixes #5407

### DIFF
--- a/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-disks.sh
+++ b/pkg/cidata/cidata.TEMPLATE.d/boot.Linux/05-lima-disks.sh
@@ -22,8 +22,12 @@ for i in $(seq 0 $((LIMA_CIDATA_DISKS - 1))); do
 	test -n "$FORMAT_DISK" || FORMAT_DISK=true
 	test -n "$FORMAT_FSTYPE" || FORMAT_FSTYPE=ext4
 
-	# first time setup
-	if [[ ! -b "/dev/disk/by-label/lima-${DISK_NAME}" ]]; then
+	# first time setup: a partition that already carries a filesystem was set up by an
+	# earlier boot (possibly by an older Lima version). This does not rely on the
+	# "lima-${DISK_NAME}" filesystem label, which mkfs silently truncates for long disk
+	# names, causing the label lookup to never match and the disk to be reformatted (and
+	# its data destroyed) on every boot.
+	if ! blkid -s TYPE -o value "/dev/${DEVICE_NAME}1" >/dev/null 2>&1; then
 		if $FORMAT_DISK; then
 			if [ "$FORMAT_FSTYPE" == "swap" ]; then
 				echo 'type=swap' | sfdisk --label gpt "/dev/${DEVICE_NAME}"
@@ -46,11 +50,11 @@ for i in $(seq 0 $((LIMA_CIDATA_DISKS - 1))); do
 	if command -v growpart >/dev/null 2>&1 && command -v resize2fs >/dev/null 2>&1; then
 		growpart "/dev/${DEVICE_NAME}" 1 || true
 		# Only resize when filesystem is in a healthy state
-		if command -v "fsck.$FORMAT_FSTYPE" -f -p "/dev/disk/by-label/lima-${DISK_NAME}"; then
+		if command -v "fsck.$FORMAT_FSTYPE" -f -p "/dev/${DEVICE_NAME}1"; then
 			if [[ $FORMAT_FSTYPE == "ext2" || $FORMAT_FSTYPE == "ext3" || $FORMAT_FSTYPE == "ext4" ]]; then
-				resize2fs "/dev/disk/by-label/lima-${DISK_NAME}" || true
+				resize2fs "/dev/${DEVICE_NAME}1" || true
 			elif [ "$FORMAT_FSTYPE" == "xfs" ]; then
-				xfs_growfs "/dev/disk/by-label/lima-${DISK_NAME}" || true
+				xfs_growfs "/dev/${DEVICE_NAME}1" || true
 			else
 				echo >&2 "WARNING: unknown fs '$FORMAT_FSTYPE'. FS will not be grew up automatically"
 			fi


AkihiroSuda:
> This sandbox cannot boot a real Lima VM, so the fix was not exercised
through an actual guest boot or this repo's BATS/hack/test-templates.sh
integration tests, which are what this repo's docs point to for
behavior that spans multiple boots.

What is "this sandbox"?
Why can't you test an actual guest boot?
Are you an AI? 🤔 

> The change is narrow (a single
detection condition plus two path substitutions) and the reasoning
above is believed sufficient, but this is disclosed as a real gap in
validation.

Believed sufficient by who?

pujitha24:
To answer directly: yes, I use an AI coding assistant (Claude) heavily in preparing PRs like this one, including drafting this reply, and I should have disclosed that up front instead of leaving that vague "sandbox" line in — sorry for the confusion. "Sandbox" meant the assistant's own dev environment, which has no VM/hypervisor access; nothing to do with Lima itself. I've rewritten the description in my own words, added an `Assisted-by: Claude` trailer per the contributing doc, and dropped the "believed sufficient" hedge — that judgment call was mine after reviewing the assistant's reasoning, not anyone else's. I still haven't run this through an actual guest boot or the BATS suite, since I don't have a real VM environment to test with — if there's a lightweight way to get that coverage, or you'd rather find someone to test it on real hardware, let me know.

AkihiroSuda:
> I still haven't run this through an actual guest boot or the BATS suite, since I don't have a real VM environment to test with

If you don't use Lima, what is motivating your contribution? 🤔

pujitha24:
Fair question. I don't run Lima day to day — I came across #5407 while looking for well-scoped open issues to work on, and this one stood out because bcollard had already nailed the root cause (the mkfs label truncation) and sketched the fix direction, so it looked like something I could implement correctly by reasoning through the script rather than needing hands-on Lima experience. I get that's a different kind of motivation than hitting the bug myself, and I understand if that means this needs more scrutiny before it's mergeable. I'm still glad to test it if there's a lightweight way to boot a VM for this, or to have someone else validate it directly.

AkihiroSuda:
> I'm still glad to test it if there's a lightweight way to boot a VM for this, or to have someone else validate it directly.

Contributors are expected to test their PRs ahead of submissions, so I'm closing your PRs.
You, or anybody else, may still resubmit them with same content after manual verification.

PR #5423, opened 2026-08-18
validate 9p securityModel, protocolVersion, and cache

Nashit-h:
## What This PR Changes

`Validate` checked only `msize` among the 9p mount options and left `securityModel`, `protocolVersion`, and `cache` unvalidated. `securityModel` is written verbatim into the host-side QEMU `-virtfs local,...,security_model=%s` argument, and `protocolVersion`/`cache` go into the guest mount options string, so a value containing a comma in a template that `limactl create`/`start` fetches from a URL injects extra options into those strings (for example `securityModel: "none,readonly=off"`). This rejects values outside the sets already documented in `docs/config/mount.md`, next to the existing `mountType` check.

N/A, small validation hardening; the accepted values are already documented for these fields.

## How I Tested This

`go test ./pkg/limayaml/...`, plus a new table test in `validate_test.go` that fails before this change and passes after: the documented non-default values are accepted, and a `securityModel`/`protocolVersion`/`cache` value carrying an injected `,option` is rejected.

### DIFF
--- a/pkg/driver/qemu/qemu_driver.go
+++ b/pkg/driver/qemu/qemu_driver.go
@@ -104,6 +104,10 @@ func validateConfig(cfg *limatype.LimaYAML) error {
 		return err
 	}
 
+	if err := validate9pMountOptions(cfg); err != nil {
+		return err
+	}
+
 	for i, nw := range cfg.Networks {
 		if unknown := reflectutil.UnknownNonEmptyFields(nw,
 			"Lima",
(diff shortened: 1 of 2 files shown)


pkg/limayaml/validate.go
```diff
@@ -30,6 +30,16 @@ import (
...
 
+// The 9p option values below are interpolated verbatim into the QEMU `-virtfs`
+// argument (securityModel) and the guest mount options (protocolVersion, cache),
+// so a value outside these sets, e.g. one containing a comma, would inject extra
+// options. Keep them in sync with the documented values in docs/config/mount.md.
+var (
+	supported9pSecurityModels   = []string{"passthrough", "mapped-xattr", "mapped-file", "none"}
+	supported9pProtocolVersions = []string{"9p2000", "9p2000.u", "9p2000.L"}
+	supported9pCaches           = []string{"none", "loose", "fscache", "mmap"}
+)
```
AkihiroSuda: These should be verified by the driver, not by limactl, as these options are specific to the QEMU driver.

AkihiroSuda (changes requested):
https://github.com/lima-vm/lima/pull/5423/changes#r3804152845

pkg/limayaml/validate.go
Nashit-h: Makes sense, 9p is QEMU-only here. Moved the checks into the driver's validateConfig (new validate9pOptions) and dropped them from limactl, test moved with it.

Nashit-h: Makes sense, these are QEMU-specific. Moved the checks into the QEMU driver's Validate path (validate9pMountOptions in qemu_driver.go) and reverted the limayaml change. Test moved along with it.

AkihiroSuda:
CI seems failing.
Please try rebasing.

PR #5426, opened 2026-08-19
strip control characters from guest cloud-init log lines

Nashit-h:
## What This PR Changes

The hostagent reads cloud-init log lines from the guest over SSH and forwards them in `CloudInitProgress` events, which `limactl start` and `limactl watch` print to the operator's terminal with a bare `%s`. A guest controls those bytes (anything in `/var/log/cloud-init-output.log` or the cloud-init journald output), so it can emit ANSI/OSC escape sequences and spoof or rewrite the terminal during the default start progress display. This strips non-printable runes at the single emit choke point (`emitCloudInitProgressEvent`) so every consumer is safe, matching the `removeControlChars` handling already used for other guest-controlled strings and the `%#q` quoting used for the sibling guest-error path.

Small self-contained security fix; no separate tracking issue.

## How I Tested This

Added a regression test that runs an escape sequence through `emitCloudInitProgressEvent` and asserts the emitted `LogLine` has no control characters. It fails on the current code (the `\x1b` survives) and passes with this change. `go test ./pkg/hostagent/...` and the repo-pinned `golangci-lint run ./pkg/hostagent/` are green.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -48,6 +48,7 @@ import (
 	"github.com/lima-vm/lima/v2/pkg/portfwd"
 	"github.com/lima-vm/lima/v2/pkg/sshutil"
 	"github.com/lima-vm/lima/v2/pkg/store"
+	"github.com/lima-vm/lima/v2/pkg/strutil"
 	"github.com/lima-vm/lima/v2/pkg/version/versionutil"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/hostagent/hostagent.go
```diff
@@ -363,6 +370,18 @@ func (a *HostAgent) emitCloudInitProgressEvent(ctx context.Context, progress *ev
 	a.emitEvent(ctx, ev)
 }
 
+// removeControlChars drops non-printable runes so guest-controlled text is safe
+// to print to the operator's terminal.
+func removeControlChars(s string) string {
```
AkihiroSuda: Duplicated code:
https://github.com/lima-vm/lima/blob/20dbdc0e8da292fa9b82178affb6310b0835a996/pkg/cidata/cidata.go#L499-L507

Should be moved to `pkg/strutil`

Nashit-h: Good point. Moved it to strutil.RemoveControlChars and switched both cidata and the hostagent path over to it.

AkihiroSuda:
Please squash the commits

Nashit-h:
Squashed down to a single commit.

AkihiroSuda (approved):
Thanks

PR #5436, opened 2026-08-22
fix(vz): stop the VM even when CanRequestStop is false

ekalinin:
## Problem

`LimaVzDriver.Stop()` returns `vz: CanRequestStop is not supported` without stopping the VM when `CanRequestStop()` is false, and it never removes the driver PID file. The qemu and krunkit drivers do remove theirs.

On an instance registered as a LaunchDaemon (`limactl autostart enable --condition=boot`), launchd sends SIGTERM to the host agent, `Stop()` fails, and `vz.pid` is left behind. Within the same boot that is harmless, because `store.ReadPIDFile()` removes the PID file of a dead process, but after a reboot the recorded PID may have been reused by an unrelated process. The instance is then reported as `Broken` with `vz driver is running but host agent is not`, and launchd with `KeepAlive` crash-loops `limactl start`.

This is bug 1 of #5087. It overlaps with the vz part of the draft PR #5088.

## Changes

- The guest is asked to shut down over SSH whenever it could not be asked through the Virtualization framework, not only for the macOS guests that ignore `RequestStop()`. That is what stops the VM in the reported case, where `CanRequestStop()` is false.
- The PID file is always removed when the driver stops, so a failure to stop the VM cannot leave a stale `vz.pid` behind.
- The wait loop is extracted into `waitForStopped()`, which also stops its ticker. The timeout is unchanged at 30 seconds.

The VM is never stopped forcibly with `VirtualMachine.Stop()`: as noted in review, that does not give the guest a chance to stop cleanly and may corrupt the disk. When the guest ignores every request to shut down, `Stop()` returns an error, as it does today.

## Not in this PR

The LaunchDaemon plist does not set `ExitTimeOut`, so launchd sends SIGKILL 20 seconds after SIGTERM, which can cut the graceful shutdown short. That belongs to the autostart code, so it is left for a separate change.

## Testing

`go test ./pkg/driver/vz/...` covers the two extracted helpers, including that the PID file is removed and that removing a missing one is not an error. `Stop()` itself needs a real VM, so it is not unit tested.

`make golangci-lint` is clean.

Assisted-by: Claude Code


pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -500,43 +500,79 @@ func (l *LimaVzDriver) requestStopViaSSH(ctx context.Context) error {
...
-				l.machine.mu.Lock()
-				stopped := l.machine.stopped
-				l.machine.mu.Unlock()
-				if stopped {
-					return nil
-				}
+	if !l.machine.CanStop() {
+		return errors.New("vz: the VM cannot be stopped")
+	}
+	if err := l.machine.Stop(); err != nil {
```
balajiv113: During initial implementation i remember doing direct stop is forceful close and may corrupt the disk. Can you confirm via the docs if its okay to use this.

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -500,43 +500,79 @@ func (l *LimaVzDriver) requestStopViaSSH(ctx context.Context) error {
...
-				logrus.WithError(err).Warn("Failed to request shutdown via SSH")
+	if l.machine.CanRequestStop() {
+		if _, err := l.machine.RequestStop(); err != nil {
+			logrus.WithError(err).Warn("vz: failed to request the guest to stop")
+		} else {
+			timeout := gracefulStopTimeout
+			if *l.Instance.Config.OS == limatype.DARWIN {
+				// macOS VM does not respond to l.machine.RequestStop(),
+				// so we need to run `shutdown -h now` in the VM via SSH for graceful shutdown.
+				if err := l.requestStopViaSSH(ctx); err != nil {
```
balajiv113: Can't we extend this itself for linux as well ? calling stop via ssh looks safe

pkg/driver/vz/vz_driver_darwin.go
balajiv113: https://developer.apple.com/documentation/virtualization/vzvirtualmachine/stop(completionhandler:)#Discussion

```
Warning
This is a destructive operation. It stops the VM without giving the guest a chance to stop cleanly.
```

pkg/driver/vz/vz_driver_darwin.go
ekalinin: Done. SSH shutdown is now used for any guest that could not be asked through the framework, not only for macOS guests.

pkg/driver/vz/vz_driver_darwin.go
ekalinin: You're right, thanks! Fixed: It is now only reached after both `RequestStop()` and `shutdown -h` now over SSH failed or timed out.

The reason I kept it: `Stop()` is called from `startRoutinesAndWait()` on the way out, and the host agent process exits right after it, so with vz the VM is destroyed by the process exit anyway — that is exactly what happens today when `Stop()` returns the `CanRequestStop is not supported error`. Calling it explicitly only makes the teardown ordered and lets us wait for the VM to be gone. If you prefer, I can drop the forced stop and return an error instead.

balajiv113: Please return error. As calling stop will corrupt the disk which is dangerous. I believe we already have error only if am not wrong

ekalinin: Removed, `Stop()` now returns the error from the wait as before.

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -500,43 +500,65 @@ func (l *LimaVzDriver) requestStopViaSSH(ctx context.Context) error {
...
 }
 
+// gracefulStopTimeout is how long to wait for the guest to shut down by itself.
+// Most Linux machines shutdown within 5 seconds, but macOS machines can take longer.
+const gracefulStopTimeout = 30 * time.Second
+
 func (l *LimaVzDriver) Stop(ctx context.Context) error {
 	logrus.Info("Shutting down VZ")
-	canStop := l.machine.CanRequestStop()
+	defer l.removePIDFile()
```
balajiv113, line 509: Do we need this explicitly now ??

If requestStop is not supported anyway we have ssh shutdown this should cover almost all cases. Even that doesn't work stop will end in error (having pid in this case is better right ??).

Just my take please let me know what you think

ekalinin: With vz the PID file holds the PID of the process that runs the VM in-process: `vz.pid` is written with `os.Getpid()` in `vm_darwin.go`. `Stop()` is only called on the way out `startRoutinesAndWait()` at `hostagent.go:530`, and for an external driver the driver process is stopped on the next line by `server.Stop()`. 
So once `Stop()` returns, with or without an error, that `PID` is dead and the file cannot mean "the VM is still running".

The cleanup is also not only about the error path: nothing ever removes `vz.pid`, not even after a successful stop. qemu removes it in `killQEMU()` and krunkit in the goroutine that waits for the process; vz is the only driver that doesn't. It goes unnoticed today because `ReadPIDFile()` drops the PID file of a dead process, so within one boot it self-heals. After a reboot the PID may have been reused, and that is the broken state reported in #5087.

pkg/driver/vz/vz_driver_darwin.go
```diff
@@ -500,43 +500,65 @@ func (l *LimaVzDriver) requestStopViaSSH(ctx context.Context) error {
...
-		_, err := l.machine.RequestStop()
-		if err != nil {
-			return err
+	requested := false
+	if l.machine.CanRequestStop() {
+		if _, err := l.machine.RequestStop(); err != nil {
+			logrus.WithError(err).Warn("vz: failed to request the guest to stop")
+		} else {
+			requested = true
 		}
```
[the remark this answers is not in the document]
ekalinin: Fixed

AkihiroSuda:
@balajiv113 Was your concern resolved?

balajiv113 (approved):
LGTM 👍

PR #5439, opened 2026-08-24
go.mod: update minimum Go version to 1.26 (remove `pkg/ptr`)

AkihiroSuda:
## What This PR Changes

Remove `pkg/ptr`.
Since Go 1.26, `new()` can be used instead of `ptr.Of()`.

None

## How I Tested This
```
go build ./...
```

## AI Usage

None


go.mod
```diff
@@ -1,7 +1,7 @@
 // gomodjail:confined
 module github.com/lima-vm/lima/v2
 
-go 1.25.7
+go 1.26
```
alexandear, line 4: Why not `go 1.26.0`?

pkg/windows/registry_windows.go
```diff
@@ -33,7 +33,7 @@ func AddVSockRegistryKey(port int) error {
 	}
 
 	if slices.Contains(used, port) {
-		return fmt.Errorf("port %#q in use", port)
+		return fmt.Errorf("port %d in use", port)
```
alexandear, line 36: It's a bugfix? Maybe we should do it in a separate PR.

alexandear (approved):

go.mod
AkihiroSuda: Does it matter here?

pkg/windows/registry_windows.go
AkihiroSuda: This is a lint fix that is necessary to pass the CI

go.mod
alexandear: `go 1.26.0` requires a released toolchain.
`go 1.26` also accepts prereleases (beta and RC).

So with `go 1.26` someone on `go1.26rc1` could build this and hit toolchain bugs we can't reproduce.

AkihiroSuda: > So with go 1.26 someone on go1.26rc1 could build this and hit toolchain bugs we can't reproduce.

We do not need to care about them

PR #5441, opened 2026-08-24
qemuimgutil: reject qcow2 external data file references

Nashit-h:
## What This PR Changes

`QemuImageUtil.Convert` only rejected a source image that had a backing file, and `AcceptableAsBaseDisk` rejected backing files and VMDK extents but not a qcow2 external data file (`format-specific.data.data-file`). `qemu-img` reads that file from the host when it opens the image, so a crafted qcow2 whose data file points at a host path such as `/etc/passwd` gets that path's contents copied into the guest disk during conversion, or read directly when the image is booted as a base disk. An image URL can come from a remote template, so those bytes are attacker-influenced. The native converter already refuses such images through go-qcow2reader's `Readable()`, and `proxyimgutil` prefers the qemu implementation when `qemu-img` is installed, so the weaker check was the one that ran. This adds the missing `data-file` field to the parsed info and routes both callees through one `rejectExternalFileReferences` helper covering backing files, the qcow2 data file, and VMDK extents.

No pre-filed issue; this is a self-contained fix for a host-file disclosure on the image path. Glad to open one first if you prefer.

## How I Tested This

`go test ./pkg/qemuimgutil/...`, with new cases asserting the `Convert` validation and `AcceptableAsBaseDisk` reject a qcow2 data file and a VMDK extent while a plain qcow2 still passes. Manual check with qemu-img 11.0:

```
qemu-img create -f qcow2 -o data_file=secret.bin,data_file_raw=on evil.qcow2 4096
# place a host file at ./secret.bin, then:
qemu-img convert -O raw evil.qcow2 out.raw   # out.raw now holds secret.bin's bytes
qemu-img info --output=json evil.qcow2       # backing-filename is empty; data-file is set
```

Before the fix the empty backing filename let the image through; after it, the data file is rejected.

### DIFF
--- a/pkg/qemuimgutil/qemuimgutil.go
+++ b/pkg/qemuimgutil/qemuimgutil.go
@@ -84,6 +84,7 @@ func (sp *InfoFormatSpecific) Vmdk() *InfoFormatSpecificDataVmdk {
 
 type InfoFormatSpecificDataQcow2 struct {
 	Compat          string `json:"compat,omitempty"`           // since QEMU 1.7
+	DataFile        string `json:"data-file,omitempty"`        // since QEMU 3.0
 	LazyRefcounts   bool   `json:"lazy-refcounts,omitempty"`   // since QEMU 1.7
 	Corrupt         bool   `json:"corrupt,omitempty"`          // since QEMU 2.2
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):
Thanks

PR #5443, opened 2026-08-26
qemu: support `nestedVirtualization` with HVF (QEMU 11.1 or later)

### LINKED ISSUE #5419: nestedVirtualization: support QEMU
https://github.com/lima-vm/lima/blob/09903953eb78ac9a1a9d9d8f1eb0f6b6b85a0e5f/templates/default.yaml#L654-L662

QEMU 11.1 seems to support nested virtualization now: https://wiki.qemu.org/ChangeLog/11.1

> hvf: support nested virtualization for the virt board

sharanrajt:
## What This PR Changes

QEMU 11.1 added nested virtualization support to the HVF accelerator for the aarch64 `virt` machine (https://wiki.qemu.org/ChangeLog/11.1).

This PR adds support for nested virtualization in the QEMU driver:
- Passes `virtualization=on` to the `virt` machine (`-machine virt,...,virtualization=on`) for `aarch64` and `armv7l` when `nestedVirtualization: true`.
- Enforces QEMU 11.1.0 or later and macOS 15 or later when using the `hvf` accelerator (since QEMU relies on macOS 15's `hv_vm_config_set_el2_enabled`).
- Emits a warning and ignores `nestedVirtualization` for architectures other than `aarch64` and `armv7l`.
- Updates `templates/default.yaml` comments to document `vmType: qemu` support on Apple M3+ / macOS 15+.

Closes #5419

## How I Tested This

- **Manual Testing:** Tested on Apple M4 / macOS 26.5 with QEMU 11.1.0: verified that the guest kernel initializes KVM in nVHE mode and a nested KVM VM boots EDK2 to the UEFI shell.
- **Unit Tests:** Added `TestValidateConfigNestedVirtualization` to `pkg/driver/qemu/qemu_test.go` and ran `go test -v ./pkg/driver/qemu/...`.
- **Build Verification:** Built successfully with `make minimal` and verified `go test ./...`.

## AI Usage

Assisted-by: Antigravity


templates/default.yaml
```diff
@@ -657,7 +657,7 @@ plain: null
 #   qemu-system-aarch64 -accel kvm -cpu host -M virt
 # - Without specifying `-cpu host`, nested virtualization may fail with the error:
 #   qemu-system-aarch64: kvm_init_vcpu: kvm_arch_init_vcpu failed (0): Invalid argument
-# - Only supported on Apple M3 or later with `vmType: vz` or `vmType: krunkit`.
+# - Only supported on Apple M3 or later with `vmType: vz`, `vmType: krunkit`, or `vmType: qemu` (QEMU 11.1 or later).
```
AkihiroSuda:
```suggestion
# - Only supported on Apple M3 or later.
```

pkg/driver/qemu/qemu.go
```diff
@@ -581,6 +602,9 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 		args = appendArgsIfNoConflict(args, "-machine", machine)
 	case limatype.ARMV7L:
 		machine := "virt,accel=" + accel
+		if y.NestedVirtualization != nil && *y.NestedVirtualization {
+			machine += ",virtualization=on"
```
AkihiroSuda: How did you test this configuration?

pkg/driver/qemu/qemu_driver.go
```diff
@@ -126,6 +126,18 @@ func validateConfig(cfg *limatype.LimaYAML) error {
 		}
 	}
 
+	if cfg.NestedVirtualization != nil && *cfg.NestedVirtualization {
```
AkihiroSuda, line 128: Can we support Linux hosts?

pkg/driver/qemu/qemu.go
sharanrajt: I didn't test it. I added it because the virt machine accepts virtualization=on for 32-bit too. But armv7l guests are`tcg`-only in Lima and Linux dropped `kvm` on 32-bit ARM hosts in 5.7, so a guest couldn't use it anyway. I'll remove the armv7l case and keep this PR to aarch64.

pkg/driver/qemu/qemu_driver.go
sharanrajt: Yes. For aarch64 the same code path applies on Linux/KVM, the macOS version check is darwin-only, and virtualization=on works with KVM given QEMU ≥ 10.1 and a ≥ 6.16 host kernel booted with kvm-arm.mode=nested (FEAT_NV hardware); otherwise QEMU fails with a clear "host kernel KVM does not support providing Virtualization extensions" error. I don't have such hardware, so this part is untested.
For x86_64/KVM, nested virt is a host-side setting exposed through -cpu host, which Lima already uses. I'll change the x86_64 path to check /sys/module/kvm_{intel,amd}/parameters/nested and error only when it's disabled, instead of warning "unsupported". I'll also document the Linux requirements in default.yaml.

AkihiroSuda:
Lint is failing

sharanrajt:
@AkihiroSuda The lint and test was failing because of using `github.com/lima-vm/lima/v2/pkg/ptr` which has been removed and fixed to use Go 1.26 `new()`.

pkg/driver/qemu/qemu.go
```diff
@@ -88,6 +88,44 @@ func minimumQemuVersion() (hardMin, softMin semver.Version) {
 	return hardMin, softMin
 }
 
+// minimumQemuVersionForNestedVirtualization is the oldest version of QEMU whose accelerator supports
+// providing the Virtualization Extensions (EL2) to an aarch64 guest (`-machine virt,virtualization=on`).
+// TCG has no such requirement.
+var minimumQemuVersionForNestedVirtualization = map[string]*semver.Version{
+	// KVM_CAP_ARM_EL2 support; also needs a host kernel (6.16 or later) booted with `kvm-arm.mode=nested`.
```
AkihiroSuda: This seems to be about ARM, while `kvmNestedVirtualizationEnabled` seems to only support x86_64

pkg/driver/qemu/qemu.go
```diff
@@ -88,6 +88,44 @@ func minimumQemuVersion() (hardMin, softMin semver.Version) {
...
+	"kvm": semver.New("10.1.0"),
+	// https://wiki.qemu.org/ChangeLog/11.1
+	"hvf": semver.New("11.1.0"),
+}
+
+// kvmNestedVirtualizationEnabled reports whether nested virtualization is enabled in the host KVM module
+// on x86_64 Linux, by reading the `nested` parameter of kvm_intel (Intel) or kvm_amd (AMD).
+func kvmNestedVirtualizationEnabled() (bool, error) {
```
AkihiroSuda: Could be moved to `*_linux_amd64.go`

pkg/driver/qemu/qemu.go
sharanrajt: Thank you for the review, the map is indeed only for aarch64 guests (EL2 / -machine virt,virtualization=on); for x86_64 guests nesting is controlled by the host KVM module instead. I've clarified the doc comment to state this explicitly and cross reference `kvmNestedVirtualizationEnabled`, and prefixed the inline comments with aarch64 Linux host: / aarch64 macOS host: so the per-arch scope is clear at a glance.

pkg/driver/qemu/qemu.go
sharanrajt: Done, moved `kvmNestedVirtualizationEnabled` into kvm_linux_amd64.go (build-tagged linux and amd64) with a stub in kvm_others.go for other platforms. parseKVMNestedParam stays in qemu.go since it's a pure parser with cross-platform tests.

pkg/driver/qemu/qemu.go
```diff
@@ -88,6 +88,31 @@ func minimumQemuVersion() (hardMin, softMin semver.Version) {
 	return hardMin, softMin
 }
 
+// minimumQemuVersionForNestedVirtualization is the oldest version of QEMU whose accelerator supports
+// providing the Virtualization Extensions (EL2) to an aarch64 guest (`-machine virt,virtualization=on`).
+// It is only consulted for aarch64 guests; for x86_64 guests, nested virtualization is controlled by the
+// host KVM module instead (see kvmNestedVirtualizationEnabled, implemented per platform).
+// TCG has no such requirement.
+var minimumQemuVersionForNestedVirtualization = map[string]*semver.Version{
```
AkihiroSuda: Should be moved to `*_arm64.go`

AkihiroSuda: You could just have `func nestedVirtualizationEnabled(accel string, qemuVer *semver.Version) (bool, error)` for each of  `qemu_{linux_darwin}_{amd64,arm64}.go` to cover both `kvmNestedVirtualizationEnabled` and `minimumQemuVersionForNestedVirtualization`

pkg/driver/qemu/qemu.go
AkihiroSuda: See https://github.com/lima-vm/lima/pull/5443#discussion_r3974889379

pkg/driver/qemu/qemu.go
sharanrajt: Thank you for the suggestion! To make sure I'm on the right track, below is my plan:

Replace both `minimumQemuVersionForNestedVirtualization` and `kvmNestedVirtualizationEnabled` with a single unified function:
``` go
func nestedVirtualizationEnabled(accel string, qemuVer *semver.Version) (bool, error)
```
with platform specific implementation we need to create the following:
qemu_linux_amd64.go -> KVM nested param (kvm_intel / kvm_amd)
qemu_linux_arm64.go -> QEMU ≥ 10.1.0 for KVM (KVM_CAP_ARM_EL2)
qemu_darwin_arm64.go -> QEMU ≥ 11.1.0 for HVF
qemu_darwin_amd64.go -> Returns error for HVF (unsupported on Intel Mac)
qemu_others.go -> Fallback stub ((true, nil))

`parseKVMNestedParam` stays in `qemu.go` since it's a pure parser tested cross-platform in `qemu_test.go`.

The caller in `Cmdline()` stays mostly the same — the AARCH64 case calls `nestedVirtualizationEnabled(accel, version)` instead of the map lookup, and the X8664 case calls it instead of `kvmNestedVirtualizationEnabled()`.

Does this match what you had in mind?

AkihiroSuda: > qemu_darwin_amd64.go -> Returns error for HVF (unsupported on Intel Mac)

Should return `(false, nil)` ?

> qemu_others.go -> Fallback stub ((true, nil))

The fallback value should be false?

sharanrajt: For `qemu_darwin_amd64.go`: Agreed, (false, nil) is cleaner and HVF not supporting nested virt on Intel Macs is a known limitation, not an error condition.

For `qemu_others.go`: Will change to (false, nil) as well, false is the safe default for unsupported platforms.

Can I also update the AARCH64 caller to check the bool return, so that (false, nil) is handled correctly (warn and skip virtualization=on).

pkg/driver/qemu/qemu_darwin_arm64.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// nestedVirtualizationEnabled reports whether the host can provide nested virtualization
+// for the given accelerator.
+//
+// On aarch64 macOS, HVF nested virtualization requires QEMU 11.1.0 or later.
+// https://wiki.qemu.org/ChangeLog/11.1
+func nestedVirtualizationEnabled(accel string, qemuVer *semver.Version) (bool, error) {
+	if accel != "hvf" {
+		return true, nil
```
AkihiroSuda: Did you mean `false` ?

pkg/driver/qemu/qemu_linux_arm64.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// nestedVirtualizationEnabled reports whether the host can provide nested virtualization
+// for the given accelerator.
+//
+// On aarch64 Linux, KVM_CAP_ARM_EL2 requires QEMU 10.1.0 or later and a host kernel
+// (6.16 or later) booted with `kvm-arm.mode=nested`.
+func nestedVirtualizationEnabled(accel string, qemuVer *semver.Version) (bool, error) {
+	if accel != "kvm" {
+		return true, nil
```
AkihiroSuda: Did you mean false?

pkg/driver/qemu/qemu_darwin_arm64.go
sharanrajt: Yes, your correct it should be `false`. I'll correct it.

pkg/driver/qemu/qemu_linux_arm64.go
sharanrajt: Yes, your correct it should be `false`. I'll correct it.

pkg/driver/qemu/qemu.go
```diff
@@ -550,6 +562,39 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
...
+				logrus.Fatalf("nested virtualization with %s: %v", strings.ToUpper(accel), err)
+			} else if ok {
+				nestedVirtAArch64 = true
+			} else {
+				logrus.Warnf("nested virtualization is not supported with accelerator %s on this platform, ignoring", strings.ToUpper(accel))
+			}
+		case limatype.X8664:
+			// On x86_64, nested virtualization is controlled by the host KVM module
+			// (the `nested` parameter of kvm_intel / kvm_amd) and is exposed to the guest via `-cpu host`.
+			if accel != "kvm" {
```
AkihiroSuda: This could be checked inside `nestedVirtualizationEnabled`

pkg/driver/qemu/qemu.go
```diff
@@ -550,6 +562,39 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
...
+				logrus.Warnf("field `nestedVirtualization` is not supported with accelerator %#q for architecture %#q, ignoring", accel, *y.Arch)
+				break
+			}
+			if enabled, err := nestedVirtualizationEnabled(accel, version); err != nil {
+				logrus.WithError(err).Warn("Failed to check whether nested virtualization is enabled in the host KVM module")
+			} else if !enabled {
+				return "", nil, errors.New("nested virtualization is disabled in the host KVM module (`nested` parameter of kvm_intel / kvm_amd)")
+			}
+			if !strings.HasPrefix(cpu, "host") {
```
AkihiroSuda: This could be checked inside nestedVirtualizationEnabled

pkg/driver/qemu/qemu.go
```diff
@@ -550,6 +562,39 @@ func Cmdline(ctx context.Context, cfg Config) (exe string, args []string, err er
 	}
 	args = appendArgsIfNoConflict(args, "-cpu", cpu)
 
+	// Nested virtualization
+	nestedVirtAArch64 := false
+	if y.NestedVirtualization != nil && *y.NestedVirtualization {
+		switch *y.Arch {
```
AkihiroSuda: This `switch` could be removed. `nestedVirtualizationEnabled` could handle the arch differences in the single place.

pkg/driver/qemu/qemu_linux_amd64.go
```diff
@@ -0,0 +1,38 @@
+//go:build linux && amd64
```
AkihiroSuda: Already implied by the file name

pkg/driver/qemu/qemu_linux_arm64.go
```diff
@@ -0,0 +1,28 @@
+//go:build linux && arm64
```
AkihiroSuda: Already implied by the file name

pkg/driver/qemu/qemu_linux_amd64.go
```diff
@@ -0,0 +1,56 @@
...
+			return false, nil
+		}
+		enabled, err := parseKVMNestedParam(string(b))
+		if err != nil {
+			return false, err
+		}
+		if !enabled {
+			return false, errors.New("nested virtualization is disabled in the host KVM module (`nested` parameter of kvm_intel / kvm_amd)")
+		}
+		if !strings.HasPrefix(cpu, "host") {
```
AkihiroSuda, line 50: This check doesn't need to be in the loop

AkihiroSuda (approved):
Thanks

pkg/driver/qemu/qemu_test.go
```diff
@@ -151,3 +153,69 @@ func TestSwtpmCmdline(t *testing.T) {
...
+		} else {
+			assert.NilError(t, err)
+		}
+	} else {
+		assert.NilError(t, err)
+	}
+}
+
+func TestNestedVirtualizationEnabled(t *testing.T) {
+	if runtime.GOOS == "darwin" && runtime.GOARCH == "arm64" {
```
AkihiroSuda, line 201: This could be moved to `*_darwin_arm64.go`

pkg/driver/qemu/qemu_driver.go
```diff
@@ -139,6 +139,18 @@ func validateConfig(cfg *limatype.LimaYAML) error {
...
```
AkihiroSuda, line 142: This check could be also moved to `*_darwin_*.go`

PR #5449, opened 2026-08-27
Implement HCS driver

### LINKED ISSUE #4961: Support for Windows hosts, without depending on QEMU
> We have to investigate whether we want to use Hyper-V directly or use HCS
> https://learn.microsoft.com/en-us/virtualization/api/hcs/schemareference
> 
> HCS is the basis of WSL2, so it might be available in the "Home Edition" of Windows 11, while Hyper-V is only available in Pro, Enterprise, etc. ?
> 
> Also, the support for 9p seems only available in HCS (correct me if I'm wrong)
> https://learn.microsoft.com/en-us/virtualization/api/hcs/schemareference#Plan9

 _Originally posted by @AkihiroSuda in [#4852](https://github.com/lima-vm/lima/issues/4852#issuecomment-4299027649)_

mie313:
## What This PR Changes
This PR addresses #4961 .
It adds HCS driver as VM driver options on Windows hosts.

#4961 

Closes #

## How I Tested This
I validated that a Linux(Ubuntu and Fedora) VM successfully launched with `vmType:hcs` on Windows 11 (x86_64).

## AI Usage
Assisted-by: Claude
I used AI as a reference when exploring some implementation ideas and generating sample code, but I worte the actual code by myself.


website/content/en/docs/releases/experimental.md
AkihiroSuda, line 18: Move `vmType:hcs` around this line

go.mod
```diff
@@ -130,3 +130,16 @@ require (
...
+require (
+	github.com/containerd/cgroups/v3 v3.1.3 // indirect
+	github.com/containerd/errdefs v1.0.0 // indirect
+	github.com/containerd/errdefs/pkg v0.3.0 // indirect
+	github.com/containerd/typeurl/v2 v2.3.0 // indirect
+	github.com/golang/groupcache v0.0.0-20241129210726-2c02b8208cf8 // indirect
+	go.opencensus.io v0.24.0 // indirect
+)
+
+replace github.com/Microsoft/hcsshim => github.com/AkihiroSuda/hcsshim v0.0.0-20260714172934-dc466dcf888f
```
AkihiroSuda: https://github.com/microsoft/hcsshim/pull/2825 is not merged yet, so you can just copy the code from https://github.com/AkihiroSuda/hcstest.tmp/blob/28a4feaa8f5d7f1f3eba59c418a60dfc5ec048fa/main_windows.go#L76-L87 instead

website/content/en/docs/config/vmtype/hcs.md
```diff
@@ -0,0 +1,65 @@
+---
+title: HCS
+weight: 5
+---
+
+> **Warning**
+> "hcs" mode is experimental
+
+| ⚡ Requirement | Lima >= TBD + (Windows >= Windows 10 OR Windows Server 2019) |
```
AkihiroSuda: This will be in v2.3

website/content/en/docs/config/vmtype/hcs.md
```diff
@@ -0,0 +1,65 @@
...
+{{% tab header="YAML" %}}
+```yaml
+# Example to run Ubuntu using vmType: hcs
+vmType: hcs
+images:
+- location: "https://cloud-images.ubuntu.com/releases/resolute/release-20260720/ubuntu-26.04-server-cloudimg-amd64.img"
+  arch: "x86_64"
+  digest: "sha256:117816726abbdefc5ef3e38902e81a76f1c76c3610e709999d0885f9d5d9b477"
+plain: true
+dns: - 8.8.8.8
```
AkihiroSuda: YAML format seems broken for DNS

AkihiroSuda:
Needs rebasing

cmd/limactl/main_wsl2.go
```diff
@@ -1,4 +1,4 @@
-//go:build !external_wsl2
+//go:build windows && !external_wsl2
```
AkihiroSuda: This line does not need to be changed

cmd/limactl/main_wsl2_windows.go
AkihiroSuda, line 1: The file name should be like `main_wsl2_windows.go`

go.mod
AkihiroSuda: This `replace` should not longer be needed

website/content/en/docs/config/vmtype/hcs.md
```diff
@@ -0,0 +1,66 @@
+---
+title: HCS
+weight: 5
+---
+
+> **Warning**
+> "hcs" mode is experimental
+
+| ⚡ Requirement | Lima >= 2.3 + (Windows >= Windows 10 OR Windows Server 2019) |
```
AkihiroSuda: No need to support Windows 10 and its server derivative

cmd/limactl/main_wsl2_windows.go
mie313: Thank you, I changed the file name. I also update `main_hcs.go` -> `main_hcs_windows.go`.

cmd/limactl/main_wsl2.go
mie313: I revised the change and I update the code in `main_hcs_windows.go` too.

website/content/en/docs/config/vmtype/hcs.md
mie313: I fixed the document. ([link](https://github.com/lima-vm/lima/pull/5449/changes#diff-63f614e0e56c1b2a618c6aae36d7d1f73793b572258ba8ba913ee1051e42cca6R30-R31)).

go.mod
mie313: Thank you, based on the file you shared, I have removed the dependency on `hcsshim/internal/computecore`.
I am still struggling to remove `hcsshim/hcn` dependency.

website/content/en/docs/config/vmtype/hcs.md
mie313: I updated the requirement to  Windows >= 11, since that's the only OS I have tested on.
Please let me know if Windows Server 2022 should be included here.

go.mod
```diff
@@ -114,7 +114,7 @@ require (
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06 // indirect
-	github.com/segmentio/asm v1.1.3 // indirect
+	github.com/segmentio/asm v1.2.1 // indirect
```
AkihiroSuda, line 123: - https://github.com/segmentio/asm/pull/86

You need to update the license list to allow MIT-0: https://github.com/lima-vm/lima/blob/fa0f5f6bb9f49ad316ff76847d5901dfd58f0c98/hack/allowed-licenses.txt#L1


CNCF has already approved MIT-0:
https://github.com/cncf/foundation/blob/87e70a07d5a0cf06cd6f8208b8d52c16f035724a/policies-guidance/allowed-third-party-license-policy.md?plain=1#L53

go.mod
```diff
@@ -130,3 +130,14 @@ require (
 )
 
 require github.com/clipperhouse/uax29/v2 v2.2.0 // indirect
+
+require github.com/Microsoft/hcsshim v0.14.1
```
AkihiroSuda: nit: merge to the existing `require ()` block

go.mod
```diff
@@ -130,3 +130,14 @@ require (
...
+
+require (
```
AkihiroSuda, line 139: nit: merge to the existing indirect `require ()` block

pkg/driver/hcs/errors_windows.go
```diff
@@ -0,0 +1,10 @@
+//go:build windows
```
AkihiroSuda: Unneeded line

go.mod
AkihiroSuda: go-licenses still doesn't seem to support MIT-0
- https://github.com/google/licenseclassifier/issues/75

So you can just add `--ignore`
https://github.com/lima-vm/lima/blob/fa0f5f6bb9f49ad316ff76847d5901dfd58f0c98/Makefile#L645-L650

Makefile
```diff
@@ -644,10 +644,10 @@ shfmt:
 
 .PHONY: go-licenses
 go-licenses:
-	# the allow list corresponds to https://github.com/cncf/foundation/blob/e5db022a0009f4db52b89d9875640cf3137153fe/allowed-third-party-license-policy.md
+	# the allow list corresponds to https://github.com/cncf/foundation/blob/87e70a07d5a0cf06cd6f8208b8d52c16f035724a/policies-guidance/allowed-third-party-license-policy.md
 	# hashicorp/hcl/v2 is MPL-2.0; covered by the CNCF license exception for hashicorp/hcl
 	# see also https://github.com/cncf/foundation/issues/1242
```
AkihiroSuda, line 649: Add a comment line to explain the reason of ignoring github.com/segmentio/asm

AkihiroSuda (dismissed):
Thanks

cmd/limactl/main_hcs_windows.go
unsuman, line 1: Should we not build it as an external driver? Since it's experimental

[EDIT]: I also see you have built the HCS driver as external below, but lima prioritises the one which is built as internal

pkg/driver/hcs/hcs_driver_windows.go
```diff
@@ -0,0 +1,417 @@
...
+		inst.Status = limatype.StatusBroken
+		inst.Errors = append(inst.Errors, fmt.Errorf("failed to inspect HCS sompute system: %w", err))
+		return inst.Status
+	}
+	if state == "Running" {
+		inst.Status = limatype.StatusRunning
+		for range 50 {
+			e, err := hcn.GetEndpointByName(endpointName(inst.Name))
+			if err != nil {
+				logrus.Errorf("error: %s", err.Error())
```
unsuman: Needs a `break` or `continue`

pkg/driver/hcs/hcs_driver_windows.go
```diff
@@ -0,0 +1,417 @@
...
+			inst.Status = limatype.StatusUnknown
+			return inst.Status
+		}
+		inst.Status = limatype.StatusBroken
+		inst.Errors = append(inst.Errors, fmt.Errorf("failed to inspect HCS sompute system: %w", err))
+		return inst.Status
+	}
+	if state == "Running" {
+		inst.Status = limatype.StatusRunning
+		for range 50 {
```
unsuman: But why are we iterating for `100s` here? If the state is running then we should just try to retrieve the IP once and error out if we don't get one

pkg/limayaml/defaults.go
```diff
@@ -616,9 +624,7 @@ func FillDefault(ctx context.Context, y, d, o *limatype.LimaYAML, filePath strin
 	if o.HostResolver.Enabled != nil {
 		y.HostResolver.Enabled = o.HostResolver.Enabled
 	}
-	if y.HostResolver.Enabled == nil {
```
unsuman, line 619: Other drivers code are not updated to fill this value. I think we should keep the global default and have the HCS driver override

cmd/limactl/main_hcs_windows.go
mie313: I thought that I have not built the HCS driver as external, because `cmd/limactl/main_hcs_windows.go` is only built when `!external_hcs` is true and I have not created `cmd/lima-driver-hcs/main.go`. Am I wrong?

pkg/limayaml/defaults.go
mie313: `HostResolver.Enabled` value was only referred to in `pkg/cidata/cidata.go` and `pkg/hostagent/hostagent.go`.
They are replaced by the helper function [`HostResolverEnabled()`](https://github.com/lima-vm/lima/pull/5449/changes/BASE..eb6e5a46f8f6abd68974d05bc7cf9c8553146164#diff-1e17bbd436fd831ebcaac290c75166c21aefc990f1be6125ddcea9832a1dd0b0R85-R88) ([here](https://github.com/lima-vm/lima/pull/5449/changes/BASE..eb6e5a46f8f6abd68974d05bc7cf9c8553146164#diff-6a911aa7574099e18369e9477b214a031077350dedc30e60b933ad0f61e022c2R297))
So, I think the behavior in other drivers remains the same as before.

mie313: After double-checking, I noticed that HostResolver.Enabled is also referred to in `pkg/limayaml/validate.go` .
So, I have update the code as follows:
[EDIT]
```
	if y.VMType != nil && *y.VMType == limatype.HCS {
		if y.HostResolver.Enabled != nil && *y.HostResolver.Enabled && len(y.DNS) > 0 {
			errs = errors.Join(errs, errors.New("field `dns` must be empty when field `HostResolver.Enabled` is true"))
		}
	} else {
		if HostResolverEnabled(y) && len(y.DNS) > 0 {
			errs = errors.Join(errs, errors.New("field `dns` must be empty when field `HostResolver.Enabled` is true"))
		}
	}
```
Thank you for pointing that out.

cmd/limactl/main_hcs_windows.go
unsuman: My bad I didn't see that! But I think for now building HCS as internal is OK, because external driver transport is not well tested on Windows and can have some bug(s).

pkg/driver/hcs/hcs_driver_windows.go
mie313: Thank you, you are completely right.
I have removed the for loop as you suggested.

unsuman (approved):
Thanks

AkihiroSuda (approved):
Thanks

PR #5454, opened 2026-08-28
reject guest agent events that report an invalid IP

SABITHSAHEB:
1. GuestAddr on the port-forward event comes from the guest-reported IPPort.Ip (guest.HostString()), an unvalidated protobuf string an untrusted guest fully controls.
2. It flows into the event stream that limactl start and limactl watch print with a bare %s, so a guest can inject ANSI/OSC escapes to spoof the operator terminal; a single AddedLocalPorts entry whose Ip is not a parseable IP takes the not-forwarding branch and emits the raw bytes.

Per review, the guest-reported addresses are validated when the event is received: processGuestAgentEvents fails the event with an error if any AddedLocalPorts or RemovedLocalPorts entry does not parse with net.ParseIP, before either forwarder derives GuestAddr or log lines from it. The Events callback in the guest agent client now returns an error so that failure propagates to the reconnect loop. The regression test streams an event with a malformed IP through an in-process GuestService and asserts the error.

### DIFF
--- a/pkg/hostagent/hostagent.go
+++ b/pkg/hostagent/hostagent.go
@@ -953,10 +953,17 @@ func (a *HostAgent) processGuestAgentEvents(ctx context.Context, client *guestag
 
 	logrus.Debugf("guest agent info: %+v", info)
 
-	onEvent := func(ev *guestagentapi.Event) {
+	onEvent := func(ev *guestagentapi.Event) error {
 		logrus.Debugf("guest agent event: %+v", ev)
 		for _, f := range ev.Errors {
 			logrus.Warnf("received error from the guest: %#q", f)
+		}
+		// IPPort.Ip is under the guest's control and ends up in PortForwardEvent.GuestAddr,
+		// which limactl start/watch print to the terminal, so reject anything that is not an IP.
+		for _, f := range slices.Concat(ev.AddedLocalPorts, ev.RemovedLocalPorts) {
+			if net.ParseIP(f.Ip) == nil {
+				return fmt.Errorf("guest agent reported invalid IP %#q", f.Ip)
+			}
 		}
 		// History of the default value of useSSHFwd:
 		// - v0.1.0:        true  (effectively)
@@ -978,6 +985,7 @@ func (a *HostAgent) processGuestAgentEvents(ctx context.Context, client *guestag
 			dialContext := portfwd.DialContextToGRPCTunnel(client)
 			a.grpcPortForwarder.OnEvent(ctx, dialContext, ev)
 		}
+		return nil
 	}
 
 	if err := client.Events(ctx, onEvent); err != nil {
--- a/pkg/guestagent/api/client/client.go
+++ b/pkg/guestagent/api/client/client.go
@@ -63,7 +63,7 @@ func (c *GuestAgentClient) Info(ctx context.Context) (*api.Info, error) {
 	return c.cli.GetInfo(ctx, &emptypb.Empty{})
 }
 
-func (c *GuestAgentClient) Events(ctx context.Context, eventCb func(response *api.Event)) error {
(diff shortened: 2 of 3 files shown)


pkg/hostagent/hostagent.go
```diff
@@ -371,6 +371,12 @@ func (a *HostAgent) emitCloudInitProgressEvent(ctx context.Context, progress *ev
 }
 
 func (a *HostAgent) emitPortForwardEvent(ctx context.Context, pfEvent *events.PortForwardEvent) {
+	// GuestAddr embeds the guest-reported IP (api.IPPort.Ip), which an untrusted
+	// guest controls and which limactl start/watch print to the operator terminal
+	// with a bare %s. Strip control characters so the guest cannot inject ANSI/OSC
+	// escape sequences, same as emitCloudInitProgressEvent.
+	pfEvent.GuestAddr = strutil.RemoveControlChars(pfEvent.GuestAddr)
```
AkihiroSuda: I don't think this change is necessary.
`pfEvent.GuestAddr` should be validated when it is initialized

SABITHSAHEB: Makes sense. Moved the check to where the guest event is received: processGuestAgentEvents now drops entries whose IPPort.Ip doesn't parse as an IP, before either forwarder builds GuestAddr or logs it, and the strip in emitPortForwardEvent is reverted. Test updated to cover the validation.

pkg/hostagent/hostagent.go
```diff
@@ -931,6 +931,22 @@ func (a *HostAgent) getOrCreateClient(ctx context.Context) (*guestagentclient.Gu
...
 
+// validateLocalPorts drops guest-reported entries whose IP does not parse.
+// IPPort.Ip is an arbitrary string under the guest's control; it ends up in
+// PortForwardEvent.GuestAddr, which limactl start/watch print to the operator
+// terminal, so anything that is not an IP address must not get that far.
+func validateLocalPorts(ipPorts []*guestagentapi.IPPort) []*guestagentapi.IPPort {
+	validated := make([]*guestagentapi.IPPort, 0, len(ipPorts))
+	for _, f := range ipPorts {
+		if net.ParseIP(f.Ip) == nil {
+			logrus.Warnf("ignoring guest-reported port with invalid IP %#q", f.Ip)
```
AkihiroSuda: Should this return an error?

pkg/hostagent/hostagent.go
```diff
@@ -931,6 +931,22 @@ func (a *HostAgent) getOrCreateClient(ctx context.Context) (*guestagentclient.Gu
 	return a.client, err
 }
 
+// validateLocalPorts drops guest-reported entries whose IP does not parse.
+// IPPort.Ip is an arbitrary string under the guest's control; it ends up in
+// PortForwardEvent.GuestAddr, which limactl start/watch print to the operator
+// terminal, so anything that is not an IP address must not get that far.
+func validateLocalPorts(ipPorts []*guestagentapi.IPPort) []*guestagentapi.IPPort {
```
AkihiroSuda: This function is not necessary.
Alternatively you can just call `net.ParseIP` directly.

SABITHSAHEB: Dropped it. The check is now inline in onEvent, just net.ParseIP over AddedLocalPorts and RemovedLocalPorts.

pkg/hostagent/hostagent.go
SABITHSAHEB: Yes, done. A malformed IP now fails the event with an error instead of being skipped with a warning. That needed the Events callback in the guest agent client to return an error, which Events propagates, so processGuestAgentEvents returns it and the reconnect loop in watchGuestAgentEvents picks it up. The test now streams an event with a bad IP through a bufconn GuestService and asserts processGuestAgentEvents returns the error.

AkihiroSuda (approved):
Thanks

PR #5456, opened 2026-08-29
Use `host` CPU type starting from QEMU 11.1.1 on Windows

### LINKED ISSUE #5455: Support QEMU `host` CPU type on Windows
### Description

This should have been supported since 11.1.0, but that version contained a bug for virtualized environments (specifically on Intel platforms), so, it is more reasonable to make a cutoff at 11.1.1 (bug was classified as regression and the fix has been backported).

arixmkii:
## What This PR Changes

Passes detected QEMU version to CPU selection routines, which allows `host` level CPU for versions greater than or equal to `11.1.1`.

Closes #5455 

## How I Tested This

Tested locally inspecting command line of QEMU processes running against QEMU 11.1.0 and 11.1.1. Only `amd64` architecture was tested.

## AI Usage

Not at all.

### DIFF
--- a/pkg/driver/qemu/qemu.go
+++ b/pkg/driver/qemu/qemu.go
@@ -412,7 +412,7 @@ func audioDevice() string {
 	return "oss"
 }
 
-func defaultCPUType() limatype.CPUType {
+func defaultCPUType(qemuVersion *semver.Version) limatype.CPUType {
 	// x86_64 + TCG + max was previously unstable until 2021.
 	// https://bugzilla.redhat.com/show_bug.cgi?id=1999700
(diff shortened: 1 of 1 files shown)


arixmkii:
The CI would run against older version, because `11.1.1` is not yet available in Winget.

arixmkii:
I also ran GH CI with a custom build of 11.1.1 to have at least basic verification of this version in CI in addition to my local tests.

unsuman (approved):
Thanks, LGTM!

AkihiroSuda (approved):
Thanks

